PR #190, opened 2021-05-18
Refactor fmt modules into a single shared embassy_fmt crate

rukai:
advantages:
*   reduces duplication
*   the `#[macro_use] extern crate embassy_fmt` ensure that we always use the defmt macros over the space expensive core macros.
    +   previously the best we could do was `use crate::fmt::{unreachable, *};` which will import `unreachable` and give a compile error if any formatting macros are used but not manually imported. But I was accidentally deleting these because of the unused warnings and I didnt realize they create a helpful compiler error instead of using a core format macro.
*   the current fmt module implementation has really strange repercussions for importing. the `pub use`d macros are imported into the fmt module while the new macro_rules macros have a global scope. This means the macros are accessible at different paths depending on if the defmt feature is enabled or not. There are two ways to resolve this issue:
     +    move the macros to a seperate crate (the method I chose here)
     +   enable the pub_macro_rules nightly feature https://github.com/rust-lang/rust/issues/78855

disadvantages:
*   introduces a new nightly feature requirement: namespaced-features

We could potentially provide embassy_fmt as a wrapper over defmt/log for end users but for now its just considered an implementation detail.
I havent changed stm32 to use it yet, I plan to do that in a follow up PR which will also add a .cargo/config and fix the warnings for stm32.

PR #227, opened 2021-06-06
fmt: make all macros `macro_rules` so scoping is consistent.

Dirbaio:
Yet another try at solving the `fmt` issues.

This keeps a `fmt.rs` copypasted per-crate, but makes it so all the macros are now `macro_rules!`. This means macro scoping works the same way no matter what features are enabled (`log`, `defmt` or neither). `use crate::fmt::xxx` imports are no longer needed, so this should help get rid of the `unused import` warnings.

Not copypasting `fmt.rs` seems impossible, due to defmt macros breaking if you reexport them. See https://github.com/embassy-rs/embassy/pull/190#issuecomment-855314011 for discussion.

cc @rukai


embassy-extras/src/fmt.rs
```diff
@@ -1,67 +1,168 @@
...
-            ()
-        };
-    }
+macro_rules! trace {
+    ($($x:tt)*) => {
+        {
+            #[cfg(feature = "log")]
+            log::trace!($($x)*);
+            #[cfg(feature = "defmt")]
+            defmt::trace!($($x)*);
```
thalesfragoso, line 112: Hmm, will this produce an unused warning when compiling without defmt nor log ? Maybe do something like smoltcp:
```
$( let _ = $x; )*
```
But not sure how that works with `tt`.

thalesfragoso (commented):
Great idea, I just left some questions.

embassy-extras/src/fmt.rs
```diff
@@ -1,67 +1,168 @@
...
 compile_error!("You may not enable both `defmt` and `log` features.");
 
-pub use fmt::*;
+macro_rules! assert {
+    ($($x:tt)*) => {
+        {
+            #[cfg(not(feature = "defmt"))]
+            core::assert!($($x)*);
+            #[cfg(feature = "defmt")]
+            defmt::assert!($($x)*);
```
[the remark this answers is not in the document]
Dirbaio: Good point! No idea if it can break without `::`, but it definitely can't hurt, so adding.

embassy-extras/src/fmt.rs
Dirbaio: Done

thalesfragoso (approved):
Thanks!

PR #230, opened 2021-06-07
Update after RCC regen and register fix

lulf:


Dirbaio (approved):
Code looks much nicer with the boolified regs :)

PR #237, opened 2021-06-09
Infer enable and reset registers based on RCC registry file

lulf:
Alternative approach that should handle split registers better by inferring register names from register file and doing a sanity check that it matches the peripheral clock name.

### DIFF
--- a/stm32-metapac/build.rs
+++ b/stm32-metapac/build.rs
@@ -156,12 +156,19 @@ fn main() {
     };
 
     // Load RCC register for chip
-    let chip_family = chip.family.to_ascii_lowercase();
-    let rcc_family = chip_family.strip_prefix("stm32").unwrap();
-    let rcc_reg_path = Path::new(&dir)
-        .join("registers")
-        .join(&format!("rcc_{}.yaml", &rcc_family[0..2]));
-    let rcc: ir::IR = serde_yaml::from_reader(File::open(rcc_reg_path).unwrap()).unwrap();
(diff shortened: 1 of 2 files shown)


stm32-metapac/build.rs
```diff
@@ -131,6 +155,14 @@ fn main() {
         peripherals: Vec::new(),
     };
 
+    // Load RCC register for chip
+    let chip_family = chip.family.to_ascii_lowercase();
+    let rcc_family = chip_family.strip_prefix("stm32").unwrap();
```
Dirbaio, line 160: It's nicer to get the RCC version from `chip.peripherals` instead of doing string operations on the family. This may not always be correct if there are exceptions.

Dirbaio: This explodes if the chip doesn't have RCC right? currently stm32-metapac allows using all the chips, it's only embassy-stm32 that only supports certain families. We should try keeping the metapac building with all chips

Dirbaio (changes requested):

stm32-metapac/build.rs
lulf: Yeah, thats a good point. I thought RCC was always there. I'll change it to look for the RCC peripheral, and Option'ize the rcc register find and clock lookup.

Dirbaio: yeah, all chips have it. I meant we don't have the regs yaml for all chips yet. For these the metapac should still build, even if incomplete.

Dirbaio (approved):
Awesome :)

PR #240, opened 2021-06-10
Enable timer clock in RCC on timer start

lulf:
* Moves the tim2-specific code into macro which always uses TIM2
* For peripherals without clock specified, attempt to locate enable and
  reset registers in the RCC block matching the peripheral name. This
  could be useful for peripherals where deducing the clock name might
  not be feasible, but it remains to be tested with more chip families
  to see if it is sufficiently accurate.

### DIFF
--- a/stm32-metapac/gen/src/lib.rs
+++ b/stm32-metapac/gen/src/lib.rs
@@ -268,10 +268,20 @@ pub fn gen(options: Options) {
                 }
 
                 if let Some(rcc) = &rcc {
-                    let mut generate_rcc_peripheral = |clock_prefix| {
+                    let clock_prefix: Option<&str> = if let Some(clock) = &p.clock {
+                        Some(clock)
+                    } else if name.starts_with("TIM") {
+                        // Not all peripherals like timers the clock hint due to insufficient information from
+                        // chip definition. If clock is not specified, the first matching register with the
+                        // expected field will be used.
+                        Some("")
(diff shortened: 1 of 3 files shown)


stm32-metapac/gen/src/lib.rs
```diff
@@ -267,12 +267,13 @@ pub fn gen(options: Options) {
                     _ => {}
                 }
 
-                if let Some(clock) = &p.clock {
-                    if let Some(rcc) = &rcc {
+                if let Some(rcc) = &rcc {
+                    let mut generate_rcc_peripheral = |clock_prefix| {
```
Dirbaio, line 271: Writing this with a closure is super weird :) You could do it with a var, like this:

```rust
let clock_prefix: &str = p.clock.as_ref().unwrap_or("");
// code from the closure here
```

Dirbaio (changes requested):
Awesome removing the l0-specific hack :)

Looks great, commented a style nit.

stm32-metapac/gen/src/lib.rs
lulf: You're right, it is super weird. I can't remember why I did that, but it must have been because I thought of calling it directly from each peripheral type "spi" etc. and then I decided to do the "TIM hack" instead. I'll change it.

Dirbaio (approved):
:rocket:

PR #241, opened 2021-06-11
Provide a way for a peripheral to query its clock frequency

lulf:
Currently this looks up the frequency in the global singleton that must be initialized by the per-chip RCC implementation. At present, this is only done for the L0 family of chips.

I wonder if this singleton type should be per-chip instead, but standardizing on the clock naming so that the clock names generated by parse.py will be matching the per-chip struct Clock fields.

NOTE: This includes commits from #240


Dirbaio (approved):
Approach looks good!

> I wonder if this singleton type should be per-chip instead, but standardizing on the clock naming

It could be either defined for all families in `rcc/mod.rs` with cfg's to only define the clocks present in the current chip, or defined per-family (per-rcc-version) in `rcc/l0/mod.rs`, `rcc/h7/mod.rs`, etc. Not sure which is best, first one will be cfg-spaghetti and the second one will be somewhat copypastey. Up to you :)

lulf:
@Dirbaio I went with the first. I have also added "minimal clock" impls for f4, l4 and wl55, only enabling HSE and HSI16, in order for peripherals to somewhat work (not tested, I don't have l4, f4 or wl55).

Dirbaio (approved):
Thank you!

PR #244, opened 2021-06-12
Make gen.py work without CSafeLoader

Tiwalun:
On my Mac, the CSafeLoader seems to be unavailable in the pyyaml package. 
With this change, CSafeLoader is still used if available, but otherwise it now falls
back to SafeLoader.﻿

### DIFF
--- a/embassy-stm32/gen.py
+++ b/embassy-stm32/gen.py
@@ -4,6 +4,12 @@ import re
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Makes sense! I was fearing this would happen due to the docs warning it's not always available. The C loader is way faster :)

PR #247, opened 2021-06-14
Eth v2

thalesfragoso:
This also includes some changes to embassy-net to play nice with stm32.


embassy-stm32/src/eth/v2/descriptors.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// Transmit Descriptor representation
+///
+/// * tdes0: transmit buffer address
+/// * tdes1:
+/// * tdes2: buffer lengths
+/// * tdes3: control and payload/frame length
+#[repr(C)]
+struct TDes {
+    tdes0: VolatileCell<u32>,
```
Dirbaio, line 45: why is VolatileCell needed? this sits in normal RAM right? so normal reads/writes + fences would be enough I think?

Dirbaio (approved):
Looks great! I haven't reviewed in depth how the DMA descriptor stuff works, I don't have enough brain power :(

It'd be nice to have an example, but feel free to merge this now and we can add it later.

Dirbaio (approved):
Awesome, thank you so much! :)

PR #250, opened 2021-06-16
Add support for generating PAC for dual cores

lulf:
* Chips that have multiple cores will be exposed as chipname_corename,  i.e. stm32wl55jc_cm4
* Chips that have single cores will use the chip family as feature name  and pick the first and only core from the list
* Add support for stm32wl55 chip family


Dirbaio (approved):
This is awesome :rocket:

PR #255, opened 2021-06-24
stm32: Allow for open drain configuration for output pin

thalesfragoso:


embassy-stm32/src/gpio.rs
```diff
@@ -99,6 +129,10 @@ impl<'d, T: Pin> Output<'d, T> {
             let n = pin.pin() as usize;
             r.pupdr().modify(|w| w.set_pupdr(n, vals::Pupdr::FLOATING));
             r.moder().modify(|w| w.set_moder(n, vals::Moder::OUTPUT));
+            if open_drain {
```
Dirbaio, line 132: if it's not openrain you still need to do the modify to set it as not opendrain, otherwise it stays if the pin has been previously used as opendrain

embassy-stm32/src/gpio.rs
```diff
@@ -115,6 +149,8 @@ impl<'d, T: Pin> Drop for Output<'d, T> {
             let n = self.pin.pin() as usize;
             r.pupdr().modify(|w| w.set_pupdr(n, vals::Pupdr::FLOATING));
             r.moder().modify(|w| w.set_moder(n, vals::Moder::INPUT));
+            r.otyper().modify(|w| w.set_ot(n, vals::Ot::PUSHPULL));
```
Dirbaio, line 152: if pin is set as floating input, aren't otyper/ospeedr ignored?

embassy-stm32/src/gpio.rs
```diff
@@ -148,6 +184,18 @@ impl<'d, T: Pin> StatefulOutputPin for Output<'d, T> {
     }
 }
 
+impl<'d, T: Pin> InputPin for Output<'d, T> {
```
Dirbaio, line 187: This is a bit dangerous, if the user is creating an Output they shouldn't be able to use it as input except in very special circumstances. I guess you want this for onewire?

Maybe it should be a new `InputOutput` driver? Maybe even `OpenDrainInputOutput` because it only makes sense for opendrain?

embassy-stm32/src/gpio.rs
```diff
@@ -148,6 +184,18 @@ impl<'d, T: Pin> StatefulOutputPin for Output<'d, T> {
...
+    type Error = Infallible;
+
+    fn is_high(&self) -> Result<bool, Self::Error> {
+        self.is_set_high()
```
Dirbaio, line 191: `is_set_high/is_set_low` return the output value we've set, not the input from outside the mcu, that's probably not what you want?

Dirbaio (changes requested):

embassy-stm32/src/gpio.rs
Dirbaio: I see you're returning it to PUSHPULL on drop... Other stuff like i2c might still leave it in OPENDRAIN, I think it's safer to always set it anyway.

embassy-stm32/src/gpio.rs
thalesfragoso: Yeah, just noticed it reads from `odr` not `idr`.

embassy-stm32/src/gpio.rs
thalesfragoso: Probably, I was just setting again to leave at a reset state for the "next driver".

embassy-stm32/src/gpio.rs
```diff
@@ -99,6 +131,7 @@ impl<'d, T: Pin> Output<'d, T> {
             let n = pin.pin() as usize;
             r.pupdr().modify(|w| w.set_pupdr(n, vals::Pupdr::FLOATING));
             r.moder().modify(|w| w.set_moder(n, vals::Moder::OUTPUT));
+            pin.set_speed(speed);
```
Dirbaio, line 134: I think this should set OTYPER, otherwise it'll stay as opendrain if it was used before.

Dirbaio (approved):
Looks great! I believe I found one last bug :)

embassy-stm32/src/gpio.rs
thalesfragoso: Agreed, and done!

Dirbaio:
Thanks! :rocket:

PR #258, opened 2021-06-25
rp/gpio: add optional pins

Dirbaio:


embassy-rp/src/spi.rs
```diff
@@ -52,7 +54,7 @@ impl<'d, T: Instance> Spi<'d, T> {
 
             // Find largest post-divide which makes output <= baudrate. Post-divide is
             // an integer in the range 1 to 256 inclusive.
-            let postdiv = (1u32..=256)
+            let postdiv = (2u32..=256)
```
thalesfragoso, line 57: Is this comment outdated now ?

embassy-rp/src/spi.rs
```diff
@@ -145,10 +156,19 @@ macro_rules! impl_instance {
 impl_instance!(SPI0, Spi0);
 impl_instance!(SPI1, Spi1);
 
-pub trait ClkPin<T: Instance>: sealed::ClkPin<T> + Pin {}
-pub trait CsPin<T: Instance>: sealed::CsPin<T> + Pin {}
-pub trait MosiPin<T: Instance>: sealed::MosiPin<T> + Pin {}
-pub trait MisoPin<T: Instance>: sealed::MisoPin<T> + Pin {}
+pub trait ClkPin<T: Instance>: sealed::ClkPin<T> + OptionalPin {}
```
thalesfragoso, line 159: Should the clock really be optional ?

thalesfragoso (commented):
Just passing by with some comments/questions, thanks for the nice work!

embassy-rp/src/spi.rs
Dirbaio: Maybe. There are crash issues when setting a too high baudrate, I still have to look at them. This diff should definitely not be here anyway, will remove.

embassy-rp/src/spi.rs
Dirbaio: Why not? it's allowed by the hardware. Could be used to output custom waveforms for RGB leds or stuff... except this SPI IP is stupid, it can't transmit bytes back-to-back so timing would be off...

embassy-rp/src/spi.rs
Dirbaio: Left it TODO for now, the calculation with the loop is stupid and needs refactoring anyway.

PR #259, opened 2021-06-28
Add BlockingTimer and features to choose tick rate

thalesfragoso:


embassy/src/time/mod.rs
```diff
@@ -28,3 +42,49 @@ pub unsafe fn set_clock(clock: &'static dyn Clock) {
...
+
+impl embedded_hal::blocking::delay::DelayMs<u8> for BlockingTimer {
+    fn delay_ms(&mut self, ms: u8) {
+        block_for(Duration::from_millis(u64::from(ms)))
+    }
+}
+
+impl embedded_hal::blocking::delay::DelayMs<u16> for BlockingTimer {
+    fn delay_ms(&mut self, ms: u16) {
+        block_for(Duration::from_millis(u64::from(ms)))
```
Dirbaio, line 56: for int casts it's nicer to do `ms as u64`

embassy/src/time/mod.rs
```diff
@@ -28,3 +42,49 @@ pub unsafe fn set_clock(clock: &'static dyn Clock) {
...
+    }
+}
+
+impl embedded_hal::blocking::delay::DelayMs<u32> for BlockingTimer {
+    fn delay_ms(&mut self, ms: u32) {
+        block_for(Duration::from_millis(u64::from(ms)))
+    }
+}
+
+#[cfg(feature = "tick-mhz-1")]
```
Dirbaio, line 66: why gate these on 1mhz clock only? If we interpret the traits as "sleeps for at least Xus" we can implement them for any tick rate, just taking care to round up.

embassy/Cargo.toml
```diff
@@ -5,7 +5,11 @@ authors = ["Dario Nieuwenhuis <dirbaio@dirbaio.net>"]
 edition = "2018"
 
 [features]
+default = ["tick-hz-32768"]
 std = ["futures/std", "embassy-traits/std"]
+tick-hz-32768 = []
+tick-hz-1000 = []
+tick-mhz-1 = []
```
Dirbaio, line 12: not sure on the naming. Maybe `tick-32768hz`, `tick-1mhz` is nicer?

Dirbaio (approved):
Looks nice! Just a few minor nits.

embassy/src/time/mod.rs
thalesfragoso: Seems a bit of a footgun though.

thalesfragoso: From the e-h docs:
```
Pauses execution for us microseconds
```

Dirbaio: Yeah... usually these delays are implicitly "or longer", they can do that for other reasons anyway, such as CPU too busy with interrupts.

For drivers that want a DelayUs for "make a RST pulse that's 100us" it'd work great, but for fancier stuff like OneWire it could be a footgun yeah.

thalesfragoso: I think that was the reason they separated the traits into `Ms` and `Us`, because some drivers just won't be able to deal with microseconds. Like ours `BlockingTimer` with 1kHz tick rate

Dirbaio: If we gate this, then we can't use drivers that require DelayUs even if they don't care about precision.

Example: https://github.com/almindor/st7789/blob/master/src/lib.rs#L101

thalesfragoso: Ok, makes sense, nothing that some docs can't solve, working on it.

Dirbaio (approved):
:rocket:

PR #270, opened 2021-07-02
Let's adjust i2c the correct way, removing the correct APBesque frequ…

bobmcwhirter:
…ency, not the i2c periph speed.


Dirbaio (approved):
Hahaha, nice bug :) 

It would be nice to change i2c to use a `non_exhasutive` Config struct, like the other peripherals, for consistency.

bobmcwhirter:
Issue created for that: https://github.com/embassy-rs/embassy/issues/271

PR #273, opened 2021-07-03
fix stm32 warnings

rukai:
I suspect deleting `fn calibrate` isn't the right option, did we maybe forget to call it somewhere? or maybe this code is just super WIP?


Dirbaio (approved):
> I suspect deleting fn calibrate isn't the right option, did we maybe forget to call it somewhere? or maybe this code is just super WIP?

I'm not sure, I think it's supposed to be public? I've added it back with a TODO for now. @bobmcwhirter knows what's going on.

PR #281, opened 2021-07-08
i2c-v2: Support transfers with more than 255 bytes

thalesfragoso:
~~Still have to find something to test these bigger transfers.~~


snakehand:
Just a questions. Wouldn't such large transfers be better served with a DMA FIFO implementation ?

Dirbaio:
DMA is work-in-progress (#282), I think the goal is to move everything over to use DMA when it's done.

PR #282, opened 2021-07-08
BDMA + DMAMUX + H7 with major help from @thalesfragoso

bobmcwhirter:


embassy-stm32/src/bdma/v2.rs
```diff
@@ -0,0 +1 @@
+pub(crate) unsafe fn init() {}
```
Dirbaio, line 1: I think BDMAv1 and BDMAv2 should use the same code, as they are exactly identical except CSELR. The CSELR bits can do `#[cfg(bdma_v2)]`

Same with DMAv1/DMAv2

embassy-stm32/src/bdma/v1.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+unsafe fn on_irq() {
+    pac::peripherals! {
+        (bdma, $dma:ident) => {
+                let isr = pac::$dma.isr().read();
+                pac::$dma.ifcr().write_value(isr);
+                let dman = <crate::peripherals::$dma as sealed::Dma>::num() as usize;
+
+                for chn in 0..7 {
```
thalesfragoso, line 151: It should be `0..8`, right ? it's my mistake

embassy-stm32/src/bdma/v2.rs
bobmcwhirter: Agreed

embassy-stm32/src/bdma/v1.rs
bobmcwhirter: Will make that change.

bobmcwhirter:
```
                for chn in 0..crate::pac::dma_channels_count!($dma) {
```

embassy-stm32/src/bdma/mod.rs
```diff
@@ -1,69 +1,440 @@
...
+                let state_num = self.state_num();
+                let regs = self.regs();
+                unsafe { transfer_p2m(regs, state_num, src, buf) }
+            }
+        }
+
+        #[cfg(dmamux)]
+        impl<T> ReadDma<T> for crate::peripherals::$channel_peri
+        where
+            Self: crate::dmamux::sealed::PeripheralChannel<T, crate::dmamux::M2P>,
```
Dirbaio, line 319: should be P2M?

embassy-stm32/src/bdma/mod.rs
```diff
@@ -1,69 +1,440 @@
...
-        impl_dma_channel!(DMA1_CH0, 0, 0);
-        impl_dma_channel!(DMA1_CH1, 0, 1);
-        impl_dma_channel!(DMA1_CH2, 0, 2);
-        impl_dma_channel!(DMA1_CH3, 0, 3);
-        impl_dma_channel!(DMA1_CH4, 0, 4);
-        impl_dma_channel!(DMA1_CH5, 0, 5);
-        impl_dma_channel!(DMA1_CH6, 0, 6);
-        impl_dma_channel!(DMA1_CH7, 0, 7);
+pac::peripherals! {
+    (bdma, DMA1) => {
```
Dirbaio, line 359: instead of copying the inner code 3 times maybe you could define a `dma_num!` macro locally that maps `DMA1 -> 0, DMA2 -> 1...`?

embassy-stm32/src/bdma/mod.rs
```diff
@@ -1,69 +1,440 @@
...
+        impl_dma!(DMA1, 0);
+        pac::bdma_channels! {
```
Dirbaio, line 361: We should avoid nesting macrotable calls because it causes `O(N*M)` cost when compiling. Granted, it's not a big difference in this case, but it can grow big in other cases.

In this case it seems to me it should be possible to do one `peripherals!` call to impl all DMAs and then a single `bdma_channels!` call outside to impl all channels for all DMAs.

embassy-stm32/src/bdma/mod.rs
```diff
@@ -1,69 +1,440 @@
...
+        impl usart::TxDma<crate::peripherals::$peri> for crate::peripherals::$channel_peri { }
+        impl usart::sealed::TxDma<crate::peripherals::$peri> for crate::peripherals::$channel_peri { }
+    };
+}
+
+#[cfg(dmamux)]
+pac::peripherals! {
+    (usart, $peri:ident) => {
+        pac::bdma_channels! {
+            ($channel_peri:ident, $dma_peri:ident, $channel_num:expr) => {
```
Dirbaio, line 425: This is doing an impl for each (usart, channel) pair... This will be a lot of impls when all peripherals are DMAified, it might cause slow builds (not sure how tolerant is rustc to impl spam).

Maybe these can be `impl<T: Channel> TxDma<$peri> for T`, so there's just one per usart?

embassy-stm32/src/usart/mod.rs
```diff
@@ -95,10 +95,10 @@ pub(crate) mod sealed {
         fn af_num(&self) -> u8;
     }
 
-    #[cfg(any(dma, dmamux))]
+    #[cfg(any(bdma, dma, dmamux))]
```
Dirbaio, line 98: isn't this cfg "always on" now that we have all?

Dirbaio (changes requested):
Commented a few minor things.

I still have to fully read through this to understand everyting. My first impression is there's a lot of traits, hopefully there's some way to simplify.

embassy-stm32/src/usart/mod.rs
bobmcwhirter: Maybe? But I don't have enough confidence that there isn't some weird chip lacking any DMA... No harm, no foul leaving it for now, yeah?

PR #286, opened 2021-07-12
Make embassy::time optional

Dirbaio:


embassy-stm32/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 resolver = "2"
 
 [dependencies]
-embassy = { version = "0.1.0", path = "../embassy" }
+embassy = { version = "0.1.0", path = "../embassy", features = ["time-tick-32768hz"] }
```
thalesfragoso, line 9: Do we want to specify the ticks here ? Binary crates won't be able to choose the right tick for their job. However, `embassy-stm32` needs a tick, which means that we will need to provide tick feature in `embassy-stm32` too :/

thalesfragoso (commented):
Thanks, I think this is a good idea overall, but not sure what is the best way to also give flexibility for the end user. I left a comment.

embassy-stm32/Cargo.toml
Dirbaio: Yeah... When we do this, embassy-stm32 will need `time-tick-X` features which enable the corresponding ones in `embassy`...

Dirbaio: The other way around would be somehow embassy configuring the clock with the enabled tick rate, but some hardware can only work with one tick rate. For example nrf's can only do 32768hz. So I think it's better if the tick rate is enabled by the hal

thalesfragoso: And a default one...
Do you want to do this on this PR or in a later one ? If later, maybe we want to bump it to 1MHz ?

Dirbaio: The current stm32 clock can only do 32768hz right? We can add flexibility later. No opinion on the default from me, except as discussed on matrix 1mhz+16bit overflows too fast, we'd probably want 1mhz+32bit.

thalesfragoso: Thinking again, 32k seems a saner default.

thalesfragoso (approved):
Thanks!

PR #290, opened 2021-07-14
First shot at async dma usart for stm32.

bobmcwhirter:


embassy-stm32/src/usart/v3.rs
```diff
@@ -99,7 +115,11 @@ impl<'d, T: Instance> Uart<'d, T> {
     }
 }
 
-impl<'d, T: Instance> embedded_hal::blocking::serial::Write<u8> for Uart<'d, T> {
+impl<'d, T: Instance, TxDma, RxDma> embedded_hal::blocking::serial::Write<u8>
+    for Uart<'d, T, TxDma, RxDma>
```
Dirbaio, line 119: You can directly specify the NoDma type to restrict the impl. This way you don't need the `NoDmaMarker` trait.

`impl<'d, T: Instance,  RxDma> embedded_hal::blocking::serial::Write<u8> for Uart<'d, T, NoDma, RxDma>`

embassy-stm32/src/usart/v3.rs
```diff
@@ -54,11 +64,17 @@ impl<'d, T: Instance> Uart<'d, T> {
...
             inner,
             phantom: PhantomData,
+            tx_dma,
+            rx_dma,
         }
     }
 
     #[cfg(any(dma, dmamux))]
-    pub async fn write_dma(&mut self, ch: &mut impl TxDma<T>, buffer: &[u8]) -> Result<(), Error> {
+    pub async fn write_dma(&mut self, buffer: &[u8]) -> Result<(), Error>
```
Dirbaio, line 73: this should probably no longer be public (?)

Dirbaio (changes requested):
:rocket:

Dirbaio (approved):
This is awesome :D

PR #300, opened 2021-07-17
stm32: Clear possible set flags after disabling DMA

thalesfragoso:


embassy-stm32/src/dma/dma.rs
```diff
@@ -67,6 +67,13 @@ pub(crate) async unsafe fn do_transfer(
             w.set_en(false);
         });
         while ch.cr().read().en() {}
+
+        // Disabling the DMA mid transfer might cause some flags to be set, clear them all for the
+        // next transfer
+        for isrn in 0..2 {
+            let isr = dma.isr(isrn).read();
```
Dirbaio, line 74: This clears the flags for all channels, not just the current one.

Dirbaio (changes requested):

embassy-stm32/src/dma/dma.rs
thalesfragoso: Oh yeah, sorry, copy-paste from the interrupt...
Will fix it, although if other channel has the flags set we would be in the interrupt (which clears it)

Dirbaio (approved):

PR #305, opened 2021-07-20
Spi dma take3

bobmcwhirter:


embassy-stm32/src/spi/v1.rs
```diff
@@ -128,9 +141,150 @@ impl<'d, T: Instance> Spi<'d, T> {
...
+    where
+        Tx: TxDmaChannel<T>,
+    {
+        unsafe {
+            T::regs().cr1().modify(|w| {
+                w.set_spe(false);
+            });
+            T::regs().cr2().modify(|reg| {
+                reg.set_rxdmaen(true);
+            });
```
thalesfragoso, line 156: is this needed here ?

embassy-stm32/src/spi/v1.rs
```diff
@@ -128,9 +141,150 @@ impl<'d, T: Instance> Spi<'d, T> {
...
+            T::regs().cr1().modify(|w| {
+                w.set_spe(false);
+            });
+        }
+
+        Ok(())
+    }
+
+    #[allow(unused)]
+    async fn read_write_dma_u8(&mut self, read: &mut [u8], write: &[u8]) -> Result<(), Error>
```
thalesfragoso, line 231: Should we check for the buffers being the same size ? What happens if `read.len() > write.len()` ?

embassy-stm32/src/spi/v3.rs
```diff
@@ -161,9 +175,167 @@ impl<'d, T: Instance> Spi<'d, T> {
...
+            T::regs().cfg1().modify(|reg| {
+                reg.set_rxdmaen(true);
+            });
+        }
+
+        let rx_request = self.rxdma.request();
+        let rx_src = T::regs().rxdr().ptr() as *mut u8;
+        let rx_f = self
+            .rxdma
+            .read(rx_request, rx_src, &mut read[0..write.len()]);
```
thalesfragoso, line 295: This seems to make more sense if compared to the other two implementations.

thalesfragoso (commented):
Thanks again! I left some comments that I think apply to both v1 and v2.

embassy-stm32/src/spi/v3.rs
thalesfragoso: Maybe we should have some indications that says that if `read.len() > write.len()` then `read` won't be fully filled ? Or maybe just require the lengths to be equal ?

thalesfragoso: Actually, I think that might panic if `write` is bigger than `read`.

bobmcwhirter: Yah I think definitely an assert but I am not sure how I feel about the slicing really. 

I'll double check which DMA directions get enabled in write-only modes on all of them.

bobmcwhirter: If read is longer then it never fills and won't complete. Which I hit a few times myself.

embassy-stm32/src/spi/v1.rs
bobmcwhirter: Added both an assert that read >= write, slicing it down as needed, and documented.

bobmcwhirter:
Also, where do we stand on the string-of-commits vs squash-it-down continuum?

Dirbaio (approved):
:rocket:

PR #306, opened 2021-07-21
Remove feature(impl_trait_in_bindings), update nightly

Dirbaio:


Dirbaio:
Will open again when  rust-lang/rust#87258 gets fixed. No point in keeping this around, it bitrots fast because it touches all files.

Dirbaio:
Reopening for visibility. Fixes #318

PR #309, opened 2021-07-21
STM32 DMA/BDMA: Use interrupt flags instead of atomics

timokroeger:
I have only tested the USART DMA example.
Any thoughts?

Code size USART DMA:
```
before (atomics)
   8804      48    1464   10316    284c usart_dma
after (interrput flags)
   8784      48    1464   10296    2838 usart_dma
```


timokroeger:
Unfortunately I do not have any boards to test the BDMA changes, hope I did not miss anything. Overall changes are very similar to DMA.

Dirbaio (approved):
Tested BDMA on stm32l476vg, it's now working with a small fix.

Looks awesome! Much simpler and the code size savings are nice. :rocket:

PR #310, opened 2021-07-22
Add a `Dbgmcu` struct capable of enabling all relevant DBGMCU.cr fields.

bobmcwhirter:
Includes the addition of a `dbgmcu!(...)` macro table which currently takes
the form of

	(cr, $fn_name:ident)

where `$fn_name` is something like `set_dbgsleep_d1` etc.

The method is unsafe, since it's performing unsafe PAC operations.

Two examples modified to demonstrate its usage.

### DIFF
--- a/embassy-stm32/src/dbgmcu/mod.rs
+++ b/embassy-stm32/src/dbgmcu/mod.rs
@@ -1,0 +1,13 @@
+pub struct Dbgmcu {}
+
+impl Dbgmcu {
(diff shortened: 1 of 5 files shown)


timokroeger:
Nice! For some families we have `enable_debug_wfe()` in the rcc module. IMO we should either migrate the existing API to use your new code or remove `enable_debug_wfe()` in favor of this PR.

PR #312, opened 2021-07-22
Enable SYSCFG clock in `exti::init()`

timokroeger:


bobmcwhirter:
Hrm, looking at the chip YAMLs, at least for the ones I checked for WB55, they seem to have clocks and should end up being `RccPeripheral`s.

e.g. `data/chips/STM32WB55CC.yaml` in `stm32-data`:

```
    SYSCFG:
      address: 0x40010000
      kind: SYS:STM32WB55_sys_v1_0
      clock: APB2
      block: syscfg_wb55/SYSCFG
```

bobmcwhirter:
Oh, but missing it from the RCC reg? Lemme check the RM!

bobmcwhirter:
Nothing in the RM or in errata. Thanks, STM <sarcasm/>.

Dirbaio (approved):
thanks for the PR, looks great!

I've added removing `syscfgen` from all examples.

Seems like wb55 indeed doesn't have RCC, I guess it's always enabled? we can merge this and fix later if it turns out it needs more stuff to work.

PR #321, opened 2021-07-28
F4 PLL

thalesfragoso:
~~Still missing PWR configuration needed if pushing some chips over 168MHz.~~


embassy-stm32/src/pwr/mod.rs
```diff
@@ -1,4 +1,5 @@
 #[cfg_attr(any(pwr_h7, pwr_h7smps), path = "h7.rs")]
+#[cfg_attr(not(any(pwr_h7, pwr_h7smps)), path = "none.rs")]
```
Dirbaio, line 2: why not do `#[cfg(pwr)]` in lib.rs so that the `pwr` module is never imported?

embassy-stm32/src/rcc/f4/max.rs
```diff
@@ -0,0 +1,268 @@
+#[cfg(any(
+    feature = "stm32f401cb",
```
Dirbaio, line 2: it'd make sense to get `build.rs` to set a cfg like `stm32f401` by substringing the feature, so you can do `#[cfg(stm32f401)]`.

Dirbaio (changes requested):

embassy-stm32/src/pwr/mod.rs
thalesfragoso: `lib.rs` already has `#[cfg(pwr)]`, but it's enabled for the F4 (since it has a pwr block configured), but I use the reg block directly instead of creating an abstraction in this folder, because it's just two register writes...

thalesfragoso: Hmm, but it would probably be better to use `#[cfg_attr(pwr_f4)` instead of `not(any(`

Dirbaio (approved):
Awesome, thank you!

PR #329, opened 2021-07-29
Misc stm32 fixes

### LINKED ISSUE #307: Clean up STM32 examples
Remove old-PAC usage for RCC if RccPeripheral is implemented enough.

So we can stop copy/pasting stuff that isn't needed into new examples.

timokroeger:
#307

### DIFF
--- a/examples/stm32l4/src/bin/usart.rs
+++ b/examples/stm32l4/src/bin/usart.rs
@@ -8,25 +8,22 @@
 
 #[path = "../example_common.rs"]
 mod example_common;
-use cortex_m::prelude::_embedded_hal_blocking_serial_Write;
-use defmt::panic;
-use embassy::executor::Spawner;
+
+use embassy_stm32::dbgmcu::Dbgmcu;
 use embassy_stm32::dma::NoDma;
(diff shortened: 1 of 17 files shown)


examples/stm32l4/src/bin/usart.rs
```diff
@@ -8,26 +8,23 @@
...
 use embassy_stm32::dma::NoDma;
 use embassy_stm32::usart::{Config, Uart};
-use embassy_stm32::{pac, Peripherals};
+use embedded_hal::blocking::serial::Write;
 use example_common::*;
 
-#[embassy::main]
-async fn main(_spawner: Spawner, p: Peripherals) {
+#[cortex_m_rt::entry]
+fn main() -> ! {
```
Dirbaio, line 19: Why change this to *not* use `embassy::main`?

timokroeger: Some examples only call blocking APIs. Changed here to match the other examples of other families.

Dirbaio (approved):
Nice! Thank you :)

PR #331, opened 2021-07-30
metapac: Improve enable bit search

timokroeger:
The number has different meanings depending on family:
stm32f0: RCC_APB2ENR - APB peripheral clock enable register 2    CLOCK: APB1
stm32f4: RCC_APB2ENR - RCC APB2 peripheral clock enable register CLOCK: APB2

Ignore the clock number and search all registers for a matching enable bit.

### DIFF
--- a/stm32-metapac-gen/src/lib.rs
+++ b/stm32-metapac-gen/src/lib.rs
@@ -118,7 +118,7 @@ impl BlockInfo {
 
 fn find_reg_for_field<'c>(
     rcc: &'c ir::IR,
-    reg_prefix: &str,
+    reg_regex: &str,
     field_name: &str,
 ) -> Option<(&'c str, &'c str)> {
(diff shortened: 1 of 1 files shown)


stm32-metapac-gen/src/lib.rs
```diff
@@ -443,19 +437,20 @@ pub fn gen(options: Options) {
                     if let Some(clock_prefix) = clock_prefix {
```
Dirbaio, line 437: `clock_prefix` is now completely unused, it should be removed. It still has the effect of skipping the search if peripheral has no clock, but I think always doing the bit search should be fine.

Dirbaio (changes requested):
Looking great! just one nit

Dirbaio (approved):
Thank you! :+1:

PR #335, opened 2021-07-31
examples: Drop build-std = ["core"]

bgamari:
Previously the cargo configurations of all of the example projects had
`build-std = ["core"]`, which forces compilation of `core` as a
code-size optimisation. However, this is strictly unnecessary and will
currently break for users who do not use `rustup` directly (e.g. nix
users).


Dirbaio (approved):
Nice, thanks!

PR #336, opened 2021-07-31
examples: Consistently use unwrap! in favor of .unwrap()

bgamari:
Unfortunately errors from `embedded_graphics` and `core` doesn't provide
the necessary instances currently.

### DIFF
--- a/embassy-macros/src/lib.rs
+++ b/embassy-macros/src/lib.rs
@@ -364,7 +364,7 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
             #chip_setup
 
             executor.run(|spawner| {
-                spawner.spawn(__embassy_main(spawner, p)).unwrap();
+                spawner.must_spawn(__embassy_main(spawner, p));
             })
 
         }
--- a/embassy/src/executor/mod.rs
+++ b/embassy/src/executor/mod.rs
@@ -56,6 +56,14 @@ impl Spawner {
         }
     }
 
+    /// Used by the `embassy_macros::main!` macro to throw an error when spawn
+    /// fails. This is here to allow conditional use of `defmt::unwrap!`
+    /// without introducing a `defmt` feature in the `embassy_macros` package,
+    /// which would require use of `-Z namespaced-features`.
+    pub fn must_spawn<F>(&self, token: SpawnToken<F>) -> () {
+        unwrap!(self.spawn(token));
+    }
+
     /// Convert this Spawner to a SendSpawner. This allows you to send the
     /// spawner to other threads, but the spawner loses the ability to spawn
     /// non-Send tasks.
--- a/embassy-stm32/src/rcc/f4/mod.rs
+++ b/embassy-stm32/src/rcc/f4/mod.rs
@@ -63,7 +63,7 @@ impl<'d> Rcc<'d> {
         }
 
         let sysclk = if sysclk_on_pll {
-            plls.pllsysclk.unwrap()
+            unwrap!(plls.pllsysclk)
         } else {
             sysclk
         };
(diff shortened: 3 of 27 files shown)


embassy/Cargo.toml
```diff
@@ -14,6 +14,7 @@ time-tick-32768hz = ["time"]
 time-tick-1000hz = ["time"]
 time-tick-1mhz = ["time"]
 
+defmt = ["dep:defmt", "embassy-macros/defmt"]
```
bgamari, line 17: Note that this apparently requires `-Z namespaced-features`.

embassy-macros/src/lib.rs
```diff
@@ -342,6 +342,15 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
 
     let chip_setup = chip::generate(&embassy_prefix, config);
 
+    #[cfg(feature = "defmt")]
+    let spawn = quote! {
+        unwrap!(spawner.spawn(__embassy_main(spawner, p)));
```
Dirbaio, line 347: This should somehow ensure the `unwrap!` macro is in-scope in the user's code, I'm not sure if it's doing it currently.

embassy/Cargo.toml
Dirbaio: In the past we've decided to not use `namespaced-features`. the problem is that it requires enabling it in end user's `.cargo/config.toml`. There's a similar situation already with `embassy-traits`.

Maybe we can add a `Spawner::must_spawn()` method so the panic is inside `embassy` so it can simply use the `defmt` featurE?

embassy-macros/src/lib.rs
bgamari: I'll admit that I'm not sure how to do this.

embassy/Cargo.toml
bgamari: This sounds plausible.

bgamari:
@Dirbaio, thanks for the review! I believe I have addressed your comments.

embassy-stm32/src/clock.rs
```diff
@@ -255,7 +255,7 @@ impl TimerInner {
         self.stop_and_reset();
 
         let psc = timer_freq.0 / TICKS_PER_SECOND as u32 - 1;
-        let psc: u16 = psc.try_into().unwrap();
+        let psc: u16 = unwrap!(psc.try_into().map_err(|_| "psc division"));
```
bgamari, line 258: This is a new change. Unfortunately, the `map_err` is necessary because the error type here doesn't have a `Format` instance.

bobmcwhirter (commented):
Seems straight-forward to me.

embassy/src/executor/mod.rs
```diff
@@ -56,6 +56,17 @@ impl Spawner {
         }
     }
 
+    /// Used by the `embassy_macros::main!` macro to throw an error when spawn
+    /// fails. This is here to allow conditional use of `defmt::unwrap!`
+    /// without introducing a `defmt` feature in the `embassy_macros` package,
+    /// which would require use of `-Z namespaced-features`.
+    pub fn must_spawn<F>(&self, token: SpawnToken<F>) -> () {
+        #[cfg(feature = "defmt")]
```
Dirbaio, line 64: This should use straight `unwrap!()`, which forwards to defmt or core depending on enabled features. See `src/fmt.rs` .

Dirbaio (changes requested):

embassy/src/executor/mod.rs
bgamari: Fixed.

Dirbaio (approved):
Awesome! Thanks :D

Dirbaio:
rustfmt'd it for you :P

PR #337, opened 2021-08-02
Introduce future types

lulf:
Instead of using inferred types for send() and recv(), introduce future types allowing them to be more easily wrapped in other types without `Box` or other unstable features.

I think this change should be sound:

* For send(), sender is already Sync + Send, and there is only a syntactic change.
* For recv(), the added lifetime argument should ensure that only 1 RecvFuture can be created at a time or else the borrow checker will complain. Moreover, making RecvFuture Send + Sync should be safe because there can exist only 1, and it also locks the underlying channel_cell.

@huntc ^^

### DIFF
--- a/embassy/src/util/mpsc.rs
+++ b/embassy/src/util/mpsc.rs
@@ -156,7 +156,7 @@ where
     /// closed by `recv` until they are all consumed.
     ///
     /// [`close`]: Self::close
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
LGTM, thank you!

huntc:
Thanks @lulf !

Issue #235, opened 2021-06-08
embassy::main macro enables RTC but doesn't expose priority for configuration


jacobrosenthal:
This panics with "1 ERROR panicked at 'sd_softdevice_enable err SdmIncorrectInterruptConfiguration`" when used together with the nrf_softdevice crate requires reserving priority 0 1 and 3 I believe

>dirbaio: the thing is it can't go in Config because that's the config for embassy::init, which is not what sets up RTC1. It's the macro that does it

PR #338, opened 2021-08-02
Remove dyn in time Clock/Alarm.

### LINKED ISSUE #322: Dyn-less clock/alarm
We can replace the current use of `dyn` with `extern Rust` functions like [this](https://github.com/embassy-rs/critical-section/blob/main/src/lib.rs#L69). 

This saves the cost of virtual calls and vtables. In a given chip, you only have one clock anyway.

Dirbaio:
Fixes #322

- nrf: updated, tested, working :heavy_check_mark: 
- STM32: updated, tested on H7, working :heavy_check_mark: 
- rp: updated, test todo :construction:


embassy-stm32/src/time_driver.rs
```diff
@@ -0,0 +1,329 @@
...
+        Self {
+            timestamp: Cell::new(u64::MAX),
+            callback: Cell::new(ptr::null()),
+            ctx: Cell::new(ptr::null_mut()),
+        }
+    }
+}
+
+struct State {
+    /// Number of 2^23 periods elapsed since boot.
```
timokroeger, line 80: 2^15

embassy-nrf/src/lib.rs
```diff
@@ -162,7 +163,10 @@ pub fn init(config: config::Config) -> Peripherals {
     while r.events_lfclkstarted.read().bits() == 0 {}
 
     // Init GPIOTE
-    crate::gpiote::init(config.gpiote_interrupt_priority);
+    gpiote::init(config.gpiote_interrupt_priority);
+
+    // init RTC time driver
+    time_driver::init();
```
timokroeger, line 169: Configurable irq priority would be nice to be consistent with `gpiote::init()`

Dirbaio: Agreed! Will do this on a follow-up PR, since the previous code couldn't do this anyway (when using `embassy::main`). There's an issue for this already (#235)

embassy-stm32/src/time_driver.rs
Dirbaio: Oopsie. Fixed, thanks!

thalesfragoso (commented):
Awesome work, thanks! I just left two comments with the smallest of nits...

PR #341, opened 2021-08-03
Add uart::Read DMA-based implementation

lulf:
* Rename existing read() to bread() (blocking)

Only tested with v2 usart on L0.

### DIFF
--- a/embassy-stm32/src/usart/v1.rs
+++ b/embassy-stm32/src/usart/v1.rs
@@ -96,7 +96,7 @@ impl<'d, T: Instance, TxDma, RxDma> Uart<'d, T, TxDma, RxDma> {
         Ok(())
     }
 
-    pub fn bread(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+    pub fn read_blocking(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
         unsafe {
             let r = self.inner.regs();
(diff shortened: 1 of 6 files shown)


bobmcwhirter:
"bread" because of function name clashes? Do appropriate trait `use` statements not disambiguate?

If we need to rename, I think I'd prefer `read_blocking(...)` to parallel `read_dma(...)` mayhap.

Dirbaio:
I'm not sure about the `bread` name either. First, it looks with bread :bread: . Second, the `b` prefix like in `bwrite` [will be removed](https://github.com/rust-embedded/embedded-hal/pull/280) in embedded-hal. `read_blocking` sounds better.

Unfortunately there's no blocking uart read trait in embedded-hal :S

lulf:
> I'm not sure about the `bread` name either. First, it reads like bread bread . Second, the `b` prefix like in `bwrite` [will be removed](https://github.com/rust-embedded/embedded-hal/pull/280) in embedded-hal. `read_blocking` sounds better.
> 

This ^^ was the reason for naming it :bread: , but I agree `read_blocking` is way better. In any case, :bread: sounds better as a type:

```rust
impl Bread {
   fn slice() -> &[BreadSlice];
}
```

Dirbaio (approved):
LGTM now that the :bread: is gone :) 

Thank you!

PR #344, opened 2021-08-04
Remove builders from Config(s) and examples.

bobmcwhirter:


examples/stm32h7/src/bin/dac.rs
```diff
@@ -54,3 +52,16 @@ fn to_sine_wave(v: u8) -> u8 {
...
+    config.rcc = rcc_config();
+    config
+}
+
+fn rcc_config() -> rcc::Config {
+    let mut config = rcc::Config::default();
+    config.sys_ck = Some(400.mhz().into());
+    config.pll1.q_ck = Some(100.mhz().into());
+    config
+}
```
timokroeger, line 67: Duplicated for each h7 example. Could be moved to `example_common`

bobmcwhirter:
Good point!

examples/stm32h7/src/bin/dac.rs
Dirbaio: I would make it a single method too. `config.rcc.sys_ck = xx`

examples/stm32h7/src/bin/eth.rs
```diff
@@ -108,16 +108,11 @@ fn main() -> ! {
...
-    rcc_config.pll1.q_ck = Some(Hertz(100_000_000));
-    let config = Config::default().rcc(rcc_config);
-
-    let mut p = embassy_stm32::init(config);
+    let mut p = embassy_stm32::init(config());
 
     // Constrain and Freeze clock
 
-    let mut rcc = Rcc::new(&mut p.RCC, RccConfig::default());
+    let mut rcc = Rcc::new(&mut p.RCC, rcc::Config::default());
```
Dirbaio, line 115: This is initializing RCC twice (though it already did that before...?), it shouldn't do so :S

examples/stm32f0/src/example_common.rs
```diff
@@ -12,7 +12,9 @@ use embassy_stm32::Config;
 pub fn config() -> Config {
     let mut rcc_config = rcc::Config::default();
     rcc_config.enable_debug_wfe = true;
-    Config::default().rcc(rcc_config)
+    let mut config = Config::default();
+    config.rcc = rcc_config;
```
Dirbaio, line 16: `config.rcc.enable_debug_wfe = true`, so you don't need two vars.

Dirbaio (approved):
:rocket: Thank you!

PR #345, opened 2021-08-05
Forward attributes to generated task function

### LINKED ISSUE #18: Doc comment on #[task] not forwarded to the function
``` rust
/// Entrypoint to the application after hardware setup has finished.
#[task]
pub async fn run() {}
```

embassy-doc-comments

Liamolucko:
Fixes #18

Syn considers doc comments to be attributes, so forwarding those fixes the issue.
It's probably a good idea to forward most attributes anyway.


Dirbaio (approved):
I didn't expect this to be that simple! Very nice, thank you!

In the future we might want to whitelist which attributes are allowed, but this already improves the status quo, so let's merge it and improve later. #346

PR #356, opened 2021-08-12
Add IRQ-driven buffered USART implementation for STM32 v2 usart

lulf:
* Implementation based on nRF UARTE, but simplified to not use DMA to
  avoid complex interaction between DMA and USART.
* Implementation of AsyncBufRead and AsyncWrite traits
* Some unit tests to ring buffer

### DIFF
--- a/embassy-hal-common/src/ring_buffer.rs
+++ b/embassy-hal-common/src/ring_buffer.rs
@@ -67,11 +67,11 @@ impl<'a> RingBuffer<'a> {
         self.empty = self.start == self.end;
     }
 
-    pub fn is_full(&mut self) -> bool {
+    pub fn is_full(&self) -> bool {
         self.start == self.end && !self.empty
     }
 
-    pub fn is_empty(&mut self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.empty
     }
 
--- a/embassy-hal-common/src/ring_buffer.rs
+++ b/embassy-hal-common/src/ring_buffer.rs
@@ -67,6 +67,14 @@ impl<'a> RingBuffer<'a> {
         self.empty = self.start == self.end;
     }
 
+    pub fn is_full(&mut self) -> bool {
+        self.start == self.end && !self.empty
+    }
+
+    pub fn is_empty(&mut self) -> bool {
+        self.empty
+    }
+
     pub fn clear(&mut self) {
         self.start = 0;
         self.end = 0;
--- a/embassy-hal-common/src/ring_buffer.rs
+++ b/embassy-hal-common/src/ring_buffer.rs
@@ -82,3 +82,40 @@ impl<'a> RingBuffer<'a> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn push_pop() {
+        let mut b = [0; 4];
+        let mut rb = RingBuffer::new(&mut b);
+        let buf = rb.push_buf();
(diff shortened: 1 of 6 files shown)


lulf:
Needs https://github.com/embassy-rs/stm32-data/pull/75

embassy-stm32/src/usart/v2.rs
```diff
@@ -153,25 +159,224 @@ impl<'d, T: Instance, RxDma> embedded_hal::blocking::serial::Write<u8>
...
+
+            state.rx_waker.register(cx.waker());
+            Poll::<Result<&[u8], embassy::io::Error>>::Pending
+        })
+    }
+    fn consume(mut self: Pin<&mut Self>, amt: usize) {
+        self.inner.with(|state| {
+            state.rx.pop(amt);
+        });
+        self.inner.pend();
```
timokroeger, line 356: Would it be enough to call pend only when the buffer was full?

embassy-stm32/src/usart/v2.rs
```diff
@@ -153,25 +159,224 @@ impl<'d, T: Instance, RxDma> embedded_hal::blocking::serial::Write<u8>
...
+                return Poll::Pending;
+            }
+
+            let n = core::cmp::min(tx_buf.len(), buf.len());
+            tx_buf[..n].copy_from_slice(&buf[..n]);
+            state.tx.push(n);
+
+            Poll::Ready(Ok(n))
+        });
+        self.inner.pend();
```
timokroeger, line 379: I think it is only required to manually pend the ISR when the buffer was empty. If there was data still in the buffer the TXE interrupt should still be enabled.

examples/stm32l0/src/bin/usart_irq.rs
```diff
@@ -0,0 +1,51 @@
...
+            interrupt::take!(USART1),
+            &mut TX_BUFFER,
+            &mut RX_BUFFER,
+        )
+    };
+
+    usart.write_all(b"Hello Embassy World!\r\n").await.unwrap();
+    info!("wrote Hello, starting echo");
+
+    let mut buf = [0; 1];
```
timokroeger, line 46: Maybe choose a larger buffer size e.g. 4, to illustrate that the buffered UART can efficiently handle multibyte receives and might read *up to* 4 bytes.

timokroeger (commented):
Looks good overall, thank you for the efforts!
I did a quick port to USARTv1 for testing (because I do not have v2 hardware) and it worked without any major changes.

Once it is merged we can think about moving the buffered UART code to a shared place where we can have it for both v1 and v2.

Furthermore I think there is the potential to reduce the interrupt overhead even further if we enable and check IRQs only as required. E.g only run `on_tx()` when the TXIE bit is set.

lulf:
@timokroeger Thanks for the review. I have address your comments. 

PR now points to stm32-data PR https://github.com/embassy-rs/stm32-data/pull/78 to fix an issue with the USART IRQ pins

embassy-hal-common/src/ring_buffer.rs
```diff
@@ -67,6 +67,14 @@ impl<'a> RingBuffer<'a> {
         self.empty = self.start == self.end;
     }
 
+    pub fn is_full(&mut self) -> bool {
```
bobmcwhirter, line 70: Probably not important, but `&mut` isn't actually required is it?

bobmcwhirter (approved):
LGTM

PR #357, opened 2021-08-16
Add H7 exti button example using correct EXTI reg block offsets.

bobmcwhirter:


lulf (approved):
Nice catch, looks good.

Dirbaio (approved):
:top:

PR #358, opened 2021-08-16
Use const fn for uart state storage

lulf:


Dirbaio:
This won't build without adding another nightly feature `#![feature(const_fn_trait_bound)]`. I'd rather keep the nightly feature count to a minimum... What's your use case? creating it directly in a static?

lulf:
@Dirbaio Yeah, I noticed that I had to add this to all examples and then abandoned it :) The purpose was to store State directly in static, and I noticed this feature was already depended on by the embassy core crate at https://github.com/embassy-rs/embassy/blob/master/embassy/src/lib.rs#L3

thalesfragoso:
> The purpose was to store State directly in static

Don't you usually need `&mut State` for it to be useful ? In that case using `Forever` seems better.

lulf:
> > The purpose was to store State directly in static
> 
> Don't you usually need `&mut State` for it to be useful ? In that case using `Forever` seems better.

That's true. I was thinking using 'native rust' constructs would be better. But it doesn't matter. I agree we shouldn't sprinkle nightly features around, and I guess maybe it can be worked around in embassy core?

PR #369, opened 2021-08-21
Add defmt support for mpsc errors.

derekdreery:


embassy/src/util/mpsc.rs
```diff
@@ -430,6 +438,20 @@ impl<T> fmt::Display for TrySendError<T> {
...
 }
 
+#[cfg(feature = "defmt")]
+impl<T> defmt::Format for TrySendError<T> {
+    fn format(&self, fmt: defmt::Formatter<'_>) {
+        defmt::write!(
+            fmt,
+            "{}",
+            match self {
+                TrySendError::Full(..) => "no available capacity",
```
Dirbaio, line 448: With this way of writing it, the strings `"no available capacity"` and `"channel closed"` will be actually stored in flash and sent char-by-char over the wire, losing the efficiency of defmt.

If you do this instead, the strings will be interned, so they won't appear in flash and only a numeric ID will be sent over the wire:


```rust
match self {
                TrySendError::Full(..) =>defmt::write!(fmt, "no available capacity"),
                TrySendError::Closed(..) => defmt::write!(fmt, "channel closed"),
}

**Review by Dirbaio — CHANGES_REQUESTED** [review]
Thanks for the PR! Commented just one small nit.

File: embassy/src/util/mpsc.rs (hunk 1 above)
Comment (derekdreery) [reply embassy/src/util/mpsc.rs:448]: I'm still trying to get an intuition for how defmt interns.

Comment (Dirbaio) [reply embassy/src/util/mpsc.rs]: Only the "first" string in write!, info!, warn! etc (the string that can contain placeholders for format args like `{}`) is interned. Everything else is the format args values, which is serialized and sent over the wire, never tokenized. If an arg is a `str`, it'll be sent char-by-char, not tokenized.

(The only exception is if you use `intern!` to get a `defmt::Str`, that one is tokenized when used as a format arg)

**Review by Dirbaio — APPROVED** [review]
Thank you!

File: embassy/src/util/mpsc.rs (hunk 1 above)
Comment (mattico) [reply embassy/src/util/mpsc.rs]: Could also do

```rust
        defmt::write!(
            fmt,
            "{=istr}",
            match self {
                TrySendError::Full(..) => defmt::intern!("no available capacity"),
                TrySendError::Closed(..) => defmt::intern!("channel closed"),
            }
        )
```

PR #370, opened 2021-08-21
Expose EASY_DMA_SIZE.

derekdreery:


embassy-nrf/src/lib.rs
```diff
@@ -67,6 +67,8 @@ mod chip;
 #[path = "chips/nrf52840.rs"]
 mod chip;
 
+pub use chip::EASY_DMA_SIZE;
```
derekdreery, line 70: I tried putting the docstring here, but it didn't get picked up. I can remove it if you prefer.

Dirbaio (approved):
Thank you!

PR #371, opened 2021-08-24
Fix blocking-write for SPI.

bobmcwhirter:


embassy-stm32/src/spi/v3.rs
```diff
@@ -559,7 +583,7 @@ impl<'d, T: Instance, Tx: TxDmaChannel<T>, Rx: RxDmaChannel<T>> traits::FullDupl
     for Spi<'d, T, Tx, Rx>
 {
     #[rustfmt::skip]
-    type WriteReadFuture<'a> where Self: 'a = impl Future<Output = Result<(), Self::Error>> + 'a;
+    type WriteReadFuture<'a> where Self: 'a = impl Future<Output=Result<(), Self::Error>> + 'a;
```
lulf, line 586: Hmm, cargo fmt likes the space

lulf (approved):
LGTM, but cargo fmt incompatible change (but skipped due to rustfmt attribute)

lulf (approved):

PR #378, opened 2021-08-27
Use our beloved Rust instead of Python

numero-744:
Done for /embassy-stm32 and /stm32-metapac
The new generator is in /stm32-gen-features
A CI check "generated features up to date" could/should be added in another MR

Draft because:

- [x] This is an unsolicited PR (surpriiiiise! :partying_face:) which may therefore be irrelevant
- [x] If this PR is approved then metapac support could/should be added in the same PR
- [x] The generated data should be sorted


lulf:
LGTM, which work on the gen_features.py is blocking this?

numero-744:
I had read in the Matrix embassy channel that someone (IIRC it was you @lulf) was working on changing some naming conventions about the cores in the features list because there was a problem with the use of the `-` or something like that. If it is okay I think I'll put the features of `stm32-metapac/gen_features` in the same Rust generator this week-end. Also expect some refactoring.

By the way I did not mention that running the Python script was about ~18s for me, now build+run of this new Rust code is (when only the root crate is modified) <6s, and without optimizations the same Rust code was running in >30s, I don't know why it gets even slower than Python. This is why I have enabled optimizations in `dev` profile, I want to minimize build+run not build OR run (and I assume that `cargo check` is fast enough if no build is needed).

Another benefit of Rust over Python here is that there are some unit tests.

Do you think it would be relevant to add a CI check "generated features are up to date" and/or "unit tests of features generator pass"?

lulf:
One comment on the code: It would be good to sort the list of chips like in python so order is deterministic.

> I had read in the Matrix embassy channel that someone (IIRC it was you @lulf) was working on changing some naming conventions about the cores in the features list because there was a problem with the use of the `-` or something like that. If it is okay I think I'll put the features of `stm32-metapac/gen_features` in the same Rust generator this week-end. Also expect some refactoring.

Ah, right :D Those changes will only impact the build.rs of stm32-metapac, so iirc it will not change the python scripts. I am focused on other tasks right now, and that fix also seems quite small anyway.

> 
> By the way I did not mention that running the Python script was about ~18s for me, now build+run of this new Rust code is (when only the root crate is modified) <6s, and without optimizations the same Rust code was running in >30s, I don't know why it gets even slower than Python. This is why I have enabled optimizations in `dev` profile, I want to minimize build+run not build OR run (and I assume that `cargo check` is fast enough if no build is needed).
> 

I believe this is anyway run on demand, and not very often, so that seems OK with me. 

> Another benefit of Rust over Python here is that there are some unit tests.
> 

Yeah, saw that, thats great!

> Do you think it would be relevant to add a CI check "generated features are up to date" and/or "unit tests of features generator pass"?

I think that check could be useful and prevent forgetting to rerun this when changing list of chip families to generate.

numero-744:
The `toml` crate I am using here does not seem to be able to sort the list of the chips… I see two solutions:

* find another `toml` crate…
* sort the output of the `toml` serialization function ?
* do not use `toml`, do some custom formatting (better IMO because, at least for this first generation script, the output is really easy to format)

For the CI I am used to GitLab not GitHub so I will try to do something, but I consider it as optional for now. Btw GH does not enable CI on my fork by default, is it *sound* to enable it?

numero-744:
/me

1. Sort the paths before inserting them into the HashMap :test_tube: 
2. Notice that the resulting toml is still not sorted :thinking: 
3. Notice that `glob` output was already sorted and that keys are inserted in order into the hashmap… :disappointed: 

I'll make the program do the `stm32-metapac/gen_features.py` stuff, and if I have time I'll replace `toml` crate by a custom formatting function to keep chips sorted, unless someone suggests me a better solution.

Dirbaio:
Yeah, Rust hashmaps don't keep order, while Python ones do... Not using `toml` seems the easiest solution here.  :+1:

lulf (approved):
These changes look good to me, thanks for fixing this!

PR #379, opened 2021-08-27
Random range

bobmcwhirter:


Dirbaio:
Perhaps this should be some `struct RngWhatever<T: Rng>` that wraps the raw `Rng` and gives extra functionality?

Adding a method to the trait has a few disadvantages: 
- each impl has to copypaste the code
- adding more of such methods is a breaking change, while adding methods to the util wrapper struct isn't.

Dirbaio:
I meant making the random utils a struct, not a trait :) 

The "get a random u32 out of a RNG that gives random bytes" is the same for all possible hardware out there.

Naming can be confusing. Maybe name the trait `RngBytes` and the util struct `Rng`.?

bobmcwhirter:
Yah I've got another commit to add. Tomorrow!

bobmcwhirter:
@Dirbaio How's this look? Renamed Random->Rng, since it is really just an RNG wrapper, and then re-created Random as a universal struct for next_x(...) helper, avoiding weird Util or Bytes suffix.

Dirbaio (approved):
Awesome, thank you so much! :)

PR #380, opened 2021-08-30
Fix STM32G0 ADC

bgamari:
The STM32G0 ADC differs quite significantly from previous series.

Depends upon https://github.com/embassy-rs/stm32-data/pull/82.


embassy-stm32/src/rcc/g0/mod.rs
```diff
@@ -18,10 +18,38 @@ pub const LSI_FREQ: u32 = 32_000;
...
+            HSI16Prescaler::Div2 => 0x01,
+            HSI16Prescaler::Div4 => 0x02,
+            HSI16Prescaler::Div8 => 0x03,
+            HSI16Prescaler::Div16 => 0x04,
+            HSI16Prescaler::Div32 => 0x05,
+            HSI16Prescaler::Div64 => 0x06,
+            HSI16Prescaler::Div128 => 0x07,
+        }
+    }
+}
```
mattico, line 50: Nit:

```rust
#[derive(Clone, Copy)]
pub enum HSI16Prescaler {
    NotDivided = 0,
    Div2 = 1,
    Div4 = 2,
    Div8 = 3,
    Div16 = 4,
    Div32 = 5,
    Div64 = 6,
    Div128 = 7,
}

impl Into<u8> for HSI16Prescaler {
    fn into(self) -> u8 {
        self as u8
    }
}
```

Dirbaio (commented):
Is `v3` only present on stm32g0? It confuses me that the file is called `v3` but only enabled if `adc_g0` which seems to suggest that yes, but `v3.rs` contains `#[cfg(rcc_g0)]`, which seems to suggest that no?

Dirbaio (approved):
All looks good now! Thank you! :rocket:

PR #381, opened 2021-09-01
Add HAL for SubGhz peripheral for STM32 WL series

lulf:
Based on the HAL from stm32wl, the peripheral driver has been
modified to fit into embassy, using the embassy APIs, providing
operation of the radio peripheral.

The initial version does not offer any async APIs, but the example
shows how the radio IRQ can be used to perform async TX of the radio.

Edit: One question I have in my mind is wrt. DMA vs. Non DMA. I would prefer if I could choose one or the other. Right now, everything requires NoDma since its just copied from stm32wl-hal. Unless anyone have any reservations about it, I'd like to change it to use DMA and flip everything to async. However, I think before that, I need to make sure https://github.com/ivajloip/rust-lorawan/issues/41 is completed, otherwise it will not be able to integrate with the lora link layer.


Dirbaio (approved):
:rocket:

PR #382, opened 2021-09-01
Typestate nrf timer

fnafnio:
useful for hooking up the PPI to an Event without needing interrupt
tested with buffered_uart example on nRF52840-DK

the names are aweful but descriptive, I'm very open to suggestions

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -15,9 +15,8 @@ use crate::gpio::sealed::Pin as _;
 use crate::gpio::{OptionalPin as GpioOptionalPin, Pin as GpioPin};
 use crate::pac;
 use crate::ppi::{AnyConfigurableChannel, ConfigurableChannel, Event, Ppi, Task};
-use crate::timer::Frequency;
 use crate::timer::Instance as TimerInstance;
-use crate::timer::Timer;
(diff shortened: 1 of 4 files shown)


fnafnio:
Of course some docs on the differences and use cases should be added, like suggesting to set a short on something like clear counter

Dirbaio:
Yeah, naming for this is quite something... I can't think of anything better either :S 

Perhaps the type aliases are not necessary? especially since the Cc struct doesn't have corresponding type aliases. They don't gain you much, `Timer<TIMER3, NotAwaitable>` is the same length as `NotAwaitableTimer<TIMER3>`...

We could also use default generic params, the default should probably be NotAwaitable. This way, the user would write `Timer<TIMER3>` for not awaitable, and `Timer<TIMER3, Awaitable>`

Another thing: if both `new` methods are named equally, the compiler will reject `Timer::new` as ambiugous. Perhaps they should be `new` and `new_awaitable`? especially if we remove the type aliases.

fnafnio:
I've implemented the suggestions and added an example for the awaitable timer.
Having less redundancy definitely improves the feel

I guess for the not awaitable timer the comment leading one to the uarte implementation should be enough?

Dirbaio (approved):
Awesome, thank you!

Yeah, docs look good, and we can always improve them later.

PR #388, opened 2021-09-06
Don't set SAF=true, do set RA=true for Ethernet.

bobmcwhirter:
Source-Address-Filtering is not helping the board to receive packets.
For unknown reasons, the Receive-All is required, when in theory
it should not be required. Until we figure it out, follow the
stm32h7xx-hal example of setting RA=true.


Dirbaio (approved):
LGTM, thank you!

thalesfragoso (approved):
Thanks!

PR #389, opened 2021-09-06
STM32H7: Ethernet: Disable RA in MAC filtering, fix order of MACA0 register writes

adamgreig:
Previously the MACA0 register was written LR and then HR, but the value across both registers only updates when LR is written, so only a partial MAC address was being configured in the core. Presumably this is why destination filtering was failing and the RA (receive all) bit needed to be set.

With this reordering of the writes, the correct MAC address is loaded into the MAC core, and so the RA bit is no longer required.

Tested on an H745ZI nucleo board.


Dirbaio (approved):
Very nice catch! Thank you :)

Tested working on H755ZI too.

PR #390, opened 2021-09-09
Wasm executor support

lulf:


embassy/src/executor/arch/wasm.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// for example by passing it as an argument to the initial tasks.
+    ///
+    /// This function requires `&'static mut self`. This means you have to store the
+    /// Executor instance in a place where it'll live forever and grants you mutable
+    /// access. There's a few ways to do this:
+    ///
+    /// - a [Forever](crate::util::Forever) (safe)
+    /// - a `static mut` (unsafe)
+    /// - a local variable in a function you know never returns (like `fn main() -> !`), upgrading its lifetime with `transmute`. (unsafe)
+    pub fn start(&'static mut self, init: impl FnOnce(Spawner) + Send) {
```
Dirbaio, line 65: Why is `Send` required on `init`?

embassy/src/executor/raw/mod.rs
```diff
@@ -10,7 +10,7 @@
 mod run_queue;
 #[cfg(feature = "time")]
 mod timer_queue;
-mod util;
+pub(crate) mod util;
```
Dirbaio, line 13: I think this pub is not used?

Dirbaio:
This is awesome! :D 

Could you add building the WASM example to CI, so that it doesn't bitrot?

embassy/src/time/mod.rs
```diff
@@ -51,6 +51,9 @@ mod timer;
 #[cfg(feature = "std")]
 mod driver_std;
 
+#[cfg(feature = "wasm")]
+mod driver_wasm;
```
Dirbaio, line 55: I think this file is missing

embassy/src/executor/arch/wasm.rs
lulf: Copy paste from some other executor :)

embassy/src/executor/raw/mod.rs
lulf: Ah, yes. I tried to use the UninitCell, but discovered it did not implement Sync + Send.

lulf:
@Dirbaio Also needs https://github.com/embassy-rs/critical-section/pull/2 right now, not sure if there is another workaround you can think of?

Dirbaio (approved):
Thank you! :)

PR #399, opened 2021-09-15
Support for STM32L1

lulf:
* Add RCC
* Fix more issues with dash in chip names
* Update stm32-data version
* Add blinky and spi example
* Add to CI


Dirbaio (approved):
Very nice!! Thank you! :rocket: 

Nice refactor on the SPI!

PR #403, opened 2021-09-24
Small adjustment to 'set_as_af' interface

mryndzionek:
Small adjustment to 'set_as_af' interface. Mostly just to remove the dependency on the raw GPIO registers in i2c.

### DIFF
--- a/embassy-stm32/src/gpio.rs
+++ b/embassy-stm32/src/gpio.rs
@@ -55,14 +55,6 @@ impl From<Speed> for vals::Ospeedr {
     }
 }
 
-/// Type settings
-#[derive(Debug)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub enum OutputType {
-    PushPull,
-    OpenDrain,
-}
-
 /// GPIO input driver.
 pub struct Input<'d, T: Pin> {
     pub(crate) pin: T,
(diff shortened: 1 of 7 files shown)


embassy-stm32/src/gpio.rs
```diff
@@ -55,6 +55,14 @@ impl From<Speed> for vals::Ospeedr {
     }
 }
 
+/// Type settings
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum OutputType {
```
Dirbaio, line 61: This is only used in the private `set_as_af` API, so it should be `pub(crate)`

Dirbaio (changes requested):
Thank you! Looks great, just one small nit

embassy-stm32/src/gpio.rs
mryndzionek: If I do this I get:
```
crate-private type `gpio::OutputType` in public interface
```

Dirbaio: oh, it's because `sealed::Pin` is "pub but not really" due to `pub(crate) mod sealed`. 

Moving it inside `mod sealed` should work.

mryndzionek: `set_as_af` is public, is it not?

mryndzionek: If I move it inside `sealed::Pin` it's not seen in all the other external modules: `i2c`, `spi`, `uart`, `bxcan`, etc

Dirbaio: `Pin` is a [sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed), to forbid external crates from implementing it.

It's split in two: a private `trait gpio::sealed::Pin`, and a public `trait gpio::Pin: gpio::sealed::Pin`. 

Both are `pub`, but `mod sealed` is `pub(crate)`. This makes it so `sealed::Pin` is inaccessible from outside the crate even if it's `pub`. This is what prevents outside crates from implementing it.

This also makes it so that methods in `sealed::Pin` are effectively crate-private, they're not accessible from outside the crate either.

Dirbaio (approved):
Thank you!

PR #405, opened 2021-09-27
Feature: CRC HAL

theunkn0wn1:
This PR implements a `embassy-stm32` HAL abstraction around the CRC peripheral.

- [x] crc_v1
- [x] crc_v2
- [x] crc_v3

I don't have access to a device with the V2 or V3 crc peripherals, however I have tested the V1 against my stm32f446RETx.


embassy-stm32/src/crc/v1.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::peripherals::CRC;
+use crate::rcc::sealed::RccPeripheral;
+
+pub struct Crc {
+    _peripheral: CRC,
+}
+
+impl Crc {
+    /// Instantiates the CRC32 peripheral and initializes it to default values.
+    pub fn new(peripheral: CRC) -> Self {
```
lulf, line 11: Most peripherals take an Unborrow, i.e. Unborrow<Target = CRC> so the HAL can be used with both &mut and owned types. Would it make sense to allow that here too?

embassy-stm32/src/crc/v2.rs
```diff
@@ -0,0 +1,175 @@
...
+pub enum PolySize {
+    Width7,
+    Width8,
+    Width16,
+    Width32,
+}
+
+impl Crc {
+    /// Instantiates the CRC32 peripheral and initializes it to default values.
+    pub fn new(peripheral: CRC, config: CrcConfig) -> Self {
```
lulf, line 67: Same as for v1

lulf (approved):
Looks good :rocket: ! I left a comment on the fn new() taking ownership vs using Unborrow

embassy-stm32/src/crc/mod.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg_attr(crc_v2, path = "v2.rs")]
+#[cfg_attr(crc_v1, path = "v1.rs")]
+#[cfg_attr(crc_v3, path = "v2.rs")]
```
Dirbaio, line 3: it's somewhat confusing that `v3` uses `v2.rs`. Could you name it `v2v3.rs`? 

Also could you sort these attrs by version?

embassy-stm32/src/crc/v1.rs
```diff
@@ -0,0 +1,55 @@
...
+        CRC::reset();
+        let mut instance = Self {
+            _peripheral: peripheral,
+        };
+        instance.init();
+        instance
+    }
+
+    /// Resets the CRC unit to default value (0xFFFF_FFFF)
+    pub fn init(&mut self) {
```
Dirbaio, line 25: This is called `reset` in v2/v3. We should try to keep the API consistent across versions. Could you name this `reset`?

embassy-stm32/src/crc/v1.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        self.read()
+    }
+    pub fn read(&self) -> u32 {
+        unsafe { PAC_CRC.dr().read() }
+    }
+
+    /// Reclaims the CRC peripheral.
+    pub fn release(self) -> CRC {
```
Dirbaio, line 50: Embassy doesn't have free/release functions, the way to get back the peripheral is through Unborros as @lulf noted.

embassy-stm32/src/crc/v2.rs
```diff
@@ -0,0 +1,175 @@
...
+pub struct Crc {
+    _peripheral: CRC,
+    _config: CrcConfig,
+}
+
+pub enum CrcConfigError {
+    InvalidPolynomial,
+}
+
+pub struct CrcConfig {
```
Dirbaio, line 15: just `Config`, for consistency, and for not repeating "crc" in `crc::CrcConfig`

embassy-stm32/src/crc/v2.rs
```diff
@@ -0,0 +1,175 @@
...
+    fn reconfigure(&mut self) {
+        unsafe {
+            // Init CRC value
+            PAC_CRC.init().write_value(self._config.crc_init_value);
+            #[cfg(crc_v3)]
+            PAC_CRC.pol().write_value(self._config.crc_poly);
+
+            // configure CR components
+            // (reverse I/O, polysize, poly)
+            PAC_CRC.cr().modify(|w| {
```
Dirbaio, line 99: This can be a `write` since you're setting all the fields. Maybe you can also reset it in the same write?

embassy-stm32/src/crc/v2.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub fn feed_bytes(&mut self, bytes: &[u8]) -> u32 {
+        for byte in bytes {
+            unsafe {
+                PAC_CRC.dr8().write_value(*byte as u32);
+            }
+        }
+        unsafe { PAC_CRC.dr().read() }
+    }
+    /// Feeds a halfword into the CRC peripheral. Returns the computed checksum.
+    pub fn feed_halfword(&mut self, byte: u16) -> u32 {
```
Dirbaio, line 144: argument `byte`/`bytes` should be halfword/word in the following functions

embassy-stm32/src/crc/v2.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Feeds an slice of halfwords into the CRC peripheral. Returns the computed checksum.
+    pub fn feed_halfwords(&mut self, bytes: &[u16]) -> u32 {
+        for byte in bytes {
+            unsafe {
+                PAC_CRC.dr16().write_value(*byte as u32);
+            }
+        }
+        unsafe { PAC_CRC.dr().read() }
+    }
+    /// Feeds a halfword into the CRC peripheral. Returns the computed checksum.
```
Dirbaio, line 159: word

Dirbaio (changes requested):
This looks very good, thanks for working on it! I've commented some details.

embassy-stm32/src/crc/v1.rs
theunkn0wn1: I am unfamiliar with the semantics on this, though armed with @Dirbaio 's comment below about `reclaim` not being part of the API I guess it makes sense.

embassy-stm32/src/crc/v2.rs
theunkn0wn1: I do not wish to write to reset here, since I don't know in what order the writes will occur.
Reset must be written to last.

Dirbaio (approved):
Thank you! :rocket:

PR #406, opened 2021-09-27
Basic STM32F1 support

mryndzionek:
`hello` and `blinky` on STM32F103C8 work okay. I've introduced the dependency on the `cast` crate to be in line with [stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/rcc.rs#L246). Had to explicitly unwrap the casts because the functions return a `Result<>`. Not sure why there are no `stm32f1xx-hal`. I had compilation errors.

The `button` and `button_exti` examples should also work. I'll add them in another PR after testing.


Dirbaio (approved):
Great work! I've tested GPIO input/output and EXTI and it all works! Very clean code too. Thank you! :) 

The only improvement I'd like to suggest is using `TryFrom` instead of `cast`, so we minimize dependencies: `unwrap!(u32::try_from(x))`,  [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f14e0c4b7a6853f0712733f98fb094a9)

(`unwrap!()` instead of `.unwrap()` is so that the panic message is printed with defmt if the user doesn't have core panic messages enabled, they cause bloat)

mryndzionek:
> The only improvement I'd like to suggest is using `TryFrom` instead of `cast`, so we minimize dependencies: `unwrap!(u32::try_from(x))`, [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f14e0c4b7a6853f0712733f98fb094a9)

There is currently one problem with this:
```
error[E0277]: the trait bound `TryFromIntError: Format` is not satisfied
```
If I try:
```rust
impl defmt::Format for core::num::TryFromIntError { ...
```
I get:
```
only traits defined in the current crate can be implemented for arbitrary **types**
```

mryndzionek:
Okay, I see that the missing implementation is added in [defmt master](https://github.com/knurling-rs/defmt/commit/ceb61cedc0688192c50b2700e4660ea8547e3f1c), but not released yet. I don't know how `defmt` avoids this `rustc(E0117)` error.

Dirbaio:
Oh, that's unfortunate. If that's the case then normal `.unwrap()` is acceptable, we can change it later when defmt 0.3 is out.

mryndzionek:
With `.unwrap()` I get:
```
error: section '.rodata' will not fit in region 'FLASH': overflowed by 39 bytes
```
Only `--release` compiles...

Dirbaio:
Yep, that's the bloat from `core::fmt`. 

You can get rid of it by removing `print-defmt` [here](https://github.com/mryndzionek/embassy/blob/stm32f1_support/examples/stm32f1/Cargo.toml#L31) and adding this to `.cargo/config`:

```toml
[unstable]
build-std = ["core"]
build-std-features = ["panic_immediate_abort"]
```

The downside is that normal `core` (non-defmt) panic messages aren't printed. The examples don't do that because it's beginner-unfriendly. `fmt` bloat is a problem of rust embedded as a whole, not just embassy.

mryndzionek:
Hey, I removed the bloat from core::fmt and rebased. Let me know if this is okay, or should I remove commit 90884eb (will the CI build then?).

Dirbaio:
> should I remove commit 90884eb 

Yes, as I said we don't do that in examples because it's somewhat unfriendly: people fork off the examples then get crashes without panic messages. 

> will the CI build then?

It just does `cargo check`, which doesn't attempt to link, so yes.

Dirbaio (approved):
Thank you! :rocket:

PR #407, opened 2021-09-27
Fix: erroneous behavior of pins  being configured into AF

theunkn0wn1:
A race condition exists when configuring pins into a peripheral's AF.

In the case of usart_v1 on a stm32f446RE, this causes an erroneous `0xF8` byte to be written to the usart's TX pin during startup.

This PR fixes the race condition by only configuring the pin into AF mode at the end of the configuration.

Minimal falsifying example:
```rs
#![no_std]
#![no_main]
#![feature(type_alias_impl_trait)]

/*
Module declartions
 */
// stuff used across binaries
#[path = "../example_common.rs"]
mod example_common;
use example_common::*;
/*
Includes
 */
use embassy::executor::Spawner;
use embassy::traits::*;
use embassy_stm32::usart::{Config, Uart};
use embassy_stm32::Peripherals;

#[embassy::main]
async fn main(_spawner: Spawner, p: Peripherals) {
    info!("Hello World!");

    let mut usart = Uart::new(p.USART1, p.PA10, p.PA9, p.DMA2_CH7, p.DMA2_CH5, config); 
}

### DIFF
--- a/embassy-stm32/src/gpio.rs
+++ b/embassy-stm32/src/gpio.rs
@@ -311,9 +311,6 @@ pub(crate) mod sealed {
             let pin = self._pin() as usize;
             let block = self.block();
             block
-                .moder()
-                .modify(|w| w.set_moder(pin, vals::Moder::ALTERNATE));
-            block
                 .afr(pin / 8)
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Thank you!

PR #420, opened 2021-10-11
Implements nRF SAADC continuous mode

huntc:
SAADC continuous mode is now supported by offering a new `Continuous` struct alongside the existing `OneShot` one. Its `sample` method returns a `ContinuousStream` (a `futures::Stream`) of buffer samples.

When a single channel is sampled, the internal SAADC timer will be used. For multiple channels, PPI will be configured to wire up the TIMER with SAADC.

A new example is provided to illustrate continuous mode. In summary:

```rust
#[embassy::main]
async fn main(_spawner: Spawner, mut p: Peripherals) {
    let channel_config = ChannelConfig::single_ended(&mut p.P0_02);
    let mut saadc = Continuous::new(p.SAADC, interrupt::take!(SAADC), config, [channel_config]);

    let mut double_buf = [[0; 500]; 2]; // A double buffer for 10_000 samples / 1/20th second
    let mut samples = saadc.sample(&mut double_buf, 1600); // 10_000 samples per second
    while let Some(buf) = samples.next().await {
        info!(
            "samples: {=i16}, {=i16}, {=i16}...",
            &buf[0], &buf[1], &buf[2]
        );
    }
}
```

TODO:

* [ ] Fire the sample task from the interrupt handler and switch buffers there having filled one buffer
* [ ] Provide a stop method
* [ ] Wire up PPI with the timer where more than one channel is to be sampled

### DIFF
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -1,6 +1,7 @@
 #![macro_use]
 
 use core::marker::PhantomData;
+use core::slice;
 use core::sync::atomic::{compiler_fence, Ordering};
 use core::task::Poll;
 use embassy::interrupt::InterruptExt;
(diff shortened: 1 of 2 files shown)


huntc:
Closing in favour of https://github.com/embassy-rs/embassy/pull/425 after discussion

PR #425, opened 2021-10-12
Implements continuous sampling for the nRF SAADC

huntc:
Implements continuous sampling for the nRF SAADC and also renames `OneShot` to `Saadc`. The one-shot behaviour is retained with the `sample` method and a new `run_sampler` method is provided for efficiently (i.e. zero copying) sampler processing. A double buffer is used for continuously sampling, which is swapped appropriately.

A sample frequency is provided and will set the internal timer of the SAADC when there is just one channel being sampled. Otherwise, PPI will be used to hook up the TIMER peripheral to drive the sampling task. Two methods are provided for this: `run_task_sampler` and `run_task_sampler` with the latter available where the compiler sees that just one channel is configured. Note that we set up the PPI and timer behaviour outside of the `Saadc` for maximum flexibility.

A callback is provided to the `run_sampler` method. This is a synchronous callback that should return in a reasonably short space of time. The SAADC could stall if it does not. A reasonable practice is to perform a small amount of processing within the callback to yield a signal, perhaps via `mpsc`. In the case of `mpsc`, the `try_send` method becomes useful.

A new example has been provided to illustrate continuous sampling, along with multiple channels and external timing:

```rust
#[embassy::main]
async fn main(_spawner: Spawner, mut p: Peripherals) {
    let channel_1_config = ChannelConfig::single_ended(&mut p.P0_02);
    let channel_2_config = ChannelConfig::single_ended(&mut p.P0_03);
    let channel_3_config = ChannelConfig::single_ended(&mut p.P0_04);
    let mut saadc = Saadc::new(
        p.SAADC,
        interrupt::take!(SAADC),
        config,
        [channel_1_config, channel_2_config, channel_3_config],
    );

    let mut timer = Timer::new(p.TIMER0);
    timer.set_frequency(Frequency::F1MHz);
    timer.cc(0).write(100); // We want to sample at 10KHz
    timer.cc(0).short_compare_clear();

    let mut ppi = Ppi::new(p.PPI_CH0);
    ppi.set_event(timer.cc(0).event_compare());
    ppi.set_task(saadc.task_sample());
    ppi.enable();

    timer.start();

    let mut bufs = [[[0; 3]; 50]; 2];

    let mut c = 0;
    let mut a: i32 = 0;

    saadc
        .run_task_sampler(&mut bufs, move |buf| {
            for b in buf {
                a += b[0] as i32;
            }
            c += buf.len();
            if c > 10000 {
                a = a / c as i32;
                info!("channel 1: {=i32}", a);
                c = 0;
                a = 0;
            }
            SamplerState::Sampled
        })
        .await;
}
```

### DIFF
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -237,8 +237,7 @@ impl<'d, const N: usize> Saadc<'d, N> {
         .await;
     }
 
-    /// Continuous sampling with double buffers. The sample buffers generally
-    /// should be a multiple of the number of channels configured.
+    /// Continuous sampling with double buffers.
     ///
     /// A task-driven approach to driving TASK_SAMPLE is expected. With a task
     /// driven approach, multiple channels can be used.
@@ -355,8 +354,7 @@ impl<'d, const N: usize> Saadc<'d, N> {
 }
 
 impl<'d> Saadc<'d, 1> {
-    /// Continuous sampling on a single channel with double buffers. The sample
-    /// buffers generally should be a multiple of the number of channels configured.
+    /// Continuous sampling on a single channel with double buffers.
     ///
     /// The internal clock is to be used with a sample rate expressed as a divisor of
     /// 16MHz, ranging from 80..2047. For example, 1600 represnts a sample rate of 10KHz
@@ -374,7 +372,8 @@ impl<'d> Saadc<'d, 1> {
     ) where
         S: FnMut(&[[i16; 1]]) -> SamplerState,
     {
-        self.run_sampler(bufs, Some(sample_rate_divisor), sampler).await;
+        self.run_sampler(bufs, Some(sample_rate_divisor), sampler)
+            .await;
     }
 }
 


huntc:
> Nice work! It would be nicer api if it would be possible to express the sample rate as the frequency rather than as a divisor IMHO, but maybe as a future improvement.

Thanks! It is certainly possible to convey a frequency enum, but with the internal timer, the value can be from 80..2047. That's a big enum. :-) Perhaps there are some common frequencies, but then what if you don't want a common one. Dunno...

lulf:
I was thinking that you could pass the sample rate
+ timer instance , and the implementation could derive the divisor from that (asking the timer for its frequency) . That way the user of the API don't have to do the calculation.

huntc:
> I was thinking that the user could provide the desired sample rate and the timer instance , and the implementation could derive the divisor from that (asking the timer for its frequency) . That way the user of the API don't have to do the calculation.

Could you please provide a sketch of what you're thinking there? Thanks.

huntc:
bors d+

PR #427, opened 2021-10-14
New nrf PPI api (with DPPI support for nRF91 & nRF53)

diondokter:
- Added _ppi and _dppi features to distinguish between the new and the old peripheral.
- Removed ConfigurableChannel and added capacity numbers to the channels
- Replaced the PPI api with a new one using the DPPI terminology (publish & subscribe)
- Updated all tasks and event registers for DPPI

My proposal for the new API.
Tested on my nRF52840 and nRF9160.

Biggest changes for nRF52 is that there's no longer a distinction made between fork task and normal task. You now subscribe tasks to a channel and at runtime it is checked whether or not there's still room for another subscription.
Same for events.

There are differences between the PPI and DPPI though:
- With the PPI you have a limited amount of tasks and events per channel, but a task or event can be used on multiple channels at the same time.
- With the DPPI you have an unlimited amount of tasks and events per channel, but every task or event can only be used on 1 channel.
This is all checked at runtime.

Currently you need to track which tasks and events are assigned to a channel in order to unassign them. For the PPI this data is stored centrally in the registers, so it would be easy to create e.g. `clear_all` and `get_subscribed_tasks` functions. But for the DPPI that data is stored decentrally and so would need some manual tracking.

If there are requests for tracking functionality, then it should be able to be made relatively easy. But for now this API is fine I think.

### DIFF
--- a/embassy-nrf/src/ppi/dppi.rs
+++ b/embassy-nrf/src/ppi/dppi.rs
@@ -1,61 +1,74 @@
-use super::{Channel, Event, Ppi, Task};
+use core::marker::PhantomData;
+
+use embassy::util::Unborrow;
+use embassy_hal_common::unborrow;
+
+use super::{Channel, ConfigurableChannel, Event, Ppi, Task};
 
 const DPPI_ENABLE_BIT: u32 = 0x8000_0000;
 const DPPI_CHANNEL_MASK: u32 = 0x0000_00FF;
 
-impl<'d, C: Channel + 'd, const EVENT_COUNT: usize, const TASK_COUNT: usize>
+impl<'d, C: ConfigurableChannel> Ppi<'d, C, 1, 1> {
+    pub fn new_one_to_one(ch: impl Unborrow<Target = C> + 'd, event: Event, task: Task) -> Self {
+        Ppi::new_many_to_many(ch, [event], [task])
+    }
+}
+
+impl<'d, C: ConfigurableChannel> Ppi<'d, C, 1, 2> {
+    pub fn new_one_to_two(
+        ch: impl Unborrow<Target = C> + 'd,
+        event: Event,
+        task1: Task,
+        task2: Task,
+    ) -> Self {
+        Ppi::new_many_to_many(ch, [event], [task1, task2])
+    }
+}
+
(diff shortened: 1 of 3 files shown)


Dirbaio:
There's one issue with the current PPI API, but which is made quite worse when introducing DPPI support: it's possible to leave tasks/events registered to a channel.

For example, this thing [here](https://github.com/embassy-rs/embassy/blob/05bc4d198e558a153d78e0fa07228c73023c4b5b/embassy-nrf/src/buffered_uarte.rs#L153) is not setting the fork task. If that PPI channel has a fork task set from before, it'll keep using it.

Of course for PPI the fix is easy: just clear the channel's tasks/events in `new`. However for DPPI this is impossible, it'd require looping over ALL publish/subscribe registers in the system. And setting tasks/events will not unset the previous ones! Every single driver will have to remember to unsubscribe/unpublish everything it did on drop, or it'll leave the channel "poisoned" for the next use. This is quite a big footgun.

WIth the API I proposed above, the impls for DPPI could store in `Self` which tasks/events are registered, and unregister them on drop. `ManyToMany` has no problem storing an arbitrary amount of tasks/events thanks to the const generics. This way everything gets cleaned up on drop automatically, drivers using PPI don't have to worry about it. Also storing stuff in Self is not needed for PPI (nrf52), just clearing the channel regs in `new` is enough.

Dirbaio:
> Right, but then you'll loose the ability to have an AnyChannel implementation. Is that ok?

Why is that? all channels in DPPI can be ManyToMany. AnyChannel would just store the channel number, like before.

diondokter:
> 
> 
> > Right, but then you'll loose the ability to have an AnyChannel implementation. Is that ok?
> 
> Why is that? all channels in DPPI can be ManyToMany. AnyChannel would just store the channel number, like before.

Ah, well, I didn't examine your code well enough. Never mind :P

diondokter:
Ok, so I think this is a lot better!
Good thing you didn't accept the changes as they were.

So, what do you guys think of this?

There is one issue though... With the DPPI, any task can only be assigned to one channel. This is now not checked for.
The way that tasks and events work now means that we can only check at runtime if a task or event is already in use.
Currently that's not being checked. Any idea how that can be added?

PR #436, opened 2021-10-19
Add support for temperature sensor peripheral

lulf:
* Add TEMP peripheral to all nRF52 chips
* Add async HAL for reading temperature values
* Add example application reading temperature values


huntc:
You wouldn't have a little video showing it working, would you? Anyhow, great work. My only other comment would be that we should try and support the 9160 out of the box, because I want that one. :-)

PR #439, opened 2021-10-20
Prevent overflow in std timer driver

### LINKED ISSUE #438: Std tick example is broken due to timer overflow
Running the std `tick` example, I get the following panic which stops the timer thread:

```
thread '<unnamed>' panicked at 'overflow when adding duration to instant', library/std/src/time.rs:390:33
stack backtrace:
[2021-10-20T06:37:06.389264000Z INFO  tick] tick
   0: rust_begin_unwind
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/panicking.rs:100:14
   2: core::panicking::panic_display
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/panicking.rs:64:5
   3: core::option::expect_failed
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/option.rs:1637:5
   4: core::option::Option<T>::expect
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/option.rs:708:21
   5: <std::time::Instant as core::ops::arith::Add<core::time::Duration>>::add
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/std/src/time.rs:390:9
   6: embassy::time::driver_std::TimeDriver::alarm_thread
             at /Users/lulf/dev/embassy/embassy/src/time/driver_std.rs:90:17
   7: core::ops::function::FnOnce::call_once
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/ops/function.rs:227:5
```

From what I can tell, the issue is that 'zero' is actually not zero in the underlying OS representation, so adding u64::MAX to it will overflow. Instead, the max value added shoul be u64::MAX - zero.

lulf:
This prevents the std time driver from overflowing when setting the next
wakeup time. If an overflow occurs, default to sleeping up to 1 second.

Fixes #438

### DIFF
--- a/embassy/src/time/driver_std.rs
+++ b/embassy/src/time/driver_std.rs
@@ -90,7 +90,7 @@ impl TimeDriver {
             // Ensure we don't overflow
             let until = zero
                 .checked_add(StdDuration::from_micros(next_alarm))
-                .unwrap_or(zero + StdDuration::from_secs(1));
+                .unwrap_or_else(|| StdInstant::now() + StdDuration::from_secs(1));
 
             unsafe { DRIVER.signaler.as_ref() }.wait_until(until);
         }


lulf:
@Dirbaio I can't find a convenient way to subtract the 'zero' time as the internal representation is hidden. Instead, this PR defaults to a 1 second wait if no alarms was set or if it would overflow.

I guess we could look into rewriting the timer logic, but I'm not sure we care enough about std for that?

lulf:
@Dirbaio I see this on apple m1, mac os x. Tracked it down to the time refactoring that happened in late August.

PR #448, opened 2021-10-24
Dummy pin implementation for Saadc internal vdd sampling

jacobrosenthal:
For instance, for reading the battery input voltage on the nrf
Api ends up looking like
`let channel_config = saadc::ChannelConfig::single_ended(saadc::VddInput::default());`

I ~haven't confirmed a sane reading yet~, but this compiles so is ready for bikeshedding
Update: It looks like Ive got sane readings

### DIFF
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -78,6 +78,29 @@ pub struct ChannelConfig<'d> {
     phantom: PhantomData<&'d ()>,
 }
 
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Looks like nrf9160 has VDD_GPIO instead of VDD... 
https://github.com/nrf-rs/nrf-pacs/blob/0b198f07356963a975cb71c3c35e0fbf555e4f22/pacs/nrf9160-pac/src/saadc_ns/ch/pselp.rs#L59-L60

Dirbaio:
Thank you! :D

PR #449, opened 2021-10-24
STM32: Add PWM support

bgamari:
Here is a first-cut at implementing PWM support for STM32 targets via the TIM peripherals. Currently this only contains pin configuration for the STM32G0 but it would be straightforward to extend to other platforms.


jacobrosenthal:
If it matters my https://github.com/embassy-rs/embassy/pull/455 just merged with a SequencePwm type (which you dont need to support), but that ended up renaming pwm to SimplePwm though @huntc isnt in love with that naming. It seems whatever name is decided on to have a constant name for the duty cycle style pwms that consistent with the e-h pwm trait whatever we choose...

bgamari:
Thanks for finishing this, @Dirbaio !

PR #455, opened 2021-10-26
simple_playback api from nrf sdk

jacobrosenthal:
Port of the nrf_drv_pwm_simple_playback call from the nordic sdk that allows you to set up a sequence to play across leds with no interaction necessary using the 'shorts' registers to trigger looping sequences


jacobrosenthal:
Im not sure an the api yet but this is working and cleaned up now

jacobrosenthal:
Oh hey the hal does have some of this functionality
https://github.com/nrf-rs/nrf-hal/blob/eb4ba6ae4204c7f58fc968e57b2995df15f5ac77/examples/pwm-demo/src/main.rs#L172


https://github.com/nrf-rs/nrf-hal/blob/eb4ba6ae4204c7f58fc968e57b2995df15f5ac77/nrf-hal-common/src/pwm.rs

jacobrosenthal:
I cant figure out if/how the hal api is usable within multiple tasks. loading and starting a sequence https://github.com/nrf-rs/nrf-hal/blob/eb4ba6ae4204c7f58fc968e57b2995df15f5ac77/nrf-hal-common/src/pwm.rs#L535 requires static mut buffers and is typestated so returns PwmSeq which you can then you can .split() to get back the constituent parts. Im trying to use it across tasks so attempting threadmodemutex<refcell<option> but fighting consider borrowing the Option's content: .as_ref() because the split requires a move...

so back to my api. Not sure if/how to handle safety with the api and the buffers. But its fun and functional and Im playing with it a few places

jacobrosenthal:
More general api chatter

First. Why is this here and not in a hal and why am I bothering you all :)
The reason this should be here and not hal is because the hal version is all type statey currently. In the same way youd have an async hal version here because its not there, we should have this api here, even though it isnt colored async

Next im ready to start criticizing the upstream master implementation of pwm..

`set_duty` fn is both the way to set the duty and the ONLY way to start the task. also it emits ONE duty cycle, so its more of a one_shot. I think a lot of people are used to apis like arduino's analogwrite apis where you expect to see that duty cycle infinitely unless you disable or change it. 

Further the existing `new` is rather opinionated. It sets a refresh of 32, which is very specific and we dont expose a way to even change it after the fact. We could add setters for refresh and enddelay and more but that feels too raw as discussed. That `new` could probably be renamed to an opinionated new function like `new_with_refresh_32` or whatever the application the current new is targeting.  Then we could have a `new_with_duty_period()` for the arduino analogwrite style configuration. Note Arduino style apis immediately start the task when analogWrite is called. would we want people to in addition to new call `enable` or `start` or should `new` functions start their task in addition to configuration?

Note my playback api very much wants to be started when its configured, as only it knows internally whether it should be starting the seq0 or seq1 task.

jacobrosenthal:
So I made a separate `PwmSeq` struct and I think this is nice? My `PwmSeq` is using the sealed state of the existing `Pwm` struct, which I don't really need since I don't need the duty array. But then wed have to `impl_pwm_seq!` for all the pwm channels as well. Do we want that?

The main constraint I need to keep is I really want to take the sequence buffer one time during `new` so as to order to dodge the ownership issues. If you want to change the sequence you can drop the pwmseq instance and remake it.

Usage looks like
```
    let pwm = unwrap!(PwmSeq::new(p.PWM0, p.P0_13, NoPin, NoPin, NoPin, config));
    let _ = pwm.start(SequenceMode::times(1);
    Timer::after(Duration::from_millis(1000)).await;
    let _ = pwm.start(SequenceMode::Infinite);
    Timer::after(Duration::from_millis(20000)).await;
    pwm.stop();
```

The current sequence config is still a little comprehensive.
```
    let config = SequenceConfig {
        counter_mode: CounterMode::UpAndDown,
        top: 12000,
        prescaler: Prescaler::Div16,
        sequence: &seq_values,
        sequence_load: SequenceLoad::Common,
        refresh: 0,
        end_delay: 0,
    };
```
If this still feels too 'raw' I could see dumping end_delay and refresh from the struct. Those are just ram saving features, as you could instead stuff the filler bytes intp your sequence. And or the 4 sequence fields could go in a separate `Sequence` configuration

jacobrosenthal:
Another more c-ey option instead of the Times enum for `start` would just be `start(count)` where count= 0 is infinite, not sure

examples/nrf/src/bin/pwm.rs
```diff
@@ -7,95 +7,42 @@ mod example_common;
...
+        pwm.set_duty(0, 2500 - 218);
+        Timer::after(Duration::from_millis(5000)).await;
+
+        info!("180 deg");
+        pwm.set_duty(0, 2500 - 250);
+        Timer::after(Duration::from_millis(5000)).await;
+
+        info!("0 deg");
+        pwm.set_duty(0, 2500 - 125);
+        Timer::after(Duration::from_millis(5000)).await;
```
huntc, line 46: Wdyt about minimising the changes to this example? I thought what we had was quite straightforward, and it’d be interesting to make a direct comparison.

jacobrosenthal: Yeah I dont much care what the examples are, this isnt final. However I disagree thats a good example. Its attempting to create a tight sequence which is literally what the new api would be for, with a loose timing scheduler which is a bad idea.

The old example also cant work anymore because it was relying on a the .refresh field to be stretching out the sequence buffer which is a very advanced api. if that sin table is recreated the old example could be returned.

jacobrosenthal: https://github.com/embassy-rs/embassy/pull/455/files#diff-c1b3acd7a9372df9109e152f99037a9d9f1c6db11ec955ca1792830adb434de3L84

jacobrosenthal: I threw a commit on which restores the example by adding a new api set_time_stretch. Not sure what other drives elsewhere call it. Its the 'refresh' register on nrf. Now other commenters have said they dont want more api surface necessarily, so Ill leave that up to someone else. But having it set in the constructor with no way to change as it is in the upstream wasnt great

embassy-nrf/src/util.rs
```diff
@@ -2,14 +2,14 @@ const SRAM_LOWER: usize = 0x2000_0000;
 const SRAM_UPPER: usize = 0x3000_0000;
 
 /// Does this slice reside entirely within RAM?
-pub(crate) fn slice_in_ram(slice: &[u8]) -> bool {
+pub(crate) fn slice_in_ram<T>(slice: &[T]) -> bool {
```
Dirbaio, line 5: if you allow arbitrary types, you need `slice.len() * size_of::<T>()`

embassy-nrf/src/pwm.rs
```diff
@@ -81,31 +335,43 @@ impl<'d, T: Instance> Pwm<'d, T> {
...
 
         r.decoder.write(|w| {
             w.load().individual();
             w.mode().refresh_count()
         });
         r.mode.write(|w| w.updown().up());
-        r.prescaler.write(|w| w.prescaler().div_1());
-        r.countertop
-            .write(|w| unsafe { w.countertop().bits(32767) });
+        r.prescaler.write(|w| w.prescaler().div_16());
```
Dirbaio, line 346: why these changes?

embassy-nrf/src/pwm.rs
```diff
@@ -152,6 +423,12 @@ impl<'d, T: Instance> Pwm<'d, T> {
             .write(|w| unsafe { w.countertop().bits(duty.min(32767u16)) });
     }
 
+    /// Additional number of PWM periods spent on each duty cycle value.
```
Dirbaio, line 426: I'm not sure if this is useful for the "simple" PWM. Since the channels are always playing the same duty (that you set just once), playing them once or many times doesn't make a difference, right?

embassy-nrf/src/pwm.rs
```diff
@@ -23,20 +25,267 @@ pub enum Prescaler {
...
+    /// Must be started by calling `set_duty`
     ///
     /// # Safety
     ///
-    /// The returned API is safe unless you use `mem::forget` (or similar safe mechanisms)
-    /// on stack allocated buffers which which have been passed to [`send()`](Pwm::send)
-    /// or [`receive`](Pwm::receive).
+    /// The returned API is safe unless you use `mem::forget` (or similar safe
+    /// mechanisms) on stack allocated buffers which which have been passed to
+    /// [`send()`](Pwm::send) or [`receive`](Pwm::receive).
```
Dirbaio, line 288: Pwm::send, Pwm::receive don't exist.

embassy-nrf/src/pwm.rs
```diff
@@ -23,20 +25,267 @@ pub enum Prescaler {
...
+impl<'d, T: Instance> PwmSeq<'d, T> {
+    /// Creates the interface to a PWM Sequence interface.
+    ///
+    /// Must be started by calling `start`
+    ///
+    /// # Safety
+    ///
+    /// The returned API is safe unless you use `mem::forget` (or similar safe
+    /// mechanisms) on stack allocated buffers which which have been passed to
+    /// [`send()`](Pwm::send) or [`receive`](Pwm::receive).
```
Dirbaio, line 70: Pwm::send, Pwm::receive don't exist.

embassy-nrf/src/pwm.rs
```diff
@@ -23,20 +25,267 @@ pub enum Prescaler {
...
+
+        // if NoPin provided writes disconnected (top bit 1) 0x80000000 else
+        // writes pin number ex 13 (0x0D) which is connected (top bit 0)
+        r.psel.out[0].write(|w| unsafe { w.bits(ch0.psel_bits()) });
+        r.psel.out[1].write(|w| unsafe { w.bits(ch1.psel_bits()) });
+        r.psel.out[2].write(|w| unsafe { w.bits(ch2.psel_bits()) });
+        r.psel.out[3].write(|w| unsafe { w.bits(ch3.psel_bits()) });
+
+        // Disable all interrupts
+        r.intenset.reset();
```
Dirbaio, line 115: this is a noop (see below)

embassy-nrf/src/pwm.rs
```diff
@@ -23,20 +25,267 @@ pub enum Prescaler {
...
+    /// Up and down counter (center-aligned PWM duty cycle)
+    UpAndDown,
+}
+
+/// Interface to the PWM peripheral
 pub struct Pwm<'d, T: Instance> {
     phantom: PhantomData<&'d mut T>,
 }
 
+pub struct PwmSeq<'d, T: Instance> {
```
Dirbaio, line 57: Perhaps this could be `SequencePwm` and the old `Pwm` could be `SimplePwm`? This way it's clearer why there is two  maybe?

examples/nrf/src/bin/pwm.rs
Dirbaio: The old PWM api and example didn't need `refresh`.  It just plays a 1-sample sequence. When the sequence is done playing (which is immediately), the last duty keeps playing "forever". This is how you "set duty".

> After the last value in the sequence has been loaded and started executing, a SEQEND[n] event is generated. The PWM generation will then continue with the last loaded value.

Dirbaio:
I like the new sequence API! 

The `SequenceConfig` struct is inconsistent with other `Config` structs, that are always `non_exhaustive` and have a `Default`. In this case it's not possible because it has the borrowed buffer, but that could be moved to an arg in `new`.

embassy-nrf/src/pwm.rs
jacobrosenthal: Doesnt so much matter, as long as theyre documented, I was just looking at what nrf sdk uses as its 'defaults' and used them

examples/nrf/src/bin/pwm.rs
jacobrosenthal: Ok maybe this is the last clarification I needed. If the sequence is 1 long, it plays forever, but if the sequence is larger than 1, it stops?

jacobrosenthal: no rather. 

it always the last value in the register.. so even on a 5 times sequence it only plays the sequence five times and also leaves the last value in the register in that case

embassy-nrf/src/util.rs
jacobrosenthal: Like this? 
```
ptr >= SRAM_LOWER && (ptr + slice.len() * core::mem::size_of::<T>()) < SRAM_UPPER
```

jacobrosenthal:
@Dirbaio 
Resolved most of those comments and fixed SequenceConfig struct to be non exhaustive as well.

Still todo:
* which examples to keep
* How to allow inverting pins in the SequenceConfig
* possibly tackle the dropping and deconfiguring of pins
* possibly get rid of state.duty

embassy-nrf/src/pwm.rs
```diff
@@ -1,16 +1,305 @@
...
+            pin.conf().write(|w| w.dir().output());
+        }
+        if let Some(pin) = ch3.pin_mut() {
+            pin.set_low();
+            pin.conf().write(|w| w.dir().output());
+        }
+
+        // if NoPin provided writes disconnected (top bit 1) 0x80000000 else
+        // writes pin number ex 13 (0x0D) which is connected (top bit 0)
+        r.psel.out[0].write(|w| unsafe { w.bits(ch0.psel_bits()) });
```
jacobrosenthal, line 93: do we have to 'disconnect' pins on drop as well?

embassy-nrf/src/util.rs
Dirbaio: :+1:

embassy-nrf/src/pwm.rs
Dirbaio: not needed, writing enable=0 already makes PSELs not apply

Dirbaio: Or maybe not? nrfx does disconnect them
https://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_uarte.c#L341
https://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_pwm.c#L207

I'm not sure if it's a "just in case" thing or the hardware actually requires it...

jacobrosenthal: I threw it in

embassy-nrf/src/pwm.rs
jacobrosenthal: Another option is to have a default for SimplePwm as well, that impls default with some values. Though still pushes the question what should the default values be, the nrf sdk or just what we had here before

huntc: FWIW re. Saadc, I put Embassy's defaults back to what the nRF sdk does.

jacobrosenthal: Ah, well then ill leave this with nrf sdk's defaults as well then and resolve this.

jacobrosenthal:
So maybe I misunderstood but  it looks like seqend does fire when the sequence ENDs not when its loaded.. So maybe its ok for simplepwm duty cycle since the sequence is up to 4, but for sequencepwm its blocking until the first go through of the sequence so ill take that back out of there. But it also does make the pwm example a bit slower too since it was probably relying on that timing a little bit

I removed the unstable feature and micromath dep, and made the pwm_sequence example be kind of a mirror of the pwm example, though Im not running the sequence across all 4 channels.

I think Ill leave the last 2 features, setting pin invert and callbacks for seq end and stuff for someone else who actually needs those features to design.

So I think this is basically ready for final review now?

Dirbaio:
> So maybe I misunderstood but it looks like seqend does fire when the sequence ENDs not when its loaded..

It fires when the *last sample is done loading*. For a 1-sample seq it's essentially immediately (time it takes for DMA to load 1 sample). For a N-sample seq it's `(time it takes to play N-1 samples) + (time it takes DMA to load 1 sample)`. 

So yes, the "spin until seqend" trick only works for the "simple" pwm, for the sequence pwm it'll be spinning all the time...

embassy-nrf/src/pwm.rs
```diff
@@ -1,16 +1,296 @@
...
+}
+
+pub struct SequencePwm<'d, T: Instance> {
+    phantom: PhantomData<&'d mut T>,
+    ch0: Option<AnyPin>,
+    ch1: Option<AnyPin>,
+    ch2: Option<AnyPin>,
+    ch3: Option<AnyPin>,
+}
+
```
huntc, line 31: Is it possible to avoid the name, "Simple". Simple is definitely a matter of personal perspective and we don't have it anywhere else. May I suggest just `Pwm` and `ExtendedPwm`? Also, I feel that both of these structs should be documented. Doing so will help yield the name.

Dirbaio (approved):
Thank you! Looks great! Thanks for taking the time to remove the `STATE` thing too :) 

bors r+

embassy-nrf/src/pwm.rs
huntc: Or, `Pwm` with an optional config and sequence.

Or, `DutyPwm` and `SequencePwm`.

Dirbaio:
bors d+

@jacobrosenthal about naming, feel free to change the names to `Pwm+SequencePwm` or something else if you want.

I'm not a fan of `DutyPwm` though, since the sequence one also deals with duty values, just a sequence of them.

jacobrosenthal:
On naming I honestly don't care so, current naming was your idea ;)  if its up to me I wouldn't change it again myself, but if you all pick something ill update it.

huntc:
> On naming I honestly don't care so, current naming was your idea ;) if its up to me I wouldn't change it again myself, but if you all pick something ill update it.

Sorry to trip this up. I've got two hours of meetings coming up but then I'll put some effort into some alternate name suggestions. Thanks for entertaining this.

embassy-nrf/src/pwm.rs
huntc: How about `SingleDutyPwm`? I think this highlights the difference between it and sequencing.

Accordingly, can we change the doc comment to:

```
/// SingleDutyPwm sets a single duty cycle on up to four channels.
```

Along the above lines, `DutySequencePwm` may be more descriptive than `SequencePwm`, with its description changed to:

```
/// DutySequencePwm specifies a sequence of duty cycles on up to four channels
/// and repeat that sequence n times.
```

PR #464, opened 2021-10-30
Simpler CI

Dirbaio:
Use xtask. Advantages:
- Single job, we don't clog up GHA's free max of 10 concurrent jobs
- Should be easier to cache. Cache isn't doing anything the current way CI is setup
- Easy to run locally.


lulf:
+1 to simplify. Is see the build is pretty fast, so maybe this suggestion isn't interesting: But if you haven't already considered, there is also the possibility of generating the GHA build matrix from xtask so you get the benefit from both worlds. I just did this in drogue-device CI:

* https://github.com/drogue-iot/drogue-device/blob/main/.github/workflows/ci.yaml#L44
* https://github.com/drogue-iot/drogue-device/blob/main/xtask/src/main.rs#L58

Dirbaio:
Wow, I didn't know you could dynamically generate the matrix like that :exploding_head: 

Anyway, the problem with the matrix is it spawns a *LOT* of jobs, and free GHA has a limit of max 10 concurrent jobs. We currently have about 30 jobs. This means they get executed in "batches" of 10 instead of in parallel, which brings up the "git push to ci results" latency up a lot. 

Most of these jobs spend time building shared deps, so a single job is not that much slower. The final latency is less because no job waits due to the 10-job limit, AND one PR's jobs don't make other PRs wait.

The problem with the approach here is that running `cargo` multiple times doesn't parallelize slow stuff like building the PACs or linking the final binaries... I'm doing some hacky experiments to try to do that, if they fail I'll probably end up merging somehting like this PR because I think it's stlil better/faster than what we have now...

Dirbaio:
Closing in favor of #489

PR #465, opened 2021-11-02
Adjust for STM32U5.

bobmcwhirter:


Dirbaio:
Could you add U5 to CI?

If we have no examples, just add building `embassy-stm32` for a u5 chip, to check it builds fine

lulf (approved):
🚀🚀

bors r+

PR #469, opened 2021-11-03
Add support for STM32H723 and fix eth example.

matoushybl:


examples/stm32h7/memory.x
```diff
@@ -0,0 +1,5 @@
+MEMORY
```
Dirbaio, line 1: this memory.x is unused, the build is using the  one generated by embassy-stm32 because  it has the `memory-x` feature

Dirbaio (approved):
LGTM other than the comment! Thanks for this! :) 

bors d+

PR #482, opened 2021-11-08
Add MCO peripheral.

matoushybl:
This PR adds an abstraction over STM32 RCC feature called MCO (Microcontroller Clock Output). The clock output can bind to several clock sources and then can be scaled using a prescaler.

Given that from the embassy ecosystem the RCC is generaly invisible to the user, the MCO was implemented as a separate peripheral bound to the pin where the clock should appear.

### DIFF
--- a/embassy-stm32/src/rcc/h7/mod.rs
+++ b/embassy-stm32/src/rcc/h7/mod.rs
@@ -1,7 +1,11 @@
 use core::marker::PhantomData;
 
 use embassy::util::Unborrow;
+use embassy_hal_common::unborrow;
+use stm32_metapac::rcc::vals::{Mco1, Mco2};
 
+use crate::gpio::sealed::Pin as __GpioPin;
+use crate::gpio::Pin;
 use crate::pac::rcc::vals::Timpre;
 use crate::pac::{RCC, SYSCFG};
 use crate::peripherals;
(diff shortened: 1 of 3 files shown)


bobmcwhirter:
Needs a rebase I think but looks pretty straightforward. 

I'll look again for detail in the AM.

examples/stm32h7/src/bin/mco.rs
```diff
@@ -0,0 +1,32 @@
...
+use embedded_hal::digital::v2::OutputPin;
+use example_common::*;
+
+#[embassy::main]
+async fn main(_spawner: Spawner, p: Peripherals) {
+    info!("Hello World!");
+
+    let mut led = Output::new(p.PB14, Level::High, Speed::Low);
+
+    let _mco = Mco::new(p.PA8, Mco1Source::Hsi, McoClock::Divided(2));
```
Dirbaio, line 21: There's no owned singleton for RCC MCO1/MCO2, so you could create a second `Mco` instance with the same Mco 1/2 that would break the previously created one.

Could you add RCC_MCO1, RCC_MCO2 peripheral singletons (they're created [here](https://github.com/embassy-rs/embassy/blob/fc75e0ec573267d7b97698ea140e1533e356e012/embassy-stm32/build.rs#L19)) and make `Mco` require them?

Dirbaio (changes requested):

embassy-stm32/src/rcc/h7/mod.rs
```diff
@@ -508,6 +510,172 @@ impl<'d> Rcc<'d> {
...
+impl Default for Mco2Source {
+    fn default() -> Self {
+        Self::SysClk
+    }
+}
+
+impl McoSource for Mco2Source {
+    fn into_raw(&self) -> u8 {
+        match self {
+            Mco2Source::SysClk => 0,
```
bobmcwhirter, line 584: I also wonder if there should be an `enum` within the `rcc_h7.yaml` register?

examples/stm32h7/src/bin/mco.rs
bobmcwhirter: May be also able to create it via YAML like the VREFBUF?

https://github.com/embassy-rs/stm32-data/blob/main/data/extra/family/STM32L4.yaml

bobmcwhirter: But maybe not, since it doesn't have an address per se?

matoushybl: @Dirbaio I designed it around the singletons for the pins (or at least I hope that I haven't messed it up), I agree that this is suboptimal.

Could you please give me more pointers about defining a custom singleton? We can discuss it on matrix so that we do not clutter the discussion here.

Dirbaio: @bobmcwhirter it's not an actual peripheral so it should go in embassy instead of stm32-data. -data should have data for the "unopinionated ground truth". Adding extra singletons because it helps making a safe api is a nopinionated decision made by embassy-stm32  :) 

@matoushybl you can check the RCC version with `peripheral_versions!` and append the right singletons to the [`singletons`](https://github.com/embassy-rs/embassy/blob/fc75e0ec573267d7b97698ea140e1533e356e012/embassy-stm32/build.rs#L29) vec, like we already do for gpio, exti, dma.

embassy-stm32/src/rcc/h7/mod.rs
Dirbaio: it'd be nice, but I wouldn't require doing it *now*, we can always add it later.

Dirbaio (approved):
Very nice! Thank you :) 

bors r+

PR #483, opened 2021-11-12
Combine into one Pwm type

huntc:
The existing nRF HAL name of `SimplePwm` does not appear to correlate with existing nRF terminology. This PR converts all examples to the `SequencePwm` struct, renames it to `Pwm` and then removes `SimplePwm`.

The changes to the examples to support "simple PWM" are minimal. One example:

```rust
    config.prescaler = Prescaler::Div1;
    config.top = 32767;
    config.sequence_load = SequenceLoad::Individual;

    let mut duty = [0; 4];
    let mut pwm = unwrap!(Pwm::new(
        p.PWM0, p.P0_13, p.P0_14, p.P0_16, p.P0_15, config, &mut duty
    ));
    info!("pwm initialized!");

    let mut i = 0;
    loop {
        i += 1;
        pwm.set_duty(0, DUTY[i % 1024]);
        pwm.set_duty(1, DUTY[(i + 256) % 1024]);
        pwm.set_duty(2, DUTY[(i + 512) % 1024]);
        pwm.set_duty(3, DUTY[(i + 768) % 1024]);
        Timer::after(Duration::from_millis(3)).await;
    }
```

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -19,6 +19,7 @@ pub struct Pwm<'d, T: Instance> {
     ch1: Option<AnyPin>,
     ch2: Option<AnyPin>,
     ch3: Option<AnyPin>,
+    sequence: &'d mut [u16],
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -51,7 +52,7 @@ impl<'d, T: Instance> Pwm<'d, T> {
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch3: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         config: Config,
-        sequence: &'d [u16],
+        sequence: &'d mut [u16],
     ) -> Result<Self, Error> {
         slice_in_ram_or(sequence, Error::DMABufferNotInDataMemory)?;
 
@@ -136,6 +137,7 @@ impl<'d, T: Instance> Pwm<'d, T> {
             ch1: ch1.degrade_optional(),
             ch2: ch2.degrade_optional(),
             ch3: ch3.degrade_optional(),
+            sequence,
         })
     }
 
@@ -205,11 +207,39 @@ impl<'d, T: Instance> Pwm<'d, T> {
         r.tasks_stop.write(|w| unsafe { w.bits(0x01) });
     }
 
+    /// Enables the PWM generator.
+    #[inline(always)]
+    pub fn enable(&self) {
+        let r = T::regs();
+        r.enable.write(|w| w.enable().enabled());
+    }
+
     /// Disables the PWM generator.
     #[inline(always)]
     pub fn disable(&self) {
         let r = T::regs();
         r.enable.write(|w| w.enable().disabled());
+    }
+
+    /// Sets duty cycle (15 bit) for a PWM channel.
+    pub fn set_duty(&mut self, channel: usize, duty: u16) {
+        let r = T::regs();
+
+        self.sequence[channel] = duty & 0x7FFF;
+
+        r.seq0
+            .ptr
+            .write(|w| unsafe { w.bits(self.sequence.as_ptr() as u32) });
+
+        // defensive before seqstart
+        compiler_fence(Ordering::SeqCst);
+
+        // tasks_seqstart() doesn't exist in all svds so write its bit instead
+        r.tasks_seqstart[0].write(|w| unsafe { w.bits(1) });
+
+        // defensive wait until waveform is loaded after seqstart
+        while r.events_seqend[0].read().bits() == 0 {}
+        r.events_seqend[0].write(|w| w);
     }
 }
 
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -44,14 +44,14 @@ impl<'d, T: Instance> Pwm<'d, T> {
     /// mechanisms) on stack allocated buffers which which have been passed to
     /// [`new()`](Pwm::new).
     #[allow(unused_unsafe)]
-    pub fn new<'a>(
+    pub fn new(
         _pwm: impl Unborrow<Target = T> + 'd,
         ch0: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch1: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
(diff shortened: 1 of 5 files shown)


huntc:
Flaky test? Changes here have nothing to do with STM:

https://github.com/embassy-rs/embassy/runs/4185170450?check_suite_focus=true

Dirbaio:
Allowing modifying the buffer while DMA reads from it is unfortunately UB (#484 )

There's another downside: playing the sequence in infinite mode causes DMA to be constantly reloading the duty value from RAM even when not needed. This is "not nice" because it creates extra bus traffic which might make other stuff slower (I guess in practice it's not likely to cause issues...?). You can make it read from RAM less often, for example setting a high `refresh`, but then you have high latency for changing the duty value. The previous `set_duty` had no latency and zero bus traffic...

huntc:
> Allowing modifying the buffer while DMA reads from it is unfortunately UB (#484 )
> 
> There's another downside: playing the sequence in infinite mode causes DMA to be constantly reloading the duty value from RAM even when not needed. This is "not nice" because it creates extra bus traffic which might make other stuff slower (I guess in practice it's not likely to cause issues...?). You can make it read from RAM less often, for example setting a high `refresh`, but then you have high latency for changing the duty value. The previous `set_duty` had no latency and zero bus traffic...

I'm just trying to discern the difference between the previous `set_duty` and what is here to further understand. With the prior `set_duty`, we did this:

```rust
        // tasks_seqstart() doesn't exist in all svds so write its bit instead
        r.tasks_seqstart[0].write(|w| unsafe { w.bits(1) });

        // defensive wait until waveform is loaded after seqstart
        while r.events_seqend[0].read().bits() == 0 {}
        r.events_seqend[0].write(|w| w);
```

...with the sequence stuff, we start the sequence having declared the loop.

```rust
                r.loop_.write(|w| unsafe { w.cnt().bits(0x1) });
                r.shorts.write(|w| w.loopsdone_seqstart0().enabled());

                r.tasks_seqstart[0].write(|w| unsafe { w.bits(0x01) });
```

Is it the loop setup here that makes the difference you describe?

Dirbaio:
Yes, if you set up the sequence to loop, DMA is constantly re-reading it from RAM. That's what allows it to "magically notice" you've changed the duty value in the 1-sample sequence.

huntc:
> Yes, if you set up the sequence to loop, DMA is constantly re-reading it from RAM. That's what allows it to "magically notice" you've changed the duty value in the 1-sample sequence.

Nice to get to the bottom of that. Thanks. Easily fixed with a re-instated `set_duty` method.

embassy-nrf/src/pwm.rs
```diff
@@ -216,15 +207,43 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
         r.tasks_stop.write(|w| unsafe { w.bits(0x01) });
     }
 
+    /// Enables the PWM generator.
+    #[inline(always)]
+    pub fn enable(&self) {
+        let r = T::regs();
+        r.enable.write(|w| w.enable().enabled());
+    }
+
```
huntc, line 216: This method went missing from the later `SequencePwm`.

embassy-nrf/src/pwm.rs
```diff
@@ -216,15 +207,43 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
+        // defensive before seqstart
+        compiler_fence(Ordering::SeqCst);
+
+        // tasks_seqstart() doesn't exist in all svds so write its bit instead
+        r.tasks_seqstart[0].write(|w| unsafe { w.bits(1) });
+
+        // defensive wait until waveform is loaded after seqstart
+        while r.events_seqend[0].read().bits() == 0 {}
+        r.events_seqend[0].write(|w| w);
+    }
```
huntc, line 243: Dropped the `max_duty`/`set_max_duty`, `period`/`set_period`, and `prescaler`/`set_prescaler` methods as we now have `Config`.

huntc:
> Allowing modifying the buffer while DMA reads from it is unfortunately UB (#484 )
> 
> There's another downside: playing the sequence in infinite mode causes DMA to be constantly reloading the duty value from RAM even when not needed. This is "not nice" because it creates extra bus traffic which might make other stuff slower (I guess in practice it's not likely to cause issues...?). You can make it read from RAM less often, for example setting a high `refresh`, but then you have high latency for changing the duty value. The previous `set_duty` had no latency and zero bus traffic...

Both of these issues are now addressed. On the latter, I've migrated `set_duty` into what was formally `SequencePwm`.

embassy-nrf/src/pwm.rs
```diff
@@ -44,25 +34,25 @@ pub enum Error {
...
+    pub fn new(
         _pwm: impl Unborrow<Target = T> + 'd,
         ch0: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch1: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch3: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
-        config: SequenceConfig,
-        sequence: &'a [u16],
+        config: Config,
+        sequence: &'d mut [u16],
```
huntc, line 55: Thanks to @jacobrosenthal for the lifetime fix

examples/nrf/src/bin/pwm_led.rs
```diff
@@ -8,40 +8,33 @@ use defmt::*;
...
+        p.P0_13,
+        NoPin,
+        NoPin,
+        NoPin,
+        config,
+        &mut seq_values
+    ));
+    let _ = pwm.start(SequenceMode::Infinite);
+
+    info!("pwm started!");
```
huntc, line 35: Thanks to @jacobrosenthal for this modified example

huntc:
Any ideas why this PR is affecting the STM code? https://github.com/embassy-rs/embassy/runs/4186168394?check_suite_focus=true#step:5:276

embassy-nrf/src/pwm.rs
jacobrosenthal: sequencepwm cant have set_duty :)
The whole point of sequence pwm is you dont get to control duty after its started (unless you stop and restart it)

huntc:
Abandoning this PR having discovered that the `start` method of `SequencePwm` is mutually exclusive to the `set_duty` method of `SimplePwm`. I still have a problem with the name, `SimplePwm` and propose `SimplifiedPwm` as I think this more accurately reflects what is going on (there's arguably nothing simple here).

PR #484, opened 2021-11-12
Fix lifetime of pwmsequence buffer argument and clarity cleanups

jacobrosenthal:
Im not sure its UB to write to the sequence values struct after having started a pwm sequence.. but its against the ethos of the api  I had in mind to monkey with it while its in motion.

~Use const generics to consume any size input instead of a borrow of slice~

cc https://github.com/embassy-rs/embassy/pull/483


embassy-nrf/src/pwm.rs
```diff
@@ -24,7 +24,7 @@ pub struct SimplePwm<'d, T: Instance> {
 
 /// SequencePwm allows you to offload the updating of a sequence of duty cycles
 /// to up to four channels, as well as repeat that sequence n times.
-pub struct SequencePwm<'d, T: Instance> {
+pub struct SequencePwm<'d, T: Instance, const N: usize> {
```
Dirbaio, line 27: The generic here will caues rustc to duplicate ALL the `SequencePwm` code if you instantiate it with different lengths, which is quite likely.

Also it doesn't allow playing sequences with length decided at runtime.

I would instead make it take a `&mut [u16]`

Dirbaio (changes requested):

embassy-nrf/src/pwm.rs
```diff
@@ -55,18 +55,18 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
-    pub fn new<'a>(
+    pub fn new(
         _pwm: impl Unborrow<Target = T> + 'd,
         ch0: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch1: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch3: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         config: SequenceConfig,
-        sequence: &'a [u16],
+        sequence: [u16; N],
```
Dirbaio, line 65: This is even "more UB" than before: you're now giving DMA a pointer to this *local variable*. I guess you meant to make it a field in Pwm, but that's stilll UB because the user can move the Pwm

huntc:
As the changes here will prevent #483 from progressing, can we please debate the merits of that first?

Dirbaio:
Unfortunately It's indeed UB to let the user write to the buffer whlie DMA is reading from it, so we do need this change.

huntc:
> Unfortunately It's indeed UB to let the user write to the buffer whlie DMA is reading from it, so we do need this change.

Would it be ok for me to continue the other PR under that assumption?

embassy-nrf/src/pwm.rs
jacobrosenthal: Yeah was totally not working :)

embassy-nrf/src/pwm.rs
```diff
@@ -55,14 +55,14 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
-    pub fn new<'a>(
+    pub fn new(
         _pwm: impl Unborrow<Target = T> + 'd,
         ch0: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch1: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch3: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         config: SequenceConfig,
-        sequence: &'a [u16],
+        sequence: &mut [u16],
```
Dirbaio, line 65: it doesn't have to be `mut`, sorry. I screwed up in my last comment. :see_no_evil: 

Also, it has to live for `'d` so that it's guaranteed to be live while DMA runs.

jacobrosenthal:
Heh so I already did consume, @huntc must have altered the signature in his pr and spooked me. 

But we found the lifetime bug in this pr! :)

huntc:
> Heh so I already did consume, @huntc must have altered the signature in his pr and spooked me.
> 
> But we found the lifetime bug in this pr! :)

I don't believe that I changed the signature of `SequencePwm`.

jacobrosenthal:
no the 'a lifetime was too loose and was allowing writes... 
Damn .. 
Nice bug hunt @huntc :)

embassy-nrf/src/pwm.rs
```diff
@@ -55,14 +55,14 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
         _pwm: impl Unborrow<Target = T> + 'd,
         ch0: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch1: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch2: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         ch3: impl Unborrow<Target = impl GpioOptionalPin> + 'd,
         config: SequenceConfig,
-        sequence: &'a [u16],
+        sequence: &'d [u16],
     ) -> Result<Self, Error> {
         slice_in_ram_or(sequence, Error::DMABufferNotInDataMemory)?;
```
huntc, line 68: Nice fix. Can I apply this to my PR? Saves me a bit of effort.

huntc (approved):

jacobrosenthal:
Added a bunch of nitpicks for clarity including:
* rename top to max_duty for constancy
* dump the pwm_sequence example since its not mirroring the pwm example and is thus confusing
* turn the pwm_led example into a pwmsequence style and call it pwm_sequence instead
* disable the pwm peripheral on stop, so its disabled on drop, thus we dont need the disable fn anymore
* few other nitpicks that came up in #483 and #485

embassy-nrf/src/pwm.rs
```diff
@@ -447,17 +438,14 @@ impl<'d, T: Instance> SimplePwm<'d, T> {
 
         self.duty[channel] = duty & 0x7FFF;
 
-        r.seq0
-            .ptr
-            .write(|w| unsafe { w.bits(&self.duty as *const _ as u32) });
```
Dirbaio, line 452: you can't remove this write: self might have been moved from between it was created and now. You don't want it to use a stale pointer.

Dirbaio (approved):
Thanks!!

bors r+

embassy-nrf/src/pwm.rs
```diff
@@ -396,7 +387,7 @@ impl<'d, T: Instance> SimplePwm<'d, T> {
 
         r.seq0
             .ptr
-            .write(|w| unsafe { w.bits(&pwm.duty as *const _ as u32) });
+            .write(|w| unsafe { w.bits((&pwm.duty).as_ptr() as u32) });
```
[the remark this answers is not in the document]
jacobrosenthal: Ah yeah, that was actually the goal, to do it only once. Late night though.. 
Ill get it in a future PR some day

PR #485, opened 2021-11-12
Rename SimplePwm to SimplifiedPwm

huntc:
I feel that renaming to `SimplifiedPwm` is more accurate - PWM isn't necessarily simple.

Also, provided an improved pwm_led example from Jacob Rosenthal leveraging sequences.

Lastly, an `enabled` method was missing and a buffer was incorrectly written to a register.

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -11,9 +11,9 @@ use crate::interrupt::Interrupt;
 use crate::pac;
 use crate::util::slice_in_ram_or;
 
-/// SimplePwm is the traditional pwm interface you're probably used to, allowing
+/// SimplifiedPwm is an Arduino-style PWM interface allowing
 /// to simply set a duty cycle across up to four channels.
-pub struct SimplePwm<'d, T: Instance> {
+pub struct SimplifiedPwm<'d, T: Instance> {
     phantom: PhantomData<&'d mut T>,
     duty: [u16; 4],
     ch0: Option<AnyPin>,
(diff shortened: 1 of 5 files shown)


embassy-nrf/src/pwm.rs
```diff
@@ -216,6 +216,13 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
         r.tasks_stop.write(|w| unsafe { w.bits(0x01) });
     }
 
+    /// Enables the PWM generator.
+    #[inline(always)]
+    pub fn enable(&self) {
```
jacobrosenthal, line 221: Hrm. The fn called `enable` is again coming from that embedded hal trait for the simplepwm

The enable fn isnt necessary for sequence because the constructor and start enable, and theres no way to disable. That said, I think ill add a disable to stop fn in my pr, maybe theres some power savings there, not sure. 

I guess technically we could rename start and stop to enable and disable but Im not that excited about that, since the do very different things and should probably have different names.

huntc: > Hrm. The fn called `enable` is again coming from that embedded hal trait for the simplepwm
> 
> The enable fn isnt necessary for sequence because the constructor and start enable, and theres no way to disable. That said, I think ill add a disable to stop fn in my pr, maybe theres some power savings there, not sure.
> 
> I guess technically we could rename start and stop to enable and disable but Im not that excited about that, since the do very different things and should probably have different names.

You already had a `disable` function... are you stating that you don't need `disable` either?

jacobrosenthal: pwmsequence start and stop are disable and enable. You could argue they could be called enable and disable, but they do a good bit more than enable and disable so maybe different naming is good. but then yes its confusing it had a disable fn at all so yes I just got rid of it a few min ago

Dirbaio:
I think I like `SimplePwm` more. 

- Shorter
- It is actually "simple" as in it's "the simplest PWM you can have: just set a duty value".

huntc:
I’m sorry I couldn’t convince you all. Unfortunately stating that something is simple can deceive the reader. Stating that something is simplified is accurate and doesn’t imply it is simple. Others have written about this. I will move on. :-)

PR #486, opened 2021-11-13
Pwm ppi events

jacobrosenthal:
More PWM yak shaving. I was going to do some safe pwm ppi events stuff but I just dont think it fits this api design.. ppi is just very low level, im not sure how safe it will be in general

* first we should probably have borrows of handlers for ppi with lifetime of the peripheral?  hal does https://github.com/nrf-rs/nrf-hal/blob/eb4ba6ae4204c7f58fc968e57b2995df15f5ac77/nrf-hal-common/src/pwm.rs#L714-L716
* in general having access to tasks can put the state in some configuration the api doesnt understand anymore. for `SequencePwm` ideally id hand you back either only seq_start0 or seq_start1 because youd only use one based on if your `Times` is even or odd.. but again we only know that with this api AFTER start has been called. I dont think were ready for typestates

SO I figured why not add the pwm ppi events but make them unsafe and commit this example since I started it.

Somewhat related drop IS removing the last duty cycle from the pin correctly, but stop DOES NOT..the only thing that sets the pin back is pin.conf() as far as I can tell, so I tried to document that better and got rid of stop for the `SimplePwm` again since that doesnt need it then. However its ackward we dont have a way to unset the pwm without setting a new sequence of 0s, or dropping the peripheral

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -308,7 +308,8 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
         Task::from_reg(&r.tasks_stop)
     }
 
-    /// Stop playback. Does NOT clear the last duty cycle from the pin.
+    /// Stop playback. Disables the peripheral. Does NOT clear the last duty
+    /// cycle from the pin.
     #[inline(always)]
     pub fn stop(&self) {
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Thank you!

PR #489, opened 2021-11-15
Faster CI with cargo-batch

Dirbaio:


huntc (commented):
Did we get rid of the cargo fmt check?

Dirbaio:
it's inside `ci.sh` now. I'll also try consolidating the other jobs too.

.github/workflows/rust.yml
```diff
@@ -2,172 +2,36 @@ name: Rust
...
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-      - name: Check fmt
-        run: find -name '*.rs' | xargs rustfmt --check  --skip-children --unstable-features --edition 2018
+          curl -L -o /usr/local/bin/cargo-batch https://github.com/embassy-rs/cargo-batch/releases/download/batch-0.1.0/cargo-batch
+          chmod +x /usr/local/bin/cargo-batch
+          ./ci.sh
```
huntc, line 29: What are we pulling down here? May be a comment would help describe?

Dirbaio:
it's not really faster :'(

PR #490, opened 2021-11-16
DCMI

matoushybl:


embassy-stm32/src/dcmi.rs
```diff
@@ -0,0 +1,178 @@
...
+pub mod ll {
+    pub use super::sealed::*;
+}
+
+mod sealed {
+    use super::*;
+
+    macro_rules! pin {
+        ($name:ident) => {
+            pub trait $name: GpioPin {
```
Dirbaio, line 14: for optional pins (D8+ and the others) you'll want `: OptionalPin` instead of `: GpioPin` because `NoPin` doesn't implement `GpioPin` (and it shouldn't, as it's not a real pin).

Dirbaio (changes requested):

embassy-stm32/src/dcmi.rs
```diff
@@ -0,0 +1,393 @@
...
+            phantom: PhantomData,
+            dma,
+        }
+    }
+
+    unsafe fn on_interrupt(_: *mut ()) {
+        let ris = T::regs().ris().read();
+        if ris.err_ris() {
+            #[cfg(feature = "defmt")]
+            defmt::error!("DCMI: Error.");
```
Dirbaio, line 141: just do `error!("DCMI: Error.");`. The macros in `src/fmt.rs` will automatically compile to using `defmt`, `log` or nothing depending on what's enabled.

embassy-stm32/src/dcmi.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    unsafe fn toggle(enable: bool) {
+        T::regs().cr().modify(|r| {
+            r.set_enable(enable);
+            r.set_capture(enable);
+        })
+    }
+
+    pub async fn capture(&mut self, buffer: &mut [u32])
```
Dirbaio, line 165: This should propagate the errors from the irq above (error/overrun) to the user with a `Result`. I'm not sure what DMA will do on overrun, but it might hang forever or do other bad stuff...

Dirbaio (changes requested):

Dirbaio:
Thanks for working on this! The driver looks great, there's one possible improvement though: have you tried writing it without using PeripheralMutex?

PeripheralMutex is annoying for the end user since they have to separate allocate and pass in the `State`, and it's annoying for the driver author too, I guess you noticed that. :) We should use it only when there's absolutely no other way.

You're using the State to send the result from the irq to main thread. There's one trick to avoid that:

When the irq fires, wake the waker, and then instead of clearing the interrupt flag on ICR, *mask* it in IER. Masking prevents the irq from firing in a loop, but leaves the irq flag intact in RIS. Main thread then gets woken and can check RIS to see which irqs fired. You're using the RIS register itself to pass info to main thread instead of having to make your own `isr_result` variable. The only thing you need to share between main and irq is the waker, which you can do with AtomicWaker without any mutexing.

You can see an example of that in the RNG driver [here](https://github.com/embassy-rs/embassy/blob/4eb9f0f3aebfa7b1d72830da2241b2a8b7ca139f/embassy-stm32/src/rng.rs#L163). (but please use the owned interrupts stuff to register the irq instead, like [this](https://github.com/embassy-rs/embassy/blob/4eb9f0f3aebfa7b1d72830da2241b2a8b7ca139f/embassy-stm32/src/i2c/v2.rs#L101)).

embassy-stm32/src/dcmi.rs
```diff
@@ -0,0 +1,465 @@
...
+        })
+    }
+
+    pub async fn capture(&mut self, buffer: &mut [u32]) -> Result<(), Error> {
+        self.inner.with(|state| {
+            let channel = &mut state.dma;
+
+            let r = state.peri.regs();
+            let src = r.dr().ptr() as *mut u32;
+            unsafe { channel.start_read(channel.request(), src, buffer) };
```
Dirbaio, line 216: Here you're starting the DMA, but never stop it. The user can start a capture and then cancel the operation (by dropping the future for this `async fn` before it completes). This releases the borrow on `buffer`, so the user can then deallocate it. However DMA has not been stopped, so it'll keep writing to it, overwriting whatever and causing UB.

If you use the safe wrapper `dma::read()` instead, it returns a future that you can await, and that *stops the DMA when dropped*. That'll cause DMA to stop automatically if the user drops the future for the `async fn capture()`, preventing the UB (because dropping a future drops all its contents recursively).

Dirbaio (approved):
Thank you! Looks much simpler without PeripheralMutex :) 

bors r+

matoushybl:
bors try

Dirbaio:
bors d+

PR #491, opened 2021-11-16
Fix interrupt_take macro by specifying path to panic macro.

matoushybl:


Dirbaio:
I don't think the change is correct. The macro's output gets compiled in the caller's context, so `crate` there means the user's crate, not the `embassy-macro` or `embassy` crates.

What's the issue you're seeing with it?

matoushybl:
The issue is with the STM32 examples, when I add `use example_common::*;`, It includes defmt macros without defmt prefix and then when I use `interrupt_take!` it fails with ambiguous name. Or at least that is my understanding of the issue.

```
error[E0659]: `panic` is ambiguous (glob import vs any other name from outer scope during import/macro resolution)
  --> src/bin/camera.rs:29:19
   |
29 |     let i2c_irq = interrupt_take!(I2C1_EV);
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^ ambiguous name
   |
   = note: `panic` could refer to a macro from prelude
note: `panic` could also refer to the macro imported here
  --> src/bin/camera.rs:18:5
   |
18 | use example_common::*;
   |     ^^^^^^^^^^^^^^^^^
   = help: consider adding an explicit import of `panic` to disambiguate
   = help: or use `crate::panic` to refer to this macro unambiguously
   = note: this error originates in the macro `interrupt_take` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0659]: `panic` is ambiguous (glob import vs any other name from outer scope during import/macro resolution)
  --> src/bin/camera.rs:53:20
   |
53 |     let dcmi_irq = interrupt_take!(DCMI);
   |                    ^^^^^^^^^^^^^^^^^^^^^ ambiguous name
   |
   = note: `panic` could refer to a macro from prelude
note: `panic` could also refer to the macro imported here
  --> src/bin/camera.rs:18:5
   |
18 | use example_common::*;
   |     ^^^^^^^^^^^^^^^^^
   = help: consider adding an explicit import of `panic` to disambiguate
   = help: or use `crate::panic` to refer to this macro unambiguously
   = note: this error originates in the macro `interrupt_take` (in Nightly builds, run with -Z macro-backtrace for more info)
```
Would a better fix be to add feature gate for defmt and other panic implementation?


Thanks for the review, @Dirbaio!

embassy-macros/src/lib.rs
```diff
@@ -250,7 +250,11 @@ pub fn interrupt_take(item: TokenStream) -> TokenStream {
             static TAKEN: ::embassy::export::atomic::AtomicBool = ::embassy::export::atomic::AtomicBool::new(false);
 
             if TAKEN.compare_exchange(false, true, ::embassy::export::atomic::Ordering::AcqRel, ::embassy::export::atomic::Ordering::Acquire).is_err() {
-                panic!("IRQ Already taken");
+                #[cfg(feature = "defmt")]
+                defmt::panic!("IRQ Already taken");
```
Dirbaio, line 254: the `cfg`s inside `quote!()` get included in the macro output literally, so they're checking for the `defmt` feature in the user's crate, not embassy-macros.  You should check them outside.

One problem is it'll emit code using `defmt::panic`, so `defmt` has to be in scope in the user's crate. This will break if the user enables `defmt` in embassy but doesn't depend on `defmt` themself. This should be rare though, so maybe it's still worth it...?

Dirbaio (changes requested):

embassy-macros/src/lib.rs
matoushybl: Closing now, this was resolved and fixed according to the discussion on Matrix - always use `core::panic`.

Dirbaio (approved):
Thank you!

bors r+

PR #498, opened 2021-11-22
Makes our various mutexes available to Signal

huntc:
Embassy provides various types of Mutex e.g. Noop. This PR allows the mutex to be provided as per what we do with the MPSC. A major benefit is that the previous situation of needing a critical section can often be avoided.

A consequence of the change is that the `new` function is no longer const. This means that, as per the MPSC, `Forever` needs to be utilised e.g.:

```rust
static SOME_SIGNAL: Forever<Signal<Noop, SomeCommand>> = Forever::new();

// Then, during initialization...
let some_signal = SOME_SIGNAL.put(Signal::new());
```

### DIFF
--- a/embassy/src/channel/signal.rs
+++ b/embassy/src/channel/signal.rs
@@ -1,3 +1,5 @@
+use crate::blocking_mutex::kind::MutexKind;
+use crate::blocking_mutex::Mutex;
 use core::cell::UnsafeCell;
 use core::future::Future;
 use core::mem;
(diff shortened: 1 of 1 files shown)


lulf (approved):
LGTM, but as to the wl55 build failure, I think using a CriticalSection mutex for that case should be correct.

hydra:
This might need rebasing now that https://github.com/embassy-rs/embassy/pull/792 is merged as it uses a `Signal`.

huntc:
Just writing up here what became the blocker: Functions in traits cannot be declared const...

The [RFC states](https://github.com/rust-lang/rfcs/blob/master/text/0911-const-fn.md):

> Traits, trait implementations and their methods cannot be const - this allows us to properly design a constness/CTFE system that interacts well with traits - for more details, see Alternatives.

I'm unsure that there's a way forward here.

Dirbaio:
Done in #961

PR #500, opened 2021-11-23
Low level DMA channel API.

matoushybl:
This should be an improved version of the PR by @theunkn0wn1.

### DIFF
--- a/embassy-stm32/src/dma/bdma.rs
+++ b/embassy-stm32/src/dma/bdma.rs
@@ -1,19 +1,28 @@
 #![macro_use]
 
-use core::future::Future;
 use core::sync::atomic::{fence, Ordering};
-use core::task::{Poll, Waker};
+use core::task::Waker;
 
 use embassy::interrupt::{Interrupt, InterruptExt};
 use embassy::waitqueue::AtomicWaker;
-use embassy_hal_common::drop::OnDrop;
-use futures::future::poll_fn;
 
-use crate::dma::{Channel, Request};
+use crate::dma::Request;
 use crate::interrupt;
 use crate::pac;
 use crate::pac::bdma::vals;
 use crate::rcc::sealed::RccPeripheral;
+
+use super::{Word, WordSize};
+
+impl From<WordSize> for vals::Size {
+    fn from(raw: WordSize) -> Self {
+        match raw {
+            WordSize::OneByte => Self::BITS8,
+            WordSize::TwoBytes => Self::BITS16,
+            WordSize::FourBytes => Self::BITS32,
+        }
+    }
+}
 
 const CH_COUNT: usize = pac::peripheral_count!(bdma) * 8;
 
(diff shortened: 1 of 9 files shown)


embassy-stm32/src/dma/mod.rs
```diff
@@ -29,27 +30,91 @@ pub trait Channel: sealed::Channel {
     where
         Self: 'a;
 
-    fn read<'a>(
+    type CompletionFuture<'a>: Future<Output = ()> + 'a
+    where
+        Self: 'a;
+
+    fn read_u8<'a>(
```
Dirbaio, line 37: the futures-based methods (`read_uX`, `write_uX`, `wait_for_completion`) can be implemented on top of the non-futures ones (`start_xxx`, `is_stopped`, `set_waker`). 

Could you move them out of the trait? Making them standalone funcs instead, like  `async fn read_u8(ch: &mut impl Channel,  ...)`. This will make the trait much smaller, simplifying the dma/bdma impls.

embassy-stm32/src/dma/mod.rs
```diff
@@ -29,27 +30,91 @@ pub trait Channel: sealed::Channel {
...
         &'a mut self,
         request: Request,
         word: &u8,
         num: usize,
-        dst: *mut u8,
+        reg_addr: *mut u32,
     ) -> Self::WriteFuture<'a>;
+
+    /// Starts this channel for writing a stream of bytes.
+    fn start_write_u8<'a>(&'a mut self, request: Request, buf: &'a [u8], reg_addr: *mut u32);
```
Dirbaio, line 88: The `'a` is not necessary in these methods.

embassy-stm32/src/dma/mod.rs
```diff
@@ -29,27 +30,91 @@ pub trait Channel: sealed::Channel {
...
+
+    /// Starts this channel for writing a stream of half-words.
+    fn start_write_u16<'a>(&'a mut self, request: Request, buf: &'a [u16], reg_addr: *mut u32);
```
Dirbaio, line 91: Maybe this could be generic over a `Word` trait impl'd for `u8, u16, u32`, so that there's just `start_read<W>, start_write<W>`. This would remove the need for the ugly macros in the impl side.

embassy-stm32/src/dma/mod.rs
```diff
@@ -29,27 +30,91 @@ pub trait Channel: sealed::Channel {
...
+    /// Stops this channel.
+    fn stop<'a>(&'a mut self);
+
+    /// Returns whether this channel is active or stopped.
+    fn is_stopped<'a>(&self) -> bool;
+
+    /// Returns the total number of remaining transfers.
+    fn remaining_transfers<'a>(&'a mut self) -> u16;
+
+    /// Sets the waker that is called when this channel completes/
```
Dirbaio, line 114: `/` -> `.`

embassy-stm32/src/dma/mod.rs
```diff
@@ -29,27 +30,91 @@ pub trait Channel: sealed::Channel {
...
     ) -> Self::WriteFuture<'a>;
+
+    /// Starts this channel for writing a stream of bytes.
+    fn start_write_u8<'a>(&'a mut self, request: Request, buf: &'a [u8], reg_addr: *mut u32);
+
+    /// Starts this channel for writing a stream of half-words.
+    fn start_write_u16<'a>(&'a mut self, request: Request, buf: &'a [u16], reg_addr: *mut u32);
+
+    /// Starts this channel for writing a stream of words.
+    fn start_write_u32<'a>(&'a mut self, request: Request, buf: &'a [u32], reg_addr: *mut u32);
```
Dirbaio, line 94: `start_read, start_write` should be `unsafe`, since they require the buffer is alive for the entire duration of the transfer, not just the duration of the function call. This requirement should be documented.

embassy-stm32/src/dma/bdma.rs
```diff
@@ -257,3 +253,162 @@ pac::interrupts! {
...
+    }
+
+    pub unsafe fn stop(dma: pac::bdma::Dma, ch: u8) {
+        let ch = dma.ch(ch as _);
+
+        // Disable the channel and interrupts with the default value.
+        ch.cr().write(|_| ());
+
+        // Wait for the transfer to complete when it was ongoing.
+        while ch.cr().read().en() {}
```
Dirbaio, line 361: Instead of spinning, it might be better to simply make `stop` request the stop, but not wait until actually stopped. (documenting it in the Channel trait method, and maybe renaming `stop` to `request_stop` or something else?)

The advantages are:

- User code decides whether to spin or not (for example using async instead. Stopping is probably very fast though, so using async is probably not worth it)
- It's possible to do stuff like stopping 2 channels concurrently: request both to stop, and *then* spin waiting for both to stop.

Dirbaio (changes requested):
I like the dirsction this is going, the new API is definitely way more powerful! I've commented a few things, mostly stuff that could be simplified.

matoushybl:
bors try

Dirbaio:
bors try

embassy-stm32/src/dma/bdma.rs
```diff
@@ -164,88 +87,87 @@ pub(crate) unsafe fn init() {
...
-                        #[cfg(dmamux)]
-                        <Self as super::dmamux::sealed::MuxChannel>::DMAMUX_REGS,
-                        #[cfg(dmamux)]
-                        <Self as super::dmamux::sealed::MuxChannel>::DMAMUX_CH_NUM,
-                    )
-                }
+        impl crate::dma::sealed::Channel for crate::peripherals::$channel_peri {
+
+            unsafe fn start_write<W: Word>(&mut self, request: Request, buf: &[W], reg_addr: *mut u32) {
+                low_level_api::reset_status(crate::pac::$dma_peri, $channel_num);
```
Dirbaio, line 93: `reset_status` could be done by `start_trasnfer`, so that there's less code inside the macro

Dirbaio (changes requested):
Refactor is looking great! A few more nits, mostly docs.

embassy-stm32/src/dma/bdma.rs
```diff
@@ -257,3 +179,91 @@ pac::interrupts! {
...
+        ch.cr().read().en()
+    }
+
+    /// Gets the total remaining transfers for the channel
+    /// Note: this will be zero for transfers that completed without cancellation.
+    pub unsafe fn get_remaining_transfers(dma: pac::bdma::Dma, ch: u8) -> u16 {
+        // get a handle on the channel itself
+        let ch = dma.ch(ch as _);
+        // read the remaining transfer count. If this is zero, the transfer completed fully.
+        ch.ndtr().read().ndt()
```
Dirbaio, line 254: This is inverted

embassy-stm32/src/dma/bdma.rs
```diff
@@ -257,3 +179,91 @@ pac::interrupts! {
...
+        // Disable the channel and interrupts with the default value.
+        ch.cr().write(|_| ());
+
+        // "Subsequent reads and writes cannot be moved ahead of preceding reads."
+        fence(Ordering::SeqCst);
+    }
+
+    pub unsafe fn is_stopped(dma: pac::bdma::Dma, ch: u8) -> bool {
+        let ch = dma.ch(ch as _);
+        ch.cr().read().en()
```
Dirbaio, line 245: This is inverted

Dirbaio (approved):
bors r+

PR #506, opened 2021-11-28
Clock cleaning

lulf:
Different STM32 RCC peripherals have different capabilities and register values. Define types for each RCC types inside each module to ensure full range of capabilities for each family can be used


Dirbaio (approved):
Sounds good! Thank you!

bors r+

PR #508, opened 2021-11-29
STM32 timers + pwm improvements

matoushybl:


matoushybl:
Hi, @Dirbaio, I've redone the whole thing, which should hopefully allow writing both in-tree drivers as well as out-of-tree ones with a low-level api. Could you please have a look?

embassy-stm32/src/timer/basic_timer.rs
```diff
@@ -0,0 +1,96 @@
...
+    timer: T,
+    phantom: PhantomData<&'d T>,
+    update_waker: WakerRegistration,
+    ready: bool,
+}
+
+unsafe impl<'d, T: super::Basic16bitInstance> Send for StateInner<'d, T> {}
+unsafe impl<'d, T: super::Basic16bitInstance> Sync for StateInner<'d, T> {}
+
+impl<'d, T: super::Basic16bitInstance> PeripheralState for StateInner<'d, T>
```
Dirbaio, line 16: It'd be nice to not use PeripheralMutex, storing the `WakerRegistration` in a `static` instead.

Dirbaio:
LGTM on the `Instance` trait classfications, we'll finally have them right! :D

The only nit is the `Timer` API would be more ergonomic without PeripheralMutex, same reason as the DCMI support you wrote a while back :)

Dirbaio (approved):
Thank you!

bors r+

Dirbaio:
oops needs some rebase. @matoushybl feel free to merge yourself :+1:

PR #509, opened 2021-11-30
Remove unsafe from nRF uarte and improve doco with rationale

huntc:
The constructors themselves are not strictly unsafe. Interactions with DMA can be generally unsafe if a future is dropped, but that's a separate issue. It is important that we use the `unsafe` keyword diligently as it can lead to confusion otherwise.

I've also provided some rationale re. the usage of [Uarte] vs [BufferedUarte].


Dirbaio (approved):
thank you! 

bors r+

PR #525, opened 2021-12-08
Applies the Uarte patch

huntc:
Applies the Nordic workaround found in the `Uarte` for the nRF9160 and nRF5340 to the `BufferedUarte`.

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -22,7 +22,7 @@ use crate::pac;
 use crate::ppi::{AnyConfigurableChannel, ConfigurableChannel, Event, Ppi, Task};
 use crate::timer::Instance as TimerInstance;
 use crate::timer::{Frequency, Timer};
-use crate::uarte::{Config, Instance as UarteInstance};
+use crate::uarte::{apply_workaround_for_enable_anomaly, Config, Instance as UarteInstance};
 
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/buffered_uarte.rs
```diff
@@ -182,6 +183,61 @@ impl<'d, U: UarteInstance, T: TimerInstance> BufferedUarte<'d, U, T> {
         }
     }
 
+    #[cfg(not(any(feature = "_nrf9160", feature = "nrf5340")))]
+    fn apply_workaround_for_enable_anomaly() {
+        // Do nothing
+    }
+
+    #[cfg(any(feature = "_nrf9160", feature = "nrf5340"))]
+    fn apply_workaround_for_enable_anomaly() {
```
Dirbaio, line 192: Could you keep this fn in one place instead of duplicating it? Maybe a standalone fn in the `uarte` mod that takes the regblock as an argument, so it can be used by both.
Thanks!

Dirbaio (changes requested):

embassy-nrf/src/buffered_uarte.rs
huntc: ah yes - wilco

Dirbaio (approved):
Thank you!

bors r+

Dirbaio:
bors d+

PR #527, opened 2021-12-08
stm32/tests: add uart, uart_dma tests.

Dirbaio:
H7 usart_dma fails with a super strange error, looks like defmt corruption. Will investigate.


Dirbaio:
oh, it seems it works.

bors r+

PR #531, opened 2021-12-08
Rust Analyzer documentation

huntc:
Intends to clarify how to make Embassy usable from within Visual Studio Code by assuming the presence of the Rust Analyzer.

Note that IntelliJ does not support the Rust Analyzer, so an open question remains as to the whether this is a good move. If it turns out not to be a good move then we should still look at clarifying the use of a workspace file.


huntc:
@lulf @bobmcwhirter I understand that you both use IntelliJ. Please comment on the impact of this PR.

bobmcwhirter:
Not Ulf just me. I’ll check in the US AM

bobmcwhirter:
I can't see that it would affect IntelliJ?

huntc:
> 

I'm assuming that IntelliJ seeks to understand the workspace level cargo.toml file. This disappears as part of the PR. A quick test would be to remove your cargo.toml and see what happens. :-)

huntc:
Cursory testing in IntelliJ appears to be fine.

I loaded Embassy up into IntelliJ along with my PR change (removed workspace cargo.toml), and it "appears" fine. I was able to do basic symbol resolution in the nrf examples, and also an stm example. I might go as far as saying that the IntelliJ integration looks a little nicer than RA given that I can have multiple targets being resolved. IntelliJ prompts me to select an appropriate Cargo.toml when it struggles for a given source (like an example).

Dirbaio (approved):
LGTM! :+1: 

bors r+

PR #533, opened 2021-12-09
Book poc

lulf:
This is a Proof of Concept for an embassy book. It's using Antora/Asciidoc.

* Asciidoc because it's a single specification with a slightly richer feature set than markdown. 
* Antora because it allows keeping content in the embassy repo, while book definition in another repo (embassy-book). 

Using antora also allows for easy embedding of embassy doc in other projects, which I think in turn increases probability of upstream contributions.

The sources of content are located in docs/ but could also be in a separate repo. However, keeping it in the embassy repo makes it easier to support one version of the book per embassy version in the future.

At present, the book is automatically built every hour from this branch and published at: https://embassy-rs.github.io/embassy-book/embassy/dev/index.html


huntc:
I think this is an excellent initiative!

Before we embark on docs though, I think it’d be useful to identify who this book is for. Obviously ref doc is catered for already (although we need to add more to the overall Rustdoc).

My take on this book is that it should “sell” Embassy to both

* newbie embedded developers
* newbie Rust developers coming from embedded C

lulf:
@huntc I agree on "selling" to people coming from the embedded world, but I think that also includes a good ref doc as well. I'm thinking of explaining the async model, the executor runtime, as well as the HALs in some higher level way. The Hubris ref doc is a good example of what I'd like to see here https://hubris.oxide.computer/reference along with some nice figures.

Dirbaio:
Can Antora/Asciidoc compile Rust snippets in CI to test they're up to date? For example, if we break some public API, CI should fail until snippets are updated.

lulf:
> > @huntc I agree on "selling" to people coming from the embedded world, but I think that also includes a good ref doc as well. I'm thinking of explaining the async model, the executor runtime, as well as the HALs in some higher level way. The Hubris ref doc is a good example of what I'd like to see here https://hubris.oxide.computer/reference along with some nice figures.
> 
> Great, but who is the target audience specifically? :-)

@huntc I'd like just some docs, okay! :) So, here is who I have in mind and who I'm motivated to write for personally:

* The curious developer who'd like to see what all this async fuss is about
* The company who is looking to "modernise" their application and evaluating options
* The developer wanting  to understand design choices of embassy and how the different parts fits together
* The developer wanting to get started with embassy on <platform>

Not sure if you're looking for anything more specific than those personas, but I think we're in a better spot if we have the above.

lulf:
> Can Antora/Asciidoc compile Rust snippets in CI to test they're up to date? For example, if we break some public API, CI should fail until snippets are updated.

The way I've done this in the past is to put these snippets in a place where CI can compile them as part of regular build, and then use the asciidoc includes to put them in the doc. I'll add a poc of this here as well today.

lulf:
@Dirbaio Inclusion example: 

* https://embassy-rs.github.io/embassy-book/embassy/dev/examples.html
* source: https://raw.githubusercontent.com/embassy-rs/embassy/book-poc/docs/modules/ROOT/pages/examples.adoc

To include snippets, I'd go about it this way: put snippets in modules that can be compiled as part of a cargo project and hook it into ci.sh. Then include them in docs using the same mechanism.

Would that work for you?

Dirbaio:
> Depending on line numbers for the snippets is a little brittle

Yeah that's my concern as well.. mdbook can compile inline snippets, see [this](https://github.com/knurling-rs/defmt/blob/main/book/src/format.md?plain=1#L7-L26) for example. It works like rustdoc, where the entire snippet has to compile but you can "hide" parts with `#`.

It's not a blocker IMO though. We can go ahead with antora, if it proves to be unmaintainable then we may consider switching.

Dirbaio:
Do you want to script deploying to `embassy.dev`? :D

I've added a secret `WEBSITE_SSH_KEY` that can rsync stuff to `embassy.dev`. Like this for example: `rsync index.html embassy@embassy.dev:index.html`.

We can deploy the book to the root, so the "frontpage" of the book can serve as a "home page".

Dirbaio (approved):
bors r+

lulf:
> Do you want to script deploying to `embassy.dev`? :D
> 
> I've added a secret `WEBSITE_SSH_KEY` that can rsync stuff to `embassy.dev`. Like this for example: `rsync index.html embassy@embassy.dev:index.html`.
> 
> We can deploy the book to the root, so the "frontpage" of the book can serve as a "home page".

Given the lack of content on embassy.dev, maybe we can do that for now? But ultimately my "recommendation" would perhaps that we have some kind of landing page that could link to the book at book.embassy.dev for instance. Antora is great for books, but not so great for writing a fine tuned web page.

Issue #536, opened 2021-12-10
Unreliable BufferedUarte read_exact calls


huntc:
I'm having an issue receiving 18 bytes via `BufferedUarte` where not all of the bytes transmitted to it by another device are being received. I have used a scope to confirm that the other device is sending the 18 bytes.

I think it is to do with both the BufferedUarte and also how I read from it in my program. I read in two parts. First, we use `read_exact` to read a 5-byte header with the 5th byte indicating the remaining payload length. Then we `read_exact` the remainder payload.

If I change things to just one `read_exact` of 18 bytes instead of the two calls outlined above then all appears well.

At first, I thought my program might be exceeding the idle time logic (20 bit/baudrate timeout implemented via PPI), but disabling that and ensuring that the rx buffer was sized exactly to 18 bytes did not appear to help.

Another observation I made is that setting the size of the receive buffer to the exact size expected to be received makes the problem surface faster. If the buffer is large, then you might not see problems for a while.

huntc:
I was able to reproduce this given a modification to the examples/nrf/buffered_uart.rs.

```rust
#![no_std]
#![no_main]
#![feature(type_alias_impl_trait)]

#[path = "../example_common.rs"]
mod example_common;

use embassy::executor::Spawner;
use embassy::io::{AsyncBufReadExt, AsyncWriteExt};
use embassy_nrf::buffered_uarte::State;
use embassy_nrf::gpio::NoPin;
use embassy_nrf::{buffered_uarte::BufferedUarte, interrupt, uarte, Peripherals};
use example_common::*;
use futures::pin_mut;

#[embassy::main]
async fn main(_spawner: Spawner, p: Peripherals) {
    let mut config = uarte::Config::default();
    config.parity = uarte::Parity::EXCLUDED;
    config.baudrate = uarte::Baudrate::BAUD115200;

    let mut tx_buffer = [0u8; 18];
    let mut rx_buffer = [0u8; 18]; // NOTE if I set this to a larger size then the problem is deferred!!!

    let irq = interrupt::take!(UARTE0_UART0);
    let mut state = State::new();
    let u = BufferedUarte::new(
        &mut state,
        p.UARTE0,
        p.TIMER0,
        p.PPI_CH0,
        p.PPI_CH1,
        irq,
        p.P0_08,
        p.P0_06,
        NoPin,
        NoPin,
        config,
        &mut rx_buffer,
        &mut tx_buffer,
    );
    pin_mut!(u);

[...]

PR #537, opened 2021-12-12
Documents the nRF BufferedUarte problem

huntc:
Please see https://github.com/embassy-rs/embassy/issues/536 for the rationale.


Dirbaio (approved):
Thank you! I agree we should do this, until (if?) we find a way to make BufferedUart work better without RTS/CTS...

bors r+

PR #543, opened 2021-12-14
Incrementally merge STM32 SPI versions, Part 3

GrantM11235:
Notable changes:
- `SPE` is now disabled before `TXDMAEN` and `RXDMAEN` are disabled. This is the "mandatory" sequence for v2 and v3 (and maybe v1 as well, but I can't find it in the reference manual).
- v1's `write_dma_u8` now waits for idle and disables `TXDMAEN` after the transfer is complete, just like everything else.


Dirbaio (approved):
LGTM, thank you! :) 

bors r+

PR #544, opened 2021-12-15
Introduces split on the nRF Uarte

huntc:
A new `split` method is introduced such that the Uarte tx and rx can be used from separate tasks. An MPSC is used in an example to illustrate how data may be passed between these tasks.

The approach taken within the `Uarte` struct is to split into tx and rx fields on calling `Uarte::new`. These fields are returned given a call to `Uarte::split`, but otherwise, if that call isn't made, then the API remains as it was before.

Here's a snippet from a new example introduced:

```rust
#[embassy::main]
async fn main(spawner: Spawner, p: Peripherals) {
    // ...

    let uart = uarte::Uarte::new(p.UARTE0, irq, p.P0_08, p.P0_06, NoPin, NoPin, config);
    let (mut tx, rx) = uart.split();

    // ...

    // Spawn a task responsible purely for reading

    unwrap!(spawner.spawn(reader(rx, s)));

    // ...

    // Continue reading in this main task and write
    // back out the buffer we receive from the read
    // task.
    loop {
        if let Some(buf) = r.recv().await {
            info!("writing...");
            unwrap!(tx.write(&buf).await);
        }
    }
}

#[embassy::task]
async fn reader(mut rx: UarteRx<'static, UARTE0>, s: Sender<'static, Noop, [u8; 8], 1>) {
    let mut buf = [0; 8];
    loop {
        info!("reading...");
        unwrap!(rx.read(&mut buf).await);
        unwrap!(s.send(buf).await);
    }
}
```

### DIFF
--- a/embassy-nrf/src/uarte.rs
+++ b/embassy-nrf/src/uarte.rs
@@ -188,60 +188,6 @@ impl<'d, T: Instance> UarteTx<'d, T> {
     }
 }
 
-impl<'d, T: Instance> Read for UarteTx<'d, T> {
-    #[rustfmt::skip]
-    type ReadFuture<'a> where Self: 'a = impl Future<Output = Result<(), TraitError>> + 'a;
-
-    fn read<'a>(&'a mut self, rx_buffer: &'a mut [u8]) -> Self::ReadFuture<'a> {
-        async move {
-            let ptr = rx_buffer.as_ptr();
-            let len = rx_buffer.len();
-            assert!(len <= EASY_DMA_SIZE);
-
-            let r = T::regs();
-            let s = T::state();
-
-            let drop = OnDrop::new(move || {
-                info!("read drop: stopping");
-
-                r.intenclr.write(|w| w.endrx().clear());
-                r.events_rxto.reset();
-                r.tasks_stoprx.write(|w| unsafe { w.bits(1) });
-
-                while r.events_endrx.read().bits() == 0 {}
-
-                info!("read drop: stopped");
-            });
-
-            r.rxd.ptr.write(|w| unsafe { w.ptr().bits(ptr as u32) });
-            r.rxd.maxcnt.write(|w| unsafe { w.maxcnt().bits(len as _) });
-
-            r.events_endrx.reset();
-            r.intenset.write(|w| w.endrx().set());
-
-            compiler_fence(Ordering::SeqCst);
-
-            trace!("startrx");
-            r.tasks_startrx.write(|w| unsafe { w.bits(1) });
-
-            poll_fn(|cx| {
-                s.endrx_waker.register(cx.waker());
-                if r.events_endrx.read().bits() != 0 {
-                    return Poll::Ready(());
-                }
-                Poll::Pending
-            })
-            .await;
-
-            compiler_fence(Ordering::SeqCst);
-            r.events_rxstarted.reset();
-            drop.defuse();
-
-            Ok(())
-        }
-    }
-}
-
 impl<'d, T: Instance> Write for UarteTx<'d, T> {
     #[rustfmt::skip]
     type WriteFuture<'a> where Self: 'a = impl Future<Output = Result<(), TraitError>> + 'a;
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/uarte.rs
```diff
@@ -275,6 +430,31 @@ impl<'d, T: Instance> Write for Uarte<'d, T> {
...
+        // Finally we can disable, and we do so for the peripheral
+        // i.e. not just rx concerns.
+        r.enable.write(|w| w.enable().disabled());
+
+        gpio::deconfigure_pin(r.psel.rxd.read().bits());
+        gpio::deconfigure_pin(r.psel.txd.read().bits());
+        gpio::deconfigure_pin(r.psel.rts.read().bits());
+        gpio::deconfigure_pin(r.psel.cts.read().bits());
+
+        info!("uarte drop: done");
```
huntc, line 454: This bit of peripheral dropping is the only bit that made me think... is it the responsibility of the tx or the rx to do this? I'm not sure that it matters.

lulf (commented):
I think it looks awesome, 🚀 🚀 🚀 

Regarding peripheral drop: I think the alternative to having RX deconfigure pins and disable the peripheral would be to have some shared per Instance refcount that Tx and Rx drop would decrement, and last decrement would disable the peripheral.

If you hand Tx and Rx around, I suspect drop order could change right?

huntc:
> If you hand Tx and Rx around, I suspect drop order could change right?

Thanks for the review!

Drop order could indeed change. We could do the ref count thing - good thinking. I’m still unsure whether it’s a problem for us though or not. Disabling the peripheral shouldn’t blow things up, but may be it will! Hmm.

Dirbaio:
We do need refcounting. Otherwise if the user drops rx then tries to keep using tx, it'll break.

huntc:
Refcounting is now implemented by commit 2493699fb339b835ad0f8ac403355eae468b73b1 and shown to work by removing the loops out of the example. Note the `uarte tx and rx drop: done` message.

```
0 INFO  uarte initialized!
└─ uart_split::__embassy_main::task::{generator#0} @ src/bin/uart_split.rs:34
1 TRACE starttx
└─ embassy_nrf::uarte::{impl#6}::write::{generator#0}::{closure#0} @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:112
2 INFO  reading...
└─ uart_split::reader::task::{generator#0} @ src/bin/uart_split.rs:64
3 TRACE startrx
└─ embassy_nrf::uarte::{impl#9}::read::{generator#0}::{closure#0} @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:112
4 INFO  wrote hello in uart!
└─ uart_split::__embassy_main::task::{generator#0} @ src/bin/uart_split.rs:46
5 INFO  uarte rx drop
└─ embassy_nrf::uarte::{impl#11}::drop @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:138
6 INFO  did_stoprx false
└─ embassy_nrf::uarte::{impl#11}::drop @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:138
7 INFO  writing...
└─ uart_split::__embassy_main::task::{generator#0} @ src/bin/uart_split.rs:54
8 TRACE starttx
└─ embassy_nrf::uarte::{impl#6}::write::{generator#0}::{closure#0} @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:112
9 INFO  uarte tx drop
└─ embassy_nrf::uarte::{impl#7}::drop @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:138
10 INFO  did_stoptx false
└─ embassy_nrf::uarte::{impl#7}::drop @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:138
11 INFO  uarte tx and rx drop: done
└─ embassy_nrf::uarte::drop_tx_rx @ /Users/huntc/Projects/hacking/embassy/embassy-nrf/src/fmt.rs:138
```

embassy-nrf/src/uarte.rs
```diff
@@ -487,19 +568,23 @@ impl<'d, U: Instance, T: TimerInstance> Write for UarteWithIdle<'d, U, T> {
...
         pub endrx_waker: AtomicWaker,
         pub endtx_waker: AtomicWaker,
+        pub tx_rx_refcount: AtomicU8,
     }
     impl State {
         pub const fn new() -> Self {
             Self {
                 endrx_waker: AtomicWaker::new(),
                 endtx_waker: AtomicWaker::new(),
+                tx_rx_refcount: AtomicU8::new(2),
```
Dirbaio, line 587: Make this `0`, so the states can go in `.bss` so it doesn't use extra flash.

Set it to `2` in `Uart::new()`. it's needed for correctness too: if you drop the uart then create it again refcount has to be reset to 2, otherwise the second drop wont work.

Dirbaio (changes requested):
LGTM other than one issue with the refcounting.

embassy-nrf/src/uarte.rs
huntc: Fixed in commit 0642eec01e23d9037e21186cfb0c169e56f9a161

Dirbaio: :rocket: !!!

Dirbaio (approved):
This is great!! Thank you!

bors r+

huntc:
> This is great!! Thank you!
> 

My absolute pleasure.

PR #546, opened 2021-12-16
Update rust-toolchain

lulf:


Dirbaio (approved):
bors r+

huntc:
This is to avoid the asm feature declarations, or also other reasons?

lulf:
@huntc just general keeping up with improvements in nightly, and in addition to asm! removal, the rustfmt::skip can now be removed.

PR #551, opened 2021-12-19
nrf/gpio: add infallible inherent methods, remove some duplication.

Dirbaio:
Add infallible inherent methods, so that users don't have to import a trait, or `.unwrap()`.

This implements Input and Output using FlexPin, to avoid some code duplication.


huntc (approved):
I was told once that it is possible for GPIO set/clear operations to fail on some hardware targets. That always puzzled me. But this interface feels better as I just tend to unwrap.

Dirbaio:
Pins still impl the fallible InputPin/OutputPin traits. This *adds* infallible methods, so when you're using HAL types directly you don't have to unwrap.

PR #552, opened 2021-12-19
embassy-nrf: migrate to embedded-hal 1.0, embedded-hal-async

Dirbaio:


embassy-nrf/Cargo.toml
```diff
@@ -47,11 +47,14 @@ embassy = { version = "0.1.0", path = "../embassy" }
 embassy-macros = { version = "0.1.0", path = "../embassy-macros", features = ["nrf"]}
 embassy-hal-common = {version = "0.1.0", path = "../embassy-hal-common" }
 
+embedded-hal-02 = { package = "embedded-hal", version = "0.2.6" }
+embedded-hal-1 = { package = "embedded-hal", version = "1.0.0-alpha.6", git = "https://github.com/embassy-rs/embedded-hal", branch = "async" }
```
huntc, line 51: I find this confusing having to depend on two flavours.

embassy-nrf/src/twim.rs
```diff
@@ -270,10 +272,6 @@ impl<'d, T: Instance> Twim<'d, T> {
 
         self.read_errorsrc()?;
 
-        if r.txd.amount.read().bits() != buffer.len() as u32 {
-            return Err(Error::Transmit);
-        }
-
```
huntc, line 276: How come this disappeared?

embassy-nrf/src/twim.rs
```diff
@@ -703,14 +674,27 @@ pub enum Error {
...
+            Self::TxBufferZeroLength => eh1::ErrorKind::Other,
+            Self::RxBufferZeroLength => eh1::ErrorKind::Other,
+            Self::DMABufferNotInDataMemory => eh1::ErrorKind::Other,
+            Self::AddressNack => eh1::ErrorKind::NoAcknowledge(eh1::NoAcknowledgeSource::Address),
+            Self::DataNack => eh1::ErrorKind::NoAcknowledge(eh1::NoAcknowledgeSource::Data),
+            Self::Overrun => eh1::ErrorKind::Overrun,
+        }
+    }
+}
+
```
huntc, line 697: Such a pity to lose this fidelity of errors.

huntc (commented):
Nice!

embassy-nrf/Cargo.toml
Dirbaio: You mean 0.2 and 1.0? This is so the HAL can be used with drivers using either 0.2 or 1.0 traits. 

The alternative would be to completely drop 0.2 and tell users to use [embedded-hal-compat](https://github.com/ryankurte/embedded-hal-compat) if they want to use a 0.2 driver. 

I prefer directly supporting both however, because 0.2 is *massively widespread* in the ecosystem, so it'll take years for everything to upgrade to 1.0, so this will be a major usability issue until then.

embassy-nrf/src/twim.rs
Dirbaio: Hmm, I thought it couldn't happen but it seems it can: https://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_twim.c#L167

I'll check what's up and fix

embassy-nrf/Cargo.toml
huntc: At what point do you think we will let get of 0.2 support? Should we at least mark it as being deprecated?

embassy-nrf/src/twim.rs
Dirbaio: It's how the EH ErrorKind is defined. 

It makes sense though: These errors are very nrf-specific. Users of embassy-nrf can still check our `Error` enum,  Hal-independent users via the traits can only check the kind, but it will have no idea about the nrf-specific errors, so there's nothing it can do anyway.

embassy-nrf/Cargo.toml
Dirbaio: No idea about the timeline, but I guess years.

Not sure if you can mark a trait impl as deprecated, but even if you could it'd be ineffective: the user getting the warning can do nothing about it, it's the driver author that should do the work of updating it.

lulf: Having gone through the embedded-hal-compat annoyance, I think the approach in this PR is much more helpful to end users 🚀

embassy-nrf/src/twim.rs
Dirbaio: These error checks are now back.

huntc:
Seems like the only majority of changes for end-users here is the removal of some trait imports in their code. Nice!

PR #554, opened 2021-12-23
Update stm32-data with F3 Timer register changes

VasanthakumarV:
Changes to `stm32-data` as part of https://github.com/embassy-rs/stm32-data/pull/112


Dirbaio (approved):
LGTM

bors r+

PR #555, opened 2021-12-23
Use cortex-m only on cortex-m archs.

Dirbaio:
Without this, build fails for iOS.


Dirbaio:
bors r

Dirbaio:
bors r+

:zany_face:

PR #559, opened 2021-12-28
USART DMA example for the STM32F7

olofwalker:
This small PR adds the USART DMA write an example for the STM32F7. The example has been tested on a Nucleo-f767zi board.

Output from `DEFMT_LOG=info cargo run --bin usart_dma`
```
Running `probe-run --chip STM32F767ZITx target/thumbv7em-none-eabihf/debug/usart_dma`
(HOST) INFO  flashing program (108 pages / 108.00 KiB)
(HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
0 INFO  Hello World!
└─ usart_dma::__cortex_m_rt_main @ src/bin/usart_dma.rs:39
1 INFO  wrote DMA
└─ usart_dma::main_task::task::{generator#0} @ src/bin/usart_dma.rs:31
```

### DIFF
--- a/examples/stm32f7/src/bin/usart_dma.rs
+++ b/examples/stm32f7/src/bin/usart_dma.rs
@@ -5,20 +5,17 @@
 #[path = "../example_common.rs"]
 mod example_common;
 use core::fmt::Write;
-use embassy::executor::Executor;
-use embassy::util::Forever;
+use embassy::executor::Spawner;
 use embassy_stm32::dma::NoDma;
 use embassy_stm32::usart::{Config, Uart};
+use embassy_stm32::Peripherals;
 use embassy_traits::uart::Write as _Write;
 use example_common::*;
 
-use cortex_m_rt::entry;
 use heapless::String;
 
-#[embassy::task]
-async fn main_task() {
-    let p = embassy_stm32::init(Default::default());
-
+#[embassy::main]
+async fn main(_spawner: Spawner, p: Peripherals) {
     let config = Config::default();
     let mut usart = Uart::new(p.UART7, p.PA8, p.PA15, p.DMA1_CH1, NoDma, config);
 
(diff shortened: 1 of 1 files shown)


examples/stm32f7/src/bin/usart_dma.rs
```diff
@@ -0,0 +1,46 @@
...
+use embassy_stm32::dma::NoDma;
+use embassy_stm32::usart::{Config, Uart};
+use embassy_traits::uart::Write as _Write;
+use example_common::*;
+
+use cortex_m_rt::entry;
+use heapless::String;
+
+#[embassy::task]
+async fn main_task() {
```
huntc, line 19: This example can be simplified by using the embassy:main macro eg https://github.com/embassy-rs/embassy/blob/4271226fc06130d30678739d4869db4fc1820fd8/examples/stm32f7/src/bin/blinky.rs#L14

huntc:
Nice!

examples/stm32f7/src/bin/usart_dma.rs
```diff
@@ -0,0 +1,46 @@
...
+    let p = embassy_stm32::init(Default::default());
+
+    let config = Config::default();
+    let mut usart = Uart::new(p.UART7, p.PA8, p.PA15, p.DMA1_CH1, NoDma, config);
+
+    for n in 0u32.. {
+        let mut s: String<128> = String::new();
+        core::write!(&mut s, "Hello DMA World {}!\r\n", n).unwrap();
+
+        usart.write(s.as_bytes()).await.ok();
```
matoushybl, line 29: The error shouldn't be ignored like this, there is the `unwrap!()` macro from defmt, that will print the error and panic.

Using the macro won't however work on the `.unwrap()` on the line 27 as an implementation of defmt::Format would be required on the Error type. That is fine though.

Nice work!

olofwalker: Thanks, will add that unwrap macro and the suggestion @huntc made and push the update.

examples/stm32f7/src/bin/usart_dma.rs
```diff
@@ -0,0 +1,30 @@
...
+use example_common::*;
+
+use heapless::String;
+
+#[embassy::main]
+async fn main(_spawner: Spawner, p: Peripherals) {
+    let config = Config::default();
+    let mut usart = Uart::new(p.UART7, p.PA8, p.PA15, p.DMA1_CH1, NoDma, config);
+
+    for n in 0u32.. {
```
huntc, line 22: Is the for-loop necessary?

huntc (approved):

Dirbaio (approved):
Thank you!

bors r+

PR #561, opened 2021-12-29
stm32/dac: Fix disable_channel

bgamari:
Previously disable_channel enabled rather than disabled the requested
channel due to an apparent copy-paste error. Refactor to eliminate this
sort of issue by construction.


Dirbaio (approved):
LGTM, thank you! 

bors r+

PR #565, opened 2022-01-04
stm32: RCC cleanups

Dirbaio:


lulf (approved):
Agreed with the changes, I think the original stuff is mostly copy paste from the stm32xx hals :+1:

PR #574, opened 2022-01-10
Add the possibility to reconfigure Spi mode and bit order configurati…

matoushybl:
…on on the fly.

I have not tested these changes. I am also not sure if the peripheral should be disabled when changing these settings. What do you think @Dirbaio ?

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -34,7 +34,8 @@ pub enum Error {
 }
 
 // TODO move upwards in the tree
-pub enum ByteOrder {
+#[derive(Copy, Clone)]
+pub enum BitOrder {
     LsbFirst,
     MsbFirst,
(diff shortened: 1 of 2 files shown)


Dirbaio:
Looks great! The idea of reusing the `Config` struct is interesting.

One possible downside is anything we add to `Config` in the future  we'll have to support both setting in `new()` and changing in `reconfigure()`. I don't think this will be an issue though, there won't be any "unchangeable"  things since you can change anything anyway by dropping and recreating the Spi.

matoushybl:
Thanks for the review. I also don't think that would be an issue.

I've refactored the code, alongside with a small change to stm32-data, that should clean up the code a bit. It still remains untested on real hardware though. Will do that asap.

Dirbaio (approved):
LGTM, feel free to merge when tested on HW :)

PR #576, opened 2022-01-12
Make exti an optional feature

lulf:


Dirbaio:
LGTM!

Could you add checking build both with+without EXTI to `ci.sh`?

Dirbaio (approved):
Awesome! :D

bors r+

PR #579, opened 2022-01-14
embassy-stm32: L1 family RCC fixes

unrelentingtech:


examples/stm32l1/memory.x
```diff
@@ -1,5 +1,5 @@
 MEMORY
 {
   FLASH : ORIGIN = 0x08000000, LENGTH = 128K
-  RAM : ORIGIN = 0x20000000, LENGTH = 32K
+  RAM : ORIGIN = 0x20000000, LENGTH = 16K
```
Dirbaio, line 4: the example is configured for a particular chip, if we change this we should change that too: https://github.com/embassy-rs/embassy/blob/2811978492463e369ecb6604751e4ee40c2b4216/examples/stm32l1/Cargo.toml#L11

Alternatively this memory.x can be deleted, so it uses the one autogenerated by embassy-stm32 (memory-x Cargo feature)

unrelentingtech: Oh nice, deletion indeed results in something that works on my 16K chip.

There shouldn't be that much difference between these chips, only some peripherals might be unavailable but the examples don't touch those. The `*cb-a` option works on c8t6 just fine.

Dirbaio (approved):
Nice, thanks for the PR! :) 

bors r+

Issue #557, opened 2021-12-23
STM32 USB (OTG) implementation


FrozenDroid:
Hi. For our upcoming project, we need a USB-OTG implementation. We are using the STM32L476

I see that there is no implementation yet. I'm creating this issue as a track issue to coordinate work.

Has anyone done any research into implementing this in the Embassy HAL?

Dirbaio:
There's this HAL-independent driver that works with all chips with OTG: https://github.com/stm32-rs/synopsys-usb-otg

It should be easy to wrap it with [this](https://github.com/embassy-rs/embassy/blob/master/embassy-hal-common/src/usb/mod.rs), similar to what #539 does with nrf-usbd.

FrozenDroid:
Oh, I entirely missed this. I'll have a look!

PR #580, opened 2022-01-14
Integrate stm32-usbd

unrelentingtech:
Not exactly #557 because other hardware, but kinda related.

Big open question: the EP memory size / 2x16 access flag / onboard pull-up flag… is there a way to get them into stm32-data? Or can it only be a `cfg()` mess? :/ (they only differ by top-level family and I've only found the info in reference manuals)

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -23,6 +23,8 @@ futures = { version = "0.3.17", default-features = false, features = ["async-awa
 rand_core = "0.6.3"
 sdio-host = "0.5.0"
 embedded-sdmmc = { git = "https://github.com/thalesfragoso/embedded-sdmmc-rs", branch = "async", optional = true }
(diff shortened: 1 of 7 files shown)


Dirbaio:
> Big open question: the EP memory size / 2x16 access flag / onboard pull-up flag… is there a way to get them into stm32-data? Or can it only be a cfg() mess? :/ (they only differ by top-level family and I've only found the info in reference manuals)

It might be mineable from the ST XMLs at https://github.com/embassy-rs/stm32-data-sources/tree/main/cubedb/mcu/IP ... but doing that, plus codegen-ing macros for it in stm32-metapac, plus using it in embassy-stm32 will probably be more work than a bit of `cfg`-spam.

Dirbaio:
Closing in favor of #709 

Thanks anyway for the PR, the new one borrows your cfg's! :)

PR #581, opened 2022-01-14
stm32: expose all functionality as inherent methods.

Dirbaio:
This is the previous step to implementing both the embedded-hal 0.2 and embedded-hal 1.0 + embedded-hal-async traits.

The equivalent in nrf was done in #552 

- Removes need for `unwrap` in gpio.
- Removes need for `use embedded_hal::whatever` in all cases.


lulf (approved):
LGTM! 🚀

PR #582, opened 2022-01-18
Make advanced timer trait not require general purpose timer trait as …

matoushybl:
…the timers are too different.

When developing pwm driver, I originally used T: GeneralPurpose16bitTimer as it could support both GP timers and advanced timer, but advanced timer requires further modifications in registers accessible only in it (BDTR - bit AOE).

This PR makes advanced timers depend on Basic16bitTimer instead, which should hopefully improve type safety and allow for better timer drivers that can distinguish between advanced timers and general purpose ones.

### DIFF
--- a/embassy-stm32/src/pwm/mod.rs
+++ b/embassy-stm32/src/pwm/mod.rs
@@ -62,9 +62,7 @@ impl From<OutputCompareMode> for stm32_metapac::timer::vals::Ocm {
 pub(crate) mod sealed {
     use super::*;
 
-    pub trait CaptureCompareCapable16bitInstance:
-        crate::timer::sealed::GeneralPurpose16bitInstance
-    {
+    pub trait CaptureCompareCapable16bitInstance: crate::timer::sealed::Basic16bitInstance {
         unsafe fn set_output_compare_mode(&mut self, channel: Channel, mode: OutputCompareMode);
 
         unsafe fn enable_channel(&mut self, channel: Channel, enable: bool);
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
timers are so cursed... Thank you! :)

bors r+

matoushybl:
Thanks for the review, they are indeed.

matoushybl:
Hi, 
I see that the problem with AOE has been fixed in a different manner, so it would make sense to make advanced timers `GeneralPurpose16bitTimer` back.

The only problem I see if I remember things correctly is that we won't be able to distinguish between those timers when implementing drivers based on the `GeneralPurpose16bitTimer`. We can't make sure that the timer is not `AdvancedTimer` since we can't have negative trait bounds. Not sure how if that case would ever be needed though.

chemicstry:
But is there a usecase where you would want timer to be general purpose and **not** advanced? If advanced timers behave identical to general purpose ones, then I do not see why that would be needed. AFAIK the only difference is that outputs are disabled by default, which can be handled by the workaround in #827. If anything else arises we can revert the revert 😄

Any opinion @Dirbaio ?

Dirbaio:
If Adv functionality/registers are a strict superset of Gp16 functionality then it makes sense to allow using an Adv timer as Gp16. (I don't know whether this is actually the case or not, you guys are the experts!)

If it is, drivers using Gp16 shouldn't have to care if the timer is really Adv, it'd always work. Do you have any examples of drivers that might want to know?

chemicstry:
Adv registers are a superset of Gp16 in `stm32-data` definitions and I double checked it in STM32F4 SVDs.

Just to reiterate, the only so far observed functional difference is that Adv timers have an additional register (BDTR) with global output enable, which is unfortunatelly disabled by default. Calling `enable_outputs(true)` from #827 makes Adv timers behave identical to Gp16.

PR #585, opened 2022-01-23
Permit many sequences to be passed

huntc:
Sequences are now passed in via the start method to avoid having to stop the PWM and restart it. Sequences continue to be constrained with the same lifetime of the Pwm struct itself. The pwm_sequence example has been extended to illustrate multiple sequences being passed around.

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -138,7 +138,7 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
 
     /// Start or restart playback
     #[inline(always)]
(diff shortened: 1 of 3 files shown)


huntc:
Converting to a draft as this present version is UB. I was expecting that the lifetime of the sequence reference would be as long as the Pwm struct. As @jacobrosenthal has pointed out though, it is only "at least as long".

I was then rather hoping that that subtypes would save me, but no cigar:

```rust
    pub fn start<'d0, 'd1: 'd0>(&'d1 self, sequence: &'d0 [u16], times: SequenceMode) -> Result<(), Error>
```

... my intent was to declare that `self` must live at least as long as `sequence`.

lulf (approved):
I think this API makes a lot more sense 🚀

PR #588, opened 2022-01-24
Add embassy-boot

lulf:
Initial take on a bootloader using embassy, borrowing design from https://github.com/Dirbaio/bootloader-sample :

* Relies on application to write new firmware and signal firmware to be switched, and marking new firmware as 'good'
* Implements a two-bank swap algorithm, with a power safe copy of data.
* Allow reverting to previous firmware if application has not marked new firmware as safe.

In addition, this implementation brings:

* A thorough documentation of the swap/revert algorithm
* A generic API that is platform agnostic
* Good unit test coverage of the core bootloader part
* An API for the application to interact with the bootloader state

The bootloader is split into multiple creates. The embassy-boot crate contains the platform independent bootloader. This part is unit tested with a dummy in-memory flash block.

The embassy-boot-nrf crate contains the platform-specific parts for nrf, which is kept at a minimum. The linker script defines a .bootloader section initialized with a known magic value.

I have also added examples that work with the bootloader.

TODO:

* nRF: Add watchdog with a configurable timeout
* Add support for STM32
* Work with off-chip flash


lulf:
Might be of interest to you @matoushybl (iirc you were asking about embassy bootloader a while back)

matoushybl:
LGTM, nice work! The only downside is that this approach most likely won't work for STM32s with large sector sizes, like the H7.

lulf:
@matoushybl The sector size should be configurable at compile time for the embassy-boot crate. For STM32, I would expect another crate embassy-boot-stm32 similar to embassy-boot-nrf

lulf:
Now works with softdevice

lulf:
After thinking about it, I'll try to get embedded-storage PR  through first so we can still have a releasable embassy

PR #589, opened 2022-01-24
stm32/i2c: allow empty writes

darkwater:
The Senseair Sunrise CO2 sensor expects a wake-up packet in the form of (START, address, STOP), which looks like a `write(addr, &[])`, but this assertion prevents sending that.

I'm not sure why the assertion is there. Sending empty packets works fine in my limited testing, at least.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -139,7 +139,7 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
     }
 
     unsafe fn master_read(address: u8, length: usize, stop: Stop, reload: bool, restart: bool) {
-        assert!(length < 256 && length > 0);
+        assert!(length < 256);
 
         if !restart {
             // Wait for any previous address sequence to end
@@ -170,7 +170,7 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
     }
 
     unsafe fn master_write(address: u8, length: usize, stop: Stop, reload: bool) {
-        assert!(length < 256 && length > 0);
+        assert!(length < 256);
 
         // Wait for any previous address sequence to end
         // automatically. This could be up to 50% of a bus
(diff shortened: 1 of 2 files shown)


Dirbaio:
I think it's there because DMA doesn't like 0-byte transfers. These funcs are used from both blocking and DMA (async) code paths. Have you tested if async works? 

If it doesn't work, it should be easy to fix: just make it not do a DMA transfer if length is zero.

darkwater:
Yeah, the async implementation with DMA is the one I tested it with. It's not causing any obvious issues on an STM32L476RG, at least.

darkwater:
Oh hold on, I might've missed something while testing

darkwater:
Yeah I didn't test it properly and it didn't actually work. Does it make sense to:

- Do what I did here, to use the non-dma implementation for empty writes?
- Add a `buf.len() > 0` assert to the DMA transfer functions instead?
- Allow empty reads in the same way here?

darkwater:
Alright I've made the other changes.

embassy-stm32/src/i2c/v2.rs
```diff
@@ -619,8 +627,8 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
         TXDMA: super::TxDma<T>,
         RXDMA: super::RxDma<T>,
     {
-        self.write_dma_internal(address, bytes, true, true).await?;
-        self.read_dma_internal(address, buffer, true).await?;
+        self.write(address, bytes).await?;
```
Dirbaio, line 630: I don't think this is correct. Old code does "write, restart, read", new code does "write, stop, start, read"

darkwater: I think the write part does the same, but I did miss a different parameter on the read call. I'll just copy over the if statement.

darkwater: Ah, the write won't do the same if it takes the empty path of course.

darkwater: Can you confirm if the old code was correct? Seems to me like it would do "write, stop, restart, read"

Dirbaio: Haven't personally tested it, it's supposed to work but it's possible it was already wrong, yes.

darkwater: I think the latest commit should have the same behaviour as the old code, at least. The calls in the empty paths in `write_read` now also match the calls in `blocking_write_read`.

Dirbaio (approved):
Looks great, Thank you! 

bors r+

PR #591, opened 2022-01-25
PWM WS2812B example and flexible sequence config

huntc:
I've permitted the PWM sequences to be mutated on stopping the PWM by associating them with a new  `SingleSequencer` structure. This is so that we can perform effects on the LEDs (and other use-cases, I'm sure!). The example has been updated to illustrate the use of this by flashing a WS2812B LED.

There's also a `Sequencer` structure for more sophisticated PWM interactions, along with a `pwm_double_sequence`  example to illustrate.

These changes should make it possible to attain all of the nRF PWM functionality available.

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -342,6 +342,14 @@ impl<'d, 's, T: Instance> SingleSequencer<'d, 's, T> {
         };
         self.sequencer.start(start_seq, times)
     }
+
+    /// Stop playback. Disables the peripheral. Does NOT clear the last duty
+    /// cycle from the pin. Returns any sequences previously provided to
+    /// `start` so that they may be further mutated.
+    #[inline(always)]
+    pub fn stop(&self) {
+        self.sequencer.stop();
+    }
 }
 
 /// A composition of sequences that can be started and stopped.
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -321,7 +321,7 @@ impl<'s> Sequence<'s> {
 /// is used.
 #[non_exhaustive]
 pub struct Sequences<'d, 's, T: Instance> {
-    _pwm: &'s mut SequencePwm<'d, T>,
+    pub pwm: &'s mut SequencePwm<'d, T>,
     sequence0: Sequence<'s>,
     sequence1: Option<Sequence<'s>>,
 }
@@ -333,7 +333,7 @@ impl<'d, 's, T: Instance> Sequences<'d, 's, T> {
         sequence1: Option<Sequence<'s>>,
     ) -> Self {
         Sequences {
-            _pwm: pwm,
+            pwm,
             sequence0,
             sequence1,
         }
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -131,9 +131,7 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
     /// Start or restart playback. Takes at least one sequence along with its
     /// configuration. Optionally takes a second sequence and/or its configuration.
     /// In the case where no second sequence is provided then the first sequence
-    /// is used. In the case where no second sequence configuration is supplied,
-    /// the first sequence configuration is used. The sequence mode applies to both
-    /// sequences combined as one.
+    /// is used. The sequence mode applies to both sequences combined as one.
     #[inline(always)]
     pub fn start(
         &mut self,
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -323,9 +323,9 @@ pub struct SingleSequencer<'d, 's, T: Instance> {
 
 impl<'d, 's, T: Instance> SingleSequencer<'d, 's, T> {
     /// Create a new sequencer
-    pub fn new(pwm: &'s mut SequencePwm<'d, T>, sequence: Sequence<'s>) -> Self {
+    pub fn new(pwm: &'s mut SequencePwm<'d, T>, words: &'s [u16], config: SequenceConfig) -> Self {
         Self {
-            sequencer: Sequencer::new(pwm, sequence, None),
+            sequencer: Sequencer::new(pwm, Sequence::new(words, config), None),
         }
     }
 
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -31,8 +31,6 @@ pub struct SequencePwm<'d, T: Instance> {
     ch1: Option<AnyPin>,
     ch2: Option<AnyPin>,
     ch3: Option<AnyPin>,
-    sequence0: Option<Sequence<'d>>,
-    sequence1: Option<Sequence<'d>>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


examples/nrf/src/bin/pwm_sequence_ppi.rs
```diff
@@ -11,26 +11,27 @@ use embassy::executor::Spawner;
...
-    config.refresh = 30;
+    let mut seq_config = SequenceConfig::default();
+    seq_config.refresh = 30;
 
     let mut pwm = unwrap!(SequencePwm::new(
         p.PWM0, p.P0_13, NoPin, NoPin, NoPin, config,
     ));
 
-    let _ = pwm.start(&seq_values, SequenceMode::Times(1));
+    let _ = pwm.start(&seq_values, seq_config, None, None, SequenceMode::Infinite);
```
huntc, line 34: Infinite seems a better demo in that button 2 can then be used to stop it.

embassy-nrf/src/pwm.rs
```diff
@@ -136,12 +128,28 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
     #[inline(always)]
-    pub fn start(&mut self, sequence: &'d [u16], times: SequenceMode) -> Result<(), Error> {
-        slice_in_ram_or(sequence, Error::DMABufferNotInDataMemory)?;
-
-        if sequence.len() > 32767 {
+    pub fn start(
+        &mut self,
+        sequence0: &'d [u16],
+        sequence_config0: SequenceConfig,
+        sequence1: Option<&'d [u16]>,
```
Dirbaio, line 142: It's a bit strange that you can pass `sequence1=Some, sequence_config1=None` or vice versa, shouldn't it be all or nothing? Perhaps move the buffer into SequenceConfig, maybe rename it to Sequence, then take `Option<Sequence<'d>>`?

huntc: Done as of commit 12ce02457438c1dd1566f3f3c43537ae4f54f699

huntc:
I've now also provided [a commit](https://github.com/embassy-rs/embassy/pull/591/commits/76cc4237c8c02537eb59ad9a47790fe1014afc3f) that permits the subsequent mutation of a PWM sequence by storing them as a mutable ref within the new `Sequence` struct. The example has been updated to illustrate the use of this by flashing the LED.

embassy-nrf/src/pwm.rs
```diff
@@ -117,55 +121,70 @@ impl<'d, T: Instance> SequencePwm<'d, T> {
...
+    pub fn start(
+        &mut self,
+        sequence0: Sequence<'d>,
+        sequence1: Option<Sequence<'d>>,
+        times: SequenceMode,
+    ) -> Result<(), Error> {
+        let alt_sequence = sequence1.as_ref().unwrap_or(&sequence0);
+
+        slice_in_ram_or(sequence0.words, Error::DMABufferNotInDataMemory)?;
+        slice_in_ram_or(alt_sequence.words, Error::DMABufferNotInDataMemory)?;
```
huntc, line 149: Because the `Sequence` struct now retains the word buffer as mutable, do we need this check? I'm thinking not...

Dirbaio: why is it mutable now? `&'d [u16]` should be enough, we just need it stays alive for the duration of the DMA, we don't need to write to it.

huntc: It is mutable so that we can subsequently write into it. You’ll see that the stop method transfers ownership back. The example illustrates the use case where we are updating the buffer so that the led pulsates. 

If we cant update the buffer then we’d need a number of stack-allocated buffers for each frame of the pulsating effect. This would be expensive on memory, particularly when more than one LED needs to be addressed.

huntc: Commit 2112b83e9d328bd662156a167e721ab28bffc6f7
d76cd5ceaf5140c48ef97180beae156c0c0e07c8 now owns the sequence buffers. This approach owns the sequence buffers which, while introducing an extra move, eliminates the need to guard the lifetime of the sequence buffer. Given ownership, the buffer will be retained until the PWM sequence task is stopped. WDYT?

huntc: I've now reverted the owning of buffers - too much copying for comfort. So, we're back to passing around &'d mut [16]`, which I think is fine given the opportunity to subsequently modify it. I'm pretty happy with things now.

huntc:
Any further feedback or opposition to merging? My project has begun to depend on the approach here, and it is working well, but I’m happy to consider an alternate direction if required.

embassy-nrf/src/pwm.rs
```diff
@@ -376,20 +418,36 @@ pub struct SequenceConfig {
...
             end_delay: 0,
         }
     }
 }
 
+/// A composition of a sequence buffer and its configuration.
+#[non_exhaustive]
+pub struct Sequence<'d> {
+    /// The words comprising the sequence. Must not exceed 32767 words.
+    pub words: &'d mut [u16],
```
lulf, line 431: Why must this buffer be mutable? It's not required by the PWM as far as I can tell? If you need to write again to the buffer, you could instead let the Sequence go out of scope in your app, which should allow you to write to the buf and create a new Sequence instance. That way feels more immutable.

lulf (commented):
Looks good in general, but I left a comment on the need for passing a mutable reference vs. having a more 'immutable' API.

embassy-nrf/src/pwm.rs
huntc: Thanks for feedback.

You can’t let the sequence go out of scope because the PWM is using it until you tell it to stop. The PWM needs to own the buffer between starting and stopping. 

Hopefully the example I provide shows why we need the mutability. When we stop the PWM we get the buffer refs back so we can then update the contents inline. In my example, we lighten and darken the colours. 

I hope I’m not missing something…

lulf: I can see the convenience argument. What I was thinking of is that instead of 
```  
        if let (Some(color_seq), Some(reset_seq)) = pwm.stop() {
            color_seq.words[color_bit] = bit_value;
            unwrap!(pwm.start(color_seq, Some(reset_seq), SequenceMode::Times(2)));
        }
```

You could do

```  
        pwm.stop();
        color_seq_words[color_bit] = bit_value;
        let color_seq = Sequence::new(&color_seq_words, SequenceConfig::default());
        let reset_seq = Sequence::new(&reset_seq_words, reset_seq_config);
        unwrap!(pwm.start(color_seq, Some(reset_seq), SequenceMode::Times(2)));
```

I.e. recreating the Sequence instances with the same buffer references. But I can see how that's a bit more work. I've not explored if it can be done more elegantly.

huntc: But you can’t subsequently modify color_seq_words once you pass it into the start method... That buffer must live until the Pwm struct is dropped or the stop method is called… it has to be owned by the Pwm struct… 

Thanks again though. If you could try modifying the example then that’d be great. I could be missing something for sure.

lulf: Right, and the pwm.stop() in the above code will allow you to modify color_seq_words again, because the PWM would no longer own it. But FWIW I think maybe your approach is easier to use, so I'm fine with merging this as is.

embassy-nrf/src/pwm.rs
```diff
@@ -376,20 +418,36 @@ pub struct SequenceConfig {
...
+}
+
 /// How many times to run the sequence
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
 pub enum SequenceMode {
-    /// Run sequence n Times total
+    /// Run sequence n Times total.
+    /// 1 = Run sequence 0 once
+    /// 2 = Run sequence 0 and then sequence 1
+    /// 3 to 4 = Run sequence 0, sequence 1, sequence 0 and then sequence 1
```
jacobrosenthal, line 448: Times is a pretty bad way of describing all these options
Times(1) runs seq 1
Times(2) runs seq 0 then seq1
times(3) runs seq1 then seq0 then seq1

Times was a good name for my api, because mine was looping the same seq n times

jacobrosenthal:
This does distort the old api. Previously because seq werent mut I could stick them in a forever and call start from any task.

Also now to use it to repeat a single sequence you need to pass that sequence in twice? Pretty bad api for the original use case
```
        let seq_confg = SequenceConfig::default();
        let seq0 = Sequence::new(&color_seq_words, seq_confg);
        let seq1 = Sequence::new(&color_seq_words, seq_confg);
        unwrap!(pwm.start(color_seq, Some(seq1), SequenceMode::Times(2)));
```

Perhaps we can keep my old api and rename it LoopingPwm and you can have the SequencePwm name, or AdvancedPwm or something

embassy-nrf/src/pwm.rs
huntc: I will correct.

I don't believe the looping behaviour has changed.

huntc:
> This does distort the old api. Previously because seq werent mut I could stick them in a forever and call start from any task.
> 

Does this not diminish the ability to reason where a peripheral can be updated if it can be updated from several tasks? I've found that tasks end up owning peripherals and coordinating activity with other tasks via channels, signals, and so forth.

> Also now to use it to repeat a single sequence you need to pass that sequence in twice? Pretty bad api for the original use case
> 
> ```
>         let seq_confg = SequenceConfig::default();
>         let seq0 = Sequence::new(&color_seq_words, seq_confg);
>         let seq1 = Sequence::new(&color_seq_words, seq_confg);
>         unwrap!(pwm.start(color_seq, Some(seq1), SequenceMode::Times(2)));
> ```

You should just pass it in once with `Times(1)` above. Please note that I've not yet pushed all of my changes so perhaps this isn't apparent.

> 
> Perhaps we can keep my old api and rename it LoopingPwm and you can have the SequencePwm name, or AdvancedPwm or something

I personally think that the less code repetition the better, and in fact there's the opportunity to consolidate both `SequencePwm` and `SimplePwm` again down the track.

It is important to me that we retain support for your use-cases though. Understanding your use-cases and design decisions hasn't been apparent to me through the code that is there, but I'm keen for that to change.

jacobrosenthal:
> You should just pass it in once with Times(1) above. Please note that I've not yet pushed all of my changes so perhaps this isn't apparent.

Now how would I do it 3 times?

embassy-nrf/src/pwm.rs
huntc: > Times(1) runs seq 1 

`Times(1)` has always run sequence 0. https://github.com/embassy-rs/embassy/blob/master/embassy-nrf/src/pwm.rs#L180

> Times(2) runs seq 0 then seq1

This is what I wrote.

> times(3) runs seq1 then seq0 then seq1

I will correct accordingly.

jacobrosenthal:
But that's a really odd API. Nobody would think times(3) would start
swapping order. Your API needs a rethink. The times is a great API for
looping a single sequence n times and makes nice sense.

huntc:
> But that's a really odd API. Nobody would think times(3) would start swapping order. Your API needs a rethink. The times is a great API for looping a single sequence n times and makes nice sense.

Is this not your API? It certainly isn't mine.

https://github.com/embassy-rs/embassy/blame/abd21ad139ad1518e6c80903e8fe80c942026358/embassy-nrf/src/pwm.rs#L175

embassy-nrf/src/pwm.rs
huntc: Sequence buffer is no longer mutable.

embassy-nrf/src/pwm.rs
huntc: The sequence buffer is no longer mutable. A new Sequences struct can easily be dropped to return ownership of the Pwm driver and then establish new sequences.

huntc:
> This does distort the old api. Previously because seq werent mut I could stick them in a forever and call start from any task.

Sequences are no longer mutable. There should be no difference to the use of Forever as you require it. Would you have a moment to verify that? Thanks.

jacobrosenthal:
Its just a bad api sorry. 
Itll need a redesign of the Times enum

I cant say it differently. someone else will have to weigh in.


Old api
loop 1 time, seq happens 1 times
    `pwm.start(seq_values, SequenceMode::Times(1).unwrap();`
loop 2 time, seq happens 2 times
    `pwm.start(seq_values, SequenceMode::Times(2).unwrap();`
loop 3 time, seq happens 3 times
    `pwm.start(seq_values, SequenceMode::Times(3).unwrap();`
    
Times is a very good enum name





new api
play 1 time, seq0 happens 1 time
```
        let seq_confg = SequenceConfig::default();
        let seq0 = Sequence::new(&color_seq_words, seq_confg);
        unwrap!(pwm.start(seq0, None), SequenceMode::Times(1)));

```

play 2 times, seq0, then seq1, so each happens 1 time, nothing happens 2x
```
        let seq_confg = SequenceConfig::default();
        let seq0 = Sequence::new(&color_seq_words, seq_confg);
        let seq1 = Sequence::new(&color_seq_words, seq_confg);
        unwrap!(pwm.start(seq0, Some(seq1), SequenceMode::Times(2)));

```

play 3 times, seq1, seq 0, seq1, all of a sudden seq1 is now first? Also seq 1 happens 2x and seq0 happens 1x. Nothing happens 3x
```
        let seq_confg = SequenceConfig::default();
        let seq0 = Sequence::new(&color_seq_words, seq_confg);
        let seq1 = Sequence::new(&color_seq_words, seq_confg);
        unwrap!(pwm.start(seq0, Some(seq1), SequenceMode::Times(3)));

```
* Its incredibly noisy
* Times probably couldnt be more unintuitive

huntc:
> Its just a bad api sorry. Itll need a redesign of the Times enum
> 

I've now introduced a `SingleSequencer` with its own `start` and `stop`, which behaves exactly as the previous `start` and `stop` methods of `SequencePwm`. There's an additional `Sequencer` that provides for more complex behaviour. Please let me know how this API looks to you now.

Dirbaio (approved):
Thank you!

bors r+

PR #593, opened 2022-01-26
traits: migrate Delay to embedded-hal 1.0+async, remove Rng and Flash.

Dirbaio:


matoushybl (approved):
LGTM, thanks! I cannot wait to use new version of embassy with traits available for the whole ecosystem!

PR #595, opened 2022-02-02
stm32f3: fix nonexistent cfg tests

unrelentingtech:
The rcc code was taken from stm32-rs which uses 'x' features, but embassy uses features with full chip names.

Add these 'x' wildcards as cfgs and use them in rcc. They will be useful for USB #580 too.

---

I don't have any F3 boards, so this is not tested. But the original cfg clearly doesn't look right…

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -125,6 +125,10 @@ fn main() {
         println!("cargo:rustc-cfg={}", &chip_name[..chip_name.len() - 2]);
     }
 
+    if chip_name.starts_with("stm32f3") {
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Nice catch, thanks for the PR :) 

generating "wildcard" cfg's like this is an Interesting trick. I wonder if we should do it for all chips. It'd incentivize matching on that instead of on stm32-metapac register versions though, so maybe doing it in a case-by-case basis is better?

bors r+

unrelentingtech:
Yeah, let's only do it for the F3 case right now, where the differences are… weird (PLL frequency calculations, USB memory size and access width)

PR #602, opened 2022-02-07
Add stm32 USB OTG peripherals

### LINKED ISSUE #557: STM32 USB (OTG) implementation

chemicstry:
Fixes #557. This is similar to #580, but for synopsys IP.

I could add examples to other chips, but I have no way of testing them. The F4 example is tested and working.


Dirbaio:
This is awesome, thank you for your hard work! Tested it myself on an F4 as well, worked like a treat.

Code is great, just one concern: otg_fs and otg_hs are almost identical, do you think you could unify them? it seems to me it should be easy, unless there's some hardware detail that means it's a bad idea, like OTG_HS being usable in both HS and FS mode.?

chemicstry:
I wasn't sure about unifying. The HS peripheral has much more interrupts and pins due to external PHY support. FS only has internal PHY. So if additional HS features are implemented in the future it will start to look way different from FS.

chemicstry:
Actually the https://github.com/stm32-rs/synopsys-usb-otg only limits the API when FS mode is selected and there are no other functional changes. So I guess we can use only the HS feature of `synopsys-usb-otg` and limit the API at `embassy` level.

chemicstry:
So the problem with FS/HS peripherals is that https://github.com/stm32-rs/synopsys-usb-otg uses two different sets of register definitions for each. They share similar registers, but no one has confirmed if HS register set is superset of FS register set (see https://github.com/stm32-rs/synopsys-usb-otg/issues/12#issuecomment-721378363). There are many reasons to believe so: the stm32h7 reference manual refers to FS peripheral as "HS with FS only capabilities", morover I tested the HS driver with FS peripheral on stm32f4 and it seems to be working fine. However, there is no guarantee for this, just experimental evidence and I don't have time at the moment to investigate register compatibilities inside `synopsys-usb-otg`.

So at this point I see two possibilities:
1. Keep FS and HS implementations separate. This means that only one can be used at a time (limitation of `synopsys-usb-otg`) and chips with FS and HS peripherals will only be able to use a single interface.
2. Use HS implementation for both FS and HS. And wait until someone reports a bug that on some chip FS peripheral does not work with HS driver.

chemicstry:
So the consensus in matrix chat was to go with route #2 and use the HS driver for both peripherals. Looking for scream testers.

While at it, I also added ULPI external PHY support.

I've heard that some STM32 chips also have internal High-Speed PHY, but I'm not sure which. So if someone finds a need for such thing it's only a matter of adding `fn new_hs(...)` function with `PhyType::InternalHighSpeed` gated behind appropriate cfg's.

Dirbaio (approved):
This is great! Thank you!

bors r+

PR #604, opened 2022-02-09
Add embassy-boot

lulf:
Continuation of https://github.com/embassy-rs/embassy/pull/588

Embassy-boot is a simple bootloader that works together with an application to provide firmware update capabilities with a minimal risk.

The bootloader consists of a platform-independent part, which implements the swap algorithm, and a platform-dependent part (currently only for nRF) that provides addition functionality such as watchdog timers softdevice support.

The bootloader is intended to be configurable for different flash sizes and architectures, and only requires that embedded-storage flash traits are implemented.

The nRF version can be configured programatically as a library, or using linker scripts to set the partition locations for DFU, ACTIVE and STATE

* DFU: Where the next firmware version should be written. This is used by the FirmwareUpdater
* ACTIVE: Where the current firmware version resides. Written by bootloader when swap magic is set
* STATE: Contains the bootloader magic and the copy progress. Can be 1-N pages long (depending on how much flash you have which will determine the copy progress index size


embassy-boot/boot/src/fmt.rs
```diff
@@ -0,0 +1,225 @@
+#![macro_use]
```
matoushybl, line 1: Is it possible to use `embassy::fmt` instead of copy-pasting this?

embassy-boot/nrf/memory.x
```diff
@@ -0,0 +1,18 @@
+MEMORY
```
matoushybl, line 1: What is the difference between this and `memory_bm.x`?

embassy-boot/nrf/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+        };
+        let ret = mbr::sd_mbr_command(&mut cmd);
+        assert_eq!(ret, NRF_SUCCESS);
+
+        let msp = *(addr as *const u32);
+        let rv = *((addr + 4) as *const u32);
+
+        trace!("msp = {=u32:x}, rv = {=u32:x}", msp, rv);
+
+        core::arch::asm!(
```
matoushybl, line 103: Please add some high level comment here on what the assembly does and why it is needed. Thanks!

embassy-boot/nrf/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+    const SCB_ICSR: *const u32 = 0xE000_ED04 as *const u32;
+    let irqn = core::ptr::read_volatile(SCB_ICSR) as u8 as i16 - 16;
+
+    panic!("DefaultHandler #{:?}", irqn);
+}
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    unsafe {
+        core::arch::asm!("udf #0");
```
matoushybl, line 43: Is this different from `cortex_m::asm::udf()`?

matoushybl (approved):
Awesome! Minor questions, but LGTM.

embassy-boot/boot/src/fmt.rs
lulf: I don't think so, I tried to do it once, but I think it wasn't possible, at least at the time. It's present in embassy-nrf, embassy-stm32 and embassy at present, so I think maybe we can fix all of these in one swoop, if possible.

embassy-boot/nrf/memory.x
lulf: None really. Maybe it's better to remove this and have a user needing to copy memory-bm.x to memory.x before compiling? Having this present just makes it easier to get started with the examples I guess.

embassy-boot/nrf/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+use embassy_nrf::nvmc::Nvmc;
+
+#[entry]
+fn main() -> ! {
+    let p = embassy_nrf::init(Default::default());
+    /*
+        for i in 0..10000000 {
+            cortex_m::asm::nop();
+        }
+    */
```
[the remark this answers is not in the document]
lulf: I'll add a comment to why it is there: when using debug probe and RTT, this is needed, otherwise writing to flash 'too early' will cause a hard fault.

PR #605, opened 2022-02-10
stm32: Remove OptionalPin

Dirbaio:
The idea behind OptionalPin has a few problems:

- you need to impl the signal traits for NoPin which is a bit weird https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/dcmi.rs#L413-L416
- you can pass any combination of set/unset pins, which needs checking at runtime  https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/dcmi.rs#L130

The replacement is to do multiple `new` constructors for each combination of pins you want to take.


Dirbaio:
bors r_

PR #611, opened 2022-02-11
Document #[embassy::main(config = ...)]

danielzfranklin:
Add an example documenting the config param based on help I got in the matrix chat.

Because the example is very short and incompatible with the other example code I just included it inline.


Dirbaio (approved):
Thank you!

PR #617, opened 2022-02-12
Add feature defmt-timestamp-uptime

danielzfranklin:
Add the feature defmt-timestamp-uptime. Enabling it adds a timestamp of the number of seconds since startup next to defmt log messages using `Instant::now`.

### DIFF
--- a/embassy/src/fmt.rs
+++ b/embassy/src/fmt.rs
@@ -195,6 +195,10 @@ macro_rules! unwrap {
     }
 }
 
+#[cfg(feature = "defmt-timestamp-uptime")]
+// defmt offers a disply hint for microseconds so we convert from millis
+defmt::timestamp! {"{=u64:us}", crate::time::Instant::now().as_millis() * 1_000 }
+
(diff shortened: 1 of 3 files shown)


embassy/src/fmt.rs
```diff
@@ -195,6 +195,10 @@ macro_rules! unwrap {
     }
 }
 
+#[cfg(feature = "defmt-timestamp-uptime")]
+// defmt offers a disply hint for microseconds so we convert from millis
+defmt::timestamp! {"{=u64:us}", crate::time::Instant::now().as_millis() * 1_000 }
```
Dirbaio, line 200: Could you add `from_micros, as_micros` to `Instant`, and use that instead? It'll be slightly more precise, especially for stm32 which currently defaults to a tick rate of 1mhz. It's also nice for consistency, because `Duration` has it.

danielzfranklin: Done

Dirbaio (approved):
Really nice! Thank you! :D

bors r+

PR #628, opened 2022-02-17
Improve logic for processing button events

ceigel:
Allow blinking to be interrupted by button presses in this sample.


Dirbaio (approved):
Thank you!

bors r+

PR #629, opened 2022-02-18
Provides an HFCLK timer driver

huntc:
The existing (default) timer driver for nRF implements one using the RTC, which uses the LFCLK. An additional feature has been provided where the HFCLK can be used for the timer driver by using TIMER1. An STM example has also been ported so that either driver implementation can be exercised.

To use the HFCLK driver, simply use the `time-driver-timer1` feature in place of the `time-driver-rtc1` one when depending on embassy-nrf.


embassy-nrf/src/time_driver.rs
```diff
@@ -83,194 +79,407 @@ impl AlarmState {
...
+        alarm_count: AtomicU8::new(0),
+        alarms: Mutex::const_new(CriticalSectionRawMutex::new(), [ALARM_STATE_NEW; ALARM_COUNT]),
+    });
+
+    impl TimerDriver {
+        fn init(&'static self, irq_prio: crate::interrupt::Priority) {
+            let r = timer();
+            r.mode.write(|w| w.mode().timer());
+            r.prescaler
+                .write(|w| unsafe { w.prescaler().bits(Frequency::F62500Hz as u8) });
```
Dirbaio, line 316: Why this particular prescaler? I thought you wanted this to have higher precision. Higher tick rate means more preicision. With 1Mhz rate you can measure down to 1us of time. With 16Mhz you can measure down to 62.5ns.

embassy-nrf/src/time_driver.rs
```diff
@@ -83,194 +79,407 @@ impl AlarmState {
...
+    }
 
-            let r = rtc();
+    impl Driver for TimerDriver {
+        fn now(&self) -> u64 {
+            // `period` MUST be read before `counter`, see comment at the top for details.
+            let period = self.period.load(Ordering::Relaxed);
+            compiler_fence(Ordering::Acquire);
+            let r = timer();
+            r.tasks_capture[3].write(|w| unsafe { w.bits(1) });
```
Dirbaio, line 398: This will overwrite whatever's in CC3 which you're currently using for (half of the) overflow handling. It's super unfortunate that you can't read the counter value directly. WTF Nordic? but it is what it is. We have to sacrifice a CC channel to read the counter >_<

This means it can't support 3 alarms unlike RTC. It can max 2 or 1 depending on how you implement overflows

Dirbaio (changes requested):
- Could you split into several files? `time_driver/{mod.rs, rtc.rs, timer.rs}`
- The overflow handling is not correct. 
   - RTC is 24bit, TIMER is 32bit. Many constants need updating, in `calc_now` and the rest of the code.
   - RTC has OVERFLW event, TIMER doesn't. We still need some way to detect overflow. Alternatives:
      - Set one CC at 0, another at 0x8000_0000. This uses 2 CCs so we can support only 1 alarm (4 ccs total, minus 1 dummy for reading the counter, minus 2 for overflow handling)
      - Set one CC at 0x8000_0000, when it fires change it to 0, when it fires change it back to 0x8000_0000 etc.

huntc:
> * Could you split into several files? `time_driver/{mod.rs, rtc.rs, timer.rs}`

Resolved in commit d99c74807a5b09fad79702aae1ba686bf8a1d8d0

Dirbaio:
closing due to outdatedness. Feature is still wanted tho!

PR #636, opened 2022-02-23
stm32: Add support for using TIM12 and TIM15 as time driver

matoushybl:
I am not sure what the effect of reducing the number of alarms will be, but these are the only timers I have available on my board.

### DIFF
--- a/embassy-stm32/src/time_driver.rs
+++ b/embassy-stm32/src/time_driver.rs
@@ -18,7 +18,11 @@ use crate::rcc::sealed::RccPeripheral;
 use crate::timer::sealed::Basic16bitInstance as BasicInstance;
 use crate::timer::sealed::GeneralPurpose16bitInstance as Instance;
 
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/time_driver.rs
```diff
@@ -320,8 +341,18 @@ macro_rules! impl_timer {
...
+);
+
+#[cfg(time_driver_tim12)]
+crate::pac::peripherals!(
+    (timer, TIM12) => { impl_timer!(TIM12, TIM8_BRK_TIM12); };
+);
+
+#[cfg(time_driver_tim15)]
+crate::pac::peripherals!(
+    (timer, TIM12) => { impl_timer!(TIM12, TIM8_BRK_TIM12); };
```
Dirbaio, line 357: this should be TIM15

Dirbaio (approved):
Thank you!

bors r+

PR #640, opened 2022-02-23
Skip EasyDMA slice location check for empty slices and copy data if necessary

### LINKED ISSUE #631: Overly aggressive easyDMA SPIM buffer location check?
When using the nRF52840 SPIM, I've noticed that the `spim.read(&mut rx)` method always throws an `Err(DMABufferNotInMemory)` no matter what. This appears to be related to the fact that it calls `async_inner` with `&[]` as the argument for the tx buffer which appears to have a memory location which `slice_in_ram` does not like — though in my mind it should not matter because it is an empty slice anyways.

Using `transfer` in the same way with an inline array as the tx buffer causes the same issue. It can only be remedied by creating a variable which holds the empty array and passing that as the buffer. See the assertions below:
image
image

Similarly, this appears to also affect the write method when using inline arrays like `spim.write(&[1, 2, 3])` — probably because the compiler dumps it into the static data section and references it from there instead of allocating it at runtime. I'm not sure there is a reasonable fix for that one.

However, the read method *should* work by either allocating a variable inside the method itself or updating the `slice_in_ram` method to ignore zero-length arrays (given that easyDMA has no problem with that; I have no idea).

TilBlechschmidt:
As discussed, this PR makes the following changes:
- Ignore pointer location of zero-length slices (fixes #631)
- Change default functions so they copy the tx buffer if it does not reside in RAM
- Introduce new variants for `write`, `transfer`, and their blocking versions which fails instead of copying
- Add documentation about the motivation behind all these variants
image

Remaining TODOs:

- [x] Change copying behaviour for other peripherals
    - [x] TWI
    - [x] UART
- [x] Add module-level documentation regarding EasyDMA and `_from_ram` method variants

@Dirbaio it probably makes sense for you to review it now before I "copy" over the changes to the other two peripherals.

### DIFF
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -8,6 +8,7 @@ use embassy::util::Unborrow;
 use embassy_hal_common::unborrow;
 use futures::future::poll_fn;
 
+use crate::chip::FORCE_COPY_BUFFER_SIZE;
 use crate::gpio::sealed::Pin as _;
 use crate::gpio::{self, AnyPin};
 use crate::gpio::{Pin as GpioPin, PselBits};
@@ -273,8 +274,9 @@ impl<'d, T: Instance> Spim<'d, T> {
             Ok(_) => Ok(()),
             Err(Error::DMABufferNotInDataMemory) => {
                 trace!("Copying SPIM tx buffer into RAM for DMA");
-                let tx_copied = tx.clone();
-                self.blocking_inner_from_ram(rx, tx_copied)
+                let mut tx_buf = [0u8; FORCE_COPY_BUFFER_SIZE];
+                tx_buf[..tx.len()].copy_from_slice(tx);
+                self.blocking_inner_from_ram(rx, &tx_buf[..tx.len()])
             }
             Err(error) => Err(error),
         }
@@ -304,8 +306,9 @@ impl<'d, T: Instance> Spim<'d, T> {
             Ok(_) => Ok(()),
             Err(Error::DMABufferNotInDataMemory) => {
                 trace!("Copying SPIM tx buffer into RAM for DMA");
-                let tx_copied = tx.clone();
-                self.async_inner_from_ram(rx, tx_copied).await
+                let mut tx_buf = [0u8; FORCE_COPY_BUFFER_SIZE];
+                tx_buf[..tx.len()].copy_from_slice(tx);
+                self.async_inner_from_ram(rx, &tx_buf[..tx.len()]).await
             }
             Err(error) => Err(error),
         }
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -31,38 +31,7 @@ pub enum Error {
 
 /// Interface for the SPIM peripheral using EasyDMA to offload the transmission and reception workload.
 ///
-/// ## Data locality requirements
-///
-/// On nRF chips, EasyDMA requires the buffers to reside in RAM. However, Rust
-/// slices will not always do so. Take the following example:
-///
-/// ```no_run
-/// // As we pass a slice to the function whose contents will not ever change,
-/// // the compiler writes it into the flash and thus the pointer to it will
-/// // reference static memory. Since EasyDMA requires slices to reside in RAM,
-/// // this function call will fail.
-/// let result = spim.write_from_ram(&[1, 2, 3]);
-/// assert_eq!(result, Error::DMABufferNotInDataMemory);
-///
-/// // The data is still static and located in flash. However, since we are assigning
-/// // it to a variable, the compiler will load it into memory. Passing a reference to the
-/// // variable will yield a pointer that references dynamic memory, thus making EasyDMA happy.
-/// // This function call succeeds.
-/// let data = [1, 2, 3];
-/// let result = spim.write_from_ram(&data);
-/// assert!(result.is_ok());
-/// ```
-///
-/// Each function in this struct has a `_from_ram` variant and one without this suffix.
-/// - Functions with the suffix (e.g. [`write_from_ram`](Spim::write_from_ram), [`transfer_from_ram`](Spim::transfer_from_ram)) will return an error if the passed slice does not reside in RAM.
-/// - Functions without the suffix (e.g. [`write`](Spim::write), [`transfer`](Spim::transfer)) will check whether the data is in RAM and copy it into memory prior to transmission.
-///
-/// Since copying incurs a overhead, you are given the option to choose from `_from_ram` variants which will
-/// fail and notify you, or the more convenient versions without the suffix which are potentially a little bit
-/// more inefficient.
-///
-/// Note that the [`read`](Spim::read) and [`transfer_in_place`](Spim::transfer_in_place) methods do not have the corresponding `_from_ram` variants as
-/// mutable slices always reside in RAM.
+/// For more details about EasyDMA, consult the module documentation.
 pub struct Spim<'d, T: Instance> {
     phantom: PhantomData<&'d mut T>,
 }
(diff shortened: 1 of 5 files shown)


embassy-nrf/src/util.rs
```diff
@@ -19,10 +19,10 @@ pub(crate) fn slice_in_ram<T>(slice: *const [T]) -> bool {
     ptr >= SRAM_LOWER && (ptr + len * core::mem::size_of::<T>()) < SRAM_UPPER
 }
 
-/// Return an error if slice is not in RAM.
+/// Return an error if slice is not in RAM. Skips check if slice is zero-length.
 #[cfg(not(feature = "nrf51"))]
 pub(crate) fn slice_in_ram_or<T, E>(slice: *const [T], err: E) -> Result<(), E> {
-    if slice_in_ram(slice) {
+    if slice.len() > 0 && slice_in_ram(slice) {
```
Dirbaio, line 25: use `slice_ptr_parts` to get the len, which works on stable

TilBlechschmidt:
For the record: I used the following command to generate the documentation (took me a while to find the right flags ^^)
```bash
cargo doc --no-deps --features nightly,nrf52840 --target thumbv8m.main-none-eabihf
```

embassy-nrf/src/spim.rs
```diff
@@ -234,7 +268,18 @@ impl<'d, T: Instance> Spim<'d, T> {
         Ok(())
     }
 
-    async fn async_inner(&mut self, rx: *mut [u8], tx: *const [u8]) -> Result<(), Error> {
+    fn blocking_inner(&mut self, rx: &mut [u8], tx: &[u8]) -> Result<(), Error> {
+        match self.blocking_inner_from_ram(rx, tx) {
+            Ok(_) => Ok(()),
+            Err(Error::DMABufferNotInDataMemory) => {
+                let tx_copied = tx.clone();
```
Dirbaio, line 275: `.clone()` here clones just the slice pointer, not the data. The data will still be in Flash.

You need to copy through a buffer in the stack. This is how nrf-hal does it for example: https://github.com/nrf-rs/nrf-hal/blob/master/nrf-hal-common/src/spim.rs#L94

TilBlechschmidt: I was wondering about that whether it would do that. Thanks for the pointer (pun very much intended)!

TilBlechschmidt: I was initially trying to do it that way but having a fixed-size buffer allocation of 255-1024 byte seems like a pretty hefty price to pay. Will the compiler optimise it away if it is not required (either by not calling the "regular" methods or only calling it with in-memory checks)?

Dirbaio: For blocking, if the buffer is in a fn that is only called when copying, it should only use the extra memory if it does get called... For async it'll probably use the memory no matter what, the size of a future is the max of all possible child futures...

Dirbaio: It is what it is :(

TilBlechschmidt: Either way, I've changed it to use a buffer based on the `chip::FORCE_COPY_BUFFER_SIZE` constant that was already present.

Dirbaio:
LGTM to the copy now!

There's one issue left: it'll panic if writing something bigger than `FORCE_COPY_BUFFER_SIZE`. Also related (and this is a bug that was already present) it'll also fail if the buffer is longer than `EASY_DMA_SIZE` (which is surprisingly low on some chips, 255 bytes on nrf52832!). We'd have to do some chunking.

Since this bug was kinda present before, I'm OK with adding the copy now and the chunking later, so up to you :)

TilBlechschmidt:
For now I'll implement the copying for TWIM and UARTE. Maybe in the future, I'll tackle the challenge of handling large reads/writes. Nonetheless that deserves its own PR. Is there already an issue for that?

Dirbaio:
Sounds great!

> Is there already an issue for that?

There isn't

TilBlechschmidt:
Apparently, this particular issue has been encountered previously. The [TWIM implementation already contains a similar "workaround"](https://github.com/embassy-rs/embassy/blob/35636953b2048394733eaff95fad17970788d08a/embassy-nrf/src/twim.rs#L435). Should I restructure them to match the layout/naming of the UARTE and SPIM interfaces?

On another note, I am not particularly happy with the way the documentation works out. Would you be fine with lifting it up to the module level (basically a subheading in the `embassy_nrf` module docs that describes how EasyDMA works and what all those `_from_ram` methods are for instead of doing it over and over again in each of the three relevant peripherals)?

Dirbaio:
> Should I restructure them to match the layout/naming of the UARTE and SPIM interfaces?

Sounds good, they should all be consistent because the underlying issue is the same.

> Would you be fine with lifting it up to the module level (basically a subheading in the embassy_nrf module docs that describes how EasyDMA works and what all those _from_ram methods are for instead of doing it over and over again in each of the three relevant peripherals)?

Sounds good as well! Perhaps add a link to it from all the affected fns, or just from the `_from_ram` variants..?

TilBlechschmidt:
I would say just the `_from_ram` variants are fine, maybe with an additional note on the struct itself that makes the reader aware that there is something to take note of.

TilBlechschmidt:
I'll rework the methods on `TWIM` to match the behaviour of those in the other two peripherals (regarding panicking that is) — even though the current handling in `TWIM` is arguably nicer, I would remove it in favour of #651.

Current implementation for reference:
```rust
if wr_buffer.len() > FORCE_COPY_BUFFER_SIZE { // <-- nice error handling which will be unnecessary once #651 lands
    return Err(Error::TxBufferTooLong);
}

// Copy to RAM
let wr_ram_buffer = &mut [0; FORCE_COPY_BUFFER_SIZE][..wr_buffer.len()];
wr_ram_buffer.copy_from_slice(wr_buffer);

self.blocking_write_read(address, wr_ram_buffer, rd_buffer)
```

embassy-nrf/src/util.rs
```diff
@@ -19,10 +19,11 @@ pub(crate) fn slice_in_ram<T>(slice: *const [T]) -> bool {
     ptr >= SRAM_LOWER && (ptr + len * core::mem::size_of::<T>()) < SRAM_UPPER
 }
 
-/// Return an error if slice is not in RAM.
+/// Return an error if slice is not in RAM. Skips check if slice is zero-length.
 #[cfg(not(feature = "nrf51"))]
 pub(crate) fn slice_in_ram_or<T, E>(slice: *const [T], err: E) -> Result<(), E> {
-    if slice_in_ram(slice) {
+    let (_, len) = slice_ptr_parts(slice);
+    if len > 0 && slice_in_ram(slice) {
```
Dirbaio, line 26: This will always fail if len==0, we want it to always succeed.

embassy-nrf/src/spim.rs
```diff
@@ -28,6 +29,40 @@ pub enum Error {
...
+/// On nRF chips, EasyDMA requires the buffers to reside in RAM. However, Rust
+/// slices will not always do so. Take the following example:
+///
+/// ```no_run
+/// // As we pass a slice to the function whose contents will not ever change,
+/// // the compiler writes it into the flash and thus the pointer to it will
+/// // reference static memory. Since EasyDMA requires slices to reside in RAM,
+/// // this function call will fail.
+/// let result = spim.write_from_ram(&[1, 2, 3]);
+/// assert_eq!(result, Error::DMABufferNotInDataMemory);
```
Dirbaio, line 45: This is `Err(Error::DMABufferNotInDataMemory)`

embassy-nrf/src/util.rs
TilBlechschmidt: Good catch. I apparently shouldn't author PRs during the middle of the night 🙈

This should do the trick:
```suggestion
    if len == 0 || slice_in_ram(slice) {
```

embassy-nrf/src/spim.rs
TilBlechschmidt: Good catch. I'll add another commit in the following days that moves the docs up to the module level — that should then include a fix for this.

TilBlechschmidt:
A couple of unfortunate trailing whitespaces later, this PR _should_ (fingers crossed) now be looking good. Feel free to give it a review now @Dirbaio 🙂

Dirbaio (approved):
This is great, thank you! I like the docs a lot.

bors r+

TilBlechschmidt:
Thank you for the timely reviews, good feedback and overall nice interaction! It has been fun to contribute 😊

PR #644, opened 2022-02-25
stm32 misc fixes

Dirbaio:
- Fix build when no DBGMCU is present (wl55 cm0 core)
- Fix multicore 2nd core having the wrong NVIC.


Dirbaio:
teleprobe spurious failure

bors r+

PR #645, opened 2022-02-25
stm32 usart: Fix RX interrupt flag handling

timokroeger:
* On v1 interrupts cannot be cleared individually.
  Instead they are cleared implicitly by reading or writing DR (which we do now).
* Multiple error flags can be set at the same time:
  Handle them all in one go intstead of re-entering the ISR for each one so that
  we do not lose any error flags on v1 hardware.
* Wake when the RX buffer becomes full: This allows fast running chips to pull data
  from the buffer before receiving the next byte.

Tested on v1 hardware, lets see if v2 still succeeds on CI.


lulf (approved):
Thanks for this! :rocket: bors r+

PR #648, opened 2022-02-26
Fix nRF Saadc continuous sampling

huntc:
Starting the sampling task prior to starting the SAADC peripheral can lead to unexpected buffer behaviour with multiple channels. We now provide an init callback at the point where the SAADC has started for the first time. This callback can be used to kick off sampling via PPI.

We also need to trigger the SAADC to start sampling the next buffer when the previous one is ended so that we do not drop samples - the major benefit of double buffering.

Given these additional tasks, we now simplify the API by passing in the TIMER and two PPI channels.

As a bonus, we provide an async `calibrate` method as it is recommended to use before starting up the sampling.

The example has been updated to illustrate these new features along with the simplified API.

The changes here have been tested on my nRF52840-DK.

### DIFF
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -10,7 +10,8 @@ use embassy_hal_common::unborrow;
 use futures::future::poll_fn;
 
 use crate::interrupt;
-use crate::ppi::{Event, Task};
+use crate::ppi::{ConfigurableChannel, Event, Ppi, Task};
+use crate::timer::{Frequency, Instance as TimerInstance, Timer};
 use crate::{pac, peripherals};
 
 use pac::{saadc, SAADC};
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/saadc.rs
```diff
@@ -263,9 +268,19 @@ impl<'d, const N: usize> Saadc<'d, N> {
 
     /// Continuous sampling with double buffers.
     ///
+    /// NOTE: It is important that the time spent within the callback supplied
```
Dirbaio, line 271: DMA overwriting the buffer that the user is using is undefined behavior, we can't have this in a safe API.

embassy-nrf/src/saadc.rs
```diff
@@ -328,7 +345,7 @@ impl<'d, const N: usize> Saadc<'d, N> {
         // wouldn't happen anyway.
         compiler_fence(Ordering::SeqCst);
 
-        r.tasks_start.write(|w| unsafe { w.bits(1) });
+        r.tasks_calibrateoffset.write(|w| unsafe { w.bits(1) });
```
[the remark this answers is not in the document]
huntc: Good point. Can I make that a separate PR though?

embassy-nrf/src/saadc.rs
```diff
@@ -368,11 +391,23 @@ impl<'d, const N: usize> Saadc<'d, N> {
...
 
+    /// Return the end event for use with PPI
+    pub fn events_end(&self) -> Event {
+        let r = Self::regs();
+        Event::from_reg(&r.events_end)
+    }
+
     /// Return the sample task for use with PPI
-    pub fn task_sample(&self) -> Task {
+    pub fn tasks_sample(&self) -> Task {
```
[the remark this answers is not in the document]
huntc: Fair enough. I’ll make sure the other PPI methods are consistent here too.

embassy-nrf/src/saadc.rs
huntc: Actually I see now that we are updating the next buffer to write to on EVENTS _STARTED. We need to update it post this event AND the callback having finished.

embassy-nrf/src/saadc.rs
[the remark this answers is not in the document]
huntc: Commit 466d66bd5bb6407a51a31d75b601022cb599acbf takes care of this

embassy-nrf/src/saadc.rs
huntc: I will push another commit shortly for this additional logic. My goal is to absolutely prevent undefined behavior and retain the safe API.

huntc: Commit c36337531238175018a531b2dfe1951417056d9b should take care of things.

embassy-nrf/src/saadc.rs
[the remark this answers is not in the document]
huntc: This was easier than I thought hence quickly getting this in. Calibration is now factored out into a separate method. Commit a42b63895c782d0607bc5d537d3bd9d29f8e61d4 takes care of this.

huntc:
Converting back to a draft as I now feel that there's something inherently wrong when starting the timer sampling before the peripheral is started. Calibration appears to help, but I don't think it is the entire solution.

embassy-nrf/src/saadc.rs
huntc: > Actually I see now that we are updating the next buffer to write to on EVENTS _STARTED. We need to update it post this event AND the callback having finished.

I was wrong about this. Double buffering logic has been restored. Commit fa3a0439740f4d06f07343a1645f54723431ebd7 takes care of this.

huntc:
> Converting back to a draft as I now feel that there's something inherently wrong when starting the timer sampling before the peripheral is started. Calibration appears to help, but I don't think it is the entire solution.

A new init callback resolves this issue.

Dirbaio:
The sampler API was quite complex before already, and adding the `init` closure makes it even more complex.

Most (all?) use of this API will be to sample on a timer. I was thinking it'd be massively simpler if it took ownership of the timer and PPI channels, instead of having the user do the setup.

huntc:
> The sampler API was quite complex before already, and adding the `init` closure makes it even more complex.
> 
> Most (all?) use of this API will be to sample on a timer. I was thinking it'd be massively simpler if it took ownership of the timer and PPI channels, instead of having the user do the setup.

I agree, noting that we give up the use of a PPI and TIMEr peripherals for anything else. Probably worth it though, as you state. 

Can we view that as a separate PR?

huntc:
> 

So, to be clear what you'd like to see here is that `Saadc` adopts an API similar to `BufferedUarte` where a timer and two PPI channels are passed in (two are needed for us too)? Here's a snippet from the `BufferedUarte` constructor:

```rust
        timer: impl Unborrow<Target = T> + 'd,
        ppi_ch1: impl Unborrow<Target = impl ConfigurableChannel + 'd> + 'd,
        ppi_ch2: impl Unborrow<Target = impl ConfigurableChannel + 'd> + 'd,
```

huntc:
> The sampler API was quite complex before already, and adding the `init` closure makes it even more complex.
> 
> Most (all?) use of this API will be to sample on a timer. I was thinking it'd be massively simpler if it took ownership of the timer and PPI channels, instead of having the user do the setup.

Commit a4e516731d399730c5fc4f6e4ad7fcaabf5759c8 takes care of this.

Dirbaio (approved):
Looks great, thank you! :D

bors r+

PR #649, opened 2022-02-28
stm32: Register access for timers now doesn't require self

matoushybl:


Dirbaio (approved):
Nice, more consistent with the rest :+1:

PR #650, opened 2022-02-28
`bxcan` improvements

timokroeger:
* Do not use wildcard reexport for `bxcan`
* `cfgs` to fix `bxcan` implementation for devices with "CAN1" but no "CAN2" peripheral

### DIFF
--- a/embassy-stm32/src/can/bxcan.rs
+++ b/embassy-stm32/src/can/bxcan.rs
@@ -98,7 +98,7 @@ foreach_peripheral!(
     (can, CAN1) => {
         cfg_if::cfg_if! {
             if #[cfg(all(
-                any(stm32l4, stm32l72, stm32l73),
+                any(stm32l4, stm32f72, stm32f73),
                 not(any(stm32l49, stm32l4a))
             ))] {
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/can/bxcan.rs
```diff
@@ -92,17 +92,28 @@ foreach_peripheral!(
...
-    // There are only 14 filter banks when CAN2 is not available.
+    // CAN1 and CAN2 is a combination of master and slave instance.
+    // CAN1 owns the filter bank and needs to be enabled in order
+    // for CAN2 to receive messages.
     (can, CAN1) => {
-        unsafe impl bxcan::FilterOwner for peripherals::CAN1 {
-            const NUM_FILTER_BANKS: u8 = 28;
+        cfg_if::cfg_if! {
+            if #[cfg(all(
+                any(stm32l4, stm32l72, stm32l73),
```
Dirbaio, line 101: there's probably some mistake here, STM32L7 is not a thing

timokroeger: embarrassing mistake, indeed... fixed

Dirbaio (approved):
Thank you! :) 

bors r+

PR #652, opened 2022-03-02
Use new stm32-data registers and fix AHB clock calculation

msamsonoff:
This is the follow-on to my PR against stm32-data that added new register enums for the G0. I have updated the G0 RCC module to use those new enums.

I have also fixed an issue with the calculation of the AHB clock rate. 32 is not available as an AHB prescaler. The sequence jumps from 16 to 64. The original bit shifting math did not account for this gap. I have replaced it with a `match` instead.


Dirbaio (approved):
Rustfmt'd it for you :) 

Thanks for the PR! Nice catch on the `32` divider, it's almost as if ST intentionally lays traps to make using their chips harder :joy: 

bors r+

PR #656, opened 2022-03-08
stm32: Refactor DMA interrupts

GrantM11235:
Previously, every dma interrupt handler called the same `on_irq`
function which had to check the state of every dma channel.

Now, each dma interrupt handler only calls an `on_irq` method for its
corresponding channel or channels.


Dirbaio:
bors try

(so HIL tests run)

Dirbaio:
bors try

Dirbaio (approved):
Thank you!

bors r+

PR #657, opened 2022-03-09
Async usb stack

Dirbaio:
TODO

- [x] Make it work on nRF
- [x] Add a way for classes to handle their own EP0 control requests - thanks @alexmoon!
- [x] Handle CONTROL OUT requests with data.
- [ ] Impl AsyncRead/AsyncWrite for CDC ACM -- will do later, it's not trivial
- [x] Cleanup unwraps/asserts/panics
- [x] Cleanup logs (make everything trace/debug, not info)
- [ ] Port synopsys-usb-otg
- [ ] Port stm32-usbd
- [ ] Add more classes? HID, MSD?


nviennot:
I'll use this PR to mention, I made a USB Host driver for the synopsys-usb-otg using embassy, and it's all async.
It reads USB flash devices.
The sources are here: https://github.com/nviennot/turbo-resin/tree/main/src/drivers/usb
If there's some appetite, I can package this nicely

Dirbaio:
@nviennot that's super interesting, indeed! It would be awesome if you did :)

Dirbaio:
> dirbaio().await: the alternative is adding all the classes to embassy-usb itself
> dirbaio().await: but in the end I thought it'd be bad, because doing a breaking change to a class's API would require major-bumping embassy-usb, which contains the driver traits, which would break all the HAL crates
> dirbaio().await: vs one crate per class, you just major-bump that one and that's it

but now that I think about it, the same applies to breaking changes in the `embassy-usb` public api itself (the core stack, not the classes...). Perhaps we could separate just the *driver traits* into their own crate `embassy-usb-driver`, and have everything else (stack+classes) in `embassy-usb`? 

Something I like about separate class crates is it enforces all the required stuff is public in `embassy-usb` for users to make their own classes out-of-tree though.

lulf:
Either way is fine I think. Having public api in one place sounds good

Dirbaio:
There's this weird issue that causes enumeration to take 5s:

```
0.004486 TRACE USB event: suspend
0.134948 TRACE usb: reset
0.134979 TRACE USB event: resume
0.262481 TRACE control request: Request { direction: In, request_type: Standard, recipient: Device, request: 06, value: 100, index: 00, length: 40 }
0.262573 TRACE   control in accept [12, 01, 10, 02, 00, 00, 00, 08, de, c0, fe, ca, 10, 00, 00, 00, 00, 01]
0.272766 ERROR ControlPipe::data_in timed out.
0.282775 ERROR ControlPipe::data_in timed out.
5.432128 TRACE usb: reset
5.556274 TRACE control request: Request { direction: Out, request_type: Standard, recipient: Device, request: 05, value: 37, index: 00, length: 00 }
5.556365 TRACE   control accept
```

Checked with a USB sniffer, the issue is on the very first control transfer, the host doesn't know the EP0 max_packet_size is 8. We respond with the descriptor in 8-byte chunks, but the host thinks the first chunk of 8 is shorter than the max, and switches to the status phase, while we're waiting to send the rest of the data.

The timeout there I put is a dirty workaround, it makes it work but slowly. If I set EP0 max_packet_size to 64 then the issue is gone.

TinyUSB also had the same issue: https://github.com/hathach/tinyusb/issues/184
They fixed it with a fun "if we don't have addr and the device descriptor would get fragmented, send the first fragment only" hack: 

https://github.com/hathach/tinyusb/blob/9c8c5c1c53214b8954bb28c33a496b5b423d5ecc/src/device/usbd.c#L973-L976

alexmoon:
> Checked with a USB sniffer, the issue is on the very first control transfer, the host doesn't know the EP0 max_packet_size is 8. We respond with the descriptor in 8-byte chunks, but the host thinks the first chunk of 8 is shorter than the max, and switches to the status phase, while we're waiting to send the rest of the data.

Could we add a new `UsbDeviceState` (maybe `AddressedConfigurationNotSent` or something)? Then a get configuration descriptor request would check the state and if it was in that state it would only send the first packet of the configuration and change the state to `Addressed` so future requests would get the whole thing?

Dirbaio:
> Could we add a new UsbDeviceState (maybe AddressedConfigurationNotSent or something)? Then a get configuration descriptor request would check the state and if it was in that state it would only send the first packet of the configuration and change the state to Addressed so future requests would get the whole thing?

Did it with `pending_address` for now, and it works!

We'll probably have to rework address handling a bit when adding support for stm32. Nrf's are special and automatically handle the address, other chips don't, that'll probably mean adding the "state" you mention.

embassy-nrf/src/usb.rs
```diff
@@ -1,43 +1,798 @@
...
+
+    fn enable(self) -> Self::Bus {
+        let regs = T::regs();
+
+        errata::pre_enable();
+
+        regs.enable.write(|w| w.enable().enabled());
+
+        // Wait until the peripheral is ready.
+        while !regs.eventcause.read().ready().is_ready() {}
```
alexmoon, line 202: The datasheet indicates this will typically take 1ms from enable to ready. That seems like a long time to block the executor. Should `Driver::enable` be async? That would require `UsbDriverBuilder::build` to become async as well.

Also, should `Bus` disable the `USBD` on drop?

Dirbaio: > That seems like a long time to block the executor. Should Driver::enable be async?

If it's only at startup maybe it's not a big deal. Not sure if there's any usecase where you'd want to create/drop it dynamically at runtime? No strong opinion, feel free to make it async if you want.

> Also, should Bus disable the USBD on drop?

Ideally yes. I think due to how the lifetimes are setup it's possible for individual endpoints to outlive the Bus, that might be tricky to handle.

alexmoon:
Ok, just FYI here's what I've ended up doing to handle USB power detected/removed events on nrf:

The main task loop:
1. Waits for USB power detected
2. Creates two futures: one for the USB power removed signal and one for USB processing (which will never return).
3. Does a `select` on those two futures then loops.

The USB processing future:
1. Creates a guard wrapper around `&mut USBD` which disables the USBD on drop and implements `Unborrow`
2. Creates a usb `Driver` with a mutable reference to that guard
3. Does all the rest of the configuring and building of the USB
4. Finally, joins the `UsbDevice::run` and USB class futures

I'm not sure if any of that could be made cleaner with changes to the async usb stack, but it works fine as is for my purposes.

Dirbaio:
I was wondering, why disable USB when unplugged? The examples running on nrf52840DK work fine if you unplug/plug without doing anything special. One reason is low-power (?), is there others?

> Creates a guard wrapper around &mut USBD which disables the USBD on drop and implements Unborrow

Why is this required? I think it should be possible to do without if `Bus` disables the peripheral when dropped.

alexmoon:
> I was wondering, why disable USB when unplugged? The examples running on nrf52840DK work fine if you unplug/plug without doing anything special. One reason is low-power (?), is there others?

I've got a battery powered device so low power is the main reason. I also have some concerns about how reliable the USBD is if you leave it enabled with no vbus. I haven't seen any problems, but the PHY is powered from the vbus and there is that ~1ms delay between vbus applied and PHY power ready. I'm just worried there could be race conditions or brownout type issues between the USBD and the PHY in rare cases. The datasheet seems to imply that Nordic expects the USBD to be enabled only when vbus is present. But I could just be being paranoid.

> > Creates a guard wrapper around &mut USBD which disables the USBD on drop and implements Unborrow
> 
> Why is this required? I think it should be possible to do without if `Bus` disables the peripheral when dropped.

This was my solution to the lifetime issues you raised. This way both the bus and the endpoints get tied to the lifetime of the guard and its drop only runs after all of them have terminated.

Dirbaio:
> The datasheet seems to imply that Nordic expects the USBD to be enabled only when vbus is present

:+1: . I wonder if we could make the stack itself handle this? so from the user's point of view the USB driver is always "created", but internally it tracks whether it's enabled or not. Maybe it's not worth the complexity..?

> This was my solution to the lifetime issues you raised. This way both the bus and the endpoints get tied to the lifetime of the guard and its drop only runs after all of them have terminated.

Hmm, I see. Maybe dropping the bus could set all endpoints to "not ready"? This way they won't try to send/receive data, so it should be OK if they're still created. And maybe something similar for the Control pipe...?

Dirbaio:
I'm going to merge now that namespaced-features is stable. The stack is in quite good shape for nRF, I'll hopefully work on stm32 soon. Thank you for all the great contributions @alexmoon ! :)

alexmoon:
> > The datasheet seems to imply that Nordic expects the USBD to be enabled only when vbus is present
> 
> 👍 . I wonder if we could make the stack itself handle this? so from the user's point of view the USB driver is always "created", but internally it tracks whether it's enabled or not. Maybe it's not worth the complexity...

That would be awesome, but I couldn't come up with a clean way of doing it. Part of the issue is that the power events come from a different peripheral, and it's even inaccessible if you're using a Softdevice (like I am). So it would probably have to be something like methods on the `UsbDevice` to tell it to disable/enable based on external events, but then you get into lifetime conflicts with the `run` method...

> > This was my solution to the lifetime issues you raised. This way both the bus and the endpoints get tied to the lifetime of the guard and its drop only runs after all of them have terminated.
> 
> Hmm, I see. Maybe dropping the bus could set all endpoints to "not ready"? This way they won't try to send/receive data, so it should be OK if they're still created. And maybe something similar for the Control pipe...?

Yeah, I think something like that could be made to work. It would probably require some additional shared state. I don't know if the control pipe is an issue, since it's only accessed from the UsbDevice, which must be dropped before the Bus is dropped.

alexmoon:
> I'm going to merge now that namespaced-features is stable. The stack is in quite good shape for nRF, I'll hopefully work on stm32 soon. Thank you for all the great contributions @alexmoon ! :)

🎉 You're very welcome!

Dirbaio:
> the power events come from a different peripheral, and it's even inaccessible if you're using a Softdevice (like I am).

Oh, oops. Very true. :(

> I don't know if the control pipe is an issue, since it's only accessed from the UsbDevice, which must be dropped before the Bus is dropped.

True! That's how it's used *today* though, it doesn't have to be this way forever, so making drivers rely on that seems a bit fragile. I wonder if the control pipe could be part of the Bus.

PR #659, opened 2022-03-14
Add API for BufferedUarte without flow control

lulf:


huntc:
We deliberately require flow control as BufferedUarte requires a rethink. Please see https://github.com/embassy-rs/embassy/issues/536.

lulf:
Hmm, I need to rewrite a lot of code then, but ok :/

PR #661, opened 2022-03-14
Add support for splitting stm32 usart into TX and RX

lulf:
* Keeps existing API for usart, but wraps it in Tx and Rx sub-types
* Adds split() method similar to nRF for getting indepdendent TX and RX
  parts
* Implements e-h traits for TX and RX types


Dirbaio (approved):
LGTM, this looks really nice :) 

Why is it WIP?

lulf:
@Dirbaio just want to test it first ;)

PR #662, opened 2022-03-14
stm32: Finish unifying spi versions

GrantM11235:
Notable changes:
- `set_word_size` is always called before disabling SPE. This is important because `set_word_size` may or may not re-enable SPE.
- The rx buffer is flushed on v1 as well. I don't know if this is required.
- All functions are now generic over word size


Dirbaio:
bors try

Dirbaio (approved):
Thank you!

bors r+

PR #665, opened 2022-03-15
Rebuild when the chip's JSON changes

nviennot:


Dirbaio:
Oooh this is very nice!

The generated PAC also depends on the register yamls from `data_dir/registers`, could you add that as well?

nviennot:
Absolutely, it's done

Dirbaio (approved):
Thank you!

bors r+

PR #666, opened 2022-03-15
stm32/spi: Clear rx fifo in blocking methods

GrantM11235:


embassy-stm32/src/spi/mod.rs
```diff
@@ -574,36 +574,90 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
...
+    fn drop(&mut self) {
+        unsafe {
+            self.sck.as_ref().map(|x| x.set_as_disconnected());
+            self.mosi.as_ref().map(|x| x.set_as_disconnected());
+            self.miso.as_ref().map(|x| x.set_as_disconnected());
+        }
+    }
+}
+
+struct ConfiguredBlocking<'a, W: Word> {
```
Dirbaio, line 613: Could you do this without a separate struct? It adds a bit of boilerplate and indirection, and since only a private API, the added type safety is not that worth it IMO.

Dirbaio (changes requested):

PR #667, opened 2022-03-15
Remove duplicate stm32-metapac/src/common.rs with chiptool

nviennot:
There's a duplicate file common.rs with the chiptool crate. This PR makes the source of truth the one in chiptool.

This PR is a good pair with https://github.com/embassy-rs/chiptool/pull/4


Dirbaio (approved):
Thank you!

bors r+

PR #669, opened 2022-03-16
Add SDMMC v1 and SDIO support

chemicstry:
SDMMC v2 peripheral is an extension of SDMMC v1 (or SDIO) so I managed to reuse most of the code, with some cfg's.

Apart from small differeces in registers, the biggest change is that v2 uses internal DMA, while v1 has to use shared DMA peripheral. This makes code a bit uglier, because DMA channel for v1 has to be passed around. Not sure if it's possible to make it any cleaner.

This also adds `TransferOptions` structure to DMA, because SDMMC v1 requires setting peripheral flow control and burst transfers. Let me know if some alternative way would be prefered.

I tested this on STM32F429ZIT6 (with sd card) and STM32H745ZIT6 (with oscilloscope).

Depends on: https://github.com/embassy-rs/stm32-data/pull/130


chemicstry:
The `sdmmc-rs` feature, which enables [embedded-sdmmc](https://github.com/rust-embedded-community/embedded-sdmmc-rs) FAT FS is currently broken, because it uses an outdated fork of asyncifyed lib. I don't think there is currently an async FS library, so maybe this feature could be removed for now.

However, I'm not sure how usable this peripheral is, without an async FS. I have yet to test if it works with `future::block_on` to use with a non-async FS library.

embassy-stm32/build.rs
```diff
@@ -483,7 +483,9 @@ fn main() {
         (("i2c", "TX"), quote!(crate::i2c::TxDma)),
         (("dcmi", "DCMI"), quote!(crate::dcmi::FrameDma)),
         (("dcmi", "PSSI"), quote!(crate::dcmi::FrameDma)),
-        (("sdmmc", "SDIO"), quote!(crate::sdmmc::SdmmcDma)),
+        // Same channel is used for RX/TX
+        (("sdmmc", "RX"), quote!(crate::sdmmc::SdmmcDma)),
+        (("sdmmc", "RX"), quote!(crate::sdmmc::SdmmcDma)),
```
Dirbaio, line 488: RX->TX?

chemicstry: SDMMCv1 uses the same DMA channel for both directions. However, `stm32-data` generates 3 signals `RX`, `TX` and `SDIO` (or `SDMMCx`) for the same DMA channel so I just picked `RX`. Maybe this could also be fixed in `stm32-data`, but idk how you call it?

chemicstry: ohhh, sorry I saw the problem xD

Dirbaio:
> The sdmmc-rs feature, which enables [embedded-sdmmc](https://github.com/rust-embedded-community/embedded-sdmmc-rs) FAT FS is currently broken, because it uses an outdated fork of asyncifyed lib. I don't think there is currently an async FS library, so maybe this feature could be removed for now.

Yeah, this is quite unfortunate. `embedded-sdmmc` is kind of a kitchen sink though: it has a BlockDevice trait, an impl for SD SPI, MBR partition parsing, and FAT filesystem. Only the SPI impl is actually related to SDMMC, the other stuff can be used with any block device really. 

Perhaps the best way forward is to add a BlockDevice trait to [embedded-storage](https://github.com/rust-embedded-community/embedded-storage), then build on top of that? the higher layers could go into a crate like `embassy-fat32`, `embassy-mbr` or something.

chemicstry:
> Yeah, this is quite unfortunate. `embedded-sdmmc` is kind of a kitchen sink though: it has a BlockDevice trait, an impl for SD SPI, MBR partition parsing, and FAT filesystem. Only the SPI impl is actually related to SDMMC, the other stuff can be used with any block device really.
>
> Perhaps the best way forward is to add a BlockDevice trait to [embedded-storage](https://github.com/rust-embedded-community/embedded-storage), then build on top of that? the higher layers could go into a crate like `embassy-fat32`, `embassy-mbr` or something.

I also had weird feelings for that crate. There is a standalone [fatfs](https://github.com/rafalh/rust-fatfs) lib, which could be used instead, without all other bloat. But it does not support async either (yet?).

I think the proposed `embedded-storage` trait would be ideal. Do you want me to submit a PR there now, before proceeding with this,, or can we fix that in the followup PR later?

Dirbaio:
> Do you want me to submit a PR there now, before proceeding with this,, or can we fix that in the followup PR later?

If the driver is already useful as-is, we can merge it and build on top later, no probs.

> There is a standalone [fatfs](https://github.com/rafalh/rust-fatfs) lib, which could be used instead, without all other bloat. But it does not support async either (yet?).

Huh, fun! Seems to be a bit more mature than `embedded-sdmmc`, it can delete files and create/delete dirs. Hmm.

Dirbaio (approved):
bors r+

PR #670, opened 2022-03-17
Make UART futures Send

chemicstry:
This is a quick fix to make `Uart` futures implement `Send`.

Previously they were `!Send`, because pointer to the data register was held across an await point. Simple rearrange fixes the issue.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -106,8 +106,8 @@ impl<'d, T: Instance, TxDma> UartTx<'d, T, TxDma> {
                 reg.set_dmat(true);
             });
         }
-        let dst = tdr(T::regs());
-        crate::dma::write(ch, request, buffer, dst).await;
+        let transfer = crate::dma::write(ch, request, buffer, tdr(T::regs()));
+        transfer.await;
         Ok(())
     }
 
@@ -150,9 +150,8 @@ impl<'d, T: Instance, RxDma> UartRx<'d, T, RxDma> {
                 reg.set_dmar(true);
             });
         }
-        let r = T::regs();
-        let src = rdr(r);
-        crate::dma::read(ch, request, src, buffer).await;
+        let transfer = crate::dma::read(ch, request, rdr(T::regs()), buffer);
+        transfer.await;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usart/mod.rs
```diff
@@ -106,8 +106,8 @@ impl<'d, T: Instance, TxDma> UartTx<'d, T, TxDma> {
                 reg.set_dmat(true);
             });
         }
-        let dst = tdr(T::regs());
-        crate::dma::write(ch, request, buffer, dst).await;
+        let transfer = crate::dma::write(ch, request, buffer, tdr(T::regs()));
```
Dirbaio, line 109: why the temp variable? I think `write(...).await;` should work

Dirbaio (commented):
Interesting!

Why do you need the futures to be Send? just curious.

embassy-stm32/src/usart/mod.rs
Dirbaio: wat? :D

chemicstry:
Ah, I'm using the `FuturesExt::boxed()` from [futures](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.boxed), which enforces the `Send`.
I guess I could just manually box them, but is there a reason for futures not to be `Send`?

embassy-stm32/src/usart/mod.rs
chemicstry: I mean, maybe the `write(ch, request, buffer, tdr(T::regs()))` is internally expanded by compiler as:
```rust
let _tmp1 = T::regs();
let _tmp2 = tdr(_tmp1);
write(ch, request, buffer, _tmp2);
```
I'm not a compiler guru so IDK :D

Dirbaio:
> Ah, I'm using the FuturesExt::boxed() from [futures](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.boxed), which enforces the Send.

Huh, no idea why that requires Send. You can box manually just fine without Send.

> is there a reason for futures not to be Send?

For most drivers, none! It's nice to make futures Send if we can. Some drivers can't Send safely, such as the ones using PeripheralMutex, but they're very few.

> An alternative would be to make dma::write(...) accept a newtype (maybe struct DmaAddress(*mut u32) ?), which wraps the raw pointer and implements the Send itself. Then you could call await directly on DMA methods.

The workaround is fine, I wouldn't make it more complicated with newtypes. It's just that it was unexpected to me that it was necessary! :) Could you add a comment explaining it? then we're good to go

Dirbaio (approved):
Thank you!

bors r+

PR #672, opened 2022-03-17
Reset peripherals on enable

chemicstry:
Add reset on initialization to peripherals that did not have it before. This fixes problems when same peripheral is reinitialized at runtime multiple times.

Some exceptions:
- ADC: all ADCs share a single reset
- DCMI: does reset before enable - couldn't find anything about the order in the reference manual. Just keep it if it works?

I also fixed safety issues where global RCC registers where accessed without critical section.


Dirbaio (approved):
bors r+

:rocket:

Dirbaio:
bors d+

PR #673, opened 2022-03-19
Inline GPIO functions

nviennot:
All GPIO functions are monomorphized (per pin). Inlining these make the ROM smaller when using opt-level="z"


Dirbaio (approved):
Thank you!

bors r+

PR #674, opened 2022-03-19
Add Dynamic GPIO for stm32

nviennot:


embassy-stm32/src/gpio.rs
```diff
@@ -378,6 +390,138 @@ impl<'d, T: Pin> Drop for OutputOpenDrain<'d, T> {
     }
 }
 
+pub enum Dynamic<'d, T: Pin> {
+    Input(Input<'d, T>),
+    Output(Output<'d, T>),
+    OutputOpenDrain(OutputOpenDrain<'d, T>),
+}
+
+impl<'d, T: Pin> Dynamic<'d, T> {
```
Dirbaio, line 399: Could you impl the embedded-hal traits as well? (both 0.2 and 1.0)

Dirbaio (changes requested):
I would prefer if it worked like nrf's `Flex`, instead of an enum:

- It has a tiny cost: it uses 1 byte of RAM for storing the enum discriminant, and has to check it at runtime.
- It panics if the pin is in wrong mode. I agree usually panicking on misuse is better, but I'd argue using the pin in the "wrong" modes is not misuse here, it has some interesting use cases:
  - You can set the output value while in input mode, so it takes effect when you switch to output. This allows changing output->input->output while "keeping" the previous output value, and saves a register write when switching to output mode.
- The enum approach is something the user can write in their own crate (and they can do so more optimized, for example just Input and Output if they're not using OpenDrain). `Flex` is lower level, there's no way for the user to write it themselves (other than using the PAC)

nviennot:
Agreed. The enum is a cost we don't need to pay.
We should do the implementation like nrf where Input/Output wraps Flex.

Dirbaio:
Closing in favor of #851

PR #677, opened 2022-03-22
nrf: nrf52832 doesn't have SPI3

Dirbaio:


matoushybl (approved):
Checked the datasheet and it doesn't. Thanks!

PR #678, opened 2022-03-27
Add minimal F2 family support

Gekkio:
Here's the bare minimum to support F2 family (207/217/205/215). A lot is missing in RCC (e.g. PLL support), but this is enough to have a working blinky example. The example is set up for a NUCLEO-F207ZG board which I don't have, but I've tested it on my custom board with a F215 and different pinout :sweat_smile: 

After looking at other RCC implementation, I noticed there's two main API styles: a "low-level" API (e.g. L0) where the `Config` struct has dividers and other low-level "knobs", and a "high-level" API (e.g. F0) where it has desired clock frequencies and the RCC implementation figures out how to achieve them. Which one is preferred? Personally I like the low-level API slightly more, because it gives you the most control and it would be easy to also provide some functions to calculate the required parameters based on desired clock frequencies.

F2 has a nasty errata: a delay or DSB instruction must be added after every RCC peripheral clock enable. I've added this workaround to build.rs, but am not sure if this is the best approach. Any comments?

I'm planning to add PLL support too once I know which kind of API is preferred. Would you prefer a separate pull request for that, or should I continue working on this one?


Dirbaio:
Thanks for the PR! Looks great overall, just one thing:

- Could you add building the F2 examples, and also embassy-stm32 for a F2 chip to CI? (`ci.sh` and `ci_stable.sh` files). Preferably the "biggest" F2 chip, if it has more stuff it's more likely to fail if there's something wrong (unfortunately there's so many chips that making CI build everything would take way too long).

> I noticed there's two main API styles: a "low-level" API (e.g. L0) where the Config struct has dividers and other low-level "knobs", and a "high-level" API (e.g. F0) where it has desired clock frequencies and the RCC implementation figures out how to achieve them. Which one is preferred?

It's due to historicla reasons, we should unify that someday. I somewhat prefer the low-level style too.

> F2 has a nasty errata: a delay or DSB instruction must be added after every RCC peripheral clock enable. I've added this workaround to build.rs, but am not sure if this is the best approach. Any comments?

Sounds good!

> I'm planning to add PLL support too once I know which kind of API is preferred. Would you prefer a separate pull request for that, or should I continue working on this one?

Smaller PRs is nicer, if individual ones are already useful for something, which this one is! :)

Gekkio:
> Thanks for the PR! Looks great overall, just one thing:
> 
>     * Could you add building the F2 examples, and also embassy-stm32 for a F2 chip to CI? (`ci.sh` and `ci_stable.sh` files). Preferably the "biggest" F2 chip, if it has more stuff it's more likely to fail if there's something wrong (unfortunately there's so many chips that making CI build everything would take way too long).

Done! At least I think so...let's see if the CI agrees.

> Smaller PRs is nicer, if individual ones are already useful for something, which this one is! :)

Cool, I'll continue in a separate PR :smile:

Dirbaio (approved):
Thank you!

bors r+

PR #679, opened 2022-03-27
Reexport unborrow macro in HALs

matoushybl:
Removes the need to depend on embassy-hal-common in the case of developing custom peripheral drivers.

### DIFF
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -115,6 +115,9 @@ pub use chip::pac;
 #[cfg(not(feature = "unstable-pac"))]
 pub(crate) use chip::pac;
 
+pub use embassy::util::Unborrow;
+pub use embassy_hal_common::unborrow;
+
 pub use chip::{peripherals, Peripherals};
 
 pub mod interrupt {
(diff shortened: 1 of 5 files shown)


embassy-nrf/src/lib.rs
```diff
@@ -115,6 +115,9 @@ pub use chip::pac;
 #[cfg(not(feature = "unstable-pac"))]
 pub(crate) use chip::pac;
 
+#[cfg(feature = "unstable-pac")]
```
Dirbaio, line 118: I don't think gating it behind `unstable-` is needed, it's part of the stable public API so it's not like we're going to change it.

matoushybl: I made it so because it cannot be used without accessing PAC. But you're right that it is stable and public.

Dirbaio:
Why just the macro, and not the `Unborrow` trait?

matoushybl:
I didn't think about exporting the trait itself as it is accessible directly using the embassy crate that everything depends on anyway, but it is a good idea to export the trait as well as the macro.

lulf:
I think the reexport makes sense. Any reason not to merge this?

matoushybl:
I don't think there is any reason not to merge it, it likely just got lost in the PRs. Re-requesting review now.

Dirbaio (approved):
bors r+

matoushybl:
Thanks, @Dirbaio!

PR #680, opened 2022-03-27
Add support for USB classes handling control requests.

alexmoon:
Added functionality to #657 to allow classes to handle their own EP0 control requests. This was needed for an HID class implementation I'm working on. It a fairly substantial change so review would be much appreciated.


Dirbaio:
Thank you! :)

PR #681, opened 2022-03-28
A couple async_usb_stack improvements.

alexmoon:
Removes UnsafeCell from cdc_acm::Control and avoids an extra memcpy in embassy_nrf::Usb when it's not necessary.


Dirbaio (approved):
Oooh the trick to remove the UnsafeCell is very interesting. It makes `State` a self-referential struct, I thought these were impossible in safe Rust :o

Thank you!

bors r+

PR #684, opened 2022-03-29
Async usb bug fixes

alexmoon:
With these changes I have a working HID class. 🎉

The HID `ControlHandler` needs to respond to some `RequestType::Standard` control requests, so I moved the handling for `GET_STATUS`, `GET_INTERFACE` and `SET_INTERFACE` into the `ControlHandler` with default implementations so we can route all control requests with a `Recipient::Interface` to the `ControlHandler`.

The `embassy-nrf::ControlPipe` implementation also needed some work to support control out requests with more than one data packet.

### DIFF
--- a/embassy-nrf/src/usb.rs
+++ b/embassy-nrf/src/usb.rs
@@ -508,6 +508,30 @@ pub struct ControlPipe<'d, T: Instance> {
 }
 
 impl<'d, T: Instance> ControlPipe<'d, T> {
+    async fn read(&mut self, buf: &mut [u8]) -> Result<usize, ReadError> {
+        let regs = T::regs();
+
+        // Wait until ready
+        regs.intenset.write(|w| w.ep0datadone().set());
+        poll_fn(|cx| {
+            EP_OUT_WAKERS[0].register(cx.waker());
+            let regs = T::regs();
+            if regs
+                .events_ep0datadone
+                .read()
+                .events_ep0datadone()
+                .bit_is_set()
+            {
+                Poll::Ready(())
+            } else {
+                Poll::Pending
+            }
+        })
+        .await;
+
+        unsafe { read_dma::<T>(0, buf) }
+    }
+
     async fn write(&mut self, buf: &[u8], last_chunk: bool) {
         let regs = T::regs();
         regs.events_ep0datadone.reset();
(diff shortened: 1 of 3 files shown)


embassy-nrf/src/usb.rs
```diff
@@ -595,29 +619,19 @@ impl<'d, T: Instance> driver::ControlPipe for ControlPipe<'d, T> {
...
-                    Poll::Ready(())
-                } else {
-                    Poll::Pending
+            let req_length = usize::from(req.length);
+            let max_packet_size = usize::from(self.max_packet_size);
+            let mut total = 0;
+            for chunk in buf.chunks_mut(max_packet_size) {
+                let size = self.read(chunk).await?;
+                total += size;
+                if size < max_packet_size || total == req_length {
```
Dirbaio, line 629: Is `total == req_length` needed? I think chunked transfers are always required to end in a packet smaller than `max_packet_size`, inserting a ZLP if it's an exact multiple.

embassy-nrf/src/usb.rs
```diff
@@ -595,29 +619,19 @@ impl<'d, T: Instance> driver::ControlPipe for ControlPipe<'d, T> {
...
-                if regs
-                    .events_ep0datadone
-                    .read()
-                    .events_ep0datadone()
-                    .bit_is_set()
-                {
-                    Poll::Ready(())
-                } else {
-                    Poll::Pending
+            let req_length = usize::from(req.length);
```
Dirbaio, line 623: I think this chunking could go in `embassy-usb` instead of in the driver. It's the same for all drivers and AFAIK no hardware does it for you. If we make it the driver's responsibility it'll end up copypasted in all the drivers.

Same applies to `accept_in` from a previous PR, so I'm OK with merging this here and moving later if you prefer.

embassy-nrf/src/usb.rs
Dirbaio: Oh huh, it's [not actually required](https://stackoverflow.com/questions/3739901/when-do-usb-hosts-require-a-zero-length-in-packet-at-the-end-of-a-control-read-t).

Dirbaio:
> With these changes I have a working HID class. :tada: 

Wow, that's awesome! Will you PR it? :eyes:  (also related, should USB class impls go in embassy-usb? or in one crate per class, like `embassy-usb-serial`, `embassy-usb-hid`?)

Thank you!

PR #685, opened 2022-03-29
Fix STM32 timer interrupt bug

chemicstry:
Previously timer overflow interrupts were not firing correctly, because Update Interrupt Enable (UIE) was not set.

The timers still worked somewhat correclty, because overflow was handled together with other interrupts.


Dirbaio (approved):
Thank you!

bors r+

PR #687, opened 2022-03-30
Refactor ControlPipe to use the typestate pattern

alexmoon:
This moves the `ControlPipe` struct into the `control` module and implements a typestate pattern to ensure the methods of `ControlPipe` are only called at valid times during the control transaction process. This eliminates the assertions used previously and moves the checks to compile time. Otherwise, functionality is unchanged.


Dirbaio (approved):
LGTM, thank you! :)

PR #688, opened 2022-03-31
Async usb hid

alexmoon:
Initial implementation of an HID class for the async USB stack


embassy-usb-hid/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+
+impl<'d, D: Driver<'d>, const N: usize> ReportReader<'d, D, N> {
+    pub async fn read(&mut self, buf: &mut [u8]) -> Result<usize, ReadError> {
+        assert!(buf.len() >= N);
+        if let Some(ep) = &mut self.ep_out {
+            let max_packet_size = usize::from(ep.info().max_packet_size);
+
+            let mut chunks = buf.chunks_mut(max_packet_size);
+
+            // Wait until we've received a chunk from the endpoint or a report from a SET_REPORT control request
```
Dirbaio, line 254: The logic here is quite complex. Is it actually needed?

The [HID specs](https://www.usb.org/document-library/device-class-definition-hid-111), page 10 say:

> If a device declares an Interrupt Out endpoint then Output reports are transmitted by the host to the device through the Interrupt Out endpoint. If no Interrupt Out endpoint is declared then Output reports are transmitted to a device through the Control endpoint, using Set_Report(Output) requests.

I guess this means that if we *do* have the INTERRUPT OUT pipe then reports MUST go through it, and we can ignore the control pipe. This'd simplify the code a lot. Hopefully it holds true in practice...

embassy-usb-hid/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+            report_descriptor,
+            request_handler,
+            ep_out,
+            ep_in,
+        )
+    }
+
+    /// Creates a new HidClass with the provided UsbBus & HID report descriptor.
+    /// See new() for more details.
+    pub fn new_ep_out(
```
Dirbaio, line 135: I think this is illegal? The [HID specs](https://www.usb.org/document-library/device-class-definition-hid-111), page 10 say the INTERRUPT IN endpoint is mandatory.

embassy-usb-hid/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// Set the idle rate for `id` to `dur`.
+    ///
+    /// If `id` is `None`, set the idle rate of all input reports to `dur`. If
+    /// an indefinite duration is requested, `dur` will be set to `Duration::MAX`.
+    fn set_idle(&self, id: Option<ReportId>, dur: Duration) {
+        let _ = (id, dur);
+    }
+}
+
+pub struct Control<'d, const OUT_N: usize> {
```
Dirbaio, line 324: this shouldn't be pub

embassy-usb-hid/src/lib.rs
alexmoon: Agreed, we should remove it.

embassy-usb-hid/src/lib.rs
alexmoon: Agreed.

embassy-usb-hid/src/lib.rs
alexmoon: I think that should be true of well behaved hosts, but I don't think the spec is quite that strict and I'm not confident all hosts will operate that way. Page 55 of the spec seems looser, for example:

> Reports may also be requested (polled) and sent through the Control pipe or sent through an optional Interrupt Out pipe.

I'd rather err on the side of making the device work consistently regardless of how the host behaves.

Dirbaio: Hmm... `usbd-hid` outright rejects all control requests and it seems to work for people... :joy: 

Another problem with handling reports from the control pipe is it needs an extra `[u8; OUT_N]` of RAM for the signal, if we could get rid of that it'd be nice...

TinyUSB seems to [handle](https://github.com/hathach/tinyusb/blob/9c8c5c1c53214b8954bb28c33a496b5b423d5ecc/src/class/hid/hid_device.c#L326) [both](https://github.com/hathach/tinyusb/blob/9c8c5c1c53214b8954bb28c33a496b5b423d5ecc/src/class/hid/hid_device.c#L410) though, but maybe it's just to support the case of not having the OUT endpoint?

Not sure what's best... Perhaps we can check whether major OSs behave properly, then do the simplification if it is the case?

embassy-usb-hid/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+impl<'a, const IN_N: usize, const OUT_N: usize> State<'a, IN_N, OUT_N> {
+    pub fn new() -> Self {
+        State {
+            control: MaybeUninit::uninit(),
+            out_signal: Signal::new(),
+        }
+    }
+}
+
+pub struct HidClass<'d, D: Driver<'d>, const IN_N: usize, const OUT_N: usize> {
```
Dirbaio, line 77: The `IN_N` generic param is unused, it could be removed?

alexmoon: It's also used to allocate a buffer on the stack for serializing input reports.

embassy-usb-hid/src/lib.rs
alexmoon: It would be awfully nice if we could count on that. My fear is that some obscure host or an application using libusb directly or something would cause a device to suddenly break. The simplest solution is to always have SET_REPORT control requests route to `RequestHandler::set_report()` and OUT endpoint reports route to `HidClass::output()`. But I expect that will mean device developers who allocate an OUT endpoint won't handle SET_REPORT requests for output reports in `RequestHandler::set_report()` and if there are bad hosts those devices will break in the field. I want to think about it some more to see if I can come up with a low to zero overhead way of merging the streams.

embassy-usb-hid/src/lib.rs
Dirbaio: oh, I was grepping for `IN_N` and didn't see it's called `N` in `ReportWriter`. Oops :rofl:

alexmoon:
Ok, I've reworked the output report handling. I reduced the memory overhead to a constant size (a few pointers) and reduced the number of memcpy's from a nominal 4 (though probably optimized to 2) down to 1. I did have to add a couple of things:

* A EndpointOut::wait_for_data_ready() method
* A new synchronization primitive (AsyncLease) to allow one task to share a buffer with another

Have a look and let me know what you think.

Thanks.

Dirbaio:
I'm afraid your new is even more complex, and I think it still doesn't solve the problem...

- The complexity leaks out, via the `EndpointOut::wait_for_data_ready()` method, which now all drivers will have to implement.
- Incoming reports on the control pipe get lost if the HID task is not waiting on it *right now*. This can be the case if the code is waiting [here](https://github.com/embassy-rs/embassy/blob/acfe3707a8921020d00176f074d20d1b2b88bd12/embassy-usb-hid/src/lib.rs#L233), or if the user is doing other async stuff between `read` calls, or if the USB task runs at higher prio than the HID task.
- `AsyncLease` is unsound if you leak the `AsyncLeaseFuture` (with `mem::forget`, `ManuallyDrop`, or others, which are all safe) because it allows freeing the buffer without having drop de-register it. This is not an absolute blocker since we already have this issue for all futures that do DMA (you can leak them then deallocate the buffer, DMA will keep writing to it). However, it's completely impossible in today's Rust to do DMA like that soundly, while this case can be made sound with a different design, so I'd like to avoid this.

What if instead we make all the "incoming reports" API callback-based?

- `ControlHandler` calls `RequestHandler::set_report()` on incoming reports over the control pipe.
- `ReportReader` has an `async fn run()` that runs forever. It calls `RequestHandler::set_report()` on incoming reports over the interrupt pipe.

This solves all the buffer issues I think. Also, IMO it's more consistent to have *only* callbacks, vs currently we have both callbacks and `async fn read()`. 

Callbacks are somewhat less flexible because you can't `.await` in them, but I think they work fine for the majority of use cases. Most of the time handling incoming reports will be just seting some variables, or turning on/off the capslock LED or something, which usually doesn't require awaiting. If the user does need async stuff they can send the reports to their own async task via a channel/signal on their own.

embassy-usb-hid/src/lib.rs
Dirbaio: I did some more reading, it seems you're right, and we can't rely on reports always coming on the interrupt pipe :( Linux has APIs to explicitly send stuff on the control endpoint, for example.

I've commented in the main thread an alternative that I think could work though :)

alexmoon:
> I'm afraid this new solution is even more complex, and I think it still doesn't solve the problem...

You're not wrong. I wanted to see what could be done in the way of a low-overhead abstraction, but it does feel overly complex and fragile.

>     * The complexity leaks out, via the `EndpointOut::wait_for_data_ready()` method, which now all drivers will have to implement.

I do think this could be a useful addition in any event, along with a similar method for `EndpointIn`. Being able to separate the waiting from the reading/writing can be useful for buffer lifetime management.

>     * Incoming reports on the control pipe get lost if the HID task is not waiting on it _right now_. This can be the case if the code is waiting [here](https://github.com/embassy-rs/embassy/blob/acfe3707a8921020d00176f074d20d1b2b88bd12/embassy-usb-hid/src/lib.rs#L233), or if the user is doing other async stuff between `read` calls, or if the USB task runs at higher prio than the HID task.

Yeah, this is a big negative.

>     * `AsyncLease` is unsound if you leak the `AsyncLeaseFuture` (with `mem::forget`, `ManuallyDrop`, or others, which are all safe) because it allows freeing the buffer without having drop de-register it. This is not an absolute blocker since we already have this issue for all futures that do DMA (you can leak them then deallocate the buffer, DMA will keep writing to it). However, it's completely impossible in today's Rust to do DMA like that soundly, while this case can be made sound with a different design, so I'd like to avoid this.

True.

> What if instead we make all the "incoming reports" API callback-based?

I don't want to force users into the callback API, because it adds the same memory overhead as the original signal solution, plus the overhead of an additional task and means output reports have to be handled in a sync, shared-reference context which in many cases would means additional synchronization overhead.

That said, I agree that it's the simplest consistent solution that covers all the bases.

Therefore, I think we should give the crate user the option. Either they can start the `run()` task to get consistent output report handling, or they can use `read()` to get output reports from the interrupt pipe and take responsibility for synchronizing handling of control request reports themselves.

Dirbaio:
> it adds the same memory overhead as the original signal solution

I'm not so sure. With `.read()` the user has to allocate a buffer, so the previous solution required 2 buffers: that one and the one in the Signal for control reqs. Now we allocate one in `.run()`, but control reqs reuse the usb stack's control buffer, so in total it requires just 1 buffer.

> plus the overhead of an additional task

Should be the same amount of tasks, the task that previously did `.read()` now does `.run()`.

I like the idea about offering both `.read()` and `.run()`! Let's merge, we can polish more later.

Dirbaio (approved):
bors r+

Dirbaio:
CI fails because Cargo namespaced features are not available in stable, but they will be in 1.60 which is out in 2 days. Manually merging.

PR #692, opened 2022-04-02
Add basic device state handling for endpoints.

alexmoon:
For protocols where the endpoint stream is stateful, it will be necessary for endpoint tasks to handle bus reset events. We should also avoid enabling endpoints until the host has set the device configuration.

This PR adds a `set_configured()` method to the `driver::Bus` trait to allow the driver to enable/disable endpoints in response to SET_CONFIGURATION control requests.

It also adds a `Disabled` error to both `WriteError` and `ReadError` which should be returned from `EndpointIn::write()` or `EndpointOut::read()` if the endpoint is disabled during the call.

Finally, it adds an `Endpoint::wait_enabled()` method to allow clients to wait for the device to be configured and the endpoint(s) enabled.

### DIFF
--- a/embassy-nrf/src/usb.rs
+++ b/embassy-nrf/src/usb.rs
@@ -17,6 +17,7 @@ use futures::future::poll_fn;
 use futures::Future;
 
 pub use embassy_usb;
+use pac::usbd::RegisterBlock;
 
 use crate::interrupt::Interrupt;
 use crate::pac;
@@ -92,11 +93,11 @@ impl<'d, T: Instance> Driver<'d, T> {
             regs.epdatastatus.write(|w| unsafe { w.bits(r) });
             READY_ENDPOINTS.fetch_or(r, Ordering::AcqRel);
             for i in 1..=7 {
-                if r & (1 << i) != 0 {
-                    EP_IN_WAKERS[i - 1].wake();
+                if r & In::mask(i) != 0 {
+                    In::waker(i).wake();
                 }
-                if r & (1 << (i + 16)) != 0 {
-                    EP_OUT_WAKERS[i - 1].wake();
+                if r & Out::mask(i) != 0 {
+                    Out::waker(i).wake();
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


Dirbaio:
What's the use case for `wait_enabled`? I would really really prefer to keep the driver API surface as minimal as possible.

alexmoon:
In my device I need to react to changes in the connection state (i.e. whether the configuration is set or not) in my endpoint handling task. So the first thing I need is for `read()`/`write()` to abort with an error if the endpoints are disabled. That also gives me an opportunity to reset any state I have associated with the endpoint when it is disconnected. Otherwise, state could carry over between connections which would be bad.

Once `read()`/`write()` start erroring when the endpoint is disabled, there needs to be some signal for when it is safe to start using the endpoint. Another option would be to add a `RequestHandler::set_configured()` callback. I think that would work but it seems like the wrong layer of abstraction to me. What you care about for the endpoint is whether it is enabled or not, and the configuration being set is only an indirect indication of that state. It would also be insufficient if the stack ever supports manually enabling or disabling endpoints at runtime (and would get more complicated if multiple configurations were supported).

The other problem I see with a `set_configured()` callback is that it would require each class to essentially create its own version of `wait_enabled()` using a `Signal` from the `RequestHandler` or similar. Otherwise, there's no way for the application to know when the class is ready to start sending/receiving data.

You can see how `wait_enabled` works in the usb-serial example. With this change, it's easy and natural for it to report connected/disconnected events.

Dirbaio:
I was thinking whether it'd be possible to specify read/write wait until EP enabled, instead of erroring out. This might make usb-serial easier to use, because you can pretend the serial port is "always there", just not receiving/sending data if disconnected.

I guess in other cases you do want writing to fail immediately though... Like in HID, ignore keypresses, instead of hanging forever and then sending a super-stale keypress to the host when you're finally plugged in?

alexmoon:
> I guess in other cases you do want writing to fail immediately though... Like in HID, ignore keypresses, instead of hanging forever and then sending a super-stale keypress to the host when you're finally plugged in?

Yeah, that's right. There can also be protocol state that spans packets that needs to get reset. If you're waiting for a packet when the connection is reset that state can get out of sync unless the read/write returns an error.

Dirbaio:
Sounds good then! :+1: :+1: 

Also if you want the serial port to look like it's "always there", the class can do that on top of `read/wait_enabled`.

PR #693, opened 2022-04-04
no_std version of `futures::future::select_all`

alexmoon:
Here's a no-std compatible version of `select_all`. It's not quite as useful as the original because it requires an array of Unpin futures to be pre-constructed instead of taking an iterator (which could return `Pin<Box<_>>` in `std`). And, of course, you don't get a `Vec` of the unfinished futures returned at completion. Still, I think it's cleaner than a long cons of select calls.

I'll leave it up to you whether this is sufficiently general purpose to include in Embassy or not.

### DIFF
--- a/embassy/src/util/select_all.rs
+++ b/embassy/src/util/select_all.rs
@@ -1,10 +1,6 @@
-#![allow(dead_code)]
-
 use core::future::Future;
 use core::pin::Pin;
 use core::task::{Context, Poll};
-
-use futures::future::FutureExt;
 
 /// Future for the [`select_all`] function.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


embassy/src/util/select_all.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Creates a new future which will select over a list of futures.
+///
+/// The returned future will wait for any future within `iter` to be ready. Upon
+/// completion the item resolved will be returned, along with the index of the
+/// future that was ready.
+///
+/// # Panics
+///
+/// This function will panic if the array specified contains no items.
+pub fn select_all<Fut: Future + Unpin, const N: usize>(arr: [Fut; N]) -> Option<SelectAll<Fut, N>> {
```
Dirbaio, line 27: You can make this not require Unpin, which makes it much easier to use (no `pin_mut!()` required!). 

(`.into_inner()` requires Unpin, maybe you can make just that one require Unpin.

alexmoon: You're right, that's much better. I think the futures version requires `Unpin` because it uses a `Vec` so the futures get moved around.

I'd appreciate a review of the safety in the `poll()` implementation, now that it has to use unsafe.

Dirbaio (approved):
Thank you! :)

PR #694, opened 2022-04-05
Add select, select3, select4.

Dirbaio:
The difference with those from the `futures` crate is they don't return the other partially-run futures, so they can work with `!Unpin` futures which makes them much easier to use.


Dirbaio:
No comments on the `mod util` in7 days. Susprising but I'll take it :see_no_evil: 

bors r+

PR #695, opened 2022-04-05
Simplify Channel.

Dirbaio:
- Allow initializing in a static, without Forever.
- Remove ability to close, since in embedded enviromnents channels usually live forever and don't get closed.
- Remove MPSC restriction, it's MPMC now. Rename "mpsc" to "channel".
- `Sender` and `Receiver` are still available if you want to enforce a piece of code only has send/receive access, but are optional: you can send/receive directly into the Channel if you want.


huntc:
I think it is confusing to have two types of channel. Perhaps we should remove the old one.

huntc:
Just to add, when I introduced the original mpsc, I weighed the importance of being API aligned with what people may be used to coming from std. Having now full developed a production based thing with Embassy, closing the channel hasn’t been important. Can we just modify the existing MPSC to avoid closing?

Dirbaio:
@huntc I thought you added close functionality because you found it useful? I haven't used it myself, but I can imagine it could be handy in some circumstances? especially in not "long-lived" channels. 

Maybe renaming the other one to "CloseableChannel" can help avoid confusion.

Dirbaio:
As [discussed on Matrix](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$F8_xeWql3SPoEO8YIKJra2PYV_8kwYwpk1znuVDisf4?via=matrix.org&via=converser.eu&via=braun-odw.eu), keeping 2 kinds of channels doesn't seem worth it. Changed the PR to update the existing channel instead of adding a new one.

embassy/src/channel/channel.rs
```diff
@@ -0,0 +1,427 @@
...
+    ///
+    /// // Declare a bounded channel of 3 u32s.
+    /// let mut channel = Channel::<NoopRawMutex, u32, 3>::new();
+    /// ```
+    #[cfg(not(feature = "nightly"))]
+    pub fn new() -> Self {
+        Self {
+            inner: Mutex::new(RefCell::new(ChannelState::new())),
+        }
+    }
```
huntc, line 259: Is the diff between these constructors simply the import?

embassy/src/channel/channel.rs
```diff
@@ -0,0 +1,427 @@
...
+//! This queue takes a Mutex type so that various
+//! targets can be attained. For example, a ThreadModeMutex can be used
+//! for single-core Cortex-M targets where messages are only passed
+//! between tasks running in thread mode. Similarly, a CriticalSectionMutex
+//! can also be used for single-core targets where messages are to be
+//! passed from exception mode e.g. out of an interrupt handler.
+//!
+//! This module provides a bounded channel that has a limit on the number of
+//! messages that it can store, and if this limit is reached, trying to send
+//! another message will result in an error being returned.
```
huntc, line 12: Mention that this is an MPMC for the purposes of familiarity for some?

examples/nrf/src/bin/channel.rs
```diff
@@ -0,0 +1,45 @@
...
+
+enum LedState {
+    On,
+    Off,
+}
+
+static CHANNEL: Channel<ThreadModeRawMutex, LedState, 1> = Channel::new();
+
+#[embassy::task]
+async fn my_task() {
```
huntc, line 24: I realise that this example doesn't need it, but I think it'd be great to show the Sender being passed in as a param to the task as that is commonly how it will be passed around.

huntc (approved):

examples/nrf/src/bin/channel.rs
Dirbaio: > that is commonly how it will be passed around.

I'm not so sure! Using the `static` directly is much easier and might save a tiny bit of RAM (no need to pass around a pointer if you can just go and get the static directly).

One advantage for using Sender/Receiver is that you can use NoopMutex instead. I'll add another example!

embassy/src/channel/channel.rs
Dirbaio: In nightly it can be `const`, in stable it can't because it requires `const_fn_trait_bounds`. Rust 1.61 will stabilize it, so we'll be able to make it const in stable.

examples/nrf/src/bin/channel.rs
huntc: My tasks are often in their own modules. Here's a declaration of one for processing analog inputs for the existing MPSC:

```rust
#[embassy::task(pool_size = 1)]
pub async fn main_task(
    mut p: AnalogInputPeripherals,
    line_flow_signal_0: &'static Signal<line_monitor::LineFlow>,
    line_flow_signal_1: &'static Signal<line_monitor::LineFlow>,
    sender: Sender<'static, NoopRawMutex, state::Command, MAX_COMMANDS_PENDING>,
)
```

While the signal is static, I still find it useful to pass in as a param as it feels clear as to what the task's requirements are.

embassy/src/channel/channel.rs
Dirbaio: Done.

examples/nrf/src/bin/channel.rs
Dirbaio: Added another example

huntc:
Wonderful! Thanks!

rise0chen:
@Dirbaio  I have a crate about atomic channel.
It is `static friendly`, `no alloc`, `lock free`.
[ach](https://github.com/rise0chen/ach/tree/main/ach)
[async_ach](https://github.com/rise0chen/async_ach)

PR #696, opened 2022-04-05
Add async Mutex.

Dirbaio:
What it says on the tin :) 

It allows sharing data between tasks when you want to `.await` stuff while holding it locked.


lulf (approved):
LGTM 🚀

PR #697, opened 2022-04-06
embassy-usb-hid bug fixes

alexmoon:
This adds `ready()` methods to `ReportReader` and `ReportWriter` to wait for their respective endpoints to be enabled. It also adds a `use_report_ids` argument to `ReportReader::run()` to control whether the first byte of reports should be used as the report ID passed to the callback or not. This is needed for full consistency with control request reports.

Finally, it includes a couple of robustness improvements to `ReportReader`:

First, it properly resets the report offset on bus reset or a disabled error.

Second, it now handles superfluous zero-length-packets. MacOS appears to send zero-length-packets even when the report is maximum size. Even worse it can interleave the ZLPs with data packets (e.g. REPORT REPORT ZLP ZLP). This is now handled properly.

### DIFF
--- a/embassy-usb-hid/src/lib.rs
+++ b/embassy-usb-hid/src/lib.rs
@@ -9,6 +9,7 @@ pub(crate) mod fmt;
 
 use core::mem::MaybeUninit;
 use core::ops::Range;
+use core::sync::atomic::{AtomicUsize, Ordering};
 
 use embassy::time::Duration;
 use embassy_usb::driver::EndpointOut;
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
LGTM, thank you! :) 

wow, the dechunking logic is getting quite hairy

alexmoon:
> wow, the dechunking logic is getting quite hairy

Yeah, it would be so much easier if everyone just followed the spec!

PR #699, opened 2022-04-06
Add ADC support for H7

matoushybl:
Seems to work on H743, ADC1 works fine, but ADC3 has different ADC_COMMON block which stm32-data does't yet support, so measuring VREFINT or TEMP doesn't work. I'd like to discuss how adding support for the different ADC_COMMON could be done.


Dirbaio:
LGTM overall, could you rebase and fix CI? Thanks!

embassy-stm32/src/rcc/h7.rs
```diff
@@ -614,6 +639,16 @@ pub(crate) unsafe fn init(mut config: Config) {
     // Peripheral Clock (per_ck)
     RCC.d1ccipr().modify(|w| w.set_ckpersel(ckpersel));
 
+    // ADC clock MUX
+    RCC.d3ccipr()
```
Dirbaio, line 643: we'll eventually need some crate-wide infra to manage all the stuf fin CCIPRx :S

Dirbaio (approved):
Thank you!

bors r+

PR #700, opened 2022-04-07
Add back support for cloning sender/receiver

lulf:
The automatic derive clone does not work because RawMutex is not Clone.


Dirbaio (approved):
Oh, whoops. 

Thank you! 

bors r+

PR #701, opened 2022-04-07
Async-ify Driver::enable and UsbDeviceBuilder::build

alexmoon:
As discussed, this prevents the 1ms blocking wait in the nrf USB driver.

I also found an issue with `ControlPipe::accept_in` returning an error when a setup packet is received immediately after `data_in` completes (before the future is polled). That's now fixed.


Dirbaio (approved):
Looks great, thank you! :)

PR #708, opened 2022-04-08
Unify ReadError and WriteError into EndpointError

Dirbaio:
Originally #698, got overwritten in a rebase I think.


Dirbaio:
spurious HIL test fail? 

bors r+

PR #709, opened 2022-04-09
embassy-usb stm32

Dirbaio:
TODO

- [x] Add cfgs for EP_COUNT, EP_MEMORY_ADDR. -- ended up doing it in stm32-data
- [x] Auto-enable PWR USV in stm32l5,
- [x] Auto-enable PWR on other families that need it -- didn't find any
- [ ] ~~Assert at runtime that RCC has valid 48Mhz clock?~~ -- won't do it yet, too messy without rcc cleanup
- [x] Tested on STM32L5
- [x] Tested on STM32F303ZE
- [x] Tested on STM32F103c8
- [x] Fix duplicated irqs on STM32F303 in stm32data, remove workaround in embassy-stm32
- [x] Test HID example
- [x] Test NCM example


votrungchi:
Hello, I am facing the unstable problem with Feature Out on STM32F407VG. The data received from the host is incorrect, unstable, and will panicked after some request. If you interesting on it, I will provide the detail information.

PR #710, opened 2022-04-09
add stm32l4 hsi48 and usb example

ant32:


Dirbaio (approved):
Thank you!

Funny coincidence, I was just now trying to get USB working on L5, and your PR solved it all, both the HSI48 and the PWR bit. :joy:  (L5 is very very similar to the L4)

bors r+

PR #711, opened 2022-04-10
Add DeviceStateHandler, DeviceCommand channel, and remote wakeup support

alexmoon:
Apologies for the size of this PR. Once I started getting into the Vbus power management side of my device I found a couple of areas of functionality missing from embassy-usb. Specifically, I need the application to be able to respond to changes in the USB device state in order to properly control the amount of power I'm drawing from Vbus. I also wanted to enable remote wakeup support for my device.

In order to enable device state monitoring, I've created a `DeviceStateHandler` trait and made it possible to pass in an optional reference a handler implementing that trait when creating the `UsbDeviceBuilder`.

Remote wakeup required a way to send commands to the bus which is exclusively owned by the `UsbDevice::run` method. This is the same problem we were discussing for enabling/disabling the device on Vbus power events. My solution is to allow an optional `Channel` to be provided to the `UsbDeviceBuilder` (via `UsbDeviceBuilder::new_with_channel`), allowing the application to send commands into the `run` method. Right now it supports enable, disable and remote wakeup commands.

Since there's now a way to dynamically enable and disable the device, I also added `Config::start_enabled` to control whether or not the `UsbDevice` should start in the enabled state. That also allowed me to make `UsbDeviceBuilder::build` sync again and move enabling the bus into `UsbDevice::run`.

This led to a few driver changes:
1. `Driver::enable` became `Driver::into_bus`
2. `Bus::enable`, `Bus::disable`, and `Bus::remote_wakeup` were added
3. I removed `Bus::reset`, `Bus::suspend`, and `Bus::resume` because they were only ever called based on the result of `Bus::poll`. It made more sense to me to have `Bus::poll` handle the driver-specific state management itself.

I've updated the `usb_hid_keyboard` example to take advantage of all these additional features.

Let me know what you think.

Thanks!


embassy-usb/src/driver.rs
```diff
@@ -57,18 +56,24 @@ pub trait Driver<'a> {
...
     where
         Self: 'a;
 
-    fn poll<'a>(&'a mut self) -> Self::PollFuture<'a>;
+    /// Enables the USB peripheral. Soon after enabling the device will be reset, so
+    /// there is no need to perform a USB reset in this method.
+    fn enable(&mut self) -> Self::EnableFuture<'_>;
+
+    /// Disables and powers down the USB peripheral.
+    fn disable(&mut self);
```
Dirbaio, line 74: disable() should be async as well, for consistency

embassy-usb/src/driver.rs
```diff
@@ -106,6 +100,16 @@ pub trait Bus {
     fn force_reset(&mut self) -> Result<(), Unsupported> {
         Err(Unsupported)
     }
+
+    /// Initiates a remote wakeup of the host by the device.
+    ///
+    /// The default implementation just returns `Unsupported`.
```
Dirbaio, line 106: there's no default impl

alexmoon: oops!  😂

embassy-usb/src/driver.rs
alexmoon: sure

embassy-usb/src/driver.rs
alexmoon: because `remote_wakeup` is async, a default impl is not possible, so I've just removed the comment.

Dirbaio:
I'm not convinced by the use of the `Channel`.
- There's not much precedent for APIs that you use by speaking into a channel. I think they're nice for internally synchronizing betewen different tasks, but for APIs it's exposing implementation details of "how" are you syncing stuff.
- It needs the Mutex generic, and I'd like to keep generics to a minimum, Also, if you pass `None` as a channel, you still have to specify the generic, Rust won't be able to infer it.

Maybe there's another way? Perhaps the user can `select` on `bus.run()` and other stuff, so when it's time to disable the bus, `run()` gets canceled and now you gain back access to the UsbDevice so you can do other stuff to it?

What happened with the other "drop stuff to disable, recreate later again" idea? It didn't work out in the end?

embassy-usb/src/lib.rs
```diff
@@ -64,82 +99,147 @@ pub struct UsbDevice<'d, D: Driver<'d>> {
...
+        }
+
         loop {
             let control_fut = self.control.setup();
             let bus_fut = self.bus.poll();
-            match select(bus_fut, control_fut).await {
-                Either::Left(evt) => match evt {
+            let commands_fut = recv_or_wait(self.commands);
+
+            match select3(bus_fut, control_fut, commands_fut).await {
```
Dirbaio, line 161: Could you switch to using #694 ?

alexmoon: sure

alexmoon:
> I'm not convinced by the use of the `Channel`.

I'm not a big fan either, but it's the best I could come up with so far.

>     * It needs the Mutex generic, and I'd like to keep generics to a minimum, Also, if you pass `None` as a channel, you still have to specify the generic, Rust won't be able to infer it.

I solved that with separate constructors (`UsbDeviceBuilder::new` vs `UsbDeviceBuilder::new_with_channel`). I also gave the `UsbDevice` a default mutex type so if you're creating a static and don't use a channel you don't have to specify it. You can see in the examples I updated (except for usb-hid-keyboard which uses the new channel) that the impact on code not using the channel is pretty much zero (the added `None`s are for the `DeviceStateHandler` not the channel).

> Maybe there's another way? Perhaps the user can `select` on `bus.run()` and other stuff, so when it's time to disable the bus, `run()` gets canceled and now you gain back access to the UsbDevice so you can do other stuff to it?

The problem there is that `run()` is not cancel-safe, so if it's dropped in the middle of processing a control request, for example, the bus state could get messed up.

> What happened with the other "drop stuff to disable, recreate later again" idea? It didn't work out in the end?

It worked fine for disabling/enabling the whole bus, but it doesn't work for triggering a remote wakeup. The same will be true for a forced reset when we get to that.

alexmoon:
Ok, I think I have an idea. It will take me a little while to prove it out. But I think I can make `UsbDevice::run` return a custom future type. I can then implement `enable`, `remote_wakeup`, etc as methods on that future. So you can `pin_mut` the future from `run`, start it running, then interrupt it and make those calls on the future itself, then resume by `await`ing the future again.

Dirbaio:
"custom" futures that you can either await or call custom methods on are quite confusing. I'd rather stick to a "standard" API with only normal `async fn`s that all you do with is `.await`.

> The problem there is that run() is not cancel-safe, so if it's dropped in the middle of processing a control request, for example, the bus state could get messed up.

Does that matter in practice though? If you're canceling `run()` to disdable or force-reset the bus, all the state gets wiped clean anyway. We could document "Canceling messes up the state, only cancel if you're going to disable or reset". 

Maybe `remote_wakeup` is trickier. Just to confirm: you can only do a remote wakeup while suspended, right? not whlie resumed or disabled?. Maybe having `run()` voluntarily return when suspended could work? so you can then wait on a `.wait_resumed()` fn (which would be cancel-safe) or do a `.remote_wakeup()`.

alexmoon:
Ok, I've given that a shot. I reviewed the await points in the `run` method and I think the only thing that's potentially problematic is if the future is dropped in the middle of handling a control request. What I've done for right now is set a flag while control request handling is in progress and then if `run` is called again with that flag set the control request will be rejected to get the bus back into the right state. I'm not 100% sure that's the correct action to take, but it seems reasonable.

Having `run` return voluntarily is a good idea and would definitely resolve the problem. The down side is that it would require all users of `run` to look like:

```rust
loop {
  usb.run().await;
  usb.wait_resumed().await;
}
```

which is a little more verbose than the current default. I'll leave it up to you which approach you prefer.

Dirbaio:
> The down side is that it would require all users of run to look like

True. Maybe `run()` would run forever, then for advanced users we can do a `run_until_suspended()`? I think there are still cases where you want to cancel `run_until_suspended()` though, so this doesn't solve all issues.

Dirbaio:
> What I've done for right now is set a flag while control request handling is in progress and then if run is called again with that flag set the control request will be rejected to get the bus back into the right state. I'm not 100% sure that's the correct action to take, but it seems reasonable.

I think it could be outright ignored? The host will retry the transfer if it needs to.

alexmoon:
> > The down side is that it would require all users of run to look like
> 
> True. Maybe `run()` would run forever, then for advanced users we can do a `run_until_suspended()`? I think there are still cases where you want to cancel `run_until_suspended()` though, so this doesn't solve all issues.

Good idea. I've implemented it.

> > What I've done for right now is set a flag while control request handling is in progress and then if run is called again with that flag set the control request will be rejected to get the bus back into the right state. I'm not 100% sure that's the correct action to take, but it seems reasonable.
> 
> I think it could be outright ignored? The host will retry the transfer if it needs to.

With `run_until_suspend` I'm comfortable doing that. I've put a note that dropping the run futures may leave the bus in an inconsistent state just to be safe.

alexmoon:
I've run some tests and it appears that if you drop the `run` future in the middle of a control request it will cause the bus to timeout for 5 seconds (at least on my system). It does eventually recover, but I think the warnings are appropriate.

Dirbaio:
That's if you don't reset/disable after, right? if you do then the host notices and it immediately recovers when you reenable?

alexmoon:
> 



> That's if you don't reset/disable after, right? if you do then the host notices and it immediately recovers when you reenable?

Correct.

Dirbaio (approved):
LGTM, very nice! THank you! :) 

bors r+

PR #712, opened 2022-04-11
Add types for channel dynamic dispatch

lulf:
* Add internal DynamicChannel trait implemented by Channel that allows
  polling for internal state in a lock safe manner and does not require
  knowing the channel size.
* Existing usage of Sender and Receiver is preserved and does not use
  dynamic dispatch.
* Add DynamicSender and DynamicReceiver types that references the
  channel using the DynamicChannel trait and does not require the const
  generic channel size parameter.

Having the ability not know the channel size is very convenient when you don't want to change all of your channel using code when tuning the size. With this change, existing usage can be kept, and those willing to pay the price for dynamic dispatch may do so.

### DIFF
--- a/embassy/src/channel/channel.rs
+++ b/embassy/src/channel/channel.rs
@@ -68,17 +68,11 @@ where
 
 /// Send-only access to a [`Channel`] without knowing channel size.
 #[derive(Copy)]
-pub struct DynamicSender<'ch, M, T>
-where
-    M: RawMutex,
-{
-    channel: &'ch dyn DynamicChannel<M, T>,
+pub struct DynamicSender<'ch, T> {
+    channel: &'ch dyn DynamicChannel<T>,
 }
 
-impl<'ch, M, T> Clone for DynamicSender<'ch, M, T>
-where
-    M: RawMutex,
-{
+impl<'ch, T> Clone for DynamicSender<'ch, T> {
     fn clone(&self) -> Self {
         DynamicSender {
             channel: self.channel,
@@ -86,7 +80,7 @@ where
     }
 }
 
-impl<'ch, M, T, const N: usize> From<Sender<'ch, M, T, N>> for DynamicSender<'ch, M, T>
+impl<'ch, M, T, const N: usize> From<Sender<'ch, M, T, N>> for DynamicSender<'ch, T>
 where
     M: RawMutex,
 {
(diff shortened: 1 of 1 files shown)


huntc:
> Having the ability not know the channel size is very convenient when you don't want to change all of your channel using code when tuning the size. With this change, existing usage can be kept, and those willing to pay the price for dynamic dispatch may do so.

Doesn’t a simple type alias also achieve the same goal ie a type alias requires a change in just one place…?

lulf:
@huntc If you control the entire code base, yes. But if you expose an API in in one crate, and want consumers of your API to not consider the queue size, I believe this is a much nicer way. Using Channel in in Drogue Actor, the library can now expose an Address handle for the actor that doesn't need to have the channel size as a type parameter. Then I can use this address handle in another library that doesn't need to care about the queue size etc.

lulf:
For the 'try' synchronous operations of the channel, I was able to work around this by adding a 'wrapper trait'. However, for async and GAT, the trait would need to have an associated type, which in turn requires specifying the associated types whenever using the trait, so that approach does not work.

Having the functionality in the Channel impl itself is the better way IMHO.

lulf:
Some performance numbers


## Static on master

#### Size:
    Finished release [optimized] target(s) in 28.68s
  13208     104    1444   14756    39a4 channel_sender_receiver

#### Stack:
(HOST) INFO  reading 254.49 KiB of RAM for stack usage estimation
(HOST) INFO  program has used at least 0.21/254.49 KiB (0.1%) of stack space


## Static on branch (should be the same as on master)

#### Size:
    Finished release [optimized] target(s) in 29.89s
  13208     104    1444   14756    39a4 channel_sender_receiver

#### Stack:
(HOST) INFO  reading 254.49 KiB of RAM for stack usage estimation
(HOST) INFO  program has used at least 0.21/254.49 KiB (0.1%) of stack space



## Dynamic on branch

#### Size:
   Compiling embassy-nrf-examples v0.1.0 (/home/lulf/dev/embassy/examples/nrf)
    Finished release [optimized] target(s) in 0.30s
  13244     104    1468   14816    39e0 channel_sender_receiver_dyn


#### Stack:
(HOST) INFO  reading 254.46 KiB of RAM for stack usage estimation
(HOST) INFO  program has used at least 0.22/254.46 KiB (0.1%) of stack space

Dirbaio:
let's do this :rocket: 

bors r+

PR #714, opened 2022-04-12
add more clock options for l4 and l5

ant32:
- added an assert so it panics if pll48div is not 48Mhz
- added MSI as a clock source for PLL
- removed hsi48 option for MCUs mentioned in l4 rcc presentation
- copied some code from l4 to l5, but don't have a way of testing it.

### DIFF
--- a/embassy-stm32/src/rcc/l4.rs
+++ b/embassy-stm32/src/rcc/l4.rs
@@ -96,6 +96,7 @@ pub enum APBPrescaler {
 pub enum PLLSource {
     HSI16,
     HSE(Hertz),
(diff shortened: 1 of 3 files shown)


ant32:
I see now that there is a PR open for some of the clock stuff for l5. If you want I can remove it from this PR

Dirbaio (approved):
Tested working on L5. USB works with either HSI48, PLL from HSI16, or PLL from MSI.

> I see now that there is a PR open for some of the clock stuff for l5. If you want I can remove it from this PR

Don't worry, yours is better, I'll rebase!

Thank you!

bors r+

PR #715, opened 2022-04-12
stm32/dac: Check proper channel

michalsrb:
Small fix. Otherwise it panics when trying to use channel 1 if channel 2 does not exist.


Dirbaio (approved):
Nice catch, thank you!

bors r+

PR #716, opened 2022-04-12
Implement giant (chunked) DMA transfers

matoushybl:
There is a breaking change in the Channel trait in the method `remaining_transfers()`, this method however hasn't been used anywhere in the codebase and the return type changed from u16 to u32, so all of the previous use cases should work.


embassy-stm32/src/dma/dma.rs
```diff
@@ -56,6 +57,93 @@ impl State {
 
 static STATE: State = State::new();
 
+struct M0AR(u32);
+struct M1AR(u32);
+struct ChunkSize(u16);
+
+struct GiantTransferState {
+    enabled: [AtomicBool; DMA_CHANNEL_COUNT],
```
Dirbaio, line 65: this would be more readable with a `ChannelState` struct with all the stuff for one channel, and then a have a `[ChannelState; DMA_CHANNEL_COUNT]`. Also fold it with the old STATE.

embassy-stm32/src/dma/dma.rs
```diff
@@ -273,9 +494,38 @@ mod low_level_api {
...
+                        dma.st(channel_num).m0ar().write_value(new_addr);
+                    } else if !current_target_memory1 {
+                        // update pointer to memory 1
+                        let mem1_addr = dma.st(channel_num).m1ar().read();
+                        let new_addr =
+                            GIANT_TRANSFER_STATE.dequeue_next_chunk(state_index, mem1_addr);
+                        dma.st(channel_num).m1ar().write_value(new_addr);
+                    }
+                } else {
+                    GIANT_TRANSFER_STATE.disable_for_channel(state_index);
```
Dirbaio, line 521: where is the DMA hardware actually stopped? AFAICT it'll keep reading...?

embassy-stm32/src/dma/dma.rs
```diff
@@ -56,6 +57,93 @@ impl State {
...
+        &self,
+        channel_index: usize,
+        data_addr: u32,
+        data_len: usize,
+        transfer_len_bytes: u8,
+    ) -> (M0AR, M1AR, ChunkSize) {
+        let mut chunk_size = data_len;
+        let mut total_chunk_count = 1;
+        while chunk_size > 0xffff {
+            chunk_size /= 2;
```
Dirbaio, line 95: this is wrong if `data_len` doesn't divide evenly (for example if it's `2^17 + 1`), you're rounding down so it'll read less bytes than wanted.

matoushybl: You are right, the giant transfer cannot be used wit odd data length, as there is no way of changing the data length without any gap between the transfers.

The Channel trait impl checks for this case and panics, I also added some asserts that checks that the data len is even.

embassy-stm32/src/dma/dma.rs
Dirbaio: Oh, is it the `cr.write(|_| ());`? I see. Would be nice to clarify it with a comment, the current comment indicates it's just to disable irqs.

Isn't it "too late" by then though? like, after you get the last TCIF, the hardware won't stop automatically, it'll swap buffers once again and start overwriting the second-to-last chunk.

embassy-stm32/src/dma/dma.rs
Dirbaio: It's not enough to check it's even. If you do 4 chunks it has to be multiple of 4, if 8 chunks it has to be multiple of 8, etc...

embassy-stm32/src/dma/dma.rs
matoushybl: Good thinking. I had a look at the Cube code and it seems to set M0AR to the initial buffer address and M1AR to first chunk address https://github.com/STMicroelectronics/STM32CubeH7/blob/b340b13929e36a3427b8d94e8b1006022f82273f/Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_dcmi.c#L1151-L1161

Which is I assume kinda works for DCMI as it doesn't matter that much if the data get overwritten? Definitely feels weird.

Dirbaio: Yeah, seems like they ignore the issue there, probably because in the case of DCMI you control how much data it will generate, so if you read exactly 1 frame it won't actually overwrite stuff. This is not the case for UART etc.

Maybe the "poison address" Idea would work here? At second-to-last TCIF, set the other MxAR to 0xFFFF_FFFF, At the last TCIF, it'll switch to writing there, and stop due to the error.

embassy-stm32/src/dma/dma.rs
matoushybl: You are right, maybe dcmi buffers lengths uphold some properties that makes it work for them. https://github.com/STMicroelectronics/STM32CubeH7/blob/b340b13929e36a3427b8d94e8b1006022f82273f/Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_dcmi.c#L419 I’ll do some calculations soon to see if it isn’t a bug in cube as well.

Dirbaio:
@matoushybl LGTM, except could you not feature-gate it? There's no downside in always having it, it causes no bloat if it's not used.

Dirbaio (approved):
Thank you!

bors r+

matoushybl:
Thanks for all the help with this!

PR #720, opened 2022-04-15
USB: New builder API 

Dirbaio:
usb: improved descriptor building API
    
    The same API call allocates interfaces/endpoints/etc and writes their descriptors.
    This means less API calls, and less possibility to screw things up.
    
    DescriptorWriter is now private.

### DIFF
--- a/embassy-usb-hid/src/lib.rs
+++ b/embassy-usb-hid/src/lib.rs
@@ -110,23 +110,13 @@ fn build<'d, D: Driver<'d>>(
     ));
 
     let len = config.report_descriptor.len();
-    let if_num = builder.alloc_interface_with_handler(control);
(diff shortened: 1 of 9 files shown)


alexmoon:
The HID changes are actually pretty close to where I started with it. I changed to the more generic heavy version in order to move potential errors from runtime to compile time (no more `Option`s or `unwrap`s). What about just making a type alias?

```rust
pub type HidClassOut<'d, D, const IN_N: usize, const OUT_N: usize> =
    HidClass<'d, D, ReportReader<'d, D, OUT_N>, IN_N>;
```

This version also seems to require the use of the `split` method in order to use the class. I think it's nicer from an API perspective to give the user the choice whether or not splitting makes sense in their application.

alexmoon:
The `Builder` changes look good to me. I like the design of nested builders, especially for something as tree-like as the USB configuration is.

PR #723, opened 2022-04-19
Add embedded-storage trait impls for QSPI

lulf:
* Adds implementations of embedded-storage and embedded-storage-async
for QSPI
* Add blocking implementations of QSPI
* Use blocking implementation in new() and embedded-storage impls
* Use async implementation in embedded-storage-async impls
* Add FLASH_SIZE const generic parameter
* Own IRQ in Qspi instance to disable it on drop.

### DIFF
--- a/embassy-nrf/Cargo.toml
+++ b/embassy-nrf/Cargo.toml
@@ -18,7 +18,7 @@ flavors = [
 [features]
 
 # Enable nightly-only features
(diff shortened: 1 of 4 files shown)


Dirbaio:
For consistency with other drivers, blocking API should be `blocking_foo` instead of `foo_blocking`.

LGTM otherwise :)

Dirbaio (approved):
awesome! Thank you

bors r+

PR #724, opened 2022-04-20
STM32 Flash + Bootloader 

lulf:
Not working.


lulf:
Bootloader/flash tested on L0, L4 and WL55, should also work on L1 and WB.

embassy-stm32/src/flash/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub(crate) const WORD_SIZE: usize = 8;
+}
+
+#[cfg(flash_wb)]
+mod config {
+    use super::*;
+    pub(crate) const FLASH_SIZE: usize = 0x100000;
+    pub(crate) const FLASH_START: usize = FLASH_BASE;
+    pub(crate) const FLASH_END: usize = FLASH_START + FLASH_SIZE;
+    pub(crate) const PAGE_SIZE: usize = 4096;
```
Dirbaio, line 31: perhaps the names `ERASE_SIZE`, `WRITE_SIZE` would be clearer. I've seen "page size" used as the write size in some places.

Dirbaio, line 31: perhaps the names `ERASE_SIZE`, `WRITE_SIZE` would be clearer. I've seen "page size" used as the write size in some places.

embassy-stm32/src/flash/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+use embedded_storage::nor_flash::{
+    ErrorType, NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash,
+};
+
+const FLASH_BASE: usize = 0x8000000;
+
+#[cfg(flash_l4)]
+mod config {
+    use super::*;
+    pub(crate) const FLASH_SIZE: usize = 0x100000;
```
Dirbaio, line 18: Is this the max size for each family? Ideally the size should be the one for the current chip.

Dirbaio, line 18: Is this the max size for each family? Ideally the size should be the one for the current chip.

embassy-stm32/src/flash/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+        unsafe {
+            pac::FLASH.pecr().modify(|w| {
+                w.set_optlock(true);
+                w.set_prglock(true);
+                w.set_pelock(true);
+            });
+        }
+    }
+
+    pub fn blocking_read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Error> {
```
Dirbaio, line 116: the embedded-storage trait specifies `offset` is relative to the start of the memory. Even if the flash is memory-mapped starting at `0x8000000`, the offset should start at `0`.

Dirbaio, line 116: the embedded-storage trait specifies `offset` is relative to the start of the memory. Even if the flash is memory-mapped starting at `0x8000000`, the offset should start at `0`.

embassy-stm32/src/flash/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+        }
+        if offset as usize % WORD_SIZE != 0 || buf.len() as usize % WORD_SIZE != 0 {
+            return Err(Error::Unaligned);
+        }
+
+        self.clear_all_err();
+
+        #[cfg(any(flash_wl, flash_wb, flash_l4))]
+        unsafe {
+            pac::FLASH.cr().write(|w| w.set_pg(true));
```
Dirbaio, line 138: if you remove the trailing semicolon, many of these `unsafe{}`s become a single line

Dirbaio, line 138: if you remove the trailing semicolon, many of these `unsafe{}`s become a single line

embassy-stm32/src/flash/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+                    return Err(Error::Seq);
+                }
+                return Ok(());
+            }
+        }
+    }
+
+    fn clear_all_err(&mut self) {
+        unsafe {
+            pac::FLASH.sr().modify(|w| {
```
Dirbaio, line 265: maybe this can be done by simply writing 0?

Dirbaio, line 265: maybe this can be done by simply writing 0?

Dirbaio (changes requested):

embassy-stm32/src/flash/mod.rs
lulf: I've used the 'max size for family' according to the RM, but the true flash size could be lower, and there are a lot of combinations. The true flash size can be found in a system page, which varies with each chip family. Since it cannot be retrieved from the header files directly (without expanding lots of C macros), I wonder if maybe we can create a static map of the page location in stm32-data, and generate a global constant with this address. 

While at it, one could add more of these per-familiy constants in that mapping I guess.

embassy-stm32/src/flash/mod.rs
lulf: Ah, interesting. I have to change some assumptions about the addresses in the linker scripts provided by the bootloader.

lulf:
@Dirbaio Thanks for the review!

embassy-stm32/src/flash/mod.rs
lulf: Hmm, they don't for me.

embassy-stm32/src/flash/mod.rs
lulf: Will test it

embassy-stm32/src/flash/mod.rs
lulf: Ok, now it's fully derived from XML

Dirbaio:
ugh why did github duplicate all my comments? :joy:

lulf:
I've updated stm32-data to provide more flash settings: erase_size, write_size and erase_byte which are all derived from the XML. It works for the chips I've tested for so far, but might require some tuning for the more advanced multi-bank chips.

This, combined with making a single Aligned type aligned to the largest write_size we know of, gets rid of the various bootloader features I've added as well.

Dirbaio (approved):
Awesome work Ulf! :100: 

bors r+

PR #725, opened 2022-04-21
remove description which not makes sense

liigo:


liigo:
A `task` contains a `core::future::Future`s tree. `Future` returns `Pull::Ready` (defined by the core lib). "Task returns `Pull::Ready`" is not an defined semantic. You may just say: after the *last* `Future` in a task returns `Pull::Ready`, an executor *may* mark the task as 'finished' and will not schedule it any more.

Dirbaio:
I see. IMO the fact that tasks can exit is important enough that we should keep it. Other than that, feel free to reword to clarify.

liigo:
I've pushed a new commit, please review. @Dirbaio

PR #729, opened 2022-04-24
Update Rust nightly.

Dirbaio:
Fixes a new opaque type error in the task macro.
Full error is "opaque type's hidden type cannot be another opaque type from the same scope".
This got disallwed by the lazy-TAIT PR: https://github.com/rust-lang/rust/pull/94081

Sadly there's now some weird type inference errors with pre-lazy-TAIT
nightlies, so support for those is dropped.

### DIFF
--- a/embassy-macros/src/macros/task.rs
+++ b/embassy-macros/src/macros/task.rs
@@ -61,10 +61,14 @@ pub fn run(args: syn::AttributeArgs, mut f: syn::ItemFn) -> Result<TokenStream, 
 
     ctxt.check()?;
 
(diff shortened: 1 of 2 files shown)


Dirbaio:
WIP, investigating why it now fails in older nightlies...???

Dirbaio:
it seems a type inference bug in pre-lazy-tait code. Can be fixed with yet another wrapper fn, but I'd rather not do it because it's more opportunities for the optimizer to fail to inline stuff. Let's drop support for old nightlies instead.

PR #732, opened 2022-04-26
macros: simplify task macro using "TAIT laundering".

Dirbaio:
This brings the macro to the state before the nightly update #729 #730, with a much cleaner workaround for the opaque type error, from https://github.com/rust-lang/rust/issues/96406


Dirbaio:
bors r+

@lulf feel free to bors r+ yourself when you approve :)

PR #733, opened 2022-04-26
Add f107 ethernet support.

davidlenfesty:
The original driver works perfectly, the only required changes were only to clock and pin configuration.

Bits that are also present and deserve some scrutiny:
- Migrated LAN8742a driver to a generic SMI version (only *functional* difference is a couple extra status checks for the link poll, which IMO weren't great anyways (i.e. it would consider the link down if it negotiated to 10M)
- Migrated f1 RCC init to "new" style. As of creating this draft PR, it is pretty much only tested on the happy path for my specific configuration, and also needs a couple things done (calculated clock speeds are hardcoded and ADC clock isn't implemented)
- Support for v1b driver (f2 and f4) was added, but not tested.

(Made a draft PR until f4 support is verified and I finish the clock init - adding testing and at least feature parity with how it was before)


davidlenfesty:
I neglected to implement the usbprescaler as well, looks like I'll need to tweak stm32-data again.

Dirbaio:
Tested on nucleo-stm32f429zi, it Just Works. :+1:  The only issue was "ETHMAC" vs "ETH" in RCC registers, I've unified them to "ETH". I've pushed it to your branch, hope you don't mind :) 

If the F1 RCC refactor is going to be complicated, could you split that to a separate PR? I'd rather have smaller PRs, and just the ETH changes already make it work on F4.

davidlenfesty:
I removed all the RCC stuff as @GrantM11235 is working on a more comprehensive (and correct :joy:) RCC implementation.

Note that I don't think you can really support any reasonable ethernet clock configs without RCC changes, so the f107 changes don't really do anything until that gets finished :P

Dirbaio:
Thank you! :rocket: 

bors r+

PR #734, opened 2022-04-26
executor: Add `Spawner::for_current_executor`.

Dirbaio:
This is needed to spawn non-Send tasks in an InterruptExecutor, after the fixes in #730 .

@matoushybl could you check if this works for your use case?


matoushybl:
@Dirbaio Thanks for this! I created a trampoline task and spawned the higher priority tasks there.

PR #738, opened 2022-04-27
Add split method to UarteWithIdle in nrf hal

ZoeyR:
I needed this method because I was writing a driver for the sim7000 modem. Its a line based protocol so the read until idle was very useful to me. Unfortunately unlike the primary `Uarte` struct the `UarteWithIdle` didn't have `split()`. So I added it 😄


huntc:
LGtM. Did you essentially copy the same approach as per the regular Uarte? It looks that way, which is good.

huntc:
I’m wondering though whether it’d be better for us to remove the UarteWithIdle type and have a split_with_idle on Uarte to reduce code…

ZoeyR:
Yeah, I basically copied the approach from the regular `Uarte`. I don't have a strong opinion on removing `UarteWithIdle`. I can do either approach.

lulf (approved):
Thanks for the PR 🚀 🚀 

bors r+

PR #743, opened 2022-04-29
Add PLL config support for F2

Gekkio:
- minor changes to make the F2 RCC API a bit more flexible
- low-level PLL config with assertions based on datasheet specs. It shouldn't be very difficult to later add a "reverse API" where you pass the clocks you want to a function and it generates a `PLLConfig` struct for you
- PLL API tested on my custom board with 12 MHz HSE as source for PLL to generate max clocks for SYSCLK/AHB/APB/APB1/PLL48
- the example *should* work but is untested since I don't have the Nucleo board :disappointed:

### DIFF
--- a/embassy-stm32/src/rcc/f2.rs
+++ b/embassy-stm32/src/rcc/f2.rs
@@ -1,7 +1,8 @@
-use core::ops::Div;
+use core::convert::TryFrom;
+use core::ops::{Div, Mul};
 
 use crate::pac::flash::vals::Latency;
-use crate::pac::rcc::vals::{Hpre, Ppre, Sw};
+use crate::pac::rcc::vals::{Hpre, Pllp, Pllsrc, Ppre, Sw};
 use crate::pac::{FLASH, RCC};
 use crate::rcc::{set_freqs, Clocks};
 use crate::time::Hertz;
@@ -20,6 +21,7 @@ pub struct HSEConfig {
 pub enum ClockSrc {
     HSE,
     HSI,
+    PLL,
 }
 
 /// HSE clock source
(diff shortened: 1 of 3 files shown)


Dirbaio:
> Move HSE config out of main clock mux.
> This makes the configuration more flexible and closer to the underlying
> configuration register structure. For example, we could use HSI for the
> system clock, but use HSE to output a clock with MCO.

This is very interesting, I hadn't thought of this before. SGTM :+1:  The downside is now you can set the mux to HSE without configuring it, I see you're checking for this and panicking.

embassy-stm32/src/rcc/f2.rs
```diff
@@ -226,30 +406,65 @@ impl Default for Config {
...
+            assert!(config.hsi, "HSI must be enabled to be used as system clock");
             (HSI, Sw::HSI)
         }
-        ClockSrc::HSE(freq, source) => {
-            enable_hse(source);
-            (freq, Sw::HSE)
+        ClockSrc::HSE => {
+            let hse_config = config
+                .hse
+                .expect("HSE must be configured to be used as system clock");
```
Dirbaio, line 461: could you use `unwrap!`, `panic!` or `assert!` or similar instead of `.expect()`? `expect()` stores the error message as a string, while the others can use defmt's string tokenization.

Gekkio: Sure! I naively thought assert did that too so I didn't give the error handling style much thought...I'll update the example too since it uses `.unwrap()` and not `unwrap!`

Gekkio:
> > Move HSE config out of main clock mux.
> > This makes the configuration more flexible and closer to the underlying
> > configuration register structure. For example, we could use HSI for the
> > system clock, but use HSE to output a clock with MCO.
> 
> This is very interesting, I hadn't thought of this before. SGTM +1 The downside is now you can set the mux to HSE without configuring it, I see you're checking for this and panicking.

True, it slightly breaks the good API design guideline "make illegal states unrepresentable" since there are dependencies between the states of the different `Config` struct fields that are checked at runtime. On the other hand, these dependencies exist in the hardware itself so it's mostly a question about the benefit and cost of abstracting and hiding away things.

I think the older design was nicer in simple and common cases, because the types nicely guided you towards a correct config. But it also added extra assumptions which made some use cases impossible...in some advanced cases we'd need to duplicate some parts of the config (and handle the possibility of config mismatches if for example HSE frequency was defined as part of both main mux and PLL configurations) or just put the shared stuff in one place and refer to it from elsewhere like I did here.

Some example advanced scenarios I've thought about:

- accurate HSE crystal for USB purposes, so PLL uses HSE and outputs an accurate PLL48 clock for USB. SYSCLK uses HSI for low power, so neither PLL or HSE configuration can be behind the main clock mux config
- HSE + PLL + MCO to output a clock to some other chip (e.g. FPGA), while the system uses HSI for low power
- HSE for accurate USB *and* I2S, while the system uses HSE directly. Note that the main PLL and PLLI2S share the same input clock mux and pre division factor, so they are not independent...IMHO the worst API design would require defining the HSE frequency three times: for main clock mux, for main PLL, and for I2S PLL. In this pull request, PLLI2S is not currently supported, but it would probably just be added to `PLLConfig`, and you'd define HSE characteristics and the shared parts of PLL config just once.

I'm still not 100% sure which way is better, but the "cost" of this more flexible version is pretty low so I decided to go with it :smile:

Dirbaio (approved):
Makes a lot of sense, yep! Thank you!

bors r+

PR #749, opened 2022-05-03
Rename spawn_allocate to spawn_mark_used

danbev:
This commit contains a suggestion to rename `TaskStorage::spawn_allocate`.

The motivation for this is when reading through the code I was
expecting something else to happen in this method, due to `allocate` in
the method name.

I'm very new to this code base so I may be wrong in thinking this
is a good change, but I wanted to open this PR to get some feedback.

### DIFF
--- a/embassy/src/executor/raw/mod.rs
+++ b/embassy/src/executor/raw/mod.rs
@@ -164,14 +164,14 @@ impl<F: Future + 'static> TaskStorage<F> {
     /// Once the task has finished running, you may spawn it again. It is allowed to spawn it
     /// on a different executor.
     pub fn spawn(&'static self, future: impl FnOnce() -> F) -> SpawnToken<impl Sized> {
-        if self.spawn_allocate() {
+        if self.spawn_mark_used() {
             unsafe { SpawnToken::<F>::new(self.spawn_initialize(future)) }
         } else {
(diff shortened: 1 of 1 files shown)


Dirbaio:
The logic behind the old names is "spawning" is two steps: "allocate" the memory for the task (mark the TaskStorage as "in use"), then "initialize" it with the future. 

I'd rather keep the "symmetry" in the names, but I agree "allocate" can be misleading because it's not using "The One Alloc" (the heap). Perhaps "spawn_reseve", "spawn_mark_used", "spawn_mark"? Or perhaps just addding comments would clarify it enough?

danbev:
> I'd rather keep the "symmetry" in the names, but I agree "allocate" can be misleading because it's not using "The One Alloc" (the heap). Perhaps "spawn_reseve", "spawn_mark_used", "spawn_mark"? Or perhaps just addding comments would clarify it enough?

Thanks for the fast reply! I like your suggestion `spawn_mark_used` and updated with the PR with it.
Adding a comment would work, the only thing it that at least for me it would be easier to read with the changed name and not have to read a comment. But this is just a suggestion and I have no problem with not changing this at all if you'd prefer that :+1:

Dirbaio (approved):
Sounds good, thank you! :) 

bors r+

PR #751, opened 2022-05-04
Use explicit return statement TaskStorage::spawn

danbev:
This commit removes the else branch in `TaskStorage::spawn`, and returns
explicitly from the if statement's branch, similar to what
[TaskPool::spawn](https://github.com/embassy-rs/embassy/blob/85c0525e01c52bbb85c7b93600a60837ee7b87dc/embassy/src/executor/raw/mod.rs#L235-L243) does.


Dirbaio (approved):
Thank you!

bors r+

PR #754, opened 2022-05-05
Correct the UarteWithIdle timeout math

### LINKED ISSUE #753: nrf UarteWithIdle uses too short a timeout.
It appears the math is wrong for the timeout in `UarteWithIdle` on the NRF. It attempts to calculate the time for 40bits of idle but it ends up about half of that, at least for 115200 baud.

At 115200 baud the current timeout is 2788 cycles. Calculated with 0x8000_0000 / (30801920 / 40);
30801920 is `Baudrate::BAUD115200` as a u32.

The timeout _should_ be about 5,555 cycles if my math is correct. Math detailed below:
- 115200 baud is 115200 bits per second
- take 1 / 115200 to get the seconds per bit
- multiply by 40 to get the total time in seconds
- multiply by the timer frequency of 16 million cycles / second to get the number of cycles
- This should be equal to 5,555.5 cycles

I noticed this because the uarte would drop bytes randomly during usage. I increased the timeout by a massive amount (to 100_000 cycles) to see if it fixed the issue and it did. Looking at the math this seems to be the culprit.

ZoeyR:
Fixes #753 

Tested on a custom board with an nrf52840 communicating via uarte to a sim7000g using hardware flow control. I also bumped the idle time to 8 bytes of idle since 2 bytes of idle cut it a bit short in my experience. Especially given rounding errors due to integer math.

### DIFF
--- a/embassy-nrf/src/uarte.rs
+++ b/embassy-nrf/src/uarte.rs
@@ -635,12 +635,12 @@ impl<'d, U: Instance, T: TimerInstance> UarteWithIdle<'d, U, T> {
         let r = U::regs();
 
         // BAUDRATE register values are `baudrate * 2^32 / 16000000`
-        // source: https://devzone.nordicsemi.com/f/nordic-q-a/391/uart-baudrate-register-values
+        // source: https://devzone.nordicsemi.com/f/nordic-q-a/39130801920/uart-baudrate-register-values
         //
-        // We want to stop RX if line is idle for 2 bytes worth of time
-        // That is 20 bits (each byte is 1 start bit + 8 data bits + 1 stop bit)
-        // This gives us the amount of 16M ticks for 20 bits.
-        let timeout = 0x8000_0000 / (baudrate as u32 / 40);
+        // We want to stop RX if line is idle for 8 bytes worth of time
+        // That is 160 bits (each byte is 1 start bit + 8 data bits + 1 stop bit)
+        // This gives us the amount of 16M ticks for 160 bits.
+        let timeout = (0xFFFF_FFFF / baudrate as u32) * 160;
 
         timer.set_frequency(Frequency::F16MHz);
         timer.cc(0).write(timeout);


ZoeyR:
Perhaps it would be better to leave the un-simplified formula in the comments so that it can be reconstructed to check for accuracy? Let me know.

Dirbaio:
As discussed in Matrix, if you're using RTS/CTS you shouldn't lose data independently of the idle timeout. For users not using RTS/CTS, there's no way to choose a timeout value that will suit everyone. Due to this, I'd rather keep the shorter timeout of 2 bytes, as it'll lead to less latency.

The original maths are correct:

```

Formula from converting BAUDRATE reg value to baudrate hz:
Source: https://devzone.nordicsemi.com/f/nordic-q-a/391/uart-baudrate-register-values
baudrate_hz = baudrate_reg * 16_000_000 / 2^32
Example: 0x01D7E000 * 16_000_000 / 2^32 = 115203.857421875 = ~115200

1 byte on the wire is 10 bits worth of time (1 start bit + 8 data bits + 1 stop bit).
2 bytes is therefore 20 bits.

That is, 20 baudrate_hz cycles.
That is, 20 / baudrate_hz seconds.

Formula for calculating CC register value from a timeout in seconds,
when the timer is running at 16Mhz:

cc = timeout_s * 16_000_000.

Putting this all together:

cc = timeout_s * 16_000_000.
cc = (20 / baudrate_hz) * 16_000_000.
cc = (20 / (baudrate_reg * 16_000_000 / 2^32)) * 16_000_000.
cc = 20 / baudrate_reg / 16_000_000 * 2^32 * 16_000_000.
cc = 20 * 2^32 / baudrate_reg.

Now, to make the math fit in a u32:

cc = 20 * 2^32 / baudrate_reg.
cc = 20 * 2 * 2^31 / baudrate_reg.
cc = 40 * 2^31 / baudrate_reg.
cc = 2^31 / (baudrate_reg / 40)

Dividing baudrate_reg / 40 is slightly less precise, but the result is rounded
down, so the end result in `cc` is rounded up. Waiting for slightly more than 2 bytes is fine.
```

Your formula is also correct. You're replacing `2^32` with `0xFFFF_FFFF` which I guess is slightly more accurate than having to divide by 40 instead of 20.

One difference is your formula rounds down, while the original one rounds up. I think it's better to always round up, so the timeout is always "at least 2 bytes". 

If you want, change the formula to the more accurate `cc = 0xFFFF_FFFF / (baudrate_reg / 20)`, otherwise I think we can close this.

Dirbaio:
I'm going to go ahead and close this given the silence, feel free to open a new PR if you think the formula with slightly better precision is worth it.

PR #755, opened 2022-05-06
Add support for flash and bootloader for F3, F7 and H7

matoushybl:


lulf (commented):
This is amazing! :rocket: :rocket: :rocket: :smile_cat: ! 

I left one comment about the flash base which you can get from the HAL/pac. 

Maybe you can also update the documentation list of supported STM32 chips in docs `docs/modules/ROOT/pages/bootloader.adoc`

lulf (approved):

PR #756, opened 2022-05-07
Add async quadrature decoder for embassy-nrf

kalkyl:
Thsi PR adds an async interface to the QDEC peripheral to embassy-nrf, that can be used for rotary encoders for example.


embassy-nrf/src/qdec.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// let irq = interrupt::take!(QDEC);
+    /// let config = qdec::Config::default();
+    /// let mut q = Qdec::new(p.QDEC, p.P0_31, p.P0_30, config);
+    /// let delta = q.read().await;
+    /// ```
+    pub async fn read(&mut self) -> i16 {
+        // In case the future is dropped, stop the task and reset events.
```
Dirbaio, line 158: Is this needed? I think if you stop QDEC then it'll stop counting. If rotation happens while hte user is not reading, it'll be lost.

It might be useful to `select()` between QDEC read and a timeout, in which case the user might want to read later and not lose the rotation count.

embassy-nrf/src/qdec.rs
```diff
@@ -0,0 +1,222 @@
...
+        let on_drop = OnDrop::new(|| {
+            let t = Self::regs();
+            t.tasks_stop.write(|w| unsafe { w.bits(1) });
+            t.events_reportrdy.reset();
+        });
+
+        let t = Self::regs();
+        t.intenset.write(|w| w.reportrdy().set());
+        unsafe { t.tasks_start.write(|w| w.bits(1)) };
```
Dirbaio, line 167: perhaps start() could be called on `new` instead, if we always want it "started" (related to the previous comment)

kalkyl:
Good points! Updated!

PR #760, opened 2022-05-08
stm32/gpio: Add support for `set_speed` for gpio v1

GrantM11235:


embassy-stm32/src/gpio.rs
```diff
@@ -511,10 +511,20 @@ pub(crate) mod sealed {
             self.set_as_analog();
         }
 
-        #[cfg(gpio_v2)]
+        #[cfg(any(gpio_v1, gpio_v2))]
```
Dirbaio, line 514: this is always, just remove the cfg

Dirbaio (approved):
Very nice, thank you! :) 

Fixes #758

PR #761, opened 2022-05-08
stm32: make set_speed unconditional.

Dirbaio:


Dirbaio:
Closing for now. `set_as_af` already sets the speed in GPIOv1 https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/gpio.rs#L445

Maybe all the `set_speed`s should be conditional instead of none, Or maybe the `set_as_af` API could be improved to always set the speed so no conditionals are needed (maybe take a config?)

PR #762, opened 2022-05-08
Allow maximal clock for F7 HCLK

matoushybl:
Fixes a typo in clock calculations, see: https://github.com/stm32-rs/stm32f7xx-hal/blob/aaf2034cab697577e6f4fc6079fc7e4691d7ad9d/src/rcc.rs#L347


Dirbaio (approved):
oh whoops, nice catch!

bors r+

PR #763, opened 2022-05-09
Misc USB improvements

Dirbaio:
The "simplify control in/out handlng" commit gives a -2kb code size improvement.


embassy-nrf/src/usb.rs
```diff
@@ -183,8 +177,10 @@ impl<'d, T: Instance> driver::Driver<'d> for Driver<'d, T> {
         &mut self,
         max_packet_size: u16,
     ) -> Result<Self::ControlPipe, driver::EndpointAllocError> {
-        self.alloc_endpoint_out(Some(0x00.into()), EndpointType::Control, max_packet_size, 0)?;
-        self.alloc_endpoint_in(Some(0x80.into()), EndpointType::Control, max_packet_size, 0)?;
+        self.alloc_in.used |= 0x01;
+        self.alloc_in.lens[0] = max_packet_size as u8;
+        self.alloc_out.used |= 0x01;
+        self.alloc_out.lens[0] = max_packet_size as u8;
```
alexmoon, line 183: Should this return an error if the control pipe has already been allocated?

alexmoon: > I don't like the fact it's possible to have this error in the first place though. I was thinking of changing it so that the control pipe is returned by `into_bus`: `fn enable(self) -> (Self::Bus, Self::ControlPipe)`, this way the error is impossible.

Yes, that's better. I don't like the `enable` name though, because the bus is not enabled by that function. I think `into_bus` is clearer.

Dirbaio: with the new API it should be `into_bus_and_control_pipe` which is not very nice. Perhaps `initialize`,  or `start`?

alexmoon: initialize could be ok. Other options:

- init 
- build
- into_inner
- finalize
- freeze

start still makes me think it does something to the bus state.

Dirbaio: :'( I think `start` is good enough. The driver is previously initialized but "being configured", `.start()` starts actual operation.

PR #765, opened 2022-05-11
Implement Output::is_set_low for embassy-rp

danbev:
This commit implements a suggestion for the method `is_set_low` which is
currently a `todo`, by reading last value written to `GPIO_OUT`.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -127,8 +127,8 @@ impl<'d, T: Pin> Output<'d, T> {
 
     /// Is the output pin set as low?
     pub fn is_set_low(&self) -> bool {
-        // todo
-        true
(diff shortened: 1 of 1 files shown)


embassy-rp/src/gpio.rs
```diff
@@ -127,8 +127,8 @@ impl<'d, T: Pin> Output<'d, T> {
 
     /// Is the output pin set as low?
     pub fn is_set_low(&self) -> bool {
-        // todo
-        true
+        // Reading from SIO: GPIO_OUT gives the last value written.
+        unsafe { self.pin.sio_out().value().read() == 0 }
```
Dirbaio, line 131: `VALUE` is a `u32` with each bit being the value of one pin. You should check just the bit of the current pin, not the whole `VALUE`.

PR #767, opened 2022-05-12
Stm32 fixes

jr-oss:
Using embassy on STM32 Discovery board showed problems with
- USART with parity
- Clock settings (CFGR) when using PLL and prescaler
- Flash ACR settings

This PR attempts to fix these


lulf:
LGTM, just some rustfmt fixes needed to pass the build.

lulf:
Awesome :rocket: 

bors r+

PR #769, opened 2022-05-12
Async shared bus for SPI & I2C + rename embassy-traits

kalkyl:
This PR adds `embedded-hal-async` compatible shared bus wrappers for SPI and I2C.
Also renames `embassy-traits` to `embassy-embedded-hal`, since this is now home for utils for embedded-hal(-async) for the embassy project.


alex13sh:
why doesn't I2cBusDevice have an i2c address? similar to how SpiBusDevice has CS

kalkyl:
> 

The I2C address is conventionally a hard-coded value inside the device drivers (or configurable via the driver) so this wrapper has no use of it. Its job is just to provide shared access to the underlying peripheral while exposing the same embedded-hal-async compatible interface so that the same device drivers can be used as with an exclusive device

Dirbaio:
Thank you! :rocket:

PR #770, opened 2022-05-13
Add open-drain support for embassy-rp

danbev:
This commit adds open-drain support for embassy-rp by adding a new type
named `embassy_rp::gpio::OutputOpenDrain`.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -86,7 +86,6 @@ pub struct Output<'d, T: Pin> {
 }
 
 impl<'d, T: Pin> Output<'d, T> {
-    // TODO opendrain
     pub fn new(pin: impl Unborrow<Target = T> + 'd, initial_output: Level) -> Self {
         unborrow!(pin);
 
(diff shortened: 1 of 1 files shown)


embassy-rp/src/gpio.rs
```diff
@@ -139,6 +138,91 @@ impl<'d, T: Pin> Drop for Output<'d, T> {
...
+    #[inline]
+    pub fn new(pin: impl Unborrow<Target = T> + 'd, initial_output: Level) -> Self {
+        unborrow!(pin);
+
+        unsafe {
+            match initial_output {
+                Level::High => {
+                    // For Open Drain high should pull low (drain).
+                    pin.pad_ctrl().write(|w| {
+                        w.set_pde(true);
```
Dirbaio, line 157: why pull down? 

Open drain means "0 = drive low, 1 = not drive". if you use a pull you want pull *up* so that it's high when nobody is driving it, not pull *down*.

Also, Pullup should be optional, you don't want it if the board has an external pullup resistor.

embassy-rp/src/gpio.rs
```diff
@@ -139,6 +138,91 @@ impl<'d, T: Pin> Drop for Output<'d, T> {
...
+        Self {
+            pin,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Set the output as high.
+    #[inline]
+    pub fn set_high(&mut self) {
+        // For Open Drain high should pull low (drain).
```
Dirbaio, line 178: this is the other way around, low = drive the pin low, high = don't drive the pin

embassy-rp/src/gpio.rs
```diff
@@ -139,6 +138,91 @@ impl<'d, T: Pin> Drop for Output<'d, T> {
...
+        unsafe {
+            self.pin.pad_ctrl().write(|w| {
+                w.set_od(false);
```
Dirbaio, line 181: you can use `gpio_oe` in `SIO` to enable/disable the output  (`SIO` is faster than all other peripherals, accesses to it are single-cycle)

embassy-rp/src/gpio.rs
danbev: > why pull down?

The reason for doing this was that my understanding of open-drain is that it has two states, `GND/drive low` or `floating/not drive`. So the pull-down in this case was for the `GND state/drive low` state. It sounds like I got this wrong though. Could you tell me what the correct way to drive the output pin low is? 

> Open drain means "0 = drive low, 1 = not drive"

Sorry, I got these mixed up and for some reason I managed to convince myself of the opposite :disappointed: 

> Also, Pullup should be optional, you don't want it if the board has an external pullup resistor.

I was under the assumption that any Pullup would be an external pull up and was intentionally left out.

Thanks for the feedback and your patience as this is very new to me!

embassy-rp/src/gpio.rs
danbev: Sorry, I got these mixed up and for some reason I managed to convince myself of the opposite :disappointed:

embassy-rp/src/gpio.rs
danbev: That makes sense, I'll update shortly. Thanks

embassy-rp/src/gpio.rs
Dirbaio: check the datasheet page 264, figure 37.

The "output" is the buffer at the top. If output_enable=1, it drives the value of output_data into the pad. If output_enable=0, it doesn't (leaves the pad floating).

The "pull up/down" is resistors at the bottom. If you enable pull up/down the corresponding resistor is enabled, driving the pad high/low *very weakly*.

A common use of open drain is to "talk" on a single wire in two directions. We set pull-up so that the wire reads high if no one is driving it, then we can either drive it ourselves low to talk, or we can not drive and "listen" to the other chip talk.

The pullup can be internal to the MCU (enabling it with that register) or it can be external (a physical resistor in the PCB, separate to the MCU).

Dirbaio: So, to do open drain:

- on new:
  - set funcsel to SIO, otherwise what you write to SIO doesn't take effect
  - Set output value to 0, leave it like that (it only takes effect if OE=1)
- When set_high, set OE=0
- When set_low, set OE=1

it'll be easier if you don't configure the pull-up for now, we can add support for it later

danbev: Thanks for the explanation, I really appreciate it!

I'll update this pr and also rebase and squash the commit so that it is easier to review.

embassy-rp/src/gpio.rs
```diff
@@ -155,6 +154,90 @@ impl<'d, T: Pin> Drop for Output<'d, T> {
...
+    /// Is the output level low?
+    #[inline]
+    pub fn is_set_low(&self) -> bool {
+        !self.is_set_high()
+    }
+
+    /// Toggle pin output
+    #[inline]
+    pub fn toggle(&mut self) {
+        if self.is_set_low() {
```
Dirbaio, line 233: you can also use `value_xor` here.

Dirbaio:
LGTM other than the comment, feel free to merge yourself when done :) 

bors d+

danbev:
> LGTM other than the comment, feel free to merge yourself when done :)

Awesome, thanks for all your help! :)

PR #776, opened 2022-05-18
Automatically set ADC clock prescaler on v2 ADC to respect max frequency

matoushybl:


Dirbaio (approved):
thank you! 



bors r+

PR #779, opened 2022-05-19
Impl ToggleableOutputPin for embassy-rp Output

danbev:


embassy-rp/src/gpio.rs
```diff
@@ -131,6 +131,16 @@ impl<'d, T: Pin> Output<'d, T> {
         let val = 1 << self.pin.pin();
         unsafe { (self.pin.sio_out().value().read() & val) == 0 }
     }
+
+    /// Toggle pin output
+    #[inline]
+    pub fn toggle(&mut self) {
+        if self.is_set_low() {
```
Dirbaio, line 138: you can do this much faster by writing to `value_xor` :)

danbev: Oh, that sounds nice, let me take a look. Thanks!

danbev: I've taken a stab at that and hope I got it right: https://github.com/embassy-rs/embassy/pull/779/commits/99c2defa76012e28d1249c206ebe11718aeba3b7

Dirbaio (approved):
Thank you!

:rocket:
 
bors r+

PR #784, opened 2022-05-27
stm32: make tick rate configurable

willglynn:
The stm32 time drivers support arbitrary tick rates but the associated Cargo features do not. Enabling any time driver presently enables `embassy/time-tick-32768hz`; instead, enable only `embassy/time`.

Additionally, `embassy/time` now functions in the absence of any `embassy/time-tick-*` feature, defaulting to 32768 Hz if left unspecified.


Dirbaio:
Sounds good to me!

Have you tried using it at 1Mhz? Does it work? It'll overflow 15 times per second, and it'll lose counts if user code blocks interrupts for longer than 1/15s, so I'm not really sure if it'll be usable in practice.

willglynn:
It works at 1 MHz… at least, it doesn't totally fail? ¯\_(ツ)_/¯ I might need to revisit the exact timer frequency as my app complexity grows, but I can't say for sure at this point. Certainly not until I get a PLL clock source working, which will be another PR soon™.

It also occurred to me that `embassy/time` could just _pick_ a tick rate in the absence of any feature flags, and that seemed like a good idea, so I added that here too.

willglynn:
Oh, as for 32-bit timers, the particular platform I'm currently targeting is an STM32G0 which [doesn't have any](https://www.st.com/resource/en/datasheet/stm32g030f6.pdf). `TIM1`, `TIM3`, `TIM14`, `TIM16`, `TIM17` are all 16-bit timer peripherals. It is still a Cortex-M0+ core, so it does still have SysTick, which as a 24 bit counter happens to be the largest one available.

Dirbaio:
Can you remove this defaulting? After thinking a bit about it, I think it's better to *not* have a default, and force the user to choose one, even if it's a breaking change now.

The 32768hz default came from nrf's RTC (it's like STM32's LPTIM, it's lowpwer but can only count at low freq), and was carried over to stm32's time driver by inertia (and because it worked OK for 16bit timers). If you use a non-lowpower high-freq timer there's no point in using a low tick rate, it's less precise for no benefit. So defaulting to it for stm32 TIMx (and for nrf's TIMx which is probably going to be added in the future #629) doesn't make much sense. 

About 32bit timers: high tick rates should warn (or error?) if you're using a 16bit timer. We should add support for 32bit timers, but we can do that later.

Also, in stm32g0 you can probably chain two 16bit timers to make a 32bit one (but that'd be yet another driver, in addition to the single 16bit and 32bit drivers...)

willglynn:
I removed the default and added a tick rate feature to all the STM32 examples.

AFAIK even the smallest STM32s have the RTC peripheral. Clock sources include LSI, LSE, or a prescaled HSE, so it's always usable. The RTC ticks internally at 32.768 kHz, supports arbitrary (subsecond) alarms, and can autonomously accumulate long durations without CPU involvement. If the user doesn't need the RTC as a wall clock, it might make a good Embassy time driver.

Dirbaio:
I think you have to update `ci.sh` as well.

re RTC: I think stm32's RTC will be annoying to use because you have to convert "Y/M/D H:M:S" back to "tick count" which is nontrivial math? (nRF's "RTC" is more like LPTIM, just an increasing tick count)

willglynn:
I think that might do it.

(Aside: I was looking for something like `cargo build …` saying what failed, and when I didn't see that I ended up converting `cargo batch` to `set +x; cargo build …; cargo build …`, which at least printed what it was doing when it was doing it. I then figured out where `cargo-batch` lives (crates.io comes up empty), saw that it prints out `features:`, and saw it in the original CI output 🤦‍♂️)

As for the STM32 RTC, yeah, the registers are all Babylonian time/Gregorian calendar BCD. I didn't say it'd be _fun_, I said it's ubiquitous 😆 The time math isn't so bad. The date math… well, in exchange for 12 bytes of RAM, the fast path can be an integer compare and the slow path only needs to happen once a day.

willglynn:
Any ideas what I missed?

```
error: could not compile `embassy` due to previous error
     features: ["default", "embedded-hal-async", "nightly", "time"]
```

PR #791, opened 2022-06-03
Channel clarification

diondokter:
This was unclear from the docs.


lulf:
Thanks :rocket: 

bors r+

PR #792, opened 2022-06-03
Add example for using a Signal.

hydra:
I didn't find an example, so I created one for the STM32H7.  Code based on the nrf mutex.rs example.

### DIFF
--- a/examples/stm32h7/src/bin/signal.rs
+++ b/examples/stm32h7/src/bin/signal.rs
@@ -1,0 +1,41 @@
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+
+use defmt::{info, unwrap};
+use embassy::executor::Spawner;
+use embassy::time::{Duration, Timer};
+use embassy_stm32::Peripherals;
+
+use defmt_rtt as _;
+use embassy::channel::Signal;
+// global logger
+use panic_probe as _;
+
+static SIGNAL: Signal<u32> = Signal::new();
+
+#[embassy::task]
+async fn my_sending_task() {
+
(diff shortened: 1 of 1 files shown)


hydra:
As a side note, the rustdocs for `Signal` says:

`Signals are generally declared as being a static const and then borrowed as required.`

But I didn't find an example of when or why you might need to borrow it.  Clarification of that statement, either as additional code in this example or in the rustdoc for Signal would be appreciated.

examples/stm32h7/src/bin/signal.rs
```diff
@@ -0,0 +1,41 @@
...
+#![feature(type_alias_impl_trait)]
+
+use defmt::{info, unwrap};
+use embassy::executor::Spawner;
+use embassy::time::{Duration, Timer};
+use embassy_stm32::Peripherals;
+
+use defmt_rtt as _;
+use embassy::channel::Signal;
+// global logger
```
Dirbaio, line 12: this comment should go next to `use defmt_rtt`

Dirbaio:
LGTM, other than the comment. Can you `rustfmt` it to make CI happy? Thanks!

hydra:
> LGTM, other than the comment. Can you `rustfmt` it to make CI happy? Thanks!

sure, let me do that and I'll update the PR.

examples/stm32h7/src/bin/signal.rs
hydra: done, I rearranged the imports a little to make things a bit clearer and then reformatted with rustfmt.

Dirbaio:
Thank you!

bors r+

PR #793, opened 2022-06-06
Add extended Signal extended example

hydra:
Adds an example of a more real-world example of using a signal between two tasks in two modules which don't share implementation details of the Signal other than through a defined API.

I'm a rust n00b, so please double check my usage of Rust types.

I can squash the commits post-review.

### DIFF
--- a/examples/stm32h7/src/bin/signal_shared.rs
+++ b/examples/stm32h7/src/bin/signal_shared.rs
@@ -2,27 +2,58 @@
 #![no_main]
 #![feature(type_alias_impl_trait)]
 
+// global logger
 use defmt::{info, unwrap};
+use defmt_rtt as _;
+
+use panic_probe as _;
+
+use embassy::channel::Signal;
 use embassy::executor::Spawner;
 use embassy::time::{Duration, Timer};
 use embassy_stm32::Peripherals;
 
-use defmt_rtt as _;
-// global logger
-use panic_probe as _;
-
+// Expected output (with defmt-print):
+//
+// 0.000030 INFO  registering sender
+// 0.000091 INFO  main task doing other things
+// 0.000152 INFO  getting sender
+// 0.000213 INFO  waiting for signal
+// 1.000152 INFO  main task doing other things
+// 1.000213 INFO  signalling, counter: 0
+// 1.000335 INFO  signalled, counter: 0
+// 2.000213 INFO  main task doing other things
+// 2.000274 INFO  signalling, counter: 1         <-- receiver misses this
+// 3.000274 INFO  main task doing other things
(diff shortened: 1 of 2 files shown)


examples/stm32h7/src/bin/signal_shared.rs
```diff
@@ -0,0 +1,125 @@
...
+    use super::{Duration, Timer};
+    use core::cell::RefCell;
+    use embassy::blocking_mutex::raw::CriticalSectionRawMutex;
+    use embassy::blocking_mutex::Mutex;
+
+    // RefCell because Mutux::lock() give a immutable reference to the closure.
+    // Option because a default value needs to be provided for a static value.
+    // 'static lifetime for the reference because the signal needs to live forever.
+    // Mutex because otherwise you get a 'cannot be shared between threads safely' due to the static lifetime requirement.
+    static SIGNAL: Mutex<CriticalSectionRawMutex, RefCell<Option<&'static Signal<u32>>>> =
```
Dirbaio, line 89: This is very non-idiomatic. Signal is designed to not need extra mutexes, it already does mutexing internally.

If you want two mods to talk to each other without one poking into the other's statics, you can pass the signal as a task argument: `#[embassy::task] pub async fn my_receiving_task(signal: &'static Signal<u32>) { ..`

Dirbaio: The cleanest would probably be `main` having the `static` and passing `&'static Signal` to both tasks. Each task declares the resources it needs, and main "wires" them together.

hydra: yeah, there's so many ways of doing it.  I was looking for a solution where main knows nothing about the internal needs of the other tasks other than they both might need an init() to be called on them.  Yes, I'm aware that I could pass the static from the `main` function, again the idea was to hide the internals.

Yes, agree that the need for an additional mutex is far from ideal.  The second mutex is only for initialisation and also so that task excution order doesn't matter.

Any other suggestions?

Dirbaio: > I was looking for a solution where main knows nothing about the internal needs of the other tasks

If main forgets to call `init_sending_task` or calls it in the wrong order, `my_receiving_task` will panic at the unwrap. So main still has to know some requirements, and on top of that they're no longer checked by the compiler (you get a runtime panic instead of a compile fail). It's not a great tradeoff.

hydra: ok, I'll poke the example around and update the PR as per your suggestions so that at least there is an example for others even though it doesn't meet my initial requirements.

hydra:
I don't currently have time to complete this, I'll open a new PR if I ever make time to revisit this.

PR #797, opened 2022-06-07
Use correct index and bank for stm32 flash

lulf:
The page index was wrong because it doesn't take FLASH_BASE into account. 

For l4, the page index register also depends on a bank selection register.


lulf:
Yes, but it's translated  to FLASH_BASE addresses in the 'outer' Flash blocking_write before handed to the family-specific code.

Dirbaio:
oh, oops! LGTM then, thank you! :) 

bors r+

PR #800, opened 2022-06-09
Implement support for LPUART

FrozenDroid:
Have tested this with BufferedUart, seems to work well. I have not tested uart v1.
Creating this PR now to allow for discussion on whether or not this is the best way to do it.

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -36,6 +36,8 @@ embassy-embedded-hal = {version = "0.1.0", path = "../embassy-embedded-hal" }
 atomic-polyfill = "1.0.1"
 defmt = { version = "0.3", optional = true }
 log = { version = "0.4.14", optional = true }
+nb = "1.0.0"
+cfg-if = "1.0.0"
 cortex-m-rt = ">=0.6.15,<0.8"
 cortex-m = "0.7.6"
 critical-section = "1.1"
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -1,5 +1,7 @@
 use core::sync::atomic::{compiler_fence, Ordering};
 
+use embassy_hal_common::impl_peripheral;
+
 use crate::pac::dma::vals;
 use crate::{pac, peripherals};
 
@@ -34,6 +36,10 @@ impl<T: Channel> Dma<T> {
         }
     }
 }
+
+pub struct NoDma;
+
+impl_peripheral!(NoDma);
 
 mod sealed {
     use super::*;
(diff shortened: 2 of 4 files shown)


embassy-stm32/build.rs
```diff
@@ -255,6 +255,16 @@ fn main() {
         (("usart", "CTS"), quote!(crate::usart::CtsPin)),
         (("usart", "RTS"), quote!(crate::usart::RtsPin)),
         (("usart", "CK"), quote!(crate::usart::CkPin)),
+        (("lpuart", "TX"), quote!(crate::usart::TxPin)),
+        (("lpuart", "RX"), quote!(crate::usart::RxPin)),
+        (("lpuart", "CTS"), quote!(crate::usart::CtsPin)),
+        (("lpuart", "RTS"), quote!(crate::usart::RtsPin)),
+        (("lpuart", "CK"), quote!(crate::usart::CkPin)),
+        (("lpuart", "TX"), quote!(crate::usart::TxPin)),
```
Dirbaio, line 263: these are duplicated

FrozenDroid: Yeah, though the usart also has duplicated ones, so I thought it was for a reason.

embassy-stm32/src/usart/mod.rs
```diff
@@ -455,55 +455,91 @@ unsafe fn clear_interrupt_flags(_r: crate::pac::usart::Usart, _sr: regs::Sr) {
...
 }
 
-pin_trait!(RxPin, Instance);
-pin_trait!(TxPin, Instance);
-pin_trait!(CtsPin, Instance);
-pin_trait!(RtsPin, Instance);
-pin_trait!(CkPin, Instance);
+pub trait BasicUartInstance: sealed::BasicUartInstance {}
+
+pub trait UartInstance: sealed::UartInstance {}
```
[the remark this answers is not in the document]
FrozenDroid: Renamed to BasicInstance and FullInstance

FrozenDroid:
> Have you checked the version of UART and LPUART matches in all chips?
I haven't, how would I go about searching through all the files to check this?

Dirbaio:
manually :smiling_face_with_tear: ... Checking one chip per family should be safe enough.

FrozenDroid:
Ah dear, already found an example of a chip (STM32L4A6) that has a v1 LPUART but a v2 USART...
🙃 

Where do we go from here?

darkwater:
I've been asked to take over here, but I'm not sure how to properly resolve the conflicts. Apparently this is a PR from our master branch, which also seems to have other work on it. Should I open a new PR from a dedicated branch, with just the LPUART commits, rebased on current upstream master?

Dirbaio:
yes! there should be no unrelated commits. If you have to open a new PR then so be it :)

PR #803, opened 2022-06-10
Initial independent watchdog implementation

FrozenDroid:


embassy-stm32/src/wdg/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+use stm32_metapac::iwdg::vals::Key;
+
+pub use stm32_metapac::iwdg::vals::Pr as Prescaler;
+
+pub struct IndependentWatchdog<'d, T: Instance> {
+    wdg: PhantomData<&'d mut T>,
+}
+
+impl<'d, T: Instance> IndependentWatchdog<'d, T> {
+    pub fn new(_instance: impl Unborrow<Target = T>, presc: Prescaler) -> Self {
```
chemicstry, line 15: This should be `impl Unborrow<Target = T> + 'd`, otherwise unborrow will only live for the length of the call to `new()` and not for the entire `IndependentWatchdog` lifetime `'d`, which is unsound

embassy-stm32/src/wdg/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+        IndependentWatchdog {
+            wdg: PhantomData::default(),
+        }
+    }
+
+    pub unsafe fn unleash(&mut self) {
+        T::regs().kr().write(|w| w.set_key(Key::START));
+    }
+
+    pub unsafe fn feed(&mut self) {
```
lulf, line 33: I've seen the term `pet` being used in nRF HAL, would be good to standardize on the same terms

lulf (commented):
Nice! 🚀

lulf (approved):
LGTM, just tell bors to r+ it when it compiles :) 

bors d+

FrozenDroid:
Not sure why it didn't compile in the first place, but let's try it again :o 

bors r+

lulf (approved):
LGTM!

PR #804, opened 2022-06-10
Preliminary RISCV support

MabezDev:
- ~~Moves the default Interrupt implementation into a cortex_m specific module~~
- Adds a RISCV32 executor based on [osobiehl](https://github.com/osobiehl)'s work in [esp32c3 mess work](https://github.com/osobiehl/riscv-embassy-mess-work)

(FYI esp implementation of embassy traits etc, is being developed [here](https://github.com/esp-rs/esp-hal/tree/feature/embassy))

[bonus ascii cinema](https://asciinema.org/a/500857 )

### DIFF
--- a/embassy/src/executor/arch/riscv32.rs
+++ b/embassy/src/executor/arch/riscv32.rs
@@ -1,0 +1,74 @@
+use core::marker::PhantomData;
+use core::ptr;
+
+use atomic_polyfill::{AtomicBool, Ordering};
+
+use super::{raw, Spawner};
+
+/// global atomic used to keep track of whether there is work to do since sev() is not available on RISCV
+///
+static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
+
+/// RISCV32 Executor
+pub struct Executor {
+    inner: raw::Executor,
+    not_send: PhantomData<*mut ()>,
+}
+
+impl Executor {
+    /// Create a new Executor.
+    pub fn new() -> Self {
+        Self {
+            // use Signal_Work_Thread_Mode as substitute for local interrupt register
+            inner: raw::Executor::new(
+                |_| {
+                    SIGNAL_WORK_THREAD_MODE.store(true, Ordering::SeqCst);
+                },
+                ptr::null_mut(),
+            ),
+            not_send: PhantomData,
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


embassy/src/executor/arch/riscv32.rs
```diff
@@ -0,0 +1,155 @@
...
+                //will only set this value to true.
+                critical_section::with(|_| {
+                    // if there is work to do, loop back to polling
+                    // TODO can we relax this?
+                    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
+                        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
+                    }
+                    // if not, wait for interrupt
+                    else {
+                        riscv::asm::wfi();
```
jessebraham, line 66: A little nit-picky, but do we really need to pull in the `riscv` dependency just for this one call? Could this be replaced with `core::arch::asm!("wfi")` instead to eliminate the dependency, or is there some reason to keep this?

MabezDev: Good call!

Dirbaio:
@MabezDev LGTM, could you squash?

Dirbaio (approved):
:rocket:!!! Thank you! :) 

bors r+

embassy/src/executor/arch/riscv32.rs
```diff
@@ -0,0 +1,74 @@
...
+        loop {
+            unsafe {
+                self.inner.poll();
+                // we do not care about race conditions between the load and store operations, interrupts
+                //will only set this value to true.
+                critical_section::with(|_| {
+                    // if there is work to do, loop back to polling
+                    // TODO can we relax this?
+                    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
+                        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
```
huntc, line 63: Would compare_exchange work here?

Dirbaio: doesn't work in targets without CAS atomic support (`riscv32imc` instead of `riscv32imac`)

huntc: I see. Thanks.

PR #805, opened 2022-06-10
Preliminary Xtensa support

MabezDev:
Based on the work in #804.

I hope non-upstream target support is acceptable :).

### DIFF
--- a/embassy/src/executor/arch/xtensa.rs
+++ b/embassy/src/executor/arch/xtensa.rs
@@ -1,0 +1,75 @@
+use core::marker::PhantomData;
+use core::ptr;
+
+use atomic_polyfill::{AtomicBool, Ordering};
+
+use super::{raw, Spawner};
+
+/// global atomic used to keep track of whether there is work to do since sev() is not available on Xtensa
+///
+static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
+
+/// Xtensa Executor
+pub struct Executor {
+    inner: raw::Executor,
+    not_send: PhantomData<*mut ()>,
+}
+
+impl Executor {
+    /// Create a new Executor.
+    pub fn new() -> Self {
+        Self {
+            // use Signal_Work_Thread_Mode as substitute for local interrupt register
+            inner: raw::Executor::new(
+                |_| {
+                    SIGNAL_WORK_THREAD_MODE.store(true, Ordering::SeqCst);
+                },
+                ptr::null_mut(),
+            ),
+            not_send: PhantomData,
+        }
+    }
+
+    /// Run the executor.
+    ///
+    /// The `init` closure is called with a [`Spawner`] that spawns tasks on
+    /// this executor. Use it to spawn the initial task(s). After `init` returns,
+    /// the executor starts running the tasks.
+    ///
+    /// To spawn more tasks later, you may keep copies of the [`Spawner`] (it is `Copy`),
+    /// for example by passing it as an argument to the initial tasks.
+    ///
+    /// This function requires `&'static mut self`. This means you have to store the
+    /// Executor instance in a place where it'll live forever and grants you mutable
+    /// access. There's a few ways to do this:
+    ///
+    /// - a [Forever](crate::util::Forever) (safe)
+    /// - a `static mut` (unsafe)
+    /// - a local variable in a function you know never returns (like `fn main() -> !`), upgrading its lifetime with `transmute`. (unsafe)
+    ///
+    /// This function never returns.
+    pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+        init(self.inner.spawner());
+
+        loop {
+            unsafe {
+                self.inner.poll();
+                // we do not care about race conditions between the load and store operations, interrupts
+                // will only set this value to true.
+                // if there is work to do, loop back to polling
+                // TODO can we relax this?
+                critical_section::with(|_| {
+                    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
+                        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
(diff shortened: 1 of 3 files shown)


embassy/src/executor/arch/xtensa.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::interrupt::{Interrupt, InterruptExt};
+use atomic_polyfill::{AtomicBool, Ordering};
+use core::marker::PhantomData;
+use core::ptr;
+
+/// global atomic used to keep track of whether there is work to do since sev() is not available on RISCV
+///
+static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
+
+/// RISCV32 Executor
```
jessebraham, line 11: A couple out-of-date comments here

embassy/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 #![cfg_attr(not(any(feature = "std", feature = "wasm")), no_std)]
-#![cfg_attr(feature = "nightly", feature(generic_associated_types, type_alias_impl_trait))]
+#![cfg_attr(
+    feature = "nightly",
+    feature(generic_associated_types, type_alias_impl_trait, asm_experimental_arch)
```
Dirbaio, line 4: can you set `asm_experimental_arch` only when `feature = "nightly"` AND building for xtensa?

embassy/Cargo.toml
```diff
@@ -40,6 +40,9 @@ defmt-timestamp-uptime = ["defmt"]
 # Enabling it directly without supplying a time driver will fail to link.
 time = []
 
+# enable the Xtensa executor
+xtensa = []
```
Dirbaio, line 44: this feature is unused (there's no `cfg(feature="xtensa")` anywhere in the code).

Instead, could you make it so that when building for xtensa, you get the executor if `nightly` is enabled, and you don't get it otherwise? There's no need to have a separate feature.

embassy/src/executor/arch/xtensa.rs
```diff
@@ -0,0 +1,72 @@
...
+                // we do not care about race conditions between the load and store operations, interrupts
+                // will only set this value to true.
+                // if there is work to do, loop back to polling
+                // TODO can we relax this?
+                if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
+                    SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
+                }
+                // if not, wait for interrupt
+                else {
+                    // waiti sets the PS.INTLEVEL therefore no critical section is needed
```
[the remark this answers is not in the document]
MabezDev: So in my testing, I've verified that the TRM is indeed correct. However, I also found some more notes on implemented critical sections on Xtensa, they should actually be done by increasing the `PS.INTLEVEL` such that `PS.INTLEVEL > any possible interrupt level` instead of disabling interrupts via the `INTENABLE` register. I need to do a bit more reading to verify that no interrupts can exist at the max (which is 15), the TRM is kinda vague on that. I guess NMI might be at that level, but what can ya do about that :D. By modifying `PS.INTLEVEL`, the `waiti` instruction can atomically reduce the `PS.INTLEVEL` whilst slipping into sleep mode, hence no race conditions :D.

So what this actually means is we need `critical_section` impls for Xtensa. It seems the path forward for `critical_section` is that every HAL/PAC will have a per chip cs impl that can handle multi core etc, so I've opened https://github.com/esp-rs/esp-hal/issues/87.

I'll make sure all this theory work actually works on hardware and get back to you :).

MabezDev:
I've now reverted to using the critical section to avoid the race condition, for reference the CS impl I used looked like this:

```rust
struct CriticalSection;
critical_section::custom_impl!(CriticalSection);

static mut VPS: u32 = 0;
unsafe impl critical_section::Impl for CriticalSection {
    unsafe fn acquire() -> u8 {
        core::arch::asm!("rsil {0}, 15", out(reg) VPS);
        0
    }

    unsafe fn release(_token: u8) {
        core::arch::asm!("wsr.ps {0}", in(reg) VPS)
    }
}
```
With the next release of `critical_section`, I believe I can get rid of the `static mut` and pass it through the variable-sized token instead to restore processor state.

Dirbaio:
Hmm, if `waiti 0` sets `INTLEVEL=0`, it's effectively exiting the critical section. I wonder if this can be used to create unsoundness, for example calling `run()` within a CS will cause it to end early, but I can't immediately see how to exploit that because run() is noreturn...

MabezDev:
I'm not sure there is anything we can do here, `waiti` doesn't restore the previous `INTLEVEL` so you're right it does effectively end the CS there. However, this is only possible because we are using unsafe and we make sure that we don't include anything that needs to be inside the CS after the `waiti` call. This is sort of similar to cortex-m in that you can call `interrupt::enable` inside the critical section to end the critical section there but of course, this method is marked as unsafe in cortex-m.

MabezDev:
FYI [this file](https://github.com/espressif/esp-idf/blob/e8d172eb87e606c52c6a995d118c1eb0f4810abc/components/xtensa/include/xtensa/xtruntime.h#L120-L136) is what I am basing my implementation on, which came straight from Tensilica (now Cadence). Might be worth a quick skim read, just in case I'm missing something obvious :sweat_smile:.

MabezDev:
To the best of my knowledge, I think this is the best we can do for now. I've added a comment to warn that the CS ends early when using `waiti`. It kinda sucks it works this way, it means we can't implement any CPU usage tracking etc. Hopefully, I can find a better way to do this in the future.

I hope this is acceptable as is, I'd be happy to manage any issues relating to Xtensa (or RISCV) if this gets merged :).

Dirbaio (approved):
cool! Let's do this then! We can always change it later if it turns out to be troublesome :rocket: 

Thank you! :) 

bors r+

PR #806, opened 2022-06-11
Add embassy-cortex-m crate.

Dirbaio:
- Move Interrupt and InterruptExecutor from `embassy` to `embassy-cortex-m`.
- Move Unborrow from `embassy` to `embassy-hal-common` (nothing in `embassy` requires it anymore)
- Move PeripheralMutex from `embassy-hal-common` to `embassy-cortex-m`.


embassy-macros/src/macros/cortex_m_interrupt_declare.rs
```diff
@@ -9,8 +9,7 @@ pub fn run(name: syn::Ident) -> Result<TokenStream, TokenStream> {
     let result = quote! {
         #[allow(non_camel_case_types)]
         pub struct #name_interrupt(());
-        unsafe impl ::embassy::interrupt::Interrupt for #name_interrupt {
-            type Priority = crate::interrupt::Priority;
+        unsafe impl ::embassy_cortex_m::interrupt::Interrupt for #name_interrupt {
```
MabezDev, line 12: Should these macros be moved into the embassy-cortex-m crate? Feels a little odd having them in here given that only `cortex-m` chips will ever use them.

Dirbaio: I kept them here because I didn't want to make yet another `embassy-cortex-m-macros` crate. :sweat_smile: 

Either way, it's not "public API", users aren't supposed to use `embassy-macros` directly, but use the reexports from the other crates.

MabezDev: Ah yes, I forgot that proc-macros need to be in their own crate. I wonder if that limitation will be lifted at some point.

MabezDev (approved):
LGTM!

examples/stm32f4/src/bin/multiprio.rs
```diff
@@ -62,11 +62,11 @@ use defmt_rtt as _; // global logger
 use panic_probe as _;
 
 use cortex_m_rt::entry;
-use embassy::executor::{Executor, InterruptExecutor};
-use embassy::interrupt::InterruptExt;
 use embassy::time::{Duration, Instant, Timer};
 use embassy::util::Forever;
+use embassy_cortex_m::executor::{Executor, InterruptExecutor};
```
lulf, line 67: Could it make sense to reexport these in embassy-stm32/embassy-nrf/embassy-rp to avoid having to add the embassy-cortex-m to dependencies?

lulf (commented):
Makes sense to split up the module. Would be nice if InterruptExecutor etc. could be reexported by the HALs to remove the need for adding another dependency in application Cargo.tomls

examples/stm32f4/src/bin/multiprio.rs
Dirbaio: Nice idea, done!

Dirbaio:
bors r+

:rocket:

PR #807, opened 2022-06-12
Rename channel to mpmc

huntc:
I've renamed the channel module for the MPMC as `mpmc`. There was a previous debate about this, but I feel that the strategy here avoids importing `channel::channel`. The change leaves `signal::Signal`, but I think that's ok. It is all a bit subjective of course. The bottom line for me is that I really like the term `mpmc` - it means something to me and aligns with broader naming e.g. in Tokio.

### DIFF
--- a/embassy/src/channel/mod.rs
+++ b/embassy/src/channel/mod.rs
@@ -1,7 +1,5 @@
 //! Async channels
 
-pub mod channel;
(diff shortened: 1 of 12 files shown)


lulf (approved):
I like the mpmc name better as well 👍

PR #812, opened 2022-06-15
Fix typo in subghz/packet_params.rs

danbev:


danbev:
Sorry, this is already fixed in https://github.com/embassy-rs/embassy/pull/811.

PR #813, opened 2022-06-15
Document remaining public APIs in embassy crate

lulf:
This also includes the README.md in the toplevel crate documentation, fixing a few formatting issues with it as well. 

Please review the mutex in detail in case I've misunderstood/missed a few things there.

### DIFF
--- a/embassy/src/blocking_mutex/mod.rs
+++ b/embassy/src/blocking_mutex/mod.rs
@@ -1,14 +1,23 @@
-//! Blocking mutex (not async)
-
+//! Blocking mutex.
+//!
+//! This module provides a blocking mutex that can be used to synchronize data.
 pub mod raw;
 
 use core::cell::UnsafeCell;
 
 use self::raw::RawMutex;
 
-/// Any object implementing this trait guarantees exclusive access to the data contained
-/// within the mutex for the duration of the lock.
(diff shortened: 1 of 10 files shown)


embassy/src/blocking_mutex/mod.rs
```diff
@@ -1,14 +1,23 @@
...
 use core::cell::UnsafeCell;
 
 use self::raw::RawMutex;
 
-/// Any object implementing this trait guarantees exclusive access to the data contained
-/// within the mutex for the duration of the lock.
-/// Adapted from <https://github.com/rust-embedded/mutex-trait>.
+/// Blocking mutex (not async)
+///
+/// Provides a blocking mutual exclusion primitive backed by an implementation of [`raw::RawMutex`].
```
huntc, line 12: One thing that Dario mentioned, which helped me a lot, was that the blocking mutex is designed for bringing your own interior mutability. They are also low level on this basis. In particular, locks are intended to be short lived and must not cross async boundaries (and their design enforces this too). OTOH, the async mutex can have a lock spanning awaits, but also comes with a little more overhead.

huntc (approved):
Nice work! You beat me to it. :-)

embassy/src/blocking_mutex/mod.rs
lulf: Thanks, I've emphasized this point

PR #814, opened 2022-06-15
Extract setting of lora modulation params

danbev:
This commit suggests extracting the lora modulation parameters into a
separate function which can then be called from both the do_tx, and
the do_rx functions.


lulf:
LGTM :rocket: 

bors r+

PR #817, opened 2022-06-16
Added a pubsub channel implementation

diondokter:
This is similar to Tokio's Broadcast channel, except that it doesn't allocate.

The publishers and subscribers are dynamic. They use an &dyn channel reference because it's really annoying to have to specify the mutex and const generics every time.
Do we need fully generic types as well?


lulf (commented):
:rocket: thanks for doing the work!

Is there a practical need for the subscriber knowing how far behind it is? I'm thinking that getting rid of the message id tracking could simplify the subscriber.

I also think there is another type of pubsub with stronger guarantees, where every subscriber would get their own queue so you'd avoid them loosing messages. The way this implementation behaves is similar to an event log like Kafka. Nothing wrong with that, just maybe we need to think if having variant with stronger guarantees could allow this variant to be simplified, if that makes sense :)

diondokter:
I don't see the advantage of every subscriber having its own queue. They can still loose messages because of the fixed queue size. So, you'd only use more memory. Can you maybe clarify what the advantage is?

WRT the message id tracking, if every subscriber had its own queue you wouldn't need it. But in this system with a shared queue, every subscriber must know which items it has already read. It could have read none of the items in the queue, some or all. So you need to keep track of that in some way.
The `Lagged` error is something we would need. For us it would be a major error if a message wasn't read, so we do need that information. How much we lagged is not very important but the info is there anyways and it's good to see how bad the situation is.
What is possible is to create a convenience method on top that ignores the lag and just tries to give you the next message, that's not a bad idea IMO.

About the name bikeshedding, I don't have strong opinions. Implementing the stream traits is a good idea!
Maybe we should keep the names the same as what Tokio does? That's probably not a bad idea... I'll check

embassy/src/channel/pubsub.rs
```diff
@@ -0,0 +1,590 @@
...
+    subscriber_index: usize,
+    /// The message id of the next message we are yet to receive
+    next_message_id: u64,
+    /// The channel we are a subscriber to
+    channel: &'a dyn PubSubBehavior<T>,
+}
+
+impl<'a, T: Clone> Subscriber<'a, T> {
+    /// Wait for a published message
+    pub fn wait<'s>(&'s mut self) -> SubscriberWaitFuture<'s, 'a, T> {
```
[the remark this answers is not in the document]
diondokter: Done

diondokter:
The `futures::Sink` trait is pretty annoying and its interface doesn't really map very nicely on the current API. Maybe let's wait with that one until somebody actually needs that...

lulf:
> I don't see the advantage of every subscriber having its own queue. They can still loose messages because of the fixed queue size. So, you'd only use more memory. Can you maybe clarify what the advantage is?
>

That's a good point. The only thing I can think of is that it simplifies the bookeeping, but I agree less memory usage is more important.
 
> WRT the message id tracking, if every subscriber had its own queue you wouldn't need it. But in this system with a shared queue, every subscriber must know which items it has already read. It could have read none of the items in the queue, some or all. So you need to keep track of that in some way. The `Lagged` error is something we would need. For us it would be a major error if a message wasn't read, so we do need that information. How much we lagged is not very important but the info is there anyways and it's good to see how bad the situation is. What is possible is to create a convenience method on top that ignores the lag and just tries to give you the next message, that's not a bad idea IMO.
> 

Yeah, I think maybe the StreamExt implementation could do that for instance. Wrt. loosing messages being a major error, wouldn't it be another option to have the publisher wait until all subscribers have consumed the message before continuing? Maybe this could be another publisher type, but it seems easier to catch/prevent the error this way.

> About the name bikeshedding, I don't have strong opinions. Implementing the stream traits is a good idea! Maybe we should keep the names the same as what Tokio does? That's probably not a bad idea... I'll check
> 

To me, `wait` and even `next` is not that descriptive, maybe `next_message` is better. I'm fine either way, just wanted to start a discussion now before the api is settled. 

> edit: Ok Tokio uses sender/receiver instead of publisher/subscriber so their naming doesn't really match

lulf:
> The `futures::Sink` trait is pretty annoying and its interface doesn't really map very nicely on the current API. Maybe let's wait with that one until somebody actually needs that...

I was thinking the `futures::Stream` trait

diondokter:
>Yeah, I think maybe the StreamExt implementation could do that for instance. Wrt. loosing messages being a major error, wouldn't it be another option to have the publisher wait until all subscribers have consumed the message before continuing? Maybe this could be another publisher type, but it seems easier to catch/prevent the error this way.

That's already kind of the case. A subscriber can only lag if a `publish_immediate` method has been called. This is a balance between practicalities. The channel needs capacity for every publisher that might wait, while no capacity is needed for a publisher that only calls `publish_immediate`.

> I was thinking the futures::Stream trait

I know! It just made me wonder about the Sink trait.
I have now implemented Stream in a way where it ignores lag results alongside some lag free methods.
I've named those methods `_pure` and I'm open to suggestions :P

embassy/src/channel/pubsub.rs
```diff
@@ -0,0 +1,651 @@
...
+struct PubSubState<T: Clone, const CAP: usize, const SUBS: usize, const PUBS: usize> {
+    /// The queue contains the last messages that have been published and a countdown of how many subscribers are yet to read it
+    queue: Deque<(T, usize), CAP>,
+    /// Every message has an id.
+    /// Don't worry, we won't run out.
+    /// If a million messages were published every second, then the ID's would run out in about 584942 years.
+    next_message_id: u64,
+    /// Collection of wakers for Subscribers that are waiting.  
+    /// The [Subscriber::subscriber_index] field indexes into this array.
+    subscriber_wakers: [Option<WakerRegistration>; SUBS],
```
Dirbaio, line 239: `WakerRegistration` is already kinda a `Option<Waker>`, no need to option-wrap it again.

embassy/src/channel/pubsub.rs
```diff
@@ -0,0 +1,651 @@
...
+    pub fn immediate_publisher(&self) -> ImmediatePublisher<'_, T> {
+        ImmediatePublisher { channel: self }
+    }
+}
+
+impl<M: RawMutex, T: Clone, const CAP: usize, const SUBS: usize, const PUBS: usize> PubSubBehavior<T>
+    for PubSubChannel<M, T, CAP, SUBS, PUBS>
+{
+    fn try_publish(&self, message: T) -> Result<(), T> {
+        self.inner.lock(|inner| {
```
Dirbaio, line 98: all the functions here acquire the mutex internally, and you call them from the future/stream impls. (for example, a subscriber will do `get_message` then `register_subscriber_waker`). This has 2 issues:

1. It's slightly slower because you're acquiring the mutex twice. 
2. It can lead to race conditions, because "in between" of the two calls, another thread can sneak in and change the state. For example, when receiving in a subscriber: 
    - Call `get_message`, there are no messages so it returns None.
    - Another thread sneaks in and pushes a message
    - Call `register_subscriber_waker`. The waker gets registered, and NOT woken because it was registered AFTER the message came.
    - The subscriber deadlocks, it yields even though there's a message in the queue and the waker never gets woken.

To fix these, you could move all these fns into `PubSubState` taking `&mut self`, and make the futs acquire the mutex themselves.

Dirbaio:
Some general throughts:

- There's A LOT of "unchecked" stuff. Have you tested it has an impact on the generated code? the compiler is very good at optimizing these out. It'd be great if we could get this to "no unsafe", like the current `Channel`.
- The forced use of `dyn` feels a bit strange. On one hand you're micro-optimizing the code with lots of "unchecked", but on the other hand you use `dyn` which probably has more (negative) impact than the (positive) impact of the "unchecked"s. Maybe you could do like the `Channel`, where publisher/subscriber don't use dyn, and the user can optionally opt into dyn if the generics are too annoying?

embassy/src/channel/pubsub.rs
diondokter: I knew that. The outer option is to keep track of which slot is populated or free.
I could make an enum `Slot` with `Taken(Waker)` and `Free`, but that seems a bit overkill to me.

embassy/src/channel/pubsub.rs
```diff
@@ -0,0 +1,613 @@
...
+    }
+}
+
+/// Warning: The stream implementation ignores lag results and returns all messages.
+/// This might miss some messages without you knowing it.
+impl<'a, T: Clone> futures::Stream for Subscriber<'a, T> {
+    type Item = T;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = unsafe { self.get_unchecked_mut() };
```
Dirbaio, line 318: unsafe is not needed if Subscriber is Unpin

diondokter: Oh cool!

embassy/src/channel/pubsub.rs
diondokter: Yes, well, I do expect people to use them long term indeed. But I can imagine shorter lived ones too.
So to make it ergonomic, I think it should be either this API or a `Split`-like api where you can get them all once at the same time.

Personally I think this API is the better of the two and it only really costs an extra option per waker, so like one byte? (Unless the compiler is able to combine them)

Dirbaio: WakerRegistration is 4 bytes (without `executor-agnostic`) and has no niches, so `Option<WakerRegistration>` will be 8 bytes due to alignment.

there's also a 3rd possibility: allow creating them dynamically but not "recycle" slots, once you've created all N you can't create more even if you drop the old ones.

diondokter: Hmmmm, that doesn't *feel* right.
But you're right, an extra 4 bytes is a lot.

I could remove the options if I could query the WakerRegistration if they have a waker registered.
That would probably be the best solution right?

Instead of tracking a fixed index for each Sub/Pub, they just track the index they're using at that moment.
Then when a spot needs to be found, it just takes the first one that has no waker registered yet.

Dirbaio: That's not a bad idea at all! :) 

Maybe have it into a "MultiWakerRegistration" where you can register up to N wakers? It could be useful in other scenarios. 

Also, If registration is "smart" in that it doesn't register an already-registered waker, you don't even need to track indices in pub/subs at all, you just register the waker every time and that's it.

diondokter: > Also, If registration is "smart" in that it doesn't register an already-registered waker, you don't even need to track indices in pub/subs at all, you just register the waker every time and that's it.

You'd still need to know which one to remove on the drop. Oh and you'd need to be able to unregister a waker.

> Maybe have it into a "MultiWakerRegistration" where you can register up to N wakers? It could be useful in other scenarios.

Sure, that'd simplify some things in the pubsub impl.

Dirbaio: If you don't unregister, it's possible the `MultiWakerRegistration` will get full, yes. In that case you can wake+clear all, then register the waker you were trying to register. If the other tasks are still interested, they'll register themselves again and the thing will settle down in 1 iteration because there can't be more than N interested tasks. 

This is exactly what the current WakerRegistration does, it wakes the old waker if there's one, then replaces it. WakerRegistration is kinda a MultiWakerRegistration of size 1.

This may seem inefficient, but it should happen very rarely, and it yields the best code size vs trying to actively "clean up" no-longer-interested wakers.

diondokter: Ah right of course!
This saves another usize from the pubs/subs

diondokter: Alright, I changed it. It costs 8 bytes extra in the channel, but it saves 4 bytes per pub/sub capacity and 4 bytes per pub/sub instance. Well worth it.

diondokter:
Ok, I think I have implemented all the feedback now.

All unsafe is gone and you can now choose between dyn channels and generic channels.

For that last bit I played a little with the PubSubBehavior trait. Pub and Sub now have a generic reference through that trait.
To save a little bit more typing, I made newtypes `DynPublisher` and `Publisher` that simply hold a `Pub` and deref it.
It's a bit verbose, but it saves on having double implementations.

lulf:
Excellent, thank you so much for writing this 🚀 🚀 

bors r+

PR #822, opened 2022-06-18
Remove the authors field from Cargo.tomls

Dirbaio:
It currently contains whoever was first to write some code for the crate,
even if many more people have contributed to it later.

The field is "sort of" deprecated, it was made optional recently:
https://rust-lang.github.io/rfcs/3052-optional-authors-field.html

Due the the reasons listed there I believe removing it is better than
setting it to generic fluff like "The Embassy contributors".


lulf (approved):
I agree. I've always found it becomes out of date at some point.

PR #824, opened 2022-06-21
embassy/time: round up by default in duration conversions. Fixes #823

### LINKED ISSUE #823: DelayUs rounded down to zero
I have synchronous driver that uses `embedded_hal::blocking::delay::DelayUs` for the [reset sequence](https://github.com/almindor/st7789/blob/24968a36149751cef2c9ec28048906f2f07f8d3f/src/lib.rs#L166). However `delay_source.delay_us(10);` gets rounded down to 0, which means the reset pin is 0 seconds active. 

I was thinking that the delay implementation should at least wait 1 tick. Would that the an acceptable solution in this project? If yes, I will create a PR for it.

Dirbaio:


Dirbaio:
I'm not sure whether to round up in `as_*` by default. It prevents the same footgun in "reverse" direction, if you do something like `delay_ms(duration.as_millis())` to sleep with something other than embassyy... but it feels weird that a 1us duration will get rounded up to 1s by `.as_secs()`...

Rust's std rounds down in `as_*`. (and doesn't need rounding in `from_*` because it represents the duration as `(secs, nanos)`)

lulf:
I think to follow POLA, it's better to follow "std" semantics by default.

lulf:
Hmm, yeah... I think round up is fine, but I don't have a strong argument either way tbh.

lulf (approved):
LGTM, I think it should be merged as is.

Dirbaio:
current code in this PR rounds *up* in `as_*` though, I guess you wanted to LGTM "round *down* in `as_*`, *up* in `from_`"?

lulf:
Uhm 😸 yes! I got confused. Whatever std does for as_  (round down). And up for from.

Dirbaio:
cool! Changed the PR to do just that :) 

bors r+

PR #827, opened 2022-06-22
Fix PWM for advanced timers

chemicstry:
Advanced timers have additional BDTR register, which has a global output enable bit and it is disabled by default.

Also added an example for F4, but it will only work once https://github.com/embassy-rs/stm32-data/pull/149 is merged. We can also move it to some other chip, but I don't have anything else to test on atm.

### DIFF
--- a/embassy-stm32/src/pwm/mod.rs
+++ b/embassy-stm32/src/pwm/mod.rs
@@ -55,6 +55,9 @@ pub(crate) mod sealed {
     use super::*;
 
     pub trait CaptureCompare16bitInstance: crate::timer::sealed::Basic16bitInstance {
+        /// Global output enable. Does not do anything on non-advanced timers.
+        unsafe fn enable_outputs(&mut self, enable: bool);
+
         unsafe fn set_output_compare_mode(&mut self, channel: Channel, mode: OutputCompareMode);
 
         unsafe fn enable_channel(&mut self, channel: Channel, enable: bool);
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/pwm/mod.rs
```diff
@@ -118,38 +123,7 @@ macro_rules! impl_compare_capable_16bit {
...
-                    .ccr(channel.raw())
-                    .modify(|w| w.set_ccr(value));
-            }
-
-            unsafe fn get_max_compare_value(&self) -> u16 {
-                use crate::timer::sealed::GeneralPurpose16bitInstance;
-                Self::regs_gp16().arr().read().arr()
-            }
-        }
+        impl_compare_capable_16bit!($inst);
```
chemicstry, line 126: Replaced duplicated code here with an existing macro. Not sure why it wasn't used before? The code is identical.

Dirbaio (approved):
Thank you!:) 

bors r+

smbolton:
Thanks, @chemicstry, I was just preparing functionally identical code to do the same. I tested mine on stm32f103c8t6, stm32f401ccu6, stm32f303vct6, and stm32h743vit6 (using embassy-stm32 only, no async, and my [embassy/stm32-data#149](https://github.com/embassy-rs/stm32-data/pull/149)). Yours should work the same, I'm sure, but I'm not going to dig the boards out again to try it.

PR #831, opened 2022-06-24
Move bootloader main to examples

lulf:
This should remove some confusion around embassy-boot-* being a library
vs. a binary. The binary is now an example bootloader instead.


Dirbaio (approved):
LGTM, I agree it's clearer now!

PR #839, opened 2022-07-01
Misc LoRaWAN improvements

timokroeger:
Trying too get `embassy-lora` running on a [LoRa-E5 Dev Board](https://wiki.seeedstudio.com/LoRa_E5_Dev_Board/).
I can see the join message arriving in the The Things Network console but the device does not receive the accept message yet.
Opening this PR anyway because I think there are some nice things to decouple the lora crate from the nucleo board.

@lulf Could you test if this PR breaks your LoRa setup? Marking as draft for the time being.

### DIFF
--- a/embassy-lora/src/stm32wl/mod.rs
+++ b/embassy-lora/src/stm32wl/mod.rs
@@ -4,7 +4,6 @@ use core::task::Poll;
 
 use embassy_hal_common::{into_ref, Peripheral, PeripheralRef};
 use embassy_stm32::dma::NoDma;
-use embassy_stm32::gpio::{AnyPin, Output};
 use embassy_stm32::interrupt::{Interrupt, InterruptExt, SUBGHZ_RADIO};
 use embassy_stm32::subghz::{
     CalibrateImage, CfgIrq, CodingRate, Error, HeaderType, Irq, LoRaBandwidth, LoRaModParams, LoRaPacketParams,
@@ -100,7 +99,7 @@ impl<'d, RS: RadioSwitch> SubGhzRadio<'d, RS> {
     async fn do_tx(&mut self, config: TxConfig, buf: &[u8]) -> Result<u32, RadioError> {
         //trace!("TX Request: {}", config);
         trace!("TX START");
-        self.switch.set_tx_lp();
+        self.switch.set_tx();
         self.configure()?;
 
         self.radio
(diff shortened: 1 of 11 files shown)


lulf:
I fixed/modified the lorawan example with more configuration options to make downlink work in #843 . I also tried your branch with that fix, and downlink works for me.

Those settings could be the reason for downlink not working. Also note the special RX1 delay required for TTN

lulf:
@timokroeger I didn't see you marked this as ready for review until now. Does this mean you got things working as expected?

timokroeger:
Yes, this is ready for review. Rebased to resolve the merge conflicts.
With a different LoRaWAN network provider I could occasionally receive downlink messages. Did not yet have the time to setup my own gateway to do any further testing.

embassy-stm32/src/spi/mod.rs
```diff
@@ -180,6 +180,18 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
         )
     }
 
+    /// Useful for on chip peripherals like SUBGHZ which are hardwired.
+    /// The bus can optionally be exposed externally with `Spi::new()` still.
+    pub fn new_internal(
```
lulf, line 185: Should be marked as pub(crate) IMO since we likely want the HAL to have some interface for those special peripherals?

examples/stm32wl/src/bin/lorawan.rs
```diff
@@ -32,11 +58,12 @@ async fn main(_spawner: embassy::executor::Spawner, p: Peripherals) {
...
-    let radio = SubGhz::new(p.SUBGHZSPI, p.PA5, p.PA7, p.PA6, NoDma, NoDma);
-
+    let radio = SubGhz::new(p.SUBGHZSPI, NoDma, NoDma);
     let irq = interrupt::take!(SUBGHZ_RADIO);
-    static mut RADIO_STATE: SubGhzState<'static> = SubGhzState::new();
-    let radio = unsafe { SubGhzRadio::new(&mut RADIO_STATE, radio, rfs, irq) };
+
+    let mut radio_config = SubGhzRadioConfig::default();
+    radio_config.calibrate_image = CalibrateImage::ISM_863_870;
+    let radio = SubGhzRadio::new(radio, rfs, irq, radio_config).unwrap();
```
lulf, line 66: It's slightly annoying that you have to specify the region both for calibration and for the lorawan stack. I wonder if we can somehow derive this setting from region, but it's good for now I think.

embassy-lora/src/stm32wl/mod.rs
```diff
@@ -198,77 +140,109 @@ impl<'a> StateInner<'a> {
             .set_preamble_len(8)
             .set_header_type(HeaderType::Variable)
             .set_payload_len(0xFF)
-            .set_crc_en(true)
+            .set_crc_en(false)
```
lulf, line 143: Any particular reason for disabling CRC?

lulf (changes requested):
Thanks for these fixes! I've left a few comments. Running the PR on my Nucleo WL55JC board works perfectly!

embassy-lora/src/stm32wl/mod.rs
timokroeger: I tried to match the C driver from the STM32 SDK but also found this table in RP002-1.0.3:
grafik

lulf:
Good news! I got it working, but I had to do some TX tuning. I'll try to pinpoint the changes I needed to do, and maybe we can make that configurable if it's not already.

lulf:
@timokroeger Ok, it was the TX configuration that was not working for my board. I've raised a PR against this branch with stuff I needed https://github.com/timokroeger/embassy/pull/1

It makes the params configurable, but leaves LP tx as default. We can use HP tx as default as well, I dunno which approach is 
better.

It also adopts to the new rust-lorawan radio interface.

lulf:
@timokroeger I've released lorawan-device 0.8.0 which contains the radio changes. If you merge this https://github.com/timokroeger/embassy/pull/1 and then rebase this PR, it should be good to go I think.

timokroeger:
Rebased to include @lulf s changes

lulf:
Looks like just some cargo fmt is needed

lulf (approved):
Thanks for doing this 🚀 

bors r+

PR #845, opened 2022-07-06
Add blocking shared bus for i2c and SPI

kalkyl:
Blocking versions of the shared buses


embassy-embedded-hal/src/shared_bus/blocking/i2c.rs
```diff
@@ -0,0 +1,69 @@
+//! Blocking shared I2C bus
+use core::cell::RefCell;
+use core::fmt::Debug;
+use core::future::Future;
+
+use embedded_hal_1::i2c;
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum I2cBusDeviceError<BUS> {
```
Dirbaio, line 9: The enums can probably be shared across blocking/async

PR #846, opened 2022-07-06
Implement scoped spawn

chemicstry:
An attempt on implementing scoped spawn.

Using regular tasks is inconvenient because they require all references to be static and this is difficult when you construct a lot of stuff in the main function. So far I've been using a huge `select_biased!` block to poll all of the anonymous futures, but this is very inefficient, because all of them share the same waker, which means that on wake up each of the futures has to be polled. This PR makes it possible to spawn anonymous futures that use references to the stack and schedules each independently.

I wasn't sure if it's possible to cancel a spawned task, so dropping an unfinished guard results in a panic.

Getting this to work was kind of tricky and I'm not sure if it is the best approach, so I'm open for suggestions.

### DIFF
--- a/embassy/src/executor/raw/mod.rs
+++ b/embassy/src/executor/raw/mod.rs
@@ -15,12 +15,13 @@ mod waker;
 
 use core::cell::Cell;
 use core::future::Future;
+use core::marker::PhantomData;
 use core::pin::Pin;
 use core::ptr::NonNull;
 use core::task::{Context, Poll};
 use core::{mem, ptr};
 
-use atomic_polyfill::{AtomicU32, Ordering};
+use atomic_polyfill::{AtomicPtr, AtomicU32, Ordering};
 use critical_section::CriticalSection;
 
 use self::run_queue::{RunQueue, RunQueueItem};
@@ -217,18 +218,18 @@ unsafe impl<F: Future + 'static> Sync for TaskStorage<F> {}
 // repr(C) is needed to guarantee that the Task is located at offset 0
 // This makes it safe to cast between TaskHeader and TaskStorage pointers.
 #[repr(C)]
-pub struct ScopedTaskStorage<'a, F: Future + 'a> {
-    header: &'a TaskHeader,
+pub struct ScopedTaskStorage<F> {
+    header: TaskHeader,
     future: UninitCell<F>, // Valid if STATE_SPAWNED
     waker: AtomicWaker,
 }
 
-impl<'a, F: Future + 'a> ScopedTaskStorage<'a, F> {
+impl<F> ScopedTaskStorage<F> {
     /// Create a new TaskStorage, in not-spawned state.
     #[cfg(feature = "nightly")]
-    pub const fn new(header: &'static TaskHeader) -> Self {
+    pub const fn new() -> Self {
         Self {
-            header,
+            header: TaskHeader::new(),
             future: UninitCell::uninit(),
             waker: AtomicWaker::new(),
         }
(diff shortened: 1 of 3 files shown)


Dirbaio:
Interesting! Some general thoughts:

1- This is adding two separate features:
- Allow spawning tasks that borrow data from another.
- Allow tasks returning a result that can be collected by the parent task.

IMO they are mostly orthogonal, for example you might want to collect a result from a non-scoped task. The result feature should be available everywhere, or we should add only scoped tasks for now and defer the result for later.

(The result feature has some tricky edge cases. For example, if the parent task doesn't collect the state, the child task stays in `STATE_RESULT_PENDING` forever, and can't be reused. Especially for non-scoped tasks, this means they all need to have a "JoinHandle" and track more info, like "is the JoinHandle live?" to avoid this issue, which adds nontrivial code size and complexity)

2- Forcing the use of `dyn` is unfortunate, this should be avoidable with TAIT + macros or perhaps a different design that allows good old Rust's type inference to do it.

3- The `TaskStorage` and `ScopedTaskStorage` duplication feels a bit ugly to me, maybe it'd be possible to unify them especially if we hold off on the "task result" feature.

embassy/src/executor/raw/mod.rs
```diff
@@ -207,6 +211,148 @@ impl<F: Future + 'static> TaskStorage<F> {
...
+            .storage
+            .header
+            .state
+            .fetch_and(!STATE_RESULT_PENDING, Ordering::Acquire)
+            & STATE_RESULT_PENDING
+            != 0
+        {
+            // Safety: STATE_RESULT_PENDING ensures that result is init
+            unsafe {
+                (*self.storage.result.get()).assume_init_drop();
```
Dirbaio, line 350: you should first drop, then clear the flag. By clearing the flag you give permission to other threads to overwrite the task storage, which will race with us doing the drop.

embassy/src/executor/raw/mod.rs
```diff
@@ -207,6 +211,148 @@ impl<F: Future + 'static> TaskStorage<F> {
...
+        let this = &*(p.as_ptr() as *const Self);
+
+        let future = Pin::new_unchecked(&mut **this.future.as_mut());
+        let waker = waker::from_task(p);
+        let mut cx = Context::from_waker(&waker);
+        match future.poll(&mut cx) {
+            Poll::Ready(res) => {
+                this.future.drop_in_place();
+                this.header.state.fetch_and(!STATE_SPAWNED, Ordering::AcqRel);
+                (*this.result.get()).write(res);
```
Dirbaio, line 293: there's a tiny window here where `state=0`, where another thread could claim this TaskStorage even if we're not done with it.

embassy/src/executor/raw/mod.rs
chemicstry: My idea was to protect this by borrowing `ScopedTaskStorage` for the duration of the guard, which would prevent any new spawn until guard is dropped and the only write is in the poll method with `Ordering::Release`. But now I see that I needed a `mut` borrow to make it exclusive, my bad.

To make it thread safe without exclusive borrow I think you would need a CAS-like mutex, which is not available on some MCUs. But I see that executor already has some CAS operations, so I guess it's not a problem?

Dirbaio: this shouldn't need CAS, all you need to do is to clear `STATE_RESULT_PENDING` after dropping, not before.

(And ordering should be `Release` I think, to prevent the drop to be moved down past it)

PR #849, opened 2022-07-06
Add EH 1.0 impls for stm32 i2c

kalkyl:
Add missing stm32 impls for embedded-hal 1.0 i2c traits


embassy-stm32/src/i2c/v2.rs
```diff
@@ -854,6 +854,51 @@ mod eh1 {
     impl<'d, T: Instance, TXDMA: TxDma<T>, RXDMA: RxDma<T>> embedded_hal_1::i2c::ErrorType for I2c<'d, T, TXDMA, RXDMA> {
         type Error = Error;
     }
+
+    impl<'d, T: Instance, TXDMA: TxDma<T>, RXDMA: RxDma<T>> embedded_hal_1::i2c::blocking::I2c
```
Dirbaio, line 858: This impl should apply when there's no DMA channels. Blocking doesn't use DMA, we don't want to force 

```rust
impl<'d, T: Instance> embedded_hal_1::i2c::blocking::I2c for for I2c<'d, T, NoDma, NoDma>
```

or the equivalent (because the generics defaults)

```rust
impl<'d, T: Instance> embedded_hal_1::i2c::blocking::I2c for for I2c<'d, T>
```

Dirbaio (approved):
:rocket: :rocket: :rocket: :rocket: 

bors r+

PR #853, opened 2022-07-10
Add embedded_hal_async support for embassy-rp

danbev:
This commit adds support for embedded-hal-async to the Embassy
Raspberry PI crate.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -1,11 +1,20 @@
 use core::convert::Infallible;
+use core::future::Future;
 use core::marker::PhantomData;
+use core::pin::Pin as FuturePin;
+use core::task::{Context, Poll};
 
+use embassy::waitqueue::AtomicWaker;
+use embassy_cortex_m::interrupt::{Interrupt, InterruptExt};
 use embassy_hal_common::{unborrow, unsafe_impl_unborrow};
 
 use crate::pac::common::{Reg, RW};
 use crate::pac::SIO;
-use crate::{pac, peripherals, Unborrow};
+use crate::{interrupt, pac, peripherals, Unborrow};
+
+const PIN_COUNT: usize = 30;
+const NEW_AW: AtomicWaker = AtomicWaker::new();
+static INTERRUPT_WAKERS: [AtomicWaker; PIN_COUNT] = [NEW_AW; PIN_COUNT];
 
 /// Represents a digital input or output level.
 #[derive(Debug, Eq, PartialEq)]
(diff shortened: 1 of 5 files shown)


tests/rp/src/bin/gpio_async.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    // Test wait_for_high
+    {
+        async_pin.set_high();
+        async_pin.set_as_output();
+        delay();
+        assert!(async_pin.is_high(), "async_pin was expected to be high");
+
+        async_pin.set_as_input();
+        async_pin.wait_for_high().await;
```
Dirbaio, line 26: The tests are run automatically in CI, they should work without a human interacting with the board. 

The board in CI has GPIO0 and GPIO1 connected together, this is how the `gpio.rs` tests work (one pino utputs something, the other pin looks that it output the right thing).

Perhaps here the best way to test it is to set GPIO0 low, start a `wait_for_high` on GPIO1, check it doesn't return, then set GPIO0 high, check that it returns.

embassy-rp/src/gpio.rs
```diff
@@ -265,6 +472,47 @@ impl<'d, T: Pin> Flex<'d, T> {
     pub fn toggle(&mut self) {
         unsafe { self.pin.sio_out().value_xor().write_value(self.bit()) }
     }
+
+    pub async fn wait_for_high<'a>(&mut self) {
+        if self.is_set_as_output() {
+            panic!("wait_for_high can't be called when in output mode");
```
Dirbaio, line 478: Don't do these checks, the whole point of Flex is it lets you do anything at the cost of some safety, so it's fine if it behaves weird if you use it in the wrong mode.

embassy-rp/src/gpio.rs
```diff
@@ -520,6 +779,35 @@ mod eh02 {
             Ok(self.toggle())
         }
     }
+
+    use core::convert::Infallible;
+
+    impl<'d, T: Pin> embedded_hal_async::digital::Wait for Flex<'d, T> {
```
Dirbaio, line 785: can you add an impl for `Input` as well?

embassy-rp/src/gpio.rs
danbev: Ah I see, I'll remove them, thanks!

embassy-rp/src/gpio.rs
danbev: I'll add it :+1:

tests/rp/src/bin/gpio_async.rs
danbev: Thanks for the details about how this works!

>Perhaps here the best way to test it is to set GPIO0 low, start a wait_for_high on GPIO1, check it doesn't return, then set GPIO0 high, check that it returns.

I've added a suggestion in https://github.com/embassy-rs/embassy/pull/853/commits/b29865b0698baa21110952b254479a77d233747c. Is something like that what you had in mind when you wrote `check it doesn't return`, or is there perhaps a way to check the future would return Poll::Pending which I'm not aware (or some other way that I'm just not "seeing")?

danbev: @Dirbaio Just a friendly ping to see if you have time to take a look at my comment above? Thanks

Dirbaio: perhaps timing the `wait_for_high` and check it takes the expected time would work:

```rust
output.set_low();
let fut1 = async {
  Timer::after(Duration::from_millis(10)).await;
  output.set_high();
};
let fut2 = async {
  let start = Instant::now();
  input.wait_for_high().await;
  let dur = Instant::now() - start;
  assert!(dur >= Duration::from_millis(10) && dur < Duration::from_millis(11));
};
join(fut1, fut2).await;

danbev: Oh, very nice! Thank you for showing me that!
I've updated this PR with a commit now.

Dirbaio (approved):
Thank you!

bors r+

PR #854, opened 2022-07-10
Implement IWDG timeout calculation

chemicstry:
Allow specifying `IndependentWatchdog` timeout as `Duration` instead of prescaler value.

Since IWDG is clocked from LSI, which differs between families, I standardized HSI/LSI definitions in RCC and used that.

### DIFF
--- a/embassy-stm32/src/rcc/f0.rs
+++ b/embassy-stm32/src/rcc/f0.rs
@@ -3,7 +3,11 @@ use crate::pac::rcc::vals::{Hpre, Pllmul, Pllsrc, Ppre, Sw, Usbsw};
 use crate::pac::{FLASH, RCC};
 use crate::time::Hertz;
 
-const HSI: u32 = 8_000_000;
+/// HSI speed
+pub const HSI_FREQ: Hertz = Hertz(8_000_000);
+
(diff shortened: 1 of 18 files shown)


Dirbaio:
can WDG clock from other clocks (LSE or others)? (if it can, supporting just LSI for now is fine, I'm just curious)

chemicstry:
> can WDG clock from other clocks (LSE or others)? (if it can, supporting just LSI for now is fine, I'm just curious)

Not for any of the chips I know. Some (stm32u5) have a selectable LSI frequency, which IDK what to do for now. Maybe it should be better to implement something similar to `RccPeripheral` for `IWDG` to be able to do `T::frequency()`?

Dirbaio:
> Maybe it should be better to implement something similar to RccPeripheral for IWDG to be able to do T::frequency()?

seems like a very niche feature, I'd say it's OK to not worry about that for now.

Dirbaio (approved):
Nice! Thank you! :) 

bors r+

chemicstry:
Woops, sorry the build failed because I was using stm32-data with https://github.com/embassy-rs/stm32-data/pull/150. Bumped.

PR #855, opened 2022-07-10
PDM microphone support for nrf

pbert519:
PDM microphones have a long startup phase, therefore the driver samples continuously and only switches the target buffer if the user requests sampling.

### DIFF
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -76,6 +76,14 @@ pub mod gpio;
 pub mod gpiote;
 #[cfg(not(any(feature = "_nrf5340", feature = "_nrf9160")))]
 pub mod nvmc;
+#[cfg(any(
+    feature = "nrf52810",
+    feature = "nrf52811",
+    feature = "nrf52833",
+    feature = "nrf52840",
+    feature = "_nrf9160"
+))]
+pub mod pdm;
 pub mod ppi;
 #[cfg(not(any(feature = "nrf52805", feature = "nrf52820", feature = "_nrf5340-net")))]
 pub mod pwm;
--- a/embassy-nrf/src/pdm.rs
+++ b/embassy-nrf/src/pdm.rs
@@ -1,0 +1,242 @@
+//! PDM mirophone interface
+
+use core::marker::PhantomData;
+use core::sync::atomic::{compiler_fence, Ordering};
+use core::task::Poll;
+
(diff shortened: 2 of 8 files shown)


embassy-nrf/src/lib.rs
```diff
@@ -76,6 +76,8 @@ pub mod gpio;
 pub mod gpiote;
 #[cfg(not(any(feature = "_nrf5340", feature = "_nrf9160")))]
 pub mod nvmc;
+#[cfg(feature = "nrf52840")]
```
lulf, line 79: Looking at Nordic infocenter, it looks like PDM should be available on the nRF 52833, 52811, 52810, 5340 and 9160 as well.

embassy-nrf/src/pdm.rs
```diff
@@ -0,0 +1,200 @@
...
+            .ptr
+            .write(|w| unsafe { w.sampleptr().bits(DUMMY_BUFFER.as_ptr() as u32) });
+        r.sample
+            .maxcnt
+            .write(|w| unsafe { w.buffsize().bits(DUMMY_BUFFER.len() as _) });
+
+        r.tasks_start.write(|w| w.tasks_start().set_bit());
+
+        Self { phantom: PhantomData }
+    }
```
lulf, line 100: Maybe this should be part of an explicit "calibrate()" method so that the peripheral isn't _on_ unless user requests it?

lulf (commented):
:rocket: Looks good, added a few comments inline.

embassy-nrf/src/pdm.rs
pbert519: But how would the user stop the "calibration"?
Maybe a start() and stop() function are better.

I would not stop the peripheral after the user samples, because we would need the startup / calibration for every sample we want to capture.

embassy-nrf/src/lib.rs
lulf: It seems only available from the application core, which I think you can guard with the feature "_nrf5340-app" (both secure and non-secure mode).

For the name, I think you should use PDM0 as used by the PAC.

embassy-nrf/src/pdm.rs
lulf: I'm a little worried user is going to forget stop(), but maybe it's good for now. With maybe also a Drop implementation that calls stop to be sure it stops.

Dirbaio (approved):
Thank you!

bors r+

PR #863, opened 2022-07-12
Add more convenience GPIO functions

chemicstry:
This reduces boilerplate code from:
```rs
if config.pin_high {
    pin.set_high()
} else {
    pin.set_low()
}
```
to
```rs
pin.set_level(config.pin_high.into());
```

### DIFF
--- a/embassy-stm32/src/gpio.rs
+++ b/embassy-stm32/src/gpio.rs
@@ -143,10 +143,7 @@ impl<'d, T: Pin> Flex<'d, T> {
 
     #[inline]
     pub fn get_level(&self) -> Level {
-        match self.is_high() {
-            true => Level::High,
-            false => Level::Low,
-        }
+        self.is_high().into()
     }
 
     #[inline]
@@ -164,10 +161,7 @@ impl<'d, T: Pin> Flex<'d, T> {
     /// What level output is set to
     #[inline]
     pub fn get_set_level(&self) -> Level {
-        match self.is_set_high() {
-            true => Level::High,
-            false => Level::Low,
-        }
+        self.is_set_high().into()
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/gpio.rs
```diff
@@ -153,6 +161,15 @@ impl<'d, T: Pin> Flex<'d, T> {
         state == vals::Odr::LOW
     }
 
+    /// What level output is set to
+    #[inline]
+    pub fn get_set_level(&self) -> Level {
```
chemicstry, line 166: I'm not sure about the name 😅

Dirbaio:
- Can you add it to nrf, rp as well? We should try to keep all 3 as close as possible.
- If you add the bool<->Level `From` impls you can simplify the code a lot (no matches).

embassy-stm32/src/gpio.rs
Dirbaio: `get_output_level`?

Dirbaio:
> could there be problems with into() not being #[inline]?

within the same crate everything can get inlined I think?

Dirbaio:
Oh and another thing, sorry :sweat: 

It'd be better if all methods on Input/Output/OutputOpenDrain would just call into the corresponding Flex method, so that all logic is in Flex.

chemicstry:
> Oh and another thing, sorry 😓
> 
> It'd be better if all methods on Input/Output/OutputOpenDrain would just call into the corresponding Flex method, so that all logic is in Flex.

On it 👷‍♂️

chemicstry:
Done, except for `embassy-rp` `OutputOpenDrain`, which changes modes between input/output to emulate open drain

Dirbaio:
LGTM then! :D
Is it ready for merging?

Dirbaio (approved):
merge it is

bors r+

PR #870, opened 2022-07-20
Async F4 flash driver

chemicstry:
When I started writing this driver I was unaware that erasing/writing flash on the same bank as program code (but different sector) causes CPU to stall. This renders the async flash driver somewhat less useful, but if you have a chip with 2x1MB flash memory, it can still be useful for storing configuration data in different bank.

I'm not personally using this yet, so have to hurry to merge this, but opened a PR anyway to maybe discuss usefulness of async flash and general implementation.

There are two things that I'm not sure about:
- Driver hijacks `FLASH` interrupt, just like `EXTI` implementation. Passing interrupt as an argument to `new()` might be better, but then we would have two different flash drivers: async and blocking.
- Async and non-async functions duplicate a lot of code (especially `write`), when the only difference is the delay between operations. Maybe blocking operations could just busy poll async futures? Although that might have some minor overhead.

### DIFF
--- a/embassy-stm32/src/flash/f4.rs
+++ b/embassy-stm32/src/flash/f4.rs
@@ -210,7 +210,7 @@ pub(crate) async unsafe fn write(offset: u32, buf: &[u8]) -> Result<(), Error> {
         w.set_pg(false);
         w.set_eopie(false);
         w.set_errie(false);
-});
+    });
 
     ret
 }
--- a/embassy-stm32/src/flash/f4.rs
+++ b/embassy-stm32/src/flash/f4.rs
@@ -1,11 +1,14 @@
 use core::convert::TryInto;
 use core::ptr::write_volatile;
+use core::task::Poll;
 
 use atomic_polyfill::{fence, Ordering};
+use embassy::waitqueue::AtomicWaker;
+use futures::future::poll_fn;
 
 use super::{ERASE_SIZE, FLASH_BASE, FLASH_SIZE};
 use crate::flash::Error;
-use crate::pac;
+use crate::{interrupt, pac};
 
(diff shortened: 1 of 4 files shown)


lulf:
In case you are unaware, there is also https://github.com/embassy-rs/embassy/blob/master/embassy-embedded-hal/src/adapter.rs#L14 which can wrap embedded_storage async flash as embedded_storage blocking flash.

chemicstry:
> In case you are unaware, there is also https://github.com/embassy-rs/embassy/blob/master/embassy-embedded-hal/src/adapter.rs#L14 which can wrap embedded_storage async flash as embedded_storage blocking flash.

Isn't it the other way around? Wraps blocking flash and exposes async flash interface. I don't think there is much use of it, since CPU will be busy looping in the blocking driver and other async tasks can not proceed.

lulf:
You're right, never mind then! (That's slightly embarrassing because I wrote that adapter code 😆). It's been slightly useful for things like embassy-boot until flash drivers started to support async (which is better, fully agreed).

chemicstry:
> You're right, never mind then! (That's slightly embarrassing because I wrote that adapter code 😆). It's been slightly useful for things like embassy-boot until flash drivers started to support async (which is better, fully agreed).

No worries, no doubt it is useful when async driver is not available (I'm still using old usb-otg 😥).

The problem with flash is that there is a lot of code duplication to support both async and blocking interfaces, hence my suggestion to only have async one and use an adapter for blocking interface. I could continue with this PR and port all families to async, but unfortunatelly I don't have enough dev boards to test everything.

Dirbaio:
Hello,

PR #873, opened 2022-07-21
Fix build of embassy-rp on stable channel

jannic:


Dirbaio (approved):
Really nice, thank you! :) 

bors r+

PR #874, opened 2022-07-22
Add inline attribute to embassy-rp async functions

danbev:
This commit adds the inline attribute to the recently added async gpio
functions. This is to enable cross-crate inlining and to be consistent
with the other functions implemented for Input and Flex.


Dirbaio (approved):
Thank you!

bors r+

PR #876, opened 2022-07-24
Add defmt support to embassy-embedded-hal errors

matoushybl:
`defmt::unwrap!()` should now work with shared buses. I tested it only with I2C as I don't have SPI in the target project.


Dirbaio:
why not use `derive(defmt::Format)`?

matoushybl:
Sorry, I didn't realise the derive macro works with generic arguments. I redid it with the macro.

Dirbaio (approved):
Thanks! :) 

bors r+

PR #878, opened 2022-07-26
Minor simplifications for `Forever`

GrantM11235:


embassy/src/util/forever.rs
```diff
@@ -87,12 +74,8 @@ impl<T> Forever<T> {
...
         }
 
-        unsafe {
-            let p = self.t.get();
-            let p = (&mut *p).as_mut_ptr();
-            p.write(val());
-            &mut *p
-        }
+        let p: &'static mut MaybeUninit<T> = unsafe { &mut *self.t.get() };
+        p.write(val())
```
Dirbaio, line 78: this code was written in that weird way to ensure the T was constructed in-place. 

The compiler would sometimes construct it in the stack and then move it into the Forever, causing a stack overflow if T was big. I found directly passing the T to a ptr write had the highest odds of getting optimized. 

Have you checked if MaybeUninit::write also gets optimized? My gut feeling is it does, but it should be checked because you never know... LGTM if it does.

GrantM11235: The asm looks the same to me https://godbolt.org/z/cGqY1nKzs

Dirbaio:
Nice! Thanks for taking the time to check :) 

bors r+

PR #879, opened 2022-07-26
Improve ADC configuration options

chemicstry:
This smooths out a few rough edges with ADC:
- Make `Resolution::to_max_count()` pub for all ADC versions. Useful if performing conversion manually.
- `VREF` and `VREFINT` were convoluted. `VREF` is an external voltage reference used for all ADC conversions and it is exposed as a separate pin on large chips, while on smaller ones it is connected to `VDDA` internally. `VREFINT` is an internal voltage reference connected to one of the ADC channels and can be used to calculate unknown `VREF` voltage.
- Add a separate `VREF_DEFAULT_MV`, equal to 3.3V, which is the usual supply voltage and should work for most simple cases.
  - For an external voltage reference `set_vref()` can be used to set a known precise voltage.
  - If no voltage reference is present, `VREFINT` calibration can be used to calculate unknown `VREF` voltage. If I understand correctly, this is only implemented for `adc_v3` (L4?), but is not currently exposed (private). If someone is interested, this can be fixed in separate PR.

### DIFF
--- a/embassy-stm32/src/adc/v2.rs
+++ b/embassy-stm32/src/adc/v2.rs
@@ -153,7 +153,7 @@ impl Prescaler {
 
 pub struct Adc<'d, T: Instance> {
     sample_time: SampleTime,
-    vref: u32,
+    vref_mv: u32,
     resolution: Resolution,
     phantom: PhantomData<&'d mut T>,
 }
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/adc/v2.rs
```diff
@@ -150,7 +153,7 @@ impl Prescaler {
 
 pub struct Adc<'d, T: Instance> {
     sample_time: SampleTime,
-    calibrated_vdda: u32,
+    vref: u32,
```
Dirbaio, line 156: if we have no typed struct for mV it'd be good to specify the unit in the field name: `vref_mv`.

Dirbaio (approved):
Thanks! :rocket: 

bors r+

PR #880, opened 2022-07-28
Add UDP socket support

arturkow2000:


Dirbaio:
To allow actually using UDP, this would need an async UdpSocket wrapper on top of the smoltcp one, like [the one we have now for TCP](https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/tcp.rs).

lulf:
The `embassy` crate got split into `embassy-executor` and `embassy-util`. Macros should use `embassy-executor`.

Otherwise LGTM

embassy-net/src/udp.rs
```diff
@@ -0,0 +1,227 @@
...
+    {
+        self.io.write(buf, remote_endpoint.into()).await
+    }
+
+    pub async fn flush(&mut self) -> Result<(), Error> {
+        self.io.flush().await
+    }
+}
+
+impl embedded_io::Error for BindError {
```
Dirbaio, line 217: these aren't needed if UDP doesn't impl the embedded-io traits.

embassy-net/src/udp.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        Self {
+            io: UdpIo {
+                stack: &stack.socket,
+                handle,
+            },
+        }
+    }
+
+    pub fn split(&mut self) -> (UdpReader<'_>, UdpWriter<'_>) {
```
Dirbaio, line 69: If you make send/recv require just `&self` then split is not needed anymore, which simplifies things a lot.

This can't be done with TCP because we want ONE reader and ONE writer, but in UDP having many writers is fine because it's packet-oriented, not stream-oriented.

embassy-net/src/udp.rs
arturkow2000: Removed

embassy-net/src/udp.rs
```diff
@@ -0,0 +1,227 @@
...
+                    s.register_send_waker(cx.waker());
+                    Poll::Pending
+                }
+                Err(udp::SendError::Unaddressable) => Poll::Ready(Err(Error::NoRoute)),
+            })
+        })
+        .await
+    }
+
+    async fn flush(&mut self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
arturkow2000: Removed

embassy-net/src/udp.rs
arturkow2000: Done

embassy-net/src/udp.rs
```diff
@@ -0,0 +1,173 @@
...
+impl Drop for UdpSocket<'_> {
+    fn drop(&mut self) {
+        // safety: not accessed reentrantly.
+        let s = unsafe { &mut *self.io.stack.get() };
+        s.sockets.remove(self.io.handle);
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct UdpIo<'a> {
```
Dirbaio, line 122: TcpIo was just to share code between TcpSocket, TcpReader, TcpWriter. Here it's not needed anymore because there's just 1 struct. Could you fold it all into UdpSocket?

arturkow2000: Done

Dirbaio (approved):
Thank you! 
bors r+

PR #882, opened 2022-07-29
Remove separation of stable vs. nightly implementations in executor/raw

jannic:
Since rust 1.61, the required const features are availabe on stable rust


Dirbaio (approved):
ooh this is really nice, thank you! :) 

bors r+

PR #888, opened 2022-08-01
Extend QSPI config with freq, delay, spi mode, and address mode

### LINKED ISSUE #886: Implementing nRF52 QSPI frequency selection
Hello there!
I would love to implement the feature of being able to select the frequency used for QSPI on the nRF52 ([see this comment](https://github.com/embassy-rs/embassy/blob/bd6bab1625d90a2dc2a4b57b40dcfaa9516bf791/embassy-nrf/src/qspi.rs#L20)).

Couple of notes up-front:
- Should we use the same re-exported `Frequency` enum present in the [SPIM module](https://github.com/embassy-rs/embassy/blob/bd6bab1625d90a2dc2a4b57b40dcfaa9516bf791/embassy-nrf/src/spim.rs#L10), re-export it again, or define our own?
- Are there limits on how low the EasyDMA controller can go? Default seems to be 32 MHz and the previously mentioned enum goes down to 125k.
- Implementation sketch
    - Add frequency parameter to initializer
    - Calculate required frequency divider based on that according to [`SCKFREQ_W`](https://docs.rs/nrf52840-pac/latest/nrf52840_pac/qspi/ifconfig1/struct.SCKFREQ_W.html)'s documentation
    - Set the value [here](https://github.com/embassy-rs/embassy/blob/bd6bab1625d90a2dc2a4b57b40dcfaa9516bf791/embassy-nrf/src/qspi.rs#L125)
    - Update the example(s)

TilBlechschmidt:
Fixes #886 among some of the TODOs which have been flying around in the source file :)

I'll make one more commit to remove some old commented out code from the file and take some time this evening to verify that it works.

### DIFF
--- a/embassy-nrf/src/qspi.rs
+++ b/embassy-nrf/src/qspi.rs
@@ -309,13 +309,10 @@ impl<'d, T: Instance, const FLASH_SIZE: usize> Qspi<'d, T, FLASH_SIZE> {
     }
 
     fn start_write(&mut self, address: usize, data: &[u8]) -> Result<(), Error> {
-        //info!("start_write ptr {}", data.as_ptr() as u32);
         assert_eq!(data.as_ptr() as u32 % 4, 0);
-        //info!("start_write OK ptr");
         assert_eq!(data.len() as u32 % 4, 0);
-        //info!("start_write OK len");
         assert_eq!(address as u32 % 4, 0);
-        //info!("start_write OK addr");
+
         if address > FLASH_SIZE {
             return Err(Error::OutOfBounds);
         }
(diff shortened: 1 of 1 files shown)


TilBlechschmidt:
Examples have not been updated since the changes are covered by `Config::default()` and the `#[non_exhaustive]` attribute — it is entirely optional to overwrite the previously applicable defaults

TilBlechschmidt:
While I do not have the tooling to confirm the exact values / frequencies, the API change appears to work (judging by transfer speeds). I'd thus consider this ready for review 🙂

PR #891, opened 2022-08-03
Async Pipe

Dirbaio:


lulf (changes requested):
LGTM, but shouldn't the original ring_buffer be removed from embassy-hal-common/src/ring_buffer.rs?

Dirbaio:
it's still used by BufferedUart (nrf, stm32), and it's slightly different (it borrows the buffer instead of owning it with const generics).

(also note the new ringbuffer is not public API)

lulf (approved):
Looks like a great addition in my view!

PR #895, opened 2022-08-08
Implement embedded-nal-async traits for embassy-net

lulf:


embassy-net/src/tcp.rs
```diff
@@ -328,3 +328,170 @@ impl<'d> embedded_io::asynch::Write for TcpWriter<'d> {
...
+    use core::ptr::NonNull;
+
+    use atomic_polyfill::{AtomicBool, Ordering};
+    use embedded_nal_async::IpAddr;
+
+    use super::*;
+
+    pub struct TcpClient<'d, D: Device, const N: usize, const TX_SZ: usize = 1024, const RX_SZ: usize = 1024> {
+        stack: &'d Stack<D>,
+        tx: &'d BufferPool<TX_SZ, N>,
```
Dirbaio, line 344: you can use a single pool for both tx+rx buffers: `Pool<([u8; TX_SZ], [u8; RX_SZ}), N>`

embassy-net/src/tcp.rs
```diff
@@ -328,3 +328,170 @@ impl<'d> embedded_io::asynch::Write for TcpWriter<'d> {
...
+        rx: &'d BufferPool<RX_SZ, N>,
+    }
+
+    impl<'d, D: Device, const N: usize, const TX_SZ: usize, const RX_SZ: usize> TcpClient<'d, D, N, TX_SZ, RX_SZ> {
+        pub fn new(stack: &'d Stack<D>, tx: &'d BufferPool<TX_SZ, N>, rx: &'d BufferPool<RX_SZ, N>) -> Self {
```
Dirbaio, line 349: the pool is an implementation detail, perhaps to avoid leaking it, wrap it in a `TcpClientState<const N: usize, const TX_SZ: usize, const RX_SZ: usize>`?

PR #896, opened 2022-08-09
Implement I2C pullup configuration

chemicstry:
I wasn't sure if I should put frequency into config struct, so left it separate as in SPI periph.

Also added Copy derives to gpio types, not sure why they weren't?

### DIFF
--- a/embassy-stm32/src/i2c/v1.rs
+++ b/embassy-stm32/src/i2c/v1.rs
@@ -4,10 +4,23 @@ use embassy_embedded_hal::SetConfig;
 use embassy_hal_common::into_ref;
 
 use crate::gpio::sealed::AFType;
+use crate::gpio::Pull;
 use crate::i2c::{Error, Instance, SclPin, SdaPin};
 use crate::pac::i2c;
 use crate::time::Hertz;
 use crate::Peripheral;
+
+#[non_exhaustive]
+#[derive(Copy, Clone)]
+pub struct Config {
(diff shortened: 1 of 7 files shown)


embassy-stm32/src/i2c/v1.rs
```diff
@@ -4,11 +4,24 @@ use embassy_embedded_hal::SetConfig;
...
+use crate::gpio::Pull;
 use crate::i2c::{Error, Instance, SclPin, SdaPin};
 use crate::pac::i2c;
 use crate::time::Hertz;
 use crate::Peripheral;
 
+#[non_exhaustive]
+#[derive(Copy, Clone)]
+pub struct Config {
+    pullup_enable: bool,
```
lulf, line 16: Looking at the nRF HAL, maybe this should be 

```
pub struct Config {
    pub sda_pullup: bool,
    pub scl_pullup: bool,
}
```

this would allow setting pullup individually on the pins + keep the naming the same.

lulf (changes requested):

embassy-stm32/src/i2c/v1.rs
chemicstry: Done. I also forgot to make config field `pub` 🙄

lulf:
Thanks! :rocket: :rocket: :rocket:

bors r+

lulf (approved):

PR #902, opened 2022-08-12
Declare std feature in embassy-util Cargo.toml

arturkow2000:


lulf:
Closed in favor of #931 now that crates was split.

PR #903, opened 2022-08-15
Update signal.rs

bobmcwhirter:
Allow `poll_wait` to be public so higher-order futures around Signal can be built.


Dirbaio (approved):
Thanks!

bors r+

PR #905, opened 2022-08-16
Rewrite documentation using correct module names

lulf:
* Remove traits section now that we have embedded-hal-async and refer to
  it.
* Explanation that embassy is multiple things.
* Bootloader description image


Dirbaio (approved):
Thanks!!! You're doing great work on the docs, I should join too :sweat_smile: :see_no_evil:  :rocket:

PR #910, opened 2022-08-18
(embassy-rp): Implement remaining logic for blocking UART

MathiasKoch:
async read/write is still `todo!()`, awaiting DMA implementation.

### DIFF
--- a/embassy-rp/src/uart.rs
+++ b/embassy-rp/src/uart.rs
@@ -151,7 +151,7 @@ impl<'d, T: Instance> UartRx<'d, T> {
 }
 
 impl<'d, T: Instance> Uart<'d, T> {
-    /// Create a new UARTE without hardware flow control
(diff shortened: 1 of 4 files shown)


embassy-rp/src/uart.rs
```diff
@@ -1,42 +1,199 @@
...
+                    }
+                };
+            }
+        }
+        Ok(())
+    }
 }
 
 impl<'d, T: Instance> Uart<'d, T> {
+    /// Create a new UARTE without hardware flow control
```
9names, line 154: UARTE is an NRF thing, is this a typo or a copy-paste mistake?

MathiasKoch: Ahh, yeah.. good catch :)

lulf (approved):
Looks good! :rocket: 

bors r+

PR #913, opened 2022-08-18
(embassy-rp): Add DMA implementation

MathiasKoch:
This PR adds everything necessary to do peripheral to memory DMA & memory to memory DMA operations.

It also adds async UART read & write, powered by DMA


embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,224 @@
...
 
-pub struct Dma<T: Channel> {
-    _inner: T,
+#[interrupt]
+unsafe fn DMA_IRQ_0() {
+    let ints0 = pac::DMA.ints0().read().ints0();
+
+    critical_section::with(|_| {
+        for channel in 0..CHANNEL_COUNT {
+            if ints0 & (1 << channel) == 1 {
```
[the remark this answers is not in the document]
MathiasKoch: Good catch :+1:

MathiasKoch:
@Dirbaio Could you give your 2 cents on this implementation? and the 3 questions in the PR body?

embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,224 @@
...
-use crate::{pac, peripherals};
+use crate::{interrupt, pac, peripherals};
 
-pub struct Dma<T: Channel> {
-    _inner: T,
+#[interrupt]
+unsafe fn DMA_IRQ_0() {
+    let ints0 = pac::DMA.ints0().read().ints0();
+
+    critical_section::with(|_| {
```
Dirbaio, line 18: This CS is not needed (AtomicWaker is already atomic)

embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,224 @@
...
+        for channel in 0..CHANNEL_COUNT {
+            if ints0 & (1 << channel) == (1 << channel) {
+                CHANNEL_WAKERS[channel].wake();
+            }
+        }
+        pac::DMA.ints0().write(|w| w.set_ints0(ints0));
+    });
+}
+
+pub fn read<'a, C: Channel, W: Word>(ch: impl Peripheral<P = C> + 'a, from: *const W, to: &mut [W]) -> Transfer<'a, C> {
```
Dirbaio, line 28: these should be unsafe because they dereference arbitrary raw pointers.

embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,224 @@
...
+    channel: PeripheralRef<'a, C>,
+}
+
+impl<'a, C: Channel> Transfer<'a, C> {
+    pub(crate) fn new(channel: impl Peripheral<P = C> + 'a) -> Self {
+        into_ref!(channel);
 
         unsafe {
-            let p = inner.regs();
+            let irq = interrupt::DMA_IRQ_0::steal();
```
Dirbaio, line 95: i'd do this only once at boot, in `embassy_rp::init`. (and yes, the steal is fine)

embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,224 @@
...
+            irq.disable();
+            irq.set_priority(interrupt::Priority::P6);
 
-            p.read_addr().write_value(from.as_ptr() as u32);
-            p.write_addr().write_value(to.as_mut_ptr() as u32);
-            p.trans_count().write_value(from.len() as u32);
+            pac::DMA.inte0().write(|w| w.set_inte0(1 << channel.number()));
```
Dirbaio, line 99: this will disable irqs for all the other channels. Either do a `.modify` (within a CS to avoid races) or use the fancy "atomic set bit" register alias https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/atomic_register_access.rs#L21

EDIT: Actually I think you can simply enable all channel's irqs on boot.

embassy-rp/src/uart.rs
```diff
@@ -116,13 +118,29 @@ impl<'d, T: Instance> UartTx<'d, T> {
     }
 }
 
-impl<'d, T: Instance> UartRx<'d, T> {
-    fn new() -> Self {
-        Self { phantom: PhantomData }
+impl<'d, T: Instance> UartTx<'d, T, Async> {
+    pub async fn write(&mut self, buffer: &[u8]) -> Result<(), Error> {
+        if let Some(ch) = &mut self.tx_dma {
```
Dirbaio, line 123: i'd use `.unwrap()`. The `M` generic should enforce this is never None, so if it is we'd like to find out instead of having this silently do nothing.

Dirbaio (changes requested):
> interrupt::set_handler vs #[interrupt] usage?

`set_handler` only works if the user has done `take!` (that's what registers the actual handler, with `#[interrupt`]`. So for irqs registered by the HAL unconditionally, use `#[interrupt]`.

> let irq = interrupt::DMA_IRQ_0::steal(); feels wrong?

it's fine for HAL-registered irqs with `#[interrupt]`, since the user can't `take!` them because they'll get a "duplicate symbol" error for the irq handler.

> Error handling on the DMA part?

in stm32 I made bus errors just panic, because the only way to get those is if the user passed an invalid pointer. There's no sane reason you'd want to handle that to recover and continue. I think same logic applies to rp2040.

MathiasKoch:
> in stm32 I made bus errors just panic, because the only way to get those is if the user passed an invalid pointer. There's no sane reason you'd want to handle that to recover and continue. I think same logic applies to rp2040.

Where would this have to be checked?

embassy-rp/src/dma.rs
```diff
@@ -1,82 +1,234 @@
...
+
+        Self { channel }
+    }
+}
+
+impl<'a, C: Channel> Drop for Transfer<'a, C> {
+    fn drop(&mut self) {
+        let p = self.channel.regs();
+        unsafe {
+            p.ctrl_trig().write(|w| w.set_en(false));
```
Dirbaio, line 122: you have to use CHAN_ABORT to stop a channel, setting EN=false isn't enough, see datasheet `2.5.5.3. Channel Abort`

Dirbaio:
> Where would this have to be checked?

Not sure. Presumably a bus error stops the channel which triggers an irq, so you can check the `**_ERROR` bits in the irq?

embassy-rp/src/dma.rs
MathiasKoch: Nice catch!

MathiasKoch:
How about the IRQ priority? Is P3 alright?

Dirbaio:
Yeah. in nrf it's configurable, but we can do that later.

Have you tested whether a DMA error triggers the IRQ?

MathiasKoch:
Not yet, i am working from home today and didn't bring any hardware with me.

MathiasKoch:
But from the datasheet:

> Logical OR of the READ_ERROR and WRITE_ERROR flags.
The channel halts when it encounters any bus error, and
always raises its channel IRQ flag.

Dirbaio:
> Do we want to know which caused the error in the panic message?

These errors will be very rare so I don't think it makes much of a difference. Up to you :)

MathiasKoch:
Fixed!

I think this should be ready to go. It is still rather untested, will not be able to test until tomorrow though.

MathiasKoch:
Cool. Yeah, I will add that.. 

I also have a PR for `bufferedUart` almost ready. Just testing they actually work.

Dirbaio (approved):
Nice work, this is awesome! Thank you :) 

bors r+

PR #914, opened 2022-08-19
(embassy-rp): Add I2C master implementation

MathiasKoch:
This PR adds both blocking and DMA based async implementations of I2C master.

Both E-H 0.2 & E-H 1.0 abstractions are implemented as well.

### Questions & concerns:
- Do we need an I2C interrupt handler (for transfer done, abort & error handling?) (async only)
- Do we need to add some automatic attempt at unblocking an I2C bus in case of failures (see ref: https://github.com/fivdi/pico-i2c-dma/blob/7ebfd553f3ce5b5b210d53102b0ecca158172633/src/i2c_dma.c#L116-L142)
- Should I add `vectored_{read, write}` implementations?

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -1,8 +1,13 @@
 use core::marker::PhantomData;
 
-use embassy_hal_common::into_ref;
+use atomic_polyfill::Ordering;
+use embassy_cortex_m::interrupt::InterruptExt;
+use embassy_hal_common::{into_ref, PeripheralRef};
 use pac::i2c;
 
+use crate::dma::{AnyChannel, Channel};
+use crate::gpio::sealed::Pin;
+use crate::gpio::AnyPin;
 use crate::{pac, peripherals, Peripheral};
 
 /// I2C error abort reason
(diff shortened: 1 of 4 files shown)


embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+
+        Self { phantom: PhantomData }
+    }
+}
+
+impl<'d, T: Instance, M: Mode> I2c<'d, T, M> {
+    /// Number of bytes currently in the RX FIFO
+    #[inline]
+    pub fn rx_fifo_used(&self) -> u8 {
```
Dirbaio, line 150: what's the use case for these? they're very low level, and I think they'll always read as "fifo empty" because the high-level read/write APIs wait until the transfer is done?

embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,733 @@
...
+        }
+
+        Ok(())
+    }
+
+    async fn read_internal(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+        let len = buffer.len();
+        self.read_blocking_internal(&mut buffer[..1], true, len == 1)?;
+
+        if len > 2 {
```
Jacobious52, line 147: I think this needs to be greater or equal to 2
```suggestion
        if len >= 2 {
```

I was trying to read 2 bytes for a 16 button keypad and was stumped why only half of it would work 🤣 This fixed it

embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,733 @@
...
+                    ch,
+                    T::regs().ic_data_cmd().ptr() as *const _,
+                    &mut buffer[1..len - 1],
+                    T::RX_DREQ,
+                )
+            };
+            transfer.await;
+        }
+
+        if len > 2 {
```
Jacobious52, line 171: Same here

```suggestion
        if len >= 2 {
```

embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,733 @@
...
+            self.read_internal(buffer).await
+        }
+    }
+
+    pub async fn write_read(&mut self, address: u8, bytes: &[u8], buffer: &mut [u8]) -> Result<(), Error> {
+        Self::setup(address.into())?;
+        if bytes.is_empty() {
+            self.write_blocking_internal(bytes, false)?;
+        } else {
+            self.write_internal(bytes, false).await?;
```
Jacobious52, line 205: send_stop to `true` here to match the standalone write behavior in `write` -> `self.write_internal(bytes, true).await`

```suggestion
            self.write_internal(bytes, true).await?;
```

Jacobious52 (commented):
Just noticed a few things using this PR to get some simple I2C working on the Pico.

Awesome driver! Hope this can get finished and merged soon 🚀

embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    unsafe fn on_interrupt(_: *mut ()) {
+        let status = T::regs().ic_intr_stat().read();
+
+        // FIXME:
+        if status.tcr() || status.tc() {
+            let state = T::state();
+            state.chunks_transferred.fetch_add(1, Ordering::Relaxed);
+            state.waker.wake();
+        }
```
jsgf, line 104: Aside from this being references to the STM registers, it's not clear to me that this ISR is needed at all. The cortex_m executor just blocks waiting for any interrupt, and then polls all tasks when any interrupt happens (not the most efficient, but very simple).

So it seems to me that you could just have the task directly poll the register states when it needs to block on a hardware transition.

embassy-rp/src/i2c.rs
```diff
@@ -0,0 +1,733 @@
...
+        Self { frequency: 100_000 }
+    }
+}
+
+const FIFO_SIZE: u8 = 16;
+
+pub struct I2c<'d, T: Instance, M: Mode> {
+    tx_dma: Option<PeripheralRef<'d, AnyChannel>>,
+    rx_dma: Option<PeripheralRef<'d, AnyChannel>>,
+    dma_buf: [u16; 256],
```
jsgf, line 59: I don't think DMA is winning much here, if we need to statically allocate this buffer and copy the payload data to it. Given that i2c is pretty slow, it wouldn't be the end of the world to take an interrupt for each byte, at least for low speed devices.

DMA might be useful for demanding applications, but it seems to me that the default impl would be better simple and low overhead. I imagine we'd want a standard i2c implemented with pio as well (and maybe pio+dma) so in principle we'd want to support multiple impls anyway.

MathiasKoch: I think I would agree with this. 

Thoughts @Dirbaio

Dirbaio: It's very unfortunate the i2c design doesn't allow DMAing directly from the buffer, yes... Whether that makes using DMA not worth it.. I dunno! I'm OK with whatever you guys decide.

Re pio-i2c, I'd imagine that'd be a different struct.

MathiasKoch:
Given that i have no hardware to test async I2C with, I am inclined to split the blocking and async parts into two PR's here, so we can merge the blocking part & continue working on the async part?

MathiasKoch:
Moved the async part into #978.

This PR with just the blocking implementation should be good to go now :+1:

Dirbaio (approved):
I'd say "write a HIL test" but there's no i2c device in the HIL rig to talk to... Something we'll have to do someday, same for stm32. I trust you tested it on real HW :) 

Thanks for your great work! :rocket: 

bors r+

Dirbaio:
bah

Dirbaio:
rebased

bors r+

PR #915, opened 2022-08-19
Implement support for LPUART

FrozenDroid:
Turns out zero commits is not quite enough to get it implemented (#800), so let's try it again ;)


FrozenDroid:
Once https://github.com/embassy-rs/stm32-data/pull/155 has been merged, and I've updated the stm32-data submodule ref, then this is ready to be merged!

PR #918, opened 2022-08-22
Doc warnings

lulf:
Fixing a few doc warnings in embassy-executor, embassy-time, embassy-usb and embassy-nrf. There are more left, but I need a break :D


Dirbaio (approved):
Thank you!

How frustrating that these errors (invalid doc links) aren't caught in CI, maybe we should do something about it (and make CI even slower :sob: )

bors r+

PR #923, opened 2022-08-23
Fix warnings after crate split

lulf:
Attempting to add doc modules that don't have any warnings to CI for at least 1 target + feature combination. Hopefully doesn't add to much CI time but catches bugs.


Dirbaio:
cargo-batch only supports `build` commands, I wonder why it doesn't error :thinking:

lulf:
Oh :D Then maybe it's better to leave it out? Or maybe a separate job?

Dirbaio:
Thanks! :D

PR #924, opened 2022-08-23
Ensure interrupt::take works without embassy-executor

lulf:
Add "rtos-trace-interrupt" feature flag on embassy-macros and enable it
for embassy-executor, to ensure that the interrupt::take! macro can be
used without depending on embassy-executor.


Dirbaio (approved):
Thanks! 

bors r+

Issue #919, opened 2022-08-22
WASM example cannot import `env`


sjmackenzie:
Steps to repro:

* checkout 1b9599025868d3a5d0d8e773593b05df8b2fecf2
* run `[/path/to/embassy/examples/wasm]$ wasm-pack build --target web`
* run `[/path/to/embassy/examples/wasm]$ simple-http-server`
* navigate to http://0.0.0.0:8000 with your browser
* click on `index.html`

Expected behaviour:
* See log entries printed via an html list

Actual behaviour:
* Blank white screen with the word "log" printed at the top but nothing else. 
* This error is printed in the console: `Uncaught TypeError: Error resolving module specifier “env”. Relative module specifiers must start with “./”, “../” or “/”.`

lulf:
This is related to critical-section no longer being 'importable'. I'll dig and see if I can fix it.

lulf:
Fixed in #925 and works for me now, please reopen if you still face some issues with it.

sjmackenzie:
The `actual behaviour` (as above) persists in 529535194d4b5d58b31fd6a7541176105e3c63f7

Changes to steps: I updated my dependencies as per https://github.com/embassy-rs/embassy/commit/bd4ae2e9526cb66c15b7335cfbc803106637c7e1 (and I'm using `wasm-pack --version 0.9.1`)

Perchance did you clear your artefacts before testing?

```

lulf:
@sjmackenzie 

What

lulf:
@sjmackenzie I started from a clean repo (so no existing artifacts), and that works for me now. Are you modifying the example with some additional dependencies? Basically this error is "unresolved symbol" error, so if a dependency is not "wasm-ready", you can get this error.

Could you share your .wasm file that is generated in `pkg/embassy_wasm_example_bg.wasm`?

sjmackenzie:
Oh, I'm being an idiot, a browser hard refresh solved the issue. It's working!

Thank you for you patch and your patience.

PR #925, opened 2022-08-23
Enable 'std' feature on critical-section for WASM

lulf:
This fixes the WASM support which was failing due to missing
critical-section implementation. This also upgrades the bindgen
dependency and ensures that tooling works.

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -47,5 +47,5 @@ cfg-if = "1.0.0"
 static_cell = "1.0"
 
 # WASM dependencies
-wasm-bindgen = { version = "0.2.76", features = ["nightly"], optional = true }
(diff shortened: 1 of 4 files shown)


lulf:
this fixes #919

Dirbaio (approved):
Great! :rocket: 

bors r+

9SMTM6:
@lulf this seems to have been reverted. The wasm example adds critical-section with that feature without any comment. That cost me a good half day.

Do you know why that feature is no longer enabled for "arch-wasm"? Either the documentation needs to be improved, or the feature flag should again enable that dependency flag.

9SMTM6:
I see now what happened. This wasn't actually enabled in the project, only in the example.

I'll make a PR that enables this blindly if arch-wasm is enabled, as that way it'll work without having to specify this in cargo.toml. Thats how arch-std does it too.

PR #926, opened 2022-08-23
nrf: PPI fixes

Dirbaio:
- Feature-gate DPPI-only stuff that should be feature-gated
- Fix unsoundness due to Task/Event ptr being public: this meant you could construct a Task/Event with an arbitrary pointer and then safely call `Ppi::new_*` which would blindly write to it. Now you have to go through `Task/Event::new_unchecked` which is unsafe, with the right safety contract.


huntc (approved):
Thanks for doing this!

PR #928, opened 2022-08-26
Ensure that the sampling is stopped

huntc:
Ensures that nRF saadc sampling is stopped and is awaited prior to exiting the two sampling methods. Not doing so causes a potential power drain and the potential for dropped buffer writes when having finished continuous sampling.

### DIFF
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -224,6 +224,8 @@ impl<'d, const N: usize> Saadc<'d, N> {
     }
 
     /// One shot sampling. The buffer must be the same size as the number of channels configured.
+    /// The sampling is stopped prior to returning in order to reduce power consumption (power
+    /// consumption remains higher if sampling is not stopped explicitly).
     pub async fn sample(&mut self, buf: &mut [i16; N]) {
         let r = Self::regs();
 
--- a/embassy-nrf/src/saadc.rs
+++ b/embassy-nrf/src/saadc.rs
@@ -3,6 +3,7 @@
 use core::sync::atomic::{compiler_fence, Ordering};
 use core::task::Poll;
 
+use embassy_hal_common::drop::OnDrop;
 use embassy_hal_common::{impl_peripheral, into_ref, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
 use futures::future::poll_fn;
@@ -225,8 +226,12 @@ impl<'d, const N: usize> Saadc<'d, N> {
 
     /// One shot sampling. The buffer must be the same size as the number of channels configured.
     /// The sampling is stopped prior to returning in order to reduce power consumption (power
-    /// consumption remains higher if sampling is not stopped explicitly).
+    /// consumption remains higher if sampling is not stopped explicitly). Cancellation will
+    /// also cause the sampling to be stopped.
     pub async fn sample(&mut self, buf: &mut [i16; N]) {
+        // In case the future is dropped, stop the task and wait for it to end.
+        let on_drop = OnDrop::new(Self::stop_sampling_immediately);
+
         let r = Self::regs();
 
         // Set up the DMA
@@ -259,6 +264,7 @@ impl<'d, const N: usize> Saadc<'d, N> {
         })
         .await;
 
+        on_drop.defuse();
         Self::stop_sampling().await;
     }
 
@@ -279,6 +285,12 @@ impl<'d, const N: usize> Saadc<'d, N> {
     /// taken to acquire the samples into a single buffer. You should measure the
     /// time taken by the callback and set the sample buffer size accordingly.
     /// Exceeding this time can lead to samples becoming dropped.
+    ///
+    /// The sampling is stopped prior to returning in order to reduce power consumption (power
+    /// consumption remains higher if sampling is not stopped explicitly), and to
+    /// free the buffers from being used by the peripheral. Cancellation will
+    /// also cause the sampling to be stopped.
+
     pub async fn run_task_sampler<S, T: TimerInstance, const N0: usize>(
         &mut self,
         timer: &mut T,
@@ -330,6 +342,9 @@ impl<'d, const N: usize> Saadc<'d, N> {
         I: FnMut(),
         S: FnMut(&[[i16; N]]) -> SamplerState,
     {
+        // In case the future is dropped, stop the task and wait for it to end.
+        let on_drop = OnDrop::new(Self::stop_sampling_immediately);
+
         let r = Self::regs();
 
         // Establish mode and sample rate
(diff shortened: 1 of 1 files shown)


peter9477 (commented):
I'm not sure this is the best approach yet.  I've been continuing to explore power consumption and ADC timing (using `DWT::cycle_count()`).  It appears that the first reading (or calibration) following starting the ADC may take an additional ~25us, and it's possible this would be undesirable for some users.  If it is caused by needing to do a START following either ENABLE or STOP, then doing a stop after every sample may not be preferable.  On the other hand, if the increased power consumption is seen up until the STOP, as opposed to the STOP+DISABLE, then someone who only infrequently needs to do a conversion would be suffering much higher power consumption than if they did a stop after each sample as this PR does.  

I don't have an answer quite yet, but I would note that the primary reason to add the STOP initially was because without it the drop() and DISABLE would actually not decrease power consumption again, as though the peripheral is not truly disabled merely by clearing the ENABLE bit. (For latecomers, this is not documented in the datasheet, but has been observed with a current monitoring device, and Nordic's own SDK is careful to do a STOP just before any disabling which suggests they are aware of this behaviour.)

huntc:
How about continuous sampling though? Isn’t there a risk here that sampling will continue post returning from it’s method?

peter9477:
A, silly me.  Correction to the above: upon reviewing my test code and notes, I see that the supposed extra delay following a fresh enable+start was actually from other activity occurring in parallel, and a false alarm.  So there's no extra delay (that I can see), and the issues is I think mainly whether we should be doing the STOP following every sample, or only just before the disable.  I'll do some more power measurements to compare whether consumption drops because of the STOP or only once one does a disable *following* a STOP.

huntc:
> A, silly me. Correction to the above: upon reviewing my test code and notes, I see that the supposed extra delay following a fresh enable+start was actually from other activity occurring in parallel, and a false alarm. So there's no extra delay (that I can see), and the issues is I think mainly whether we should be doing the STOP following every sample, or only just before the disable. I'll do some more power measurements to compare whether consumption drops because of the STOP or only once one does a disable _following_ a STOP.

Again DMA writing into buffers post exiting the sampling, be it single or continuous, is my main concern. I might take another look to see whether I can prove any UB.

huntc:
> So there's no extra delay (that I can see), and the issues is I think mainly whether we should be doing the STOP following every sample, or only just before the disable. 

Thinking on this more, I can’t see a downside on explicitly stopping. We aren’t blocking here given the async wait for the stop, so others tasks can proceed. And then, if the delay for stopping causes some jitter, continuous sampling offers an alternative.

peter9477:
Okay, further power results: 

1. After enabling and running a sample (which in my rev doesn't yet do STOP), power consumption increases by 750uA.  If I go idle for 10s, that consumption persists the whole time.
2. After then doing a STOP, consumption drops back to the previous level, 750uA lower, showing that the increase comes merely from starting the ADC and not directly from actively running conversions.  (The datasheet says consumption during acquisition(?) and conversion is more like 1.24mA but I'm sampling power too infrequently to catch that.)
3. Following the DISABLE, power doesn't appear to change noticeably.  It might be on the order of a few microamps at most, from what I can see, if anything.  This may be a peripheral where being merely being enabled, but not active (i.e. STOPPED) consumes basically no extra power.

That suggests that the stop should be done between samples for users who want to reduce power consumption, but I think it might be worth adding documentation noting this is being done as it will cause the return of the sample value to be delayed.  It might be reasonable to provide two options, like `sample_lowpower()` and `sample()`, with a note that the one of them will reduce power but at the cost of a few microseconds (?) increased latency for the reading as a result of it waiting for STOPPED.  

I may be overthinking this for typical use cases, but I have two conflicting ones.  I need to read a battery level infrequently, so a single sample() followed by an immediate stop, but since I'll also want to reconfigure the Saadc I'll want to drop() it anyway.  For the other case, I'll be enabling then running a tight loop reading samples (one at a time) repeatedly, and processing each one in real time (cannot buffer and post-process in this case) so I would not want it to be doing a wasteful STOP between each sample.  I'd use `sample_lowpower()` (or `sample_then_stop()`?) for the one, and would want something like the existing minimal `sample()` for the other.

huntc:
On reading the datasheet again, it states:

> EasyDMA is finished with accessing RAM when events END or STOPPED are generated. The register [RESULT.AMOUNT](https://infocenter.nordicsemi.com/topic/ps_nrf52840/saadc.html?cp=4_0_0_5_22#register.RESULT.AMOUNT) can then be read, to see how many results have been transferred to the result buffer in RAM since the START task was triggered.

So, with one-shot sampling, given that it cannot return prior to receiving the END event, we're ok. DMA is complete and no more bytes should be written to the buffer passed in. Given power consumption though, I think that we should keep the STOP in there. I've pushed a commit with the additional doc for awareness as suggested.

With continuous sampling and given double buffering, it is possible to exit while an END hasn't been received for the "active" buffer. Thus, I think it is best that we explicitly stop from a DMA perspective and avoid UB.

I don't imagine that issuing a STOP and waiting for it to end is going to be costly. I also believe that continuous sampling with a handoff to another task, perhaps via a signal or a channel, is going to be a low latency alternative. Profiling the cost of waiting for a STOP to complete vs continous with signalling would be interesting.

If we find that waiting for a STOP is costly then we can always add an additional method as suggested by @peter9477. For now though, I would prefer that we keep the methods to just the one without this evidence.

huntc:
I've also added an additional commit to have future cancellation cause the stopping of the sampling task.

embassy-nrf/src/saadc.rs
```diff
@@ -403,6 +427,56 @@ impl<'d, const N: usize> Saadc<'d, N> {
...
+
+        r.events_stopped.reset();
+        r.tasks_stop.write(|w| unsafe { w.bits(1) });
+
+        while r.events_stopped.read().bits() == 0 {}
+        r.events_stopped.reset();
+    }
+
+    // Stop sampling and wait for it to stop in a non-blocking fashino
+    async fn stop_sampling() {
```
Dirbaio, line 449: how long does stopping take? if it's fast (<100 clock cycles) then it's not worth doing it async because an asyn context switch is longer than that. Doing it blocking would be faster and smaller code.

peter9477: I've measured it at well under 45 cycles, probably 22, as measured by DWT::cycle_count() and comparing a code path that stops and disables but doesn't wait for EVENTS_STOPPED (45 cycles) with one that does (23 cycles). Give or take a few cycles.  This isn't documented, but I assume at least in the case of `sample()` where it would stop only following the completion of the requested number of samples, that it won't ever take longer.  Stopping while it's actively converting (i.e. interrupting a requested sequence) would likely take longer (possibly requiring waiting for at least the current conversion to complete).

huntc: Sounds like we should just stop and block wait then. I’ll do the mod. It’ll make things even simpler!

huntc: Commit https://github.com/embassy-rs/embassy/pull/928/commits/30641d05640a7a3cd4d71c70c09694a3eef825e0 takes care of always stopping the sampling whether the futures are canceled or not.

Dirbaio (approved):
LGTM, thank you! :) Also thanks @peter9477 for the feedback and the measurements. 

bors r+

Issue #702, opened 2022-04-07
STM32 DMA double-buffering


Dirbaio:
We want to add some form of first-class double-buffering support, to allow endless streaming of data.

User 

Currently you can 

## Example use cases

- Streaming samples from ADC
- Streaming samples to DAC
- Streaming to/from I2S/SAI

## Requirements

1. Support double-buffering.
2. Gap between transfers must be as small as possible (Ideally none, max the latency of an irq. The latency of a wake is too much.)
3. There must not be UB even if irqs are delayed arbitrarily long (DMA must not wrap around and start overwriting the slice the user code is touching)

## How to do this?

Satisfying the requirements is tricky. `3` essentially means we can't use DMA modes that "wrap around by default". For example, with circular buffer you might do this:

    Start read onto a buffer in circular mode
    Loop {
        Wait for HTIE, this means the 1st half is filled
        Hand the 1st half to the user, they process it
        Wait for TCIE, this means the 2nd half is filled
        Hand the 2nd half to the user, they process it
    }}

However, if user takes too long to process the 1st half, DMA might wrap around and overwrite it from under them -> UB. 

Unfortunately I believe it's "fundamentally impossible" to wrap DMA circular mode in a safe rust API :'(

The way we use DMA has to be something like "start writing to buf1, queue a write to buf2. When you're done with buf1 or buf2 tell me. but DO NOT wrap around back to buf1 until I tell you to do so", so if user code takes too long, DMA just stops (and maybe loses data) but there's no UB.

### Idea 1: use M0AR/M1AR

There's some interesting ideas around on how to use M0AR/M1AR for this: writing a "poison" address to the next buffer (like 0xFFFF_FFFF) to get DMA to error and stop, then overwrite the poison with the real addr when it's safe to continue.

I'm not sure if this actually works in practice, or if it does it avoid UB in all cases.

Disadvantages:
- Only works when the two bufs have the same length. Hardware has 2 addr regs but only 1 len reg :(
- and only on chips with M0AR/M1AR (F2, F4, F7, H7, L5)

### Idea 2: transfer queuing

Add a way in `trait Channel` to queue transfers. You start one transfer, queue the next. When a transfer finishes, the IRQ handler starts the next transfer if queued. DMA stops if there's no queued transfer.

This allows code (e.g. the ADC hal) to:
- Start transfer to buf1
- Queue transfer to buf2
- When buf1 is filled, hand it to user code, then queue it again
[...]

PR #932, opened 2022-08-29
STM32 circular BDMA - single buffer

AntoineMugnier:
related to issue #702 

The purpose of this PR is to allow Peripheral->Mem transfer using circular BDMA in the embassy-stm32 DMA library.
The new method `start_circular_read()` fulfill this purpose.

As a first implementation of this new method, the `read_continuous()` method has been developped in the adc f1 library (stm32f103) and a new example has been developed in the stm32f1 example directory (adc_dma.rs)


embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +234,75 @@ impl<'d, T: Instance> Adc<'d, T> {
         self.convert()
     }
 
+    pub async fn read_continuous<S, const N: usize, U>(
+        &mut self,
+        pin: &mut impl AdcPin<T>,
+        mut rxdma: PeripheralRef<'_, U>,
```
AntoineMugnier, line 240: I chosed to pass the DMA channel as a reference instead of sending it via the peripheral structure constructor.
As such, the dma channel can be re-used  safely for another peripheral once the transfer is done.
Sending a reference to the dma channel when constructing the peripheral structure means the channel is "bound forever" to the peripheral, which is not good in my opinion

embassy-stm32/src/dma/bdma.rs
```diff
@@ -270,10 +364,31 @@ mod low_level_api {
 
         if isr.teif(channel_num) {
             panic!("DMA: error on BDMA@{:08x} channel {}", dma.0 as u32, channel_num);
+        } else if STATE.data_ready[index].load(Ordering::SeqCst) {
+            if cr.read().dir() == vals::Dir::FROMPERIPHERAL {
+                panic!("DMA: Concurrent access on the same buffer half between CPU and BDMA@{:08x} channel {}. User data processing may be too slow", dma.0 as u32, channel_num);
+            } else {
+                panic!("DMA: Concurrent access on the same buffer half between CPU and BDMA@{:08x} channel {}.Writting on the buffer by the user may be too slow", dma.0 as u32, channel_num);
```
chemicstry, line 371: This direction is not technically an UB, just random data being sent out to peripheral. Maybe it would be better to report an error instead of panic? Although, it would complicate code a bit and might be better left for future PR.

AntoineMugnier: So on your opinion, I should just delete it for now ?

embassy-stm32/src/adc/f1.rs
lulf: Another approach to that could be to drop the adc peripheral when you want to 'free' your channel for use by another peripheral. At least that's the approach taken for other peripherals afaict.

AntoineMugnier: Yeah but if you want to alternate dma transfer with two peripherals with the same channel, you constantly need to drop and re-construct them, which is I find not ideal.
But I think this situation would be kind of rare, so I'll change that to be homogenous with other peripherals.

embassy-stm32/src/dma/bdma.rs
chemicstry: No, I think it's okay as is. It can be addressed in another PR adding different transfer direction (i.e. DAC) to keep this concise.

embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +238,75 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+        sampler: &mut S,
+    ) where
+        S: FnMut(&[u16; N]) -> SamplerState,
+        RXDMA: RxDma<T>,
+    {
+        let rx_request = self.rxdma.request();
+        let rx_src = T::regs().dr().ptr() as *mut u16;
+
+        unsafe {
+            self.rxdma
```
lulf, line 254: Phase 1, start the DMA

embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +238,75 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+
+        // Configure the channel to sample
+        unsafe { T::regs().sqr3().write(|reg| reg.set_sq(0, pin.channel())) }
+
+        //Enable ADC
+        unsafe {
+            T::regs().cr2().modify(|reg| {
+                reg.set_adon(true);
+                reg.set_swstart(true);
+            });
```
lulf, line 283: Phase 2: Setup ADC. This could be the first closure

embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +238,75 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+                    buf_index = !buf_index & 0x01; // switch the buffer index (0/1)
+                    return Poll::Pending;
+                } else {
+                    return Poll::Ready(());
+                }
+            } else {
+                return Poll::Pending;
+            }
+        })
+        .await;
```
lulf, line 304: Phase 3, the outer part of setting waker, checking if data is ready etc. is generic, but the inner is not.

embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +238,75 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+            if self.rxdma.is_data_ready() {
+                let sampler_state = sampler(&data[buf_index]);
+                self.rxdma.set_data_processing_done();
+
+                if sampler_state == SamplerState::Sampled {
+                    buf_index = !buf_index & 0x01; // switch the buffer index (0/1)
+                    return Poll::Pending;
+                } else {
+                    return Poll::Ready(());
+                }
```
lulf, line 299: Phase 4, specific to ADC, could be the second closure.

embassy-stm32/src/adc/f1.rs
```diff
@@ -221,6 +238,75 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+            } else {
+                return Poll::Pending;
+            }
+        })
+        .await;
+
+        self.rxdma.request_stop();
```
lulf, line 306: Phase 5: Generic

lulf (commented):
I've tried to mark the different phases of the operation in the code comments, and I think the 'circular DMA' pattern could be captured into something like this that implements the common setup, polling and finalizing of the operation:

```
self.rxdma.receive_continuous(|| { PHASE 2: ADC SETUP CODE }, || { PHASE 4: ADC CHECK CODE} ).await;
```

The advantage being that the DMA API exposes less of the internals, and that one could potentially reuse this for other circular DMA uses.

Dirbaio:
closing since DMA already has ringbuffer support nowadays

PR #933, opened 2022-08-29
embassy-rp: Add async SPI

kalkyl:


MathiasKoch:
Nice!

You should probably add a HIL test as well?

Ideally one without DMA and one with. This way we're guaranteed it'll never break :)

Quote @Dirbaio :

> To run the tests locally you'll need to install [teleprobe](https://github.com/embassy-rs/teleprobe), then cargo run will work because teleprobe is confgured as a [runner](https://github.com/embassy-rs/embassy/blob/master/tests/rp/.cargo/config.toml#L7). (probe-run doesn't work because it can't boot from RAM)

lulf (commented):
I'm not sure if this is possible with the rp2040, but the stm32 SPI allows enabling DMA for only TX or only RX:

```
Spi<'d, T: Instance, Tx, Rx>
```

Where Tx or Rx can be NoDma or a Channel. Ignore me if that doesn't make sense for rp2040 :) 

Similarily, it would be good to keep the argument order consistent between HALs, currently the stm32 has (peri, sck, mosi, miso, txdma, rxdma, freq, config) and rp will have (peri, txdma, rxdma, clk, mosi, miso, config).

Dirbaio:
> which pins are MOSI/MISO for the RP2040 in his HIL though.

Just added a wire between GP3 and GP4 :) 

> the stm32 SPI allows enabling DMA for only TX or only RX: `Spi<'d, T: Instance, Tx, Rx>`

when @MathiasKoch implemented the UART we decided to not have generic params for DMA channels, using `AnyChannel` instead. In stm32 we're forced to because there's no `AnyChannel` yet (it's harder to implement because H7 has both DMA and BDMA). 

I actually like the embassy-rp way more, less generics is better. (and I'd like to try to make it work on stm32 someday). The downside is you can't have DMA for only one direction, but I think that's aceptable.

embassy-rp/src/spi.rs
```diff
@@ -57,8 +64,8 @@ fn calc_prescs(freq: u32) -> (u8, u8) {
     ((presc * 2) as u8, (postdiv - 1) as u8)
 }
 
-impl<'d, T: Instance> Spi<'d, T> {
-    pub fn new(
+impl<'d, T: Instance, M: Mode> Spi<'d, T, M> {
+    pub fn new_blocking(
```
Dirbaio, line 68: to be consistent, `new`s need some rearranging:

```rust
impl<'d, T: Instance> Spi<'d, T, Blocking> {
    fn new_blocking
    fn new_blocking_rxonly
    fn new_blocking_txonly
}
impl<'d, T: Instance> Spi<'d, T, Async> {
    fn new
    fn new_rxonly
    fn new_txonly
}
```

Blocking constructors are inside an `impl` block that fixes `M=Blocking`, same for async ones. This ensures you can't construct an async SPI with the blocking constructor or vice versa.

embassy-rp/src/spi.rs
```diff
@@ -78,26 +87,48 @@ impl<'d, T: Instance> Spi<'d, T> {
 
     pub fn new_txonly(
         inner: impl Peripheral<P = T> + 'd,
+        tx_dma: Option<PeripheralRef<'d, AnyChannel>>,
```
Dirbaio, line 90: these should be `impl Peripheral<P = impl Channel> + 'd` so that you can't pass None. We don't want to allow creating async SPIs without DMA.

embassy-rp/src/spi.rs
```diff
@@ -225,19 +261,109 @@ impl<'d, T: Instance> Spi<'d, T> {
...
+            // is held across an await and makes the future non-Send.
+            crate::dma::read(rx_ch, self.inner.regs().dr().ptr() as *const _, rx_buffer, T::RX_DREQ)
+        };
+        join(tx_transfer, rx_transfer).await;
+        Ok(())
+    }
+
+    pub async fn transfer_in_place(&mut self, words: &mut [u8]) -> Result<(), Error> {
+        let (ptr, len) = crate::dma::slice_ptr_parts(words);
+        let tx_buffer = unsafe { core::slice::from_raw_parts(ptr as *const _, len) };
```
Dirbaio, line 340: it is UB to create overlapping slices like this. Best fix is probably using raw pointers, like the stm32 spi,

embassy-rp/src/spi.rs
```diff
@@ -346,7 +485,56 @@ mod eh1 {
     }
 }
 
-impl<'d, T: Instance> SetConfig for Spi<'d, T> {
+cfg_if::cfg_if! {
```
Dirbaio, line 488: use `#[cfg(...)] mod eha { ... }`. 

(Old code uses `cfg_if!` because rustc still parses code inside disabled cfg's, and Rust stable at that time couldn't parse the new GAT where-clause location. This is not the case anymore)

Dirbaio (changes requested):

Dirbaio:
bors try

(to check HIL tests work)

embassy-rp/src/spi.rs
kalkyl: You mean to change the dma fns to work with raw pointers instead?

Dirbaio: yes... I can't think of any other way to make it sound :(

embassy-rp/src/spi.rs
```diff
@@ -225,19 +201,240 @@ impl<'d, T: Instance> Spi<'d, T> {
...
+            Some(mosi.map_into()),
+            Some(miso.map_into()),
+            None,
+            config,
+        )
+    }
+
+    pub fn new_txonly(
+        inner: impl Peripheral<P = T> + 'd,
+        tx_dma: impl Peripheral<P = impl Channel> + 'd,
```
Dirbaio, line 289: for consistency, the dma args should be after the pin args. (same for new_rxonly and new_inner)

Dirbaio (approved):
Thank you! :rocket: 

bors r+

matoushybl:
bors d+

PR #934, opened 2022-08-30
(embassy-rp): Add Buffered UART implementation

MathiasKoch:
### Questions & concerns: 
- ~~Would it make sense to add `RxBufferedUart` and `TxBufferedUart`, for cases where you would want to only buffer one way?~~
- ~~Do I need to be monitoring more interrupt flags than `Receive` & `Receive timeout`?~~

This PR adds working `BufferedUart` implementation, along with `RxBufferedUart` and `TxBufferedUart`. The implementation leaves room for improvement with respect to performance, as it still does not utilize DMA nor the internal UART buffers.


lulf:
> ### Questions & concerns:
> 
>     * Would it make sense to add `RxBufferedUart` and `TxBufferedUart`, for cases where you would want to only buffer one way?
> 

IMHO it would be very useful, not only for "selective buffering", but also because you can have RX and TX work independently, for instance in an AT command driver.

>     * Do I need to be monitoring more interrupt flags than `Receive` & `Receive timeout`?

MathiasKoch:
Exactly my thoughts. 

And i have precisely the need for bufferedRx, but i don't really care about the buffer for TX in my AT application

MathiasKoch:
@Dirbaio Ready for review

Only thing is
> The implementation leaves room for improvement with respect to performance, as it still does not utilize DMA nor the internal UART buffers.

But I still think this is much better than nothing, and then we can iterate on the performance.

embassy-rp/Cargo.toml
```diff
@@ -7,9 +7,7 @@ edition = "2021"
 src_base = "https://github.com/embassy-rs/embassy/blob/embassy-rp-v$VERSION/embassy-rp/src/"
 src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-rp/src/"
 features = ["nightly", "defmt", "unstable-pac", "unstable-traits"]
-flavors = [
-    { name = "rp2040", target = "thumbv6m-none-eabi" },
-]
+flavors = [{ name = "rp2040", target = "thumbv6m-none-eabi" }]
```
Dirbaio, line 10: could you not auto-reformat?

embassy-rp/src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub mod dma;
 pub mod gpio;
 pub mod interrupt;
 pub mod spi;
+#[cfg(feature = "time-driver")]
```
Dirbaio, line 11: these changes are unrelated. Making the time-driver optional sounds good to me, but could you do it in a separate PR?

MathiasKoch:
That should do it :+1:

embassy-rp/src/uart/buffered.rs
```diff
@@ -0,0 +1,535 @@
...
+    }
+}
+
+impl<'d, T: Instance + 'd> embedded_io::asynch::Read for RxBufferedUart<'d, T> {
+    type ReadFuture<'a> = impl Future<Output = Result<usize, Self::Error>>
+    where
+        Self: 'a;
+
+    fn read<'a>(&'a mut self, buf: &'a mut [u8]) -> Self::ReadFuture<'a> {
+        poll_fn(move |cx| {
```
lulf, line 339: This seems to share logic with BufferedUart Read and seems like something that shouldn't be repeated. Could they somehow be merged/share this implementation?

embassy-rp/src/uart/buffered.rs
```diff
@@ -0,0 +1,535 @@
...
+}
+
+impl<'d, T: Instance + 'd> embedded_io::asynch::BufRead for RxBufferedUart<'d, T> {
+    type FillBufFuture<'a> = impl Future<Output = Result<&'a [u8], Self::Error>>
+    where
+        Self: 'a;
+
+    fn fill_buf<'a>(&'a mut self) -> Self::FillBufFuture<'a> {
+        poll_fn(move |cx| {
+            self.inner.with(|state| {
```
lulf, line 415: Same as with the non-buffered versions, this also shares code with the BufferedUart impl.

embassy-rp/src/uart/buffered.rs
```diff
@@ -0,0 +1,535 @@
...
+    }
+}
+
+impl<'d, T: Instance + 'd> embedded_io::asynch::Write for TxBufferedUart<'d, T> {
+    type WriteFuture<'a> = impl Future<Output = Result<usize, Self::Error>>
+    where
+        Self: 'a;
+
+    fn write<'a>(&'a mut self, buf: &'a [u8]) -> Self::WriteFuture<'a> {
+        poll_fn(move |cx| {
```
lulf, line 497: Also shares code with BufferedUart

lulf (commented):
LGTM but maybe the duplicated code for TxBufferedUart, RxBufferedUart and BufferedUart should be put into some internal functions they can share?

MathiasKoch:
@lulf Valid points. Should be addressed now :+1: 

@Dirbaio Think this is ready to go?

embassy-rp/src/uart/buffered.rs
```diff
@@ -0,0 +1,498 @@
+use core::future::Future;
+use core::task::{Poll, Waker};
+
+use atomic_polyfill::{compiler_fence, Ordering};
+use embassy_cortex_m::peripheral::{PeripheralMutex, PeripheralState, StateStorage};
+use embassy_hal_common::ring_buffer::RingBuffer;
+use embassy_sync::waitqueue::WakerRegistration;
+use futures::future::poll_fn;
```
Dirbaio, line 8:
```suggestion
use core::future::poll_fn;
```

Dirbaio:
LGTM overall. Same as stm32, ideally BufferedUart would use DMA... but that's difficult to achieve so irq-driven is fine for now if it's useful to people I guess.

Could you add a HIL test?

MathiasKoch:
> Could you add a HIL test?

Done :+1:

tests/rp/src/bin/uart_buffered.rs
```diff
@@ -0,0 +1,37 @@
...
+    let config = Config::default();
+    let uart = Uart::new_blocking(uart, tx, rx, config);
+
+    let irq = interrupt::take!(UART0_IRQ);
+    let tx_buf = &mut [0u8; 32];
+    let rx_buf = &mut [0u8; 32];
+    let mut state = State::new();
+    let mut uart = BufferedUart::new(&mut state, uart, irq, tx_buf, rx_buf);
+
+    let data = [0xC0, 0xDE];
```
Dirbaio, line 28: BufferedUart can buffer more than the 2 bytes in the hardware fifo, it'd be a good test that it is actually buffering.

Dirbaio:
Hmm the test is not passing for me locally. Let's try CI just in case..?

bors try

MathiasKoch:
Huh, actually forgot to run it locally 🤣 just assumed it would work, as i am using the bufferedUart elsewhere.

Will correct your comment and make it pass tomorrow morning ☺️

Dirbaio:
Rebased for you. 

bors try
bors d+

MathiasKoch:
bors try

MathiasKoch:
bors try

PR #935, opened 2022-08-30
Remove generic const expressions from embassy-boot

lulf:
* Remove the need for generic const expressions and use buffers provided in the flash config.
* Extend embedded-storage traits to simplify generics.
* Document all public APIs
* Add toplevel README
* Expose AlignedBuffer type for convenience.
* Update examples

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -1,53 +1,54 @@
 #![feature(type_alias_impl_trait)]
 #![feature(generic_associated_types)]
-#![feature(generic_const_exprs)]
-#![allow(incomplete_features)]
 #![no_std]
-///! embassy-boot is a bootloader and firmware updater for embedded devices with flash
-///! storage implemented using embedded-storage
-///!
-///! The bootloader works in conjunction with the firmware application, and only has the
-///! ability to manage two flash banks with an active and a updatable part. It implements
-///! a swap algorithm that is power-failure safe, and allows reverting to the previous
-///! version of the firmware, should the application crash and fail to mark itself as booted.
-///!
-///! This library is intended to be used by platform-specific bootloaders, such as embassy-boot-nrf,
-///! which defines the limits and flash type for that particular platform.
-///!
+#![warn(missing_docs)]
+#![doc = include_str!("../../README.md")]
 mod fmt;
 
-use embedded_storage::nor_flash::{NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
+use embedded_storage::nor_flash::{ErrorType, NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
 use embedded_storage_async::nor_flash::AsyncNorFlash;
 
 const BOOT_MAGIC: u8 = 0xD0;
 const SWAP_MAGIC: u8 = 0xF0;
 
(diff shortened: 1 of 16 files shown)


lulf:
Tested on nrf52-dk and nucleo-wl55jc

embassy-boot/nrf/src/lib.rs
```diff
@@ -1,19 +1,21 @@
...
 use embassy_nrf::nvmc::{Nvmc, PAGE_SIZE};
 use embassy_nrf::peripherals::WDT;
 use embassy_nrf::wdt;
 use embedded_storage::nor_flash::{ErrorType, NorFlash, ReadNorFlash};
 
+/// A bootloader for nRF devices.
 pub struct BootLoader {
-    boot: embassy_boot::BootLoader<PAGE_SIZE>,
+    boot: embassy_boot::BootLoader,
+    magic: AlignedBuffer<4>,
```
matoushybl, line 17: Is this equal to the WRITE_SIZE? What do you think about extracting it into a public constant (since the length will be used for magic buffer initialization in user's code)?

embassy-boot/boot/src/lib.rs
```diff
@@ -599,7 +643,13 @@ impl FirmwareUpdater {
...
 
-    // Write to a region of the DFU page
+    /// Write data to a flash page.
+    ///
+    /// The buffer must follow alignment requirements of the target flash and a multiple of page size big.
+    ///
+    /// # Safety
+    ///
+    /// Failing to meet alignment and size requirements may result in a panic.
     pub async fn write_firmware<F: AsyncNorFlash>(
```
matoushybl, line 653: Unrelated to this refactor - should the user be able to write to the DFU partition when the state is SWAP? I added precondition for it into my FirmwareUpdater, because then in the case of a reboot, the firmware didn't boot.

matoushybl (approved):
Looks good to me, there are only some minor notes 🚀

embassy-boot/nrf/src/lib.rs
lulf: Good idea. I'll add the constant to embassy_nrf which is 'hiding' it in the embedded_storage trait impl.

embassy-boot/boot/src/lib.rs
```diff
@@ -1,53 +1,54 @@
...
-///!
+#![warn(missing_docs)]
+#![doc = include_str!("../../README.md")]
 mod fmt;
 
-use embedded_storage::nor_flash::{NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
+use embedded_storage::nor_flash::{ErrorType, NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
 use embedded_storage_async::nor_flash::AsyncNorFlash;
 
 const BOOT_MAGIC: u8 = 0xD0;
```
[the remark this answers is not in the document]
lulf: Maybe we can improve the FirmwareUpdater to allow this mode of operation instead? Feels like an implementation detail that we shouldn't leak if we can avoid it.

embassy-boot/boot/src/lib.rs
lulf: Yup, maybe that's a good check. I'll add it.

PR #936, opened 2022-08-30
Add split() method to BufferedUarte in embassy-nrf

ZoeyR:
I haven't completed testing this yet. I'm creating this PR early so that I can get corrected if I went way off course.

This PR adds a `split()` method to `BufferedUarte` as discussed on matrix.

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -207,7 +207,8 @@ impl<'d, U: UarteInstance, T: TimerInstance> BufferedUarte<'d, U, T> {
     async fn inner_read<'a>(&'a self, buf: &'a mut [u8]) -> Result<usize, core::convert::Infallible> {
         poll_fn(move |cx| {
             let mut do_pend = false;
-            let res = self.inner.borrow_mut().with(|state| {
+            let mut inner = self.inner.borrow_mut();
+            let res = inner.with(|state| {
                 compiler_fence(Ordering::SeqCst);
                 trace!("poll_read");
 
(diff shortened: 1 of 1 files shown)


Dirbaio:
LGTM overall.

It'd be nice if the write/read code wasn't duplicated. Perhaps you can make the Tx/Rx structs have a `&BufferedUarte`, and add `read`, `write` async fns to `BufferedUarte` that take `&self`, and then call these from all the trait impls?

ZoeyR:
Yeah, that should be workable. I assume for now we'd want those to be internal to the module?

Dirbaio:
yes

ZoeyR:
Factored those out!

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -232,22 +227,64 @@ impl<'d, U: UarteInstance, T: TimerInstance> embedded_io::asynch::Read for Buffe
...
     }
-}
 
-impl<'d, U: UarteInstance, T: TimerInstance> embedded_io::asynch::BufRead for BufferedUarte<'d, U, T> {
-    type FillBufFuture<'a> = impl Future<Output = Result<&'a [u8], Self::Error>>
-    where
-        Self: 'a;
+    async fn inner_write<'a>(&'a self, buf: &'a [u8]) -> Result<usize, core::convert::Infallible> {
+        poll_fn(move |cx| {
+            let res = self.inner.borrow_mut().with(|state| {
```
Dirbaio, line 240: avoid borrowing twice: `let inner = self.inner.borrow_mut();` then use `inner`.

ZoeyR: I did it for the functions that double-borrowed. I left the ones with a single borrow intact. Should I switch those too?

Dirbaio (approved):
LGTM, feel free to merge yourself when you've tested it out! Thanks :)

bors d+

ZoeyR:
Okay, did some testing and it seems to work!

bors r+

PR #938, opened 2022-08-31
Do not use cfg_if for embedded-hal-async feature gates.

Dirbaio:
Old code used `cfg_if!` because rustc still parses code inside disabled cfg's, and Rust stable at that time couldn't parse the new GAT where-clause location. This is not the case anymore.

bors r+


Dirbaio:
absdgasbg;jasf


bors r+

PR #943, opened 2022-09-03
Use embassy_executor::main in runtime.adoc

danbev:
This commit replaces `embassy::main` with `embassy_executor::main` in the
runtime documentation page.

Refs: https://embassy.dev/dev/runtime.html


lulf (approved):
thank you!

bors r+

PR #946, opened 2022-09-12
sync/signal: wake old waker on overflow instead of panicking.

Dirbaio:
This makes behavior consistent with `WakerRegistration`. It allows canceling `wait` in one task and then calling `wait` in another. If two tasks are `wait`ing concurrently the signal will be received by only one of them, randomly.


Dirbaio:
solves the [original issue](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$MnPjGUOVDC2XGvV8FRcTeLVydNxFLOUXx_egHnulKWY?via=matrix.org&via=grusbv.com&via=community.rs), so merging.

bors r+

PR #949, opened 2022-09-16
(embassy-rp): Implement RealTimeClock

MathiasKoch:
Basically a 1:1 port of the great implementation effort made by `rp-hal`

### DIFF
--- a/embassy-rp/src/rtc/filter.rs
+++ b/embassy-rp/src/rtc/filter.rs
@@ -1,0 +1,100 @@
+use super::DayOfWeek;
+use crate::pac::rtc::regs::{IrqSetup0, IrqSetup1};
+
+/// A filter used for [`RealTimeClock::schedule_alarm`].
+///
+/// [`RealTimeClock::schedule_alarm`]: struct.RealTimeClock.html#method.schedule_alarm
+#[derive(Default)]
+pub struct DateTimeFilter {
+    /// The year that this alarm should trigger on, `None` if the RTC alarm should not trigger on a year value.
+    pub year: Option<u16>,
+    /// The month that this alarm should trigger on, `None` if the RTC alarm should not trigger on a month value.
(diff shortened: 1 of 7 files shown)


embassy-rp/src/rtc/filter.rs
```diff
@@ -0,0 +1,100 @@
...
+        if let Some(day_of_week) = self.day_of_week {
+            w.set_dotw_ena(true);
+            let bits = super::datetime::day_of_week_to_u8(day_of_week);
+
+            w.set_dotw(bits);
+        }
+        if let Some(hour) = self.hour {
+            w.set_hour_ena(true);
+            w.set_hour(hour);
+        }
```
lulf, line 90: Should there be an else statement to these or are they cleared before each filter is configured?

```
        if let Some(hour) = self.hour {
            w.set_hour_ena(true);
            w.set_hour(hour);
        } else {
            w.set_hour_ena(false);
        }
```

lulf (commented):
LGTM, one question on filter set/unset

embassy-rp/src/rtc/filter.rs
MathiasKoch: Hmm.. it's a 100% copy paste from rp-hal, so i think it should be okay.. will give it a check though

MathiasKoch: It's not super obvious at first, but they are indeed reset automatically by using `write` rather than `modify` here: https://github.com/embassy-rs/embassy/pull/949/files/feead3ae89d57e9f0ff7d7a264136d3e89aaebcf#diff-2e8895df0110ae9c757ad2b245f8f997a24d7c78ccbf071571336c530768108fR141-R143

MathiasKoch:
@Dirbaio Ready to go!

lulf (approved):
LGTM

bors r+

PR #950, opened 2022-09-16
Add .into_inner() and .get_mut() to Mutex

hulthe:
Similar to the methods on std Mutex, these methods allow accessing the underlying data without locking the mutex when you have exclusive access to it.


Dirbaio (approved):
Nice! Thank you :D

bors r+

PR #951, opened 2022-09-16
(embassy-rp): Implementation of generic flash mutation access

MathiasKoch:
I have attempted to utilize the work done in `rp2040-flash` by implementing `embedded-storage` traits on top, for RP2040.

Concerns:
1. ~~Should the DMA be paused where I have put a FIXME note? `DMA_CHx.ctrl_trig().write(|w| { w.set_en(false) })`? If so, how to properly do that without have control over the peripheral for the DMA channels? And if so, I assume we should only re-enable/unpause the ones that were enabled before?~~
2. ~~Should I make sure core2 is halted as part of this code? I am not sure if https://github.com/jannic/rp2040-flash/blob/ea8ab1ac807a7ab2b28a18bb5ca2e42495bb744d/examples/flash_example.rs#L103-L109 is heavy/slow code to run?~~
3. ~~Any good way of making this configurable over `FLASH_SIZE`, `WRITE_SIZE` and `ERASE_SIZE` without doing it as generics or parameters, as those make it possible to do differing configs throughout the same program, which feels wrong? Preferably, a compile-time option?~~

**EDIT:**
I have implemented the flash API here under the assumption that all external QSPI nor flashes are infact `Multiwrite` capable, as this makes it possible to use the ROM function for writes of 1 bytes at a time.

I have also added a HIL test for this, but because HIL tests are running 100% from RAM and I wanted to make sure it still works when running from flash, I have also added an example testing erase/write cycles of entire sectors, as well as single bytes in multi-write style.

Ping @Dirbaio


Dirbaio:
> Should the DMA be paused 

Oof this is cursed. I guess yes, or this will be a problem when DMAing from flash. 

> If so, how to properly do that without have control over the peripheral for the DMA channels?

If the flash code globally disables irqs and stops the other core you have the guarantee that no other code will race you, so you can go ahead and do it unsafely I guess...

> And if so, I assume we should only re-enable/unpause the ones that were enabled before?

Yes...

Also, are we sure EN=0/EN=1 pauses/resumes the channel fully seamlessly?

Alternatively we can disallow DMAing from flash... nRF has this limitation in hardware and it's not *that* bad, it means you have to copy through a RAM buffer though. Perhaps it could be feature-gated, so you can either "DMA from flash" OR "write to flash", but not both in the same binary...

> Should I make sure core2 is halted as part of this code? I

In the final version yes, but I'd say it's OK to not do it for now since we don't support muticore yet. 

Note that code *resets* core2, we want to *pause* it instead. It'd have to send an irq to the other core (using the FIFO maybe) so that the other core "parks" itself in a RAM irq handler. Also note it's possible to do flash ops from core2, in this case it should park core1 :sweat: .

> Any good way of making this configurable over FLASH_SIZE, WRITE_SIZE and ERASE_SIZE without doing it as generics or parameters

Cargo features with a range of possible options? :P They're always powers of 2 so there shouldn't be too many "reasonable" options...?

MathiasKoch:
>Also, are we sure EN=0/EN=1 pauses/resumes the channel fully seamlessly?

From the RP2040 datasheet:

> DMA Channel Enable.
> When 1, the channel will respond to triggering events,
> which will cause it to become BUSY and start transferring
> data. When 0, the channel will ignore triggers, stop issuing
> transfers, and **pause the current transfer sequence (i.e.
> BUSY will remain high if already high)**

As I read that, it actually fully pauses any ongoing transfer, so I don't think this is too bad.

> In the final version yes, but I'd say it's OK to not do it for now since we don't support muticore yet.
>
> Note that code resets core2, we want to pause it instead. It'd have to send an irq to the other core (using the FIFO maybe) so that the other core "parks" itself in a RAM irq handler. Also note it's possible to do flash ops from core2, in this case it should park core1.

Noted.

> Cargo features with a range of possible options? :P They're always powers of 2 so there shouldn't be too many "reasonable" options...?

Hmm, yeah. I'm not super fond of that way of doing it. It just doesn't seem to scale very well :(

lulf:
> I have attempted to utilize the work done in `rp2040-flash` by implementing `embedded-storage` traits on top, for RP2040.
> 
> Concerns:
> 
>     1. Should the DMA be paused where I have put a FIXME note? `DMA_CHx.ctrl_trig().write(|w| { w.set_en(false) })`? If so, how to properly do that without have control over the peripheral for the DMA channels? And if so, I assume we should only re-enable/unpause the ones that were enabled before?
> 
>     2. Should I make sure core2 is halted as part of this code? I am not sure if https://github.com/jannic/rp2040-flash/blob/ea8ab1ac807a7ab2b28a18bb5ca2e42495bb744d/examples/flash_example.rs#L103-L109 is heavy/slow code to run?
> 
>     3. Any good way of making this configurable over `FLASH_SIZE`, `WRITE_SIZE` and `ERASE_SIZE` without doing it as generics or parameters, as those make it possible to do differing configs throughout the same program, which feels wrong? Preferably, a compile-time option?
> 

Although a slightly different situation, I've had a similar dilemma in embassy-boot generalizing these, and ended up using const generics for them. What about a combination that allows both uses: a GenericFlash using const generics, and a type alias `Flash` based on cargo features? Or would that be too confusing?

MathiasKoch:
> > I have attempted to utilize the work done in `rp2040-flash` by implementing `embedded-storage` traits on top, for RP2040.
> > 
> > Concerns:
> > 
> >     1. Should the DMA be paused where I have put a FIXME note? `DMA_CHx.ctrl_trig().write(|w| { w.set_en(false) })`? If so, how to properly do that without have control over the peripheral for the DMA channels? And if so, I assume we should only re-enable/unpause the ones that were enabled before?
> > 
> >     2. Should I make sure core2 is halted as part of this code? I am not sure if https://github.com/jannic/rp2040-flash/blob/ea8ab1ac807a7ab2b28a18bb5ca2e42495bb744d/examples/flash_example.rs#L103-L109 is heavy/slow code to run?
> > 
> >     3. Any good way of making this configurable over `FLASH_SIZE`, `WRITE_SIZE` and `ERASE_SIZE` without doing it as generics or parameters, as those make it possible to do differing configs throughout the same program, which feels wrong? Preferably, a compile-time option?
> > 
> 
> Although a slightly different situation, I've had a similar dilemma in embassy-boot generalizing these, and ended up using const generics for them. What about a combination that allows both uses: a GenericFlash using const generics, and a type alias `Flash` based on cargo features? Or would that be too confusing?

That was actually exactly what I was thinking about.
Either chip specific exports like boot2 does for their bootloaders, or '1M_4_1' kind of definitions, where 'FLASH-SIZE_ERASE-SIZE_READ-SIZE'

Dirbaio:
> What about a combination that allows both uses: a GenericFlash using const generics, and a type alias Flash based on cargo features? Or would that be too confusing?

It should be one or the other. Having 2 ways of doing the same thing is very confusing.

MathiasKoch:
Given the kind of extreme limitations of the rom-data flash API, that aims at wide flash support, my suggestion would be to make a repo similar to https://github.com/rp-rs/flash-algo where we can implement a few different implementations (QSPI, Dual-SPI, SPI) in various configurations, compile and link them fully for RAM, and then `include_bytes()` one of them similarly to https://github.com/rp-rs/rp2040-boot2.

Then we can do fallback to the flash API in rom-data, if users do not wish the added flash size of an optimized flash API.


W.R.T flash configurations & cargo features vs const generics, how about adding a "custom" peripheral for flash, such that you can only instantiate it once (panic on re-entrance), where you can pass flash-size, erase-size etc., and then you would have to pass around the flash struct like all other microcontrollers?

Thoughts @Dirbaio @lulf ?

Dirbaio:
Sounds good to me! :+1:  I think having to write RAM-linked "flash driver" blobs is inevitable given the ROM limitations...

> adding a "custom" peripheral for flash, such that you can only instantiate it once (panic on re-entrance), where you can pass flash-size, erase-size etc., and then you would have to pass around the flash struct like all other microcontrollers?

Not sure what you mean by "custom"? I'd make a normal FLASH peripheral (just a struct, no functionality) and then a Flash *driver* that you pass all that config when creating it. If we do this, the API would have the same structure as any other peripheral.

MathiasKoch:
> Not sure what you mean by "custom"? I'd make a normal FLASH peripheral (just the singleton struct, no functionality) and then a Flash driver that you pass all that config when creating it. If we do this, the API would have the same structure as any other peripheral.

By custom i just meant that it does not come from a pac peripheral. Not sure what the requirements are for the `peripheral!` macro in embassy?

Dirbaio:
Ah! then all our peripherals are already "custom". They're all invented by the HAL, we don't use the PAC ownership model. (rp and stm32 PACs don't have owned singletons at all because they're generated by chiptool. nrf PACs do because we're using the `nrf-rs` ones, but we just ignore them)

MathiasKoch:
Perhaps it would make sense to just have the RAM-blobs implement the CMSIS-Pack ABI? That way it would be re-usable by probe-rs once they figure out how they want to tackle https://github.com/probe-rs/probe-rs/issues/1212

MathiasKoch:
I've started looking into this again, and came up with a thought here:

Would it make sense to make N different flash RAM blobs that works in the same way as the boot2, where you have to `include_bytes!()` one of them, that is linked into a ram location, where each of the N versions would be e.g. QSPI, Dual SPI etc. based on chip capabilities. These RAM blobs would implement be implemented as the CMSIS-Pack ABI just like https://github.com/rp-rs/flash-algo/

What is left is then the considerations around `ERASE_SIZE`, `WRITE_SIZE`, `READ_SIZE`, `FLASH_SIZE`.

Would it be possible to make those configurable through a macro, in the same way as `defmt::timestamp!()`? (https://github.com/knurling-rs/defmt/blob/ce350f767ffe44bc12513e411a0440c47d5ba0f4/macros/src/items/timestamp.rs#L34-L59) 
Basically a macro that fills three different link symbols, and can only be called once in the entire binary (checked at link time)

Something like `embassy-rp::flash!(QSPI, 16 * 1024 * 1024, 2048, 4, 1)`, where `embassy-rp::flash!(FLASH VARIANT, FLASH SIZE, ERASE_SIZE, WRITE_SIZE, READ_SIZE)`? One could then do a fallback to the ROM function variant, if `embassy-rp::flash!()` is not called? 

Thoughts @Dirbaio @lulf

MathiasKoch:
@Dirbaio Ready for review!

ci.sh
```diff
@@ -121,7 +121,7 @@ cargo batch  \
     --- build --release --manifest-path tests/stm32/Cargo.toml --target thumbv7em-none-eabi --features stm32h755zi --out-dir out/tests/nucleo-stm32h755zi \
     --- build --release --manifest-path tests/stm32/Cargo.toml --target thumbv7em-none-eabi --features stm32wb55rg --out-dir out/tests/nucleo-stm32wb55rg \
     --- build --release --manifest-path tests/stm32/Cargo.toml --target thumbv7em-none-eabi --features stm32u585ai --out-dir out/tests/iot-stm32u585ai \
-    --- build --release --manifest-path tests/rp/Cargo.toml --target thumbv6m-none-eabi --out-dir out/tests/rpi-pico \
+    --- build --release --manifest-path tests/rp/Cargo.toml --target thumbv6m-none-eabi --features embassy-rp/time-driver --out-dir out/tests/rpi-pico \
```
Dirbaio, line 124: why is this needed? 

if it's needed it should go in tests/rp/Cargo.toml instead. Tests shouldn't require extra flags to succesfully build.

embassy-rp/src/rom_data.rs
```diff
@@ -7,7 +7,6 @@
 //! > on the device, as well as highly optimized versions of certain key
 //! > functionality that would otherwise have to take up space in most user
 //! > binaries.
-
```
Dirbaio, line 10: accidentally removed newline

embassy-rp/src/intrinsics.rs
```diff
@@ -1,5 +1,4 @@
 #![macro_use]
-
```
Dirbaio, line 2: accidentally removed newline

MathiasKoch:
Resolved

embassy-rp/src/flash.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// # Safety
+    ///
+    /// Nothing must access flash while this is running.
+    /// Usually this means:
+    ///   - interrupts must be disabled
+    ///   - 2nd core must be running code from RAM or ROM with interrupts disabled
+    ///   - DMA must not access flash memory
+    ///
+    /// `addr` and `len` parameters must be valid and are not checked.
+    pub unsafe fn flash_range_erase_and_program(addr: u32, data: &[u8], use_boot2: bool) {
```
Dirbaio, line 300: this is unused?

MathiasKoch: Currently, yes. 

Not sure if I should just remove it? It was merely to cover the functions provided by the ROM data by RP2040.

embassy-rp/src/flash.rs
```diff
@@ -0,0 +1,429 @@
...
+            unsafe { self.in_ram(|| ram_helpers::flash_range_program(unaligned_offset as u32, &pad_buf, true)) }
+        }
+
+        let remaining_len = bytes.len() - start_padding;
+        let end_padding = start_padding + PAGE_SIZE * (remaining_len / PAGE_SIZE);
+
+        // Write aligned slice of length in multiples of 256 bytes
+        // If the remaining bytes to be written is more than a full page.
+        if remaining_len >= PAGE_SIZE {
+            let aligned_data = &bytes[start_padding..end_padding];
```
Dirbaio, line 166: This will fail if `bytes` is a slice that points to FLASH (the compiler will do that if it's constant data). I think we'll have to buffer it through RAM.

embassy-rp/src/flash.rs
MathiasKoch: It might make sense to use it, if one were to implement `embedded_storage::Storage` traits.
That way the RMW operations could be one read + one `erase_and_program`

embassy-rp/src/flash.rs
MathiasKoch: Good catch!

MathiasKoch: Hmm, that kinda sucks as it would mean that large writes would result in many calls to `flash_range_program`, rather than the 3 it is capped at currently.

On the other hand it would make implementation much cleaner, as i could use `chunks_exact(PAGE_SIZE)`, and its `remainder()`

embassy-rp/src/flash.rs
Dirbaio: does it actually erase? code is identical to `flash_range_program` :rofl:

MathiasKoch: The difference lies within the `true, true` vs `false, true` in `flash_function_pointers`

Dirbaio: ah scratch that, there's the `flash_function_pointers(true, true)`, didn't see it.

dunno then, we can keep it if you want. Up to you

MathiasKoch:
@Dirbaio Ready for merge!

In case you didn't see it on matrix, I figured out the issue was due to the RTT header residing in flash:
>  debugger looking for the RTT header which caused the flash access which in turn caused a fault?
I had the same issue with rp2040-flash, with the following workaround: https://github.com/jannic/rp2040-flash/blob/ea8ab1ac807a7ab2b28a18bb5ca2e42495bb744d/examples/flash_example.rs#L88-L92 
>
>For defmt-rtt, it should be fixed with version 0.4.0, which includes https://github.com/knurling-rs/defmt/pull/683

Dirbaio (approved):
bors d+

MathiasKoch:
I don't understand why the bors test is failing? 
It works 100% of the times with teleprobe locally?
@Dirbaio

Dirbaio:
the CI has a Pico W, not sure if it matters..?

MathiasKoch:
Hmm. I just tested it on a Pico I had lying around here (non W version), and it works great there as well. Teleprobe tests pass 100% of times.

Not sure what is different with the CI, but I don't have a Pico W to test it on :(

PR #958, opened 2022-09-21
Implement proper `Drop` for `BufferedUarte`

ZoeyR:
The drop method in `BufferedUarte` was prone to hanging indefinitely and also didn't actually disable the peripheral. I mostly copied over the drop method from `Uarte` with some modifications since `BufferedUarte` could have a transmit lasting indefinitely.

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -429,14 +429,15 @@ impl<'a, U: UarteInstance, T: TimerInstance> Drop for StateInner<'a, U, T> {
     fn drop(&mut self) {
         let r = U::regs();
 
+        self.timer.stop();
+
         r.inten.reset();
         r.events_rxto.reset();
         r.tasks_stoprx.write(|w| unsafe { w.bits(1) });
-
         r.events_txstopped.reset();
         r.tasks_stoptx.write(|w| unsafe { w.bits(1) });
-        while r.events_txstopped.read().bits() == 0 {}
 
+        while r.events_txstopped.read().bits() == 0 {}
         while r.events_rxto.read().bits() == 0 {}
 
         r.enable.write(|w| w.enable().disabled());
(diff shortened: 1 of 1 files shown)


embassy-nrf/src/buffered_uarte.rs
```diff
@@ -427,23 +427,26 @@ impl<'u, 'd: 'u, U: UarteInstance, T: TimerInstance> embedded_io::asynch::Write
 
 impl<'a, U: UarteInstance, T: TimerInstance> Drop for StateInner<'a, U, T> {
     fn drop(&mut self) {
+        debug!("oh no, dropping uarte");
         let r = U::regs();
 
         // TODO this probably deadlocks. do like Uarte instead.
```
lulf, line 433: Remove this I guess?

lulf (commented):
LGTM

embassy-nrf/src/buffered_uarte.rs
ZoeyR: ahahahaha, whoops

ZoeyR:
Build issues are fixed, could I get an approval on this?

ZoeyR:
@lulf is there anything else that I need to do for this PR?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -429,21 +429,22 @@ impl<'a, U: UarteInstance, T: TimerInstance> Drop for StateInner<'a, U, T> {
     fn drop(&mut self) {
         let r = U::regs();
 
-        // TODO this probably deadlocks. do like Uarte instead.
+        r.inten.reset();
+        r.events_rxto.reset();
+        r.tasks_stoprx.write(|w| unsafe { w.bits(1) });
 
-        self.timer.stop();
```
lulf, line 434: I noticed this part is removed. Shouldn't the timer be stopped as well anymore or is that done implicitly?

lulf (commented):
Just one comment re: timer

embassy-nrf/src/buffered_uarte.rs
ZoeyR: I'll double check. I figured it was done implicitly since the normal Uarte also doesn't explicitly stop the timer.

ZoeyR: Turns out I was wrong. The normal Uarte did this with an OnDrop guard. Fixed.

lulf (approved):
LGTM, thanks!

bors r+

PR #959, opened 2022-09-21
Generic, executor-agnostic queue implementation

ivmarkov:
Hopefully relatively well documented.

Implementation relies on a fixed-size `SortedLinkedList` from `heapless`. (By default, for up to 128 timer schedules, but we can lower this number to - say - 64.)

As discussed earlier, on queue overflow, the `WakerRegistration` approach is utilized, whereas the waker that is ordered first in the queue is awoken to make room for the incoming one (which might be the waker that would be awoken after all!). Wakers are compared with `Waker::will_wake`, so the queue should actually not fill up that easily, if at all.

I've left provisions for the user to manually instantiate the queue using a dedicated macro - `generic_queue!` so that users willing to adjust the queue size, or users (like me) who have to use the queue in a complex "on-top-of-RTOS-but-the-timer-driver-calling-back-from-ISR" scenario can customize the mutex that protects the queue.

The one thing I'm not completely happy with is the need to call `{ embassy_time::queue::initialize() }` early on before any futures using embassy-time are polled, which is currently on the shoulders of the user. I'm open to any ideas where we can get rid of this and do it on the first call to `_embassy_time_schedule_wake`, without introducing very complex combinations of critical sections, atomics and whatnot.


embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+            queue: SortedLinkedList::new_u8(),
+            alarm_at: Instant::MAX,
+        }
+    }
+
+    fn schedule(&mut self, at: Instant, waker: &Waker, alarm_schedule: &AtomicU64) {
+        self.queue
+            .find_mut(|timer| timer.waker.will_wake(waker))
+            .map(|mut timer| {
+                timer.waker = waker.clone();
```
Dirbaio, line 101: if `timer.waker.will_wake(waker)` then it's *not* necessary to clone the waker.

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+        self.update_alarm();
+    }
+
+    fn handle_alarm_callback(ctx: *mut ()) {
+        unsafe { (ctx as *const Self).as_ref().unwrap() }.handle_alarm();
+    }
+}
+
+unsafe impl<const N: usize, R: RawMutex + 'static> Send for Queue<N, R> {}
+unsafe impl<const N: usize, R: RawMutex + 'static> Sync for Queue<N, R> {}
```
Dirbaio, line 239: This is not right because `Cell` is not Sync. This can be a problem in practice if the user calls `.initialize()` from one thread and `.update_alarm()` from another (through polling a timer), which would cause a data race on the `alarm` Cell.

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+//!
+//! # Initializing the queue
+//! - Call ```unsafe { embassy_time::queue::initialize(); }``` early on in your program, before any of your futures that utilize `embassy-time` are polled.
+//!
+//! # Customizing the queue
+//! - It is possible to customize two aspects of the queue:
+//!   - Queue size:
+//!     By default, the queue can hold up to 128 timer schedules and their corresponding wakers. While it will not crash if more timer schedules are added,
+//!     the performance will degrade, as one of the already added wakers will be awoken, thus making room for the new timer schedule and its waker.
+//!   - The mutex (i.e. the [`RawMutex`](embassy_sync::blocking_mutex::raw::RawMutex) implementation) utilized by the queue:
```
Dirbaio, line 19: This feels way too tricky to push onto the user's responsibility. IMO the time driver should be "critical-section compliant", meaning that it won't call arbitrary code (embassy's code) from a context that "skips" the critical-section protections. 

For example, if the CS impl in use is "lock the RTOS scheduler to avoid the current thread getting switched out, but not disable IRQs" then the time driver shouldn't call the alarm callback from an ISR, only from regular threads.

This would mean we can hardcode `CriticalSectionMutex` here. Which is exactly what `critical-section` is for, libs should be able to blindly rely on it.

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+    }
+
+    fn check_initialized(&self) {
+        if self.alarm.get().is_none() {
+            panic!("Queue is not initialized yet");
+        }
+    }
+
+    fn update_alarm(&self) {
+        // Need to set the alarm when we are *not* holding the mutex on the inner queue
```
Dirbaio, line 215: I'm starting to think we should forbid `set_alarm` from calling the callback synchronously. It could return an "instant already in the past" error instead, so the timer queue knows it's running behind and it does another "round" of expirations. 

It'd save a lot of trouble here, because it'd allow putting all the state in `InnerQueue`. It wouldn't make the executor-tied impl much more complex either, it'd just mean doing another executor loop on set_alarm error.

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+//!
+//! # Enabling the queue
+//! - Enable the Cargo feature `generic-queue`. This will automatically instantiate the queue.
+//!
+//! # Initializing the queue
+//! - Call ```unsafe { embassy_time::queue::initialize(); }``` early on in your program, before any of your futures that utilize `embassy-time` are polled.
+//!
+//! # Customizing the queue
+//! - It is possible to customize two aspects of the queue:
+//!   - Queue size:
```
Dirbaio, line 16: Maybe queue size could be configured via Cargo features? I don't like that there's 2 ways of initializing the queue, either implicitly or via the macro. This'd allow doing it always implicitly.

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,617 @@
...
+//! This module provides a timer queue that works with any executor.
+//!
+//! In terms of performance, this queue will likely be less efficient in comparison to executor-native queues,
+//! like the one provided with e.g. the `embassy-executor` crate.
+//!
+//! # Enabling the queue
+//! - Enable the Cargo feature `generic-queue`. This will automatically instantiate the queue.
+//!
+//! # Initializing the queue
+//! - Call ```unsafe { embassy_time::queue::initialize(); }``` early on in your program, before any of your futures that utilize `embassy-time` are polled.
```
Dirbaio, line 12: IMO we shouldn't require the user to explicitly initialize anything. At every schedule, check if the alarm handle is None and create it if it is. I don't think it'd add much overhead, the current `.unwrap()` is already doing the "check if the alarm handle is None" part at every shedule.

embassy-time/src/queue.rs
ivmarkov: Agreed and fixed.

embassy-time/src/queue.rs
ivmarkov: That's precisely the reason why `Queue::initialize` is unsafe - because it has to be called *before* any futures that use `embassy-time` are polled.

I can surely protect the `Cell` instance with a mutex as well, but this mutex - at least in the RTOS case - might need to be of a different type than the one protecting the inner queue structure, I'm afraid. I'll elaborate on that further down, where the `initialize` method itself is being discussed.

embassy-time/src/queue.rs
ivmarkov: I don't think we can blindly simplify it and hard-code it to "the CS impl the HAL provides to the `critical-section` crate should be compatible with the Driver impl the same HAL provides to the `embassy-time` crate".

For example:
- I believe that the `critical-section` impl that should be provided by the ESP-IDF HAL should be based on a FreeRTOS mutex (which is equivalent to a STD mutex), even if this mutex cannot be used from an ISR context. This is so because in most of the cases user code is **not** executed in an ISR context. (ISR contexts are reserved for the minority, who has the  "honor" to deal with device drivers, timer queues and executors). So for the regular "user-space" cases, FreeRTOS mutex is the right approach:
  - It does not introduce latencies when handling higher prio tasks or interrupts, as user-space code can be interrupted while holding the mutex
  - It can be locked for arbitrary long intervals
  - Arbitrary user code can be executed in the mutex. Including calls to `libc`, like `malloc`
  - Contrast this with all the restrictions imposed by a "disable-all-interrupts" mutex. I for one am not sure what part of the STD API is safe to call while holding such a mutex!

On the other hand, I also believe that the Driver impl provided by the ESP-IDF HAL **should** be calling back from an ISR. This is because this gives you the lowest latency (we are talking high res timers after all!) *and* we control what is being called - the timer queue and then on top only the executor code which schedules the corresponding task if it was not scheduled already. No user code, no `malloc`s, nothing.

So your suggestion to hard-code the CS impl used in the timer queue puts me between a rock and a hard place:
- I either have to expose a "disable-all-interrupts" CS impl for `critical-section`, with all the consequences for the user
- Or I have to use a less efficient Driver impl, which does not call from an ISR.

Hence why I made it configurable, with some defaults.

Dirbaio: > and we control what is being called

No, you don't. You're calling the `embassy-time` alarm callback, which can can in turn exec arbitrary code from any crate. For example, if:
  - the end user is using the time driver directly (it's public API), or
  - the end user is using the raw executor, the alarm callback calls the `signal_fn` executor callback which is supplied by the user: https://github.com/embassy-rs/embassy/blob/master/embassy-executor/src/raw/mod.rs#L289

And even if the callback executed only `embassy-time` code, I don't think it's `embassy-time`'s responsibility to publicly guarantee "it's OK to call the callback from a context that violates the `critical-section` guarantees"...

> So your suggestion to hard-code the CS impl used in the timer queue puts me between a rock and a hard place:

Yes, I agree it's quite unfortunate... There's a third option: supply a timer queue impl yourself, in that case you *do* control all code that's being called.

embassy-time/src/queue.rs
ivmarkov: With the circular dependencies that are being introduced by using `extern "Rust"` magic, the *implicits might not always work* (that's the topic we discussed on the forum last week). I'll elaborate more on that in a separate yet-to-be-opened-issue, but for now I'm relatively convinced that you are not hitting the linker issue only because you don't have yet a more elaborate crate dependency graph that I do. But the consequences are that - unless we want to change the whole `extern "Rust"` magic approach with something else (does this "else" even exist, I wonder) - in some cases user* should be calling the following macros *explicitly from their binary crate* so as not to run into linking issues:
- `critical_section::set_impl!`
- `embassy_time::driver_impl!`
- Possibly `embassy_time::generic_queue!`

So if we are to expose these macros as a "backup" approach anyway, why not keep the configuration that way as well?

embassy-time/src/queue.rs
ivmarkov: I agree and that would solve the "unprotected `Cell`" issue from your first comment.

Only problem is that the creation of the alarm handle then would be protected by the same potentially "disable-all-interrupts" critical-section that protects the inner queue. Have to see if that's possible on the ESP-IDF, as allocating an alarm there calls a bunch of code, that does `malloc` and whatnot. Let me check though.

embassy-time/src/queue.rs
ivmarkov: +1

embassy-time/src/queue.rs
Dirbaio: if a particular impl needs initializing stuff at runtime, it can still declare a `fn initialize()` itself and ask the user to call it. I don't think we should require it from `embassy-time` itself if most impls won't need it.

embassy-time/src/queue.rs
ivmarkov: > > and we control what is being called
> 
> No, you don't. You're calling the `embassy-time` alarm callback, which can can in turn exec arbitrary code from any crate. For example, if:
> 
> * the end user is using the time driver directly (it's public API), or
> * the end user is using the raw executor, the alarm callback calls the `signal_fn` executor callback which is supplied by the user: https://github.com/embassy-rs/embassy/blob/master/embassy-executor/src/raw/mod.rs#L289
> 
> And even if the callback executed only `embassy-time` code, I don't think it's `embassy-time`'s responsibility to publicly guarantee "it's OK to call the callback from a context that violates the `critical-section` guarantees"...

I think I see where you are going with that. *You are not even talking about the generic queue implementation actually*. You are saying - a `Driver` provider should make sure that users can safely use the critical sections provided by the `critical-section` crate in the user code that the `Driver` calls-back via `set_alarm_callback`, as the Driver API has its uses beyond (a) timer queue. And I now agree to that - the opposite would be very surprising indeed. Can you confirm I'm on the right track?

> Yes, I agree it's quite unfortunate... There's a third option: supply a timer queue impl yourself, in that case you _do_ control all code that's being called.

Except what happens on `waker.wake()` but we can probably assume it is a minimal piece of code that does just task scheduling.
I'll try to expose a minimal public API where the user can take the generic `Queue` impl and instantiate it with their own alarm callbacks that are disconnected from the `Driver` ones. If this proves too unwieldy, oh well, we'll just copy paste.

ivmarkov: Made the changes and then realized that even with a custom timer queue, `embassy-time` still has a hard-dependency on `Driver` because of the `_embassy_time_now` call, which is provided by the driver, and not by the queue.

ivmarkov: @Dirbaio Sorry for pinging but this might be the last question remaining with the new design.
Basically:
- Who should provide the `now()` method?
- a) `Driver` as it still does (+) less changes (-) folks contributing their custom timer queue that is not based on `Driver` alarms still need to contribute a `Driver`, just because of the `now()` method provided by the driver
- b) `TimerQueue` (-) a bit more changes (+) somewhat more symmetric (+) `Driver` impl becomes optional, at least w.r.t. contributing a custom `TimerQueue` impl
- c) A third entity - say - a `Now` trait;
- d) Both `TimerQueue` and `Driver` should have a `now` method, but  `Instant::now` will use `TimerQueue::now` instead of `Driver::now`

embassy-rp/src/timer.rs
```diff
@@ -68,9 +68,16 @@ impl Driver for TimerDriver {
         })
     }
 
-    fn set_alarm(&self, alarm: AlarmHandle, timestamp: u64) {
+    fn set_alarm(&self, alarm: AlarmHandle, timestamp: u64) -> bool {
         let n = alarm.id() as usize;
         critical_section::with(|cs| {
+            let now = self.now();
```
Dirbaio, line 74: I think changing the order here is not correct. If you check `timestamp <= t` first, and set it after, `timestamp` might have elapsed between both, causing the alarm to never fire.

embassy-time/src/queue.rs
ivmarkov: Marking this conversation as resolved. The agreement is to initially go with (a) as it is the simplest variant, and in most cases folks would anyway use either the generic-queue implementation this PR contributes, or the embassy-executor custom one. Both of these require a `Driver` impl anyway.

embassy-time/src/queue.rs
ivmarkov: This is addressed in the latest version of the PR, where the contract of `Driver::set_alarm` is changed.

embassy-rp/src/timer.rs
ivmarkov: I've tried to address your feedback for all the three drivers. 

They are now doing almost exactly what they did before my changes (checks for timestamp-in-the-past are restored to be at their original places in the code and they are disarming the alarm as well. Only difference now is they return `false` rather than calling the callback synchronously). Code is a bit more verbose as disarming the alarm is not happening as a side effect of calling `trigger_alarm` but we are talking a line or two extra, so I hope this is acceptable.

embassy-time/src/queue_generic.rs
```diff
@@ -0,0 +1,453 @@
...
+            alarm: None,
+            alarm_at: Instant::MAX,
+        }
+    }
+
+    fn schedule_wake(&mut self, at: Instant, waker: &Waker) {
+        self.queue
+            .find_mut(|timer| timer.waker.will_wake(waker))
+            .map(|mut timer| {
+                timer.at = at;
```
Dirbaio, line 77: this needs to keep the min of old and new `at`s, not overwrite

embassy-time/src/queue.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Schedules a waker in the queue to be awoken at moment `at`.
+    /// If this moment is in the past, the waker might be awoken immediately.
+    fn schedule_wake(&'static self, at: Instant, waker: &Waker);
+}
+
+/// Set the TimerQueue implementation.
+///
+/// See the module documentation for an example.
+#[macro_export]
+macro_rules! timer_queue_impl {
```
Dirbaio, line 49: could you change embassy-executor to use this macro too? this way the magic name _embassy_time_schedule_wake is contained within this crate for real

embassy-time/src/queue_generic.rs
ivmarkov: If you can elaborate why, that would be great! I can't see why myself.

embassy-time/src/queue.rs
ivmarkov: Makes sense, will do.

embassy-time/src/queue_generic.rs
Dirbaio: A single task is running two different futures with `join`, each waits on a different timer. Fut1 waits for instant 50, fut2 waits for instant 100. When polling the task, `join` polls fut1 then fut2. 

Polling fut2 will overwrite `at=100`, so the task will be polled at instant `100`, but we want it to get polled at 50.

ivmarkov: Agreed! Will fix. 

(I thought for a second we have a bigger problem if a timer future is dropped and a new timer future is created by the same task, but fortunately, each time a timer is polled, it reschedules itself so your "min" would work just fine.)

Dirbaio:
> If you can point me to any additional tests you would like me to run, please let me know.

Tested it on nRF, works fine! Executor changes LGTM too. I guess the only test left now is the scream test :D

Also, did some code size comparisons:

```
nrf blinky example
12224     104    1536   13864    3628 blinky --release
13228     328    1520   15076    3ae4 blinky --release generic-queue-8
 8096     104    1536    9736    2608 blinky --release lto=fat opt-level=z
 9120     328    1520   10968    2ad8 blinky --release lto=fat opt-level=z generic-queue-8
```

+1kb code, +224bytes RAM. Not a bad price to pay for it to be executor-agnostic. 

This is great, thanks a lot for your hard work :) 

bors r+

Dirbaio:
bors cancel

sorry, one more last thing. :see_no_evil: The timer queue ends up in `.data` so it's using both flash and ram, ideally it should be in `.bss`. For this, the `static` intiial value must be all either 0 or uninit bytes. 

```
20000000 g       .data	00000000 __sdata
20000000 g     O .data	00000030 _SEGGER_RTT
20000030 l     O .data	00000038 _ZN11embassy_nrf11time_driver6DRIVER17h7c72554b6f45e707E
20000068 l     O .data	000000e0 _ZN12embassy_time13queue_generic5QUEUE17hddb72db82cb43749E   <<<<<< THIS
20000148 g       .data	00000000 __edata
```

ivmarkov:
> Hmm, gave it a try. Queue is now in `.bss` but code is now +1.5kb... I wonder if SortedLinkedList is too complex?

Might be the case indeed. :(

What about switching to a simple unordered `[Option<Timer>; N]` array and then paying `O(n)` not only when scheduling a timer, but also when the alarm fires as we need to traverse the whole thing to (a) evict the timers which are due and (b) to find the minimal time afterwards for the alarm?

This is all in a critical section which is kind of annoying, but then, perhaps lowering `N` to 32 by default and just doing the brute-force-traverse-end-to-end-unordered might be good enough?

ivmarkov:
On the other hand, I wonder - if you are fine with 1Kb code size, you should probably be fine with 1.5Kb as well, no? :)

Dirbaio:
Changed it to an unsorted `heapless::Vec`. (uses less RAM than `[Option<Timer>; N]`, and ops are `O(len)` instead of `O(n)`)

`dispatch` changes `O(1) -> O(n)`, `schedule_wake` keeps being `O(n)`. Since both have to take a CS I don't think this will make much of a difference in practice..? Both approaches have `O(n)` stuff inside CS's. 

```
12224     104    1536   13864    3628 blinky --release
13228     328    1520   15076    3ae4 blinky --release generic-queue-8
13904     104    1752   15760    3d90 blinky --release generic-queue-8 new
12728     104    1680   14512    38b0 blinky --release generic-queue-8 new vec
 8096     104    1536    9736    2608 blinky --release lto=fat opt-level=z
 9120     328    1520   10968    2ad8 blinky --release lto=fat opt-level=z generic-queue-8
 9688     104    1760   11552    2d20 blinky --release lto=fat opt-level=z generic-queue-8 new
 9484     104    1688   11276    2c0c blinky --release lto=fat opt-level=z generic-queue-8 new vec
```

embassy-time/src/queue_generic.rs
```diff
@@ -96,35 +95,35 @@ impl InnerQueue {
         loop {
             let now = Instant::now();
 
-            while self.queue.peek().filter(|timer| timer.at <= now).is_some() {
-                self.queue.pop().unwrap().waker.wake();
+            let mut next_alarm = Instant::MAX;
+
+            let mut i = 0;
+            while i < self.queue.len() {
```
ivmarkov, line 101: This O(n^2) though.

Dirbaio: is it? `swap_remove` is `O(1)`, it swaps with the last element instead of shifting all after.

ivmarkov: Yep - just checked - sorry!

Dirbaio:
so @ivmarkov if my changes look good to you, let's merge! :D

bors d+

Dirbaio:
I still don't get why it's now +500b without LTO but +1.4kb with... I can't think of any other improvements to try :S

It's still a slight improvement with LTO over SortedLinkedList and uses less RAM, and the simpler code is likely faster.

ivmarkov:
I agree!

PR #961, opened 2022-09-22
Parameterize Signal with RawMutex

ivmarkov:
The `RawMutex` parameter is deliberately chosen to be the second one, so as it can take as a default `CriticalSectionRawMutex`. This way backwards compatibility is preserved, and users utilizing the `critical-section` crate everywhere can just continue to use the more ergonomic single-generic-parameter version of Signal.

I'm thinking we should probably do the same for `Channel`, and move the `RawMutex` parameter as the last one in the list, with a `CriticalSectionRawMutex` being its default. But that's a backwards-incompatible change of course.

### DIFF
--- a/embassy-sync/src/signal.rs
+++ b/embassy-sync/src/signal.rs
@@ -1,8 +1,11 @@
 //! A synchronization primitive for passing the latest value to a task.
-use core::cell::UnsafeCell;
+use core::cell::Cell;
 use core::future::{poll_fn, Future};
 use core::mem;
 use core::task::{Context, Poll, Waker};
+
+use crate::blocking_mutex::raw::{CriticalSectionRawMutex, RawMutex};
+use crate::blocking_mutex::Mutex;
 
 /// Single-slot signaling primitive.
 ///
@@ -28,8 +31,11 @@ use core::task::{Context, Poll, Waker};
 ///
 /// static SOME_SIGNAL: Signal<SomeCommand> = Signal::new();
 /// ```
-pub struct Signal<T> {
-    state: UnsafeCell<State<T>>,
+pub struct Signal<T, R = CriticalSectionRawMutex>
+where
+    R: RawMutex,
+{
+    state: Mutex<R, Cell<State<T>>>,
 }
 
 enum State<T> {
@@ -38,24 +44,27 @@ enum State<T> {
     Signaled(T),
 }
 
-unsafe impl<T: Send> Send for Signal<T> {}
-unsafe impl<T: Send> Sync for Signal<T> {}
-
-impl<T> Signal<T> {
+impl<T, R> Signal<T, R>
+where
+    R: RawMutex,
+{
     /// Create a new `Signal`.
     pub const fn new() -> Self {
         Self {
-            state: UnsafeCell::new(State::None),
+            state: Mutex::new(Cell::new(State::None)),
         }
     }
 }
 
-impl<T: Send> Signal<T> {
+impl<R, T: Send> Signal<T, R>
+where
+    R: RawMutex,
+{
     /// Mark this Signal as signaled.
     pub fn signal(&self, val: T) {
-        critical_section::with(|_| unsafe {
-            let state = &mut *self.state.get();
-            if let State::Waiting(waker) = mem::replace(state, State::Signaled(val)) {
+        self.state.lock(|cell| {
+            let state = cell.replace(State::Signaled(val));
+            if let State::Waiting(waker) = state {
                 waker.wake();
             }
         })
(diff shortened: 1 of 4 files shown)


lulf (approved):
Thanks for fixing this! I guess this replaces #498, and the original issue in that PR no longer applies since Mutex new can be const now.

Some CI build errors, but I think you can merge when fixed.

bors d+

ivmarkov:
> Thanks for fixing this! I guess this replaces #498, and the original issue in that PR no longer applies since Mutex new can be const now.
> 
> Some CI build errors, but I think you can merge when fixed.
> 
> bors d+

Hm, those build errors were not supposed to happen in the first place. Let me look into those...

ivmarkov:
@lulf I'm struggling to understand the existing paper cuts in type inference in the presence of default type parameters, but the net result anyway would be that we'll no longer be able to just type:
`let signal = Signal::new();`.

It should be changed to either
`let signal = Signal::<_>::new();`
or to
`let signal: Signal<_> = Signal::new();` (as done in the [usb_hid_keyboard example fix](https://github.com/ivmarkov/embassy/blob/signal/examples/nrf/src/bin/usb_hid_keyboard.rs#L81)).

Note that the above applies *only* to local variables. In static contexts as well as when your signal is a field member of a `struct` type you anyway have to provide the type `T` of `Signal`, so we don't have an ergonomic regression there, as far as I understand.

So if you believe the benefits of having symmetry with the mpmc Queue and configurability of the mutex type outweigh this disadvantage, I'll push.

lulf:
Maybe test it with a newer nightly version first.

Dirbaio:
It's https://faultlore.com/blah/defaults-affect-inference/

TLDR is `Signal::new()` means "infer both T, R", which fails because R can't be inferred, while `Signal::<_>::new()" means "infer T, use default for R" which works.

ivmarkov:
Thanks for the clarification. So what's the verdict? :) Shall I merge?

embassy-sync/src/signal.rs
```diff
@@ -28,8 +31,11 @@ use core::task::{Context, Poll, Waker};
 ///
 /// static SOME_SIGNAL: Signal<SomeCommand> = Signal::new();
 /// ```
-pub struct Signal<T> {
-    state: UnsafeCell<State<T>>,
+pub struct Signal<T, R = CriticalSectionRawMutex>
+where
+    R: RawMutex,
+{
+    state: Mutex<R, RefCell<State<T>>>,
```
Dirbaio, line 38: I think it should be possible to use `Cell` instead of `RefCell`.

ivmarkov: Let me try.

Dirbaio:
I'm not sure if the "confusingness" of using a default type parameter *and* being inconsistent with the rest of the stuff in embassy-sync is worth it. 
- Backwards-compat is a non-concern since embassy-sync is not released yet (and it's not compatible in all cases due to these inference issues).
- I'm not convinced the CS is the right default. It's what you want for irq->main, or task->task at different prios, but it's not for task->task at same prio, or fut->fut within the same task (with select/join). In my experience I usually write more of the latter, even.

ivmarkov:
> I'm not sure if the "confusingness" of using a default type parameter _and_ being inconsistent with the rest of the stuff in embassy-sync is worth it.
> 
> * Backwards-compat is a non-concern since embassy-sync is not released yet (and it's not compatible in all cases due to these inference issues).
> * I'm not convinced the CS is the right default. It's what you want for irq->task, or task->task at different prios, but it's not for task->task at same prio, or fut->fut within the same task (with select/join). In my experience I usually write more of the latter, even.

I'm fine either way as I'm customizing the mutex in my own use cases.
If you are inclined towards removing the default, shall I also swap the places of the `T` and `R` parameters, so that the code is symmetric with the ordering in mpmc Queue?

Dirbaio:
> If you are inclined towards removing the default, shall I also swap the places of the T and R parameters, so that the code is symmetric with the ordering in mpmc Queue?

Yup!

(I wonder if T should go first in all structs though, since it's the more "important" one...?)

embassy-sync/src/signal.rs
ivmarkov: Yep, works fine.

ivmarkov:
> > If you are inclined towards removing the default, shall I also swap the places of the T and R parameters, so that the code is symmetric with the ordering in mpmc Queue?
> 
> Yup!
> 
> (I wonder if T should go first in all structs though, since it's the more "important" one...?)

As in `Channel<M, T, const N: usize>` becoming `Channel<T, const N: usize, M>`? Yes, seems kind of the right ordering for me too.

But then why not put some default there, for `M` (or `R`, in Signal)? `CriticalSectionRawMutex` might not be the most optimized mutex (I'm with you w.r.t. the argument on signaling within the boundary of just one executor), but then at least for beginners having a default that works in all cases might be useful.

Dirbaio:
> why not put some default there

Because it breaks with a confusing error if you try to do `Channel::new()`

ivmarkov:
> > why not put some default there
> 
> Because it breaks with a confusing error if you try to do `Channel::new()`

Which brings the question how often users would be hit by this given this comment in the `Signal` docu: 
"Signals are generally declared as `static`s and then borrowed as required."

I mean, if you declare it as a static, you *have* to provide the type for `T` anyway, but you can still skip on `R`. Ditto for structs.

ivmarkov:
OK. Maybe let's just *pick* a way if we are unsure which way is the best way? As I mentioned I'm fine either way. So shall I:
1) Leave it as it is (`T`, then `R`) + `CriticalSectionRawMutex` as  the default;
2) Leave it as it is (`T`, then `R`) but remove the default;
3) Remove the default and reorder (i.e. first `R` then `T`)?

(1 and 2 would imply changes to mpmc Queue but perhaps these can be done later.)

One more - perhaps more important - thing: are we deploying fearlessly this stuff without any unit tests? Don't get me wrong, not willing to write these post-factum, just asking if I'm blind.

Dirbaio:
I'd say 3 for now, then later discuss if we do 1 or 2 on all structs?

> are we deploying fearlessly this stuff without any unit tests?

Yes, there's no tests for Signal. It's so simple I wouldn't worry too much, but feel free to write tests if you want.

ivmarkov:
> feel free to write tests if you want.

Not that type of guy. :)

OK, I do 3 and then merge.

PR #963, opened 2022-09-22
Remove some uses of the `futures` crate

Dirbaio:
- use `Future` from `core`
- `poll_fn` is now stable in `core` since Rust 1.64 (out today)
- Use `join` from `embassy-futures`.


Dirbaio:
bors r_

PR #965, opened 2022-09-23
(embassy-rp): add RP2040 ROM functions and intrinsics aliases

MathiasKoch:
Add RP2040 ROM functions described in section **2.8.3.1. Bootrom Functions** of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf

Make all ROM functions (normal and floating point) provide both a direct
call that does the operation and a module with a ptr() function to get
the function pointer.

Add a feature to enable automatic caching of the result of ROM table
function lookups.

Add a check for a V2 bootrom when using floating point functions that
require it.  Panic when it's not present.

Add a standardized macro for intrinsics export and connect the simple
ROM functions to intrinsics.

Direct copy from `rp-hal`! Full credit to those guys for all the heavy lifting.


Dirbaio (approved):
:+1: 

bors r+

PR #966, opened 2022-09-23
(embassy-boot): make magic constants public

MathiasKoch:
To allow writing custom `FirmwareUpdater` implementations without redefining magic constants.

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -7,8 +7,13 @@ mod fmt;
 use embedded_storage::nor_flash::{ErrorType, NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
 use embedded_storage_async::nor_flash::AsyncNorFlash;
 
-const BOOT_MAGIC: u8 = 0xD0;
-const SWAP_MAGIC: u8 = 0xF0;
+/// Magic constant used to indicate to the bootloader, that the active image is
+/// valid and should be booted
+pub const BOOT_MAGIC: u8 = 0xD0;
+
+/// Magic constant used to indicate to the bootloader, that it should swap the
+/// active and DFU images
+pub const SWAP_MAGIC: u8 = 0xF0;
 
 /// A region in flash used by the bootloader.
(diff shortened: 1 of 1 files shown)


lulf:
What are the current limitations of the FirmwareUpdater that makes it necessary to expose these? Can it can be modified to support those use cases rather than exposing these values, because I think it would be better if was handled internally so that one can be sure boot magic is written/read correctly.

MathiasKoch:
My main issue is that I receive my firmware in chunks (streaming OTA), so the current write function would need to split into `prepare_update()` (that does erase), and `write_block()` that can be called N times before calling `mark_updated()`.

Furthermore, I would need a blocking version of those as it stands right now.

If you prefer those changes, I can make those instead?

There would be some concerns around enforcing `prepare_update()` before `write_block()` so I would probably do something like

```rust
impl FirmwareUpdater {
    pub fn prepare_update<'a>(&mut self) -> UpdateWriter<'a> {
        erase();
        UpdateWriter
    }
    
    pub fn write_firmware<F: AsyncNorFlash>(
        &mut self,
        offset: usize,
        data: &[u8],
        flash: &mut F,
        block_size: usize,
    ) -> Result<(), F::Error> {
    
        let mut writer = self.prepare_update();
        
        let mut write_offset = self.dfu.from + offset;
        for chunk in data.chunks(block_size) {
            trace!("Wrote chunk at {}: {:?}", write_offset, chunk);
            writer.write_block(write_offset as u32, chunk).await?;
            write_offset += chunk.len();
        }
    }
}

struct UpdateWriter<'a> { }

impl<'a> UpdateWriter<'a> {
    pub fn write_block() { ... }
}
```

having the public functions in both blocking and async

lulf:
I think making this changes would be perfectly fine, so I would prefer that. I've worked around this in my own code by writing to a 'temporary' page-sized buffer and flushing it once the entire buffer is full, but I think the API you propose is better, so I fully support that, thanks!

MathiasKoch:
Replacing this PR with: https://github.com/embassy-rs/embassy/pull/971

PR #971, opened 2022-09-26
(embassy-boot): add blocking API to FirmwareUpdater

MathiasKoch:
Also add a split `prepare_update` + `write_firmware` API, to allow for an optimized update API at the exchange of added complexity. 
The old API is left in place to allow users to choose the complexity level that fits their needs.

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -604,6 +604,21 @@ impl FirmwareUpdater {
         self.dfu.len()
     }
 
+    /// Obtain the current state.
+    ///
+    /// This is useful to check if the bootloader has just done a swap, in order
+    /// to do verifications and self-tests of the new image before calling
+    /// `mark_booted`.
+    pub async fn get_state<F: AsyncNorFlash>(&mut self, flash: &mut F, aligned: &mut [u8]) -> Result<State, F::Error> {
+        flash.read(self.state.from as u32, aligned).await?;
+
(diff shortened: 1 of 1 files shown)


embassy-boot/boot/src/lib.rs
```diff
@@ -679,7 +688,156 @@ impl FirmwareUpdater {
...
+
+        Ok(())
+    }
+
+    /// Prepare for an incoming DFU update by erasing the entire DFU area and
+    /// returning a `FirmwareWriter`.
+    ///
+    /// Using this instead of `write_firmware` allows for an optimized API in
+    /// exchange for added complexity.
+    pub fn prepare_update<F: NorFlash>(&mut self, flash: &mut F) -> Result<FirmwareWriter, F::Error> {
```
lulf, line 703: From what I can tell, this is identical to prepare_update_blocking, so it should perhaps instead be async and take the async NorFlash trait instead?

lulf (commented):
Looks good, I think one minor change is needed for prepare_update to be async.

embassy-boot/boot/src/lib.rs
MathiasKoch: Huh, nice catch. A bit too fast on the copy-paste :)

MathiasKoch:
@lulf Fixed :+1:

lulf:
Thanks, great work :rocket: 

bors r+

PR #979, opened 2022-09-27
usb: make HALs depend only on embassy-usb-driver.

Dirbaio:
Follow up to #972 

bors r+


guillaume-michel:
I can confirm that this PR fixes the problem I had trying to upgrade embassy to commit after f27a47a37b59bf3b9079f4d4d5f43caf7b7872f8

PR #980, opened 2022-09-27
Experiment: remove Instance generics.

Dirbaio:
This is an experiment of removing "peripheral instance" generics on driver structs. For example, uart is now `Uart<'d, Blocking>` instead of `Uart<'d, UART1, Blocking>`.

Everything is still checked at compile time as before (such as pin functions). The `T` generic is moved from the struct to the `new` methods, not removed entirely.

The main advantage is you can now write code that works with, e.g, both UART0 and UART1 without generics or macros.

```rust
#[embassy_executor::main]
    let config = uart::Config::default();
    let uart0 = uart::Uart::new_with_rtscts_blocking(p.UART0, p.PIN_0, p.PIN_1, p.PIN_3, p.PIN_2, config);
    let uart1 = uart::Uart::new_with_rtscts_blocking(p.UART1, p.PIN_4, p.PIN_5, p.PIN_7, p.PIN_6, config);
    spawner.spawn(hello_task(uart0)).unwrap();
    spawner.spawn(hello_task(uart1)).unwrap();
}

#[embassy_executor::task(pool_size = 2)]
async fn hello_task(mut uart: Uart<'static, Blocking>) {
    uart.blocking_write("Hello World!\r\n".as_bytes()).unwrap();

    loop {
        uart.blocking_write("hello there!\r\n".as_bytes()).unwrap();
        Timer::after(Duration::from_secs(1)).await;
    }
}
```

### DIFF
--- a/embassy-hal-common/src/atomic_ring_buffer.rs
+++ b/embassy-hal-common/src/atomic_ring_buffer.rs
@@ -1,0 +1,154 @@
+use core::slice;
+use core::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
+
+pub struct RingBuffer {
+    buf: AtomicPtr<u8>,
+    len: AtomicUsize,
+    start: AtomicUsize,
+    end: AtomicUsize,
+}
+
+pub struct Reader<'a>(&'a RingBuffer);
+pub struct Writer<'a>(&'a RingBuffer);
+
+impl RingBuffer {
+    pub const fn new() -> Self {
+        Self {
+            buf: AtomicPtr::new(core::ptr::null_mut()),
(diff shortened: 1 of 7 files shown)


Dirbaio:
code size comparison:

```
  14656      48    1552   16256    3f80 old, uart0
  14720      48    1568   16336    3fd0 new, uart0: +64 bytes
  16312      48    1616   17976    4638 old, uart0+uart1 (copypaste)
  15032      48    1656   16736    4160 new, uart0+uart1: -1280 bytes
```

Dirbaio:
this is hopelessly outdated. Needs redoing, probably copying the final design we ended up using for stm32.

PR #981, opened 2022-09-28
(embassy-stm32): add E-H1 uart blocking & nb implementation

MathiasKoch:


embassy-stm32/src/usart/mod.rs
```diff
@@ -359,6 +359,79 @@ mod eh1 {
     impl<'d, T: BasicInstance, RxDma> embedded_hal_1::serial::ErrorType for UartRx<'d, T, RxDma> {
         type Error = Error;
     }
+
+    
+    impl<'d, T: BasicInstance, RxDma> embedded_hal_1::serial::nb::Read for UartRx<'d, T, RxDma> {
+        fn read(&mut self) -> nb::Result<u8, Self::Error> {
```
Dirbaio, line 365: could this call the `eh02` code, to not copypaste? or better, make it an inherent private fn and call it from trait impls, this way we don't have any logic in the trait impls.

Dirbaio (approved):
Thank you!

bors r+

Dirbaio:
needs updating to `alpha.9`

bors d+

PR #978, opened 2022-09-27
(embassy-rp): async i2c master implementation

MathiasKoch:
Build on top of #914 to add the async implementation of I2C

PR #984, opened 2022-09-28
rp pico async i2c implementation

jsgf:
This implements an interrupt-driven async i2c master. It is based on https://github.com/embassy-rs/embassy/pull/914, a bit of https://github.com/embassy-rs/embassy/pull/978 and @ithinuel's https://github.com/ithinuel/rp2040-async-i2c.git

This is still work-in-progress, and is currently untested.

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -406,9 +406,9 @@ impl<'d, T: Instance> I2c<'d, T, Async> {
         self.read_async_internal(buffer, false, true).await
     }
 
-    pub async fn write_async(&mut self, addr: u16, buffer: &[u8]) -> Result<(), Error> {
+    pub async fn write_async(&mut self, addr: u16, bytes : impl IntoIterator<Item = u8>) -> Result<(), Error> {
         Self::setup(addr)?;
-        self.write_async_internal(buffer.iter().copied(), true).await
+        self.write_async_internal(bytes, true).await
     }
 }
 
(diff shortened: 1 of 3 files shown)


MathiasKoch:
This is great to see!

One thing springs to mind already. Why would you limit the blocking implementations to `Mode = Blocking` ? That seems like an unnecessary limitation to me?

jsgf:
@MathiasKoch 

Because I was concerned that it would be a footgun - if you started an async operation and then also did a blocking operation while it is outstanding, then you'd screw up the hardware state for the async operation.

But maybe that's moot and the lifetimes prevent you from having access to the I2c object while the future is still pending - it's easy to relax again if it seems useful.

I guess the other question is why you would want blocking if you have a non-blocking context anyway?

jsgf:
OK, this is working as much as I've tested it, so I think its good for review.

It still disallows blocking IO on an async I2c - I'm happy to reverse that if there's a reason, but I don't personally see one.

kalkyl:
Can confirm it's working great on hardware over here aswell!

Dirbaio:
> It still disallows blocking IO on an async I2c - I'm happy to reverse that if there's a reason, but I don't personally see one.

Yes, it should be allowed. All drivers on embassy HALs allow mixing blocking+async like this. 

(One reason why it's wanted is it allows using a blocking driver and an async driver on the same i2c at the same time (for 2 different devices on the same i2c bus), without having to recreate the bus.)

jsgf:
Updated to allow blocking ops on Async I2c

jsgf:
More updates:
- tweak tx abort error handling to make it consistent with other impls
- trigger tx fifo refill a little earlier to stop bus stalls

jsgf:
ping @Dirbaio - this is ready for review (or can someone else review? @MathiasKoch @ithinuel ?)

Dirbaio (approved):
Thank you!

bors r+

PR #985, opened 2022-09-28
Create Sx126X LORA driver

ceekdee:
Implementation features:

- update embassy-lora to support Semtech SX126X chips, specifically the RAK4631 chip (nrf52480 and sx1262).

- support additional SX126X packages by adding a feature (reference feature rak4631) and updating the board specific Rust file.  To enable feature rak4631, settings.json must currently enable "rust-analyzer.linkedProjects" for "examples/nrf/Cargo.toml".

- provide tx/rx examples in examples/nrf to show compatibility with the interface provided by the SX127X LORA implementation.

Only LORA P2P communication has been tested.  Implementation lines marked with ??? indicate areas for further investigation.  Furthermore, I question whether the DIO1 handler is adequate for catching all interrupt sequences.

This implementation is patterned after the C/C++ implementation provided by Semtech, available through the RAK-nRF52-RUI developers platform.


embassy-lora/src/lib.rs
```diff
@@ -7,6 +7,8 @@ pub(crate) mod fmt;
 
 #[cfg(feature = "stm32wl")]
 pub mod stm32wl;
+#[cfg(feature = "rak4631")]
```
lulf, line 10: Feature should be named sx126x instead I think.

lulf (commented):
I had a quick look, and a couple of questions:

* From my very quick glance the sx127x and sx126x follow a similar structure. Do you think it's possible/impossible to change the sx127x into 'sx12xx'?
* For the board-specific antenna switches, maybe it could make use of a RadioSwitch trait/type so users can reconfigure the sx126x to match their board?

embassy-lora/src/lib.rs
ceekdee: - Creating sx12xx : 1) Semtech does not appear to have attempted consolidating the drivers for sx126x and sx127x; 2) Is it worth the migration pain for existing users of the sx127x driver?  I imagine developers will stick with the sx127x for existing products, but going forward will choose sx126x for improved battery performance, if given the option.  That said, https://github.com/Lora-net/LoRaMac-node/tree/master/src/radio would be a good place to evaluate the differences to be overcome in an sx12xx implementation.  I'll take a look.
- RadioSwitch:  Will look into it.  Thanks!
- The rak4631 feature was intended to indicate the flavor of sx126x package so that the correct implementation is used in board_specific.rs.  I will evaluate whether a new feature is needed to identify sx126x separately.

One thing I failed to mention in the initial post is that the GFSK aspect of the sx126x chip is not implemented in the current driver.  If a GFSK driver is needed, it seems like it should be implemented in a GFSK-specific portion of the Embassy code-base as a separate concern.

I greatly appreciate having other sets of eyes look at this.

lulf: Thanks for looking into this!

1) The reason for my question was partially because I've used https://github.com/lthiery/rust-sx12xx/ wrapper before, and because it seems the chips have lot of registers in common, and lots of the settings are similar. But keeping them separate is also fine if you think that's better overall.
2) Since this is early stage and not even released on crates.io, we don't give any API guarantees just yet. Given the main purpose of embassy-lora to integrate with the lorawan-device crate, the API surface used in practice by users are ::new() to create the driver before wrapping it in lorawan-device. 

I've ignored GFSK in the sx127x driver, but I think there are some remains from the original driver it was based off. I think we can ignore/remove the GFSK part, as it is probably better served by the original sx127x driver anyway, but if you want to keep it that's fine too.

ceekdee: As shown in Semtech's LORA repository (https://github.com/Lora-net/LoRaMac-node/tree/master/src/radio), the four different radio implementations (lr1110, sx126x, sx1272, and sx1276) mostly support the same LORA radio API.  It gets a little muddied in the sx1272 and sx1276 implementations, but the API is clear in the lr1110 and sx126x implementations:

Init
GetStatus
SetModem
SetChannel
IsChannelFree
Random
SetRxConfig
SetTxConfig
CheckRfFrequency
TimeOnAir
Send
Sleep
Standby
Rx
StartCad
SetTxContinuousWave
Rssi
Write
Read
WriteBuffer
ReadBuffer
SetMaxPayloadLength
SetPublicNetwork
GetWakeupTime
IrqProcess

Optional depending on radio:
RxBoosted,
SetRxDutyCycle

- In the implementation being evaluated here, embassy-lora/src/sx126x/sx126x_lora/mod.rs supports this interface, with a couple of exceptions related to GFSK (CheckRfFrequency and SetPublicNetwork).
- The file embassy-lora/src/sx126x/mod.rs provides the PhyRxTx trait and other plumbing to support existing Embassy LORAWAN functionality.

If we are to move forward with an sx12xx implementation and believe (as Semtech apparently does) that the LORA API above identifies a solid interface to LORA PHY functionality, the current Embassy sx127x implementation would need significant refactoring.  It is certainly feasible, but I am not sure if the implementation effort is warranted since all current radios support the PhyRxTx interface.

ceekdee: I am unclear on the intended use of the RadioSwitch trait.  It would be great if some examples of potential use can be provided, so that I can determine how the current antenna handling can be improved.

On a related note, I just realized that output pins can also be degraded, as I did for input pins.  That might allow me to significantly lessen the generic jumbles like:

impl<SPI, CS, RESET, ANTRX, ANTTX, WAIT, BUS> Sx126xRadio<SPI, CS, RESET, ANTRX, ANTTX, WAIT, BUS>

CS, RESET, ANTRX, and WAIT can be condensed to one generic.  Let me know if I am generically jumbled myself.

lulf: Agreed on letting the sx127x implementation stay as is for now maybe, but maybe someone who's willing can merge them in the future. Speaking of that, the existing STM32WL SubGhz radio, is actually a Sx126x as well (https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/subghz/mod.rs), but with a 'special' SPI bus. Probably not worth merging that either, but it might serve as an implementation guide at least :)

On radio switch: By the looks of it, the ANTRX and ANTTX would be part of the RadioSwitch implementation, like this: https://github.com/embassy-rs/embassy/blob/master/examples/stm32wl/src/bin/lorawan.rs#L19

In the case of the sx127x, a board based on RAK811 needs a RadioSwitch implementation different to a board like the STM32 L0 Lora Discovery. But the driver stays the same, and does not need to provide feature flags for different boards.

The RadioSwitch is used in the embassy-lora to configure the antenna pins for TX or RX or sleep, like this https://github.com/embassy-rs/embassy/blob/master/embassy-lora/src/stm32wl/mod.rs#L134

ceekdee: I agree that it would be beneficial to put all LORA radio physical layer processing under one architectural structure, regardless whether the driver for the radio is stm or nrf.  Furthermore, it would be good to provide a configuration mechanism - maybe features, maybe not - to associate generic LORA physical layer processing with specific processing for each type of Semtech chip.

The sx126x implementation provides a generic API, a set of generic subroutines, and a file to hold board-specific functionality chosen by feature.  I believe the generic API (listed in a previous comment) is a decent starting point, but there are multiple ways to bind specific functionality to this generic API, based on stm/nrf chip and Semtech chip.

I will be unable to do much additional LORA implementation until December; however, I would be interested in implementing a more cohesive LORA physical layer across chips.  If you would like to lay out an overall architecture in the interim, I would be willing to implement it.  My biggest deficiency is that I don't have the equipment to test chip combinations.  But I believe that with your guidance I can get the implementation close enough that we can bring it to fruition.

In the meantime, I will incorporate RadioSwitch processing for antennas and look at simplifying my use of generics.  I assume it is still reasonable to get the current sx126x implementation into the pot so folks can start turning up the heat on it.

embassy-lora/Cargo.toml
```diff
@@ -9,13 +9,15 @@ src_base = "https://github.com/embassy-rs/embassy/blob/embassy-lora-v$VERSION/em
 src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-lora/src/"
 features = ["time", "defmt"]
 flavors = [
+    { name = "rak4631", target = "thumbv7em-none-eabihf",   features = ["rak4631"] },
```
lulf, line 12: Could you rename this feature to sx126x ?

embassy-lora/src/lib.rs
lulf, line 10: Same rename here since it's not really board-specific.

examples/nrf/Cargo.toml
```diff
@@ -17,6 +18,10 @@ embassy-nrf = { version = "0.1.0", path = "../../embassy-nrf", features = ["defm
 embassy-net = { version = "0.1.0", path = "../../embassy-net", features = ["defmt", "tcp", "dhcpv4", "medium-ethernet", "pool-16"], optional = true }
 embassy-usb = { version = "0.1.0", path = "../../embassy-usb", features = ["defmt"], optional = true }
 embedded-io = "0.3.0"
+embassy-lora = { version = "0.1.0", path = "../../embassy-lora", features = ["rak4631", "time", "defmt"], optional = true }
```
lulf, line 21: Same

lulf (commented):
I think it looks good and ready to be merged, but the feature should be renamed to reflect this is not really a rak-specific driver.

Great work, thank you for the contribution 🚀

embassy-lora/src/lib.rs
lulf: I would prefer if there was nothing board-specific, but if you're committed to improve that, I think we can rename the feature sx126x and fix the implementation to not be board-specific. 

I don't want board-specific feature flags to creep into embassy-lora unless it can be avoided, and if we add those flags it can start off in the wrong direction if people start adding more feature flags for different boards.

embassy-lora/Cargo.toml
ceekdee: Implementation updated to use feature sx126x  and remove rak4631 feature usage in the board specific processing.  I ran into puzzling issues attempting to implement the RadioSwitch functionality, so updated the board specific processing to include antenna handling for the time being.  This will get shaken out when we implement sx12xx processing.

A correction was made to RSSI computation, but that may need more work.  Implementation marked with ??? remains to be investigated.

The two P2P examples have been simplified to support a wider range of nrf/sx126x combinations.

I will be available through October to address discovered issues, but will be unavailable for much of November.  I look forward to working toward an sx12xx solution starting in December.  Thank you for reviewing this submission.

lulf (approved):
LGTM, thanks for working on it! Let's merge it.

bors r+

PR #986, opened 2022-09-29
(embassy-stm32): implement RTC peripheral for all stm32 families with rtc

MathiasKoch:
This is only tested with `rtc_v2l4`, except for it passing the cargo-batch ci.

I feel like there is a lot of code duplication here, and for some of the families they could share the same implementation file for this PR, but I think that will change once alarms are added to the mix, so I kept them separate for now based on the various `rtc_vxxx` configurations.


MathiasKoch:
@Dirbaio Just attempting to revisit this PR, but it is a fucking nightmare to feature-gate properly!

Any pointers on how to improve my quality of life on this? Would it be faster/more correct for me to add the proper enums to stm32-data? And if so, can you give me a quick rundown on how to do it properly?
It seems like most of it is missing enums in RCC, and some enums that vary between being located in `stm32_metapac::rtc::vals::*` or stm32_metapac::rcc::vals::*` for different families?

Also, stuff like some chips having `pac::PWR` as a raw pointer, and others as a struct?


It also seems impossible, or at least very hard for me to add families/chips incrementally, rather than this big-bang all at once PR?

Dirbaio:
Thanks for your work @MathiasKoch :) Sorry I haven't been able to help, I'm swamped with work and PRs :(

Closing in favor of #1371

MathiasKoch:
Awesome 🤩 Just happy to see it finalized! ❤️

PR #987, opened 2022-09-29
(embassy-stm32): uart flowcontrol

MathiasKoch:
Add RTS & CTS flow control to stm32 UARTs


embassy-stm32/src/usart/mod.rs
```diff
@@ -241,6 +240,29 @@ impl<'d, T: BasicInstance, TxDma, RxDma> Uart<'d, T, TxDma, RxDma> {
...
+        tx_dma: impl Peripheral<P = TxDma> + 'd,
+        rx_dma: impl Peripheral<P = RxDma> + 'd,
+        config: Config,
+    ) -> Self {
+        into_ref!(cts, rts);
+
+        unsafe {
+            rts.set_as_af(rts.af_num(), AFType::OutputPushPull);
+            cts.set_as_af(cts.af_num(), AFType::Input);
+            T::regs().cr3().write(|w| {
```
Dirbaio, line 258: Are you sure this works? `new()` is doing RCC reset+enable *after* this write, so it looks a bit suspicious.

MathiasKoch: Hmm, I am not sure that I see how RCC reset and enable alters the UART control regs?

MathiasKoch: I will be able to verify this on monday though.

Dirbaio: Writes to peripheral regs simply don't work if the peripehral is not rcc-enabled, and the rcc-reset clears all the registers.

MathiasKoch:
@Dirbaio Took a long time for me to get around to resolving this, but it should be good to go now.

Dirbaio (approved):
Cool, thank you! 

bors r+

PR #988, opened 2022-09-29
Update embedded-hal versions and explicitly pin

lulf:
Pinning to ensure CI don't accidentally break.


huntc (approved):

huntc:
Thanks for attending to this so quickly!

PR #989, opened 2022-09-29
Re-instate direction discriminant values

huntc:
The previous commit of f4f58249722bc656a13865e06535d208440c3e4a had removed explicit discriminant values from the direction enums, causing USB serial to fail as `0x80` is expected for the inbound direction. A number of places cast a variant into a u8 via "as u8". To my knowledge, there's no way of preventing these numeric casts to occur, so we need to reinstate the explicit discriminant values.

Tested on the nRF USB examples.

### DIFF
--- a/embassy-usb/src/control.rs
+++ b/embassy-usb/src/control.rs
@@ -106,7 +106,7 @@ impl Request {
         let recipient = rt & 0b11111;
 
         Request {
-            direction: if rt & 0x80 == 0 { Direction::Out } else { Direction::In },
+            direction: rt.into(),
             request_type: unsafe { mem::transmute((rt >> 5) & 0b11) },
             recipient: if recipient <= 3 {
                 unsafe { mem::transmute(recipient) }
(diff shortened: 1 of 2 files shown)


embassy-usb/src/control.rs
```diff
@@ -106,7 +106,7 @@ impl Request {
         let recipient = rt & 0b11111;
 
         Request {
-            direction: if rt & 0x80 == 0 { Direction::Out } else { Direction::In },
+            direction: rt.into(),
```
huntc, line 109: I think it makes sense to keep the conversions local, but I have a separate commit for this in case the previous is preferred.

alexmoon:
I think if the intent is to allow casting the enum to its discriminant value, the `#[repr(u8)]` attribute should be restored as well.

However, I wonder if the better fix would be to address the places where the enum is being cast? Maybe an inherent `fn to_u8` or similar?

Dirbaio:
I removed the discriminant values because they feel like "implementation details" (the fact that the direction is in bit 7 in some places is an implementation detail), so I wanted to keep that out of the `-driver` crate. My intent was to remove all the `as u8` casts, seems like I forgot some, oops.

Dirbaio:
Fixed in #991

PR #990, opened 2022-09-29
Small pubsub improvements

diondokter:
- Futures in pub & sub are now awaited instead of returned
- Added functions for reading how many messages are available

This helps people get better compiler diagnostics. For example, I forgot to call await on a future and the compiler didn't complain.
This also helps with making some decisions based on the state of the channels.

### DIFF
--- a/embassy-sync/src/pubsub/mod.rs
+++ b/embassy-sync/src/pubsub/mod.rs
@@ -562,7 +562,7 @@ mod tests {
     async fn correct_available() {
         let channel = PubSubChannel::<NoopRawMutex, u32, 4, 4, 4>::new();
 
-        let mut sub0 = channel.subscriber().unwrap();
+        let sub0 = channel.subscriber().unwrap();
         let mut sub1 = channel.subscriber().unwrap();
         let pub0 = channel.publisher().unwrap();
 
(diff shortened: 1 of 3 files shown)


alexmoon:
I would personally prefer to have nameable types for futures where feasible. Could your issue be addressed with a `#[must_use]` attribute instead?

diondokter:
> I would personally prefer to have nameable types for futures where feasible. Could your issue be addressed with a `#[must_use]` attribute instead?

Yes, that is possible. I'll change it.
Can you explain that preference? I've gone back and forth about this myself

alexmoon:
> Yes, that is possible. I'll change it. Can you explain that preference? I've gone back and forth about this myself

It's not a very strong preference. There are just occasions where being able to name the type is useful (e.g. storing in a static, using a turbofish on a call, etc). In general a nameable type is strictly more useful than an unnameable type. I don't think it's usually worth the extra trouble to create a nameable type when an async fn has actual logic in it, but when it's just awaiting a struct it doesn't cost anything to return the nameable type instead.

Dirbaio (approved):
LGTM, thank you! 

bors r+

PR #992, opened 2022-09-30
(embassy-stm32): remove flash lock/unlock public API from stm32 flash

MathiasKoch:
Instead, perform the unlocking and locking automatically on erase and write operations.

This makes the `embedded-storage` abstraction actually useable in libraries, while still keeping the flash peripheral locked the majority of the time.


lulf (approved):
Thank you 🚀 

bors r+

Issue #1002, opened 2022-10-07
License statements missing in crates


chrysn:
Many crates (eg. embassy-usb) do not have a machine readable `license = "..."` statement. (They're probably the ones not yet on crates.io, for crates.io requires some information to be present around there). From the repository structure it appears like the licensing intention is "Apache-2.0 OR MIT" for the whole project.

Is that correct?

If so, please consider stating that explicitly in the crates. (If it's just about the editing work, I can provide a PR to that effect, but I'd only do that on invitation, for it's a statement I'd be making in my branch already).

In terms of use case, this is motivated by embassy crates having raised flags on a `cargo deny` license check that would otherwise just pass.

Dirbaio:
Yes, it's all Apache+MIT. 

we can add it to all crates even if not realeased on crates.io yet, yes. PR welcome :)

PR #1003, opened 2022-10-07
all Cargo.toml: Add license to all crate Cargo.toml files

chrysn:
This sets the license to "MIT OR Apache-2.0" in a machine readable form on all crates, as it was already in human readable form in the README.md file, and reaffirmed in #1002.

(The statements on all the individual examples might not be strictly essential for the `cargo deny` use case as they are leaf crates, but other tools might use that information).


lulf (approved):
LGTM, thanks! 🚀 

bors r+

PR #1004, opened 2022-10-07
Fix internal channels for adc v2

chemicstry:
Internal channel reading was broken on adc_v2, because `Adc::read()` requires gpio pin trait, which was not implemented by `VrefInt`, `Temperature`, `Vbat`. The required configuration bits `tsvrefe`, `vbate` were not enabled either. This PR makes it a bit closer to how adc_v4 works.

While at it, I also changed adc_v2 to use `RccPeripheral` instead of permanently enabling all ADCs.

### DIFF
--- a/embassy-stm32/src/adc/mod.rs
+++ b/embassy-stm32/src/adc/mod.rs
@@ -30,9 +30,9 @@ pub(crate) mod sealed {
     }
 }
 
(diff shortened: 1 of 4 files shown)


lulf (approved):
Thanks for fixing it 🚀 

bors r+

PR #1005, opened 2022-10-08
Expose Pin::pin() and Pin::bank() as public

mkj:


lulf (approved):
Thank you! 🚀 

bors r+

PR #1006, opened 2022-10-09
Removes some of the code duplication for UarteWithIdle

huntc:
This PR removes some of the code duplications for `UarteWithIdle` at the slight expense of requiring a split when using idle processing. As the nRF example illustrates though given the LoC removed, this expense seems worth the benefit in terms of maintenance, and the avoidance of copying over methods. My main motivation for this PR was actually due to the `event_endtx` method not having been copied across to the idle-related code.

Tested the uart_idle example on my nRF52840-dk, and from within my app. Both appear to work fine.

### DIFF
--- a/embassy-nrf/src/uarte.rs
+++ b/embassy-nrf/src/uarte.rs
@@ -173,6 +173,61 @@ impl<'d, T: Instance> Uarte<'d, T> {
         (self.tx, self.rx)
     }
 
+    /// Split the Uarte into a transmitter and receiver that will
+    /// return on idle, which is determined as the time it takes
+    /// for two bytes to be received.
+    pub fn split_with_idle<U: TimerInstance>(
+        self,
+        timer: impl Peripheral<P = U> + 'd,
+        ppi_ch1: impl Peripheral<P = impl ConfigurableChannel + 'd> + 'd,
+        ppi_ch2: impl Peripheral<P = impl ConfigurableChannel + 'd> + 'd,
+    ) -> (UarteTx<'d, T>, UarteRxWithIdle<'d, T, U>) {
+        let mut timer = Timer::new(timer);
+
+        into_ref!(ppi_ch1, ppi_ch2);
+
+        let r = T::regs();
+
+        // BAUDRATE register values are `baudrate * 2^32 / 16000000`
+        // source: https://devzone.nordicsemi.com/f/nordic-q-a/391/uart-baudrate-register-values
+        //
+        // We want to stop RX if line is idle for 2 bytes worth of time
+        // That is 20 bits (each byte is 1 start bit + 8 data bits + 1 stop bit)
+        // This gives us the amount of 16M ticks for 20 bits.
+        let baudrate = r.baudrate.read().baudrate().variant().unwrap();
+        let timeout = 0x8000_0000 / (baudrate as u32 / 40);
+
+        timer.set_frequency(Frequency::F16MHz);
+        timer.cc(0).write(timeout);
+        timer.cc(0).short_compare_clear();
+        timer.cc(0).short_compare_stop();
+
+        let mut ppi_ch1 = Ppi::new_one_to_two(
+            ppi_ch1.map_into(),
+            Event::from_reg(&r.events_rxdrdy),
+            timer.task_clear(),
+            timer.task_start(),
+        );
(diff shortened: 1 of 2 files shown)


huntc:
cc @lulf @ZoeyR Given previous interest.

embassy-nrf/src/uarte.rs
```diff
@@ -929,6 +762,75 @@ impl<'d, U: Instance, T: TimerInstance> UarteRxWithIdle<'d, U, T> {
         Ok(n)
     }
 }
+
```
huntc, line 765: Nothing changed in the following two functions - just moved to help the code layout a bit.

embassy-nrf/src/uarte.rs
```diff
@@ -173,6 +173,61 @@ impl<'d, T: Instance> Uarte<'d, T> {
...
+        let r = T::regs();
+
+        // BAUDRATE register values are `baudrate * 2^32 / 16000000`
+        // source: https://devzone.nordicsemi.com/f/nordic-q-a/391/uart-baudrate-register-values
+        //
+        // We want to stop RX if line is idle for 2 bytes worth of time
+        // That is 20 bits (each byte is 1 start bit + 8 data bits + 1 stop bit)
+        // This gives us the amount of 16M ticks for 20 bits.
+        let baudrate = r.baudrate.read().baudrate().variant().unwrap();
```
huntc, line 197: Please double-check this - it appears correct to me of course. My aim here is to avoid storing the baud rate again i.e. just pull it out of the register. It should also have been assigned earlier so it is a bug if we get a None.

embassy-nrf/src/uarte.rs
```diff
@@ -173,6 +173,61 @@ impl<'d, T: Instance> Uarte<'d, T> {
         (self.tx, self.rx)
     }
 
+    /// Split the Uarte into a transmitter and receiver that will
```
huntc, line 176: The following implementation was mostly taken from what was the constructor for `UarteWithIdle`.

embassy-nrf/src/uarte.rs
lulf: It appears correct o me too, I cannot find a code path where split can be called with an uninitialized baudrate register.

embassy-nrf/src/uarte.rs
```diff
@@ -1126,26 +1012,4 @@ mod eha {
...
-        fn write<'a>(&'a mut self, buffer: &'a [u8]) -> Self::WriteFuture<'a> {
-            self.write(buffer)
-        }
-
-        type FlushFuture<'a> = impl Future<Output = Result<(), Self::Error>> + 'a where Self: 'a;
-
-        fn flush<'a>(&'a mut self) -> Self::FlushFuture<'a> {
-            async move { Ok(()) }
-        }
-    }
```
lulf, line 1150: Shouldn't these implementations be kept for UarteWithIdle?

lulf (commented):
It looks good in general, I have one question wrt. the e-h implementations.

embassy-nrf/src/uarte.rs
huntc: UarteWithIdle is gone. :-)

lulf: I meant UarteRxWithIdle

huntc: Ah. Well, UarteWithIdleRx was previously there and didn’t have these e h implementations. I don’t think I’ve taken anything away.

huntc: We could add it though. It was probably an omission from before.

lulf: Ok, I guess that can be fixed later. I think it can be merged, thanks for cleaning it up!

lulf (approved):

Dirbaio (approved):
bors r+

I'd appreciate it if in future PRs you separated the "substance" changes (removing UarteWithIdle) from the "boring" stuff (renaming `T/U`, moving the workaround fns). Ideally in separate PRs, but at least in separate commits. It'd help review a lot. Thanks!

huntc:
Sorry if I made it difficult to review. I shall endeavour to make it easier in future.

PR #1012, opened 2022-10-17
rp i2c: have separate wakers for each i2c unit

jsgf:
If they both share one waker, there's the possibility that some wakeups could get lost.


Dirbaio:
can you use a per-instance `State` struct instead, like this? 
https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/i2c/mod.rs#L49
https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/i2c/v2.rs#L135

The advantage is since each is a separate `static`, the compiler can remove the state for unused i2c instances. With a single static it can't.

PR #1014, opened 2022-10-18
Add memory barriers to H7 flash driver to mitigate PGSERR errors

matoushybl:
The stm32h7xx-hal uses only the ordering barrier, while the CubeMX uses the DSB and ISB instructions, to be on the safe side, both are used here.

Without the barrier, the PG bit is not set, when the writes are being done, resulting in an error.


lulf (approved):
Good catch :rocket: 

bors r+

PR #1015, opened 2022-10-18
Enable defmt in embassy-hal-common when defmt is enabled in stm32 HAL

matoushybl:


Dirbaio (approved):
thNKS!

BORS R+

Dirbaio:
oops caps lock

bors r+

PR #1016, opened 2022-10-19
stm32/dma: Store wakers in "associated statics"

GrantM11235:
I'm sure that I'm not the first person to use this technique, but I don't think I have seen it anywhere else.

Instead of storing all the wakers in a single static array, each channel singleton gets an associated function that returns a static reference to its waker
```rust
fn waker() -> &'static AtomicWaker {
    static WAKER: AtomicWaker = AtomicWaker::new();
    &WAKER
}
```

Another alternative would be to use the `foreach_dma_channel!` macro to generate normal statics like this:
```rust
static CH1_WAKER: AtomicWaker = AtomicWaker::new();
static CH2_WAKER: AtomicWaker = AtomicWaker::new();
// etc...
```
But I think that it is cleaner to tie the statics directly to the channel singletons.

Based on my very limited testing, this is also a small code-size win. I checked the stm32f1 adc example in debug (with `opt-level = "s"`) and release. `text` is 86 bytes smaller in debug and 64 bytes smaller in release, `data` and `bss` were unchanged.

### Future possibilities
If unused interrupt handlers could be optimized away, the unused statics should be optimized away too. I have an idea for this, it involves dynamically assigning interrupt handlers at runtime.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -602,29 +602,15 @@ fn main() {
         }
     }
 
-    let mut dma_channel_count: usize = 0;
-    let mut bdma_channel_count: usize = 0;
-    let mut gpdma_channel_count: usize = 0;
-
     for ch in METADATA.dma_channels {
         let mut row = Vec::new();
         let dma_peri = METADATA.peripherals.iter().find(|p| p.name == ch.dma).unwrap();
         let bi = dma_peri.registers.as_ref().unwrap();
 
-        let num;
         match bi.kind {
-            "dma" => {
-                num = dma_channel_count;
-                dma_channel_count += 1;
-            }
-            "bdma" => {
-                num = bdma_channel_count;
-                bdma_channel_count += 1;
-            }
-            "gpdma" => {
-                num = gpdma_channel_count;
-                gpdma_channel_count += 1;
-            }
+            "dma" => {}
+            "bdma" => {}
+            "gpdma" => {}
             _ => panic!("bad dma channel kind {}", bi.kind),
         }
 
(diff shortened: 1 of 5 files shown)


guillaume-michel:
Looking forward for this change

lulf (approved):
I think this is a nice improvement, easier to follow as well, thanks!

Dirbaio:
This will prevent implementing `AnyChannel` in the future, which I think we do want.

 `embassy-rp` already [has it](https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/dma.rs#L241). One of the biggest advantages is it allows removing all instance/channel generics from drivers (#980) which is a big win in simplicity, and in code size if you use multiple instances in the same program.

Dirbaio:
Actually it's still possible to apply this idea to `AnyChannel`: making it store `number: u8, waker: &'static AtomicWaker`. It'll take 8 bytes intead of 1 byte however, so I'm not sure if it's a win...

GrantM11235:
Another way to implement `AnyChannel` would be with a macro generated match statement like this:
```rust
impl AnyChannel {
    fn waker(&self) -> &'static AtomicWaker {
        match self.num {
            1 => Ch1::waker(),
            // etc...
            _ => unreachable!(),
        }
    }
}
```

This keeps the size of `AnyChannel` to one byte, but it it wouldn't allow the unused wakers to be optimized away. I'm not sure which is the optimal solution, but this PR works with either, and it would be an internal implementation detail that we can change whenever we want.

Dirbaio:
That'd produce somewhat big code, and would prevent optimizing away wakers for unused channels (which I believe was one of the advantages of the original approach over the array?)

GrantM11235:
This is very stale. I might revisit it at some point, but for now I will close it.

PR #1021, opened 2022-10-21
rp usb: wait for accept() completion

mkj:
This fixes failures when `defmt` feature isn't enabled (timing related).


Dirbaio (approved):
Oooh, nice, you found it! Thank you for taking the time for investigating it :) 

bors r+

PR #1022, opened 2022-10-22
Implement I2C timeout

chemicstry:
This PR implements timeouts for all I2C methods.

Without timeouts, there are cases where I2C peripheral can be stuck in an infinite loop (i.e. SCL line is shorted to ground). This allows handling such errors without freezing entire application.

In case anyone is wondering, the `TIMEOUT` flag in `SR1` register is only available in SMBus mode, hence for regular I2C timeouts must be handled in software.

I'm not sure about the required `embassy-time` dependency, though. `embassy-nrf` I2C uses `time` feature gating, but this would cause a lot of code duplication. I believe compiler should be able to optimise out all timeout checks if `Config::timeout` is never set.

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -34,7 +34,7 @@ flavors = [
 [dependencies]
 embassy-sync = { version = "0.1.0", path = "../embassy-sync" }
 embassy-executor = { version = "0.1.0", path = "../embassy-executor" }
-embassy-time = { version = "0.1.0", path = "../embassy-time", optional = true }
+embassy-time = { version = "0.1.0", path = "../embassy-time" }
 embassy-futures = { version = "0.1.0", path = "../embassy-futures" }
 embassy-cortex-m = { version = "0.1.0", path = "../embassy-cortex-m", features = ["prio-bits-4"]}
 embassy-hal-common = {version = "0.1.0", path = "../embassy-hal-common" }
@@ -84,7 +84,7 @@ exti = []
 
 # Features starting with `_` are for internal use only. They're not intended
 # to be enabled by other crates, and are not covered by semver guarantees.
-_time-driver = ["dep:embassy-time"]
+_time-driver = []
 time-driver-any = ["_time-driver"]
 time-driver-tim2 = ["_time-driver"]
 time-driver-tim3 = ["_time-driver"]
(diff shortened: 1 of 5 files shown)


chemicstry:
I tested `i2cv1` on `STM32F429ZIT6`, but don't have any `i2cv2` board to confirm that it works

lulf:
It's a bit unfortunate to require embassy-time I think. Is there a real requirement of using blocking operations with timeout or should we just require async (embassy_time::with_timeout()) for those use cases?

chemicstry:
I've checked multiple HALs and almost all of them have open issues regarding this problem. So far I've seen mbed (C) and stm32f1xx-hal implement delays for this.

Using async timeout is a possibility, however, the most common use for I2C bus is with `embedded-hal` device drivers so timeout must be supported by embassy's EH implementation. You could write an EH wrapper from the user side, but IMO it's quite a common requirement to not let I2C bus freeze your application.

An alternative could be to implement blocking EH traits using async `embassy_time::with_timeout()` and busy polling future. This would greatly simplify I2C driver code, but would still require `embassy-time`.

chemicstry:
Here's another idea. Leave `I2c` implementation as is and add a `struct TimeoutI2c<'d, T> { i2c: I2c<'d, T> }`, which is gated behind `time` feature. It would implement blocking EH traits using async timeout with block_on(). How about this?

Dirbaio:
> @lulf: Is there a real requirement of using blocking operations with timeout or should we just require async (embassy_time::with_timeout()) for those use cases?

Yes, you might want to use a blocking-only driver and still prevent these i2c hangs. This is why I added it to nrf.

> @chemicstry Leave I2c implementation as is and add a struct TimeoutI2c<'d, T> { i2c: I2c<'d, T> }, which is gated behind time feature

SGTM, better than requiring `embassy-time` in the main `i2c`.

> @chemicstry It would implement blocking EH traits using async timeout with block_on(). 

`async + block_on` leads to much bigger code than simple blocking. It's better to "integrate" the timeout into the blocking busywaiting, like nrf does.

(The async traits OTOH can be impl'd with `with_timeout` just fine, yes)

To avoid code duplication perhaps the internal fns could take a `check_timeout: impl Fn() -> Result<(), Error>`, so you can supply a nop in the "without timeout" ops.

chemicstry:
Closing in favor of #1025

PR #1024, opened 2022-10-24
stm32/adc: Remove voltage and temperature conversions

GrantM11235:
The current conversion utilities are confusing and a bit of a footgun. (Two out of the three examples got it wrong! They didn't measure vref at all, so all the conversions are completely wrong if vcca isn't 3.3v)

I think we should eventually have some sort of conversion utilities in the HAL, but for now I think it is best to just remove it and let the users do their own math.

cc @chemicstry

### DIFF
--- a/embassy-stm32/src/adc/v2.rs
+++ b/embassy-stm32/src/adc/v2.rs
@@ -80,15 +80,6 @@ impl super::sealed::InternalChannel<ADC1> for Temperature {
 }
 
 impl Temperature {
-    /// Converts temperature sensor reading in millivolts to degrees celcius
-    pub fn to_celcius(sample_mv: u16) -> f32 {
-        // From 6.3.22 Temperature sensor characteristics
-        const V25: i32 = 760; // mV
-        const AVG_SLOPE: f32 = 2.5; // mV/C
-
-        (sample_mv as i32 - V25) as f32 / AVG_SLOPE + 25.0
-    }
-
     /// Time needed for temperature sensor readings to stabilize
     pub fn start_time_us() -> u32 {
         10
@@ -172,7 +163,6 @@ impl Prescaler {
 
 pub struct Adc<'d, T: Instance> {
     sample_time: SampleTime,
-    vref_mv: u32,
     resolution: Resolution,
     phantom: PhantomData<&'d mut T>,
 }
@@ -200,7 +190,6 @@ where
         Self {
             sample_time: Default::default(),
             resolution: Resolution::default(),
-            vref_mv: VREF_DEFAULT_MV,
             phantom: PhantomData,
         }
     }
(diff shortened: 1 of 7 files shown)


embassy-stm32/src/adc/v2.rs
```diff
@@ -80,15 +80,6 @@ impl super::sealed::InternalChannel<ADC1> for Temperature {
 }
 
 impl Temperature {
-    /// Converts temperature sensor reading in millivolts to degrees celcius
-    pub fn to_celcius(sample_mv: u16) -> f32 {
-        // From 6.3.22 Temperature sensor characteristics
-        const V25: i32 = 760; // mV
-        const AVG_SLOPE: f32 = 2.5; // mV/C
```
chemicstry, line 87: Maybe we could still have these as associated constant on types? Same with voltage on `VrefInt`.

Although, it has the problem of what type to use. I.e. should AVG_SLOPE be 2500 uV/C u16 or 2.5 mV/C f32.

What do you think?

GrantM11235: I eventually would like to include these constants for all chips in the hal or the metapac, but as far as I can tell, the information is only available in the datasheets. The values aren't necessarily the same even between chips in same family, and I'm not prepared to look through all 159 PDFs to find out :fearful:

chemicstry: Fair point 👍

examples/stm32f4/src/bin/adc.rs
```diff
@@ -24,19 +24,40 @@ async fn main(_spawner: Spawner) {
     // Startup delay can be combined to the maximum of either
     delay.delay_us(Temperature::start_time_us().max(VrefInt::start_time_us()));
 
+    let vref_sample = adc.read_internal(&mut vrefint);
+
+    let convert_to_millivolts = |sample| {
+        // From http://www.st.com/resource/en/datasheet/DM00071990.pdf
+        // 6.3.24 Reference voltage
+        const VREF_MILLIVOLTS: u32 = 1210; // mV
```
chemicstry, line 32: Could you rename this to `VREFINT_MILLIVOLTS` and other variables that refer to `VrefInt`? Also in other examples.

So that it's not confused with actual `Vref`, which is an external pin (on smaller chips connected internally to VDDA).

examples/stm32f4/src/bin/adc.rs
```diff
@@ -24,19 +24,40 @@ async fn main(_spawner: Spawner) {
...
         // Read internal temperature
         let v = adc.read_internal(&mut temp);
-        let celcius = Temperature::to_celcius(adc.to_millivolts(v));
+        let celcius = convert_to_celcius(v);
         info!("Internal temp: {} ({} C)", v, celcius);
 
         // Read internal voltage reference
         let v = adc.read_internal(&mut vrefint);
-        info!("VrefInt: {} ({} mV)", v, adc.to_millivolts(v));
+        info!("VrefInt: {} ({} mV)", v, convert_to_millivolts(v));
```
chemicstry, line 60: Since you are using `VrefInt` to convert voltage there is not point in converting itself, you will always get your defined constant if Vref is stable

chemicstry: Maybe you can calculate Vref (VDDA) here instead

chemicstry (approved):

Dirbaio (approved):
bors r+

PR #1025, opened 2022-10-24
Implement I2C timeouts, second attempt

chemicstry:
This is an alterrnative to #1022 as discussed there.

Timeouts are implemented using suggested `check_timeout: impl Fn() -> Result<(), Error>` function, which does not depend on `embassy-time` by default and is a noop for regular I2C.

This also adds `time` feature like in `embassy-nrf` to enable `embassy-time` dependencies. While at it, I also gated some other peripherals that depend on `embassy-time`, notably `usb` and (partially) `subghz`.

`TimeoutI2c` is currently only implemented for i2cv1, because i2cv2 has additional complications:
- Async methods still use a lot of busy waiting code in between DMA transfers, so simple `with_timeout()` will not work and it will have to use both types of timeouts. It could probably be rewritten to replace busy waits with IRQs, but that's outside the scope of this PR.
- I2C definition `I2c<'d, T, TXDMA, RXDMA>` is different from i2cv1 `I2c<'d, T>` making it hard to share single `TimeoutI2c` wrapper. A couple of options here:
  - Duplicate `TimeoutI2c` code
  - Add dummy `TXDMA`, `RXDMA` types to i2cv1 considering that in the future it should also support DMA

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -82,9 +82,12 @@ memory-x = ["stm32-metapac/memory-x"]
 subghz = []
 exti = []
 
+# Enables additional driver features that depend on embassy-time
+time = ["dep:embassy-time"]
+
 # Features starting with `_` are for internal use only. They're not intended
 # to be enabled by other crates, and are not covered by semver guarantees.
-_time-driver = ["dep:embassy-time"]
+_time-driver = ["time"]
 time-driver-any = ["_time-driver"]
 time-driver-tim2 = ["_time-driver"]
 time-driver-tim3 = ["_time-driver"]
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -7,7 +7,9 @@ use crate::interrupt::Interrupt;
 mod _version;
 pub use _version::*;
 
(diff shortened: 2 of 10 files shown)


Dirbaio:
:+1: 

> TimeoutI2c is currently only implemented for i2cv1, because i2cv2 has additional complications:
> Async methods still use a lot of busy waiting code in between DMA transfers

ideally v1 and v2 should be at feature parity. I know they already aren't but I'd love to not introduce yet another disparity! Perhaps you could add `TimeoutI2c` to `v2` but supporting blocking operation only, as a start? 

> Add dummy TXDMA, RXDMA types to i2cv1 considering that in the future it should also support DMA

Sounds good to me, since we'll need it in the future anyway.

chemicstry:
Done. `TimeoutI2c` should work with both v1 and v2 in blocking mode, but I could only test v1. Btw, is the name `TimeoutI2c` OK? I couldn't think of anything better.

It's also possible to add async gated behind v2 with that weird blocking + async timeout combo. Should I? Tbh, I'm not even sure if it's possible (or worth) to properly fix the driver to not use any busy waits.

Dirbaio (approved):
Nice! Thanks for working on this! :D

bors r+

PR #1026, opened 2022-10-25
stm32/adc: New single channel API

GrantM11235:
Based on #1024, cc @chemicstry 

```rust
let mut input = adc.single_channel(&mut pin, SampleTime::default(), Resolution::TwelveBit);
loop {
        let v = input.read();
        info!("Adc value: {}", v);
}
```

The preferred sample time and resolution are no longer stored inside the `Adc` struct with setter methods, those parameters are now supplied by the user directly whenever the driver needs to configure them.

The `Adc::single_channel` method configures the adc for the specified pin, sample time, and resolution, and returns a `SingleChannel` struct. This struct can be used to quickly and efficiently read from the adc multiple times without re-configuring each time.

I kept the `Adc::read` method because it is slightly more ergonomic in some cases:
```rust
adc.read(pin, sample_time, resolution);
// is a slightly shorter way to write this:
adc.single_channel(pin, sample_time, resolution).read();
```

### Future possibilities

```rust
let mut input = adc.single_channel(&mut pin, SampleTime::default(), Resolution::TwelveBit);

input.dma_read(dma_channel, &mut buffer).await;
```

### DIFF
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -86,7 +86,6 @@ pub use sample_time::SampleTime;
 
 pub struct Adc<'d, T: Instance> {
     sample_time: SampleTime,
-    calibrated_vdda: u32,
     phantom: PhantomData<&'d mut T>,
 }
 
@@ -122,7 +121,6 @@ impl<'d, T: Instance> Adc<'d, T> {
 
         Self {
             sample_time: Default::default(),
-            calibrated_vdda: VDDA_CALIB_MV,
             phantom: PhantomData,
         }
     }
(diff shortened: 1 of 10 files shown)


chemicstry:
Shouldn't resolution still remain on `Adc`, because it is global and can't be configured per-channel?

I'm not too keen on `single_channel()` taking `&mut self`. I think usually you want to measure more than one channel so the API should have that as first-class support. Unfortunatelly this is hard to express using a struct which references `Adc`. Is there anything wrong with modifying original `set_sample_time` to be per-channel? I.e. `Adc::set_sample_time(&mut self, pin: &..., sample_time: SampleTime)`.

Since it's a major refactor, it would also be nice to think how multi-channel DMA API would look like. But it's hard to reason without a PoC. Ideally it would take an array of pins instead of one like in #932, but that requires their types to be erased.

GrantM11235:
This is very stale, so I'm closing it

PR #1029, opened 2022-10-26
Update nrf pacs

diondokter:


Dirbaio (approved):
THanks!

bors r+

diondokter:
Ok, this change to USB should be alright. I checked that the variant is correct for each

embassy-nrf/src/usb.rs
```diff
@@ -313,7 +313,7 @@ impl<'d, T: Instance, P: UsbSupply> driver::Bus for Bus<'d, T, P> {
                 }
             })
             .await;
-            regs.eventcause.write(|w| w.ready().set_bit()); // Write 1 to clear.
+            regs.eventcause.write(|w| w.ready().clear_bit_by_one());
```
Dirbaio, line 316: `clear_bit_by_one`? wtf :o

diondokter: Yeah, I saw that as well. It's pretty verbose which is nice :)

Dirbaio: IMO the name is not obvious at all what it does. I'd say `write_one_to_clear`, or `clear_by_writing_one` or something. But yeah it's just bikeshedding :shrug:

PR #1032, opened 2022-10-27
stm32/adc: Misc refactoring

GrantM11235:
Noteworthy changes:

- Fixed a few typos in the `SampleTime`s
- `set_resolution` now writes directly to the configuration register. This also fixed a bug in v3 where the resolution was changed while `ADEN` is enabled, which the datasheet says isn't allowed.


Dirbaio (approved):
bors r+

PR #1037, opened 2022-10-31
Add uart async task example

miathedev:
Dear Embassy Team,

here i propose an additional async uart pass-through example for the STM32WL.

Because im quite new to Rust, is there something like **interfaces**?

The code: 
```
mut usart1: Uart<
    'static,
    embassy_stm32::peripherals::USART1,
    embassy_stm32::peripherals::DMA1_CH3,
    embassy_stm32::peripherals::DMA1_CH4,
>,
mut usart2: Uart<
    'static,
    embassy_stm32::peripherals::LPUART1,
    embassy_stm32::peripherals::DMA1_CH5,
    embassy_stm32::peripherals::DMA1_CH6,
>,
```
is quite ugly in my opinion. I would like to allow any Type of DMA and USART/UART as argument. Is this possible somehow?
Im open to any feedback.

With love,
Mia

### DIFF
--- a/examples/stm32wl/src/bin/uart_async.rs
+++ b/examples/stm32wl/src/bin/uart_async.rs
@@ -1,0 +1,60 @@
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_stm32::interrupt;
+use embassy_stm32::usart::{Config, Uart};
+use {defmt_rtt as _, panic_probe as _};
+
+/*
+Pass Incoming data from LPUART1 to USART1
+Example is written for the LoRa-E5 mini v1.0,
+but can be surely changed for your needs.
+*/
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let mut config = embassy_stm32::Config::default();
+    config.rcc.mux = embassy_stm32::rcc::ClockSrc::HSE32;
+    let p = embassy_stm32::init(config);
+
+    defmt::info!("Starting system");
+
+    let mut config1 = Config::default();
+    config1.baudrate = 9600;
+
+    let mut config2 = Config::default();
+    config2.baudrate = 9600;
+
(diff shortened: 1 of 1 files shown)


lulf (changes requested):
Thanks for the PR! I added a few comments which should remove the need for the 'ugly' types. 

In general: Yes, there are interfaces in rust, called traits, but you need the full type in this case since the size of the toplevel embassy tasks must be known. However, you could also remove the embassy_executor::task macro and run the function from your main loop (but I suggest copying it all into the main loop instead).

The Uart peripherals implement some traits from the embedded-hal-async crate, similar to HALs for other chips. By importing `embedded_hal_async::serial::{Read, Write}`, you can create async functions like this:

```
async fn bridge<F, T>(from: F, to: T)
    where F: Read, T: Write
 {
    // Read from from, write to to.
}
```

examples/stm32wl/src/bin/uart_async_task.rs
```diff
@@ -0,0 +1,89 @@
...
+    let mut usart1 = Uart::new(p.USART1, p.PB7, p.PB6, irq, p.DMA1_CH3, p.DMA1_CH4, config1);
+
+    unwrap!(usart1.write(b"Hello Embassy World!\r\n").await);
+    unwrap!(usart2.write(b"Hello Embassy World!\r\n").await);
+
+    //PB5 OnBoard-LED on LoRa-E5 mini v1.0
+    unwrap!(spawner.spawn(uart_passthrough(usart1, usart2, p.PB5.degrade())));
+
+    loop {
+        Timer::after(Duration::from_secs(10)).await;
```
lulf, line 46: Rather than having an idle loop here, consider copying the loop from the passthrough task.

miathedev:
So. i hope i changed it correctly as you requested.
Btw. im totally now checking out Rust traits. Thank you so much!

lulf (approved):
Looks good to me! Maybe squash your commits, and then you can merge it yourself :)

bors d+

miathedev:
bors d+

PR #1039, opened 2022-11-01
stm32-metapac-gen: Use `serde_json` to parse json files

GrantM11235:
This makes stm32-metapac-gen over twice as fast. A full run on my desktop goes from about six and a half seconds to about three seconds. Suprisingly, it also reduces the fresh compile time by almost a second.

### DIFF
--- a/stm32-metapac-gen/Cargo.toml
+++ b/stm32-metapac-gen/Cargo.toml
@@ -9,5 +9,7 @@ license = "MIT OR Apache-2.0"
 regex = "1.5.4"
 chiptool = { git = "https://github.com/embassy-rs/chiptool", rev = "28ffa8a19d84914089547f52900ffb5877a5dc23" }
 serde = { version = "1.0.130", features = [ "derive" ] }
+serde_json = "1.0.87"
 serde_yaml = "0.8.21"
 proc-macro2 = "1.0.29"
+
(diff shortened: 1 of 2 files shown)


Dirbaio:
How much does cold build time increase due to having to build both `serde_yaml` and `serde_json`?

GrantM11235:
Build time methodology:
`cargo clean && cargo build --release` in the `stm32-metapac-gen` directory, using the compile time reported by cargo.

Run time methodology:
`time ./target/release/stm32-metapac-gen` in the `stm32-metapac-gen` directory, using the "real" time reported by `time`.

I measure the build time a few more times, and it is actually more than a second faster on average:
- Before:
  - 16.02
  - 16.15
  - 15.87
  - 15.86
  - 16.27
- After:
  - 14.64
  - 14.67
  - 14.85
  - 14.78
  - 14.85

Dirbaio:
I meant, for example, what about `cargo clean && cargo build --release --bin usart` in `examples/stm32f4`?

GrantM11235:
Basically no change:
- Before
	- 16.63
	- 16.66
	- 16.70
- After
	- 16.75
	- 16.60
	- 16.44

Dirbaio (approved):
Interesting, I guess there's no downside then! Thanks for taking the time to do the measurements! :) 

bors r+

PR #1042, opened 2022-11-04
embassy-nrf: Add SPIS module

kalkyl:
Untested on hardware


huntc (commented):
Nice. 

Would it make sense to structure the SPI types in a similar fashion to Uarte ie with split, SpiRx, SpiTx etc?

kalkyl:
Verified to be working on nrf9160

kalkyl:
> Nice. 
> 
> 
> 
> Would it make sense to structure the SPI types in a similar fashion to Uarte ie with split, SpiRx, SpiTx etc?

Do you mean for typestate reasons, so you can't call read on "txonly" instance etc?

huntc:
> Do you mean for typestate reasons, so you can't call read on "txonly" instance etc?

Purely for consistency.

kalkyl:
> > Do you mean for typestate reasons, so you can't call read on "txonly" instance etc?
> 
> 
> 
> Purely for consistency. I don’t know if it makes sense or not though. 

As for now i chose to be consistent with the SPIM driver, so we'd have to change that one aswell if going that route... I have no strong opinion on it :)

Dirbaio:
@huntc splitting an SPI driver makes no sense, because unlike UART you can't start/stop TX and RX independently. Every SCK pulse, you're both TXing and RXing one bit. You can't RX without TXing or vice versa. All you can do is ignore the rx'd bits or tx zeros.

embassy-nrf/src/lib.rs
```diff
@@ -267,5 +268,12 @@ pub fn init(config: config::Config) -> Peripherals {
     #[cfg(feature = "_time-driver")]
     time_driver::init(config.time_interrupt_priority);
 
+    // Disable UARTE (enabled by default for some reason)
```
Dirbaio, line 271: Are you sure this is not an SPM bug? 

Either way, could you split this to a separate PR?

embassy-nrf/src/spis.rs
```diff
@@ -0,0 +1,539 @@
...
+            config,
+        )
+    }
+
+    pub fn new_txonly(
+        spis: impl Peripheral<P = T> + 'd,
+        irq: impl Peripheral<P = T::Interrupt> + 'd,
+        cs: impl Peripheral<P = impl GpioPin> + 'd,
+        sck: impl Peripheral<P = impl GpioPin> + 'd,
+        mosi: impl Peripheral<P = impl GpioPin> + 'd,
```
Dirbaio, line 80: since we're on the slave side, txonly should be MISO, not MOSI?

embassy-nrf/src/spis.rs
```diff
@@ -0,0 +1,539 @@
...
+            config,
+        )
+    }
+
+    pub fn new_rxonly(
+        spis: impl Peripheral<P = T> + 'd,
+        irq: impl Peripheral<P = T::Interrupt> + 'd,
+        cs: impl Peripheral<P = impl GpioPin> + 'd,
+        sck: impl Peripheral<P = impl GpioPin> + 'd,
+        miso: impl Peripheral<P = impl GpioPin> + 'd,
```
Dirbaio, line 100: same

embassy-nrf/src/spis.rs
```diff
@@ -0,0 +1,539 @@
...
+                trace!("Copying SPIS tx buffer into RAM for DMA");
+                let tx_ram_buf = &mut [0; FORCE_COPY_BUFFER_SIZE][..tx.len()];
+                tx_ram_buf.copy_from_slice(tx);
+                self.async_inner_from_ram(rx, tx_ram_buf).await
+            }
+            Err(error) => Err(error),
+        }
+    }
+
+    /// Reads data from the SPI bus without sending anything. Blocks until the buffer has been filled.
```
Dirbaio, line 343: Does this return before filling the buf if CS gets deasserted? If that's the case, "Blocks until the buffer has been filled." is not really true.

embassy-nrf/src/spis.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// If necessary, the write buffer will be copied into RAM (see struct description for detail).
+    /// Returns number of bytes written.
+    pub fn blocking_write(&mut self, data: &[u8]) -> Result<usize, Error> {
+        self.blocking_inner(&mut [], data).map(|n| n.1)
+    }
+
+    /// Same as [`blocking_write`](Spis::blocking_write) but will fail instead of copying data into RAM. Consult the module level documentation to learn more.
+    /// Returns number of bytes written.
+    pub fn blocking_write_from_ram(&mut self, data: &[u8]) -> Result<usize, Error> {
+        self.blocking_inner(&mut [], data).map(|n| n.1)
```
Dirbaio, line 379: should be `_from_ram`

embassy-nrf/src/spis.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// If necessary, the write buffer will be copied into RAM (see struct description for detail).
+    /// Returns number of bytes transferred `(n_rx, n_tx)`.
+    pub fn blocking_transfer(&mut self, read: &mut [u8], write: &[u8]) -> Result<(usize, usize), Error> {
+        self.blocking_inner(read, write)
+    }
+
+    /// Same as [`blocking_transfer`](Spis::blocking_transfer) but will fail instead of copying data into RAM. Consult the module level documentation to learn more.
+    /// Returns number of bytes transferred `(n_rx, n_tx)`.
+    pub fn blocking_transfer_from_ram(&mut self, read: &mut [u8], write: &[u8]) -> Result<(usize, usize), Error> {
+        self.blocking_inner(read, write)
```
Dirbaio, line 359: should be `_from_ram`

Dirbaio (approved):
Thank you!

bors r+

embassy-nrf/src/lib.rs
kalkyl: Not sure, tried a couple different SPM:s with same result but they may all have the same bug... Will submit this as a separate PR (note that SPIS is unusable without this)

PR #1043, opened 2022-11-06
Rpi Pico PIO driver

fluffware:
This is a driver for the two PIOs in the Pico. I've tried using the type system to get as much compile time checks as possible. There is asynchronous support for FIFOs an Irqs. No DMA support yet. There's an example that shows how to use the driver asynchronously.

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -26,6 +26,7 @@ time-driver = []
 rom-func-cache = []
 intrinsics = []
 rom-v2-intrinsics = []
+pio = ["dep:pio", "dep:pio-proc"]
 
 # Enable nightly-only features
 nightly = ["embassy-executor/nightly", "embedded-hal-1", "embedded-hal-async", "embassy-embedded-hal/nightly", "dep:embassy-usb-driver", "dep:embedded-io"]
@@ -64,3 +65,10 @@ embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["un
 embedded-hal-1 = { package = "embedded-hal", version = "=1.0.0-alpha.9", optional = true}
 embedded-hal-async = { version = "=0.2.0-alpha.0", optional = true}
 embedded-hal-nb = { version = "=1.0.0-alpha.1", optional = true}
+
+paste = "1.0"
+pio-proc = {version= "0.2", optional = true}
+pio = {version= "0.2", optional = true}
+
+[patch.crates-io]
+pio = {git = "https://github.com/rp-rs/pio-rs.git"}
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -476,11 +476,13 @@ impl<'d, T: Pin> Flex<'d, T> {
         unsafe {
             self.pin.pad_ctrl().modify(|w| {
                 w.set_ie(true);
-                match pull {
-                    Pull::Up => w.set_pue(true),
-                    Pull::Down => w.set_pde(true),
-                    Pull::None => {}
-                }
+                let (pu, pd) = match pull {
+                    Pull::Up => (true, false),
+                    Pull::Down => (false, true),
+                    Pull::None => (false, false),
+                };
+                w.set_pue(pu);
+                w.set_pde(pd);
             });
         }
     }
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -48,6 +48,21 @@ pub enum Pull {
     Down,
 }
 
+/// Drive strength of an output
+#[derive(Debug, Eq, PartialEq)]
+pub enum Drive {
+    _2mA,
+    _4mA,
+    _8mA,
+    _12mA,
+}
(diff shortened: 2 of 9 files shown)


kalkyl:
Awesome! Already successfully using this in a current project :)

nvx (commented):
I noticed a mix-up of in/out in one of the examples based on my testing.

It also seems that we're missing the ability to specify the jmp pin. Looks like `set_jmp_pin` exists under `execctrl` register rather than `pinctrl` which I imagine is how it got missed.

Finally is there any reason the PioPin pin() func is private? Being public would be a little more convenient in some cases (saves having to save a copy of the pin number before converting it to a PioPin)

Awesome functionality in general though!

nvx:
I discovered `JMP` instructions behaving incorrectly when loaded at non-zero origins (the only jmp in the example applications is coincidentally in a program loaded at origin 0 so it behaves correctly there).

I initially thought this was a bug with the Rust PIO assembler (https://github.com/rp-rs/pio-rs/issues/34), but looks like its behaviour is consistent with the C SDK and it's up to the PIO loader to correct the JMP offsets.

lulf:
Could you rebase the PR?

nvx:
@fluffware did you have any thoughts on the relative jmp issue before this gets merged? It looks like it shouldn't be too hard to fix (can leverage code from the other rust program loader). Happy to submit a PR if needed (PR to your ksb_pio branch I guess?)

fluffware:
> @fluffware did you have any thoughts on the relative jmp issue before this gets merged? It looks like it shouldn't be too hard to fix (can leverage code from the other rust program loader). Happy to submit a PR if needed (PR to your ksb_pio branch I guess?)

I think that belongs in pio-rs, either as an option when parsing or a fix-up afterwards (like a link-editor).

nvx:
> > @fluffware did you have any thoughts on the relative jmp issue before this gets merged? It looks like it shouldn't be too hard to fix (can leverage code from the other rust program loader). Happy to submit a PR if needed (PR to your ksb_pio branch I guess?)
> 
> I think that belongs in pio-rs, either as an option when parsing or a fix-up afterwards (like a link-editor).

Currently rp-hal does this fixup not pio-rs when not using embassy - see the code at https://github.com/rp-rs/rp-hal/blob/4b2b33b950d06ad690cde7088db2545e39f02221/rp2040-hal/src/pio.rs#L193

I believe this mimics the behaviour of the C pico sdk as well. We can certainly decide to do things differently, but I daresay there would have to be a good reason for deviating from way everywhere else does it.

fluffware:
> > > @fluffware did you have any thoughts on the relative jmp issue before this gets merged? It looks like it shouldn't be too hard to fix (can leverage code from the other rust program loader). Happy to submit a PR if needed (PR to your ksb_pio branch I guess?)
> > 
> > 
> > I think that belongs in pio-rs, either as an option when parsing or a fix-up afterwards (like a link-editor).
> 
> Currently rp-hal does this fixup not pio-rs when not using embassy - see the code at https://github.com/rp-rs/rp-hal/blob/4b2b33b950d06ad690cde7088db2545e39f02221/rp2040-hal/src/pio.rs#L193
> 
> I believe this mimics the behaviour of the C pico sdk as well. We can certainly decide to do things differently, but I daresay there would have to be a good reason for deviating from way everywhere else does it.

I think everyone else is doing it wrong :-) The fixup manipulates the instructions and needs knowledge of how they are encoded. This has little to do with getting the code to run on the hardware.
Ok, how about this? Let's wrap the program in a new class, let's call it ResolvedProgram, which gives a new view of the program  with all the addresses adjusted. This way we don't mix the encoding aware code with the rest. Hopefully that class could be part of pio-rs at some point.

lulf (approved):
Looking forward to play with this, thanks for working on it!

bors r+

kalkyl:
All i wanted for xmas!

lulf (commented):
I wonder if this is breaking the rp GPIO CI test. Does the test need to be modified or is it catching a real issue?

kalkyl:
> 

I saw this PR changes `write` into `modify` in gpio `set_pull`, this is not right? I guess that's why the CI fails because it's not reseting the pulls...

fluffware:
> > 
> 
> I saw this PR changes `write` into `modify` in gpio `set_pull`, this is not right? I guess that's why the CI fails because it's not reseting the pulls...
Unless I've completely misunderstood how modify works this is correct. Using write would overwrite any other settings done on the pad control register. What is not correct is that the unused pulls are not cleared.  I'll fix that.

kalkyl:
> > > 
> 
> > 
> 
> > I saw this PR changes `write` into `modify` in gpio `set_pull`, this is not right? I guess that's why the CI fails because it's not reseting the pulls...
> 
> 
> 
> Unless I've completely misunderstood how modify works this is correct. Using write would overwrite any other settings done on the pad control register. What is not correct is that the unused pulls are not cleared.  I'll fix that.
> 
> 

Ahh yes sorry, didn't see you added those other fns that also mess with the pad_ctrl reg, so your latest addition should be right yeah :)

PR #1044, opened 2022-11-06
rp/uart: use lockfree ringbuffer.

Dirbaio:
This gets rid of another PeripheralMutex usage.

Extracted out of #980

### DIFF
--- a/embassy-hal-common/src/atomic_ring_buffer.rs
+++ b/embassy-hal-common/src/atomic_ring_buffer.rs
@@ -1,0 +1,331 @@
+use core::slice;
+use core::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
+
+/// Atomic reusable ringbuffer
+///
+/// This ringbuffer implementation is designed to be stored in a `static`,
+/// therefore all methods take `&self` and not `&mut self`.
+///
+/// It is "reusable": when created it has no backing buffer, you can give it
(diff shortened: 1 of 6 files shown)


embassy-hal-common/src/atomic_ring_buffer.rs
```diff
@@ -0,0 +1,156 @@
...
+        (unsafe { buf.add(start) }, n)
+    }
+
+    pub fn pop_done(&self, n: usize) {
+        trace!("  ringbuf: pop {:?}", n);
+
+        let start = self.0.start.load(Ordering::Relaxed);
+        self.0.start.store(self.0.wrap(start + n), Ordering::Release);
+    }
+}
```
huntc, line 156: Would it be possible to get some tests for this? It looks quite testable. Thanks.

huntc (approved):
Thanks for this - just one comment, but LGTM.

lulf (approved):
LGTM, I really like the ergonomics of not having to manage the State type in the app code!

I agree with @huntc it would be nice to have some tests for the core ringbuffer part - it's the kind of code one can accidentally make bad changes too in the future. Also, if you have some references or explanation of why the barriers are valid, that would be helpful to any readers I think. Then again, it hopefully won't need to change...

Dirbaio:
Thanks for the feedback! Added docs, tests, comments explaining the atomic orderings.

bors r+

PR #1047, opened 2022-11-07
Ensure embassy-lora stm32wl supports log crate

lulf:


embassy-lora/src/stm32wl/mod.rs
```diff
@@ -172,7 +172,7 @@ impl<'d, RS: RadioSwitch> SubGhzRadio<'d, RS> {
                 self.radio.read_buffer(ptr, &mut buf[..len as usize])?;
                 self.radio.set_standby(StandbyClk::Rc)?;
 
-                trace!("RX done: {=[u8]:#02X}", &mut buf[..len as usize]);
+                trace!("RX done: {:02x?}", &mut buf[..len as usize]);
```
Dirbaio, line 175: defmt needs `:02x`, log needs `:02x?`

Very unfortunate https://github.com/knurling-rs/defmt/issues/648

lulf: 😭

lulf: what's the general preferred workaround? #[cfg()] or just derive LowerHex (if that works?)

Dirbaio: I've done cfg's for this in the past... A wrapper type deriving defmt and LowerHex could work but I think it's more annoying...

PR #1049, opened 2022-11-09
embassy-nrf: Add I2S module

chris-zen:
This PR adds I2S support for the nrf52 series (`nrf52832`, `nrf52833`, `nrf52840`).

We could only test it in a `nrf52840` in master mode for an output stream (see `i2s_waveform` example), using a clone of the [Adafruit I2S Stereo Decoder - UDA1334A](https://learn.adafruit.com/adafruit-i2s-stereo-decoder-uda1334a/overview).

We were wondering if this could be a welcome addition to embassy, as we are working on this very informally and don't have much free time for it.

https://user-images.githubusercontent.com/932644/202316609-e53cd912-e463-4e01-839e-0bbdf37020da.mp4

Co-authored-by: @brainstorm <brainstorm@nopcode.org>

### DIFF
--- a/embassy-nrf/src/i2s.rs
+++ b/embassy-nrf/src/i2s.rs
@@ -327,6 +327,7 @@ impl<'d, T: Instance> I2S<'d, T> {
     pub fn start(&self) -> &Self {
         let r = T::regs();
         self.enable();
+        trace!("START");
         r.tasks_start.write(|w| unsafe { w.bits(1) });
         self
     }
@@ -487,8 +488,8 @@ impl<'d, T: Instance> I2sOutput<'d, T> {
         let s = T::state();
 
         let seq = s.seq.fetch_add(1, Ordering::Relaxed);
-        if r.events_txptrupd.read().bits() != 0 && seq > 0 {
-            info!("XRUN!");
+        if r.events_txptrupd.read().bits() != 0 && seq > 1 {
+            warn!("XRUN!");
             loop {}
         }
 
@@ -505,8 +506,8 @@ impl<'d, T: Instance> I2sOutput<'d, T> {
         });
 
         trace!("[{}] PTR", s.seq.load(Ordering::Relaxed));
-        r.txd.ptr.write(|w| unsafe { w.ptr().bits(ptr as u32) });
         r.rxtxd.maxcnt.write(|w| unsafe { w.bits(maxcnt) });
+        r.txd.ptr.write(|w| unsafe { w.ptr().bits(ptr as u32) });
 
         compiler_fence(Ordering::SeqCst);
 
@@ -517,7 +518,7 @@ impl<'d, T: Instance> I2sOutput<'d, T> {
                 r.events_txptrupd.reset();
                 r.intenset.write(|w| w.txptrupd().set());
                 let overruns = s.overruns.fetch_sub(1, Ordering::Relaxed);
-                if overruns - 1 != 0 {
+                if overruns != 0 {
                     warn!("XRUN: {}", overruns);
                     s.overruns.store(0, Ordering::Relaxed)
                 }
--- a/embassy-nrf/src/i2s.rs
+++ b/embassy-nrf/src/i2s.rs
@@ -2,17 +2,18 @@
 
 //! I2S
 
-//use core::future::poll_fn;
-//use core::sync::atomic::{compiler_fence, Ordering};
-//use core::task::Poll;
+use core::future::poll_fn;
+use core::sync::atomic::{compiler_fence, Ordering};
+use core::task::Poll;
 
-//use embassy_hal_common::drop::OnDrop;
+use embassy_cortex_m::interrupt::{InterruptExt, Priority};
+use embassy_hal_common::drop::OnDrop;
 use embassy_hal_common::{into_ref, PeripheralRef};
 
 //use crate::gpio::sealed::Pin as _;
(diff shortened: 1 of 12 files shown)


chris-zen:
Moved from description to a comment for historical tracking:

We are facing some problems because the output signal contains glitches, and we haven't been able to identify the cause. We were wondering if more people would be interested in helping us with this problem before we add more features. I'll explain our research so far:

1) When the buffer size can fit an integer number of waveform periods, the sound is clean and everything works fine.



2) When the buffer size doesn't fit an integer number of waveform periods, then we observe glitches, possible due to discontinuities in the signal from one buffer to the next one.







Using a sigrok capture of the i2s data pin into WAV, I confirmed that the last two samples of each buffer appear always with wrong values. This is an example for a mono buffer of 50 samples:



3) I initially thought that it was caused by overruns (when the preparation of the next buffer takes more than the reproduction of the previous one), which might be making the I2S peripheral to repeat the last buffer in absence of the new one. But I haven't been able to detect any overrun (you can look for the word `XRUN` in the code to see how we looked for that).
4) The logging seems pretty good and everything happens as expected.

    ```
    77.620361 TRACE [1] PTR (The PTR register is set)
    77.620391 TRACE [1] POLL Pending (The future is polled but the event TXPTRUPD  has not happened yet)
    77.657440 TRACE [1] INT (The event TXPTRUPD happens, which triggers the interrupt, which wakes the future)
    77.657501 TRACE [1] POLL Ready (The future is polled, and returns ready)
    ```

5) I have tested the `SineOsc` waveform generator in the desktop and it works as expected.
6) I have checked the MAXCNT value and it coincides with the buffer size in 32 bits words.
7) It looks like a misalignment between the signal values between successive buffers, but I can't figure out why.

Any help would be appreciated ;-)

chris-zen:
We were overrunning the DMA buffers. Now we keep 3 buffers alternating sequentially, so there are 2 used by the DMA, while the third is being written by the application.

brainstorm:
Anything else we could do to get this reviewed, @lulf @Dirbaio?

embassy-nrf/src/i2s.rs
```diff
@@ -0,0 +1,1051 @@
...
+}
+
+/// I2S input
+pub struct InputStream<'d, T: Instance> {
+    _p: PeripheralRef<'d, T>,
+}
+
+impl<'d, T: Instance> InputStream<'d, T> {
+    /// Prepare the initial buffer and start the I2S transfer.
+    pub async fn start<S>(&self, buffer: &mut [S]) -> Result<(), Error>
```
lulf, line 697: Should this be &mut to prevent multiple concurrent uses of it?

embassy-nrf/src/i2s.rs
```diff
@@ -0,0 +1,1051 @@
...
+}
+
+/// I2S full duplex stream (input & output)
+pub struct FullDuplexStream<'d, T: Instance> {
+    _p: PeripheralRef<'d, T>,
+}
+
+impl<'d, T: Instance> FullDuplexStream<'d, T> {
+    /// Prepare the initial buffers and start the I2S transfer.
+    pub async fn start<S>(&self, buffer_out: &[S], buffer_in: &mut [S]) -> Result<(), Error>
```
lulf, line 746: Same as for InputStream::start

embassy-nrf/src/i2s.rs
```diff
@@ -0,0 +1,1051 @@
...
+}
+
+/// I2S output
+pub struct OutputStream<'d, T: Instance> {
+    _p: PeripheralRef<'d, T>,
+}
+
+impl<'d, T: Instance> OutputStream<'d, T> {
+    /// Prepare the initial buffer and start the I2S transfer.
+    pub async fn start<S>(&self, buffer: &[S]) -> Result<(), Error>
```
lulf, line 648: Same as for InputStream::start

lulf (commented):
I think it looks good "structurally", but I feel too inexperienced with i2s in general to review the correctness of how the peripheral is used.

I added some comments on the use of non-mut references that I thought looked dangerous.

chris-zen:
@lulf thank you very much for the review.

@lulf @Dirbaio the current design allows the application to write on buffers that are still configured in the DMA (that's why I had problems as documented initially). I was following the product specification literally, and it allows for a very intuitive way to stream data in a loop without requiring copying data, ring buffers or other complex synchronisation mechanisms. Given that the I2S peripheral has a double buffer internally, there will always be two buffers that are being used by the DMA, and a third one that will be free and could be used by the application to prepare the next buffer to be sent. The way to protect users from misusing the buffers would be to make the I2S module the responsible to manage the memory of the buffers and its rotation completely. The counter side is that the application would loose control of that memory (but that's probably fine, I haven't been able to think on any case where that's a problem).

I'm willing to make the changes to move the management of the memory into the I2S module rather than leaving it for the application if you are happier with that. Looking forward for your feedback.

chris-zen:
@lulf @Dirbaio I consider this work ready for review and merge. I'd appreciate if you could have a look ;-)

I've included 3 examples for output only, input only and full duplex.

lulf (approved):
I think my initial comments were addressed, the code reads nicely the APIs seem to enforce safe use, so I think it's in a mergeable state. I'm not sure what terminology is appropriate, but it's consistent with existing code so I don't think it makes sense to make this an outlier.

Thanks for putting in the work to support this!

bors r+

PR #1051, opened 2022-11-10
Adapted nvmc so it can be used for all nrf targets

diondokter:
Title says it all pretty much.
It's a bit annoying with the configs though.

I've made this a draft because I haven't really tested this yet.

### DIFF
--- a/embassy-nrf/src/nvmc.rs
+++ b/embassy-nrf/src/nvmc.rs
@@ -10,8 +10,12 @@ use embedded_storage::nor_flash::{
 use crate::peripherals::NVMC;
 use crate::{pac, Peripheral};
 
+#[cfg(not(feature = "_nrf5340-net"))]
 /// Erase size of NVMC flash in bytes.
 pub const PAGE_SIZE: usize = 4096;
+#[cfg(feature = "_nrf5340-net")]
+/// Erase size of NVMC flash in bytes.
+pub const PAGE_SIZE: usize = 2048;
 
 /// Size of NVMC flash in bytes.
 pub const FLASH_SIZE: usize = crate::chip::FLASH_SIZE;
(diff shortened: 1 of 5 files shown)


lulf (approved):
Changes look good to me, and I think this change would enable embassy-boot support for those targets as well. 

I can confirm it works on nrf52, but I've not been able to test on nrf91 (yet, I'm trying to figure out how to flash the devkit). When you've been able to test it, feel free to merge it.

bors d+

embassy-nrf/src/nvmc.rs
```diff
@@ -55,6 +59,56 @@ impl<'d> Nvmc<'d> {
...
+    }
+
+    #[cfg(any(feature = "_nrf9160", feature = "_nrf5340"))]
+    fn erase_page(&mut self, page: u32) {
+        #[cfg(not(feature = "_nrf5340-net"))]
+        const FLASH_START_ADDR: u32 = 0;
+        #[cfg(feature = "_nrf5340-net")]
+        const FLASH_START_ADDR: u32 = 0x100_0000;
+
+        let first_page_word = (FLASH_START_ADDR + page * PAGE_SIZE as u32) as *mut u32;
```
balderk, line 86: I think there is a bug here. **page** is the address off the page, and not the page number, thus the logic can be simplified to ```let firs_page_word = page as *mut u32```.

This also renders the **const FLASH_START_ADDR: ...** not needed.

Should the **page** argument be renamed to **page_addr** to improve clarity?

diondokter: Ah yes, you're right! Well spotted. `page_addr` is a better name yes.
Still didn't have the time to test this out. Hopefully before the next year.

diondokter:
Tested it out and it works great :D
Thanks for catching the bug @balderk !

bors r+

PR #1052, opened 2022-11-10
stm32: Fix watchdog division by zero for 256 prescaler, add watchdog …

matoushybl:
…example for H7

The problem is that `2u8.powi(8) == 0`, which causes division by zero.


lulf:
Thanks!

bors r+

PR #1053, opened 2022-11-10
Disable MMC interrupts

### LINKED ISSUE #594: stm32 eth/v2: mask unneeded interrupts
I believe our driver has the same bug.

https://matrix.to/#/!RxaQWWMYJTsVdHEKmL:matrix.org/$Wpng1U0beGru3uulUVjO0Pt2g6DbWyJUjl49uq0Vwj0?via=matrix.org&via=rubdos.be&via=t2bot.io

> ryan-summers We found an issue in the H7 where the ethernet peripheral has by-default enabled counters that trigger an ISR whenever they're half full, so after you get 2^31 good packets, the ISR flips and your ETH handler now never exits
> ryan-summers That was a fun one to debug when it only failed after ~5 days of continuous heavy data livestreaming :^)
> ryan-summers There's a workaround in the H7 HAL now, I think it got released in 0.11
> ryan-summers Well, by workaround I mean we disable the ISRs
> ryan-summers Ref https://github.com/stm32-rs/stm32h7xx-hal/pull/295

matoushybl:
MMC interrupts can cause firmware hangup - refer to: https://github.com/stm32-rs/stm32h7xx-hal/issues/275 for more information

Fixes #594

### DIFF
--- a/embassy-stm32/src/eth/v2/mod.rs
+++ b/embassy-stm32/src/eth/v2/mod.rs
@@ -116,6 +116,24 @@ impl<'d, T: Instance, P: PHY, const TX: usize, const RX: usize> Ethernet<'d, T, 
 
         mac.macqtx_fcr().modify(|w| w.set_pt(0x100));
 
+        // disable all MMC RX interrupts
+        mac.mmc_rx_interrupt_mask().write(|w| {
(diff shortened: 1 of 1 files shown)


lulf:
Thanks!

bors r+

PR #1054, opened 2022-11-11
riscv fixes

swolix:
With these changes I can run embassy on our RISC-V processor, please consider merging this, feedback is very welcome.

I don't fully understand the code in the executor, but I have implemented a critical section by globally disabling interrupts, which means the wfi inside the critical section will hang the whole thing.

### DIFF
--- a/embassy-executor/src/arch/riscv32.rs
+++ b/embassy-executor/src/arch/riscv32.rs
@@ -54,20 +54,7 @@ impl Executor {
         loop {
             unsafe {
                 self.inner.poll();
-                // we do not care about race conditions between the load and store operations, interrupts
-                //will only set this value to true.
-                critical_section::with(|_| {
-                    // if there is work to do, loop back to polling
-                    // TODO can we relax this?
-                    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
-                        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
-                    }
-                    // if not, wait for interrupt
-                    else {
-                        core::arch::asm!("wfi");
-                    }
-                });
-                // if an interrupt occurred while waiting, it will be serviced here
+                core::arch::asm!("wfi");
             }
         }
     }
--- a/embassy-executor/src/arch/riscv32.rs
+++ b/embassy-executor/src/arch/riscv32.rs
@@ -54,7 +54,12 @@ impl Executor {
         loop {
             unsafe {
                 self.inner.poll();
-                core::arch::asm!("wfi");
+                // we do not care about race conditions between the load and store operations, interrupts
+                // will only set this value to true.
+                // if there is work to do, loop back to polling
+                if !SIGNAL_WORK_THREAD_MODE.fetch_and(false, Ordering::SeqCst) {
+                    core::arch::asm!("wfi");
+                }
             }
         }
     }
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -25,6 +25,7 @@ flavors = [
 default = []
 std = ["embassy-macros/std", "critical-section/std"]
 wasm = ["dep:wasm-bindgen", "dep:js-sys", "embassy-macros/wasm"]
+riscv = ["embassy-macros/riscv"]
 
 # Enable nightly-only features
 nightly = []
--- a/embassy-macros/Cargo.toml
+++ b/embassy-macros/Cargo.toml
@@ -16,6 +16,7 @@ proc-macro = true
 [features]
 std = []
 wasm = []
+riscv = []
 
 # Enabling this cause interrupt::take! to require embassy-executor
 rtos-trace-interrupt = []
--- a/embassy-macros/src/macros/main.rs
+++ b/embassy-macros/src/macros/main.rs
@@ -45,7 +45,7 @@ pub fn run(args: syn::AttributeArgs, f: syn::ItemFn) -> Result<TokenStream, Toke
         }
     };
 
-    #[cfg(all(feature = "std", not(feature = "wasm")))]
+    #[cfg(all(feature = "std", not(feature = "wasm"), not(feature = "riscv")))]
     let main = quote! {
         fn main() -> ! {
             let mut executor = ::embassy_executor::Executor::new();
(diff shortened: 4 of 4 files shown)


embassy-executor/src/arch/riscv32.rs
```diff
@@ -54,20 +54,7 @@ impl Executor {
...
-                    // TODO can we relax this?
-                    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
-                        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
-                    }
-                    // if not, wait for interrupt
-                    else {
-                        core::arch::asm!("wfi");
-                    }
-                });
-                // if an interrupt occurred while waiting, it will be serviced here
```
swolix, line 70: I think I break something here, but I am not sure what / how to fix.
The original code is broken if the critical section disables interrupts

swolix: I wonder if the code should not remain the same but without the critical section, if this is a proper atomic, the critical section is not needed? I *assume* atomic instructions are emitted for the AtomicBool?

Or

Why is the store dependent on the load at all, can't the store be unconditional?

swolix: I noticed this code is *probably* copied from xtensa which (according to the comments) ends the critical section on the wfi (equivalent) code. This is not true for RISC-V I believe

embassy-executor/src/arch/riscv32.rs
```diff
@@ -55,19 +55,11 @@ impl Executor {
...
+                // will only set this value to true.
+                // if there is work to do, loop back to polling
+                if !SIGNAL_WORK_THREAD_MODE.fetch_and(false, Ordering::SeqCst) {
+                    core::arch::asm!("wfi");
+                }
```
swolix, line 62: the wfi in the critical section hangs on RISC-V, I think this code does the same, but I don't fully understand what it is for

Dirbaio: The critical section is to avoid a race where:
- Main thread checks SIGNAL_WORK_THREAD_MODE, sees it's `false`.
- Then an interrupt happens, sets `SIGNAL_WORK_THREAD_MODE` to `true`.
- Main thread goes to `wfi` anyway, because it had checked `SIGNAL_WORK_THREAD_MODE` before and seen `false`.
- `wfi` sleeps forever, the interrupt doesn't interrupt it because it has already happened in the past
- Everything hangs!

When this code was written, we thought that an interrupt becoming pending still causes `wfi` to exit, even if the interrupt doesn't immediately get handled because interrupts were disabled. Perhaps this is not the case? or perhaps only in some chips? iirc the riscv spec was somewhat tight-lipped regarding this.

cc @MabezDev @ivmarkov

swolix: Reading riscv spec again makes me think you are right and wfi should just resume, even if interrupts are disabled (which my critical section implementation does).

However, I think my proposed change still makes sense; wfi would / should exit on pending interrupts, not on new interrupts alone. The critical section in the original code is not needed *I think* since we can atomically "fetch_and"?

Dirbaio: Your code is still vulnerable to the race if things execute in this order:

- main: does `fetch_and`, sees the flag is `false`
- irq: fires, sets the flag to `true`.
- main: does `wfi`, hangs.

This is why interrupts have to be disabled between checking the flag and `wfi`.

swolix: Ah, I get it now, thanks!

I'll revert these changes, is the other stuff ok for merge? (then I'll rebase those on top of latest master)

lulf:
Lets merge this, and I'll update #1071 to include the riscv changes as well. Thank you for the fix!

bors r+

swolix:
@Dirbaio This is our own implementation, I (now) believe it is our implementation of the WFI instruction that is at fault (and the embassy code to be correct), so we'll have to fix this.
Thanks for merging the other changes!

elpiel:
@swolix would it be possible to share an example with embassy?
I don't want to get started from scratch since you've seen to have it working.
I also suppose you're talking about ESP32-C3?

swolix:
@elpiel; I would, but I use our own RISC-V core (including our own SVD generator). If you are willing you can make a new pull-req adding an example, I'll review and help get it up and running, but I do not have a ESP32-C3, so I cannot test. I can give you some pointers, hopefully this get you started;

The ESP32 is an IMC implementation, so you probably need to use the "riscv32i-unknown-none-elf" toolchain, and possibly turn on individual processor features, in `.cargo/config`;

```
[target.riscv32i-unknown-none-elf]
rustflags = [
  "-C", "link-arg=-Tmemory.x",
  "-C", "link-arg=-Tlink.x",
  "-C", "link-arg=-Tdefmt.x",
  "-C", "target-feature=+m",
  "-C", "target-feature=+c",
]

[build]
target = "riscv32i-unknown-none-elf"
```
`rust-toolchain.toml`;
```
[toolchain]
channel = "nightly"
targets = [ "riscv32i-unknown-none-elf" ]
```
`Cargo.toml`;
```
[dependencies]
riscv-rt = { version = "0.9.*", git = "https://github.com/rust-embedded/riscv-rt.git" }
riscv = "0.8.*"
critical-section = { version = "1.1.1", features = ["restore-state-bool"] }

[patch.crates-io]
atomic-pool = { version = "1.0.0", git = 'https://github.com/swolix/atomic-pool.git' }
```

Note that I am not sure about the atomic pool fix yet; I don't know if this is correct (and whether I should get those merged), but I need this for atomic support (to get it compiling).

Feel free to contact me if you have further questions

MabezDev:
@elpiel If you want an esp32c3 example (or any esp32 actually) checkout the example I created here: https://github.com/esp-rs/esp-hal/blob/main/esp32c3-hal/examples/embassy_hello_world.rs

elpiel:
@swolix and @MabezDev thank you both for providing some info.
I've ordered an ESP32-C3 from a local open-source HW company and the moment I have it, I will try to set it up with embassy.

elpiel:
@swolix and @MabezDev thank you both for providing some info.
I've ordered an ESP32-C3 from a local open-source HW company and the moment I have it, I will try to set it up with embassy.

elpiel:
I've made this template repository based on the original esp32 template.
It still has some work to be done in order to be used with `cargo generate` but it's a good starting point for anyone looking to use ESP32-C3 and embassy:

https://github.com/LechevSpace/embassy-esp32-template

PR #1056, opened 2022-11-13
embassy-nrf: Add TWIS module

kalkyl:
Verified to be working on nrf9160


embassy-nrf/src/twis.rs
```diff
@@ -0,0 +1,745 @@
...
+
+        // Enable TWIS instance.
+        r.enable.write(|w| w.enable().enabled());
+
+        // Disable all events interrupts
+        r.intenclr.write(|w| unsafe { w.bits(0xFFFF_FFFF) });
+
+        // Set address
+        r.address[0].write(|w| unsafe { w.address().bits(config.address0) });
+        r.config.modify(|_r, w| w.address0().enabled());
```
Dirbaio, line 138: `config` state from before is unknown. It's probably best to do a `.write()` with the two bits values.

embassy-nrf/src/twis.rs
```diff
@@ -0,0 +1,745 @@
...
+    }
+
+    fn clear_errorsrc(&mut self) {
+        let r = T::regs();
+        r.errorsrc
+            .write(|w| w.overflow().bit(true).overread().bit(true).dnack().bit(true));
+    }
+
+    /// Returns matched address for latest command.
+    pub fn address_match(&self) -> u8 {
```
Dirbaio, line 246: Perhaps it might be easier to use if you return the address index (first or second) matched, instead of the actual address? I'm thinking of a use case where you're emulating 2 different devices, and the address for both is determined at runtime somehow.

If you return the index, user code can go "oh it's the 2nd address? then I go do 2nd-device stuff".  If you return the address instead, the user has to check which one it is against the dynamically calculated ones.

Or perhaps we can do both! :D

embassy-nrf/src/twis.rs
```diff
@@ -0,0 +1,745 @@
...
+        self.blocking_wait_timeout(timeout)
+    }
+
+    // ===========================================
+
+    /// Listen asynchronously for commands from an I2C master.
+    ///
+    /// The buffer must have a length of at most 255 bytes on the nRF52832
+    /// and at most 65535 bytes on the nRF52840.
+    pub async fn listen(&mut self, buffer: &mut [u8]) -> Result<Command, Error> {
```
Dirbaio, line 661: Huh, this API is very *interesting*!

- Docs could be epanded a bit. It took me a bit to figure out what "listen" means, and the fact that `buffer` is provided just in case the master does a write, but might go unused.
- It's confusing that `Command` variants are named from the master's PoV (write meaning a master->slave transfer) and the `Twis::write` method is named from the slave's PoV (write meaning a slave->master transfer).
- It could return the address (or the address index) inside Command, or maybe document "if you want to know which one it was, call `.address_match()`.

Dirbaio: The alternative would be to name everything from the slave's PoV but I can see that being confusing too, ugh. 

"Master wants to *read* from us so we should *write* data to it". vs "Master wants us to *write* to it so we do so"

embassy-nrf/src/twis.rs
kalkyl: Good point! Will add address_match_index so we have both :)

embassy-nrf/src/twis.rs
kalkyl: Hmm, to me a Command is the same regardless of PoV, since it's the "state of the bus"... But yes it might be confusing that the rx can also "write", maybe this `write` fn should be renamed to something else, like `respond`?
Is there also a better name for `listen` that would make more sense to you?

kalkyl: ...but if other people find reversing the Command would be more intuitive, i could totally live with that :)

Dirbaio: > maybe this write fn should be renamed to something else, like respond or provide?

Oooh I like this. `respond_to_read` maybe, so it's more explicit?

> Is there also a better name for listen that would make more sense to you? wait?

Noo, `listen` is great!, my confusion was due to no docs.

Dirbaio (approved):
Niiiice!! Thanks for your hard work!

bors r+

PR #1057, opened 2022-11-14
stm32g0: Fix ADC for channels above 14

jaxter184:
using the CHSELR register in sequence mode does not support ADC channels above 14. Also, it seems like the sequencer itself wasn't being used anyway, so I turned it off (maybe the whole block from L72..L76 could be removed?) and used a bit shift.


Dirbaio (approved):
Thank you!

bors r+

PR #1059, opened 2022-11-15
embassy-rp: Add basic ADC module

kalkyl:
Oneshot ADC

### DIFF
--- a/embassy-rp/src/adc.rs
+++ b/embassy-rp/src/adc.rs
@@ -120,14 +120,9 @@ impl<'d> Adc<'d> {
 
     pub fn blocking_read<PIN: Channel<Adc<'d>, ID = u8>>(&mut self, _pin: &mut PIN) -> u16 {
         let r = Self::regs();
-        let ch = PIN::channel();
         unsafe {
-            if ch == 4 {
-                r.cs().modify(|w| w.set_ts_en(true))
-            }
-            while !r.cs().read().ready() {}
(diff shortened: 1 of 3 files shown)


embassy-rp/src/adc.rs
```diff
@@ -0,0 +1,173 @@
...
+        irq.unpend();
+        irq.enable();
+
+        Self { phantom: PhantomData }
+    }
+
+    async fn wait_for_ready() {
+        let r = Self::regs();
+        unsafe {
+            r.inte().modify(|w| w.set_fifo(true));
```
Dirbaio, line 80: This can race with the irq handler. It's OK right now if we only use one interrupt I think, but it might not be in the future. 

It'd be cool to use the "atomic bitset/bitclear" register aliases. We don't have any functions for them for now, but we could. See rp-hal: https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/atomic_register_access.rs

Dirbaio: Actually the ADC only *has* one interrupt so this is guaranteed to not be a problem in the future :rofl: 

Maybe change it to a `.write()` (and in the irq handler too) then? we don't even need RMW :)

Dirbaio (approved):
LGTM other than the commented nit! Feel free to merge yourself. Thank you! :)

bors r+

Dirbaio:
bors cancel
bors d+

Issue #1041, opened 2022-11-03
embassy-lora `Sx126xRadio` pin types


jbeaurivage:
I'm currently trying to use `Sx126xRadio` that #985 has brought. As defined [here](https://github.com/embassy-rs/embassy/blob/b99533607ceed225dd12ae73aaa9a0d969a7365e/embassy-lora/src/sx126x/mod.rs#L26), it seems that there is only one type parameter for all output pins, and one type parameter for all interrupt-input pins:

```rust
impl<SPI, CTRL, WAIT, BUS> Sx126xRadio<SPI, CTRL, WAIT, BUS>
where
    SPI: SpiBus<u8, Error = BUS> + 'static,
    CTRL: OutputPin + 'static,
    WAIT: Wait + 'static,
    BUS: Error + Format + 'static,
{
    pub async fn new(
        spi: SPI,
        cs: CTRL,
        reset: CTRL,
        antenna_rx: CTRL,
        antenna_tx: CTRL,
        dio1: WAIT,
        busy: WAIT,
        enable_public_network: bool,
    ) -> Result<Self, RadioError<BUS>> {
        let mut lora = LoRa::new(spi, cs, reset, antenna_rx, antenna_tx, dio1, busy);
        lora.init().await?;
        lora.set_lora_modem(enable_public_network).await?;
        Ok(Self { lora })
    }
}
```

@lulf and @ceekdee, since you've worked on that PR: Unless I'm mistaken, does this mean that all the `CTRL` pins would need to have the same type, and all the `WAIT` pins likewise? Would it make more sense to have a unique parameter for each of these pins?

lulf:
Both the nRF and STM32 HALs provide an AnyPin type you can create, i.e. using p.PB4.downgrade(). In general, I think we want to have a RadioSwitch parameter for the antenna pins at least, because it can differ from board to board. I think @ceekdee is planning to fix that.

jbeaurivage:
Ah, fair enough. I want to use it with atsamd-hal, but it has a similar DynPin API.

Regarding the antenna pins, I believe my board has a single switch for both TX and RX. How is it recommended to handle that case?

lulf:
Until the API has been refactored, I think you maybe need to wrap your pin in some custom shared pin type containing an inner type with a RefCell<AnyPin> that the driver can use. Since TX and RX is not done at the same time, it should be 'safe' to do this.

[...]

Issue #1023, opened 2022-10-23
Consolidate all LORA physical layer implementations behind one API


ceekdee:
This issue arises out of the discussion regarding the recent addition of an implementation to support the Semtech sx126x LORA chips:

https://github.com/embassy-rs/embassy/pull/985#discussion_r983568457

I hope to start working on this project in December, or to contribute to other initiatives that start before then.  In the meantime, this issue serves as a launchpad for a discussion on expectations developers have for a LORA physical layer implementation and API that supports Semtech LORA chips and MCU variants.  It would be most helpful if envisioned use cases can be added here to shape the discussion.

My personal focus at this time is on a point-to-point (P2P) implementation that requires only the LORA physical layer.  However, I see great value in interfacing the physical layer implementation with - for example - the Drogue IoT project.

lulf:
Commenting here instead of #1041

> A brief synopsis of architectural thoughts to this point for the implementation in issue 1023. A bit premature, but I will be unavailable to work on an implementation until the end of November and want to encourage other thoughts.

> * produce a trait that describes LoRa API functions that are prevalent across Semtech chips (the base API, largely present in the sx126x implementation - https://github.com/embassy-rs/embassy/blob/master/embassy-lora/src/sx126x/sx126x_lora/mod.rs - which closely followed the Semtech C implementation);
> * produce additional traits for LoRa API functions unique to different flavors of Semtech chips;
> *implement enums and structs which support the various concerns represented across the full API, making each as generic as possible to support present and future LoRa chips (this will build on the separation of concerns in the stm32wl driver);
> *  implement an orchestrator for each MCU/LoRa chip combination. For example, the rak4631 will have an orchestrator, as will a STM32 nucleo board;
> * each orchestrator is a lightweight implementation which grabs the device-specific pins and SPI bus and assembles the necessary enums and structures to support the MCU/LoRa chip combination;
> * the LoRa generic and board-specific API for an MCU/LoRa chip combination will be accessed through the orchestrator.


I agree on the approach. However, in terms of where this happens, I think we should leave board-specifics out of embassy-lora itself, and perhaps surface the orchestrator concept, at least initially, in a separate create (some kind of BSP?). Most likely outside the embassy main repo? I think what you're describing is not only a solution for lora specifically, but a pattern that applies to other types of networking.

[...]

PR #1064, opened 2022-11-21
Fix LoRaWAN PHY settings for SX126x driver

jbeaurivage:
While working on #1023 / #1041, I noticed that the `lorawan_device::PhyTxRx` implementation does not conform to the LoRaWAN standard, and therefore devices using this driver could never communicate with a gateway. This PR backports the changes I've made to fix the offending parameters, and I can confirm that the driver now works with LoRaWAN networks.

* Set preamble length to 8 symbols
* Set polarity to inverted for received messages


lulf (approved):
Thanks for taking the time to look into and fix it 🚀 

bors r+

PR #1065, opened 2022-11-22
embassy-nrf: Default disable UARTE (nrf9160)

kalkyl:
Uarte is enabled by default on the nrf9160, which is both bad for power consumption and renders the other "shared" peripherals unusable. This might be an SPM bug, but had the same issue with all pre-compiled SPM:s available out there, so adding this fix until we figure out.


Dirbaio (approved):
Thanks!
bors r+

PR #1068, opened 2022-11-22
Add Default to some types

mkj:
These are a couple of places I've found `Default` to be handy


Dirbaio (approved):
Makes sense. Thank you!

bors r+

PR #1071, opened 2022-11-22
refactor: autodetect macro variant

lulf:
Apply heuristics using target_arch, target_os and target_family to determine which variant of the entry point to use.


Dirbaio (approved):
Awesome! Thanks for working on it! :D

bors r+

PR #1072, opened 2022-11-23
Add size of bytes received as return param for uart reads

miathedev:
Hi fellow ambassador Rustoceans :crab:!

Just a minor change,

for being able to implement a tty cli using embassy, the information how many bytes have been received is important.
The information is actually available, but not forwarded to the user. Now it is.

With kind regards,
Mia


lulf:
In this case though, the number of bytes received will always be buffer.len(), because idle line detection is not used. We could change the API I guess, but it doesn't give any info the user of the API doesn't already have.

PR #1074, opened 2022-11-23
Added blinky example for stm32f0

imrank03:
Hi, I have added **blinky** example for `stm32f0` and tested with Nucleo board `STM32F091RC`.

- Can I add more example for stm32f0?

### DIFF
--- a/examples/stm32f0/.cargo/config.toml
+++ b/examples/stm32f0/.cargo/config.toml
@@ -1,5 +1,5 @@
 [target.thumbv6m-none-eabi]
-runner = 'probe-run --chip STM32F030F4Px'
+runner = 'probe-run --chip STM32F091RCTX'
 
 [build]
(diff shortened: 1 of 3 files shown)


lulf:
Hey, thanks for the PR! :rocket: 

I think the example is fine, but I noticed it's a different board being set, I'm unsure if we should keep the existing target or adopt a new one.

imrank03:
Hello @lulf Thanks for the response.
Yes the target board is different I don't have the `stm32f030f4` for testing, I have the `stm32f091 Nucleo board` and I have written many examples for it and if you are okay I will push the commit.

lulf:
I think that's fine, from the commit history there doesn't seem to be anyone pushing for a different board, so I think if you want to add some more examples it makes sense to switch the board for the examples.

bors r+

PR #1075, opened 2022-11-23
fix: add required metadata for embassy-boot

lulf:


Dirbaio:
LGTM! Could you enable the `defmt` feature (for both embassy-docs and docs.rs)? This way docs show which structs derive defmt::Format.

PR #1078, opened 2022-11-27
usb: dispatch vendor requests

ReinoutHeeck:
I needed support for USB vendor requests, so here's a patch.

Vendor requests are free-form, so Embassy cannot interpret the recipient and index fields. 
Therefor all vendor requests are routed to interface 0.


Dirbaio:
Interesting. What protocol/class are you implementing that requires vendor requests?

Supporting vendor requests sounds good to me. Forwarding them to interface 0 is strange though. If the class that needs to handle them doesn't get registered first, it won't get interface ID 0 and things will silently not work.

I think it'd make more sense for a class to be able to set a "vendor control handler". I guess only one is enough, so it could error if when setting a handler when one is already set. This way, a class that needs it will either get it, or error out, but never silently not work.

ReinoutHeeck:
> Interesting. What protocol/class are you implementing that requires vendor requests?

WebUSB - but really any any USB gadget that does not fit 100% inside the predefined classes needs access to this protocol extension point, it is the go-to solution for things that don't fit.


> Forwarding them to interface 0 is strange though. If the class that needs to handle them doesn't get registered first, it won't get interface ID 0 and things will silently not work.

Other USB stacks handle this by having all classes get to see all control requests and have each filter out what is intended for each specifically. It works well enough there, so why not here - it also fits the current dispatch logic nicely, no surprises so no cognitive overhead ;-)

> I think it'd make more sense for a class to be able to set a "vendor control handler". 

That would create overhead that all other USB gadgets will need to lug around, even if they don't need it. Maybe that is why other stacks implemented it this lighter / less engineered way.

Dirbaio:
oh! Delivering it to all interfaces (not just interface 0) sounds good as well.

ReinoutHeeck:
> oh! Delivering it to all interfaces (not just interface 0) sounds good as well.

YAGNI :-) 

Just sending it to each class gives a point to pick it up. Where a class needs to distribute a request  to all interfaces it is always possible to add that extra level of dispatch overhead / churn locally only. Which requests would you expect to distribute to all interfaces?

Dirbaio:
#1203 decouples the control handlers from the interfaces, and routes vendor requests to all handlers.

The problem with the previous implementation was that there's no way to "send it to all classes", hence why I suggested sending it to all interfaces.  But yeah, you made me realize the root issue is it doesn't make sense to tie ControlHandlers to interfaces, since they also have to handle stuff that doesn't belong to one interface in particular.

I'm going to close this in favor of #1203, but thanks anyway for the contribution!

PR #1079, opened 2022-11-27
Async function in trait cleanup

yodaldevoid:
Some issues I ran across after the AFIT stuff was merged.


Dirbaio (approved):
Thanks for the fixes!

Unfortunately they slipped through because CI wasn't testing these features at all. I've added them.

bors r+

PR #1081, opened 2022-11-28
Fix GPIO pull ups on RP2040

ExplodingWaffle:


Dirbaio:
`.write()` defaults all bits to zeros (unlike `.modify()` which does a read-modify-write) so as far as I can tell, the old code and the new one are equivalent.

Are you seeing an issue on actual hardware?

ExplodingWaffle:
Whoops- I've just realised what went wrong. Yes I thought I was having a hardware issue, turns out I simplified my test case too hard, and didn't let the pin I thought should have a pull up [live long enough](https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/gpio.rs#L587) so it was getting it's pullup cleared. Sorry for wasting your time 🙃

PR #1082, opened 2022-11-28
stm32: Add basic support for DMA priority settings

matoushybl:
This adds very basic support for specifying priority for DMA interrupts. Unfortunately, the patch now doesn't allow for specifying different priorities for DMA1/DMA2, or BDMA1/BDMA2, which I didn't know how to support.


lulf (approved):
Looks like a nice way to do it

bors r+

PR #1083, opened 2022-11-28
stm32: Fix H7 unaligned erase

matoushybl:
This PR simplifies erasing sectors on the H7, which was buggy.


lulf:
Thanks for fixing :rocket: 
bors r+

PR #1086, opened 2022-11-30
rp: Add an RngCore impl based on ROSC.RANDOMBIT

yodaldevoid:
This has the potential to not be random, but it should not be an issue if default clock settings are used.

### DIFF
--- a/embassy-rp/src/clocks.rs
+++ b/embassy-rp/src/clocks.rs
@@ -5,7 +5,7 @@ use crate::{pac, reset};
 const XOSC_MHZ: u32 = 12;
 
 /// safety: must be called exactly once at bootup
-pub unsafe fn init() {
+pub(crate) unsafe fn init() {
     // Reset everything except:
     // - QSPI (we're using it to run this code!)
     // - PLLs (it may be suicide if that's what's clocking us)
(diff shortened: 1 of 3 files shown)


embassy-rp/src/clocks.rs
```diff
@@ -196,3 +196,40 @@ unsafe fn configure_pll(p: pac::pll::Pll, refdiv: u32, vco_freq: u32, post_div1:
...
+///
+/// This will not produce random values if the ROSC is stopped or run at some
+/// harmonic of the bus frequency. With default clock settings these are not
+/// issues.
+pub struct RoscRng;
+
+impl RoscRng {
+    fn next_u8() -> u8 {
+        let random_reg = pac::ROSC.randombit();
+        let mut acc = unsafe { random_reg.read().randombit() as u8 };
```
Dirbaio, line 210: to avoid reading the reg in 2 places, you could start with `0` and loop for `0..u8::BITS`

yodaldevoid: I was about to say that that would not work due to too many shifts, but I just realized that's only the case if the shift is after the access. Once again I am reminded that coding should not be done while sleepy.

Dirbaio (approved):
Thank you!

bors r+

yodaldevoid:
> Thank you!
> 
> bors r+

I realized at the last minute that I left the loop as `1..u8::BITS`, but thankfully caught it before this was merged.

See previous statement about coding while sleepy.

PR #1088, opened 2022-11-30
stm32: Enable fifo for buffered uart

matoushybl:
This PR enables fifo for buffered uart where it is available. This should hopfully get rid of some overrun errors. I tried it in my application where it worked, but more intensive testing is probably required.


lulf:
Why not, lets try it :rocket: 

bors r+

PR #1089, opened 2022-12-01
feat: embassy-boot for rp2040

lulf:
Add embassy-boot support for RP2040, with examples for the Raspberry Pi Pico.

### DIFF
--- a/examples/boot/application/rp/memory.x
+++ b/examples/boot/application/rp/memory.x
@@ -1,0 +1,15 @@
+MEMORY
+{
+  /* NOTE 1 K = 1 KiBi = 1024 bytes */
+  BOOT2                             : ORIGIN = 0x10000000, LENGTH = 0x100
+  BOOTLOADER_STATE                  : ORIGIN = 0x10006000, LENGTH = 4K
+  FLASH                             : ORIGIN = 0x10007000, LENGTH = 512K
+  DFU                               : ORIGIN = 0x10087000, LENGTH = 516K
+  RAM                               : ORIGIN = 0x20000000, LENGTH = 256K
+}
+
+__bootloader_state_start = ORIGIN(BOOTLOADER_STATE) - ORIGIN(BOOT2);
+__bootloader_state_end = ORIGIN(BOOTLOADER_STATE) + LENGTH(BOOTLOADER_STATE) - ORIGIN(BOOT2);
+
+__bootloader_dfu_start = ORIGIN(DFU) - ORIGIN(BOOT2);
+__bootloader_dfu_end = ORIGIN(DFU) + LENGTH(DFU) - ORIGIN(BOOT2);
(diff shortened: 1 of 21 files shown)


lulf:
@danbev FYI ^^

MathiasKoch:
Lol.. didn't see this before my PR landed.. Doh!

MathiasKoch:
`PAGE_SIZE` & `WRITE_SIZE` shouldn't be needed as generics on the bootloader for rp, as those are given by the implementation, and re-exported by `embassy-rp`

lulf:
@MathiasKoch Ah, I wasn't aware you had this in the works either lol. I have verified that the example works on the rpi pico. I'm fine with whatever PR being merged really, I have done some doc updates here as well, and I'll fix the page and write size.

MathiasKoch:
I think if you change the page and write size, your PR looks more complete than mine :+1: 

I really just wanted to upstream this, to reduce the amount of "custom" code we have to maintain, now that I had a minute to do it :+1: 

Will just look through your implementation, and close my PR in favor of this. Great work! :+1:

examples/boot/application/rp/memory.x
```diff
@@ -0,0 +1,15 @@
+MEMORY
+{
+  /* NOTE 1 K = 1 KiBi = 1024 bytes */
+  BOOT2                             : ORIGIN = 0x10000000, LENGTH = 0x100
```
MathiasKoch, line 4: Why have boot2 defined for the application?

examples/boot/bootloader/rp/memory.x
```diff
@@ -0,0 +1,19 @@
+MEMORY
+{
+  /* NOTE 1 K = 1 KiBi = 1024 bytes */
+  BOOT2                             : ORIGIN = 0x10000000, LENGTH = 0x100
+  FLASH                             : ORIGIN = 0x10000100, LENGTH = 24K
+  BOOTLOADER_STATE                  : ORIGIN = 0x10007000, LENGTH = 4K
```
MathiasKoch, line 6: why `0x10007000` with a bootloader of `24K`? shouldn't `0x10006000` suffice, if you do
`FLASH                             : ORIGIN = 0x10000100, LENGTH = 24K - 0x100`

MathiasKoch (approved):

examples/boot/application/rp/memory.x
lulf: It's used in the linker env settings below, and I think it's great for documentation. Yes you can avoid it, but I think for an example it's useful to leave as much information in there as possible.

examples/boot/bootloader/rp/memory.x
lulf: Agreed!

lulf (commented):
I've addressed the comments and pushed an update.

PR #1090, opened 2022-12-02
feature(embassy-boot): Add implementation & example of embassy-boot for rp2040 (embassy-rp)

MathiasKoch:
Currently, untested in this particular configuration, but is a copy-paste of the implementation we successfully use on our devices.

If someone has an RP Pico handy, a simple test would be extremely welcomed as part of reviewing.


MathiasKoch:
Closing again in favor of the overlooked #1089 :)

PR #1093, opened 2022-12-04
Add random example

miathedev:
Thanks Lulf for the help!

This should be added as example so other people can look it up easily.

With love,
Mia


barafael:
Hi, I think it might be useful to comment what the register setting in your unsafe code block (line 18) does?

PR #1098, opened 2022-12-06
Reset sdmmc clock on card init

chemicstry:
Set clock back to 400kHz in `init_card()`, which allows card reinitialization if it was removed


Dirbaio (approved):
Thanks!

bors r+

PR #1101, opened 2022-12-07
[stm32] Enable flash cache by default

henrikssn:


Dirbaio:
closing since F3 RCC is already enabling prefetch, and f4/f7 is already enabled by default by the hardware.

PR #1102, opened 2022-12-09
usb: Make driver generic over max interfaces

yodaldevoid:
It would be nice if a default value could be used so old code would mostly just work, but defaults are not used while determining types so adding a default does nothing here.

It would also be nice if we could calculate the number of interfaces needed at compile time, but I'll leave that as an exercise for later.

https://github.com/rust-lang/rust/issues/95486
https://github.com/rust-lang/rust/issues/99727
https://github.com/rust-lang/rust/issues/27336

### DIFF
--- a/embassy-usb-logger/src/lib.rs
+++ b/embassy-usb-logger/src/lib.rs
@@ -67,7 +67,7 @@ impl<const N: usize> UsbLogger<N> {
         config.device_protocol = 0x01;
         config.composite_with_iads = true;
 
-        let mut builder = Builder::new(
+        let mut builder = Builder::<'d, D, 4>::new(
             driver,
             config,
             &mut state.device_descriptor,
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -4,7 +4,7 @@ use crate::control::ControlHandler;
 use crate::descriptor::{BosWriter, DescriptorWriter};
 use crate::driver::{Driver, Endpoint, EndpointType};
 use crate::types::*;
-use crate::{DeviceStateHandler, Interface, UsbDevice, MAX_INTERFACE_COUNT, STRING_INDEX_CUSTOM_START};
+use crate::{DeviceStateHandler, Interface, UsbDevice, STRING_INDEX_CUSTOM_START};
 
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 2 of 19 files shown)


embassy-usb-logger/src/lib.rs
```diff
@@ -67,7 +67,7 @@ impl<const N: usize> UsbLogger<N> {
         config.device_protocol = 0x01;
         config.composite_with_iads = true;
 
-        let mut builder = Builder::new(
+        let mut builder = Builder::<'d, D, 2>::new(
```
lulf, line 70: Why hardcoding this to '2'?

yodaldevoid: `CdcAcmClass` only requires two descriptors to initialize. I did however forget that the old default was 4, and `CdcNcmClass` requires four descriptors.

alexmoon:
I’m not a big fan on the generics/turbofish spam this generates. Could you use an associated const in the driver to define the max interfaces a driver supports and use that in the builder? The builder should be short lived and on the stack so a little extra memory usage shouldn’t be a big problem there. The build method could then be generic on the number of interfaces for the device with a default value.

yodaldevoid:
> I’m not a big fan on the generics/turbofish spam this generates. Could you use an associated const in the driver to define the max interfaces a driver supports and use that in the builder? The builder should be short lived and on the stack so a little extra memory usage shouldn’t be a big problem there. The build method could then be generic on the number of interfaces for the device with a default value.

I am also not happy with the noise this creates, though it should only require one annotation per device. It is amplified here due to the number of examples in this project that use USB. The only way I can think to decrease this noise in the typical case where a user doesn't need many interfaces and doesn't care about minimizing the number of interfaces is to set a default, but as stated in the top post this doesn't work at the moment as we would like.

As for your suggestion, the maximum number of interfaces isn't dependent on the device but rather how much memory you want to give to store descriptors and interface handlers.

alexmoon:
> As for your suggestion, the maximum number of interfaces isn't dependent on the device but rather how much memory you want to give to store descriptors and interface handlers.

Ah, right. I was thinking of endpoints not interfaces. Oh well. In that case I can't think of a good alternative to avoid the generic spam and still allow arbitrary numbers of interfaces without excess memory consumption. So this looks ok to me.

alexmoon:
@Dirbaio suggested on matrix that we use features to allow for compile time configuration of the max interface value rather than const generics to avoid the generics spam. For example: smoltcp-rs/smoltcp#742

Dirbaio:
I'm going to close this PR in favor of #1202, which does that. Thanks anyway for the contribution! :)

PR #1103, opened 2022-12-09
embassy-stm32: add rs485 driver enable to uart

FrozenDroid:


FrozenDroid:
Might want to be able to set the driver enable timings using DEAT and DEDT at some point, but this is good enough for now for us.

matoushybl:
Thanks!
bors r+

PR #1104, opened 2022-12-09
embassy-sync: Add new Signal inspectors

zeroeightysix:
This PR adds `idle` and `waiting` to Signal, serving as complements to the existing `signaled` to inspect Signal state.

### DIFF
--- a/embassy-sync/src/signal.rs
+++ b/embassy-sync/src/signal.rs
@@ -65,6 +65,20 @@ where
     }
 }
 
+macro_rules! state_matches {
+    ($self:ident, $pattern:pat_param) => {
+        $self.state.lock(|cell| {
+            let state = cell.replace(State::None);
+
+            let res = matches!(state, $pattern);
+
+            cell.set(state);
+
+            res
+        })
+    };
+}
+
 impl<M, T: Send> Signal<M, T>
 where
     M: RawMutex,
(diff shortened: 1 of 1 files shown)


Dirbaio:
What's the use case for knowing whether a signal is being waited on?

Also, note that the "waiting" state just means "has a waker". If you wait with `.wait().await` and then cancel it, the waker does *not* go away, so `.waiting()` will still return `true`, until it gets signaled.

Exposing this feels very "implementation detail" to me, I don't think we should do it unless there's a very strong use case for it.

zeroeightysix:
I want to use it in my key handler, to determine whether or not a key press should be handled by the "key inhibitor", or a more specific per-key `Signal`. Something like this:

```rust
pub fn press_key(k: Key) {
    if INHIBITOR.waiting() {
        INHIBITOR.signal(k);
    } else {
        get_key_signal(k).signal(())
    }
}
```

I hadn't considered that the `Waiting` state isn't always 'valid' - oops!

I could also just have `INHIBITOR` be something like `Mutex<Option<Signal>>`, but because `Signal` already stores some kind of 'none/something' state, I thought these methods would better fill the job.

Dirbaio:
I assume you're no longer interested in the feature as currently designed, because of the `Waiting `issue? I'm going to close this for now

PR #1105, opened 2022-12-10
embassy-rp: Add multicore support

kalkyl:
This PR adds multicore support + critical-section impl using hardware spinlocks.
Based on the rp2040-hal implementation.

### DIFF
--- a/embassy-rp/src/multicore.rs
+++ b/embassy-rp/src/multicore.rs
@@ -1,0 +1,266 @@
+//! Multicore support
+//!
+//! This module handles setup of the 2nd cpu core on the rp2040, which we refer to as core1.
+//! It provides functionality for setting up the stack, and starting core1.
+//!
+//! The entrypoint for core1 can be any function that never returns, including closures.
+
+use core::mem::ManuallyDrop;
+use core::sync::atomic::{compiler_fence, Ordering};
+
+use crate::pac;
+
+/// Errors for multicore operations.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    /// Operation is invalid on this core.
+    InvalidCore,
+    /// Core was unresponsive to commands.
+    Unresponsive,
+}
+
+/// Core ID
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum CoreId {
+    Core0,
+    Core1,
+}
+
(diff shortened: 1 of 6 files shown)


yodaldevoid:
Should the flash write functionality be updated pause core 1 in this PR, left to a follow up PR, or left as an exercise for the user?

kalkyl:
> Should the flash write functionality be updated pause core 1 in this PR, left to a follow up PR, or left as an exercise for the user?

Ahh, for `in_ram`? I'll look into it

embassy-rp/src/multicore.rs
```diff
@@ -0,0 +1,343 @@
...
+const PAUSE_TOKEN: u32 = 0xDEADBEEF;
+const RESUME_TOKEN: u32 = !0xDEADBEEF;
+static IS_CORE1_INIT: AtomicBool = AtomicBool::new(false);
+
+/// Errors for multicore operations.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    /// Core was unresponsive to commands.
+    Unresponsive,
```
Dirbaio, line 50: IMO this could be a panic, because there's no way this happens in normal usage, right? Like, if it happens, something in the HAL or the silicon is very broken and the user can't do anything about it anyway.

kalkyl: True!

embassy-rp/src/multicore.rs
```diff
@@ -0,0 +1,343 @@
...
+            cortex_m::interrupt::enable();
+            // Signal to CORE0 that execution is resumed
+            fifo_write(RESUME_TOKEN);
+        }
+    }
+}
+
+impl Core1 {
+    /// Spawn a function on this core
+    pub fn spawn<F>(&mut self, stack: &'static mut [usize], entry: F) -> Result<(), Error>
```
Dirbaio, line 148: this could take `stack: &'static mut Stack<N>`, so that it enforces it's aligned.

Dirbaio:
Could you add a HIL test? Nothing fancy, maybe something like "core0 pings core1, core1 pings back, core0 exits with test success"

kalkyl:
> Could you add a HIL test? Nothing fancy, maybe something like "core0 pings core1, core1 pings back, core0 exits with test success"

Check!

Dirbaio (approved):
This is great! Excited to finally have multicore support! :rocket:

PR #1111, opened 2022-12-13
fix net README.md

overheat:
examples/std/src/bin/net.rs connects to TCP port 8000, you need to have something listening there. For example nc -l 8000

Besides, need to add `-- --static-ip` becuase `nc` not support DHCP.


Dirbaio (approved):
Thank you!

PR #1117, opened 2022-12-18
modify RP2040 adc example to get inside biased bipolar diode voltage and convert to Celsius degree.

overheat:
modify RP2040 adc example to get inside biased bipolar diode voltage, then convert this temperature sensor data into Celsius degree, according to chapter 4.9.5. Temperature Sensor in RP2040 datasheet.

PR #1119, opened 2022-12-20
add convert_to_celsius function in the adc module

overheat:
modify RP2040 adc example to get inside biased bipolar diode voltage,
    then convert this temperature sensor data into Celsius degree,
    according to chapter 4.9.5. Temperature Sensor in RP2040 datasheet.


overheat:
#1117

lulf (approved):
Thanks!

bors r+

kalkyl:
Looks like CI fails because you need to run `cargo fmt` for your info...

overheat:
Sure.

PR #1122, opened 2022-12-22
embassy-rp: Add split() to BufferedUart

kalkyl:


examples/rp/.cargo/config.toml
```diff
@@ -5,4 +5,4 @@ runner = "probe-run --chip RP2040"
 target = "thumbv6m-none-eabi"        # Cortex-M0 and Cortex-M0+
 
 [env]
-DEFMT_LOG = "trace"
```
Dirbaio, line 8: Why change the log level?

kalkyl: Trace just totally overflooded the console from the atomic-ring-buffer logs and then i thought since the examples don't depend on trace logs it might be a better default? But happy to change if you disagree :)

Dirbaio: True. Perhaps `debug`?

Dirbaio:
bors d+

examples/rp/.cargo/config.toml
kalkyl: Good idea! I'll change to that

PR #1123, opened 2022-12-23
[WIP] Icmp and ping example.

overheat:
I try to implement the icmp in `embassy-net` and a example in `std/net_icmp.rs`, but a compile error I cannot finger it out, because I think I did the same as UDP wrapper and they are have almost the same `new` method.

Please give me some advises. Thanks.

```
$ cargo c --bin net_icmp
...
error[E0308]: arguments to this function are incorrect
  --> src/bin/net_icmp.rs:82:22
   |
82 |     let mut socket = IcmpSocket::new(stack, &mut rx_meta, &mut rx_buffer, &mut tx_meta, &mut tx_buffer);
   |                      ^^^^^^^^^^^^^^^
   |
...
```

### DIFF
--- a/embassy-net/Cargo.toml
+++ b/embassy-net/Cargo.toml
@@ -20,6 +20,7 @@ defmt = ["dep:defmt", "smoltcp/defmt"]
 nightly = ["dep:embedded-io", "embedded-io?/async", "dep:embedded-nal-async"]
 unstable-traits = []
 
+icmp = ["smoltcp/socket-icmp"]
 udp = ["smoltcp/socket-udp"]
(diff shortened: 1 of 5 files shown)


Dirbaio:
`PacketMetadata` is a different type for UDP and ICMP. The exported one is [for UDP](https://github.com/embassy-rs/embassy/blob/1bd6c954c23b16041b382243a844a53727f6cc9c/embassy-net/src/lib.rs#L42).

The ideal fix I think is to remove that `pub use` and move it to the socket modules, so that there's both `embassy_net::udp::PacketMetadata` and `embassy_net::icmp::PacketMetadata`.

PR #1124, opened 2022-12-23
Fix two SPI bugs for stm32

### LINKED ISSUE #1095: stm32 HAL DMA SPI read broken
Performing a transfer in place correctly reads data from the sensor. Doing the equivalent with a write and then read sends/receives the same bytes over the wire, but the `read` call returns the wrong data.

```
    // In my tests this call was actually done twice, as shown in the logic analyzer trace below
    let mut buf = [0u8, 0];
    embedded_hal_async::spi::SpiDevice::transfer_in_place(&mut spi_dev_1, &mut buf).await.unwrap();
    println!("{:?}", buf);

    let mut buf = [0u8];
    embedded_hal_async::spi::transaction!(&mut spi_dev_1, move |bus| async move {
        bus.write(&buf).await?;
        bus.read(&mut buf).await
    }).await.unwrap();
    println!("{:?}", buf);
```

```
// Output (note the third read, done with a write then read, returns 0 rather than 8)

0.000213 [14, 8]
└─ nucleo_f446re_embassy::____embassy_main_task::{async_fn#0}::{closure#0} @ src/main.rs:62
0.000549 [14, 8]
└─ nucleo_f446re_embassy::____embassy_main_task::{async_fn#0}::{closure#0} @ src/main.rs:66
0.000915 [0]
└─ nucleo_f446re_embassy::____embassy_main_task::{async_fn#0}::{closure#0} @ src/main.rs:73
```

Note each transaction sends/receives the same bits:

image

rmja:
This PR fixes two bugs:
* It fixes #1095 by ensuring that pin speed is VeryHigh for all spi versions. I am on stm32f429 which seems to be spi_v1, and it also needs the VeryHigh pin speed. Otherwise bit errors on the "last bit in every byte" can happen.
* It also fixes a lifetime bug for the tx buffer when sending "write_repeated". The issue can be seen when doing spi.write where the clock byte changes during a transmission because the buffer handled to the dma must live throughout the entire transfer.

### DIFF
--- a/embassy-stm32/src/dma/mod.rs
+++ b/embassy-stm32/src/dma/mod.rs
@@ -59,7 +59,7 @@ pub(crate) mod sealed {
         unsafe fn start_write_repeated<W: super::Word>(
             &mut self,
             request: Request,
-            repeated: &[W; 1],
+            repeated: *const [W],
             count: usize,
             reg_addr: *mut W,
             options: TransferOptions,
@@ -246,7 +246,7 @@ mod transfers {
     pub fn write_repeated<'a, W: Word>(
         channel: impl Peripheral<P = impl Channel> + 'a,
         request: Request,
-        repeated: &[W; 1],
+        repeated: *const [W],
         count: usize,
         reg_addr: *mut W,
     ) -> impl Future<Output = ()> + 'a {
--- a/embassy-stm32/src/dma/mod.rs
+++ b/embassy-stm32/src/dma/mod.rs
@@ -59,7 +59,7 @@ pub(crate) mod sealed {
         unsafe fn start_write_repeated<W: super::Word>(
             &mut self,
             request: Request,
-            repeated: *const [W],
+            repeated: *const W,
             count: usize,
             reg_addr: *mut W,
             options: TransferOptions,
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/dma/mod.rs
```diff
@@ -59,7 +59,7 @@ pub(crate) mod sealed {
         unsafe fn start_write_repeated<W: super::Word>(
             &mut self,
             request: Request,
-            repeated: W,
+            repeated: &[W; 1],
```
Dirbaio, line 62: We should use raw pointers for consistency with the other methods: `repeated: *const W`.

(The reason to use raw pointers is that the buffer is required to live for the whole transfer, using references gives the impression that living for the duration of the function call only is enough)

rmja:
@Dirbaio makes sense. I have updated the pr.

Dirbaio:
I meant `*const W`, not `*const [W]`, since it's just one.

rmja:
@Dirbaio whoops, here you go:)

Dirbaio (approved):
Thanks!

bors r+

PR #1127, opened 2022-12-25
clean up doc comment generation

Weshnaw:
I noticed that when I created doc comments for my tasks that the doc comments got included on the inner function but not the outer functions, I personally prefer keeping the documentation as clean as possible so this PR aims to hide the inner function and then add doc comments to the outer function.

The actual changes include:
* adding #[doc(hidden)] onto the `task_inner` function
  * I flip flopped on this one because I could imagine someone may want this in their docs, but decided to include but I think arguments could be made either way
* copy the attributes from `task_inner` to `task_outer`
  * I don't work with proc_macros often so I am not entirely sure if the way I went about it is correct but it seems to work fine
    * specifically: using `parse_quote` to create the `task_outer` as a `ItemFn` then duplicating the attributes from `task_inner` to `task_outer`
  * I also am not sure if it's a good idea to duplicate all attributes over, but I honestly wasn't sure how to just get the just doc comment attributes

OLD doc
NEW doc

### DIFF
--- a/embassy-macros/src/macros/task.rs
+++ b/embassy-macros/src/macros/task.rs
@@ -1,6 +1,7 @@
 use darling::FromMeta;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
+use syn::{parse_quote, ItemFn};
 
 use crate::util::ctxt::Ctxt;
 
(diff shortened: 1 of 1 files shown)


Dirbaio:
I think in the end what we want is a list of "known" attributes with what to do with them (put them on inner, outer, or both), and error out on unknown attributes. Duplicating the attributes should be fine for now though, making doc comments work is a nice improvement.

Thanks!

bors r+

PR #1128, opened 2022-12-25
Add missing SPI pins

pferreir:
The SPI definitions lack the pins which are not accessible on the pico (but are so e.g. on the stamp).


Dirbaio (approved):
Thanks!

bors r+

PR #1130, opened 2022-12-26
USB serial (CDC-ACM) improvements

timokroeger:
* Remove unused call management descriptor
* Set flag for supported capabilities

The rp `usb_serial` example still works with windows hosts.


Dirbaio (approved):
Thanks!

bors r+

PR #1133, opened 2022-12-27
rp: Fill and empty FIFOs in buffered uart interrupt

timokroeger:
Fixes an issue where only the first byte was transmitted. Should improve throughput aswell.


Dirbaio (approved):
Very nice! Thank you! :) 

bors r+

PR #1135, opened 2022-12-29
Add Clone & Copy on embassy_nrf::gpio::Level

mdevlamynck:
This simply adds the Clone and Copy derive traits on embassy_nrf::gpio::Level.


Dirbaio (approved):
Thanks!

bors r+

PR #1138, opened 2023-01-03
embassy-boot (rp): Add WatchdogFlash

kalkyl:


Dirbaio:
Integrating it into the flash feels a bit hacky.

Perhaps we should add support for passing a watchdog into the bootloader? It would be yet another generic param though, so I'm not sure...

kalkyl:
> Integrating it into the flash feels a bit hacky.
> 
> Perhaps we should add support for passing a watchdog into the bootloader? It would be yet another generic param though, so I'm not sure...

Yeah, i just mirroring the functionality of the nrf version here - i've been using it like this myself for a while and just thought i'd upstream it, but i agree it feels a little hacky...

Dirbaio:
Oh, I wasn't aware nrf already did it this way. If so, I'm fine with doing it for rp too for now. We can revisit in the future.

Dirbaio (approved):
bors r+

PR #1139, opened 2023-01-04
Wdt config changes

huntc:
Per commits:

* By passing WDT config around we can control it more easily and promote sharing it between files.

* The memory layout of the s140 crept into a number of memory files, which can cause confusion (well, it did for me!).

* Obtaining the current WDT config is useful so that we do not have to duplicate configurations around the place. A constructor method has been introduced that attempts to return the current running WDT config from the WDT peripheral. The bootloader example has also been updated to show how the watchdog can be obtained and used.


lulf (approved):
LGTM!

bors r+

PR #1140, opened 2023-01-04
feat(stm32): Let uart implement embedded-io Write trait

rmja:


Dirbaio:
The `embedded-io` Read trait is designed for "std-like" data streams, which have some buffer or backpressure mechanism so data received doesn't get lost while you're not actively reading. This means you should be able to, for example, read a line from a text stream by reading byte-by-byte until you encounter `\n`. 

Implementing `Read` for the non-buffered uart does not satisfy that. The rx hardware is stopped while no `.read().await` call is in progress, so bytes received then are very likely to get lost (yes, the hardware has a buffer, but it's tiny, just 1-4 bytes depending on the uart version). The byte-by-byte "read line" would fail, for example.

If you need `embedded-io` `Read` you should use BufferedUart.

This is why the existing code doesn't impl `embedded-io` `Read` on the raw UART, and why we shouldn't add it. We can implement just `embedded-io` `Write` for the raw uart, as that doesn't have this problem.

rmja:
Year, i realized that after i made this PR. That is the reason why i made #1150. It properly implements Read. I will update this PR to only implement Write.

rmja:
@Dirbaio I updated the PR. Let me know if you need me to anymore here. And, please let me know your thoughts on #1150 :)

Dirbaio (approved):
Thanks!

bors r+

PR #1142, opened 2023-01-04
More rp2040 BufferedUart fixes

timokroeger:
* Refactor init code
* Make it possible to drop RX without breaking TX (or vice versa)
* Correctly handle RX buffer full scenario

### DIFF
--- a/embassy-rp/src/uart/buffered.rs
+++ b/embassy-rp/src/uart/buffered.rs
@@ -38,6 +38,38 @@ pub struct BufferedUartTx<'d, T: Instance> {
     phantom: PhantomData<&'d mut T>,
 }
 
+fn init<'d, T: Instance + 'd>(
+    irq: PeripheralRef<'d, T::Interrupt>,
+    tx: Option<PeripheralRef<'d, AnyPin>>,
+    rx: Option<PeripheralRef<'d, AnyPin>>,
+    rts: Option<PeripheralRef<'d, AnyPin>>,
+    cts: Option<PeripheralRef<'d, AnyPin>>,
+    tx_buffer: &'d mut [u8],
+    rx_buffer: &'d mut [u8],
+    config: Config,
+) {
+    let regs = T::regs();
+    unsafe {
+        regs.uartimsc().modify(|w| {
+            w.set_rxim(true);
+            w.set_rtim(true);
+            w.set_txim(true);
+        });
+    }
+
+    super::Uart::<'d, T, Async>::init(tx, rx, rts, cts, config);
+
+    let state = T::state();
+    let len = tx_buffer.len();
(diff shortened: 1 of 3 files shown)


embassy-rp/src/uart/buffered.rs
```diff
@@ -367,40 +305,31 @@ pub(crate) unsafe fn on_interrupt<T: Instance>(_: *mut ()) {
...
-            });
         }
         if ris.oeris() {
             warn!("Overrun error");
-            r.uarticr().write(|w| {
-                w.set_oeic(true);
-            });
         }
+        // Clear any error flags
+        r.uarticr().write(|w| {
```
Dirbaio, line 325: if an error happens between reading `uartimsc` and this write, it'll get lost.

embassy-rp/src/uart/buffered.rs
```diff
@@ -72,66 +100,25 @@ impl<'d, T: Instance> BufferedUart<'d, T> {
...
-        let state = T::state();
-        let regs = T::regs();
-
-        let len = tx_buffer.len();
-        unsafe { state.tx_buf.init(tx_buffer.as_mut_ptr(), len) };
-        let len = rx_buffer.len();
-        unsafe { state.rx_buf.init(rx_buffer.as_mut_ptr(), len) };
-
-        unsafe {
-            regs.uartimsc().modify(|w| {
```
Dirbaio, line 116: this `uartimsc` write is gone in the new code. I think it's still necessary.

embassy-rp/src/uart/buffered.rs
```diff
@@ -155,48 +143,24 @@ impl<'d, T: Instance> BufferedUartRx<'d, T> {
...
-        irq.set_handler(on_interrupt::<T>);
-        irq.unpend();
-        irq.enable();
-
         Self { phantom: PhantomData }
     }
 
     fn read<'a>(buf: &'a mut [u8]) -> impl Future<Output = Result<usize, Error>> + 'a {
         poll_fn(move |cx| {
+            unsafe { T::Interrupt::steal() }.pend();
```
Dirbaio, line 162: We shouldn't always pend the irq, it wastes cpu time. It should be pended only if we we actually popped bytes off the buffer. Or if we want to be more precise, only when we popped bytes which caused the buffer to go from full to not-full.

Or alternatively, on pop enable the rx irq, and let the hardware trigger the irq when data actually arrives.

Same for `fill_buf` and `write`.

embassy-rp/src/uart/buffered.rs
timokroeger: fixed by only clearing the pending flags

Dirbaio (approved):
Thanks!

bors r+

PR #1143, opened 2023-01-04
rp2040: add {tx,rx}-only constructors to UART

pferreir:
As discussed with @henrik-alser on Matrix. I also added an example, feel free to remove it if it's too much.

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -98,7 +98,19 @@ pub struct UartRx<'d, T: Instance, M: Mode> {
 }
 
 impl<'d, T: Instance, M: Mode> UartTx<'d, T, M> {
-    fn new(tx_dma: Option<PeripheralRef<'d, AnyChannel>>) -> Self {
+    /// Create a new DMA-enabled UART which can only send data
+    pub fn new(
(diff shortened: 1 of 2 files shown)


Dirbaio:
For buffereduart, this is already available as `BufferedUart{RT}x::new()`. I don't think we should have 2 different constructors to do the same thing.

They're missing for the non-buffered uart indeed. Could you change them to `Uart{RT}x::new(), like the existing BufferedUart ones?

pferreir:
> They're missing for the non-buffered uart indeed. Could you change them to `Uart{RT}x::new(), like the existing BufferedUart ones?

Right. I was following @henrik-alser's idea of `.new_{t,r}xonly methods`, but if you guys think it's better that way, no problem.

pferreir:
PR Updated.

Dirbaio (approved):
Thanks!

for context: the `_txonly/_rxonly` naming was from before we had `split()`, so the `UartRx/UartTx` structs didn't exist. Now that they do, it's cleaner to put the constructors there since that's what you're creating in the end.

bors r+

pferreir:
Perfect. Thanks for reviewing this so quickly!

PR #799, opened 2022-06-09
WIP: stm32 usb otg

Dirbaio:

PR #1145, opened 2023-01-05
STM32 USB OTG #2

chemicstry:
This is a continuation of #799

The usb serial example is already working!

TODO:
- [x] Add critical sections to registers shared with IRQ
- [x] Fix `disable()`
- [x] ~Implement cable disconnect detection~ - postponed
- [x] Fix `endpoint_set_enabled()`
- [x] ~Endpoint `wait_enabled`~ USB OTG does not have enable delay (?)
- [x] HS internal and HS ULPI PHY - untested


chemicstry:
I think this should be ready for review (I'll mark it when CI succeeds).

Notes:
- Tested only on `NUCLEO-F429ZI` board. Tests on other boards would be very welcome.
- Cable connect (VBUS) detection is not implemented, because it is wildly different for each chip. I suggest leaving this for future improvements.
- External/internal high-speed PHY infrastructure is there, but since I don't have any way to test it, it is likely to not work. At least the internal HS PHY requires additional clock and RCC configuration.
- Internal static `State` size is set to `MAX_EP_COUNT` (8), because `State<{Self::ENDPOINT_COUNT}>` requires `feature(generic_const_exprs)`
- I suggest squashing the merge 😄

Dirbaio:
checking nothing HIL broke...

bors try

Dirbaio:
I've squashed the commits, keeping the the the non-otg fixes separate.

Thank you @chemicstry for your hard work! It's exciting to finally have OTG support :star_struck: 

bors r+

Dirbaio:
Tested working on:
- nucleo-f429zi
- nucleo-f767zi
- nucleo-h755zi
- nucleo-l4a6zg
- nucleo-l4r5zi
- B-U585I-IOT02A

mattico:
I tested this on an STM32H743XIH6 board with an external USB334x PHY. The usb_serial example worked straight away after making the requisite changes:

- interrupt = `OTG_HS`
- `driver::new_hs_ulpi(...pins)`
- `CdcAcmClass::new(max_packet_size: 512)`
- `ep_out_buffer = [0u8; 1024]`

The CDC max packet size must be increased from 64 when in HS mode or else Windows says the configuration descriptor is invalid.

Thanks for working on this @chemicstry!

PR #1147, opened 2023-01-06
Support codesigning in the firmware updater

huntc:
This PR provides a method to verify that firmware has been SHA-512 hashed and signed with a private key given its public key. The implementation provides both [`ed25519-dalek`](https://github.com/dalek-cryptography/ed25519-dalek/blob/main/Cargo.toml) and [`salty`](https://github.com/ycrypto/salty) as the signature verifiers. Either of the `ed25519-dalek` and `ed25519-salty` features is required to enable the functionality from `embassy-boot`.

The `verify_and_mark_updated` method is used in place of `mark_updated` when signing is used via its feature. This avoids the accidental omission of validation where it has been declared as required at compile time. It also keeps the parity of calls at the same number to the previous situation.

The PR permits other types of signature verifiers in the future on the proviso that the [Signature trait](https://github.com/RustCrypto/traits/tree/master/signature) is supported.

Finally, I've updated the CI to include testing `embassy-boot`, which it was doing before. In addition, I've included a unit test for verification based on a `ed25519-dalek` documentation example. This tests both the `dalek` and `salty` implementations.

In terms of code size comparisons, `dalek` adds about 68KiB and `salty` adds about 20KiB. I'm using `salty` myself. I've also tested this out by signing my code with the OpenBSD `signify` utility and then verify it during firmware upload using `salty`.

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -52,6 +52,16 @@ pub enum BootError {
     BadMagic,
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for BootError {
+    fn format(&self, fmt: defmt::Formatter) {
+        match self {
+            BootError::Flash(_) => defmt::write!(fmt, "BootError::Flash(_)"),
+            BootError::BadMagic => defmt::write!(fmt, "BootError::BadMagic"),
+        }
+    }
+}
+
 impl<E> From<E> for BootError
 where
     E: NorFlashError,
@@ -557,6 +567,33 @@ where
         self.state
     }
 }
+/// Errors returned by FirmwareUpdater
+#[derive(Debug)]
+pub enum FirmwareUpdaterError {
+    /// Error from flash.
+    Flash(NorFlashErrorKind),
+    /// Signature errors.
+    Signature(signature::Error),
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for FirmwareUpdaterError {
+    fn format(&self, fmt: defmt::Formatter) {
+        match self {
+            FirmwareUpdaterError::Flash(_) => defmt::write!(fmt, "FirmwareUpdaterError::Flash(_)"),
+            FirmwareUpdaterError::Signature(_) => defmt::write!(fmt, "FirmwareUpdaterError::Signature(_)"),
+        }
+    }
+}
+
+impl<E> From<E> for FirmwareUpdaterError
+where
+    E: NorFlashError,
+{
+    fn from(error: E) -> Self {
+        FirmwareUpdaterError::Flash(error.kind())
+    }
+}
 
 /// FirmwareUpdater is an application API for interacting with the BootLoader without the ability to
 /// 'mess up' the internal bootloader state
@@ -609,7 +646,11 @@ impl FirmwareUpdater {
     /// This is useful to check if the bootloader has just done a swap, in order
     /// to do verifications and self-tests of the new image before calling
     /// `mark_booted`.
-    pub async fn get_state<F: AsyncNorFlash>(&mut self, flash: &mut F, aligned: &mut [u8]) -> Result<State, F::Error> {
+    pub async fn get_state<F: AsyncNorFlash>(
+        &mut self,
+        flash: &mut F,
+        aligned: &mut [u8],
+    ) -> Result<State, FirmwareUpdaterError> {
         flash.read(self.state.from as u32, aligned).await?;
 
         if !aligned.iter().any(|&b| b != SWAP_MAGIC) {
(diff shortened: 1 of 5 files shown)


embassy-boot/boot/src/lib.rs
```diff
@@ -619,12 +640,62 @@ impl FirmwareUpdater {
...
+        assert_eq!(_aligned.len(), F::READ_SIZE);
+        assert!(_end <= self.dfu.to);
+
+        #[cfg(feature = "ed25519")]
+        {
+            use ed25519_dalek::{Digest, PublicKey, Sha512, Signature, SignatureError, Verifier};
+
+            let into_signature_error = |e: SignatureError| FirmwareUpdaterError::Signature(e.into());
+            let public_key = PublicKey::from_bytes(_public_key).map_err(into_signature_error)?;
+            let signature = Signature::from_bytes(_public_key).map_err(into_signature_error)?;
```
lulf, line 670: should probably be from_byte(_signature)?

embassy-boot/boot/src/lib.rs
```diff
@@ -727,12 +809,62 @@ impl FirmwareUpdater {
...
```
lulf, line 839: should probably be from_bytes(_signature)

lulf (commented):
Seems like you're on the right track to me :)

I think doing key management outside the BL as done here is probably the most flexible. I'm wondering if the verification methods should be available for the bootloader as well so users can optionally verify the dfu _before_ attempting to swap and/or load the active? Seems like that way you can also protect against a bad verification in the application.

huntc:
> Seems like you're on the right track to me :)

Thanks!

> 
> I think doing key management outside the BL as done here is probably the most flexible. I'm wondering if the verification methods should be available for the bootloader as well so users can optionally verify the dfu _before_ attempting to swap and/or load the active? Seems like that way you can also protect against a bad verification in the application.

The verification is still happening before the swap of the active… right? It is the active performing the verification and, all being well, then swapping. 

My original idea was to place the verification within the boot loader, but you then have to store the signature and public key in flash. This’ll require another flash region for the boot loader. 

I guess the old program can always screw something up in terms of verification, but then it can screw up anything around interacting with the firmware updater anyway.

I’m not seeing any particular advantage to the boot loader performing the verification presently. Hope I’m not missing something.

huntc:
OK, I've got this working and have written a unit test for it. In addition, I've now got CI including the unit tests for embassy-boot. I'll do some real-world testing on my device on Monday and then ready the PR for a final review.

huntc:
> An alternative could be that if you compile with signatures, the mark_updated() would require the additional parameters for verifying the signature (done before marking it as swapped) to be passed rather than having a separate method for that. This way you'd have to leave out the compiler flag to forget checking the signature (and you'd get a compilation error for not providing the signature if you tried).

Sounds reasonable. I can try that.

huntc:
> An alternative could be that if you compile with signatures, the mark_updated() would require the additional parameters for verifying the signature (done before marking it as swapped) to be passed rather than having a separate method for that. This way you'd have to leave out the compiler flag to forget checking the signature (and you'd get a compilation error for not providing the signature if you tried).

I like this - the API is more streamlined and ergonomic. Thanks for pushing this one. Here's what my project's code is now looking like:

```rust
let mut aligned = [0; 4];

if firmware_updater
    .verify_and_mark_updated(
        &mut nvmc,
        &public_signing_key,
        &firmware_signature.to_be_bytes(),
        firmware_len,
        &mut aligned,
    )
    .await
    .is_ok()
{
    debug!("Updating firmware...");
    cortex_m::peripheral::SCB::sys_reset();
} else {
    debug!("Firmware signature does not verify or flash error... Dropping.");
}
```

embassy-boot/boot/src/lib.rs
```diff
@@ -619,12 +660,120 @@ impl FirmwareUpdater {
...
+    /// The `_read_aligned` buffer must have a size of F::READ_SIZE, and follow the alignment rules for the flash being read from.
+    /// The `_write_aligned` buffer must have a size of F::WRITE_SIZE, and follow the alignment rules for the flash being written to.
+    #[cfg(feature = "_verify")]
+    pub async fn verify_and_mark_updated<F: AsyncNorFlash>(
+        &mut self,
+        _flash: &mut F,
+        _public_key: &[u8],
+        _signature: &[u8],
+        _update_len: usize,
+        _read_aligned: &mut [u8],
```
lulf, line 686: Maybe instead of having to pass both a read_aligned and a write_aligned buffer, pass a single buffer that is write_aligned? Having to pass two buffers requires additional RAM for no reason.

I've not seen a case where write alignment didn't also cover read alignment for nrf, stm32 and rp, so it should be fine.

huntc: `READ_SIZE` appears to be 1 on the nRF.

lulf (commented):
Great work! 🚀 Could you add a note to docs/modules/ROOT/pages/bootloader.adoc about this as well?

huntc:
> Great work! 🚀 Could you add a note to docs/modules/ROOT/pages/bootloader.adoc about this as well?

I promise to do that over the weekend as there a few things to write about; particularly in terms of how to generate and sign code outside of Embassy. Meanwhile, if you approve, can we merge this so I can then merge in my work related PR? Thanks.

lulf (approved):
bors r+

PR #1149, opened 2023-01-08
Add samples for nrf5340

Tiwalun:
Samples for the nrf5340, copied from the existing nrf samples.

Not sure if copying them is the best way of adding support, or using features in the existing samples would be better?
The code is mostly the same, with some different peripherals and pin mappings for the DK.


lulf (commented):
I'm not sure how to think about making nrf5340 separate from nrf. We don't do that for nrf52, nrf52840 etc, so why treat nrf5340 differently from those?

Tiwalun:
So far all the samples seem to be for the nrf52840, I haven't seen anything else. The nrf5340 is a different architecture, and it needs different features of `embassy-nrf`.

I think having a separate folder, similar to the stm32 samples, is easiest to use for users. But it's more effort to maintain the samples, so probably best not to copy all the samples, but just a subset.

lulf:
> So far all the samples seem to be for the nrf52840, I haven't seen anything else. 

I think the reason for that is for maintainability as you say, but I think in this case it would be appropriate to do one of the following alternatives:

* rename examples/nrf to examples/nrf52840 to be consistent with nrf5340
* rename nrf to nrf52, and rename this PR to nrf53
* use cargo features add #[cfg(feature = "nrf5340")] to the examples

Tiwalun:
I think I will remove most of the examples, and just keep two or three, to showcase LEDs and UART on the DK. That should be enough to get started, but hopefully not too much of a maintenance burden.

And I'll rename the existing folder to nrf52840 to make it more consistent. Does that sound good?

lulf:
Sounds good to me @Tiwalun, thanks!

PR #1150, opened 2023-01-12
feat(stm32): Add DMA based, ring-buffer based rx uart

rmja:
This PR adds a ring-buffer based background receive capability to the stm32 uart.

It should be used something like this:
```rust
let mut usart_config = usart::Config::default();
usart_config.baudrate = 115200;
usart_config.data_bits = DataBits::DataBits8;
usart_config.stop_bits = StopBits::STOP1;
usart_config.parity = Parity::ParityNone;

let usart2_irq = interrupt::take!(USART2);
let usart2 = usart::Uart::new(
        p.USART2,
        p.PA3,
        p.PA2,
        usart2_irq,
        p.DMA1_CH6,
        p.DMA1_CH5,
        usart_config,
);
let (_, rx) = usart2.split();
let mut dma_buf = [0; 256];
let mut rx = rx.into_ring_buffered(&mut dma_buf);

// One can now either start the receiver explicitly by calling rx.start()
// Alternatively, the receiver is called on the first call to read()
loop {
    let mut buf = [0; 16];
    let received = rx.read(&mut buf).await.unwrap();
    assert!(received > 0);
}
```

From the last assertion it is given that the call to `read()` only completes when at least one byte is received. If no data is ever received the call will not complete.
This allows for an easy wait of reading from the ring buffer without having to poll.

The implementation works something like this:
* The provided `dma_buf` is registered in the dma controller and the controller is started.
* All calls to `read()` reads from this buffer. If the call to `read()` is not fast enough such that the dma buffer is overwriting undread bytes then an `Error::Overrun` error is returned.
* If there are no bytes currently in the ring buffer then wait for some uart activity and return the just read portion.

It internally uses a `DmaRingBuffer` which uses the TCIF flag and the `NDTR` register to keep track of the available portions within the ring buffer and whether the DMA writer has overflowed.

I have tested the implementation here on my desk and it seems to work just fine. There are also a bunch of `DmaRingBuffer` tests. I do not know how to run these tests within the `embassy-stm32` crate, so the way that I run them is to copy out the `dma_ringbuffer.rs` into an empty lib crate and then run them there. Let me know if this can be done differently.

cc @chemicstry


rmja:
I do not know why the the build fails. Can anyone have a look at this and let me know what I need to change for it to not fail?

embassy-stm32/src/usart/dma_ringbuffer.rs
```diff
@@ -0,0 +1,413 @@
...
+        }
+    }
+
+    /// Copy from the dma buffer at `data_range` into `buf`
+    fn copy_to(&mut self, buf: &mut [u8], data_range: Range<usize>) -> usize {
+        // Limit the number of bytes that can be copied
+        let length = usize::min(data_range.len(), buf.len());
+
+        // Copy from dma buffer into read buffer
+        buf[0..length].copy_from_slice(&self.dma_buf[data_range.start..data_range.start + length]);
```
chemicstry, line 211: It is undefined behavior to use `&[u8]`, which can be overwritten by circular DMA. If I remember correctly from the previous bikeshedding discussions, the only sound solution is to iterate over `&[AtomicU8]` and load each with `Ordering::Relaxed`. Even the `ptr::read_volatile()` is undefined behavior on data that can potentially be overwritten by DMA.

Isolated `copy_from_slice` within driver may be less dangerous than exposing DMA buffer to the user, where anything can happen, but I'm not sure. @GrantM11235 knows about this stuff. For me, it feels like the least "naughty" option would be to fill buffer with:
```
for i in 0..length {
    buf[i] = core::ptr::read_volatile(self.dma_buf.as_ptr().offset(i));
}
```

chemicstry (commented):
Thanks for the PR, that was fast!

I haven't had time to test this yet, but have a couple of general questions:
- I see that circular DMA is not supported by BDMA. Does this exclude some chips that don't have any other DMA except BDMA?
- Have you considered the double-buffer approach? It would allow zero-copy UART.

chemicstry:
> I do not know why the the build fails. Can anyone have a look at this and let me know what I need to change for it to not fail?

I think it's formatting. Did you run `cargo fmt`?

rmja:
Regarding BDMA: I can see that there is another PR proposing circular buffer for bdma here: https://github.com/embassy-rs/embassy/pull/932. Maybe it is possible somehow but just not added in embassy yet.

Yes, I considered the double buffer approach and I actually came quite long in the implementation until I realized that it would not work. My findings are:

* For double buffer mode, both buffers must be the same length. This basically adds the requirement that the driver must own both buffers. A buffer provided by the user cannot be used.
* If we assumed that on each call to read() we swapped the pre-registered buffers, then it seem impossible to safely "1) get the number of bytes currently written to the active buffer, and 2) swap the buffers".  I could not get by head around how to do this safely without a race.
* Additionally I think that an API like `read() -> &[u8]` is more difficult to use because it basically requires the user to be able to use the entire returned slice. When implementing something like the Read trait from embedded-io, then the provided buffer to `read(buf: &mut[u8])` must be large enough to contain the entire slice, or the implementer must have an intermediate buffer that stores the part from the slice that the caller did not have space for. Alternatively the api should be `read(max: usize) -> &[u8]` but then the dma buffers cannot be easily swapped, as both buffers now contain user data.

Does this make sense?

embassy-stm32/src/usart/dma_ringbuffer.rs
rmja: I was not aware of this, thanks! I have updated the PR with the change.

rmja:
@chemicstry To ease the review of this, I have now added a test app and a host utility.

Try and run `usart_rx_ring_buffered` on the device. When it says "Ready to receive..." then run the `saturate_serial <some-port>` utility on the host. This utility writes 0,1,2,...,255,0,1,... and the device verifies that all the received bytes have this pattern. I have added some randomization with delays and buffer sizes to make it as complete as possible.

chemicstry:
Sorry for not replying earlier, I'm a bit busy hopping around different projects.

Thank you for the wide test coverage, they definitely make the ringbuffer implementation less scary. However, I would like to discuss how this integrates into the embassy-stm32 as a whole instead of being an ad-hoc solution.

First problem is the whole circular DMA stuff. There are currently 3 implementations of circular DMA: DCMI (in-tree), #932 (pending) and yours. For example, your `usart/dma_ringbuffer.rs` begs to live in `dma` instead of `usart`, but it depends if anything else except UART will use it. In general, it seems that there are 3 different use cases for circular DMA RX:
1. Single read of large buffer that exceeds 16bit NDTR. Only possible on DMA with M0AR, M1AR. Used by DCMI.
2. Continuous reading of arbitrary length data. Used by UART.
3. Continuous reading of chunked data (buffer swapping). Used by ADC, I2S, SPI (?).

I was thinking that 2 and 3 could be combined, but after looking at the code it no longer seems worth it. So if there are no more use cases for 2 except USART, then maybe the ringbuffer implementation is okay as it is.

The second thing that I see troublesome is 5 different UART implementations: `UartRx`, `UartTx`, `BufferedUartRx`, `BufferedUartTx` and now `RingBufferedUartRx`. This is getting really confusing and I think some of it should be pruned. IMO `BufferedUartTx` is useless (unless you are short on DMA channels?). And `RingBufferedUartRx` is now superior to `BufferedUartRx`. So my suggestion would be to remove old `BufferedUart` and only have `RingBufferedUartRx` as an extension to `UartRx`. What do you think @Dirbaio ?

embassy-stm32/src/usart/rx_ringbuffered.rs
```diff
@@ -0,0 +1,296 @@
...
+        Ok(len)
+    }
+
+    /// Wait for uart data
+    async fn wait_for_data(&mut self, old_ndtr: usize) -> Result<(), Error> {
+        let r = T::regs();
+
+        // make sure USART state is restored to neutral state when this future is dropped
+        let _drop = OnDrop::new(move || {
+            // SAFETY: only clears Rx related flags
```
chemicstry, line 188: This is unsafe, but it seems that the problem is also present in the original UART driver. It doesn't matter if you don't touch TX flags, the read-modify-write (RMW) operation is performed on the whole register and can be interrupted by UART IRQ, which also uses some of these registers. This can also happen if TX/RX parts are used in different executors.

rmja:
I completely agree in your points. The use for `BufferedUart{Rx,Tx}` seems really narrow. I do not have the insights into embassy to judge where the things should go, but I agree that the `usart/dma_ringbuffer.rs` does not have any dependency on the uart, and should therefore probably be in the dma module.

Regarding the use cases - I think that sampling from ADC also could gain from your second point: "continously reading of arbitrary length data".

Also note that it is the `DmaRingBuffer` that controls the clearing of the dma transfer completed interrupt flag. Interrupt for that event must therefore not be enabled! The flag is used to detect overruns.

Also, it would be really nice if it was possible to actually run the tests on the pc. The way that I currently run the tests is to copy out the `dma_ringbuffer.rs` into a new lib and run from there. Any idea on how to do this?

embassy-stm32/src/usart/rx_ringbuffered.rs
rmja: Is there something in embassy that allows this to be done atomically with retry? Drone OS does this: https://github.com/drone-os/drone-core/blob/master/src/reg/atomic.rs#L88-L125

Dirbaio:
okay @rmja after #1348 i've (finally!!) taken a deep look at this! (sorry!)

I've done for you a first attempt to port it to the new DMA architecture here: https://github.com/embassy-rs/embassy/commit/502d1aac76bce8a60754e71f0823548bd7b583e1 . Just changed structure, the logic is unchanged. There's now a dedicated `RingBuffer` DMA driver (alongside the classic `Transfer`). It internally uses the `DmaRingBuffer`, which can be shared across all 3 DMA versions. The HIL test passes on F4.

The overall logic looks great to me. Some thoughts/TODOs:

- It feels to me the logic in `read` and `wait_for_data` could be simplified a bit, there's a bit of duplication. I think it should be possible to do reads with one `poll_fn` roughly like the following. (it might do unnecessary irqs/wakes, but it's OK, the only effect is it hurts perf a bit, we can optimize later if we see it's an issue but I'm pretty sure it won't be).

```rust
poll_fn {
   set waker
   enable rxne
   if self.ringbuf.read() len != 0 {  
      disable rxne
      return Ready
   } else {
      return Pending // rxne irq will wake us
   }
}
```
- The `RingBuffer` DMA driver should be self-contained, hiding all the DMA implementation details. UART code shouldn't have to know abotu `ndtr`.
- Add `RingBuffer` support to BDMA (should be easy)
- Add `RingBuffer` support to GPDMA (If you want, I can take a look at it)

rmja:
@Dirbaio you are awesome. I will have a look and get back!

rmja:
This PR is replaced by https://github.com/embassy-rs/embassy/pull/1404

PR #1153, opened 2023-01-14
Documentation on verifying firmware

huntc:
The documentation has been enhanced to describe the verification of firmware with the firmware updater. Examples have also been provided that describe how keys can be generated and how firmware can be signed.


lulf (approved):
Wow, great stuff 🚀

bors r+

PR #1159, opened 2023-01-18
stm32 usb otg bug fixes

chemicstry:
This fixes a couple of usb otg bugs that surfaced with `usb_ethernet` example from nrf:
- Properly implemented `Endpoint::wait_enabled()`
- Return `EndpointError::Disabled` when neccessary in `Endpoint::write()`


Dirbaio (approved):
Thank you!

bors r+

PR #1160, opened 2023-01-18
Stop sampling when exiting the Saadc methods

huntc:
Prior to this commit, the onDrop function was being dropped immediately and not on exiting the Saadc sampling methods.

A few other places have been corrected also.

Thanks to Peter Hansen for pointing out this issue.


peter9477:
I reviewed these and concur.  Note for posterity and other Rust noobs, the reason the saadc.rs OnDrop wasn't working properly before (confirmed by still seeing high current after using `sample()`) is because by not assigning it to a variable (in a "let" statement), it was just a "temporary" and the scope of a temporary is basically the statement it's within.  That meant it was being dropped immediately after it was created, before any other code in the function ran.

Issue #1154, opened 2023-01-16
Allow to configure the `ConfigStrategy::Dhcp` `smoltcp::socket::dhcpv4::Socket`


Czocher:
Hello,
From what I noticed the `smoltcp::socket::dhcpv4::Socket` is [created as-is](https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/lib.rs#L152) and cannot be further configured. I believe adding some option which would allow the socket to be further configured would be beneficial.

Is this something that I could provide as a PR? Feel free to give any hints you deem necessary.

Otherwise if you're aware of any way to change the socket settings can you share it?

PR #1161, opened 2023-01-18
WIP: Smoltcp socket config

Czocher:
This PR updates the smoltcp version to the newest master one as well as implements the feature specified in #1154 - to allow the dhcpv4 socket to be configured.

Currently it should be considered a WIP PR - requires testing.

@Dirbaio can you have a look and check if this is compatible with what we discussed on the matrix channel?

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -288,11 +288,7 @@ impl<D: Driver + 'static> Inner<D> {
             cx: Some(cx),
             inner: &mut self.device,
         };
-        if !s.iface.poll(timestamp, &mut smoldev, &mut s.sockets) {
-            // If poll() returns error, it may not be done yet, so poll again later.
-            cx.waker().wake_by_ref();
-            return;
-        }
+        s.iface.poll(timestamp, &mut smoldev, &mut s.sockets);
 
         // Update link up
         let old_link_up = self.link_up;
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -23,11 +23,11 @@ use embassy_sync::waitqueue::WakerRegistration;
 use embassy_time::{Instant, Timer};
 use futures::pin_mut;
 use heapless::Vec;
+#[cfg(feature = "medium-ethernet")]
+use smoltcp::iface::Routes;
 #[cfg(feature = "dhcpv4")]
 use smoltcp::iface::SocketHandle;
 use smoltcp::iface::{Interface, InterfaceBuilder, SocketSet, SocketStorage};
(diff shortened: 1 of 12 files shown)


embassy-net/src/lib.rs
```diff
@@ -45,25 +47,16 @@ use crate::device::DriverAdapter;
 const LOCAL_PORT_MIN: u16 = 1025;
 const LOCAL_PORT_MAX: u16 = 65535;
 
-pub struct StackResources<const ADDR: usize, const SOCK: usize, const NEIGHBOR: usize> {
-    addresses: [IpCidr; ADDR],
+pub struct StackResources<const SOCK: usize, const NEIGHBOR: usize> {
```
Dirbaio, line 50: NEIGHBOR is unsued, could now be removed.

embassy-net/src/lib.rs
```diff
@@ -45,25 +47,16 @@ use crate::device::DriverAdapter;
...
+    addresses: [IpCidr; 5],
```
Dirbaio, line 51: `addresses` is now owned by the smoltcp iface, there's no need to have it in this struct (smoltcp iface borrowed it before)

embassy-net/src/lib.rs
```diff
@@ -75,10 +68,34 @@ pub struct Config {
...
+            max_lease_duration: Default::default(),
+            retry_config: Default::default(),
+            ignore_naks: Default::default(),
+            server_port: smoltcp::wire::DHCP_SERVER_PORT,
+            client_port: smoltcp::wire::DHCP_CLIENT_PORT,
+        }
+    }
+}
+
 pub enum ConfigStrategy {
```
Dirbaio, line 95: Now that there's config structs for DHCP and static it makes sense to rename like this so they're more consistent:

Config -> StaticConfig
ConfigStrategy -> Config

embassy-net/src/lib.rs
```diff
@@ -105,21 +122,20 @@ impl<D: Driver + 'static> Stack<D> {
     pub fn new<const ADDR: usize, const SOCK: usize, const NEIGH: usize>(
```
Dirbaio, line 122: ADDR, NEIGH should be removable.

Dirbaio (approved):
Thanks for the PR!

You had some auto-formatting tool that had formatted some Cargo.tomls. I've reverted it for you, please make sure it's disabled next time.

I've further updated it to a newer smoltcp and did some more fixes. I've tested it working, so merging.

bors r+

Czocher:
Thanks @Dirbaio sorry, for the autoformat, I'll take that into account next time!

PR #1162, opened 2023-01-19
rp gpio: make pin_bank() inline

mkj:
This allows set_high() etc to be inlined, toggling pins should be much faster.


Dirbaio (approved):
Thanks!

bors r+

PR #1166, opened 2023-01-20
fix embedded-sdmmc integration.

Dirbaio:
- Renamed feature to `embedded-sdmmc`.
- Move embedded-sdmmc fork repo to the embassy-rs org.
- Remove unused features in the fork
- Fix impl in embassy-stm32
- Add to CI so it doesn't break again.

bors r+


Dirbaio:
bors cancel

PR #1170, opened 2023-01-23
nrf: add support for UICR configuration.

Dirbaio:
- APPROTECT enable/disable. Notably this fixes issues with nrf52-rev3 and nrf53 from locking itself at reset.
- Use NFC pins as GPIO.
- Use RESET pin as GPIO.

NFC and RESET pins singletons are made available only when usable as GPIO, for compile-time checking.

TODO: test
- [x] nrf52 rev1
- [x] nrf52 rev3
- [x] nrf53
- [x] nrf91

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -46,8 +46,8 @@ cargo batch  \
     --- build --release --manifest-path embassy-nrf/Cargo.toml --target thumbv7em-none-eabi --features nightly,nrf52810,gpiote,time-driver-rtc1 \
     --- build --release --manifest-path embassy-nrf/Cargo.toml --target thumbv7em-none-eabi --features nightly,nrf52811,gpiote,time-driver-rtc1 \
     --- build --release --manifest-path embassy-nrf/Cargo.toml --target thumbv7em-none-eabi --features nightly,nrf52820,gpiote,time-driver-rtc1 \
-    --- build --release --manifest-path embassy-nrf/Cargo.toml --target thumbv7em-none-eabi --features nightly,nrf52832,gpiote,time-driver-rtc1 \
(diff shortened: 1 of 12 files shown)


huntc (approved):
Looks good. I just compared it with my code and it appears to correlate. No surprise given that my code was originally derived from your suggested code! :-)

Nice feature. Thanks.

peter9477:
Is the choice of naming correct, or is there a problem with the design concept here (or implementation)?  

My understanding is that the default (at least for nRF52840, but I assume it's the same for others) is for the NFC pins to be in NFC mode (not GPIO), but for the RESET pin to be in GPIO (not act as a reset).  These correspond to the default (flash erased) "1" bits.  

All that can be done without erasing the entire chip is to change those from 1 to 0, meaning that while "nfc-pins-as-gpio" makes sense as a non-optional feature, "reset-pin-as-gpio" would not because it could not actually perform the required change.  Instead I suspect it should be "reset-pin-as-reset", or some such.

Dirbaio:
@peter9477 yes, the default from the hardware's PoV is different for the two flags. That doesn't mean the Cargo feature default has to be the same though. 

I think enabling the reset pin by default is better:
- I've seen a few people ask why isn't reset working in matrix already. 
- If you disable it when you want it enabled, troubleshooting is hard: reset doesn't work with no indication why. If you enable it when you want it disabled, troubleshooting is easy: your firmware won't compile because the pin singleton is unavailable.

I agree it's a footgun that the setting won't apply if UICR is already programmed wrong though! I've added code to check for that, and warn if it's the case. Hopefully that mitigates it!

Tested working on all chips.

bors r+

PR #1173, opened 2023-01-24
nRF examples crates names

davidedellagiustina:
Fixed nRF examples crates' names: they had the same names and they were conflicting during compilation (Cargo warning).


lulf:
thanks!

bors r+

PR #1178, opened 2023-01-27
rp: allow isochronous USB endpoints to be up to 1023 bytes in size

nitroxis:
The datasheet allows isochronous USB endpoints to be up to 1023 bytes in size (see "4.1.2.5. DPSRAM"). This PR changes the check to allow this and also changes the length computation to align to 64 bytes (instead of hardcoded 64 bytes).

Embassy does not yet support isochronous USB endpoints, however I'm investigating adding support. This change is simple enough and should be correct according to the datasheet, so maybe future implementers don't run into this issue.


Dirbaio (approved):
Thanks!

bors r+

PR #1179, opened 2023-01-27
LoRa/SX1276: adjust Rx window offset and duration

samueltardieu:
After a transmission, two receive windows Rx1 and Rx2 are opened for one second each, one right after the other, after a fixed delay (for example 5s). The Rx window offset is added to the starting date of each window and the Rx window duration represents the maximum delay we will wait for an incoming message before declaring that a timeout occurred.

A value of -500ms for the offset and 800ms for the duration means that instead of having Rx1 = [5000, 6000[ and Rx2 = [6000, 7000[ we get Rx1 = [4500, 5300[ and Rx2 = [5500, 6300[. We only cover 30% of the expected windows.

The maximum time a SX127x can take before the Rx side is ready is TS_HOP + TS_RE = 50µs + 2.33ms. Using 3ms for the offset and 1003ms for the duration will give much better time windows: Rx1 = [4997, 5997[ and Rx2 = [5997, 7000]. Note that the
lorawan-device crate caps Rx1 end date to Rx2 start date.

This change allows a previously failing Murata CMWX1ZZABZ-091 module (STM32L + SX1276) to connect to the TTN LoRa network.

### DIFF
--- a/embassy-lora/src/sx127x/mod.rs
+++ b/embassy-lora/src/sx127x/mod.rs
@@ -70,10 +70,10 @@ where
     RFS: RadioSwitch + 'static,
 {
     fn get_rx_window_offset_ms(&self) -> i32 {
-        -500
+        -3
     }
     fn get_rx_window_duration_ms(&self) -> u32 {
-        800
+        1003
     }
 }
 


samueltardieu:
Note that the same issue exists in the SX126x code but I have no way to check the fix so I've left it untouched.

The issue also exists for the STM32WL code, but since there the window duration is set to 3000ms (instead of the expected 1000ms) the Rx2 window is covered in full so the second transmission (during Rx2) will be received. However, the chip will listen on Rx1 frequencies only during the first half of the expected Rx1 window. This makes it very inefficient as it will systematically wait for the Rx2 window, and if there is nothing to receive the radio will be on for a total time of 4s instead of 2s.

lulf:
Thanks for the PR. The current values are derived from a bit of trial-and-error with a STM32 Lora Discovery board. I'm happy to see some more correct numbers. I will give it a try before merging, but looks good in general.

lulf:
Worked for me!

bors r+

PR #1180, opened 2023-01-27
usb: allow adding isochronous endpoints

nitroxis:
This adds (basic) support for isochronous endpoints. In theory, isochronous endpoints have guaranteed bandwidths and are supposed to adhere to strict timings. But it seems that nothing bad happens if you don't follow the specs closely in this regard. Better handling could also still be added in the future.


Dirbaio (approved):
Thanks :rocket: !!

bors r+

nitroxis:
Note that for this to work on the RP2040, [this](https://github.com/embassy-rs/embassy/pull/1178) PR is required, otherwise it might panic because of too big packets.

I have not tested this on any other boards.

PR #1182, opened 2023-01-29
executor: Replace `NonNull<TaskHeader>` with `TaskRef`

GrantM11235:


embassy-executor/src/raw/mod.rs
```diff
@@ -424,9 +451,9 @@ impl Executor {
 /// # Safety
 ///
 /// `task` must be a valid task pointer obtained from [`task_from_waker`].
-pub unsafe fn wake_task(task: NonNull<TaskHeader>) {
+pub unsafe fn wake_task(task: TaskRef) {
```
Dirbaio, line 454: this can be safe now!

Dirbaio (approved):
Thank you! 

bors r+

PR #1185, opened 2023-01-31
Add DNS socket to embassy-net

lulf:
* [x] TODO: Test it :D

### DIFF
--- a/embassy-net/src/dns.rs
+++ b/embassy-net/src/dns.rs
@@ -1,19 +1,10 @@
 //! DNS socket with async support.
-use core::cell::RefCell;
-use core::future::poll_fn;
-use core::mem;
-use core::task::Poll;
-
-use embassy_hal_common::drop::OnDrop;
-use embassy_net_driver::Driver;
 use heapless::Vec;
-use managed::ManagedSlice;
-use smoltcp::iface::{Interface, SocketHandle};
-pub use smoltcp::socket::dns::DnsQuery;
-use smoltcp::socket::dns::{self, GetQueryResultError, StartQueryError, MAX_ADDRESS_COUNT};
+pub use smoltcp::socket::dns::{DnsQuery, Socket, MAX_ADDRESS_COUNT};
+pub(crate) use smoltcp::socket::dns::{GetQueryResultError, StartQueryError};
 pub use smoltcp::wire::{DnsQueryType, IpAddress};
 
-use crate::{SocketStack, Stack};
+use crate::{Driver, Stack};
 
 /// Errors returned by DnsSocket.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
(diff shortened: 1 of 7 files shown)


lulf:
Tested that it works now

embassy-net/src/dns.rs
```diff
@@ -0,0 +1,114 @@
...
+pub use smoltcp::wire::{DnsQueryType, IpAddress};
+
+use crate::{SocketStack, Stack};
+
+/// Errors returned by DnsSocket.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    /// No available query slot
+    NoFreeSlot,
```
Dirbaio, line 22: `NoFreeSlot` could async-wait until there is one, instead of failing.

embassy-net/src/dns.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Make a query for a given name and return the corresponding IP addresses.
+    pub async fn query(&mut self, name: &str, qtype: DnsQueryType) -> Result<Vec<IpAddress, MAX_ADDRESS_COUNT>, Error> {
+        let query = match { self.with_mut(|s, i| s.start_query(i.context(), name, qtype)) } {
+            Ok(handle) => handle,
+            Err(e) => return Err(e.into()),
+        };
+
+        poll_fn(|cx| {
```
Dirbaio, line 96: if this is async-canceled, it'll leak the query slot inside the socket. It should cal `cancel_query`.

lulf: I'm always overlooking the cancellation :see_no_evil:

embassy-net/src/dns.rs
lulf: Since the query slots are provided in new() and DnsSocket::query is &mut self, would it actually be possible to have a waiter unless query() changes to &self?

Dirbaio: ah good point, seems like no.

I was thinking about the case of creating one DnsSocket and then sharing it "system-wide" to do DNS queries for all tasks. You can't create two DnsSockets, the first one will process and discard all packets for the 2nd one.

Perhaps the socket should be owned by the Stack, and the stack should have a `&self`  method for doing a DNS query? so multiple tasks can do multiple queries concurrently?

lulf: > ah good point, seems like no.
> 
> I was thinking about the case of creating one DnsSocket and then sharing it "system-wide" to do DNS queries for all tasks. You can't create two DnsSockets, the first one will process and discard all packets for the 2nd one.
> 

Is this a limit internally in smoltcp? 

> Perhaps the socket should be owned by the Stack, and the stack should have a `&self` method for doing a DNS query? so multiple tasks can do multiple queries concurrently?

Given the above, yes maybe that's a better approach, I can give it a go.

Dirbaio: > Is this a limit internally in smoltcp?

Yes, the packets are given to the first socket that accepts it https://github.com/smoltcp-rs/smoltcp/blob/739e7a58a9a41ce027eaf1e8857dc507b70f4756/src/iface/interface/mod.rs#L1380
and the DnsSocket accepts all DNS packets https://github.com/smoltcp-rs/smoltcp/blob/739e7a58a9a41ce027eaf1e8857dc507b70f4756/src/socket/dns.rs#L340

lulf: I've changed it to use a single DnsSocket (separate type just for convenience in imports and 'hiding' the stack impl), implemented on the Stack type, which I think makes more sense after implementing it, because it ties into the DNS server updates nicely.

I didn't add the waker in the case of new slots, and created slots equal to the number of sockets (SOCK on the resources). We could add wakers as well, but I'm wondering if it's really needed, as it would just 'postpone' the problem of no free slots by N wakers, which I think just complicates types and config. Or am I missing something?

embassy-net/src/lib.rs
```diff
@@ -46,15 +48,22 @@ use crate::device::DriverAdapter;
 
 const LOCAL_PORT_MIN: u16 = 1025;
 const LOCAL_PORT_MAX: u16 = 65535;
+const MAX_QUERIES: usize = 2;
 
 pub struct StackResources<const SOCK: usize> {
     sockets: [SocketStorage<'static>; SOCK],
+    #[cfg(feature = "dns")]
+    queries: Option<[Option<dns::DnsQuery>; MAX_QUERIES]>,
```
Dirbaio, line 56: I don't think this is ever `None`?

embassy-net/src/lib.rs
```diff
@@ -107,6 +116,8 @@ struct Inner<D: Driver> {
     config: Option<StaticConfig>,
     #[cfg(feature = "dhcpv4")]
     dhcp_socket: Option<SocketHandle>,
+    #[cfg(feature = "dns")]
+    dns_socket: Option<SocketHandle>,
```
Dirbaio, line 120: I think the Option could be removed. If DNS support is enabled, there should always be a socket.

embassy-net/src/lib.rs
```diff
@@ -208,6 +230,59 @@ impl<D: Driver + 'static> Stack<D> {
         .await;
         unreachable!()
     }
+
+    #[cfg(feature = "dns")]
+    async fn dns_query(
```
Dirbaio, line 235: why not make this `pub`? The only reason you'd want to go through `DnsSocket` is if you need embedded-nal, otherwise directly calling this would be less boilerplate.

embassy-net/src/dns.rs
Dirbaio: Just 1 waker should be enough. If there's two tasks waiting they'll fight over the WakerRegistration, just like in `embassy-sync` stuff, but it should all still work. This'd work for an infinite amount of tasks waiting.

embassy-net/src/lib.rs
lulf: It needs to be moved into the socket, so its taken when stack is crated.

Dirbaio: The socket should borrow it. If you move it, the `Into<ManagedSlice<..>>`impl gives you a `ManagedSlice::Owned` which doesn't work without alloc.

lulf: First time i used managed slice, so it was a bit confusing. But now it should be good I think.

Dirbaio (approved):
LGTM!

Dirbaio:
Tested working on nrf52840 usb-ethernet. Thanks for the PR, this is great! :D

bors r+

Dirbaio:
(force-merging, teleprobe flakiness)

overheat:
@lulf Do I need to setup a DNS server on tap0 (192.168.69.100) when testing with static ip?

Dirbaio:
You have to set the DNS server IP in `StaticConfig`. https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/lib.rs#L76

If you have configured `tap0` to be able to route out into the public internet, any public DNS server IP will do. If not, you'll have to setup one on your PC and use that IP.

lulf:
@overheat I followed https://github.com/smoltcp-rs/smoltcp#hosted-usage-examples to setup my environment, and used the DNS servers.

overheat:
@lulf Is it support no_std?
I try to add dns to rp2040, but panicked at 'adding a socket to a full SocketSet', when add dns feature to embassy-net.

lulf:
@overheat It works on no_std, here is an example I just tested works on my rp pico w https://github.com/embassy-rs/cyw43/compare/master...dns-example

Make sure you are not running out of sockets, this is the param to the StackResources type.

overheat:
@lulf 
Yes, I think you are right. I am running out of sockets.

I need to change `StackResources` from `2` to `3` even only enable `dns` feature for `embassy-net` in Cargo.toml, otherwise it will crash.

PR #1187, opened 2023-02-01
executor: Minor refactoring

GrantM11235:
The third commit may be slightly more controversial than the first two. Personally, I think it makes the code more readable and easier to reason about, but I can drop it if you disagree.

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -141,25 +141,14 @@ impl<F: Future + 'static> TaskStorage<F> {
     /// Once the task has finished running, you may spawn it again. It is allowed to spawn it
     /// on a different executor.
     pub fn spawn(&'static self, future: impl FnOnce() -> F) -> SpawnToken<impl Sized> {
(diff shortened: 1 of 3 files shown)


Dirbaio:
>  Set poll_fn in `TaskStorage::new` 

`TaskStorage::new` intentionally returned a TaskStorage with all zero-or-uninit bytes, so that it can to in `.bss` instead of in `.data`. Have you checked whether `#[embassy::task]` tasks are still getting placed in `.bss`?

> Replace the pointer in TaskHeader with an Option<&Executor> 

LGTM! nice :) 

checking the 3rd..

GrantM11235:
> `TaskStorage::new` intentionally returned a TaskStorage with all zero-or-uninit bytes, so that it can to in `.bss` instead of in `.data`. Have you checked whether `#[embassy::task]` tasks are still getting placed in `.bss`?

Oh, I didn't realize that. I changed it back to be lazily initialized in a `Cell<Option>`, which should go in `.bss`

Dirbaio (approved):
Thank you!

bors r+

PR #1188, opened 2023-02-02
LoRa timings for SX126x/STM32WL

samueltardieu:
Those timings open Rx time windows covering 99.7% of the one expected
by the antenna while allowing 3ms for the Rx subsystem to start listening.

### DIFF
--- a/embassy-lora/src/stm32wl/mod.rs
+++ b/embassy-lora/src/stm32wl/mod.rs
@@ -260,10 +260,10 @@ impl From<embassy_stm32::spi::Error> for RadioError {
 
 impl<'d, RS> Timings for SubGhzRadio<'d, RS> {
     fn get_rx_window_offset_ms(&self) -> i32 {
-        -500
+        -3
     }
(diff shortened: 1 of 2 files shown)


lulf (approved):
I don't have a way to test those right now, but I think it makes sense to align it with the other drivers.

bors r+

CBJamo:
This change appears to have broken the sx126x, at least on my hardware. A little experimentation gives me -45ms as the smallest working offset. I'm not sure why that would be. The sx126x's specified worst case rx wakeup time is on the order of 300us, and while I wouldn't be surprised if the clock on my micro was running fast, I would be surprised if it was gaining 45ms over 5s.

I suspect, but do not know, that gateways tend to transmit near the beginning of the rx window, and that it would be better to bias the window in that direction. I'm not really sure where to look/ask for that information though.

samueltardieu:
@CBJamo Don't hesitate to use -50/1050 as values and submit a patch (given that you're in the best position to test it).

I should get my hand on a STM32WL soon and will do the same testing there.

PR #1189, opened 2023-02-02
USB: Add MS OS Descriptors (alternate implementation)

alexmoon:
This is an alternate API for #1152 based on the work of @mattico. By switching to a writer-style API instead of a builder API some compile-time guarantees are lost, but it integrates better into the usb `Builder` and makes an api that can be used by USB device class implementations.

It also adds a feature flag so there is zero cost to the MS OS descriptors for devices that don't need to use them.

I've added an example based on `usb_serial` which tells Windows to use the generic `WinUSB` driver instead of the serial port driver for the device.

Comments are welcome. It would be nice to see either this or #1152 merged as my project is going to require the MS OS Descriptors soon.

### DIFF
--- a/embassy-usb/Cargo.toml
+++ b/embassy-usb/Cargo.toml
@@ -13,7 +13,7 @@ target = "thumbv7em-none-eabi"
 [features]
 defmt = ["dep:defmt", "embassy-usb-driver/defmt"]
 usbd-hid = ["dep:usbd-hid", "dep:ssmarshal"]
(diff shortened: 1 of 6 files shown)


alexmoon:
Right now there's no way to create a multi-configuration device with embassy-usb so I don't think its necessary to support multiple configurations in the MS OS descriptor until embassy-usb adds support.

Dirbaio (approved):
LGTM! :rocket: 

bors r+

PR #1190, opened 2023-02-03
stm32: add window_watchdog timer

pattop:
WindowWatchdog configures a timer channel to capture both edges of an
input and compare the time between edges to a duration range. In other
words, it expects a ~50% duty cycle square wave on its input.

One possible use for this is to monitor whether another device is
functional.

If VALID_THRESHOLD (currently 2) edge intervals are within range, ok()
returns true, otherwise, if an edge is outside of the valid range ok()
returns false.

state_change() runs the calling task whenever the ok() state changes.

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -1,7 +1,7 @@
 #![no_std]
 #![cfg_attr(
     feature = "nightly",
-    feature(type_alias_impl_trait, async_fn_in_trait, impl_trait_projections)
+    feature(type_alias_impl_trait, async_fn_in_trait, impl_trait_projections, int_roundings)
 )]
 #![cfg_attr(feature = "nightly", allow(incomplete_features))]
 
--- a/embassy-stm32/src/timer/window_watchdog.rs
+++ b/embassy-stm32/src/timer/window_watchdog.rs
@@ -1,0 +1,150 @@
+use core::cell::RefCell;
+use core::future::poll_fn;
+use core::marker::PhantomData;
+use core::task::Poll;
+
+use embassy_cortex_m::peripheral::{PeripheralMutex, PeripheralState, StateStorage};
+use embassy_sync::waitqueue::WakerRegistration;
+use embassy_time::Duration;
+use stm32_metapac::timer::vals;
+
+use crate::gpio::sealed::AFType;
+use crate::pwm::*;
+use crate::Peripheral;
+
+/* requires VALID_THRESHOLD edges within range before reporting ok */
+const VALID_THRESHOLD: u8 = 2;
+
+pub struct State<'a, T: CaptureCompare16bitInstance>(StateStorage<StateInner<'a, T>>);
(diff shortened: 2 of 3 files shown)


pattop:
@Dirbaio as discussed on chat.

Dirbaio:
@pattop can you fix CI? the dep on `embassy-time` is optional. I think it's reasonable to require it for this feature, so the fix would be to add cfg's in the right places

pattop:
Will do

embassy-stm32/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![no_std]
 #![cfg_attr(
     feature = "nightly",
-    feature(type_alias_impl_trait, async_fn_in_trait, impl_trait_projections)
+    feature(type_alias_impl_trait, async_fn_in_trait, impl_trait_projections, int_roundings)
```
Dirbaio, line 4: We shouldn't add new nightly features. The already present ones are absolutely required for async traits, and likely to be stabilized soon. The goal is for the whole of Embassy to work on stable when that happens.

embassy-stm32/src/timer/window_watchdog.rs
```diff
@@ -0,0 +1,149 @@
...
+struct StateInner<'a, T: CaptureCompare16bitInstance> {
+    phantom: PhantomData<&'a mut T>,
+    waker: WakerRegistration,
+    min: u16,
+    count: u8,
+}
+unsafe impl<'a, T: CaptureCompare16bitInstance> Send for StateInner<'a, T> {}
+
+pub struct WindowWatchdog<'a, T: CaptureCompare16bitInstance> {
+    inner: RefCell<PeripheralMutex<'a, StateInner<'a, T>>>,
```
Dirbaio, line 34: Could you not use `PeripheralMutex`? We're trying to move away from using it, I'd really like to not introduce new uses. You can pass a pointer to the irq with [set_handler_context](https://docs.embassy.dev/embassy-cortex-m/git/thumbv8m.main-none-eabihf/interrupt/trait.InterruptExt.html#tymethod.set_handler_context), then make that struct usable with `&` instead of `&mut`. (using atomics, mutexes, AtomicWaker etc)

embassy-stm32/src/timer/window_watchdog.rs
```diff
@@ -0,0 +1,149 @@
...
+                count: 0,
+            })),
+        }
+    }
+
+    pub fn ok(&self) -> bool {
+        self.inner.borrow_mut().with(|state| state.ok())
+    }
+
+    pub async fn state_change(&self, old_state: bool) {
```
Dirbaio, line 93: i'd name this `wait_for_state_change`, so it's more obvious what it does

Dirbaio:
Could you add docs describing operation? What is considered as a pulse (rising edge? falling edge?), what exactly triggers the "ok" state to change?

Dirbaio:
abandoned, closing.

PR #1193, opened 2023-02-06
stm32/usart: fix LPUART clock multiplier

pattop:
According to RM0351 Rev 9 (L4) and RM0399 Rev 3 (H7):

baud = (256 * clock) / LPUARTDIV


Dirbaio (approved):
Interesting off-by-one bug! Thank you!

bors r+

PR #1194, opened 2023-02-06
stm32/usart: improve baud rate calculation

pattop:
Add support for 8x oversampling and clock prescaler on v2 hardware.
Fail if baud rate error is > 2.5%.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -1,5 +1,6 @@
 #![macro_use]
 
+use core::cmp;
 use core::future::poll_fn;
 use core::marker::PhantomData;
 use core::sync::atomic::{compiler_fence, Ordering};
@@ -44,6 +45,16 @@ pub enum StopBits {
     STOP1P5,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Oversampling {
+    #[doc = "Automatically determine oversampling ratio"]
+    Auto,
+    #[doc = "Use 8x oversampling"]
+    By8,
+    #[doc = "Use 16x oversampling"]
+    By16,
+}
+
 #[non_exhaustive]
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub struct Config {
@@ -55,6 +66,7 @@ pub struct Config {
     /// read will abort, the error reported and cleared
     /// if false, the error is ignored and cleared
     pub detect_previous_overrun: bool,
+    pub oversampling: Oversampling,
 }
 
 impl Default for Config {
@@ -66,6 +78,7 @@ impl Default for Config {
             parity: Parity::ParityNone,
             // historical behavior
             detect_previous_overrun: false,
+            oversampling: Oversampling::Auto,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usart/mod.rs
```diff
@@ -55,6 +56,8 @@ pub struct Config {
     /// read will abort, the error reported and cleared
     /// if false, the error is ignored and cleared
     pub detect_previous_overrun: bool,
+    // None: automatically determine best oversampling ratio
+    pub force_oversampling: Option<u8>,
```
Dirbaio, line 60: I think this would be more idiomatic as an enum, with cases for "auto, 8, 16". This way you can't set invalid values at all.

pattop: Done.

pattop:
I tried to figure out why CI is failing. I think it's a combination of me using #[cfg] incorrectly (I think we need cfg_if on the presc register write?) and my lack of understanding of how the metapac works :)

Anyway, I'm pretty sure the prescaler register should be implemented for lpuart_v2 and usart_v2, so the #[cfg] blocks probably need to change.

Dirbaio:
you can see which failed in the CI logs:

```
 error: could not compile `embassy-stm32` due to 2 previous errors
     features: ["_time-driver", "defmt", "embedded-hal-1", "embedded-hal-async", "embedded-storage-async", "exti", "nightly", "stm32l476vg", "time", "time-driver-any", "unstable-traits"]
```

so you can repro locally with:
```
cd embassy-stm32
cargo build --target thumbv7em-none-eabi --features defmt,embedded-hal-1,embedded-hal-async,embedded-storage-async,exti,nightly,stm32l476vg,time,time-driver-any,unstable-traits
```

stm32l476vg uses:
- usart v2: https://github.com/embassy-rs/stm32-data/blob/cc93f9d10395077770bebefb6b9488e06b0e5811/data/chips/STM32L476VG.json#L5269
- lpuart v1: https://github.com/embassy-rs/stm32-data/blob/cc93f9d10395077770bebefb6b9488e06b0e5811/data/chips/STM32L476VG.json#L2439

[usart_v2](https://github.com/embassy-rs/stm32-data/blob/main/data/registers/usart_v2.yaml) has GTPR (not PRESC!) and [lpuart_v1](https://github.com/embassy-rs/stm32-data/blob/main/data/registers/lpuart_v1.yaml) hasn't.

The issue is the driver is sharing code between U(S)ART and LPUART by casting all register blocks into LPUART. This was fine because the LPUART regs are a subset of USART.... until now, where we need something that's available in USART and not LPUART.

This'll need a bit of refactoring. Perhaps pass extra info into `configure` so it knows which kind of uart it is. If it does have PRESC/GTPR then you can cast the regblock back from LPUART to USART/UART and do the reg write. Maybe the clock calc needs to change depending on whether the prescaler is available or not (?)

pattop:
I think that's a mistake in the PAC -- USART definitely has a USART_PRESC register in H7.

pattop:
BTW, I tested this on H7 and F4.

I think GTPR is only used in IrDA/Smartcard modes.

pattop:
Looks like we need to define usart_v3 for the prescaler register. It's in at least L4+, H7 and G0 , but not in L4 (v2), so looks like 2016 onwards or so.

Dirbaio:
ah yes, my bad, GTPR is completely unrelated to this. Sorry for the confusion.

Yes, we do need usart_v3. And hopefully there's no chips where USART has it and LPUART doesn't, then :sweat:

pattop:
Is there a way to have compile time assertions to validate that the USART->LPUART conversion is valid?

Dirbaio:
no :(

PR #1199, opened 2023-02-07
STM32 SPI: Set clk-pin pull-up/-down to match spi clock polarity

### LINKED ISSUE #1094: stm32 HAL DMA SPI CLK pin floats during start of transfer
## Problem description

The embassy stm32 HAL for DMA SPI lets the clock line float briefly during the setup of a transaction, rather than continuing to drive it (it should be driven low, in the configuration shown here). This results in the sensor reading errant data (see below). The MOSI line also floats during this time, but that is less of a problem.

## Reproduction

```
#[embassy_executor::main]

    let spi_bus_1 = {
        let sck = p.PA5;
        let miso = p.PA6;
        let mosi = p.PA7;
        let spi = Spi::new(
            p.SPI1,
            sck,
            mosi,
            miso,
            p.DMA2_CH3,
            p.DMA2_CH0,
            mhz(1),
            spi::Config::default(),
        );
        SPI_BUS.init(Mutex::<ThreadModeRawMutex, _>::new(spi))
    };
    let csn = Output::new(p.PB6, Level::High, Speed::High);
    let mut spi_dev_1 = SpiDevice::new(spi_bus_1, csn);

    let mut buf = [0u8, 0];
    embedded_hal_async::spi::SpiDevice::transfer_in_place(&mut spi_dev_1, &mut buf).await.unwrap();
    println!("{:?}", buf);
}
```

## Relevant code

There may not be an easy fix for this, since control of the chip select line is handled outside this HAL code - and the stm32 reference manual says you are required to disable SPI before setting up a DMA transfer. The specific line that lets the pin start floating is below:

https://github.com/embassy-rs/embassy/blob/5e94b8060b7f1af8118ca1f943181ce70be92057/embassy-stm32/src/spi/mod.rs#L512

## Potential solutions

A few solutions to this were discussed in the matrix chat.

- Keep the DMA transaction open across calls to read/write
  - This may be problematic if the user changes bus settings between calls, and also the reference manual says the chip should not be placed into low power mode while SPI is enabled
- Set the CLK (and maybe MOSI) pins as pull-down on setup (or pull-up, depending on config - and this would need to be updated if the user modified the config)
  - This is less good than driving the pin to the correct value, but may be better than nothing
- Document this and require users fix it themselves (add a pull-up/down resistor - or configure the pins as pull-up/pull-down before passing them into SPI setup)
- Dig around in the reference manual and determine if there is a better way to start/stop a DMA transaction while keeping active control of the clock the whole time
  

## Extra data

Expected (taken from stm32f4xx-hal)

nice-spi

Embassy stm32 HAL w/ DMA (note the long CLK "pulse" at the beginning)

buggy-embassy-hal-spi

jr-oss:
Fixes #1094 

There are some proposed solutions in #1094 

> Keep the DMA transaction open across calls to read/write
        This may be problematic if the user changes bus settings between calls, and also the reference manual says the chip should not be placed into low power mode while SPI is enabled

As already described, this is problematic and against reference manual recommendation

>    Set the CLK (and maybe MOSI) pins as pull-down on setup (or pull-up, depending on config - and this would need to be updated if the user modified the config)
        This is less good than driving the pin to the correct value, but may be better than nothing

That is also my preferred solution. See below citation from reference manual.

>    Document this and require users fix it themselves (add a pull-up/down resistor - or configure the pins as pull-up/pull-down before passing them into SPI setup)

Setting internal pull-up/-down won't work, because `sck.set_as_af()` will change the gpio pull mode to none: https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/gpio.rs#L552-L555

>    Dig around in the reference manual and determine if there is a better way to start/stop a DMA transaction while keeping active control of the clock the whole time

I haven't found a better way

------
From ST reference manual RM0394 (L4) 
(Same note in RM0399 (H7) / RM0038 (L1) / RM0316 /F3)):

    40.4.6
    Communication formats
    ...
    The idle state of SCK must correspond to the polarity selected in the SPIx_CR1 register (by
    pulling up SCK if CPOL=1 or pulling down SCK if CPOL=0).

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -10,7 +10,7 @@ pub use embedded_hal_02::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MO
 use self::sealed::WordSize;
 use crate::dma::{slice_ptr_parts, Transfer};
 use crate::gpio::sealed::{AFType, Pin as _};
-use crate::gpio::AnyPin;
+use crate::gpio::{AnyPin, Pull};
 use crate::pac::spi::{regs, vals, Spi as Regs};
 use crate::rcc::RccPeripheral;
 use crate::time::Hertz;
@@ -93,8 +93,14 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
         config: Config,
     ) -> Self {
         into_ref!(peri, sck, mosi, miso);
+
+        let sck_pull_mode = match config.mode.polarity {
+            Polarity::IdleLow => Pull::Down,
+            Polarity::IdleHigh => Pull::Up,
+        };
+
         unsafe {
-            sck.set_as_af(sck.af_num(), AFType::OutputPushPull);
+            sck.set_as_af_pull(sck.af_num(), AFType::OutputPushPull, sck_pull_mode);
             sck.set_speed(crate::gpio::Speed::VeryHigh);
             mosi.set_as_af(mosi.af_num(), AFType::OutputPushPull);
             mosi.set_speed(crate::gpio::Speed::VeryHigh);


Dirbaio (approved):
Thanks! 

bors r+

PR #1200, opened 2023-02-07
feat(stm32): Add 16 data bit fmc ctor

rmja:
This has been validated with the Is42s16400j sdram on stm32f429.


Dirbaio (approved):
thanks!

bors r+

PR #1202, opened 2023-02-07
usb: make max interface count configurable at compile time.

Dirbaio:
New approach to #1102, not using const generics.


Dirbaio:
merging as part of #1203

PR #1205, opened 2023-02-08
stm32/rng Fix rng generation lock-up

lucasgranberg:
This PR fixes a problem where the device gets locked in case of rng errors.

The PR also includes a hack for stm32wl based devices where the more complicated RNG peripheral can get stuck on seed errors.

### DIFF
--- a/embassy-stm32/src/rng.rs
+++ b/embassy-stm32/src/rng.rs
@@ -32,8 +32,9 @@ impl<'d, T: Instance> Rng<'d, T> {
     }
 
     pub fn reset(&mut self) {
+        // rng_v2 locks up on seed error, needs reset
         #[cfg(rng_v2)]
-        if unsafe { T::regs().sr().read().seis()} {
+        if unsafe { T::regs().sr().read().seis() } {
             T::reset();
         }
         unsafe {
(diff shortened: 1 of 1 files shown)


Dirbaio:
Could you add rng_v2 to stm32-data, and do `cfg(rng_v2)` instead?

I see L5, U5 also have the CONDRST bit, so presumably they're also affected by the hang. Fixing it in stm32-data and doing the reset based on the peripheral version ensures the bug won't reappear for future chip families.

lucasgranberg:
I can try to get the peripheral version working. My first attempt at porting the CubeWL code was not successful.
I have the code for stm32-data so Ill start with a PR for that then.

Dirbaio:
If the RCC reset works, we can leave that for now, no need to port the weird st's reset sequence.

I meant that we should have the right version in stm32-data so that we can do `cfg(rng_v2)`, even if we don't use the new v2 registers yet.

lucasgranberg:
I created a PR for v2 for stm32-data
https://github.com/embassy-rs/stm32-data/pull/169

Dirbaio (approved):
Thanks! :rocket: 

bors r+

PR #1208, opened 2023-02-09
nrf/buffered_uarte: make it work without rts/cts, and lock-free.

Dirbaio:
> dirbaio: so I was checking how zephyr does UARTE RX on nRF
> dirbaio: because currently we have the ugly "restart DMA on line idle to flush it" hack
> dirbaio: because according to the docs "For each byte received over the RXD line, an RXDRDY event will be generated. This event is likely to occur before the corresponding data has been transferred to Data RAM."
> dirbaio: so as I understood it, the only way to guarantee the data is actually transferred to RAM is to stop+restart DMA
> dirbaio: well, guess what?
> dirbaio: they just count RXDRDY's, and process that amount of data without restarting DMA
> dirbaio: with a timer configured as counter https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/serial/uart_nrfx_uarte.c#L650-L692
> dirbaio: 🤔🤷⁉️
> dirbaio: someone saying you can do the "hook up rxdrdy to a counter" trick, someone else saying it's wrong 🤪 https://devzone.nordicsemi.com/f/nordic-q-a/28420/uarte-in-circular-mode

So we're going to do just that!

- BufferedUarte is lock-free now. No PeripheralMutex.
- The "restart DMA on line idle to flush it" hack is GONE. This means
  - It'll work correctly without RTS/CTS now.
  - It'll have better throughput when using RTS/CTS.

~Very WIP, it doesn't work yet.~


embassy-nrf/src/buffered_uarte.rs
```diff
@@ -13,82 +8,128 @@
...
+            tx_buffer,
+        )
+    }
+
+    /// Create a new BufferedUarte with hardware flow control (RTS/CTS)
+    ///
+    /// # Panics
+    ///
+    /// Panics if `rx_buffer.len()` is odd.
+    pub fn new_with_rtscts(
```
huntc, line 127: Small question re. convention, and I realise this was already there, is it more the convention to have just `with_rtscts` i.e. without the `new_`?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -111,92 +185,206 @@ impl<'d, U: UarteInstance, T: TimerInstance> BufferedUarte<'d, U, T> {
...
+            r.intenclr.write(|w| w.rxdrdy().clear());
+            r.events_rxdrdy.reset();
+            s.rx_waker.wake();
+        }
+
+        //if r.events_endrx.read().bits() != 0 {
+        //    r.events_endrx.reset();
+        //    //trace!("  irq_rx: endrx");
+        //    r.shorts.write(|w| w.endrx_startrx().clear_bit());
+        //}
```
huntc, line 309: Remove?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -111,92 +185,206 @@ impl<'d, U: UarteInstance, T: TimerInstance> BufferedUarte<'d, U, T> {
...
-        // BAUDRATE register values are `baudrate * 2^32 / 16000000`
-        // source: https://devzone.nordicsemi.com/f/nordic-q-a/391/uart-baudrate-register-values
-        //
-        // We want to stop RX if line is idle for 2 bytes worth of time
-        // That is 20 bits (each byte is 1 start bit + 8 data bits + 1 stop bit)
-        // This gives us the amount of 16M ticks for 20 bits.
-        let timeout = 0x8000_0000 / (config.baudrate as u32 / 40);
+        // Configure byte counter.
+        let mut timer = Timer::new_counter(timer);
+        timer.cc(1).write(tx_buffer.len() as u32 * 2);
```
huntc, line 241: We're using the tx_buffer len here. Given that this byte counter is associated with receiving, should it be the rx buffer?

huntc (commented):
Looks good, but I'm still not quite clear on the principle of operation. Does it cope with that use-case you highlighted where you can start receiving before setting the second rx buffer?

embassy-nrf/src/buffered_uarte.rs
Dirbaio: Where's this convention from? I've seen `from_foo` instead of `new_from_foo`, but not `with_foo`. 

Also to me `with_rtscts` sounds like you're adding RTS/CTS to an already created instance, builder-style, which is not what's happening here.

embassy-nrf/src/buffered_uarte.rs
Dirbaio: :+1:

embassy-nrf/src/buffered_uarte.rs
Dirbaio: Whooops. Very nice catch, thanks!

PR #1209, opened 2023-02-10
Time: Add from_hz function for Duration.

CBJamo:
I found myself doing things like this

```rust
    let rate_us = 1_000_000 / rate_hz;
    let mut ticker = Ticker::every(Duration::from_micros(rate_us));
```

Several times, and figured it was worth adding a little convenience function to handle that. This also makes the calculation const, which is a nice little upside. The compiler might have been doing that already, but this makes sure. 

Speaking of const, would it be better to give hz as a float? Obviously we'd want to avoid that at runtime since many targets don't have a fpu, but if it's at compile time that doesn't matter and a float may be more ergonomic.

### DIFF
--- a/embassy-time/src/duration.rs
+++ b/embassy-time/src/duration.rs
@@ -81,6 +81,11 @@ impl Duration {
         }
     }
 
+    /// Creates a duration corresponding to the specified Hz.
+    pub const fn from_hz(hz: u64) -> Duration {
+        Duration { ticks: TICK_HZ / hz }
+    }
+
     /// Adds one Duration to another, returning a new Duration or None in the event of an overflow.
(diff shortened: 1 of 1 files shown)


Dirbaio:
Thanks for the PR! This is definitely something that'd be nice to have.

- Could you round to nearest instead of to zero? This reduces the average error.
- Perhaps it should force a minimum of 1 tick, in case the duration is rounded to exactly 0 ticks? This'd prevent the Ticker from becoming a hot loop if you set a too high freq.

> would it be better to give hz as a float?

Not in the default `from_hz` function because there's chips with lack of a FPU. We could do something like `from_hz_f32` though.

Dirbaio (approved):
Thank you!

bors r+

PR #1211, opened 2023-02-12
Fix rcc prescaler for wl55 HCLK1

chrenderle:
fix "prescaler none" which incorrectly set "prescaler divided by 3"
Issue: #1168


lulf:
Thanks!

bors r+

PR #1215, opened 2023-02-13
Add clone to embassy_rp::gpio::Level

Slushee-a:
Allows you to wite a cleaner state change detector. Example:
```rs
let mut button_state: Level = Level::Low;
let mut prev_button_state: Level = button_state;

loop {
    button_state = button.get_level();

    if prev_button_state != button_state {
        led.set_level(button_state);  // Takes ownership of button_state. 
    }

    prev_button_state = button_state; // Can't be done since the ownership has been moved.
                                      // Adding Clone makes this code possible
}
```


Dirbaio:
Could you add `Copy` as well? For tiny enums like this, the best is `Copy+Clone`.

Dirbaio (approved):
Thank you!

bors r+

PR #1216, opened 2023-02-14
Lora: Make sx126x antenna rx/tx pins optional

CBJamo:
On our hardware, we're using the alt mode function of dio2 to control rx/tx, but not controlling the crystal with dio3, the library currently assumes just the inverse.

The part of this patch addressing the antenna switching is pretty reasonable, but is a breaking change. The part disabling the crystal control is a horrible hack because I'm not sure what the most idiomatic way to approach making that optional is.


ceekdee:
Related issues:

- https://github.com/embassy-rs/embassy/issues/1041 (no fix proposed yet)
- https://github.com/embassy-rs/embassy/issues/1023 (an in-progress consolidation of LoRa PHY functionality which addresses  optional antenna control but does not yet address optional DIO2/DIO3 processing for sx126x).

Would it be possible to share the specific mcu/sx126x combination for which this dio2/dio3 flexibility is needed so that it can be considered for the consolidated LoRa implementation?

lulf:
@ceekdee ^^ This looks good to me, but maybe it's already handled in your rewrite? Maybe it makes sense to merge it for now?

ceekdee:
As stated in the initial comment, this is a breaking change.  In particular, the following two examples are affected:

- https://github.com/embassy-rs/embassy/blob/master/examples/nrf52840/src/bin/lora_p2p_report.rs#L36
- https://github.com/embassy-rs/embassy/blob/master/examples/nrf52840/src/bin/lora_p2p_sense.rs#L71

In addition, I assume commenting out DIO3 setup will break support for some mcu/sx126x combinations.  A feature could be used to remove this functionality, but I understand there is hesitancy to add features.

CBJamo:
Ceekdee's work will certainly supersede this, so I'll close this in favor of #1023

PR #1218, opened 2023-02-17
Lora: sx126x: Change timing window to match values found experimentally.

CBJamo:
As mentioned in #1188.


ceekdee:
@lulf: Is my understanding correct that the timing window is not a LoRa physical layer concern and should therefore move to the MAC layer as we move forward with a new, consolidated LoRa PHY implementation and updates to the LoRa MAC layer to calculate the windows?

PR #1219, opened 2023-02-17
stm32/sdmmc: Fix SDIOv1 writes

chemicstry:
This fixes writes on sdmmc v1 (SDIO). I'm pretty sure I tested writes in #669, but maybe I was just lucky or I just forgot.

There were two problems:
- Writes require DMA FIFO mode, otherwise SDIO FIFO is under/overrun depending on sdio/pclk2 clock ratio.
- Hardware flow control is broken for sdmmc v1 (I checked F1 and F4 erratas). This causes clock glitches above 12 MHz and results in write CRC errors.


Dirbaio (approved):
Wow, this must've been hard to figure out. Thank you for the effort!

bors r+

PR #1224, opened 2023-02-21
nrf: new interrupt binding traits/macro

Dirbaio:
Introduce a new interrupt binding system.

Goals:
- Remove the indirect call trampoline of the previous system.
- Still catch mistakes like "forgot to bind interrupt" and "bound the wrong interrupt" at compile-time
- Allow binding one interrupt to multiple handlers (this is needed for STM32).
- Allow picking between different handlers for the same interrupt (e.g. `Uarte` and `BufferedUarte`) at compile time.

The mechanism is a `bind_interrupts!` macro that generates the handler, and a "token" struct with `Binding` impls that you can pass to drivers to "prove" at compile time that you've bound the required interrupts.

The new system is more flexible: you can manually write the handler and the `unsafe impl Binding` yourself. Useful if you want to do custom stuff in the interrupt handler.

Originally prototyped as part of #980


embassy-nrf/src/interrupt.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// # Safety
+    ///
+    /// This function must ONLY be called from the interrupt handler for `I`.
+    unsafe fn on_interrupt();
+}
+
+/// Compile-time assertion that an interrupt has been bound to a handler.
+///
+/// Most of the time[`bind_interrupts`](crate::bind_interrupts)
```
lulf, line 29: Seems like part of this sentence is missing

lulf (approved):
I think it's a nice improvement!

I'm wondering if this could be an alternative, though more painful to implement:


```rust
#[bind_interrupts(SERIAL0 => uarte::InterruptHandler<SERIAL0>)]
struct Irqs;
```

Dirbaio:
> I'm wondering if this could be an alternative, though more painful to implement:
> 
> ```rust
> #[bind_interrupts(SERIAL0 => uarte::InterruptHandler<SERIAL0>)]
> struct Irqs;
> ```

I thought about it, but:
- `macro_rules!` allow using `$crate`, which avoid the issue of the current `take!` macro working if you use it like `use embassy_nrf::interrupt; interrupt::take!(FOO);` but failing if you use it like `use embassy_nrf::interrupt::take; take!(FOO);`
- `macro_rules!` allow autocomplete in rust-analyzer to somewhat work.
- proc macros suck to implement.

korken89:
From our old discussions, you can find my experiments here: https://github.com/korken89/async-nrf52832-hal/blob/master/nrf52832-hal/src/gpiote.rs#L12

I was quite happy with the usability of it, and might give some inspiration.

lulf:
> I'm not too happy about the syntax though. Any ideas for alternatives? Maybe this?
> 
> ```rust
> bind_interrupts!(struct Irqs {
>     SERIAL0 => uarte::InterruptHandler<SERIAL0>;
> });
> ```

I think this looks better

PR #1226, opened 2023-02-22
embassy-net: Implement flush for TcpSocket

### LINKED ISSUE #1223: embassy-net: TCP sockets discards unsent data on Drop
I have a relatively simple program that
1. accepts a new connection
2. writes some bme280 measurements to the socket
3. drops the socket and waits for a new connection

This results in no data being transmitted. As far as I can tell, the problem is that Drop immediately unregisters the socket's buffers from the network stack so that the stack never gets to execute the write.

I would have expected all outdoing data to be sent. Alternatively, there should be a method that waits for this.

```rust
let mut rx_buffer = [0; 4096];
let mut tx_buffer = [0; 4096];
let mut buf = [0; 4096];

loop {
    let mut socket = TcpSocket::new(stack, &mut rx_buffer, &mut tx_buffer);
    if let Err(e) = socket.accept(1234).await {
        defmt::warn!("accept error: {:?}", e);
        continue;
    }
    let m = bme.measure(&mut Delay).await.unwrap();
    let n = serde_json_core::to_slice(&m, &mut buf).unwrap();

    match socket.write_all(&buf[..n]).await {
        Ok(()) => {}
        Err(e) => {
            defmt::warn!("write error: {:?}", e);
            break;
        }
    }
}
```

kbleeke:
Implements flush for TcpSocket by checking the send queue. 

Flushing is implemented by checking if smoltcp's send_queue/tx_buffer is empty. The flush is completed when all outstanding octets are acknowledged. Smoltcp wakes the send waker [here](https://docs.rs/smoltcp/latest/src/smoltcp/socket/tcp.rs.html#1712) when ACKs are processed and data is removed from the send buffer. So we can re-check in our flush implementation, if the buffer is now empty.

fixes #1223

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -254,10 +254,19 @@ impl<'d> TcpIo<'d> {
         .await
     }
 
-    #[allow(unused)]
     async fn flush(&mut self) -> Result<(), Error> {
-        poll_fn(move |_| {
-            Poll::Ready(Ok(())) // TODO: Is there a better implementation for this?
+        poll_fn(move |cx| {
+            self.with_mut(|s, _| {
+                // If there are outstanding send operations, register for wake up and wait
+                // smoltcp issues wake-ups when octets are dequeued from the send buffer
+                if s.send_queue() > 0 {
+                    s.register_send_waker(cx.waker());
+                    Poll::Pending
+                // No outstanding sends, socket is flushed
+                } else {
+                    Poll::Ready(Ok(()))
+                }
+            })
         })
         .await
     }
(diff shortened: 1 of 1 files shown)


Dirbaio:
Could you add `pub async fn flush()` to `TcpSocket`, `TcpWriter` as well? 

The reason is that async traits are not yet available in Rust stable, so we add inherent (non-trait) methods for all the functionality. There was no point in having the `flush` methods before if they were noops, but now there is.

Thanks!

kbleeke:
Right, added.

Dirbaio (approved):
Thank you!

bors r+

PR #1227, opened 2023-02-22
stm32/dma: fix spurious transfer complete interrupts

pattop:
DMA interrupts must be acknowledged by writing to the DMA_{L,H}IFCR
register.

Writing to the CR register is unnecessary as the channel (EN bit) is
disabled by hardware on completion of the transfer.


Dirbaio (approved):
Thanks! 

bors r+

PR #1228, opened 2023-02-23
stm32/sdmmc: Implement proper clock configuration

chemicstry:
This implements proper clock configuration for sdmmc based on chip family, because `RccPeripheral::frequency()` is almost always incorrect. This can't be fixed in PAC, because sdmmc uses two clock domains, one for memory bus and one for sd card. `RccPeripheral::frequency()` usually returns the memory bus clock, but SDIO clock calculations need sd card domain clock. Moreover, chips have multiple clock source selection bits, which makes this even more complicated. I'm not sure if it's worth implementing all this logic in `RccPeripheral::frequency()` instead of cfg's in sdmmc.

Some chips (Lx, U5, H7) require RCC updates to expose required clocks. I didn't want to mash everything in a single PR so left a TODO comment. I also left a `T::frequency()` fallback, which seemed to work in H7 case even though the clock is most certainly incorrect.

In addition, added support for clock divider bypass for sdmmc_v1, which allows reaching a maximum clock of 48 MHz. The peripheral theoretically supports up to 50 MHz, but for that ST recommends setting pll48 frequency to 50 MHz 🤔

### DIFF
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -140,10 +140,21 @@ cfg_if::cfg_if! {
         /// Calculate clock divisor. Returns a SDMMC_CK less than or equal to
         /// `sdmmc_ck` in Hertz.
         ///
-        /// Returns `(clk_div, clk_f)`, where `clk_div` is the divisor register
-        /// value and `clk_f` is the resulting new clock frequency.
-        fn clk_div(ker_ck: Hertz, sdmmc_ck: u32) -> Result<(u8, Hertz), Error> {
-            let clk_div = match ker_ck.0 / sdmmc_ck {
+        /// Returns `(bypass, clk_div, clk_f)`, where `bypass` enables clock divisor bypass (only sdmmc_v1),
+        /// `clk_div` is the divisor register value and `clk_f` is the resulting new clock frequency.
+        fn clk_div(ker_ck: Hertz, sdmmc_ck: u32) -> Result<(bool, u8, Hertz), Error> {
+            // sdmmc_v1 maximum clock is 50 MHz
+            if sdmmc_ck > 50_000_000 {
+                return Err(Error::BadClock);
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/sdmmc/mod.rs
```diff
@@ -550,6 +563,44 @@ impl<'d, T: Instance, Dma: SdmmcDma<T>> Sdmmc<'d, T, Dma> {
...
+                    crate::rcc::get_freqs().pll48
+                }).expect("PLL48 is required for SDIO")
+            } else if #[cfg(stm32f7)] {
+                critical_section::with(|_| unsafe {
+                    use core::any::TypeId;
+                    let sdmmcsel = if TypeId::of::<T>() == TypeId::of::<crate::peripherals::SDMMC1>() {
+                        crate::pac::RCC.dckcfgr2().read().sdmmc1sel()
+                    } else {
+                        crate::pac::RCC.dckcfgr2().read().sdmmc2sel()
+                    };
```
chemicstry, line 590: Is there a better way to express this?

Dirbaio: make the `kernel_clock()` method part of the `trait Instance`, implement it inside the `foreach_peripheral!`.

To make it generate different code for sdmmc1/2 without having to copypaste everything, you could do this:

```rust
macro_rules! kernel_clock {
    (sdmmc1) => {
        // code here
    };
    (sdmmc2) => {
        // code here
    };
}

foreach_peripheral!(
    (sdmmc, $inst:ident) => {
        impl sealed::Instance for peripherals::$inst {
            ...
            fn kernel_clock() -> Hertz {
                kernel_clock!($inst);
            }
        }
    };
);
```

Super ugly, but I think it's less bad than using `TypeId`.

Dirbaio (approved):
LGTM, feel free to merge when CI's green!

bors d+

PR #1231, opened 2023-02-23
embassy-time: Implement conversions to/from core::time::Duration for embassy-time::Duration

kbleeke:
I chose microseconds for the conversion as the lowest resolution that embassy provides. A new Error-type did not seem that useful but I can add one, if necessary.


Dirbaio (approved):
Thanks!

bors r+

PR #1238, opened 2023-02-25
embassy-net: DNS resolver detects when name is just an IP address

### LINKED ISSUE #1237: embassy_net: should DNS resolve IP adresses without asking the DNS server?
I'm using an address like "192.168.178.1:8086" with reqwless. 
This fails with a DNS error since neither reqwless or embassy recognizes this is an IP address and not a hostname.

I think ether reqwless or embassy should handle this case and immediately return an IP if the query is for an IP. Should embassy handle this case? 

I'd add something like this to `Stack::dns_query`
```rust
match qtype {
    dns::DnsQueryType::A => {
        if let Ok(ip) = name.parse().map(IpAddress::Ipv4) {
            return Ok(iter::once(ip).collect());
        }
    }
    #[cfg(feature = "proto-ipv6")]
    dns::DnsQueryType::AAAA => {
        if let Ok(ip) = name.parse().map(IpAddress::Ipv6) {
            return Ok(iter::once(ip).collect());
        }
    }
    _ => {}
}
```

kbleeke:
fixes #1237 

I am not sure, if this is the right place to put this code. Alternatively, It could be in dns::DnsSocket or reqwless (and all other libraries that need to maybe resolve hostnames).

Are there other DNS query-types where it would make sense to try to parse an IP address?

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -236,6 +236,22 @@ impl<D: Driver + 'static> Stack<D> {
     /// Make a query for a given name and return the corresponding IP addresses.
     #[cfg(feature = "dns")]
     pub async fn dns_query(&self, name: &str, qtype: dns::DnsQueryType) -> Result<Vec<IpAddress, 1>, dns::Error> {
+        // For A and AAAA queries we try detect whether `name` is just an IP address
+        match qtype {
+            dns::DnsQueryType::A => {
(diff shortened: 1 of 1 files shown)


lulf (approved):
I think this should cover most use cases for embassy-net. There is also the https://crates.io/crates/itsdns crate which would benefit from this shortcut.

Thanks for the fix!

bors r+

Issue #745, opened 2022-05-02
Embassy HAL implementation requirements


MabezDev:
We're looking to implement the required traits/executors for Espressif chips, do you have any pointers on what we'll need to implement in our HAL's?

So far it looks like we'll need:

- Embassy executor for RISCV arch.
- Embassy executor for Xtensa arch.
- Add embedded-hal alpha traits.
- Implement embedded-hal-async where appropriate.

lulf:
Great to see the interested of running embassy on espressif! :rocket:  

> We're looking to implement the required traits/executors for Espressif chips, do you have any pointers on what we'll need to implement in our HAL's?
> 
> So far it looks like we'll need:
> 
>     * Embassy executor for RISCV arch.
> 
>     * Embassy executor for Xtensa arch.
> 

:+1: ! The 'core' of the executor is arch-independent, but the the different arch-dependent executor parts can be found here:

https://github.com/embassy-rs/embassy/tree/master/embassy/src/executor/arch

The cortex_m one might be the most relevant to look at. 

>     * Add embedded-hal alpha traits.
> 
>     * Implement embedded-hal-async where appropriate.

In addition to the embedded-hal(-async) traits, embassy has the concept of an opt-in 'timer driver' system timer that is implemented for most HALs, which makes async timers just work. For instance: 

https://github.com/embassy-rs/embassy/blob/master/embassy/src/time/driver.rs

and an implementation:
https://github.com/embassy-rs/embassy/blob/master/embassy-nrf/src/time_driver.rs#L122-L126

There is also the `critical-section` crate that must have implementations for the target architecture (https://crates.io/crates/critical-section). Risc V seems to have some support there already.

There is also some hardware specific stuff in the macros that's needed for users to use the `#[embassy::main]` macro:

https://github.com/embassy-rs/embassy/blob/master/embassy-macros/src/macros/main.rs#L8-L15

MabezDev:
Thanks for the pointers!

For HAL peripheral implementations I also saw the `Unborrow` trait used sporadically, I understand what it's doing but I don't know why it's needed? Could you point me in the direction of some docs to explain the requirement of this trait and when to use it?

Dirbaio:
[...]

PR #1240, opened 2023-02-26
Use atomics using atomic-polyfill

psykora:
Replace use of `core::sync::atomic` by `atomic_polyfill`

Motivation: When playing with rust and `esp-rs/esp-wifi`, I noticed that the examples does not compile with the embassy master. 
This is quick fix for the problem. I noticed that the polyfill is used as drop-in replacement on other places

- Target: riscv32imc-unknown-none-elf
- Device: esp32c3

### DIFF
--- a/embassy-executor/src/arch/riscv32.rs
+++ b/embassy-executor/src/arch/riscv32.rs
@@ -1,6 +1,7 @@
 use core::marker::PhantomData;
 use core::ptr;
-use core::sync::atomic::{AtomicBool, Ordering};
+
+use atomic_polyfill::{AtomicBool, Ordering};
 
 use super::{raw, Spawner};
 
--- a/embassy-hal-common/Cargo.toml
+++ b/embassy-hal-common/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 [features]
 
 [dependencies]
+atomic-polyfill = { version = "1.0" }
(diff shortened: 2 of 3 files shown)


Dirbaio:
See discussion here https://github.com/embassy-rs/embassy/issues/745#issuecomment-1445069621

psykora:
Hi Dirbaio,

Thanks for pointing out there is another workaround. I didn't know it. (To be honest my experience with rust on MCU is close to zero 😳).

However, I still don't understand why we can't use `atomic-polyfill` here. What is different from other cases? 

Rationale:
- It is not newly introduced to embassy. 
  It is used on numerous places (just quickly looked for occurrences of `atomic_polyfill` in this repository) for example.:
  [embassy-executor/src/raw/run_queue.rs](https://github.com/embassy-rs/embassy/blob/c21cc21c622de5527668671c61f677baf7924cf3/embassy-executor/src/raw/run_queue.rs)
- To my understanding [atomic-polyfill](https://github.com/embassy-rs/atomic-polyfill) should just reexport `core::sync::atomic::` on targets where there is native support.

Thanks for understanding,

Dirbaio:
It has overhead on targets without hardware CAS support, because it has to acquire a critical section for each operation. This is also needed in load/store operations to prevent them from racing with CAS operations. If you never do CAS on that atomic value, it's useless overhead.

That code you linked does do CAS operations.

PR #1242, opened 2023-02-26
WIP: usb mass storage (bulk-only, SCSI)

chemicstry:
This is currently a mess, but I'm publishing this as a WIP to avoid working on the same thing in parallel.

Example status:
- `usb_msc_ram` - should work fine on both windows/linux.
- `usb_msc_flash` - does not work. I guess STM32 flash is too slow and OS timeouts on writes.
- `usb_msc_sdmmc` - seems to work on linux, but causes weird hangs and freezes on windows.

I'm going to clean this up once I get it working properly, but I would like to get feedback on some important parts.

SCSI protocol is big-endian and uses a lot of bitfields with weird sizes and alignments so it's a nightmare to work without a struct packing library. I couldn't find a suitable packing library for this, because most of them can't do lazy deserialization, which is really desirable for SCSI as it contains a lot of obsolete or unused fields. The closest I could find was [modular-bitfield](https://github.com/Robbepop/modular-bitfield), but it does not support u8 arrays inside structs. So instead I came up with a declarative macro only packing solution. It uses const generics so compiled accessor functions are optimized by compiler to a few instructions.

The whole SCSI thing is really big and I'm not sure if `embassy-usb` is the right place for it. Would it be better if I published it as a separate crate?


gak:
This is amazing work! I can't wait to see it merged (or as a separate crate). I have a use for this code and can try out the PR in the next few weeks to give feedback.

chemicstry:
This is still far from complete and probably needs some work to update with latest embassy changes. I still have a use case for this so it's not forgotten. Maybe in the next couple of months I will return to this, unless someone else wants to pick it up.

Tbh, I'm not sure if it's better included in embassy or a separate crate. The SCSI part begs to be in a separate crate, but as it can be seen from other existing SCSI crates, it is very hard to cover all use cases and environments with a single crate.

mchodzikiewicz:
Hi, I very much interested in msc for embassy.

Very common usecase I can see is combining it with a filesystem and USB provisioning an IoT device by mounting and drag-and-droping config file - if there are no issues with `usb_msc_ram` maybe it could already land as a separate PR already?

I will be happy to contribute provisioning example if I get it working

Dominaezzz:
This is fantastic work and thanks for contributing it!

Just wanted to say that I tested this on an ESP32S3 with an SD Card over SPI, and my Linux desktop was able to read the FAT32 filesystem just fine. Access was quite slow but that's likely due to the `BlockDevice` trait only reading one block at a time. Ideally the trait would be able to read `n` blocks at a time. The scsi implementation could also start reading the next block whilst the previous block is being sent to the host. But these are optimisations that could be applied in another PR perhaps.

I'll keep playing around with this.

Wassasin:
I forked this, bumped to the most recent `main`, and wrote examples for the NRF52840. It works fine on Linux. I'll test it with Windows later. I noted that ejecting the device does not work at the moment, and I'll look into how this relates to SCSI.

Wassasin:
I have tested the NRF52840 implementation on both Windows and Linux, and it worked without issue. I was not able to test the STM32F4 as Fedex has misplaced the package with the Nucleo I just ordered, and I'll be on holiday the next couple of days. At least the examples for the STM32F4 compile once more.

This is my first encounter with implementing SCSI, so forgive me if I misunderstand something. Reading this MR over, it seems correct & fine to me, and I am unsure what exactly is missing. Most documentation I found list a minimal set of commands that should be implemented, and I believe this MR implements them. I only fixed up some stuff to correctly use a recent version of main, added documentation or made parts of the implementation private, and removed some constants (like the serial number).

In this branch we advertise that the SCSI device is removable. Ejecting a SCSI device is interesting, in that the USB-sticks I looked at seemed to shut down the USB endpoints, until VBUS is re-asserted. It is not clear to me what we should do in this case. For now I opted to implement the command as a no-op, except giving the implementing party a chance to flush any pending operations.

I was unsure whether all STM32 Flash peripherals and/or their alternative regions are `MultiwriteNorFlash`. In my experience with F7, H7, G0 and L4, they are. For now I only marked the main Flash peripheral to be this.

As I am not a maintainer of Embassy, you can find my branch @ https://github.com/Wassasin/embassy/tree/usb_msc .

Wassasin:
I received the STM32F429ZI nucleo board and updated all the examples to run. The `usb_msc_flash` example did not work because the `RmwMultiwriteNorFlashStorage` buffer did not fit alongside the stack. Initializing the BSS section causes the firmware to hang. Another issue is that the flash page size for the STM32F429ZI depends on the flash region addressed. The example now uses a specific flash region that has a small page size.

The SDMMC example works, but I noticed that a FAT32 or EXT4 filesystem uses a lot of blocks, both at creation and when reading the partition. Currently loading such a filesystem is really slow. Creating a smaller file system (128MB compared to 32GB) in this case makes it way faster.

Dirbaio:
Hello,

Issue #1241, opened 2023-02-26
embassy-rp Ability to change UART baudrate during runtime


andres-hurtado-lopez:
As far as I have worked with Embassy, is of my understanding that  Peripherals and UART interfaces are built under the concept of the UART interface consuming the interface pins from the the Peripherals singleton. 

In rust if a member of a struct is consumed there is no way back. Which means as soon as the UART is created you can't create another instance with a different baudrate. This is not the case of more more high level frameworks, _(sorry for the example I'm going to give here I which I understand is a totally different tool)_ the arduino **Processing** environment on which to change the speed is only a matter of creating again a new UART instance with a new speed configuration.

There is a case in which there are serial devices that work with different buadrates during runtime _(for intance configure the device at 9600 8-N-1 and then go 115200 for regular operation after configuration)_. On the `embassy_rp::uart::Uart` methods there isn't any kind of aid to change the baudrate neither any other configuration of the UART during runtime. I'm missing something on how I'm thinking on how to solve the problem or this is a feature that is missing on the UART functionality ?

Dirbaio:
- You can create the uart driver with `&mut` on the peripheral and pins singleton, which means that when the driver is dropped the borrow on the singletons is released, and you can use them again.
- There's no "set baudrate" method currently, but we could have it. [nRF already has it.](https://github.com/embassy-rs/embassy/blob/bc71230cd07296468f2e03c00f9ceddbab67c9d9/embassy-nrf/src/buffered_uarte.rs#LL190C12-L190C24). PR's welcome!.

andres-hurtado-lopez:
> * You can create the uart driver with `&mut` on the peripheral and pins singleton, which means that when the driver is dropped the borrow on the singletons is released, and you can use them again.
> * There's no "set baudrate" method currently, but we could have it. [nRF already has it.](https://github.com/embassy-rs/embassy/blob/bc71230cd07296468f2e03c00f9ceddbab67c9d9/embassy-nrf/src/buffered_uarte.rs#LL190C12-L190C24). PR's welcome!.

Alright I have just created the `set_baudrate`:


```
/// sets baudrate on runtime    
pub fn set_baudrate(&mut self, baudrate: u32) {
	
	let r = T::regs();
	
        let clk_base = crate::clocks::clk_peri_freq();
	
        let baud_rate_div = (8 * clk_base) / baudrate;
        let mut baud_ibrd = baud_rate_div >> 7;
[...]

PR #1243, opened 2023-02-27
RP-PICO UART adding set_baudrate

andres-hurtado-lopez:
The following PR attepts to bring fuctionality to allow change od UART baudrate changes during runtime.

Changes where created under @Dirbaio supervision and discussed on issue:

[https://github.com/embassy-rs/embassy/issues/1241#issuecomment-1445500175]( https://github.com/embassy-rs/embassy/issues/1241#issuecomment-1445500175)

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -299,7 +299,7 @@ impl<'d, T: Instance> Uart<'d, T, Async> {
     }    
 }
 
-impl<'d, T: Instance, M: Mode> Uart<'d, T, M> {
+impl<'d, T: Instance + 'd, M: Mode> Uart<'d, T, M> {
     fn new_inner(
         _uart: impl Peripheral<P = T> + 'd,
         mut tx: PeripheralRef<'d, AnyPin>,
@@ -350,7 +350,7 @@ impl<'d, T: Instance, M: Mode> Uart<'d, T, M> {
                 pin.pad_ctrl().write(|w| w.set_ie(true));
             }
 
-	    Uart::<T,M>::set_baudrate_inner(config.baudrate);
+	    Self::set_baudrate_inner(config.baudrate);
 
             let (pen, eps) = match config.parity {
                 Parity::ParityNone => (false, false),
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -296,7 +296,7 @@ impl<'d, T: Instance> Uart<'d, T, Async> {
             Some(rx_dma.map_into()),
             config,
         )
-    }
+    }    
 }
 
 impl<'d, T: Instance, M: Mode> Uart<'d, T, M> {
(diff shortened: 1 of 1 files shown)


Dirbaio:
Thanks! Could you try the approach in the [last comment](https://github.com/embassy-rs/embassy/issues/1241#issuecomment-1445500175)?

embassy-rp/src/uart/mod.rs
```diff
@@ -369,11 +350,7 @@ impl<'d, T: Instance, M: Mode> Uart<'d, T, M> {
                 pin.pad_ctrl().write(|w| w.set_ie(true));
             }
 
-	    let (baud_ibrd, baud_fbrd) = Uart::<T,M>::baudrate_calculations(config.baudrate);
-
-            // Load PL011's baud divisor registers
-            r.uartibrd().write_value(pac::uart::regs::Uartibrd(baud_ibrd));
-            r.uartfbrd().write_value(pac::uart::regs::Uartfbrd(baud_fbrd));
+	    Uart::<T,M>::set_baudrate_inner(config.baudrate);
```
Dirbaio, line 353: It's nicer to do `Self::set_baudrate_inner`, to not repeat the generic params.

andres-hurtado-lopez: Sure thing boss just a sec

andres-hurtado-lopez: I know but the compiler complaints about not allowing Generics over Self unless I add a lifetime specifier which I'm not confident could alter many other stuff inside the crate itself. Either way I send it through.

Just to let you know here is the error I was telling, in case you want to know.

```
cargo check
    Checking embassy-rp v0.1.0 (/home/andresh/Projects/embassy/embassy-rp)
error[E0109]: type arguments are not allowed on self type
   --> src/uart/mod.rs:353:13
    |
353 |         Self::<T,M>::set_baudrate_inner(config.baudrate);
    |         ----   ^ ^ type argument not allowed
    |         |
    |         not allowed on self type
    |
note: `Self` is of type `uart::Uart<'d, T, M>`
   --> src/uart/mod.rs:85:12
    |
85  | pub struct Uart<'d, T: Instance, M: Mode> {
    |            ^^^^ `Self` corresponds to this type
...
302 | impl<'d, T: Instance, M: Mode> Uart<'d, T, M> {
    | --------------------------------------------- `Self` is on type `Uart` in this `impl`
help: the `Self` type doesn't accept type parameters, use the concrete type's name `Uart` instead if you want to specify its type parameters
    |
353 |         Uart::<T,M>::set_baudrate_inner(config.baudrate);
    |         ~~~~

```

andres-hurtado-lopez: The Self is there are we good now ?

Dirbaio:
LGTM. Can you fix the CI failures?

embassy-rp/src/uart/mod.rs
andres-hurtado-lopez: > LGTM. Can you fix the CI failures? (run `cargo fmt`. I recommend setting your IDE to do that automatically on save)

Sorry mate, gave it a try but the CI seems to be failing on a thing that is not displayed on the log. The only thing that shows is:

1) the curl pulling the batch
2) the setup of the toolchain
3) A long diff with the changes (I guess)
3) a plain vanilla error message without any references or explaination. that says

`
Error: Process completed with exit code 123.
`
And there isn't any thing further. 

Looking at eh `ci.sh` script,  looks like is some stuff related with the `run_elf` function which is for testing on the probes I guess, that is out of my reach unless you teach me.

Dirbaio: Failure is due to not having formatted your code with rustfmt.

andres-hurtado-lopez: Alright mate Done. 

You live you learn !!!!!

First time for me on a PR

Dirbaio (approved):
Thank you!

bors r+

TFleury:
It seems you missed something. `set_baudrate` does not effectively change baudrate at runtime.
It needs an additional register write to latch in the divisors. (See [official SDK](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_uart/uart.c#L93-L95))

PR #1247, opened 2023-02-28
`PacketQueue::init()`

davidedellagiustina:
`PacketQueue` is pretty big, so I added a method to initialize it without requiring an allocation on the stack (which could in fact overflow). Before this PR, the only solution would be to declare a `PacketQueue` instance as a `static mut`, while now one could for example have a `Box<MaybeUninit<PacketQueue<...>>>` and then use `init()` on it.

Ideally, the same work would need to be done for all those structures that own big arrays which could overflow the stack.

### DIFF
--- a/embassy-stm32/src/eth/mod.rs
+++ b/embassy-stm32/src/eth/mod.rs
@@ -5,6 +5,7 @@
 mod _version;
 pub mod generic_smi;
 
+use core::mem::MaybeUninit;
 use core::task::Context;
 
 use embassy_net_driver::{Capabilities, LinkState};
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/eth/mod.rs
```diff
@@ -30,13 +30,42 @@ pub struct PacketQueue<const TX: usize, const RX: usize> {
...
+                tx_buf: [Packet([0; TX_BUFFER_SIZE]); TX],
+                rx_buf: [Packet([0; RX_BUFFER_SIZE]); RX],
+            }
+        }
+    }
+
+    // Allow to initialize a Self without requiring it to go on the stack
+    #[cfg(feature = "nightly")]
+    pub const unsafe fn init(this: &mut core::mem::MaybeUninit<Self>) {
+        let this: &mut Self = unsafe { this.assume_init_mut() };
```
Dirbaio, line 57: creating a `&mut` to uninitialized data is undefined behavior. You have to do all writes to initialize using raw pointers, only after that you can create the `&mut`.

Dirbaio:
Perhaps it's easier to zerofill the whole struct with `mem::zeroed()` or `ptr::write_bytes(0)`?  If it's done within the crate, there's no future-compat issues, because we can guarantee zerofilling is fine.

embassy-stm32/src/lib.rs
```diff
@@ -1,7 +1,13 @@
 #![no_std]
 #![cfg_attr(
     feature = "nightly",
-    feature(type_alias_impl_trait, async_fn_in_trait, impl_trait_projections)
+    feature(
+        type_alias_impl_trait,
+        async_fn_in_trait,
+        impl_trait_projections,
+        const_mut_refs,
```
Dirbaio, line 8: We shouldn't add more nightly features. The plan is for the whole Embassy to work on stable once AFIT/TAIT is stable, which is hopefully near.

davidedellagiustina:
> Perhaps it's easier to zerofill the whole struct with `mem::zeroed()` or `ptr::write_bytes(0)`? If it's done within the crate, there's no future-compat issues, because we can guarantee zerofilling is fine.

Would something like [this](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field) work for you?

davidedellagiustina:
Maybe, but it would definitely be more "obscure" and maybe cumbersome if it needs changes in the future. Tell me which you prefer, and I'll be happy to implement it :D

Dirbaio:
okay! please do the zerofilling then

embassy-stm32/src/lib.rs
davidedellagiustina: So `new()` should just duplicate the code instead of using `init()`? Because there's no way to make `init()` `const` then.

Dirbaio: Yes... It's not too bad if we do zerofilling. `new` can use `mem::zeroed` and `init` can use `ptr::write_bytes(0)`

davidedellagiustina: I just noticed that `core::mem::zeroed()` it not `const`, so not really an option to use in `new()`...

Dirbaio: :( Then we can leave `new` as-is and then do zerofilling for `init` I guess

davidedellagiustina: Implemented in the last commit

embassy-stm32/src/eth/mod.rs
```diff
@@ -39,6 +40,11 @@ impl<const TX: usize, const RX: usize> PacketQueue<TX, RX> {
             rx_buf: [Packet([0; RX_BUFFER_SIZE]); RX],
         }
     }
+
+    // Allow to initialize a Self without requiring it to go on the stack
+    pub unsafe fn init(this: &mut MaybeUninit<Self>) {
```
Dirbaio, line 45: I think this doesn't need to be unsafe, there's no safety invariant the user has to satisfy?

Dirbaio (commented):
LGTM other than the comment. Feel free to `bors r+` yourself when done. 

Thanks for the PR!

bors d+

embassy-stm32/src/eth/mod.rs
davidedellagiustina: Done in last commit

PR #1248, opened 2023-02-28
embassy-time: add async tick() method to Ticker

kbleeke:
Small QOL change so you don't have to add a direct dependency on futures-util to use the Ticker


kbleeke:
sure

Dirbaio (approved):
Thanks!

bors r+

PR #1250, opened 2023-03-01
stm32 spi rxonly slave

pattop:
- stm32/spi: fix occasional data corruption
- stm32/spi: add receive only slave

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -115,7 +115,7 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
             Some(miso.map_into()),
             txdma,
             rxdma,
-            freq,
+            Some(freq),
             config,
         )
     }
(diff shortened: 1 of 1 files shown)


Dirbaio:
SPI slave should be a separate driver, not a mode in the existing driver. 
See https://github.com/embassy-rs/embassy/pull/885#issuecomment-1200500372 and https://github.com/rust-embedded/embedded-hal/pull/396

(they can share code internally, sure, but the user-facing API has to be different, so we shouldn't squash both APIs into the same struct)

embassy-stm32/src/spi/mod.rs
```diff
@@ -476,6 +475,7 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
         let tx_f = crate::dma::write_repeated(&mut self.txdma, tx_request, &clock_byte, clock_byte_count, tx_dst);
 
         unsafe {
+            set_rxdmaen(T::REGS, true);
```
Dirbaio, line 532: there's some versions where the RM explicitly says you must set RXDMAEN=true, start DMAs, set TXDMAEN=true in that exact order. It was F4 iirc. And indeed if you don't do that it doesn't work.

PR #1256, opened 2023-03-02
hal-common/atomic_ring_buffer: add push_bufs() push_slices()

pattop:
This allows a writer to access all of the free space in the buffer, even
when it's wrapping around.


Dirbaio (approved):
Thanks!

PR #1257, opened 2023-03-02
docs: add esp-hal

MabezDev:
I can also add some examples if you think that is a good idea, but I don't want to add any maintenance burden in doing so; of course, I am more than happy to maintain the examples myself.


lulf (approved):
Makes sense I think, thanks for adding it!

bors r+

Dirbaio:
Adding a "dummy" `embassy-esp` directory feels a bit strange. I think the note in the top-level README should be enough.

Dirbaio:
Thanks!

PR #1258, opened 2023-03-02
Swap debug! for trace! in rp gpio

CBJamo:
When using gpio pin changes for things like peripheral interrupts these debug! calls flood defmt, making it difficult to find what you're actually looking for.


chemicstry:
Not sure if you know this, but you can set per-module log level with `DEFMT_LOG=embassy_rp::usb=debug`.

CBJamo:
I only knew about doing it at the crate level. Being able to do it per-module probably makes this unnecessary, yeah.

I think trace is more appropriate for these messages. Possibly one debug message would be appropriate, but getting multiple messages for each pin change feels excessive.

Dirbaio:
Agree they're very verbose, so `trace` makes more sense.  Thanks for the PR!

PR #1259, opened 2023-03-04
embassy_usb: Add split() for cdc_acm

dalegaard:
This adds a `split(self)` function for the `CdcAcmClass`, allowing separate send and receive tasks.


Dirbaio (approved):
Looks great! Thank you!

PR #1262, opened 2023-03-06
bump embedded-storage-async to 0.4

mehmetalianil:
I just haven't found a way to revert the altered stm-metapac contents due to building.


Dirbaio:
you can fix it like this:

```
git fetch
git rebase origin/master
git reset origin/master
```

this'll remove the commits you did, but still leave the changes on disk, pending to be committed. Then you can add the files you do want to commit with `git add` and then do `git commit` and then force-push the changes because this rewrites the history with `git push -f`

lulf (commented):
~The embedded-storage-async changes look good, but the changes to stm32-metapac should not be committed.~

mehmetalianil:
How do we fix merge conflicts in this workflow? Does it wait me to merge it on my local?
Shall I merge embassy/master to this branch?

lulf:
Yes, you can do that. Ideally do a rebase, but I've not looked into how complex resolving the conflict is.

mehmetalianil:
@lulf shall I touch boot/lib.rs test_verify?

lulf:
@mehmetalianil Yes, the in memory 'test flash' needs to be updated as well

mehmetalianil:
So it seems that in order to use async functions in traits we need to use an unstable feature. 
In several parts in embassy this is guarded by 
```rust
#![cfg_attr(
    feature = "nightly",
    feature(async_fn_in_trait)
)]
#![cfg_attr(feature = "nightly", allow(incomplete_features))]
```

Well, I am going to try this out on the bootloader but if you think that it is too intrusive to alter even the package manifest, I would understand.

mehmetalianil:
The issue is that test_verify would then need to be feature gated by nightly, which might be suboptimal. 
I think this change actually enforces nightly for this test to run with async storage traits .Are we okay with that?

lulf:
> So it seems that in order to use async functions in traits we need to use an unstable feature. In several parts in embassy this is guarded by
> 
> ```rust
> #![cfg_attr(
>     feature = "nightly",
>     feature(async_fn_in_trait)
> )]
> #![cfg_attr(feature = "nightly", allow(incomplete_features))]
> ```
> 
> Well, I am going to try this out on the bootloader but if you think that it is too intrusive to alter even the package manifest, I would understand.

That sounds good to me, go for it

Dirbaio (approved):
Thank you!

bors r+

PR #1264, opened 2023-03-07
Add multicast to udp socket

28Smiles:
This should make it possible to implement mDNS for embassy-net, which in turn is essential for smart home

### DIFF
--- a/embassy-net/src/udp.rs
+++ b/embassy-net/src/udp.rs
@@ -50,10 +50,7 @@ impl<'a, D: Driver> UdpSocket<'a, D> {
             udp::PacketBuffer::new(tx_meta, tx_buffer),
         ));
 
-        Self {
-            stack,
-            handle,
-        }
+        Self { stack, handle }
     }
 
     pub fn bind<T>(&mut self, endpoint: T) -> Result<(), BindError>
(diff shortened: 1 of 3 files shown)


embassy-net/src/udp.rs
```diff
@@ -26,13 +25,13 @@ pub enum Error {
     NoRoute,
 }
 
-pub struct UdpSocket<'a> {
-    stack: &'a RefCell<SocketStack>,
+pub struct UdpSocket<'a, D: Driver> {
```
Dirbaio, line 28: It is intentional that the socket is not generic on the driver. It prevents generics spam proliferation.

embassy-net/src/udp.rs
```diff
@@ -143,8 +139,29 @@ impl<'a> UdpSocket<'a> {
     }
 }
 
-impl Drop for UdpSocket<'_> {
+#[cfg(feature = "igmp")]
+impl<'a, D: Driver + smoltcp::phy::Device + 'static> UdpSocket<'a, D> {
+    pub fn join_multicast_group<T>(&self, addr: T) -> Result<bool, smoltcp::iface::MulticastError>
```
Dirbaio, line 144: I don't think these methods should be in UdpSocket. Them being here implies the "joining" is scoped to the socket, ie if you drop the socket you un-join the group. This is not the case though.

these should be in `Stack` instead, mirroring smoltcp.

28Smiles: I made the previous helpers on Stack public and reverted the changes on the Socket.

Dirbaio (approved):
Thank you!

bors r+

PR #1274, opened 2023-03-11
Example using the PIO to drive WS2812 aka Neopixel RGB leds

CBJamo:
This example also uses a pio program compiled at runtime, rather than one built at compile time. There's no reason to do that, but it's probably useful to have an example that does this as well.


Dirbaio (approved):
thanks!
 
 
bors r+

PR #1284, opened 2023-03-17
Fix APB clock calculation for several STM32 families

### LINKED ISSUE #1283: Frequency for APB clocks is miscalculated on some STM32 families
Hi!

I noticed that the calculation of the APB 2 and 3 clock domains on some of the STM32 families is incorrect, namely the L0, L1, L4, L5, U5, WB and WL families.

It appears that the calculation of the register value to a prescaler divisor value was applied twice. For the case where a prescaler divisor of 2 is applied the code will panic as the divisor is correctly calculated to be 2, but then when the calculation is performed the second time, incorrectly using 2 as the register value an overflowing subtraction occurs. In cases where a divisor of 4 or higher is selected the result is an incorrect value is stored in the global clocks variable, which results in incorrect peripheral clock scaling.

ericyanush:
This PR fixes #1283

### DIFF
--- a/embassy-stm32/src/rcc/l0.rs
+++ b/embassy-stm32/src/rcc/l0.rs
@@ -313,7 +313,7 @@ pub(crate) unsafe fn init(config: Config) {
         pre => {
             let pre: Ppre = pre.into();
             let pre: u8 = 1 << (pre.0 - 3);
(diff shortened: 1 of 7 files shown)


Dirbaio (approved):
Nice catch, thank you! 

bors r+

PR #1286, opened 2023-03-18
fix(rp2040): ADC accordingly to datasheet

elpiel:
I'm opening this PR to continue the discussion based on the last PR for fixing the ADC:
https://github.com/embassy-rs/embassy/pull/1245

I still need to tested if ADC now works well with this change as with disabling the Pull resistors as before but it is what's stated in the Datasheet.


Dirbaio:
done in #1623

PR #1288, opened 2023-03-20
fix(rp): spi transfer

### LINKED ISSUE #1181: RP2040 SPI asserts equal length for transfer
It seems that the current implementation of Spi requires the read and write to be of equal lengths:

https://github.com/embassy-rs/embassy/blob/7e251a25509a02f9388a8352522e1a279ad857b1/embassy-rp/src/spi.rs#L385-L388

However, in [embedded_hal_async:0.2.0-alpha.0](https://docs.rs/embedded-hal-async/0.2.0-alpha.0/embedded_hal_async/spi/trait.SpiBus.html#tymethod.transfer) it is stated that:

> It is allowed for read and write to have different lengths, even zero length

How should we fix this misalignment?

elpiel:
Fixes #1181

### DIFF
--- a/embassy-rp/src/spi.rs
+++ b/embassy-rp/src/spi.rs
@@ -404,13 +404,8 @@ impl<'d, T: Instance> Spi<'d, T, Async> {
 
                 if from_len > to_len {
                     let write_bytes_len = from_len - to_len;
-                    // disable incrementation of buffer
-                    tx_ch.regs().ctrl_trig().modify(|ctrl_trig| {
-                        ctrl_trig.set_incr_write(false);
-                        ctrl_trig.set_incr_read(false);
-                    });
-
                     // write dummy data
+                    // this will disable incrementation of the buffers
                     crate::dma::write_repeated(tx_ch, p.dr().ptr() as *mut u8, write_bytes_len, T::TX_DREQ).await
                 }
             }
--- a/embassy-rp/src/spi.rs
+++ b/embassy-rp/src/spi.rs
@@ -384,8 +384,8 @@ impl<'d, T: Instance> Spi<'d, T, Async> {
     }
 
     async fn transfer_inner(&mut self, rx_ptr: *mut [u8], tx_ptr: *const [u8]) -> Result<(), Error> {
-        let (_, from_len) = crate::dma::slice_ptr_parts(tx_ptr);
-        let (_, to_len) = crate::dma::slice_ptr_parts_mut(rx_ptr);
+        let (_, tx_len) = crate::dma::slice_ptr_parts(tx_ptr);
+        let (_, rx_len) = crate::dma::slice_ptr_parts_mut(rx_ptr);
 
         unsafe {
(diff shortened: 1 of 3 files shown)


embassy-rp/src/spi.rs
```diff
@@ -385,26 +386,59 @@ impl<'d, T: Instance> Spi<'d, T, Async> {
...
+        // is held across an await and makes the future non-Send.
+        let tx_transfer = async {
+            let p = self.inner.regs();
+            unsafe {
+                crate::dma::write(&mut tx_ch, tx_ptr, p.dr().ptr() as *mut _, T::TX_DREQ).await;
+
+                if from_len > to_len {
+                    let write_bytes_len = from_len - to_len;
+                    // disable incrementation of buffer
+                    tx_ch.regs().ctrl_trig().modify(|ctrl_trig| {
```
Dirbaio, line 408: is this needed? doesn't `write_repeated` already do it?

Dirbaio:
Looks good overall!

- Have you tested on hardware?
- Could you expand the SPI test to test the new scenarios? Make sure to do test doing more transfers after the uneven transfers to check that the FIFO clearing works fine.

embassy-rp/src/spi.rs
elpiel: yes, you're correct. I will remove it.

elpiel:
@Dirbaio regarding the test, should I just connect the 2 pins together in order to run it?

elpiel:
@Dirbaio fixed a bug and update the test.

One thing to note: The `ram_memory.x` file does not work for me atm because it's giving me this error:

On the other hand, I've been using a different `memory.x` file for other projects and I'm not sure what is the problem.
Any clue why?

`memory.x`:
```
MEMORY {
    BOOT2 : ORIGIN = 0x10000000, LENGTH = 0x100
    FLASH : ORIGIN = 0x10000100, LENGTH = 2048K - 0x100
    RAM   : ORIGIN = 0x20000000, LENGTH = 256K
}

EXTERN(BOOT2_FIRMWARE)

SECTIONS {
    /* ### Boot loader */
    .boot2 ORIGIN(BOOT2) :
    {
        KEEP(*(.boot2));
    } > BOOT2
} INSERT BEFORE .text;
```

PS: I'm running picoprobe with probe-run and I've also experienced issues when using probe-run but haven't had time to investigate that issue - sometimes (could be after 10 minutes, could be after 5 seconds) the debugging serial logging breaks because of a architecture specific error.

Dirbaio:
> One thing to note: The ram_memory.x file does not work for me atm because it's giving me this error:

It's a special linker script that puts all code in RAM, not FLASH. This is so that it boots faster (no need to flash) and preserves flash life (we're not flashing every time the test runs). probe-run doesn't support booting from RAM, you have to use `teleprobe local run`. `.cargo/Config.toml` is configured to do that.

Dirbaio (approved):
LGTM! 

Could you fix rustfmt? then feel free to `r+` yourself.

bors d+

PR #1290, opened 2023-03-20
Add async can for STM32

ericyanush:
I'm beginning work on adding an async CAN implementation for the STM32 devices within Embassy. Before I go too far with this, I'd like to get some feedback on the direction I'm heading. This is a barebones implementation right now, that only handles async TX, and doesn't cover off any of the errors that can occur. Once I know the general idea that I'm chasing is good, adding appropriate error handling and expanding this to cover off all the error scenarios should be easy enough.

Thanks!

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -68,6 +68,7 @@ stm32-fmc = "0.2.4"
 seq-macro = "0.3.0"
 cfg-if = "1.0.0"
 embedded-io = { version = "0.4.0", features = ["async"], optional = true }
+heapless = { version = "0.7.5", default-features = false }
 
 [dev-dependencies]
 critical-section = { version = "1.1", features = ["std"] }
--- a/embassy-stm32/src/can/bxcan.rs
+++ b/embassy-stm32/src/can/bxcan.rs
@@ -80,7 +80,6 @@ impl<'d, T: Instance> Can<'d, T> {
     }
 
     pub async fn transmit_async(&mut self, frame: &Frame) {
-        defmt::info!("Staring async frame transmission");
         let tx_status = self.queue_transmit(frame).await;
         self.wait_transission(tx_status.mailbox()).await;
     }
@@ -88,10 +87,8 @@ impl<'d, T: Instance> Can<'d, T> {
     async fn queue_transmit(&mut self, frame: &Frame) -> bxcan::TransmitStatus {
         poll_fn(|cx| {
             if let Ok(status) = self.can.transmit(frame) {
-                defmt::info!("Frame queued successfully in mb{}", status.mailbox());
                 return Poll::Ready(status);
             }
-            defmt::info!("Mailboxes full, waiting to queue");
             T::state().tx_waker.register(cx.waker());
             Poll::Pending
         })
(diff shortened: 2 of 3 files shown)


timokroeger:
cc @chemicstry

chemicstry:
I have written an out-of-tree async can driver for use with canadensis library: https://gist.github.com/chemicstry/594725b6b6dfdd2fbaed98669a0a4186
Be warned that it's a mess, but you might find something useful. `Queue` type is just a ringbuffer with async support.

Some things that I observed:
- You really want a queue for receiving. At 1mbps, the inter-frame time can be as low as 67us and with just 3 receive mailboxes per FIFO you will be missing frames if using only wakers.
- The `bxcan` library was getting in my way a lot, so it might be worth investigating if using just the registers from PAC would be cleaner.
- Some protocols (i..e. cyphal) need RX timestamps, so it would be a nice to have under `time` feature gate.

ericyanush:
@chemicstry thanks for the pointers! I'm fairly new to the world of async in this context, and hadn't considered the latency of the wakers in terms of filling the rx mailboxes.

I'll take a look at if it makes more sense to build onto of the bxcan lib, or if it's smoother to just the pac directly. Not that it can't be solved with a little extra feature gating, but I know that not all bxcan peripherals support the time triggered functionality. I know for sure that it's not supported in the L432KC/KB and L442KC, as it's listed in the errata sheet as not supported.

chemicstry:
IIRC time triggered can (TT-CAN) is not working on any of the chips. It doesn't have any documentation, just a few registers that somehow made it through to the reference manual. What I do instead is `Instant::now()` in RX IRQ, which is precise enough if you set IRQ priorities right.

xoviat:
@chemicstry What do you think about this as an option? A `Can` struct created with a fixed read and write buffer at initialization that has one method: `create_subscriber(ids: [u8])`, which returns a `Subscriber` struct. The `Subscriber` struct has two methods: `read` and `write`. The read method blocks-async until a message with one of the subscriber's ids is available, while the write method returns immediately unless the buffer is full, in which case it async-blocks until there is space on the write buffer. The can struct updates the buffer and wakes the appropriate subscribers, as needed. This model would allow multiple tasks to use the bus, mirroring embedded-hal's spi-device.

ericyanush:
Closing this until I've completed a reasonable implementation to prevent notification spam for pushing to the branch I based this draft pr on.

xoviat:
@ericyanush Don't let my previous comments stop you from moving forward on this.

Dirbaio:
Also don't worry about notification spam! Github has tools for people to manage their notifications, like unwatching one particular PR.

Having a PR open for WIP code is good to give it visibility, so people know it's being worked on so they can help, or avoid starting duplicate work, or give feedback.

schphil:
Hi, thanks for the work so far @ericyanush.  I'm interested in using the async can implementation in a project. `async fn transmit` and `async fn receive_event` seem to work already. What would be the next steps from your point of view? If I can be of any help let me know.

ericyanush:
@schphil There is certainly some work left to do; I feel like the public API I have exposed is a bit rough and should be refined. Otherwise, I have a version of the code you see, plus a few updates running in a pre-production network of devices, and it has been stable thus far. I will try over the next week or so to find time to extract the updated version the async code out into a public repo so it can be looked at and enable suggestions/enhancements from others.

PR #1294, opened 2023-03-23
Add support for `QSPI` in `stm32`

Mirror0:
Implemented with help of Tomasz Grześ <tomasz.grzes@gmail.com>.

Tested only on stm32f777zi.

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -48,6 +48,8 @@ pub mod crc;
 ))]
 pub mod flash;
 pub mod pwm;
+#[cfg(quadspi)]
+pub mod qspi;
 #[cfg(rng)]
 pub mod rng;
 #[cfg(sdmmc)]
@@ -60,7 +62,6 @@ pub mod usart;
 pub mod usb;
 #[cfg(otg)]
 pub mod usb_otg;
-
 #[cfg(iwdg)]
 pub mod wdg;
 
--- a/embassy-stm32/src/qspi/mod.rs
+++ b/embassy-stm32/src/qspi/mod.rs
@@ -1,0 +1,342 @@
+#![macro_use]
+
+use embassy_hal_common::{into_ref, PeripheralRef};
+
+use crate::dma::TransferOptions;
+use crate::gpio::sealed::AFType;
+use crate::gpio::AnyPin;
+use crate::pac::quadspi::Quadspi as Regs;
+use crate::rcc::RccPeripheral;
+use crate::{peripherals, Peripheral};
+
+pub struct QspiWidth;
+
+#[allow(dead_code)]
+impl QspiWidth {
+    pub const NONE: u8 = 0b00;
+    pub const SING: u8 = 0b01;
+    pub const DUAL: u8 = 0b10;
+    pub const QUAD: u8 = 0b11;
+}
+
+struct QspiMode;
+
+#[allow(dead_code)]
+impl QspiMode {
+    pub const INDIRECT_WRITE: u8 = 0b00;
+    pub const INDIRECT_READ: u8 = 0b01;
+    pub const AUTO_POLLING: u8 = 0b10;
+    pub const MEMORY_MAPPED: u8 = 0b11;
+}
+
+pub struct QspiTransaction {
+    pub iwidth: u8,
+    pub awidth: u8,
+    pub dwidth: u8,
+    pub instruction: u8,
+    pub address: Option<u32>,
+    pub dummy: u8,
+    pub data_len: Option<usize>,
+}
+
+impl Default for QspiTransaction {
+    fn default() -> Self {
+        Self {
+            iwidth: QspiWidth::NONE,
+            awidth: QspiWidth::NONE,
+            dwidth: QspiWidth::NONE,
+            instruction: 0,
+            address: None,
+            dummy: 0,
+            data_len: None,
+        }
+    }
+}
+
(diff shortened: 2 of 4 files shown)


Mirror0:
Hello @Dirbaio,

How can I add the required features to the CI, so that the `build.rs` sees the `qspi` module? Is there some script where I can declare them?

embassy-stm32/src/lib.rs
```diff
@@ -48,6 +48,8 @@ pub mod crc;
 ))]
 pub mod flash;
 pub mod pwm;
+#[cfg(all(quadspi, feature = "unstable-pac"))]
```
Dirbaio, line 51: this shouldn't be gated on `unstable-pac`

Dirbaio:
The `quadspi` feature gets defined here, if the chip has some peripheral that has `quadspi` registers: https://github.com/Mirror0/embassy/blob/ad4970ae03190a6c966fed2c4d4a72b196cd9f1c/embassy-stm32/build.rs#L26

I think the issue is the comment I wrote. CI tries to build without `unstable-pac`, in which case the pin/dma impls get generated anyway but `crate::qspi` doesn't exist.

embassy-stm32/src/lib.rs
Mirror0: Unfortunately I don't think it can be passed without `unstable-pac` feature, as the module uses `crate::gpio::low_level::AFType` which without `unstable-pac` cannot resolve `low_level` module and `AFType` from it is used in configuration here: https://github.com/Mirror0/embassy/blob/ad4970ae03190a6c966fed2c4d4a72b196cd9f1c/embassy-stm32/src/qspi/mod.rs#L103

Have you have any suggestion that I can use to avoid this dependencies? Because as far as I can see, only this prevents the build without `unstable-pac`

Dirbaio: you can use `crate::gpio::sealed` which is always available (within the crate)

Mirror0:
Thank you @Dirbaio your suggestion solves the issue with building without `unstable-pac`, but there is another problem. CI fails with an error saying that `Peripheral::QUADSPI` does not satisfy some trait bounds. But if I build the same code manually from `embassy-stm32` root with my board selected (`cargo build -F stm32f777zi`) the build is successful.

Can you help me with this? Because I cannot see where the underlying issue may be and what is different between the local stm32 module build and the build performed by CI.

Dirbaio:
CI prints the set of features that caused the failure:

```
error: could not compile `embassy-stm32` due to 14 previous errors
     features: ["_time-driver", "defmt", "embedded-hal-1", "embedded-hal-async", "exti", "nightly", "stm32f413vh", "time", "time-driver-any", "unstable-traits"]
```

If you try that combination of features locally, you can repro the failure:

```
[dirbaio@mars embassy-stm32]$ git checkout stm32-qspi-support 
Already on 'stm32-qspi-support'
Your branch is up to date with 'Mirror0/stm32-qspi-support'.
[dirbaio@mars embassy-stm32]$ cargo check --target thumbv7em-none-eabi --features _time-driver,defmt,embedded-hal-1,embedded-hal-async,exti,nightly,stm32f413vh,time,time-driver-any,unstable-traits
...
error[E0277]: the trait bound `peripherals::QUADSPI: rcc::RccPeripheral` is not satisfied
...etc
```

The issue is actually that ST names the peripheral `QUADSPI` but the RCC regs are sometimes `QUADSPIEN` and `QSPIEN`, so stm32-data wasn't picking them up in the latter case. Fixed in https://github.com/embassy-rs/stm32-data/commit/78c43e0dba6750d2f66b66663d52b05b5f7fd520, this should work now if you rebase on top of #1302

embassy-stm32/src/qspi/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+        unsafe {
+            T::REGS.cr().modify(|v| v.set_dmaen(false));
+            self.setup_transaction(QspiMode::INDIRECT_WRITE, &transaction);
+
+            while !T::REGS.sr().read().tcf() {}
+            T::REGS.fcr().modify(|v| v.set_ctcf(true));
+        }
+    }
+
+    pub fn read(&mut self, buf: &mut [u8], transaction: QspiTransaction) {
```
Dirbaio, line 184: for consistency with the other drivers, could you name the blocking functions `blocking_`? `blocking_write`, `blocking_write_dma`, etc.

embassy-stm32/src/qspi/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+
+#[allow(dead_code)]
+impl QspiMode {
+    pub const INDIRECT_WRITE: u8 = 0b00;
+    pub const INDIRECT_READ: u8 = 0b01;
+    pub const AUTO_POLLING: u8 = 0b10;
+    pub const MEMORY_MAPPED: u8 = 0b11;
+}
+
+pub struct QspiTransaction {
```
Dirbaio, line 32: I think naming this `TransferConfig` is better. It's the analogue to `Config`, but for each transfer.

Mirror0: Fixed

embassy-stm32/src/qspi/mod.rs
Mirror0: Fixed

Dirbaio (approved):
Thank you! :)

bors r+

PR #1295, opened 2023-03-23
(embassy-rp): Add embedded-io blocking Read + Write for BufferedUart

MathiasKoch:


Dirbaio (approved):
:+1: 

bors r+

MathiasKoch:
Hmm... missed a place, will fix right away

PR #1296, opened 2023-03-25
Allow embassy_sync::Pipe to handle wraparounds (better)

jamesmunns:
Currently, if you are at the wraparound condition, reads and writes will only fill up to the wraparound, even if there is additional space after the wraparound.

Though the caller COULD manually handle this case (see the `manual_wraparound` test) - this could lead to split messages in the case of lock contention.

This PR adds logic to allow read and write to proactively "try again" in case of a potential wraparound condition, which would prevent needing to handle the logic manually, and (IMO) follows a more expected behavior from an API perspective.

I noticed the former behavior in my testing, where messages appeared truncated in my usage of Pipe, which wasn't wrong, but I assumed that incomplete read/writes were "fatal", at least until additional write/reads occured.

This logic COULD potentially live in `ring_buffer` instead, but I'm not sure it would work well with the current API there.

### DIFF
--- a/embassy-sync/src/pipe.rs
+++ b/embassy-sync/src/pipe.rs
@@ -179,11 +179,23 @@ impl<const N: usize> PipeState<N> {
     }
 
     fn try_read_with_context(&mut self, cx: Option<&mut Context<'_>>, buf: &mut [u8]) -> Result<usize, TryReadError> {
+        // If the user provided buffer is empty, there's nothing to do here.
+        if buf.is_empty() {
+            return Ok(0);
+        }
+
+        // If the buffer WAS full prior to reading, wake the writer in case someone
+        // was waiting to push more data.
         if self.buffer.is_full() {
             self.write_waker.wake();
         }
 
+        // We try (up to) twice, in case we wrap around, and can steal from the tail
+        // AND the head of the ring buffer.
         let available = self.buffer.pop_buf();
+
+        // If we're empty on first try, nothing to be done. Register a waker
+        // and return an empty error.
         if available.is_empty() {
             if let Some(cx) = cx {
                 self.read_waker.register(cx.waker());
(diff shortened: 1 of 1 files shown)


Dirbaio:
`write` and `read` in `Pipe` work like `std::io::Read/Write`. They only guarantee is "if there's space, at least 1 byte is written".

In the soupstone case, the fix is to use `.write_all` (from `embedded-io`) here instead of `.write`:

```rust
STDOUT.write(b"hello, world!\r\n").await;
```

- About docs: I agree they're suboptimal indeed, that's probably the source of the confusion. There was some language copypasted from Channel talking about "messages" or "values" which is wrong. I've fixed this in #1299 .
- About the change in this PR: Both the current behavior (without this PR) and the new behavior (with) are correct w.r.t the `Read/Write` guarantees. So it comes down about perf and principle-of-least-surprise:
   - perf: chances are the user is already writing their own loop (or using `write_all`, `read_exact` etc) so having `write/read` do 2 tries is not needed, I think this is a slight perf loss (probably small enough to not matter)
   - principle-of-least-surprise: currently `write` can return early in 2 cases: wraparound and full. After this PR, it only does inthe full case. This makes code that uses `write` by mistake (that should be using `write_all`) less likely to fail, but it can still fail. I'm not sure if "hiding more" the footgun is a win, if it doesn't fully eliminate it.

> this could lead to split messages in the case of lock contention.

I'm not very convinced that this is a disadvantage. 
- Using `write` with this PR is not enough to ensure no corruption, messages can still be truncated if the pipe buffer fills up.
- To avoid that, you have to use `write_all`, and that can interleave messages even with this PR (because it does multiple calls to `write`).
- The full fix is to use `write_all`, and share a single writer with an async mutex, instead of having two writers. Once you do that, it also works fine

jamesmunns:
That's fair! I agree that nothing is *wrong* about the current behavior, but I was surprised that a Pipe with enough free space (albeit with wraparound) would consume only a portion of the provided read/write. I think #1299 helps a lot to clear this up.

Totally understand if you want to keep the internals interface simple, and require users to do their own looping. The suggestion to use a split reader/writer with an async mutex I think would also work in my case.

No hard feelings if you close this :)

Dirbaio:
Yeah, I agree it's surprising... 

though arguably, if all code doing a single `write` like that is likely buggy, isn't it better to surface the bug ASAP to the user?

Like, currently the bug will happen on wraparound, which is likely to be noticed. With this PR it'll only happen on buffer full, which might be hard to reproduce, or only happen when the system is under load...

PR #1297, opened 2023-03-25
feat(stm32): Support multiple flash regions

rmja:
This depends on https://github.com/embassy-rs/stm32-data/pull/176

This is a general overhaul of the flash module to support multiple erase sizes.
Overall this PR contains:
* Move complex sector computation to embassy-hal-common to allow for tests
* Implement `FlashRegion` trait for each available flash region
* Add Flash::into_regions() to get each region.
* Implement embedded-storage traits for each region to support different erase sizes
* Split family write operations into begin/do/end
* Protection against simultaneous writes/erases for each split region is done through a global mutex


rmja:
Something that could be interesting to add is to check if the area to erase is already erased.
This check could potentially ensure that the flash does not break with too many erases in case where, during an erase, and because the mcu is stalled while this happens, the app may panic because of some unrelated timeout. When it re-boots after the panic then the erase is in fact completed, ready to do some writes which takes way faster, and therefore are likely not to cause any unrelated timeout panics.

embassy-hal-common/src/lib.rs
```diff
@@ -10,4 +10,5 @@ mod macros;
 mod peripheral;
 pub mod ratio;
 pub mod ring_buffer;
+pub mod stm32;
```
Dirbaio, line 13: `hal-common` is for chip-independent stuff shared between all HALs. It shouldn't have stm32 stuff.

I know there's no way to run the tests there... but still, please put them there, and they'll run if someday we figure out how to run them.

embassy-stm32/src/flash/mod.rs
```diff
@@ -17,74 +19,175 @@ pub struct Flash<'d> {
...
+            family::clear_all_err();
+            family::unlock();
+            let res = family::blocking_erase(start_address, end_address);
+            family::lock();
+            res
+        }
+    }
+}
+
+fn take_lock_spin() -> MutexGuard<'static, CriticalSectionRawMutex, ()> {
```
Dirbaio, line 183: never use spinlocks. They can cause deadlocks if you spin on a higher priority context waiting for a lower priority context to release it.

In this particular case, it should be enough to wrap all the operations that could race with `critical_section::with`.

embassy-stm32/src/flash/mod.rs
```diff
@@ -17,74 +19,175 @@ pub struct Flash<'d> {
...
 }
 
+pub trait FlashRegion {
+    const BASE: usize;
+    const SIZE: usize;
+    const ERASE_SIZE: usize;
+    const WRITE_SIZE: usize;
+    const ERASE_VALUE: u8;
+
+    fn blocking_read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Error> {
```
Dirbaio, line 124: it'd be cool if these could call the methods from `Flash` (or both call some shared method) to avoid duplication.

embassy-stm32/src/flash/other.rs
```diff
@@ -0,0 +1,7 @@
+pub trait FlashRegion {
```
Dirbaio, line 1: you can put this in `mod.rs`, that's where all stuff that's shared across versions goes.

embassy-stm32/src/lib.rs
```diff
@@ -47,6 +47,13 @@ pub mod crc;
     flash_l0, flash_l1, flash_wl, flash_wb, flash_l4, flash_f3, flash_f4, flash_f7, flash_h7
 ))]
 pub mod flash;
+#[cfg(not(any(
```
Dirbaio, line 50: this is a bit strange. Perhaps it's better to remove the `cfg` here, and put them inside `flash/mod.rs`. So `flash/mod.rs` contains the always-available stuff (`FlashRegion`), and then there's a cfg-guarded `flash/common.rs` with all the stuff that's available only on supported chips (mostly what's now in `flash/mod.rs`.)

embassy-stm32/src/flash/mod.rs
```diff
@@ -17,74 +19,175 @@ pub struct Flash<'d> {
...
 
+static REGION_LOCK: Mutex<CriticalSectionRawMutex, ()> = Mutex::new(());
+
 impl<'d> Flash<'d> {
     pub fn new(p: impl Peripheral<P = FLASH> + 'd) -> Self {
         into_ref!(p);
         Self { _inner: p }
     }
 
+    pub fn into_regions(self) -> FlashRegions {
```
Dirbaio, line 30: The lifetime here needs to carry over into `FlashRegions` and then into the individual regions.

Otherwise you can create a `Flash` with borrowed (`&mut`) peripherals, then convert it to `FlashRegions`, which releases the borrow because the lifetime is no longer active, then create *another* `Flash` without dropping the `FlashRegions` first.

embassy-stm32/build.rs
```diff
@@ -103,6 +103,70 @@ fn main() {
...
+        })
+        .unzip();
+
+    flash_regions.extend(quote! {
+        pub struct FlashRegions {
+            #(#fields),*
+        }
+
+        impl FlashRegions {
+            pub(crate) const fn take() -> Self {
```
Dirbaio, line 160: `take` sounds like it's removing the original (e.g. from an Option). I'd name this `new`.

embassy-hal-common/src/stm32/flash/f4.rs
```diff
@@ -0,0 +1,119 @@
...
+pub const SECOND_BANK_SECTOR_OFFSET: u8 = 12;
+
+#[derive(Debug, PartialEq)]
+pub struct FlashSector {
+    pub index: u8,
+    pub start: u32,
+    pub size: u32,
+}
+
+pub fn get_sector(address: u32, dual_bank: bool, flash_size: u32) -> FlashSector {
```
Dirbaio, line 14: This is hardcoding the sectors for F4/F7. Could you instead generate some static array in `build.rs` and then iterate it? This way the same code would work for all chips, including those with uniform sectors.

Something like `(size, sector_count, bank)` could work. For example for an F4 it could be:

```rust
static SECTORS: &[Sector] = &[
   Sector{ size: 16*1024, count: 4, bank: 0},
   Sector{ size: 64*1024, count: 1, bank: 0},
   Sector{ size: 128*1024, count: 3, bank: 0},
   Sector{ size: 16*1024, count: 4, bank: 1},
   Sector{ size: 64*1024, count: 1, bank: 1},
   Sector{ size: 128*1024, count: 3, bank: 1},
];
```

and for a single-bank uniform-sector chip it could be:


```rust
static SECTORS: &[Sector] = &[
   Sector{ size: 1*1024, count: 128, bank: 0},
];
```

(it's fine if the code is more complex than needed for chips with uniform sector sizes. similarly, it's fine if the `bank` stays unused for single-bank chips. We can optimize for simplicity for now, optimize later).

embassy-stm32/src/flash/f4.rs
```diff
@@ -34,91 +34,59 @@ pub(crate) unsafe fn unlock() {
...
+            return false;
+        }
+        address += sector.size;
+    }
+    address == end_address
+}
 
-    while addr < to {
-        let sector = get_sector(addr, dual_bank);
+pub(crate) unsafe fn blocking_erase(start_address: u32, end_address: u32) -> Result<(), Error> {
```
Dirbaio, line 76: perhaps it'd help simplify things if the family-specific code takes a sector number to erase, and the responsibility of looping is on the common code? same for `is_eraseable_range`.

rmja:
@Dirbaio Thank you for your time and the comments! I will align to the suggestions asap.

embassy-hal-common/src/lib.rs
rmja: I have moved these back - and, I found a way to actually run the tests from within embassy-stm32.

embassy-stm32/src/flash/mod.rs
rmja: Ahh, I did not realize that p could be a reference. This is now fixed.

embassy-stm32/build.rs
rmja: Renamed to new...

embassy-hal-common/src/stm32/flash/f4.rs
rmja: This is not as easy as it sounds:) The reason is that for the F4 family, it is determined at runtime in a register setting if we are using a single or dual bank configuration. I have made an alternative proposal where we have a get_sector() per family.

I can also go with something where each family has a `get_sectors() -> &'static [Sector]` that returns the sector table depending on the the bank config, but that would require at least `#[feature(const_for)]` for a reasonable implementation.

Let me know what sounds best here...

embassy-stm32/src/flash/mod.rs
rmja: As stated above, I could use a suggestion on how to do this without `critical_section::with` as I need some way of:
1. unlock()
2. write()/erase()
3. lock()

where I am guaranteed that I do the entire sequence. BUT, I do not want to disable interrupt during the entire sequence. The reason is that it may take a very long time to write/erase. I know that the mcu may stall during e.g. erase, but only per sector. It will wake up and handle higher priority interrupts between each sector. Wrapping the entire sequence with a cs will multiply the time with the sector count which I think we can do better.

embassy-hal-common/src/stm32/flash/f4.rs
rmja: I found a way to do this:-)

embassy-stm32/src/flash/mod.rs
Dirbaio: Agreed, and it should be doable, yes! probably just taking a CS fo each operation? 

Also perhaps lock/unlock for each operation. It feels dangerous to let irqs fire while flash is unlocked (though the current code already does that, so idk). lock/unlock is instant compared to erase/write, so it shouldn't make a difference in perf.

```rust
for sector in some_sector_iter {
   critical_section::with(|_| {
      unlock();
      erase(sector);
      lock();
   });
}
```

rmja: Done!

Dirbaio:
@rmja LGTM. I've released metapac v3 with the stm32-data changes, but seems there's still some build failures. Could you take a look?

rmja:
@Dirbaio I have fixed the issue, but bootloader examples still fail to compile. My plan is to align these depending on https://github.com/embassy-rs/embassy/pull/1314 which enables for a bootloader block size decoupled from the erase sizes of the dfu and active flashes. Let me know if that is ok.

rmja:
@Dirbaio one more thing that I forgot to mention that changed in this PR. I have set ReadNorFlash::READ_SIZE = 1. Before it was 4. Are you aware of any chips that cannot read a single flash byte at a time?

Dirbaio:
> My plan is to align these depending on https://github.com/embassy-rs/embassy/pull/1314 which enables for a bootloader block size decoupled from the erase sizes of the dfu and active flashes. Let me know if that is ok.

Okay, will take a look.

> I have set ReadNorFlash::READ_SIZE = 1. Before it was 4. Are you aware of any chips that cannot read a single flash byte at a time?

yeah that should be fine. They all can read bytes, I don't know why it was set to 4.

Dirbaio (approved):
Let's gooo :rocket: :rocket: 
Thanks for your work, this is great! :) 

bors r+

PR #1303, opened 2023-03-27
Add logging and interface for debugging USB buffer usage

jamesmunns:


embassy-usb/src/lib.rs
```diff
@@ -165,6 +165,25 @@ struct Interface {
...
+    pub device_descriptor_used: usize,
+    /// Number of config descriptor bytes used
+    pub config_descriptor_used: usize,
+    /// Number of bos descriptor bytes used
+    pub bos_descriptor_used: usize,
+    /// Number of msos descriptor bytes used
+    ///
+    /// Will be `None` if the "msos-descriptor" feature is not active.
+    /// Otherwise will return Some(bytes).
+    pub msos_descriptor_used: Option<usize>,
```
jamesmunns, line 182: I decided to go with an option here to differentiable between "msos available but not used" and "msos not active", rather than feature flagging (since this is not on a hot path/persistent usage or anything). I figured that would make printing and such less tedious for folks.

Let me know if you'd prefer a feature flag'd usize instead.

embassy-usb/src/builder.rs
```diff
@@ -201,6 +201,20 @@ impl<'d, D: Driver<'d>> Builder<'d, D> {
         self.config_descriptor.end_configuration();
         self.bos_descriptor.end_bos();
 
+        info!("USB: device_descriptor used: {}", self.device_descriptor.position());
+        info!("USB: config_descriptor used: {}", self.config_descriptor.position());
+        info!("USB: bos_descriptor_buf used: {}", self.bos_descriptor.writer.position());
+        #[cfg(feature = "msos-descriptor")]
+        info!("USB: device_descriptor used: {}", msos_descriptor.len());
+        if self.control_buf.len() != self.config.max_packet_size_0.into() {
+            warn!(
```
jamesmunns, line 210: I'm not sure if this should be an error (either the log level, a panic, or otherwise). Let me know if you have thoughts.

Dirbaio: I'm sorry, I completely brainfarted on Matrix. I told you the `control_buf` size must be equal to the EP0 MPS, but that's not true at all. Control payloads get fragmented into multiple MPS-sized packets, so the buffer size can be anything (usually bigger, but could be smaller too). So this check is not correct :(

embassy-usb/src/lib.rs
Dirbaio: lgtm. I think we should remove that feature anyway, or at least not make it change the `Builder::new()` signatur, because that's quite annoying.

embassy-usb/src/builder.rs
jamesmunns: Will remove that warning!

Dirbaio (approved):
thank you!

bors r+

PR #1307, opened 2023-03-28
(embassy-stm32): add embedded-io blocking Read + Write for BufferedUart

MathiasKoch:


embassy-stm32/src/usart/buffered.rs
```diff
@@ -197,6 +197,40 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
         .await
     }
 
+    fn inner_blocking_read<'a>(&'a self, buf: &'a mut [u8]) -> Result<usize, Error> {
```
Dirbaio, line 200: you can remove this lifetime

embassy-stm32/src/usart/buffered.rs
```diff
@@ -236,6 +270,39 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
         .await
     }
 
+    fn inner_blocking_write<'a>(&'a self, buf: &'a [u8]) -> Result<usize, Error> {
```
Dirbaio, line 273: same

Dirbaio (approved):
LGTM other than the lifetime nit. Thank you!

bors d+

PR #1309, opened 2023-03-30
Preliminary support for bootloader regions

rmja:
This PR makes it easier to add support for multiple flash regions in the bootloader. It contains:
* Redefintion of the `Flash` trait. It now only includes required fields. Any blocking or async storage traits must be added manually.
* Let there be both a `FlashConfig` and a `BlockingFlashConfig`
* Let Bootloader and FirmwareUpdater have the flash config as a member so that they internally can request mut ref's for each of the flash regions depending on their need.
* Remove Bootloader::default() and FirmwareUpdater::default() - they must now be new'ed with their respective config.
* Let firmware writer track the written number of bytes instead of relying on passing the offset

I have only aligned the embassy-boot-stm32 project with the changes. Depending on feedback I can align the others as well.

I am splitting the PR here to avoid too many changes when supporting multiple regions.

cc @lulf

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -243,89 +243,89 @@ impl BootLoader {
     /// |       DFU |            3 |      3 |      2 |      1 |      3 |
     /// +-----------+--------------+--------+--------+--------+--------+
     ///
-    pub fn prepare_boot<P: BlockingFlashConfig>(
+    pub fn prepare_boot<F: BlockingFlashConfig>(
         &mut self,
-        p: &mut P,
+        flash: &mut F,
         magic: &mut [u8],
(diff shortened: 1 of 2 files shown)


lulf (commented):
I think the changes to BootLoader type is fine, as it's not generally wrapped into the application, so the added generics doesn't get in the way.

However, I think adding generics to the FirmwareUpdater makes it harder to store and pass as a type. For instance for DFU over BLE, it would mean I now have to use generics or create newtypes instead of just using the updater as is.


I'm not sure how it makes it easier to support multiple regions this way, what is your plan for that?

rmja:
Closed in favor of https://github.com/embassy-rs/embassy/pull/1310

PR #1310, opened 2023-03-30
Let FirmwareWriter internally track the number of written bytes

rmja:
This also removes the block_size parameter to the writer. It seems to be an additional size that is not really used. The `NorFlash` should internally support writing to chunks of its write size.

### DIFF
--- a/embassy-boot/boot/src/lib.rs
+++ b/embassy-boot/boot/src/lib.rs
@@ -831,7 +831,6 @@ impl FirmwareUpdater {
         offset: usize,
         data: &[u8],
         flash: &mut F,
-        block_size: usize,
     ) -> Result<(), FirmwareUpdaterError> {
         assert!(data.len() >= F::ERASE_SIZE);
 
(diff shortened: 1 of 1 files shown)


embassy-boot/boot/src/lib.rs
```diff
@@ -1097,99 +1095,48 @@ impl FirmwareUpdater {
...
-
-        let mut write_offset = self.0.from + offset;
-        for chunk in data.chunks(block_size) {
-            trace!("Wrote chunk at {}: {:?}", write_offset, chunk);
-            flash.write(write_offset as u32, chunk).await?;
-            write_offset += chunk.len();
-        }
-        /*
-        trace!("Wrote data, reading back for verification");
-
```
lulf, line 1136: FWIW, the reason for this chunking is that some external flashes cannot write the full page size, so it need to be chunked.

rmja: As far as I can tell that is a violation of the NorFlash trait then. It does not limit the number of blocks to write at once. I wonder if the block size for those devices could be included in their implementation of the trait?

lulf: Hmm, maybe. But the flash implementation is a generic 'Qspi' implementation like this https://github.com/embassy-rs/embassy/blob/master/embassy-nrf/src/qspi.rs#L571 where the 'block size' is not known by Qspi, and I'm not sure if  maybe that would be too specific to put in there.

lulf: Why did you close the PR btw?

PR #1311, opened 2023-03-31
Split bootloader implementation into multiple files

rmja:
No other change has been made. I think that I have all the types re-exported in lib.rs as before.


lulf:
Agreed, this should've been done a long time ago :) Thanks :rocket: 

bors r+

PR #1312, opened 2023-03-31
Let bootloader partition have read/write/erase operations

rmja:
This change should not have any breaking changes.


Dirbaio (approved):
thanks!

bors r+

PR #1313, opened 2023-03-31
(embassy-stm32): rework bufferedUart to get rid of PeripheralMutex

MathiasKoch:
New implementation is very similar to the implementation of embassy-nrf & embassy-rp. 

Also adds embedded-hal traits to bufferedUart.

**NB**: Still needs testing on actual hardware


Dirbaio (approved):
ooh, this is great. Much appreciated, thank you! :) 

bors r+

PR #1314, opened 2023-03-31
Let the bootloader block-size be decoupled from the flash

rmja:
This changes the block copy operation sequence.
Previously the assumption was that we could have an entire flash-erase-size in memory. Now, erases are done on a per-need basis, and writes are always tracked on a well-defined block size.

This allows the block-size tracked by the bootloader to be much smaller than the flash-erase-size.

The work is based on https://github.com/embassy-rs/embassy/pull/1312
cc @lulf

### DIFF
--- a/embassy-boot/boot/src/boot_loader.rs
+++ b/embassy-boot/boot/src/boot_loader.rs
@@ -42,8 +42,14 @@ pub trait FlashConfig {
     /// The update progress is tracked in blocks of this size.
     ///
     /// The size of a block must be such that
-    /// 1) BLOCK_SIZE >= ACTIVE::WRITE_SIZE && ACTIVE::WRITE_SIZE % BLOCK_SIZE == 0 && ACTIVE::capacity() % BLOCK_SIZE == 0
-    /// 2) BLOCK_SIZE >= DFU::WRITE_SIZE && DFU::WRITE_SIZE % BLOCK_SIZE == 0 && DFU::capacity() % BLOCK_SIZE == 0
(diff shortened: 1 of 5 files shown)


rmja:
This is not yet ready for review!

rmja:
Here are some suggestions for further work, but I want to limit this PR for now:
* Move `FlashConfig::BLOCK_SIZE` to a const generic on `BootLoader`. This would allow for `aligned_block_buffer: &mut [u8; BLOCK_SIZE]` instead of `aligned_block_buffer: &mut [u8]`.
* Remove `FirmwareWriter` now when `Partition` exposes the write operations. `FirmwareWriter` currently has a "max-write-size" parameter, but I really think that is a limitation that should be handled by the chunking inside `NorFlash`, and not "up here" in user code.
* Let `Partition` fields be `u32` instead of `usize` in case we ever want to support 16 bit platforms with 32 bit flash.
* Add `FirmwareUpdater::incremental_hash` and `FirmwareUpdater::incremental_hash_blocking` to support computing any hash over the dtu area by providing a closure to the hash update function.

rmja:
I have now improved the validation in the tests such that we validate that the parts being written are erased. This is done by having a `verify_erased_before_write` range inside `MemFlash` which is a range that is asserted to be FF's when being written. This range can be limited with `MemFlash::with_limited_erase_before_write_verification`, as for example writing the state magic is writing all 0's to the magic before the state is erased. This allows for this behavior.

I think the pr is now ready for review. Let me know your thoughts!

rmja:
Closed in favor of https://github.com/embassy-rs/embassy/pull/1329

PR #1318, opened 2023-04-01
rp: Allow zero len reads for buffered uart

timokroeger:
Prevents the read methods from getting stuck forever.

cc @MathiasKoch can you test if this fixes the problem you described in the chat?


MathiasKoch:
Will give it a test first thing Monday morning 👌

Dirbaio:
I think `write` has the same bug?

timokroeger:
Indeed, added the same checks for the write methods.

Dirbaio (approved):
Thank you!

bors r+

PR #1321, opened 2023-04-02
executor: add Pender, rework Cargo features.

Dirbaio:
This introduces a `Pender` struct with enum cases for thread-mode, interrupt-mode and
custom callback executors. This avoids calls through function pointers when using only
the thread or interrupt executors. Faster, and friendlier to `cargo-call-stack`.

`embassy-executor` now has `arch-xxx` Cargo features to select the arch and to enable
the builtin executors (thread and interrupt).


hydra:
> This introduces a `Pender` struct with enum cases for thread-mode, interrupt-mode and custom callback executors. This avoids calls through function pointers when using only the thread or interrupt executors.

Very happy to see this abstraction, for a year or so now in my async-executor, which is loosely based on the embassy one, and uses my own preemptive scheduler I've had this:

```
impl Executor {
    pub fn new(on_idle_fn: OnIdleFn, on_enqueue_fn: OnEnqueueFn) -> Self {...}
}
```

and in threads:
```
        fn on_enqueue_fn(_executor: Pin<&Executor>, _task: &mut TaskHeader) -> () {
            Scheduler::wake(thread_handle());
        }


        fn on_idle_fn(executor: Pin<&Executor>) -> () {
            critical_section::with(|cs| {
                if executor.is_idle() {
                    match executor.next_deadline() {
                        Some(deadline) => Scheduler::suspend_until_cs(cs, deadline),
                        None => Scheduler::suspend_immediately_cs(cs),
                    }
                }
            });
        }
```

Of note, waking a thread is one thing, telling the pre-emptive scheduler when there's no work to do in another thing for which you might also want to consider a similar abstraction, perhaps `Idler`, thoughts?

Dirbaio:
> telling the pre-emptive scheduler when there's no work to do in another thing for which you might also want to consider a similar abstraction, perhaps Idler, thoughts?

Not sure if I understand the purpose? but: embassy's raw executor handles "no work to do" by returning from `poll()`, so user code can do whatever. There's no need to pass an object into the executor to tell it what to do on idle.

hydra:
> Not sure if I understand the purpose?  but: embassy's raw executor handles "no work to do" by returning from `poll()`, so user code can do whatever. There's no need to pass an object into the executor to tell it what to do on idle.

The purpose of `on_idle_fn` it in my system is so that a pre-emptive scheduler can sleep the thread and pend a context switch, the waking of threads needs to be co-ordinated with the waking of tasks (the deadline/expiry) because the 'alarm' code is in the pre-emptive scheduler.

Yes, you're right in embassy the 'poll()' method returns when there's no work to - the loop exits if there are no queued tasks, I guess in the embassy code-base there might not be a need for an `Idler`.  Carry on! 😄

PR #1322, opened 2023-04-03
Remove FirmwareWriter

rmja:
FirmwareWriter currently has a "max-write-size" parameter, but this is a limitation that should be handled by chunking inside the NorFlash driver, and not "up here" in user code. In case that the driver (e.g. qspi driver) is unaware of any max-write limitations, one could simply add an intermediate NorFlash adapter providing the chunk'ing capability.


lulf (approved):
Agreed! Thanks!

bors r+

PR #1323, opened 2023-04-03
Add hash functions to FirmwareUpdater

rmja:
This adds support for computing any hash over the update in the dtu area by providing a closure to the hash update function.


huntc:
I’ve only had a brief look, but the idea of the Digest trait is already to not tie ourselves to a specific implementation. I think I’d prefer that we stick with the trait over a hashing function.

huntc:
> I’ve only had a brief look, but the idea of the Digest trait is already to not tie ourselves to a specific implementation. I think I’d prefer that we stick with the trait over a hashing function. The trait I’d from the Rust crypto project and is already understood.

Oh hang on… it isn’t about the digest, it’s about the hash. Will look further.

embassy-boot/boot/src/firmware_updater.rs
```diff
@@ -182,6 +176,22 @@ impl FirmwareUpdater {
...
+        &mut self,
+        dfu_flash: &mut F,
+        update_len: u32,
+        aligned: &mut [u8],
+        mut update: impl FnMut(&[u8]),
+    ) -> Result<(), FirmwareUpdaterError> {
+        for offset in (0..update_len).step_by(aligned.len()) {
+            self.dfu.read(dfu_flash, offset, aligned).await?;
+            let len = core::cmp::min((update_len - offset) as usize, aligned.len());
+            update(&aligned[..len]);
```
huntc, line 190: I originally wanted to use [Rust Crypto's `Digest` trait](https://github.com/RustCrypto/traits/tree/master/digest) but didn't as the ed25519 libs I chose didn't support it directly. However, those libs will support the trait in the future, and may even now do so. I think that, instead of this `update` closure, we should require something that implements the `Digest` trait instead, and provide our own implementations for the libs we are presently using. That shouldn't be hard to do.

rmja:
@huntc I completely agree, that is much better. I have updated the PR.

embassy-boot/boot/src/firmware_updater.rs
```diff
@@ -118,38 +119,37 @@ impl FirmwareUpdater {
         _state_and_dfu_flash: &mut F,
         _public_key: &[u8],
         _signature: &[u8],
-        _update_len: usize,
+        _update_len: u32,
```
huntc, line 122: Unsure why we are changing this to a `u32`. Can you expand on why?

embassy-boot/boot/src/digest_adapters/ed25519_dalek.rs
huntc, line 1: Why not implement the convenience `Digest` trait instead? I realise they amount to the same thing, mostly curious.

https://docs.rs/digest/latest/digest/trait.Digest.html

embassy-boot/boot/src/firmware_updater.rs
```diff
@@ -118,38 +119,37 @@ impl FirmwareUpdater {
...
 
-            let mut digest = Sha512::new();
-            for offset in (0.._update_len).step_by(_aligned.len()) {
-                self.dfu.read(_state_and_dfu_flash, offset as u32, _aligned).await?;
-                let len = core::cmp::min(_update_len - offset, _aligned.len());
-                digest.update(&_aligned[..len]);
-            }
+            let mut message = [0; 64];
+            self.hash::<_, Sha512>(_state_and_dfu_flash, _update_len, _aligned, &mut message)
+                .await?;
```
huntc, line 143: I think we're doing an additional copy over what we were doing before, although I may be misreading this. With the `Digest` trait, you should be able to return a ref to the result of `finalize`.

embassy-boot/boot/src/firmware_updater.rs
```diff
@@ -327,39 +341,36 @@ impl FirmwareUpdater {
...
-        _update_len: usize,
+        _update_len: u32,
         _aligned: &mut [u8],
     ) -> Result<(), FirmwareUpdaterError> {
-        let _end = self.dfu.from + _update_len;
         let _read_size = _aligned.len();
 
         assert_eq!(_aligned.len(), F::WRITE_SIZE);
-        assert!(_end <= self.dfu.to);
+        assert!(_update_len <= self.dfu.len() as u32);
```
huntc, line 350: Why are we changing the assertion here?

huntc (commented):
Looking great! Just some questions.

embassy-boot/boot/src/firmware_updater.rs
rmja: I have reverted to usize. My plan is to change this to u32 throughout the bootloader, but that should probably be done in a separate pr. The reason for the u32 is to support 16 bit mcu's with flashes larger than 64k. Some of the msp430's have this.

embassy-boot/boot/src/firmware_updater.rs
rmja: This is a left over from https://github.com/embassy-rs/embassy/pull/1312

embassy-boot/boot/src/firmware_updater.rs
rmja: I think that would require that the digest instance is returned then as returning a reference it would not work. I am doing something very similar to https://github.com/RustCrypto/traits/tree/master/digest#generic-code

embassy-boot/boot/src/digest_adapters/ed25519_dalek.rs
rmja: That was to make it easy for myself:-) Implementing the individual parts seemed way easier.

huntc (approved):
Thanks so much for the contribution here!

Approved. Can you please confirm if the changes have been tested on actual hardware? I get nervous about bootloader changes. :-)

rmja:
It works just fine here on my stm32f429👍

PR #1329, opened 2023-04-04
Reduce bootloader memory usage

rmja:
By not requiring that the provided buffer must be able to contain one "erase-size" bytes.

This PR is the successor of #1314 and replaces it entirely.


embassy-boot/boot/src/boot_loader.rs
```diff
@@ -30,23 +30,18 @@ where
...
-    /// size of the flash, but for external QSPI flash modules, this can be lower.
-    const BLOCK_SIZE: usize;
-    /// The erase value of the flash. Typically the default of 0xFF is used, but some flashes use a different value.
-    const ERASE_VALUE: u8 = 0xFF;
-}
-
-/// Trait defining the flash handles used for active and DFU partition
+/// Trait defining the flash handles used for active and DFU partition.
+/// The ACTIVE and DFU erase sizes must be equal. If this is not the case, then consider adding an adapter for the
+/// smallest flash to increase its erase size such that they match. See e.g. [`crate::large_erase::LargeErase`].
```
Dirbaio, line 35: this doesn't exist?

rmja: Fixed. It turned out not to be a requirement.

embassy-boot/boot/src/boot_loader.rs
```diff
@@ -169,87 +175,93 @@ impl BootLoader {
...
-        &mut self,
-        p: &mut P,
-        magic: &mut [u8],
-        page: &mut [u8],
-    ) -> Result<State, BootError> {
+    pub fn prepare_boot<P: FlashConfig>(&mut self, p: &mut P, aligned_buf: &mut [u8]) -> Result<State, BootError> {
         // Ensure we have enough progress pages to store copy progress
-        assert_partitions(self.active, self.dfu, self.state, page.len(), P::STATE::WRITE_SIZE);
-        assert_eq!(magic.len(), P::STATE::WRITE_SIZE);
+        assert_eq!(0, P::page_size() % aligned_buf.len());
```
Dirbaio, line 180: I think we should also assert that `aligned_buf.len()` is a multiple of the two WRITE_SIZEs?

rmja: Good point

Dirbaio (approved):
LGTM! Thank you! :) 

bors r+

Dirbaio:
feel free to merge when green

bors d+

PR #1330, opened 2023-04-05
stm32/pwm: add complementary pwm

xoviat:
This implements complementary PWM with dead time on many supported targets. The specific dead-time programming functions are passed through directly to the user, which is a bit ugly but the best compromise I could reach for now.

### DIFF
--- a/embassy-stm32/src/pwm/mod.rs
+++ b/embassy-stm32/src/pwm/mod.rs
@@ -1,3 +1,4 @@
+pub mod complementary_pwm;
 pub mod simple_pwm;
 
 #[cfg(feature = "unstable-pac")]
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/pwm/mod.rs
```diff
@@ -67,6 +70,14 @@ pub(crate) mod sealed {
         unsafe fn get_max_compare_value(&self) -> u16;
     }
 
+    pub trait ComplementaryCaptureCompare16bitInstance: CaptureCompare16bitInstance {
```
Dirbaio, line 73: Can you add these methods to `CaptureCompare16bitInstance` instead, and not add traits for `Complementary`?

The purpose of these traits is to keep track which timer supports which features. If all `CaptureCompare16bitInstance` timers also impl `ComplementaryCaptureCompare16bitInstance` then 1 trait is enough. The complementary driver can use the `CaptureCompare16bitInstance` trait.

xoviat: All `CaptureCompare16bitInstance` do not impl. `ComplementaryCaptureCompare16bitInstance`, only advanced timers do in most (all?) chips.

Dirbaio: oh, my bad! I misread the `foreach_interrupt!`. I see GP16/GP32 implement `CaptureCompare16bitInstance`, ADV impls both. Sorry!

Dirbaio (approved):
LGTM, thank you! :)

xoviat:
For most (all?) chips, `TIM_ADV` has complementary functionality, `TIM_GP16` does not and `TIM_GP32` does not. 

`TIM_ADV`: 

image

`TIM_GP16` and `TIM_GP32`:

image

PR #1331, opened 2023-04-05
Let bootloader partition be u32

rmja:
This is probably controversial but hear me out:)

The idea about changing from usize to u32 is to enable support for 16 bit mcu's with large flash, e.g. MSP430F5529. Its usize is only 16 bit, but its flash is larger than 64k. Hence, to address its entire flash, it needs the flash address space to be u32.

Missing from the PR is `update_len` in the verification methods. There is currently [a different PR](https://github.com/embassy-rs/embassy/pull/1323) that contains changes in those methods, and I will align depending on the merge order of the two.

The general distinction between u32 and usize is:
* If it is a size or address that only ever lives in flash, then it is u32.
* If the offset or size is ever representable in memory, then usize.

### DIFF
--- a/embassy-boot/boot/src/boot_loader.rs
+++ b/embassy-boot/boot/src/boot_loader.rs
@@ -50,13 +50,13 @@ pub trait FlashConfig {
 }
 
 trait FlashConfigEx {
-    fn page_size() -> usize;
+    fn page_size() -> u32;
 }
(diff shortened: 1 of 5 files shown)


Dirbaio (approved):
LGTM, can you rebase?

bors d+

PR #1332, opened 2023-04-05
executor: Replace unsound critical sections with atomics

GrantM11235:
I couldn't figure out the correct orderings, so I just left them as SeqCst for now.


Dirbaio (approved):
LGTM. Thanks!

This probably hurts perf in chips without CAS (like thumbv6), but I agree removing unsoundness is more important :). We can do another impl that uses *only* CSs later.

bors r+

PR #1336, opened 2023-04-07
stm32/usart: fix unstable traits

xoviat:
The update to embedded-hal-nb led to warnings as discussed in chat. These changes appear to fix the problem.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -589,11 +589,23 @@ mod eh1 {
         type Error = Error;
     }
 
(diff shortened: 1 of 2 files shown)


xoviat:
And the build fails. What's interesting is that:

- My local computer considers `eh1::serial::ErrorKind` to be separate from `ehnb::serial::ErrorKind`, but CI considers them to be the same type.
- My local computer is unable to fetch ehnb from crates.io, and instead must fetch it from the git repository.

Dirbaio:
the nb issue is https://github.com/rust-lang/cargo/issues/11939 , something in crates.io broke.

The other issues I'm not sure. If it works in CI it's maybe you have some local change?

PR #1339, opened 2023-04-07
update trait implementations for Uuart

hdoordt:
Fixes compile errors now that embedded-hal-nb is updated


xoviat:
See #1336.

Dirbaio:
this should be fixed now https://github.com/rust-lang/cargo/issues/11939#issuecomment-1500841310

PR #1342, opened 2023-04-09
feat(rp): add `Wait` impl to `OutputOpenDrain`

TheLostLambda:
A while ago `OutputOpenDrain` was made to implement `InputPin`, something that allowed drivers for various one-wire protocols to be written, but it's been lacking a `Wait` implementation — something that's needed to write async versions of these drivers.

This commit also adds `get_level()` to `OutputOpenDrain`, since `is_high()` and `is_low()` were already implemented, but `get_level()` itself was missing.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -437,6 +437,37 @@ impl<'d, T: Pin> OutputOpenDrain<'d, T> {
     pub fn is_low(&self) -> bool {
         self.pin.is_low()
     }
+
+    /// Returns current pin level
+    #[inline]
+    pub fn get_level(&self) -> Level {
+        self.is_high().into()
+    }
+
+    #[inline]
+    pub async fn wait_for_high(&mut self) {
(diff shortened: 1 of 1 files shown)


TheLostLambda:
Hmm, a quick poke around actually reveals some inconsistency between the different HALs developed by Embassy. The SMT32 one, for example, does already implement `get_level()` but also lacks a `Wait` implementation.

Is there a hardware reason for that? I've tested that my Pi Pico is happy being in `OutputOpenDrain` mode with `Wait` interrupts, but I don't know if all micro-controllers support that?

Sorry if that's a silly question to ask, I'm just getting started with proper embedded development!

Dirbaio:
> Is there a hardware reason for that? I've tested that my Pi Pico is happy being in OutputOpenDrain mode with Wait interrupts, but I don't know if all micro-controllers support that?

not afaik, not even in stm32. it's just not implemented.

Thanks for the PR!

bors r+

Issue #1343, opened 2023-04-09
Use STM32 SPI  as Neopixel WS2812b driver


smeenka:
To use the spi device for neopixel control it would be nice if the SPI device would support:
* a new constructor (new_neopixels) which only uses the Mosi pin, all other pins are None
* Is able to set the bitsize (dff register) to 5 bits.

Then with a clock of 4 Mhz, each spi transfer will clock out one bit for the neopixel, with an exact clock of 800 Khz.
Combined with the DMA transfer, it is an ideal way of controlling a neopixel string.

I will create a fork to implement this requested behavior from the staging branch, including an example file how to use it, for the STM32G070 (nucleo-g070RB board)

Dirbaio:
> a new constructor (new_neopixels) which only uses the Mosi pin, all other pins are None

Sounds good! nrf already has support for that, make sure to keep the API and function names consistent https://github.com/embassy-rs/embassy/blob/master/embassy-nrf/src/spim.rs#L103-L110

> Is able to set the bitsize (dff register) to 5 bits.

Sounds good as well! A field in `config` would work.

smeenka:
Here is my pull request. https://github.com/embassy-rs/embassy/pull/1345

I was not able to set the bitsize in the config. As it is tighly coupled to the datasize. On each write the datasize is set, depending on the data type. I was not sure If I could change this behavior

Dirbaio:
done in https://github.com/embassy-rs/embassy/pull/1345.

but see https://github.com/embassy-rs/embassy/issues/2902

PR #1345, opened 2023-04-10
Added a neopixel constructor to spi, with an example in the stm32g0 d…

smeenka:
For Spi I added the possibility to use the Spi device as a Neopixel driver, with very precise timing, and not dependent on software during the transfer. Even without the --release flag, the timing is perfect.
Note that between the bursts of data the Mosi line should stay on low level. A resistor of 10k can guarantee that, as it seems the the pin is floating between the bursts (on the nucleo-G070RB platform I use).
I created an example for the STM32G0 family.
This example  does contain a very simple Neopixel driver, only for one type. But this Neopixel driver can easy be adapted to different types (for example RGBW types).

### DIFF
--- a/embassy-stm32/src/dma/mod.rs
+++ b/embassy-stm32/src/dma/mod.rs
@@ -21,6 +21,8 @@ pub use gpdma::*;
 #[cfg(dmamux)]
 mod dmamux;
 
+pub mod word;
+
 use core::mem;
 
 use embassy_cortex_m::interrupt::Priority;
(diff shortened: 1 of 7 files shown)


smeenka:
Note that the choice to use i8 to select for 5 bits is up to discussion. But I did not see another way to select the bit width, as it is tightly coupled to the data type in the HAL in Embassy.

embassy-stm32/src/spi/mod.rs
```diff
@@ -178,6 +178,23 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
         )
     }
 
+    pub fn new_neopixel_driver(
```
Dirbaio, line 181: - can you make it consistent with nrf, as I pointed out in the issue? https://github.com/embassy-rs/embassy/issues/1343#issuecomment-1501211790
- If it's txonly it needs no rxdma.

embassy-stm32/src/dma/mod.rs
```diff
@@ -143,6 +143,13 @@ pub trait Word: sealed::Word {
     fn bits() -> WordSize;
 }
 
+impl sealed::Word for i8 {}
```
Dirbaio, line 146: I see, if the bit size is coupled to the type then it makes sense to stick to that, not add it in `Config`. Using `i8` is not nice, it's not intuitive and we're out of options if we want to add 6bit, 7bit, etc in the future. It might be better to define a new type for a 5-bit word:

`#[repr(transparent)] pub struct U5(pub u8);`

smeenka: Created an U8 type and this type selects a word_size of 5 bits. Placed NoDma in the test case as I did not saw a possibility to remove RxDma from the constructor.
A wish: Would be nice if the public constructor does support Some() for all pins and channels. Then less constructors are needed, as the user can make its own selection. Freedom to the masses!

Dirbaio: `Option<impl XxxPin<..>>` doesn't work well with type inference with the pin trait setup. Also, different constructors allow tracking the SPI capabilities in the type so that it only impls `SpiBusWrite` or `SpiBusRead`. (this is not implemented in nRF though)

Dirbaio: > I did not saw a possibility to remove RxDma from the constructor.

Should be doable if you put it in a different `impl` block that fixes RxDma=NoDma instead of being generic.

Dirbaio:
I just realized the existing new_txonly/new_rxonly already have the same useless DMA, so this is not a new issue in this PR. Merging it as is sounds good to me then, we can refactor later.

Renamed `new_neopixel_driver` to `new_txonly_nosck`, since it's not really neopixel-specific, it's just txonly spi with no sck.

Thank you for your contribution!

bors r+

Dirbaio:
In the end I've refactored it a bit, to add support for all Uxx word sizes, and all SPI versions.

bors r+

PR #1348, opened 2023-04-11
stm32/dma: refactor

Dirbaio:
- The shared `Channel` trait is not scaling well.
  - It has methods for oneshot, double-buffered, circular...
  - Double-buffered, circular is not implementable in GPDMA as written.
  - There's already runtime panics saying "this is not supported in BDMA/GPDMA"
  - #1150 adds more stuff for circular mode
  - We'd like to support the fancy new GPDMA linked lists, but that's a completely different API.
- Shared `Channel` trait makes it hard to implement `AnyChannel`, because H7 has a mix of DMA+BDMA
- `AnyChannel` would be useful to remove the `TXDMA, RXDMA` generics on drivers.

So this refactor does:
- Different `Channel` trait for DMA/BDMA/GPDMA.
- Different "driver" for each DMA use case. Currently there's `Transfer` for one-shot transfers, in the future there'll be `CircularTransfer` etc.
- The new driver api is nicer, expose it publically and get rid of `low_level`.

TODO: 
- [x] BDMA
- [x] DMA
- [x] GPDMA
- [x] BDMA+DMA in the H7
- [x] DMA double-buffer mode (needed in DCMI)
- [x] update DCMI
- [x] update SDMMC
- [ ] check how to reduce duplication (hard to do without `include!` hacks, i'd rather not do hacks...)

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -260,7 +260,7 @@ fn main() {
     // ========
     // Generate DMA IRQs.
 
-    let mut dma_irqs: HashMap<&str, Vec<(&str, &str)>> = HashMap::new();
+    let mut dma_irqs: HashMap<&str, Vec<(&str, &str, &str)>> = HashMap::new();
 
     for p in METADATA.peripherals {
         if let Some(r) = &p.registers {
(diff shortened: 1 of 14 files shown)


Dirbaio:
cc @rmja this will unblock #1150

PR #1360, opened 2023-04-12
stm32/rcc: add i2s pll on some f4 micros

xoviat:
Adds the i2s pll on some f4 micros.

### DIFF
--- a/embassy-stm32/src/rcc/f4.rs
+++ b/embassy-stm32/src/rcc/f4.rs
@@ -28,6 +28,8 @@ pub struct Config {
     pub sys_ck: Option<Hertz>,
     pub pclk1: Option<Hertz>,
     pub pclk2: Option<Hertz>,
+
+    #[cfg(not(any(stm32f410, stm32f411, stm32f412, stm32f413, stm32f423, stm32f446)))]
     pub plli2s: Option<Hertz>,
 
     pub pll48: bool,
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/rcc/f4.rs
```diff
@@ -28,11 +28,71 @@ pub struct Config {
...
     pub pclk1: Option<Hertz>,
     pub pclk2: Option<Hertz>,
+    pub plli2s: Option<Hertz>,
 
     pub pll48: bool,
 }
 
-unsafe fn setup_pll(pllsrcclk: u32, use_hse: bool, pllsysclk: Option<u32>, pll48clk: bool) -> PllResults {
+#[cfg(any(
+    feature = "stm32f411",
```
xoviat, line 37: Is there a way to select all of the chip variants without adding them all here? Something like a regex?

Dirbaio: build.rs defines some cfgs with cut down versions of the chip name https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/build.rs#L837

so this works:

```rust
#[cfg(any(
    stm32f411,
    stm32f412,
    stm32f413,
    stm32f423,
    stm32f446,
))]
```

embassy-stm32/src/rcc/f4.rs
```diff
@@ -28,11 +28,65 @@ pub struct Config {
     pub sys_ck: Option<Hertz>,
     pub pclk1: Option<Hertz>,
     pub pclk2: Option<Hertz>,
+    pub plli2s: Option<Hertz>,
```
Dirbaio, line 31: In chips without i2s pll, this field is still present, but does nothing. That might be confusing for the user. Could you change it to put the `cfg` in the field?

Dirbaio (approved):
thanks!

bors r+

PR #1361, opened 2023-04-13
Executor: Replace unnecessary atomics in runqueue

GrantM11235:
Only the head pointer needs to be atomic. The `RunQueueItem` pointers are only loaded and stored, and never concurrently

### DIFF
--- a/embassy-executor/src/raw/run_queue.rs
+++ b/embassy-executor/src/raw/run_queue.rs
@@ -4,15 +4,16 @@ use core::ptr::NonNull;
 use atomic_polyfill::{AtomicPtr, Ordering};
 
 use super::{TaskHeader, TaskRef};
+use crate::raw::util::SyncUnsafeCell;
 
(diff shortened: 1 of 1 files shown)


Dirbaio:
This could use the `core` (not `atomic-polyfill`) AtomicPtr instead. It has no overhead for loads/stores, and there's less unsafe.

The only downside is it's not available for riscv, but that should [get solved soon](https://github.com/rust-lang/rust/issues/99668#issuecomment-1508757127).

GrantM11235:
I don't think that using AtomicPtr is actually any safer in this case. Any bug that could cause a data race here would almost certainly also lead to UB somewhere else. I think that using an atomic here actually makes it harder to reason about, because it falsely implies that there can/will be concurrent accesses.

Dirbaio (approved):
Fair! SGTM1 Also using `SyncUnsafeCell` allows using `Option` which is nice. 

bors r+

PR #1366, opened 2023-04-14
Add ability to invert UART pins

jakewins:
This is useful in some cases where the surrounding circuit for some reason inverts the UART signal, for instance if you're talking to a device via an optocoupler.

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -6,7 +6,7 @@ use crate::dma::{AnyChannel, Channel};
 use crate::gpio::sealed::Pin;
 use crate::gpio::AnyPin;
 use crate::{pac, peripherals, Peripheral};
-use crate::pac::io::vals::Inover;
+use crate::pac::io::vals::{Inover, Outover};
 
 #[cfg(feature = "nightly")]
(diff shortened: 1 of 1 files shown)


embassy-rp/src/uart/mod.rs
```diff
@@ -381,19 +394,31 @@ impl<'d, T: Instance + 'd, M: Mode> Uart<'d, T, M> {
         let r = T::regs();
         unsafe {
             if let Some(pin) = &tx {
-                pin.io().ctrl().write(|w| w.set_funcsel(2));
+                pin.io().ctrl().write(|w| {
+                    w.set_funcsel(2);
+                    w.set_inover(if config.invert_tx { Inover::INVERT } else { Inover::NORMAL });
```
Dirbaio, line 399: TX, RTS should set `outover`, not `inover`.

jakewins:
@Dirbaio corrected! But, testing this out over here, it doesn't seem to work as I expected. I'm a bit short on time to debug, gonna just drop the optocoupler for now so signal isn't inverted; will leave this branch on my fork, and see if I can get this working after summer when things maybe are a bit calmer :)

PR #1369, opened 2023-04-14
Lora AFIT

Dirbaio:
Extracted out of #1367 

Probably we should wait until `rust-lorawan` is merged+released?

### DIFF
--- a/examples/stm32l0/Cargo.toml
+++ b/examples/stm32l0/Cargo.toml
@@ -15,8 +15,8 @@ embassy-time = { version = "0.1.0", path = "../../embassy-time", features = ["de
 embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32l072cz", "time-driver-any", "exti", "unstable-traits", "memory-x"]  }
 embassy-lora = { version = "0.1.0", path = "../../embassy-lora", features = ["sx127x", "time", "defmt"], optional = true}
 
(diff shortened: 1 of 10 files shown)


lulf:
Yup, I can look into that. There is an additional incoming change there as well.

examples/stm32l0/Cargo.toml
```diff
@@ -15,8 +15,8 @@ embassy-time = { version = "0.1.0", path = "../../embassy-time", features = ["de
 embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32l072cz", "time-driver-any", "exti", "unstable-traits", "memory-x"]  }
 embassy-lora = { version = "0.1.0", path = "../../embassy-lora", features = ["sx127x", "time", "defmt"], optional = true}
 
-lorawan-device = { version = "0.8.0", default-features = false, features = ["async"], optional = true }
-lorawan = { version = "0.7.1", default-features = false, features = ["default-crypto"], optional = true }
+lorawan-device = { version = "0.9.0", default-features = false, features = ["async"] }
```
Dirbaio, line 18: the L0 examples are used in `ci_stable.sh` to check an stm32 project builds in stable, that's why these were here.

lulf: The optional got lot lost in the copy-paste, thanks!

PR #1370, opened 2023-04-15
stm32/i2c: fix races when using dma.

xoviat:
This change addresses two races:

1. It removes the `chunks_transferred` state variable that is modified inside the interrupt. Analysis of the code reveals that the only time the waker can be woken is when `chunks_transferred` is incremented. Therefore, waking is enough to signal the `poll_fn` that the `chunks_transferred` has incremented. Moving to `remaining_len` clarifies the code, since there is no need to track how many chunks are remaining.
2. It moves the start of the transfer until after the waker is registered, which could theoretically occur if the clock speed is very low, but probably never would even if this wasn't fixed.

There is another race that I noticed: between writes the waker may not yet be registered. In that case, the code would simply be stuck and the `poll_fn` would never be woken. There is no way to resolve this without broadening the scope of the analysis, and this will likely never occur.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -485,6 +485,7 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
         poll_fn(|cx| {
             state.waker.register(cx.waker());
 
+            let isr = unsafe { T::regs().isr().read() };
             if remaining_len == total_len {
                 // NOTE(unsafe) self.tx_dma does not fiddle with the i2c registers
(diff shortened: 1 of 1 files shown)


xoviat:
Tested on L4.

Dirbaio:
> Analysis of the code reveals that the only time the waker can be woken is when chunks_transferred is incremented. Therefore, waking is enough to signal the poll_fn that the chunks_transferred has incremented

This is not true. Rust async allows "spurious wakes". Your future can be polled even if you didn't wake it. One case where it happens is with `select`. When one of the two futures wakes the waker, both are polled.

Dirbaio (approved):
Thank you!

bors r+

PR #1371, opened 2023-04-16
RTC

xoviat:
This adds RTC for most of the stm32 chips. Nearly all of the work was not done by me, but I took it the last bit by disabling the chips that weren't working. I think it would be easier to enable them in future PRs if requested.


embassy-stm32/Cargo.toml
```diff
@@ -66,6 +66,7 @@ stm32-fmc = "0.2.4"
 seq-macro = "0.3.0"
 cfg-if = "1.0.0"
 embedded-io = { version = "0.4.0", features = ["async"], optional = true }
+chrono = { version = "^0.4", default-features = false, optional = true}
```
Dirbaio, line 69: could you add some CI tests with chrono enabled?

embassy-stm32/src/rtc/mod.rs
```diff
@@ -0,0 +1,238 @@
+//! RTC peripheral abstraction
+use core::marker::PhantomData;
+
+#[cfg_attr(feature = "chrono", path = "datetime_chrono.rs")]
```
Dirbaio, line 4: It's strange that the API changes with or without chrono. We should try to make features additive. Could you change this so that we always use our own `DateTime`, and the `chrono` feature adds `From` impls to convert in both directions?

(Yes, many Cargo features we have are not additive, but it's still nice to make them additive when possible, and in this case it is)

Dirbaio (approved):
Thank you!

bors r+

PR #1372, opened 2023-04-16
rp: add division intrinsics

pennae:
rp2040-hal adds division intrinsics using the hardware divider unit in the SIO, as does the pico-sdk itself. using the hardware is faster than the compiler_rt implementations, and more compact too.

since embassy does not expose the hardware divider in any way (yet?) we could go even further an remove the state-saving code rp2040-hal needs, but that doesn't seem to be worth it.


Dirbaio (approved):
Thank you!

bors r+

PR #1374, opened 2023-04-16
stm32: remove TIMX singleton when used on timer driver

xoviat:
After multiple ways of looking at this, this is the best solution I could think of.


Dirbaio (approved):
Yeah, I can't think of a better approach either. LGTM! Thank you :) 

bors r+

Dirbaio (approved):
bors r+

PR #1376, opened 2023-04-17
rtc: cleanup and consolidate

xoviat:
This removes an extra file that I left in, adds an example, and consolidates the files into one 'v2' file.


MathiasKoch:
Given both this & the former PR with RTC that was merged for stm32, have you tried it on actual hardware?

It seems to me like it is still missing some LS(I/E) Rcc configs?

Dirbaio:
RTC is the perfect thing to write HIL tests for, since it doesn't require external hardware. If any of you guys is up for it I can give you teleprobe access, so you can remotely run the tests on any of the boards in the test farm.

xoviat:
bors try

Dirbaio (approved):
LGTM! Thanks!

bors r+

PR #1379, opened 2023-04-18
enable inline-asm feature for cortex-m in examples

pennae:
inline assembly is supported since rust 1.59, we're way past that. enabling this makes the compiled code more compact, and on rp2040 even decreses memory usage by not needing thunks in sram.


Dirbaio (approved):
Thanks!

bors r+

PR #1381, opened 2023-04-18
fix: `embassy-executor` - use `atomic_polyfill` for `riscv` & `xtensa`

elpiel:
Updating to latest executor in `esp-wifi` (see https://github.com/esp-rs/esp-wifi/pull/163#issuecomment-1513792161) it ended up failing builds for examples since riscv apparently does not support `AtomicBool` from core

### DIFF
--- a/embassy-executor/src/arch/riscv32.rs
+++ b/embassy-executor/src/arch/riscv32.rs
@@ -6,8 +6,8 @@ pub use thread::*;
 #[cfg(feature = "executor-thread")]
 mod thread {
     use core::marker::PhantomData;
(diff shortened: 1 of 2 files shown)


Dirbaio:
I'm sorry, but this is not the right fix for the problem. See previous discussion in #1278, #1240, https://github.com/embassy-rs/embassy/issues/745#issuecomment-1445069621, https://github.com/rust-lang/rust/issues/99668, particularly the fix in https://github.com/embassy-rs/embassy/pull/1278#issuecomment-1468192815

PR #1382, opened 2023-04-19
rp: hook up softfloat intrinsics to bootrom

pennae:
rp-hal has done this very well already, so we'll just copy their entire impl again. only div.rs needed some massaging because our sio access works a little differently, everything else worked as is.

includes a minor bit of refactoring to make it easier to check which bits we've copied from rp2040-hal and which we haven't.

### DIFF
--- a/embassy-rp/src/float/add_sub.rs
+++ b/embassy-rp/src/float/add_sub.rs
@@ -1,0 +1,92 @@
+// Credit: taken from `rp-hal` (also licensed Apache+MIT)
+// https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/float/add_sub.rs
+
+use super::{Float, Int};
+use crate::rom_data;
+
+trait ROMAdd {
+    fn rom_add(self, b: Self) -> Self;
+}
+
+impl ROMAdd for f32 {
+    fn rom_add(self, b: Self) -> Self {
+        rom_data::float_funcs::fadd(self, b)
+    }
+}
+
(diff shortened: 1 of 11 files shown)


Dirbaio:
Thanks!

- Could you add a test [here](https://github.com/embassy-rs/embassy/tree/master/examples/rp/src/bin) for these? Doing some math operations and checking they give the right result. This will ensure the black magic for defining the intrinsics links correctly etc, and also that it works.
- Could you keep the intrinsics-related code in `intrinsics.rs`, instead of `sio.rs`? SIO is a kitchen sink that has tons of unrelated stuff that the designers wanted to make single-cycle-accessible. It's nicer organize code according to what it does, not to where it is on the chip. For example SIO also has GPIO stuff, but code for that is in the `gpio` module.
- Could you remove `typelevel.rs`?
   - One of embassy HALs' design goals is to *not* do such complicated typelevel stuff.
   - For `Sealed`, we use a slightly different pattern, which is doing a local `mod sealed`. Example here: https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/spi.rs#L442-L473
   - The other stuff is AFAICT unused, and the docs don't apply to embassy-rp anyway.

Dirbaio:
good point. Perhaps that's okay? assuming bugs that causing them to be silently not used would be rarer, vs bugs making them not work.

Dirbaio (approved):
Thanks!

bors r+

PR #1383, opened 2023-04-19
embassy-boot: Add nightly flag

sawi97:
This adds "nightly" as a flag to embassy-boot and embassy-boot-nrf which gates features requiring nightly, enabled by default.
Makes it possible to build the bootloader with the stable compiler when setting `default-features=false`.

It should be straight forward to do this for stm32 and rp as well, but I am not been able to test it.


embassy-boot/boot/Cargo.toml
```diff
@@ -45,8 +45,11 @@ default_features = false
 features = ["rand", "std", "u32_backend"]
 
 [features]
+default = ["nightly"]
```
Dirbaio, line 48: In general we try to stay away from default features, because it's common that some library crate that depends on embassy-boot forgets to add `default-features=false` and then the end user has no way to disable them.

Could you remove it, and explicitly enable `nightly` in the examples? Thanks!

Dirbaio:
Thank you for the PR!

- Could you add the boot crates to `ci_stable.sh` so that they don't regress?
- Could you update stm32, rp as well for consistency? it's fine if you don't test it on real hardware, the changes are very straightforward. Just check that it builds is fine.

PR #1384, opened 2023-04-19
rp: optimize rom-func-cache for runtime

pennae:
storing a full function pointer initialized to a resolver trampoline lets us avoid the runtime cost of checking whether we need to do the initialization. this also slightly reduces flash usage due to a slightly more space-efficient initialization procedure.

### DIFF
--- a/embassy-rp/src/rom_data.rs
+++ b/embassy-rp/src/rom_data.rs
@@ -56,50 +56,11 @@ macro_rules! declare_rom_function {
         fn $name:ident( $($argname:ident: $ty:ty),* ) -> $ret:ty
         $lookup:block
     ) => {
-        #[doc = r"Additional access for the `"]
-        #[doc = stringify!($name)]
-        #[doc = r"` ROM function."]
-        pub mod $name {
-            /// Retrieve a function pointer.
-            #[cfg(not(feature = "rom-func-cache"))]
-            pub fn ptr() -> extern "C" fn( $($argname: $ty),* ) -> $ret {
-                let p: *const u32 = $lookup;
-                unsafe {
-                    let func : extern "C" fn( $($argname: $ty),* ) -> $ret = core::mem::transmute(p);
-                    func
-                }
-            }
-
-            /// Retrieve a function pointer.
-            #[cfg(feature = "rom-func-cache")]
-            pub fn ptr() -> extern "C" fn( $($argname: $ty),* ) -> $ret {
-                use core::sync::atomic::{AtomicU16, Ordering};
-
-                // All pointers in the ROM fit in 16 bits, so we don't need a
-                // full width word to store the cached value.
-                static CACHED_PTR: AtomicU16 = AtomicU16::new(0);
-                // This is safe because the lookup will always resolve
-                // to the same value.  So even if an interrupt or another
-                // core starts at the same time, it just repeats some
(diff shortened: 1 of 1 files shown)


embassy-rp/src/rom_data.rs
```diff
@@ -56,56 +56,30 @@ macro_rules! declare_rom_function {
...
-                    val => val as *const u32,
-                };
-                unsafe {
-                    let func : extern "C" fn( $($argname: $ty),* ) -> $ret = core::mem::transmute(p);
-                    func
-                }
-            }
+        declare_rom_function!{
+            __internal ,
+            fn $name( $($argname: $ty),* ) -> $ret
```
Dirbaio, line 61: Nice trick to avoid duplication!

I think you have to propagate `$outer` too?

embassy-rp/src/rom_data.rs
```diff
@@ -114,43 +90,74 @@ macro_rules! declare_rom_function {
...
+
+                #[link_section = concat!(".text.rom_func_cache.", stringify!([<$name _cache>]), ".trampoline")]
+                extern "C" fn lookup() -> *const u32 {
+                    $lookup
+                }
+
+                core::arch::global_asm!{
+                    r#" .section .data.rom_func_cache.{cache}
+                        .type {cache}, %object
+                        .align 2
```
Dirbaio, line 121: should be `.align 4`?

embassy-rp/src/rom_data.rs
```diff
@@ -114,43 +90,74 @@ macro_rules! declare_rom_function {
...
+                        {cache}:
+                            .word rom_func_cache.{cache}.trampoline
```
Dirbaio, line 123: why is it needed to define this in ASM instead of Rust? Doing it in rust would allow letting rustc mangle it, which would allow getting rid of `paste!`.

embassy-rp/src/rom_data.rs
```diff
@@ -114,43 +90,74 @@ macro_rules! declare_rom_function {
...
+                        .section .text.rom_func_cache.{cache}.trampoline
+                        .thumb_func
+                        rom_func_cache.{cache}.trampoline:
+                            push {{r0-r4}}
+                            mov r4, lr
+                            bl {lookup}
+                            mov ip, r0
+                            mov lr, r4
+                            pop {{r0-r4}}
+                            bx ip
```
Dirbaio, line 134: shouldn't this also write the lookup result to `cache`?

embassy-rp/src/rom_data.rs
Dirbaio: ah okay! TIL.

embassy-rp/src/rom_data.rs
Dirbaio: I mean the variable, not the function. I think it should be possible to define the variable as a regular Rust `static mut` and then have the ASM use it?

Dirbaio:
:sweat_smile:

I still think the idea of "initialize the cache pointing to a fn that does lookup then overwrites the cache with the real fn" is really nice. It's still doable in pure Rust, perhaps it still helps perf there.

Up to you, I'm fine with either!

Dirbaio (approved):
thank you!

bors r+

PR #1385, opened 2023-04-20
feat: add embassy-boot-rp to the doc builder

elpiel:


lulf (approved):
Thanks for fixing!

bors r+

PR #1387, opened 2023-04-20
rp: add PWM api

pennae:
add PWM api ~~including interrupts and async support.~~

depends on https://github.com/embassy-rs/rp-pac/pull/1

**TODO**:

- [x] example
- [x] test
- [x] move divmode to typelevel
- [x] deduplicate `new_*` functions

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -60,8 +60,9 @@ chrono = { version = "0.4", default-features = false, optional = true }
 embedded-io = { version = "0.4.0", features = ["async"], optional = true }
 embedded-storage = { version = "0.3" }
 rand_core = "0.6.4"
+fixed = "1.23.1"
 
-rp-pac = { version = "1", features = ["rt"] }
+rp-pac = { version = "2", features = ["rt"] }
 
 embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["unproven"] }
 embedded-hal-1 = { package = "embedded-hal", version = "=1.0.0-alpha.10", optional = true}
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -1,0 +1,338 @@
+//! Pulse Width Modulation (PWM)
+
+use embassy_embedded_hal::SetConfig;
+use embassy_hal_common::{into_ref, Peripheral, PeripheralRef};
+use fixed::traits::ToFixed;
+use fixed::FixedU16;
+use pac::pwm::regs::{ChDiv, Intr};
+use pac::pwm::vals::Divmode;
+
+use crate::gpio::sealed::Pin as _;
+use crate::gpio::{AnyPin, Pin as GpioPin};
+use crate::{pac, peripherals, RegExt};
+
+#[non_exhaustive]
+#[derive(Clone)]
+pub struct Config {
+    pub invert_a: bool,
+    pub invert_b: bool,
+    pub phase_correct: bool,
+    pub enable: bool,
+    pub divider: fixed::FixedU16<fixed::types::extra::U4>,
+    pub compare_a: u16,
+    pub compare_b: u16,
+    pub top: u16,
+}
+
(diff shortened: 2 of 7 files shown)


embassy-rp/Cargo.toml
```diff
@@ -25,6 +25,7 @@ critical-section-impl = ["critical-section/restore-state-u8"]
 unstable-pac = []
 
 time-driver = []
+pwm-driver = []
```
Dirbaio, line 28: No need to feature gate it.

(`time-driver` is optional because there can only be one `embassy-time` driver in a given binary, build fails otherwise. The feature is needed if the user wants to use `embassy-rp` but use something else as a time driver. This is not the case for PWM)

embassy-rp/src/pwm.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    pub fn wrap<'f>(&'f mut self) {
+        unsafe {
+            while pac::PWM.intr().read().0 & self.bit() == 0 {}
+            pac::PWM.intr().write_value(Intr(self.bit() as _));
+        }
+    }
+}
+
+impl<'d, T: Channel> Pwm<'d, T, Async> {
```
Dirbaio, line 189: as discussed in Matrix, let's not do "async wait for wrap" for now.

embassy-rp/src/pwm.rs
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    pub fn into_async(self) -> Pwm<'d, T, Async> {
+        Pwm {
+            _mode: Async::enable(self.inner.number()),
+            inner: self.inner,
+        }
+    }
+
+    pub fn wrap<'f>(&'f mut self) {
```
Dirbaio, line 181: perhaps a more descriptive name like `wait_for_wrap()` would be better.

Also, the lifetime is not needed.

embassy-rp/src/pwm.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    fn bit(&self) -> u32 {
+        1 << self.inner.number() as usize
+    }
+}
+
+impl<'d, T: Channel> Pwm<'d, T, Blocking> {
+    pub fn new_blocking_a(
+        inner: impl Peripheral<P = T> + 'd,
+        a: impl Peripheral<P = impl PwmPin<T> + 'd> + 'd,
```
Dirbaio, line 148: I think the inner `+ 'd` is not needed.

Dirbaio:
isn't `a` the output and `b` the input? so if you take `a,b` then it's `output, input`? This is why I was also suggesting adding `out_` and `in_` to the arg names, helps make it clearer.

Dirbaio:
bors try

TheLostLambda:
Should this implement `pwm::SetDutyCycle` from `embedded-hal`? Is that something that now *will* be back in 1.0 @Dirbaio ?

https://github.com/rust-embedded/embedded-hal/pull/430

Sorry if I'm behind on things! All of this work seems to move at an impressive pace!

Dirbaio:
retrying now that https://github.com/rp-rs/pio-rs/pull/39 is released.

bors r+

Dirbaio:
bors cancel
bors try

embassy-rp/Cargo.toml
```diff
@@ -60,8 +60,10 @@ chrono = { version = "0.4", default-features = false, optional = true }
 embedded-io = { version = "0.4.0", features = ["async"], optional = true }
 embedded-storage = { version = "0.3" }
 rand_core = "0.6.4"
+fixed = "1.23.1"
+typenum = { version = "1.14", default-features = false }
```
Dirbaio, line 64: `typenum` is unused

Dirbaio: `default-features = false` doesn't disable default features, it just doesn't enable them. If another crate depends on `typenum` without `default-features = false` then they get enabled.

Dirbaio (approved):
Thank you!

bors r+

PR #1388, opened 2023-04-22
rcc: deduplicate and combine

xoviat:
This is an incremental attempt to deduplicate combine rcc logic, eventually leading to a generalized `const` rcc solver. The attempt is incremental and public, so that work is not wasted and I can receive feedback on the approach. I expect this to take a long time.


xoviat:
```
// STM32C0: -
// STM32F0: m, n
// STM32F1: m, n
// STM32F2: cm, n, p, q, n, r
// STM32F3: om, n
// STM32F4: cm, n, p, q
// STM32H5: m, n, p, q+
```

Dirbaio:
As discussed on Matrix, I don't think the public API should have `BoundedInteger`s. The TLDR is they complicate the API and in practice they don't add much compile-time safety, since the user will do `42.try_into().unwrap()` which is also a runtime panic.

PR #1392, opened 2023-04-23
embassy-rp : Fix for division intrinsics clashing with rp2040-hal

peterkrull:
Commit [7a682ec](https://github.com/embassy-rs/embassy/commit/7a682ec02af50026d31296ce5cac6383580f5e55#diff-f121955242a67342004444b26214e5d1d591c3182dcd0fedf4329ad472cd1200) may break compilation if also using `rp2040-hal`. It seems that the rp2040-hal does have a feature flag for [disabling intrinsics](https://github.com/rp-rs/rp-hal/blob/2c9921cdc578bfcbd5c6e0eb0085bb91f9a7e229/rp2040-hal/src/sio.rs#L323), but I still cannot seem to compile with that enabled. Adding these flags fixes it for me.

### DIFF
--- a/embassy-rp/src/intrinsics.rs
+++ b/embassy-rp/src/intrinsics.rs
@@ -352,7 +352,6 @@ core::arch::global_asm!(
     ".endm",
 );
 
-#[cfg(all(target_arch = "arm", feature = "intrinsics"))]
 macro_rules! division_function {
     (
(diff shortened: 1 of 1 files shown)


peterkrull:
Yes that seems to work fine for me.

Dirbaio (approved):
bors r+

PR #1395, opened 2023-04-25
rp/pio: bit of a rework

pennae:
the pio module is currently in a Bit of a State. this is far from all that's needed to make it more useful, but it's a start.


Dirbaio (approved):
thanks! 

bors r+

PR #1396, opened 2023-04-25
Add an external LoRa physical layer feature

ceekdee:
The original LoRa drivers have been deprecated and examples associated with them deleted; however, the original LoRa drivers are still available to allow a gentle transition to the external lora-phy crate.


lulf (commented):
Amazing piece of work, this is great @ceekdee !

I'd like to hear your thoughts on the following before merging:

* Bit-rot of the existing drivers. I think it can be left out of the PR and I can take on the job of removing them unless you're already planning on that. 
* Personal opinion: the types for the interface variants tend to get a bit long and noisy when reading the code.

ceekdee:
> * Bit-rot of the existing drivers. I think it can be left out of the PR and I can take on the job of removing them unless you're already planning on that.
_I am planning to remove them once folks have had a reasonable period of time to make the transition._
> * Personal opinion: the types for the interface variants tend to get a bit long and noisy when reading the code.
_I am certainly open to suggestions.  Fortunately, the name is only used once in setup.  From then on, iv is the variable name._

I do not understand the Rust/build-stable check error.  I assume that needs to be resolve before merge?

lulf:
I think you're adding 'unstable-traits' flag by default in some cases, which requires nightly compiler.

Dirbaio:
I'd outright remove the old drivers. We've already removed/broken things without notice often. There's no stability guarantees on git master. Also, removing code from git doesn't cause it to immediately stop working, people can stay pinned to old commits and upgrade later at any time they wish.

Dirbaio:
stable CI builds the nrf52840, stm32l0 examples, without the `nightly` feature https://github.com/embassy-rs/embassy/blob/master/ci_stable.sh#L73-L74

You need to make sure that when the `nightly` feature in the examples crate is not enabled, no nightly feature/crate is used for dependencies. So don't enable nightly features unconditionally, enable them based on the crate's nightly feature.

```toml
[features]
nightly = [..., "embassy-time/nightly", "embassy-time/unstable-traits"]
```

ceekdee:
@Dirbaio :  rp and stm32wl did not have a feature specification, so I added:

[features]
default = ["nightly"]
nightly = ["embassy-time/nightly", "embassy-time/unstable-traits"]

adding default to match those in stm23l0 and nrf52840.  Is it appropriate to add the default to nightly?

ceekdee:
I would like to separate the removal of the original LoRa drivers into a separate PR, since there are several minor details to attend to.  I plan on working on that immediately after this PR.  Ok?

Dirbaio:
>  rp and stm32wl did not have a feature specification

these are not tested in stable CI, so they don't need a `nightly` feature, please don't add it. Only nrf52840 and stm32l0 are.

ceekdee:
Build stable is making me unstable ... :)  Guidance appreciated!

Dirbaio:
you can debug it like this:

```
cd examples/stm32l0 (or nrf52840)
cargo tree --no-default-features --format '{p} {f}'
```

and then check which nightly-only crates or features are enabled.

ceekdee:
For stm32l0 examples:

nightly = ["embassy-stm32/**nightly**", "embassy-time/nightly", "embassy-time/unstable-traits",
           "embassy-lora", "lora-phy", "lorawan-device", "lorawan", "embedded-io/async"]

embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32l072cz", "time-driver-any", "exti", "**unstable-traits**", "memory-x"]  }  <== no "nightly"

Should it be:

nightly = ["embassy-stm32/**unstable-traits**", "embassy-time/nightly", "embassy-time/unstable-traits",
           "embassy-lora", "lora-phy", "lorawan-device", "lorawan", "embedded-io/async"]

Dirbaio:
`unstable-traits` is not nightly-only. It enables the `embedded-hal 1.0` traits, which are not async and therefore works on stable.

`embedded-hal-async` traits are enabled only if both `nightly` and `unstable-traits` are set. So it should be fine to enable `unstable-traits` unconditionally.

lulf:
@ceekdee These fixes should help with the stable build https://github.com/ceekdee/embassy/pull/1

Dirbaio (approved):
bors r+

PR #1399, opened 2023-04-26
fix stm32f7 example runner command for probe-rs-cli

mbrieske:
Just a minor thing: `-v` option does not work with probe-rs-cli


Dirbaio (approved):
nice find, thanks!

Issue #1397, opened 2023-04-26
Add support  for radio coprocessor and IPCC on stm32wb55


OueslatiGhaith:
Hello, 

I noticed that embassy does not support IPCC or the radio coprocessor on stm32wb55. Is there any plans to add them in the future?

This would be a useful feature for many users who want to use Bluetooth or Zigbee on their stm32wb55 devices.

I have noticed that this issue was addressed by  [GH-24 ](https://github.com/embassy-rs/embassy/pull/24) though it was closed. The example there does work with RTIC although it is quite outdated

Dirbaio:
Yes, this would be a great feature to have. No one is working on it or plans to AFAIK though. So PR's welcome :) 

The code in #24 is probably a good starting point, yes!

PR #24, opened 2021-01-16
Add STM32WB55

eupn:
This PR adds initial STM32WB55 support with a primary focus on BLE, but also adds an async I2C with DMA.

Beacon example was tested in hardware:

image

PR #1401, opened 2023-04-26
added IPCC, TL MBOX, and BLE to stm32wb55

OueslatiGhaith:
Hello,

This pull request is related to [#1397](https://github.com/embassy-rs/embassy/issues/1397), it aims to add support for IPCC, TL Mbox, and BLE support for STM32WB55.

This is basically a port of [GH-24](https://github.com/embassy-rs/embassy/pull/24) that has the same issues mentioned there, as I have literally zero experience in writing HALs, though I think this might be helpful to anyone that wants to spend some time to improve upon it.

This pull request adds:
- support of the IPCC peripheral
- support for TL Mbox
- an implementation of [bluetooth_hci](https://crates.io/crates/bluetooth-hci)
- GATT server example

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -29,7 +29,6 @@ time-driver = []
 rom-func-cache = []
 intrinsics = []
 rom-v2-intrinsics = []
-pio = ["dep:pio", "dep:pio-proc"]
 
 # Enable nightly-only features
 nightly = ["embassy-executor/nightly", "embedded-hal-1", "embedded-hal-async", "embassy-embedded-hal/nightly", "dep:embassy-usb-driver", "dep:embedded-io"]
@@ -70,5 +69,5 @@ embedded-hal-async = { version = "=0.2.0-alpha.1", optional = true}
 embedded-hal-nb = { version = "=1.0.0-alpha.2", optional = true}
 
 paste = "1.0"
-pio-proc = {version= "0.2", optional = true}
-pio = {version= "0.2.1", optional = true}
+pio-proc = {version= "0.2" }
+pio = {version= "0.2.1" }
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -11,20 +11,16 @@ mod critical_section_impl;
 mod intrinsics;
 
 pub mod adc;
+pub mod clocks;
 pub mod dma;
+pub mod flash;
 mod float;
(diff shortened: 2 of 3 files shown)


xoviat:
This PR is failing because bbqueue fails to compile. It may be possible to use the embassy-sync channel for similar functionality, instead.

Dirbaio:
- Have you tested this as working on real hardware?
- This is a LOT of code. it will have to be split in several PRs to go through review. Are you up for the task of splitting it, then doing the required fixes based on the review feedback? :)

Some thoughts about higher level structure:

The STM32 radio firmware supports two ways of operation:
  - Standard HCI commands, for managing connections and sending raw link-layer packets over them. The bluetooth host stack on top of it has to implement the higher layer protocols (ATT, GATT, L2CAP...)
  - Vendor-specific HCI commands that let you use a built-in implementation of the higher layer protocols.

The example code uses the latter. We should also support the former (hooking it up to a host stack, like [bleps](https://github.com/bjoernQ/bleps) (written Rust one!), or C ones like btstack, bluedroid), it has some advantages like being portable over all bluetooth chips that implement HCI.

I think it would make sense to make `embassy-stm32` only contain the HCI layer: just send/receive raw HCI packets (as `&[u8]`). Then the `bluetooth-hci` implementation with the vendor-specific commands could go into a separate crate, say, `embassy-stm32-ble`, and the user could hook up `embassy-stm32` to another BLE stack if they want to.

Also, I'm not sure if the `bluetooth-hci` crate is a good choice
  - Last release 3 years ago
  - Uses `nb`
  - It has code size problems (see https://github.com/embassy-rs/embassy/pull/24#issuecomment-791095746)

OueslatiGhaith:
Thanks for taking a look at this :)

@xoviat, I haven't considered that, that might be a good idea, thanks :D 

I have tested this on an stm32wb55rg and it seems to be working, at least on my end, I think dockerizing the dev flow might help with the whole "it works on my machine problem".

@Dirbaio, I am not sure about the best way to split this code, since it's not evident how can we test, for example, the IPCC peripheral without the mailbox or a Bluetooth implementation.
I'm willing to work on this if I can get some guidance since I'm somewhat of a noob when it comes to embedded rust.

About `bluetooth-hci`, I haven't found any other HCI implementation in the wild, do you know anything that might be a good replacement, or do you think it's better to reimplement the whole spec from scratch?

PR #1404, opened 2023-04-27
feat(stm32): Add DMA based, ring-buffer based rx uart, v3

rmja:
This PR replaces #1150. Comparing to that PR, this one has the following changes:

* The implementation now aligns with the new stm32 dma module, thanks @Dirbaio!
* Calls to `read()` now returns on either 1) idle line, or 2) ring buffer is at most half full. This is different from the previous pr, which would return a lot of 1 byte reads. Thank you @chemicstry for making me realize that it was actually not what I wanted. This is accomplished using half-transfer completed and full-transfer completed interrupts. Both seems to be supported on both dma and bdma.

The implementation still have the issue mentioned here: https://github.com/embassy-rs/embassy/pull/1150#discussion_r1094627035

Regarding the todos here: https://github.com/embassy-rs/embassy/pull/1150#issuecomment-1513905925. I have removed the exposure of ndtr from `dma::RingBuffer` to the uart so that the uart now simply calls `ringbuf::reload_position()` to align the position within the ring buffer to that of the actual running dma controller. BDMA and GPDMA is not implemented. I do not have any chips with those dma controllers, so maybe someone else should to this so that it can be tested.

The `saturate_serial` test utility inside `tests/utils` has an `--idles` switch which can be used to saturate the uart from a pc, but with random idles.

Because embassy-stm32 now can have tests, we should probably run them in ci. I do this locally to test the DmaRingBuffer:  `cargo test --no-default-features --features stm32f429ig`.

cc @chemicstry @Dirbaio

### DIFF
--- a/embassy-stm32/src/dma/dma.rs
+++ b/embassy-stm32/src/dma/dma.rs
@@ -4,6 +4,7 @@ use core::pin::Pin;
 use core::sync::atomic::{fence, Ordering};
 use core::task::{Context, Poll, Waker};
 
+use atomic_polyfill::AtomicUsize;
 use embassy_cortex_m::interrupt::Priority;
 use embassy_hal_common::{into_ref, Peripheral, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
@@ -129,13 +130,16 @@ impl From<FifoThreshold> for vals::Fth {
 
 struct State {
     ch_wakers: [AtomicWaker; DMA_CHANNEL_COUNT],
+    complete_count: [AtomicUsize; DMA_CHANNEL_COUNT],
 }
 
 impl State {
     const fn new() -> Self {
+        const ZERO: AtomicUsize = AtomicUsize::new(0);
         const AW: AtomicWaker = AtomicWaker::new();
         Self {
             ch_wakers: [AW; DMA_CHANNEL_COUNT],
+            complete_count: [ZERO; DMA_CHANNEL_COUNT],
         }
     }
 }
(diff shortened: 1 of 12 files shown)


xoviat:
These changes are needed to fix the build: rmja/embassy@99ade50f96dac95b9ca3bfb1c6ec49b2ee02833d. Other than that, I'm not pleased that we'll have a `buffered.rs` and an `rx_ringbuffered.rs` file in the uart mod. Perhaps a better name would be `buffered_v2.rs`? I also think it's unfortunate that write was not implemented, but that's understandable.

Anyway, the `ringbuffer.rs` is `pub(crate)`, which is good; it can be modified internally to add bdma and gpdma later. And the implementation appears to be correct, and it is tested. So I don't have really anything to add there.

Dirbaio:
I'm working on adding BDMA+GPDMA right now, so no need to feature-gate it.

About naming: I've been thinking about it, and I think we need to make the buffered uart the "default", and name the non-buffered one "raw". See the discussions for the `embedded-hal` serial traits, where "buffered vs unbuffered" was discussed and the conclusion was most of the people are better off with the buffered impls. https://github.com/rust-embedded/embedded-hal/pull/349

So it could be something like:
- `embassy_stm32::uart` -> DMA-buffered (the one added in this PR)
- `embassy_stm32::uart::irq_driven` -> IRQ-buffered (the current "buffered uart")
- `embassy_stm32::uart::raw` -> raw, unbuffered

I think we should do it in a follow-up PR though: this one has been delayed enough, it's RX-only for now, and the change should be done in rp, nrf at the same time.

embassy-stm32/src/dma/dma.rs
```diff
@@ -578,18 +613,194 @@ impl<'a, C: Channel, W: Word> DoubleBuffered<'a, C, W> {
...
+    fn ndtr(&self) -> usize {
+        let ch = self.regs().st(self.num());
+        unsafe { ch.ndtr().read() }.ndt() as usize
+    }
+
+    fn get_complete_count(&self) -> usize {
+        let dma = self.regs();
+        let channel_num = self.num();
+        let index = self.index();
+        // Manually process tcif in case transfer was completed and we are in a higher priority task.
```
Dirbaio, line 640: All the existing drivers already have this "problem", Making the interrupt a lower priority than the task that uses it is not supported. I wouldn't take measures to try to mitigateit.

rmja: The reason I made this was that I at some point called `get_complete_count()` from a uart interrupt, and the priority relation between the uart and the dma controller is not strict. I can see now that I do not do this anymore, and so this could probably be removed.

embassy-stm32/src/dma/dma.rs
```diff
@@ -183,13 +188,43 @@ pub(crate) unsafe fn on_irq_inner(dma: pac::dma::Dma, channel_num: usize, index:
...
     }
 }
 
+unsafe fn process_tcif(dma: pac::dma::Dma, channel_num: usize, index: usize) -> bool {
+    let isr_reg = dma.isr(channel_num / 4);
+    let cr_reg = dma.st(channel_num).cr();
+
+    // First, figure out if tcif is set without a cs.
+    if isr_reg.read().tcif(channel_num % 4) && cr_reg.read().tcie() {
+        // Make tcif test again within a cs to avoid race when incrementing complete_count.
```
Dirbaio, line 212: what race is this avoiding?

Dirbaio:
- removed `process_tcif` in DMA.
- Added BDMA support
- Feature-gated ringbuffer on chips with GPDMA for now. Circular mode in GPDMA is not trivial :sweat: 
- cleaned up test so that it passes in HIL.

bors try

Dirbaio:
let's see if HIL tests pass. :pray: 

bors r+

embassy-stm32/src/dma/dma.rs
rmja: The race is not possible if it is guaranteed that the DMA interrupt always have the highest priority.

PR #1405, opened 2023-04-27
add IPCC peripheral for stm32wb

OueslatiGhaith:
Hello again,

This pull request is related to #1397 and #1401, inspired by #24, and was tested on an stm32wb55rg.

This pull request aims to add the IPCC peripheral for stm32wb microcontrollers.
I am debating whether this should be included in the public API, since the IPCC peripheral would be typically managed by the TL Mailbox, not by the app directly.


xoviat:
Usually we don't have the `constrain` api (hal objects are constructed with their constructors), but other than that, I don't see any problems.

OueslatiGhaith:
@xoviat , I should've commented more on my reasoning there, I figured that the IPCC peripheral wouldn't be used directly by users, but rather by the TL mailbox (which I plan to make a pull request for after this one as per #1401). The constrain method will come useful later.

I can remove it for now if you think that would be better.

PR #1406, opened 2023-04-27
rp: DMA behaviour during flash operations

kalkyl:
This PR changes the old behaviour during flash operations where all DMA transfers were paused during the flash operation.
The new approach is to wait for any DMA operating in flash region to finish and let RAM transfers continue.


Dirbaio (approved):
Thank you!

bors r+

PR #1410, opened 2023-04-28
Add `Transactional` trait to rp's i2c impl

### LINKED ISSUE #1409: I2C Raspberry pi pico
Hello! I am using embassy with my raspberry pi pico and it looks pretty good! I want to use it with my Pn532 (NFC Chip) via I2C but i have a problem. 
The library enforces the I2C (in this case `embassy_rp::i2c::I2c::new_blocking()`) to implement `embedded_hal::blocking::i2c::Transactional` together with `Read`, `Write` and `ReadWrite` traits.
I see that all the other required traits are implemented except `Transactional`. So my question is how hard it would be to implement that and whether this is pretty easy and can be easily made into embassy or if i can write a simple wrapper which implements this (however this will work).

Thanks :)

DasLixou:
Fixes #1409

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -617,16 +617,22 @@ mod eh02 {
             self.blocking_write_read(address, bytes, buffer)
         }
     }
-    
+
     impl<'d, T: Instance, M: Mode> embedded_hal_02::blocking::i2c::Transactional for I2c<'d, T, M> {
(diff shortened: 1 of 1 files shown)


DasLixou:
I have no clue whats up with that error message in `build-nightly`. Have i done smth wrong?

Dirbaio (approved):
Thank you!

bors r+

PR #1412, opened 2023-04-29
stm32/uart: abort on error

xoviat:
This PR aborts the DMA transfer in the event of a UART error. Otherwise, the transfer will never complete, and an error will not be returned.


xoviat:
Tested on f7.

Dirbaio (approved):
thanks!

bors r+

PR #1414, opened 2023-04-30
rp: report errors from buffered and dma uart receives

pennae:
neither of these reported errors so far, which is not ideal. add error reporting to both of them that matches the blocking error reporting as closely as is feasible, even allowing partial receives from buffered uarts before errors are reported where they would have been by the blocking code. dma transfers don't do this, if an errors applies to any byte in a transfer the entire transfer is nuked (though we probably could report how many bytes have been transferred).

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -763,7 +763,7 @@ mod sealed {
         fn regs() -> pac::uart::Uart;
 
         #[cfg(feature = "nightly")]
-        fn state() -> &'static buffered::State;
+        fn buffered_state() -> &'static buffered::State;
     }
     pub trait TxPin<T: Instance> {}
     pub trait RxPin<T: Instance> {}
@@ -801,7 +801,7 @@ macro_rules! impl_instance {
             }
 
             #[cfg(feature = "nightly")]
-            fn state() -> &'static buffered::State {
+            fn buffered_state() -> &'static buffered::State {
                 static STATE: buffered::State = buffered::State::new();
                 &STATE
             }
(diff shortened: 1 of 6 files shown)


embassy-rp/src/uart/mod.rs
```diff
@@ -146,6 +153,29 @@ impl<'d, T: Instance, M: Mode> UartTx<'d, T, M> {
...
+
+    pub fn busy(&self) -> bool {
+        unsafe { T::regs().uartfr().read().busy() }
+    }
+
+    /// Assert a break condition after waiting for the transmit buffers to empty,
+    /// or release a previously asserted break condition. This condition must be
+    /// asserted for at least two frame times to be effective, otherwise it may
+    /// be received as a normal character.
+    pub fn set_break(&mut self, brk: bool) {
```
Dirbaio, line 165: why not handle the timing ourselves, so the user does just `uart.send_break()` and doesn't have to worry about the timing?

embassy-rp/src/uart/mod.rs
```diff
@@ -243,19 +308,95 @@ impl<'d, T: Instance> UartRx<'d, T, Blocking> {
     }
 }
 
+static UART_RX_WAKERS: [AtomicWaker; 2] = [AtomicWaker::new(), AtomicWaker::new()];
```
Dirbaio, line 311: could you do this with the "State" struct pattern? This is what the buffered uart does. and most other drivers do. 

Probably rename the buffered uart `state` to `buffered_state.`

embassy-rp/src/uart/mod.rs
Dirbaio: just use `embassy-time`, so no need to mess with sysclk.

Dirbaio (approved):
Thank you! Love how exhaustive the tests are <3

bors r+

PR #1422, opened 2023-05-02
rp: remove leftovers from #1414

pennae:
forgot to remove these when they were no longer necessary or useful. oops.


Dirbaio (approved):
Thanks!

bors r+

PR #1423, opened 2023-05-02
rp: fix gpio InputFuture and inefficiencies

pennae:
InputFuture could not wait for edges without breaking due to a broken From impl, but even if the impl had been correct it would not have worked correctly because raw edge interrupts are sticky and must be cleared from software. also replace critical sections with atomic accesses, and do nvic setup only once.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -9,7 +9,7 @@ use embassy_sync::waitqueue::AtomicWaker;
 
 use crate::pac::common::{Reg, RW};
 use crate::pac::SIO;
-use crate::{interrupt, pac, peripherals, Peripheral};
+use crate::{interrupt, pac, peripherals, Peripheral, RegExt};
 
 const PIN_COUNT: usize = 30;
 const NEW_AW: AtomicWaker = AtomicWaker::new();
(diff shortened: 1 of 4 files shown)


embassy-rp/src/gpio.rs
```diff
@@ -136,16 +136,11 @@ pub enum InterruptTrigger {
...
-            3 => Some(InterruptTrigger::EdgeLow),
-            4 => Some(InterruptTrigger::EdgeHigh),
-            _ => None,
-        }
-    }
+pub(crate) unsafe fn init() {
+    let irq = interrupt::IO_IRQ_BANK0::steal();
+    irq.disable();
+    irq.set_priority(interrupt::Priority::P3);
+    irq.enable();
```
Dirbaio, line 143: This is run on core0 only, so it enables the interrupt in core0's nvic only. When waiting on gpio from core1, the interrupts will be routed to core1 but the interrupt won't fire since it's disabled in nvic.

Dirbaio: yeah that should work. The other way is to route all interrupts to core0. That's what DMA is doing. I don't think it makes much of a difference.

Dirbaio (approved):
thank you!

bors r+

Dirbaio:
bors cancel
bors d+

PR #1425, opened 2023-05-02
rp pio, round 2

pennae:
another round of bugfixes for pio, and some refactoring. in the end we'd like to make pio look like all the other modules and not expose traits that provide all the methods of a type, but put them onto the type itself. traits only make much sense, even if we added an AnyPio and merged the types for the member state machines (at the cost of at least a u8 per member of Pio).


Dirbaio (approved):
LGTM! Thanks for your work, and thanks for splitting them in separate commits so nicely! 

bors r+

Issue #1427, opened 2023-05-02
stm32: rx dma ringbuf is flaky


Dirbaio:
steps to reproduce:
- checkout https://github.com/embassy-rs/embassy/commit/a21a2901acc839b512f40beeee4f77eb5ee8913d
- `cargo run --release --features stm32f103c8 --bin usart_rx_ringbuffered`
- after a few seconds, it stops transmitting data. No overrun error, no corrupted data.

sample output: https://gist.github.com/Dirbaio/49fae6ef02747bcdbd3ad2df5e5f6671

rmja:
I cannot reproduce this on f4.

Here is my local diff relative to current master:

<details>
<summary>Patch</summary>

```
diff --git a/tests/stm32/build.rs b/tests/stm32/build.rs
index 7ae31177..d3f1f39a 100644
--- a/tests/stm32/build.rs
+++ b/tests/stm32/build.rs
@@ -9,7 +9,11 @@ fn main() -> Result<(), Box<dyn Error>> {
     println!("cargo:rustc-link-arg-bins=--nmagic");
 
     // too little RAM to run from RAM.
-    if cfg!(any(feature = "stm32f103c8", feature = "stm32c031c6")) {
+    if cfg!(any(
+        feature = "stm32f429zi",
+        feature = "stm32f103c8",
+        feature = "stm32c031c6"
+    )) {
         println!("cargo:rustc-link-arg-bins=-Tlink.x");
         println!("cargo:rerun-if-changed=link.x");
     } else {
diff --git a/tests/stm32/src/bin/usart_rx_ringbuffered.rs b/tests/stm32/src/bin/usart_rx_ringbuffered.rs
index 9d75dbe5..066c0937 100644
--- a/tests/stm32/src/bin/usart_rx_ringbuffered.rs
+++ b/tests/stm32/src/bin/usart_rx_ringbuffered.rs
@@ -10,7 +10,7 @@ use defmt::{assert_eq, panic};
 use embassy_executor::Spawner;
 use embassy_stm32::interrupt;
 use embassy_stm32::usart::{Config, DataBits, Parity, RingBufferedUartRx, StopBits, Uart, UartTx};
-use embassy_time::{Duration, Timer};
+use embassy_time::{Duration, Timer, Instant};
 use example_common::*;
 use rand_chacha::ChaCha8Rng;
 use rand_core::{RngCore, SeedableRng};
@@ -35,9 +35,9 @@ mod board {
 }
[...]

PR #1428, opened 2023-05-02
stm32/uart: rework ringbuf

xoviat:


xoviat:
bors try

xoviat:
bors try

Dirbaio:
- This PR is quite big. Could you split it into "fix RX" and "add TX support"?
- I think the `DmaRingBuffer` struct should be split into two separate structs for TX and RX. They share almost no code, and having a `const M: bool` generic param where `false` is TX and `true` is RX is very awkward.

About the RX changes:
- Is this supposed to fix #1427? If so, can you expand the tests to check that the repros don't fail anymore?

About the design for TX:
- I would prefer to use `embassy_hal_common::atomic_ring_buffer` instead of writing the ringbuffer implementation from scratch. The TX code is a standard ringbuffer, no hacks needed unlike for RX. Using the atomic ringbuffer wll make it easier to start the DMA transfers from the IRQ in the future (when we make it possible to run custom code in the DMA IRQs)
- Since you do control when DMA is started and stopped, and which part of the buffer is being TX'd, volatile copies are not needed. The CPU should only write to memory that's not being touched by DMA now.

xoviat:
The tx changes have been taken out and the struct refactor has been reverted. The tests were commented out because they rely on the broken behavior of storing ndtr, and I did not want to rewrite them and then have to throw that work away. I may bring them back in a future PR, once this has been merged.

xoviat:
Alright, tests are back. Not as much work as I thought.

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -54,151 +53,130 @@ pub trait DmaCtrl {
...
+            /*
+                first, check if the dma has wrapped around more than once
 
-            self.first = (self.first + len) % self.dma_buf.len();
-            Ok(len)
+                then, get the current position of of the dma write and check
+                if it's inside data we could have copied
+            */
+            let pos = self.pos(dma.get_remaining_transfers());
+            if dma.get_complete_count() > 1 || pos > self.start || pos < end {
```
rmja, line 144: Again, what was wrong with `pos > self.start || dma.get_complete_count() > 1` like it was?
I think that `dma.get_complete_count() > 1` should always be the last condition tested in these or's, as that can change behind our back by the dma controller. The last condition `pos < end`, what does it capture?. That can only happen if `dma.get_complete_count() > 1` which should be tested as the last condition.

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -54,151 +53,130 @@ pub trait DmaCtrl {
...
 
-            Ok(0)
-        } else if self.first < end {
-            // The available, unread portion in the ring buffer DOES NOT wrap
+                then, get the current position of of the dma write and check
+                if it's inside data we could have copied
+            */
+            let (pos, complete_count) =
+                critical_section::with(|_| (self.pos(dma.get_remaining_transfers()), dma.get_complete_count()));
+            if (complete_count > 0 && pos >= end) || complete_count > 1 || (pos >= self.start && pos < end) {
```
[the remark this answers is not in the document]
rmja: Do we need some `compiler_fence(Ordering::SeqCst)` to guarantee the ordering here?

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -54,151 +53,130 @@ pub trait DmaCtrl {
...
+            /*
+                first, check if the dma has wrapped around more than once
 
-                self.first = (self.first + len) % self.dma_buf.len();
-                Ok(len)
+                then, get the current position of of the dma write and check
+                if it's inside data we could have copied
+            */
+            let pos = self.pos(dma.get_remaining_transfers());
+            if dma.reset_complete_count() > 1 || pos > self.start || pos < end {
```
rmja, line 172: Maybe this should be `dma.reset_complete_count() > 1 || pos > self.start || dma.reset_complete_count() > 0` ?

embassy-stm32/src/usart/ringbuffered.rs
```diff
@@ -58,8 +66,8 @@ impl<'d, T: BasicInstance, RxDma: super::RxDma<T>> RingBufferedUartRx<'d, T, RxD
                 w.set_rxneie(false);
                 // enable parity interrupt if not ParityNone
                 w.set_peie(w.pce());
-                // disable idle line interrupt
-                w.set_idleie(false);
+                // enable idle line interrupt
+                w.set_idleie(true);
```
rmja, line 70: This is much better!

embassy-stm32/src/usart/ringbuffered.rs
```diff
@@ -111,96 +118,49 @@ impl<'d, T: BasicInstance, RxDma: super::RxDma<T>> RingBufferedUartRx<'d, T, RxD
...
-            self.wait_for_data_or_idle().await?;
+            match self.ring_buf.read(buf) {
+                Ok((0, _)) => {}
+                Ok((len, _)) => {
+                    return Ok(len);
+                }
+                Err(_) => {
+                    return self.stop(Error::Overrun);
+                }
+            }
```
rmja, line 137: Of cause, putting this in the loop is way better.

embassy-stm32/src/usart/ringbuffered.rs
```diff
@@ -111,96 +118,49 @@ impl<'d, T: BasicInstance, RxDma: super::RxDma<T>> RingBufferedUartRx<'d, T, RxD
...
+                true => Poll::Ready(()),
+            };
 
-            self.ring_buf.reload_position();
-            if !self.ring_buf.is_empty() {
-                // Some data is now available
-                Poll::Ready(())
-            } else {
-                Poll::Pending
-            }
```
rmja, line 203: Not having to rely on this in the dma interrupt is good idea!

embassy-stm32/src/usart/ringbuffered.rs
```diff
@@ -243,11 +186,7 @@ impl<'d, T: BasicInstance, RxDma: super::RxDma<T>> RingBufferedUartRx<'d, T, RxD
 
         match select(dma, uart).await {
             Either::Left(((), _)) => Ok(()),
-            Either::Right((Ok(()), _)) => Ok(()),
-            Either::Right((Err(e), _)) => {
-                self.teardown_uart();
```
rmja, line 248: Is it on purpose that the uart is not stopped on error?

embassy-stm32/src/usart/ringbuffered.rs
```diff
@@ -210,28 +170,11 @@ impl<'d, T: BasicInstance, RxDma: super::RxDma<T>> RingBufferedUartRx<'d, T, RxD
...
-
-            // SAFETY: only clears Rx related flags
-            unsafe {
-                // This read also clears the error and idle interrupt flags on v1.
-                rdr(r).read_volatile();
-                clear_interrupt_flags(r, sr);
-            }
+            // Critical section is needed so that IDLE isn't set after
+            // our read but before we clear it.
+            let sr = critical_section::with(|_| clear_idle_flag(T::regs()));
```
rmja, line 175: Instead of having this here in a critical section, could we put it directly in the usart on_interrupt? and then from there capture the SR register in an AtomicU32 that can be safely swapped and analyzed here?

rmja (commented):
I just took some coffee and went through this:-)
It looks really good and it has improved to the better! I have added some comments throughout. I may not be correct on all of them, but I am learning here, and I really would like to understand the reasoning behind some of the changes - hope that is ok.

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -41,7 +40,7 @@ pub struct OverrunError;
 pub trait DmaCtrl {
     /// Get the NDTR register value, i.e. the space left in the underlying
     /// buffer until the dma writer wraps.
-    fn ndtr(&self) -> usize;
+    fn get_remaining_transfers(&self) -> usize;
```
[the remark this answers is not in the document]
xoviat: This was adjusted because `get_remaining_transfers` is used elsewhere in the dma mod for the same thing.

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -54,151 +53,130 @@ pub trait DmaCtrl {
...
-    /// The buffer end position
-    fn end(&self) -> usize {
-        self.dma_buf.len() - self.ndtr
-    }
-
-    /// Returns whether the buffer is empty
-    pub fn is_empty(&self) -> bool {
-        self.first == self.end()
+    /// The capacity of the ringbuffer
+    pub const fn len(&self) -> usize {
```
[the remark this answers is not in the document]
xoviat: Will do.

embassy-stm32/src/dma/ringbuffer.rs
```diff
@@ -54,151 +53,130 @@ pub trait DmaCtrl {
...
+            buffer, we will do a bit more work than we have to, but algorithms should not be optimized for error
+            conditions.
+
+            After we've done our work, we confirm that we haven't overrun more than a full buffer, and also that
+            the dma has not overrun within the data we could have copied. We check the data we could have copied
+            rather than the data we actually copied because it costs nothing and confirms an error condition
+            earlier.
+        */
+        let end = self.pos(dma.get_remaining_transfers());
+        if self.start == end && dma.get_complete_count() == 0 {
```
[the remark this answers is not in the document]
xoviat: This is a distinction between my implementation and the current implementation. In the event that the DMA has filled the buffer completely and then stopped (unlikely, but possible), my implementation can recover all of the data from the ringbuffer, whereas I think the current implementation would return overflow.

Supposed the ringbuffer wrapped all the way around, and the dma stopped, but the buffer was never read. In that case, complete_count would be `==1` because the buffer had wrapped, but all data would still be valid (end slices after; start slices before). The race is still solve because we always check again after we've read data.

embassy-stm32/src/dma/ringbuffer.rs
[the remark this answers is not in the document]
xoviat: > the dma is currently writing before the block that we have copied out.

We don't know that. Assuming we still had valid data, the dma could still be writing after the end or it could have wrapped. If it has wrapped (`complete_count > 0` guarantees it has wrapped at least once but we cannot make any other claim), then we know for certain than if the position is after the end of our buffer, the dma has overwritten data that could have been valid.

Reconciling this logical condition with that condition, this condition can detect a case where `complete_count` has not yet 
 been updated but the dma has overwritten the data. In other words, it has a greater tolerance for interrupt lag. Whether this would actually happen is debatable but it is still correct.
 
 >  I dont think the cs is needed, but I may be wrong.
 
 It is needed because I could not forsee the impact that a long interrupt could have if it occurred between the assignment of pos and complete count.
 
 > Do we need some compiler_fence(Ordering::SeqCst) to guarantee the ordering here?
 
 You may be right about this. I have no objection to this change.

embassy-stm32/src/dma/ringbuffer.rs
xoviat: You're right the `complete_count` should be the last condition tested.

> The last condition pos < end, what does it capture?. That can only happen if dma.get_complete_count() > 1 which should be tested as the last condition.

No, it can't. You can *never* assume that an interrupt will fire, ever. The most you can say is that, after some fixed amount of time, it is statistically certain to have fired, but it is never guaranteed. And these are practically free checks because these variables are already on the stack.

embassy-stm32/src/dma/ringbuffer.rs
xoviat: The benefit of that condition would be marginal, in my opinion.

embassy-stm32/src/usart/ringbuffered.rs
xoviat: It's stopped in the read function.

embassy-stm32/src/usart/ringbuffered.rs
xoviat: Register access is fast, atomics are slow. However, the critical section may be slower. I personally don't like sharing state between interrupt handlers, though.

embassy-stm32/src/dma/ringbuffer.rs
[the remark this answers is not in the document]
rmja: That is a good point. Thanks for clarifying this!.

embassy-stm32/src/dma/ringbuffer.rs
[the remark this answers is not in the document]
rmja: Year, maybe I just have to give up in understanding this. I was specifically requested to take out the code that handled interrupt tolerance in `get_complete_count()` which was previously [there](https://github.com/rmja/embassy/blob/45843034ec7c53f3f619bf1dcc6c145c428b0601/embassy-stm32/src/dma/dma.rs#L636-L643) in the PR - it tested the tcif flag, and adjusted complete_count accordingly, not relying on the dma to actually have run its isr.

embassy-stm32/src/dma/ringbuffer.rs
rmja: The same goes here. I think it is ok to rely on `get_complete_count()` to actually return the correct value, as it can easily ensure that without relying on that the interrupt has actually fired.

embassy-stm32/src/dma/ringbuffer.rs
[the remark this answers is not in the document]
rmja: Year, ok. Still a bad name though:)

xoviat:
@rmja Would you be okay if I merged these changes? I will cleanup the commits first.

rmja:
Yes, lets do that. Should we add the fix to get_complete_count, such that it always returns the sum of the complete_count atomic + the tcif interrupt flag?

PR #1429, opened 2023-05-03
rp pio, √9

pennae:
another mix of refactoring and soundness issues. most notably pio pins are now checked for being actually accessible to the pio blocks, are constructible from not just the owned peripherals but refs as well, and have their registrations to the pio block reverted once all state machines and the common block has been dropped.

state machines are now also stopped when dropped, and concurrent rx+tx using dma can finally be done in a sound manner. previously it was possible to do, but allowed users to start two concurrent transfers to the same fifo using different dma channels, which obviously would not have the expected results on average.

### DIFF
--- a/embassy-rp/src/pio.rs
+++ b/embassy-rp/src/pio.rs
@@ -97,13 +97,13 @@ pub(crate) unsafe fn init() {
 /// Future that waits for TX-FIFO to become writable
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub struct FifoOutFuture<'a, 'd, PIO: PioInstance, const SM: usize> {
-    sm: &'a mut PioStateMachine<'d, PIO, SM>,
+    sm_tx: &'a mut PioStateMachineTx<'d, PIO, SM>,
     value: u32,
 }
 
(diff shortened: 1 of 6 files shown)


embassy-rp/src/pio.rs
```diff
@@ -306,157 +289,251 @@ impl<PIO: PioInstance> PioPin<PIO> {
...
+    rx: PioStateMachineRx<'d, PIO, SM>,
+    tx: PioStateMachineTx<'d, PIO, SM>,
+}
+
+impl<'d, PIO: PioInstance, const SM: usize> Drop for PioStateMachine<'d, PIO, SM> {
+    fn drop(&mut self) {
+        unsafe {
+            PIO::PIO.ctrl().write_clear(|w| w.set_sm_enable(1 << SM));
+        }
+        on_pio_drop::<PIO>();
```
Dirbaio, line 458: dropping one state machine will deconfigure all PIO pins. I think this could be a footgun, if you only use one SM, and drop all the others early?

Dirbaio (approved):
LGTM! Neat :)  Thank you!


bors r+

PR #1430, opened 2023-05-04
Handle SUBGHZSPI as async

ceekdee:
For STM32WL, simplify configuration for the use of SUBGHZSPI to perform LoRa operations.  Use Rx/Tx DMA on SPI to enable async functionality.

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -195,15 +195,20 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
     }
 
     /// Useful for on chip peripherals like SUBGHZ which are hardwired.
-    /// The bus can optionally be exposed externally with `Spi::new()` still.
     #[allow(dead_code)]
     pub fn new_subghz(
         peri: impl Peripheral<P = T> + 'd,
         txdma: impl Peripheral<P = Tx> + 'd,
         rxdma: impl Peripheral<P = Rx> + 'd,
-        freq: Hertz,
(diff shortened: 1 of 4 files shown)


ceekdee:
LGTM

lulf (approved):
Thanks!

bors r+

embassy-stm32/src/spi/mod.rs
```diff
@@ -195,15 +195,20 @@ impl<'d, T: Instance, Tx, Rx> Spi<'d, T, Tx, Rx> {
...
     /// Useful for on chip peripherals like SUBGHZ which are hardwired.
-    /// The bus can optionally be exposed externally with `Spi::new()` still.
     #[allow(dead_code)]
     pub fn new_subghz(
         peri: impl Peripheral<P = T> + 'd,
         txdma: impl Peripheral<P = Tx> + 'd,
         rxdma: impl Peripheral<P = Rx> + 'd,
-        freq: Hertz,
-        config: Config,
+        pclk3_freq: u32,
```
Dirbaio, line 203: You can obtain pclk3_freq with `RccPeripheral` from inside `embassy-stm32`, there should be no need to make the user pass it in.

ceekdee: I tried that, but then I would have to include a feature to prevent an error on - for example - stm32l0 build.

ceekdee: Correct; I was worried about ramifications in other areas if I tried to push that through on this PR.  I'll work on another PR to address this.

PR #1432, opened 2023-05-04
Support PLLXTPRE switch.

MrOscarLoplate:
See figure 2. Clock tree page 12 DS5319 Rev 18
https://www.st.com/resource/en/datasheet/stm32f103cb.pdf

### DIFF
--- a/embassy-stm32/src/rcc/f1.rs
+++ b/embassy-stm32/src/rcc/f1.rs
@@ -148,10 +148,8 @@ pub(crate) unsafe fn init(config: Config) {
     }
 
     if let Some(pllmul_bits) = pllmul_bits {
-        {
-            let pllctpre_flag: u8 = if config.pllxtpre { 1 } else { 0 };
-            RCC.cfgr().modify(|w| w.set_pllxtpre(Pllxtpre(pllctpre_flag)));
-        }
+        let pllctpre_flag: u8 = if config.pllxtpre { 1 } else { 0 };
+        RCC.cfgr().modify(|w| w.set_pllxtpre(Pllxtpre(pllctpre_flag)));
 
         // enable PLL and wait for it to be ready
         RCC.cfgr().modify(|w| {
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rcc/f1.rs
```diff
@@ -24,10 +24,15 @@ pub struct Config {
     pub pclk1: Option<Hertz>,
     pub pclk2: Option<Hertz>,
     pub adcclk: Option<Hertz>,
+    pub pllxtpre: Option<bool>,
```
Dirbaio, line 27: I think this should be a `bool`, not an `Option<bool>`? `true/false` mean use it / don't use it, `None` doesn't make much sense.

embassy-stm32/src/rcc/f1.rs
```diff
@@ -145,7 +148,11 @@ pub(crate) unsafe fn init(config: Config) {
     }
 
     if let Some(pllmul_bits) = pllmul_bits {
-        RCC.cfgr().modify(|w| w.set_pllxtpre(Pllxtpre(if config.pllxtpre {1u8} else {0u8})));
+        {
```
Dirbaio, line 151: the `{}` aren't needed

embassy-stm32/src/rcc/f1.rs
```diff
@@ -28,7 +28,10 @@ pub struct Config {
 }
 
 pub(crate) unsafe fn init(config: Config) {
-    let pllsrcclk = config.hse.map(|hse| hse.0 / if config.pllxtpre {2} else {1}).unwrap_or(HSI_FREQ.0 / 2);
+    let pllsrcclk = config
+        .hse
+        .map(|hse| if config.pllxtpre { hse.0 / 2 } else { hse.0 })
```
Dirbaio, line 33: could you split this `if` to its own variable too?

Dirbaio (approved):
Thank you!

bors r+

PR #1435, opened 2023-05-06
Added example for multi priority executors rp2040

### LINKED ISSUE #1413: Multiple executor example to run tasks at different priority levels on RP2040.
I have two different things need to manage, time critical motor/sensor control and networking. One fellow suggest using an InterruptExecutor for higher prio tasks with embassy.

I only found a example for nrf52840, [here](https://github.com/embassy-rs/embassy/blob/master/examples/nrf52840/src/bin/multiprio.rs
)

Need example for rp2040

fakusb:
I added an example for multiple priorities of tasks on rp2040 by adjusting [examples/nrf52840/src/bin/multiprio.rs](https://github.com/embassy-rs/embassy/blob/master/examples/nrf52840/src/bin/multiprio.rs) .

This needs https://github.com/embassy-rs/rp-pac/pull/2 , and this commit also adds the 6 new interrupt handlers for software interrupts to embassy-rs.

We might need to change the git path for rp-pac in [embassy-rp/Cargo.toml](https://github.com/embassy-rs/embassy/compare/master...fakusb:rp2040-multiprio-executor?expand=1#diff-47463ea358745927ecdb686f52feab816fde5d402a9628a136c116f34a802ab0)

Closes #1413


Dirbaio:
LGTM, thanks! I've just released `rp-pac v3` with the changes from the PR, could you update Cargo.toml to point to that?

Dirbaio (approved):
Thank you!

bors r+

PR #1436, opened 2023-05-08
rp: Clock configuration

CBJamo:
Draft of a more complete clock config for the 2040.

I also extended and made public the clk_<name>_freq functions. I know at least the ws2812 pio example would like to get the sys clock at runtime rather than just using a constant. I suspect most pio-based peripherals will want access to the clocks.

Open questions:

1. Best way to handle the 3 external clock frequencies. I think the XIN (aka crystal) freq should just be set by the init function then never changed, though if it's an external clock that could change? I'm not sure anyone would ever want to do that but maybe it should be handled just in case? The other two should probably be set by the application.
2. Better estimation of ROSC frequency. Right now it's really just a lookup table of the speed from the single sample I did this testing on, and only uses the frequency range and div, drive strength is ignored.
3. Probably some kind of warning should be generated if the random bit from the rosc won't be useful, not sure how to do that.
4. Should clocks only be allowed to be configured at init, or should they be modifiable at runtime? For example, switching the RTC to a clock in pin when a pps source is available.

Bonus feature to support clock output. I only implemented the bare minimum, and only for gpout0. I'm sure there's a clean way with macros to impl all 4 without just copy/paste, but I haven't learned macros yet.


embassy-rp/src/clocks.rs
```diff
@@ -2,10 +2,139 @@ use pac::clocks::vals::*;
 
 use crate::{pac, reset};
 
-const XOSC_MHZ: u32 = 12;
+// TODO fix terrible use of global here
+static mut XIN_HZ: u32 = 0;
+
+pub struct ClockConfig {
+    rosc_config: Option<RoscConfig>,
```
Dirbaio, line 9: i'd remove `_config` from the field names, if the struct is named `ClockConfig` it's obvious they're all config.

embassy-rp/src/clocks.rs
```diff
@@ -2,10 +2,139 @@ use pac::clocks::vals::*;
...
+pub struct XoscConfig {
+    hz: u32,
+    clock_type: ExternalClock,
+    sys_pll: Option<PllConfig>,
+    usb_pll: Option<PllConfig>,
+}
+
+pub struct RoscConfig {
+    range: pac::rosc::vals::FreqRange,
+    drive_strength_0: u8,
```
Dirbaio, line 97: perhaps it'd make sense to make this an array? 

(disclaimer: I don't know what it is / what it's used for, lol)

embassy-rp/src/clocks.rs
```diff
@@ -189,14 +629,56 @@ unsafe fn configure_pll(p: pac::pll::Pll, refdiv: u32, vco_freq: u32, post_div1:
...
     // Turn on post divider
     p.pwr().modify(|w| w.set_postdivpd(false));
 }
 
+pub struct Gpout0 {
+    _pin: crate::peripherals::PIN_21,
+}
+
+impl Gpout0 {
+    pub fn new(pin: crate::peripherals::PIN_21) -> Self {
```
Dirbaio, line 645: this should have a `trait GpoutPin`, and take a `pin: impl Peripheral<P = impl GpoutPin> + 'd`. 

This way the same struct can be used for all 4 GPOUTs, and can be created with a borrowed pin so you can drop it and get the pin back Same as with all other drivers that use pins.

embassy-rp/src/clocks.rs
```diff
@@ -2,10 +2,139 @@ use pac::clocks::vals::*;
...
 use crate::{pac, reset};
 
-const XOSC_MHZ: u32 = 12;
+// TODO fix terrible use of global here
+static mut XIN_HZ: u32 = 0;
+
+pub struct ClockConfig {
+    rosc_config: Option<RoscConfig>,
+    xosc_config: Option<XoscConfig>,
+    ref_clk_config: (RefClkSrc, u8),
```
Dirbaio, line 11: i'd use structs here instead of tuples. With tuples you can't name the fields, so it ends up being not very self-documenting. There's no way to know the `u8` is the divider just from reading the types.

Dirbaio:
> I think the XIN (aka crystal) freq should just be set by the init function then never changed, though if it's an external clock that could change? I'm not sure anyone would ever want to do that but maybe it should be handled just in case? The other two should probably be set by the application.

I wouldn't worry about it, it seems an incredibly rare corner case.

> Probably some kind of warning should be generated if the random bit from the rosc won't be useful, not sure how to do that.

When we add a driver for that RNG, it could have `fn new()` that panics if ROSC is out of spec, and `unsafe fn new_unchecked()` that doesn't. Printing a warning is probably not enough, it's possible that the user hasn't enabled either defmt nor log, so they won't see it.

> Should clocks only be allowed to be configured at init, or should they be modifiable at runtime? For example, switching the RTC to a clock in pin when a pps source is available.

Changing clocks at runtime is a giant can of worms. I'd say allowing setting them at init only is good enough for now. This is the case for `embassy-stm32` today, there hasn't been much complaining.

I think, if we did, the API could be a `set_config(config: ClockConfig)`, that takes the same struct as `init` and applies it. The problem is there will be instantiated drivers. that have already done clock calculations, that are now wrong. Having `set_config` recalculate all the dividers for all the drivers is not feasible. Using typelevel hacks to enforce all drivers are dropped when changing clocks is not feasible either. Perhaps it's not worth trying? We could just document it in `set_config`, perhaps make it `unsafe`.

embassy-rp/src/clocks.rs
CBJamo: I was going back and forth on this. Partially because I felt the huge number of structs was a bit too verbose. But, I don't have a super strong preference so I switched it.

embassy-rp/src/clocks.rs
CBJamo: I'm going to need some help with this. I think I addressed the drop part, but I'm not sure how to handle the 4 pins with out a huge mess of match statements to route the commands to the right registers and such.

embassy-rp/src/clocks.rs
CBJamo: Doh' of course this should be an array. 

The drive strength is the input current for each stage of the ring oscillator. Basically bigger number = higher freq. Honestly I doubt anyone will ever use it. When the rosc is set to the high freq range it's already pushing the max freq of the core.

embassy-rp/src/clocks.rs
CBJamo: Done

embassy-rp/src/clocks.rs
Dirbaio: oh I see, the `GPOUTx` regs are duplicated in `rp-pac`. 

You can try adding some transforms [here](https://github.com/embassy-rs/rp-pac/blob/main/svd/rp2040.yaml) (`MergeEnums`, `MergeFieldsets`, `MakeRegisterArray` should do the trick), then run `update.sh`.

Once the regs are arrayified, it should be easy to do. For example, add a `fn number() -> u8` to `GpoutPin`, store the num in `self`, so you can use it as an array index.

CBJamo:
Broadly agree about changing clocks at runtime. I'm happy to leave it for now and let whoever first needs it figure it out. I only brought it up because I considered breaking out each clock gen into it's own function for the sake of changing just one at runtime.

There's a good chance I'll end up being the one who needs to do that. In particular the pps into rtc case. Luckily that means only reconfiguring one peripheral, so hopefully it won't be too bad.

embassy-rp/src/clocks.rs
CBJamo: Fixed up, but should certainly be re-reviewed.

CBJamo:
Now requires https://github.com/embassy-rs/rp-pac/pull/3

embassy-rp/src/clocks.rs
```diff
@@ -15,124 +172,411 @@ pub(crate) unsafe fn init() {
...
 
-pub(crate) fn _clk_sys_freq() -> u32 {
-    125_000_000
+pub fn xosc_freq() -> u32 {
+    unsafe { XIN_HZ }
 }
 
-pub(crate) fn clk_peri_freq() -> u32 {
-    125_000_000
+pub fn gpin0_freq() -> u32 {
```
Dirbaio, line 382: if these are TODO i'd just remove them

embassy-rp/src/clocks.rs
```diff
@@ -15,124 +172,411 @@ pub(crate) unsafe fn init() {
...
+
+    let div = unsafe { c.clk_rtc_div().read() };
+    let int = if div.int() == 0 { 65536 } else { div.int() };
+    // TODO handle fractional clock div
+    let _frac = div.frac();
+
+    base / int
+}
+
+pub fn clk_gpout_freq<T: GpoutPin>(gpout: &Gpout<T>) -> u32 {
```
Dirbaio, line 547: this could be a method in `Gpout`, I think it'd be cleaner.

embassy-rp/src/clocks.rs
```diff
@@ -189,14 +633,147 @@ unsafe fn configure_pll(p: pac::pll::Pll, refdiv: u32, vco_freq: u32, post_div1:
...
+pub struct Gpout<'d, T: GpoutPin> {
+    gpout: PeripheralRef<'d, T>,
+}
+
+impl<'d, T: GpoutPin> Gpout<'d, T> {
+    pub fn new(gpout: impl Peripheral<P = T> + 'd) -> Self {
+        into_ref!(gpout);
+
+        unsafe {
+            let p = pac::IO_BANK0.gpio(gpout.pin_number()).ctrl();
```
Dirbaio, line 723: if you make `trait GpoutPin: crate::gpio::Pin` then you can remove `pin_number()`, you can access `ctrl` through the `Pin` trait like `gpout.io().ctrl()`.

example:
https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/spi.rs#L475
https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/spi.rs#L98

Dirbaio: Another reason for adding the supertrait is that `GpoutPin` should be sealed. (A trait becomes sealed if you add a supertrait that's already sealed)

embassy-rp/src/clocks.rs
```diff
@@ -1,11 +1,168 @@
...
 use pac::clocks::vals::*;
 
-use crate::{pac, reset};
+use crate::{pac, reset, Peripheral};
 
-const XOSC_MHZ: u32 = 12;
+// TODO fix terrible use of global here
+static mut XIN_HZ: u32 = 0;
+
+pub struct ClockConfig {
```
Dirbaio, line 9: - fields should be public, otherwise you can't change them
- add `#[non_exhaustive]`, so that we can add fields in the future without it being a breaking change.

embassy-rp/src/clocks.rs
```diff
@@ -189,14 +633,147 @@ unsafe fn configure_pll(p: pac::pll::Pll, refdiv: u32, vco_freq: u32, post_div1:
...
+            }
+        }
+    };
+}
+
+impl_gpinpin!(PIN_20, 20, 0);
+impl_gpinpin!(PIN_22, 22, 1);
+
+pub struct Gpin<'d, T: GpinPin> {
+    gpout: PeripheralRef<'d, T>,
```
Dirbaio, line 666: gpout -> gpin

embassy-rp/src/clocks.rs
```diff
@@ -189,14 +633,147 @@ unsafe fn configure_pll(p: pac::pll::Pll, refdiv: u32, vco_freq: u32, post_div1:
...
+            });
+        }
+    }
+}
+
+impl<'d, T: GpoutPin> Drop for Gpout<'d, T> {
+    fn drop(&mut self) {
+        unsafe {
+            let p = pac::IO_BANK0.gpio(self.gpout.pin_number()).ctrl();
+            p.write(|w| w.set_funcsel(pac::io::vals::Gpio0ctrlFuncsel::NULL.0));
```
Dirbaio, line 772: this could also disable it in `clk_gpout_ctrl`. Not sure if there's any advantage (powr consumption maybe?), but I'd do it just to be safe.

embassy-rp/src/clocks.rs
Dirbaio: Same for Gpin btw

examples/rp/src/bin/gpout.rs
```diff
@@ -0,0 +1,34 @@
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_rp::{clocks, pac};
```
Dirbaio, line 7: `embassy_rp::pac` is only available behind the `unstable-pac` Cargo feature. `embassy_rp::clocks` should reexport all the needed enums, so the user can do everything without `embassy_rp::pac`.

embassy-rp/src/clocks.rs
Dirbaio: :+1: :ok_hand:

embassy-rp/src/clocks.rs
CBJamo: I'd like to leave this in. The alternative is that in the various functions that get clock frequencies we either todo for those two sources, or they just sorta vanish into the `_ => unreachable!()`, which is worse.

Long term this gets messy, as only the application knows what the gpin frequency is. Some kind of shared state is probably necessary.

embassy-rp/src/clocks.rs
CBJamo: Good point, done.

embassy-rp/src/clocks.rs
CBJamo: I thought something like this was probably possible, but I wasn't sure where to look for it. Fixed.

examples/rp/src/bin/gpout.rs
CBJamo: Done, but since naming is the hardest thing in our profession I'd appreciate a sanity check on what I picked.

embassy-rp/src/clocks.rs
Dirbaio: ah okay! I had missed that it's used by us, not just the user.

Dirbaio (approved):
LGTM! Can you rebase to fix the conflict? Feel ree to merge when done!

bors d+

PR #1438, opened 2023-05-08
Update GitHub Actions CI

striezel:
The following updates are performed:
* update [`actions/cache`](https://github.com/actions/cache) to v3
* update [`actions/checkout`](https://github.com/actions/checkout) to v3

Still using the outdated actions will generate several warnings in CI runs, for example in https://github.com/embassy-rs/embassy/actions/runs/4917900672:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/cache@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

The PR will get rid of those warnings.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -29,7 +29,7 @@ jobs:
     concurrency: doc-${{ matrix.crates }}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Thank you!

bors r+

PR #1439, opened 2023-05-09
rp: use rp2040-boot2 to provide the boot2 blob

pennae:
we're currently shipping an old boot2 that runs the flash at half speed. use the more recent version instead, and allow user to choose between the different supported boot2 versions for different flash chips if they need that.


Dirbaio (approved):
Thanks!

bors d+
bors r+

PR #1444, opened 2023-05-10
feat(rp): SPI config builder

elpiel:
Sadly even with default, it's not possible to create a non-exhaustive struct like:
```rust
Config {
frequency: ...,
..Default::default()
}
```

This means that if you want to create a Config you must use:
```
config.frequency = ....;
....
```

Introducing a builder for the `Config` will greatly improve the usage and creation of it.

I couldn't find any `no_std` crates that support derives for builders. Let me know if there is a good solution we can use for `embassy`

### DIFF
--- a/embassy-rp/src/spi.rs
+++ b/embassy-rp/src/spi.rs
@@ -18,12 +18,61 @@ pub enum Error {
     // No errors for now
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+// #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Builder {
+    pub frequency: Option<u32>,
+    pub phase: Option<Phase>,
+    pub polarity: Option<Polarity>,
+}
+
+impl Builder {
+    pub fn frequency(self, frequency: u32) -> Self {
+        self.frequency = frequency;
+        self
+    }
+
+    pub fn phase(self, phase: Phase) -> Self {
+        self.phase = phase;
+        self
+    }
+
+    pub fn polarity(self, polarity: Polarity) -> Self {
+        self.polarity = polarity;
+        self
+    }
+
+    pub fn build(self) -> Config {
+        Config {
+            frequency: self.frequency.unwrap_or(1_000_000),
+            phase: self.frequency.unwrap_or(Phase::CaptureOnFirstTransition),
+            polarity: self.frequency.unwrap_or(Polarity::IdleLow),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


embassy-rp/src/spi.rs
```diff
@@ -18,6 +18,49 @@ pub enum Error {
     // No errors for now
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+// #[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
elpiel, line 22: These structs still don't `derive(Format)`

embassy-rp/src/spi.rs
```diff
@@ -18,6 +18,49 @@ pub enum Error {
...
+pub struct Builder {
+    pub frequency: Option<u32>,
+    pub phase: Option<Phase>,
+    pub polarity: Option<Polarity>,
+}
+
+impl Builder {
+    pub fn frequency(self, frequency: u32) -> Self {
```
MathiasKoch, line 30: I would personally think having these setter functions directly on the config struct is nicer.. such that you do ‘Config::default().frequency(1234)‘ and don't need the ‘build()‘ nor the builder struct

Dirbaio:
Using `non_exhaustive` structs instead of builders is an intentional API design decision.

Both `non_exhaustive` structs and builders do a good job at allowing adding fields without breaking changes. (If breaking changes weren't a concern, we could just use a regular struct).

The syntax difference between them is only personal aesthetics taste. Either

```rust
config.frequency = 1_000_000;
config.polarity = Polarity::IdleLow;
config.phase = Phase::CaptureOnFirstTransition;
```

or 

```rust
let config = Config::builder().frequency(1_000_000).polarity(Polarity::IdleLow).phase(Phase::CaptureOnFirstTransition).build();
```

do the job equally fine. One person might like one more than the other, but neither is inherently better than the other.

`non_exhaustive` structs do have some inherent advantages otherwise:
- require zero boilerplate or macros to implement.
- are arguably more "rust-native", it's a builtin Rust feature instead of a "design pattern" convention.
- You can read fields or take references to them, instead of just setting them.
- More similar to a regular struct, which is what we'd do if it were not for semver concerns.

This is the reason why all config structs in Embassy are `non_exhaustive` structs instead of builders.

elpiel:
Yes, it makes sense indeed.
My only inconvenience with them is that I personally wrap the mut config in a scope to set all values and have an immutable variable afterwards.

I'll close this PR as I agree that it's not needed

PR #1450, opened 2023-05-13
Work around xtensa deadlock

### LINKED ISSUE #1449: executor: multi-core support in xtensa
# Original issue, now resolved:

On multi-core systems, the [critical section implementation](https://github.com/esp-rs/esp-hal/blob/fa3627c1fda334ac89181ae293f762eab82bb307/esp-hal-common/src/lib.rs#L192) provided by esp-hal looks like the following: 
 - mask interrupts
 - acquire a cross-core lock
 - release cross-core lock
 - unmask interrupts

The [executor implementation](https://github.com/embassy-rs/embassy/blob/82f7e104d90a6628d1873017ea5ef6a7afb3b3f7/embassy-executor/src/arch/xtensa.rs#L78) relies on the `waiti` instruction to unmask interrupts, releasing the critical section without releasing the cross-core lock. This can lead to deadlocks where, after waiti ends the critical section, an interrupt handler tries to acquire the multi-core lock that hasn't been released. The following screenshot is an example of this deadlock:

image

Edit: looking at the backtrace and considering how the cross-core lock is reentrant (thanks, dirbaio!), I think triggering this lockup relies on two interrupts firing at specific points in time. One interrupt to wake the CPU, which then handles the interrupt handler and returns to code that is no longer inside a critical section, then the other interrupt triggering at the exact point where the cross-core lock has cleared the owner CPU's id, but before releasing the actual lock.

# Remaining issues

 - Xtensa has no ISA-only mechanism to wake up other cores
 - current implementation uses a single atomic flag to signal that something woke up since the last `waiti`. Each thread-mode executor needs its own flag for this mechanism to be correct.

bugadani:
Fixes (works around) #1449

This PR is intended as a temporary workaround until a proper solution can be found. This is unnecessary and wasteful on single-core MCUs, and also probably wasteful on dual-core ones. As dirbaio explained, `esp-hal` should probably provide a specialized executor implementation for dual-core Xtensa CPUs that maybe only disables interrupts without a complete cross-core lock.

Hacky and probably shouldn't be merged.

### DIFF
--- a/embassy-executor/src/arch/xtensa.rs
+++ b/embassy-executor/src/arch/xtensa.rs
@@ -63,26 +63,29 @@ mod thread {
             loop {
                 unsafe {
                     self.inner.poll();
+
+                    // Manual critical section implementation that only masks interrupts handlers.
+                    // We must not acquire the cross-core on dual-core systems because that would
+                    // prevent the other core from doing useful work while this core is sleeping.
+                    let token: critical_section::RawRestoreState;
+                    core::arch::asm!("rsil {0}, 5", out(reg) token);
+
                     // we do not care about race conditions between the load and store operations, interrupts
                     // will only set this value to true.
                     // if there is work to do, loop back to polling
-                    // TODO can we relax this?
-                    critical_section::with(|_| {
-                        if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
-                            SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
-                        } else {
-                            // waiti sets the PS.INTLEVEL when slipping into sleep
-                            // because critical sections in Xtensa are implemented via increasing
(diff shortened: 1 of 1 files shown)


Dirbaio:
How does this fix it? Even if you re-disable interrupts, the interrupt can fire during the `waiti`, right?

bugadani:
The whole point of this patch is to prevent a race condition in the critical section implementation. Disabling the interrupts after wakeup means it's no longer possible for an interrupt handler to trigger while the core is in the process of exiting the critical section. It is no longer possible to try to lock the cross-core lock at the same point as the exiting critical section tries to release it.

bugadani:
Interestingly, this doesn't work. It reduces the likelihood of the deadlock, and in fact there is no more deadlock I can see, but the CPU locks up at the `rsil` (in take 1) or the `waiti` (in take 2).

PR #1451, opened 2023-05-13
Work around xtensa deadlock, take 2

bugadani:
This PR is another go at trying to do something with #1449. The commit was part of the previous attempt but mistakenly discarded as I still experienced lockups. However, after further testing, it looks like that lockup is caused by something else.

This is a manual, "cpu-local" critical section impl that should be good enough on dual-core CPUs, although the implementation still contains `SIGNAL_WORK_THREAD_MODE` which is absolutely not correct on dual-core. This approach was chosen because:
 - not taking the global lock technically allows the second core to run
 - wrapping the signal read and the sleep in a critical section prevents a race condition that would cause the CPU to sleep longer than ideal if an interrupt hits after reading, but before sleeping.

### DIFF
--- a/embassy-executor/src/arch/xtensa.rs
+++ b/embassy-executor/src/arch/xtensa.rs
@@ -63,21 +63,29 @@ mod thread {
             loop {
                 unsafe {
                     self.inner.poll();
+
+                    // Manual critical section implementation that only masks interrupts handlers.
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

bors r+

PR #1453, opened 2023-05-14
stm32 uart: Fix error flag handling for blocking operations

### LINKED ISSUE #1452: stm32: Serial error flags not cleared on v3/v4 USART
Once there is an overrun error with blocking read, the flag won't be cleared and all subsequent calls return overrun error too. 
Tested on STM32F3 recovery board with the RX pin connected back to the TX pin.

Example to reproduce:
``` rust
#[embassy_executor::main]
    let p = embassy_stm32::init(embassy_stm32::Config::default());

    let uart_config = Config::default();
    let uart_irq = interrupt::take!(USART1);
    let uart = Uart::new(p.USART1, p.PA10, p.PA9, uart_irq, p.DMA1_CH4, p.DMA1_CH5, uart_config);
    let (mut tx, mut rx) = uart.split();

    let tx_buf = b"abcd";
    let mut rx_buf = [0_u8; 4];

    let n = 2;
    info!("Sending {=usize} bytes: {=[u8]:02X}", n, tx_buf[..n]);
    unwrap!(tx.blocking_write(&tx_buf[..n]));
    unwrap!(tx.blocking_flush());
    for _ in 0..3 {
        if let Err(e) = rx.blocking_read(&mut rx_buf[..n]) {
            error!("{:?}", e);
        } else {
            info!("Received {=usize} bytes: {=[u8]:02X}", n, &rx_buf[..n]);
            break;
        }
    }
}
```
Output:
```
0.000335 INFO Sending 2 bytes: [61, 62]
└─ src\bin\usart_fail.rs:24
0.000762 ERROR Overrun
└─ src\bin\usart_fail.rs:29
0.000976 ERROR Overrun
└─ src\bin\usart_fail.rs:29
0.001159 ERROR Overrun
└─ src\bin\usart_fail.rs:29
```

timokroeger:
Clear and report the error flags one by one and pop the data byte only after all error flags were handled.

For v1/v2 we emulate the v3/v4 behaviour by buffering the status register because a read to the data register clears all flags at once which means we might loose all but the first error.

Only tested on stm32f3 discovery board with loopback. Let‘s see what CI says for the other families.
Fixes #1452

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -109,6 +109,8 @@ pub struct UartRx<'d, T: BasicInstance, RxDma = NoDma> {
     _peri: PeripheralRef<'d, T>,
     rx_dma: PeripheralRef<'d, RxDma>,
     detect_previous_overrun: bool,
+    #[cfg(any(usart_v1, usart_v2))]
+    buffered_sr: stm32_metapac::usart::regs::Sr,
 }
 
 impl<'d, T: BasicInstance, TxDma> UartTx<'d, T, TxDma> {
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Thanks!

bors r+

PR #1424, opened 2023-05-02
add TL maibox for stm32wb

OueslatiGhaith:
Hello,

This pull request is related to #1397 and #1401, inspired by #24, build upon the work done in #1405, and was tested on an stm32wb55rg.

This pull request aims to add the transport layer mailbox for stm32wb microcontrollers. For now it's only capable of initializing it and getting the firmware information

PR #1457, opened 2023-05-15
TL Mbox read and write for stm32wb

OueslatiGhaith:
Hello,

This pull request is related to #1397 and #1401, inspired by #24, built upon the work done in #1405 and #1424, and was tested on an stm32wb55rg.

This pull request aims to add read and write functionality to the TL mailbox for stm32wb microcontrollers

### DIFF
--- a/examples/stm32wb/src/bin/tl_mbox_tx_rx.rs
+++ b/examples/stm32wb/src/bin/tl_mbox_tx_rx.rs
@@ -27,6 +27,7 @@ async fn main(_spawner: Spawner) {
         - Once complete, in the Release_Notes.html, find the memory address that corresponds to your device for the
           stm32wb5x_BLE_Stack_full_fw.bin file. It should not be the same memory address.
         - Select that file, the memory address, "verify download", and then "Firmware Upgrade".
+        - Select "Start Wireless Stack".
         - Disconnect from the device.
         - In the examples folder for stm32wb, modify the memory.x file to match your target device.
         - Run this example.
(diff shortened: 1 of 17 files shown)


Dirbaio:
What happened with the tests in #1424 @OueslatiGhaith @xoviat? I see there's a test there but it's not enabled. I'd like to have at least basic "it works" tests, so we don't break it in the future.

If I flashed the radio stack wrong, or it needs another version, please let me know and I'll fix it!

examples/stm32wb/src/bin/tl_mbox_tx_rx.rs
```diff
@@ -0,0 +1,97 @@
...
+            code,
+            payload_len,
+            payload[3..]
+        );
+    }
+
+    mbox.shci_ble_init(&mut ipcc, Default::default());
+
+    info!("resetting BLE");
+    mbox.send_ble_cmd(&mut ipcc, &[0x01, 0x03, 0x0c]);
```
xoviat, line 78: When running on hw, the example stalls here. Is this desired behavior?

xoviat:
TODO (for me): adjust the linker file for the tests for wb55

examples/stm32wb/src/bin/tl_mbox_tx_rx.rs
OueslatiGhaith: no, should reset the ble stack and return a command complete packet, I'll look into it

OueslatiGhaith: it was an incorrect payload, the cmd should be [0x01, 0x03, 0x0c, 0x00, 0x00]

xoviat:
That part works now. There are some issues, but they can be addressed in the next pr. I think HIL will pass.

bors r+

PR #1460, opened 2023-05-15
rp/multicore: ensure stack is 8-byte aligned.

Dirbaio:
bors r+

### DIFF
--- a/embassy-rp/src/multicore.rs
+++ b/embassy-rp/src/multicore.rs
@@ -153,7 +153,12 @@ where
         psm.frce_off().modify(|w| w.set_proc1(false));
     }
 
-    let mem = unsafe { core::slice::from_raw_parts_mut(stack.mem.as_mut_ptr() as *mut usize, stack.mem.len() / 4) };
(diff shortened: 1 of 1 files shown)


embassy-rp/src/multicore.rs
```diff
@@ -153,7 +153,12 @@ where
         psm.frce_off().modify(|w| w.set_proc1(false));
     }
 
-    let mem = unsafe { core::slice::from_raw_parts_mut(stack.mem.as_mut_ptr() as *mut usize, stack.mem.len() / 4) };
+    // The ARM AAPCS ABI requires 8-byte stack alignment.
+    // #[align] on `struct Stack` ensures the bottom is aligned, but the top could still be
```
ithinuel, line 157: because the stack grows from the end of the ram to the beginning, isn't the "top" of the stack close to the start of the memory and the "bottom" at the end of the array instead?

Dirbaio: yeah, it's ambiguous... the code was already using "stack bottom = lower address" terminology so I stuck with that. Not sure if other places use the reverse one.

PR #1463, opened 2023-05-16
Allow for an optional user-defined entry macro when targeting RISC-V

jessebraham:
In [esp-hal](https://github.com/esp-rs/esp-hal) we use our own custom runtime crate, [esp-riscv-rt](https://github.com/esp-rs/esp-riscv-rt). This PR adds the ability to optionally specify an entry macro when using `embassy_executor::main`. The following forms are both accepted:

```rust
#[embassy_executor::main] // uses `riscv_rt::entry` by default
async fn main() {}

#[embassy_executor::main(entry = "esp_riscv_rt::entry")]
async fn main() {}
```

I attempted to get this working without needing to quote the entry macro argument, however I was not able to get this working. Based off some reading I did this may not be possible, however I am rather inexperienced with proc macros. Happy to change this if anybody has any insight.

### DIFF
--- a/embassy-macros/src/lib.rs
+++ b/embassy-macros/src/lib.rs
@@ -79,6 +79,8 @@ pub fn main_cortex_m(args: TokenStream, item: TokenStream) -> TokenStream {
 /// * The function must not use generics.
 /// * Only a single `main` task may be declared.
 ///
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Thank you!

bors r+

PR #1464, opened 2023-05-16
rp: Add system reset fn via watchdog

kalkyl:


Dirbaio (approved):
:eyes: :dog2: 

bors r+

PR #1465, opened 2023-05-17
rp: continue clock rework

pennae:
vastly reduce the code size of initial clock config (over 700 bytes saved!), at the cost of about 48 bytes of ram used to store the frequencies of all clocks in the system. also stop exporting unstable pac items for clock config, fix a few settings that were out of spec, and add missing features (most notably gpin source information).

### DIFF
--- a/examples/rp/src/bin/gpout.rs
+++ b/examples/rp/src/bin/gpout.rs
@@ -17,14 +17,14 @@ async fn main(_spawner: Spawner) {
     gpout3.enable();
 
     loop {
-        gpout3.set_src(clocks::GpoutSrc::CLK_SYS);
+        gpout3.set_src(clocks::GpoutSrc::Sys);
         info!(
             "Pin 25 is now outputing CLK_SYS/1000, should be toggling at {}",
             gpout3.get_freq()
         );
         Timer::after(Duration::from_secs(2)).await;
 
-        gpout3.set_src(clocks::GpoutSrc::CLK_REF);
+        gpout3.set_src(clocks::GpoutSrc::Ref);
         info!(
             "Pin 25 is now outputing CLK_REF/1000, should be toggling at {}",
             gpout3.get_freq()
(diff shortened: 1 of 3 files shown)


examples/rp/src/bin/gpout.rs
```diff
@@ -17,14 +17,14 @@ async fn main(_spawner: Spawner) {
...
-        gpout3.set_src(clocks::GpoutSrc::CLK_SYS);
+        gpout3.set_src(clocks::GpoutSrc::Sys);
         info!(
             "Pin 25 is now outputing CLK_SYS/1000, should be toggling at {}",
             gpout3.get_freq()
         );
         Timer::after(Duration::from_secs(2)).await;
 
-        gpout3.set_src(clocks::GpoutSrc::CLK_REF);
+        gpout3.set_src(clocks::GpoutSrc::Sys);
```
Dirbaio, line 27: Ref?

Dirbaio:
No... I can't think of an easy way to do that... Perhaps it can be tested by wiring a GPOUT to a GPIN, clocking something from the GPIN, and checking that it's alive?

Dirbaio:
> that would require doing clock config twice

ah, true. :( Perhaps we can put a 2nd Pico with a fixed helper firmware that outputs a clock using GPOUT?

CBJamo:
I don't think it's a good idea to configure gpin in the main init. Many applications for an external clock will require you to first enable the clock from an MCU (imagine a TXCO), basically requiring runtime clock init. This gets hairy fast, but I don't think configuring it at init is a safe assumption to make.

Dirbaio:
If we later allow changing the config at runtime (applying a second `Config`) then that problem is solved, right? You can first apply a first `Config` without GPIN, use SPI/I2C/whatever to configure and start the TCXO, then apply a second `Config` that switches to it.

CBJamo:
Hmm, that's a good point Dirbaio. I had assumed that reconfig would be separate functions. But if it's always the one struct/function then it'll be harder to do your reconfig out of order and halt the system.

Dirbaio (approved):
Thanks! 

bors r+

PR #1471, opened 2023-05-21
embassy-net: Make TcpSocket::abort() async

mkj:
This lets callers ensure that the reset packet is sent to the remote host. Otherwise there isn't a way to wait for the smoltcp stack to send the reset.

Requires changes to smoltcp to wake after the reset has been sent, see https://github.com/smoltcp-rs/smoltcp/compare/master...mkj:smoltcp:abort-wake

This commit also adds a "TCP accept" demo of the problem. Without the `.await` for abort() it gets dropped before the RST packet is emitted.

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -95,7 +95,8 @@ impl<'a> TcpWriter<'a> {
 
     /// Flushes the written data to the socket.
     ///
-    /// This waits until all data has been sent, and ACKed by the remote host.
+    /// This waits until all data has been sent, and ACKed by the remote host. For a connection
+    /// closed with [`abort()`](TcpSocket::abort) it will wait for the TCP RST packet to be sent.
     pub async fn flush(&mut self) -> Result<(), Error> {
         self.io.flush().await
     }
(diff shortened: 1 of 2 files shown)


Dirbaio:
I'm not sure if making `abort()` async is the best solution. It adds overhead if you don't need it (ie if you weren't going to drop the socket immediately anyway), and it can have confusing edge cases (for example, what if you cancel aborting the socket? If you cancel a read or a write, it's guaranteed to have no side effects, but not in this case).

An alternative would be to leave `abort()` as is, and make `.flush()` wait for sending the RST. (and mention this behavior in abort's and flush's docs). Having to call flush is surprising, but it is already needed for regular data transfer (if you write some data then drop the socket, the data doesn't get sent, while it does get sent on desktop OSs like Linux), so it's not *that* much more surprising to make it needed for `abort()` as well.

Thoughts?

mkj:
I've pushed changes to use `.flush()` instead. That was actually what I'd tried in the first place so it makes sense. (It didn't work initially since I hadn't made the smoltcp change).

Dirbaio:
lgtm! Just to check: https://github.com/smoltcp-rs/smoltcp/pull/788 is still required, right? to wake when the RST is sent.

Dirbaio (approved):
thank *you*!

bors r+

PR #1472, opened 2023-05-21
ci: build only what's required for prs

xoviat:
This scans the diff and if files in the nrf, stm32, or rp folders are modified, only applicable crates will be built. If any other files are modified, all files will be built, which is the current behavior.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -8,14 +8,49 @@ export DEFMT_LOG=trace
 
 TARGET=$(rustc -vV | sed -n 's|host: ||p')
 
-BUILD_EXTRA=""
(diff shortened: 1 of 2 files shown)


Dirbaio:
It scares me a bit to have to replicate the whole dependency graph into the `.sh` scripts. I'm not sure if it's going to be maintainable in the long run, especially as we add more crates and merge more repos (cyw43, nrf-softdevice, ..) into the main repo.

I'm going to setup a selfhosted CI on a decent machine (i7, 16gb RAM) which will make CI much faster. Hopefully with better caching too. I propose waiting to see how it goes, If it's still too slow then we can look at doing it.

PR #1473, opened 2023-05-22
Protect default implementations for FirmwareUpdater and BootLoader

rmja:
It seems as if the arm compiler can does not care about whether the bootloader symbols are undefined if the default() function is never used. The x64 compiler does care however, so this change ensures that we can instantiate the types from tests.


lulf (approved):
LGTM!

bors r+

Issue #1061, opened 2022-11-20
Executor is not fair if spi dma transfer are used


MauroMombelli:
Minimal example to reproduce the issue follow.

EXPECTED:
START A and START B are both printed (order can be inverted)
A and B are printed, in different order/frequency

OBSERVED:
START A and A are NEVER printed.
START B and B are always printed.

FUN FACT:
Adding a third task, is also never run. "spi.transfer().await;" seems to monopolize the executor
Adding a little delay with "Timer::after(Duration::from_millis(100)).await;"  to the spi's task will let START A and A get printed

CODE:

`#![no_std]
#![no_main]
#![feature(type_alias_impl_trait)]

use defmt::*;
use embassy_executor::Spawner;
use embassy_futures::yield_now;
use embassy_stm32::{spi::{Config, Spi}, peripherals::{SPI1, DMA1_CH3, DMA1_CH2}};

use {defmt_rtt as _, panic_probe as _};

// Declare async tasks
#[embassy_executor::task]
async fn a() {
    info!("START A");
    
    loop {
        info!("A");
        yield_now().await;
    }
}

#[embassy_executor::task]
async fn spi_poll(mut spi: Spi<'static, SPI1, DMA1_CH3, DMA1_CH2>) {
    info!("START B");
    
    loop {
        info!("B");
        let write: [u8; 7] = [0; 7];
        let mut read: [u8; 7] = [0; 7];
        let _ = spi.transfer(&mut read, &write).await;
    }
}

#[embassy_executor::main]
async fn main(spawner: Spawner) {
    
    let p = embassy_stm32::init(Default::default());

    info!("Hello World!");

    spawner.spawn(a()).unwrap();
    
    let spi = Spi::new(
        p.SPI1,
        p.PA5,
        p.PA7,
        p.PA6,
        p.DMA1_CH3,
        p.DMA1_CH2,
        embassy_stm32::time::Hertz(1_000_000),
        Config::default(),
    );

    spawner.spawn(spi_poll(spi)).unwrap();
}
`

yodaldevoid:
I have also run into this issue, but when using `embassy_rp` instead. Adding a `yield_now` to my SPI task also addressed the problem.

nikvoid:
This is happening because without `yield_now().await` you just do not have any await-points in `a` task, so it blocks entire executor in this endless loop
```
    loop {
        info!("A");
        yield_now().await; // await point, here executor can switch tasks
    }
```

[...]

PR #1475, opened 2023-05-22
Add YieldingAsync adapter

rmja:
This PR calls `yield_now()` for long blocking `NorFlash` read and erase operations.
The motivation for this change is to allow for other tasks on the same executor to get something done between these long running operations, for example a task that feeds a watchdog. This will allow the watchdog to have a timer relative to e.g. one sector erase, instead of all sector erase.

### DIFF
--- a/embassy-embedded-hal/Cargo.toml
+++ b/embassy-embedded-hal/Cargo.toml
@@ -14,10 +14,10 @@ target = "x86_64-unknown-linux-gnu"
 [features]
 std = []
 # Enable nightly-only features
-nightly = ["embedded-hal-async", "embedded-storage-async"]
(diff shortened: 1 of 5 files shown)


Dirbaio:
- If we do this, it should be done for all the BlockingAsync impls, it should be consistent. Not just flash.
- This has a non-negligible perf/codesize impact. Perhaps it'd be best to create a 2nd adapter with this behavior?

Dirbaio:
Oh, so that it you can combine with async impls too, not just blocking. That's very intersting. It'd fix (workaround?) issues like #1061

rmja:
@Dirbaio I have updated the pr to now have a `YieldingAsync<T>`.

Dirbaio (approved):
LGTM, thank you!

bors r+

PR #1478, opened 2023-05-23
stm32: Minor fixes in flash regions for F4 dual bank layout

rmja:
This PR has the following fixes:
* Ensure that `FlashRegion` instances can only be created within the embassy-stm32 crate.
* Remove `Drop` trait for `AltFlashLayout`, as it is hard to use, as one cannot take the individual regions out from the struct. Instead of going back to single bank mode on `Drop`, we instead transition to single bank mode when calling `Flash::into_regions()`.
* Add missing `otp_region` to the dual bank layout and implement `NorFlash` for the alternate regions.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -206,6 +206,7 @@ fn main() {
                 erase_size: #erase_size,
                 write_size: #write_size,
                 erase_value: #erase_value,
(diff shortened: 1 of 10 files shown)


Dirbaio (approved):
Thanks!

bors r+

PR #1480, opened 2023-05-24
stm32: Async flash support for F4

rmja:
This PR depends on https://github.com/embassy-rs/embassy/pull/1478.

It adds async write/erase operations to the F4 series based on the work in https://github.com/embassy-rs/embassy/pull/870 but aligned to the new flash regions.

If one considers the entire `Flash` then nothing has changed other than the async operations have been added.


rmja:
This is now aligned with the new bind_interrupt changes.

embassy-stm32/src/flash/common.rs
```diff
@@ -1,44 +1,86 @@
...
+        flash_irq.enable();
+
+        Self {
+            inner: p,
+            #[cfg(all(feature = "nightly", flash_f4))]
+            blocking_only: false,
+        }
+    }
+
+    pub fn new_blocking_only(p: impl Peripheral<P = FLASH> + 'd) -> Self {
```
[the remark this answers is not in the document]
rmja: Same thing. `new_blocking()` indicates that `new()` does not support blocking, which in fact it does.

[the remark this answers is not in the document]
Dirbaio: this is already the case currently in all drivers that have `new` and `new_blocking`, for example in `embassy-rp`:
- `new_blocking` is blocking only
- `new` supports async, and also blocking if mixing is possible (it usually is).

Dirbaio (approved):
LGTM! 
bors d+

embassy-stm32/src/flash/mod.rs
```diff
@@ -1,15 +1,24 @@
 use embedded_storage::nor_flash::{NorFlashError, NorFlashErrorKind};
 
+#[cfg(all(feature = "nightly", flash_f4))]
```
Dirbaio, line 3: async support shouldn't be gated to nightly. 

Only the async trait impls (for `embedded-storage-async` ) need nightly, Inherent methods don't need it.

rmja: Year, the problem is that there are some of the chips that do not have the FLASH interrupt. Is there some cfg() that can determine if that is the case, or do I need to go through all the variants and determine which ones it is?

Dirbaio: gating to `flash_f4` is fine, I was referring to the nightly gate only.

rmja: Of cause!

PR #1482, opened 2023-05-24
Add ConcatFlash utility

rmja:
This PR adds a `ConcatFlash` utility that can be used to concatenate two `NorFlash` flashes. This is especially useful when concatenating multiple flash regions with unequal erase size.


Dirbaio (approved):
Thanks!

bors r+

PR #1485, opened 2023-05-25
Align with updated stm32 metapac

rmja:
This now depends on https://github.com/embassy-rs/stm32-data/pull/192 as it seems that ppre1 is using the enum, but ppre2 does not.

Please have a good look at this. I do not have a G4 chip so I cannot verify this change.


Dirbaio:
thanks for the fixes! stm32-metapac v9 released.

Dirbaio (approved):
thanks!

bors r+
bors d+

PR #1486, opened 2023-05-25
feature(embassy-stm32): add RTC MUX selection to embassy-stm32 L4 family

MathiasKoch:
To select and setup LSE and/or LSI


Dirbaio (approved):
thanks!

bors r+

PR #1490, opened 2023-05-26
sync: do will_wake check in MultiWakerRegistration.

Dirbaio:


petegibson:
Thanks @Dirbaio, this resolved the issue for me. Now works with `SUBS=2` which matches the number of `PubSubChannel.subscriber()` calls.

PR #1492, opened 2023-05-26
Remove ability to, during runtime, set alt layout - it does not work.

rmja:
I wasted yesterday trying to get this to work. It seems that erase operations does not work when in db1m mode. The erase operations, both bank and sector, runs and completes without any error flags being set. It also takes expected amount of time, but the flash is not actually erased.
I found [this](https://community.st.com/s/question/0D50X00009XkWrASAV/is-it-possilbe-to-checkand-may-be-writethe-db1m-option-byte-from-the-firmware-i-am-using-an-stmf4427ig-1mb-with-double-bank-via-db1m-option-byte) on the topic, at it seems that one have to reset the mcu for the change to take effect, so it does not make any sense to have it as runtime configurable.

### DIFF
--- a/embassy-stm32/src/flash/asynch.rs
+++ b/embassy-stm32/src/flash/asynch.rs
@@ -34,7 +34,7 @@ impl<'d> Flash<'d, Async> {
     }
 
     pub fn into_regions(self) -> FlashLayout<'d, Async> {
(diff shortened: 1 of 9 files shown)


Dirbaio (approved):
bors d+

PR #1493, opened 2023-05-26
Disable ppi group in BufferedUarte drop

linusharberg:
Solve the issue where power consumption stays high after dropping BufferedUarte


Dirbaio (approved):
Thank you!

bors r+

PR #1494, opened 2023-05-26
Create flash partition for shared flash access

rmja:


embassy-embedded-hal/src/flash/partition.rs
```diff
@@ -0,0 +1,150 @@
...
+impl<M: RawMutex, T: ErrorType> ErrorType for Partition<'_, M, T> {
+    type Error = Error<T::Error>;
+}
+
+#[cfg(feature = "nightly")]
+impl<M: RawMutex, T: ReadNorFlash> ReadNorFlash for Partition<'_, M, T> {
+    const READ_SIZE: usize = T::READ_SIZE;
+
+    async fn read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Self::Error> {
+        if self.offset % T::READ_SIZE as u32 != 0 || self.size % T::READ_SIZE as u32 != 0 {
```
Dirbaio, line 54: I think it'd be better to check in `new` that offset/size are multiple of ERASE_SIZE. We can asume ERASE_SIZE is a multiple of READ_SIZE/WRITE_SIZE, so this covers read/write as well. This way we don't have to do checks at runtime.

Dirbaio: oh but then it needs `where T: NorFlash` in `new()`, so we can't support readonly flashes. 

Perhaps that's okay? Readonly flash is rare. Also, we could add support if someone needs it by adding `ReadonlyPartition` or something.

Dirbaio (approved):
thanks!

bors r+

PR #1496, opened 2023-05-27
Add BlockingPartition

rmja:
~~This currently does not compile as the blocking mutex does not provide a `&mut T`?~~ This should be ready.


Dirbaio (approved):
thanks!

PR #1497, opened 2023-05-27
Partition generic over blocking/async

rmja:
Is this a good idea? cc @Dirbaio


Dirbaio:
I think it's much simpler to do two separate structs instead. There's no code sharing between blocking and async, they're 2 separate sets of impls that are mutually exclusive. 

Also, this is the same issue as the mutex in shared buses, and there we already have separate structs for blocking/async.

rmja:
Closed in favor of #1496

PR #1498, opened 2023-05-27
Remove bootloader partitions

rmja:
Now when we have `Partition` and soon a `BlockingPartition` (this PR depends on #1496 for tests to run), then we don't need the local Partition type in the bootloader, nor the FlashConfig trait for dispatching the flash.

I have not migrated the FirmwareUpdate yet; I would like to know if this is a path that we want to make before I put in that effort.
I think that when it comes to the FirmwareUpdater, then it would have to be split into two types - a blocking and async, which should be ok.

Let me know what you think!


lulf (commented):
I like the simplifications, looks good!

rmja:
Ok I think this is ready. Are the bootloader examples part of CI? They probably should be.
Let me know what you think.

Dirbaio:
> Are the bootloader examples part of CI?

Yes [they are](https://github.com/embassy-rs/embassy/blob/1d34078fa11839f88dd2e47a9355c6b35755128f/ci.sh#L115-L128)

Dirbaio (approved):

PR #1501, opened 2023-05-29
async can

xoviat:
> You really want a queue for receiving. At 1mbps, the inter-frame time can be as low as 67us and with just 3 receive mailboxes per FIFO you will be missing frames if using only wakers.
> The bxcan library was getting in my way a lot, so it might be worth investigating if using just the registers from PAC would be cleaner.
> Some protocols (i.e. cyphal) need RX timestamps, so it would be a nice to have under time feature gate.

- [x] Switch to new irq handlers
- [x] Add timestamp to RX messages
- [x] Implement message queue
- [x] Implement message priority


schphil:
@xoviat have you thought about adding a split method  `CAN -> TX & RX`  to send and receive simultaneously ? i did that in my repo. the only problem I have with this that I loose the ability the reconfigure since there is nothing like a reunite. what do you think?

embassy-stm32/src/can/bxcan.rs
```diff
@@ -26,30 +110,236 @@ impl<'d, T: Instance> Can<'d, T> {
...
+        self.can.modify_config().set_bit_timing(bit_timing).leave_disabled();
+    }
+
+    /// Queues the message to be sent but exerts backpressure
+    pub async fn write(&mut self, frame: &Frame) -> bxcan::TransmitStatus {
+        poll_fn(|cx| {
+            if let Ok(status) = self.can.transmit(frame) {
+                return Poll::Ready(status);
+            }
+            T::state().tx_waker.register(cx.waker());
```
Dirbaio, line 164: This has a race condition: if the interrupt happens after `can.transmit()` but before `tx_waker.register()`, the task won't get woken and will get stuck.

You have to register the waker first.

xoviat:
@Dirbaio Sorry, that wasn't my code and I just missed that issue before I marked this as ready. Thanks for the catch.

Dirbaio (approved):
thanks! :rocket:

PR #1504, opened 2023-05-29
Add read-only partition support

rmja:


embassy-embedded-hal/src/flash/partition/asynch.rs
```diff
@@ -23,20 +23,33 @@ impl<'a, M: RawMutex, T: NorFlash> Partition<'a, M, T> {
...
+            panic!("Partition size must be a multiple of the read, write and erase size");
         }
         Self { flash, offset, size }
     }
 }
 
-impl<M: RawMutex, T: NorFlash> ErrorType for Partition<'_, M, T> {
+impl<'a, M: RawMutex, T: ReadNorFlash> Partition<'a, M, T> {
+    /// Create a new read-only partition
+    pub const fn new_readonly(flash: &'a Mutex<M, T>, offset: u32, size: u32) -> Self {
```
Dirbaio, line 37: This allows you to call `new_readonly()` with a readwrite flash to obtain a `Partition<..., MyReadWriteFlash>`, which doesn't check WRITE_SIZE/ERASE_SIZE, and then still call the read-write NorFlash methods on it.

PR #1509, opened 2023-05-30
embassy-usb modifications

pcbzzz:
`BosWriter` doesn't write valid attributes, seems like it was just added as an example. BOS needs more data than is currently being passed to the builder, so I don't think it's even really necessary as a builder param until it's more flushed out. Though I've left is as optional for now.

### DIFF
--- a/embassy-usb/src/class/hid.rs
+++ b/embassy-usb/src/class/hid.rs
@@ -1,6 +1,6 @@
 //! USB HID (Human Interface Device) class implementation.
 
-use core::mem::MaybeUninit;
+use core::mem::{self, MaybeUninit};
 use core::ops::Range;
 use core::sync::atomic::{AtomicUsize, Ordering};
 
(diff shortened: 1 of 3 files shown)


embassy-usb/src/class/hid.rs
```diff
@@ -369,8 +369,30 @@ impl<'d, D: Driver<'d>, const N: usize> HidReader<'d, D, N> {
...
+    /// Boot protocol, custom reports are ignored
+    Boot = 0,
+    /// Custom reports
+    Report = 1,
+}
+
 /// Handler for HID-related control requests.
 pub trait RequestHandler {
+    /// Gets the current HID protocol
+    fn get_protocol(&self) -> Option<Protocol> {
```
Dirbaio, line 386: None here means "return 1", it has no special meaning. I think it'd make more sense to make this method not return an Option, and make the default impl return 1.

Dirbaio:
closing due to staleness. 

- The HID protocol seems a useful addition, feel free to repoen a PR with just that.
- I'm not sure about the BOS thing.

PR #1529, opened 2023-06-02
Fix #1528 HS USB on STM32F7

### LINKED ISSUE #1528: STM32F7 High-Speed USB device does not appear (ULPI via USB3320)
I'm working with an STM32F746 Discovery board and am having no USB connectivity at all via the HS USB port. It has a USB3320 for HS via ULPI. The device has 2 separate USB ports for HS and FS. I'm on the main branch running Windows 10.

"No connectivity" as in Windows does not show any new devices or notifications. Also using a tool called USB Device Tree Viewer which clearly shows new/removed devices even if they're connected for a short amount of time, or have failed a negotiation etc.

Based off [a comment](https://github.com/embassy-rs/embassy/pull/1145#issuecomment-1379497618) for the USB OTG PR, I've modified the [usb_serial example](https://github.com/embassy-rs/embassy/blob/main/examples/stm32f7/src/bin/usb_serial.rs) and tried various things to try get it to work with no luck. The [modified code here](https://gist.github.com/gak/0a2db8299abd10ff8155c9db78b45467#file-usb_serial-rs), along with logs and cargo files.

Things I tried/Notes: (I'm new to Embassy and STM32 so these might not be appropriate debugging attempts)
- Various clock speeds, buffer sizes.
- ~~Set all related pins to VeryHigh.~~ (Looks like Embassy does this in `config_ulpi_pins!`)
- ~~Various USB cables, ports, hubs, direct to PC. (All USB 2.1/3.x)~~ (Not the issue)
- ~~There are jumper settings for which USB device powers the board, and changed between them.~~ (Not the issue)
- Reboots, lots of reboots.
- Checked the GPIO pins multiple times in the code. Tried different ones intentionally and awesomely failed to compile. Nice!
- Using FS only, successfully detected USB on the original usb_serial example with no problems.
- The board schematics has a different pin for ULPI_DIR to what CubeMX suggests. Tried both.
- DISCO appears to use USB3320’s CLKOUT as the clock source. Not sure if relevant.
- The official STM32 C examples work, e.g. [HID_Standalone](https://github.com/STMicroelectronics/STM32CubeF7/tree/master/Projects/STM32746G-Discovery/Applications/USB_Device/HID_Standalone)

Unfortunately everything is so small on the PCB I can't get a decent probe in there to see what's actually talking!

Any suggestions would be appreciated. In the mean time I'll dig into the code some more.

gak:
#1528 

`w.set_usb_otg_hsen(ulpien);` seems to not be needed, but leaving it in because I don't know if any clock configuration would disable it.


xoviat (approved):
Given that these changes mirror the h7 logic, I will merge this after a bit of time if no one else objects, or if someone else does not get to it first.

PR #1530, opened 2023-06-03
sync/pipe: write all user data to pipe

cumthugo:
Hello, old _try_write_with_context_ function only write minimum available size to pipe, some data will be lost from user.

It can workaround by user side base on return bytes from write(buf).await, but it's complex.

This patch will wait until all bytes are written to pipe.

### DIFF
--- a/embassy-sync/src/pipe.rs
+++ b/embassy-sync/src/pipe.rs
@@ -294,6 +294,16 @@ where
         WriteFuture { pipe: self, buf }
     }
 
+    /// Write all bytes to the pipe.
+    ///
+    /// This method writes all bytes from `buf` into the pipe
+    pub async fn write_all(&self, mut buf: &[u8]) {
+        while !buf.is_empty() {
+            let n = self.write(buf).await;
+            buf = &buf[n..];
+        }
+    }
+
     /// Attempt to immediately write some bytes to the pipe.
     ///
     /// This method will either write a nonzero amount of bytes to the pipe immediately,


Dirbaio:
It works like this by design. `write` and `read` in `Pipe` work like `std::io::Read/Write`. They only guarantee is "if there's space, at least 1 byte is written".

If you want to write until the full slice is written, use `.write_all` (from `embedded-io`) instead of `.write`:

See previous discussion in #1296

cumthugo:
It still makes me confused. My patch is also compatible with "if there's space, at least 1 byte is written".

`async write` in `Pipe` should make user easy to use, user who want to use `Pipe/write` directly must follow the way like `write_all` in [embedded-io](https://github.com/embassy-rs/embedded-io).

There is no use case to return part of bytes in **async mode**, so I think it's different with sync mode in `std::io::Read/Write`. And this issue is triggered by ringbuffer which is the internal/detail implementation. 
As a user of `Pipe`, there's no need to go deep into the implementation of ringbuffer to find the root cause unless provide `write_all` in `Pipe`.

Dirbaio:
> There is no use case to return part of bytes in async mode

There are some use cases, when the user is going to loop calling `write` anyway. If you make `write` loop writing all chunks, you end up with two nested loops which is an unnecessary cost.

> unless provide write_all in Pipe.

Adding a builtin `Pipe::write_all()` sounds good to me. I think some of the confusion is because `write` is available directy but you need to import `embedded_io` for `write_all`. Having `write_all` builtin could help?

One thing is sure, `write` should behave like embedded-io `write`, not `write_all`, for consistency.

cumthugo:
I agree with you, adding `Pipe::write_all()` help me to use `Pipe` correctly.

I'm newer of Rust and Github, please help to check the updated changes, thanks.

embassy-sync/src/pipe.rs
```diff
@@ -293,7 +293,19 @@ where
     pub fn write<'a>(&'a self, buf: &'a [u8]) -> WriteFuture<'a, M, N> {
         WriteFuture { pipe: self, buf }
     }
-
+    
+    /// Write all bytes to the pipe.
+    ///
+    /// This method writes all bytes from `buf` into the pipe
+    pub async fn write_all<'a>(&'a self, buf: &'a [u8]) -> Result<(), TryWriteError> {
```
Dirbaio, line 300: - explicitly writing the `'a` lifetime is not needed
- this can't fail, so it should return nothing, not a Result.

embassy-sync/src/pipe.rs
```diff
@@ -293,7 +293,19 @@ where
...
+        let mut buf = buf;
```
Dirbaio, line 301: intead of creating another `let` binding, you can make the function arg itself mut: `fn write_all(..., mut buf: &[u8])`

Dirbaio:
Thanks!

Could you run rustfmt, and also squash all commits so the other approach doesn't appear in the git history?

Dirbaio:
Thank you!

PR #1531, opened 2023-06-03
wip: tl_mbox

xoviat:


xoviat:
@OueslatiGhaith let me know if these changes look good to you.

I've refactored the IPCC so that messages can now be read in the main thread. I've also somewhat reversed the static methods so that the individual subsystems can be accessed; it's now possible to do `tl_mbox.ble_subsystem.read().await`. This is so that the methods are organized more easily and that all methods are not put on `tl_mbox`.

I've also added a draft mac and thread subsystem. They don't work currently, but with your changes, they can be made to work fairly easily I think.

Issue #1198, opened 2023-02-07
STM32 G4 RCC implementation is missing PLL


robbym:
Like the title suggests, the [file implementing clock configuration](https://github.com/embassy-rs/embassy/blob/master/embassy-stm32/src/rcc/g4.rs) for G4 devices is missing PLL support (my device in particular is the [STM32G491RE](https://www.st.com/en/microcontrollers-microprocessors/stm32g491re.html)):
```rust
/// Clocks configutation
pub struct Config {
    pub mux: ClockSrc,
    pub ahb_pre: AHBPrescaler,
    pub apb1_pre: APBPrescaler,
    pub apb2_pre: APBPrescaler,
    pub low_power_run: bool,
}
```

Other implementations ([like F4](https://github.com/embassy-rs/embassy/blob/ba18656e94ceffc3fe10a7c47c9b93a1d9d4b2d6/embassy-stm32/src/rcc/f4.rs#L27)) have PLL support and will even calculate the PLL parameters given a source clock, source frequency, and target frequency, which is nice.

I'm fairly new to this project so I don't know if someone is already working on this in particular or on generalizing the notion of PLL configuration, which could _probably_ be done from the STM32 devices I've seen. What should the next steps be?

lulf:
Thanks for raising the issue to track this. I'm not aware of anyone working on this, the next steps would be:

* Extend the ClockSrc to support PLL configuration similar to other device classes. In general, the RCC configuration differs between different device classes, so I think it's fine to keep a g4 PLL configuration separate from the f4 configuration, rather than sprinkling the code with `#[cfg(..)]`. 
* Check the RM to ensure that the PLL configuration map to the correct register values.
* Implement and raise PR (profit!) :)

robbym:
No problem, thanks for the next steps. I'll take a crack at it!

robbym:
@lulf The G4 implementation currently specifies its own enums (e.g. AHBPrescaler, APBPrescaler) whereas other implementations specify it in their rcc_xx.yaml file and have them generated. Should that be done before this? And should that be a separate issue?

lulf:
I didn't realize there was a difference there. I think generally the generated PAC types should not be exposed from the HAL, so I think it's best to define the enums in the HAL.

kevswims:
[...]

PR #1532, opened 2023-06-04
stm32/rcc: Add basic PLL support for G4 series chips

cstlaurent:
As discussed in issue https://github.com/embassy-rs/embassy/issues/1198 , PLL support is missing on G4 series chip.

This PR adds basic PLL support for sysclk. It also adds a PLL example. It was tested working with multiple configurations on a board with an STM32G431CB. 

For some reason i'm not able to go above 132Mhz, but I hardly suspect a board issue since all configurations below that works perfectly... [edit: fixed]


cstlaurent:
I think I see why the processor lock up above 132Mhz, it is probably because of the flash wait state and/or Core power boost mode registers which need to be configured according to the sysclk freq.

Will try to add this configuration

cstlaurent:
Ok I added Vcore boost mode and proper flash wait state configuration. Everything works up to the max frequency of 170Mhz

kevswims:
Thanks for putting this up @cstlaurent I did some testing and this works on one of my boards that has an 8MHz external oscillator as well.

embassy-stm32/src/rcc/g4.rs
```diff
@@ -15,6 +17,7 @@ pub const LSI_FREQ: Hertz = Hertz(32_000);
 pub enum ClockSrc {
     HSE(Hertz),
     HSI16,
+    PLL(PllSrc, PllM, PllN, PllClkDiv),
```
kevswims, line 20: Should this be `PLLR` instead of just `PLL`?

Additionally, is there an idea of how to change this to support configuring the PLL P and Q outputs for driving other peripherals?

embassy-stm32/src/rcc/g4.rs
```diff
@@ -41,6 +44,128 @@ pub enum APBPrescaler {
...
+    fn into(self) -> Pllsrc {
+        match self {
+            PllSrc::HSE(..) => Pllsrc::HSE,
+            PllSrc::HSI16 => Pllsrc::HSI16,
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum PllClkDiv {
```
kevswims, line 64: Might want to specify which divider this is for. The P output can be divided between 2 and 31 so this only applies to R and Q dividers.

embassy-stm32/src/rcc/g4.rs
cstlaurent: Hi, thanks for the review,

Yeah, I was unsure with the naming convention, since I partially copied the implementation from the l4 one as well as input from the ref-manual and the CubeIDE UI.

I was thinking, maybe we should strictly follow the CubeIDE/ref-manual naming as shown in the screenshot below:

CleanShot 2023-06-07 at 19 57 13@2x

This would mean changing `PLL` to `PLLCLK` (as refered on the System clock mux) and `PLLClkDiv` would be renamed to `PllR`.


As for the P and Q, I cheaped out a bit on the implementation since I didn't need them 😄 . But with the rename I think it would be quite easy to add in. (which I can also do in this PR)

Let me know what you think, I will do the required changes.

embassy-stm32/src/rcc/g4.rs
cstlaurent: As explained above, renaming to `PllR` would be simpler to understand

embassy-stm32/src/rcc/g4.rs
kevswims: Those changes sound good to me. My board can use the PLLQ output for CAN or USB related clocking so having those usable would be great.

Dirbaio (approved):
Very nice! Thanks for the contribution :)

PR #1534, opened 2023-06-05
Replace Into<bool> for Level with From<Level> for bool

ExplodingWaffle:
See #1171 but for STM and RP also


Dirbaio (approved):
Thank you!

PR #1535, opened 2023-06-05
Add IPv6 to Embassy net, make IPv4 optional

rubdos:
This puts all IPv4-related code behind the `proto-ipv4` feature flag, and allows the use of IPv6 in the Embassy net stack. This is an initial, untested version.


rubdos:
Sorry for the noise, I should find out why rustfmt did not automatically run, and I ran into recent conflicts. This should be it :-)

ci.sh
```diff
@@ -25,11 +25,15 @@ cargo batch  \
...
     --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,dhcpv4,medium-ethernet,unstable-traits \
     --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,dhcpv4,medium-ethernet,nightly \
     --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,dhcpv4,medium-ethernet,unstable-traits,nightly \
+    --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,proto-ipv6,medium-ethernet \
+    --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,proto-ipv6,medium-ethernet,unstable-traits \
+    --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,proto-ipv6,medium-ethernet,nightly \
+    --- build --release --manifest-path embassy-net/Cargo.toml --target thumbv7em-none-eabi --features defmt,tcp,udp,dns,proto-ipv6,medium-ethernet,unstable-traits,nightly \
```
Dirbaio, line 36: could you add some tests enabling *both* IPv4 and IPv6?

embassy-net/src/lib.rs
```diff
@@ -113,8 +128,12 @@ impl Default for DhcpConfig {
 
 /// Network stack configuration.
 pub enum Config {
-    /// Use a static IP address configuration.
-    Static(StaticConfig),
+    /// Use a static IPv4 address configuration.
```
Dirbaio, line 131: To support using IPv4 and IPv6 at the same time, the config would have to look a bit different:

- ipv4 config: an enum of (static, dhcp, none)
- ipv6 config: an enum of (static, none) (smoltcp doesn't support DHCPv6 or IPV6 RAs yet)

"None" is useful in case you want to compile the firmware with both IPv4+IPv6 but the user can choose the configuration at runtime, perhaps using IPv4 only or IPv6 only.

Dirbaio: > I imagine the stack should be able to support multiple adresses of a single kind too

Oh, that's true, I didn't think about that. A `Vec<Config>` isn't quite right either though, because you can't have multiple default gateways of the same kind, for example. If we wanted to add support for this it'd make more sense to make the CIDR a Vec. So you can have multiple addrs, but only one default gateway. This is what NetworkManager does, for example.

> So what you propose is to split Config into ConfigV4 and ConfigV6, and make both optional, is that correct?

yeah, something like:

```rust
struct Config {
   ipv4: Ipv4Config,
   ipv6: Ipv6Config,
}
enum Ipv4Config {
   None,
   Static(StaticIpv4Config),
   Dhcp(Dhcpv4Config),
}
enum Ipv6Config {
   None,
   Static(StaticIpv6Config),
}
```

I think adding a `None` case to the enums is nicer than using Option, so that there is only one "level" of enums, not two.

rubdos: I have also added some convenience methods to `Config`, as not to have to manually construct the struct with a bunch of feature flags everywhere.

Dirbaio:
Looks great! Thank you for your work :)

PR #1540, opened 2023-06-06
Added can_recv for TcpSocket

RussHewgill:
The `may_recv` function returned True when the socket was open, but had no pending data. The smoltcp `can_recv` function instead only returns True when there is pending data.


Dirbaio (approved):
thanks!

bors r+

PR #1560, opened 2023-06-14
Feature/stm32g4 pll enhancements - Add PLL support for the P and Q outputs for G4 series chips

kevswims:
This PR adds onto https://github.com/embassy-rs/embassy/pull/1532 and I believe it completes https://github.com/embassy-rs/embassy/issues/1198 by supporting the rest of the PLL functionality. 

This also adds a USB Serial example for the G4 that uses the other PLL output to ensure it is working correctly. 

This has been tested on a custom STM32G473 board that I have.

@cstlaurent I made a few tweaks to the API you had created to add the other output configurations here. These changes align much closer with the API used for the H5 PLL config.

### DIFF
--- a/embassy-stm32/src/rcc/g4.rs
+++ b/embassy-stm32/src/rcc/g4.rs
@@ -17,7 +17,7 @@ pub const LSI_FREQ: Hertz = Hertz(32_000);
 pub enum ClockSrc {
     HSE(Hertz),
     HSI16,
-    PLLCLK(PllSrc, PllM, PllN, PllR),
+    PLL,
 }
 
 /// AHB prescaler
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/rcc/g4.rs
```diff
@@ -240,11 +337,80 @@ impl Default for Config {
...
+            Hertz(internal_freq / div_q.to_div())
+        });
+
+        let pll_r_freq = pll_config.div_r.map(|div_r| {
+            RCC.pllcfgr().modify(|w| {
+                w.set_pllr(div_r.into());
+                w.set_pllren(true);
+            });
+            Hertz(internal_freq / div_r.to_div())
+        });
```
cstlaurent, line 401: Should we add asserts to make sure the P,Q,R clock domains do not go over 170Mhz like proposed in the ref manual p295:
CleanShot 2023-06-14 at 14 06 18@2x

cstlaurent (commented):
Wow nice PR @kevswims , looks much cleaner, I based mine on the L4, but I guess the H5 is the newest development and the proper direction, I'm all in for the change.

Only one small comment. I will test this out tonight on my board.

examples/stm32g4/src/bin/usb_serial.rs
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    config.rcc.mux = ClockSrc::PLL;
+
+    let p = embassy_stm32::init(config);
+    info!("Hello World!");
+
+    unsafe {
+        pac::RCC.ccipr().write(|w| w.set_clk48sel(0b10));
+    }
```
cstlaurent, line 43: Do you think we should add this HSI48 clk select in `config.rcc` as well.It seems to be the only way of having more than 144Mhz on the main CPU while using usb

CleanShot 2023-06-14 at 14 21 53@2x

embassy-stm32/src/rcc/g4.rs
kevswims: @cstlaurent I like the idea of checking for invalid frequencies here but I'm not quite sure what the best approach for these asserts is. The reference manual notes not to exceed 170 MHz but there's a lot of invalid frequencies below that depending on which voltage ranges the part is operating in. See the following specs from the datasheet:

image

examples/stm32g4/src/bin/usb_serial.rs
kevswims: I'll double check this. I'd like to have this example show how to use both the HSI48 and the PLL output. I'm not sure if there's any more setup that has to be done with the RCC to accomplish that or not. I know the HSI48 requires the clock recovery system to be setup to adjust it's frequency but I will take a look at it.

kevswims: Here's an example of using the CRS and the 48 MHz internal clock for USB: https://github.com/kevswims/stm32g4xx-hal/blob/develop/examples/usb_serial.rs

I'll see if I can get this added into the RCC as well though I don't think it needs to be part of the initial configuration since it is pretty independent of everything else.

embassy-stm32/src/rcc/g4.rs
cstlaurent: Right, I missed this table, not sure why only the 170Mhz was specified in those config section tho. Maybe there is a chance to damage the device 🤷 ?

examples/stm32g4/src/bin/usb_serial.rs
cstlaurent: Similarly to the H5 config there is a boolean in the config to enable/disable the hsi48. But I agree the CK48 mux could be independent

kevswims: I'll add a flag to match that API then. That seems reasonable to me.

cstlaurent: From the generated STM32Cube code it just set the enable bit of the HSI48, no need to configure the other frequency recovery registers (I guess default parameters are fine). Basically the same as what is done in the H5

CleanShot 2023-06-14 at 15 09 05@2x

and 

CleanShot 2023-06-14 at 15 10 38@2x

kevswims: The clock recovery is not required to turn it on but it is required to use it for USB applications since the oscillator itself can drift too much. The CRS uses info from the USB host to trim the frequency to be within the specs that are needed for USB operation.

cstlaurent: Ok I see, I was not aware of that part, good to know 😄

Dirbaio (approved):
Thank you!

PR #1563, opened 2023-06-16
embassy-executor: introduce `InterruptExecutor::spawner()`

kaspar030:
This PR adds a `spawner()` method to the interrupt executor.
This allows spawning a task from somewhere (not only where the executor has been started).


Dirbaio (approved):
ooh, this is very useful. Thank you! :)

PR #1566, opened 2023-06-17
tl-mbox: switch to new ipcc mechanism

xoviat:
It works!


xoviat:
I assume that this will work with more complex commands, but if that isn't the case, we may have to back out. Most indications are positive at this point, though.

PR #1569, opened 2023-06-19
wpan: misc. cleanup and add mac

xoviat:
This fixes a few bugs I found with the last pr and adds basic mac functionality.

One thing I noticed was that mac was a bit different than ble: it does not use the linked list for events. In response, I modified the code structure a bit to mask the difference with ble so that there is a consistent api between the two.


xoviat:
@OueslatiGhaith just have a look at this to see if it looks reasonable.

embassy-stm32-wpan/src/mac.rs
```diff
@@ -0,0 +1,109 @@
...
+            });
+        }
+
+        Self { phantom: PhantomData }
+    }
+
+    /// SAFETY: passing a pointer to something other than a managed event packet is UB
+    pub(crate) unsafe fn drop_event_packet(_: *mut EvtPacket) {
+        // Write the ack
+        CmdPacket::write_into(MAC_802_15_4_CMD_BUFFER.as_mut_ptr(), TlPacketType::OtAck, 0, &[]);
```
xoviat, line 45: bug!

PR #1572, opened 2023-06-19
BDMA: request stop after busy loop in blocking_wait()

whitequark:
Otherwise the channel cannot be used again, since CR.EN remains set and the DMA channel registers are read-only while it is set.


JuliDi:
I can confirm that it does not work with the current implementation and this PR fixes it on my STM32L432KC

PR #1574, opened 2023-06-19
Prevent bootloader revert and add missing examples to CI

lulf:
Prevent accidental revert when using firmware updater
    
    This change prevents accidentally overwriting the previous firmware before
    the new one has been marked as booted.

Add firmware updater examples to CI
    
    CI was not building the a.rs application due to the requirement of b.bin
    having been built first. Add a feature flag to examples so that CI can
    build them including a dummy application.
    
    Update a.rs application examples so that they compile again.

I was tempted to give a go at adding the magic buffer to FirmwareUpdater type but decided against it for now and will probably raise a separate PR for that.


Dirbaio (approved):
Thanks!

PR #1576, opened 2023-06-22
wpan: add BLE HCI

OueslatiGhaith:
this reexports a forked version of [bluetooth-hci](https://github.com/danielgallagher0/bluetooth-hci) that removes `nb`.
The fork still has the same code size issues, but work is being done to reduce it


embassy-stm32-wpan/src/evt.rs
```diff
@@ -105,6 +106,14 @@ impl EvtBox {
         Self { ptr }
     }
 
+    pub fn evt<'a>(&self) -> &'a [u8] {
```
Dirbaio, line 109: This is unsound. It allows you to hold on to the slice after you've dropped the packet.

You probably want `pub fn evt<'a>(&'a self) -> &'a [u8]`.
Or `pub fn evt(&self) -> &[u8]` which is the same using lifetime elision.

OueslatiGhaith: oh this is unused, forgot to remove it

embassy-stm32-wpan/src/evt.rs
```diff
@@ -124,6 +133,20 @@ impl EvtBox {
             slice::from_raw_parts(p_payload, payload_len as usize)
         }
     }
+
+    /// writes an underlying [`EvtPacket`] into the provided buffer.
+    /// Returns the number of bytes that were written.
+    /// Returns an error if event kind is unknown or if provided buffer size is not enough.
+    pub fn serial<'a>(&self) -> &'a [u8] {
```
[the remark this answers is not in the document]
OueslatiGhaith: copy on read should solve this, as the result of this operation may live longer that the packet itself

xoviat:
alright, looks good!

PR #1577, opened 2023-06-22
disable pull-up and down resistors for rp adc blocking_read

dbdbc:
Changed ```Adc::blocking_read``` to behave the same way as the async ```Adc::read``` by disabling the pull-up and pull-down resistors before starting the adc.

Note that this increases the time it takes the function to run slightly, in my measurements with default settings it took on average 3.00 µs instead of 2.89 µs.


Dirbaio (approved):
thanks!

PR #1578, opened 2023-06-23
stm32 can split method

schphil:
This is a first draft implementing a split method which allows to simultaneously transmit and receive data via the can interface. 

- [x] make it work
- [x] add an [example](https://github.com/schphil/embassy/blob/can-split/examples/stm32f7/src/bin/can.rs)
- [x] how to change the baudrate after splitting an instance

Any thoughts on how this can be improved or made cleaner are welcome.


xoviat:
@schphil Can you rustfmt your code? Do you need help with that?

As far as changing the baudrate, I would suggest that you can drop `CanTx` and `CanRx`, change the baud-rate, and then get them back. In the application, you can store `Option<CanTx>` and `Option<CanRx>`.

schphil:
@xoviat thanks for the fmt hint. 

yep, i also thought about doing it the way you said, but I didn't think it through to the end. that should suit my needs. thanks.

xoviat:
Are two lifetime parameters required for the tx and rx structs? Could they be combined?

schphil:
not sure if I get you right. are you referring to removing the lifetime c? something like that: 

```rust 
pub fn split(&'d self) -> (CanTx<'d, T>, CanRx<'d, T>) {
   (CanTx { can: &self.can }, CanRx { can: &self.can })
 }
```
```rust 
pub struct CanTx<'d, T: Instance> {
    can: &'d RefCell<bxcan::Can<BxcanInstance<'d, T>>>,
}
```
```rust 
pub struct CanRx<'d, T: Instance> {
    can: &'d RefCell<bxcan::Can<BxcanInstance<'d, T>>>,
}
```

xoviat:
I was wrong about the lifetimes; two are required.

PR #1582, opened 2023-06-25
Add the high resolution timer

xoviat:
Not everything is added, but it's a good start.


embassy-stm32/src/hrtim/traits.rs
```diff
@@ -0,0 +1,200 @@
...
+        /// Set the dead time as a proportion of max_duty
+        fn set_channel_dead_time(channnel: usize, dead_time: u16);
+
+        //        fn enable_outputs(enable: bool);
+        //
+        //        fn enable_channel(&mut self, channel: usize, enable: bool);
+    }
+}
+
+pub trait HighResolutionCaptureCompare16bitInstance:
```
Dirbaio, line 112: this could be just `Instance`.

embassy-stm32/src/hrtim/traits.rs
```diff
@@ -0,0 +1,200 @@
+use crate::rcc::sealed::RccPeripheral;
+use crate::time::Hertz;
+
+#[derive(Clone, Copy)]
+pub(crate) enum HighResolutionControlPrescaler {
```
Dirbaio, line 5: it's already known it's "HighResolution" now that it's in the `hrtim` module. It could be just "ControlPrescaler"

embassy-stm32/src/hrtim/mod.rs
```diff
@@ -0,0 +1,409 @@
+mod traits;
```
Dirbaio, line 1: missing `pub use traits::Instance`.

embassy-stm32/src/hrtim/traits.rs
```diff
@@ -0,0 +1,200 @@
...
+            HighResolutionControlPrescaler::Div16 => 16,
+            HighResolutionControlPrescaler::Div32 => 32,
+            HighResolutionControlPrescaler::Div64 => 64,
+            HighResolutionControlPrescaler::Div128 => 128,
+        }
+    }
+}
+
+impl From<HighResolutionControlPrescaler> for u8 {
+    fn from(val: HighResolutionControlPrescaler) -> Self {
```
Dirbaio: it is very strange that converting to u32 gives the prescaler value and converting to u8 gives the register value. Perhaps this is a missing enum that could be added to stm32-metapac?

xoviat: I will add it later, but I would not like to wait for another metapac release.

xoviat: Also, this code doesn't leave this crate.

Dirbaio: :smiling_face_with_tear:

Dirbaio (approved):
thank you!

PR #1583, opened 2023-06-25
Allow arbitrary expressions as `task(pool_size)`

bugadani:
This PR relaxes `pool_size` to allow arbitrary expressions. Unfortunately, this also means that the attribute now accepts arbitrary paths, meaning the attribute may cause compile errors not constructed by syn/darling/embassy (e.g. use of undeclared module, etc.).

The intent of this PR is to allow controlling both the pool size, and the length of some global resource array with the same constant.

### DIFF
--- a/embassy-macros/src/macros/task.rs
+++ b/embassy-macros/src/macros/task.rs
@@ -1,3 +1,4 @@
+use darling::export::NestedMeta;
 use darling::FromMeta;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
@@ -11,8 +12,8 @@ struct Args {
     pool_size: Option<usize>,
 }
 
-pub fn run(args: syn::AttributeArgs, f: syn::ItemFn) -> Result<TokenStream, TokenStream> {
-    let args = Args::from_list(&args).map_err(|e| e.write_errors())?;
+pub fn run(args: &[NestedMeta], f: syn::ItemFn) -> Result<TokenStream, TokenStream> {
+    let args = Args::from_list(args).map_err(|e| e.write_errors())?;
 
     let pool_size: usize = args.pool_size.unwrap_or(1);
 
--- a/embassy-macros/src/macros/task.rs
+++ b/embassy-macros/src/macros/task.rs
@@ -1,21 +1,24 @@
 use darling::export::NestedMeta;
 use darling::FromMeta;
-use proc_macro2::TokenStream;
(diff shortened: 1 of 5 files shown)


xoviat:
What about allowing pool size to be a const?

embassy-macros/src/macros/task.rs
```diff
@@ -45,8 +79,10 @@ pub fn run(args: syn::AttributeArgs, f: syn::ItemFn) -> Result<TokenStream, Toke
         },
     }
 
-    if pool_size < 1 {
-        ctxt.error_spanned_by(&f.sig, "pool_size must be 1 or greater");
+    if let PoolSize::Literal(pool_size) = pool_size {
+        if pool_size < 1 {
```
Dirbaio, line 83: I wonder if we could/should get rid of this check?
- If you're controlling the pool size based on some global const based on some calculation or Cargo feature, it might make sense to set it as zero in some build configs. (like, use 0 TCP server tasks if this product variant has no Ethernet)
- It's weird that zero length is allowed via a const, but not if you put the literal 0 in the meta.

This would allow parsing the attr as just "expression", instead of "literal or expression", which would simplify this a lot I think?

Dirbaio:
Could you add an example (any chip) that uses the new pool_size style? That'll double a CI check so we don't break it in the future.

Also, update to syn 2. nice! :heart:

embassy-macros/src/macros/task.rs
bugadani: > It's weird that zero length is allowed via a const

Yeah, that is actually an unintended side effect. I don't think proc macros can access the value of a constant. This also means we can't reject paths to non-constants here.

I guess I can try arbitrary expressions. Not sure how much simpler things can get, we'll see.

Dirbaio: ah I missed you're parsing it as "literal or path", not "literal or expression" Yeah, I'm not sure if arbitrary expressions is a good idea. Even if it still accepts only literals or paths, removing the check should allow removing the enum which simplifies things a bit at least :sweat_smile:

bugadani:
Huh, I got spoofed by a local failure but this seems to work well. We can now do stuff like `#[task(pool_size = CONSTANT - 1)]` for example, as long as the expression evals to a `usize`.

Dirbaio (approved):
Very nice! Thank you :)

PR #1591, opened 2023-06-28
Make StackResources::new() const

bugadani:
So, I don't actually know for a fact if this works, but since the constructor seems to only initialize arrays, I don't see a reason why not. CI shall tell us :)


Dirbaio (approved):
CI said yes! :D

PR #1592, opened 2023-06-28
G4 RCC CRS support and Working CRS USB Example

kevswims:


kevswims:
Has a WIP tag until I get the CRS stuff documented properly.

embassy-stm32/src/rcc/g4.rs
```diff
@@ -512,3 +512,12 @@ pub(crate) unsafe fn init(config: Config) {
         apb2_tim: Hertz(apb2_tim_freq),
     });
 }
+
+pub(crate) unsafe fn enable_hsi48() {
+    RCC.crrcr().modify(|w| w.set_hsi48on(true));
+
+    info!("Waiting for HSI48 to turn on");
```
kevswims, line 519: Take these logs out.

embassy-stm32/src/crs.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn new(peripheral: impl Peripheral<P = CRS> + 'd, config: Config) -> Self {
+        unsafe {
+            enable_hsi48();
+        }
+
+        CRS::enable();
+
+        match config.sync_source {
+            SyncSource::Gpio => {
+                unimplemented!("GPIO CRS Sync Source is not supported");
```
kevswims, line 43: I don't have a good setup to test these but I can at least set the source bits properly.

embassy-stm32/src/crs.rs
```diff
@@ -0,0 +1,67 @@
+use embassy_hal_common::PeripheralRef;
+
+use crate::pac::CRS as PAC_CRS;
+use crate::peripherals::CRS;
+use crate::rcc::enable_hsi48;
+use crate::rcc::sealed::RccPeripheral;
+use crate::{pac, Peripheral};
+
+pub struct Config {
+    sync_source: SyncSource,
```
kevswims, line 10: What other config options need to be added here? The default reset behavior is enough for USB. Do we need to support the other configuration stuff?

embassy-stm32/src/rcc/g4.rs
```diff
@@ -512,3 +512,12 @@ pub(crate) unsafe fn init(config: Config) {
         apb2_tim: Hertz(apb2_tim_freq),
     });
 }
+
+pub(crate) unsafe fn enable_hsi48() {
```
kevswims, line 516: Need to look at which other chips have the CRS module and add this to them as well.

embassy-stm32/src/crs.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+pub enum SyncSource {
+    Gpio,
+    Lse,
+    Usb,
+}
+
+pub struct Crs<'d> {
```
Dirbaio, line 19: I think CRS should be fields in the RCC `Config` struct, instead of a user-facing driver struct. It's kind of part of RCC, and as a user you don't do much with it directly other than configuring it at boot.

kevswims: Works for me, I went back and forth on how the config should work for this a few times. I'll get that changed.

embassy-stm32/src/rcc/g4.rs
kevswims: Since the CRS work is moved into the g4 RCC module I'm not going to tackle this right now. If someone has a G0 device with USB that I could test with I would be willing to port this over.

Dirbaio:
Thank you! :rocket:

PR #1599, opened 2023-06-29
feature(1355): Add trigger to task, triggered + clear to Event

ilikepi63:
Added utility methods to NRF Task + Event structs.

### DIFF
--- a/embassy-nrf/src/ppi/mod.rs
+++ b/embassy-nrf/src/ppi/mod.rs
@@ -137,6 +137,11 @@ impl Task {
         Self(ptr)
     }
 
+    // Triggers this task.
+    pub unsafe fn trigger(&mut self) {
+        self.0.write(|w| unsafe { w.bits(1) });
+    }
+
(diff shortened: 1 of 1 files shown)


Dirbaio:
Looks good! I wouldn't mark the functions as `unsafe` though, since there is no safety requirement the user has to fulfill to call them.

embassy-nrf/src/ppi/mod.rs
```diff
@@ -137,9 +137,9 @@ impl Task {
         Self(ptr)
     }
 
-    // Triggers this task.
+    /// Triggers this task.
     pub unsafe fn trigger(&mut self) {
-        self.0.write(|w| unsafe { w.bits(1) });
+        *self.0.as_ptr() = 1;
```
Dirbaio: register access needs to be volatile reads/writes, not regular pointer reads/writes. (needed to prevent the compiler from reoderding / optimizing them out).

Dirbaio (approved):
Thank you! :)

PR #1600, opened 2023-06-29
feature(1354): Added lifetimes to Event + Tasks

ilikepi63:
Added in lifetimes for ppi Events + Tasks

### DIFF
--- a/embassy-nrf/src/ppi/mod.rs
+++ b/embassy-nrf/src/ppi/mod.rs
@@ -15,6 +15,7 @@
 //! many tasks and events, but any single task or event can only be coupled with one channel.
 //!
 
+use core::marker::PhantomData;
 use core::ptr::NonNull;
 
 use embassy_hal_common::{impl_peripheral, into_ref, PeripheralRef};
@@ -125,16 +126,16 @@ const REGISTER_DPPI_CONFIG_OFFSET: usize = 0x80 / core::mem::size_of::<u32>();
 /// When a task is subscribed to a PPI channel, it will run when the channel is triggered by
 /// a published event.
 #[derive(PartialEq, Eq, Clone, Copy)]
-pub struct Task(NonNull<u32>);
+pub struct Task<'d>(NonNull<u32>, PhantomData<&'d ()>);
 
-impl Task {
+impl<'d> Task<'_> {
     /// Create a new `Task` from a task register pointer
     ///
     /// # Safety
     ///
     /// `ptr` must be a pointer to a valid `TASKS_*` register from an nRF peripheral.
     pub unsafe fn new_unchecked(ptr: NonNull<u32>) -> Self {
-        Self(ptr)
+        Self(ptr, PhantomData)
     }
 
     /// Triggers this task.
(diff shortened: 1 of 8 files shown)


ilikepi63:
@Dirbaio I've implemented this. I realized that we potentially should be treating the tasks and events with their own lifetimes, but bounded to the driver's lifetime. 

This achieves what we want to achieve and removes the need to try and solve multiple references as the case was when we have a ppi_channel made up of a ppi_group disable_all task that is also references from the ppigroup. 

Hope this works!

Dirbaio:
@ilikepi63: the extra lifetime param is not needed.

```rust
fn task_capture<'s: 'd>(&self) -> Task<'s>
```

behaves in practice the same as 

```rust
fn task_capture(&self) -> Task<'d>
```

`Task` is [covariant](https://doc.rust-lang.org/nomicon/subtyping.html), i.e. you can convert from `Task<'long>` to `Task<'short>`. So, you can call `.task_capture()` to get a `Task<'d>` and then implicitly convert it to a task with a shorter lifetime. You don't need an explicit `'s` in the signature to allow that.

Dirbaio:
You're missing the lifetimes in PWM and GPIOTE. I think that should fix the CI issue.

embassy-nrf/src/ppi/mod.rs
```diff
@@ -125,16 +126,16 @@ const REGISTER_DPPI_CONFIG_OFFSET: usize = 0x80 / core::mem::size_of::<u32>();
 /// When a task is subscribed to a PPI channel, it will run when the channel is triggered by
 /// a published event.
 #[derive(PartialEq, Eq, Clone, Copy)]
-pub struct Task(NonNull<u32>);
+pub struct Task<'d>(NonNull<u32>, PhantomData<&'d ()>);
 
-impl Task {
+impl<'d> Task<'_> {
```
Dirbaio, line 131:
```suggestion
impl<'d> Task<'d> {
```

embassy-nrf/src/ppi/mod.rs
```diff
@@ -156,26 +160,29 @@ impl Task {
...
 
 /// Represents an event that a peripheral can publish.
 ///
 /// An event can be set to publish on a PPI channel when the event happens.
 #[derive(PartialEq, Eq, Clone, Copy)]
-pub struct Event(NonNull<u32>);
+pub struct Event<'d>(NonNull<u32>, PhantomData<&'d ()>);
 
-impl Event {
+impl<'d> Event<'_> {
```
Dirbaio, line 171:
```suggestion
impl<'d> Event<'d> {
```

embassy-nrf/src/ppi/ppi.rs
```diff
@@ -3,12 +3,12 @@ use embassy_hal_common::into_ref;
 use super::{Channel, ConfigurableChannel, Event, Ppi, StaticChannel, Task};
 use crate::{pac, Peripheral};
 
-impl Task {
+impl<'d> Task<'_> {
```
Dirbaio, line 6:
```suggestion
impl<'d> Task<'d> {
```

embassy-nrf/src/ppi/ppi.rs
```diff
@@ -3,12 +3,12 @@ use embassy_hal_common::into_ref;
...
     fn reg_val(&self) -> u32 {
         self.0.as_ptr() as _
     }
 }
-impl Event {
+impl<'d> Event<'_> {
```
Dirbaio, line 11:
```suggestion
impl<'d> Event<'d> {
```

ilikepi63:
Thanks for the help on the lifetimes :) It seems I have quite a bit of reading to do on it. 

Looks like you were right on the missing stuff. Thanks!

Dirbaio (approved):
Looks great! Thanky ou

PR #1610, opened 2023-07-03
Address I2C lockup issue

ShakenCodes:
Fixes issue where I2C driver will not timeout (as timeout functions were Ok(()) noops.)
Ensures master_stop() is called when an error occurs.
Does not identify or address the I2C issue that exposed these weaknesses.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -382,35 +382,59 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
         // I2C start
         //
         // ST SAD+W
-        Self::master_write(
+        match Self::master_write(
             address,
             write.len().min(255),
             Stop::Software,
             last_chunk_idx != 0,
             &check_timeout,
-        )?;
+        ) {
+            Ok(()) => {
+                for (number, chunk) in write.chunks(255).enumerate() {
+                    if number != 0 {
+                        Self::master_continue(chunk.len(), number != last_chunk_idx, &check_timeout)?;
+                    }
 
-        for (number, chunk) in write.chunks(255).enumerate() {
-            if number != 0 {
-                Self::master_continue(chunk.len(), number != last_chunk_idx, &check_timeout)?;
+                    for byte in chunk {
+                        // Wait until we are allowed to send data
+                        // (START has been ACKed or last byte when
+                        // through)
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/i2c/v2.rs
```diff
@@ -382,35 +383,59 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
         // I2C start
         //
         // ST SAD+W
-        Self::master_write(
+        match Self::master_write(
```
Dirbaio, line 386: please write these with "early return" style. This avoids too much neested indentation and keeps the error handling close to the source errors.


```rust
if let Err(e) = Self::master_write(..) {
   ... handle it
   return Err(e)
}

Dirbaio:
Sorry for the delay, we weren't sure what the best API for timeouts would be. 

Could you make the timeout configurable, instead of hardcoded to 100ms? It should be a `pub timeout: Option<Duration>` in `Config`. So even if the user enables the `time` feature, they can choose to not enable timeouts. 

Other than that, PR looks good so if the timeout is configurable it should be good to go. Thanks! :)

Dirbaio:
Sorry for the delay, we weren't sure what the best API for timeouts would be. 

Could you make the timeout configurable, instead of hardcoded to 100ms? It should be a `pub timeout: Option<Duration>` in `Config`. So even if the user enables the `time` feature, they can choose to not enable timeouts. 

Other than that, PR looks good so if the timeout is configurable it should be good to go. Thanks! :)

ShakenCodes:
Thank you for the feedback. I will make the requested change shortly. (I am on a small work emergency right now, so it might be a few days...)

ShakenCodes:
I was looking at incorporating the requested improvement to the I2C driver and noticed a file in the directory named [timeout.rs](http://timeout.rs/) that seems to be doing what was intended all along. Have I just been using the wrong instance of the I2C? It looks like if you use TimeoutI2c::new() around I2c::new(), you get the timeout functionality I thought was missing.
If this is correct, I am happy to cancel/retract this PR.

Dirbaio:
oh, oops. I hadn't noticed that either. Then yes, we shouldn't add timeouts to the main i2c.

The "do stop on error" still seems a change worth doing. Could you send a PR with just that?

ShakenCodes:
This is unneeded due to existing TimeoutI2c feature.
Will open a separate PR for calling master_stop(0 consistently.

PR #1612, opened 2023-07-04
Add stm32 FMC support for bank 1 16-bit RAM

wyager:
We only had support for 16-bit modules on bank 2.


Dirbaio (approved):
Thanks! :rocket:

PR #1617, opened 2023-07-04
stm32/rcc: allow const-propagation

xoviat:
This allows the UB time-travel optimization to propagate clock frequencies.


embassy-stm32/src/rcc/mod.rs
```diff
@@ -97,8 +97,11 @@ pub mod low_level {
 }
 
 pub(crate) mod sealed {
+    use super::Clocks;
+
     pub trait RccPeripheral {
         fn frequency() -> crate::time::Hertz;
+        fn clocks() -> &'static Clocks;
```
Dirbaio, line 104: Why is this method needed? It does not make much sense to have it in `RccPeripheral`. "Get *my* clock" makes sense (that's what `frequency()` does) but "get *all* clocks" doesn't.

Dirbaio (approved):

PR #1621, opened 2023-07-05
Remove bad semicolon in time driver example struct declaration

rasmuspeders1:
Discovered this when I simply copy pasted the example from the time driver example.
The semicolon is not allowed with struct declarations with braces.

The doc test compiles fine for some reason!? (am I misunderstanding something here?)

Thought this might help others copy pasting the example in the future.


Dirbaio (approved):
Thanks

PR #1623, opened 2023-07-05
rp/adc: rewrite the module

pennae:
- don't require an irq binding for blocking-only adc
- abstract adc pins into an AnyPin like interface, erasing the actual peripheral type at runtime.
- add pull-up/pull-down functions for adc pins
- add a test (mostly a copy of the example, to be honest)
- configure adc pads according to datasheet
- drop embedded-hal interfaces. embedded-hal channels can do neither AnyPin nor pullup/pulldown without encoding both into the type

incorporates #1286, which seems to cause no problems. should not interfere with other drivers because during pin drop we reset the IE/OD bits to defaults.

### DIFF
--- a/tests/rp/src/bin/adc.rs
+++ b/tests/rp/src/bin/adc.rs
@@ -1,0 +1,86 @@
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+#[path = "../common.rs"]
+mod common;
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_rp::adc::{Adc, Config, InterruptHandler, Pin};
+use embassy_rp::bind_interrupts;
+use embassy_rp::gpio::Pull;
+use {defmt_rtt as _, panic_probe as _};
+
+bind_interrupts!(struct Irqs {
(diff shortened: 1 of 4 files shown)


dbdbc:
If I'm not mistaken there is currently no way to disable the temperature sensor once it was activated. I think there should be one, as it permanently draws power. Possible ways to do this could be
- automatically deactivating it again after measuring
- letting the user manually disable it
- a solution similar to the one the rp-hal repository used (handing a ```TempSense``` struct to the user).

tests/rp/src/bin/adc.rs
```diff
@@ -0,0 +1,84 @@
...
+#![no_main]
+#![feature(type_alias_impl_trait)]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_rp::adc::{Adc, Config, InterruptHandler, Pin};
+use embassy_rp::bind_interrupts;
+use embassy_rp::gpio::Pull;
+use {defmt_rtt as _, panic_probe as _};
+
```
Dirbaio, line 11: missing `#[path = "../common.rs"] mod common;`

Dirbaio (approved):

PR #1627, opened 2023-07-07
rp/pio: use bind_interrupts for irqs

### LINKED ISSUE #1338: embassy-rp PIO drivers should use `irq!()` macro
Right now, there is no way for the user to access the PIO interrupts when the `pio` feature is used. I want to implement a communication protocol that requires fairly tight timing, so I'd rather have access to the raw interrupts of one PIO instance instead of using `irq(n).await`. Right now, the PIOs are the only peripherals that has all interrupts blocked by embassy:
- DMA: while `DMA_IRQ_0` is implemented by embassy, `DMA_IRQ_1` is left to the user
- GPIO and `multicore`: I can see that embassy always claims these interrupts, though I'd love to see some way to hook into the GPIO IRQ and execute some timing critical user code before the embassy interrupt handler
- Timer: without the `time-driver` feature, the user can implement this interrupt handler

While the embassy PIO interrupt handlers can be disabled using the `pio` feature, there is no way to expose one PIO completely to the user while leaving the other managed by embassy.

I propose two solutions:
- pass in the interrupts with the `irq!()` macro when creating an instance of the PIO driver
- have two separate feature flags `pio0` and `pio1`

pennae:
closes #1338

### DIFF
--- a/embassy-rp/src/pio.rs
+++ b/embassy-rp/src/pio.rs
@@ -16,12 +16,12 @@ use pio::{SideSet, Wrap};
 use crate::dma::{Channel, Transfer, Word};
 use crate::gpio::sealed::Pin as SealedPin;
 use crate::gpio::{self, AnyPin, Drive, Level, Pull, SlewRate};
-use crate::interrupt::InterruptExt;
+use crate::interrupt::typelevel::{Binding, Handler, Interrupt};
 use crate::pac::dma::vals::TreqSel;
 use crate::relocate::RelocatedProgram;
-use crate::{interrupt, pac, peripherals, pio_instr_util, RegExt};
+use crate::{pac, peripherals, pio_instr_util, RegExt};
 
-struct Wakers([AtomicWaker; 12]);
+pub struct Wakers([AtomicWaker; 12]);
 
 impl Wakers {
     #[inline(always)]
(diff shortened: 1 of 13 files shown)


embassy-rp/src/pio.rs
```diff
@@ -85,42 +85,20 @@ const RXNEMPTY_MASK: u32 = 1 << 0;
...
-    let ints = pac::PIO1.irqs(0).ints().read().0;
-    for bit in 0..12 {
-        if ints & (1 << bit) != 0 {
-            WAKERS[1].0[bit].wake();
+impl<PIO: Instance> Handler<PIO::Interrupt> for InterruptHandler<PIO> {
+    unsafe fn on_interrupt() {
+        let ints = PIO::PIO.irqs(0).ints().read().0;
+        for bit in 0..12 {
+            if ints & (1 << bit) != 0 {
+                WAKERS[0].0[bit].wake();
```
Dirbaio, line 97: this needs to use `WAKERS[1]` for PIO1.

It'd be nice to change it so that each PIO has its own wakers static, so that if you only use PIO0 the static for PIO1 can get optimized out.

Dirbaio (approved):
Thanks! :rocket:

Issue #1298, opened 2023-03-25
STM32H7 HAL: missing support for "wide VCO range"


diggit:
Hi,
STM32 H7 HAL does not seem to support VCO Wide Range mode. As a result, maximum frequency is capped at 420 MHz (H730 can do 550 MHz). Wide VCO range is 192 to 836 MHz.

Whole `pll` module would require some changes.
- VCO config: https://github.com/embassy-rs/embassy/blob/299689dfa2a5e160dbd6aa474772a9317a219084/embassy-stm32/src/rcc/h7.rs#L759
- asserts: https://github.com/embassy-rs/embassy/blob/299689dfa2a5e160dbd6aa474772a9317a219084/embassy-stm32/src/rcc/h7.rs#L715
- etc

I am total Rust noob (literally started 3 days ago), so I am not able to create PR with decent code quality.

I have one maybe a bit OT question, but would it be possible to make RCC config more compile time matter (optionally)?
At this moment all math is done in runtime and whole RCC config stuff takes over 1kiB on size optimized build.
In the end all checks could be done in compile time and then only registers would get filled (if config is compile time constant).
In C++, I'd leverage `constexpr` for this stuff, but I don't know options in rust and current HAL design.

diggit:
Also, PLL channel output P is always required even though, it is possible to disable it via config bit.
affected fcn: https://github.com/embassy-rs/embassy/blob/299689dfa2a5e160dbd6aa474772a9317a219084/embassy-stm32/src/rcc/h7.rs#L788
Should I open separate issue for this?

Dirbaio:
Please open PRs! Don't worry if you're new to Rust, we can work things out on PR review. :)

Dirbaio:
Fixed in embassy-stm32 v0.2.0

Issue #1092, opened 2022-12-03
`embassy-stm32::init` never ends execution on `STM32H7A3ZIT6Q`


indietyp:
I have been trying to get `embassy` running on the [STM32H7A3 Nucleo-144](https://www.st.com/en/evaluation-tools/nucleo-h7a3zi-q.html), where initialization never seems to end, after a lot of ~~pain and suffering~~ trying to debug the issue, I think I found the culprit.

Code without embassy seems to work fine, but during the initialization phase, the `flash_setup` this specific function call never seems to return/end:

https://github.com/embassy-rs/embassy/blob/f109e73c6d7ef2ad93102b7c8223f5cef30ef36f/embassy-stm32/src/rcc/h7.rs#L257-L262

Specifically, `FLASH.acr().write(|w| /* */)` seems to hang. The closure executes just fine, but no code is executed once the closure returns.

I took the `blinky` example and adapted it to my chip. This roughly meant: changing the probe-run chip in .cargo to `STM32H7A3ZITxQ`, removing the `memory.x`, and enabling the `"stm32h7a3zi"` and `"memory-x"` features for embassy-stm32.

The configuration for the peripherals is still the same and only uses the default. I extracted some additional "data" from the `flash_setup` function: `wait_states`: `0`, `progr_delay`: `0`, `rcc_aclk_mhz`: `31`.

indietyp:
The minimal rust code I was working with:

```rust
#![warn(
    missing_docs,
    clippy::pedantic,
    clippy::nursery,
    clippy::undocumented_unsafe_blocks
)]
#![no_std]
#![no_main]
#![feature(type_alias_impl_trait)]
#![feature(provide_any)]
#![feature(alloc_error_handler)]
#![feature(lint_reasons)]

use defmt_rtt as _;
use embassy_executor::Spawner;
use panic_probe as _;

// TODO: I2C Timeout
#[embassy_executor::main]
async fn main(_spawner: Spawner) {
    let _p = embassy_stm32::init(Default::default());
    defmt::panic!("Successfully initialized");
}
```

PR #1629, opened 2023-07-08
Correctly calculate target VCO frequency from multipliers

davidpurser:
The STM32H7 reference manual (RM0433) specifies PLL output frequency with these formulas:

`F_ref_ck = F_pll_src / DIVM`
`F_vco = F_ref_ck * DIVN`
`F_pll_p_ck = F_vco / (DIVP+1)`

... where `F_pll_p_ck` is the output of the PLL postdivider P.

Expanding this and solving for `F_vco`, we get:

`F_vco = F_pll_p_ck * (DIVP+1)`

i.e. `vco_ck = output * pll_x_p`

But the formula in the H7 RCC is incorrectly specified as `vco_ck = output + pll_x_p`.  Therefore asking for a PLL output frequency of 80 MHz causes `vco_ck` to be calculated as `80000005`, which is nonsense (you cannot actually realize a VCO frequency of 80000005 with only integer multipliers), and leads to a panic as `vco_ck` is out of range.

With this fix, asking for a PLL output frequency of 80 MHz with a PLL source clock of 16 MHz results in the correct set of parameters:

```
pll_src = 16 MHz
pll_x_m = 8
pll_x_n = 200
pll_x_p = 5
output = 16 MHz / 8 * 200 / 5 = 80000000
```

Similarly, asking for 400 MHz output also gives correct parameters:

```
pll_src = 16 MHz
pll_x_m = 8
pll_x_n = 200
pll_x_p = 1
output = 16 MHz / 8 * 200 / 1 = 400000000
```

I believe `*` was the intended operation here, and it seems to be working properly for me with the change.  Hoping someone else can confirm my logic.

### DIFF
--- a/embassy-stm32/src/rcc/h7.rs
+++ b/embassy-stm32/src/rcc/h7.rs
@@ -740,7 +740,7 @@ mod pll {
             }
         };
 
-        let vco_ck = output + pll_x_p;
+        let vco_ck = output * pll_x_p;
 
         assert!(pll_x_p < 128);
         assert!(vco_ck >= VCO_MIN);


xoviat:
Thanks!

JuliDi:
I suspect this PR breaks the clock on the H7A3ZI. My project has been working with commit fa2cda81db03f66c803c4ae931e1603973750548 and with dff9bd9711205fd4cd5a91384072ab6aa2335d18 it now doesn't.

This used to be working (though if I set 250 MHz, it would break, as well)
```rust
    let mut config = embassy_stm32::Config::default();
    config.rcc.sys_ck = Some(mhz(200));
    config.rcc.hclk = Some(mhz(200));
    config.rcc.pll1.q_ck = Some(mhz(100));
 ```

Unfortunately RM0433 is not applicable to all H7s, you would also need to consider RM0455 :/

davidpurser:
@JuliDi  sorry about that!

I do see in RM0455 that the PLL block is a little different than RM0433.

For instance DIVN can only be 8-420, whereas it can be 4-512 on the older H7 chips. I don’t recall seeing code to handle this.

Also the PLL seems to have different VCO range limits in the datasheet.

It looks like a different PLL block than the older H7 chips, and I suspect the PLL block is different enough on that chip that a small feature gate might be needed to handle both. I can take a closer look today though.

JuliDi:
@davidpurser No worries, I know this is a very difficult thing because of how different the chips all are.

If you could have a look at it, that would be great. I have close to no experience with that part of the ST chips... 
Happy to help if I can.

davidpurser:
@JuliDi @xoviat  It looks like the rcc/h7.rs code for PLL setup is heavily based on this code:

https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/pll.rs

... which has since added support for RM0455/RM0468 chips (STM32H7A3, 723/733/725/735, etc), as well as selecting VCO range (see #1298), properly configuring flash wait states for RM0468 chips (see #1092), and fractional PLL dividers via a couple of different strategies.

Looks like the PLL code was originally brought into embassy-rs by @thalesfragoso.  To properly fix these issues we probably need to implement similar solutions here to what stm32h7xx-hal has done.  This isn't really my area of expertise, so I just want to check with everyone for input before I start making larger changes here to try to match stm32h7xx-hal's PLL logic.

xoviat:
I don't really regret merging this because I don't think we would have found these issues without this bump. That said, what's preferred is to have the user provide prescalers rather than compute them in rcc. Since this is old code, that isn't done.

If you're unable to manage that, then you're welcome to go ahead and port what you can from h7 hal. I'll try to look at the next PR more closely, if I have time.

Dirbaio:
I wrote the H5 RCC is in the "new style" of asking the user to provide the prescalers. It should be easy to backport to H7 since they're very similar.

davidpurser:
> I wrote the H5 RCC is in the "new style" of asking the user to provide the prescalers. It should be easy to backport to H7 since they're very similar.

Thanks! I think that makes more sense and is actually something I was thinking about anyway. I will take a look at the H5 RCC for inspiration and get started.

PR #1633, opened 2023-07-10
stm32 impl embedded can error for bus error

schphil:
I'm not sure where is the correct place for the embedded-can in the Cargo.toml. If I'm wrong please let me know.


Dirbaio:
what's the use case for implementing *just* the error, and not the main traits?

schphil:
as far as I know there are no async embedded can traits, if I'm right?  also I'm a bit confused about using the right api, since bxcan for example uses `embedded_hal_02::can;` and not `embedded_can` . bxcan seems not to be frequently maintained atm. 

for my case, I have my own traits `CanTx` and `CanRx` implementing `async fn transmit()` and `async fn receive()` respectively, since my lib uses splitted can instances. i like to implement the `BusError` for `embedded_can::Error` to keep my lib generic. 

any suggestions or thoughts on the first point from your side?

schphil:
this was not thought through to the end. sorry for bringing this up.

PR #1637, opened 2023-07-10
Ensure I2C master_stop() called after error

ShakenCodes:
This is a redo of the previous I2C lockup related PR, but without introducing a timeout mechanism.


Dirbaio (approved):
thanks~!

PR #1638, opened 2023-07-11
stm32: add stop mode

xoviat:


embassy-stm32/build.rs
```diff
@@ -81,6 +81,18 @@ fn main() {
         singletons.push(c.name.to_string());
     }
 
+    // ========
+    // Handle low power.
+
+    let low_power = match env::vars()
```
Dirbaio, line 87: it's probably easier to unconditionally codegen code that has `#[cfg(feature="low_power")]`

embassy-stm32/src/rcc/mod.rs
```diff
@@ -73,6 +75,25 @@ pub struct Clocks {
     pub adc: Option<Hertz>,
 }
 
+#[cfg(feature = "low-power")]
+static mut CLOCK_REFCOUNT: AtomicU32 = AtomicU32::new(0);
```
reitermarkus, line 79: Why `static mut` when it's an atomic?

PR #1639, opened 2023-07-11
rp/gpio: fix is_set_high/is_set_low, expand tests.

Dirbaio:


henrikberg:
If only all fixes was this straight forward! HB

PR #1642, opened 2023-07-12
RP2040 Rtc update with example 

henrikberg:
- RTC example.
- Remove start time from Rtc::new() for more flexibility.
- Not reset RTC in Clock::init().
- Rename to Rtc from RealTimeClock to match STM32 implementation.
- Add missing set_/get_scratchN to watchdog for reboot safe data.
- Add save/restore to RP2040 Rtc.

### DIFF
--- a/embassy-rp/src/watchdog.rs
+++ b/embassy-rp/src/watchdog.rs
@@ -107,4 +107,100 @@ impl Watchdog {
             w.set_trigger(true);
         })
     }
+
+    pub fn set_scratch0(&mut self, value: u32) {
+        let watchdog = pac::WATCHDOG;
+        watchdog.scratch0().write(|w| {
+            *w = value;
+        })
+    }
+
+    pub fn get_scratch0(&mut self) -> u32 {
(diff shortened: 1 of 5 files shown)


embassy-rp/src/watchdog.rs
```diff
@@ -107,4 +107,100 @@ impl Watchdog {
             w.set_trigger(true);
         })
     }
+
+    pub fn set_scratch0(&mut self, value: u32) {
```
Dirbaio, line 111: this should be `get/set_scratch(&mut self, index: usize, ..)` If you could turn that into an array in `rp-pac` too it'd be even more awesome.

Dirbaio:
perhaps the RTC is reset due to this? 
https://github.com/embassy-rs/embassy/blob/864202a23a1fa4362445d8f0b2fcda5003e38131/embassy-rp/src/clocks.rs#L320

In which case it'd be best to not reset it there, instead of adding this save/restore mechanism?

Dirbaio (approved):
LGTM, thank you! :)

PR #1643, opened 2023-07-12
Reset `expires_at` of finished task

bugadani:
Currently, it is possible for finished tasks to remain in a non-zero state (namely, `STATE_TIMER_QUEUED`). This PR resets the `expires_at` field to allow the timer queue to remove the bit, which will in turn allow the task's pool slot to be reused.


Dirbaio (approved):
Thanks!

PR #1648, opened 2023-07-13
Don't require Embassy users to depend on `cortex-m-rt` directly

hannobraun:
When setting up my first Embassy app, I got an error about `cortex-m-rt` from the `main` macro. I figured out quickly what was wrong, but still, requiring users to depend on a library they might otherwise not be using directly seemed unintuitive and unnecessary. So I fixed it :smile: 

Please let me know, if I'm missing something! I'm just getting started with Embassy, so I might be totally on the wrong track here.

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -32,7 +32,7 @@ features = ["nightly", "defmt", "pender-callback", "arch-cortex-m", "executor-th
 # Architecture
 _arch = [] # some arch was picked
 arch-std = ["_arch", "critical-section/std"]
-arch-cortex-m = ["_arch", "dep:cortex-m"]
+arch-cortex-m = ["_arch", "dep:cortex-m", "dep:cortex-m-rt"]
 arch-xtensa = ["_arch"]
 arch-riscv32 = ["_arch"]
 arch-wasm = ["_arch", "dep:wasm-bindgen", "dep:js-sys"]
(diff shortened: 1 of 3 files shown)


Dirbaio:
interesting. 

The downside is that updating to a new major version of  `cortex-m-rt` will require a new major version of the `embassy-executor` crate. (`cortex-m-rt` uses `links` in Cargo.toml to prevent multiple versions to be used at the same time, so it can break projects that do depend on 0.7 directly)

If we do this we should do it for riscv, xtensa as well, for consistency. Major-bumping every time one of the `-rt` crates is major-bumped could be a lot of bumps.

hannobraun:
That's a good point. I didn't consider the other architectures.

> If we do this we should do it for riscv, xtensa as well, for consistency. Major-bumping every time one of the `-rt` crates is major-bumped could be a lot of bumps.

Looking at the releases of the three `-rt` crates, it seems like [`cortex-m-rt`](https://crates.io/crates/cortex-m-rt/versions) is pretty stable, [`xtensa-lx-rt`](https://crates.io/crates/xtensa-lx-rt/versions) releases very frequently, and [`riscv-rt`](https://crates.io/crates/riscv-rt/versions) is somewhere in the middle.

I guess we could sacrifice consistency temporarily for a more convenient and polished experience on Cortex-M, and give the other platforms the same treatment whenever we think those have settled down enough. Not sure if that's the right call, just floating the possibility.

Dirbaio:
Unfortunately I'd prefer to keep requiring the user to bring their own cortex-m-rt version. Thanks anyway for the PR, I appreciate it! :)

hannobraun:
Understood. Thanks for considering!

PR #1649, opened 2023-07-13
rp: Check intrstatus before signalling suspended

simmsb:
I'm not sure if this is also needed elsewhere, but currently on rp2040 the device gets stuck if the usb suspends because Bus::poll will always yield Ready(Suspended)


Dirbaio:
Thanks!

PR #1650, opened 2023-07-13
RP examples gets file description

henrikberg:
- Add descriptions to all RP2040 examples. 
- Some need hardware that was not specified.
- Round ADC temp example to make more sense.


Dirbaio (approved):
Nice, thank you! :)

PR #1653, opened 2023-07-15
stm32/eth: solve busy-loop polling

### LINKED ISSUE #1646: embassy_net::Stack::run thrashes
Calling `stack.run()` thrashes the CPU (including preventing lower-priority tasks from running on rtic).

I added some logging here to test it on my chip - you can look at the diff from main to see what logging I added.

https://github.com/wyager/embassy/tree/wyager-ethernet-thrashing

This function gets run thousands of times a second, starving other stuff on the system: https://github.com/embassy-rs/embassy/blob/main/embassy-net/src/lib.rs#L666

@Dirbaio thought it might be this TODO https://github.com/embassy-rs/embassy/blob/82d765689aa0f922d0b43e402eaa3d911b2d461d/embassy-stm32/src/eth/mod.rs#L84

xoviat:
fixes #1646

### DIFF
--- a/embassy-stm32/src/eth/generic_smi.rs
+++ b/embassy-stm32/src/eth/generic_smi.rs
@@ -1,5 +1,7 @@
 //! Generic SMI Ethernet PHY
 
+use futures::task::Context;
+
 use super::{StationManagement, PHY};
 
 #[allow(dead_code)]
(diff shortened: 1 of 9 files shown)


embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -38,7 +42,22 @@ mod phy_consts {
...
 /// Generic SMI Ethernet PHY
-pub struct GenericSMI;
+pub struct GenericSMI {
+    #[cfg(feature = "time")]
+    poll_interval: Duration,
+}
+
+impl GenericSMI {
+    #[cfg(feature = "time")]
+    pub fn new(poll_interval: Duration) -> Self {
```
Dirbaio, line 52: we should try to make Cargo features additive, so working code doesn't break if you enable `time`.

 Instead of adding/removing the param, could you make `new()` default to e.g. 500ms and then add a `.set_poll_interval(Duration)` fn?

Dirbaio (approved):
Looks great, thanks! :rocket:

PR #1658, opened 2023-07-16
implement most infra for wpan mac

xoviat:
todo: 

- [ ] cleanup control and runner files
- [ ] resolve `msdu_ptr` lifetime issue
- [ ] fix ieee802154 address propagation
- [ ] validate control loop
- [ ] add example


Dirbaio:
The mailbox to the radio is already a queue of buffers, right? In that case, it might be better to make the `embassy-net` push/pull packets from it directly, instead of using `embassy-net-driver-channel`. With the channel, there's 2 queues, one in the channel and another in the mailbox, and there's one "useless" copy between them. It'd be faster/smaller if there was only 1 queue.

PR #1659, opened 2023-07-16
Add usb-hid keyboard example for rp pico.

maximedeboeck:
This PR includes a port of the usb_hid_keyboard.rs example from the nrf52840, with several modifications. I've omitted the hardware VBUS detector implementation as it doesn't seem to be available in embassy-rp.

Also, the remote_wakeup functionality has been disabled due to issues encountered with bus resets following suspensions. This issue led to the disabling of remote_wakeup—though I'm currently unsure if this is an issue specific to the Pico, Windows, or my particular setup.

In its current state, the example waits for a high input signal on pin 16, and then simulates pressing the 'a' key until a low signal is detected. The dependency on usbd-hid in Cargo.toml is necessary for handling KeyBoard reports.

### DIFF
--- a/examples/rp/Cargo.toml
+++ b/examples/rp/Cargo.toml
@@ -40,6 +40,7 @@ display-interface = "0.4.1"
 byte-slice-cast = { version = "1.2.0", default-features = false }
 smart-leds = "0.3.0"
 heapless = "0.7.15"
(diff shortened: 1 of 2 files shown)


Dirbaio:
Thank you!

PR #1665, opened 2023-07-17
stm32/eth: fix cfg(not(time))

xoviat:


embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -45,20 +48,19 @@ use self::phy_consts::*;
 pub struct GenericSMI {
     #[cfg(feature = "time")]
     poll_interval: Duration,
+    #[cfg(not(feature = "time"))]
+    phantom: PhantomData<()>,
```
Dirbaio, line 52: why is this needed? To prevent the user from constructing a GenericSMI with `GenericSMI{}` because it has no fields? 

in that case the PhantomData and making it conditionally compiled is not needed, you could just do `_private: ()`

xoviat: I'd picked up that pattern from stm32f4xx-hal, but I'll start using the pattern you suggested instead.

Dirbaio (approved):

PR #1666, opened 2023-07-17
Embassy-rp I2C: Fix 1664 (async transaction failure)

alexferro:
Change embassy-rp i2c.rs impl of embedded_hal_async::i2c::I2c::transaction to only do the call to setup() for address once per call to transactions. Calling setup multiple times results in I2C transactions being skipped on the bus, even across calls to transaction() or devices.

This fixes the issue I opened, but I do not know if there is an underlying issue in the driver that causes this.

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -716,6 +716,9 @@ mod nightly {
         async fn transaction(&mut self, address: A, operations: &mut [Operation<'_>]) -> Result<(), Self::Error> {
             let addr: u16 = address.into();
 
+            if operations.len() > 0 {
+                Self::setup(addr)?;
+            }
             let mut iterator = operations.iter_mut();
 
             while let Some(op) = iterator.next() {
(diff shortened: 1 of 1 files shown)


Dirbaio:
Thanks for the PR!

Just to confirm: this capture for the fixed code from the issue 
image

corresponds to two `write_read()`s (equivalent two two `transaction([Write, Read])`? Or does it correspond to one `transaction([Write, Read, Write, Read])`? 

If it's the former, it should do "start, write, restart, read, stop. start, write, restart, read, stop", which is what the capture shows.
If it's the latter, it should do "start, write, restart, read, restart, write, restart, read, stop", which is something different.

alexferro:
> corresponds to two `write_read()`s (equivalent two two `transaction([Write, Read])`? Or does it correspond to one `transaction([Write, Read, Write, Read])`?

Yeah, sorry to be ambiguous with that. The trace represents two calls to transaction(). That was clearer in early drafts of my issue that were not as clear in other ways.




> moving the setup call out of the loop is definitely a good thing, no other multi-op method does this. doing it conditionally breaks with the blocking transaction though, perhaps we should skip the check so they differ mostly in additional `.await` sprinkled around?

TBH I didn't even look at the blocking implementation for this change. I added the check because it didn't feel right to me to prep the transaction with a zero element list of operations, especially since I was already chasing an issue with extra calls to that function. 
To some extent, that makes me argue for adding that check to the other implementations. But if that isn't a concern, it could also be removed.

alexferro:
Of course, I'm worrying about a case that probably never happens, but the check feels "safer" to me. To avoid doing anything to the HW unless we know we actually will be.

Dirbaio (approved):
Thanks for the clarification. looks good to me then!

PR #1667, opened 2023-07-18
nrf/pdm: Add continuous sampling API

quentinmit:
I modeled this API off the continuous sampling API provided by the `saadc` module.  The existing one-shot sampling API does not make it possible to stream continuous samples because it throws away at least one sample between each invocation.

It also fleshes out the `Config` struct to allow configuring more of the PDM block's settings.


examples/nrf/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
```
Dirbaio, line 1: accidentally added file

Dirbaio:
unfortunately `pdm.rs` had CRLF endings, and your editor auto-changed it to all LF endings so the diff is not useful (it shows all lines as edited). I've fixed the CRLF stuff in #1669, could you rebase? You might get ugly conflicts, the easy way to fix them is to copy `pdm.rs` out before rebasing and pasting it back in.

examples/nrf/Cargo.toml
quentinmit: Ack, sorry. That was a merge error. Fixed.

quentinmit:
Done, I also ran `cargo fmt` since it looked like I was failing a CI check for that. And I fixed some compatibility problems the CI found with other PACs.

Dirbaio (approved):
Thanks!

PR #1668, opened 2023-07-18
Make dual-stack work in embassy-net

bjoernQ:
While it is possible to configure both, IPv4 and IPv6, it isn't possible to use both at the same time because the IP address was always set as the first element


Dirbaio (approved):
Thanks!

PR #1670, opened 2023-07-18
Fix multicast support

ivmarkov:
The original [multicast PR](https://github.com/embassy-rs/embassy/pull/1264/files#diff-dd4daeb7962fbcb73155aa6ce2683eba0c7182796936378e1e8d4dc22d04b2b5R308) implemented multicast support by requiring the Driver to implement `smoltcp::phy::Device`. 

However, I don't think the "implements `smoltcp::phy::Device`" contract is fulfilled when the driver is based on `embassy-net-driver-channel` (e.g., [the W5500 driver](https://github.com/embassy-rs/embassy/blob/main/embassy-net-w5500/Cargo.toml#L13)), rendering multicast support unavailable on these drivers.

Moreover, the current API is blocking, which is weird. To my understanding, joining/leaving a multicast group involves sending a packet, so these should be async as the sending might fail (queue full)?

Disclaimer: Not sure if I arranged and understood everything correctly though. First contribution to `embassy-net`.

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -479,30 +479,78 @@ impl<D: Driver + 'static> Stack<D> {
 }
 
 #[cfg(feature = "igmp")]
-impl<D: Driver + smoltcp::phy::Device + 'static> Stack<D> {
+impl<D: Driver + 'static> Stack<D> {
     /// Join a multicast group.
-    pub fn join_multicast_group<T>(&self, addr: T) -> Result<bool, MulticastError>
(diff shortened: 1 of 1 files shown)


Dirbaio:
Oh yep, this should've definitely not used `smoltcp::phy::Device` directly. 

The current smoltcp multicast API is somewhat weird, it's the only thing that emits packets directly. That makes it hard to wrap, it should instead be a new socket type or some state in `Interface`, so it takes care of sending the packets for you when it can, instead of you having to retry. Oh well. With the current smoltcp design, LGTM. Thanks!

PR #1673, opened 2023-07-20
rp: disable adc hardware on Adc drop

pennae:
the adc constantly pulls a small but significant amount of current while the hardware is enabled. this can have quite an effect on sleeping devices that also use the adc.


Dirbaio (approved):
LGTM, THX :rocket:

PR #1674, opened 2023-07-20
Rp stack guards

pennae:
fix up an incorrectly placed bit in the stack guard mpu setup, and allow installing stack guards for core0 as well. that'll allow using dedicated sram for the core0 stack without losing overflow protection entirely.

### DIFF
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -219,6 +219,74 @@ select_bootloader! {
     default => BOOT_LOADER_W25Q080
 }
 
+/// Installs a stack guard for the CORE0 stack in MPU region 0.
+/// Will fail if the MPU is already confgigured. This function requires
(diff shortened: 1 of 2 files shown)


Dirbaio:
ah okay I had missed the new `pub fn install_stack_guard` intended to be called from core0.

I'm not sure if it's UB, it requires owning a `&mut Stack` for core0 while core0 is already running from it. It could be used to overwrite arbitrary memory of the stack. (vs for core1 you hand the `&mut Stack` to `embassy-rp` and then no longer own it when core1 is running).

Maybe it makes more sense to do it with magic symbols set by the linker script? `__stack_start`, `__stack_end` for core0, `__core1_stack_start`, `__core1_stack_end` for core1

Dirbaio (approved):
thanks!

PR #1608, opened 2023-07-01
WIP: rp continuous dma transfer and free-running adc

dbdbc:
Hey everyone,
this is an implementation of continuous dma transfer using two dma channels and chaining. The user has to provide the next buffer to write to in time, otherwise the channel will simply stop and be restarted once the buffer is provided. The user will be notified if this happened too late.

Also, free-running sampling of the adc is implemented, with options for round robin, and setting sampling speed. Either a single dma channel with a fixed size buffer or continuous sampling using the continuous dma transfer can be used.

I included two examples.

Since I don't have a lot of experience with rust or embedded, I'd highly appreciate your feedback.

PR #1675, opened 2023-07-21
rp: add dma-from-adc

pennae:
minimal dma-from-adc implementation to get things started that can hopefully be the basis for an update of #1608. we just happen to need dma adc a bit earlier and without continuous sampling support, so here we are :)

### DIFF
--- a/embassy-rp/src/adc.rs
+++ b/embassy-rp/src/adc.rs
@@ -223,28 +223,21 @@ impl interrupt::typelevel::Handler<interrupt::typelevel::ADC_IRQ_FIFO> for Inter
 }
 
 mod sealed {
-    pub trait AdcChannel {
-        fn channel(&mut self) -> u8;
-    }
+    pub trait AdcChannel {}
 }
 
 pub trait AdcChannel: sealed::AdcChannel {}
 pub trait AdcPin: AdcChannel + gpio::Pin {}
 
 macro_rules! impl_pin {
-    ($pin:ident, $channel:expr) => {
-        impl sealed::AdcChannel for peripherals::$pin {
-            fn channel(&mut self) -> u8 {
-                $channel
-            }
-        }
-
+    ($pin:ident) => {
+        impl sealed::AdcChannel for peripherals::$pin {}
         impl AdcChannel for peripherals::$pin {}
         impl AdcPin for peripherals::$pin {}
     };
 }
 
-impl_pin!(PIN_26, 0);
-impl_pin!(PIN_27, 1);
-impl_pin!(PIN_28, 2);
-impl_pin!(PIN_29, 3);
+impl_pin!(PIN_26);
+impl_pin!(PIN_27);
+impl_pin!(PIN_28);
+impl_pin!(PIN_29);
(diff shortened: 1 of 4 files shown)


embassy-rp/src/adc.rs
```diff
@@ -42,24 +46,55 @@ impl<'p> Pin<'p> {
             w.set_pue(pull == Pull::Up);
             w.set_pde(pull == Pull::Down);
         });
-        Self { pin: pin.map_into() }
+        Self(Source::Pin(pin.map_into()))
+    }
+
+    pub fn new_sensor(s: impl Peripheral<P = ADC_TEMP_SENSOR> + 'p) -> Self {
```
Dirbaio, line 52: name should mention "temperature" somehow, just "sensor" doesn't tell what it does.

embassy-rp/src/adc.rs
```diff
@@ -176,19 +198,71 @@ impl<'d> Adc<'d, Async> {
...
+        while !r.cs().read().ready() {}
+        while !r.fcs().read().empty() {
+            r.fifo().read();
         }
-        let result = Self::sample_async(4).await;
-        r.cs().modify(|w| w.set_ts_en(false));
-        result
+
+        // set up fifo for dma
+        r.fcs().write(|w| {
```
Dirbaio, line 222: needs setting `err` to true if S=Sample, false otherwise.

embassy-rp/src/adc.rs
```diff
@@ -42,24 +46,55 @@ impl<'p> Pin<'p> {
...
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(transparent)]
+pub struct Sample(u16);
+
+impl Sample {
+    pub fn good(&self) -> bool {
+        self.0 < 0x8000
```
Dirbaio, line 93: datasheet says the "error" bit is bit 12, not bit 15

embassy-rp/src/adc.rs
```diff
@@ -176,19 +198,71 @@ impl<'d> Adc<'d, Async> {
...
-        Self::sample_async(pin.channel()).await
-    }
-
-    pub async fn read_temperature(&mut self) -> Result<u16, Error> {
+    pub async fn read_many<S: AdcSample>(
+        &mut self,
+        ch: &mut Channel<'_>,
+        buf: &mut [S],
+        dma: impl Peripheral<P = impl dma::Channel>,
+    ) -> S::Result {
```
Dirbaio, line 206: same as in #1715 , changing the return type like this is very surprising/unconventional. Can you change this to always return `Result<(), Error>`?

embassy-rp/src/adc.rs
Dirbaio: hah, I had an old version of the datasheet :joy: 
old:
screenshot-2023-07-31_02-51-32
new:
screenshot-2023-07-31_02-53-33


the bit is indeed 15 (it's the same in both versions)
screenshot-2023-07-31_02-52-13

embassy-rp/src/adc.rs
Dirbaio: > the possible overhead of unwrapping for an error

that overhead is tiny, and the compiler will probably optimize it out anyway.

> we'd prefer to split the method in two

that sounds good too!

Dirbaio (approved):
thanks!

Issue #334, opened 2021-07-31
embassy_stm32 ADC and DAC drivers don't configure pin correctly


bgamari:
To use the ADC and DAC peripherals on external pins, it is expected that the user set the pin's alternative function to `ANALOG`. However, currently neither `embassy_stm32`'s ADC nor DAC drivers do this.

Moreover, it's unclear how the ADC interface as it exists today would efficiently do so given that the it does not take ownership of the pins which it samples.

bgamari:
In the case of the ADC this will require interface faces. I propose that the `AdcPin` `impl`s of the existing GPIO pins are dropped and replaced with (following the example of `gpio::Input`):
```
struct AdcInput<'d, T> {
    pub(create) pin: T,
    phantom: PhantomData<&'d mut T>,
}

// generated
impl<'d, T: AdcPin> AdcInput<'d, T> {
    pub fn new(pin: impl Unborrow<Target = T> + 'd) -> Self {
         unborrow!(pin);

        cortex_m::interrupt::free(|_| unsafe {
            let r = pin.block();
            let n = pin.pin() as usize;
            r.moder().modify(|w| w.set_moder(n, vals::Moder::ANALOG));
        });

        Self {
            pin,
            phantom: PhantomData,
        }
    }
}
```

bobmcwhirter:
Do the pins not start/default to analog?

bobmcwhirter:
We also have `set_as_analog()` on pin to shorten that, which we attempt to do when non-analog peripherals drop their pins, to return them to that mode.

bgamari:
@bobmcwhirter, it turns out that nearly all pins *do* start in analog state.

I am fine with relying on `set_as_analog()` to maintain the invariant that unclaimed pins are always returned to analog state. However, this invariant must be documented. There is currently no way someone unfamiliar with the code could know this.

Dirbaio:
We shouldn't rely on unclaimed pins being in a particular state, we should consider them undefined. Some don't start in analog mode (I think?) and there could also be bootloaders mucking with pins before we run.

`AdcInput` is a "typestate", the only thing it does is assert the pin is in a certain mode. To be consistent with the way digital pin modes are handled, it probably makes more sense to keep `read` taking the unclaimed pin, and having it set Analog mode.

IfitNaoned:
When i run the ADC example on NUCLEO-4L76RG with a potentionmeter like below , the value readed is always the same ( 26 ).

[...]

PR #1676, opened 2023-07-22
stm32: Fix DAC examples

adamgreig:
Currently the DAC examples don't work on F4 or L4, because the DAC driver defaults to enabling the trigger (useful for DMA) but doesn't set the trigger selection, so it's left at the reset value of 0 for `TIM6`. The examples then try to software trigger each new value, but as the DAC isn't configured for software trigger, nothing happens.

One option is to change the trigger selection from `TIM6` to `SOFTWARE`, but as this disables the channel you then have to re-enable it, and since we just trigger immediately after writing each word, it's more sensible to just disable the trigger.

On the H7, in _theory_ the example would work as-is, because the default value for the trigger selection is `SOFTWARE`. However, the PAC is wrong for the H7 and puts the trigger enable bit TEN into the trigger select field instead, so the HAL ends up configuring a different trigger (`TIM1`) instead of enabling trigger mode (which is left disabled), so the software writes go through immediately anyway. Which is good, because the PAC also has the wrong enum for the triggers, so trying to set it to `SOFTWARE` would actually set it to `TIM8`. I think it's best to update the example to match the F4/L4, but at least it still works, and then fix the PAC/HAL separately.

I also updated the DAC driver to set the pin to analogue mode as recommended in the reference manual. The DAC _works_ without this, but would contend with the digital driver if the pin was in output mode, or needlessly trigger the input Schmitt trigger in default input mode, so it's better to set it to analog. I'm not sure if the added constraint on the pin type is the best way to do this though. This issue is covered in #334, but this PR doesn't change the ADC behaviour (maybe it's already fixed, though?).

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -213,8 +213,9 @@ impl<'d, T: Instance, Tx> DacCh1<'d, T, Tx> {
     pub fn new(
         peri: impl Peripheral<P = T> + 'd,
         dma: impl Peripheral<P = Tx> + 'd,
-        _pin: impl Peripheral<P = impl DacPin<T, 1>> + 'd,
+        pin: impl Peripheral<P = impl DacPin<T, 1>> + crate::gpio::sealed::Pin + 'd,
     ) -> Self {
+        pin.set_as_analog();
         into_ref!(peri, dma);
         T::enable();
         T::reset();
@@ -324,8 +325,9 @@ impl<'d, T: Instance, Tx> DacCh2<'d, T, Tx> {
     pub fn new(
         _peri: impl Peripheral<P = T> + 'd,
         dma: impl Peripheral<P = Tx> + 'd,
-        _pin: impl Peripheral<P = impl DacPin<T, 2>> + 'd,
+        pin: impl Peripheral<P = impl DacPin<T, 2>> + crate::gpio::sealed::Pin + 'd,
     ) -> Self {
+        pin.set_as_analog();
         into_ref!(_peri, dma);
         T::enable();
         T::reset();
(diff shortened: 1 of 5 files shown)


JuliDi:
@adamgreig thanks for spotting these mistakes in my changes to the implementation and the examples!

Also the H7 not working with DMA was giving me headaches and at the end I just excluded the H7s DMA/BDMA/DMAMUX mess from the implementation. Apparently I didn't search for the error in the right places after all...

So if we could make it work with the H7 now (especially with circular DMA) that would be fantastic for my current project. Happy to do some more work/testing on this if necessary and update the examples accordingly.

adamgreig:
If you're up for fixing the stm32-data issue (https://github.com/embassy-rs/stm32-data/issues/222) that would be great as I've not got time to look into it at all, and I think with that fixed everything in embassy-stm32 would probably be a lot smoother!

Besides that maybe a HIL test that outputs some voltages on the DAC and reads them with the ADC would be a nice way to check everything keeps working. I haven't looked at the HIL tests before so don't know what's involved there, though.

JuliDi:
Sure, I will look into that issue. Totally makes sense fixing it at the origin.

Regarding HIL I have no idea either, but it would be nice to habe. Maybe @Dirbaio can tell us whether that would work.

JuliDi:
Is there anything missing in this PR that prevents a merge?

Dirbaio (approved):
Thank you!

Dirbaio:
> Regarding HIL I have no idea either, but it would be nice to habe. Maybe @Dirbaio can tell us whether that would work.

Yeah, it's definitely doable. It'd need just one wire from a DAC pin to an ADC pin. It you want to work on it please let me know. I can give you access to teleprobe so you can run tests locally.

Issue #1604, opened 2023-07-01
Feature Request: DMA/BDMA/RingBuf: read_exact (or similar)


alexferro:
I'm working on a project with circular buffering, but not using the USART (it's audio) and the DMA RingBuf implementation is a fairly decent fit to my needs, but I'd rather work with blocks of data, rather than sample by sample. Thanks to the wake behavior of the DMA interrupt, if I successfully get into the Pending state, the next time my task gets woken is when I will be able to read a full block of data. However, given data processing time, by the time I am back to waiting on the DMA, I've usuaulyl gotten in at least one sample of data, and as far as I can tell, there's no longer a way (after #1428 was merged) to determine if read with a given size will succeed or not without trying it and getting a partial read. This means I have a slightly ugly loop through the future to do a partial read and then enter pending until I know that my actual data is all ready.

However, read_exact, has a few pitfalls that read does not. Specifically it is likely to misbehave (I don't think unsafely) is such a way that you could effectively end up hanging or overruning if the buffer to read_exact isn't a pure fraction of the internal ring buffer size.

I've included my working (this has been modified slightly to be standalone, not part of my project, so it's missing type annotations properly, since they came from elsewhere in my codebase) code that does this as part of an async fn.  This was pieced together from the usart_ringbuf driver. 

For me, this is not a high priority, as I have this workaround, even if I feel it's a bit ugly. I've quite enjoyed using embassy otherwise, and I was expecting this portion to be ugly since I saw #702 and agree that the general case is hard to make safe.

```rust
pub struct MyDmaRingBuffer<DMA_CH, W> {
    ring: RingBuffer<'static, DMA_CH, W>,
}

impl MyDmaRingBuffer {
    pub async fn read_exact(&mut self, buffer: &mut [W]) {
        use core::future::poll_fn;
        use core::sync::atomic::{compiler_fence, Ordering};
        use core::task::Poll;

        let mut read_data = 0;
        let buffer_len = buffer.len();

        let dma = poll_fn(|cx| {
            self.ring.set_waker(cx.waker());

            compiler_fence(Ordering::SeqCst);

            if let Ok((len, _remaining)) = self.ring.read(&mut buffer[read_data..buffer_len]) {
                read_data += len;
                if read_data == buffer_len {
                    Poll::Ready(())
                } else {
                    Poll::Pending
                }
            } else {
                defmt::panic!("SAI DMA RingBuffer Read Overrun");
            }
        });
        dma.await
    }
}
```

[...]

PR #1681, opened 2023-07-23
Add a STM32/DMARingBuffer::read_exact helper

alexferro:
Add a STM32/DMARingBuffer::read_exact helper

This provides a helper function with an async implementation, that
will only return (or error) when it was able to read that many bytes,
sleeping until ready.

Additionally, corrected the documentation for Ringbuffer functions to use
"elements" instead of "bytes" as the types were already generic over the
word/element size.

This is in reference to #1604.

Adding this helper function is right on the edge of what felt acceptable to me to duplicate, rather than refactor. I'm fine with this being merged as is, but it feels like there's a better way and I'm just not experienced enough to know what is best, especially in a project other people develop and use.
<details>
<summary>Implementation Choices</summary>

I started down the path of refactoring DMA to unify the code, but ran into a number of paths.
Implementation options I can see:
- Duplicate code: Simple to understand and reason about. Bad for maintenance. (Chose this, to open a discussion)
- Add a public DmaRingBuf Trait providing the public API of DmaRingBuf, and implement it by both DMA and BDMA. Additionally have read_exact an async function that is defaulted by the trait: No code duplication. All existing users, both in and out of tree, must bring the trait into scope, which feels odd when most chips only have one available anyway. Required nightly and async_trait.
- DmaRingBuf Trait and DmaRingBufEx with the nightly cfg flag enabled: See above, but this avoids the nightly issue, just even more traits needed.
- Add an Adapter that turns a DmaRingBuf into an DmaRingBufExact (via a method `into_read_exact(self)` or similar). Requires a trait still to use in the bounds of the generic for the helper object: This one feels the most discoverable, and does not require nightly for async_trait, but still has the downsides of turning the API of DmaRingBuffer into a trait.
- Macros?: I know there's macros used for some of this code duplication elsewhere in embassy, but this feels like a bit of a hack, and I'm not sue I know enough to get this right.

I suppose it could also be just the bare minimum of `set_waker` and `read` needed by the trait, and those could be duplicated inside the trait impl, but that seems ugly and confusing ("why do these functions exist twice, which one should I use?")

</details>

I also changed the word choice (after I noticed I was subconsciously doing it too) in the comments for the ringbuffer to substitute 'elements' for 'bytes' as the DMA types are generic over byte/half/word. While the trait bound used for the buffer element is 'Word' I chose to avoid it due to the overloaded meaning of 'word' outside of the context of the DMA peripheral/driver.

<details>
<summary>Why read_exact is a "async fn" not a regular function returning impl Future</summary>
I had originally written this PR in such a way that it was not using an async function, just returning a Future. See this code block:

```rust
pub fn read_exact<'a>(
    &'a mut self,
    buffer: &'a mut [W],
) -> impl core::future::Future<Output = Result<(), OverrunError>> + 'a {
    use core::future::poll_fn;
    use core::sync::atomic::{compiler_fence, Ordering};
    use core::task::Poll;

    let mut read_data = 0;
    let buffer_len = buffer.len();

    poll_fn(move |cx| {
        self.set_waker(cx.waker());

        match self.read(&mut buffer[read_data..buffer_len]) {
            Ok((len, _remaining)) => {
                read_data += len;
                if read_data == buffer_len {
                    Poll::Ready(Ok(()))
                } else {
                    Poll::Pending
                }
            }
            Err(e) => Poll::Ready(Err(e)),
        }
    })
}
```
However, while that had worked in my original prototyped testing (in my project I had a wrapper type), when I added it directly to RingBuffer, I ran into several lifetime issues that did not make sense to me, and even as I fixed them locally, then they just extended into the point of use. At least at my current understanding of them, this was a non-starter. I think I was ending up with telling the compiler that the buffer had to outlive the DMA or something.
However, as I went back to my original async fn I had written for my project, I found it worked properly, without any lifetime issues, probably as the created future is definitely consumed internally.

</details>

### DIFF
--- a/embassy-stm32/src/dma/bdma.rs
+++ b/embassy-stm32/src/dma/bdma.rs
@@ -466,13 +466,51 @@ impl<'a, C: Channel, W: Word> RingBuffer<'a, C, W> {
         self.ringbuf.clear(DmaCtrlImpl(self.channel.reborrow()));
     }
 
-    /// Read bytes from the ring buffer
+    /// Read elements from the ring buffer
     /// Return a tuple of the length read and the length remaining in the buffer
-    /// If not all of the bytes were read, then there will be some bytes in the buffer remaining
-    /// The length remaining is the capacity, ring_buf.len(), less the bytes remaining after the read
+    /// If not all of the elements were read, then there will be some elements in the buffer remaining
+    /// The length remaining is the capacity, ring_buf.len(), less the elements remaining after the read
     /// OverrunError is returned if the portion to be read was overwritten by the DMA controller.
     pub fn read(&mut self, buf: &mut [W]) -> Result<(usize, usize), OverrunError> {
         self.ringbuf.read(DmaCtrlImpl(self.channel.reborrow()), buf)
+    }
+
+    /// Read an exact number of elements from the ringbuffer.
+    ///
+    /// Returns the remaining number of elements available for immediate reading.
+    /// OverrunError is returned if the portion to be read was overwritten by the DMA controller.
+    ///
+    /// Async/Wake Behavior:
+    /// The underlying DMA peripheral only can wake us when its buffer pointer has reached the halfway point,
+    /// and when it wraps around. This means that when called with a buffer of length 'M', when this
+    /// ring buffer was created with a buffer of size 'N':
+    /// - If M equals N/2 or N/2 divides evenly into M, this function will return every N/2 elements read on the DMA source.
+    /// - Otherwise, this function may need up to N/2 extra elements to arrive before returning.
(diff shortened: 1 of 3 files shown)


xoviat (approved):
LGTM.

xoviat:
With respect to the implementation choices, I poked this bear and wasted a bunch of time. This is honestly the best implementation to put in at this point, since a major refactor would probably get bogged down.

> but it feels like there's a better way and I'm just not experienced enough to know what is best, especially in a project other people develop and use.

There is a better way, but it's not really your problem at this point, as it would require a refactor. So don't worry about it.

alexferro:
I guess it's good to know my sense of "this is getting ugly" is starting to develop for Rust. 
Thanks!

PR #1684, opened 2023-07-23
stm32/rcc: move rcc logic from ipcc

xoviat:
This moves the bulk of the rcc configuration out of the ipcc module, and also frontloads clock computation.


xoviat:
This time, the HIL failure is not my fault.

PR #1685, opened 2023-07-24
stm32/can: bxcan async enable

chemicstry:
This PR implements async enable for bxcan.

Using the default enable from bxcan library can lead to infinite loops if there is a hardware problem (i.e. canbus shorted to VCC, or faulty transceiver).

Unfortunatelly, bxcan seems to only have interrupt for entering sleep mode, but not for leaving (SLAK = 0) so I used `yield_now()`.


Dirbaio (approved):
Thanks!

PR #1687, opened 2023-07-25
stm32/can: implement proper RX timestamps

chemicstry:
This PR implements bxcan receive timestamps based on `embassy_time`.

The previous implementation used `u16` time register from bxcan peripheral, which is part of TT-CAN (time triggered CAN) feature set and is broken on STM32 chips (See [ STM32F4 Errata 2.13.1](https://www.st.com/resource/en/errata_sheet/es0206-stm32f427437-and-stm32f429439-device-errata-stmicroelectronics.pdf)).

I also implemented loopback latency measurement in tests, which should allow catching any performance regressions.


xoviat:
Why is the u16 timestamp not sufficient? Does it not reflect the current time, or what is the problem?

chemicstry:
Internal bxcan timestamp is not tied to any timer inside STM32 and datasheet has zero documentation about it. So even the timer frequency is unknown. It is just some free running timer inside the peripheral, which values have no absolute meaning. Not to mention that errata says to not use it, so it probably has hadrware bugs.

I do not see any uses for such timer, but please let me know if there is some application note or something that I missed.

xoviat:
The bxcan internal timer is not subject to irq delay, and does not require the time feature, which is why I wanted to use it. The frequency of the timer can be determined by testing it; I will try to write a test to determine the frequency and adjust the timestamp. 

You may be right that it doesn't work, but the tx functionality is not used and rx timestamp still may work. I would like to try to get it working first before we reject it.

chemicstry:
I guess it's possible to reverse engineer how it works, but then there are other issues like synchronizing it with other timers (Instant) if you want to get absolute timestamp. Then you have to figure out when the timer is actually started, if it ever stops and how it differs between chips. Maybe some other chips that use bxcan peripheral have more information, dunno if such exist.

I've used this timestamp in IRQ method with UAVCAN v0 and it was precise enough for +-1us clock synchronization over CAN, but using bxcan timestamp could be better, depending on the clock frequency. Let me know how it goes.

timokroeger:
I, too, tried to use the bxcan timestamp and can confirm everything @chemicstry wrote. Biggest issue is that it's free running, overflows often und there is no way to sync it.

xoviat:
The timer ticks at the CAN bitrate, overflows at u16::MAX, and starts after enable is called. By recording the instant at enable and the bitrate, irq delays as long as 65ms at a 1mhz bitrate can be compensated for. However, I think it's probably reasonable to make timestamps dependent on the time feature and set them to zero if time is not enabled. 

In order to synchronize the timer, it would be better to make enable a blocking method so the the instant can be collected without an interrupt delay. I am skeptical that can initialization needs to be an async method.

If all of this sounds good to you, I can try to implement these changes on top of this PR and you can see if they work. If it slows down the irq handler too much from the extra math, it might not be worth it.

chemicstry:
The problem with synchronous enable is that a hadrware/wiring failure can cause an infinite loop and freeze the entire system. Maybe we could do something similar to I2C driver, where blocking methods have a timeout.

Otherwise, if timers can be synchronized properly I'm all in for using the bxcan timer.

Dirbaio:
what's the status on this/ I haven't been following the discussions, is this the approach we want to follow?

Dirbaio:
LGTM, can you rebase?

chemicstry:
Updated.

For future reference, the problem with #1698 was that frequency difference between `embassy-time` and bxcan timer was fractional and accumulated error over time. I still think it is possible to implement if you rearrange the math, but for the time being this PR should do. The API should be compatible if bxcan timer is implemented in the future, so no breaking changes will be neccessary.

chemicstry:
Seems that the test timing requirements are too tight. I will take a look

Dirbaio (approved):
thank you!

PR #1688, opened 2023-07-25
DAC ADC Test STM32H7

JuliDi:
Some boards have a pin that is both DAC output pin and ADC input pin. So starting the DAC on this pin and then stealing the peripheral again to initialise the ADC on the same pin, we can measure the ADC output.

Discussed in #1676


xoviat:
I would patch metapac to a rev, if possible.

JuliDi:
> I would patch metapac to a rev, if possible.

done

examples/stm32h7/src/bin/dac_dma.rs
```diff
@@ -0,0 +1,140 @@
...
+    dac.select_trigger(embassy_stm32::dac::Ch1Trigger::Tim6).unwrap();
+    dac.enable_channel().unwrap();
+
+    TIM6::enable();
+    TIM6::regs().arr().modify(|w| w.set_arr(reload as u16 - 1));
+    TIM6::regs().cr2().modify(|w| w.set_mms(Mms::UPDATE));
+    TIM6::regs().cr1().modify(|w| {
+        w.set_opm(Opm::DISABLED);
+        w.set_cen(true);
+    });
```
adamgreig, line 63: TIM6 and TIM7 are supported by the `Basic16bitInstance` HAL driver, so you can skip all the frequency and register access and instead just:

```rust
    TIM7::enable();
    tim7.set_frequency(Hertz(8000));
    TIM7::regs().cr2().write(|r| r.set_mms(timer::vals::Mms::UPDATE));
    tim7.start();
```

JuliDi: Why did I not know? What a mess. Thanks for the hint.

Then I guess it's best if I close this PR and do two separate PRs, one for the HIL test and one for the DAC DMA examples cleanup...

JuliDi: For some reason, this does work quite well on the STM32H7 but not on the L4...

JuliDi: Feels like the implementation with the set_frequency is somehow different. If I call `TIM6::frequency()` I will always get the same value, no matter what frequency I've set with `set_frequency`. Probably really best to separate these two PRs.

JuliDi:
Closing in favor of #1697.

PR #1690, opened 2023-07-25
rp: Fix ROM cache ptr() returning the trampoline

Sizurka:
When #1384 changed the ROM cache optimization it unfortunately introduced some subtle breakage.  The flash functions assume that a call to `::ptr()` returns the actual ROM address (since they need to operate without any calls into XIP), but after the optimization the first call will return the trampoline.  The trampoline itself is (likely) located in flash, so when the flash operation functions try to use it with XIP disabled, they immediately hard fault.

So this just separates out the `::ptr()` function into a version that uses the full optimization for the direct calls and one that uses a slightly less efficient version but will always return a ROM pointer.


Dirbaio:
thanks for the fix! :rocket:

PR #1692, opened 2023-07-26
[DRAFT] stm32/can: fdcan support

davidpurser:
This is currently a WIP / draft.

STM32 FDCAN peripheral support implemented as an async wrapper around the `fdcan` crate.  `fdcan` implements quite a bit of logic and types that I don't feel it makes sense to duplicate.  Therefore `fdcan` library is used for all functions except interrupt handler.

Currently using `fdcan::pac` for register access -- partly for consistency with `fdcan` create and partly because `stm32-metapac` has various issues with FDCAN:
- FDCAN interrupt register bits (IE/IR registers) differ between G0/G4/L5 chips and H7 chips
- stm32-metapac `TRP` register definition only has bits [7:0], while RM0433 defines bits [31:0] (have not checked if this is an H7-specific issue)
- On H7, FDCAN RCC bits are shared between FDCAN1 and FDCAN2.  stm32-metapac has them listed under FDCAN1

While embassy's `bxcan` implementation contains an internal 32-frame queue, I have opted not to implement that for `fdcan` for a couple of reasons:
- `fdcan` peripheral has its own FIFO storage which can contain a few messages
- 32 CAN-FD frames is quite a bit of storage :)
- implementing an internal queue may defeat some prioritization/ordering features (e.g. if priority messages are sent to fifo0 vs fifo1)
- as with other peripherals, any additional queueing is easily implemented on the application side with a dedicated task, which gives the application flexibility as far as queue length, type, and storage location

TODO:
- [ ] fix metapac issues for FDCAN, and in particular FDCAN on H7 chips
- [ ] revert dependency change in Cargo.toml (stop getting metapac from `stm32-data-generated` git once it is fixed and released)
- [ ] implement receive from rx fifo 1
- [ ] helpers for clocks and bit timing?
- [ ] error handling and error interrupts
- [ ] implement all valid mode transitions from `fdcan` crate
- [ ] remove defmt debug messages
- [ ] simplify some interfaces

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -57,9 +57,11 @@ sdio-host = "0.5.0"
 embedded-sdmmc = { git = "https://github.com/embassy-rs/embedded-sdmmc-rs", rev = "a4f293d3a6f72158385f79c98634cb8a14d0d2fc", optional = true }
 critical-section = "1.1"
 atomic-polyfill = "1.0.1"
-stm32-metapac = "12"
+#stm32-metapac = "12"
+stm32-metapac = { git = "https://github.com/embassy-rs/stm32-data-generated", rev = "ee7f153b2bc99db1f3f2540c2950bca5f1fcf1ba" }
 vcell = "0.1.3"
 bxcan = "0.7.0"
+fdcan = { version = "0.2.0", features = ["fdcan_h7"]}
 nb = "1.0.0"
 stm32-fmc = "0.2.4"
 seq-macro = "0.3.0"
@@ -74,7 +76,8 @@ critical-section = { version = "1.1", features = ["std"] }
 [build-dependencies]
 proc-macro2 = "1.0.36"
 quote = "1.0.15"
-stm32-metapac = { version = "12", default-features = false, features = ["metadata"]}
+#stm32-metapac = { version = "12", default-features = false, features = ["metadata"] }
+stm32-metapac = { git = "https://github.com/embassy-rs/stm32-data-generated", rev = "ee7f153b2bc99db1f3f2540c2950bca5f1fcf1ba", default-features = false, features = ["metadata"] }
 
 [features]
 default = ["rt"]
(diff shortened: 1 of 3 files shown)


xoviat:
I'm not inclined to reject any relevant improvements to embassy-stm32. That said, the queue will probably be brought back in subsequent work (not necessarily yours) because of the same reasons it was put there in the first place: the hardware queue is not sufficient at a high bitrate.

davidpurser:
@xoviat  Fair enough. I can put the queue back in, there are only a few lines of code needed for it. The H7 buffers can be configured to hold 32 messages in the hardware buffer for each Rx fifo, but I think some other STM series are more limited so you may be right that the software buffer is needed.

I’m considering allowing the user to choose a queue size (and maybe type) via type generic parameter on the Fdcan. That would remove the fixed 32 size and allow the user some flexibility. This seems like a potentially worthwhile improvement to the bxcan design as well.

Obviously I still have some work to do on the implementation, this is more or less a first pass at it, but I wanted to get the PR open for feedback. Thanks for taking a look.

kevswims:
@davidpurser have you made any further progress on this? We need to get CAN-FD support added to the G4 series so we can help out with getting this in.

tcbennun:
> fdcan: add H7 support; fix generation and registers for G0, G4, H5, L5, U5 embassy-rs/stm32-data#308

I'm working with FDCAN on H5 (whose FDCAN looks the same as G4).

For the PAC I have a PR here: https://github.com/embassy-rs/stm32-data/pull/308

I haven't got too far with the embassy-stm32 driver but I'm starting to work on it. This PR only supports H7 which is very different!

blaa:
I gathered all noteworthy attempts at FDCAN integration I could find here in the issue:
https://github.com/embassy-rs/embassy/issues/2447
I'll leave the link here so that the current progress is easily discoverable.

plaes:
@Dirbaio Seems like now that #2447 and #2571 have landed, this PR is obsolete...

cc: @blaa

PR #1694, opened 2023-07-27
Use lora-phy v1.2.1; modify embassy-lora dependencies

ceekdee:
Use lora-phy v1.2.1 rather than a patch.

Remove unnecessary dependencies for embassy-lora.

Place the embassy-lora dependency on rust-lorawan behind the time feature so that users of embassy-lora have the option to not pull in rust-lorawan.

### DIFF
--- a/embassy-lora/Cargo.toml
+++ b/embassy-lora/Cargo.toml
@@ -12,7 +12,7 @@ target = "thumbv7em-none-eabi"
 
 [features]
 stm32wl = ["dep:embassy-stm32"]
-time = []
+time = ["dep:embassy-time", "dep:lorawan-device"]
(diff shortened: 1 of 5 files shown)


ceekdee:
Even though all checks now pass, I am concerned about the following:

time = ["embassy-time", "lorawan-device"]
defmt = ["dep:defmt", "lorawan-device/defmt"]

If the time feature is not enabled, eliminating lorawan-device, defmt still includes "lorawan-device/defmt".  Is this an issue?

Dirbaio (approved):
thanks for the release! :)

ceekdee:
PRs 1693 and 1694 both made changes to the same Cargo.toml files, and PR 1694 lost in the merge queue.  I resolved the conflicts and placed PR 1694 back on the merge queue.

PR #1698, opened 2023-07-28
stm32/can: add timestamps based on time

xoviat:
@chemicstry This implementation seems to work, but the downside is that it requires 64-bit math and requires the user to use the set_bitrate method. Let me know what you think.


chemicstry:
Thanks for working on this. I haven't checked the math yet (need more coffee for that), but some initial thoughts:
- Should `enable` be in a critical section? IRQ firing in the middle of it could throw the timer offset way off. On the other hand, the critical section would be quite long so timeout value should be as short as possible to not interfere with other things.
- IIRC from I2C timeout PR there was a consensus that `time` *must* be an optional feature. That's why in my PR I used `Envelope` struct, where you could remove timestamp field when `time` is not enabled.

Dirbaio:
the general policy is `embassy-time` is optional, but if the user doesn't enable it's fine for them to simply don't get the feature. In other words, we don't go out of our way to provide alternatives for the user to "plug" their own time.

xoviat:
Yes to both points, which is why this is currently draft. I wanted you to check though whether it actually works for your use case. I've run a few simple tests but was hoping you could run more and double check the math.

xoviat:
@chemicstry With this updated PR, I received this output:


```
INFO - 0.000030 Hello World!
INFO - 0.000152 Configuring can...
INFO - 0.010375 Can configured
INFO - 0.020538 Transmitting frame...
INFO - 0.020690 now (tx): 676
INFO - 0.020935 Receiving frame...
TRACE - 0.021118 now: 690
TRACE - 0.021331 now (time): 685
...
INFO - 0.394561 loopback latency 0
INFO - 0.434814 Transmitting frame...
INFO - 0.434967 now (tx): 14252
INFO - 0.435241 Receiving frame...
TRACE - 0.435424 now: 14266
TRACE - 0.435668 now (time): 14250
INFO - 0.436004 loopback latency 0
INFO - 0.476257 Test OK
```

I think because the hardware timer is triggered on the rising edge of the first bit, the theoretical latency may be zero.

PR #1699, opened 2023-07-28
RP2040: PIO UART example

mvniekerk:


examples/rp/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 embassy-embedded-hal = { version = "0.1.0", path = "../../embassy-embedded-hal", features = ["defmt"] }
+embassy-hal-common = { version = "0.1.0", path = "../../embassy-hal-common", features = ["defmt"] }
```
Dirbaio, line 10: `embassy-hal-common` is for internal use only, no crate should depend on it except embassy hals.

mvniekerk: Removed

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+use embassy_usb::driver::EndpointError;
+use embassy_usb::{Builder, Config};
+use embedded_io::asynch::{Read, Write};
+use {defmt_rtt as _, panic_probe as _};
+
+use crate::uart::PioUart;
+use crate::uart_rx::PioUartRx;
+use crate::uart_tx::PioUartTx;
+
+bind_interrupts!(struct UsbIrqs {
```
Dirbaio, line 33: you can bind multiple interrupts in a single call, then use `Irqs` for both.

```rust
    USBCTRL_IRQ => InterruptHandler<USB>;
    PIO0_IRQ_0 => PioInterruptHandler<PIO0>;
});
```

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+
+    // Run the USB device.
+    let usb_fut = usb.run();
+
+    // PIO UART setup
+    let uart = PioUart::new(9600, p.PIO0, p.PIN_4, p.PIN_5).await;
+    let (mut uart_tx, mut uart_rx) = uart.split();
+
+    // Channels setup
+    static USB_CHANNEL_TX: Channel<ThreadModeRawMutex, u8, 20> = Channel::<ThreadModeRawMutex, u8, 20>::new();
```
Dirbaio, line 97: - `Pipe` would be more appropriate here, it allows enqueuing/dequeuing many bytes at a time, not just one-by-one so it might be a bit faster. It implements the `embedded-io` traits too.
- Since you're not sharing these across tasks, you can put them in local variables instead of `static`s, which allows using `NoopRawMutex` which is slightly faster.

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+        trace!("USB OUT: {:x}", data);
+        usb_tx.write_packet(&data).await?;
+    }
+}
+
+/// Read from the UART and write it to the USB TX send channel
+async fn uart_read<'a>(
+    uart_rx: &mut PioUartRx<'a>,
+    usb_tx_send: &mut embassy_sync::channel::Sender<'a, ThreadModeRawMutex, u8, 20>,
+) -> Result<(), Disconnected> {
```
Dirbaio, line 182: if `uart_read`/`uart_write` never return, it'd be more idiomatic to make them `-> !`

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+    use crate::uart_tx::PioUartTx;
+    use crate::PioIrqs;
+
+    pub struct PioUart<'a> {
+        tx: PioUartTx<'a>,
+        rx: PioUartRx<'a>,
+    }
+
+    impl<'a> PioUart<'a> {
+        pub async fn new(
```
Dirbaio, line 225: does not need to be async

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+                usb_write(&mut usb_tx, &mut usb_channel_tx_recv),
+            )
+            .await;
+            info!("Disconnected");
+        }
+    };
+
+    // Read + write from UART
+    let uart_future = async {
+        loop {
```
Dirbaio, line 124: loop is not needed if if uart_read/uart_write never return

examples/rp/src/bin/pio_uart.rs
```diff
@@ -0,0 +1,431 @@
...
+            (Self { sm_tx }, offset)
+        }
+
+        pub async fn write_u8(&mut self, data: u8) {
+            self.sm_tx.tx().wait_push(data as u32).await;
+        }
+    }
+
+    impl Io for PioUartTx<'_> {
+        type Error = PioUartError;
```
Dirbaio, line 331: if reading/writing never fails, the convention is to do `type Error = core::convert::Infallible`. It's an uninhabited enum (0 variants) which means it's impossible to create a value of it.

Dirbaio (commented):
Thanks for the PR! Love how complete the uart implementation is :D

I've commented a few minor things to make the example more idiomatic.

examples/rp/src/bin/pio_uart.rs
mvniekerk: Fixed

examples/rp/src/bin/pio_uart.rs
mvniekerk: * Used Pipe
* Will be fighting the borrow checker. The uart pipe will be dropped when usb is dropped (that will drop the reader). Same goes usb pipe vs uart. Kept as static?

examples/rp/src/bin/pio_uart.rs
mvniekerk: Fixed

examples/rp/src/bin/pio_uart.rs
mvniekerk: Fixed

examples/rp/src/bin/pio_uart.rs
mvniekerk: Fixed

examples/rp/src/bin/pio_uart.rs
mvniekerk: Fixed

examples/rp/src/bin/pio_uart.rs
Dirbaio: Fixed it for you :) the issue is you were tying the pipe lifetimes to the usb in the functions. using `'_` works.

Dirbaio (approved):

examples/rp/src/bin/pio_uart.rs
mvniekerk: Nice, thank you!

PR #1702, opened 2023-07-28
Allow ethernet and 802.15.4 to coexist

rubdos:
/cc @xoviat 

I had some duplicate work lying around with @xoviat's #1655 (and I still have some more), and this is part of the result of some rebase conflicts. @xoviat, to what extent are you planning to add sixlowpan to Embassy?


Dirbaio:
is the issue a build failure when both `medium-ethernet` and `medium-ieee802154` features are enabled? If so, could you add a check to `ci.sh` so it doesn't break again?

rubdos:
Yes indeed. It's not that we needed this, but we had that code lying around in this form...

Dirbaio (approved):
thanks!

xoviat:
@rubdos it's partially implemented but a full implementation is whenever I get around to it.

PR #1704, opened 2023-07-28
Expose IEEE802.15.4 address in Driver

rubdos:
(I'm sorry for the PR-spam. I'm trying to get the "finished" things we have cleaned up, atomic and mergeable.)

This allows a net::Driver to be an IEEE802.15.4 device.

I wrote this in two stages.

1. Added `ieee802154_address()` to the trait
2. Merged `ethernet_address()` and `ieee802154_address` into `hardware_address()`.

This allows the community (mainly @Dirbaio, cc @xoviat) to decide which way is better. HardwareAddress means we have a bunch of match-statements at runtime, for a property that's typically known at compile time (but that's the case anyway with smoltcp, currently), but it's a lot more compact and future proof. It's a (big-ish) breaking change, however.

CI will probably complain, but I'll clean that up after a decision on the above is taken :-)

### DIFF
--- a/embassy-net-driver/Cargo.toml
+++ b/embassy-net-driver/Cargo.toml
@@ -22,4 +22,3 @@ features = ["defmt"]
 
 [dependencies]
 defmt = { version = "0.3", optional = true }
-smoltcp = { version = "0.10", default-features = false }
--- a/embassy-net-driver/Cargo.toml
+++ b/embassy-net-driver/Cargo.toml
@@ -22,3 +22,4 @@ features = ["defmt"]
 
 [dependencies]
 defmt = { version = "0.3", optional = true }
+smoltcp = { version = "0.10", default-features = false }
--- a/embassy-net-driver/src/lib.rs
+++ b/embassy-net-driver/src/lib.rs
@@ -4,7 +4,15 @@
 
 use core::task::Context;
 
-use smoltcp::wire::HardwareAddress;
+/// Representation of an hardware address, such as an Ethernet address or an IEEE802.15.4 address.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum HardwareAddress {
+    /// A six-octet Ethernet address
+    Ethernet([u8; 6]),
+    /// An eight-octet IEEE802.15.4 address
+    Ieee802154([u8; 8]),
+}
 
 /// Main `embassy-net` driver API.
 ///
--- a/embassy-net-driver/src/lib.rs
+++ b/embassy-net-driver/src/lib.rs
@@ -53,6 +53,9 @@ pub trait Driver {
 
     /// Get the device's Ethernet address.
     fn ethernet_address(&self) -> [u8; 6];
+
+    /// Get the device's IEEE 802.15.4 address.
+    fn ieee802154_address(&self) -> [u8; 8];
 }
 
 impl<T: ?Sized + Driver> Driver for &mut T {
@@ -77,6 +80,9 @@ impl<T: ?Sized + Driver> Driver for &mut T {
     }
     fn ethernet_address(&self) -> [u8; 6] {
         T::ethernet_address(self)
+    }
+    fn ieee802154_address(&self) -> [u8; 8] {
+        T::ieee802154_address(self)
     }
 }
 
--- a/embassy-net-driver/src/lib.rs
+++ b/embassy-net-driver/src/lib.rs
@@ -12,6 +12,8 @@ pub enum HardwareAddress {
     Ethernet([u8; 6]),
     /// An eight-octet IEEE802.15.4 address
     Ieee802154([u8; 8]),
+    /// Indicates that a Driver is IP-native, and has no hardware address
+    Ip,
 }
 
 /// Main `embassy-net` driver API.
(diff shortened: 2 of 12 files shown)


xoviat:
I'll take a look at this later if someone else doesn't. I want to clarify that the main thing I care about is that the implementation remains zero-copy, as is currently done in the runner implementation. I don't really have a strong opinion on changes that aren't performance-relevant.

rubdos:
> the implementation remains zero-copy

That's also a requirement for us, although not strictly in an initial phase. I don't think this change specifically is really concerning regarding zero-copy. Thanks for weighing in!

xoviat:
Yeah I don't think this is performance relevant, I was just discussion in general.

xoviat:
I just looked and I agree that hardware address is the away forward. Make the CI pass and we can merge.

xoviat:
Re-export the hardware address type though

xoviat:
Re-export the hardware address type though

embassy-net-driver/Cargo.toml
```diff
@@ -21,4 +21,5 @@ target = "thumbv7em-none-eabi"
 features = ["defmt"]
 
 [dependencies]
-defmt = { version = "0.3", optional = true }
\ No newline at end of file
+defmt = { version = "0.3", optional = true }
+smoltcp = { version = "0.10", default-features = false }
```
Dirbaio, line 25: `embassy-net-driver` should not depend on smoltcp. This is so smoltcp can be upgraded without requiring every driver crate to update. Can you copypaste the `HardwareAddress` enum instead?

rubdos: > Can you copypaste the `HardwareAddress` enum instead?

IMO that loses part of the meaning of using `HardwareAddress`, since the whole purpose is to be able to use it as-is with `smoltcp`. It would be cool if we could just [depend on smoltcp_wire](https://github.com/smoltcp-rs/smoltcp/pull/812) instead, I suppose, although even that doesn't alleviate the concern.

I get the concern however, it felt pretty awkward to put smoltcp as a dependency.

Another option that I didn't mention in my OP, is to fix the hardware address type at compile time through a generic.

If after my plea here you still rather have the definition copied, I'll happily do that!

rubdos:
> Re-export the hardware address type though

I could reuse the one from `embassy_net`, I suppose, since it's reexported there already.

xoviat:
You can make hardware address an enum with ethernet and ieee802154 variants. In any case, a driver should only need to have embassy-net-driver as a dependency in cargo.toml, nothing else.

embassy-net-driver/Cargo.toml
rubdos: I slept on it, and I think I agree. So I'll rewrite it and fix CI. If we're happy with the result, I'll squash the results such that Cargo.toml remains unchanged.

embassy-net-driver/src/lib.rs
```diff
@@ -4,6 +4,16 @@
 
 use core::task::Context;
 
+/// Representation of an hardware address, such as an Ethernet address or an IEEE802.15.4 address.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum HardwareAddress {
```
Dirbaio, line 10: could you add an `Ip` variant with no contents, like in smoltcp? otherwise drivers with IP medium have to return a "dummy" hardware address which is a bit awkward.

(I know this was already "wrong" before, but now that you're changing this it would be the ideal time to fix it. Thank you! :pray: )

rubdos:
```diff
diff --git a/embassy-net/src/lib.rs b/embassy-net/src/lib.rs
index 746acae4..dbefb0ef 100644
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -252,23 +252,7 @@ impl<D: Driver + 'static> Stack<D> {
         resources: &'static mut StackResources<SOCK>,
         random_seed: u64,
     ) -> Self {
-        #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
-        let medium = device.capabilities().medium;
-
-        let hardware_addr = match medium {
-            #[cfg(feature = "medium-ethernet")]
-            Medium::Ethernet => to_smoltcp_hardware_address(device.hardware_address()),
-            #[cfg(feature = "medium-ip")]
-            Medium::Ip => HardwareAddress::Ip,
-            #[cfg(feature = "medium-ieee802154")]
-            Medium::Ieee802154 => to_smoltcp_hardware_address(device.hardware_address()),
-            #[allow(unreachable_patterns)]
-            _ => panic!(
-                "Unsupported medium {:?}. Make sure to enable it in embassy-net's Cargo features.",
-                medium
-            ),
-        };
-        let mut iface_cfg = smoltcp::iface::Config::new(hardware_addr);
+        let mut iface_cfg = smoltcp::iface::Config::new(to_smoltcp_hardware_address(device.hardware_address()));
         iface_cfg.random_seed = random_seed;

         let iface = Interface::new(
```

Including this, I suppose.

Dirbaio: yup! :D

rubdos:
I suggest I still squash some things up here, but this is the general idea, right? :-)

PR #1705, opened 2023-07-28
[STM32] H7 DAC DMA example and feature documentation

JuliDi:
This PR adds the DAC DMA example from the STM32L4's examples to the STM32H7, as well.

It also removes the memory.x files from other examples and uses the embassy-stm32 `memory-x` feature instead.
To make this more transparent, it also adds documentation for various features using https://crates.io/crates/document-features

Screenshot from 2023-07-28 16-58-51

If someone with more insight into the various features could add some better explanations that would be great.


JuliDi:
imo this can be merged and more documentation added on the go

Dirbaio (approved):
thanks!

PR #1710, opened 2023-07-28
rp: add async flash

Sizurka:
This adds an async mode to RP flash operations.  This mode requires a DMA channel but can utilize the background best effort read mechanism.  This mechanism avoids stalling actual code execution from XIP as well as avoiding filling the cache with flash data (that's going to RAM anyway).  Write and erase remain blocking, however.

See the [Pico SDK Example](https://github.com/raspberrypi/pico-examples/blob/master/flash/xip_stream/flash_xip_stream.c) and  section (2.6.3.4) of the manual.

### DIFF
--- a/embassy-boot/rp/src/lib.rs
+++ b/embassy-boot/rp/src/lib.rs
@@ -6,7 +6,7 @@ mod fmt;
 #[cfg(feature = "nightly")]
 pub use embassy_boot::FirmwareUpdater;
 pub use embassy_boot::{AlignedBuffer, BlockingFirmwareUpdater, BootLoaderConfig, FirmwareUpdaterConfig, State};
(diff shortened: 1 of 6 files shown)


Dirbaio:
Very interesting! Shame the hardware can't do the same for writes :smiling_face_with_tear:.

Can you add code that tests it to https://github.com/embassy-rs/embassy/blob/main/tests/rp/src/bin/flash.rs ?

Dirbaio:
looks good! Can you fix the Ci failures?

Sizurka:
Should be fixed now.  Sorry, I wasn't thinking about the other cases where the flash signature change would be needed.

Dirbaio (approved):
thank you! :rocket:

PR #1713, opened 2023-07-30
stm32f2 PLL overflow with crystal

MabezDev:
With a significant enough HSE input frequency, the VCO clock calculation will overflow a u32. Therefore, we must use the inner value and cast to u64 in this specific case to ensure the mul isn't clipped before applying the divider.

FYI the input source on my board is a 25mhz external crystal.


Dirbaio (approved):
thanks!

(mabez using stm32?! wut?)

PR #1714, opened 2023-07-30
stm32/dma: add writable ringbuf

xoviat:


xoviat:
@adamgreig Can you confirm this works for you? The control loop should just have to repeatedly call `write_exact`, which will exert backpressure.

PR #1715, opened 2023-07-30
rp: add dormant-wake functionality for Input

pennae:
this temporarily takes ownership of pins, mostly because we want to clean up the dormant-wake bits after a wakeup and because doing anything *else* with the input while we're waiting for a wakeup isn't possible at all. doing it like this lets us not impose any cost on those who don't use dormant wakes.

the error type for dormant-wake setup is dependent on the static type of the pin being used. this allows us to support dormant-wake for AnyPin and still allowing eg unwrap-infallibe in cases where pin types are statically known.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -75,6 +75,15 @@ pub enum Bank {
     Qspi = 1,
 }
 
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct DormantWakeConfig {
+    pub edge_high: bool,
+    pub edge_low: bool,
+    pub level_high: bool,
+    pub level_low: bool,
+}
+
 pub struct Input<'d, T: Pin> {
     pin: Flex<'d, T>,
 }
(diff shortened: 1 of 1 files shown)


Dirbaio:
- This only sets the flags on the GPIO, there's no API for going dormant. Will you PR that later?
- Why is dormant wake on QSPI disallowed? checking the datasheet the QSPI bank does have the dormant_wake_inte register too.
- The typelevel contraption to change the error type based on the bank is somewhat complicated. I propose one of the following:
   - Always return an error, it just won't happen for IO bank pins.
   - Require a `trait DormantWakePin`, implement it for only IO bank pins. This disallows `AnyPin` though.
- Error should not be `()`, it should be something like `DormantWakeNotSupported`
- Why not allowing just setting the dormant wake flags, without taking ownership of the pin? and clear them when dropping the `Input`. This'd allow configuring them once at program startup, while still using the pin for other stuff, without having to create/drop the DormantWake struct every time.

embassy-rp/src/gpio.rs
```diff
@@ -639,15 +653,59 @@ impl<'d, T: Pin> Flex<'d, T> {
...
+            w.set_edge_high(idx % 8, true);
+            w.set_edge_low(idx % 8, true);
+            w.set_level_high(idx % 8, true);
+            w.set_level_low(idx % 8, true);
+        });
+    }
+}
+
+pub struct DormantWake<'r, 'd, T: Pin> {
+    pin: &'r mut Flex<'d, T>,
```
Dirbaio, line 692: if you store `PeripheralRef<'a, AnyPin>` instead, you can get `DormantWake` to have 1 lifetime param instead of 2 .

PR #1719, opened 2023-07-30
Prevent UART from only getting 0s from the output

mvniekerk:


Dirbaio:
accidentally added `.idea` dir

Dirbaio:
can you squash the 2 commits together so it doesn't show up in history?

mvniekerk:
> accidentally added `.idea` dir

Fixed, saw that as well

mvniekerk:
Will try to squash the commits

PR #1721, opened 2023-07-30
stm32f2: Avoid resetting RTC backup domain

MabezDev:
Also ensure the pwr is enabled before trying to initialize. For the F2 series, this is in a separate clock control register which has to be enabled.


Dirbaio (approved):
thanks

PR #1724, opened 2023-07-31
Added STM32G0 Flash Support

bguruprasath5:
This pull request adds flash support for STM32G0 series in Embassy. The implementation has been tested and works as expected. Also, added a working example. Feedback is welcome.


Dirbaio (approved):
Looks great, thank you! :)

PR #1727, opened 2023-07-31
stm32: add hrtim v2

Dirbaio:


Dirbaio:
@xoviat could you take a look? they seem to be identical except v2 has an extra channel (F). 

I haven't tested on hadware, but this is needed to fix stm32 docs build so I'd rather add support for v2 even if untested.

RPdenBoer:
Not sure if this is the right place, but it seems `stm32-metapac-16.0.0/src/peripherals/hrtim_v2.rs` has a heap of hardcoded values that prevent F channel actually being used. eg. `assert!(n < 5usize);`
```
#[doc = "Timer X Output Enable"]
        #[inline(always)]
        pub fn set_t1oen(&mut self, n: usize, val: bool) {
            assert!(n < 5usize);
            let offs = 0usize + ([0usize, 2usize, 4usize, 6usize, 8usize][n] as usize);
            self.0 = (self.0 & !(0x01 << offs)) | (((val as u32) & 0x01) << offs);
        }
```

PR #1728, opened 2023-07-31
rp: fix qspi gpio interrupts, make qspi gpio optional

pennae:
qspi gpio interrupts simply would not have worked, at all. this alone has us suspect that nobody uses qspi pins for anything other than qspi, so we'll make the qspi gpio implementations optional as well to avoid the mostly-unnecessary overhead they come with. this does seem like much of a downside, anyone using qspi pins for gpio must already acknowledge this by using a different boot2 config.


Dirbaio (approved):
thank you! :upside_down_face:

PR #1730, opened 2023-08-01
LSE requires setting PWREN bit on STM32L4

bartekkowalski:
According to the reference manual RM0394, section 5.1.4, Backup domain Access:
To make changes to the RTC registers, you first need to enable the power interface clock by setting the PWREN bits.

This is missing from the current code and the program will hang on while !RCC.csr().read().lsirdy() {} every time on my STM32L431.

Adding the line to set the bit fixes the issue.


Dirbaio (approved):
thanks!

PR #1741, opened 2023-08-03
rp: fix adc test flakiness

pennae:
GP29 is connected to the cyw43 SCK pin. cyw43 is selected by default (due to rp2040 pins being input/pulldown by default), so the wifi chip is always selected and watches the SCK pin. this little bit of load on the SCK pin is enough to disturb the 300k voltage divider used for VSYS sensing, making the test flaky.


Dirbaio (approved):
ooh, a pin conflict :joy:

Dirbaio:
LGTM, can you fix CI?

PR #1743, opened 2023-08-04
stm32/dma: consolidate ringbuf

xoviat:
this has less savings than the other pr, but does not impact performance in a way that the compiler can't optimize


Dirbaio (approved):
thanks!

PR #1745, opened 2023-08-04
rp: add generic dormant-sleep functionality

pennae:
this is "generic" in that it doesn't require the user to set up anything specific to go to dormant sleep, unlike the C sdk which requires clock sources to be configured explicitly and doesn't much care about PLLs. we will instead take a snapshot of the current clock configuration, switch to a known clock source (very slow rosc, in this case), go to sleep, and on wakeup undo everything we've done (ensuring stability of PLLs and such).

tested locally, but adding tests to HIL seems infeasible. we'd need at least another pico or extensive modifications to teleprobe since dormant-sleep breaks SWD (except to rescue-dp), neither of which is feasible at this point. if we *did* want to add tests we should check for both rtc wakeups (with an external rtc clock source) and gpio wakeups.

### DIFF
--- a/embassy-rp/src/clocks.rs
+++ b/embassy-rp/src/clocks.rs
@@ -1,3 +1,4 @@
+use core::arch::asm;
 use core::marker::PhantomData;
 use core::sync::atomic::{AtomicU16, AtomicU32, Ordering};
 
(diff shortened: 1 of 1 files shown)


embassy-rp/src/clocks.rs
```diff
@@ -873,3 +875,131 @@ impl rand_core::RngCore for RoscRng {
...
+            // magic or even boot2 may place it into ram.
+            let pc: usize;
+            asm!(
+                "mov {pc}, pc",
+                pc = out (reg) pc
+            );
+            pc < 0x20000000
+        };
+        if is_in_flash {
+            asm!(
```
Dirbaio, line 983: these are asm so that rustc doesn't insert calls to other functions that might not be in flash/ram? can you add a comment explaining that?

Dirbaio (approved):
thanks!

PR #1746, opened 2023-08-04
wip: enc28j60 driver.

Dirbaio:
based on https://github.com/japaric/enc28j60

TODO:

- [x] Test that it works
- [x] Write an example
- [ ] Test in HIL
- [ ] Make it use a background task and `embassy-net-driver-channel`
- [ ] Make it use the irq pin, not spin CPU at 100%.
- [ ] Perhaps find some other solution for the regs, the current code is very verbose.

no time to do the remaining items, will do later. Driver works fine as-is if you're OK with 100% cpu usage, I've been using in production for years.


Qix-:
Did this get the IRQ pin working prior to merging?

Dirbaio:
no

Issue #1731, opened 2023-08-01
pend() called for no reason inside executor poll


lorislibralato:
https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/raw/mod.rs#L412
`dequeue_expired()` call `wake_task()` on all expired tasks, moving them into run_queue, but if run_queue is empty `wake_task()` will call `pend()` (resulting in another call to `poll()` when using `InterruptExecutor`).
A possible fix could be creating a function wake_task_no_pend() that call directly self.run_queue.enqueue() instead of self.enqueue().
I'll be happy to open a PR if the bug is confirmed.

Dirbaio:
fixed in https://github.com/embassy-rs/embassy/pull/1748

PR #1748, opened 2023-08-05
Avoid calling pend() when waking expired timers

lorislibralato:
Created a function similar to `wake_task()` that instead of calling `executor.enqueue()` directly call `executor.run_queue.enqueue()` to avoid unnecessary calls to `pend()`
Issue: #1731


Dirbaio (approved):
thanks!

PR #1749, opened 2023-08-05
Add Send and Sync impl to allow UsbDevice to be used on Core1 of RP2040

Meigs2:
I was working on my project when I ran into an issue where I could not send my UsbDevice to Core1 of my RP2040.

I identified that UsbDevice does not implement Sync and Send, and therefore could not be passed between threads. Implementing Send and Sync for UsbDevice appears to resolve this issue.

I've also added a test similar to the current multicore test. Attempting to compile without implementing Sync and Send causes the build to fail. Adding Send and Sync causes the build to succeed.

Note that I was unable to get the test for this working locally, but I was able to verify by copying the file to the rp2040 examples folder and removing lines 4-5.

### DIFF
--- a/embassy-usb/src/lib.rs
+++ b/embassy-usb/src/lib.rs
@@ -217,6 +217,9 @@ struct Inner<'d, D: Driver<'d>> {
     msos_descriptor: crate::msos::MsOsDescriptorSet<'d>,
 }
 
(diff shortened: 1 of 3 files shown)


Meigs2:
I'm having issues now apparently, closing until I get this fixed proper

PR #1750, opened 2023-08-06
enc28j60: check LSTAT on both bits 2 and 10

Qix-:
On rev6 of the enc28j60, pin 10 is indeed LSTAT. It appears 2 is also LSTAT depending on the revision, so this PR checks both (as 2 is specified to be 'always 0' in the correct case).


xoviat:
@Qix- can you rustfmt?

Qix-:
`cargo fmt` isn't showing any changes, is there a special command I should run?

Dirbaio:
root cause of the issue found: https://github.com/embassy-rs/embassy/pull/1746/commits/075a48c61dc588a94097c61add7088e82a989f65

so thankfully the workaround is no longer needed. Thanks anyway for the PR! :)

Qix-:
Ahhhhhh nice :D never been so satisfied to have a PR closed 😂 nice work.

PR #1756, opened 2023-08-06
Make usb-logger read and discard input data

dhylands:
This allows normal linux terminal emulators, like screen or picocom to be used with the usb_logger. Without this, calling `tcsetattr` with `TCSAFLUSH` will hang.


Dirbaio (approved):
thanks for the investigation and the fix! :)

PR #1758, opened 2023-08-06
Update `Handler` and `RequestHandler` traits to enable cross-core usb

Meigs2:
I was working on my project and while I was trying to get Core1 of my RP2040 to handle dealing with usb packets and message serialization/deserialization, I could not get the classes across to the other core.

I ended up having to add the `Sync` and `Send` traits to the `Handler` and `RequestHandler` traits in `embassy_usb`. After these modifications, I am able to build and use the `CdcAcmClass` successfully on Core1 of my RS2040.

I'm not fully aware of the ramifications of adding these traits bounds to the traits, but it appears to enable the functionality I need.

I also added a test, however I'm not sure if that is needed or wanted. If the fixes are not added, then the test should not compile.

It should also be noted that I could not get the test to run locally for some reason. I'm probably missing a tool. However, copying the files to the embassy-rp examples folder and removing lines 4-5 did effectively the same thing as running the test.

### DIFF
--- a/embassy-usb/src/class/hid.rs
+++ b/embassy-usb/src/class/hid.rs
@@ -370,7 +370,7 @@ impl<'d, D: Driver<'d>, const N: usize> HidReader<'d, D, N> {
 }
 
 /// Handler for HID-related control requests.
-pub trait RequestHandler {
+pub trait RequestHandler: Sync + Send {
     /// Reads the value of report `id` into `buf` returning the size.
     ///
     /// Returns `None` if `id` is invalid or no data is available.
--- a/embassy-usb/src/lib.rs
+++ b/embassy-usb/src/lib.rs
@@ -85,7 +85,7 @@ const STRING_INDEX_CUSTOM_START: u8 = 4;
 ///
 /// All methods are optional callbacks that will be called by
 /// [`UsbDevice::run()`](crate::UsbDevice::run)
-pub trait Handler {
+pub trait Handler: Sync + Send {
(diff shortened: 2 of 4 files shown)


Dirbaio:
Adding Send+Sync to Handler/RequestHandler has the implication that everyone writing a handler must make it thread-safe, even if they never intend to send the usb across cores/threads. It forces the overhead of using critical sections when it wouldn't be needed in theory. This is a somewhat strong restriction, I'm not sure all classes and examples meet that.

For your use case, are you trying to share the USB across cores (ie run one class on core0 and another on core1)? Or do you want to use it on core1 only? If it's the latter, you can send just the singleton (`p.USB`) and then construct everything on core1, so you don't  need any Send/Sync.

(for the tests: to run them you need [teleprobe](https://github.com/embassy-rs/teleprobe) and edit `.cargo/config.toml` to do `teleprobe local run`. Also note the tests are intended to verify functionality in CI, your test doesn't actually test USB is working. If it's just an "check it compiles" test, it should be an example.)

Meigs2:
Ah, so there's ramifications to adding those traits. Ok.

The use case that prompted this is indeed to have core1 handle all of my communications over USB to my client (usb + serialization), however I was having issues even doing that, sending just p.USB and constructing everything on core1. Granted, I'm new to rust + embassy overall so I'm sure I'm doing something wrong anyways. I'll give this another shot to see if I can get it all constructed on core1 and see if I can get it working without this fix.

PR #1761, opened 2023-08-08
Mark `Ticker` stream as fused

sgoll:
The `Ticker` stream never stops yielding values and can thus be marked as fused. This makes it more versatile to use in situations where a fused stream is required (such as `select_biased!`), without having to resort to `StreamExt::fuse()`.


Dirbaio (approved):
oh, nice! Thanks! :)

PR #1763, opened 2023-08-09
Refactor Channel/Sender/Receiver poll methods

rubdos:
(Was "Expose `poll_ready_to_{send,receive}` in `Sender`/`Receiver`")

These are useful methods when manually writing Futures.

@lulf suggested on Matrix to also expose `try_{recv,send}_with_context` in `Receiver` and `Sender`, but those are private in `Channel`.

### DIFF
--- a/embassy-sync/src/channel.rs
+++ b/embassy-sync/src/channel.rs
@@ -69,7 +69,7 @@ where
     /// Allows a poll_fn to poll until the channel is ready to send
     ///
     /// See [`Channel::poll_ready_to_send()`]
-    pub fn poll_ready_to_send(&self, cx: &mut Context<'_>) -> bool {
+    pub fn poll_ready_to_send(&self, cx: &mut Context<'_>) -> Poll<()> {
         self.channel.poll_ready_to_send(cx)
     }
 }
@@ -117,7 +117,7 @@ impl<'ch, T> DynamicSender<'ch, T> {
     /// Allows a poll_fn to poll until the channel is ready to send
     ///
     /// See [`Channel::poll_ready_to_send()`]
-    pub fn poll_ready_to_send(&self, cx: &mut Context<'_>) -> bool {
+    pub fn poll_ready_to_send(&self, cx: &mut Context<'_>) -> Poll<()> {
         self.channel.poll_ready_to_send(cx)
     }
 }
@@ -162,7 +162,7 @@ where
     /// Allows a poll_fn to poll until the channel is ready to receive
     ///
     /// See [`Channel::poll_ready_to_receive()`]
-    pub fn poll_ready_to_receive(&self, cx: &mut Context<'_>) -> bool {
+    pub fn poll_ready_to_receive(&self, cx: &mut Context<'_>) -> Poll<()> {
         self.channel.poll_ready_to_receive(cx)
     }
 }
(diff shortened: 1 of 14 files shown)


lulf (commented):
I think it's better to make the try_x_with_context public rather than adding another fn specific to this use case, but in any case it would be better to return the Poll type than a bool IMHO

rubdos:
> it would be better to return the Poll type than a bool IMHO

Agreed, I had to dive in the code to look what `true` and `false` meant here. `Poll<()>` would be better. Is that something we refactor in this PR?

Dirbaio:
the convention is frob, try_frob, poll_frob:

```rust
async fn frob(&mut self, arg: u32) {..}
fn try_frob(&mut self, arg: u32) -> Result<(), TryFrobError> {..}
fn poll_frob(&mut self, cx: &mut Context, arg: u32) -> Poll<()> {..}
```

where `poll_frob` will actually do the frobbing if ready.

Unless we have a good reason, these are the ones we should expose.

lulf:
Makes sense to have poll_frob do the frobbing. That way you prevent a misuse of the API in the event some conditions changed between poll() and try().

rubdos:
If I interpret your frobbing correctly, this would require a refactoring to `poll_recv(..) -> Poll<()>`?

Dirbaio:
yes!

rubdos:
In case of send, there's a slight problem:

```rust
fn poll_send(&mut self, cx: &mut Context, message: T) -> Poll<()> {}
```

You move the message into the `poll_send` method, but the method may fail and have to move the message back. A `poll_send` method might not make sense in this context.

embassy-sync/src/channel.rs
```diff
@@ -144,6 +158,20 @@ where
     pub fn try_recv(&self) -> Result<T, TryRecvError> {
         self.channel.try_recv()
     }
+
+    /// Allows a poll_fn to poll until the channel is ready to receive
+    ///
+    /// See [`Channel::poll_ready_to_receive()`]
+    pub fn poll_ready_to_receive(&self, cx: &mut Context<'_>) -> Poll<()> {
```
Dirbaio, line 165: we should be consistent on `recv` vs `receive`. Not sure which though. `receive` is nicer because no abbreviations, but `recv` would avoid a breaking change.

rubdos: If you want, we can move on to receive, and add a `#[deprecated]` note.

rubdos: Another argument for changing: we're breaking the API already by moving to `Poll<()>` over `bool`.

rubdos: I took the liberty of commiting my suggestion in 96a5888, I can easily drop/alter it again if we decide on something else.

Dirbaio:
`recv` is still mentioned in `DynamicReceiver`, and in `DynamicRecvFuture`, and in some links in doc comments. Could you change these?

I'd outright do the rename, without adding `deprecated` aliases. It's less work removing the m later, and next `embassy-sync` is probably going to be `0.3` anyway.

rubdos:
Done, I think! :-) `rg recv embassy-sync` comes up empty now.

Dirbaio (approved):
LGTM, thank you!

PR #1766, opened 2023-08-10
stm32/rtc: add start/stop wakeup

xoviat:
TODO: 

- [x] autocompute prescaler and set to minimize time before requested duration
- [x] compute actual time elapsed according to prescaler adjustment

### DIFF
--- a/embassy-stm32/src/rtc/v2.rs
+++ b/embassy-stm32/src/rtc/v2.rs
@@ -23,6 +23,73 @@ impl super::Rtc {
     }
 
     #[allow(dead_code)]
+    #[cfg(all(feature = "time", stm32wb))]
+    // start the wakeup alarm with the given duration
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/rtc/v2.rs
```diff
@@ -22,6 +22,73 @@ impl super::Rtc {
...
+        let rtc_ticks = duration.as_ticks() * rtc_hz / TICK_HZ;
+        let rtc_ticks = if rtc_ticks > u16::MAX as u64 {
+            u16::MAX
+        } else {
+            rtc_ticks as u16
+        };
+
+        while !RTC::regs().isr().read().wutf() {}
+
+        RTC::regs().isr().modify(|w| w.set_wutf(false));
```
xoviat, line 50: This procedure should only be done if the wakeup timer is enabled.

PR #1768, opened 2023-08-10
fix rng ced toggling sequence on reset.

lucasgranberg:
This fixes an error where the first call to async_fill_bytes gives an clock error. The manual (for stm32wl) states that the ced bit has to be toggled on and off in this manner.

NISTC should also be "CUSTOM" ie 1 for CONFIGA.


Dirbaio (approved):
thanks!

PR #1770, opened 2023-08-10
Executor: prepare 0.2.1 release

bugadani:
I think it's okay to have this as a point release, as it only seems to contain a bunch of fixes and a new API.


Dirbaio (approved):
Thank you!

bugadani:
No, thank YOU! :)

PR #1771, opened 2023-08-11
embassy_sync::zero_copy_channel

rubdos:
Discussion dump from Matrix

```
        ⤷ ╡ My eye just fell on zerocopychannel in that crate, would that be something to expose in embassysync? Looks like a useful
          ╡ primitive
        ⤷ ╡ (especially since atm I'm working with Channel and zero-copy was one concern in the current prototype; was actually
          ╡ planning on writing something like zero-copy-channel)
 &dirbaio ╡ maybe, why not
        ⤷ ╡ I already wrote it to make it independent from embassy-net
        ⤷ ╡ there's a few weird things about the design though, that's why I postponed putting it in embassy-sync
   rubdos ╡ I'll probably take the initiative tomorrow or on friday for it, then.
        ⤷ ╡ Just make a preliminary PR, and we can see from there :)
 &dirbaio ╡ mainly, it's inherently different from Channel in that you don't send ownership of the objects
   rubdos ╡ I mean, that's kind of the point, right? :D
 &dirbaio ╡ you have to give it a buffer of alreayd-initialized T's, and you only get &mut when sending and receiving
        ⤷ ╡ the items get "recycled" infinitely
        ⤷ ╡ wich is a bit weird
        ⤷ ╡ so you can't use it to send e.g. a Vec<T>
        ⤷ ╡ but perhaps that's fine
        ⤷ ╡ the alternative of having the user use MaybeUninit and unsafe is worse
   rubdos ╡ Would you want to refactor it a bit to use MaybeUninit around your types?
        ⤷ ╡ Maybe T: Default and some MaybeUninit magic could clean it up?
        ⤷ ╡ even without Default, I suppose
 &dirbaio ╡ like, it'd have to say "on send, you must logically move in an owned value into the MaybeUninit. On recv, you must
          ╡ logically move out or drop. The channel won't drop for you, if you don't drop you'll have a leak"
        ⤷ ╡ actually, if the user wants to do that, they can already do it with a ZerocopyChannel<MaybeUninit<T>>
   rubdos ╡ If T: Default, then the user can write directly into the channel.  For some nice T::default() implementations, that would
          ╡ be kinda clean, I presume
        ⤷ ╡ But I get what you say, it'll be messy either way. I'll have a close look to the code, play with it a bit, and I'll save
          ╡ this discussion for the PR
        ⤷ ╡ My head and hands are somewhere else atm :-)
```


rubdos:
This current version passes CI (with docs), but the interface is indeed a bit clumsy. I'm trying some design out now, to make the Sender and Receiver design slightly less clumsy.

rubdos:
See 7aba82c for a suggested design and PoC.

Dirbaio:
I'm not convinced about the visitor API.

You might want to obtain a buffer to start an in-place read into it, but then abort the operation. This is not possible with the visitor API. There's cases in the embassy-net drivers that do this.  

Also it makes it too easy to accidentally send bad data on future cancel. When you cancel (drop) a future, the visitor will get dropped causing half-filled data to be sent into the channel.

rubdos:
> I'm not convinced about the visitor API.

Neither was I... It's very bulky and the types leak everywhere...

Dirbaio:
I've reverted the visitor/guard API. Thanks for the PR!

PR #1772, opened 2023-08-11
Add compound descriptor set tracking to the usb descriptor writer

alisomay:
# Content

While writing a USB Audio Class (MIDI) I've realized that the initial class specific descriptor requires the total length of the whole class specific descriptor set. 

Page: `38`, Section `B.3.2`
https://www.usb.org/sites/default/files/USB%20MIDI%20v2_0.pdf

This is also true for class specific descriptors from the Video class.
https://www.xmos.ai/download/AN00127:-USB-Video-Class-Device(2.0.2rc1).pdf#subsubsection.2.7.2

I'd like to add at least the audio class to `embassy-usb` step by step.
This PR aims to provide the underlying machinery to unlock having dynamic length and types of class specific descriptors depending on the user configuration which might come from lets say.. a builder pattern. 

I have chosen to call these kinds of descriptor chains a "compound set" of descriptors where the initial descriptor needs a total length of the following ones and the following ones may vary in length and type.

It is mainly about tracking the total length of the compound set and rewriting it to the initial descriptor. 

# Notes

- This will be my first contribution to this code base so don't hesitate to nitpick. I don't know the style here yet :) 
- I've written long function names, verbose comments and didn't think so much about publicity of the functions. I'd gladly update the code according to your advises.
- If you see a better path of implementing this functionality or a better api to expose to the user. I'd love to hear it 🙏 

Thanks for your review 🙏

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -439,6 +439,48 @@ impl<'a, 'd, D: Driver<'d>> InterfaceAltBuilder<'a, 'd, D> {
         self.builder.config_descriptor.write(descriptor_type, descriptor)
     }
 
+    /// Add a custom descriptor to this alternate setting,
+    /// assuming that the descriptor is the initial one which will be followed by variable number
+    /// of descriptors of the same type forming a compound descriptor set.
+    ///
+    /// Descriptors are written in the order builder functions are called. Note that some
+    /// classes care about the order.
+    ///
+    /// # Details
+    ///
+    /// In the USB specification, when it comes to class-specific descriptors, the presence of a "total length" field is not universally common across all classes.
+    /// However, it is found in several class specifications, particularly in those where a variable number of descriptors or a variable configuration might follow.
+    ///
+    /// For instance:
+    ///
+    ///  - **Audio Class:** The class-specific header descriptor in the Audio class contains a "Total Length" field, which indicates the entire length of the class-specific AudioControl interface descriptor. This is useful because the AudioControl interface can have a variable configuration depending on the number and type of audio channels, terminal interfaces, etc.
+    ///
+    ///  - **Video Class:** The Video class's class-specific header descriptor also contains a "Total Length" field.
+    ///
+    /// The purpose behind the "Total Length" (or equivalent) field in these class-specific descriptors is to provide a straightforward way to determine the full length of a compound descriptor set.
+    /// This can be especially helpful for parsing or skipping over the entirety of the descriptor set if needed.
+    ///
+    /// While it's common in certain classes like Audio and Video,
+    /// it's not a universal feature of all USB class-specific descriptors.
+    pub fn start_writing_compound_set_tracking_total_length(&mut self, descriptor_type: u8, descriptor: &[u8]) {
+        let descriptor_writer = &mut self.builder.config_descriptor;
+        // Start tracking the total length of the compound descriptor set.
+        descriptor_writer.start_tracking_total_length_of_compound_descriptor_set(descriptor_writer.position());
+        // Write the initial descriptor of that particular set.
+        self.builder.config_descriptor.write(descriptor_type, descriptor);
+    }
+
+    /// End writing a compound descriptor set updating the total length area in the initial descriptor.
+    ///
+    /// The offset of the total length bytes should be provided.
+    /// This number may change depending on the class, please derive it from the specification.
+    pub fn end_writing_compound_set_updating_total_length(&mut self, offset: usize) {
+        // End tracking the total length of the compound descriptor set and write it to the initial descriptor of that set on provided offset.
+        self.builder
+            .config_descriptor
+            .end_tracking_total_length_of_compound_descriptor_set_and_update_the_initial_descriptor(offset);
+    }
+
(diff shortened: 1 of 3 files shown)


Dirbaio:
Thanks for the PR!

I'm not sure if building this functionality into the descriptor writer itself is the best solution. The issue with `wTotalLength` is part of a bigger issue, which is the descriptor writer currently doesn't allow modifying already-written descriptors. 

There's other cases where you need modifying previous descriptors. For example, you sometimes need to write the interface number of an interface that you haven't allocated yet. The CDC ACM class has this [hack](https://github.com/embassy-rs/embassy/blob/main/embassy-usb/src/class/cdc_acm.rs#L185) where it assumes the interfaces are allocated sequentially, it adds 1 to "guess" the next interface number. If it could modify the already-written descriptor, this wouldn't be needed.

It'd be adding these functions:

```rust
fn position(&self) -> usize;
fn overwrite(&mut self, position: usize, data: &[u8]);
```

alisomay:
I agree, this was more thought with a patching approach.
I also think that building this functionality would make more sense with a generic approach. 
I'll rework the PR to come up with that since I get the green light that the functionality makes sense.

alisomay:
The new commits
- [7b91597](https://github.com/embassy-rs/embassy/pull/1772/commits/7b91597e9c96410be72918860eee24ef1de6c8a5)

Removes the case specific implementation and adds overwrite functionality to the `InterfaceAltBuilder`.

- [965e234](https://github.com/embassy-rs/embassy/pull/1772/commits/965e234b8dca1dd1377bb7171769374600afa736)
- [1ca3c8c](https://github.com/embassy-rs/embassy/pull/1772/commits/1ca3c8c4e6ff171f5b3e7de1d3cf69fef0d5625c)

Suggests to use interior mutability in fields where other builders would like to mutate in the root `Builder`.

# The background about suggesting interior mutability

The other builders such as `FunctionBuilder` or `InterfaceAltBuilder` etc. mutably borrow the whole root builder.
If this pattern is preferred I think using interior mutability in the root builder is not a bad idea. 
This part of the code is concerned with the initialization so I assume that extreme runtime optimizations are not necessary so using some `RefCell`s shouldn't hurt us much.

Also by introducing interior mutability we expose a more flexible interface to the user when writing descriptors. 
Lines such as `drop(func)` shouldn't be necessary. Also overwriting descriptors from earlier interfaces when defining multiple interfaces in the class implementation becomes seamless and careful thought about the borrow checker behaviour is not necessary.

What do you think @Dirbaio? 
I don't see any disadvantages with this pattern, I wonder if you see some.

Thanks for your time and review again.

Dirbaio:
builders must require `&mut self` because order of descriptors matters, for example to determine which endpoint belongs to which interface. If we allow `&self` you can do things like:

```rust
let iface1 = func.interface();
let alt1 = iface1.alt_setting(...);
let iface2 = func.interface();
let alt2 = iface2.alt_setting(...);

// creates an endpoint on iface1, but the host will think the endpoint belongs to iface2.
alt1.endpoint_bulk_out(...); 
```

alisomay:
Yes it does 😕. Seems I disregarded that. Thanks for your patience. I'll rework it with that in mind.

Dirbaio:
What was the motivation to allow `&self`? I think everything can be done with `&mut`.

alisomay:
I was refactoring the `cdc_acm.rs` to use the `overwrite`.

Then using `overwrite` on the first `InterfaceAltBuilder` instance after creating the second instance of `InterfaceAltBuilder` (it needs to come after that to take the number of the second interface) created a problem for double mutable borrows. 

When I was playing with that, I questioned the need for `&mut self` forgetting that it keeps the order as a side effect and wanted to see how things would feel without it. 

Then you reminded me the good side effect of it 😅

alisomay:
Sure!

One can easily work around this by doing.

```rust
        // Data interface
        let mut iface = func.interface();
        let data_if = iface.interface_number();
        let data_if_number_byte = data_if.0;

        let mut alt = iface.alt_setting(USB_CLASS_CDC_DATA, 0x00, CDC_PROTOCOL_NONE, None);
        let read_ep = alt.endpoint_bulk_out(max_packet_size);
        let write_ep = alt.endpoint_bulk_in(max_packet_size);

        // Fill in the subordinate interface byte with the right interface number.
        alt.overwrite(subordinate_interface_position, &[data_if_number_byte]);

        drop(func);
```

Since they point always to the same writer.

alisomay:
> ah... I was thinking something along the lines of: add a method on all builders that gives `&mut` on the "raw" descriptor writer. So you can access it through whatever builder you have "active" now, it doesn't have to be the same one as before.

About that, why not. 
That would require to make the writer public across the boundry of the crate. 
I just thought it might be too distruptive to the current design. 
If you think making that public is fine, it might give more control over the buffer.

Dirbaio:
> That would require to make the writer public across the boundry of the crate.

Ah, that's true it's not public. Probably better to not expose it. Adding `position()`, `overwrite()` to the builders themselves sounds good to me.

Dirbaio:
sounds good to me!

Dirbaio:
Hello,

PR #1777, opened 2023-08-13
embassy-net:tcp:write_with/read_with

MabezDev:
- Add async versions of smoltcp's `send` and `recv` closure based API.

Two things I'm not sure about:
- I copied the docs for the functions from smoltcp, is this correct, or should I link to the smoltcp docs instead?
- How to handle `Err(tcp::RecvError::Finished)`, I don't think this is an error, but we can't return `Ok()` because the type is defined by the return type closure. For now I just treated it as an error, because it should only happen when the socket is closed.

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -82,6 +82,17 @@ impl<'a> TcpReader<'a> {
     pub async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
         self.io.read(buf).await
     }
+
+    /// Call `f` with the largest contiguous slice of octets in the receive buffer,
+    /// and dequeue the amount of elements returned by `f`.
+    ///
+    /// If no data is available, it waits until there is at least one byte available.
+    pub async fn read_with<F, R>(&mut self, f: F) -> Result<R, Error>
+    where
+        F: FnOnce(&mut [u8]) -> (usize, R),
+    {
+        self.io.read_with(f).await
+    }
 }
 
 impl<'a> TcpWriter<'a> {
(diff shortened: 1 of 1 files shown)


embassy-net/src/tcp.rs
```diff
@@ -359,6 +417,55 @@ impl<'d> TcpIo<'d> {
...
+    }
+
+    async fn recv<F, R>(&mut self, f: F) -> Result<R, Error>
+    where
+        F: FnOnce(&mut [u8]) -> (usize, R),
+    {
+        let mut f = Some(f);
+        poll_fn(move |cx| {
+            self.with_mut(|s, _| {
+                if s.recv_queue() == 0 {
```
Dirbaio, line 450: you should do more checks here, so that "This function errors if the receive half of the connection is not open." is true. As it is now it'll hang instead of erroring.

embassy-net/src/tcp.rs
```diff
@@ -359,6 +417,55 @@ impl<'d> TcpIo<'d> {
...
     }
 
+    async fn send<F, R>(&mut self, f: F) -> Result<R, Error>
+    where
+        F: FnOnce(&mut [u8]) -> (usize, R),
+    {
+        let mut f = Some(f);
+        poll_fn(move |cx| {
+            self.with_mut(|s, _| {
+                if s.send_queue() == s.send_capacity() {
```
Dirbaio, line 427: Same as recv, this needs more checks. Doc says "This function returns `Err(Error::Illegal)` if the transmit half of the connection is not open". This'll hang instead of erroring in the case where the transmit half is not open but the buffer is full.

Dirbaio:
- We shouldn't mix write/send, read/recv names. I propose `write_with`, `read_with`.
- Docs should document that these functions async-wait until there's at least 1 byte available to read/write.

PR #1779, opened 2023-08-15
Forward capacity impls

MabezDev:
raw & icmp sockets are not supported in embassy_net, so those are missing here.


Dirbaio (approved):
:ok_hand: :pray:

PR #1784, opened 2023-08-16
stm32f2: Check RCC settings before applying

sgoll:
This makes sure to check the RCC settings and derived clocks for hardware compatibility before applying them. Previously the check would be run _after_ setting the parameters, possibly making the execution unreliable when parameters are out of bounds.


Dirbaio (approved):
makes sense, thank you! :)

Issue #1516, opened 2023-04-12
cyw43: Bluetooth support


segfaultsourcery:
Are there plans to support bluetooth as well? As far as I understand, the CYW43439 chip on the Pico W supports it.

Meigs2:
Bluetooth would be nice! 

I took a look at the driver we use currently for the cyw43 and it doesn't seem like it supports bluetooth, but that's just a cursory search through the repo.

Meigs2:
Well actually, it seems like the driver does have a structure for some bluetooth stuff:

https://infineon.github.io/wifi-host-driver/html/structwhd__btwt__config__params__t.html

brandonros:
https://github.com/raspberrypi/pico-sdk/releases/tag/1.5.1 looks like it was "added" here

I think what we'd need to translate/implement is here: https://github.com/raspberrypi/pico-sdk/tree/master/src/rp2_common/pico_btstack

Not sure if a new .bin/different .bin would be needed here: https://github.com/embassy-rs/embassy/tree/main/cyw43-firmware

Dirbaio:
The meat for sending/receiving HCI packets over SPI is here https://github.com/raspberrypi/pico-sdk/tree/master/src/rp2_common/pico_cyw43_driver/cybt_shared_bus

The issue is that will get us only "send/receive HCI packets to the bluetooth controller". We then need a bluetooth host stack on top of that. (the host stack is what implements the higher layers of BLE, like L2CAP, ATT, GATT...). 

Pico-sdk is using BTStack as the host stack. It is proprietary, requires licensing to use. RPi and BlueKitchen (the company behind BTStack) have entered some agreement to [license BTStack for free on Pico W-based hardware only](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_btstack/LICENSE.RP). 

This license does NOT cover other hardware, and we're in the same situation for other hardware: STM32WB[^1] and nRF52/53[^2]'s controller blobs also speak HCI, we also need a host stack for them.

[^1]: they have some kind of higher-level vendor-specific commands for higher-layer protocols. I dunno how complete they are or how nice they're to use compared to a real BLE host stack.
[^2]: the "old" softdevice (what's implemented in `nrf-softdevice`) does have implementations for higher-layer protocols, but it's in "maintenance mode" according to Nordic, doesn'd support nRF53 nor newer Bluetooth features. The newer "softdevice controller" blob is HCI only.

[...]

PR #1789, opened 2023-08-17
cyw43: split up control init firmware/wifi

brandonros:
I'm looking at what needs to be done to start https://github.com/embassy-rs/embassy/issues/1516

Leaving `init` alone doesn't break it but allows me to call `init_firmware` and skip the `init_wifi` stuff.

Thoughts?


cyw43/src/runner.rs
```diff
@@ -116,6 +116,29 @@ where
...
+    }
+
+    pub(crate) async fn init_bluetooth(&mut self) {
+        // cybt_wait_bt_ready
+        loop {
+            let val = self
+                .bus
+                .bp_read32(CHIP.pmu_base_address + REG_BACKPLANE_BT_CTRL_REG_ADDR)
+                .await;
+            if val & BTSDIO_REG_BT_AWAKE_BITMASK == 1 {
```
brandonros, line 133: i can't tell why this is never being hit based on https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_cyw43_driver/cybt_shared_bus/cybt_shared_bus.c#L112-L155

are they doing something different with the firmware upload/download?

Issue #1778, opened 2023-08-14
STM32L4 (others?) RTC clock source config also requires setting PWREN bit


alexferro:
Issue #1711 (and the accompanying pull request) fixed this in the case of using the LSE for the RTC clock source, but while the LSI can be started without the PWREN bit set, the call to unlock_registers in rtc_v2 waits indefinitely due to the PWR clock being disabled. 

Since the RTC is being unconditionally initialized, the PWREN bit needs to be also set unconditionally, rather than just in the LSE branch.

KorribanMaster:
Have you uploaded your patch for the l4 already so i can try if it fixes it on the f4 (f446) as well ?

PR #1793, opened 2023-08-18
stm32l4: set pwren in rcc regardless of clock source

ARizzo35:
Fixes https://github.com/embassy-rs/embassy/issues/1778


Dirbaio (approved):
thanks!

PR #1797, opened 2023-08-18
stm32: allow setting PWM duty cycle to 100%

aurelj:
Setting the compare_value to max_compare_value make the PWM output go low when the timer reach the max_compare_value and go high again on the next timer clock, when the value roll back to 0.
So to allow a 100% PWM that never goes low, the compare_value must be set to max_compare_value + 1.


Dirbaio (approved):
thanks!

PR #1802, opened 2023-08-18
Add error translation to tcp errors

### LINKED ISSUE #1801: Properly Propagate Protocol Problems (tcp::Error -> embeded_io::Error)
Currently there's no error translation for converting from embassy_net::tcp::Error to embedded_io::Error.
It is rather concerning to get a "Other" as an Error value instead of useful information that you have to get by breaking at the conversion spot in tcp.rs https://github.com/embassy-rs/embassy/blob/97da34595cb9943affadb1ce8785556108b977df/embassy-net/src/tcp.rs#L515

I was seeing the embedded_io Other error when under the hood the actual error was ConnectionReset, which the function above converts to Other which loses all meaning.

avlec:
Translation of tpc client ConnectError and Error to the appropriate embedded_io_async errors

Closes #1801

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -384,13 +384,20 @@ mod embedded_io_impls {
 
     impl embedded_io_async::Error for ConnectError {
         fn kind(&self) -> embedded_io_async::ErrorKind {
-            embedded_io_async::ErrorKind::Other
+            match self {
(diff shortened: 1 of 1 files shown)


Dirbaio:
LGTM, can you rustfmt?

Dirbaio (approved):
thank you! :upside_down_face:

PR #1803, opened 2023-08-18
configure flash latency after axi clock and  handle different flash on STM32H7A/B devices

randomplum:
Flash latency was being calculated and set based on requested clock, but before the clock was actually configured.
STM32H7A/B devices has slightly different FLASH peripheral.

### DIFF
--- a/embassy-stm32/src/flash/mod.rs
+++ b/embassy-stm32/src/flash/mod.rs
@@ -65,9 +65,11 @@ impl FlashRegion {
 #[cfg_attr(flash_f7, path = "f7.rs")]
 #[cfg_attr(flash_g0, path = "g0.rs")]
 #[cfg_attr(flash_h7, path = "h7.rs")]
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/flash/mod.rs
```diff
@@ -65,9 +65,11 @@ impl FlashRegion {
 #[cfg_attr(flash_f7, path = "f7.rs")]
 #[cfg_attr(flash_g0, path = "g0.rs")]
 #[cfg_attr(flash_h7, path = "h7.rs")]
+#[cfg_attr(flash_h7ab, path = "h7ab.rs")]
```
Dirbaio, line 68: since both are so similar, it's nicer to have a single `h7.rs` file handling both. Also notice both versions have CRCRDERR, I snuck in the fix for that in your stm32-data PR.

randomplum: good point

Dirbaio (approved):
thank you!

PR #1807, opened 2023-08-20
Fix hardcoded buffer length in USB NCM causing broken link on USB 2.0.

romixlab:
NCM is working without any issues with hardcoded buffer length of 64 on USB FullSpeed. But on USB HighSpeed, where maximum packet size for bulk endpoints is 512, it is failing. Even if 64 is provided to the CdcNcmClass:new().
Probably the underlying USB driver or USB hardware is still sending full packets even if length 64 is configured, leaving majority of the packet filled with zeros.

Tested on STM32H743ZI in FS mode and on STM32H747XI in HS mode (ULPI). Pings and simple TCP loopback are working fine now, Wireshark also do not show any apparent issues.

### DIFF
--- a/embassy-usb/src/class/cdc_ncm/mod.rs
+++ b/embassy-usb/src/class/cdc_ncm/mod.rs
@@ -248,6 +248,8 @@ pub struct CdcNcmClass<'d, D: Driver<'d>> {
     write_ep: D::EndpointIn,
 
     _control: &'d ControlShared,
+
+    max_packet_size: usize,
 }
 
 impl<'d, D: Driver<'d>> CdcNcmClass<'d, D> {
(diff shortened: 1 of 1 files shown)


romixlab:
I see no easy way to avoid hardcoding the scratchpad buffer length, though. It has to be 512 to accommodate both FS and HS. But maybe compiler can see this and optimize when 64 is actually passed to CdcNcmClass::new().

Dirbaio (approved):
thank you!

> I see no easy way to avoid hardcoding the scratchpad buffer length, though. It has to be 512 to accommodate both FS and HS. But maybe compiler can see this and optimize when 64 is actually passed to CdcNcmClass::new().

It could be done with const generics, but I guess it's not worth the complexity for now. If someone really wants these 448 bytes back in the future, they can send a PR :)

romixlab:
Awesome :)

PR #1812, opened 2023-08-21
Make `AvailableTask` public

bugadani:
This change should allow custom TaskStorage implementations

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -147,10 +147,7 @@ impl<F: Future + 'static> TaskStorage<F> {
     pub fn spawn(&'static self, future: impl FnOnce() -> F) -> SpawnToken<impl Sized> {
         let task = AvailableTask::claim(self);
         match task {
-            Some(task) => {
-                let task = task.initialize(future);
-                unsafe { SpawnToken::<F>::new(task) }
-            }
+            Some(task) => task.initialize(future),
             None => SpawnToken::new_failed(),
         }
     }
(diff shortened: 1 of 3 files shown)


embassy-executor/src/raw/mod.rs
```diff
@@ -186,25 +183,33 @@ impl<F: Future + 'static> TaskStorage<F> {
...
         task.raw
             .state
             .compare_exchange(0, STATE_SPAWNED | STATE_RUN_QUEUED, Ordering::AcqRel, Ordering::Acquire)
             .ok()
             .map(|_| Self { task })
     }
 
-    fn initialize(self, future: impl FnOnce() -> F) -> TaskRef {
+    /// Initialize the [`TaskStorage`] to run the given future.
+    pub fn initialize<S>(self, future: impl FnOnce() -> F) -> SpawnToken<S> {
```
Dirbaio, line 204: if this API is safe, it can't let the user choose the `S`. It can be used to send non-Send things. It has to be restricted like the current `spawn()` public API.

bugadani:
@Dirbaio this is working pretty well. You can see a crude demo [here](https://github.com/card-io-ecg/card-io-fw/pull/95/commits/827b92bfd10f660aaf4e76611edb9c44b0491fb9) if you'd like, it's working pretty well.

I've taken the liberty of moving the wall of explanative text into AvailableTask, as I believe it is now a better place for it. I've also deduplicated the functions that looked the same.

Dirbaio (approved):
thanks!

PR #1813, opened 2023-08-22
Modified the STM Uart baud calculation to be fully 32-bit

diondokter:
This saves me 620 bytes on my G0 CM0+ where I don't use any u64's anywhere.

I've tested the new calculation with this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d617401ffcb409b8bb0cb2f25567b453

It produces identical results compared to the old calculation (at least with the parameters in my test).


Dirbaio (approved):
thank you!

PR #1808, opened 2023-08-20
cyw43: upload bluetooth firmware to ram during init

brandonros:
attempt #2 of https://github.com/embassy-rs/embassy/pull/1799 of trying to convert parts of https://github.com/raspberrypi/pico-sdk/commit/c8ccefb9726b89d35d230d194d6977705908ffe7 to Rust so we can plug the HCI layer into bleps

expected (pico-sdk with cyw43-driver):

```
reading BT_CTRL_REG_ADDR should be 0 a few times until it is finally 0x01000100 while waiting in a loop and sleeping 1ms
```

actual (Rust logs):

```
1.632834 DEBUG wait_bt_ready
└─ /Users/brandon/Desktop/embassy/cyw43/src/fmt.rs:127
1.634490 DEBUG BT_CTRL_REG_ADDR = 8fc0f42a
└─ /Users/brandon/Desktop/embassy/cyw43/src/fmt.rs:127
1.637088 DEBUG BT_CTRL_REG_ADDR = 00000000
└─ /Users/brandon/Desktop/embassy/cyw43/src/fmt.rs:127
1.639505 DEBUG BT_CTRL_REG_ADDR = 00000000
└─ /Users/brandon/Desktop/embassy/cyw43/src/fmt.rs:127
```

Do we know what the bits inside `BT_CTRL_REG_ADDR` are trying to tell us/represent or is that part of the datasheet not public?

PR #1818, opened 2023-08-23
cyw43: hci outline

brandonros:
as expected, we can't upload the bluetooth firmware in a different location than the C library does because althought `cybt_ready` register check passes, when you go to read `WLAN_RAM_BASE_ADDR` or whatever, since the CLM isn't loaded or the core isn't up (guessing...), you do not get a valid address back

I'm struggling with how to break up `Runner` and `Control` and `Bus`. I'm going to keep banging on it until something clicks. I've tried a few different approaches and keep coming up short.

We should probably focus on landing https://github.com/embassy-rs/embassy/pull/1808 to be good enough to merge and this will go after, maybe?


cyw43/src/ble_connector.rs
```diff
@@ -0,0 +1,64 @@
...
+    runner: Runner<'a, PWR, SPI>,
+}
+
+impl<'a, PWR, SPI> BleConnector<'a, PWR, SPI>
+where
+    PWR: OutputPin,
+    SPI: SpiBusCyw43,
+{
+    pub fn new(runner: Runner<'a, PWR, SPI>) -> BleConnector<'a, PWR, SPI> {
+        return BleConnector { runner };
```
brandonros, line 15: @Dirbaio sorry for the ping but I could use some guidance. I think you know as well as I do we can't use `runner` here due to borrowing checking. I know before you mentioned maybe putting `Bus` behind a mutex but I didn't really have great success with that either.

How would you recommend an `HciConnector` makes its way back to `runner.bus` to be able to do `hci_read()` and `hci_write()`? A channel? Waker like ioctl does?

Same for refactoring the bluetooth firmware init, it "has to happen" after CLM load.

examples/rp/Cargo.toml
```diff
@@ -53,6 +53,8 @@ pio-proc = "0.2"
 pio = "0.2.1"
 rand = { version = "0.8.5", default-features = false }
 
+bleps = { path = "../../../bleps/bleps", features = ["macros", "async", "defmt"] }
```
brandonros, line 56: will update this once https://github.com/bjoernQ/bleps/pull/22/files is merged

brandonros:
going to close this in favor of somoething a bit easier to understnad/more clean

PR #1820, opened 2023-08-24
cyw43: bluetooth hci support

brandonros:
i don't know how to finish this up given that:

`runner` calls `bus.init()` then takes ownership of it

then

`runner` has a `select3!` thing going on that polls `ioctl` + `net_driver_channel` + `bus.status wait for event interrupt` going on, where it runs in its own task

not sure how to attach + expose an HCI connector that would plug into bleps given this


brandonros:
@Dirbaio could you please check the comment description of the PR and suggest how i can "plug" this in/give any guidance? to be clear, we're past cyw43 "unknown just need to poke at it" issues and are at the point of extending existing design

Dirbaio:
@brandonros haven't looked very deep, but I imagine a design similar to `embassy-net-driver-channel` could work: allocate packet queues for rx/tx (either a [`Pipe`](https://docs.embassy.dev/embassy-sync/git/default/pipe/struct.Pipe.html) if it's a byte stream like an uart, or a [`zerocopy_channel::Channel`](https://docs.embassy.dev/embassy-sync/git/default/zerocopy_channel/index.html) if it's packet-based.

This allows the cyw43 task to transfer packets between the queues and the cyw43, so a separate BLE task can send/receive packets and handle them with the BLE stack.

cyw43/src/bus.rs
```diff
@@ -48,44 +48,90 @@ where
...
+                | 0x4 << (8 * REG_BUS_RESPONSE_DELAY)
+                | INTR_WITH_STATUS << (8 * REG_BUS_STATUS_ENABLE),
         )
         .await;
 
+        debug!("read REG_BUS_CTRL");
         let val = self.read8(FUNC_BUS, REG_BUS_CTRL).await;
         trace!("{:#b}", val);
 
+        // TODO: C doesn't do this? i doubt it messes anything up
```
kbleeke, line 96: This is bascially just a test to ensure that the write to REG_BUS_CTRL sets settings (e.g. endianess) as expected.

cyw43/src/bus.rs
```diff
@@ -48,44 +48,90 @@ where
...
+        debug!("Make sure error interrupt bits are clear");
+        self.write8(
+            FUNC_BUS,
+            REG_BUS_INTERRUPT,
+            (IRQ_DATA_UNAVAILABLE | IRQ_COMMAND_ERROR | IRQ_DATA_ERROR | IRQ_F1_OVERFLOW) as u8,
+        )
+        .await;
+
+        // Enable a selection of interrupts
+        // TODO: why not all of these F2_F3_FIFO_RD_UNDERFLOW | F2_F3_FIFO_WR_OVERFLOW | COMMAND_ERROR | DATA_ERROR | F2_PACKET_AVAILABLE | F1_OVERFLOW | F1_INTR
```
kbleeke, line 123: No point in enabling interrupts that we don't handle. 

When I initially implemented this, I just needed IRQ_F2_PACKET_AVAILABLE to stop busy-polling the status register ;)

cyw43/src/bus.rs
```diff
@@ -48,44 +48,90 @@ where
...
+        debug!("read REG_BUS_TEST_RW");
         let val = self.read32(FUNC_BUS, REG_BUS_TEST_RW).await;
         trace!("{:#x}", val);
         assert_eq!(val, TEST_PATTERN);
+
+        debug!("write SPI_RESP_DELAY_F1 CYW43_BACKPLANE_READ_PAD_LEN_BYTES");
+        self.write8(FUNC_BUS, SPI_RESP_DELAY_F1, WHD_BUS_SPI_BACKPLANE_READ_PADD_SIZE)
+            .await;
+
+        // TODO: Make sure error interrupt bits are clear?
```
kbleeke, line 112: Seems reasonable, didn't I do this somewhere?

cyw43/src/control.rs
```diff
@@ -78,64 +79,70 @@ impl<'a> Control<'a> {
...
-        self.ioctl_set_u32(142, 0, 0).await; // SET_BAND = any
+            Timer::after(Duration::from_millis(100)).await;
 
-        Timer::after(Duration::from_millis(100)).await;
+            self.state_ch.set_ethernet_address(mac_addr);
+        }
 
-        self.state_ch.set_ethernet_address(mac_addr);
+        if bluetooth_enabled {
+            // TODO: call runner.init_bluetooth somehow and pass it bluetooth_firmware?
```
kbleeke, line 142: I assume this is here because it has to be done after `load_clm`? 

If there will more bluetooth commands in the future (like turning it off, changing some settings) that are not HCI, you could use a similar construct as the Ioctl module. 

This is basicalily how the interaction works. You allocate some space for the response, send an event to the runner. And then wait for the runner to call your waker once the response to the operation is ready.

cyw43/src/hci_connector.rs
```diff
@@ -0,0 +1,36 @@
...
+impl embedded_io_async::Read for HciConnector {
+    async fn read(&mut self, buf: &mut [u8]) -> Result<usize, HciConnectorError> {
+        // TODO: how to get all the way to runner.hci_read()?
+        Ok(0)
+    }
+}
+
+impl embedded_io_async::Write for HciConnector {
+    async fn write(&mut self, buf: &[u8]) -> Result<usize, HciConnectorError> {
+        // TODO: how to get all the way to runner.hci_write()?
```
kbleeke, line 33: I don't know anything about HCI but the general idea would probably be to turn the select3 in the runner into select4 with another channel for bluetooth output. So bytes that are to be written here are pushed into a channel. The runner is then awoken on this channel and do the `hci_write`.

For reads, I assume it starts with an interrupt? After an interrupt, you check why you were interrupted. Previously there was basically only wifi data available. So then you would need another channel/buffer there that you push the data into. The other side would then be woken up and can consume the bytes.

cyw43/src/bus.rs
```diff
@@ -48,44 +48,90 @@ where
...
         trace!("{:#x}", val);
         assert_eq!(val, TEST_PATTERN);
 
-        let val = self.read32_swapped(REG_BUS_CTRL).await;
+        debug!("read REG_BUS_CTRL");
+        let val = self.read32_swapped(FUNC_BUS, REG_BUS_CTRL).await;
         trace!("{:#010b}", (val & 0xff));
 
         // 32-bit word length, little endian (which is the default endianess).
+        // TODO: C library is uint32_t val = WORD_LENGTH_32 | HIGH_SPEED_MODE| ENDIAN_BIG | INTERRUPT_POLARITY_HIGH | WAKE_UP | 0x4 << (8 * SPI_RESPONSE_DELAY) | INTR_WITH_STATUS << (8 * SPI_STATUS_ENABLE);
```
kbleeke, line 78: There was probably a reason why I added INTERRUPT_WITH_STATUS even though it might seem unnecessary. I remember, that I would sometimes not get woken up without this setting.

cyw43/src/runner.rs
```diff
@@ -222,6 +274,261 @@ where
...
+        let b2h_buf_in_addr = wlan_ram_base_addr + BTSDIO_OFFSET_BT2HOST_IN;
+        let b2h_buf_out_addr = wlan_ram_base_addr + BTSDIO_OFFSET_BT2HOST_OUT;
+        self.bus.bp_write32(h2b_buf_in_addr, 0).await;
+        self.bus.bp_write32(h2b_buf_out_addr, 0).await;
+        self.bus.bp_write32(b2h_buf_in_addr, 0).await;
+        self.bus.bp_write32(b2h_buf_out_addr, 0).await;
+    }
+
+    async fn bt_bus_request(&mut self) {
+        // TODO: CYW43_THREAD_ENTER mutex?
```
kbleeke, line 463: We don't need any mutex here because we cannot get interrupted by the device while doing these operations. (yay cooperative multitasking).

About the interrupt disable, I don't know. If leaving them enabled doesn't corrupt the internal state of the chip, it probably doesn't need to be disabled. We only consume interrupts when we are back on the `select3` call.

kbleeke (commented):
Hi! I worked on this driver a bit earlier this year (interrupts among other things). Very excited to see bluetooth support. 

I left some comments with my unterstanding of the driver/chip, I hope it helps.

isaac-mcfadyen:
Hey! Any update on this PR? Very excited for BLE support. 😄

brandonros:
probably need somebody else to reattempt this better/nicer/cleaner/get it working

PR #1823, opened 2023-08-25
RFC: Introduce type state based control for cyw43.

Frostie314159:
This PR primarily aims at preventing users from calling functions on the `Control` struct, which would be invalid at that point in time. As a second goal, we could introduce functions to leave a WiFi network and stop hosting an AP. All the changes currently made should have zero impact on performance and could in certain cases even improve it, by moving checks which were previously executed at run-time, back to compile-time.

### DIFF
--- a/cyw43/src/control.rs
+++ b/cyw43/src/control.rs
@@ -1,9 +1,11 @@
 use core::cmp::{max, min};
+use core::marker::PhantomData;
 
 use ch::driver::LinkState;
 use embassy_net_driver_channel as ch;
 use embassy_time::{Duration, Timer};
 
+use self::control_states::{Associated, Down, HostingAP, Idle, Uninit};
 pub use crate::bus::SpiBusCyw43;
 use crate::consts::*;
 use crate::events::{Event, EventSubscriber, Events};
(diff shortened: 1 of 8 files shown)


Dirbaio:
I don't think adding typestates to Control is something we want. Yes, it adds some safety, but it requires workarounds in many cases. 

- If you want to e.g. control WiFi from one task and GPIO from another, sharing the Control with a Mutex,
- If you want to dynamically switch between AP and STA with e.g. a button press
- if you want some "network manager" that watches connection state and dynamically tries scanning, connecting to known networks, reconnecting when connection goes down.

with these you'll likely want to store the Control in a struct. With the current API it's trivial, but with typestates you'll have to do workarounds like `enum MyControl{ Idle(Control<Idle>), Associated(Control<Associated>), None }` and dynamically take the Control in/out of that enum when you want to change state, which requires panicking asserts/unwraps anyway.

Also, the cyw43 supports doing one STA and one AP concurrently. It's not implemented in this driver yet, but when we do implement it, Control won't neatly fit into one "state" anymore.

There's *one* case where I'd be OK with typestates, which is for forcing the user to do `init()` before anything else. The above downsides don't apply, because you pretty much *always* immediately init, and you never dynamically change states between init/uninit. If we do this I'd do two structs `UninitControl` and `Control` for simplicity though, instead of a typestate generic param.

Frostie314159:
Ok. I definitely understand the concerns you have. My suggestion would be to do something similar to the `AnyPin` struct in the hal. In practice this could mean creating a `AnyControl` struct.

PR #1826, opened 2023-08-26
Avoid dead code warning

bugadani:
This warning snuck in with #1774 and went unnoticed since. The constant is not needed if `executor-thread` is not active, so we can just move it into the `thread` module.


Dirbaio:
shame CI didn't catch it. Could you add to `ci.sh` all (reasonable) feature/arch combinations so it's caught in the future? Thank you :)

bugadani:
I didn't add _all_ but probably more than you meant.

Do you want to install an xtensa-compatible compiler on the runner by any chance? :P

Dirbaio:
> Do you want to install an xtensa-compatible compiler on the runner by any chance? :P

maybe later :see_no_evil:

bugadani:
> shame CI didn't catch it. Could you add to `ci.sh` all (reasonable) feature/arch combinations so it's caught in the future? Thank you :)

Either you had a good hunch with this, or you noticed I missed the `pub(super)`. Either way, good call!

Dirbaio (approved):
no, I meant shame the CI didn't catch the original warning (it denies warnings). hadn't checked the code yet.

PR #1829, opened 2023-08-27
Fix DateTime Conversion and RTC Example

aidant:
### Problem 

The `day_of_week_from_u8` implementation for `stm32` expects a number 0-6 which represents Mon-Sun, the current implementation for converting `chrono::Weekday` to `DayOfWeek` uses the `chrono` function `number_from_monday` which returns 1-7 representing Mon-Sun. This causes the following error to be thrown on Sundays and every other day the weekday to be off-by-one.

```
called `Result::unwrap()` on an `Err` value: InvalidDayOfWeek(7)
```

### Solution

Replace the `number_from_monday` function call with `num_days_from_monday` which returns the expected 0-6 representing Mon-Sun.

### Problem 

The example RTC code hangs on the line `let mut rtc = Rtc::new(p.RTC, RtcConfig::default());`.

### Solution

Configure the RTC clock source to use `LSI`.

### DIFF
--- a/embassy-stm32/src/rtc/datetime.rs
+++ b/embassy-stm32/src/rtc/datetime.rs
@@ -89,7 +89,7 @@ pub enum DayOfWeek {
 #[cfg(feature = "chrono")]
 impl From<chrono::Weekday> for DayOfWeek {
     fn from(weekday: Weekday) -> Self {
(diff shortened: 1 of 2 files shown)


Dirbaio:
LGTM. Can you fix CI? needs running rustfmt.

PR #1834, opened 2023-08-28
Fix timing on RP2040 pio_ws2812.rs example

dave-andersen:
The example spins too fast so it doesn't appear to change; it's delaying for microseconds instead of milliseconds. This commit slows it down and adds a comment noting the pin mapping for the Adafruit feather rp2040+RFM95 LoRA module, which has its Neopixel on pin 4 instead of 16.

### DIFF
--- a/examples/rp/src/bin/pio_ws2812.rs
+++ b/examples/rp/src/bin/pio_ws2812.rs
@@ -138,8 +138,9 @@ async fn main(_spawner: Spawner) {
     const NUM_LEDS: usize = 1;
     let mut data = [RGB8::default(); NUM_LEDS];
 
-    // For the thing plus, use pin 8
-    // For the feather, use pin 16
+    // Common neopixel pins:
+    // Thing plus: 8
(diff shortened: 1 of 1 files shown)


CBJamo:
On my boards (I have a thing plus and a feather) it flashes between the colors very quickly, but the color does change. If it's not changing at all that might hint at a deeper bug, can you give more detail about what's happening?

Besides that, this change does make the color a nicer pulsing color so I don't see a problem with it.

CBJamo:
Hmm, there is a lot of variation in ws2812 and ws2812-compatible parts. I wonder if the right thing to do would be to include the 50us delay in the write function itself, that way we the write function ensures the bus is a good state. Then replace the timer in the main loop with a ticker to make the color change timing consistent.

Dirbaio:
Since it seems this fixes the example on at least some neopixels, I'm going to merge. I agree with @CBJamo though, there should be a delay in `write` so it always works regardless of what delays the user uses.

PR #1835, opened 2023-08-28
stm32: fix wait for RNG data

oll3:
If no data was available to read then the loop would wait for an interrupt and skip to the next chunk without writing the current one. This could cause the given slice to only be partially filled with random data.

Fixed by moving the wait to before actually writing data to the chunk.


lucasgranberg (approved):
This makes sense to me and would avoid a hard to find rng fault.

PR #1838, opened 2023-08-28
cyw43: Add utility functions.

Frostie314159:
In its current state, this PR introduces separate up/down functions, which are just used internally for now. As well as introducing a function for leaving an infra.

### DIFF
--- a/cyw43/src/events.rs
+++ b/cyw43/src/events.rs
@@ -1,5 +1,5 @@
 #![allow(dead_code)]
-#![allow(non_camel_case_types)]
+#![allow(non_camel_case_types, non_upper_case_globals)]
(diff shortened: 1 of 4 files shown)


cyw43/src/events.rs
```diff
@@ -1,5 +1,5 @@
 #![allow(dead_code)]
-#![allow(non_camel_case_types)]
+#![allow(non_camel_case_types, non_upper_case_globals)]
```
Dirbaio, line 2: ths is a false positive in `rust-analyzer`, we shouldn't workaround it in the repo.

Dirbaio (approved):
thank you!

PR #1844, opened 2023-08-30
stm32/rtc: autocompute prescalers

xoviat:
this adjusts the prescalers appropriately if the lsi or hse is used.


xoviat:
failed merge because of pico

PR #1846, opened 2023-08-30
Stm32 timer prevent hardfault when frequency is 0

diondokter:
A frequency of 0 is wrong, but it shouldn't hardfault.
Now we simply panic instead.


Dirbaio (approved):
:+1: :ok_hand: :pray:

PR #1847, opened 2023-08-30
RP2040: XOSC delay multiplier

mvniekerk:


embassy-rp/src/clocks.rs
```diff
@@ -624,12 +626,12 @@ pub fn clk_rtc_freq() -> u16 {
...
 }
 
-fn start_xosc(crystal_hz: u32) {
+fn start_xosc(crystal_hz: u32, delay_multiplier: u32) {
     pac::XOSC
         .ctrl()
         .write(|w| w.set_freq_range(pac::xosc::vals::CtrlFreqRange::_1_15MHZ));
 
-    let startup_delay = ((crystal_hz / 1000) + 128) / 256;
+    let startup_delay = (((crystal_hz / 1000) + 128) * delay_multiplier) / 256;
```
Dirbaio, line 634: The `(x + 128)/256` is to divide by 256 rounding to nearest. I'd try to keep it: `((crystal_hz / 1000 + delay_multiplier) + 128) / 256`

embassy-rp/src/clocks.rs
```diff
@@ -94,6 +94,7 @@ impl ClockConfig {
                     post_div1: 6,
                     post_div2: 5,
                 }),
+                delay_multiplier: 1,
```
Dirbaio, line 97: perhaps we should default to a higher multiplier so that it works on all boards out of the box? so the user can lower it explicitly if they know their board's xosc stabilizes faster and they care about boot speed

embassy-rp/src/clocks.rs
mvniekerk: Check f69e845 for improvement

embassy-rp/src/clocks.rs
mvniekerk: Check f69e845 - made it 128

embassy-rp/src/clocks.rs
Dirbaio: oops sorry! I wrote `+ delay_multiplier` but it should be `* delay_multiplier`

Dirbaio (approved):
thank you!

PR #1850, opened 2023-08-31
rp2040: I2cDevice

CBJamo:
This was a cavalcade of race conditions, confusing interrupts, and pain, but I'm pretty sure it's ready to go.

I'll whip up a use example soon-ish and put that in another PR.


Dirbaio:
Looks good overall! i'd do just a few changes:

- It should be called i2c slave (in the HALs we use the vendor names, and that's the name Raspberry Pi uses in the datasheet and SDK)
- i'd split it into two top-level modules `i2c` and `i2c_slave`, so docs are less cluttered, for example if you look at `i2c` you see only the stuff needed for using i2c in master mode whcih is what most people do. Also this way it matches `embassy-nrf`

CBJamo:
Renamed and split per suggestion. I also added a simple use example and did a little cleanup.

Ps. Sorry about the git noise.

Dirbaio (approved):
Looks great! Thanks for all your work :rocket:

PR #1854, opened 2023-09-02
embassy-{net, sync, time}: Use fmt::unwrap

bugadani:
This PR tries to avoid including panic strings in the binary by preferring defmt::unwrap if available.


Dirbaio (approved):
thanks!

bugadani:
> assertion failed: down > TEST_EXPECTED_DOWNLOAD_KBPS

was that me?

PR #1860, opened 2023-09-04
initial support for STM32G4 ADC

daehyeok:
I also create PR embassy-rs/stm32-data#244 for STM32G4 ADC.
 but my stm32-data PR will break embassy-stm32 build until this PR is merged.

 What is the recommended workflow for this case?

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -58,7 +58,7 @@ sdio-host = "0.5.0"
 embedded-sdmmc = { git = "https://github.com/embassy-rs/embedded-sdmmc-rs", rev = "a4f293d3a6f72158385f79c98634cb8a14d0d2fc", optional = true }
 critical-section = "1.1"
 atomic-polyfill = "1.0.1"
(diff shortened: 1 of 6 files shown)


xoviat:
Great job working on this! I will get to this ASAP!

xoviat:
The path forward for this is to: 
1. Update the stm32-data version in this PR so that it passes the ci. This can be done by updating the tag in Cargo.toml.
2. The manual implementation of RccPeripheral and the dependency on paste needs to be removed and the refcount needs to be updated to use the new refcount statics. You can cherry-pick and adjust this commit: https://github.com/embassy-rs/embassy/pull/1849/commits/56b1607d98f785508d992b622aa4557e1161da73, or wait for the adc f3 pr to be merged and then pull in those changes.

If you're unable to do this, I can take this forward when I get a chance, as I am eager to see adc implemented for g4.

daehyeok:
Cargo.toml is updated and passed CI. I wil check f3 PR and update code  later.
 
P.S Do we have some contribution guide? I can't find how run CI test on local before upload commit.

xoviat:
I'm going to merge #1867 which will provide ADC12 on g4 but not ADC345. For ADC345, stm32-data-gen rcc.rs needs to updated to generate rcc for those peripherals on g4 (embassy-rs/stm32-data#247).

PR #1862, opened 2023-09-04
stm32: refcount peripheral enable/disable

xoviat:
This appears to work as designed, but the peripherals may currently leak refcounts. That will need to be examined before this can  be merged.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -308,7 +308,8 @@ fn main() {
     // ========
     // Generate RccPeripheral impls
 
-    let refcounted_peripherals = HashSet::from(["USART", "SPI", "I2C"]);
+    // TODO: maybe get this from peripheral kind? Not sure
+    let refcounted_peripherals = HashSet::from(["UART", "USART", "SPI", "I2C"]);
     let mut refcount_statics = HashSet::new();
 
     for p in METADATA.peripherals {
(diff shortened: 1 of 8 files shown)


embassy-stm32/build.rs
```diff
@@ -379,6 +413,21 @@ fn main() {
...
+    let mut refcount_mod = TokenStream::new();
+    for refcount_static in refcount_statics {
+        refcount_mod.extend(quote! {
+            pub(crate) static #refcount_static: AtomicU8 = AtomicU8::new(0);
+        });
+    }
+
+    g.extend(quote! {
+        mod refcount_statics {
+            use atomic_polyfill::AtomicU8;
```
Dirbaio, line 425: `atomic_polyfill` is deprecated, I would try to not introduce more uses of it.

Since the refcounting always happens within a critical section, you can use `static mut` with no atomic instructions at all.

xoviat:
skipping the adc refcount so that I can handle it later

embassy-stm32/build.rs
```diff
@@ -308,6 +308,10 @@ fn main() {
     // ========
     // Generate RccPeripheral impls
 
+    // TODO: maybe get this from peripheral kind? Not sure
+    let refcounted_peripherals = HashSet::from(["UART", "USART", "SPI", "I2C"]);
```
Dirbaio, line 312: why refcount spi, i2c? these have no split.
also LPUARTs should be refcounted I think

xoviat: lpuarts are already included

PR #1864, opened 2023-09-05
ADC support for STM32WLx

oll3:
Waiting for https://github.com/embassy-rs/stm32-data/pull/246.

Will also remove the use of adccommon_v3 CCR register for STM32G0x. Since the CCR register definition for this device seems to have some bits/names swapped (VBATEN/TSEN) compared to the common one.

### DIFF
--- a/embassy-stm32/src/rcc/wl.rs
+++ b/embassy-stm32/src/rcc/wl.rs
@@ -1,4 +1,5 @@
 pub use super::bus::{AHBPrescaler, APBPrescaler, VoltageScale};
+use crate::pac::rcc::vals::Adcsel;
 use crate::pac::{FLASH, PWR, RCC};
 use crate::rcc::bd::{BackupDomain, RtcClockSource};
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/rcc/wl.rs
```diff
@@ -116,6 +140,7 @@ pub struct Config {
     pub enable_lsi: bool,
     pub enable_rtc_apb: bool,
     pub rtc_mux: RtcClockSource,
+    pub adc_clock_source: AdcClockSource,
```
xoviat, line 143: The adc clock source should be configured in the rcc mod and the clock frequency added to the clocks adc struct.

xoviat (changes requested):

xoviat:
Can you rebase and update the metapac version?

oll3:
> Can you rebase and update the metapac version?

Yep, rebased and bumped stm32-metapac!

PR #1870, opened 2023-09-06
nrf: Use SetConfig trait internally where possible

plaes:
We can call SetConfig's set_config internally reducing code duplication and possibly reduce hard-to-detect issues with device reconfiguration in the future.


Dirbaio (approved):
thank you!

Issue #460, opened 2021-10-26
Mitigate nrf52832 dma errata


jacobrosenthal:
https://infocenter.nordicsemi.com/pdf/nRF52_PAN_109_add_v1.1.pdf

plaes:
Anomaly 109 - DMA access transfers might be corrupted
Affects following peripherals on nrf52832:
* PWM
* SPIS TX
* SPIM TX
* TWIS
* UARTE
* TWIM TX

PR #1871, opened 2023-09-07
nrf: spim: Anomaly 109 workaround for SPIM peripheral (#460)

plaes:
This implements SPIM TX workaround suggested from section 3.8.1 from Anomaly 109 addendum.

In workaround case we first keep track of original maxcnt values, then initiate "fake" transfer with zero-length maxcnt values. Once the "fake" transfer is triggered, we handle it, fill in the original maxcnt values and restart the transmission.


Dirbaio (approved):
thanks!

PR #1872, opened 2023-09-07
Remove feature(impl_trait_projections) from async traits.

Dirbaio:
It's possible `async_fn_in_traits` will get stabilized before `impl_trait_projections` (see [zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/187312-wg-async/topic/sync.20meeting.202023-09-07)). 

Trying out whether it's feasible to use `async_fn_in_traits` only. Do not merge yet. Since it makes the code a bit uglier, we should only merge this if indeed `async_fn_in_traits` gets stabilized before `impl_trait_projections`.

EDIT: maybe `impl_trait_projections` does get stabilized before AFIT: https://github.com/rust-lang/rust/pull/115659


Dirbaio:
bender run

Dirbaio:
`impl_trait_projections` will be stable on 1.75, so this is not needed anymore.

PR #1874, opened 2023-09-08
embassy-net: add async wait_config_up

JuliDi:
This PR adds a function `wait_config_up` to the embassy_net `Stack`.
It will check whether `is_config_up` is true and if not register a waker that is woken in `apply_static_config` (e.g. by DHCP).

I added a not-working example to the docstring. Not sure whether we can make this run somehow.

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -365,7 +365,7 @@ impl<D: Driver + 'static> Stack<D> {
         v4_up || v6_up
     }
 
-    /// Get for the network stack to obtainer a valid IP configuration.
+    /// Wait for the network stack to obtaine a valid IP configuration.
     pub async fn wait_config_up(&self) {
         if self.is_config_up() {
             return;
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -226,6 +226,7 @@ struct Inner<D: Driver> {
     static_v6: Option<StaticConfigV6>,
     #[cfg(feature = "dhcpv4")]
     dhcp_socket: Option<SocketHandle>,
+    config_waker: WakerRegistration,
     #[cfg(feature = "dns")]
     dns_socket: SocketHandle,
     #[cfg(feature = "dns")]
@@ -297,6 +298,7 @@ impl<D: Driver + 'static> Stack<D> {
             static_v6: None,
             #[cfg(feature = "dhcpv4")]
             dhcp_socket: None,
+            config_waker: WakerRegistration::new(),
             #[cfg(feature = "dns")]
             dns_socket: socket.sockets.add(dns::Socket::new(
                 &[],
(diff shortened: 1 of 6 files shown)


embassy-net/src/lib.rs
```diff
@@ -363,6 +366,56 @@ impl<D: Driver + 'static> Stack<D> {
         v4_up || v6_up
     }
 
+    /// Wait for the network stack to obtain a valid IP configuration.
+    /// Returns instantly if [`Stack::is_config_up`] returns `true`.
+    ///
+    /// ## Watch out:
```
MabezDev, line 372: I think this comment can be removed, or perhaps simplified to something like "Ensure `Stack::run` has been called" and defer explaining why to  `run`'s documentation.

JuliDi:
Updated examples are tested on STM32H743

embassy-net/src/lib.rs
```diff
@@ -363,6 +366,56 @@ impl<D: Driver + 'static> Stack<D> {
...
+    /// The Future is polled only when the [`Stack`] is running,
+    /// e.g. call `spawner.spawn(net_task(stack))`.
+    ///
+    /// `await`ing before will never yield!
```
MabezDev, line 376: There is a chance it will never yield anyway, for example in the case of DHCP. Maybe we should reword this to say as much, and that the end user should handle the timeout?

MabezDev (commented):
Just a small drive-by review, feel free to ignore my comments. 

Very happy to see this added though, always bothered me when testing esp-wifi that the blocking versions resolved DHCP quicker :D.

embassy-net/src/lib.rs
JuliDi: That's right. I guess if DHCP does not assign an IP, the user will have to handle it anyway. I'll think of something and maybe add an example.

JuliDi: I've rephrased the docstring. Do you think it is better now?

embassy-net/src/lib.rs
```diff
@@ -363,6 +366,55 @@ impl<D: Driver + 'static> Stack<D> {
         v4_up || v6_up
     }
 
+    /// Wait for the network stack to obtain a valid IP configuration.
+    ///
+    /// ## Notes:
+    /// - Ensure [`Stack::run`] has been called before using this function.
+    ///
+    /// - This function may never yield (e.g. if no configuration is obtained through DHCP).
```
Dirbaio, line 374:
```suggestion
    /// - This function may never return (e.g. if no configuration is obtained through DHCP).
```

"yield" in async means the future suspending on an .await, not returning.

JuliDi: good catch, thanks

Dirbaio (approved):
thanks!

PR #1880, opened 2023-09-10
rp2040: BOOTSEL button support

phire:
This module adds support for polling the BOOTSEL button found on the Pi Pico and most other rp2040 boards.

~~It's theoretically possible to create a fully safe interface, but it would be quite invasive (it needs to consume and wrap all DMA channels, CORE1 and FLASH). I don't think it's worth the effort.~~

### DIFF
--- a/embassy-rp/src/bootsel.rs
+++ b/embassy-rp/src/bootsel.rs
@@ -1,0 +1,81 @@
+//! Boot Select button
+//!
+//! The RP2040 rom supports a BOOTSEL button that is used to enter the USB bootloader
+//! if held during reset. To avoid wasting GPIO pins, the button is multiplexed onto
+//! the CS pin of the QSPI flash, but that makes it somewhat expensive and complicated
+//! to utilize outside of the rom's bootloader.
+//!
+//! This module provides functionality to poll BOOTSEL from an embassy application.
+
+use crate::flash::in_ram;
+
+/// Polls the BOOTSEL button. Returns true if the button is pressed.
+///
+/// Polling isn't cheap, as this function waits for core 1 to finish it's current
+/// task and for any DMAs from flash to complete
+pub fn poll_bootsel() -> bool {
+    let mut cs_status = Default::default();
+
+    unsafe { in_ram(|| cs_status = ram_helpers::read_cs_status()) }.expect("Must be called from Core 0");
+
+    // bootsel is active low, so invert
+    !cs_status.infrompad()
+}
+
+mod ram_helpers {
+    use rp_pac::io::regs::GpioStatus;
+
+    /// Temporally reconfigures the CS gpio and returns the GpioStatus.
+
+    /// This function runs from RAM so it can disable flash XIP.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure flash is idle and will remain idle.
+    /// This function must live in ram. It uses inline asm to avoid any
+    /// potential calls to ABI functions that might be in flash.
+    #[inline(never)]
+    #[link_section = ".data.ram_func"]
+    #[cfg(target_arch = "arm")]
+    pub unsafe fn read_cs_status() -> GpioStatus {
+        let result: u32;
+
+        // Magic value, used as both OEOVER::DISABLE and delay loop counter
+        let magic = 0x2000;
+
+        core::arch::asm!(
+            ".equiv GPIO_STATUS, 0x0",
+            ".equiv GPIO_CTRL,   0x4",
+
+            "ldr {orig_ctrl}, [{cs_gpio}, $GPIO_CTRL]",
+
+            // The BOOTSEL pulls the flash's CS line low though a 1K resistor.
+            // this is weak enough to avoid disrupting normal operation.
+            // But, if we disable CS's output drive and allow it to float...
+            "str {val}, [{cs_gpio}, $GPIO_CTRL]",
+
+            // ...then wait for the state to settle...
+            "1:", // ~4000 cycle delay loop
+            "subs {val}, #8",
+            "bne 1b",
+
+            // ...we can read the current state of bootsel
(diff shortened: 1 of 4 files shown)


phire:
Or am I confused about where the safety boundary is?

I guess it's not possible to start code on CORE1 or any DMA transfer at all without an unsafe block. I notice that `embassy-rp::flash`'s `in_ram` function already assumes that stopping core1 and waiting for outstanding DMA transfers is enough to ensure safety.

Dirbaio:
Thanks for the PR! Some feedback:

- Flash code has a helper called [`in_ram`](https://github.com/embassy-rs/embassy/blob/7d6edd7b15d2209ac0b96ff8814ecefce2964e36/embassy-rp/src/flash.rs#L206) that does all the "prepare both cores so we can stop XIP" preparations. Reading bootsel should call that instead of reimplementing it.
- Even if you make the function `#[link_section = ".data.ram_func"]`, it's possible the functions you call (for example `cortex_m::asm::delay`) are in flash. Even if you write it so it doesn't call any function, the compiler can insert function calls for you (for example to `memcpy`, or the `_aeabi_*` intrinsics, or to Rust stuff like `drop` etc). From fighting these issues in the flash code, the conclusion was that the only way to avoid it is to [write it in ASM](https://github.com/embassy-rs/embassy/blob/7d6edd7b15d2209ac0b96ff8814ecefce2964e36/embassy-rp/src/flash.rs#L756). For the same reasons, this one should be written in ASM too.

Unfortunately everything that has to do with rp2040 XIP is extremely tricky... :smiling_face_with_tear: 

> I notice that embassy-rp::flash's in_ram function already assumes that stopping core1 and waiting for outstanding DMA transfers is enough to ensure safety.

Yes, with these checks, reading bootsel doesn't need to be `unsafe`.

phire:
Hmm... I missed that `cortex_m::asm::delay` is only marked `#[inline]` instead of `#[inline(always)]`. 

And I didn't think about the fact that the compiler might insert extra function calls. That's pretty infuriating. 

> the conclusion was that the only way to avoid it is to [write it in ASM](https://github.com/embassy-rs/embassy/blob/7d6edd7b15d2209ac0b96ff8814ecefce2964e36/embassy-rp/src/flash.rs#L756).

Surely there is a better way. Could we put all ram functions into one or more static libraries and then use the linker to put the entire static library in ram? Then any leaf functions of these ram_funcs will be safely duplicated into RAM.  

> Unfortunately everything that has to do with rp2040 XIP is extremely tricky... 🥲

I've been thinking about this and have a proposal. 
If it was just bootsel I wouldn't worry. Just warn people away from using it. But this problem affects flash too. 

The problem stems from the fact that XIP is a peripheral too. A peripheral that conflicts with FLASH and is implicitly used from boot to enable reading code/data from flash.

So we should probably be representing it as an actual peripheral. When the `run-from-ram` feature is enabled, it will be in the main peripheral struct with everything else, and any code that wishes to use XIP or FLASH can do so with zero runtime overhead.

Otherwise, `embassy-rp` will  take ownership of XIP at boot, and expose a RefCell like API  to temporally borrow XIP. Immutable borrows can be used to soundly implement DMAs via the XIP address range or XIP's streaming interface. Accessing flash or bootsel directly will require XIP to be borrowed mutably, probably with a critical-section style API that pauses core1, disables interrupts and throws an error if there are outstanding immutable borrows.

This has the added benefit that `in_ram` no longer needs to check for outstanding DMAs.

Then we can refactor `embassy-rp::flash`. We should create a read-only version of flash that only does streaming reads via XIP. That way, an `XipCriticalSection` is only needed for writing to flash, or checking bootsel.

Dirbaio:
> Could we put all ram functions into one or more static libraries and then use the linker to put the entire static library in ram? Then any leaf functions of these ram_funcs will be safely duplicated into RAM.

that could work, but I think it'd be a bit fiddly to get right. 
- you'd have to call `rustc` from `build.rs` to build this staticlib separately
- i'm not sure if linking the staticlib into the final binary will work. You'll run into duplicate symbol errors, or you'll have to fiddle with link order to get the ram symbols to take precedence over the others.
- alternatively compile to a binary, convert it to bytes and include them in the final binary as data. This requires knowing the final addres though...

would be cool to try, I agree writing ASM by hand sucks.

> So we should probably be representing it as an actual peripheral. 

I don't think this'll be very ergonomic for big real-world projects. You'd have to write code to coordinate between the task that wants to write to flash and all tasks that might be using DMA. The flash task would have to ask them to stop and notify back when stopped, then proceed to flash. If you do something wrong in this process, you get a runtime panic from the RefCell.

phire:
> I don't think this'll be very ergonomic for big real-world projects. 

I was thinking embassy-rp would provide an optional "dma-manager" that handled the complexity of tracking in-progress DMAs from flash. The static typing would allow the programmer zero-cost access to either DMAs from XIP or direct flash access without the "dma-manager", it would only be needed if the programmer wanted both.  
Static typing can probably avoid the potential for panics too.

However... With further reflection, I'm really not sure if the concept of DMAing from XIP can ever be useful enough to expose via safe APIs. The timings are unpredictable depending on a cache hit, and usually slower than other methods. I can't really think of a use case where a programmer wouldn't be better off using streaming background transfers or disabling XIP and using proper QSPI.

Maybe we should just proactively declare that DMAs from XIP memory will never be sound. That would have the added benefit of removing the need to check for any active XIP DMAs in `flash::in_ram`

phire:
Anyway, I've rewritten the bootsel implementation based on feedback. 

It reuses the in_ram helper from flash and I've rewritten the ram function with inline assembly.

Dirbaio:
LGTM, thank you! :)

could you add a test [here](https://github.com/embassy-rs/embassy/tree/main/tests/rp/src/bin) that calls the function and asserts it returns false? No way to check it returns true when pressed, but at least it'll check it doesn't crash.

PR #1881, opened 2023-09-10
Handle stbiterr in 4bit wide mode for sdmmc_v1

MabezDev:
Draft until https://github.com/embassy-rs/stm32-data/pull/250 is merged.


xoviat:
Can you rebase and update metapac?

MabezDev:
I think I've updated the tag, but CI still seems to be failing.

PR #1883, opened 2023-09-10
embassy-net: Improve error when socket is not bound

JuliDi:
When I forgot to call `bind` before sending, I received a "NoRoute" error, which was confusing.
This PR attempts to provide a more helpful error in this case.

Maybe there are other cases where this may happen. If so, please let me know so I can add them.


Dirbaio (approved):
thanks!

PR #1892, opened 2023-09-12
stm32/usart: fix read_until_idle not wake up issue

cumthugo:
When use `RartRx::read_until_idle` to read from uart, sometime it can't wake up. 

**Reason**:  while `abort` future is polling which is triggered by idle interrupt , at the same time the next data is start transferring, the `sr.idle()` return false, and the `abort` future still return `Poll::Pending` state, so it can't be wake up anymore.

**Solution**: Always enable idle interrupt when poll `abort` future

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -545,6 +545,13 @@ impl<'d, T: BasicInstance, RxDma> UartRx<'d, T, RxDma> {
             unsafe { rdr(r).read_volatile() };
             clear_interrupt_flags(r, sr);
 
+            if enable_idle_line_detection {
+                // enable idle interrupt
+                r.cr1().modify(|w| {
+                    w.set_idleie(true);
+                });
+            }
+
             compiler_fence(Ordering::SeqCst);
 
             let has_errors = sr.pe() || sr.fe() || sr.ne() || sr.ore();


embassy-stm32/src/usart/mod.rs
```diff
@@ -518,19 +518,6 @@ impl<'d, T: BasicInstance, RxDma> UartRx<'d, T, RxDma> {
...
-            // This read also clears the error and idle interrupt flags on v1.
-            unsafe { rdr(r).read_volatile() };
-            clear_interrupt_flags(r, sr);
-
-            // enable idle interrupt
-            r.cr1().modify(|w| {
-                w.set_idleie(true);
-            });
-        }
-
```
xoviat, line 533: I'm not sure this code needs to be removed. If errors exist before the method is called, then they need to be cleared.

xoviat (changes requested):

embassy-stm32/src/usart/mod.rs
cumthugo: 1. If errors exist, it will be cleared correctly, the code is located in line 500~503
2. There is no difference when `enable_idle_line_detection` is false
3. When `enable_idle_line_detection` is true, line 529~523 will do the same thing such as  the code I removed.

cumthugo:
@xoviat You're right, the `idle` flag will not be cleared by removing the code.

PR #1893, opened 2023-09-12
Wait for high for DIO1 for rp2040-based LoRa boards

ceekdee:
Handle rp2040-based LoRa boards in the same manner as LoRa boards for other MCUs with regard to detecting an IRQ on the DIO1 pin - wait for high rather than rising edge.

This is PR is a result of analysis within the following issue:

https://github.com/embassy-rs/lora-phy/issues/27.


Dirbaio (approved):
thanks!

PR #1895, opened 2023-09-13
[STM32] Handle STM32H7 "_C" pins

JuliDi:
As discussed in the chat, the `split-pxy` features are introduced for use with https://github.com/embassy-rs/stm32-data/pull/254

Ideas for improvement:
- Verify this actually works with the split functions enabled (the traits are there, but I have no device that actually exposes the PC2/3 pins)
-  Add a testcase / CI step for compilation with these flags. Any suggestions where that should go? Or maybe a HIL test on an H7 board?

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -81,6 +81,16 @@ fn main() {
         singletons.push(c.name.to_string());
     }
 
+    // Extra analog switch pins available on most H7 chips
+    #[cfg(feature = "split-pa0")]
+    singletons.push("PA0_C".into());
+    #[cfg(feature = "split-pa1")]
+    singletons.push("PA1_C".into());
+    #[cfg(feature = "split-pc2")]
+    singletons.push("PC2_C".into());
+    #[cfg(feature = "split-pc3")]
+    singletons.push("PC3_C".into());
+
     // ========
     // Handle time-driver-XXXX features.
 
@@ -670,7 +680,31 @@ fn main() {
                 let key = (regs.kind, pin.signal);
                 if let Some(tr) = signals.get(&key) {
                     let mut peri = format_ident!("{}", p.name);
-                    let pin_name = format_ident!("{}", pin.pin);
+                    let pin_name = {
+                        #[allow(unused_mut)]
+                        let mut pin_name = pin.pin;
+
+                        #[cfg(not(feature = "split-pa0"))]
+                        if pin.pin == "PA0_C" {
+                            pin_name = "PA0";
+                        }
+                        #[cfg(not(feature = "split-pa1"))]
+                        if pin.pin == "PA1_C" {
+                            pin_name = "PA1";
+                        }
+
+                        #[cfg(not(feature = "split-pc2"))]
+                        if pin.pin == "PC2_C" {
+                            pin_name = "PC2";
+                        }
+                        #[cfg(not(feature = "split-pc3"))]
+                        if pin.pin == "PC3_C" {
+                            pin_name = "PC3";
+                        }
+
+                        format_ident!("{}", pin_name)
+                    };
+
                     let af = pin.af.unwrap_or(0);
 
                     // MCO is special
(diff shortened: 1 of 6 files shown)


Dirbaio:
> Add a testcase / CI step for compilation with these flags. Any suggestions where that should go? 

yes, would be good to add a few test cases of building embassy-stm32 with/without the split-xx features to `ci.sh`.

> Or maybe a HIL test on an H7 board?

perhaps later, since we don't even have ADC hil tests for now...

embassy-stm32/build.rs
```diff
@@ -81,6 +81,16 @@ fn main() {
         singletons.push(c.name.to_string());
     }
 
+    // Extra analog switch pins available on most H7 chips
```
Dirbaio, line 84: this should check the pins actually exist for the currently-being-built chip

JuliDi: That would be good. But how?
Enabling the split feature if there is no _C pin should fail asap, anyway. At the latest when modifying the syscfg in init().

embassy-stm32/src/lib.rs
```diff
@@ -180,6 +180,18 @@ pub fn init(config: Config) -> Peripherals {
     }
 
     unsafe {
+        #[cfg(feature = "_split-pins-enabled")]
```
JuliDi, line 183: @Dirbaio this is where it will fail anyway if the user enables a split feature for a wrong chip. Not sure whether we can mitigate this risk.

embassy-stm32/build.rs
Dirbaio: Letting a random register write deep in embassy fail to compile is not great, the compiler error ideally should tell you you've enabled an incompatible set of features.

Currently checking if a pin exists needs scanning all the gpio mappings to see if `PC2_C` or similar is mentioned somewhere. That signals it "exists".

Alternatively we could add a "pins: [...]" array to stm32-data that would list all pins that exist (and perhaps some properties of them, like package location?). This is something I've always wanted to do, because there are chips where only "part" of GPIO ports are implemented, this would allow preserving that. (Currently we assume if GPIOx exists then all Px[0..15] exist). It's a more involved refactor though, so perhaps not for now.

JuliDi: Having package information in there would be ideal. As we know now, some packages even expose the PC2/3 Pins (actually only one, the huge BGA package). And for those, closing the analog switch and splitting the pins should probably not go hand in hand.

But yeah, that's going to be a bigger set of changes. Do you have a suggestion what am intermediate solution/compiler error should look like? (Or would the pin/signal scanning be sufficient already?)

Another quick solution that comes to mind:
We could catch 99% of the cases with something like "if (split feature enabled and not stm32h7 enabled) then throw compiler error".
Still, there are some H7s that have the PA0/1 with analog switch and afaik also some H7 that don't have any. Just the usual STM32 mess...

Dirbaio: > Or would the pin/signal scanning be sufficient already?

yes, that will work nicely. Just collect all seen pin names in a set, then error if `split-pc2` is enabled and the set doesn't contain `PC2_C`.

JuliDi: Alright, I should have some time to do this tomorrow.

JuliDi: Argh, while implementing it I realized that I never properly tested with the split activated. There need to be more changes to how the `pin_table` is generated.

Is this what it is supposed to look like? Then if so, I would put that in a loop.
```rust
                    if pin_name == "PC2" {
                        pins_table.push(vec![
                            "PC2_C".to_string(),
                            p.name.to_string(),
                            port_num.to_string(),
                            "2".to_string(),
                            format!("EXTI{}", 2),
                        ]);
                    }
```

I will push the changes with the "scan pins and put into a set" to check the validity of the features.
The code that iterates over Portletter + `for pin_num in 0u32..16 ` is another story. I can look into it in 2 weeks. For now I suggest leaving the PR open as WIP.

JuliDi:
Won't be able to work on this for the next 2 weeks or so, but I'm happy to receive feedback in the meantime and make a proper implementation later on.

JuliDi:
closed in favor of #1988

PR #1896, opened 2023-09-14
Write to TaskStorage::future via never-inlined fn to encourage RVO

haileys:
In programs that contain large future objects, for example those which declare large buffers as locals in async functions:

```rust
async fn run() -> Result<(), RunError> {
    let mut rx_meta = [PacketMetadata::EMPTY; 16];
    let mut rx_buffer = [0; 4096];

    let mut tx_meta = [PacketMetadata::EMPTY; 16];
    let mut tx_buffer = [0; 4096];

    // ...
```

`TaskPool::spawn_impl` can in some cases use an enormous amount of stack to instantiate the future into a `TaskStorage`:

```sh-session
$ cargo build --release
   Compiling embassy-executor v0.3.0 (/home/hailey/code/embassy/embassy-executor)
   Compiling bark-esp v0.1.0 (/home/hailey/code/bark-esp)
    Finished release [optimized + debuginfo] target(s) in 1.36s
$ stack-usage target/xtensa-esp32-none-elf/release/bark-esp
FUNCTION                                                                               STACK SIZE
embassy_executor::raw::TaskPool<F,_>::spawn_impl::h3954ad8fc673878e                         20304
```

Rust _does_ support [return value optimisation](https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization), and it's unclear to me why it is not applying this optimisation in this case. Sure enough though, the disassembly for this instantiation of `TaskPool::spawn_impl` shows that the future object is returned by the future constructor to the stack before being separately memcpy'd into the `TaskStorage` struct:

image

I found that writing to `TaskStorage::future` via a `#[inline(never)]` function successfully coaxes the compiler into doing RVO here. Perhaps it's because a smaller, simpler function makes it more obvious to the optimizer that RVO is a valid optimization here.

The disassembly for `TaskPool::spawn_impl` _after_ applying this change:

image

And the disassembly for this particular `UninitCell::write_in_place` instantiation:

image

Interestingly, as well as performing RVO, the compiler is now able to see that it can validly only partially initialize the future object, by only setting what I assume is the discriminant field, even though the future object is still just as large.

### DIFF
--- a/embassy-executor/src/arch/wasm.rs
+++ b/embassy-executor/src/arch/wasm.rs
@@ -73,9 +73,10 @@ mod thread {
         pub fn start(&'static mut self, init: impl FnOnce(Spawner)) {
             unsafe {
                 let executor = &self.inner;
-                self.ctx.closure.write(Closure::new(move |_| {
+                let future = Closure::new(move |_| {
                     executor.poll();
-                }));
+                });
+                self.ctx.closure.write_in_place(|| future);
                 init(self.inner.spawner());
             }
         }
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -203,7 +203,7 @@ impl<F: Future + 'static> AvailableTask<F> {
     fn initialize_impl<S>(self, future: impl FnOnce() -> F) -> SpawnToken<S> {
         unsafe {
             self.task.raw.poll_fn.set(Some(TaskStorage::<F>::poll));
-            self.task.future.write(future());
+            self.task.future.write_in_place(future);
 
             let task = TaskRef::new(self.task);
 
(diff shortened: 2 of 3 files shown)


embassy-executor/src/arch/wasm.rs
```diff
@@ -73,9 +73,10 @@ mod thread {
         pub fn start(&'static mut self, init: impl FnOnce(Spawner)) {
             unsafe {
                 let executor = &self.inner;
-                self.ctx.closure.write(Closure::new(move |_| {
+                let future = Closure::new(move |_| {
                     executor.poll();
-                }));
+                });
+                self.ctx.closure.write_in_place(|| future);
```
haileys, line 79: This code is technically fine to just use `write` since the [`Closure` struct is only small](https://docs.rs/wasm-bindgen/0.2.87/src/wasm_bindgen/closure.rs.html#241-244), but I figure it's safest to remove `write` and instead make all callers go through `write_in_place`, since futures are _in general_ of unbounded size, and leaving a more convenient `write` function in place could cause this bug to slip back in as the code changes in future.

Dirbaio (approved):
very interesting find, thank you!

PR #1906, opened 2023-09-14
update metapac and add low-power for l0

xoviat:


Dirbaio:
can you make `ci.sh` try building some L0 with `low-power` enabled, so the build never breaks? Same for F4, I see it wasn't added.

PR #1908, opened 2023-09-15
Fix low-power feature for STM32L0

### LINKED ISSUE #1891: feature "low-power" of embassy-stm32: hint for (not) support
I wanted to use the feature "low-power" for my stm32l0. I quickly found out, that it is not working because of some compiler errors. Apparently this is not supported for the stm32l0 right now. I tested all stm32 examples and as far as I can tell it is only supported for stm32f4 right now. For every stm32 line there are different compiler errors. 

Before I want to add support for this feature for the stm32l0, I think it makes sense to add a hint that this feature is not supported. Instead of some compiler errors which you have to investigate.

I don't know of any way to do this in a nice way in Cargo.toml as a dependency or something. So I think the best way would be like this (in embassy-stm32/src/low-power.rs):
image

Also I would add this condition on rtc_f4 for every other #[cfg(feature = "low-power")] to not get any more compiler warnings, but only the one compile error shown in the image.

Is this the best approach or is there a better one?

chrenderle:
Fixes #1891

### DIFF
--- a/embassy-stm32/src/rtc/v2.rs
+++ b/embassy-stm32/src/rtc/v2.rs
@@ -298,7 +298,7 @@ impl sealed::Instance for crate::peripherals::RTC {
             crate::pac::PWR.cr().read();
         }
 
-        #[cfg(any(rtc_v2f0))]
+        #[cfg(any(rtc_v2f0, rtc_v2l0))]
         {
             // enable peripheral clock for communication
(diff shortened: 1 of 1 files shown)


xoviat:
Sorry, the place for this is enable_peripheral_clock in the rtc mod. I missed that because I wanted to have something to test on hardware quickly.

PR #1917, opened 2023-09-16
stm32: Implement `set_config` for Uart structs

MabezDev:
Allows the uart to be reconfigured after initialization.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -114,6 +114,30 @@ pub struct BufferedUartRx<'d, T: BasicInstance> {
     phantom: PhantomData<&'d mut T>,
 }
 
+impl<'d, T: BasicInstance> SetConfig for BufferedUart<'d, T> {
+    type Config = Config;
+
+    fn set_config(&mut self, config: &Self::Config) {
+        self.set_config(config)
+    }
+}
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/usart/buffered.rs
```diff
@@ -304,6 +332,10 @@ impl<'d, T: BasicInstance> BufferedUartRx<'d, T> {
             T::Interrupt::pend();
         }
     }
+
+    pub fn reconfigure(&mut self, config: &Config) {
```
Dirbaio, line 336: this should be named `set_config` for consistency. (the overlap between inherent and trait method is fine. Rust will silently prefer the inherent one, and since they have the same signature that will always work.)

MabezDev: I was following from SPI: https://github.com/embassy-rs/embassy/blob/3ddc9cd110cb56f2ba9ac491b42df19a9b9c3fd4/embassy-stm32/src/spi/mod.rs#L1062-L1067. Would you like me to change that method to set_config too?

Dirbaio: oh oops. yes please :sweat_smile:

Dirbaio (approved):
thanks!

PR #1921, opened 2023-09-17
stm32: update configure_ls and misc cleanup

xoviat:
as agreed in chat.

### DIFF
--- a/embassy-stm32-wpan/src/sub/ble.rs
+++ b/embassy-stm32-wpan/src/sub/ble.rs
@@ -1,4 +1,3 @@
-use core::marker::PhantomData;
 use core::ptr;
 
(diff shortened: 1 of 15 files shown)


embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -41,21 +41,18 @@ mod phy_consts {
 }
 use self::phy_consts::*;
 
+#[non_exhaustive]
 /// Generic SMI Ethernet PHY
 pub struct GenericSMI {
     #[cfg(feature = "time")]
     poll_interval: Duration,
-    #[cfg(not(feature = "time"))]
```
Dirbaio, line 48: this is needed so the struct doesn't become publically-constructible with `GenericSMI{}` if `time` is disabled.

PR #1939, opened 2023-09-22
spim: Fix nRF52832 SPIM with polling executors

nilfit:
On nRF52832, a workaround is used to avoid corruption of SPIM DMA transfers (anomaly 109). Because the logic setting up the real transfers was in the interrupt handler, naive executors that just poll (like  `embassy_futures::block_on`) would resolve without making (or starting) the transfer.

Move the workaround code from the interrupt handler to a `poll_fn` so that the it runs regardless of whether the event was detected through polling or by the interrupt handler triggering the waker.


plaes:
This unfortunately breaks libraries which utilize embedded-hal-0.2.x. I'll see whether I can figure it out.

plaes:
Seems to be now working in both of my testcases: async + blocking case with library using embedded-hal-0.2.x.

I noticed that I had botched the longer commit message from the initial commit, though...

korken89:
Great work on this! ♥️

Dirbaio (approved):
thanks!

PR #1947, opened 2023-09-25
Update rust-lorawan and embassy to use lora-phy version 2

ceekdee:
Update rust-lorawan and embassy to use lora-phy version 2.  This includes the latest released version of embedded-hal-async.


ceekdee:
With this update, if the embedded-hal-async crate is updated while I am away until late October, the update to lora-phy to support it needs to be made on the main branch rather than the v1.x branch.

Dirbaio (approved):
very nice! Thank you! :D

PR #1949, opened 2023-09-25
stm32/rtc: cleanup and fixes

xoviat:


embassy-stm32/src/rtc/mod.rs
```diff
@@ -10,6 +10,7 @@ use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
 use embassy_sync::blocking_mutex::Mutex;
 
 pub use self::datetime::{DateTime, DayOfWeek, Error as DateTimeError};
+#[allow(unused_imports)]
```
Dirbaio, line 13: instead of allowing unused imports, it's a bit moremaintainable to use the fully-qualified function call syntax:

```rust
<RTC as crate::rcc::sealed::RccPeripheral>::enable();
```

for example so if the `enable` call is removed in the future we're not left with a really unused import that triggers no warning.

PR #1958, opened 2023-09-28
Add support for input capture function

dberlin:
This pull requests add support to embassy_stm32 for the input capture mode
registers. It covers most of them. I will add the remainder in a future PR.

It is a fairly straightforward approach. The only non-obvious things are:

1. The input capture polarity is controlled by two registers, so we introduce an enum
   to represent it

2. The TI selection register has the wrong definitions in the SVD file, and is often
   confusing to users, so we introduce an enum to make obvious what it is controlling.

### DIFF
--- a/embassy-stm32/src/timer/mod.rs
+++ b/embassy-stm32/src/timer/mod.rs
@@ -15,6 +15,7 @@ pub mod low_level {
 }
 
 pub(crate) mod sealed {
+
     use super::*;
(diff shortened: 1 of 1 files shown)


xoviat:
So, would you like us to merge this? If so, please take it out of draft.

dberlin:
> So, would you like us to merge this? If so, please take it out of draft.

Whoops, sorry about that.

PR #1959, opened 2023-09-28
stm32: add eth tests

Dirbaio:


Dirbaio:
bender run

Dirbaio:
bender run

tests/nrf/src/bin/wifi_esp_hosted_perf.rs
```diff
@@ -23,6 +21,10 @@ bind_interrupts!(struct Irqs {
     RNG => embassy_nrf::rng::InterruptHandler<peripherals::RNG>;
 });
 
+// Test-only wifi network, no internet access!
```
plaes, line 24: Would it be possible to pick these up from ENV variables compile-time [core](https://doc.rust-lang.org/core/index.html)::[env](https://doc.rust-lang.org/core/macro.env.html#) instead of hardcoding these?

ci.sh
```diff
@@ -183,6 +183,7 @@ cargo batch  \
     --- build --release --manifest-path tests/riscv32/Cargo.toml --target riscv32imac-unknown-none-elf \
     $BUILD_EXTRA
 
+rm out/tests/nrf52840-dk/wifi_esp_hosted_perf
```
plaes, line 186: It would be nice to have a TODO comment, about the reason... :)

Dirbaio: done

Dirbaio:
bender run

PR #1962, opened 2023-09-28
Add 80MHz tick rate

bugadani:
This PR attempts fixing [this](https://github.com/esp-rs/esp-hal/blob/9f76798e64323c4f9f9f01943cd154c1642516d1/esp32s2-hal/Cargo.toml#L70). The ESP32S2 seems to need to work with 80MHz SysTimer tick rates.

I didn't overcomplicate things, just added the single frequency manually. Maybe 2^x multiples of 10MHz could be added in general.


Dirbaio:
> Maybe 2^x multiples of 10MHz could be added in general.

yes, now that it's shown that we have "holes" in the heuristic i'd just add more ways. Perhaps all multiples of 10mhz up to 300mhz, then multiples of 20 from there up to 1000mhz?

bugadani:
Might be a bit excessive until needed, don't you think? :)

Dirbaio (approved):
it's "just" a 2x increase. i'd rather not keep reviewing "add XXX tick rate" PRs :smiling_face_with_tear:

PR #1964, opened 2023-09-28
Add support for STM32 input capture filter

dberlin:
This PR adds support for the input capture filter.


dberlin:
This is, i believe, the only thing that was still missing.

I'll add some input capture measurement examples in a followup PR, just finishing testing them on a few variants.

PR #1969, opened 2023-09-29
Forgot set_count_direction and set_clock_division in 32 bit instance

dberlin:
Writing examples found I forgot to copy two methods into the 32 bit general purpose instance.


Dirbaio (approved):

dberlin:
This was actually wrong, I was not importing the right trait in the right place.
The 32 bit inherits from the 16 bit here so it didn't need these two functions (and it makes them ambiguous when both traits are used)
I'll revert it.

PR #1970, opened 2023-09-29
Revert "Forgot set_count_direction and set_clock_division in 32 bit instance"

dberlin:
Reverts embassy-rs/embassy#1969

This was unfortunately wrong, and later tests I wrote showed it. 
The 32 bit trait inherits from the 16 bit trait in this case, so it doesn't need these functions (and adding them makes it ambiguous).


xoviat:
Are you sure about this?

PR #1978, opened 2023-09-30
wip, signpost

elasticdotventures:
this WIP will be a signpost & WIP that spans:
https://github.com/esp-rs/esp-hal/pull/834
https://github.com/drogue-iot/drogue-device/pull/388


Dirbaio:
IMO we shouldn't add esp32 examples to the Embassy repo. The esp-32 hal is maintained at https://github.com/esp-rs/esp-hal, that repo is much better positioned to host the examples since they can check in CI that they're always updated to the latest esp-hal version. If we add the examples here, they'll fall out of date unless we invest recurring effort in updating them. Out-of-date examples are confusing for the users, it's best to just point them to the esp-hal repo.

lulf:
Agreed with @Dirbaio. I think a useful contribution would be to add references to the esp-hal embassy examples in the [documentation](https://embassy.dev/book/dev/hal.html)

barafael:
@Dirbaio looks like this can be closed?

PR #1984, opened 2023-09-30
stm32: require use of setconfig to set config

xoviat:


embassy-stm32/src/usart/mod.rs
```diff
@@ -273,7 +276,7 @@ impl<'d, T: BasicInstance, TxDma> UartTx<'d, T, TxDma> {
         })
     }
 
-    pub fn set_config(&mut self, config: &Config) -> Result<(), ConfigError> {
+    fn set_config(&mut self, config: &Config) -> Result<(), ConfigError> {
```
MabezDev, line 279: I think this should have stayed public so that users don't need to import the trait to use it.

PR #1985, opened 2023-09-30
WIP: Add alarms for RTC

chrenderle:
Didn't compile it yet.
Didn't test it on any target yet.
Only written for STM32L0 and probably breaks other targets.

TODO:
- [ ] make it compile ready
- [ ] test on STM32L0
- [ ] test compilation on other targets
- [ ] write drivers which allow setting the RTC alarm

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -11,7 +11,7 @@ use crate::time_driver::{get_driver, RtcDriver};
 
 const THREAD_PENDER: usize = usize::MAX;
 
-use crate::rtc::Rtc;
+use crate::rtc::{Rtc, EXIT_LINE_RTC_ALARM, EXTI_LINE_RTC_WKUP};
 
 static mut EXECUTOR: Option<Executor> = None;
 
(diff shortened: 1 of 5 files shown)


xoviat:
> For the STM32L0 the interrupt routine is the same as the wkup timer so there has to be some overlap.

If that is the case, the low-power wakeup irq handler needs to be converted to bind_interrupts and moved to the rtc mod, and needs to call the `on_wakeup_irq` function when the alarm wakes the chip.

> The code in low power and time driver is only there since for the low-power feature the low-power executor has ownership of the rtc.

The low-power feature is only intended to have ownership over part of the rtc: it's wakeup functionality and timekeeping. The alarm functionality can be split off like this:

```rust
let rtc = Rtc::new(...)
let alarm1 = rtc.alarm1.take().unwrap();

let rtc=make_static!(rtc);
stop_with_rtc(rtc);

alarm1.wait_for_alarm(day, hour, minute, second).await;
```

In fact, low-power should be able to work without embassy-time at all (in that case, it would not even set the wakeup alarm) and alarms should be available to work without the low-power feature (in that case, the core wouldn't stop, but sleep until the alarm).

Dirbaio:
Hello,

PR #1994, opened 2023-10-02
feat (rp2040): allow schmitt, slew, and drive strength be set from Flex, Input, Output

jcdickinson:
Allows the schmitt, slew and drive strength to be set from Flex. Input and Output[OpenDrain] also expose the appropriate setters.


Dirbaio (approved):
thank you!

Dirbaio:
bender run

PR #1995, opened 2023-10-02
[STM32] QUADSPI support bank 2

JuliDi:
This PR supports using QSPI bank 2 on stm32 by introducing a second constructor that accepts the pins implementing the BK2 traits.
It does not yet allow for using both banks, but at least you can use bank 2 for now, e.g., if BK1 pins are used by another peripheral or so.

Further, a "precautionary step" according to the STM32H7 errata sheet is added. It looks very odd, but it is what the recommended "workaround" looks like. Hope it avoids some unwanted surprises in the future.


Dirbaio:
bender run

PR #1997, opened 2023-10-02
stm32: avoid useless endian conversion in rng

dzamlo:
Most stm32 are little endian. The code in `rng.rs` was converting the generated `u32`s to big_endian before using their bytes. This is not needed.

I did some ad hoc measurement of the speed of `async_fill_bytes`and saw a small but measurable improvement.


Dirbaio (approved):
thank you!

Issue #1791, opened 2023-08-17
NoRoute error using the Pico-W on wifi


InfinityDevTech:
Hello!

Im trying to connect to my Node TCP server on port `80` with ip addr of `192.168.0.2`, but i keep getteing a NoRoute error, here is the relevant code.

Im pretty new to embedded development, so go a little easy on me lol

```rust
let fw = include_bytes!("../cyw43-firmware/43439A0.bin");
    let clm = include_bytes!("../cyw43-firmware/43439A0_clm.bin");

    let pwr = Output::new(peripherals.PIN_23, Level::Low);
    let cs = Output::new(peripherals.PIN_25, Level::High);
    //let white_led = Output::new(p.PIN_15, Level::Low);
    let mut c: embassy_rp::pwm::Config = embassy_rp::pwm::Config::default();
    c.top = 0x8000;
    c.compare_b = 8;
    let white_led_pwm = Pwm::new_output_a(peripherals.PWM_CH7, peripherals.PIN_14, c.clone());
    let mut pio = Pio::new(peripherals.PIO0, Irqs);
    let spi = PioSpi::new(
        &mut pio.common,
        pio.sm0,
        pio.irq0,
        cs,
        peripherals.PIN_24,
        peripherals.PIN_29,
        peripherals.DMA_CH0,
    );
    let state = make_static!(cyw43::State::new());
    let (net_device, mut control, runner) = cyw43::new(state, pwr, spi, fw).await;
    unwrap!(spawner.spawn(init::wifi_task(runner)));
    control.init(clm).await;
    control
        .set_power_management(cyw43::PowerManagementMode::PowerSave)
        .await;
    control.gpio_set(0, true).await;

    let config = Config::dhcpv4(Default::default());
    let seed = 0x0123_4567_89ab_cdef;

    // Init network stack
    let stack = make_static!(Stack::new(
        net_device,
        config,
        make_static!(StackResources::<4>::new()),
        seed
    ));

    loop {
        //control.join_open(WIFI_NETWORK).await;
        match control.join_wpa2(WIFI_NETWORK, WIFI_PASSWORD).await {
            Ok(_) => {
                info!("Joined the network! {}", WIFI_NETWORK);
                break;
            }
            Err(err) => {
                info!("join failed with status={}", err.status);
                flash_onboard(control).await;
                return;
            }
        }
    }

    unwrap!(spawner.spawn(init::net_task(stack)));

    let mut rx_buffer = [0; 4096];
[...]

Issue #1234, opened 2023-02-24
embassy-net: NoRoute error on connect


LimpidCrypto:
## Source
https://github.com/LimpidCrypto/em-as-net/blob/main/src/core/tcp/mod.rs#L56

## Overview
I'm getting a `ConnectError::NoRoute` error when trying to connect. The error get thrown when `smoltcp` [`get_source_address`](https://github.com/smoltcp-rs/smoltcp/blob/master/src/iface/interface/mod.rs#L1079) returns `None`, when trying to choose a local address automatically. This results in an `smoltcp::tcp::ConnectError::Unaddressable` error. I can't really get my head around what I'm doing wrong. Can someone please help me out here and point me in the right direction what could cause this error?

## Notes
- I can ping the ip
- I'm using the default settings for DHCP

---
**This might also be more of question of the `smoltcp` library. If so, I will ask there :)**

LimpidCrypto:
Seems like the default config of DHCP is not working for me. Using static config works doesn't throw an error anymore.
```rust
// `TcpConfig` is just a wrapper for `embassy_net::Config`
let config = TcpConfig::new_static(
    Ipv4Cidr::new(Ipv4Address::new(192, 168, 69, 2), 24),
    Some(Ipv4Address::new(192, 168, 69, 1)),
    Vec::from_iter([
        Ipv4Address::new(8, 8, 8, 8),
        Ipv4Address::new(9, 9, 9, 9),
        Ipv4Address::new(1, 1, 1, 1),
   ]),
);
```
But now connecting to the remote ip takes forever. Any advices?

kbleeke:
Are you waiting for DHCP to be finished? I have something like this before I attempt do use TCP

```rust
while !stack.is_config_up() {
    yield_now().await;
}
```

LimpidCrypto:
> Are you waiting for DHCP to be finished? I have something like this before I attempt do use TCP
> 
> 
> 
> ```rust
> 
> while !stack.is_config_up() {
> 
>     yield_now().await;
> 
> }
> 
> ```

Ah nice, I'm not doing that currently. I will give it a try later, thank you @kbleeke

LimpidCrypto:
Unfortunately it doesn't resolve my problem. It also turned out, that not `TcpSocket.connect` but `Stack.run` takes forever to complete.

Dirbaio:
`run` runs forever, you have to launch it in a background task. Please check how the examples do it.

[...]

PR #1999, opened 2023-10-02
cyw43: Update examples

KingCol13:
- Update firmware blob sizes in `from_raw_parts` to current size
- Add wait for DHCP for `wifi_tcp_server` (caused confusion in #1234, #1791)

Love this project, thanks!


Dirbaio (approved):
thank you!

PR #2009, opened 2023-10-03
TCP bug demo

wyager:
To trigger the hangup bug:

```
cat /dev/random | head -c 2048 | nc 192.168.2.3 1234 
```

Sending one byte more or less works fine.


wyager:
Oops, sorry, didn't realize this would open a PR here instead of on my fork

PR #2015, opened 2023-10-06
stm32: u5: implement >55 MHz clock speeds

willglynn:
This commit allows STM32U5 devices to operate at 160 MHz.

On STM32U5, MSIS can run at 48 MHz and HSE can reach 50 MHz. Faster clocks require using PLL1's R output, though PLL1 can serve other functions besides using the R output for the system clock. This commit extracts a public `PllConfig` struct, primarily to place associated constructors on that type, but also with an eye towards enabling the P and Q outputs in a later commit.

STM32U5 PLLs have various frequency requirements on each stage: after the `m` prescaler, after the `n` multiplier, and after the `r` divider. This commit implements the associated checks as assertions.

This commit fixes clock calculation and PLL register configuration errors in PLL initialization.

STM32U5 has a PWR peripheral which can be configured to push Vcore into different voltage ranges. System clocks exceeding 55 MHz require range 2, and system clocks exceeding 110 MHz require range 1. This commit adds `voltage_range` to `Config` and configures PWR as directed.

The voltage range implies different performance limits on various clock signals, including inside a PLL. This commit implements voltage range <-> frequency range checks as assertions, and extracts the otherwise-repeated MSIS, HSI16, and HSE initialization into private methods on `Config`.

STM32U5 frequencies above 55 MHz require using the PWR EPOD booster. The EPOD booster requires configuring a second `m` term for PLL1, `mboost`, such that it falls in a particular range. (Recall that >50 MHz cannot be reached without PLL1, so there is no scenario where EPOD is needed but PLL1 is not.) This commit configures and enables the EPOD booster automatically as required.

### DIFF
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -1,7 +1,7 @@
-use stm32_metapac::rcc::vals::{Msirange, Msirgsel, Pllm, Pllsrc, Sw};
+use stm32_metapac::rcc::vals::{Msirange, Msirgsel, Pllm, Pllmboost, Pllrge, Pllsrc, Sw};
 
 pub use super::bus::{AHBPrescaler, APBPrescaler};
-use crate::pac::{FLASH, RCC};
+use crate::pac::{FLASH, PWR, RCC};
 use crate::rcc::{set_freqs, Clocks};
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
thank you!

PR #2018, opened 2023-10-06
Add some uncontroversial derives to Error types

jamesmunns:
I did a search for `enum Error`, and added `PartialEq, Eq` where it was missing. You could maybe also make an argument for adding `Clone`, open to thoughts!


jamesmunns:
Currently failing due to some missing downstream missing Eq, will look more later.

Dirbaio (approved):
thx!

PR #2020, opened 2023-10-06
Add MCO support for stm32wl family

ShakenCodes:
Per discussions in the chat forum, this exposes the MCO feature for the STM32WL. It exposes the McoPin trait on any processor that has an MCO peripheral, by making the selection based on the pin signal starting with MCO.


Dirbaio:
superseded by #2021 (contains this PR)

PR #2030, opened 2023-10-09
stm32: Implement embedded-hal-async I2C trait for TimeoutI2c

andresovela:
I also made `TimeoutI2c` take ownership of the I2C peripheral. I didn't see the point in taking a mutable reference, it made it unnecessarily harder to use IMO.


andresovela:
Closing in favor of #2033

PR #2032, opened 2023-10-10
stm32 flash: check lock bit before unlocking

andresv:
It hardfaults if already unlocked flash is unlocked again.

Tested on stm32g081 where I always got hardfault when unlocking and then realised that flash was actually already unlocked. Not sure why it is unlocked after flashing on my case, but those changes prevent hardfault if this is the case.

At least on stm32g081 erase read write are still working after those changes.


Dirbaio (approved):
thank you!

PR #2033, opened 2023-10-10
stm32: add timeout to I2C driver

andresovela:


embassy-stm32/Cargo.toml
```diff
@@ -33,7 +33,7 @@ flavors = [
 
 [dependencies]
 embassy-sync = { version = "0.3.0", path = "../embassy-sync" }
-embassy-time = { version = "0.1.3", path = "../embassy-time", optional = true }
+embassy-time = { version = "0.1.3", path = "../embassy-time" }
```
lulf, line 36: Won't this require everyone to enable a time driver in order to use the stm32 HAL? 🤔 I don't think that should be enforced.

andresovela: Alternatively I'd suggest to add a `dma` feature that requires `time` in order to implement timeout support. Not having timeout functionality is too big of a footgun IMO.

andresovela: Otherwise I'm open to suggestions to implement timeouts with a different approach.

lulf: I've not spent much time with stm32 i2c, but is it not possible to handle timeouts externally, like this?

```
match select(Timer::after(Duration::from_secs(10)), i2c.write(...)).await {
    Either::First(_) => {} // Timeout!
    Either::Second(_) => {} // Success
}
```

Iirc, dropping the write future should do the required cancellation

lulf: Thanks, I think I understand your arguments better now. So, I think these are potential ways forward then:

* Guard I2C fns using timeouts behind the 'time' feature and make embassy-time optional again. This way you _have_ to enable timekeeping to use i2c. I think it's fine if there is never a case where you don't need timeouts. 
* If there is a valid use case to not use timeouts, then I think the better approach is to provide 2 methods: foo() and foo_with_timeout() where the second one is guarded by the time feature.

I don't know the entire set of i2c use cases, so I don't know which one of these options are best. If there are no reservations from others against the first option, I would say go for that. Most important I think is to allow the use case of not enabling embassy-time, even if that means no i2c.

andresovela: I think guarding the APIs behind a feature guard is a reasonable fix. The question the would be, do we want them to be guarded by a `time` feature or by an `async-i2c` (or something else) feature that requires `time`?

lulf: Good question... I personally think using the `time` feature is fine, but probably warrants a comment in the code?

embassy-stm32/src/i2c/mod.rs
```diff
@@ -2,16 +2,13 @@
 
 use crate::interrupt;
 
+// I2C V2 is gated on the `time` feature because timing facilities are necessary
+// to provide timeout functionality in order to prevent some APIs from stalling indefinitely
 #[cfg_attr(i2c_v1, path = "v1.rs")]
-#[cfg_attr(i2c_v2, path = "v2.rs")]
+#[cfg_attr(all(i2c_v2, feature = "time"), path = "v2.rs")]
 mod _version;
 pub use _version::*;
```
lulf, line 10: @andresovela You probably also need to gate the pub use.... I think. Consider that before this, it would always have a 'mod _version' that was valid, but this is no longer the case.

andresovela: Makes sense

andresovela: CI still complains about the same thing

lulf: So, basically any types, like State is defined within v1 or v2, so without time, it means these are missing. Maybe you'll have an easier time if you add cfg guards within v2.rs on the specific bits that rely on types from embassy-time?

andresovela: Oh I understand what the error is saying now. Cool, thanks :)

andresovela:
Finally! Made CI happy. I'm not very happy with all the guards everywhere now, but I guess that's what needs to be done? Let me know if you know of better solutions.

lulf (commented):
It looks good to me! Could you perhaps add at least one chip build to ci.sh that tried to build embassy-stm32 without time so it doesn't bit-rot?

andresovela:
Done

lulf (approved):

chemicstry:
Unfortunatelly this removed timeout feature from i2c_v1 🫤

PR #2035, opened 2023-10-10
STM32: combine RccPeripherals reset() and enable() to enable_and_reset()

### LINKED ISSUE #2034: Not possible to use both ADCs with STM32H7
Hello,

i'm trying to use both ADC1 and ADC2 on a STM32H7A3 chip.
Unfortunately as soon as the second ADC is enabled (order does not matter) the other ADC is reseted.

I was able to trace it `new` function, which enables the RCC Clock and afterwards resets the peripheral.
```impl<'d, T: Instance> Adc<'d, T> {
    pub fn new(adc: impl Peripheral<P = T> + 'd, delay: &mut impl DelayUs<u16>) -> Self {
        embassy_hal_internal::into_ref!(adc);
        T::enable();
        T::reset();
```
In this case, both ADCs share the reset:
`RCC_AHB1RSTR::ADC12RST`
and the already initalised ADC is reseted when the new() function is called for the second ADC.
The RccPeripheral trait is implemented as part of the generated stuff.

I'm open to fix this issue myself, however i'm not familiar with the metapac process.
Are they any similar Cases for other peripherals? How are you usually handle it, if the generated code does not fit?
Any hints in the right direction are appreciated.

pbert519:
Close  #2034

Combining the standalone functions `reset()` and `enable()` into `enable_and_reset()`.
 
 This are a lot of breaking changes, maybe we should add `reset_and_enable()`, but keep `enable()` and `reset()`. 
 And port only the obvious usages.
 I'm not able to test everything on real hardware.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -452,10 +452,8 @@ fn main() {
                     let rst_reg = format_ident!("{}", rst.register.to_ascii_lowercase());
                     let set_rst_field = format_ident!("set_{}", rst.field.to_ascii_lowercase());
                     quote! {
-                        critical_section::with(|_| {
-                            crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(true));
-                            crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(false));
-                        });
+                        crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(true));
+                        crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(false));
                     }
                 }
                 None => TokenStream::new(),
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -556,14 +556,14 @@ fn main() {
                     fn frequency() -> crate::time::Hertz {
                         #clock_frequency
                     }
-                    fn enable() {
+                    fn enable_and_reset() {
                         critical_section::with(|_cs| {
                             #before_enable
-                            #rst
                             #[cfg(feature = "low-power")]
                             crate::rcc::clock_refcount_add(_cs);
                             crate::pac::RCC.#en_reg().modify(|w| w.#set_en_field(true));
                             #after_enable
+                            #rst
                         })
                     }
                     fn disable() {
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -559,6 +559,7 @@ fn main() {
                     fn enable() {
                         critical_section::with(|_cs| {
                             #before_enable
+                            #rst
                             #[cfg(feature = "low-power")]
                             crate::rcc::clock_refcount_add(_cs);
                             crate::pac::RCC.#en_reg().modify(|w| w.#set_en_field(true));
@@ -572,9 +573,6 @@ fn main() {
                             #[cfg(feature = "low-power")]
                             crate::rcc::clock_refcount_sub(_cs);
                         })
-                    }
-                    fn reset() {
-                        #rst
                     }
                 }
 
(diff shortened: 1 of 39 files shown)


pbert519:
The DAC peripheral has also shared RCC bits for DAC1 und DAC2 in some h7 controllers.
I tried to remove the workaround, however the build.rs script did not generate impl RccPeripherals for DAC1 for the stm32h742xx-cm7 target, all other targets seems to work.
Not sure how to proceed there

Dirbaio:
Thanks for the PR! 

The code should first enable the peripherals then reset them. Reset requires the clocks to be running, otherwise peripheral state can't change. Could you change that, and rename the function?

> the build.rs script did not generate impl RccPeripherals for DAC1 for the stm32h742xx-cm7 target, all other targets seems to work.

the data comes from stm32-data. see [here](https://github.com/embassy-rs/stm32-data-generated/blob/main/data/chips/STM32H742ZI.json#L905), DAC1 is missing the `rcc` field. Compare with [this](https://github.com/embassy-rs/stm32-data-generated/blob/main/data/chips/STM32G431KB.json#L612) which is on another chip that does have it.

The code that does the matching is [here](https://github.com/embassy-rs/stm32-data/blob/main/stm32-data-gen/src/rcc.rs). There's exceptions hardcoded for ADC, seems DAC needs some too.

(if you want to merge this PR now and fix that later that's fine, this is already a big improvement even if it doesn't fix everything. Up to you!)

pbert519:
> The code should first enable the peripherals then reset them. Reset requires the clocks to be running, otherwise peripheral state can't change. Could you change that, and rename the function?

Done. It's probably better to assume the reset is synchronous. It's a pity that st is not able to define this.

I'll have a look into the DAC topic, but we should split it into a dedicated PR.

embassy-stm32/src/sai/mod.rs
```diff
@@ -531,10 +531,13 @@ pub struct SubBlock<'d, T: Instance, C: Channel, W: word::Word> {
 pub struct SubBlockA {}
 pub struct SubBlockB {}
 
+pub struct SubBlockAPeripheral<'d, T>(PeripheralRef<'d, T>);
```
Dirbaio, line 534: could you take these unrelated changes out? thanks!

embassy-stm32/build.rs
```diff
@@ -433,13 +433,14 @@ fn main() {
...
                     }
-                    fn enable() {
+                    fn enable_and_reset() {
                         critical_section::with(|_| {
                             #before_enable
                             #[cfg(feature = "low-power")]
                             crate::rcc::clock_refcount_add();
                             crate::pac::RCC.#en_reg().modify(|w| w.#set_en_field(true));
                             #after_enable
+                            #rst
```
Dirbaio, line 443: since this is now inside a CS, the CS inside the `rst` code is no longer needed.

embassy-stm32/src/sai/mod.rs
pbert519: This aren't unrelated changes.
While the SAI is only one Peripheral Block, it contains two complety independent subblocks.
Currently it is possible to create a subblock directly, or frist create a SAI peripheral, which just copies the PeripheralRef, so we can initalise both subblocks.
Question, where do we call enable_and_reset()?

This changes force the user to create a SAI first, which enables clock and reset, and then use the take_sub_block to get a SubBlockXPeripheral to initalise the Subblock

embassy-stm32/build.rs
pbert519: Fixed

embassy-stm32/src/sai/mod.rs
Dirbaio: ahh got it, makes sense, thank you!

Dirbaio (approved):
thank you!

PR #2038, opened 2023-10-11
fix (rp i2c): set i2c slew rate to spec value

jcdickinson:
The RP2040 datasheet indicates that I2C pins should have a limited slew rate (Page 440 - 4.3.1.3.). This configures that for both `I2c` and `I2cSlave`.

In addition, the pin configuration has been centralized to a single fn.


Dirbaio (approved):
thank you!

PR #2039, opened 2023-10-11
fix (rp i2c): fix restart/stop flags for i2c master methods

jcdickinson:
Update the start and stop flags for all read/write/read_write methods to match those in the default blocking implementation of these methods (as well as other RP2040 I2C implementations, and expected I2C behavior).

Flags used correctly elsewhere: https://github.com/embassy-rs/embassy/blob/022d870d616575805ace9b37ea87bae1fd90cb84/embassy-rp/src/i2c.rs#L561

rp-rs implementation: https://github.com/rp-rs/rp-hal/blob/c4d3fa882acc5b19269d20ff74f65f1e53e81ee9/rp2040-hal/src/i2c/controller.rs#L313

Also adds a write_read_async method that doesn't require using embedded-hal, as this is required to use I2C in an idiomatic fashion (see TI Application Report [SLVA704](https://www.ti.com/lit/an/slva704/slva704.pdf)).

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -295,12 +295,23 @@ impl<'d, T: Instance> I2c<'d, T, Async> {
 
     pub async fn read_async(&mut self, addr: u16, buffer: &mut [u8]) -> Result<(), Error> {
         Self::setup(addr)?;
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #2040, opened 2023-10-11
fix (rp i2c): stretch the clock during read transfers

jcdickinson:
Stretches the clock during read transfers.

I tested this against the in-repo I2c slave example, as well as my prototype.

Please allow @CBJamo to weigh in, possibly test, before merging. I'm pretty green at embedded, so may interrupt/register juggling might be off.

### DIFF
--- a/embassy-rp/src/i2c_slave.rs
+++ b/embassy-rp/src/i2c_slave.rs
@@ -67,6 +67,25 @@ pub struct I2cSlave<'d, T: Instance> {
     phantom: PhantomData<&'d mut T>,
 }
 
+macro_rules! trace_stat {
+    ($wher: expr, $stat: expr, $sstat: expr) => {
+        defmt::trace!(
+            "{}: tx_empty {} rd_req {} rx_done {} tx_abrt {} activity {} stop_det {} start_det {} gen_call {} restart_det {} rfne {}",
+            $wher,
+            $stat.tx_empty(),
+            $stat.rd_req(),
+            $stat.rx_done(),
+            $stat.tx_abrt(),
+            $stat.activity(),
+            $stat.stop_det(),
+            $stat.start_det(),
+            $stat.gen_call(),
+            $stat.restart_det(),
+            $sstat.rfne()
+        );
+    };
+}
+
 impl<'d, T: Instance> I2cSlave<'d, T> {
     pub fn new(
         _peri: impl Peripheral<P = T> + 'd,
@@ -94,6 +113,7 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
             w.set_ic_slave_disable(false);
             w.set_tx_empty_ctrl(true);
             w.set_rx_fifo_full_hld_ctrl(true);
+            w.set_stop_det_ifaddressed(true);
         });
 
         // Set FIFO watermarks to 1 to make things simpler. This is encoded
@@ -113,7 +133,7 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
         p.ic_enable().write(|w| w.set_enable(true));
 
         // mask everything initially
-        p.ic_intr_mask().write_value(i2c::regs::IcIntrMask(0));
+        Self::set_intr_mask(|_| {});
         T::Interrupt::unpend();
         unsafe { T::Interrupt::enable() };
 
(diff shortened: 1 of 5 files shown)


embassy-rp/src/i2c_slave.rs
```diff
@@ -181,14 +181,28 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
...
                         Poll::Ready(Ok(Command::Write(len)))
                     } else if stat.rd_req() {
                         Poll::Ready(Ok(Command::Read))
+                    // If restart_det() is set but not rd_req() the register can
+                    // get stuck, and we can enter a tight loop.
+                    } else if stat.restart_det() && !stat.rd_req() {
+                        p.ic_clr_restart_det().read();
+                        Poll::Pending
+                    // Likewise if stop_det() is set and len is 0.
+                    } else if stat.stop_det() && p.ic_rxflr().read().rxflr() == 0 {
```
jcdickinson, line 203: I'm seeing this a lot with i2c across two distinct chips (it does not appear with the loopback). I'm not sure if I've done something wrong, or if it's a race condition.

jcdickinson:
I have isolated the issues that I am seeing to the slave side, the master seems to be working right now (switching the master out with one written with rp-rs doesn't change anything).

Note that everything works in the loopback/example scenario.

The current code is demonstrating run-ons:

```
DEBUG write to backlight: [1, 1, 0]
└─ src/main.rs:134
DEBUG write to backlight: [1, 1]
└─ src/main.rs:134
DEBUG unknown read
└─ src/main.rs:129
```

There should be no naked reads occurring (every access is either a `write_read` or a `write`), and writes should be exactly two bytes long.

jcdickinson:
I want to complete my kbd firmware before tackling this, but I do want to get back to it. If someone wants to take over the branch prior to that, then please feel free to.

Some notes:
* You will need to have two RP2040s to properly test this. The loopback example is overly forgiving. I was using these:
   * https://codeberg.org/jcdickinson/sporeboard/src/commit/95569a14d3f3ed477c88a419126fa04498ffef3a - `main` Embassy impl
   * https://codeberg.org/jcdickinson/sporeboard/src/commit/88af16d1c5139fa349d4bc17dc87807609b283f4 - `rs-rp` rs-rp impl
   * I will probably leave those branches alone, as I do want to have the final kbd firmware use Embassy.
* `rd_req` is how the clock gets stretched. When the RP2040 notices a read it will start stretching and set this flag. Once this flag is cleared with `ic_clr_rd_rq` the stretching will stop.
* Writes are running across stops right now (probably sans any reads being involved, but I haven't checked that). This is definitely wrong. The master and the slave must always agree on starts, stop, restarts, and which bytes occur between them.
* I have validated that the in-repo controller is doing the right thing against `rs-rp`. 
* In the above example:
  * The writes should be exactly 2 bytes long and equal to `[1, 1]` every time.
  * The write_reads should be `[0, read]` every time.
* Super-helpful document on how I2C is supposed to work: https://www.ti.com/lit/an/slva704/slva704.pdf

jcdickinson:
I have created a parity implementation of the `i2c_slave` example (as per my updates) using `rp-rs`. The problem definitely exists in slave:

|                    | **rp-rs slave**    | **embassy slave** |
|--------------------|--------------------|-------------------|
| **rp-rs master**   | :white_check_mark: | :x:               |
| **embassy master** | :white_check_mark: | :x:               |

```
4.981705 TRACE listen: tx_empty true rd_req true rx_done false tx_abrt false activity true stop_det false start_det true gen_call false restart_det true rfne false
└─ embassy_rp::i2c_slave::{impl#1}::listen::{async_fn#0}::{closure#1} @ /home/jono/Code/embassy/embassy-rp/src/i2c_slave.rs:72
4.982174 TRACE len 5 [da, bb, ad, 0, c8]
└─ embassy_rp::i2c_slave::{impl#1}::listen::{async_fn#0}::{closure#1} @ /home/jono/Code/embassy/embassy-rp/src/fmt.rs:117

4.982523 INFO  device recieved write read: [da, bb, ad, 0, c8]
└─ i2c_slave::__device_task_task::{async_fn#0} @ src/bin/i2c_slave.rs:51

4.982751 ERROR Invalid Write Read da
└─ i2c_slave::__device_task_task::{async_fn#0} @ src/bin/i2c_slave.rs:79
```

Isolated line should read:

```
4.982523 INFO  device recieved write read: [da, bb, ad, 0]
└─ i2c_slave::__device_task_task::{async_fn#0} @ src/bin/i2c_slave.rs:51
```

jcdickinson:
I figured it out. I will break this PR up into separate PRs.

PR #2050, opened 2023-10-12
Remove unused ringbuffer

lulf:
Less is more? Looked like this was not in used anywhere.


Dirbaio (approved):
oh yeah this was for the buffereduart before we moved everything to the atomic one. nice find.

PR #2053, opened 2023-10-13
STM32: Fix regression in advanced timer to enable output of PWM signa…

### LINKED ISSUE #2004: pwm for advanced timers inop
> Hi, I have been playing around with the timers for PWM generation on stm32. I have gotten PWM working on general purpose timers(TIM2 to TIM17. I have notice that PWM is not working on advanced timers (TIM1 and TIM8). Upon further investigation I found that this is due to the MOE bit not being set in BDTR register. I have gotten it working with direct memory access now. What is the best way to add it to the HAL?. I can create a pull request if needed

jr-oss:
…l by partly reverting commit 74eb519

Fixes #2004

### DIFF
--- a/embassy-stm32/src/timer/complementary_pwm.rs
+++ b/embassy-stm32/src/timer/complementary_pwm.rs
@@ -71,7 +71,7 @@ impl<'d, T: ComplementaryCaptureCompare16bitInstance> ComplementaryPwm<'d, T> {
         this.inner.set_frequency(freq);
         this.inner.start();
 
-        this.inner.enable_outputs(true);
+        this.inner.enable_outputs();
 
         this.inner
             .set_output_compare_mode(Channel::Ch1, OutputCompareMode::PwmMode1);
--- a/embassy-stm32/src/timer/mod.rs
+++ b/embassy-stm32/src/timer/mod.rs
@@ -173,7 +173,7 @@ pub(crate) mod sealed {
                 }
             });
         }
-        fn enable_outputs(&mut self, _enable: bool) {}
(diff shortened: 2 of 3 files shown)


Dirbaio:
thanks for the PR!

Could you change `enable_outputs` to not take a `bool`? so it's just `fn enable_outputs(&mut self);`. 

The reason is it's misleading: having the method gives the illusion `CaptureCompare16bitInstance` supports turning outputs on/off, while it's not true, only the advanced ones do. If there's no `bool` arg, the function becomes just a "thing you must call to make it work", so it makes more sense for it to be noop in non-advanced timers, since there's nothign required to "make it work".

jr-oss:
Do you want it to look like this?
```rust 
           fn enable_outputs(&mut self) {
                use crate::timer::sealed::AdvancedControlInstance;
                let r = Self::regs_advanced();
                r.bdtr().modify(|w| w.set_moe(true));
            }
```
That would be a breaking change with some follow-up
Currently, there is still the bool parameter for all timers as the default implementation:
https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/timer/mod.rs#L176

Dirbaio:
yes, like that. breaking change is fine.

jr-oss:
I just updated the branch.
Not that many places to change.

Do you want to have it squashed?

Dirbaio:
doesn't matter, up to you!

jr-oss:
I prefer 2 commits to make it obvious, that it not only fixes a regression, but also changes the implementation.

PR #2057, opened 2023-10-14
Wake Stack's polling task after queueing a DNS query

bugadani:
The query is transmitted on the next `Stack::poll` which may or may not be happening in a timely manner. This PR is an attempt to fix this delay.


Dirbaio (approved):
thanks!

PR #2060, opened 2023-10-14
stm32: expand rcc mux to g4 and h7

xoviat:


embassy-stm32/src/rcc/g4.rs
```diff
@@ -357,5 +357,8 @@ pub(crate) unsafe fn init(config: Config) {
         adc: adc12_ck,
         adc34: adc345_ck,
         rtc,
+        sysclk: None,
+        noclk: None,
```
Dirbaio, line 361: "noclk" is not a clock, i'd suggest treating it specially in the match, just panic.

xoviat: sorry, I should have marked this as draft. this change was not going to be merged

PR #2068, opened 2023-10-15
Constify UsbDevice Config::new (and clippy fixes) in embassy-usb

barafael:
Just a constification of this one function I think I need in a const context. Plus a whole load of clippy fixes. The earlier commit is just normal clippy, the later one contains more pedantic fixes, some of them really do improve the quality IMHO, but you be the judge and let me know.


Dirbaio (approved):
thanks1

PR #2083, opened 2023-10-17
fiix stm32wl rng example to use MSI instead of LSI

lucasgranberg:
Small change to avoid confusion for people trying the random example for stm32wl.


Dirbaio:
I just did it in #2082 :sweat_smile:

lucasgranberg:
oh :D

PR #2086, opened 2023-10-17
Fixed FPU setting for stm32g4

Tupelov:


Dirbaio:
This is intentional to make docs build faster. If some chips are -eabi and others are -eabihf, all the dependencies are built twice, which is slower for no benefit since -eabihf doesn't change docs output.

PR #2090, opened 2023-10-18
bypass `ODEN` in `rcc::init()` if chip doesn't have it

eZioPan:


eZioPan:
don't know if there are any other side effects after this change, at least it allow me run `embassy_stm32::init()` without blocking.

Dirbaio (approved):
looks good, yup! Thank you! :)

PR #2098, opened 2023-10-21
stm32: fix opamp bug in docs build

xoviat:


Dirbaio:
Could you add the previously-failing chip to `ci.sh`? so that it doesn't break again.

PR #2099, opened 2023-10-21
stm32/rtc: misc and low-power cleanup 

xoviat:
- move to bypassshadow by default
- cleanup misc. stop and rtc code


Dirbaio:
oops failed because I deployed teleprobe right now. sorry.

PR #2104, opened 2023-10-23
net: Reset DHCP socket when the link up is detected

glaeqen:
Previously, because DHCP DISCOVER is sent before the link is established, socket has to timeout first. Which takes extra 10 s.

Now if the state of the link changed to up, socket is explicitly reset so the DISCOVER is repeated much earlier and DHCP configuration is acquired much faster.


Dirbaio (approved):
oh, nice find! thank you! :)

PR #2105, opened 2023-10-23
Fix stm32 uart set_config

andresv:
* `CR1` `UE` must be false when setting word length and parity bits.
* use `type ConfigError = ConfigError` instead of `type ConfigError = ()`
* In mod.rs there is `fn configure()` which uses write to set `CR1`. This disables `RXNEIE` and `IDLEIE` in `CR1` and those must be re-enabled again. Otherwise receive does not work anymore.

# TODO (done)

- [x]  What to do with `fifoen`, which one is correct?

https://github.com/embassy-rs/embassy/blob/ca27590afac33019dbc81489117b33fd7e3063a5/embassy-stm32/src/usart/mod.rs#L998-L999
vs
https://github.com/embassy-rs/embassy/blob/ca27590afac33019dbc81489117b33fd7e3063a5/embassy-stm32/src/usart/buffered.rs#L236-L237

Edit: `usart_v4` is correct.

# TEST
Tested with stm32g081 by reconfiguring baudrate, parity and stop bits.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -116,28 +116,28 @@ pub struct BufferedUartRx<'d, T: BasicInstance> {
 
 impl<'d, T: BasicInstance> SetConfig for BufferedUart<'d, T> {
     type Config = Config;
-    type ConfigError = ();
+    type ConfigError = ConfigError;
 
-    fn set_config(&mut self, config: &Self::Config) -> Result<(), ()> {
-        self.set_config(config).map_err(|_| ())
+    fn set_config(&mut self, config: &Self::Config) -> Result<(), Self::ConfigError> {
+        self.set_config(config)
     }
 }
 
 impl<'d, T: BasicInstance> SetConfig for BufferedUartRx<'d, T> {
     type Config = Config;
-    type ConfigError = ();
+    type ConfigError = ConfigError;
 
-    fn set_config(&mut self, config: &Self::Config) -> Result<(), ()> {
-        self.set_config(config).map_err(|_| ())
+    fn set_config(&mut self, config: &Self::Config) -> Result<(), Self::ConfigError> {
+        self.set_config(config)
     }
 }
 
 impl<'d, T: BasicInstance> SetConfig for BufferedUartTx<'d, T> {
     type Config = Config;
-    type ConfigError = ();
+    type ConfigError = ConfigError;
 
-    fn set_config(&mut self, config: &Self::Config) -> Result<(), ()> {
-        self.set_config(config).map_err(|_| ())
+    fn set_config(&mut self, config: &Self::Config) -> Result<(), Self::ConfigError> {
+        self.set_config(config)
     }
 }
 
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -233,7 +233,7 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
         configure(r, &config, T::frequency(), T::KIND, true, true)?;
 
         r.cr1().modify(|w| {
-            #[cfg(lpuart_v2)]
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/usart/buffered.rs
```diff
@@ -254,7 +254,17 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
     }
 
     pub fn set_config(&mut self, config: &Config) -> Result<(), ConfigError> {
-        reconfigure::<T>(config)
+        reconfigure::<T>(config)?;
+
+        T::regs().cr1().modify(|w| {
+            #[cfg(lpuart_v2)]
```
xoviat, line 260: Is it `#[cfg(any(lpuart_v2, usart_v4))] `?

Dirbaio: lpuart doesn't exist as a peripheral anymore, it's usart_v4.

Dirbaio, line 260: should be usart_v4

embassy-stm32/src/usart/buffered.rs
```diff
@@ -254,7 +254,17 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
     }
 
     pub fn set_config(&mut self, config: &Config) -> Result<(), ConfigError> {
-        reconfigure::<T>(config)
+        reconfigure::<T>(config)?;
+
+        T::regs().cr1().modify(|w| {
```
Dirbaio, line 259: could you put this reg write in `reconfigure` instead of pasting it 3 time?

Dirbaio (changes requested):

embassy-stm32/src/usart/buffered.rs
andresv: `reconfigure` is universally used for all uart flavors. For example when buffered uart is created it calls `configure` and then sets other registers that are needed for this driver:
https://github.com/embassy-rs/embassy/blob/ca27590afac33019dbc81489117b33fd7e3063a5/embassy-stm32/src/usart/buffered.rs#L233-L241

So putting it to `confgure` is not a good place. I'll check if I can refactor it a little to remove copy-pasting.

embassy-stm32/src/usart/buffered.rs
```diff
@@ -233,9 +233,6 @@ impl<'d, T: BasicInstance> BufferedUart<'d, T> {
         configure(r, &config, T::frequency(), T::KIND, true, true)?;
 
         r.cr1().modify(|w| {
-            #[cfg(lpuart_v2)]
-            w.set_fifoen(true);
-
```
andresv, line 238: This happens in `configure`.

embassy-stm32/src/usart/buffered.rs
andresv: I cleaned it up a little by removing redundant `w.set_fifoen(true); ` which is set in `configure` anyway.
There isn't a good place where to put those copy pasted lines without introducing more complexity.
`configure` configures all the basic stuff which is used in all uart flavours and then extra configuration is done in `buffered.rs` which is needed for this specific impl.

I would keep it as is without adding extra stuff to basic configuration.
Although I was thinking that maybe we can remove those `self.set_config` functions altogether and only rely on `SetConfig` trait as in [here](https://github.com/embassy-rs/embassy/blob/ca27590afac33019dbc81489117b33fd7e3063a5/embassy-stm32/src/usart/buffered.rs#L117C18-L142).

Then all the configuration related stuff is neatly together and there is less chance that somebody forgets to update something. However then again `SetConfig` trait must be imported by the user which is inconvenient.

Dirbaio: oh okay! I had missed that, sounds good then! thanks :)

Dirbaio (approved):
thank you!

PR #2110, opened 2023-10-24
Lora phy spidevice

CBJamo:
This depends on the breaking change in embassy-rs/lora-phy#35, thus the patch sections and draft status. 

I suspect that the generics and error handling can be done better in the new SubGhzSpiDevice, but I want to get something up, as I don't have the hardware to test this.

### DIFF
--- a/embassy-embedded-hal/src/shared_bus/asynch/spi.rs
+++ b/embassy-embedded-hal/src/shared_bus/asynch/spi.rs
@@ -66,15 +66,15 @@ where
         let mut bus = self.bus.lock().await;
         self.cs.set_low().map_err(SpiDeviceError::Cs)?;
 
-        let op_res: Result<(), BUS::Error> = try {
+        let op_res: Result<(), Self::Error> = try {
             for op in operations {
                 match op {
-                    Operation::Read(buf) => bus.read(buf).await?,
-                    Operation::Write(buf) => bus.write(buf).await?,
-                    Operation::Transfer(read, write) => bus.transfer(read, write).await?,
-                    Operation::TransferInPlace(buf) => bus.transfer_in_place(buf).await?,
+                    Operation::Read(buf) => bus.read(buf).await.map_err(SpiDeviceError::Spi)?,
+                    Operation::Write(buf) => bus.write(buf).await.map_err(SpiDeviceError::Spi)?,
+                    Operation::Transfer(read, write) => bus.transfer(read, write).await.map_err(SpiDeviceError::Spi)?,
+                    Operation::TransferInPlace(buf) => bus.transfer_in_place(buf).await.map_err(SpiDeviceError::Spi)?,
                     #[cfg(not(feature = "time"))]
-                    Operation::DelayUs(_) => return Err(SpiDeviceError::DelayUsNotSupported),
+                    Operation::DelayUs(_) => Err(SpiDeviceError::DelayUsNotSupported)?,
                     #[cfg(feature = "time")]
                     Operation::DelayUs(us) => embassy_time::Timer::after_micros(*us as _).await,
                 }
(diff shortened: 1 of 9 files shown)


embassy-lora/src/stm32wl.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        let op_res: Result<(), stm_spi::Error> = try {
+            for op in operations {
+                match op {
+                    Operation::Read(buf) => self.bus.read(buf).await?,
+                    Operation::Write(buf) => self.bus.write(buf).await?,
+                    Operation::Transfer(read, write) => self.bus.transfer(read, write).await?,
+                    Operation::TransferInPlace(buf) => self.bus.transfer_in_place(buf).await?,
+                    #[cfg(not(feature = "time"))]
+                    Operation::DelayUs(_) => return Err(SpiDeviceError::DelayUsNotSupported),
```
Dirbaio, line 74: try blocks catch `?` but not `return`. so this will return without deasserting NSS.

embassy-lora/src/lib.rs
```diff
@@ -1,12 +1,17 @@
...
+//#[cfg(feature = "stm32wl")]
+#![feature(try_blocks)]
 //! embassy-lora holds LoRa-specific functionality.
 
 pub(crate) mod fmt;
 
 /// interface variants required by the external lora physical layer crate (lora-phy)
 pub mod iv;
 
+//#[cfg(feature = "stm32wl")]
```
lucasgranberg, line 12: this is probably not supposed to be commented

embassy-lora/src/stm32wl.rs
Dirbaio: oh, oops. yes please!  :D

lucasgranberg:
Tested on wio-e5 module (stm32wle). Join sent to network so it should be good to go.

embassy-lora/src/lib.rs
CBJamo: Absolutely correct, fixed.

embassy-lora/src/stm32wl.rs
CBJamo: I took a stab at this, it works but I feel like there should be a less verbose/cleaner way of doing it.

embassy-embedded-hal/src/shared_bus/asynch/spi.rs
```diff
@@ -66,15 +66,19 @@ where
...
 
-        let op_res: Result<(), BUS::Error> = try {
+        let op_res: Result<(), Self::Error> = try {
             for op in operations {
                 match op {
-                    Operation::Read(buf) => bus.read(buf).await?,
-                    Operation::Write(buf) => bus.write(buf).await?,
-                    Operation::Transfer(read, write) => bus.transfer(read, write).await?,
-                    Operation::TransferInPlace(buf) => bus.transfer_in_place(buf).await?,
+                    Operation::Read(buf) => bus.read(buf).await.map_err(|e| SpiDeviceError::Spi(e))?,
```
Dirbaio, line 72: you can do `t.map_err(SpiDeviceError::Spi)` which is slightly shorter.

lulf (commented):
Nice work, looks good to me. Would it make sense to move it to embassy-stm32 eventually when this and the bits have landed?

Dirbaio:
yes! there was some discussion in the lora matrix about this. I think the best would be to split the trait to a separate `lora-phy-driver` crate so embassy-stm32 can depend just on that, so it needs bumping less often. 

And then remove embassy-lora, and perhaps move the lora examples to the lora repo.

plaes:
@CBJamo I guess we can close this now as all these resources have been migrated under [lora-rs](https://github.com/lora-rs/lora-rs/).

PR #2111, opened 2023-10-24
time: Add tick rates in multiples of 10 kHz

yodaldevoid:
I needed a tick rate of 40 kHz for the in order to use the watchdog counters (they are badly named, I know) running off the internal low-speed oscillator of the Infineon PSoC 4100S microcontrollers as a time source. I figured I'd just fill the X * 10000 Hz "hole" while I was at it.


Dirbaio (approved):
thanks!

PR #2113, opened 2023-10-25
WIP: NRF NFCT Driver

i404788:
Currently submitting this as an early draft to receive feedback & to notify of it being in-progress.

NFC frame parsing may be split off into a separate crate, but for now I'm adding here as well.


badrbouslikhin:
Hi, I'm currently working on integrating NFC into one of my projects as well.
I noticed this open NFC PR and would be happy to review the changes and contribute in any way I can to help get it merged. 

What's the current status?

i404788:
@badrbouslikhin The HAL-side should be mostly complete but untested, the tag emulation/protocol/parsing is entirely boilerplate or missing. I haven't ended up needing it yet so I didn't continue development.

The main issue I had was that I needed host-type (initiator) NFC rather than a tag-type (responder) NFC (nrf only supports tag-type), so I ended up using a ST25R39xx.

badrbouslikhin:
Thanks for the feedback @i404788.
I'll dig a little bit more and get back to this.

turbocool3r:
Hey @badrbouslikhin, I'd like to use that in a project of mine and could work on this during weekends, but I don't have much rust embedded experience so I would appreciate some review and guidance. That PR being upstream would be really nice

turbocool3r:
I've spent a couple of weekends trying to bring up NFCT on embassy and so far I see no hope. If you're going to try, you better try hard.

My idea was to port the firmware of [this](https://github.com/RfidResearchGroup/ChameleonUltra) nice device to avoid the memory corruption issues that are present in the original firmware written in C and to also add nice unit tests and minimal fuzzing for card emulators which were only possible with a proxmark before.

After a month I ended up with a couple of problems: 

1. Transmitting a response to a command is delayed for some reason and only happens after the next command is received. 
2. Sometimes the RXD.AMOUNT register is filled with 1 bits which is impossible since the largest DMA buffer size is 257 bytes.
3. Sometimes frames contain garbage data.

While the last two problems could be ignored, I couldn't solve the first one in any way.

What I tried:

1. A delay after activating the peripheral (setting `TASKS_ACTIVATE`). That worked and prevented garbage data from being received before activation. That thing is mentioned in the errata.
2. Setting undocumented `TASKS_STOPTX` before transmitting something or after receiving a frame without answering it. Didn't work even though this is the official solution in the SDK.
3. Inserting a small delay before processing the interrupt. Didn't work.
4. Setting STARTTX in a loop. Obviously failed.
5. Setting registers related to NFCT erratas that are also set by `SystemStart` in the SDK. Didn't work.
6. Disabling all the features of the ChameleonUltra firmware and seeing if NFCT breaks. It doesn't.

So far I'm in despair.

i404788:
Hey @turbocool3r,

Sorry to hear you are having a hard time. For your problem 1 (maybe you already considered this, I haven't looked at your fork): this sounds like it might be a task scheduling issue, in async rust whenever you await a function it will check the `poll()` and see if it's ready, if not it will continue to the next active task. This also means that you can get weird concurrency & timing issues if you aren't careful about things like spin loops. It's possible that you have the following situation:
```mermaid
sequenceDiagram
    participant Scheduler
    box Blue Task 1
    participant Rx1
    participant Tx1
    end
    box Purple Task2
    participant Rx2
    participant Tx2
    end
   Scheduler -->> Rx1: Blocking
   Rx1 -->> Tx1: NotReady (Defer)
   Scheduler -->> Rx2: Blocking
   Rx2 -->> Tx2: NotReady (Defer)
   Scheduler -->> Tx1: Ready (Now Tx1 is processed)
```

If I find some time I'm happy to take a look at your code.

turbocool3r:
hey @i404788, thanks, that's very much appreciated! just pushed it to https://github.com/turbocool3r/embassy in case you want to take a look (I've also merged main, but the changes are in `nfc.rs` mostly). 

As for the race condition I doubt this is the case since since that last commit I pushed I also tried doing everything in the interrupt handler and I observed the same thing.

i404788:
@turbocool3r I think you forgot to push your changes (at least I can't see any difference in nfc.rs). 

If not a race/blocking condition the only other thing I can imagine it being given your description is a register trigger that somehow happens in the wrong place or time (e.g. Recv somehow has the DMA start for the TX side).

turbocool3r:
Indeed I didn't commit. Just pushed the changes. Here's the project using this code https://gist.github.com/turbocool3r/11d50a786357db5e79f0129cffd33e1e.

turbocool3r:
@i404788 I just tried it on nrf5340-dk and it seems the same thing is happening there. Did you have any success?

i404788:
I just picked it back up from your message, it seems like (at least in my simplified example) it's not sending the response fast enough (generates a `FRAMEDELAYTIMEOUT` error status) when polling tx. From the product spec 7.20.7 it seems like there could be multiple reasons for this, so I'll need to investigate further.

i404788:
I've continued testing, I'm now past auto-collision/selection (using the built-in autocol resolver). I think the issue was that it was both using the internal auto-collision and transmitting frames response to it manually. I am able to receive the `E0 RATS` command consistently; however it seems fairly fragile, changes in tx_frame can cause issues in rx_frame without tx_frame ever having been called, so I suspect there is a memory corruption somewhere. 

It also seems like the ATS responses I've tried so far are either invalid or too late (field_lost hit or ignoring that just no follow up from the initiator). After that it just goes back to REQA polling. I'll see if I can get it past select.

turbocool3r:
@i404788 I've noticed that sometimes RX.AMOUNT register is all ones when a frame is received. The hardware autocollision also worked for me but after one or two frames there is usually some issue which causes tx interrupt to only arrive when the next fielddetected event happens

i404788:
@turbocool3r Hmm I don't think I've encountered the `RX.AMOUNT` issue mine has been fairly accurate as far as I can tell. I see what you mean with the fielddetect thing, I think it has to do with the `Frame timing controller` settings, setting it to `FreeRun` before starting the tx_frame/rx_frame seems to fix that.

Dirbaio:
:wave:  i'm also interested in this. Took a look at the code, i've ported it to the new PAC here https://github.com/embassy-rs/embassy/tree/nfct

Now trying to get it to work...

turbocool3r:
@Dirbaio wow, it works! Just made it emulate mifare ultralight and android accepts it. There seem to be some problems when you don't send a response, but otherwise that works great.

Dirbaio:
#3552

PR #2118, opened 2023-10-27
Feature i2c slave

smeenka:
Hereby I created a proposal for an async i2c slave implementation. It is not finished yet, but working for my use case. Work in progress.
Requirements are:
* Must be able to handle 2 addresses  read and write.
* Must be able to handle arbitration lost
Implementation is fully autonomous, after enabling. It runs fully in the interrupt context, is fast (no buffer copy in interrupt context)

Work to do still:
* Not based on a generic hal interface. (not available at this moment as far as I know)
* Handling of NACK not fully under control. I do not yet understand fully all the reload / sbc / nack combinations.But for the slave side the case is working for an actual implementation.
* Master write_read transaction not yet working
* Buffer size is allocated at compile time, and all buffers are equal. I can imagine this should be done with a macro and that the buffer size for address 1  can be different than address 2

For test purpose I did create 3 tests in the STM32G0 directory, for target nucleo-g070rb. Sometimes you have to modify the led and the uart as I also use a slightly different target.


smeenka:
Unclear why Oamsk not found in i2c::vals.

kalkyl:
Cool! It would be nice if we could have the same API as in the i2c slave implementations in embassy-nrf (twis) and embassy-rp, could you look into that?
https://github.com/embassy-rs/embassy/blob/main/examples/nrf52840/src/bin/twis.rs
https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/i2c_slave.rs
Cheers!

smeenka:
I looked into the api of the  embassy-nrf. But that is not the kind of interface I propose.
The i2c slave should be able to handle bursts of transactions in real time. The api in embassy-nrf does not support this.
One even has to decide in user code if the next transaction is read or write. Thats very limiting, as the slave cannot know this on beforehand, if the master code is a bit more complex than trivial.

I completely rewritten the code. Now it is very much in embassy style, using a channel to communicate beween interrupt context and user context.
Separated the slave code from the master code. Transaction ownership goes from user code, into the device, into a channel then back to the user code.
I understand that in the current form it cannot be merged with master, as too far from the proposed api.
You can see my code as a proposal, as a piece of research. One step further in this approach and the async api is fully based on channels. 
You can close this merge request, no problem for me.

kalkyl:
> I looked into the api of the embassy-nrf. But that is not the kind of interface I propose. The i2c slave should be able to handle bursts of transactions in real time. The api in embassy-nrf does not support this. One even has to decide in user code if the next transaction is read or write. Thats very limiting, as the slave cannot know this on beforehand, if the master code is a bit more complex than trivial.
> 
Hmm, i don't know if you misunderstood something about the nrf and rp api:s, this is not really how it works. The slave doesnt need to know anything in advance about the next transfer, you simply tell it how to react to different commands issued by the master, so it's all "real time". Or maybe i don't understand what you mean? :) However the current API i've been using in production in various projects for over a year and it's been very nice to work with. But maybe you have some special use case it doesn't cover?

smeenka:
Of course I do not have full understanding of the implementation of the embassy-nrf implementation. I have no hands on experience with it. So possible I miss something, so possible the nrf-implementation does support real time transactions. 

As long as master and slave both are made with embassy its fine, the interface like embassy-nrf will work fine. When the master is written with a RTOS like RTIC, or a NO OS like Arduino framework, the master can be faster (in bursts) than the slave in embassy can handle. You will have a chance to get an overrun for write or an under run for read. Not always but sometimes. This leads to difficult to find errors in the total application (master + slave). 

What I try to tackle with my proposed implementation is to guarantee real time support up to a certain degree. 
Note that my proposed implementation also can have overruns: make the 10 ms sleep in the master lower, and the slave will run into NoTransaction errors for a read. transaction. The long sleep is needed for the printing in the slave, which costs a lot of time.

As said before, keep my proposal as a proposal. Or cherry pick some ideas.

kalkyl:
I've had no problems interfacing to non-embassy masters, it uses clock-stretching as per the i2c spec until the response is written. How did you test it and how did it fail?

smeenka:
With RTIC all user code runs in interrupt context, with the help of i2c clock stretching. This is not the case with embassy. When the interrupt is finished (and stop signal is sent by the master) the master is ready for the next transaction. The slave is not ready, because the user task still has to be awoken, and the user code probably has to prepare for the next transaction. Awaken the user code can take at least 350 us, and probably much more, in a busy system. Within 350 us the master can finish a transaction of 2 bytes, where the slave still has to start handling the previous transaction.
So if the master is faster then the slave, you will encounter over or under runs. Its easy to simulate with the tests I did write. Decrease the wait time of 10 ms between each test in the master, and the master read tests will fail.
Because the slave test is very slow due to the printing to serial.

Dirbaio:
The slave should stretch the clock to signal the master it has to slow down. With it, communication should never fail no matter how slow the slave is.

If it doesn't, it's a bug and should be investigated. The hardware is capable of clock stretching both in the master and slave modes.

kalkyl:
It definately does do clock stretching (see "suspend" and "resume" in the nrf code).
Can you share some example code where it fails and what hardware version, maybe there's something in the errata for the chip you're using or maybe has to do with something else in how it's being used? As i said, i've been using it a lot on nrf52840 and nrf9160 and haven't seen any issues so far, so would be great to be able to reproduce and fix it!

smeenka:
I do not have any hands-on with nrf chips, nor do I have any nrf devices. Only stm32 devices.
Looking into the code for the nrf chip I seed that you are fully right, clock stretching is done until a stop condition is detected. There is no bug or whatever, only a bit misunderstanding from my side. Sorry for that.

examples/stm32g0/src/bin/i2c_slave_arbitration.rs
```diff
@@ -0,0 +1,127 @@
...
+          p.DMA1_CH2,
+          usart::Config::default(),
+      )
+      .unwrap();
+    */
+    let (tx, _rx) = uart.split();
+
+    let mut writer = SerialWriter::new(tx);
+
+    writeln!(
```
Dirbaio, line 70: this should use defmt for logging, like all the other examples in the repo.

bsodmike:
Hi @smeenka thanks for your work on this; what's stopping this from being merged into Embassy, I.e what's left that needs to be sorted out or discussed?

Thanks!

PR #2120, opened 2023-10-29
time: add `MockDriver` for testing purposes

andresovela:
This PR adds a new driver to embassy-time that exposes an API to manually advance the time.

The goal is to be able to use this driver to test code that works with `Instant` and `Duration`. I think if support for `Timer` functionality is needed, people would probably be well served with the driver provided by the `std` feature.


Dirbaio (approved):
thank you!

PR #2122, opened 2023-10-30
stm32: Add some documentation to `low_power`

bgamari:
This begins to explain the intended operation of the low-power executor.

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -1,3 +1,50 @@
+/// The STM32 line of microcontrollers support various deep-sleep modes which exploit clock-gating
+/// to reduce power consumption. `embassy-stm32` provides a low-power executor, [`Executor`] which
+/// can use knowledge of which peripherals are currently blocked upon to transparently and safely
+/// enter such low-power modes (currently, only `STOP2`) when idle.
+///
+/// The executor determines which peripherals are active by their RCC state; consequently,
+/// low-power states can only be entered if all peripherals have been `drop`'d. There are a few
+/// exceptions to this rule:
+///
+///  * `GPIO`
+///  * `RCC`
+///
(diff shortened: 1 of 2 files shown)


bgamari:
One thing that needs to be addressed here is the reference to `STOP2`. Specifically, this reference is naturally platform-dependent and I have been unable to work out which platform this reference is intended to refer to. I can see no such state in the STM32L0 or STM32G0 reference manuals.

@xoviat, perhaps you can clarify?

xoviat:
> One thing that needs to be addressed here is the reference to STOP2. Specifically, this reference is naturally platform-dependent and I have been unable to work out which platform this reference is intended to refer to. I can see no such state in the STM32L0 or STM32G0 reference manuals.

The conceptual understanding I have is that STOP1 means STOP with the LP* peripherals like LPUART, etc., and STOP2 is STOP without the LP* peripherals. STOP2 consumes less power than STOP1. If the chip has no LP* peripherals, then there is no STOP1, there is only STOP2 (or STOP), which requires all peripherals to be dropped.

The reason that GPIO is not included in the refcount is that (EXTI aside, which is another matter), GPIO is not async. The specific requirement (aside from time, which is different), is that the chip cannot enter STOP when inside of an `async fn` that depends on the activation of a peripheral. The reason for this is that the interrupt is suspended for the peripheral during STOP, which means that the `async fn` would fail.


With that said, would you like to amend this PR given that explanation, or would you like me to merge it, or explain more? Either way is fine.

bgamari:
Sorry for the latency, @xoviat; I was away from civilization last week.

My concern was rather that STOP2 is mentioned nowhere in the TRMs that I have read. Perhaps you rather meant STOP1 (which is defined in both the STM32L0 and STM32G0 TRMs)? In general, which MCU line are you developing against?

PR #2123, opened 2023-10-30
Re-add impl_trait_projections

bugadani:
Unfortunately, Xtensa (ESP32) needs to use a forked compiler. This compiler follows the 6-week release schedule of stable rustc, but with nightly features enabled. This means, for us it would be best if the feature gates could be kept until the feature hits stable rust, so we could benefit of the latest embassy changes, too.


bugadani:
@Dirbaio this should be ready for you now

Dirbaio (approved):
thank you!

PR #2125, opened 2023-10-30
stm32: RTC fix DoW range off-by-one error

MabezDev:
Unsure if this is true for other chips, but for the stm32f2 RTC, the weekday units field is between `0b001` and `0b111`.

Always fun to find a bug that only happens on a Sunday :upside_down_face:.


lucasgranberg:
Nice catch. Same on stm32wl
image

Dirbaio (approved):
that RM has a bug... Monday *is* the forbidden day of the week :skull: 

jk. LGTM, thank you!

PR #2130, opened 2023-10-31
time: fix MockDriver::now()

andresovela:
When I implemented this I used the std driver as a reference, but now I realize that the implementation of `now()`should use `as_ticks()` instead of `as_micros()`.


Dirbaio (approved):
thanks!

PR #2133, opened 2023-11-01
Reinstate rcc::Config adc_clock_source field

ShakenCodes:
The rcc::Config adc_clock_source was removed from the STM32WL when the rcc/wl.rs was consolidated into rcc/l4l5.rs. This field was omitted from the rcc::Config on the other family members covered in rcc/l4l5.rs. This brings the field and supporting enums back into rcc::Config.
Changes per discussion with dirbiao in matrix channel.

### DIFF
--- a/embassy-stm32/src/rcc/l4l5.rs
+++ b/embassy-stm32/src/rcc/l4l5.rs
@@ -4,8 +4,8 @@ pub use crate::pac::rcc::vals::Clk48sel as Clk48Src;
 #[cfg(any(stm32wb, stm32wl))]
 pub use crate::pac::rcc::vals::Hsepre as HsePrescaler;
 pub use crate::pac::rcc::vals::{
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/rcc/l4l5.rs
```diff
@@ -52,7 +52,30 @@ pub struct Pll {
     pub divr: Option<PllRDiv>,
 }
 
-/// Clocks configutation
+#[derive(Clone, Copy)]
```
Dirbaio, line 55: please use the enum from the PAC directly in the public API. `pub use crate::pac::rcc::vals::{ Adcsel as AdcClockSource, ... };`

(old code redefined all the enums due to semver concerns about reexporting stuff from the PAC, but with time we've seen it's hard to maintain because these enums vary a lot across families, so the benefit doesn't outweigh the costs)

ShakenCodes: Fixed

Dirbaio (approved):
Thank you!

PR #2138, opened 2023-11-02
rp: Add PWM input example

kalkyl:
Example how to measure the frequency of an input signal using the PWM module.


Dirbaio (approved):
:heart_eyes:

PR #2139, opened 2023-11-04
Update RP2040 memory.x

jamesmunns:
The RP2040 has 264KiB of memory, not 256KiB.

Citation: <https://www.raspberrypi.com/documentation/microcontrollers/rp2040.html#technical-specification>

> 264kB on-chip SRAM in six independent banks


lulf (commented):
Good catch, thanks! Could you update these too? 

* `boot/bootloader/rp/memory.x`
* `boot/application/rp/memory.x`

Athorus:
Above 0x20040000, the 8 KB of SRAM are not striped ! So I am not sure it is a good idea to mix non striped SRAM with striped SRAM.
See page 24 of the datasheet https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf

jamesmunns:
@Athorus that's fair! on one hand: using unstriped ram primarily for the core might be good, if it is uncontended with DMA access.

On the other hand: we have no control over that for user code, and if we only give 4K of stack for each core, that might not reasonably be enough.

Dirbaio:
> Good catch, thanks! Could you update these too?
> 
>     * `boot/bootloader/rp/memory.x`
> 
>     * `boot/application/rp/memory.x`

ping @jamesmunns :)

Dirbaio:
> Above 0x20040000, the 8 KB of SRAM are not striped ! So I am not sure it is a good idea to mix non striped SRAM with striped SRAM. See page 25 of the datasheet https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf

examples are supposed to be simple, and "the default memory.x allows me to use all RAM" makes a lot of sense for examples. The datasheet specifically says the ram is laid out like that so you can pretend it's a contiguous 264k block. for the majority of applications it's likely to increase perf (moves core activity to sram4-5, while dma is most likely to use sram0-3). and only for very particular circumstances (several DMAs adding up more than 1 word/cycle out of stack) it might reduce perf. So IMO it's worth doing this change.



> With 256K, at the end of the program I got as expected : (HOST) INFO program has used at least 26.55/141.34 KiB (18.8%) of stack space
> 
> But with 264K I got this strange message ! (HOST) WARN program has used at least 4.00/4.00 KiB (100.1%) of stack space
> 
> In both cases, the program seems to work fine

This is https://github.com/knurling-rs/probe-run/issues/415 (the regions are separate in the probe-rs target definition). it's a bit moot now that probe-run is deprecated though.

Dirbaio (approved):
thank you!

Athorus:
> This is [knurling-rs/probe-run#415](https://github.com/knurling-rs/probe-run/issues/415) (the regions are separate in the probe-rs target definition). it's a bit moot now that probe-run is deprecated though.

Shoud I use probe-rs  instead of probe-run ?

Dirbaio:
yes https://ferrous-systems.com/blog/probe-run-deprecation/

PR #2150, opened 2023-11-06
Ensure TcpIo not blocking when reading into empty slice

bugadani:
`embedded_io_async::Read` should return immediately if its output buffer can't hold any data. The easiest way to ensure that is to handle the case at `TcpIo`'s level, changing its semantics a bit, but probably for the better.


Dirbaio (approved):
nice find, thank you!

PR #2151, opened 2023-11-06
check PLL settings before set VOS

eZioPan:
this is a tiny fix after commit [stm32/rcc: set highest VOS on some F4s with no overdrive.](https://github.com/embassy-rs/embassy/commit/b8679c0cc85e5eb65bd996ee18deac4a952b1b10)

VOS can be set when use PLL (depending on target speed) on any chips, not only the chips without over-drive.


Dirbaio (approved):
thank you1

PR #2155, opened 2023-11-07
stm32: resolve eth/v2 security bug

### LINKED ISSUE #2129: stm32/eth: v2 is vulnerable to memory corruption with vlan tags
See this https://www.mattkeeter.com/blog/2023-10-31-dma/

We should use the tail pointer register (`ETH_DMACRXDTPR`) to prevent DMA from reading descriptors too soon.

xoviat:
fixes #2129


xoviat:
Do not merge yet

PR #2157, opened 2023-11-07
rp: Add USB raw example + msos-descriptor to examples and usb-logger

kalkyl:
Added the msos-descriptor feature to the examples to allow WinUSB support.
This also led to having to add this feature to embassy-usb-logger aswell.


Dirbaio (approved):
thanks!

PR #2158, opened 2023-11-07
rp: Add USB raw bulk example

kalkyl:
Example of using USB without a pre-defined class, but instead using raw USB bulk transfers.


examples/rp/Cargo.toml
```diff
@@ -12,6 +12,7 @@ embassy-executor = { version = "0.3.1", path = "../../embassy-executor", feature
 embassy-time = { version = "0.1.5", path = "../../embassy-time", features = ["nightly", "unstable-traits", "defmt", "defmt-timestamp-uptime"] }
 embassy-rp = { version = "0.1.0", path = "../../embassy-rp", features = ["defmt", "unstable-traits", "nightly", "unstable-pac", "time-driver", "critical-section-impl"] }
 embassy-usb = { version = "0.1.0", path = "../../embassy-usb", features = ["defmt", "msos-descriptor"] }
+embassy-usb-driver = { version = "0.1.0", path = "../../embassy-usb-driver" }
```
Dirbaio, line 15: it's nicer to use use the `embassy_usb::driver` reexport instead of adding a direct dep on driver.

kalkyl: Ahh sorry!

Dirbaio (approved):

PR #2169, opened 2023-11-10
fix: make RP2040 I2C slave aware of serial writes

jcdickinson:
In general, this commit allows the I2C slave to recognize when writes are occuring in serial - and seperate them out into independant commands. In order to achieve this it:

* Corrects some setup parameters for the I2C state machine.
  - Enables clock stretching
  - Enables recommendations from the datasheet
* Participates in clock stretching. This is done by leaving the `R_RD_REQ` register of `IC_INTR_STAT` set until the slave has responded with some bytes (in `respond_to_write`).
* Recognizes `FIRST_DATA_BYTE` of `IC_DATA_CMD`, which indicates to a slave-receiver than the current byte is part of a new transaction.
* Uses a state machine heavily inspired by the one in rp-rs as an implementation detail. This is no more correct than the existing approach, but makes the code significantly easier to reason about.
* Two additional errors were added, both indicating that the provided buffer is too small for the message from the master-sender. As a convenience, they have the same assosciated data so that matching can be performed in the event that the call site handles messages in a consistent way.

### DIFF
--- a/embassy-rp/src/i2c_slave.rs
+++ b/embassy-rp/src/i2c_slave.rs
@@ -4,6 +4,7 @@ use core::task::Poll;
 
 use embassy_hal_internal::into_ref;
 use pac::i2c;
+use pac::i2c::vals::Speed;
 
 use crate::i2c::{
     i2c_reserved_addr, set_up_i2c_pin, AbortReason, Instance, InterruptHandler, SclPin, SdaPin, FIFO_SIZE,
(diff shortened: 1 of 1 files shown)


embassy-rp/src/i2c_slave.rs
```diff
@@ -161,53 +236,139 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
...
+    fn rx_not_empty(&self) -> bool {
         let p = T::regs();
+        self.pending_byte.is_some() || p.ic_status().read().rfne()
+    }
 
-        p.ic_clr_intr().read();
-        // set rx fifo watermark to 1 byte
-        p.ic_rx_tl().write(|w| w.set_rx_tl(0));
+    #[inline(always)]
+    async fn next_event(&mut self) -> I2cSlaveEvent {
```
Dirbaio, line 246: I don't think this state makes the logic clearer.

- There's events that are unused (start, restart...). 
- At start of `listen`, we know the bus state is always idle. Therefore there should be no need to save the state in `I2cSlave`. Saving it means at start of `listen` the state could be anything now. This makes reasoning about the core harder. 
- There's now two "layered" state machines, `next_event` and `listen`.

Could you do the required changes for the sisues you found without this refactor?  Thank you!

PR #2170, opened 2023-11-10
Fix for embassy net dns entries not being extensible

korken89:
Not sure if this is a good fix perse, but it allows for extending DNS setting in `smoltcp` up to 8 entries. 
If this is not done these interfaces are hard linked to whatever the return of `smoltcp` is.

Other ways to achieve this is welcome. :)

### DIFF
--- a/embassy-net/CHANGELOG.md
+++ b/embassy-net/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - Avoid never resolving `TcpIo::read` when the output buffer is empty.
+- Update to `smoltcp` git.
+- Forward constants from `smoltcp` in DNS query results so changing DNS result size in `smoltcp` properly propagates.
 
 ## 0.2.1 - 2023-10-31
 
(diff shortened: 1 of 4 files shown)


lulf (commented):
How about `pub async fn query<const N: usize>(...) -> Result<Vec<IpAddress, N>>`?

This also reminds me that heapless is here used in a public API, which we should try to avoid, but one thing at a time perhaps :)

korken89:
Hi @lulf 

I tried that first but you get this error then:

```
error[E0308]: mismatched types
   --> src/lib.rs:592:9
    |
592 |         res
    |         ^^^ expected `N`, found `1`
    |
    = note: expected enum `Result<Vec<_, N>, dns::Error>`
               found enum `Result<Vec<_, 1>, _>`

For more information about this error, try `rustc --explain E0308`.
```

As this only can infer the `N` from external code, not internal that we get from `smoltcp`.

nilfit:
With https://github.com/smoltcp-rs/smoltcp/pull/855, we can use `smoltcp::config::DNS_MAX_RESULT_COUNT` for the size instead of choosing a new hard-coded size. Then dependent crates can freely choose which `smoltcp/dns-max-result-count-*` to use.

lulf (approved):

lulf:
Thanks for the fix 🚀

PR #2172, opened 2023-11-10
A simple_pwm update event interrupt to be able to synchronize with pwm

smeenka:
To able to synchronize with a pwm (for example sample motor current with ADC) I propose hereby an update interrupt for simple_pwm.
A test can be found in stm32g0 example directory, which shows the functionality.


examples/stm32g0/Embed.toml
```diff
@@ -0,0 +1,43 @@
+[default.probe]
```
Dirbaio, line 1: examples don't use cargo-embed, please don't add an Embed.toml

smeenka: Will remove. Also change the implementation, not to use the global static signal, but a signal in interrupt state context

smeenka: Added a reset function to pwm to be able to synchronize multiple timers to have the same phase.

Dirbaio:
LGTM, except could you change `State` to contain an `AtomicWaker` instead of a `Signal`? t's slightly faster/smaller. 

This means you'll have to use a manually-written poll_fn from the non-interrupt code:

- Enable interrupt
- poll_fn that does
   - register waker
   - if interrupt flag is set, clear it, return Poll::Ready
   - else return Poll::Pending

and the interrupt handler doe

- disable interrupt (NOT clear the interrupt flag like you'd usually do, leave it there so the task can see it)

This pattern is used by a lot of drivers in embassy-stm32 to handle interrupts.

Dirbaio:
Hello,

PR #2173, opened 2023-11-10
STM32 I2C: expose async API without needing "time" feature.

andresv:
Recently timeouts were added to STM32 I2C API but unfortunately after that inorder to use any async I2C API user must always opt in for `feature "time"` even if user wanted to use async API without timeouts.

Many people are using `embassy-stm32` with RTIC executor which has its own Monotonic time implementation and they might not want to opt in additional timer (using `feature "time"`), but still would like to use async I2C API (but without timeout support).

This PR exposes following methods even if `feature "time"` is not used:
```rust
pub async fn write(&mut self, address: u8, write: &[u8]) -> Result<(), Error>
pub async fn write_vectored(&mut self, address: u8, write: &[&[u8]]) -> Result<(), Error>
pub async fn read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Error>
pub async fn write_read(&mut self, address: u8, write: &[u8], read: &mut [u8]) -> Result<(), Error>
```
Notice that `feature "time"` enables additional methods with timeouts:
```rust
pub async fn write_timeout(&mut self, address: u8, write: &[u8], timeout: Duration) -> Result<(), Error>
pub async fn write_vectored_timeout(&mut self, address: u8, write: &[&[u8]], timeout: Duration) -> Result<(), Error>
pub async fn read_timeout(&mut self, address: u8, buffer: &mut [u8], timeout: Duration) -> Result<(), Error>
pub async fn write_read_timeout(&mut self, address: u8, write: &[u8], read: &mut [u8], timeout: Duration) -> Result<(), Error>
```

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -162,7 +162,6 @@ pub struct I2c<'d, T: Instance, TXDMA = NoDma, RXDMA = NoDma> {
     tx_dma: PeripheralRef<'d, TXDMA>,
     #[allow(dead_code)]
     rx_dma: PeripheralRef<'d, RXDMA>,
-    #[cfg(feature = "time")]
     timeout: Duration,
 }
 
@@ -226,6 +225,8 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
             rx_dma,
             #[cfg(feature = "time")]
             timeout: config.transaction_timeout,
+            #[cfg(not(feature = "time"))]
+            timeout: Duration::dummy_duration(),
         }
     }
 
@@ -339,21 +340,12 @@ impl<'d, T: Instance, TXDMA, RXDMA> I2c<'d, T, TXDMA, RXDMA> {
     }
 
     fn flush_txdr(&self) {
-        //if $i2c.isr.read().txis().bit_is_set() {
-        //$i2c.txdr.write(|w| w.txdata().bits(0));
-        //}
-
         if T::regs().isr().read().txis() {
             T::regs().txdr().write(|w| w.set_txdata(0));
         }
         if !T::regs().isr().read().txe() {
             T::regs().isr().modify(|w| w.set_txe(true))
         }
-
-        // If TXDR is not flagged as empty, write 1 to flush it
-        //if $i2c.isr.read().txe().is_not_empty() {
-        //$i2c.isr.write(|w| w.txe().set_bit());
-        //}
     }
 
     fn wait_txe(&self, check_timeout: impl Fn() -> Result<(), Error>) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


andresv:
Previous discussion: https://github.com/embassy-rs/embassy/pull/2033.

andresv:
@Dirbaio comments from matrix:

> I think the steps to untangle are
> 
> remove blocking busy-loops in async i2c (like the one I linked). Might need duplicating more code between blocking and async, but I think that's OK.
> This'll allow implementing timeouts with a embassy_time::with_timeout() at the top level. No need to pass check_timeout closures in async anymore, only in blocking.
> Remove the _timeout public methods.
> Change it so the regular methods (without _timeout) suffix use the timeout from config if time is enabled, no timeout otherwise.
> should be doable with just a dummy fn with_timeout that does nothing if time is not enabled, not just a whole dummy time module
> alternatively, i'd be OK with a PR that does steps 2-4 now, leaving blocking loops (step 1) in for now
> leaving the blocking loops with no timeout for now

andresv:
Removed pub `_timeout` API. Currently dummy_time is still there because various place still need to know about Instant and Duration, but I try to get rid of also these, although this means more copy pasting.

andresv:
Current state:
* messy `dummy_time` module is gone
* async `_timeout` API is removed, now only way to change timeout is using `Config`
* without `time` feature async API is available but there are no timeouts

I did not remove blocking busy loops. It is more tricky. To do it correctly essentially those:

```rust
while !T::regs().isr().read().tcr() {
    check_timeout()?;
}
```

Should be handled with interrups and in `poll_fn` which adds quite some complexity.
So I would rather tackle it with separate PR.

Let me know what are your thoughs about this and if current PR is reasonabe.

andresovela:
I think the `_timeout` APIs shouldn't be removed. They are useful on their own.

For example, on a given application it might be reasonable to configure your I2C driver to timeout after 100 ms. Until one day you need to add a 32 kB I2C write to support FW updates on a USB PD controller. This write transaction may take around 2.5 seconds on an I2C bus running at 100 KHz.

Without these APIs, the only option a user has is to increase the timeout for _all_ I2C transactions to 3 seconds or so, which may not be reasonable for that application. I'm sure there are other use cases besides the one above, it's just the first thing that I could think of.

ExplodingWaffle:
I’m pretty sure it’s possible to use [SetConfig](https://docs.embassy.dev/embassy-stm32/git/stm32f423vh/i2c/struct.I2c.html#impl-SetConfig-for-I2c%3C'd,+T%3E) and/or [I2cDeviceWithConfig](https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/i2c/struct.I2cDeviceWithConfig.html) for exactly this. (might even be able to jack up the clock speed for faster downloads, too)
Actually, now that I’m looking at it, maybe those shared_bus types should also implement SetConfig? or have some other way of changing it on the fly, like the peripheral does.

andresovela:
I don't see anything in `SetConfig` that affects the timeouts:

https://github.com/embassy-rs/embassy/blob/39c737162185adb4f30f18f700da08a55be6b55a/embassy-stm32/src/i2c/v2.rs#L1188-L1203

That being said, I can imagine `SetConfig` being improved to handle this as well, which I'd see as a good alternative as well.

Dirbaio (approved):
Yes, the idea is we'd allow changing the config at runtime, which includes the timeout.

PR #2180, opened 2023-11-13
stm32/gpdma: fix drop() to use documented method for aborting transfer

### LINKED ISSUE #2171: stm32/usart: RX DMA transfer hangs when RX pin is floating(?)
I have a STM32H503 connected via a USART to an external sensor. Here's my driver:

```rust
async fn driver<'d, T: usart::BasicInstance>(
    uart: impl Peripheral<P = T> + 'd,
    irqs: impl Binding<T::Interrupt, usart::InterruptHandler<T>> + 'd,
    rx: impl Peripheral<P = impl usart::RxPin<T>> + 'd,
    tx: impl Peripheral<P = impl usart::TxPin<T>> + 'd,
    rx_dma: impl Peripheral<P = impl usart::RxDma<T>> + 'd,
    tx_dma: impl Peripheral<P = impl usart::TxDma<T>> + 'd,
) {
    let config = {
        let mut config = usart::Config::default();
        config.baudrate = 115200;
        config
    };

    let mut uart = usart::Uart::new(uart, rx, tx, irqs, tx_dma, rx_dma, config).unwrap();

    // activate sensor
    uart.write(&[0xCD, 0x01, 0x0A, 0x0B]).await.unwrap();

    loop {
        let mut buf = [0u8; 8];
        uart.read(&mut buf).await.unwrap();
        info!("read: {:?}", buf);
    }
}
```

Everything works until I physically disconnect the sensor from the board. At this point, upon the next read, the application hangs at:

https://github.com/embassy-rs/embassy/blob/b3367be9c8f0eb54d500dc2e6f652ade05859088/embassy-stm32/src/dma/gpdma.rs#L337.

This is the read transfer trying to drop. It's as though `request_stop()` is ineffective!

I'm yet to get the scope on it but I assume on my board, the RX pin is floating. Presumably I should be pulling it high. Nonetheless, this behaviour shouldn't be happening. I would expect to see no RX, or garbage RX, not a hang.

tcbennun:
See e.g. STM32H503 RM section 15.4.4.

1. Write 1 into GPDMA_CxCR.SUSP
2. Poll GPDMA_CxSR.SUSPF until it is 1
3. Write 1 into GPDMA_CxCR.RESET

This fixes #2171 for me - instead of hanging forever, my UART read fails with a framing error (as it should!)

This *should* be the correct behaviour in any instance where a GPDMA transfer needs to be aborted.

## Why it was broken

Old behaviour step by step:

1. Reset all bits on CxCR except relevant interrupt enables.

This has the effect of writing 0 to EN, RESET and SUSP. Writing 0 to EN and RESET has no effect, per RM. Writing 0 to SUSP resumes the channel if it was already suspended (irrelevant in this case).

2. Poll TCF, waiting for it to be set

This is fine if one expects the transfer to proceed, but that may not be the case, and if Transfer is being dropped, the application will never see the data anyway.

In my case - #2171 - the transfer is being dropped due to an error in the source peripheral, so I would never expect TCF to be set! Hence, the hang on this tight polling loop.

### DIFF
--- a/embassy-stm32/src/dma/gpdma.rs
+++ b/embassy-stm32/src/dma/gpdma.rs
@@ -299,19 +299,15 @@ impl<'a, C: Channel> Transfer<'a, C> {
 
     pub fn request_stop(&mut self) {
         let ch = self.channel.regs().ch(self.channel.num());
-
-        // Disable the channel. Keep the IEs enabled so the irqs still fire.
-        ch.cr().write(|w| {
-            w.set_tcie(true);
-            w.set_useie(true);
-            w.set_dteie(true);
-            w.set_suspie(true);
+        ch.cr().modify(|w| {
+            w.set_susp(true);
         })
     }
 
     pub fn is_running(&mut self) -> bool {
         let ch = self.channel.regs().ch(self.channel.num());
-        !ch.sr().read().tcf()
+        let sr = ch.sr().read();
+        !sr.tcf() && !sr.suspf()
     }
 
     /// Gets the total remaining transfers for the channel


Dirbaio:
Thanks for tracking this down!

Could you try not changing the public API? We should try to keep it compatible between DMA/BDMA and GPDMA. Seems nothing in the HAL is relying on that because it mainly uses the async API, but users outside the HAL might. 

I think making `request_stop()` do suspend and `is_running()` return `!tcf && !suspf` should work.

I don't think reset should be required because `new_inner` already resets, but if it is I wouldn't make it public API.

tcbennun:
My bad @Dirbaio , all done.

You're right, the `new_inner` reset looks sufficient.

Dirbaio (approved):
thank you!

Issue #2176, opened 2023-11-12
cyw43: does not respond to IPv6 Neighbor Solicitation requests


sammko:
This is because the cyw43 is not configured with the required HW address for the solicited-node ipv6 multicast address. This is reasonably easy to do manually, provided that a method is added for this in cyw43 Control, something along the lines of:
```rust
    pub async fn add_mcast(&mut self, addr: [u8; 6]) {
        let mut buf = [0; 64];
        self.get_iovar("mcast_list", &mut buf).await;
        let mut n = u32::from_le_bytes([buf[0], buf[1], buf[2], buf[3]]);
        // check if already present
        for i in 0..n as usize {
            let start = 4 + i * 6;
            if buf[start..start + 6] == addr {
                return;
            }
        }
        let start_new = 4 + n as usize * 6;
        buf[start_new..start_new + 6].copy_from_slice(&addr);
        n += 1;
        buf[..4].copy_from_slice(&n.to_le_bytes());

        self.set_iovar_v::<80>("mcast_list", &buf).await
    }
```
and then:
```rust
    let address = link_local_from_hw(netdev.hardware_address());
    let mut mcast_hw = [0x33, 0x33, 0, 0, 0, 0];
    mcast_hw[2..].copy_from_slice(&address.address().solicited_node().0[12..]);

    control.add_mcast(mcast_hw).await;
```
Doing this works for me and the device starts responding to neighbor solicitations.

Ideally this would be done automatically by embassy-net, but I don't believe a suitable abstraction is in place. I'm not sure how that would look, I guess the Device would have a method for this and embassy-net would call it when an ipv6 config is provided?

Dirbaio:
oh wow, nice catch. I guess it slipped through because no tmany people are using ipv6 ;( 

It's very annoying that the multicast address depends on our ipv6 address. :( The the easiest solution would be to declare that embassy-net assumes the underlying device is in a "sane" configuration for ipv4/ipv6 including the multicast stuff, so we could just hardcode enabling multicast in cyw43 init, without adding any API. However that's not possible if cyw43 doesn't know our ipv6 addr. 

Maybe it's possible to enable multicast for the entire solicited-node range? at the cost of performance of course.

Otherwise we'd have to add that API yes...

sammko:
Adding all solicited-node multicast addresses is not viable, there's 2^24 of them (33:33: prefix with 32 bits of the solicited-node address, those being ff: and 24 bits from the original address).

[...]

PR #2184, opened 2023-11-14
cyw43: Add Control method to add multicast HW address

sammko:
Notably, a method to remove addresses is missing. I can add this if wanted but I ran out of time for now.

In reference to #2176.

### DIFF
--- a/cyw43/src/control.rs
+++ b/cyw43/src/control.rs
@@ -1,4 +1,5 @@
 use core::cmp::{max, min};
+use core::iter::zip;
 
 use embassy_net_driver_channel as ch;
 use embassy_net_driver_channel::driver::{HardwareAddress, LinkState};
@@ -14,6 +15,12 @@ use crate::{countries, events, PowerManagementMode};
 #[derive(Debug)]
 pub struct Error {
     pub status: u32,
+}
+
+#[derive(Debug)]
+pub enum AddMulticastAddressError {
+    NotMulticast,
+    NoFreeSlots,
 }
 
 pub struct Control<'a> {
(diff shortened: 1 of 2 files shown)


cyw43/src/control.rs
```diff
@@ -316,6 +322,50 @@ impl<'a> Control<'a> {
...
+        free[0] = address;
+        let n = n + 1;
+        buf[..4].copy_from_slice(&(n as u32).to_le_bytes());
+
+        self.set_iovar_v::<80>("mcast_list", &buf).await;
+        Ok(n)
+    }
+
+    /// Retrieve the list of configured multicast hardware addresses.
+    pub async fn list_mulistcast_addresses<'b>(&mut self, buf: &'b mut [u8; 64]) -> &'b [[u8; 6]] {
```
sammko, line 360: I'm not sure about this pattern of using a caller-provided buffer and returning a typed slice inside it. However, without `heapless::Vec` or something equivalent, I don't see a better solution.

cyw43/src/control.rs
```diff
@@ -16,6 +16,12 @@ pub struct Error {
     pub status: u32,
 }
 
+#[derive(Debug)]
+pub enum AddMulticastAddressError {
+    NotMulticast,
+    NoFreeSlots,
+}
+
```
sammko, line 24: Is this error handling strategy appropriate in this library? I.e. a separate error type for this use-case, not mixed with the other one above. I know it's a matter of convention, let me know what is best here.

Dirbaio: :+1:

cyw43/src/control.rs
Dirbaio: it's a bit strange yes, especially changing the type. in these cases it's more common to take a buffer and return how many slots have been filled:

`pub async fn list_mulistcast_addresses(&mut self, addresses: &mut [[u8; 6]; 10]) -> usize`

cyw43/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![no_std]
 #![no_main]
-#![feature(async_fn_in_trait, type_alias_impl_trait, concat_bytes)]
+#![feature(async_fn_in_trait, type_alias_impl_trait, concat_bytes, slice_as_chunks)]
```
Dirbaio, line 3: We shouldn't add more nightly features, the goal is for Embassy to work on stable when AFIT and TAIT are stable.

cyw43/src/control.rs
sammko: Thing is I need spaces in front of the addresses for the 4 bytes length. Maybe we can just pull in `heapless`, as it's a dependency of `embassy-net` anyway? I think that would offer the nicest API.

cyw43/src/lib.rs
sammko: Noted, I'll do it without. 👍

cyw43/src/control.rs
Dirbaio: hmm perhaps read to a stack buffer and copy the interesting part to the user buffer then? it uses 2x the memory, but you need that anyway if you return a heapless::Vec instead. Also, the compiler tends to be dumb about returning big things by value, also causing multiple copies of it on the stack.

sammko:
Should all be addressed now.

Dirbaio (approved):
thank you!

PR #2185, opened 2023-11-14
stm32/usart: impl {eio, eio_async}::Read for Uart, UartRx

tcbennun:
Hopefully self-explanatory. 🙂

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -1129,6 +1129,13 @@ where
     type Error = Error;
 }
 
+impl<T, RxDma> embedded_io::ErrorType for UartRx<'_, T, RxDma>
+where
+    T: BasicInstance,
+{
+    type Error = Error;
+}
+
 impl<T, TxDma, RxDma> embedded_io::Write for Uart<'_, T, TxDma, RxDma>
 where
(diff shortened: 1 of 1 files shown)


Dirbaio:
The raw Uart can't impl `embedded_io::Read` for a few reasons:

- It is is unbuffered, this means it can only receive while the task is actively waiting on a `.read().await` call. Data that's received between read calls is lost. The `embedded_io::Read` contract requires that data is not so easily lost, drivers using it can assume that.
- It must return early if less than `buf.len()` bytes are received. Drivers can call `read()` with a buffer bigger than the data they're expecting (for example, if they don't know the size of the data in advance) and get back a partially filled buffer with the data that arrived and process that. With this impl, they'd hang instead.


if you want `embedded_io::Read` there's [BufferedUartRx](https://docs.embassy.dev/embassy-stm32/git/stm32f303re/usart/struct.BufferedUartRx.html) (irq-driven) or [RingBufferedUartRx](https://docs.embassy.dev/embassy-stm32/git/stm32f303re/usart/struct.RingBufferedUartRx.html) (dma-driven)

tcbennun:
Gotcha. Worth a shot.

Might look at impling the ring-buffered driver for GPDMA then

PR #2186, opened 2023-11-14
Refactor bxcan split.

T-Bakker:
In this PR I have changed the way splitting the can bus works. This because I encountered a problem with the fact that the old implementation would result in `Mutex<_, CanTx<_>>` not being `Sync`. This is because:
- `Mutex<_, _>` is `Sync` when its content is `Send` (https://docs.rs/embassy-sync/latest/embassy_sync/mutex/struct.Mutex.html#impl-Sync-for-Mutex%3CM,+T%3E)
- `RefCell<_>` is `Send` but not `Sync` (https://doc.rust-lang.org/std/cell/struct.RefCell.html#impl-Send-for-RefCell%3CT%3E)
- A reference is `Send` when the value it points to is `Sync` (https://doc.rust-lang.org/nomicon/send-and-sync.html)

This means that putting a `CanTx` in a `Mutex` will result in the mutex being not `Sync`, which means that it is impossible to have multiple tasks share the mutex and concurrently access the can bus for Tx operations, which I need.

I have changed the `split(&self)` to `split(&mut self)`, which allows using the `bxcan` `split_by_ref(&mut)` function. This eliminates the need for the `RefCell` as the actual `bxcan` types are split. Once both `CanTx` and `CanRx` are dropped, `Can` becomes available again.

With this change is it no longer possible to share both a `Can` and a `CanTx`/`CanRx`/, nor is it possible to have multiple splits. I would argue that this is desirable:
- As the `CanTx` uses a `RefCell`, it will panic whenever there are concurrent accesses. This means that although the compiler allows multiple splits, the runtime does not and this limitation is also hidden. It is up to the application to guarantee that no concurrent accesses occur. I would argue that it is better to then use either a mutex of a cell to share a single `CanTx` as this makes the limitation clear to the user/application.
- In case of having both a `Can` and a `CanTx`/`CanRx`, I would argue that again this is not desirable as the limitations imposed by the `RefCell` still apply. A concurrent access to both the `Can` and the `CanTx` will result in a panic.

As the `flush_` functions are `&self`, they can no longer use the split as it is now `&mut self`. I have opted to use private static `_flush_` functions that thus can be used in a `&self` context.

I have tested this change and did not encounter problems.

### DIFF
--- a/embassy-stm32/src/can/bxcan.rs
+++ b/embassy-stm32/src/can/bxcan.rs
@@ -1,4 +1,3 @@
-use core::cell::{RefCell, RefMut};
 use core::future::poll_fn;
 use core::marker::PhantomData;
 use core::ops::{Deref, DerefMut};
@@ -84,7 +83,7 @@ impl<T: Instance> interrupt::typelevel::Handler<T::SCEInterrupt> for SceInterrup
 }
 
 pub struct Can<'d, T: Instance> {
-    pub can: RefCell<bxcan::Can<BxcanInstance<'d, T>>>,
+    pub can: bxcan::Can<BxcanInstance<'d, T>>,
 }
 
 #[derive(Debug)]
@@ -175,14 +174,12 @@ impl<'d, T: Instance> Can<'d, T> {
         tx.set_as_af(tx.af_num(), AFType::OutputPushPull);
 
         let can = bxcan::Can::builder(BxcanInstance(peri)).leave_disabled();
-        let can_ref_cell = RefCell::new(can);
-        Self { can: can_ref_cell }
+        Self { can }
     }
 
     pub fn set_bitrate(&mut self, bitrate: u32) {
         let bit_timing = Self::calc_bxcan_timings(T::frequency(), bitrate).unwrap();
         self.can
-            .borrow_mut()
             .modify_config()
             .set_bit_timing(bit_timing)
             .leave_disabled();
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/can/bxcan.rs
```diff
@@ -431,8 +429,13 @@ impl<'c, 'd, T: Instance> CanTx<'c, 'd, T> {
...
     }
 
+    /// Waits for a specific transmit mailbox to become empty
+    pub async fn flush(&self, mb: bxcan::Mailbox) {
+        Self::_flush(mb).await
+    }
+
     /// Waits until any of the transmit mailboxes become empty
-    pub async fn flush_any(&self) {
+    pub async fn _flush_any() {
```
Dirbaio, line 438: `_flush_any`, `_flush_all` should be private.

Also don't prepend with `_`, in Rust it denotes "supress warnings that it's unused", not "it's private" like in Python :)

Dirbaio:
LGTM other than the comment. Also, could you fix CI? Thanks!

embassy-stm32/src/can/bxcan.rs
mickvangelderen: Any suggestions as to what would be a preferred name? `flush_all_impl`?

Dirbaio: yeah! or `flush_all_inner`.

mickvangelderen:
> LGTM other than the comment. Also, could you fix CI? Thanks!

I'm trying to figure out how to interpret the logs from CI. 

The build fails with: `run failed: exited with code 123`. 

The tests fail to build for `embassy-net-adin1110` with: 
```
error[E0706]: functions in traits cannot be declared `async`
```
which can be solved with a rebase on main.

mickvangelderen:
@T-Bakker I want to help you get your change merged :). Hope you're okay with that.

mickvangelderen:
@T-Bakker @Dirbaio a `git rebase embassy/main` and `cargo fmt` fixed it! https://github.com/embassy-rs/embassy/pull/2207.

Dirbaio:
merged #2207

Issue #2167, opened 2023-11-09
stm32h7 adc hangs with possible fix: adc clock configured to pll that's not enabled


RobertTDowling:
On the stm32h743zi Nucleo board, I found this line of code hung, I believe in an infinite loop waiting for a ready pin to change:
`    let mut adc = Adc::new(p.ADC1, &mut delay);`

I believe the stm32h7 uses `adc/v4.rs`.

Looking at h7 reference manual, adc can use either of two clock sources

- `adc_ker_ck_input` (with CKMODE[1:0]==0) or 
- `adc_sclk`  (with CKMODE[1:0] != 0) 

A comment in `rcc/h.rs` at least hinted to me that CKMODE[1:0]==0 is being used:
Line 171 `            adc_clock_source: AdcClockSource::from_bits(0), // PLL2_P on H7, HCLK on H5`
Looking at PLL2, it appears to not be enabled.
```
impl Default for Config {
    fn default() -> Self {
        Self {
            hsi: Some(HSIPrescaler::DIV1),
            hse: None,
            csi: false,
            hsi48: Some(Default::default()),
            sys: Sysclk::HSI,
            pll1: None,
            pll2: None,
            #[cfg(any(rcc_h5, stm32h7))]
            pll3: None,
```

**So I changed the line to this**, presumably picking SYSCLK/2 with CKMODE[1:0]==2
Line 171 `            adc_clock_source: AdcClockSource::from_bits(2), // sysclk/2 on H7, HCLK on H5`

**And now the ADC works for me.** Or at least it does not hang during my call to `new` and I get reasonable values from the ADC.

I randomly picked 2 and it worked.  It could mean 64MHz/2=32MHz, which is less than 50MHz, from `adc/v4.rs`
```
#[cfg(stm32h7)]
const MAX_ADC_CLK_FREQ: Hertz = Hertz::mhz(50);
```

I have no idea at all if I did this right or not, and maybe setting up the PLL would be better.

Let me know how I can help make this better.  I could make a PR with the one line change if you like.

Robert

RobertTDowling:
Oops, I just realized that fix breaks the H5, because it probably wants 0 in those bits

Robert

RobertTDowling:
Fixed

RobertTDowling:
Fixed

PR #2192, opened 2023-11-16
stm32h7 ADC: Fix stalled clock in default h7 config

RobertTDowling:
Naive fix to clock source of ADC on STM32H7 (issue #2167)

In a nutshell, the default configuration for the v4 ADC selects clock source `ADCSEL[1:0]=0`. For the H5, that might be fine, but for the H7, that selects `PLL2_P` which is not configured. Using `ADCSEL[1:0]=2` selects the peripheral clock, which happens to be acceptable.

I hope to get advice on rust side of how to do this better.
- I just hard coded the value `2`. Note, the original hard coded `0` so I don't feel that bad.
- The mechanism to vary H7 from H5 (conditional compilation) is probably not done the way it should be
- I'm not sure the H7 and H5 are the only micros that use the v4 ADC.  If not, I've certainly done this wrong and should use something like `#[cfg(!stm32h7)]` if there is such a syntax
- I don't even know if ADCSEL[1:0]=0 works for the H5... it might need a similar fix.

I have tested on the H7, but I don't have an H5 yet so that part is untested. I can pick up an H5 before this is all over

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -170,9 +170,9 @@ impl Default for Config {
             per_clock_source: PerClockSource::HSI,
 
             #[cfg(stm32h5)]
-            adc_clock_source: AdcClockSource::from_bits(0), // HCLK on H5
+            adc_clock_source: AdcClockSource::HCLK1,
             #[cfg(stm32h7)]
-            adc_clock_source: AdcClockSource::from_bits(2), // PCLK on H7
+            adc_clock_source: AdcClockSource::PER,
 
             timer_prescaler: TimerPrescaler::DefaultX2,
             voltage_scale: VoltageScale::Scale0,
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -168,7 +168,12 @@ impl Default for Config {
             apb4_pre: APBPrescaler::DIV1,
 
             per_clock_source: PerClockSource::HSI,
-            adc_clock_source: AdcClockSource::from_bits(0), // PLL2_P on H7, HCLK on H5
+
+            #[cfg(stm32h5)]
+            adc_clock_source: AdcClockSource::from_bits(0), // HCLK on H5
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rcc/h.rs
```diff
@@ -168,7 +168,12 @@ impl Default for Config {
             apb4_pre: APBPrescaler::DIV1,
 
             per_clock_source: PerClockSource::HSI,
-            adc_clock_source: AdcClockSource::from_bits(0), // PLL2_P on H7, HCLK on H5
+
```
Dirbaio, line 171: could you use concrete enum names, like `AdcClockSource::HCLK`? the `from_bits` was to have one line of code that would work on both H5 and H7, but if there's `cfg`s there's no point.

RobertTDowling: **Sure!** I was hoping to test this on H5 to see if they could share the same (2) value, but...
...I can't compile ADC code on H5 to test this. I get this for both the stm32h503rb and the stm32h563zi: 
```
12 | use embassy_stm32::adc::Adc;
   |                    ^^^ could not find `adc` in `embassy_stm32`
```
I haven't given up completely, but I don't see what I'm doing wrong when the H7 compiles fine and I've adjusted Cargo.toml.
```
embassy-stm32 = { version = "*", path = "../embassy-stm32", features = ["nightly", "defmt", "stm32h503rb", "time-driver-any", "exti", "memory-x", "unstable-pac", "unstable-traits"] }

```
and .cargo/config.toml
```
[target.thumbv8m.main-none-eabihf]
runner = 'probe-rs run --chip STM32H503RBTx'

[build]
target = "thumbv8m.main-none-eabihf"
```

Dirbaio: ADC is not implemented for H5, it's normal it fails.

RobertTDowling: Done

Dirbaio (approved):
thank you!

PR #2197, opened 2023-11-18
Priority channel

MabezDev:
Adds a priority channel which will sift higher priority items to the front of the channel.

### Questions

- Is it okay to use heapless items in the public API? I'm currently using `Kind`, `Min` and `Max` from heapless publicly.
- Can the duplication between the normal channels be reduced further?
- More tests? I don't want to end up duplicating the tests within `heapless::BinaryHeap`, but if you have other test case ideas let me know I'll add them.

### DIFF
--- a/embassy-sync/src/channel.rs
+++ b/embassy-sync/src/channel.rs
@@ -321,7 +321,7 @@ impl<'ch, T> Future for DynamicSendFuture<'ch, T> {
 
 impl<'ch, T> Unpin for DynamicSendFuture<'ch, T> {}
 
-trait DynamicChannel<T> {
+pub(crate) trait DynamicChannel<T> {
     fn try_send_with_context(&self, message: T, cx: Option<&mut Context<'_>>) -> Result<(), TrySendError<T>>;
 
     fn try_receive_with_context(&self, cx: Option<&mut Context<'_>>) -> Result<T, TryReceiveError>;
--- a/embassy-sync/src/channel.rs
+++ b/embassy-sync/src/channel.rs
@@ -76,7 +76,7 @@ where
 
 /// Send-only access to a [`Channel`] without knowing channel size.
 pub struct DynamicSender<'ch, T> {
-    channel: &'ch dyn DynamicChannel<T>,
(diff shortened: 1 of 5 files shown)


embassy-sync/src/channel.rs
```diff
@@ -29,6 +29,8 @@ use crate::blocking_mutex::raw::RawMutex;
 use crate::blocking_mutex::Mutex;
 use crate::waitqueue::WakerRegistration;
 
+pub mod priority;
```
Dirbaio, line 32: could you put this in `embassy_sync::priority_channel`, to be consistent with `embassy_sync::zerocopy_channel`?

embassy-sync/src/channel/priority.rs
```diff
@@ -0,0 +1,614 @@
...
+//!
+//! Similar to a [`Channel`](crate::channel::Channel), however [`PriorityChannel`] sifts higher priority items to the front of the queue.
+//! Priority is determined by the `Ord` trait. Priority behavior is determined by the [`Kind`](heapless::binary_heap::Kind) parameter of the channel.
+
+use core::cell::RefCell;
+use core::future::Future;
+use core::pin::Pin;
+use core::task::{Context, Poll};
+
+use heapless::binary_heap::Kind;
```
Dirbaio, line 11: could you `pub use` everything needed so the user can use this without imprting `heapless` directly? I think it's `Kind`, `Min`, `Max`.

Dirbaio:
> Is it okay to use heapless items in the public API? I'm currently using Kind, Min and Max from heapless publicly.

embassy-sync is not a trait crate like embedded-hal/nal so IMO it's fine. We can major-bump when heapless major-bumps.

>Can the duplication between the normal channels be reduced further?

I wouldn't try :joy: I don't think there's a way to deduplicate the public API and docs without macro crimes.

> More tests? I don't want to end up duplicating the tests within heapless::BinaryHeap, but if you have other test case ideas let me know I'll add them.

LGTM, it's fine to assume heapless is not broken yeah.

Dirbaio (approved):
thank you!

PR #2198, opened 2023-11-18
STM32: Don't enable opamps in new(), wait until configured

adamgreig:
Previously we enabled the opamp as soon as `new()` is called, but for all supported use cases the user has to then call `buffer_int` or `buffer_ext` (which also set the relevant enable bit).

However, that means if someone is going to call `buffer_int`, the opamp is briefly enabled in external output mode, where its output drives a pin, until it's configured for internal mode. This glitch could potentially cause issues if that pin is being used for something else.

You can't do anything with the opamp object even if it's enabled, except manually use the PAC to configure it in ways we don't support - in which case you can also manually use the PAC to enable it.

### DIFF
--- a/embassy-stm32/src/opamp.rs
+++ b/embassy-stm32/src/opamp.rs
@@ -31,12 +31,9 @@ impl From<OpAmpSpeed> for crate::pac::opamp::vals::OpampCsrOpahsm {
 
 /// OpAmp external outputs, wired to a GPIO pad.
 ///
-/// The GPIO output pad is held by this struct to ensure it cannot be used elsewhere.
-///
 /// This struct can also be used as an ADC input.
-pub struct OpAmpOutput<'d, 'p, T: Instance, P: OutputPin<T>> {
+pub struct OpAmpOutput<'d, T: Instance> {
     _inner: &'d OpAmp<'d, T>,
-    _output: &'p mut P,
 }
 
 /// OpAmp internal outputs, wired directly to ADC inputs.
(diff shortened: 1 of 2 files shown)


Dirbaio:
this seems a bit counterintuitive. I think we should do multiple constructors instead: `new_buffer_internal`, `new_buffer_external`. 

Also, what does calling `new()` and nothing else do? from your comment it seems it'll start driving the output pin, *without taking ownership of it*, which shouldn't be possible.

adamgreig:
> this seems a bit counterintuitive. I think we should do multiple constructors instead: new_buffer_internal, new_buffer_external.

I like this, but we'd need to change the structure of the driver. Currently there's `OpAmp`, `OpAmpOutput`, `OpAmpInternalOutput`; the `OpAmp` holds the peripheral, the `OpAmpOutput` holds a reference to the `OpAmp` and the output pin, and serves as an ADC channel for reading from the opamp. We could change so the `OpAmpOutput` holds the peripheral, deleting the `OpAmp` entirely, if that sounds OK?

> Also, what does calling new() and nothing else do? from your comment it seems it'll start driving the output pin, without taking ownership of it, which shouldn't be possible.

Yes, when the opamp is enabled (as currently happens in `new()`) it begins driving its output, which by default is connected to a physical pin. The other methods just configure the _inputs_ and gain, as well as consuming the output pin.

adamgreig:
@Dirbaio I've updated this after our discussion yesterday:

* Disable-on-drop moved to `OpAmpOutput`/`OpAmpInternalOutput` instead of `OpAmp`
* Changed `buffer_int`/`buffer_ext` to take `impl Peripheral` instead of pins directly, and also updated `OpAmpOutput` to not hold the pin but instead bind the pin lifetime to the same 'd as the borrow on the OpAmp (this removed a lot of generics!)

Dirbaio (approved):
:+1:  thank you!

PR #2199, opened 2023-11-19
STM32 DAC: update to new PAC, combine ch1/ch2, fix trigger selection

adamgreig:
To be considered alongside https://github.com/embassy-rs/stm32-data/pull/306

The current trigger selection is wrong for almost all STM32s and needlessly splits into TSEL1/2 even though no chips make a distinction based on channel (!). However, there is significant difference in the TSEL enum between chips, so it's very burdensome to define it in the PAC. The PAC PR above reworks DAC support to improve accuracy and compatibility between all STM32s, and changes TSEL to always take an integer instead of an enum.

This PR adds an enum of trigger sources for every STM32, accurate except for where triggers differ between DAC1/2 and where triggers can be remapped in SYSCFG (indicated with doc comments), and updates the current driver to use it.

There's a few other changes:

* The existing `mode: u8` is changed to a `Mode` enum, since it's now common across supported DAC versions
* `build.rs` was cleaned up and some new configs added for `flashsize_X`, `package_X`, `stm32l4_nonplus`, `stm32l4_plus`, needed for the new trigger selection enums

Still to-do:
* [x] Test on hardware
* [x] Look for opportunities to refactor the two channel structs into one now that tsel is common
* [x] The driver needs to enable the now-available HFSEL when required (peripheral clock above some threshold on some chips)

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -996,8 +996,8 @@ fn main() {
         // SDMMCv1 uses the same channel for both directions, so just implement for RX
         (("sdmmc", "RX"), quote!(crate::sdmmc::SdmmcDma)),
         (("quadspi", "QUADSPI"), quote!(crate::qspi::QuadDma)),
-        (("dac", "CH1"), quote!(crate::dac::DmaCh1)),
-        (("dac", "CH2"), quote!(crate::dac::DmaCh2)),
+        (("dac", "CH1"), quote!(crate::dac::DacDma1)),
+        (("dac", "CH2"), quote!(crate::dac::DacDma2)),
     ]
     .into();
 
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1376,6 +1376,15 @@ fn main() {
         println!("cargo:rustc-cfg=flashsize_{}", &chip_name[10..11]);
     }
 
+    // Mark the L4+ chips as they have many differences to regular L4.
+    if &chip_name[..7] == "stm32l4" {
+        if "pqrs".contains(&chip_name[7..8]) {
+            println!("cargo:rustc-cfg=stm32l4_plus");
+        } else {
+            println!("cargo:rustc-cfg=stm32l4_nonplus");
+        }
+    }
+
     println!("cargo:rerun-if-changed=build.rs");
 }
 
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -65,7 +65,6 @@ fn main() {
             match r.kind {
                 // Generate singletons per pin, not per port
                 "gpio" => {
-                    println!("{}", p.name);
                     let port_letter = p.name.strip_prefix("GPIO").unwrap();
                     for pin_num in 0..16 {
                         singletons.push(format!("P{}{}", port_letter, pin_num));
@@ -1352,15 +1351,6 @@ fn main() {
 
     if let Some(core) = core_name {
         println!("cargo:rustc-cfg={}_{}", &chip_name[..chip_name.len() - 2], core);
-    } else {
-        println!("cargo:rustc-cfg={}", &chip_name[..chip_name.len() - 2]);
-    }
-
-    // ========
-    // stm32f3 wildcard features used in RCC
-
-    if chip_name.starts_with("stm32f3") {
-        println!("cargo:rustc-cfg={}x{}", &chip_name[..9], &chip_name[10..11]);
     }
 
     // =======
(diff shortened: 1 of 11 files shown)


adamgreig:
It looks like the CI is still using the old PAC (references to `set_tsel1()` which doesn't exist in https://github.com/embassy-rs/stm32-data/pull/306), despite changing `embassy-stm32`'s `Cargo.toml`. Did I miss something?

Dirbaio:
change it in `dependencies` and `build-dependencies`. `patch` only applies if you build `embassy-stm32` as the root crate, which breaks when building the examples because it's a dep there, not a root crate.

embassy-stm32/build.rs
```diff
@@ -1354,16 +1344,25 @@ fn main() {
     if &chip_name[..8] == "stm32wba" {
         println!("cargo:rustc-cfg={}", &chip_name[..8]); // stm32wba
         println!("cargo:rustc-cfg={}", &chip_name[..10]); // stm32wba52
+        println!("cargo:rustc-cfg=package_{}", &chip_name[10..11]);
+        println!("cargo:rustc-cfg=flash_{}", &chip_name[11..12]);
```
Dirbaio, line 1348: this could be confused with the cfgs for the FLASH peripheral (like `flash_f4`), could you pick another name? maybe `mem` or `memory`? (I think that letter also affects RAM size?)

adamgreig: could `size_` work? ST call it "flash memory size" typically. It sometimes influences SRAM too but not usually, so I think saying `mem` sounds a bit too much like it's talking about RAM when it's not. Could do `density` which is how ST used to refer to this on older devices.

Dirbaio: `size` sounds good. `density` please no :rofl:

adamgreig: I've gone for `flashsize` for now, plain `size` seemed a big ambiguous.

adamgreig:
The latest (WIP) commit is a fairly major rework of the DAC driver: there's now just one `DacChannel` type, instead of separate channel 1/2 structs, which is possible now there's just one TSEL and it reduces code duplication.

* I've added critical sections where the previous driver didn't have them; I think they're necessary since the control registers are shared between channels 1 and 2
* I've retained enabling the channel on creation. Ideally we wouldn't, so it's possible for users to configure a channel without causing it to drive its output (because DACs can be set to internal-output-only), but right now we require the output pin in `new()`, so it's OK. I think ideally I would change this so the pin is only required when enabling an external output mode, but that will require rethinking the mode enum a bit too.
* I've disabled trigger mode by default, because as it stands you don't know what the default trigger source is (it's different between DAC versions). We could default to software trigger, but that seems pointless for the default case because you might as well just have triggering off and update the DAC as soon as you call `set()`. Using software triggering is a fairly niche case, I think, and for the rest of the triggering use cases I think it makes sense to select a trigger source before enabling triggering.

I'm not sure about the `T::enable_and_reset()` reference counting; is it correct to call it twice, since each channel will then call disable on drop? The old driver didn't call it twice or ever call disable. It also seems like the old driver has to work around the "DAC12EN" bit in H7, but its implementation of `enable_and_reset_with_cs()` and `disable_with_cs()` doesn't touch the reference count.

Finally, the DMA trait business seems a bit unsatisfying as-is. If I just have one `DmaChannel` trait, I can put the `fn write()` inside the general impl block with a bound on `DMA: DmaChannel`, and everything works for most devices, but for some reason on H7 (and maybe others), implementing the DMA trait for both DAC CH1 and CH2 causes a conflicting implementations error. Instead, I kept a DMA trait per channel, then duplicate the impl for either concrete N=1 or N=2 with different DMA trait bounds each time. I think this works, and it's better than the total duplication we had before, but I wonder if there's a nicer way to do it.

Still to-do:
* [ ] Write HFSEL as required
* [ ] Test on hardware
* [ ] Document `Dac` methods, add methods for using both DACs together (simultaneous data updates, DMA into shared data register, etc)

adamgreig:
This is tested on F4 and H7 successfully, which covers the two main variants of the DAC peripheral at least.

* I've implemented all the features I wanted to. Someone could add DMA into the dual registers but perhaps it can wait until someone needs it.
* I added type aliases for `DacCh1`/`DacCh2` which match the old type names, which makes updating code a bit easier.
* I'm not sure about the H7 RccPeripheral refcounting business that I mentioned above.
* I added `new_internal()` to support creating without an output pin, which then sets the DAC to a mode where only the internal peripherals are driven. However, a user could manually change the mode back to one where the output pin _is_ driven. Since we allow safe writing to the PAC registers anyway I'm not sure there's really much to be done about this; I think only a "sophisticated" user is likely to use it, and for them it's useful as otherwise there's no way to have ch1 be external and ch2 be internal, for example. Still, it is a potential footgun. As it stands we don't support setting the sample-and-hold times or other configuration needed for it, so they'd be using the PAC directly anyway.

I haven't changed the `Value` enum; it's a little awkward but probably still better than having separate methods for each type of data alignment.

Dirbaio:
bender run

Dirbaio:
bender run

PR #2208, opened 2023-11-21
Add FAQ with one question from chat

jamesmunns:
Just provoking @Dirbaio to make his life easier in the future :)


Dirbaio (approved):
:heart:

PR #2211, opened 2023-11-21
ci: cache lockfiles

Dirbaio:
- Removes the slow "updating crates.io index, updating git" at start of the job.
- Avoids CI being randomly slower if a widely-used dep (like serde) gets bumped causing rebuilding everything.


Dirbaio:
bender run

PR #2213, opened 2023-11-22
introduce `OptionalPeripherals`

kaspar030:
I needed a way to spread peripheral initialization over multiple `init()` functions.
Passing `&mut Peripherals` around won't work as it's field cannot be moved out.
But `steal()`ing `Peripherals` in multiple places doesn't seem right.

So this PR augments `peripherals_struct!()` to also create a `struct OptionalPeripherals {}` that contains all fields of `Peripherals`, but wrapped in `Option`.

This is quite wasteful (1b per Peripheral, e.g., ~119b on nrf52840). I hope that this will be on stack only, or *edit* we'll optimize this into some kind of bit-field but with a similar interface, or even just compile it out.

*edit* ~~this only exposes `OptionalPeripherals` for nrf52840 now. If I get positive feedback on the idea, I'll add re-exporting for the other crates. Marking as draft for now.~~ now supports nrf, stm32 and rp. We have a similar PR to esp-rs.

### DIFF
--- a/embassy-hal-internal/src/macros.rs
+++ b/embassy-hal-internal/src/macros.rs
@@ -46,6 +46,16 @@ macro_rules! peripherals_struct {
             )*
         }
 
+        /// Struct containing all the peripheral singletons wrapped in `Option`.
+        #[allow(non_snake_case)]
+        pub struct OptionalPeripherals {
+            $(
+                #[doc = concat!(stringify!($name), " peripheral")]
+                $(#[$cfg])?
+                pub $name: Option<peripherals::$name>,
+            )*
+        }
+
         impl Peripherals {
             ///Returns all the peripherals *once*
             #[inline]
@@ -82,6 +92,19 @@ macro_rules! peripherals_struct {
                     $(
                         $(#[$cfg])?
                         $name: peripherals::$name::steal(),
+                    )*
+                }
+            }
+        }
+
+        impl OptionalPeripherals {
+            /// Create an `OptionalPeripherals`, consuming a `Peripherals`
+            #[inline]
+            pub fn from(p: Peripherals) -> Self {
+                Self {
+                    $(
+                        $(#[$cfg])?
+                        $name: Some(p.$name),
                     )*
                 }
             }
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -172,7 +172,7 @@ macro_rules! bind_interrupts {
 pub use chip::pac;
 #[cfg(not(feature = "unstable-pac"))]
 pub(crate) use chip::pac;
-pub use chip::{peripherals, Peripherals, EASY_DMA_SIZE};
+pub use chip::{peripherals, OptionalPeripherals, Peripherals, EASY_DMA_SIZE};
 pub use embassy_hal_internal::{into_ref, Peripheral, PeripheralRef};
 
 pub use crate::chip::interrupt;
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -153,7 +153,7 @@ macro_rules! bind_interrupts {
 }
 
 // Reexports
-pub use _generated::{peripherals, Peripherals};
+pub use _generated::{peripherals, OptionalPeripherals, Peripherals};
 pub use embassy_hal_internal::{into_ref, Peripheral, PeripheralRef};
 #[cfg(feature = "unstable-pac")]
 pub use stm32_metapac as pac;


AdinAck:
Could you show an example of why this is useful?

kaspar030:
> Could you show an example of why this is useful?

Sure!

Basically I'm trying to provide high-level features ("net", "usb", "usb_ethernet", ...) that an application can select, together with target board features (e.g., "nrf52840dk", "rpi-pico"), and then minimize code duplication, having shared initialization functions.

```rust
  #[cfg(feature = "nrf52840dk")]                                                                                                                               
  mod nrf52840dk {                                                                                                                                             
      pub type UsbDriver = Driver<'static, peripherals::USBD, HardwareVbusDetect>;                                                                             
      pub fn init_usb(p: Peripherals) -> UsbDriver {                                                                                                           
          use embassy_nrf::peripherals;                                                                                                                        
          use embassy_nrf::usb::{vbus_detect::HardwareVbusDetect, Driver};                                                                                     
          pub fn usb_driver(p: &mut Peripherals) -> UsbDriver {                                                                                                
              let usbd = p.USBD; // cannot move out                                                                                                            
              use super::Irqs;                                                                                                                                 
              Driver::new(usbd, Irqs, HardwareVbusDetect::new(Irqs))                                                                                           
          }                                                                                                                                                    
      }                                                                                                                                                        
  }                                                                                                                                                            
                                                                                                                                                               
  #[cfg(feature = "rpi-pico")]                                                                                                                                 
  pub mod rpi_pico {                                                                                                                                           
      use embassy_rp::peripherals;                                                                                                                             
      use embassy_rp::usb::Driver;                                                                                                                             
      pub type UsbDriver = Driver<'static, peripherals::USB>;                                                                                                  
      pub fn usb_driver(p: &mut Peripherals) -> UsbDriver {                                                                                                    
          let usb = p.USB; // cannot move out                                                                                                                  
          Driver::new(usb, super::Irqs)                                                                                                                        
      }                                                                                                                                                        
  }                                                                                                                                                            
                                                                                                                                                               
  #[cfg(feature = "nrf52840dk")]                                                                                                                               
  use nrf52840dk as board;                                                                                                                                     
                                                                                                                                                               
  #[cfg(feature = "rpi-pico")]                                                                                                                                 
  use rpi_pico as board;                                                                                                                                       
                                                                                                                                                               
  fn init(p: Peripherals) {                                                                                                                                    
      let usb_driver = board::usb_driver(&mut p);                                                                                                              
  }
```

(pseudo code and mostly conceived example)

In this example, I'd like to move the initialization of `usb_driver` out of `init()` into functions (of modules, or elsewhere). With `USBD` (or `USB`) being fields of `Peripheral`, there's no way. If `usb_driver()` takes `&mut Peripheral`, the field cannot move out. If it consumes the peripheral, that cannot be used for any other initialization.

For this example, there's ways to work around this (e.g., by passing `p.USBD` to `board::usb_driver()`), but already there'd be two versions because of different naming of the USB peripheral field.

Down the line there'll be more variants, more peripheral types, more places to pull `init_foo(p: Peripherals)` from, ...

adamgreig:
I recently published [`assign-resources`](https://crates.io/crates/assign-resources) which is another way to solve this problem: you can split `Peripherals` into multiple structs, e.g. you'd have one for `fn usb_driver()`, and you only write the name of the type in one place, so you can very easily `#[cfg]` multiple versions for different boards. In `main()`, you split up peripherals into the smaller structs, and pass each one to the relevant function, which then gets to own its peripherals.

For example:

```rust
#[cfg(feature = "nrf52840dk")]                                                                                                                               
assign_resources! {
    usb_driver: UsbResources {
        usb: USBD,
    }
}

#[cfg(feature = "rpi-pico")]                                                                                                                                 
assign_resources! {
    usb_driver: UsbResources {
        usb: USB,
    }
}

pub fn make_usb_driver(r: UsbResources) -> UsbDriver {
    // it's always called `r.usb` in `UsbResources` and is an owned peripheral, not a reference
    Driver::new(r.usb, Irqs, #[cfg(feature = "nrf52840dk")] HardwareVbusDetect::new(Irqs))
}

fn init(p: Peripherals) {
    let r = split_resources!(p);
    // pass `make_usb_driver()` all the USB related resources (pins, peripherals, DMA channels, etc)
    let usb_driver = make_usb_driver(r.usb_driver);
}
```

kaspar030:
> I recently published [assign-resources](https://crates.io/crates/assign-resources) which is another way to solve this problem: you can split Peripherals into multiple structs

kaspar030:
Just so this doesn't get cleaned: we might still want to use this, and are still searching for alternatives or better justification. :)

kaspar030:
Some update:

[We](https://github.com/future-proof-iot/RIOT-rs)'ve been using `OptionalPeripherals` in our fork of embassy for a while now. We're pairing it with a variant of `assign_resources!()` (we call it [`riot_rs::define_peripherals!()`](https://github.com/future-proof-iot/RIOT-rs/blob/main/src/riot-rs-embassy/src/define_peripherals.rs)).

In combination, it allows writing applications that don't fully control `Peripherals`, which is kinda the thing that RIOT-rs (turning into an opinionated distribution of embassy) adds on top of embassy.

I'm realizing that we don't really have a succinct example showcasing this. Probably our [hid test application](https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/embassy-usb-keyboard) comes closest.

The general idea is that an (application or library) crate can do sth like this:

```Rust

use riot_rs::...;

riot_rs::define_peripherals!(Buttons {
    btn1: P0_11,
    btn2: P0_12,
    btn3: P0_24,
    btn4: P0_25,
});

#[riot_rs::task(autostart, peripherals)]
async fn button_user(buttons: Buttons) {
  // this task now owns the peripherals in `Buttons`
...
}
```

This should look quite familiar to a "regular" embassy application. Here, the riot-rs crates drive MCU setup & startup (controlling `main()` and `Peripherals`), take care (using `OptionalPeripherals`) of extracting `Buttons` in a safe (works or panics) way, and of spawning the task.
Potentially there are other tasks which use other `Peripherals`, started elsewhere in the dependency tree, with ownership tracked at runtime using `OptionalPeripherals`. Like, enabling a feature on the `riot-rs` crate also enables the network stack.

We'd love to upstream this so we might eventually not use a fork of embassy. And we don't see a way to implement this in our code, the `peripherals_struct!()` is just the right place. :smile:

Personally I think the name needs bike shedding. Apart from that, I hope embassy finds this useful enough to merge, or unintrusive enough to merge because we want to use it.

Dirbaio:
I'm sorry, but I don't think this is something we should merge. 

- Solutions like `assign_resources!` already offer ways of splitting initialization between multiple methods, without potential "unwrap of None" panics.
- I'm not a fan of having two ways of doing the same thing, with subtle tradeoffs between them.

so I'm going to close this. Thanks anyway for the PR!

kaspar030:
Hm, this is very unfortunate, we depend on this quit a bit in [Ariel OS](https://github.com/ariel-os/ariel-os), and we'd love to not have to use a fork at some point.

What the runtime tracking allows us to do, and what no other solution that we know of does, is that it solves what we've dubbed the "partial move problem" in code that is independent at the type level.

`assign_resources!()` allows splitting initialization between multiple methods **from one method that knows all the types**, as the compiler can reason about the partial moves. Only that one method can move fields out of `Peripherals`.

`OptionalPeripherals` **allows runtime tracking of the actually moved out peripherals**, which allows us to pass such a struct to arbitrary methods without knowing the exact peripherals those methods would take out.

I think `assign_resources!()` actually solves an orthogonal problem (bundling some Peripherals fields and partially moving them out of `Peripherals`, whereas we need to fully move/take them out, elsewhere).

(In Ariel, we're trying to mitigate the "unwrap of None" panics by having some defined "initialization phase", but we're not too happy about that either. )

We'd love to handle this in our code base, but don't see how. Suggestions welcome. :)

@DirBaio, is this sufficiently different to make you reconsider?

Or, can we maybe ease taking this change in somehow?

Some options I see:

0. rename the struct (we're not happy with it, we're converging to `TrackedPeripherals`)

1. use some macro to create the wrapped struct (I found https://github.com/roignpar/optfield). hide its use (and the necessary reexports) behind a feature.
   This would add the optional dependency, but reduce code impact to a handful of feature-gated lines

3. Find (must exist, else build) a macro that exports the structure of Peripherals (create a macro that creates a macro that creates `Peripherals` which embassy exports, which we can use in our code to wrap it in `optfield`)

PR #2215, opened 2023-11-24
stm32 i2c: allow EHA traits without time feature

andresv:
`embedded-hal-async` works without needing `time`.


Dirbaio (approved):
thanks!

PR #2216, opened 2023-11-24
executor: add support for main/task macros in stable Rust

Dirbaio:
- embassy-embedded-hal: don't use feature(try_blocks).
- executor: Remove non-functional rtos-trace-interrupt.
- executor: add some tests.
- executor: add support for main/task macros in stable (allocates tasks in an arena)


embassy-executor/src/arch/xtensa.rs
```diff
@@ -8,7 +8,6 @@ mod thread {
     use core::marker::PhantomData;
     use core::sync::atomic::{AtomicBool, Ordering};
 
-    #[cfg(feature = "nightly")]
     pub use embassy_macros::main_riscv as main;
```
bugadani, line 11: might wanna _not_ reexport the riscv macro?

Dirbaio: it was already broken before, will fix in a separate PR.

PR #2224, opened 2023-11-27
Add support for HSE Oscillator in stm32g0

cbruiz:
* Set "NSS pulse off" in stm32g0 for Software Slave Mode compliancy.
* stm32g0b1 usb serial sample with spi (sdcard)


Dirbaio:
This PR has 3 things in one. Could you do *just* the "add HSE oscillator support" changes?

- Why is `w.set_nssp(false);` required? It only has an effect when using hardware NSS mode, while our driver always uses software NSS mode.
- The example is a bit niche IMO, we already have examples hsowing how to use USB and SPI. I don't think adding an example showing how to use SPI and USB at the same time is worth it.

PR #2228, opened 2023-11-27
embassy-embedded-hal: shared_bus/blocking/spi: fix build and behaviour when "time" feature disabled

### LINKED ISSUE #2227: embassy-embedded-hal fails to build with "time" feature disabled
Compile error occurs twice in this file: https://github.com/embassy-rs/embassy/blob/3a8a950bb8e8c8054aeb1338d17a3fa2e14c1b58/embassy-embedded-hal/src/shared_bus/blocking/spi.rs

```
62 |               let op_res = operations.iter_mut().try_for_each(|op| match op {
   |  __________________________________________________________________-
63 | |                 Operation::Read(buf) => bus.read(buf),
   | |                                         ------------- this is found to be of type `Result<(), <BUS as embedded_hal::spi::ErrorType>::Error>`
64 | |                 Operation::Write(buf) => bus.write(buf),
   | |                                          -------------- this is found to be of type `Result<(), <BUS as embedded_hal::spi::ErrorType>::Error>`
65 | |                 Operation::Transfer(read, write) => bus.transfer(read, write),
   | |                                                     ------------------------- this is found to be of type `Result<(), <BUS as embedded_hal::spi::ErrorType>::Error>`
66 | |                 Operation::TransferInPlace(buf) => bus.transfer_in_place(buf),
   | |                                                    -------------------------- this is found to be of type `Result<(), <BUS as embedded_hal::spi::ErrorType>::Error>`
67 | |                 #[cfg(not(feature = "time"))]
68 | |                 Operation::DelayUs(_) => Err(SpiDeviceError::DelayUsNotSupporte...
   | |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<(), ...>`, found `Result<_, ...>`
...  |
73 | |                 }
74 | |             });
   | |_____________- `match` arms have incompatible types
   |
   = note: expected enum `Result<(), <BUS as embedded_hal::spi::ErrorType>::Error>`
              found enum `Result<_, SpiDeviceError<_, _>>`
   = help: consider constraining the associated type `<BUS as embedded_hal::spi::ErrorType>::Error` to `SpiDeviceError<_, _>`
   = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
```

The error is one thing, but the intended behaviour here is not ideal. If the crate is built without support for `DelayUs`, I do not think we should attempt the sequence of operations before shortcircuiting when a `DelayUs` is reached. It would be cleaner to no-op the whole transaction. For instance, this could be done at the start of `transaction()`:

```rust
        if (
            cfg!(not(feature = "time")) &&
            operations.iter().find(|op| matches!(op, Operation::DelayUs(_))).is_some()
        ) {
            return Err(SpiDeviceError::DelayUsNotSupported);
        }
```

and then in the `match`, we kill off the error thus:

```rust
                #[cfg(not(feature = "time"))]
                Operation::DelayUs(_) => unreachable!(),
```

tcbennun:
New behaviour: check for `DelayUs` presence in operations instead of shortcircuiting.

Closes #2227

### DIFF
--- a/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
+++ b/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
@@ -55,6 +55,10 @@ where
     CS: OutputPin,
 {
     fn transaction(&mut self, operations: &mut [Operation<'_, u8>]) -> Result<(), Self::Error> {
+        if cfg!(not(feature = "time")) && operations.iter().any(|op| matches!(op, Operation::DelayUs(_))) {
+            return Err(SpiDeviceError::DelayUsNotSupported);
+        }
+
         self.bus.lock(|bus| {
             let mut bus = bus.borrow_mut();
             self.cs.set_low().map_err(SpiDeviceError::Cs)?;
@@ -65,7 +69,7 @@ where
                 Operation::Transfer(read, write) => bus.transfer(read, write),
                 Operation::TransferInPlace(buf) => bus.transfer_in_place(buf),
                 #[cfg(not(feature = "time"))]
-                Operation::DelayUs(_) => Err(SpiDeviceError::DelayUsNotSupported),
+                Operation::DelayUs(_) => unreachable!(),
                 #[cfg(feature = "time")]
                 Operation::DelayUs(us) => {
                     embassy_time::block_for(embassy_time::Duration::from_micros(*us as _));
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thank you!

PR #2236, opened 2023-12-01
no_std and no_main can be used on stable

tshepang:


lulf (approved):
Thanks for the improvement!

PR #2237, opened 2023-12-01
stm32/rcc: make h7 rm0399 power supply configurable

badrbouslikhin:
This pull request introduces power supply configuration support for the STM32H745/755 and STM32H747/757.

It includes adding two additional enums `SupplyConfig` and `SMPSSupplyVoltage`, allowing to specify different power supply configurations and SMPS output voltage levels. 
The modifications are in accordance with the RM0433 Rev 4 7.4 specifications.

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -117,6 +117,57 @@ impl From<TimerPrescaler> for Timpre {
     }
 }
 
+/// Power supply configuration
+/// See RM0433 Rev 4 7.4
+#[cfg(pwr_h7rm0399)]
+pub enum SupplyConfig {
+    /// Default power supply configuration.
+    /// V CORE Power Domains are supplied from the LDO according to VOS.
+    /// SMPS step-down converter enabled at 1.2V, may be used to supply the LDO.
+    Default,
+
+    /// Power supply configuration using the LDO.
+    /// V CORE Power Domains are supplied from the LDO according to VOS.
+    /// LDO power mode (Main, LP, Off) will follow system low-power modes.
+    /// SMPS step-down converter disabled.
+    LDO,
+
+    /// Power supply configuration directly from the SMPS step-down converter.
+    /// V CORE Power Domains are supplied from SMPS step-down converter according to VOS.
+    /// LDO bypassed.
+    /// SMPS step-down converter power mode (MR, LP, Off) will follow system low-power modes.
+    DirectSMPS,
+
+    /// Power supply configuration from the SMPS step-down converter, that supplies the LDO.
+    /// V CORE Power Domains are supplied from the LDO according to VOS
+    /// LDO power mode (Main, LP, Off) will follow system low-power modes.
+    /// SMPS step-down converter enabled according to SDLEVEL, and supplies the LDO.
+    /// SMPS step-down converter power mode (MR, LP, Off) will follow system low-power modes.
+    SMPSLDO,
+
+    /// Power supply configuration from SMPS supplying external circuits and potentially the LDO.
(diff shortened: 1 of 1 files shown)


Dirbaio:
bender run

embassy-stm32/src/rcc/h.rs
```diff
@@ -194,14 +253,101 @@ pub(crate) unsafe fn init(config: Config) {
...
+                    w.set_sden(true);
+                    w.set_ldoen(false);
+                    w.set_bypass(false);
+                });
+            }
+            SupplyConfig::SMPSLDO => {
+                PWR.cr3().modify(|w| {
+                    match config.smps_supply_voltage {
+                        Some(SMPSSupplyVoltage::V1_8) => {
+                            PWR.cr3().modify(|w| w.set_sdlevel(0b01));
```
Dirbaio, line 295: don't do a modify inside a modify, the effect will be reversed when the outer modify does the write.

embassy-stm32/src/rcc/h.rs
```diff
@@ -194,14 +253,101 @@ pub(crate) unsafe fn init(config: Config) {
         w.set_bypass(false);
     });
 
-    #[cfg(any(pwr_h7rm0399, pwr_h7rm0455, pwr_h7rm0468))]
+    #[cfg(any(pwr_h7rm0455, pwr_h7rm0468))]
```
Dirbaio, line 256: SMPS is also available for those RMs, could you enable it too?

It is identical in all 3 RMs, as a general rule if a feature is identical it should be enabled in all, to avoid accumulating maintenance debt. 

If you only have rm0399 hardware it's fine, CI test has others.

embassy-stm32/src/rcc/h.rs
```diff
@@ -194,14 +253,101 @@ pub(crate) unsafe fn init(config: Config) {
...
+                PWR.cr3().modify(|w| {
+                    match config.smps_supply_voltage {
+                        Some(SMPSSupplyVoltage::V1_8) => {
+                            PWR.cr3().modify(|w| w.set_sdlevel(0b01));
+                        }
+                        Some(SMPSSupplyVoltage::V2_5) => {
+                            PWR.cr3().modify(|w| w.set_sdlevel(0b10));
+                        }
+                        None => {
+                            panic!("Supply configuration SMPSExternalLDOBypass requires a supply voltage to be set.");
```
Dirbaio, line 339: if you make the voltage a field in the `SupplyConfig` variants that need it, it becomes impossible to not supply it:

```rust
pub enum SupplyConfig {
    Default,
    LDO,
    DirectSMPS,
    SMPSLDO(SMPSSupplyVoltage),
    SMPSExternalLDO(SMPSSupplyVoltage),
    SMPSExternalLDOBypass(SMPSSupplyVoltage),
}
```

embassy-stm32/src/rcc/h.rs
badrbouslikhin: Ok. I double-checked, and all 3 RMs are identical indeed.

Currently `Direct SMPS` is hardcoded. Apparently, that's what works on Nucleos, how do you want to handle that? 
Should we make that the default on config init?

badrbouslikhin:
Hey @Dirbaio, thanks for your review! I've taken care of all your comments and made the necessary changes. 
There's just one thing we're still unsure about - the default configuration.

embassy-stm32/src/rcc/h.rs
Dirbaio: Yes, why not. default to it on the chips that do have SMPS.

Dirbaio (approved):
thank you!

PR #2241, opened 2023-12-03
STM32: Add UID driver

adamgreig:
Based on #1253 but updated to use the now-available UID peripheral from stm32-metapac.

I'm not 100% sure if it's nicer to return the `&'static` references or just return directly a `[u8; 12]` and `[u8; 24]`; we can avoid the critical section and all the unsafe code if we just return arrays, but it wastes a bit of stack. For the `&str`, I don't see any other way, and I guess a major use case will be passing this to the USB serial number, so perhaps it's worth keeping as-is.

I added the `LOADED` check to prevent needing a critical section and computation loop on every call of `uid_hex()`/`uid_hex_bytes()`, but probably the most common case is to only call it once, in which case the memory and load/store is wasted, so I could remove it.

Tested only on stm32g4 hardware.

### DIFF
--- a/embassy-stm32/src/uid.rs
+++ b/embassy-stm32/src/uid.rs
@@ -1,0 +1,29 @@
+/// Get this device's unique 96-bit ID.
+pub fn uid() -> &'static [u8; 12] {
+    unsafe { &*crate::pac::UID.uid(0).as_ptr().cast::<[u8; 12]>() }
+}
+
+/// Get this device's unique 96-bit ID, encoded into a string of 24 hexadecimal ASCII digits.
+pub fn uid_hex() -> &'static str {
+    unsafe { core::str::from_utf8_unchecked(uid_hex_bytes()) }
+}
+
+/// Get this device's unique 96-bit ID, encoded into 24 hexadecimal ASCII bytes.
+pub fn uid_hex_bytes() -> &'static [u8; 24] {
+    const HEX: &[u8; 16] = b"0123456789ABCDEF";
+    static mut UID_HEX: [u8; 24] = [0; 24];
+    static mut LOADED: bool = false;
+    critical_section::with(|_| unsafe {
+        if !LOADED {
+            let uid = uid();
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/uid.rs
```diff
@@ -0,0 +1,33 @@
...
+pub fn uid_hex() -> &'static str {
+    unsafe { core::str::from_utf8_unchecked(uid_hex_bytes()) }
+}
+
+/// Get this device's unique 96-bit ID, encoded into 24 hexadecimal ASCII bytes.
+pub fn uid_hex_bytes() -> &'static [u8; 24] {
+    const HEX: &[u8; 16] = b"0123456789ABCDEF";
+    static mut UID_HEX: [u8; 24] = [0; 24];
+    static LOADED: AtomicBool = AtomicBool::new(false);
+    if !LOADED.load(Ordering::Acquire) {
```
Dirbaio, line 18: You need atomic CAS for this to be sound, otherwise you can have htis:

- thread A: sees LOADED=false
- thread B: sees LOADED=false
- thread B: writes to UID_HEX
- thread B: returns
- thread B: user code starts to read from UID_HEX
- thread A: writes to UID_HEX
- One thread writing and another reading without synchronization -> UB

the CS doesn't save you there, the semantics for `critical_section` are "no other CS can run concurrently", not "no other code can run concurrently". (the cortex-m single-core impl does guarantee the latter, but not necessarily all do)

Since CAS is annoying on M0's I'd just make LOADED a `static mut` too and access it inside the CS. A bit slower because you now have to take the CS if already loaded, but i'd say it's fine because this operation should be rare.

adamgreig: Good point, I've swapped as you suggested. Do you still prefer this over just computing it and returning an array directly?

Dirbaio: it's definitely a bit odd, but it's convenient to be able to get a `&str` and you only get the memory usage if you actually use it, so there's little downside I guess?

PR #2242, opened 2023-12-03
remove portable-atomic

lz1998:


Dirbaio:
what's the rationale for removing it? Is it because https://github.com/rust-lang/rust/pull/114499 is now merged, or is it something else? 

If it's the former, we can't remove portable-atomic yet. When Rust 1.75 is released at the end of the month, Embassy will switch to fully supporting stable, and that PR won't be released until Rust 1.76.

Dirbaio (approved):
1.76 has been out of a while, we can merge this now. i've rebased it for you. 

Thanks for the PR!

PR #2251, opened 2023-12-04
[FAQ] Add section about speed optimization

JuliDi:
As previously discussed in the chat, these are the speed optimizations recommended by @Dirbaio 
Thanks for the help!

### DIFF
--- a/docs/modules/ROOT/pages/faq.adoc
+++ b/docs/modules/ROOT/pages/faq.adoc
@@ -135,3 +135,19 @@ embassy-time = { git = "https://github.com/embassy-rs/embassy.git", rev = "e5fdd
 ----
 
 Note that the git revision should match any other embassy patches or git dependencies that you are using!
+
+== How can I optimize the speed of my embassy-stm32 program?
+
+* Make sure RCC is set up to go as fast as possible
+* Make sure link:https://docs.rs/cortex-m/latest/cortex_m/peripheral/struct.SCB.html[flash cache] is enabled
+* build with `--release`
+* Set the following keys for the release profile in your `Cargo.toml``:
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/faq.adoc
```diff
@@ -135,3 +135,19 @@ embassy-time = { git = "https://github.com/embassy-rs/embassy.git", rev = "e5fdd
...
+
+== How can I optimize the speed of my embassy-stm32 program?
+
+* Make sure RCC is set up to go as fast as possible
+* Make sure link:https://docs.rs/cortex-m/latest/cortex_m/peripheral/struct.SCB.html[flash cache] is enabled
+* build with `--release`
+* Set the following keys for the release profile in your `Cargo.toml``:
+    ** `opt-level=s`
+    ** `lto=fat`
+    ** `build-std=core`
```
jamesmunns, line 147: Are these stable in `Cargo.toml` now? Or do they still need to be in `.cargo/config.toml`?

JuliDi: Good question, I'm using nightly and it works from Cargo.toml. Though I got one notification in rustrover that `build-std=core` is unused. However, when I added it, it seemed like it was building std. I can look into it tomorrow.

JuliDi: Thank you, very good catch @jamesmunns ! Looks like it is still unstable and has to go to `.cargo/config.toml`, see https://doc.rust-lang.org/cargo/reference/unstable.html

I will adapt the faq entry accordingly

Dirbaio (approved):

PR #2254, opened 2023-12-05
embassy-executor 0.4

jessebraham:
Not really sure what the full release process is, but this is at least working with `esp-hal`. If you would like any further changes made please let me know and I'm happy to make them.


Dirbaio:
`portable-atomic` is still needed in 1.74 and 1.75, we shouldn't remove it yet. see #2242 

Could you add this to changelog too? sorry I forgot

```
- Added an arena allocator for tasks, allowing using the `main` and `task` macros on Rust 1.75 stable. (it is only used if the `nightly` feature is not enabled. When `nightly` is enabled, `type_alias_impl_trait` is used to statically allocate tasks, as before).
```

jessebraham:
Sorry, didn't notice that CI was red yesterday. Have updated all references to `embassy-executor@0.3.x`, so hopefully should be good to go here (pending the final CI job).

PR #2255, opened 2023-12-05
stm32/h7: fix h7 examples

badrbouslikhin:
This PR fixes the STM32H7 Ethernet examples.

The examples enable `dhcpv4` and `dns` features. Both use one socket, meaning that we need one extra socket for our firmware TCP socket.


Dirbaio (approved):
thank you!

PR #2257, opened 2023-12-06
[WIP] Add hardware chip select to SPI on STM32

SerenaLynas:
Adds hardware chip select to the spi peripheral with additional constructors (`new_cs`, `new_txonly_cs`, etc.). I've marked this as WIP because I don't have a physical board at the moment to test with. I'll test it when I get the boards in, but if anyone is eager to test this to make sure it works, be my guest.


SploxFox:
Seems like it doesn't build for all targets. I'll take a look tomorrow/overmorrow

Dirbaio:
we should be careful about how this interacts with the embedded-hal traits. There's 2 SPI traits, [SpiBus and SpiDevice](https://docs.rs/embedded-hal/1.0.0-rc.2/embedded_hal/spi/index.html). SpiBus must be implemented only when no CS pin is being managed (a bare bus), and SpiDevice must be implemented for SPIs with a CS pin.

So, the `Spi` struct, when created with hardware CS, should implement SpiDevice and not SpiBus. With the current code in this PR, it'll implement SpiBus. To do this we'd need a "mode" generic param, or perhaps do 2 different structs.

Also the CS pin should be toggled at start/end of the entire transaction, not individual transfers. I think STM32 hardware CS toggles for each transfer. So we'd have to do some hacks to override that.

What's your moativation for hardware CS? Performance?

Dirbaio:
Hello,

PR #2270, opened 2023-12-08
Assert CR2.USV on stm32wb55, wb35 series chips

Redrield:
As discussed in matrix, value must be asserted for USB to function properly.

https://www.st.com/resource/en/reference_manual/rm0434-multiprotocol-wireless-32bit-mcu-armbased-cortexm4-with-fpu-bluetooth-lowenergy-and-802154-radio-solution-stmicroelectronics.pdf pg. 175 for reference

### DIFF
--- a/embassy-stm32/src/usb/usb.rs
+++ b/embassy-stm32/src/usb/usb.rs
@@ -263,7 +263,7 @@ impl<'d, T: Instance> Driver<'d, T> {
 
         let regs = T::regs();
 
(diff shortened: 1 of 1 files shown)


Dirbaio:
can you do `#[cfg(any(stm32l5, stm32wb))]`? it's simpler, it's fine if it matches WB's without usb because this file isn't compiled at all there.

Dirbaio (approved):
thank you!

PR #2272, opened 2023-12-09
Add GP TIM9 and TIM11 to be used as time_driver candidates

cbruiz:
Add GP TIM9 and TIM11 to be used as time_driver


Dirbaio:
bender run

Dirbaio (approved):
thank you!

PR #2273, opened 2023-12-10
initial support for avr

djdisodo:
initial support for avr

tested

[embassy_time driver for avr](https://github.com/djdisodo/avr_tc0_embassy_time)
[example](https://github.com/djdisodo/avr_tc0_embassy_time)

image


lulf (approved):
Nice, thanks for the PR! :rocket:

lulf (changes requested):
Just one thing, could you add this to ci.sh so that it gets built for PRs?

djdisodo:
> Just one thing, could you add this to ci.sh so that it gets built for PRs?
got some problem here on ci/build
`error: unexpected argument '-Z' found`

janderholm:
> > The executor is polling in a hot loop. Is there an AVR equivalent of the "wait for interrupt" instruction? If so, it'd be good to use it. It helps both power consumption, and latency/jitter.
> 
> seems like there are some sleep modes(MCUCR) but i'm not sure how it should be implemented
> 
> how does it exit from wfi mode? will ANY interrupt do it?

As long as as you haven't changed the sleep mode select register, it should wake up on any interrupt:

https://microchipdeveloper.com/xwiki/bin/view/products/mcu-mpu/8-bit-avr/structure/sleep/

So an implementation similar to riscv should work:
https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/arch/riscv32.rs

But interrupts must be enabled before calling sleep so I don't think critical_section can be used:
https://onlinedocs.microchip.com/pr/GUID-A834D554-5741-41A3-B5E1-35ED7CD8250A-en-US-5/index.html?GUID-825A28EE-C4E4-4C03-864F-92AA9BA41231

djdisodo:
@janderholm 
in my implementation it wouldn't even sleep that long since it has to copy data from timer to sram quite often
is it still ok?
no drawbacks?

janderholm:
It should be fine. Waking from idle should take 4 clock cycles before servicing the interrupt.
There are other power saving states where oscillators are stopped which is going to take quite some time to stabilize on startup. But idle basically keeps everything running except for the CPU and flash clock.

https://microchipdeveloper.com/xwiki/bin/view/products/mcu-mpu/8-bit-avr/structure/sleep/

Dirbaio (approved):
LGTM! 

- Can you fix CI?
- Can you rebase, so no merge commits show up in the git history?

Thanks!

djdisodo:
i decided to not touch interrupt or se bit, this makes user decide weather to use sleep or not

djdisodo:
i think i found possible problem,

assume one interrupt is scheduled

1. executor reads SIGNAL_WORK_THREAD_MODE false and jumps to sleep code
2. interrupt kicks in and send SIGNAL_WORK_THREAD_MODE true
3. executor sleeps anyway
4. never wakes up or takes unexpectedly long to wake up since no interrupt is scheduled

janderholm:
Yeah unfortunately sei has to be used just before sleep. For future reference there is another trick where you let __pender just clear the se bit. That way the sleep instruction is a no-op. 

Then you can set the se bit just before poll and it will go to sleep after handling all waiting tasks if __pender is not executed again. 

But it's not clear that is any better since you need to set both se and sm at the same time and the implementation is chip dependent I think.

embassy-executor/src/arch/avr.rs
```diff
@@ -0,0 +1,72 @@
...
+        /// - a local variable in a function you know never returns (like `fn main() -> !`), upgrading its lifetime with `transmute`. (unsafe)
+        ///
+        /// This function never returns.
+        pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+            init(self.inner.spawner());
+
+            loop {
+                unsafe {
+                    avr_device::interrupt::disable();
+                    if SIGNAL_WORK_THREAD_MODE.swap(false, Ordering::SeqCst) {
```
Dirbaio, line 61: isn't it inverted? if `SIGNAL_WORK_THREAD_MODE` is true, it means there is more work to do, so it shouldn't sleep.

djdisodo: rushed on that, please squash commits

brainstorm:
Hello 👋🏻 

Thanks for this initial AVR support on embassy!

I've tried to write a very minimal example in https://github.com/brainstorm/embassy-as5600/commit/eb5f8488b4cdfa96aa1d889adb4421e6d37ae4db, but I'm running into the error described in the commit for both stable and nightly Rust... I'm most probably doing something wrong, but I thought whether `#[feature(impl_trait_in_assoc_type)]` need to be added to `embassy_executor` crate itself?

I took a peek at the [example linked in this PR](https://github.com/djdisodo/avr_tc0_embassy_time) and I see that it compiles fine, but it does not seem to have an entrypoint (main)... my plan was to write an embassy example, so it must contain an `executor::main` like any of the [embassy examples](https://github.com/embassy-rs/embassy/tree/main/examples)?

karioja:
@djdisodo I would really much like to try this out. However being a rust / embassy newbie I'm having hard time setting up the project. Could you please make the example (main.rs + Cargo.toml) available in some other form than as a screenshot?

simmsb:
Hey, if anyone's interested I've also been playing around with embassy on avr devices, I have a project here which is some firmware for flashlights: https://github.com/simmsb/tyrfing

djdisodo:
would be great if we had spi and i2c driver

PR #2276, opened 2023-12-11
Add example for using CAN with STM32F103 (BluePill) with a real CAN

cschuhen:
Add example for using CAN with STM32F103 (BluePill) with a real CAN. None of the existing examples focus on using a real transceiver. Also none of them map to alternate pins so this is easy to miss. In addition F1 STM's have slightly different IRQ names.


Dirbaio (approved):
thank you!

PR #2277, opened 2023-12-11
Introduce reset_{at|after} functions for Ticker.

Frostie314159:
These functions perform a similar job to reset, by reseting the timer either after a specified duration or at an instant.


Dirbaio:
Could you expand documentation a bit, what happens if the reset expiration is in the future by further than the tick rate, or in the past?

Frostie314159:
I expanded it a bit, if you have any improvements let me know.

Issue #2279, opened 2023-12-13
embassy-boot: Hitting assert in `assert_partitions` when DFU partition size equals ACTIVE partition size


plaes:
I feel that asserts defined in the [`assert_partitions`](https://github.com/embassy-rs/embassy/blob/14f41a71b6ea9dedb4ee5b9c741fe10575772c7d/embassy-boot/boot/src/boot_loader.rs#L401) function need some kind of explanations as I can't figure out the need for this one I'm hitting: `assert!(dfu.capacity() as u32 - active.capacity() as u32 >= page_size);`

In the linker configuration I have set following:
```
MEMORY
{
  /* NOTE 1 K = 1 KiBi = 1024 bytes */
  /* Softdevice - s132_nrf52_7.3.0: 152kB (0x26000 bytes including MBR) */
  MBR                               : ORIGIN = 0x00000000, LENGTH = 4K
  SOFTDEVICE                        : ORIGIN = 0x00001000, LENGTH = 148K
  ACTIVE                            : ORIGIN = 0x00026000, LENGTH = 164K 
  DFU                               : ORIGIN = 0x0004f000, LENGTH = 164K
...
```

Dirbaio:
DFU needs to be 1 page bigger than ACTIVE. This is needed for the swap algorithm to work.

lulf:
@plaes There is also some docs here that might help explaining why:

* https://embassy.dev/book/dev/bootloader.html
* https://blog.drogue.io/firmware-updates-part-1/

plaes:
I think we should leave this one still open to document the partition requirements in the top-level library.
For example that we need at least 4 partitions at minimum (FLASH, BOOTLOADER_STATE, ACTIVE and DFU) and how things need to be swapped around between app / bootloader.

lulf:
Fixed in #2283

PR #2280, opened 2023-12-13
embassy-boot: Add explanation to dfu vs active size assertion

plaes:
Some basic comments to deal with #2279.

Ideally it would be nice to have some documentation for `memory.x` as well, but this seems all to be target-specific (nrf vs pico vs stm).


lulf (approved):
Good clarifications, thanks!

PR #2281, opened 2023-12-13
allow for optional override of `Suspend` event for a UsbDevice

dstric-aqueduct:
- allow for optional override of `Suspend` event for a UsbDevice

Issue: 

- when testing the `usb_serial` example for the RP2040, the USB bus fails to respond after sleeping and then waking the OS (Win 10). 
- disabling the suspend behavior fixes the issue 

Fix:

- disabling handling of the suspend behavior fixes the issue
- behavior set by new `susependable` field on `embassy_usb::builder::Config`

### DIFF
--- a/embassy-rp/src/usb.rs
+++ b/embassy-rp/src/usb.rs
@@ -363,7 +363,7 @@ impl<'d, T: Instance> driver::Bus for Bus<'d, T> {
             let siestatus = regs.sie_status().read();
             let intrstatus = regs.intr().read();
 
-            if siestatus.resume() {
+            if siestatus.resume() || intrstatus.dev_resume_from_host() {
                 regs.sie_status().write(|w| w.set_resume(true));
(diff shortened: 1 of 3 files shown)


Dirbaio:
This is a workaround, we should instead actually fix the bug that is causing this.

dstric-aqueduct:
Adding a check of the DEV_RESUME_FROM_HOST bit in the USB INTR register in the `poll` method for the `Bus` appears to have fixed the issue.

Dirbaio:
Nice! Can you do just that change, and not add the `suspendable` flag? Thanks!

dstric-aqueduct:
> Nice! Can you do just that change, and not add the `suspendable` flag? Thanks!

Done - thanks for all your work on this wonderful project.

Dirbaio (approved):
thakn you!

PR #2282, opened 2023-12-13
feat: Extended the Scan API

umgefahren:
I have extended the scan APIs. I followed the mentioned reference implementation.

It is not tested though and it would be very kind if someone could before merging, my debug probe hasn't arrived yet.

### DIFF
--- a/cyw43/src/control.rs
+++ b/cyw43/src/control.rs
@@ -3,7 +3,7 @@ use core::iter::zip;
 
 use embassy_net_driver_channel as ch;
 use embassy_net_driver_channel::driver::{HardwareAddress, LinkState};
-use embassy_time::Timer;
+use embassy_time::{Duration, Timer};
 
 use crate::consts::*;
 use crate::events::{Event, EventSubscriber, Events};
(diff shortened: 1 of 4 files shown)


umgefahren:
Tested it, works!

cyw43/src/control.rs
```diff
@@ -1,5 +1,6 @@
 use core::cmp::{max, min};
 use core::iter::zip;
+use core::time::Duration;
```
Dirbaio, line 3: please use `embassy_time::Duration` instead.

(`core::time::Duration` is a bit bloated for embedded usage: it has u64 seconds + u32 nanoseconds, and requires expensive multiply/divide by 10^9 when doing math with it)

cyw43/src/control.rs
```diff
@@ -29,6 +30,45 @@ pub struct Control<'a> {
...
+#[derive(Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ScanType {
+    Active {
+        /// Period of time to wait on each channel when active scanning.
+        dwell_time: Option<Duration>,
+    },
+    Passive {
+        /// Period of time to wait on each channel when passive scanning.
+        dwell_time: Option<Duration>,
```
Dirbaio, line 42: maybe `dwell_time` could be a field in `ScanOptions` instead, since the two scan types have it equally?

Dirbaio:
I've backed out the BssInfo changes because they ended up conflicting with #2461 , i'm merging the rest.

PR #2285, opened 2023-12-14
embassy-boot: Fix formatting for tables

plaes:
Tables describing the a-b flashing were all garbled up in the cargo doc output, so fix up the syntax.


lulf (approved):
:rocket:  Thanks!

PR #2290, opened 2023-12-15
add ws2812 example for stm32f4 with PWM and DMA

eZioPan:


Dirbaio (approved):
thank you! 

very fancy, using PWM+DMA like this :eyes:

PR #2294, opened 2023-12-16
STM32: Enable flash support for STM32G4

adamgreig:
I checked the G4 RM and the required operations match up with what's already done for G0, so I combined the two. It seems like there's probably quite a lot of crossover with other families too, but I didn't attempt a refactor.

Many G4 devices have dual flash banks, but as far as I can tell the current stm32-data only knows about single banks for G4. By default, G4s with dual banks have the DBANK bit reset, so they can be treated as having one large bank, so I believe everything should work until someone wants to actually use the dual banks. At the moment I only have a single-bank G4 to test with so I can't find out for sure. I'll leave adding dual bank registers and HAL support for G4 to someone else...

I have tested this successfully on an STM32G431CB.


lulf (approved):
🚀🚀🚀

PR #2296, opened 2023-12-17
(yet another) stm32f4 ws2812 example ...

eZioPan:
... with spi,
and more doc on TIM&DMA version,
also remove useless TIM APRE settings,
and use for loop instead of manually flip the index bit,
and replace `embassy_time::Timer` with `embassy_time::Ticker`, for more constant time interval,
clean DMA FEIF if set after each DMA Transfer


eZioPan:
should do more fine tune on PWM example ...

Dirbaio (approved):
thank you!

PR #2300, opened 2023-12-17
STM32: Fix race in alarm setting, which impacted scheduling.

RobertTDowling:
Detect potential race condition (should be rare) and return false back to caller, allowing them to handle the possibility that either the alarm was never set because it was in the past (old meaning of false), or that in fact the alarm was set and may have fired within the race window (new meaning of false). In either case, the caller needs to make sure the callback got called.


Dirbaio (approved):
thank you!

PR #2306, opened 2023-12-18
Fix nb on rp uart

### LINKED ISSUE #2305: embassy-rp UART e_hal_nb recv has duplicate case
https://github.com/embassy-rs/embassy/blob/88e77c733cc3d52a2bd20e9e25446fe1d37c830b/embassy-rp/src/uart/mod.rs#L820-L839

Here, `dr.fe()` is checked twice, and I think we can never actually yield data?

Not sure if it is supposed to be more like this (from the ehal 0.2 impl):

```rust
impl<'d, T: Instance, M: Mode> embedded_hal_02::serial::Read<u8> for UartRx<'d, T, M> {
    type Error = Error;
    fn read(&mut self) -> Result<u8, nb::Error<Self::Error>> {
        let r = T::regs();
        if r.uartfr().read().rxfe() {
            return Err(nb::Error::WouldBlock);
        }

        let dr = r.uartdr().read();

        if dr.oe() {
            Err(nb::Error::Other(Error::Overrun))
        } else if dr.be() {
            Err(nb::Error::Other(Error::Break))
        } else if dr.pe() {
            Err(nb::Error::Other(Error::Parity))
        } else if dr.fe() {
            Err(nb::Error::Other(Error::Framing))
        } else {
            Ok(dr.data())
        }
    }
}
```

jamesmunns:
Closes #2305 

I'm not sure how to test this, but it now matches ehal v0.2 impl.

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -820,6 +820,10 @@ impl<'d, T: Instance, M: Mode> embedded_hal_nb::serial::ErrorType for Uart<'d, T
 impl<'d, T: Instance, M: Mode> embedded_hal_nb::serial::Read for UartRx<'d, T, M> {
     fn read(&mut self) -> nb::Result<u8, Self::Error> {
         let r = T::regs();
+        if r.uartfr().read().rxfe() {
+            return Err(nb::Error::WouldBlock);
+        }
+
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

Dirbaio:
bender run

PR #2307, opened 2023-12-18
stm32/can: docs, cleanup interrupt handling.y

Dirbaio:
- stm32/can: cleanup interrupt traits.
- stm32/can: docs.


Dirbaio:
bender run

PR #2311, opened 2023-12-18
Add a basic RP "read to break" function

jamesmunns:
This adds a function called "read_to_break" that receives until one of the following happen:

1. We receive the whole buffer amount
2. We receive a line break interrupt

In the latter case, we figure out how many bytes we DID get, and then return that subslice. In the former case, we return the whole subslice.

All other errors are returned the same.

One note:

This is basically a big copy/paste of `read`. I should probably clean that up, I could do the DMA reg check in a function that basically does:

```rust
pub async fn read_to_break<'a>(&mut self, buf: &'a mut [u8]) -> Result<&'a mut [u8], Error> {
    match self.read(buf) {
        Ok(()) => Ok(buf),
        Err(Break) => {
            // get bytes received...
            Ok(&mut buf[..count])
        }
        Err(e) => Err(e)
    }
}
```

But then we'd have to unwrap `self.rx_dma` again, which is annoying. Open to thoughts.

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -125,6 +125,7 @@ pub enum Error {
 pub enum ReadToBreakError {
     /// Read this many bytes, but never received a line break.
     MissingBreak(usize),
+    /// Other, standard issue with the serial request
     Other(Error),
 }
 
@@ -936,6 +937,9 @@ impl<'d, T: Instance> Uart<'d, T, Async> {
         self.rx.read(buffer).await
     }
 
+    /// Read until the buffer is full or a line break occurs.
+    ///
+    /// See [`UartRx::read_to_break()`] for more details
     pub async fn read_to_break<'a>(&mut self, buf: &'a mut [u8]) -> Result<usize, ReadToBreakError> {
         self.rx.read_to_break(buf).await
     }
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -278,13 +278,16 @@ impl<'d, T: Instance, M: Mode> UartRx<'d, T, M> {
 
     /// Read from UART RX blocking execution until done.
     pub fn blocking_read(&mut self, mut buffer: &mut [u8]) -> Result<(), Error> {
-        while buffer.len() > 0 {
+        while !buffer.is_empty() {
             let received = self.drain_fifo(buffer).map_err(|(_i, e)| e)?;
             buffer = &mut buffer[received..];
         }
         Ok(())
     }
 
+    /// Returns Ok(len) if no errors occurred. Returns Err((len, err)) if an error was
+    /// encountered. in both cases, `len` is the number of *good* bytes copied into
+    /// `buffer`.
     fn drain_fifo(&mut self, buffer: &mut [u8]) -> Result<usize, (usize, Error)> {
         let r = T::regs();
         for (i, b) in buffer.iter_mut().enumerate() {
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -116,6 +116,10 @@ pub enum Error {
     Parity,
     /// Triggered when the received character didn't have a valid stop bit.
     Framing,
+    /// There was an issue when calculating the number of transferred items
+    /// in an aborted DMA transaction. This is likely an error in the
+    /// driver implementation, please open an embassy issue.
+    Calculation,
 }
 
 /// Internal DMA state of UART RX.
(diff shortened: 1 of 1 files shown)


embassy-rp/src/uart/mod.rs
```diff
@@ -411,6 +411,90 @@ impl<'d, T: Instance> UartRx<'d, T, Async> {
...
+            w.set_feim(true);
+        });
+        T::regs().uartdmacr().write_set(|reg| {
+            reg.set_rxdmae(true);
+            reg.set_dmaonerr(true);
+        });
+        let transfer = unsafe {
+            // If we don't assign future to a variable, the data register pointer
+            // is held across an await and makes the future non-Send.
+            crate::dma::read(&mut ch, T::regs().uartdr().as_ptr() as *const _, sbuffer, T::RX_DREQ)
```
jamesmunns, line 454: Note that I take `&mut ch` here, so I can re-use `ch` later in the calculations.

embassy-rp/src/uart/mod.rs
```diff
@@ -411,6 +411,90 @@ impl<'d, T: Instance> UartRx<'d, T, Async> {
...
+        let mut ch = self.rx_dma.as_mut().unwrap();
+        T::regs().uartimsc().write_set(|w| {
+            w.set_oeim(true);
+            w.set_beim(true);
+            w.set_peim(true);
+            w.set_feim(true);
+        });
+        T::regs().uartdmacr().write_set(|reg| {
+            reg.set_rxdmae(true);
+            reg.set_dmaonerr(true);
```
jamesmunns, line 449: I *think* this halts DMA on errors like line breaks, good to double check tho.

jamesmunns:
> what about calling it `read_until_idle` like stm32 already does?

So @lights0123, this interface waits for a **line break** interrupt (the line is held LOW for >2 character times), rather than an **idle** interrupt (where the line is held HIGH with no data for >32 bit times).

In theory, the RP2040 also has an idle interrupt, but it is difficult to use with DMA (the interrupt doesn't fire if the the FIFO is empty, which it often will be DMA), which is why I wanted to use line breaks instead.

jamesmunns:
Also, I did try reading some other registers, but it seems they get invalidated when the transaction is aborted:

```rust
// Begin "James is a chicken" region - I'm not certain if there is ever
// a case where the write addr WOULDN'T exist between the start and end.
// This assert checks that and hasn't fired (yet).
let sval = buffer.as_ptr() as usize;
let eval = sval + buffer.len();
// Note: the `write_addr()` is where the NEXT write would be, BUT we also
// received one extra byte that represents the line break.
let val = ch.regs().write_addr().read() as usize - 1;
assert!((val >= sval) && (val <= eval));
let taken = val - sval;
let txns = ch.regs().trans_count().read();
let a = ch.regs().dbg_ctdreq().read().0;

defmt::println!("taken: {=usize} {=u32} {=u32}", taken, txns, a);
```

Gives:

```
...
taken: 70 0 0
taken: 84 0 0
taken: 14 0 0
taken: 28 0 0
taken: 42 0 0
taken: 56 0 0
...
```

Where the values calculated from write_addr seem to always be correct.

embassy-rp/src/uart/mod.rs
```diff
@@ -100,6 +100,10 @@ pub enum Error {
     Parity,
     /// Triggered when the received character didn't have a valid stop bit.
     Framing,
+    /// There was an issue when calculating the number of transferred items
+    /// in an aborted DMA transaction. This is likely an error in the
+    /// driver implementation, please open an embassy issue.
+    Calculation,
```
Dirbaio, line 106: for "should never happen unless there's a bug in the driver" scenarios, we should panic instead of returning an error to the caller.

jamesmunns: Got it! Will switch back to that.

embassy-rp/src/uart/mod.rs
```diff
@@ -411,6 +437,138 @@ impl<'d, T: Instance> UartRx<'d, T, Async> {
...
+
+    /// Read from the UART, until one of the following occurs:
+    ///
+    /// * We read `buffer.len()` bytes without a line break
+    ///     * returns `Ok(buffer)`
+    /// * We read `n` bytes then a line break occurs
+    ///     * returns `Ok(&mut buffer[..n])`
+    /// * We encounter some error OTHER than a line break
+    ///     * returns `Err(Error)`
+    pub async fn read_to_break<'a>(&mut self, buffer: &'a mut [u8]) -> Result<&'a mut [u8], Error> {
```
Dirbaio, line 449: could you change this to return `usize` instead of a subslice? makes the signature simpler, and is more consistent with other APIs in embassy and embedded-io.

jamesmunns: I really don't like that signature, but I can change it back if you'd like. Will discuss in chat.

jamesmunns:
Note that https://github.com/raspberrypi/pico-feedback/issues/367 is related to some of the impl details of this PR.

jamesmunns:
Just noting that this should be ready for review whenever.

jamesmunns:
Also noting that my usage of UARTRSR *IS* consistent with the Arm PrimeCell UART PL011 docs, and it appears this was just a deficiency of the RP2040 datasheet which didn't copy all of the details from the upstream docs:

https://developer.arm.com/documentation/ddi0183/g/programmers-model/register-descriptions/receive-status-register---error-clear-register--uartrsr-uartecr?lang=en

Comment:

https://github.com/raspberrypi/pico-feedback/issues/367#issuecomment-1900195055

lulf (approved):
I think this is good to go. Earlier review comments seems to be addressed and I don't really have anything to add. Thanks for the excellent code comments, it really helped reviewing the code as well.

lulf:
bender run

Dirbaio:
bender run

jamesmunns:
@Dirbaio back at you

Dirbaio:
@jamesmunns missing docs :D

PR #2312, opened 2023-12-19
Extend RTC low power mode for STM32G0

Ralim:
Hello,

_First time contributing, so please let me know where I can improve here. Super open to fixes or better style_

Trying to enable the `low-power` mode support for the STM32G0 series.

I've based this off `RM0454 Rev 5` for the STM32G070 series.

This does build and run locally and appears to work; but I haven't done in depth testing (are there any guides on how to best test this otherwise, or add automatic testing?).


Dirbaio (approved):
thank you!

PR #2314, opened 2023-12-19
stm32: i2c: Clean up conditional code a bit

plaes:
By moving conditional code inside the functions, we can reduce duplication and in one case we can even eliminate one...


Dirbaio (approved):
nice! thanks

PR #2327, opened 2023-12-19
fix stm32 rtc year from 1970 base 2000 to make leap years work

dragonnn:
STM32 RTC does store year as last two digits of a year and is 2000 year based, not 1970 unix time start.
Having the offset set right is important for the calendar to properly handle leap years.
With 1970 base setting the RTC time to 2024-02-28 23:59:59 and waiting one sec does make it jump to 2024-03-01 23:59:50 with is wrong because 2024 is a leap year.
With 2000 base it works correctly and doesn't skip 02-29.
Tested on STM32F401


Dirbaio (approved):
thanks! and nice find, this would've been a tricky bug :smiling_face_with_tear:

PR #2328, opened 2023-12-20
feat(embassy-sync): Add try_take() to signal

rmja:


embassy-sync/src/signal.rs
```diff
@@ -111,6 +111,17 @@ where
         poll_fn(move |cx| self.poll_wait(cx))
     }
 
+    /// non-blocking method to try and take the signal value.
+    pub fn try_take(&self) -> Option<T> {
+        self.state.lock(|cell| {
+            let state = cell.replace(State::None);
```
Dirbaio, line 117: This will remove the `Waiting` state, so if there was another task waiting with async `wait()` it'll hang.

rmja: Good catch

Dirbaio (approved):

PR #2331, opened 2023-12-20
track DMA circular mode in `Transfer` struct, when `poll`, just return `Ready` without checking `.is_running()`

eZioPan:
Since in circular mode, DMA will keep running until manually stop. But when user `.await` a circular `Transfer`, they got no chance (in the same thread) to manually stop it.

Unresolved:
* Should we read `Circ` bit from register every time it pulls, or record it in struct?
* Should we reject when user set both `FlowControl::Peripheral` and `circular` mode in one `TransferOptions`?
  * If we do reject setting them at the same time, then we can safely store the `circular` mode in `Transfer` struct, and reduce an AHB access on each poll (also make `Transfer` a little bigger)


eZioPan:
Ok, it makes no sense to return on circular mode set. Since when `Transfer` return and dropped, the DMA Channel just turn off.
We might need a doc to tell user spawn the another task for circular mode DMA.

PR #2338, opened 2023-12-21
docs: add a note about circular mode DMA

eZioPan:


lulf (approved):
Good addition, thanks!

lulf:
bender run

PR #2340, opened 2023-12-21
Add #[test] macro for arch-std

t-moe:
Copied and adapted from:
* https://github.com/drogue-iot/ector/blob/main/ector/src/testutils.rs
* https://github.com/drogue-iot/drogue-device/blob/main/macros/src/lib.rs

which could probably both be simplified with this PR

### DIFF
--- a/embassy-executor-macros/src/macros/test.rs
+++ b/embassy-executor-macros/src/macros/test.rs
@@ -1,0 +1,66 @@
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{self, ReturnType};
+
+use crate::util::ctxt::Ctxt;
+
+pub fn run(f: syn::ItemFn) -> Result<TokenStream, TokenStream> {
+    let ctxt = Ctxt::new();
+
+    if f.sig.asyncness.is_none() {
+        ctxt.error_spanned_by(&f.sig, "test functions must be async");
+    }
+    if !f.sig.generics.params.is_empty() {
+        ctxt.error_spanned_by(&f.sig, "test functions must not be generic");
+    }
+
+    let args = f.sig.inputs.clone();
+
+    if args.len() > 1 {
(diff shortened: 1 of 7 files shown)


lulf (approved):
Seems like a useful addition to me!

embassy-executor-macros/src/macros/test.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        async fn #embassy_test_name(#args) {
+            #task_fn_body
+        }
+
+        #[test]
+        fn #test_name() {
+            let r = ::embassy_executor::_export_testutils::TestRunner::default();
+
+            let r1: &'static mut ::embassy_executor::_export_testutils::TestRunner = unsafe { core::mem::transmute(&r) };
```
Dirbaio, line 56: This is unsound, the executor really must live for `'static`.  This is because the executor will hand out `Waker`s pointing to it, and `Waker` is `'static`.

Dirbaio: Also transmuting from `&` to `&mut` is unsound too.

t-moe: Ah, I thought that this might be ok, since the executor will never live longer than the test function....

Is there a way to fix this, without providing a custom test runner (1 executor for all tests)?

Dirbaio: the executor fundamentally requires `'static`. (other executors refcount tasks/wakers, to avoid dangling pointers while still allowing them to be deallocated. embassy's just doesn't refcount and assumes things live forever, this is what allows it to be so small/fast)

A global executor for all tests would work, but if you spawn one task per test you might still run out of RAM if you have lots of tests, because each will permanently occupy a chunk of RAM even while not runniing. 

Due to this I'm not sure if `embassy-executor` is a good fit for embedded testing...?

t-moe: I see. thanks for you feedback. I'll focus on embedded-test then or use generic-queue when testing embassy related code on std....

PR #2342, opened 2023-12-21
ci: use beta, add secondary nightly ci.

Dirbaio:


lulf (approved):
Matches a lot of the changes I was looking at trying beta myself :+1:

Issue #2344, opened 2023-12-22
Async Radio driver for embassry_nrf


guissalustiano:
Hi! I'm Interested in developing some Bluetooth tools for Rust (as my bachelor's thesis),
as a first step, I would like to implement the Radio driver with async support. Do you think it makes sense?

I started the development [here](#2351) but I'm unsure about the best API approach. The radio hardware varies between the boards, a quick resume:

| Board     | BLE 1M | BLE 2M | BLE 512K | BLE 128K | 802.15.4 | Nordic 1M | Nordic 2M |
|-------------|------------|-------------|---------------|---------------|-------------|----------------|---------------|
| nrf52805 | ✅ | ✅  | ❌ | ❌ | ❌ | ✅ | ✅ |
| nrf52810 | ✅ | ✅  | ❌ | ❌ | ❌ | ✅ | ✅ |
| nrf52811 | ✅ | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| nrf52820 | ✅ | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| nrf52832 | ✅ | ✅  | ❌ | ❌ | ❌ | ✅ | ✅ |
| nrf52833 | ✅ | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| nrf52840 | ✅ | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| nrf5340   | ✅ | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| nrf9160   | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |

## Desing
Given the uncertainty of the best interface and based on the feedback I will move development to another project, in order to iterate faster, and perhaps in the future create a generic library for any framework.
However, for this, I need some primitives to access the radio and wait for the interruption, and this must be implemented here as it requires access to the poll executor.

## Steps
First, implement POC to understand the necessary register and interrupts, and based on that implement a low-level API (specify for BLE now, but could be extended to others formats later)
- [X] Implement blocking transmission
- [X] Implement blocking receive
- [X] Implement async transmission
- [x] Implement async receive
- [x] Define a low-level API on the embassy and move the work to another repository

## Previous related issues:
- #432
- #1800

## References
[...]

PR #2351, opened 2023-12-22
Radio transmiter driver for embassry_nrf

guissalustiano:
As described in #2344, this is the first step to a full async radio driver for the nrf.
I'm still not so sure about the interface, but probably when someone starts to build something on top of them we can discover one more ergonomic API.

This is my first time contributing, so please let me know where I can improve here. Super open to fixes or better style.

I only have one nrf52840DK to test, but shouldn't be any different look on the datasheet.

Here is the example output, looked from one nrf52840 dongle ble sniffer
image


lulf (commented):
Nice work! I don't have any specific comments at the moment, but I think adding a impl Drop for disabling the radio should be added too.

As for the interface I think you need a more complex interaction with tx and rx to flesh that one out. 

Instead of adding a separate TxRadio, you could add a `fn blocking_send(&mut self, buffer: &[u8], tx_config: &TxConfig)` which imho is easier to use.

guissalustiano:
For now I finish the work here, but I will keep as a draft until have a 0.1 release for the [jewel library](https://github.com/guissalustiano/jewel-rs) witch this implementation depends.

embassy-nrf/Cargo.toml
```diff
@@ -150,6 +153,8 @@ embedded-storage-async = "0.4.0"
 cfg-if = "1.0.0"
 document-features = "0.2.7"
 
+jewel = { version = "0.1.0", git = "https://github.com/jewel-rs/jewel", optional = true }
```
lulf, line 156: No need to have this dependency.

embassy-nrf/Cargo.toml
```diff
@@ -161,3 +166,6 @@ nrf52840-pac = { version = "0.12.0", optional = true }
 nrf5340-app-pac = { version = "0.12.0", optional = true }
 nrf5340-net-pac = { version = "0.12.0", optional = true }
 nrf9160-pac = { version = "0.12.0", optional = true }
+
+[patch.crates-io]
+embassy-time = { version = "0.3.0", path = "../embassy-time", optional = true }
```
lulf: This should not be necessary, only in the binary crates.

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,432 @@
...
+
+use core::future::poll_fn;
+use core::sync::atomic::{compiler_fence, Ordering};
+use core::task::Poll;
+
+use embassy_hal_internal::drop::OnDrop;
+use embassy_hal_internal::{into_ref, PeripheralRef};
+use jewel::phy::{
+    AdvertisingChannel, Channel, ChannelTrait, HeaderSize, Mode, Radio as BleRadio, ADV_ADDRESS, ADV_CRC_INIT,
+    CRC_POLY, MAX_PDU_LENGTH,
```
lulf, line 22: Hmm, I was not imagining that embassy-nrf would depend on jewel. I think it should be the other way around, any constants or trait implementations should be in jewel.

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,431 @@
...
+            w.statlen()
+                .bits(0)
+                // MaxLen configures the maximum packet payload plus add-on size in
+                // number of bytes that can be transmitted or received by the RADIO. This feature can be used to ensure
+                // that the RADIO does not overwrite, or read beyond, the RAM assigned to the packet payload. This means
+                // that if the packet payload length defined by PCNF1.STATLEN and the LENGTH field in the packet specifies a
+                // packet larger than MAXLEN, the payload will be truncated at MAXLEN
+                //
+                // To simplify the implementation, I'm setting the max length to the maximum value
+                // and I'm using only the length field to truncate the payload
```
lulf, line 68: This feels nitpicky, but try to not comment things as 'first person'. Also leave a TODO marker for things you think can be improved later, it makes it easier to find for newcomers to the code wanting to contribute.

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,431 @@
...
+        let drop = OnDrop::new(|| {
+            trace!("radio drop: stopping");
+
+            r.intenclr.write(|w| w.end().clear());
+            r.events_end.reset();
+
+            r.tasks_stop.write(|w| w.tasks_stop().set_bit());
+
+            // The docs don't explicitly mention any event to acknowledge the stop task
+            // So I guess it's the same as end
```
lulf, line 163: Again, last line a bit unnecessary :)

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,431 @@
...
+            compiler_fence(Ordering::SeqCst);
+
+            // Acknowledge it
+            r.events_disabled.reset();
+        }
+    }
+
+    /// Set the radio mode
+    ///
+    /// The radio must be disabled before calling this function
```
lulf, line 223: I see several functions that require radio to be disabled. I think it would be good to add an assertion at the start of the function to check that for these functions.

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,431 @@
...
+        let ptr = buffer.as_ptr();
+
+        // Configure the payload
+        r.packetptr.write(|w| unsafe { w.bits(ptr as u32) });
+
+        Ok(())
+    }
+
+    /// Send packet
+    pub async fn transmit(&mut self) {
```
lulf, line 392: Should the buffer be passed as part of transmit/receive to ensure that it lives for the duration of transmission, or is set_buffer useful outside of 'transmit/receive'?

lulf (commented):
Great stuff, I've left a few comments inline. Apologies if there are good reasons for the way things are wrt. set_buffer.

embassy-nrf/src/radio/ble.rs
guissalustiano: That makes complete sense, thank you very much for the feedback!

embassy-nrf/src/radio/ble.rs
guissalustiano: On legacy advertising, the radio transmits the same data multiple times across multiple channels, this ways It's possible to change set the buffer just once and make a cheaper transmission call ([example on jewel](https://github.com/jewel-rs/jewel/blob/90a2e498db4766c816cdf4f1451387756071cb05/jewel/src/ll/mod.rs#L76-L112)).
But I agree that this is a micro-optimization and unnecessary for this case and maybe we should prefer a simpler interface. 
What do you think?

guissalustiano: I already answered myself in my own comment, so I already made the changes.

embassy-nrf/src/radio/ble.rs
```diff
@@ -0,0 +1,438 @@
...
+pub use pac::radio::mode::MODE_A as Mode;
+use pac::radio::pcnf0::PLEN_A as PreambleLength;
+use pac::radio::state::STATE_A as RadioState;
+pub use pac::radio::txpower::TXPOWER_A as TxPower;
+
+use crate::interrupt::typelevel::Interrupt;
+use crate::radio::*;
+use crate::util::slice_in_ram_or;
+
+/// UART error.
```
lulf, line 18: UART => RADIO

lulf (approved):
Thanks for addressing the comments, I think it looks good to merge unless you have more things you'd like to add.

PR #2361, opened 2023-12-26
stm32: USB IN endpoints use IN wakers

### LINKED ISSUE #2360: stm32 usb_cdc split wait_connection doesn't work
on stm32g431cb, calling `CdcAcmClass::split` returns two structs, which are designed to allow reading and writing in separate tasks. They each provide a function, `wait_connection()`, to wait until a USB host is connected and communications are ready. However, this function has a problem: the first task to call this function has its future complete, while the other waits forever and is never woken. My guess is that wakers are overwritten?

lights0123:
fixes #2360


Dirbaio (approved):
nice find, thank you!

PR #2364, opened 2023-12-27
STM32: Re-enable I2Cv1 Interrupts if poll-fn is not ready

jamesmunns:
This seems to help with my hanging problem, though I'm unsure if this
is the "right" solution.

If it is, it is possible we should do this in more places, since interrupts seem to usually disable themselves after firing.


Dirbaio (approved):
thanks for finding and debugging this! :D

PR #2366, opened 2023-12-28
Sethiam f4 adc v2 dma

sethkrie:
Opening a PR for adc_v2 DMA + Continuous + Scanning API.

Still very quick and dirty, but it works. Tested out 36 conversions of 6 channels for a single DMA transfer, which takes about 400-440 uS. No turbo wakers here, either & extra delays created by trace!() calls.

Issue is removing the trace!() seems to break the poll_fn(). But basic functionality is there!


Dirbaio:
- You're adding the RXDMA generic param to `Adc`, which makes it harder to use: everyone using ADC must now specify it, even if they don't want DMA. I think the API should be designed in a different way, something like: main `Adc` struct has no DMA, but you can call a method on it to convert it into another "ADC with DMA" struct. Or perhaps the method itself does the DMA transfer, so you don't need a second struct at all. You can borrow DMA channels as an argument for a function, so it doesn't have to be on the entire struct. See https://github.com/embassy-rs/embassy/pull/2367 which did this for PWM DMA.
- `get_dma_buf` is reading from `self.data` while DMA is writing to it, which is undefined behavior.
- Can you fix CI? I prefer CI to be passing before doing a full review, since it'll require updating all examples to the new API, which can reveal usability issues.

sethkrie:
It looks like ci/build fails for an unrelated reason? I haven't touched wb or RTC.

sethkrie:
Using the new ReadableRingBuf API & read-exact(). This should resolve the issues earlier regarding undefined behavior in DMA buffer reads.

The example shows the use of two DMA scan-mode & continuous ADCs running concurrently. Each ADC reads 2 channels & with minimal defmt prints (i.e. just printing dt), the pair of 512 sample buffers containing 256 samples / channel can be filled and read in 475us. I.e. 1024 Sa/475us = 2.15 MSa/s or 0.54 MSa/s per channel. I checked that the reads are RT & correct.

sethkrie:
@Dirbaio The CI fails here are outside of the scope of this PR. How can I address them?

Dirbaio:
failure is because you've changed the signature of ADC `.read()`, you have to update the test.

embassy-stm32/src/dma/dma.rs
```diff
@@ -0,0 +1,1049 @@
...
+use crate::interrupt::typelevel::Interrupt;
+use crate::interrupt::Priority;
+use crate::pac::dma::{regs, vals};
+use crate::{interrupt, pac};
+
+/// DMA transfer options.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct TransferOptions {
```
andresv, line 23: I am also looking into ADC. Why separate `dma.rs` file was needed?

liarokapisv:
I am also interested in adding dma/scanning support for the H7 series.
It would probably be best to collaborate to unify the interface across the series, but I don't want to delay the merge of these changes.

Personally, I would prefer to focus towards a "conversion" interface, similar to that of the Uart drivers. It is more feature-proof and can encode invariants that make methods more performant due to not having to check that incompatible methods were called. Think continuous vs discontinuous scanning and similar.

andresv:
@liarokapisv at the moment everything is open and it would be actually preferable to get competing PRs to get API right.

jas1339:
I was wondering if this pr was works? I was interested in using the dma for the adc on stm32h7.

sethkrie:
@jas1339 This does work and it's fast, the PR just needs some cleanup to not break the existing API and pass CI. Just haven't gotten around to it cause this a project outside of my working hours.

Problems I see are the lack of a strict timer-based sampling interrupt. It's just free running, so there's gonna be some drift in sample time unless that can be set up. Otherwise, if the use case isn't too sensitive to that, you'll generally get conversion around the specified sampling time.

embassy-stm32/src/dma/dma.rs
andresv: @sethkrie do you remember why additional `dma.rs` file was added.

andresv:
@sethkrie I can see that this PR adds `dma.rs` under adc, which is a lot like [dma_bdma.rs](https://github.com/embassy-rs/embassy/blob/b3252c3d4fbc5b00db2346f20130e468cf5ad05f/embassy-stm32/src/dma/dma_bdma.rs). What was the main issue why new impl was needed and `dma_bdma.rs` was not sufficient?

liarokapisv:
I made a [similar PR for the v4](https://github.com/embassy-rs/embassy/pull/3106). Mainly cross-posting here to compare and discuss the API design.

andresv:
This one can be close, it is superseded by: https://github.com/embassy-rs/embassy/pull/3116

PR #2367, opened 2023-12-28
implement PWM waveform generating with DMA

eZioPan:
This is almost just a generalized version of (original, will be gone/simplify after this PR) example/stm32f4/ws2812_pwm_dma.rs

### DIFF
--- a/embassy-stm32/src/timer/simple_pwm.rs
+++ b/embassy-stm32/src/timer/simple_pwm.rs
@@ -55,11 +55,12 @@ channel_impl!(new_ch3, Ch3, Channel3Pin);
 channel_impl!(new_ch4, Ch4, Channel4Pin);
 
 /// Simple PWM driver.
-pub struct SimplePwm<'d, T> {
+pub struct SimplePwm<'d, T, Dma> {
     inner: PeripheralRef<'d, T>,
+    dma: PeripheralRef<'d, Dma>,
 }
 
-impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
+impl<'d, T: CaptureCompare16bitInstance, Dma> SimplePwm<'d, T, Dma> {
     /// Create a new simple PWM driver.
     pub fn new(
         tim: impl Peripheral<P = T> + 'd,
@@ -69,16 +70,22 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
         _ch4: Option<PwmPin<'d, T, Ch4>>,
         freq: Hertz,
         counting_mode: CountingMode,
+        dma: impl Peripheral<P = Dma> + 'd,
     ) -> Self {
-        Self::new_inner(tim, freq, counting_mode)
+        Self::new_inner(tim, freq, counting_mode, dma)
     }
 
-    fn new_inner(tim: impl Peripheral<P = T> + 'd, freq: Hertz, counting_mode: CountingMode) -> Self {
-        into_ref!(tim);
+    fn new_inner(
+        tim: impl Peripheral<P = T> + 'd,
+        freq: Hertz,
+        counting_mode: CountingMode,
+        dma: impl Peripheral<P = Dma> + 'd,
+    ) -> Self {
+        into_ref!(tim, dma);
 
         T::enable_and_reset();
 
-        let mut this = Self { inner: tim };
+        let mut this = Self { inner: tim, dma };
 
         this.inner.set_counting_mode(counting_mode);
         this.set_frequency(freq);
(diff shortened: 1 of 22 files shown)


eZioPan:
Finally, get compiling pass.

I feel a little bit sad, that I do want to make a example to demonstrate how to manually use a `dma::Transfer`. But in the end, it turns out to become another driver/method. I guess every demo about `dma::Transfer` will eventually turn into a driver/method. So nothing to be sad about.

eZioPan:
should create PwmChannel as DacChannel do

lights0123:
hah, you made this less than 5 hours after I ported your f4 example to my g4 (and learned to look up DMA channels in the datasheet which is obviously much simplified now) :).

> I do want to make a example to demonstrate how to manually use a dma::Transfer

What about an example that continuously refills a circular buffer to write to more LEDs than RAM available for a buffer (like [many C versions do](https://github.com/MaJerle/stm32-ws2811-ws2812-ws2812b-ws281x-tim-pwm-dma-timer))? I imagine you could use the similar `WritableRingBuffer` in a high priority task, or possibly normal transfer APIs while hooking into half-transfer-complete interrupts.

eZioPan:
> What about an example that continuously refills a circular buffer to write to more LEDs than RAM available for a buffer (like many C versions do)? I imagine you could use the similar WritableRingBuffer in a high priority task, or possibly normal transfer APIs while hooking into half-transfer-complete interrupts.

Good idea! But I don't have much time these days(sorry...). I appreciate if you can write an example about dma::Transfer.

Dirbaio:
looks great!

Just one suggestion on the API: You can keep the `SimplePwm` API simple if you pass the DMA channel directly to `gen_waveform`. This way it doesn't "get in the way" if you don't want DMA (the param in `new`, the `Dma` generic param..). Something like this should work:

```rust
pub async fn gen_waveform(&mut self, dma: impl Peripheral<P = impl super::UpDma<T>>, channel: Channel, duty: &[u16]) {
```


it allows passing a `&mut` to the DMA channel, so you can call it multiple times without losing access to the singleton.

embassy-stm32/src/timer/simple_pwm.rs
```diff
@@ -135,13 +152,100 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
...
-impl<'d, T: CaptureCompare16bitInstance> embedded_hal_02::Pwm for SimplePwm<'d, T> {
+impl<'d, T: CaptureCompare16bitInstance + Basic16bitInstance, Dma> SimplePwm<'d, T, Dma>
+where
+    Dma: super::UpDma<T>,
+{
+    /// Generate a sequence of PWM waveform
+    pub async fn gen_waveform(&mut self, channel: Channel, duty: &[u16]) {
+        assert!(duty.iter().all(|v| *v <= self.get_max_duty()));
+
+        #[cfg_attr(any(stm32f334, stm32f378), allow(clippy::let_unit_value))]
```
Dirbaio, line 176: if the warning gets in the way, just always allow it, don't mention explicit chips.

embassy-stm32/src/timer/simple_pwm.rs
```diff
@@ -135,13 +152,100 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
...
+        let req = self.dma.request();
+
+        #[cfg(not(any(bdma, gpdma)))]
```
Dirbaio, line 179: we should avoid touching DMA registers from outside the `embassy_stm32::dma` code. Could you move the fifo issue handling into DMA code? Maybe always clear it, maybe add an option.

Also, are you sure it's reuqired to clear FEIF? I'd say it's fine to leave it set forever, it doesn't cause DMA to stop working (or does it?)

eZioPan: According to my test, FEIF doesn't affect DMA. I just thought we can take response to clear it.

I just remove all calls to dma regs, leave a note about UDE and FEIF. Once `dma::Transfer::clear_irqs()` or something similar get publiced, we can use it clean the flag.

embassy-stm32/src/timer/simple_pwm.rs
eZioPan: Done

eZioPan:
> You can keep the `SimplePwm` API simple if you pass the DMA channel directly to `gen_waveform`.

@Dirbaio done, thanks for suggestion!

embassy-stm32/src/timer/simple_pwm.rs
```diff
@@ -135,12 +139,90 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
...
+        self.inner.get_compare_value(channel)
+    }
+
     /// Set the output polarity for a given channel.
     pub fn set_polarity(&mut self, channel: Channel, polarity: OutputPolarity) {
         self.inner.set_output_polarity(channel, polarity);
     }
 }
 
+impl<'d, T: CaptureCompare16bitInstance + Basic16bitInstance> SimplePwm<'d, T> {
```
Dirbaio, line 155: this shouldn't need a separate impl block anymore

eZioPan: So, we will require `T: CaptureCompare16bitInstance + Basic16bitInstance` to use every other associate-functions&methods. I don't know whether this will add extra (maybe unneeded) restraint to those associate-functions&methods other than `.gen_waveform()`.

Dirbaio: CaptureCompare16bitInstance inherits from GeneralPurpose16bitInstance which inherits from Basic16bitInstance.

So `T: CaptureCompare16bitInstance + Basic16bitInstance` is the same as `T: CaptureCompare16bitInstance`

eZioPan: But we still need to expicitly write `T: CaptureCompare16bitInstance + Basic16bitInstance` instead of `T: CaptureCompare16bitInstance` right?
I find if I use the less constrait one, rustc will put a error "the trait bound `T: Basic16bitInstance` is not implemented for `T`"
(Sorry I still get confused on rust trait system)

Dirbaio: ah, there were a few missing supertrait bounds (the lowlevel traits did inherit, the public ones didn't).  I've added them.

Dirbaio (approved):
thank you!

PR #2368, opened 2023-12-28
Add low-power support for stm32l5

chrenderle:
This PR adds support for the low-power executor for the STM32L5 series. I tested this with a NUCLEO-L552ZE-Q board.

This PR depends on this PR in stm32-data: embassy-rs/stm32-data#325


Dirbaio (approved):
thank you!

PR #2369, opened 2023-12-28
dbgmcu: set bits to false when disabled

chrenderle:
Apparently the dbg_stop bits etc keep the value true over a reset. This means when it is set to true once you have to power off the mcu to reset it to false. I was able to reproduce this on my NUCLEO-L552ZE-Q.

With this PR when setting config.enable_debug_during_sleep to false it sets the dbg_stop bits etc. to false.


Dirbaio (approved):
thanks!

PR #2374, opened 2023-12-29
Reset rx_started state of nrf buffered_uarte on init

ftilde:
This was likely forgotten as part of c46418f12. Without this, when creating a uarte instance, dropping it and then creating another instance, this instance would never receive any bytes.

Please note that I've not yet understood the whole uarte interrupt machinery in it's entirety, but only made this change based on "pattern matching" and the observation that `rx_started` is not reset otherwise. I have observed tested that this change makes the "enable -> disable -> enable" use case work again after c46418f12 in my project, though. Maybe @Dirbaio as the author of c46418f12 can review this?


Dirbaio (approved):
thank you!

PR #2380, opened 2023-12-31
Extend the task macro to allow cfging arguments away

### LINKED ISSUE #2378: embassy task: cannot cfg away task arguments
```rust
#[embassy_executor::task]
pub async fn test(
    a: u32,
    b: u32,
    #[cfg(feature = "c")] c: u32,
) {

}
```

Fails with `cannot find value c in this scope` when the feature `c` is not enabled. It seems argument `cfg`'s are not applied during the transformation.

MabezDev:
Closes #2378

### DIFF
--- a/embassy-executor-macros/src/macros/task.rs
+++ b/embassy-executor-macros/src/macros/task.rs
@@ -49,7 +49,7 @@ pub fn run(args: &[NestedMeta], f: syn::ItemFn) -> Result<TokenStream, TokenStre
         },
     }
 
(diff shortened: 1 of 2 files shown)


Dirbaio (commented):
LGTM! 

Could you add a test? You can use `#[cfg(any())]` to simulate a feature that's off and `#[cfg(all())]` to simulate one that's on.

Dirbaio (approved):
thank you!

PR #2387, opened 2024-01-02
low-power: add stop support for stm32l5

chrenderle:
This PR adds support for the low-power modes STOP1 and STOP2. 

In this image enable_debug_during_sleep is true (by default) and the power-consumption is around 700uA. This is already better than normal because of the sleep mode.
l5-debug-true

In this image enable_debug_during_sleep is set to false and the power-consumption is much lower (170uA). The blinky example has a pretty high interval (300ms). With a lower interval I tested a power-consumption of ~12uA (no wakeup in this measurement). With an interval of 1s I measured around 50uA (active led uses 60uA).
l5-debug-false


Dirbaio:
LGTM! Can you fix CI?

Dirbaio (approved):
thank you!

PR #2388, opened 2024-01-03
Add Missing Clocks on G0 to `Clocks`

AdinAck:
Many clocks in the G0 were not listed nor configurable, this resulted in the PWM frequency calculation being wrong, amongst other things.


Dirbaio (approved):
thank you!

PR #2390, opened 2024-01-03
Add High Frequency (128MHz) Timer Example For G0

AdinAck:
Add an example for using the high frequency timer feature with the now functioning clock controls.


Dirbaio:
thanks!

Issue #1890, opened 2023-09-12
ReadReady and WriteReady support


haijd:
Please add support for the `ReadReady` and `WriteReady` traits from `embedded-io-async` to the `TcpSocket` in `embassy-net`.

The documentation for `ReadReady` can be found at <https://docs.rs/embedded-io-async/0.5.0/embedded_io_async/trait.ReadReady.html>.

The documentation for `WriteReady` can be found at <https://docs.rs/embedded-io-async/0.5.0/embedded_io_async/trait.WriteReady.html>.

PR #2394, opened 2024-01-03
impl ReadReady and WriteReady for tcp

swanandx:
ref: #1890 

This PR aims to implement `embedded_io_async::ReadReady` and `embedded_io_async::WriteReady` traits for `TcpSocket` as well as `TcpReader`/`TcpWriter`

I saw there are already existing fns ( `may_recv()` and `may_send()` ) which are basically the same, so I just wrapped them in trait impl ( like it is done for other traits )

Though, traits expects to return an `Error` ( whenever there is io error ), but implementation of `may_recv()` or `may_send()` return a `bool` instead of `Result<>`! please let me know if something else was expected here!

Thank you :smile:

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -547,6 +547,12 @@ mod embedded_io_impls {
         }
     }
 
+    impl<'d> embedded_io_async::ReadReady for TcpSocket<'d> {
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thank you!

swanandx:
Thank you so much for such quick response :rocket: looking forward to contribute more to embassy :)

PR #2396, opened 2024-01-03
Update STM32 RCC U5 to support P and Q dividers

tyler-gilbert:
This MR adds support for setting the P and Q dividers in the STM32 U5 RCC. This is used when setting the SAI clock.

### DIFF
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -45,6 +45,18 @@ pub struct PllConfig {
     /// The multiplied clock – `source` divided by `m` times `n` – must be between 128 and 544
     /// MHz. The upper limit may be lower depending on the `Config { voltage_range }`.
     pub n: Plln,
+    /// The divider for the P output.
+    ///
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/rcc/u5.rs
```diff
@@ -45,6 +45,18 @@ pub struct PllConfig {
     /// The multiplied clock – `source` divided by `m` times `n` – must be between 128 and 544
     /// MHz. The upper limit may be lower depending on the `Config { voltage_range }`.
     pub n: Plln,
+    /// The divider for the P output.
+    ///
+    /// When used to drive the system clock, `source` divided by `m` times `n` divided by `r`
```
Dirbaio, line 50: this comment is not relevant for p/q, only r can drive the system clock.

tyler-gilbert: Sounds good. I will update.

tyler-gilbert: Updated the comments.

Dirbaio (approved):

PR #2397, opened 2024-01-03
Add write_immediate() function to STM32 DMA ringbufer API

tyler-gilbert:
Add write_immediate() function to STM32 DMA ringbuffer API to pre-fill the buffer before starting the DMA.


lulf (approved):
LGTM

PR #2399, opened 2024-01-03
add async stop() function to STM32 dma ringbuffers

tyler-gilbert:
This MR adds an async stop() function to the STM32 DMA ringbuffers. This allows the users of the ring buffers to async await while the hardware writes all of the samples or until the hardware reads the entire buffer.


tyler-gilbert:
@Dirbaio Does this one look OK? Any other reviewers you could suggest?

lulf (approved):
Looks good to me, could you rebase on main?

tyler-gilbert:
This did not rebase like I wanted. It is now a simpler change. I will close this MR and create a new one.

PR #2401, opened 2024-01-04
New embassy-net release 0.3

MabezDev:
Bumping to 0.3 because the removal of the nightly feature is a breaking change. Also removed some left over cfg's.


Dirbaio:
can you update changelog? we're no longer using smoltcp git, there's also #1890

PR #2402, opened 2024-01-04
Expose rx_delay in nrf qspi config

ftilde:
Being able to modify this is important especially when running the interface at higher frequencies. In my case stuff got glitchy at a frequency of 32MHz which makes sense considering the default rx_delay is 2/64Mhz = 1/32MHz.

Regarding the implementation I was not sure if it would be better to add an additional enum for the config value (since valid values are between 0 and 7) or if this would be overkill. So far I went with an u8 and documented the valid value range in a comment, but we can of course change that.

### DIFF
--- a/embassy-nrf/src/qspi.rs
+++ b/embassy-nrf/src/qspi.rs
@@ -80,6 +80,8 @@ pub struct Config {
     pub frequency: Frequency,
     /// Value is specified in number of 16 MHz periods (62.5 ns)
     pub sck_delay: u8,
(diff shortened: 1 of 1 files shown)


Dirbaio:
u8 sounds good to me. can you rustfmt?

ftilde:
Hm, weird. I have my editor configured to run rustfmt on save, but it seems like rustfmt does not pick up the edition specified in the Cargo.toml and thus fails to format files with async fns.

Anyway, I've run it manually now and updated the proposed commit.

Dirbaio (approved):
thanks!

PR #2403, opened 2024-01-04
Fix an error in compiling examples for rp2040

itdiuna:


Dirbaio:
i'm not sure how can this cause an error, but upgrading a dep is always nice. thanks! :)

itdiuna:
Apologize I haven't provided any evidence. Trying to rebuild the project without this change and reproduce the issue.
It used to use 0.6.0 version of embedded-io which was not compatible with its usage by embedded-io-async (trait was not public).

Once I manage to reproduce it I'll provide the proof.

PR #2404, opened 2024-01-05
tests: Use unified link_ram_cortex_m.x linker script

plaes:
Firstly, sync link_ram.x file with upstream (only tested with nrf52840 target) and then switch all Cortex-M targets to use single file instead of 3 copies.


Dirbaio:
bender run

tests/link_ram_cortex_m.x
```diff
@@ -125,14 +133,14 @@ SECTIONS
...
     __sdata = .;
-    __edata = .;
     *(.data .data.*);
     . = ALIGN(4); /* 4-byte align the end (VMA) of this section */
   } > RAM
   /* Allow sections from user `memory.x` injected using `INSERT AFTER .data` to
    * use the .data loading mechanism by pushing __edata. Note: do not change
    * output region or load region in those user sections! */
   . = ALIGN(4);
+  __edata = .;
```
Dirbaio, line 143: this change was intended. Normally starting contents of .data are in flash and cortex-m-rt copies them from flash to ram on boot. It uses __sdata, __edata, __sidata to know what to copy. 

This is actually not needed when running from RAM, because .data is already in RAM. By setting __sdata=__edata we make cortex-m-rt copy 0 bytes without having to patch it.

Dirbaio (approved):
thanks!

PR #2405, opened 2024-01-05
stm32: Add G0 USB RCC and example

Sizurka:
This wires up the USB clock in STM32G0 RCC, with that done USB can be used on the models that support it.  It also adds a version of USB-serial example using the CRS.

Tested on a custom board with a STM32G0B1CBT6.


Dirbaio (approved):
thank you!

PR #2410, opened 2024-01-06
impl waveform with TIM OC Channel DMA

eZioPan:
Notice:

- Remove TIM6 DMA exclusion from *build.rs*, though a few chip still need extra setup to guaranty TIM DMA actually work, (won't implement until someone actually encounter this problem). 
- Rename `.gen_waveform()` to `.waveform_up()`, to match up with new `.waveform_chx()`

P.S.
I have to copy&paste™ entire "waveform for UP" implement, and do a few modification. It's too tricky to exclude some code with a `$()?` inside a *MacroTranscriber*:
`.waveform_up()` requires user input a `timer::Channel`, but `.waveform_chx()` don't. I don't know how to achieve this inside one macro.

### DIFF
--- a/embassy-stm32/src/timer/simple_pwm.rs
+++ b/embassy-stm32/src/timer/simple_pwm.rs
@@ -155,7 +155,7 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
     ///
     /// Note:  
     /// you will need to provide corresponding TIMx_UP DMA channel to use this method.
-    pub async fn gen_waveform(
+    pub async fn waveform_up(
         &mut self,
         dma: impl Peripheral<P = impl super::UpDma<T>>,
         channel: Channel,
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/timer/simple_pwm.rs
```diff
@@ -221,6 +221,95 @@ impl<'d, T: CaptureCompare16bitInstance> SimplePwm<'d, T> {
...
+                //
+                // optional TODO:
+                // clean FEIF after disable UDE
+                if !original_cc_dma_enabled {
+                    self.inner.set_cc_dma_enable_state(cc_channel, false);
+                }
+
+                if !original_cc_dma_on_update {
+                    self.inner.set_cc_dma_selection(Ccds::ONCOMPARE)
+                }
```
lights0123, line 302: This also would apply to your original PR, but I think this needs special logic to also apply on `Drop`: if one uses e.g. `embassy_time::with_timeout`, the `Transfer` will be canceled as part of `Transfer::drop`, but this will not. Rust's std often creates a simple struct that executes a bunch of code in its drop handler to solve this problem, but you could probably make one that takes a closure.

lights0123:
And what's the point of the macro again? Are there some DMA channels that only work with some timer channels?

eZioPan:
> Are there some DMA channels that only work with some timer channels?

At least for F4 MCU, the answer is yes. TIM seems randomly pick the DMA channel for TIMx_CHy.

embassy-stm32/src/timer/simple_pwm.rs
eZioPan: yeah, but we still need to wait `dma::Transfer` somehow expose a function to clean the flag.

PR #2413, opened 2024-01-07
Add buffer size info to read_packet docs

JomerDev:
Just a tiny addition to the docs after I spent way too long to get `read_packet` to work with a too small buffer


Dirbaio (approved):
thanks!

PR #2414, opened 2024-01-07
Adds function and macro to create usb logger without device (also fixes a logger issue)

JomerDev:
This can be used in cases where the user wants to log something via usb but the usb connection is already in use.
It also serves as an example on how to create multiple usb classes with one device.

Finally this fixes an issue in the usb logger where messages that were exactly MAX_PACKET_SIZE (in this case 64) bytes long wouldn't show up in the log straight away, as explained in the [CdcAcmClass](https://docs.embassy.dev/embassy-usb/git/default/class/cdc_acm/struct.CdcAcmClass.html) (see the third point about messages with exactly the length of MAX_PACKET_SIZE)

### DIFF
--- a/embassy-usb-logger/src/lib.rs
+++ b/embassy-usb-logger/src/lib.rs
@@ -6,7 +6,7 @@ use core::fmt::Write as _;
 
 use embassy_futures::join::join;
 use embassy_sync::pipe::Pipe;
-use embassy_usb::class::cdc_acm::{CdcAcmClass, State};
+use embassy_usb::class::cdc_acm::{CdcAcmClass, Receiver, Sender, State};
 use embassy_usb::driver::Driver;
 use embassy_usb::{Builder, Config};
 use log::{Metadata, Record};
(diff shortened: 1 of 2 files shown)


embassy-usb-logger/src/lib.rs
```diff
@@ -105,6 +105,34 @@ impl<const N: usize> UsbLogger<N> {
             join(run_fut, join(log_fut, discard_fut)).await;
         }
     }
+
+    // Creates the futures needed for the logger from a given class
+    pub async fn create_future_from_class<'d, D>(&'d self, class: CdcAcmClass<'d, D> )
```
JomerDev, line 114: I tried to use `create_future_from_class` inside `run` to not duplicate the code, but couldn't make it work since neither the class nor sender and receiver can be copied. If somebody finds a good way to clean this up, feel free to let me know

embassy-usb-logger/src/lib.rs
```diff
@@ -105,6 +108,38 @@ impl<const N: usize> UsbLogger<N> {
...
+            };
+            let discard_fut = async {
+                let mut discard_buf: [u8; MAX_PACKET_SIZE as usize] = [0; MAX_PACKET_SIZE as usize];
+                receiver.wait_connection().await;
+                loop {
+                    let _ = receiver.read_packet(&mut discard_buf).await;
+                }
+            };
+            join(log_fut, discard_fut).await;
+        }
```
lulf, line 141: What if you make this an internal fn that takes a &mut sender and &mut receiver in order to reuse code?

JomerDev: Great idea! Done

lulf (approved):
LGTM, thanks!

PR #2417, opened 2024-01-08
Add example of pin sharing between tasks

vasilNnikolov:
I'm very new to embassy and am enjoying it very much. This PR contains a small example showcasing how to share a pin between two tasks. I took me some time to figure it out so I figure it may be useful to another beginner in the future. Please let me know if the global `Mutex<Option<PeripheralRef<AnyPin>>>` type is the idiomatic way to do it, as well as if something else can be improved.

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -6,6 +6,7 @@
 * xref:layer_by_layer.adoc[Bare metal to async]
 * xref:runtime.adoc[Executor]
 * xref:delaying_a_task.adoc[Delaying a Task]
+* xref:sharing_peripherals.adoc[Sharing peripherals between tasks]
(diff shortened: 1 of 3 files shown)


plaes:
Ideally this should be as a section in [Embassy Documentation](https://embassy.dev/book/dev/index.html) as well.

vasilNnikolov:
Completely agree, I will look into adding a reference to this example in the docs.

lulf (approved):
Great docs, I really like how you explain the mutex type. I think it would also be good to consider other synchronization primitives like Channels, Signals etc. in the docs and their tradeoffs, because mutex is not always the best solution. But it has to start somewhere and thank you for doing the work.

vasilNnikolov:
Thanks for the review and merge❤️

PR #2420, opened 2024-01-09
Extend mock time driver to support alarms

### LINKED ISSUE #2412: Extend mock time driver to support alarms
I'm trying to unit test some async functions which make use of Timer and Ticker. Would you accept a contribution which adds alarm support to the mock driver? Or would you recommend a different approach?

chrisprice:
Fixes #2412.

I think the most contentious change is utilising `MockDriver` in the tests inside `queue_generic.rs`, but it seemed a shame not to.


chrisprice:
Interestingly I don't see the problem fixed by 3db8655e25bf40d0f0b72e557a8d6e3d156119c8 that caused CI to fail locally. I've put that down to running on an M1 mac as it worked on another x86 box.

An alternative fix would be to disabled the registration of the driver using `#[cfg(not(test))]`.

lulf (approved):

PR #2422, opened 2024-01-10
Cleanup docs and example

lulf:


examples/rp/.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "probe-rs run --chip RP2040"
+runner = "probe-rs run --chip RP2040 --probe 1209:4853:0e0039001450563641333620"
```
plaes, line 2: @lulf  You sure about this?

lulf: Oops, thank you!

PR #2436, opened 2024-01-11
Embassy nrf docs and nrf91 example

lulf:
It occurred to me we don't have nrf91 example. The nrf91 example assumes the SPM module (for instance installed using https://github.com/tweedegolf/nrf9160-rust-starter/). Not sure where to best explain that?


Dirbaio:
i'm not sure if using an SPM should be best practice. It's extra complexity, and wastes a ton of flash space :( 

The tweedegolf repo uses NS mode because nrf9160-hal supports only NS mode, but `embassy-nrf` supports both S and NS mode equally well.

The only other reason to use NS mode is nordic's modem blob is written for NS mode. However I've confirmed the hardware does allow talking to the the modem from S mode if you [set IPC and the shared RAM to non-secure](https://github.com/Dirbaio/nrf91-rust/blob/master/src/main.rs#L52-L66). I *think* also the blob should work from S mode in these conditions, but I haven't verified it.

lulf:
Ok, so S mode should work for blinky at least. I can change to that :+1:

Dirbaio:
canyou remove unused deps, and add the example to ci.sh?

Dirbaio (approved):
thanks <3

PR #2442, opened 2024-01-12
stm32/adc: Allow Adc::read to take !Sized arg

boblehest:
tl;dr: The default bound of `Sized` is unnecessarily strict, and creates an obstacle for me.

My use case is that I want to loop over many ADC pins and read them, so I want to put all my pins in an array:
```rs
let mut pins : [?; 2] = [
    &mut p.PC0,
    &mut p.PC1,
];
```
and here I want some type which I can "upcast" all my pins to (`?` in the snippet above). In certain other interfaces in Embassy you can just use `AnyPin` if you want a "type erased" pin, but in the case of `Adc::read` we need to pass a parameter which implements `AdcPin<T>`, so I ended up with this approach (simplified somewhat from my real code):
```rs
#[embassy_executor::main]
    let mut p = embassy_stm32::init(Default::default());

    let mut detectors : [&mut dyn AdcPin<ADC2>; 2] = [
       &mut p.PC0,
       &mut p.PC1,
    ];

    let mut adc = Adc::new(p.ADC2, &mut Delay);

    loop {
        for d in &mut detectors {
            let value = adc.read(*d);
            dbg!(value);
        }
    }
}
```
But this also does not compile, because `Adc::read` expects the pin parameter to be a sized type, but my trait objects are of course not sized. Hence I'd like to relax this bound to make this compile.

There might be more elegant solutions to this problem; any suggestions are welcome.

### DIFF
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -136,7 +136,7 @@ impl<'d, T: Instance> Adc<'d, T> {
         T::regs().dr().read().0 as u16
     }
 
-    pub async fn read(&mut self, pin: &mut impl AdcPin<T>) -> u16 {
+    pub async fn read(&mut self, pin: &mut (impl AdcPin<T> + ?Sized)) -> u16 {
         Self::set_channel_sample_time(pin.channel(), self.sample_time);
         T::regs().cr1().modify(|reg| {
             reg.set_scan(false);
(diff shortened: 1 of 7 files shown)


Dirbaio:
I think a more elegant solution would be to add `AnyAdcPin` (or maybe `AnyAdcChannel` or similar, since it works for the other ADC channels that are not pins). It'd make the code smaller too, since it'd avoid the overhead of the `dyn` vtable.

boblehest:
> I think a more elegant solution would be to add `AnyAdcPin` (or maybe `AnyAdcChannel` or similar, since it works for the other ADC channels that are not pins). It'd make the code smaller too, since it'd avoid the overhead of the `dyn` vtable.

Thank you for the fast reply! I agree that your suggestion is preferable -- I might attempt implementing it once I've familiarized myself with the code base.

I'm closing this PR on the assumption that there will be no need for this use case then.

PR #2446, opened 2024-01-13
Implement embedded_hal_02 traits for SpiDeviceWithConfig.

bdolgov:
There are still many libraries depending on embedded-hal 0.2 types (for example, [display-interface-spi](https://docs.rs/display-interface-spi/latest/display_interface_spi/)). Let's allow using `SpiDeviceWithConfig` with them.

### DIFF
--- a/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
+++ b/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
@@ -199,3 +199,46 @@ where
         })
     }
 }
+
+impl<'d, M, BUS, CS, BusErr, CsErr> embedded_hal_02::blocking::spi::Transfer<u8> for SpiDeviceWithConfig<'_, M, BUS, CS>
+where
+    M: RawMutex,
+    BUS: embedded_hal_02::blocking::spi::Transfer<u8, Error = BusErr> + SetConfig,
+    CS: OutputPin<Error = CsErr>,
+{
+    type Error = SpiDeviceError<BusErr, CsErr>;
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Self::Error> {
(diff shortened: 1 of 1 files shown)


Dirbaio:
the embedded-hal 0.2 traits are equivalent to the 1.0 SpiBus trait, ie a bare bus with no CS pin. See [here](https://docs.rs/embedded-hal/latest/embedded_hal/spi/index.html) for docs on the difference between SpiDevice and SpiBus.

Due to this, drivers already manage CS on their own. If you implement the eh0.2 traits for SpiDeviceWithConfig, you'll have *two* CS pins, the one managed by the driver and the one managed by SpiDeviceWithConfig. This would be incorrect.

The HAL crates (embassy-stm32, embassy-nrf, embassy-rp) already implement the embedded-hal 0.2 traits for SPI, you can use e-h0.2 drivers through that. 

That does mean you can't share the SPI bus with other things, but there's no way to do sharing with the 0.2 traits. Sharing needs CS to be managed by the impl, not by the driver. That's why the SpiDevice trait was added in 1.0.

bdolgov:
I see, thank you for the explanation. It is somewhat confusing that `SpiDevice` (without config) still [implements](https://github.com/embassy-rs/embassy/blob/686069b4c97b201e4fd444785d784b9836ee2271/embassy-embedded-hal/src/shared_bus/blocking/spi.rs#L113) `spi::Write`.

I guess it's more clean to update my libraries to use the 1.0 traits, so I'll close this PR.

Dirbaio:
> It is somewhat confusing that SpiDevice (without config) still [implements](https://github.com/embassy-rs/embassy/blob/686069b4c97b201e4fd444785d784b9836ee2271/embassy-embedded-hal/src/shared_bus/blocking/spi.rs#L113) spi::Write.

oh indeed, it shouldn't!

PR #2457, opened 2024-01-17
stm32-wpan: Pin dep to work around regression

jamesmunns:


jamesmunns:
Working around https://github.com/OueslatiGhaith/stm32wb-hci/issues/8

PR #2461, opened 2024-01-19
Add more fields to the BssInfo packet struct

raymanfx:
Taken from the Infineon WHD repository:
https://github.com/Infineon/wifi-host-driver/blob/04ee318cc96bffa7d69a1e076c008e2364453f82/WiFi_Host_Driver/inc/whd_types.h#L814

FYI I seem to be getting sound values for RSSI and noise, but the SNR field is always zero for some reason.


Dirbaio (approved):
thanks!

PR #2465, opened 2024-01-19
Implement MII interface with feature flag

sgasse:
The eth/v2 module of embassy-stm32 currently only supports RMII (reduced media independent interface) interfaces with 9 pins. However some chips only/better support a MII interface with 14 pins. With this PR, I want to propose adding support for MII interfaces to `embassy`.

Credits to figuring this out go to my colleague @KloolK, who sat down last year and implemented the necessary changes in a commit [here](https://github.com/embassy-rs/embassy/commit/49c356ab413f424d7cc14e281fe814a373a3d2fa) for an older version of `embassy`.

This PR is meant as discussion basis. Introducing MII as a feature was the least noisy to show the necessary changes. However in the current setup, the features are not additive. Having separate structs for MII and RMII may be a better solution.


Dirbaio:
very nice!

instead of using Cargo features, could you make it into separate `new_mii`, `new_rmii` constructors? This makes it consistent with how other drivers work, e.g. uart has `new`, `new_with_rtscts` etc, FMC has separate constructors for the amount of bits/banks.

embassy-stm32/src/eth/v2/mod.rs
```diff
@@ -239,6 +187,142 @@ impl<'d, T: Instance, P: PHY> Ethernet<'d, T, P> {
     }
 }
 
+/// Create a new MII Ethernet driver using 14 pins.
+pub fn new_mii<'d, T: Instance, P: PHY, const TX: usize, const RX: usize>(
```
Dirbaio, line 191: please keep these as Ethernet methods, not freestanding functions.

embassy-stm32/src/eth/v2/mod.rs
```diff
@@ -35,11 +35,11 @@ impl interrupt::typelevel::Handler<interrupt::typelevel::ETH> for InterruptHandl
 }
 
 /// Ethernet driver.
-pub struct Ethernet<'d, T: Instance, P: PHY> {
+pub struct Ethernet<'d, T: Instance, P: PHY, const NUM_PINS: usize> {
```
Dirbaio, line 38: we should strongly avoid adding generic params, even if it requires doing things at runtime. 

There's a few options. Use `heapless::Vec` with a max capacity of 14, or do an enum:

```rust
enum Pins<'d> {
   Rmii([PeripheralRef<'d, AnyPin>; 9]),
   Mii([PeripheralRef<'d, AnyPin>; 14]),
}

Comment (sgasse) [reply embassy-stm32/src/eth/v2/mod.rs:38]: Thank you, I really like the idea with the enum :slightly_smiling_face:

File: embassy-stm32/src/eth/v2/mod.rs (hunk 1 above)
Comment (sgasse) [reply embassy-stm32/src/eth/v2/mod.rs:191]: Sure, without the const generic parameter, I will move them back.

**Comment by Dirbaio** [comment]
code looks good, can you fix CI?

> What is the rcc_h5 config about? I cannot find it in the Cargo.toml of embassy-stm32. There is one part of the code guarded by #[cfg(rcc_h5)], but for my board, it takes the `#[cfg(not(rcc_h5))] path so I cannot even test if the RMII setting is fine there.

they come from here https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/build.rs#L54

in particular rcc_h5 is active on stm32h5 chips.

**Comment by sgasse** [comment]
Thanks for the info @Dirbaio ! I'd like to fix the CI, but I have a hard time reproducing the issue locally. Which steps are necessary to run `ci.sh` locally? I could not find `cargo batch` with either `cargo install cargo-batch`, `cargo search cargo-batch` or on Google. Running the commands in isolation asks for a nightly compiler due to the `out-dir` parameter, so feel like there is too much difference between what CI does and how I attempt to reproduce it.

As far as I can tell, the issue in CI stems from the fact that I changed the public API of `Ethernet::new` only in v2 to the two functions `Ethernet::new_rmii` and `Ehternet::new_mii`. Unfortunately, I could not find a combination of cfg directives to make CI happy and without being able to reproduce it locally, I do not want to waste CI minutes by pushing again and again.

[Here](https://github.com/embassy-rs/embassy/pull/2465/files#diff-73726863d95e33dc7daf87c4494604a331655d33c68f3545f7fc6701c074d4a2R4) in the ethernet module, we conditionally include either the v1 or v2 version of the ethernet module implementation. In the test, I tried a few combinations and ultimately settled on this:

```rust
#[cfg(not(eth_v2))]
let device = Ethernet::new(
    ...
);

#[cfg(eth_v2)]
let device = Ethernet::new_rmii(
    ...
);
```

From the error in CI, it looks like the test code tries to use the v1 function `Ethernet::new` where the module only provides the v2 function `Ethernet::new_rmii`.:

```
error[E0599]: no function or associated item named `new` found for struct `embassy_stm32::eth::Ethernet` in the current scope
   --> /ci/code/tests/stm32/src/bin/eth.rs:75:28
    |
75  |     let device = Ethernet::new(
    |                            ^^^ function or associated item not found in `Ethernet<'_, _, _>`

...

error: could not compile `embassy-stm32-tests` (bin "eth") due to previous error
     features: ["chrono", "eth", "rng", "stm32h563zi"]
```

I imagine the problem is that the flags passed to build the test binary are different from the flags passed to build the `embassy-stm32` library. However I did not find any mention of `eth_v2` in the `build.rs`, `Cargo.toml` or anywhere outside of the one usage in `embassy-stm32/src/eth/mod.rs`, thus I am not sure for which test case I should add them (if that is at all the right approach). Do you have a hint?

Dirbaio:
> Which steps are necessary to run ci.sh locally?

- `cargo install --git https://github.com/embassy-rs/cargo-batch`
- `./ci.sh`

> Unfortunately, I could not find a combination of cfg directives to make CI happy and without being able to reproduce it locally, I do not want to waste CI minutes by pushing again and again.

CI tells you at the end

```
error: could not compile `embassy-stm32-tests` (bin "eth") due to previous error
     features: ["chrono", "eth", "rng", "stm32h563zi"]
```

so to reproduce

```
cd tests/stm32
cargo build --release --target thumbv7em-none-eabi --features stm32h563zi
```

Dirbaio:
ah yes the issue is the `eth_v2` cfgs are only available within the `embassy-stm32` crate. so it'll always pick the `not(eth_v2)` one.

This reveals an inconsistency between v1 and v2 though! and consistency between versions matters. Maybe we should name v1 `new_rmii` too? or perhaps name the functions in v2 `new` and `new_mii`. since RMII seems way more widely used.

Dirbaio:
> or perhaps name the functions in v2 new and new_mii

ended doing this. Also fixed some fun issue with PC2_C pins in stm32-data...

thanks for the PR!

PR #2468, opened 2024-01-20
add pio_i2s example for RP2040

exoticorn:
This is an example for generating audio in real-time and playing it back through an i2s DAC.

Note that the double buffered dma transfer assumes that `pio::StateMachineTx::dma_push` does immediately start the dma transfer and that polling the returned future is only required to wait for the transfer to finish. I hope this is a safe assumption.


Dirbaio (approved):
thank you!

PR #2469, opened 2024-01-20
feat: add basic support for nRF51 chips to embassy-nrf

lulf:
TODO:

* [ ] HIL test
* [ ] Docs update
* [ ] More peripherals (can be added on demand later I would say)

@Dirbaio would you be able to hook your nrf51-dk into the HIL setup? My own dk is fried, but I verified code is running on my microbit v1.


examples/nrf51/.cargo/config.toml
```diff
@@ -0,0 +1,9 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+# replace nRF82840_xxAA with your chip as listed in `probe-rs chip list`
```
plaes, line 2: Not sure whether nrf52840 makes sense here?

Dirbaio:
I think we could (should?) avoid adding the dependency on portable-atomic, 
- RNG could use a `Mutex<RefCell<State>>` (and change the AtomicWaker to a WakerRegistration). It'd be faster because if you modify multiple fields it'll acquire a single CS, not one for each field.
- The time driver already does a CS for `next_period` so we can do "read, modify, write" manually which needs no CAS.

Dirbaio:
added nrf51-dk to HIL rig. Wire connected between P0.13 and P0.14 for GPIO etc testing.

lulf:
@Dirbaio Thanks! I'll add a HIL test as well as addressing the comments soon!

lulf:
bender run

lulf:
@Dirbaio Two issues remaining:

* Running from RAM doesn't work, it looks related to the size and the available RAM because a simple test that doesn't do gpio or timer runs fine from RAM. 
* I've set it to run from flash, which works locally, but fails with `ERROR - Run failed: Flash content verification failed.` I'm not entirely sure why this is only happening on CI.

Dirbaio:
> Running from RAM doesn't work, it looks related to the size and the available RAM because a simple test that doesn't do gpio or timer runs fine from RAM.

that's fine. we already have some stm32's that do HIL from flash because not enough RAM too...

> I've set it to run from flash, which works locally, but fails with ERROR - Run failed: Flash content verification failed. I'm not entirely sure why this is only happening on CI.

Fixed in https://github.com/probe-rs/probe-rs/pull/2119

bender run

Dirbaio:
duh of course it doesn't work if I don't put back the nrf51-dk in the HL rig.

bender run

Dirbaio (approved):
thank you! :partying_face:

PR #2471, opened 2024-01-22
gpio: remove generics.

Dirbaio:


Meigs2:
I know this is old, but great change!

Issue #2447, opened 2024-01-14
FDCAN (CAN-FD) peripheral support on STM32 (G0, G4, H5, L5, U5 and H7)


blaa:
Many STM32 devices with "generic" CAN2.0 peripheral are currently supported via the bxcan library. The ones listed in the title have a different peripheral. Additionaly the H7 is a bit different flavor because why not. Those FDCAN peripherals are currently not supported.

There's plenty of semi-coordinated progress on making the FDCAN embassy driver and many partial not-merged solutions. I've created this issue to summarize all information I could gather.

1. There's a DRAFT pull-request focusing on H7: https://github.com/embassy-rs/embassy/pull/1692/files last changes 07.2023.
2. There's some work by @tcbennun  for other non-H7 devices from 21.11.2023:
   https://github.com/MaxiluxSystems/embassy/blob/feature/fdcan/embassy-stm32/src/can/fdcan.rs
   https://github.com/tcbennun/embassy-fdcan-mvp/blob/master/src/main.rs                      
   Which was a basis for some other forks.
3. TheEppcJR from 08.12.2023 made some corrections to tcbennun. Commit descriptions aren't really certain about the change though. ;)
   https://github.com/TheEppicJR/embassy-fdcan-mvp/blob/h750-can-test/Cargo.toml
   https://github.com/TheEppicJR/embassy/commits/fdcan-exports/                 
4. There's work by Adam Morgan based on tcbennun change from 21.12.2023, which mentions G4:
https://github.com/lucimobility/embassy/tree/feature/rebased-fdcan

So my current understanding is to go with the (1) if using H7 or probably (4) if using the rest of devices. But I can't be sure. In near future I'll try (4) with stm32g431 and see how it goes.

blaa:
I gathered changes from (2) and (4) on new branch to create something that might be mergeable - no debug prints, cleanly split commits, no commented code. Moved the example from FDCAN2/3 to FDCAN1. It compiles example for stm32g4, I haven't yet tried it on hardware. 

Lacking:
- stm32h7 example and maybe some changes (rcc?). Maybe from (1).
- Docs; there's plenty of warnings about missing docs. I can probably fill those if this approach makes sense.

https://github.com/blaa/embassy/commits/feature/fdcan/

Two not included parts (were commented) are:
```
+    /*  
+    pub async fn flush(&self, mb: bxcan::Mailbox) { 
+        poll_fn(|cx| { 
[...]

PR #2475, opened 2024-01-24
Support for FDCAN peripheral as found on newer STM32 micros.

cschuhen:
Discussion tracked in:
https://github.com/embassy-rs/embassy/issues/2447

This has been a group effort.


blaa:
Commit reorganization I'd consider:
- merging stm32h5 and h7 examples to a one commit. 
- Merge changes from `formatting` to documentation or original commits. 
- Merge the `revert example chip` back to `Example working on real 250K bus with STM32G431CB.` so there's no back-and-forth changes.

blaa:
Question I was asking on #Embassy was whether that should be behind a `unstable-fdcan` feature or if the embassy is unstable enough to not bother with it. I believe the API might change, but it's not particularly extensive.

embassy-stm32/Cargo.toml
```diff
@@ -693,373 +695,373 @@ stm32f779ai = [ "stm32-metapac/stm32f779ai" ]
...
+stm32g491rc = [ "stm32-metapac/stm32g491rc", "fdcan/fdcan_g0_g4_l5" ]
+stm32g491re = [ "stm32-metapac/stm32g491re", "fdcan/fdcan_g0_g4_l5" ]
+stm32g491vc = [ "stm32-metapac/stm32g491vc", "fdcan/fdcan_g0_g4_l5" ]
+stm32g491ve = [ "stm32-metapac/stm32g491ve", "fdcan/fdcan_g0_g4_l5" ]
+stm32g4a1ce = [ "stm32-metapac/stm32g4a1ce", "fdcan/fdcan_g0_g4_l5" ]
+stm32g4a1ke = [ "stm32-metapac/stm32g4a1ke", "fdcan/fdcan_g0_g4_l5" ]
+stm32g4a1me = [ "stm32-metapac/stm32g4a1me", "fdcan/fdcan_g0_g4_l5" ]
+stm32g4a1re = [ "stm32-metapac/stm32g4a1re", "fdcan/fdcan_g0_g4_l5" ]
+stm32g4a1ve = [ "stm32-metapac/stm32g4a1ve", "fdcan/fdcan_g0_g4_l5" ]
+stm32h503cb = [ "stm32-metapac/stm32h503cb", "fdcan/fdcan_g0_g4_l5" ]
```
kevswims, line 895: Should this feature be renamed to just `fdcan/fdcan` or something more generic since it also applies to h5 and u5 series?

blaa: in embassy-stm32 fdcan.rs differentiates H7 and !H7 but not by feature (?) - by chip. H5 being in the second group. 

Those above AFAIK relate to selecting a proper feature in fdcan crate that is beyond our control. The problem is obviously that this needs to be selected per chip. But there are differences between those two hw versions.

cschuhen:
> Commit reorganization I'd consider:
> 
> * merging stm32h5 and h7 examples to a one commit.
> * Merge changes from `formatting` to documentation or original commits.
> * Merge the `revert example chip` back to `Example working on real 250K bus with STM32G431CB.` so there's no back-and-forth changes.

Yes, I was thinking the same. I just ran out of time before work and needed to read up on how to split commits in rebase. I usually use mercurial rather than git.

tests/stm32/src/bin/fdcan.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let peripherals = embassy_stm32::init(config);
+
+    let mut can = can::Fdcan::new(peripherals.FDCAN1, peripherals.PB8, peripherals.PB9, Irqs);
+
+    // 250k bps
+    can.set_bitrate(250_000);
+
+    let mut can = can.into_external_loopback_mode();
+    //let mut can = can.into_normal_mode();
```
blaa, line 30: This comment and the ones from lines 36, 63 can probably by removed. 

I've got my "lab" soldered and I'll shortly play with adding CAN frame filters. Would be cool to add them then to this test as well, but I don't consider this a requirement.

cschuhen: Done. I also figured out that I can add you as a direct collaborator.

tests/stm32/Cargo.toml
```diff
@@ -14,6 +14,7 @@ stm32f429zi = ["embassy-stm32/stm32f429zi", "chrono", "eth", "stop", "can", "not
 stm32f446re = ["embassy-stm32/stm32f446re", "chrono", "stop", "can", "not-gpdma", "dac", "sdmmc"]
 stm32f767zi = ["embassy-stm32/stm32f767zi", "chrono", "not-gpdma", "eth", "rng"]
 stm32g071rb = ["embassy-stm32/stm32g071rb", "cm0", "not-gpdma", "dac"]
+stm32g431cb = ["embassy-stm32/stm32g431cb", "chrono", "stop", "not-gpdma", "rng", "fdcan"]
```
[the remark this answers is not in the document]
Dirbaio: yeah this will do nothing because the HIL test farm doesn't have a stm32g431cb. Please add it to the existing chips instead. If there's any wire I have to add for loopback please let me know.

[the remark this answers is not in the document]
kevswims: Loopback can be configured internally on the part, it will still output signals on the TX pin so this just needs to be set to use a TX pin that is unused or won't interfere with anything else that is connected.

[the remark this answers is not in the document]
Dirbaio: it's better to use external loopback if possible, because it also tests the GPIO AF configuration etc are working correctly. it's "closer" to normal mode.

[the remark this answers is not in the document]
kevswims: With CAN you would need to have transceivers and have two separate instances talk to each other to test a full loopback externally otherwise there is not another node on the bus to ACK the message and it will consider the transmission failed. With internal loopback it ignores the missing ACK.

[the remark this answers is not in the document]
blaa: I believe this could be improved with a CAN frame filtering. Configure a filter or two, send a message - match a filter - and receive it. Then send another NOT matching message and assert that it's not being received. This should cover a lot of CAN API without touching hardware.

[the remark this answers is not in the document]
blaa: On a very short distance it's possible to do a CAN without transceivers AFAIK - using a diode instead. For testing it should be possible.

embassy-stm32/src/can/fdcan.rs
```diff
@@ -1,40 +1,759 @@
...
+
+        T::enable_and_reset();
+
+        rx.set_as_af(rx.af_num(), AFType::Input);
+        tx.set_as_af(tx.af_num(), AFType::OutputPushPull);
+
+        let mut can = fdcan::FdCan::new(FdcanInstance(peri)).into_config_mode();
+
+        unsafe {
+            T::configure_msg_ram();
```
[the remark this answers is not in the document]
cschuhen: configure_msg_ram can be taken out of the unsafe block it seems. Looks like upend can be too but I'm not sure that it's worth it.

embassy-stm32/src/can/fdcan.rs
```diff
@@ -1,40 +1,759 @@
...
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Instance> interrupt::typelevel::Handler<T::IT1Interrupt> for IT1InterruptHandler<T> {
+    unsafe fn on_interrupt() {}
+}
+
+/// Bus error
+#[derive(Debug)]
+pub enum BusError {
```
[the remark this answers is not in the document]
cschuhen: Ok, I shifted this into enums.rs  While I was at it, I also shifted out the two calc bit timing functions into util.rs. What are your thoughts on this?

embassy-stm32/src/can/util.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub prescaler: NonZeroU16,
+    /// Valid values are 1 to 128.
+    pub seg1: NonZeroU8,
+    /// Valid values are 1 to 255.
+    pub seg2: NonZeroU8,
+    /// Valid values are 1 to 128.
+    pub sync_jump_width: NonZeroU8,
+}
+
+/// Calculate nomnial CAN bit timing based on CAN bitrate and periphial clock frequency
```
Dirbaio, line 19:
```suggestion
/// Calculate nominal CAN bit timing based on CAN bitrate and periphial clock frequency
```

embassy-stm32/Cargo.toml
```diff
@@ -693,373 +695,373 @@ stm32f779ai = [ "stm32-metapac/stm32f779ai" ]
...
-stm32h7b0zb = [ "stm32-metapac/stm32h7b0zb" ]
-stm32h7b3ai = [ "stm32-metapac/stm32h7b3ai" ]
-stm32h7b3ii = [ "stm32-metapac/stm32h7b3ii" ]
-stm32h7b3li = [ "stm32-metapac/stm32h7b3li" ]
-stm32h7b3ni = [ "stm32-metapac/stm32h7b3ni" ]
-stm32h7b3qi = [ "stm32-metapac/stm32h7b3qi" ]
-stm32h7b3ri = [ "stm32-metapac/stm32h7b3ri" ]
-stm32h7b3vi = [ "stm32-metapac/stm32h7b3vi" ]
-stm32h7b3zi = [ "stm32-metapac/stm32h7b3zi" ]
+stm32g030c6 = [ "stm32-metapac/stm32g030c6", "fdcan/fdcan_g0_g4_l5" ]
```
Dirbaio, line 698: can you change these to this?

```toml
stm32g030c6 = [ "stm32-metapac/stm32g030c6", "dep:fdcan", "fdcan/fdcan_g0_g4_l5" ]
```

the `dep:fdcan` thing makes it so Cargo doesn't create an implicit `fdcan` feature for the crate, We don't want the users to be able to explicitly enable fdcan when depending on `embassy-stm32`.

Dirbaio:
bender run

embassy-stm32/src/can/fdcan.rs
```diff
@@ -1,40 +1,640 @@
...
+            }
+            if let Ok(err) = LastErrorCode::try_from(lec) {
+                return BusError::try_from(err);
+            }
+        }
+        None
+    }
+}
+impl<'d, T: Instance, M: FdcanOperatingMode> Deref for Fdcan<'d, T, M> {
+    type Target = RefCell<fdcan::FdCan<FdcanInstance<'d, T>, M>>;
```
Dirbaio, line 495: derefing to a RefCell is a bit risky, because it makes too easy to run into borrow errors together with async. For example, rx task borrows the refcell then awaits something else while holding it, then the tx task gets polled and tries to borrow the refcell again which fails.

If you use fdcan's `split_by_ref` you can make FdcanTx/FdcanRx own the Tx/Rx halves without having to use RefCells. Then Fdcan can deref to the full `fdcan::FdCan` (and maybe you want the tx/rx halves to deref to the fdcan Tx/Rx structs maybe?)

Dirbaio:
Looks great overall. and HIL test is passing! :D

It'd be great to extend the HIL test to other chips. Ideally all chips in the HIL test farm that have FDCAN. I can give teleprobe access to anyone interested, so you can run the tests locally without having to wait for CI. Please ping me on Matrix :) 

---

Also, This is more of a general observation than a review comment but: I wonder if long-term it would be better to maintain the bxcan+fdcan drivers in-crate. There's a few negative consequences of using the bxcan/fdcan crates:

- The fact that fdcan needs different Cargo features per chip. It's not nice to not be able to use the stm32-metapac metadata versions to automatically wire it. Also, we should try to keep Cargo.toml small, because every feature dependency bloats the crates.io index, which recently turned out to cause scaling problems, prompting the crates.io team to [set a limit of 300 features](https://blog.rust-lang.org/2023/10/26/broken-badges-and-23k-keywords.html). We have ~1500 features, we've been granted an exception but i'd like not to abuse it.
- The OperatingMode typestates go against the embassy-stm32 api design principles (reduce generics as much as possible), but we can't avoid them without modifying `fdcan`.
- The tx/rx halves require 2 lifetime params, while if we rolled our own we could have just one.
- The bxcan crate seems unmaintained (last commit 2y ago, and jonas-schievink has very publicly exited the Rust scene)

I'm just wondering long term, i'm not saying we should do in-crate CAN drivers now, it'd be a lot of work, i'm OK with these limitations for now, and `fdcan` gets us working fdcan support *now*, which is nice.

embassy-stm32/src/can/fdcan.rs
cschuhen: I'll have a play with this. As you might have seen me mention, I'm pretty new to rust so these type of changes result in me fighting with the compiler quite a bit :) I'll use bxcan as a reference.

cschuhen:
@Dirbaio  Thanks! I appreciate your time reviewing.

Interesting comments on use of fdcan. This does sound like a decent bit of work. My mind is wondering about forking fdcan driver by copy-pasting it into embassy-stm32 and modify from there. Still might mean a fair bit of rework here now to get any reasonable use but do you think that is a reasonable approach? Or fork fdcan crate to make one that works better with stm32-metapac.

cschuhen:
I certainly followed your issues with the feature limit. Really, itas a NOOB looking in my first thought is the feature system is broken(or limited). Shouldn't it allow chip selection to use just one feature CHIP=VALUE? Why do features have to be boolean.

I do wonder about the whole split thing... Longer term, I really think we need a buffered fdcan. Especially for read, dropping RX packets is really bad for most use of CAN so we really need the interrupt routine to putting things into a buffer. Often, this is a requirement for TX too.

Could we have internal RX(backed by Channel or PubSubChannel)  and TX(backed by Channel)  buffers and just hand out senders for TX and receivers for RX? Would this do away with the need for split at all? My biggest issue is specifying the size of buffers, this would either add to features(the way arena size does) or add generic arguments.

embassy-stm32/Cargo.toml
cschuhen: Done...

embassy-stm32/src/can/fdcan.rs
cschuhen: I actually noticed yesterday when trying to add some tests after a split (feature/fdcan_experimental branch) that I could still access the non-split version after the split.

embassy-stm32/Cargo.toml
cschuhen: I now get this error when building examples:
the package `embassy-stm32g4-examples` depends on `embassy-stm32`, with features: `fdcan` but `embassy-stm32` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.


failed to select a version for `embassy-stm32` which could resolve this conflict

cschuhen: I'm going to revert this change because it broke builds and I don't know right now how to fix it.

embassy-stm32/src/can/fdcan.rs
cschuhen: Ok, that was easier than I thought it would be. There was less random trying of mut's and &' before and after :'s than I expected.

cschuhen:
Okay, so I have pushed updates to the PR:

1. Remove RefCell
2. Includes a change reverting the cargo changes that you asked for because example builds didn't work then.
3. Includes these that I held off on including yesterday:
4. a) Include Embassy timestamps
5. b) Read from fifo1 also

embassy-stm32/Cargo.toml
cschuhen: Ok. Turns out it was just the g4. All I had to do is remove the explicit fdcan dep.

kevswims:
@Dirbaio 
> Also, This is more of a general observation than a review comment but: I wonder if long-term it would be better to maintain the bxcan+fdcan drivers in-crate. There's a few negative consequences of using the bxcan/fdcan crates:

Agreed that this isn't a problem to solve today but I really like this idea. The drivers for the CAN peripherals are not really that complex and this would make it easier to have a consistent API across the different hardware.

cschuhen:
> @Dirbaio
> 
> > Also, This is more of a general observation than a review comment but: I wonder if long-term it would be better to maintain the bxcan+fdcan drivers in-crate. There's a few negative consequences of using the bxcan/fdcan crates:
> 
> Agreed that this isn't a problem to solve today but I really like this idea. The drivers for the CAN peripherals are not really that complex and this would make it easier to have a consistent API across the different hardware.

Not to mention that the fdcan crate does not use what it can(excuse pun) from embedded hal like the ID definitions.

blaa:
I tested the code during the weekend and in general - it works. I'd not block the merge, but plan for some changes shortly. I'd assume the API not yet stable. ;) Unsure if it's worth a feature.

Mostly:
1. Get API parity with bxcan (try_write, flush_any, try_read, wait_not_empty)
2. Add API for adding CAN message filters.
3. Probably add some error counters that are available through the API. Buffer Overrun for example could be counted - even with additional dedicated buffer overrun can happen. This is simple enough and uses not much additional RAM.
4.  I found a bug where write().await can block indefinitely: With high contention on BUS with the same CAN ID messages - so a lot of arbitration error. The HW can quit the normal mode and enter BusOff state. Then no further interrupts will push the transmit forward and it will just hang. Below is a patch that could fix that, but we can add it later.

```diff
@@ -154,6 +157,21 @@ impl<T: Instance> interrupt::typelevel::Handler<T::IT0Interrupt> for IT0Interrup
 
         let ir = regs.ir().read();
 
+        if ir.bo() {
+            /* We entered/left busoff state. This would block the awaited TX forever. Force retry. */
+            regs.ir().write(|w| w.set_bo(true));
+
+            let cccr = regs.cccr().read();
+            if cccr.init() {
+                // BusOff - so many TX errors that interface got detached.
+                regs.cccr().write(|w| w.set_init(false));
+            }
+            //return;
+        }

(...)

@@ -323,11 +343,15 @@ impl<'d, T: Instance> Fdcan<'d, T, fdcan::ConfigMode> {
         can.enable_interrupt(fdcan::interrupt::Interrupt::RxFifo0NewMsg);
         can.enable_interrupt(fdcan::interrupt::Interrupt::RxFifo1NewMsg);
         can.enable_interrupt(fdcan::interrupt::Interrupt::TxComplete);
+        can.enable_interrupt(fdcan::interrupt::Interrupt::BusOff);
```

This will cause the HW to wait for 129 Bus Idle states and then retry transmissions. RM0440 entry:
```
The Bus_Off recovery sequence (see CAN Specification Rev. 2.0 or ISO11898-1) cannot be shortened by setting or resetting CCCR[INIT]. If the device goes Bus_Off, it sets CCCR.INIT of its own, stopping all bus activities. Once CCCR[INIT] has been cleared by the CPU, the device then waits for 129 occurrences of Bus Idle (129 × 11 consecutive recessive bits) before resuming normal operation. At the end of the Bus_Off recovery sequence, the Error Management Counters are reset. During the waiting time after the reset of CCCR[INIT], each time a sequence of 11 recessive bits has been monitored, a Bit0 Error code is written to PSR[LEC], enabling the CPU to readily check up whether the CAN bus is stuck at dominant or continuously disturbed and to monitor the Bus_Off recovery sequence. ECR[REC] is used to count these sequences.
```

embassy-stm32/src/rcc/h.rs
```diff
@@ -585,7 +589,9 @@ pub(crate) unsafe fn init(config: Config) {
 
         RCC.ccipr5().modify(|w| {
             w.set_ckpersel(config.per_clock_source);
-            w.set_adcdacsel(config.adc_clock_source)
+            w.set_adcdacsel(config.adc_clock_source);
+            // We can't use PAC method here because H5 and H50 define different name
+            w.0 = (w.0 & !(0x03 << 8usize)) | (((config.fdcan_clock_source.to_bits() as u32) & 0x03) << 8usize);
```
Dirbaio, line 594: i've fixed this in https://github.com/embassy-rs/stm32-data/commit/ab2bc2a739324793656ca1640e1caee2d88df72d and #2495 , if you rebase you can use `fdcan12sel` in all h5 now.

cschuhen: Fantastic. I was planning to research how to make changes in these other crates but it was a bit down on my queue.

cschuhen: I pushed the change to PR to use the new method.

cschuhen:
> I tested the code during the weekend and in general - it works. I'd not block the merge, but plan for some changes shortly. I'd assume the API not yet stable. ;) Unsure if it's worth a feature-gate.
> 
> Mostly:
> 
> 1. Get API parity with bxcan (try_write, flush_any, try_read, wait_not_empty)
> 2. Add API for adding CAN message filters.
> 3. Probably add some error counters that are available through the API. Buffer Overrun for example could be counted - even with additional dedicated buffer overrun can happen. This is simple enough and uses not much additional RAM.
> 4. Maybe splitting TX and RX between two separate interrupts?
> 5. I found a bug where write().await can block indefinitely: With high contention on BUS with the same CAN ID messages - so a lot of arbitration error. The HW can quit the normal mode and enter BusOff state. Then no further interrupts will push the transmit forward and it will just hang. Below is a patch that could fix that, but we can add it later.

These are my thoughts:

1. Yes, they can be added later because they shouldn't be 'breaking changes' we are just adding additional API. Users could probably emulate something similar by creating small reader and writer tasks that are backed by Channel.
2. Can't this already be done using the fdcan create API directly? I kind of touched on some of this in the HIL test. I think to take this a step further we need to analise if we can have a consistent API between fdcan and bxcan.
3. Yes, it would be good to find a way to notify of this too. We could add it to BusError and return one of those when detected. I'm open to other ideas.
4. Yes, no issues with that. It should be an internal change.
5. Yes, we can keep investigating this further as a bug. I think we need to return an error when we get into BusOff state for both readers and writers. If we do a 'restart' like your patch, I think it should be a config option that users can turn off.

cschuhen:
Hi @Dirbaio. @blaa and I have discussed and we believe the patch is ready to merge from our point of view.

Dirbaio (approved):
let's merge then. Thank you everyone who helped make this happen, you're awesome :heart: :rocket:

PR #2476, opened 2024-01-24
Add parameter for enabling pull-up and pull-down in RP PWM input mode

pawel00100:
I created this simple change that allows pin B (the one that can be used as PWM counter input) to be set with pull-up or pull-down internal resistor.

I made tests for two most important cases. I was thinking about adding extra test for checking if the input is floating by default, but it even in case of code being wrong test could on occasion pass so I decided not to add it:
```
    // PWM input is floating by default
    {
        let pin2 = Input::new(&mut p11, Pull::Down);
        let pwm = Pwm::new_input(&mut p.PWM_CH3, &mut p7, Pull::None, InputMode::FallingEdge, cfg.clone());
        Timer::after_millis(1).await;
        assert!(pin2.is_low());
    }
    {
        let pin2 = Input::new(&mut p11, Pull::Up);
        let pwm = Pwm::new_input(&mut p.PWM_CH3, &mut p7, Pull::None, InputMode::FallingEdge, cfg.clone());
        Timer::after_millis(1).await;
        assert!(pin2.is_high());
    }
```
Alternative test for this case could be to use ADC with pullup/down and check voltage levels, in case the voltage was in the middle it would mean that an opposite pullup/down resistor was enabled when it shouldn't, but it would require rewiring test board and would be in my opinion a bit overkill.

Please note that I'm barely experienced in Rust and I'm an amateur in EE/embedded field so I'd advice extra caution reviewing the proposed change.

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -6,7 +6,7 @@ use fixed::FixedU16;
 use pac::pwm::regs::{ChDiv, Intr};
 use pac::pwm::vals::Divmode;
 
-use crate::gpio::{AnyPin, Pin as GpioPin, SealedPin as _};
+use crate::gpio::{AnyPin, Pin as GpioPin, Pull, SealedPin as _};
 use crate::{pac, peripherals, RegExt};
 
 /// The configuration of a PWM slice.
@@ -92,6 +92,7 @@ impl<'d, T: Slice> Pwm<'d, T> {
         inner: impl Peripheral<P = T> + 'd,
         a: Option<PeripheralRef<'d, AnyPin>>,
         b: Option<PeripheralRef<'d, AnyPin>>,
+        b_pull: Pull,
         config: Config,
         divmode: Divmode,
     ) -> Self {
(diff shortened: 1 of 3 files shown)


embassy-rp/src/pwm.rs
```diff
@@ -93,6 +93,7 @@ impl<'d, T: Channel> Pwm<'d, T> {
         inner: impl Peripheral<P = T> + 'd,
         a: Option<PeripheralRef<'d, AnyPin>>,
         b: Option<PeripheralRef<'d, AnyPin>>,
+        b_pull: Pull,
```
Dirbaio, line 96: could you make this a field in `Config`? this way the number of params doesn't explode and it doesn't increase verbosity for users that don't need the pull.

(yes, divmode should've been a field in config too...)

Dirbaio: actually no, it's better to have it separate so it's only present in the `new_` constructots that use inputs, since it doesnt' make sense for outputs. 

(same for `divmode`, that's why it was that way...)

Dirbaio (approved):
thank you!

PR #2480, opened 2024-01-26
Fix the backticks on the getting started page

ssnover:
image

There was a little mistake in the formatting on the Getting Started page of the docs. This is just a quick fix to what I think the intended markup was.


jamesmunns (approved):
Thanks!

PR #2487, opened 2024-01-27
PWM Improvements

Murmele:
1) implement retriving the current duty
2) Do not wait for seqend if the pwm is disabled, because then an infinity loop occurs

I will finish implementing my driver (Driving a Buzzer) to see if I have more improvements


Dirbaio (approved):
thanks!

Murmele:
Thanks for merging

PR #2491, opened 2024-01-29
fix: typo in netcat command for std example

ValouBambou:
The previous given command `nc -l 8000` doesn't let me see anything and lead to a "WARN connect error: ConnectionReset". By explicitly changing the `local-port` of `nc` with the `-p` I can now see the `Hello` message printed, and the warning log disappeared.


Dirbaio (approved):
this is a bsd netcat vs gnu netcat difference, but seems `-lp` works for both so LGTM. Thanks!

PR #2492, opened 2024-01-29
Fix underflow in embassy-rp uartrx.read_to_break

### LINKED ISSUE #2490: panic in embassy-rp's uartrx.read_to_break
I've got a panic in embassy-rp's uartrx.read_to_break.
[Line 608](https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/uart/mod.rs#L608) returns a usize that is calculated through doing `(next_addr - 1) - sval`, however in my case it appears that both `next_addr` and `sval` are the same value meaning that with the minus one we try to return `-1` as usize.
There is a check [a few lines above](https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/uart/mod.rs#L588) which checks if `next_addr` is smaller than `sval` but it doesn't check if they are the same

JomerDev:
This fixes #2490.

In rare cases the line break can be triggered without any data having been read which causes both `next_addr` and `sval` to be the same value. Subtracting 1 and `sval` from `next_addr` causes an underflow panic in debug mode and  a high chance of panic in release mode when the user uses the returned `usize::Max` value to try and index their data

### DIFF
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -604,8 +604,8 @@ impl<'d, T: Instance> UartRx<'d, T, Async> {
             return match (all_full, last_was_break) {
                 (true, true) | (false, _) => {
                     // We got less than the full amount + a break, or the full amount
-                    // and the last byte was a break. Subtract the break off.
-                    Ok((next_addr - 1) - sval)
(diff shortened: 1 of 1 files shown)


jamesmunns (approved):
Thank you!

PR #2501, opened 2024-01-30
[WIP] Added example of an  executor that tracks the idle time.

fakusb:
An example for the rp pico towards https://embassy.dev/book/dev/faq.html#_how_can_i_measure_resource_usage_cpu_ram_etc

1) Does anyone have comments on the example code itself?
2) If I'm not mistaken, the executor would treat time spend during an interrupt (e.g. all the tasks in an InterruptExecutor) as idle time, because the interrupt handler is executed before the non-interupt-level code after the wfe-instruction is executed. This kind of caveat should probably be made even more explicit?

### DIFF
--- a/examples/rp/src/bin/core_usage.rs
+++ b/examples/rp/src/bin/core_usage.rs
@@ -1,0 +1,150 @@
+//! This example shows how to measure the idle time of the CPU when using the simple thread mode executor.
+//! For this, we copied the thread mode executor (embassy-executor/src/arch/cortex_m.rs), and adjusted it to keep track of the idled duration in some static variable.
+//!
+//! Note that the idle time might be larger than the actual idle time since interrupt handlers run before the idle-duration-timer is stopped for each executor-loop.
+//!
+//! As an example, we have one task that idles for ~2.3 seconds, and then blocks for ~1 second.
+//!
+//! Sample output below.
+//!
+//!
+//! ```not_rust
+//! [AWAIT]
+//! In the last 1.000311 sec, core idled for 0.999716 sec (0.9994052%).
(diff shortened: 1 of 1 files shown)


examples/rp/src/bin/core_usage.rs
```diff
@@ -0,0 +1,151 @@
...
+    let mut ticker = Ticker::every(Duration::from_secs(1));
+    let mut now_pretick = Instant::now();
+    let mut total_idle_pretick = idle_duration.get();
+    loop {
+        ticker.next().await;
+        let now_posttick = Instant::now();
+        let total_idle_posttick = idle_duration.get();
+        let delta_t = now_posttick.duration_since(now_pretick);
+        let delta_idle_dur = total_idle_posttick.sub(total_idle_pretick);
+        let idle_percentage = delta_idle_dur.as_ticks() as f32 / delta_t.as_ticks() as f32;
```
Dirbaio, line 140: this is missing a *100

Dirbaio:
> If I'm not mistaken, the executor would treat time spend during an interrupt (e.g. all the tasks in an InterruptExecutor) as idle time, because the interrupt handler is executed before the non-interupt-level code after the wfe-instruction is executed. This kind of caveat should probably be made even more explicit?

there's a way around this: you can do the `wfe` inside a critical section. If an interrupt becomes pending, it still causes the `wfe` to exit, but the interrupt isn't handled yet because it's disabled. You can then count the elapsed time as idle then exit the critical section, so the interrupt time is not counted as idle.

Dirbaio:
i'm marking as draft, please mask as ready to review when it's no longer WIP.

PR #2502, opened 2024-01-30
NRF: handle `uarte` RX errors

### LINKED ISSUE #2500: Support for UART specific errors in `embassy_nrf` `uarte` and `buffered_uarte`
Currently, in `embassy_nrf`, UART errors are not handled. For example, if configuring a `Uarte` with the wrong baud, when trying to read `n` bytes, the future never completes instead of returning a parity or framing error.

By UART errors, I mean:
* Parity errors
* Framing errors
* Buffer overrun errors.

jbeaurivage:
This PR adds a mechanism to handle RX errors for UART RX transactions.

Closes #2500.

### DIFF
--- a/embassy-nrf/src/uarte.rs
+++ b/embassy-nrf/src/uarte.rs
@@ -52,6 +52,39 @@ impl Default for Config {
     }
 }
 
+bitflags::bitflags! {
+    /// Error source flags
+    pub struct ErrorSource: u32 {
+        /// Buffer overrun
+        const OVERRUN = 0x01;
+        /// Parity error
+        const PARITY = 0x02;
+        /// Framing error
+        const FRAMING = 0x04;
+        /// Break condition
+        const BREAK = 0x08;
+    }
+}
+
+impl TryFrom<ErrorSource> for () {
+    type Error = Error;
+
+    #[inline]
+    fn try_from(errors: ErrorSource) -> Result<Self, Self::Error> {
+        if errors.contains(ErrorSource::OVERRUN) {
+            Err(Error::Overrun)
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/uarte.rs
```diff
@@ -52,6 +52,39 @@ impl Default for Config {
...
+        /// Parity error
+        const PARITY = 0x02;
+        /// Framing error
+        const FRAMING = 0x04;
+        /// Break condition
+        const BREAK = 0x08;
+    }
+}
+
+impl TryFrom<ErrorSource> for () {
```
Dirbaio, line 69: this is a bit strange. I'd suggest adding a method `check(self) -> Result<(), Error>` to `ErrorSource` instead.

embassy-nrf/src/uarte.rs
```diff
@@ -601,17 +662,25 @@ impl<'d, T: Instance> UarteRx<'d, T> {
...
 
         trace!("startrx");
         r.tasks_startrx.write(|w| unsafe { w.bits(1) });
 
-        poll_fn(|cx| {
+        let result = poll_fn(|cx| {
             s.endrx_waker.register(cx.waker());
+
+            if let Err(e) = self.check_and_clear_errors() {
+                return Poll::Ready(Err(e));
```
Dirbaio, line 680: This needs to stop the DMA transfer, otherwise DMA will keep writing to the buffer after returning, which is UB.

jbeaurivage:
@Dirbaio some valid points. All fixed now.

Dirbaio (approved):
thanks!

PR #2505, opened 2024-01-31
Add timeout_at convenience function and example.

CBJamo:
Small qol function. I was using with_timeout to do what I'm doing in the example, but with that method the timeouts are additive, this just makes it a little easier to think about.

### DIFF
--- a/embassy-time/src/timer.rs
+++ b/embassy-time/src/timer.rs
@@ -8,7 +8,7 @@ use futures_util::{pin_mut, Stream};
 
 use crate::{Duration, Instant};
 
-/// Error returned by [`with_timeout`] on timeout.
+/// Error returned by [`with_timeout`] and [`timeout_at`] on timeout.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


embassy-time/src/timer.rs
```diff
@@ -26,6 +26,19 @@ pub async fn with_timeout<F: Future>(timeout: Duration, fut: F) -> Result<F::Out
     }
 }
 
+/// Runs a given future with a deadline time.
+///
+/// If the future completes before the deadline, its output is returned. Otherwise, on timeout,
+/// work on the future is stopped (`poll` is no longer called), the future is dropped and `Err(TimeoutError)` is returned.
+pub async fn timeout_at<F: Future>(at: Instant, fut: F) -> Result<F::Output, TimeoutError> {
```
Dirbaio, line 33: i'd call it `with_deadline`, feels more symmetrical to `with_timeout`

CBJamo: Excellent point, done.

Dirbaio (approved):
thank you!

PR #2511, opened 2024-02-01
STM32WBA's high speed external clock has to run at 32 MHz

Ecco:
STM32WBA's HSE speed cannot be configured: it *has* to run at 32 MHz.

I'm not sure the `PllSource` enum is still needed at this point but since this is my first contribution I went for something minimal. Let me know if `PllSource` should just go.

As a side note, because their frequencies are fixed, the datasheet of that chip refers to the high-speed clocks as HSI16 and HSE32. I'd be willing to contribute a change in that direction if it'd be useful.

### DIFF
--- a/embassy-stm32/src/rcc/wba.rs
+++ b/embassy-stm32/src/rcc/wba.rs
@@ -6,26 +6,28 @@ use crate::time::Hertz;
 
 /// HSI speed
 pub const HSI_FREQ: Hertz = Hertz(16_000_000);
+// HSE speed
+pub const HSE_FREQ: Hertz = Hertz(32_000_000);
 
 pub use crate::pac::pwr::vals::Vos as VoltageScale;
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

Dirbaio:
> I'm not sure the PllSource enum is still needed at this point but since this is my first contribution I went for something minimal. Let me know if PllSource should just go.

doesn't really matter since PLL support is WIP... when it's done they'll be different though, because ClockSrc can point to PLL while PllSrc can't.

> As a side note, because their frequencies are fixed, the datasheet of that chip refers to the high-speed clocks as HSI16 and HSE32. I'd be willing to contribute a change in that direction if it'd be useful.

we used the HSI16 name a while ago to match ST naming, and it became annoying to cfg HSI/HSI16 depending on the chip, so I intentionally renamed everything to HSI. HSE/HSE32 would be the same.

PR #2512, opened 2024-02-01
Add support for STM32WBA's HSE pre-scaler

Ecco:
This adds support for STM32WBA's prescaler. Several notes:
 - `RCC_CR` should probably be `modify`-ed than `write`-en. But that's another PR.
 - We may want to add something similar for STM32WB (in `l.rs`), because on that chip the `prescaler` option just doesn't do anything.

Also note that the `Hse` struct only has one field, but I did this to match the structure of other STM32s (as implemented in `l.rs`).

### DIFF
--- a/embassy-stm32/src/rcc/wba.rs
+++ b/embassy-stm32/src/rcc/wba.rs
@@ -1,5 +1,6 @@
 use stm32_metapac::rcc::vals::{Pllsrc, Sw};
 
+pub use crate::pac::rcc::vals::Hsepre as HsePrescaler;
 use crate::pac::{FLASH, RCC};
 use crate::rcc::{set_freqs, Clocks};
 use crate::time::Hertz;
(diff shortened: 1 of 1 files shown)


Dirbaio:
can you change HSI/HSE so it works like `l.rs`? i.e an `hse` field that turns on HSE if Some, an `hsi` field that turns HSI if true, then `mux` selects which of them is in use (which must be turned on, of course). https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/rcc/l.rs#L41-L42

Please try to keep the RCC API as close as possible as the "newer" RCCs (`l.rs`, `h.rs`, `f.rs`). The WBA RCC is not using the newest API design, it would be preferable if the new features you implement already work like the new design. See [this issue](https://github.com/embassy-rs/embassy/issues/2515) for the context on the different API designs we have and why the newer one is better.

Ecco:
Superseded by #2520

PR #2513, opened 2024-02-01
F2 ADC fixes

Gekkio:
I recently tested the ADC on STM32F215, and it seems to work great with the exception of these two small differences compared to F4/F7.

PS. F2 ADC is listed as "?" in the stm32 register versions Google sheet. I quickly compared the F2 reference manual with ADC v2 register definitions in stm32-data and couldn't find differences, so at least to me it looks like a match and the current v2 assumption seems correct.


Dirbaio (approved):
thanks!

PR #2518, opened 2024-02-02
Fix incorrect D1CPRE max for STM32H7 RM0468

### LINKED ISSUE #2445: STM32H7 RM0468 incorrect D1CPRE max
According to the reference manual (fig. 49), the D1CPRE clock max is 550MHz. This matches what the clock configuration in STM32CubeMX told me before. However, when trying to set a 544MHz clock, I get a crash at runtime, complaining about d1cpre being larger than its maximum value, which in your code is 520MHz:

https://github.com/embassy-rs/embassy/blob/6ff0e4bcf5fbcccd8ae52cc83be7ed9f83b66fde/embassy-stm32/src/rcc/h.rs#L483

I cannot seem to find a reason why the clocks maximum would be 520MHz.

msrd0:
Fixes #2445

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -455,7 +455,14 @@ pub(crate) unsafe fn init(config: Config) {
     };
     #[cfg(pwr_h7rm0468)]
     let (d1cpre_clk_max, hclk_max, pclk_max) = match config.voltage_scale {
-        VoltageScale::Scale0 => (Hertz(550_000_000), Hertz(275_000_000), Hertz(137_500_000)),
+        VoltageScale::Scale0 => {
+            let d1cpre_clk_max = if pac::SYSCFG.ur18().read().cpu_freq_boost() {
+                550_000_000
+            } else {
+                520_000_000
+            };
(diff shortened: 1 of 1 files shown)


Dirbaio:
seems it was not 10)% a mistake. looking through the datasheet, you need to set the CPUFREQ_BOOST option byte in order to reach 550mhz. I think the reason the max is set to 520 is there's no support for setting that option byte yet.

perhaps a fix is to read CPUFREQ_BOOST: and set the limit to 520 or 550 dynamically. So it won't set the option byte for you for now, but if you do set it yourself via some other means (like when factory programming) you can use 550mhz?

screenshot-2024-02-02_13-57-20

msrd0:
Oh I see it's part of the datasheet and not the reference manual :sweat_smile:

msrd0:
Reading (or writing) that bit seems a bit harder as it is part of `SYSCFG_UR18`

image

... which is not currently a part of the pac crate: https://docs.embassy.dev/stm32-metapac/git/stm32h723zg/syscfg/struct.Syscfg.html

Dirbaio (approved):
thanks! :rocket:

PR #2524, opened 2024-02-03
docs: Embassy in the wild - add air quality monitoring system

matoushybl:
I developed a simple system for air quality monitoring (CO2, temperature and humidity) based on nRF52 using the nrf-softdevice. It could help as an example for people starting with the softdevice.

OT: Do you want Embassy in the wild section to also have links to commecial/proprietary projects (of course with descriptions to what is embassy used for there), or should it be kept to open-source projects only?


Dirbaio:
Thanks!

> OT: Do you want Embassy in the wild section to also have links to commecial/proprietary projects (of course with descriptions to what is embassy used for there), or should it be kept to open-source projects only?

I think having a "case studies" section featuring commercial projects using Embassy could be useful, yes!

PR #2525, opened 2024-02-03
[DO NOT MERGE] ci-probing

eZioPan:
DO NOT MERGE


eZioPan:
The panic of `fdcan` feature missing still not show up.

PR #2540, opened 2024-02-06
feat: enhance bootloader for multiple flash support

badrbouslikhin:
## Summary
This pull request introduces a enhancement to the `BootLoaderConfig` and `FirmwareUpdaterConfig` in the `embassy-boot` crate. The primary goal is to implement support for multiple flash partitions, enabling the bootloader and firmware updater to handle `ActiveFlash`, `DFUFlash`, and `StateFlash` independently. This change is particularly beneficial for systems with dual-bank or external flash setups.

## Key Changes
- **Configurations Refactored**: Both `BootLoaderConfig` and `FirmwareUpdaterConfig` have been updated to support different flash partition types.
- **Partition Handling**: Adjustments made in various modules for alignment with the new multi-flash support, specifically in the instantiation of `BlockingPartition` and `Partition`.
- `bootloader/stm32` and `bootloader/stm32wb-dfu` have been updated and a new example `bootloader/stm32-dual-bank` has been introduced.

## Testing
- The changes have been tested in the blocking flavor on an STM32H747XI.

## TODO

- [x] Update all examples
- [x] Test on other targets (is the CI capable of performing this?)
- [x] Improve `from_linkerfile_blocking` api?

Your review would be greatly appreciated!

Specifically, I am not fully satisfied with the `from_linkerfile_blocking` when the same flash is used:
```rust
let config = BootLoaderConfig::from_linkerfile_blocking(&flash, &flash, &flash);
```
I see two ways to handle this: a builder pattern or wrapping `&flash` with an Option.

### DIFF
--- a/embassy-boot/src/boot_loader.rs
+++ b/embassy-boot/src/boot_loader.rs
@@ -79,7 +79,7 @@ impl<'a, ACTIVE: NorFlash, DFU: NorFlash, STATE: NorFlash>
     /// A `BootLoaderConfig` instance with `BlockingPartition` instances for the active, DFU, and state partitions.
     ///
     /// # Example
-    /// ```no_run
+    /// ```ignore
     /// // Assume `active_flash`, `dfu_flash`, and `state_flash` all share the same flash memory interface.
     /// let layout = Flash::new_blocking(p.FLASH).into_blocking_regions();
     /// let flash = Mutex::new(RefCell::new(layout.bank1_region));
--- a/embassy-boot/src/boot_loader.rs
+++ b/embassy-boot/src/boot_loader.rs
@@ -49,19 +49,19 @@ pub struct BootLoaderConfig<ACTIVE, DFU, STATE> {
     pub state: STATE,
 }
 
-impl<'a, ActiveFlash: NorFlash, DFUFlash: NorFlash, StateFlash: NorFlash>
+impl<'a, ACTIVE: NorFlash, DFU: NorFlash, STATE: NorFlash>
     BootLoaderConfig<
-        BlockingPartition<'a, NoopRawMutex, ActiveFlash>,
-        BlockingPartition<'a, NoopRawMutex, DFUFlash>,
-        BlockingPartition<'a, NoopRawMutex, StateFlash>,
+        BlockingPartition<'a, NoopRawMutex, ACTIVE>,
+        BlockingPartition<'a, NoopRawMutex, DFU>,
+        BlockingPartition<'a, NoopRawMutex, STATE>,
     >
 {
     /// Create a bootloader config from the flash and address symbols defined in the linkerfile
     // #[cfg(target_os = "none")]
     pub fn from_linkerfile_blocking(
-        active_flash: &'a Mutex<NoopRawMutex, RefCell<ActiveFlash>>,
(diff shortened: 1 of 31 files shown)


badrbouslikhin:
Ah, that explains why the Embassy bootloader doc. page says it `the bootloader supports both internal and external flash`!

Once I get your thoughts on the changes and `from_linkerfile_blocking` api, I'll proceed with the rest :).

embassy-boot/src/boot_loader.rs
```diff
@@ -49,16 +49,20 @@ pub struct BootLoaderConfig<ACTIVE, DFU, STATE> {
     pub state: STATE,
 }
 
-impl<'a, FLASH: NorFlash>
+impl<'a, ActiveFlash: NorFlash, DFUFlash: NorFlash, StateFlash: NorFlash>
```
lulf, line 52: Could you change the names ACTIVE, DFU and STATE which is also used elsewhere in the code.

embassy-boot/src/boot_loader.rs
```diff
@@ -49,16 +49,20 @@ pub struct BootLoaderConfig<ACTIVE, DFU, STATE> {
...
-        BlockingPartition<'a, NoopRawMutex, FLASH>,
+        BlockingPartition<'a, NoopRawMutex, ActiveFlash>,
+        BlockingPartition<'a, NoopRawMutex, DFUFlash>,
+        BlockingPartition<'a, NoopRawMutex, StateFlash>,
     >
 {
     /// Create a bootloader config from the flash and address symbols defined in the linkerfile
     // #[cfg(target_os = "none")]
-    pub fn from_linkerfile_blocking(flash: &'a Mutex<NoopRawMutex, RefCell<FLASH>>) -> Self {
+    pub fn from_linkerfile_blocking(
```
lulf, line 61: Personally I don't think this API is too bad. Using a builder doesn't provide that much gain, but perhaps a better rustdoc order explaining the parameters would be appropriate

embassy-boot/src/boot_loader.rs
badrbouslikhin: Done in af2b4df.

embassy-boot/src/boot_loader.rs
badrbouslikhin: Done in 634c409.

badrbouslikhin:
I updated and compiled all examples. 
Some of them required changes in the `memory.x` most likely due to recent code inflation (they were not compiling before this change). I adjusted linker files but I do not have the boards to validate the change.

Side note: is it useful to maintain that many `boot/application` STM32 examples? They look very similar.

I am not sure how to fix CI failures though.

lulf:
There is no HIL test for the bootloader, but any examples should be added to the ci.sh to ensure they always compile. It's unfortunate that the size increased previously though, and it goes undetected because in ci it compiles with --features skip-include to avoid the double pass.

The ci/test is failing on the unit tests in the bootloader though https://ci.embassy.dev/jobs/d45560625688

badrbouslikhin:
Alright, CI is happy now. Let me know if there's anything else I need to do.

lulf (approved):
LGTM, thanks!

PR #2541, opened 2024-02-06
fix: rtos-trace time missing

xgroleau:
Could not compile when `rtos-trace` was enabled due to missing embassy_time feature. Note that I added `embassy-time` for the `rtos-trace` feature

We could probably derive from `embassy-time-driver::now()` if it's preferred and don't want to depend on `embassy-time`

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -598,7 +598,7 @@ impl rtos_trace::RtosTraceOSCallbacks for Executor {
     #[cfg(feature = "integrated-timers")]
     fn time() -> u64 {
         const GCD_1M: u64 = gcd(embassy_time_driver::TICK_HZ, 1_000_000);
-        embassy_time_driver::now() * (1_000_00 / GCD_1M) / (embassy_time_driver::TICK_HZ / GCD_1M);
+        embassy_time_driver::now() * (1_000_000 / GCD_1M) / (embassy_time_driver::TICK_HZ / GCD_1M);
     }
     #[cfg(not(feature = "integrated-timers"))]
     fn time() -> u64 {
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -581,6 +581,15 @@ impl embassy_time_queue_driver::TimerQueue for TimerQueue {
 #[cfg(feature = "integrated-timers")]
 embassy_time_queue_driver::timer_queue_impl!(static TIMER_QUEUE: TimerQueue = TimerQueue);
 
+#[cfg(all(feature = "rtos-trace", feature = "integrated-timers"))]
+const fn gcd(a: u64, b: u64) -> u64 {
+    if b == 0 {
+        a
+    } else {
+        gcd(b, a % b)
+    }
+}
+
 #[cfg(feature = "rtos-trace")]
 impl rtos_trace::RtosTraceOSCallbacks for Executor {
     fn task_list() {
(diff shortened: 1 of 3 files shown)


Dirbaio:
embassy-executor intentionally doesn't depend on `embassy-time`, only on `embassy-time-driver`, so that it works with any `embassy-time` version (as long as both have the same `embassy-time-driver` version in common). Can you change this to use the "raw" `embassy-time-driver` API to get the current time?

xgroleau:
That's what I was feeling. Move the GCDs to the embassy-time-driver, but didn't reexport it in the embassy-time crate.

embassy-time-driver/src/lib.rs
```diff
@@ -70,6 +70,23 @@ mod tick;
 /// This value is specified by the [`tick-*` Cargo features](crate#tick-rate)
 pub const TICK_HZ: u64 = tick::TICK_HZ;
 
+const fn gcd(a: u64, b: u64) -> u64 {
```
Dirbaio, line 73: please just copypaste this instead of adding more public API to `embassy-time-driver`.

I would prefer to keep the `embassy-time-driver` public API surface ruthlessly barebones, to reduce the chances it needs a major version bump.

xgroleau: I see, that's understandable. Fixed it to avoid changes in time-driver.

Dirbaio (approved):
thanks!

embassy-executor/src/raw/mod.rs
```diff
@@ -581,14 +581,26 @@ impl embassy_time_queue_driver::TimerQueue for TimerQueue {
...
 #[cfg(feature = "rtos-trace")]
 impl rtos_trace::RtosTraceOSCallbacks for Executor {
     fn task_list() {
         // We don't know what tasks exist, so we can't send them.
     }
     #[cfg(feature = "integrated-timers")]
     fn time() -> u64 {
-        Instant::now().as_micros()
+        const GCD_1K: u64 = gcd(embassy_time_driver::TICK_HZ, 1_000);
+        let ms_per_tick = (1000 / GCD_1K) / (embassy_time_driver::TICK_HZ / GCD_1K);
```
Dirbaio, line 601: original code uses microseconds, you're now doing milliseconds. [Docs](https://docs.rs/rtos-trace/latest/rtos_trace/trait.RtosTraceOSCallbacks.html#tymethod.time) say it should be microseconds.

Dirbaio: the order of operations here will accumulate error if tick rate is not a clean divisor of 1e6. Note:

- embassy_time: `(ticks * (1000 / GCD_1K)) / (embassy_time_driver::TICK_HZ / GCD_1K);`
- this PR: `ticks * ((1000 / GCD_1K) / (embassy_time_driver::TICK_HZ / GCD_1K));`

if you do the division first, it'll get rounded a bit, and that error will accumulate for every second passed. If you multiply first then divide, the rounding is just +/- 1us independently of the absolute value of the uptime.

embassy-executor/src/raw/mod.rs
```diff
@@ -581,14 +581,26 @@ impl embassy_time_queue_driver::TimerQueue for TimerQueue {
...
 #[cfg(feature = "rtos-trace")]
 impl rtos_trace::RtosTraceOSCallbacks for Executor {
     fn task_list() {
         // We don't know what tasks exist, so we can't send them.
     }
     #[cfg(feature = "integrated-timers")]
     fn time() -> u64 {
-        Instant::now().as_micros()
+        const GCD_1M: u64 = gcd(embassy_time_driver::TICK_HZ, 1_000_000);
+        let us_per_tick = (1000 / GCD_1M) / (embassy_time_driver::TICK_HZ / GCD_1M);
```
Dirbaio, line 601: 1000 -> 1_000_000

also: the order of operations here will accumulate error if tick rate is not a clean divisor of 1e6. Note:

- embassy_time: `(ticks * (1000 / GCD_1K)) / (embassy_time_driver::TICK_HZ / GCD_1K);`
- this PR: `ticks * ((1000 / GCD_1K) / (embassy_time_driver::TICK_HZ / GCD_1K));`

if you do the division first, it'll get rounded a bit, and that error will accumulate for every second passed. If you multiply first then divide, the rounding is just +/- 1us independently of the absolute value of the uptime.

xgroleau: You're right, just fixed it. 

Sorry about the back in forth in this PR., I'll take my time for the future ones

Dirbaio (approved):
thanks!

PR #2544, opened 2024-02-08
Feat/adc l0

shufps:
Adds support for the ADC of L0s (which is a slightly modified but mostly compatible `v1` ADC) and provides an example.


Dirbaio (approved):
nice! thank you! :D

PR #2547, opened 2024-02-08
nrf: spim/spis: Add size checks for EasyDMA buffer

plaes:
On most nRF chips, maximum buffer size for EasyDMA is 255, thus we never got any data when attempting to use 256 bytes as RX/TX buffer.

Maybe it makes sense to expose maximum tx/rx buffer size?

### DIFF
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -13,7 +13,7 @@ pub use embedded_hal_02::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MO
 pub use pac::spim0::config::ORDER_A as BitOrder;
 pub use pac::spim0::frequency::FREQUENCY_A as Frequency;
 
-use crate::chip::FORCE_COPY_BUFFER_SIZE;
+use crate::chip::{EASY_DMA_SIZE, FORCE_COPY_BUFFER_SIZE};
 use crate::gpio::sealed::Pin as _;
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Would be even nicer to automatically chunk the transfer, but I agree checking it is better than silently doing the wrong thing for now.

plaes:
> Would be even nicer to automatically chunk the transfer, but I agree checking it is better than silently doing the wrong thing for now.

Well, chunking might open up another can of worms.
For example, when interfacing with SPI-NOR flash supporting byte/page program from 1...256 bytes (that's how I ran into this issue), you'll still end up with two SPI writes: 255 bytes and 1 byte, as writes could be limited due to page address boundaries. (And this specific chip's datasheet also mentions also: "if more than 256 bytes of data are sent to the device, then only the last 256 bytes sent are latched into the internal buffer").

Dirbaio:
the user does "assert CS, send 256 bytes, deassert CS"
embassy-nrf would see the "send 256 bytes", and chunk it into 2 DMA transfers, of 255 and 1 bytes.
The NOR flash chip will still see it as a single 256-byte transfer (with a bit of delay in the boundary between the chunks, but it shouldn't care). 


SPI is just "wiggle SCK while shifting out bits". On the wire it's indistinguishable whether it's 1 DMA transfer of 256, or 2 DMA transfers of 255+1.

PR #2548, opened 2024-02-09
Faq fixes due to embassy-time crate split

plaes:
I ran into the dependency graph conflicts when I wanted to pull in embassy-nrf from specific git revision (after #2547 was merged), and while at it, also found some irrelevant stuff in other areas of docs.

PS. What's the release-plan going to be for embassy-nrf crate? I guess there should be a fix version up in crates.io. (And another thing I noticed between 0.1.0 and master was API change regarding dropped GPIO generics).


lulf (approved):
Looks good. I think we should release 0.2 of embassy-nrf, I don't think there are any other big api changes in the plan. Any remaining un-genericsifying up your sleeve @Dirbaio ?

PR #2549, opened 2024-02-09
nrf: spim: Expose buffer length limits

plaes:
Exposing this feels a bit awkward, but still better than nothing.

One use-case for this would be pushing data to SPI-NOR flash at maximum bandwidth supported by either bus/device. For example in case of nRF52832 and Renesas SPI-NOR flash(AT25FF081A):
* nRF52832 supports transfers up to 255 bytes
* Flash can be programmed either in Byte/Page mode (1..256 bytes), but we cannot cross the 256-byte boundaries.

Therefore the most sensible way to operate in this case would be to push out data as two 128 byte chunks to program a single page.

### DIFF
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -20,6 +20,11 @@ use crate::interrupt::typelevel::Interrupt;
 use crate::util::{slice_in_ram_or, slice_ptr_parts, slice_ptr_parts_mut};
 use crate::{interrupt, pac, Peripheral};
 
+/// Maximum receive buffer size for single transfer
+pub const MAX_RX_LEN: usize = EASY_DMA_SIZE;
+/// Maximum transfer buffer size for single transfer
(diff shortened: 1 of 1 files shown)


Dirbaio:
as discussed in #2547 I believe the HAL should auto-chunk transfers, effectively removing the limit.

plaes:
> as discussed in #2547 I believe the HAL should auto-chunk transfers, effectively removing the limit.

[From #2547](https://github.com/embassy-rs/embassy/pull/2547#issuecomment-1936180767)
> the user does "assert CS, send 256 bytes, deassert CS"
embassy-nrf would see the "send 256 bytes", and chunk it into 2 DMA transfers, of 255 and 1 bytes.
The NOR flash chip will still see it as a single 256-byte transfer (with a bit of delay in the boundary between the chunks, but it shouldn't care). It's still a single CS assert+deassert.

OK, indeed it there's single CS assert/deassert then it should work. 
I had a false impression that chunking would also change state of `CS` pin.

PR #2556, opened 2024-02-10
Fix unaligned buffer in async updater

### LINKED ISSUE #2542: Unaligned buffer in embassy-boot async firmware updater for RP2040
When using the async firmware updater for RP2040, all operations to update the magic code fails with an unaligned error.

This can be traced to changes made in #2411, after which the aligned buffer must have a size of 4 due to READ_SIZE being the larger size, and being defined as 4 for RP2040.

https://github.com/embassy-rs/embassy/blob/2c5426aa5c41563f3b225d99d39793a765ec9204/embassy-boot/src/firmware_updater/asynch.rs#L229-L230

This becomes a problem for the set magic function, which calls the flash read function for both offset 0 (line 274) as well as 1 (line 278). This is because WRITE_SIZE is defined as 1 for RP2040.

https://github.com/embassy-rs/embassy/blob/2c5426aa5c41563f3b225d99d39793a765ec9204/embassy-boot/src/firmware_updater/asynch.rs#L273-L278

This eventually triggers an error in the flash background read function for RP2040, which requires that the offset is a multiple of 4 (lines 286 to 288).

https://github.com/embassy-rs/embassy/blob/2c5426aa5c41563f3b225d99d39793a765ec9204/embassy-rp/src/flash.rs#L270-L288

chamburr:
This PR fixes #2542.

I didn't know how to implement the change in magic size everywhere as suggested by lulf in the issue. What I've done here is to keep the magic size at size of WRITE_SIZE, and slice the buffer for writes to also be size of WRITE_SIZE. To make sure that reads are aligned to READ_SIZE, I avoid the second read for progress whenever the data is available in the first read. I have tested this on RP2040 and it works.

### DIFF
--- a/embassy-boot/src/firmware_updater/asynch.rs
+++ b/embassy-boot/src/firmware_updater/asynch.rs
@@ -276,16 +276,25 @@ impl<'d, STATE: NorFlash> FirmwareState<'d, STATE> {
     async fn set_magic(&mut self, magic: u8) -> Result<(), FirmwareUpdaterError> {
         self.state.read(0, &mut self.aligned).await?;
 
-        if self.aligned.iter().any(|&b| b != magic) {
+        if self.aligned[..STATE::WRITE_SIZE].iter().any(|&b| b != magic) {
             // Read progress validity
-            self.state.read(STATE::WRITE_SIZE as u32, &mut self.aligned).await?;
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #2558, opened 2024-02-10
RP: Add explicit reset time to ws2812 write fn.

CBJamo:
This was discussed back in #1834, but I never got around to actually fixing it. I found in testing that using exactly 50us wasn't reliable, thus 55us.


Dirbaio (approved):
thanks!

PR #2563, opened 2024-02-12
Print panics via defmt per default for the stm32f0 example

nbars:
Hey,

This is the default for nearly all examples and helps locate bugs.


Dirbaio (approved):
makes sense, thanks! :)

PR #2565, opened 2024-02-13
STM32 HMAC

caleb-garrett:
Adds HMAC support to the STM32 hash peripheral. An example is appended to the STM32F7 hash example.

`HmacKey` is defined as an optional type. Supplying a value of `Some` will generate an HMAC, while supplying a value of `None` will simply calculate a hash digest.


Dirbaio:
can you make the test also test HMAC?

Dirbaio:
thanks! 

(force-merging because f103 is still dead)

PR #2567, opened 2024-02-13
FAQ add hint to embassy-time linker error to include HAL in linking

jr-oss:


docs/modules/ROOT/pages/faq.adoc
```diff
@@ -118,6 +118,13 @@ features = [
 ]
 ----
 
+If you are in the early project setup phase and not using anything from the HAL, make sure the HAL is passed to the linker by adding this line to your source:
```
jamesmunns, line 121:
```suggestion
If you are in the early project setup phase and not using anything from the HAL, make sure the HAL is explicitly used to prevent the linker removing it as dead code by adding this line to your source:
```

What do you think of this change @jr-oss?

jamesmunns (approved):
Looks good, thank you! Will merge after a 👍 or 👎 from the author on one suggestion.

docs/modules/ROOT/pages/faq.adoc
jr-oss: Looks good to me.
Thanks for improving the description.

jamesmunns:
Looks like intermittent failure with the STM32F103 board?

PR #2570, opened 2024-02-14
Add missing TIM for time-driver; reorder time-driver selection when use "time-drvier-any"

eZioPan:
I find there are some discussion about missing "TIM2" when use with "time-driver-any".
In my opinion, we should pick the capable TIM with minimal function first, to avoid consume more-powerful/common-used TIM for time-driver.

Other changes:

1. `TIM11` is `TIM_1CH`, so not capable for time-driver
2. Add missing `TIM20`(ADV), `TIM23`(GP32), `TIM24`(GP32) for candidators
3. `TIM_ADV` can also be use as time-driver (but yes, should be last ones to pick)

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -183,40 +183,33 @@ fn main() {
 
     let time_driver_singleton = match time_driver.as_ref().map(|x| x.as_ref()) {
         None => "",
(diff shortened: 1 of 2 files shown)


Dirbaio:
bender run

Dirbaio (approved):
thanks!

PR #2571, opened 2024-02-14
Port fdcan HAL module to use PAC directly instead of fdcan crate.

cschuhen:
- Don't use fdcan crate dep anymore, instead some files have been forked from fdcan crate
- Provide separate FDCAN capable and Classic CAN API's
- Provide embedded-can traits.


cschuhen:
@Dirbaio, Apologies for another cursed big PR.

When you get a chance, perhaps do a quick scan of the patch first to let me know if the overall approach is OK. 
The files in fd/message_ram/ are almost exactly copies from the fdcan crate, it may be only rustfmt changes and not much more. config.rs and filter.rs are also not much different from fdcan crate.

Most of my effort went into peripheral.rs and fdcan.rs. peripheral.rs contains stuff taken from fdcan crate with the register access changed to use 'our' PAC.

Note, that in message_ram, there is a lot of register access that does NOT use our 'PAC'. Firstly, I don't think we even have those defined for G4. And also, as I understand it, for H5/H7 they are configurable. This HAL implementation just sets H5/H7 up to match what G4 does. We could make it more flexible but that would be a whole lot more work.

My eventual goal is to make an option to have read/write done by the interrupt handler(s). I got this working as a POC already but it needed more direct access to parts of what went into peripheral.rs so I kind of had to half do this PAC conversion anyway.

This leads me to the biggest API change. There is now 2 variants of read/write with the extras being read_fd and write_fd. Strictly these are not necessary, it just means an extra 56 bytes in the frame structs. However:

1. I estimate, most users will use classic CAN, so most users would pay this penalty for nothing.
2. My primary personal motivation leads back to my desire to add buffer(channel) backed IRQ mode, in that case the extra data size is a big factor. Users can have 5* the buffer length(frames) for the same number of bytes.
3. Also, a nice follow up step would be to use the same Frame structs for embassy's bxcan implementation. bxcan certinally does not need to waste the extra data, especially as many of these older chips are quite ram constrained (STM32F042 for example).

tcbennun:
Thanks for getting this kicked off, would be awesome to remove the fdcan dep.

FYI, I added message RAM register defs and message RAM addresses to the PAC, but it's not well tested. https://github.com/embassy-rs/stm32-data/pull/308

Should in theory work with G0, G4, L5, H5 and U5.

cschuhen:
Thanks Torin... I thought about investigating getting the PAC updated for the message RAM registers. It was the H7 that held me off, as you say with the flexibility. Thinking about how to deal with that. I'm sure we can figure out a way to make it work...

tcbennun:
I'd recommend having a trait in the HAL which has methods for getting the different regions. For non-H7 the impl uses the PAC. For H7 the impl takes a configuration struct which defines the regions. The same config struct can be used to set up the regions in driver init. The user could then even set up regions themselves, possibly

Dirbaio:
maybe an alternative is: always treat the FDCANRAM as "a bunch of bytes" in the PAC, then in the HAL have a "configuration" of how things are laid out. Then you can make the configuration user-settable in H7 and "fixed" in all other chips. This way the code that accesses FDCANRAM can be the same in all chips. 

(i'm assuming the non-H7 fixed layout is a "special case" of the H7 configurable layout? hopefully that's the case, but this being ST I wouldn't take it for granted).

tcbennun:
> maybe an alternative is: always treat the FDCANRAM as "a bunch of bytes" in the PAC, then in the HAL have a "configuration" of how things are laid out. Then you can make the configuration user-settable in H7 and "fixed" in all other chips. This way the code that accesses FDCANRAM can be the same in all chips.

In terms of responsibilities, I think I prefer the PAC describing the layout for non-H7 chips, rather than the HAL, but I recognise that it makes the HAL more complicated. I'm on the fence.

> (i'm assuming the non-H7 fixed layout is a "special case" of the H7 configurable layout? hopefully that's the case, but this being ST I wouldn't take it for granted).

In short, yes.

Example from RM0492 (H503, but mirrors other H5s):

image

Example from RM0455 (H7A3/B3/B0, but mirrors other H7s):

image

H7 has a single message RAM block shared between all FDCAN modules. On non-H7s, each module has its own message RAM, but they are adjacent so again, it can be treated as a "special case" of an H7 setup.

Dirbaio:
> In terms of responsibilities, I think I prefer the PAC describing the layout for non-H7 chips, rather than the HAL, but I recognise that it makes the HAL more complicated. I'm on the fence.

yeah we can leave the non-H7 registers in the PAC, they can be useful if someone wants to use the PAC directly and doesn't care about H7. The HAL can still "ignore" them and do everything with pointer math based on the "config". 

it's just a random idea, you guys are the experts, so feel free to choose the solution you think its best though!

cschuhen:
I do see some benefit in allowing custom configuration for H7, especially if the design lets you make more smaller buffers for classic CAN. However personally, it's lower on my priority list than buffering option which helps all users. I bought a H723 board so I could (and did) test this module with real CAN but my own projects are targeting G4.

But yes, to do this, the H7 will need a config struct that is used by the msg ram reader/writer functions. In the current design, there is a function to return a ref to an individual buffer. The type of this is from the code that I copied from the fdcan create. For non-H7, this could return the PAC type. For H7, perhaps it could return something with the same name(type alias) as the non-H7 PAC crate has. I need to look into the specifics more, I didn't look at Torin's changes yet.

Where do you both stand on this issue of message ram? Are we talking for down the track? Or do we need to do something for this PR to pass through?

Dirbaio:
i'm fine with whatever you think is best.

cschuhen:
@tcbennun, I looked again at the PAC. It seems that you only added methods to get access to the individual FIFO elements but didn't break that down further into the actual registers? Is that the case or am I looking in the wrong place?

cschuhen:
In general though, I'm happy to leave it as it is for now. It won't impact the existing API, only perhaps add some more config options for H7 but that should be non-breaking.

This is pending whatever further review comments come up for this patch.

tcbennun:
> @tcbennun, I looked again at the PAC. It seems that you only added methods to get access to the individual FIFO elements but didn't break that down further into the actual registers? Is that the case or am I looking in the wrong place?

Yes, currently it's just broken down into regions.

embassy-stm32/src/can/fd/message_ram/enums.rs
```diff
@@ -0,0 +1,233 @@
+// Note: This file is copied and modified from fdcan crate by Richard Meadows
+
+/// Datalength is the message length generalised over
+/// the Standard (Classic) and FDCAN message types
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum DataLength {
+    Standard(u8),
+    Fdcan(u8),
```
blaa, line 9: If this is part of public interface consider maybe using Extended instead of Fdcan? It's probably related to how fdcan crate named it though.

embassy-stm32/src/can/enums.rs
```diff
@@ -28,3 +29,19 @@ pub enum BusError {
     ///  At least one of error counter has reached the Error_Warning limit of 96.
     BusWarning,
 }
+impl TryFrom<u8> for BusError {
+    type Error = ();
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
```
blaa, line 34: Returning an error while converting an error seemed a bit convoluted on first sight.

Maybe having `BusError::NoError` and dropping TryFrom is not so bad? But it has it's semantical drawbacks. Unsure how rest of code handles cases like this.

Maybe using From and returning Option<BusError> instead of Result would be simpler and still convey complete meaning?

embassy-stm32/src/can/fd/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Module containing that which is speciffic to fdcan hardware variant
```
blaa, line 1: spelling

embassy-stm32/src/can/fd/message_ram/enums.rs
cschuhen: Yes, though I think Classic. 
because it could be either Extended  Standard ID's,
Seems we are using Classic for any CAN 2.0B packets i.e. not Fd

Dirbaio:
bender run

embassy-stm32/src/can/fd/message_ram/enums.rs
cschuhen: Pushed.

embassy-stm32/src/can/enums.rs
cschuhen: Pushed.
This was only used in one place, where reading the registers anyway so I moved it to peripheral.rs and just made it a vanilla 'method'.

Dirbaio:
I was deploying a teleprobe fix for the h5 and screwed it up while this was running.

bender run

cschuhen:
Ok, thanks. I have pushed fixes for Tomasz comments and rebased.

cschuhen:
Hi all, I have added another PR that builds on this one: This branch builds on https://github.com/embassy-rs/embassy/pull/2571

Dirbaio:
bender run

embassy-stm32/src/can/fd/filter.rs
```diff
@@ -0,0 +1,379 @@
+//! Definition of Filter structs for FDCAN Module
+//! Note: This file is copied and modified from fdcan crate by Richard Meadows
```
Dirbaio, line 2: make this a regular comment instead of a doc comment. Keeping credits in the code is nice, but the user who's browsing the docs doesn't care about it, they just want to get their FDCAN widget working.

cschuhen:
Regarding your comments:
1) Yes! Actually I did it already, please see follow-up PR here: https://github.com/embassy-rs/embassy/pull/2588 After getting the buffered mode working. 
2) Yes, understand. I kind of didn't really like having bxcan and fdcan files mixed in one directory so I was hoping to clean it up further later and figured starting a pattern for now was a good way to start. But that's OK, I can move it back.
3) Ok, I'll have a play with that and see if I can figure it out. 

Do you have time to check the approach of my Buffered implementation at least quickly to see if it seems reasonable(does it suck). There is an enum-based decision in the IRQ handler to deal with both options. Not sure if this is considered unacceptable. I guess this is a minor extra CPU cost in the IRQ handler as well as possibly a bit of overhead in the data portion of the enum union. I figured the size of  channel sender/receivers is probably smaller, than the wakers for non-buffered mode anyway.

Essentially, I want to know if I can apply these changes to the buffered branch and move forward with that. Or should I backtrack to this head and rebase the typestate change, working from there.

Dirbaio:
> I kind of didn't really like having bxcan and fdcan files mixed in one directory

separating in modules internally is fine. You can have the module private, and do `pub use fd::*`. I was referring to the public API only.

> Actually I did it already, please see follow-up PR here: https://github.com/embassy-rs/embassy/pull/2588 

> Essentially, I want to know if I can apply these changes to the buffered branch and move forward with that. Or should I backtrack to this head and rebase the typestate change, working from there.

oh okay, I hadn't looked at that one yet. I'm OK with merging it later, don't worry about rebasing.

Dirbaio:
does the hardware somehow require you can't change the configuration while it's running? if so I think it makes sense!

cschuhen:
Yes, many (all) registers say they are locked when not in config mode.

PR #2572, opened 2024-02-14
NRF: Implement chunked DMA transfers for SPIM

plaes:
On some chips (notably nrf52832), the maximum DMA transfer length was 255 which caused some hard to understand issues when interfacing with various devices over SPI bus.

Tested with nrf52832 + SPI flash (+ DFU flashing over BLE with SPI-NOR page size of 256 bytes).

### DIFF
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -17,7 +17,7 @@ use crate::chip::{EASY_DMA_SIZE, FORCE_COPY_BUFFER_SIZE};
 use crate::gpio::sealed::Pin as _;
 use crate::gpio::{self, AnyPin, Pin as GpioPin, PselBits};
 use crate::interrupt::typelevel::Interrupt;
-use crate::util::{slice_in_ram_or, slice_ptr_parts, slice_ptr_parts_mut};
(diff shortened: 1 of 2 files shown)


plaes:
Marked as draft until the `nrf52832_anomaly_109` handling is done once per operation.

Dirbaio:
the implementation feels a bit complex to me.

IMO it'd be cleaner to rename `blocking_inner_from_ram` to `blocking_inner_from_ram_chunk` as-is, then write a new `blocking_inner_from_ram` that calls `blocking_inner_from_ram_chunk` N times. This way the chunking logic is self-contained, and completely separate from the DMA stuff. Same for async.

plaes:
> the implementation feels a bit complex to me.
> 
> IMO it'd be cleaner to rename `blocking_inner_from_ram` to `blocking_inner_from_ram_chunk` as-is, then write a new `blocking_inner_from_ram` that calls `blocking_inner_from_ram_chunk` N times. This way the chunking logic is self-contained, and completely separate from the DMA stuff. Same for async.

@Dirbaio Thanks for the insight, PR ahas been reworked. I also had to introduce `utils::slice_ptr_len` in order look up the rx/tx slice lengths.

plaes:
@nilfit Could you please test this as well?

lulf (approved):
Looks good to me!

PR #2576, opened 2024-02-15
Add GSM0710 AT multiplexer implementation

MathiasKoch:


MathiasKoch:
I have been testing this out for a while now, and it seems to work most of the time, with a few hick-ups here and there.

I still have a few TODO comments, especially around hangup handling / V.24 signalling that I cannot seem to figure out how to get an ergonomic UX on, without it blowing up. **Suggestions very much appreciated**

Apart from that, I have cases where it feels like my application just stops receiving RX packets all together, resulting in infinite retransmissions until timeout. So far, my suspicion would be, that the underlying issue is the signalling handling above. 

If you want to test it out, that would be nice

CC @Dirbaio

MathiasKoch:
@Dirbaio I am back from summer vacation, and ready to see if I can drive this over the finish line, if you can find the time to test it out at your convenience sometime.

MathiasKoch:
@Dirbaio 
Revisiting this once again.

I just realized you own https://crates.io/crates/gsm0710
Do you think that would be a better fit for this, considering that it doesn't strictly have anything to do with `embassy`?

PR #2578, opened 2024-02-15
Improve rp2040 i2c slave

CBJamo:
This commit takes the fixes and error reporting improvements from jcdickinson's work in #2169 and applies them without overlaying a software state machine on top of the hardware state machine.

Also allows configuration of response to 'general call' writes.


CBJamo:
The run after I fixed the unused imports appears to have failed on an unrelated test (bootsel) but I don't see the i2c test in the log in there either.

Dirbaio:
loose wire spurious failure :P

bender run

Dirbaio (approved):
thanks!

Issue #2515, opened 2024-02-01
stm32: RCC refactor tracking issue


Dirbaio:
## The 3 designs

Note RCC in embassy-stm32 is a bit "in flux". Currently we have three API designs for RCC depending on the family, for historical reasons. Some have been ported from stm32-rs, some have been developed for embassy but before we knew what the best design is.

- v1: user specifies the target frequencies, RCC tries to solve for the best config (prescalers etc) to achieve that. [example](https://github.com/embassy-rs/embassy/blob/c4839e4671ade4b69f47e5436b5e44fae9109159/embassy-stm32/src/rcc/f1.rs#L19-L23)
- v2: user specifies prescalers/dividers instead of target freqs. Sysclk mux is an enum whose variants contain the config for the picked clock source. [example](https://github.com/embassy-rs/embassy/blob/e8c998aad882d766988ac2c0cb0c357c600b28c1/embassy-stm32/src/rcc/g0.rs#L21-L28)
- v3: user specifies prescalers/dividers. There's separate fields for each clock source (HSE, HSI, MSI, PLL..) to enable and configure them. Sysclk mux is an enum without data that just chooses the source (which the user must ensure is enabled, of course) [example](https://github.com/embassy-rs/embassy/blob/c4839e4671ade4b69f47e5436b5e44fae9109159/embassy-stm32/src/rcc/l.rs#L39-L54)

## Rationale for the v3 design

- Directly specifying frequencies (v1 api design) turned out to be quite troublesome. Solving for the best RCC config backwards from the target frequencies is hard, requires a lot of code that the compiler isn't always able to optimize out, and it's often impossible to exactly match the requested freqs, so we have to pick the closest one making it quite unpredictable for the end user.
- The v2 design rules out some configurations that the hardware allows. For example you might want to run sysclk from PLL from HSE, but run some peripheral from HSI (with a CCIPR register). The hardware does allow having HSE and HSI on at the same time, but you can't do it if the mux enum forces you to pick. With the v3 design you can enable all of HSI, HSE, PLL just fine, then pick one for sysclk and a different one for a particular peripheral.

## The refactor

We should port all RCCs to the v3 design, both for consistency between chips and to get the v3 benefits.

- [x] STM32F1: v1 #2564
- [x] STM32F2: v3
- [x] STM32L1: v3
- [x] STM32F4: v3
- [x] STM32F0: v1 #2564
- [x] STM32F3: v1 #2560 
- [x] STM32F7: v3
[...]

PR #2579, opened 2024-02-15
[embassy-stm32]: started stm32g4 RCC refactor

barnabywalters:
Part of #2515

* Copied API from f.rs where applicable
* HSE and HSI independantly configurable
* Boost mode set by user rather
* Added HSE, pll1_q and pll1_p frequencies to set_clocks call
* Stubbed max module based on f.rs, needs cleanup

I updated the examples and could successfully run blinky.rs on hardware. usb_serial.rs compiled and flashed successfully, was able to view the serial port using serial-monitor, text appeared to be echoed but I’m unsure exactly what should have happened.

Probably needs some cleanup, but if you’re happy with the API and the overall approach, I might be up for trying to make a g.rs driver generic over G0 and C0 if they are indeed similar in structure. I don’t have hardware to test those on though.

### DIFF
--- a/embassy-stm32/src/rcc/g4.rs
+++ b/embassy-stm32/src/rcc/g4.rs
@@ -1,10 +1,10 @@
 use stm32_metapac::flash::vals::Latency;
-use stm32_metapac::rcc::vals::{Adcsel, Pllsrc, Sw};
+use stm32_metapac::rcc::vals::{Adcsel, Sw};
 use stm32_metapac::FLASH;
 
 pub use crate::pac::rcc::vals::{
-    Adcsel as AdcClockSource, Fdcansel as FdCanClockSource, Hpre as AHBPrescaler, Pllm as PllM, Plln as PllN,
-    Pllp as PllP, Pllq as PllQ, Pllr as PllR, Ppre as APBPrescaler,
+    Adcsel as AdcClockSource, Clk48sel, Fdcansel as FdCanClockSource, Hpre as AHBPrescaler, Pllm as PllPreDiv,
+    Plln as PllMul, Pllp as PllPDiv, Pllq as PllQDiv, Pllr as PllRDiv, Pllsrc, Ppre as APBPrescaler, Sw as Sysclk,
 };
 use crate::pac::{PWR, RCC};
 use crate::time::Hertz;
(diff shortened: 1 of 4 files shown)


barnabywalters:
Oh and I have no idea how to integrate this with the embassy tests, help/advice about that would be appreciated.

Dirbaio:
> Oh and I have no idea how to integrate this with the embassy tests, help/advice about that would be appreciated.

we're not writing tests specifically for RCC currently. It's still tested "indirectly" thanks to the other tests needing working RCC to boot.

embassy-stm32/src/rcc/g4.rs
```diff
@@ -73,39 +65,53 @@ pub enum Clock48MhzSrc {
...
+            sys: Sysclk::HSI,
+            hsi48: Some(Default::default()),
+            pll: None,
             ahb_pre: AHBPrescaler::DIV1,
             apb1_pre: APBPrescaler::DIV1,
             apb2_pre: APBPrescaler::DIV1,
             low_power_run: false,
-            pll: None,
-            clock_48mhz_src: Some(Clock48MhzSrc::Hsi48(Default::default())),
+            clk48_src: Some(Clock48MhzSrc::Hsi48(Default::default())),
```
Dirbaio, line 109: the `Clock48MhzSrc::Hsi48` variant should not contain the HSI48 config, since it's already a separate field in the confg.

embassy-stm32/src/rcc/g4.rs
```diff
@@ -179,73 +225,56 @@ pub(crate) unsafe fn init(config: Config) {
...
     RCC.cfgr().modify(|w| {
         w.set_sw(sw);
         w.set_hpre(config.ahb_pre);
         w.set_ppre1(config.apb1_pre);
         w.set_ppre2(config.apb2_pre);
     });
 
-    let ahb_freq = sys_clk / config.ahb_pre;
-
     let (apb1_freq, apb1_tim_freq) = match config.apb1_pre {
```
Dirbaio, line 278: you can use `super::util::calc_pclk` for this. check other RCCs how they do it.

embassy-stm32/src/rcc/g4.rs
```diff
@@ -262,20 +291,22 @@ pub(crate) unsafe fn init(config: Config) {
         }
     };
 
-    // Setup the 48 MHz clock if needed
-    if let Some(clock_48mhz_src) = config.clock_48mhz_src {
-        let source = match clock_48mhz_src {
-            Clock48MhzSrc::PllQ => {
+    // Configure the 48MHz clock source for USB and RNG peripherals.
+    {
```
Dirbaio, line 295: these `{}` are not needed.

barnabywalters: was intended to prevent `source` leaking into the parent namespace, but I’ll shift it around to not require binding to a variable at all

embassy-stm32/src/rcc/g4.rs
```diff
@@ -262,20 +291,22 @@ pub(crate) unsafe fn init(config: Config) {
...
+        let source = match config.clk48_src {
+            Clk48Src::PLL1_Q => {
                 // Make sure the PLLQ is enabled and running at 48Mhz
                 let pllq_freq = pll_freq.as_ref().and_then(|f| f.pll_q);
                 assert!(pllq_freq.is_some() && pllq_freq.unwrap().0 == 48_000_000);
```
Dirbaio, line 300: these asserts now mean it's now impossible to *not* have a 48mhz clock somewhere. IMO this should let through whatever invalid config (clock off or bad freq) and propagate it to `set_clocks!`. `l.rs` [does this.](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/rcc/l.rs#L270-L283).

Then we can add an assert to USB on startup, that fails if clock is off or wrong. This way it only fails if you actually want USB.

barnabywalters: Okay, and the same for the RNG peripheral which also uses the 48MHz clock?

Dirbaio: yeah. I think that one doesn't require exactly 48mhz though, or at least not in all families. 

By "Then we can add an assert to USB on startup" I mean later, not in this PR. For that to work the clock muxes for USB must be well implemented on all families, which they aren't right now :(

barnabywalters: Yep, I assumed you meant the USB clock assert would happen elsewhere. IIRC some stm32-rs HALs have some kind of `assert_usb_clock_valid!()` macro they use for this

Dirbaio: yeah, it'd be when you create a USB driver.

Dirbaio (approved):
thank you! :partying_face:

PR #2580, opened 2024-02-15
cyw43: Reuse buf to reduce stack usage

DafabHoid:
By reusing the buffer in `cyw43::Runner::run()`, two buffers can be removed. This reduces stack usage significantly by 4 KiB, and therefore the task pool size for Runner by the same amount.

I'm not 100% sure why separate buffers were used before. Maybe they're aliased somewhere through unsafe code, but I could not spot anything.

### DIFF
--- a/cyw43/src/runner.rs
+++ b/cyw43/src/runner.rs
@@ -242,13 +242,12 @@ where
                         cmd,
                         iface,
                     }) => {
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
nice find! thank you :D

there's no reason to use 2 buffers, it was just not well optimized.

PR #2583, opened 2024-02-16
Enable PLL Clocks for TIMx peripherals on STM32F3xx Chips

OroArmor:
Things still to do:
- [x] Make sure that with the new metapac the frequencies are set correctly and also read with this change.
- [ ] Test on an STM32F303RE for real functionality.

Opening the PR now to get basic feedback.

Documents Referenced:
- [RM0313](https://www.st.com/resource/en/reference_manual/rm0313-stm32f37xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) STM32F37xx Chips - No TIMSW available
- [RM0316](https://www.st.com/resource/en/reference_manual/rm0316-stm32f303xbcde-stm32f303x68-stm32f328x8-stm32f358xc-stm32f398xe-advanced-armbased-mcus-stmicroelectronics.pdf) STM32F303xx/STM32F3x8xx Chips - Section 9.2.10
- [RM0364](https://www.st.com/resource/en/reference_manual/rm0364-stm32f334xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) STM32F334xx Chips - Section 8.2.10
- [RM0365](https://www.st.com/resource/en/reference_manual/rm0365-stm32f302xbcde-and-stm32f302x68-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) STM32F302xx Chips - Section 9.2.10
- [RM0366](https://www.st.com/resource/en/reference_manual/rm0366-stm32f301x68-and-stm32f318x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) STM32F301xx/STM32F318xx Chips - Section 7.2.10

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -510,11 +510,7 @@ fn main() {
                     let field_name = format_ident!("{}", field_name);
                     let enum_name = format_ident!("{}", enum_name);
 
-                    rcc_cfgr_regs.insert((
-                        fieldset_name.clone(),
-                        field_name.clone(),
-                        enum_name.clone(),
-                    ));
+                    rcc_cfgr_regs.insert((fieldset_name.clone(), field_name.clone(), enum_name.clone()));
 
                     let match_arms: TokenStream = enumm
                         .variants
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -601,7 +601,7 @@ fn main() {
             .iter()
             .map(|(_fieldset, fieldname, enum_name)| {
                 quote! {
-                    pub #fieldname: Option<crate::pac::rcc::vals::#enum_name>
+                    pub #fieldname: Option<#enum_name>
                 }
             })
             .collect();
@@ -627,23 +627,32 @@ fn main() {
             })
             .collect();
 
-        g.extend(quote! {
-            #[derive(Clone, Copy)]
-            pub struct ClockMux {
-                #( #struct_fields, )*
-            }
+        let enum_names: BTreeSet<_> = rcc_cfgr_regs
+            .iter()
+            .map(|(_fieldset, _fieldname, enum_name)| enum_name)
+            .collect();
 
-            impl Default for ClockMux {
-                fn default() -> Self {
-                    Self {
-                        #( #field_names: None, )*
+        g.extend(quote! {
+            pub mod mux {
+                #(pub use crate::pac::rcc::vals::#enum_names as #enum_names; )*
+
+                #[derive(Clone, Copy)]
+                pub struct ClockMux {
+                    #( #struct_fields, )*
+                }
+
+                impl Default for ClockMux {
+                    fn default() -> Self {
+                        Self {
+                            #( #field_names: None, )*
+                        }
                     }
                 }
-            }
 
-            impl ClockMux {
-                pub fn init(self) {
-                    #( #inits )*
+                impl ClockMux {
+                    pub fn init(self) {
+                        #( #inits )*
+                    }
                 }
             }
         });
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -430,7 +430,7 @@ fn main() {
 
     let mut clock_names = BTreeSet::new();
 
-    let mut rcc_cfgr_regs = BTreeMap::new();
+    let mut rcc_cfgr_regs = BTreeSet::new();
 
     for p in METADATA.peripherals {
         if !singletons.contains(&p.name.to_string()) {
@@ -510,11 +510,7 @@ fn main() {
                     let field_name = format_ident!("{}", field_name);
                     let enum_name = format_ident!("{}", enum_name);
 
-                    if !rcc_cfgr_regs.contains_key(mux.register) {
-                        rcc_cfgr_regs.insert(mux.register, Vec::new());
-                    }
-
-                    rcc_cfgr_regs.get_mut(mux.register).unwrap().push((
+                    rcc_cfgr_regs.insert((
                         fieldset_name.clone(),
                         field_name.clone(),
                         enum_name.clone(),
@@ -602,10 +598,10 @@ fn main() {
         }
     }
 
-    for (rcc_cfgr_reg, fields) in rcc_cfgr_regs {
-        println!("cargo:rustc-cfg={}", rcc_cfgr_reg.to_ascii_lowercase());
+    if !rcc_cfgr_regs.is_empty() {
+        println!("cargo:rustc-cfg=clock_mux");
 
-        let struct_fields: Vec<_> = fields
+        let struct_fields: Vec<_> = rcc_cfgr_regs
             .iter()
             .map(|(_fieldset, fieldname, enum_name)| {
                 quote! {
@@ -614,12 +610,12 @@ fn main() {
             })
             .collect();
 
-        let field_names: Vec<_> = fields
+        let field_names: Vec<_> = rcc_cfgr_regs
             .iter()
             .map(|(_fieldset, fieldname, _enum_name)| fieldname)
             .collect();
 
-        let inits: Vec<_> = fields
+        let inits: Vec<_> = rcc_cfgr_regs
             .iter()
             .map(|(fieldset, fieldname, _enum_name)| {
                 let setter = format_ident!("set_{}", fieldname);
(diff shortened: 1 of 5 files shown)


OroArmor:
Since a lot of the code for enabling the registers is duplicated, would a macro make sense?

Also I realized that the current implementation for that is wrong.

OroArmor:
Currently the TIMx peripherals will just get the PLL clock, and not have the multiple of two applied. See: https://github.com/embassy-rs/embassy/blob/26fc17e8a7db0957896db28cb2299ab9cf89aea5/embassy-stm32/build.rs#L511-L523. Not sure if this should be hardcoded in, or if the metapac should be updated to include a scalar value or something

Dirbaio:
I was planning on autogenerating these for `stm32-data`. We already have the data ([example](https://github.com/embassy-rs/stm32-data-generated/blob/cade675b5f00c11da2fe56b31c661efff065a3e5/data/chips/STM32F334R8.json#L2191-L2194)), and we're [already using it](https://github.com/embassy-rs/embassy/blob/9352621058bb4e96b8897ec431cc939838729442/embassy-stm32/build.rs#L494) in build.rs so peripheral drivers use the actually configured frequencies in the mux registers.

the missing bit is an API for actually setting the mux registers (that's not direct PAC usage). I'm afraid hand-writing it like you did in this PR for every single chip family and peripheral is not maintainable. The ideal solution would be to make `build.rs` generate a struct with all the mux fields for the current chip, and code to apply it to the RCC regs, and just call that from the RCC hand-written code.

OroArmor:
> peripheral drivers use the actually configured frequencies in the mux registers.

Yep! That stuff looks great, but it doesn't seem to include the doubling of the frequency, unless I am missing something.
<br />

> I'm afraid hand-writing it like you did in this PR for every single chip family and peripheral is not maintainable.

Yeah... I wasn't sure about this, since it seems to have just generated enough `#[cfg()]`s to drive someone crazy. 

<br />
> The ideal solution would be to make `build.rs` generate a struct with all the mux fields for the current chip, and code to apply it to the RCC regs, and just call that from the RCC hand-written code.

I would assume that this would need some form of update to the metapac, but I would be interested in helping with that. Any pointers on where to start?

Dirbaio:
> Yep! That stuff looks great, but it doesn't seem to include the doubling of the frequency, unless I am missing something.

frequency is only doubled when running from PCLK, not PLL. It's handled here :

https://github.com/embassy-rs/embassy/blob/9352621058bb4e96b8897ec431cc939838729442/embassy-stm32/src/rcc/mod.rs#L99
https://github.com/embassy-rs/embassy/blob/9352621058bb4e96b8897ec431cc939838729442/embassy-stm32/src/rcc/f013.rs#L231

that goes into the `pclkX_tim` clock, which can then be selected with the mux.

> I would assume that this would need some form of update to the metapac

the data is already there in the metapac, youcan already use it from embassy-stm32 build.rs

OroArmor:
Ah perfect! I'll pivot this PR to generate the API for setting the clock source!

Dirbaio:
bender run

embassy-stm32/build.rs
```diff
@@ -508,6 +510,16 @@ fn main() {
                     let field_name = format_ident!("{}", field_name);
                     let enum_name = format_ident!("{}", enum_name);
 
+                    if !rcc_cfgr_regs.contains_key(mux.register) {
+                        rcc_cfgr_regs.insert(mux.register, Vec::new());
+                    }
+
+                    rcc_cfgr_regs.get_mut(mux.register).unwrap().push((
```
Dirbaio, line 517: you can do this with `.entry(..).or_insert_with(..)`. Shorter and avoids the unwrap.

Dirbaio:
looking great!

- I'd do one struct for all "clock selection" fields, not one per register. Which register a field is in is in an implementation detail the user doesn't care about. I'd name it `Mux` perhaps, and the Config field `mux`.
- Please reexport that struct under `embassy_stm32::rcc`. Otherwise it's not accessible in rustdoc, and it's awkward that the user can access it through the config field but can't name it.
- You'll have to deduplicate the fields, there's chips where a single "clock select" field is used for multiple peripherals, causing your code to emit it multiple times. (this is the cause of the CI failure).

examples/stm32f334/src/bin/pwm.rs
```diff
@@ -27,7 +27,10 @@ async fn main(_spawner: Spawner) {
         config.rcc.ahb_pre = AHBPrescaler::DIV1;
         config.rcc.apb1_pre = APBPrescaler::DIV2;
         config.rcc.apb2_pre = APBPrescaler::DIV1;
+
+        // TODO: The two lines here do the same thing
         config.rcc.hrtim = HrtimClockSource::PllClk;
+        config.rcc.mux.hrtim1sw = Some(embassy_stm32::pac::rcc::vals::Timsw::PLL1_P);
```
OroArmor, line 33: Not sure what to do here. Both of these modify the same system, but I don't want to remove the old API. 
In addition, the `Timsw` enum does seem to be easily importable, something should probably be done for that.

OroArmor:
Alright, I tested this on an STM32F303RE, and it seems like it is grabbing the wrong frequency (ie, with PLL1_P selected, it uses the pclk frequency, so pwm output has double the frequency).

examples/stm32f334/src/bin/pwm.rs
Dirbaio: yes, please remove the old API. breaking changes are fine.

about imports: yes, they shouldn't require `embassy_stm32::pac`... perhaps you can generate a `mod mux` and put it in `embassy_stm32::rcc::mux`, so you'd do `embassy_stm32::rcc::mux::Timsw::PLL1_P`. I wouldn't put all the enums in the main `embassy_stm32::rcc` mod because bigger chips have lots of them, which will clutter docs.

OroArmor: Should I re-export the enums from `pac` or make new ones that reference the values from `pac`? If I should do the latter, should I also use the same names, or should I make better names?

Dirbaio: reexporting is fine, RCC already reexports many enums from the PAC like for PLL config etc. 

I don't tihnk it'd be easy to make build.rs automatically come up with better names? i'd just use the same names.

Dirbaio (approved):
i'm going to merge this since I want to extend it to other chips too. Thanks for the PR! :)

Dirbaio:
> Alright, I tested this on an STM32F303RE, and it seems like there is no doubling effect calculated (ie it uses the value of PLL1_P and not 2 * PLL1_P)

as far as I can tell from reading the RM, the hardware only doubles the frequency if it's clocked from PCLKx. Other clocks don't get the doubling. (The "doubling" is actually "divide one time less the APB prescaler" which is why there's no doubling if APB prescaler is 1. 

Can you point me to what documentation says the doubling applies to PLL1_P too? or, have you tested this on hardware?

OroArmor:
Ah sorry I didn't make it super clear but the frequencies of the Tim peripherals were set to pll clk abut not 2 * pll clk, and I think that data needs to come from the meta pac

OroArmor:
Made the comment at the same time. Give me a while to get to my computer to write something up.

OroArmor:
From RM0316: image

In addition, STMCube shows this: image

> Can you point me to what documentation says the doubling applies to PLL1_P too? or, have you tested this on hardware?

I don't think it entirely doubles `PLL_1P`, but makes a new clock that is `PLL1_P * 2`. I did end up testing it on hardware, and was able to get the correct behavior by adding ` * 2u32` to the `frequency` function on the `TIM1` peripheral like this:
```patch

Dirbaio:
oh wow okay, this is a F3-specific exception then :face_with_spiral_eyes: . Classic ST trap.

> Thinking about it, I believe the best way to do this (for the STM32F3xx series at least), is to make a new PLL1_P2 clock and reference that clock in the metapac instead of the PLL1_P clock.

yes, sounds good! name it `PLL1_P_MUL_2` for consistency, we already have other `_DIV_X` or `_MUL_X` clocks.

OroArmor:
Tried taking a crack at it, but I don't think I fully understand the auto-generation that exists (See https://github.com/embassy-rs/stm32-data/pull/422). I can make the PR for this repository, since it is fairly straight-forward.

PR #2596, opened 2024-02-19
feat(nrf/spim): allow specifying drive of SPI pins

NBonaparte:
Makes nrf SPIM pin drives configurable, similar to the TWIM peripheral.
Apparently, high drive messes up interfacing with 3-wire SPI.

### DIFF
--- a/embassy-nrf/src/gpio.rs
+++ b/embassy-nrf/src/gpio.rs
@@ -189,7 +189,7 @@ impl<'d> Output<'d> {
     }
 }
 
-fn convert_drive(drive: OutputDrive) -> DRIVE_A {
+pub(crate) fn convert_drive(drive: OutputDrive) -> DRIVE_A {
     match drive {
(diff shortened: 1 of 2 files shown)


lulf (commented):
Is 'high drive' the only value that makes sense here, or could could the gpio::OutputDrive  type be used instead of a boolean?

NBonaparte:
I originally had `OutputDrive` but I changed it to bool. AFAIK, both levels have to be driven for SPI, and I'm not sure if there's a case where you would want standard drive for 0 and high for 1 (or vice versa).

Dirbaio:
TWI only has a `bool` because in TWI (I2C) *has* to be open-drain, you *never* want to drive a line high. For SPI you can do anything, I imagine one might want a weird "open-drain SPI" in niche applications, perhaps when bitbanging another protocol. I would make the config setting the full `OutputDrive`.

Also, please make the default high drive to match the previous behavior.

PR #2597, opened 2024-02-19
stm32/rcc: reset RTC on stm32l0

fe1es:


Dirbaio:
I think the issue you're seeing is due to `#[cfg(not(any(rcc_l0, rcc_l0_v2, ...)))]` a few lines above is not matching because RCC names for L0 were changed in https://github.com/embassy-rs/embassy/pull/2564 and weren't updated here.

Can you try changing it to `#[cfg(not(any(stm32l0, ...)))]` and seeing if it fixes it? the intent of this code is to reset the entire backup domain, not just RTC.

fe1es:
Do you mean changing `#[cfg(not(any(rcc_l0, rcc_l0_v2, rcc_l1, stm32h5, stm32c0)))]` to `#[cfg(not(any(stm32l0, rcc_l1, stm32h5, stm32c0)))]`? Unfortunately that didn’t fix it.

The issue I ran into is that if you run a program with rcc::LsConfig::default_lsi first, then change it to default_lse and run it again without dropping power, it gets stuck on Rtc::new.

I found that the RTC_ISR_INITF bit is never set, so it’s polled forever. It seems the RTC is not working here, as RTCSEL can’t be switched until RTCRST is set or POR is occurred once it’s selected, and LSION is cleared by system reset.

Dirbaio (approved):

PR #2600, opened 2024-02-19
Flash support for STM32H50x [wip]

tcbennun:
goals for h50:

- [x] basic working impl
- [ ] support for "disabling" dual-bank (i.e. pretending there's only 1 bank)


tcbennun:
Basic impl tested on a modified `flash.rs` example, works fine.

Dirbaio:
thank you!

PR #2605, opened 2024-02-20
USB needs PWR_CR2 USV set on STM32L4

Gekkio:
Confirmed to be needed on an STM32L422, and based on a quick look at L4/L4+ reference manuals, this bit is present and required to be set on all L4 chips that have some kind of USB peripheral (USB or OTG_FS). The `usb_otg` driver already sets it for `cfg(stm32l4)` and we should do the same thing here.


Dirbaio:
thank you!

PR #2606, opened 2024-02-20
stm32/dma: add AnyChannel, add support for BDMA on H7.

Dirbaio:
Things done now:

- Add AnyChannel
- Unify Channel/AnyChannel for H7 DMA+BDMA. This will finally allow using peripherals that are wired to BDMA.
- Remove TxDma/RxDma generics from a few drivers.

Later will do:

- Allow removing the TxDma/RxDma generics ALL drivers, thanks to AnyChannel.


Dirbaio:
bender run

PR #2607, opened 2024-02-21
Add block-device-driver impl for use with embedded-fatfs

MabezDev:
I'm slowly working towards the first embedded-fatfs release! The biggest changes have been that the device trait and helpers are now outside of embedded-fatfs itself. This means I can 

* PR impls before releasing embedded-fatfs
* make breaking changes in embedded-fatfs _without_ making HAL implementors update
* the block device is now more useful outside of embedded-fatfs, for instance someone could use these building blocks to implement another filesystem, or even there own.

This PR is mostly inert, but if you'd like to try it out you can use [the esp32c6 example I have](https://github.com/MabezDev/embedded-fatfs/tree/master/examples/esp32c6) in embedded-fatfs proper, which you can use a reference to craft an stm example. I have tested with my project and once embedded-fatfs has a release I plan on adding a stm32f2 example either here or in embedded-fatfs.

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -88,6 +88,8 @@ static_assertions = { version = "1.1" }
 volatile-register = { version = "0.2.1" }
 bitflags = "2.4.2"
 
+block-device-driver = { version = "0.2" }
+aligned = "0.4.1"
 
 [dev-dependencies]
 critical-section = { version = "1.1", features = ["std"] }
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -1511,3 +1511,44 @@ foreach_peripheral!(
         }
     };
 );
+
+impl<'d, T: Instance, Dma: SdmmcDma<T> + 'd> block_device_driver::BlockDevice<512> for Sdmmc<'d, T, Dma> {
+    type Error = Error;
+    type Align = aligned::A4;
+
+    async fn read(
+        &mut self,
+        mut block_address: u32,
+        buf: &mut [aligned::Aligned<Self::Align, [u8; 512]>],
+    ) -> Result<(), Self::Error> {
+        // FIXME/TODO because of missing read_blocks multiple we have to do this one block at a time
+        for block in buf.iter_mut() {
+            // safety aligned by block device
+            let block = unsafe { &mut *(block as *mut _ as *mut crate::sdmmc::DataBlock) };
+            self.read_block(block_address, block).await?;
+            block_address += 1;
+        }
+
+        Ok(())
+    }
+
+    async fn write(
+        &mut self,
+        mut block_address: u32,
+        buf: &[aligned::Aligned<Self::Align, [u8; 512]>],
+    ) -> Result<(), Self::Error> {
+        // FIXME/TODO because of missing read_blocks multiple we have to do this one block at a time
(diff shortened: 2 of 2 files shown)


embassy-stm32/src/sdmmc/mod.rs
```diff
@@ -1478,3 +1478,39 @@ foreach_peripheral!(
         impl Instance for peripherals::$inst {}
     };
 );
+
+impl<'d, T: Instance, Dma: SdmmcDma<T> + 'd> block_device_driver::BlockDevice<512> for Sdmmc<'d, T, Dma> {
+    type Error = Error;
+
+    async fn read(&mut self, mut block_address: u32, buf: &mut [[u8; 512]]) -> Result<(), Self::Error> {
+        assert!(&buf[0] as *const _ as u32 % 4 == 0);
```
Dirbaio, line 1486: if the trait allows unaligned buffers, this code should work with unaligned buffers IMO. Otherwise you get asserts at runtime, have to add hacks to trait users so they align the buffer even if it's not compile-time checked, and bad compatibility across impls.

(There's a similar problem with nRF's flash: it also asserts at runtime the buffer is 4-aligned when the embedded-storage traits don't require it, and it has caused quite a bit of pain. I see it as a bad mistake that I regret and should fix someday...)

MabezDev: Hmm. I wanted to keep the block device trait as simple as possible, but maybe adding some kind of alignment constraint is worthwhile.

I've already explored this with https://docs.rs/block-device-adapters/latest/block_device_adapters/struct.BufStream.html which you'll need anyways to use embedded-fatfs, so in _this_ case the assert is more of fail safe. However other fs impls may choose to use the block device directly.

I think I could just clean up and use [AlignedBuffer](https://github.com/MabezDev/embedded-fatfs/blob/11bd13515886c539922087903a7d35657e8b9a3f/block-device-adapters/src/buf_stream.rs#L271-L277), but I'd need to explore if can still be cast from byte slice to block slice and vice versa :thinking:. Generic alignment can be a bit tricky in Rust.

MabezDev: I think that buffer alignment is going to be common, I looked at the sdmmc peripheral for the esp's and for DMA usage they require word-aligned buffers too, I'm going to look into adding some kind of alignment constraint to the trait and report back.

MabezDev:
This PR has been updated to use a new, but unreleased version of `block-device-driver` which includes buffer alignments within the trait itself https://github.com/MabezDev/embedded-fatfs/pull/23.

Release of the new version is blocked on https://github.com/japaric/aligned/pull/18.

Dirbaio:
is this still draft for a reason?

embassy-stm32/Cargo.toml
```diff
@@ -88,6 +88,9 @@ static_assertions = { version = "1.1" }
 volatile-register = { version = "0.2.1" }
 bitflags = "2.4.2"
 
+fdcan = { version = "0.2.0", optional = true }
```
Dirbaio, line 91: the fdcan dep is unused. rebase mishap maybe?

MabezDev:
> is this still draft for a reason?

I had some issues using embassy main in my project, but I've recently sorted those and can confirm this PR works as expected now :D

Dirbaio:
:rocket:

Dirbaio:
manually stomping over spurious failure

PR #2610, opened 2024-02-22
Add Feature Descriptions to Docs

AdinAck:
I was adding a feature list page to [cargo-embassy](https://github.com/adinack/cargo-embassy) and decided to copy it over here as well, since this has been talked about as something people wanted.


Dirbaio:
I think we should focus on documenting them on the `Cargo.toml` files, and include them in the crate readme with the `document-features` crate. Some of the crate (like embassy-stm32, embassy-nrf) already do this.

If we document them in the book separately, they're likely to fall out of date easily.

AdinAck:
I didn't even know the `document-features` crate was a thing lol. *Of course* that is the right way to go about this.

PR #2613, opened 2024-02-22
ADC support for H5

tcbennun:
Most similar to `adc_v3` so the impl lives there. Blocking only for now.

Depends on https://github.com/embassy-rs/stm32-data/pull/407.

Tested temperature reads, seems fine


Dirbaio:
merged fast to get stm32-metapac updated.

thanks for the PR!

PR #2614, opened 2024-02-22
stm32: time_driver: allow use of TIM1 for driver

tcbennun:
I'm sure there's a great reason for this being omitted, typo included, <s>but the thing seems to work</s> [bit eager there] and I really need TIM2 & 3 free, sooooo

edit: PR's scope now includes correctly using the TIMx_CC interrupt source for advanced timers (TIM1/8/20)

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -189,7 +189,6 @@ fn main() {
         Some("tim5") => "TIM5",
         Some("tim8") => "TIM8",
         Some("tim9") => "TIM9",
-        Some("tim11") => "TIM11",
         Some("tim12") => "TIM12",
         Some("tim15") => "TIM15",
         Some("tim20") => "TIM20",
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -189,6 +189,7 @@ fn main() {
         Some("tim5") => "TIM5",
         Some("tim8") => "TIM8",
         Some("tim9") => "TIM9",
+        Some("tim11") => "TIM11",
         Some("tim12") => "TIM12",
         Some("tim15") => "TIM15",
         Some("tim20") => "TIM20",
--- a/embassy-stm32/src/time_driver.rs
+++ b/embassy-stm32/src/time_driver.rs
@@ -16,6 +16,8 @@ use crate::pac::timer::vals;
 use crate::rcc::sealed::RccPeripheral;
 #[cfg(feature = "low-power")]
 use crate::rtc::Rtc;
+#[cfg(any(time_driver_tim1, time_driver_tim8, time_driver_tim20))]
+use crate::timer::sealed::AdvancedControlInstance;
 use crate::timer::sealed::{CoreInstance, GeneralPurpose16bitInstance as Instance};
 use crate::{interrupt, peripherals};
 
@@ -76,6 +78,14 @@ foreach_interrupt! {
             DRIVER.on_interrupt()
         }
     };
+    (TIM1, timer, $block:ident, CC, $irq:ident) => {
+        #[cfg(time_driver_tim1)]
+        #[cfg(feature = "rt")]
+        #[interrupt]
+        fn $irq() {
+            DRIVER.on_interrupt()
+        }
+    };
     (TIM2, timer, $block:ident, UP, $irq:ident) => {
         #[cfg(time_driver_tim2)]
         #[cfg(feature = "rt")]
@@ -116,6 +126,14 @@ foreach_interrupt! {
             DRIVER.on_interrupt()
         }
     };
+    (TIM8, timer, $block:ident, CC, $irq:ident) => {
+        #[cfg(time_driver_tim8)]
+        #[cfg(feature = "rt")]
+        #[interrupt]
+        fn $irq() {
+            DRIVER.on_interrupt()
+        }
+    };
     (TIM9, timer, $block:ident, UP, $irq:ident) => {
         #[cfg(time_driver_tim9)]
         #[cfg(feature = "rt")]
@@ -141,6 +159,14 @@ foreach_interrupt! {
         }
     };
     (TIM20, timer, $block:ident, UP, $irq:ident) => {
+        #[cfg(time_driver_tim20)]
+        #[cfg(feature = "rt")]
+        #[interrupt]
+        fn $irq() {
+            DRIVER.on_interrupt()
+        }
+    };
+    (TIM20, timer, $block:ident, CC, $irq:ident) => {
         #[cfg(time_driver_tim20)]
         #[cfg(feature = "rt")]
         #[interrupt]
@@ -280,6 +306,14 @@ impl RtcDriver {
 
         <T as CoreInstance>::Interrupt::unpend();
         unsafe { <T as CoreInstance>::Interrupt::enable() };
+
+        #[cfg(any(time_driver_tim1, time_driver_tim8, time_driver_tim20))]
+        {
+            <T as AdvancedControlInstance>::CaptureCompareInterrupt::unpend();
+            unsafe {
+                <T as AdvancedControlInstance>::CaptureCompareInterrupt::enable();
+            }
+        }
 
         r.cr1().modify(|w| w.set_cen(true));
     }
--- a/embassy-stm32/src/time_driver.rs
+++ b/embassy-stm32/src/time_driver.rs
@@ -38,7 +38,7 @@ cfg_if::cfg_if! {
     }
 }
 
-#[cfg(time_drvier_tim1)]
+#[cfg(time_driver_tim1)]
 type T = peripherals::TIM1;
 #[cfg(time_driver_tim2)]
 type T = peripherals::TIM2;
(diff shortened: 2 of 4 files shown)


tcbennun:
Okay I actually get weird timing issues with it, almost like it wasn't available for a reason 🤔 

Would be great to get to the bottom of it

eZioPan:
Would you do me a favor, copy the rest of TIMs to Cargo.toml? Thanks in advance! There is actually no other reasons hide them, I just forgot to add them to Cargo.toml last time.

tcbennun:
> Would you do me a favor, copy the rest of TIMs to Cargo.toml? Thanks in advance! There is actually no other reasons hide them, I just forgot to add them to Cargo.toml last time.

Happy to... any idea why TIM1 might not be working properly out of the box?

eZioPan:
I remember since we use 16-bit timer and interrupt to handle time-driver. The minimal delay cannot be lower than a few ms. Don't you if you set time frequency too high. The default value 1_000_000 hz can be too high for 16-bit timer.

tcbennun:
> I remember since we use 16-bit timer and interrupt to handle time-driver. The minimal delay cannot be lower than a few ms. Don't you if you set time frequency too high. The default value 1_000_000 hz can be too high for 16-bit timer.

Yeah, I saw that.

I set the `embassy-time` freq to 32,768 Hz and I'm delaying only 500 ms in my minimal example.

This is what I see with TIM1 in use. Very weird, almost like no delays happen except every 4 delays / 2 seconds:

```
2.000000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.000030 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.000061 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.000091 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
4.000000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
4.000030 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
4.000061 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
4.000091 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
6.000000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
6.000030 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
6.000061 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
6.000091 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
```

When I select feature `time-driver-tim3` instead, I get the expected output.

```
0.500000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
1.000000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
1.500000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.000000 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
```

tcbennun:
I should mention this is specifically the STM32H503. Default RCC except no LSI/LSE selected.

embassy-stm32/build.rs
```diff
@@ -189,6 +189,7 @@ fn main() {
         Some("tim5") => "TIM5",
         Some("tim8") => "TIM8",
         Some("tim9") => "TIM9",
+        Some("tim11") => "TIM11",
```
eZioPan, line 192: ah, I made a mistake. I actually remove TIM11 from time-driver last time, since this not capable be a time-driver. TIM11 only have 1 CC channel, but time-driver require at least 2. Would you remove it, and also remove it from Cargo.toml?

plaes: Maybe add comment about TIM11 being unusable due to reasons mentioned above.

eZioPan:
> Very weird, almost like no delays happen except every 4 delays / 2 seconds:

hum... it is weird... I can't tell why for now... It looks like every delays actually happends, but output 4 delays at a time

tcbennun:
If I set `embassy-time` to 1 MHz, the delay happens each time, but wrongly:

```
0.524301 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
1.048589 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
1.507341 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.031629 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
2.555917 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
3.014669 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
3.538957 INFO  beep
└─ g2_leaf::____embassy_main_task::{async_fn#0} @ src/main.rs:235
```

Those are alternating time diffs of 0.524288 and 0.458752 seconds.

eZioPan:
I can confirm the same situation on STM32F413 with TIM1: on "tick-hz-32_768" it print 4 ticks at a time, on "tick-hz-1_000_000" tick isn't stable.

tcbennun:
I injected this into `RtcDriver::on_interrupt()`:

```rust
if sr.uif() && sr.ccif(0) {
    defmt::info!("Update & CC1");
} else if sr.uif() {
    defmt::info!("Update");
} else if sr.ccif(0) {
    defmt::info!("CC1");
}
```

I ran a test program with no delays awaited. With TIM3, you get the update and CC1 events alternating every half-period, as you'd expect. With TIM1, instead both the update and CC1 events fire at the same time, every whole period.

```
0.000000 INFO  Update & CC1
2.000000 INFO  Update & CC1
```
etc.

tcbennun:
If I omit the following call in `RtcDriver::init()`:
```rust
r.ccr(0).write(|w| w.set_ccr(0x8000));
```
then I get the same behaviour in TIM1 and TIM3.

I'll check the register mappings for TIM1.

tcbennun:
The register mappings are fine, as usual the reg offsets are the same (would be surprised if that was the issue).

There's gotta be something different about the default event generation.

eZioPan:
By the way, I have try `TIM_2CH`, `TIM_GP16` and `TIM_GP32`, they are all good.
Only timer left is `TIM_2CH_CMP`.

I also try CC trigger interrupt with PAC, and it's work as expect.

tcbennun:
@eZioPan and TIM1 is none of those, right - it's classified as `ADV`?

I think what's happening is, the CC interrupt is never firing. If I comment out
```rust
w.set_ccie(0, true);
```
it makes no difference and breaks TIM3 in a similar way.

eZioPan:
> TIM1 is none of those, right - it's classified as `ADV`?

yes, TIM1 is ADV.

I'm trying match stm32-data timer.yaml with RM again, let see if I can find some bug inside it.

tcbennun:
Ha. TIM1 has separate interrupts, not a global one like TIM2 and TIM3. The CC interrupt is generated on `TIM1_CC`. I'll give that a go.

eZioPan:
I also find a extra register in timer_v1.yaml and timer_l0.yaml, will remove it.

tcbennun:
right, I've had success with that.

@eZioPan please review! Do you foresee any trouble with having both the UP and CC interrupt sources handled by the same function?

tcbennun:
Unsure if this change will break TIM15/16/17, I think these were given advanced timer impls but they only have the global interrupt

embassy-stm32/src/time_driver.rs
```diff
@@ -150,6 +166,14 @@ foreach_interrupt! {
             DRIVER.on_interrupt()
         }
     };
+    (TIM20, timer, $block:ident, CC, $irq:ident) => {
+        #[cfg(time_driver_tim1)]
```
eZioPan, line 170: should be tim20

Dirbaio:
thanks! :rocket:

tcbennun:
> yeah, I can confirm it works on STM32F413VG, and output correct timing.
> 
> > Unsure if this change will break TIM15/16/17
> 
> but also STM32F413 doesn't have `2CH_CMP`, so I can't tell either.

@eZioPan Why are `GeneralPurpose2ChannelComplementaryInstance` and `AdvancedControlInstance` implemented for `1CH_CMP`, and `AdvancedControlInstance` for `2CH_CMP` in the first place, in `mod.rs`?

Dirbaio (approved):
thank you! :100:

eZioPan:
> Unsure if this change will break TIM15/16/17

I just get a stm32h562vi, and I can confirm time-driver on TIM15 (`2CH_CMP`) works perfectly. (And of course, TIM16/17 is not working since they only got 1 CC channel).

PR #2618, opened 2024-02-23
[embassy-stm32] G4 RCC refactor amendments and additions

barnabywalters:
* Added assertions for a variety of clock frequencies, based on the reference manual and stm32g474 datasheet. The family and numbers are consistent enough that I’m assuming these numbers will work for the other chips.
* Corrected value of pll1_q in set_clocks call, added pll1_r value

The G4 family supports different sets of frequencies at different voltage scaling values. The current `max` module accepts the absolute min/max values (generally in range1 boost mode) and leaves correct frequency selection in other voltage scaling modes up to the user.


Dirbaio (approved):
thanks!

PR #2619, opened 2024-02-23
STM32 Crypto Accelerator

caleb-garrett:
Adds support for the CRYP peripheral in all STM32s. This does not include support for any implementation of the AES peripheral, which is functionally a subset of the CRYP IP but is an entirely different implementation. (That said, I believe the design of these two IP are close enough that AES implementations could share this API.)

Only a blocking mode is implemented initially. I anticipate this mode is likely to be more commonly used than an async DMA implementation since most crypto traits, such as the Rust Crypto AEAD traits, are not async. All supported cipher modes are implemented. 

The API is structured differently than a typical implementation of these algorithms to support suspend/resume functionality and incremental computation. This allows the application to feed payload and associated data for encryption/decryption on a block-by-block basis. This is useful in a microcontroller, where encrypting or decrypting a large stream of data such as a large file or secure firmware update image is not practical or possible to perform in one go. It is also useful for a concurrent send/receive scenario, where an application is simultaneously sending and receiving packetized data that must be encrypted and decrypted simultaneously.

For ciphers that do not support partial blocks (ECB and CBC modes), no padding scheme (such as PKCS#7) is implemented. Thus, only complete blocks may be supplied in these modes. The application must manage padding in these modes to ensure all blocks supplied to the processor are complete, otherwise a panic will occur. Panics due to malformed parameters are largely avoided through the use of traits that enforce key, IV, and tag length constraints.

An example for the STM32F7 series is included that demonstrates use of this API for the AES-GCM-128 cipher. In my testing, this example encryption/decryption ran >10x faster with hardware acceleration compared to the aes-gcm crate implementation!

This is a work in progress. All algorithms have been trivially verified but tests still need to be added.


caleb-garrett:
CI Build issue is apparently due to an STM32H7 RCC register naming issue. Addressed in a PR here: https://github.com/embassy-rs/stm32-data/pull/411

Dirbaio (approved):
thank you! :partying_face:

PR #2621, opened 2024-02-23
RP2040: Introduce an alternative "yielding flush" option

### LINKED ISSUE #2464: embassy-rp: UART flush isn't flushy enough
The blocking uart's blocking+async flush only waits until the buffered data is enqueued, NOT until the data is all out on the wire:

https://github.com/embassy-rs/embassy/blob/871d82de7165781ca6164de72664c9a6a90b3e66/embassy-rp/src/uart/buffered.rs#L419-L427

Instead, you also need to busy loop on the `uart.busy()` flag that reports whether the transmitter is active. I do that here:

https://github.com/jamesmunns/erdnuss-pub/blob/3d98a55aa7a90141a525e4443c927400f8516dd2/source/rp2040/src/lib.rs#L74-L77

There is no interrupt for this state. AFAICT, you have to busy poll from the time of interrupt (when DMA has completed and all bytes enqueued), maybe past the interrupt when the FIFO falls low again, and then busy loop until the transmitter is idle.

This is not compliant with `embedded-io`'s `flush` guarantees. For blocking, this is NBD, we can just busy loop. It's annoying for async tho.

jamesmunns:
This is a straightforward and "least bad" approach to covering the fact that the "done" interrupt only gets us as far as "last data is enqueued".

Closes #2464

### DIFF
--- a/embassy-rp/src/uart/buffered.rs
+++ b/embassy-rp/src/uart/buffered.rs
@@ -378,7 +378,9 @@ impl<'d, T: Instance> BufferedUartTx<'d, T> {
         })
     }
 
-    fn flush() -> impl Future<Output = Result<(), Error>> {
+    /// Flush until all bytes are enqueued, which is as far as we can get with
+    /// actual interrupt notifications.
+    fn flush_to_enqueued() -> impl Future<Output = Result<(), Error>> {
         poll_fn(move |cx| {
             let state = T::buffered_state();
             if !state.tx_buf.is_empty() {
@@ -421,8 +423,17 @@ impl<'d, T: Instance> BufferedUartTx<'d, T> {
         loop {
             let state = T::buffered_state();
             if state.tx_buf.is_empty() {
-                return Ok(());
+                break;
             }
+        }
+        while self.busy() {}
+        Ok(())
+    }
+
+    /// Flush until the sender is no longer busy, using a yield-spinloop
+    pub async fn yielding_busyloop(&self) {
+        while self.busy() {
+            yield_now().await;
         }
     }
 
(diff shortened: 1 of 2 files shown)


embassy-rp/src/uart/mod.rs
```diff
@@ -187,9 +188,21 @@ impl<'d, T: Instance, M: Mode> UartTx<'d, T, M> {
     pub fn blocking_flush(&mut self) -> Result<(), Error> {
         let r = T::regs();
         while !r.uartfr().read().txfe() {}
+        while self.busy() {}
         Ok(())
     }
 
+    /// Flush the UART TX, doing a yield-spinloop until done
+    pub async fn yielding_flush(&self) {
```
Dirbaio, line 196: this should be `pub async fn flush()` for consistency (other async fns aren't prefixed like `async_` or `yielding_` , only blocking is prefixed with `blocking_`.)

embassy-rp/src/uart/buffered.rs
```diff
@@ -421,9 +423,18 @@ impl<'d, T: Instance> BufferedUartTx<'d, T> {
...
-                return Ok(());
+                break;
             }
         }
+        while self.busy() {}
+        Ok(())
+    }
+
+    /// Flush until the sender is no longer busy, using a yield-spinloop
+    pub async fn yielding_busyloop(&self) {
```
Dirbaio, line 434: every time you do `flush` you also want to do this. I think it'd make sense to make it part of `fn flush()` instead of remembering to call this at every `flush()` call.

Dirbaio:
@jamesmunns ping :upside_down_face:

jamesmunns:
@Dirbaio sorry haven't gotten back to this, the CI failures surprised me. I'll be AFK for a week, feel free to close, or I can move back to WIP/draft status.

PR #2622, opened 2024-02-23
Fix CCE and Init set to false in can fd peripheral.rs

andtsa:
Configuration Change Enable set to false before Init set to false, fixed.
Previously when trying to configure CAN, process would get stuck in leaving init mode (I imagine) because 
``self.regs.cccr().modify(|w| w.set_init(false));`` didn't work after ``self.regs.cccr().modify(|w| w.set_cce(false));`` was run.

### DIFF
--- a/embassy-stm32/src/can/fd/peripheral.rs
+++ b/embassy-stm32/src/can/fd/peripheral.rs
@@ -475,9 +475,9 @@ impl Registers {
     #[inline]
     fn leave_init_mode(&mut self, config: FdCanConfig) {
         self.apply_config(config);
-
-        self.regs.cccr().modify(|w| w.set_cce(false));
+        
         self.regs.cccr().modify(|w| w.set_init(false));
+        self.regs.cccr().modify(|w| w.set_cce(false));
         while self.regs.cccr().read().init() == true {}
     }
 


cschuhen:
So, here is the code from the FDCAN crate which is what was being used before the merge yesterday:
`    #[inline]
    fn leave_init_mode(&mut self) {
        self.apply_config(self.control.config);

        let can = self.registers();
        can.cccr.modify(|_, w| w.cce().clear_bit());
        can.cccr.modify(|_, w| w.init().clear_bit());
        while can.cccr.read().init().bit_is_set() {}
    }
`
@andtsa, presumably you had the same issue before the merge yesterday? Did you happen to try on older branch state?

cschuhen:
Here is text from the STMH742 etc RM:

_Software initialization is started by setting INIT bit in FDCAN_CCCR register, either by
software or by a hardware reset, or by going Bus_Off. While INIT bit in FDCAN_CCCR
register is set, message transfer from and to the CAN bus is stopped, the status of the CAN
bus output FDCAN_TX is recessive (high). The counters of the error management logic
(EML) are unchanged. Setting INIT bit in FDCAN_CCCR does not change any configuration
register. Clearing INIT bit in FDCAN_CCCR finishes the software initialization. Afterwards
the bit stream processor (BSP) synchronizes itself to the data transfer on the CAN bus by
waiting for the occurrence of a sequence of 11 consecutive recessive bits (Bus_Idle) before
it can take part in bus activities and start the message transfer.
Access to the FDCAN configuration registers is only enabled when both INIT bit in
FDCAN_CCCR register and CCE bit in FDCAN_CCCR register are set.
CCE bit in FDCAN_CCCR register can only be set/cleared while INIT bit in FDCAN_CCCR
is set. CCE bit in FDCAN_CCCR register is automatically cleared when INIT bit in
FDCAN_CCCR is cleared._

The last paragraph is key. To me, this implies:

1. The old behaviour was correct w.r.t. this particular RM and set of chips
2. For this particular RM and set of chips, we don't need to clear CCE at all, it should happen automatically.

cschuhen:
@Dirbaio, @andtsa I also checked the H723 and G4 RM's and saw the same description.

cschuhen:
So, I tested with both my H743 and G431. For both:

This works:
`        self.regs.cccr().modify(|w| w.set_cce(false));
        self.regs.cccr().modify(|w| w.set_init(false));`

This also works:
`        //self.regs.cccr().modify(|w| w.set_cce(false));
        self.regs.cccr().modify(|w| w.set_init(false));`

This hangs:
`        self.regs.cccr().modify(|w| w.set_init(false));
        self.regs.cccr().modify(|w| w.set_cce(false));`

So, I can't agree with this PR based on my observations.

I was going to say, that I can't see how the HIL tests didn't fail (only rustfmt did) however it seems that they didn't even run.

Dirbaio:
bender run

Dirbaio:
> however it seems that they didn't even run.

ah yep :sweat_smile: . indeed they don't pass

Dirbaio:
going to close this since it seems it's not the right fix

PR #2631, opened 2024-02-27
stm32: can: fd: fix SID read/write and BRS setting for TX

tcbennun:
- standard ID is stored in ID[28:18] for TX and RX elements
- set BRS bit correctly in TX elements (only set it if transmitting a FD frame *and* the user specified BRS)


cschuhen:
Thanks Torin, changes look good to me.

PR #2632, opened 2024-02-27
stm32: can: fd: allow TX buffers in FIFO mode

tcbennun:
Uses enum in pac introduced here: https://github.com/embassy-rs/stm32-data/pull/429

Need to figure out what this breaks. First glance suggests `BufferedCan` should work fine with it at least

### DIFF
--- a/embassy-stm32/src/can/fd/config.rs
+++ b/embassy-stm32/src/can/fd/config.rs
@@ -287,6 +287,24 @@ impl Default for GlobalFilter {
     }
 }
 
+/// TX buffer operation mode
+#[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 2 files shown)


tcbennun:
Made a behaviour change: when in FIFO mode and attempting to transmit when buffers are full, we abort instead of dropping a pending frame.

Opening for review, but the build will depend on a metapac update

cschuhen:
There seems to be an ordering issue here. That register is configured in into_config_mode which is called on construction. Unless I miss something, you can't change any config until after that so the register is not reconfigured with the new setting from config. Need to move this so that it happens when start is called.

tcbennun:
Closing in favour of successor #2647

Issue #2620, opened 2024-02-23
CAN examples for stm32h7 not working 


FranciscoAmaro24:
I've been trying to run the can example provided in the stm32h7 folder but the program seems to constantly get stuck on the CAN initialization line.
What could be the issue?

maiaherringfish:
I am having similar issues. Will documentation be released at some point for the FDCAN HAL module to aid debugging?

FranciscoAmaro24:
@maiaherringfish If you get it to work it would be amazing if you could share the solution :D.
We are in desperate need of this module in our application.

maiaherringfish:
I think I am closer to getting it working today. Have you seen the new example they released last week? Unlike the one they had before, I'm able to configure this one (rather than it hanging on initialization). However, it seems to only configure without issues if I have it set up with an HSE--otherwise the code hardfaults. Working on diagnosing this issue further

FranciscoAmaro24:
Hey
We kinda got it to work on internal loop back mode but on the normal can
configuration we were still having issues.
We have been fucking around a lot and hope by tomorrow or the day after it
works.
We’re also in contact with the guy that made can on embassy but he is
Australian so the time zones are quite different.

maiaherringfish:
That is great to hear! Can I ask if you are using the HSE or a different clock? This seems to be the cause of some issues for me

FranciscoAmaro24:
We tried but ended up falling back on the example set up of HSE

maiaherringfish:
I think my teammate and I found the issue. We noticed there is no logic in rcc/h.rs to set the fdcansel field of the d2ccip1r register, which prevents a clock from being selected correctly and causes a default to HSE.

I still have to verify whether or not the FDCAN frequency is then being identified correctly based on the clock but this should be the start of a solid fix to this issue. When I tested it I no longer received hardfaults from accessing FDCAN::frequency() on config.

cschuhen:
Please met me know how you both get on with this. Also, b.t.w. exactly which chips are you using? And what boards, in particular, what is connected to HSE. Keep in mind that it is recommended that CAN should always operate with a crystal or similar source for timing. So you want to use a clock that is either HSE or derived from it probably.

FranciscoAmaro24:
For us we are using a nucleo the stm32h743zi6 but we want to make it work
also on our custom PCB that uses the same chip.
Do you think would be a problem?

cschuhen:
[...]

PR #2634, opened 2024-02-27
adding FDCANSEL logic for STM32H7x

maiaherringfish:
Currently, using any clock other than the HSE with FDCAN causes hardfaults due to the FDCANSEL register not being set correctly.
See [Issue 2620](https://github.com/embassy-rs/embassy/issues/2620). 

This PR is not complete, as logic may still be needed to set the frequency correctly.

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -557,6 +557,9 @@ pub(crate) unsafe fn init(config: Config) {
         RCC.d3ccipr().modify(|w| {
             w.set_adcsel(config.adc_clock_source);
         });
+        RCC.d2ccip1r().modify(|w| {
+            w.set_fdcansel(config.fdcan_clock_source);
+        });
     }
     #[cfg(stm32h5)]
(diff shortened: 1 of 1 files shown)


Dirbaio:
oh so this was missing for H7. Makes sense that it didn't work :sweat_smile: 

(FYI these `*_clock_source` fields will be replaced with an autogenerated `config.rcc.mux` soon in https://github.com/embassy-rs/embassy/pull/2626 . That'll still take a few days so I'm OK with merging this so it works now without waiting)

Dirbaio:
> This PR is not complete, as logic may still be needed to set the frequency correctly.

have you tested whether this is working? In theory it should work out as build.rs is already generating code that fetches the correct frequency according to the selected clock in FDCANSEL, all that's missing is an API to actually set it.

maiaherringfish:
Sorry for the slow reply! yes it does work, I can see the clock selection being updated in the appropriate register. Today I will hopefully add code to set the frequency so it can be accessed

PR #2635, opened 2024-02-27
stm32: can: fd: some refactoring

tcbennun:
Trying to untangle some duplicated code, etc.


Dirbaio (approved):

cschuhen:
Thanks @tcbennun, combining those read and write workhorse functions again was on my list. I'm still just a bit slow when it comes to generics :).

PR #2637, opened 2024-02-28
Buffer is not big enough for FD frames.

cschuhen:
https://github.com/embassy-rs/embassy/pull/2635 Did some nice refactoring however the combined reader function only declared an 8 byte data buffer as per ClassicCan instead of the larger 64 byte that is needed in order to receive full-sized CAN FD frames.


tcbennun:
Oops - thank you!

cschuhen:
No problems... My FD capable USB CAN adaptors just arrived today so I got to test them out.

PR #2647, opened 2024-03-02
stm32: can: fd: allow TX buffers in FIFO mode

cschuhen:
Uses enum in pac introduced here: embassy-rs/stm32-data#429

Based on : https://github.com/embassy-rs/embassy/pull/2632
with added fix.

### DIFF
--- a/embassy-stm32/src/can/fd/config.rs
+++ b/embassy-stm32/src/can/fd/config.rs
@@ -287,6 +287,24 @@ impl Default for GlobalFilter {
     }
 }
 
+/// TX buffer operation mode
+#[derive(Clone, Copy, Debug)]
+pub enum TxBufferMode {
+    /// TX FIFO operation
+    Fifo,
+    /// TX queue operation
+    Queue,
+}
+
+impl From<TxBufferMode> for crate::pac::can::vals::Tfqm {
+    fn from(value: TxBufferMode) -> Self {
+        match value {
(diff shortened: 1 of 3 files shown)


Dirbaio (approved):
thanks! :rocket:

cschuhen:
I wondered what would happen if I sneaked in another change :)

cschuhen:
The words "Queue" and "FIFO" are too close to synonyms for my liking.

Dirbaio (approved):
:rofl:

Dirbaio:
seems  h7 hangs

cschuhen:
Hmm, strange. I've got to head out, I'll have a look later.

cschuhen:
@tcbennun FYI. Any comments on this? 

The patch is significant in length but really it is just pushing a bunch of things from happening at the end of the initial construction to the start of the start method. This is done to preserve order and is a good change in general I think for the future. For example if we want to allow more customisation later for things like H7 fdcanram.

tcbennun:
> @tcbennun FYI. Any comments on this?
> 
> The patch is significant in length but really it is just pushing a bunch of things from happening at the end of the initial construction to the start of the start method. This is done to preserve order and is a good change in general I think for the future. For example if we want to allow more customisation later for things like H7 fdcanram.

LGTM if it builds and runs alright (I can test it Monday), thanks for the fixes!

PR #2648, opened 2024-03-02
Add constructor for dynamic channel

peterkrull:
I think that doing
```rust
let s = c.dyn_sender();
```
 is a bit more ergonomic than
```rust
let s: DynamicSender<'_, _> = c.sender().into();
```


Dirbaio (approved):
makes sense! thank you :)

Issue #2503, opened 2024-01-31
OnceLock-like type


robertbastian:
I'm looking for a type like `std`'s [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html). In an async world it would have an API like 
* `pub const fn new() -> Self`
* `pub async fn get(&self) -> &T`
* `pub fn set(&self, T) -> Result<(), T>`

That is, it can be placed in a `static`, everyone waits for someone to put a value value in, and then everyone gets a shared reference. Any further calls to `get` resolve immediately, any further calls to `set` fail.

Does this exist? I found `Signal`, but its `wait` returns `T` by value, so it must remove the value  (the doc explanation is cyclic but the type is pretty clear). Can this be trivially assembled from existing types in embassy? Is this impossible to do?

asasine:
I think this is completed as of #2651.

PR #2651, opened 2024-03-03
Add `OnceLock` sync primitive

peterkrull:
This covers #2503

### DIFF
--- a/embassy-sync/src/once_lock.rs
+++ b/embassy-sync/src/once_lock.rs
@@ -1,0 +1,237 @@
+//! Syncronization primitive for initializing a value once, allowing others to await a reference to the value.
+
+use core::cell::Cell;
+use core::future::poll_fn;
+use core::mem::MaybeUninit;
+use core::sync::atomic::{AtomicBool, Ordering};
+use core::task::Poll;
+
+/// The `OnceLock` is a synchronization primitive that allows for
+/// initializing a value once, and allowing others to `.await` a
+/// reference to the value. This is useful for lazy initialization of
+/// a static value.
+///
+/// **Note**: this implementation uses a busy loop to poll the value,
+/// which is not as efficient as registering a dedicated `Waker`.
+/// However, the if the usecase for is to initialize a static variable
+/// relatively early in the program life cycle, it should be fine.
+///
+/// # Example
+/// ```
+/// use futures_executor::block_on;
+/// use embassy_sync::once_lock::OnceLock;
+///
+/// // Define a static value that will be lazily initialized
+/// static VALUE: OnceLock<u32> = OnceLock::new();
+///
+/// let f = async {
+///
+/// // Initialize the value
+/// let reference = VALUE.get_or_init(|| 20);
+/// assert_eq!(reference, &20);
+///
+/// // Wait for the value to be initialized
+/// // and get a static reference it
+/// assert_eq!(VALUE.get().await, &20);
+/// 
+/// };
+/// block_on(f)
+/// ```
+pub struct OnceLock<T> {
+    init: AtomicBool,
+    data: Cell<MaybeUninit<T>>,
+}
+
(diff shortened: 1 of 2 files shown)


peterkrull:
Thanks for taking a look. I do not believe so, since with the signal there is only ever one recipient of a "signaled" signal. And the recipient gets the value itself, not a reference to it. So if you wanted multiple recipients of a reference, you would need a task to keep filling the signal with a clone of that value. The OnceLock holds the value, and guarantees that a reference to it stays valid, by not allowing the value to be overwritten.

A common use case would be that the OnceLock is statically allocated, one task is responsible for initializing it, and multiple other tasks can either await (or try_get) a reference to the underlying value. Alternatively, this can be done with something like #2568, which is essentially a multi-recipient signal, where the value sent is an already static reference from e.g. a StaticCell.

lulf:
@peterkrull That makes sense. Still, I wonder when you'd use this vs. a StaticCell in an embedded application. Do you have a real world use case you could share on how you're using it? (Multicore?)

peterkrull:
I think it is overall more ergonomic than using a StaticCell for the same purpose, since the reference can be obtained from anywhere in the program, without needing to keep it defined in task and function interfaces. Also tasks can await the value to be set, or try_get if, for example, the value is only relevant once it is initialized. Eg:

```rust
static VALUE: OnceLock<u32> = OnceLock::new();

#[embassy_executor::main]
    spawner.must_spawn(load_config(p.FLASH));
    spawner.must_spawn(some_task());
    spawner.must_spawn(other_task());
}

#[embassy_executor::task]
async fn config_task(flash: FLASH) {
    let storage = Flash::new(flash) // example, ignore specifics here
    // It might take some time to search for the parameter in flash,
    // some_task can simultaneously complete its setup process.
    let value = storage.load( 0x10000..0x20000, 127 ).async; 

    // Initialize static values, assert so we are for sure the only ones doing it
    assert!(VALUE.init(value).is_ok());
}

#[embassy_executor::task]
async fn some_task() { // We do not need to have initialized and passed the reference at this point
    // Here we can do some setup that does not yet need the value

    // Get the value reference as late as possible
    let value_ref: &u32 = VALUE.get().await;
    loop { /* here we may use the reference */ }
}

#[embassy_executor::task]
async fn other_task() {
    loop { 
        if let Some(value_ref) = VALUE.try_get() {
            // In this task we only care about the value if it is initialized, not otherwise
        }
    }
}
```
It makes it easy to modify some task to get the reference, without needing to pass it through whichever tasks spawned the task in the first place. Also, it should just work on multi-core since a critical section is used to set the value. And since it uses an AtomicBool to signal whether the value is set, we do not need to make a critical section to get the value, which should make `if let Some(value_ref) = VALUE.try_get() {...}` fine even when performance is important.

peterkrull:
But to be honest I am not currently using this in my own project because most of my parameters and configuration need to be modifiable during the life of the program. I also mostly made it as an exercise after I saw that other issue requesting it, and have not looked much at it since. So arguably the PR could be marked as draft until someone more enthusiastic about it comes through.

lulf:
Thanks for the update, I think it makes sense, and I think I could make use of it.

embassy-sync/src/once_lock.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// };
+/// block_on(f)
+/// ```
+pub struct OnceLock<T> {
+    init: AtomicBool,
+    data: Cell<MaybeUninit<T>>,
+}
+
+unsafe impl<T> Sync for OnceLock<T> {}
```
elenaf9, line 45: Hi, just a question for my own understanding:

Why is it safe to implement `Sync` here even when `&T` isn't `Send`?
The `std::sync::OnceLock` it only implements `Sync` for `T: Send + Sync`.

PR #2652, opened 2024-03-03
STM32 USB Type-C/USB Power Delivery Interface (UCPD)

timokroeger:
Draft until my [hardware for testing arrives](https://www.cnx-software.com/2024/03/01/weact-stm32g4-tiny-board-stm32g4-mixed-signal-microcontroller/).

- [x] Type-C Pull-Up/Pull-Down resistor configuration
- [x] Type-C CC voltage detection
- [x] PD message reception
- [x] PD message transmission
- [x] Check for received hard reset in `transmit()` and `receive()`
- [x] Method to send a hard reset
- [x] Dead-battery resistor disable for other families than G4, see #2683
- [x] Allow to change and monitor CC resistors and voltages when PD phy is enabled (required for PD 3.0 collision detection, PD2.0 does not need this)
- [x] ~~Support for SOP' and SOP'' cable messages~~ Follow up PR
- [x] Testing!

### DIFF
--- a/embassy-stm32/src/ucpd.rs
+++ b/embassy-stm32/src/ucpd.rs
@@ -22,7 +22,7 @@ use embassy_hal_internal::drop::OnDrop;
 use embassy_hal_internal::{into_ref, Peripheral, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
 
-use crate::dma::AnyChannel;
+use crate::dma::{AnyChannel, Request, Transfer, TransferOptions};
 use crate::interrupt;
 use crate::pac::ucpd::vals::{Anamode, Ccenable, PscUsbpdclk};
 pub use crate::pac::ucpd::vals::{Phyccsel as CcSel, TypecVstateCc as CcVState};
@@ -99,7 +99,7 @@ impl<'d, T: Instance> Ucpd<'d, T> {
             // 1.75us * 17 = ~30us
             w.set_ifrgap(17 - 1);
 
-            // TODO: Only receive SOP messages
+            // TODO: Currently only SOP messages are supported.
             w.set_rxordseten(0x1);
 
             // Enable DMA and the peripheral
@@ -186,6 +186,14 @@ impl<'d, T: Instance> Ucpd<'d, T> {
         tx_dma: impl Peripheral<P = impl TxDma<T>> + 'd,
         cc_sel: CcSel,
     ) -> (PdRx<'_, T>, PdTx<'_, T>) {
+        let r = T::REGS;
+
+        // Enable the receiver on one of the two CC lines.
+        r.cr().modify(|w| {
+            w.set_phyccsel(cc_sel);
+            w.set_phyrxen(true);
+        });
+
         into_ref!(rx_dma, tx_dma);
         let rx_dma_req = rx_dma.request();
         let tx_dma_req = tx_dma.request();
@@ -204,6 +212,16 @@ impl<'d, T: Instance> Ucpd<'d, T> {
     }
 }
 
+/// Receive Error.
+#[derive(Debug, Clone, Copy)]
+pub enum RxError {
+    /// Incorrect CRC or truncated message (a line becoming static before EOP is met).
+    Crc,
+
+    /// Provided buffer was too small for the received message.
+    Overrun,
+}
+
 /// Power Delivery (PD) Receiver.
 pub struct PdRx<'d, T: Instance> {
     _ucpd: &'d Ucpd<'d, T>,
@@ -214,6 +232,79 @@ pub struct PdRx<'d, T: Instance> {
 impl<'d, T: Instance> Drop for PdRx<'d, T> {
     fn drop(&mut self) {
         T::REGS.cr().modify(|w| w.set_phyrxen(false));
+    }
+}
+
+impl<'d, T: Instance> PdRx<'d, T> {
+    /// Receives a PD message into the provided buffer.
+    ///
+    /// Returns the number of received bytes or an error.
+    pub async fn receive(&mut self, buf: &mut [u8]) -> Result<usize, RxError> {
+        let r = T::REGS;
+
+        // Check if a message is already being received. If yes, wait until its
+        // done, ignore errors and try to receive the next message.
+        if r.sr().read().rxorddet() {
+            let _ = self.wait_rx_done().await;
+            r.rxdr().read(); // Clear the RX buffer.
+        }
+
+        // Keep the DMA transfer alive so its drop code does not stop it right away.
+        let dma = unsafe {
+            // Disable the DMA complete interrupt because the end of packet is
+            // signaled by the UCPD receiver. When the DMA buffer is too short
+            // DMA stops by itself and the overrun RXOVR flag of UCPD is set.
+            let mut transfer_options = TransferOptions::default();
+            transfer_options.complete_transfer_ir = false;
+
+            Transfer::new_read(
+                &self.dma_ch,
+                self.dma_req,
+                r.rxdr().as_ptr() as *mut u8,
+                buf,
+                transfer_options,
+            )
+        };
+
+        self.wait_rx_done().await?;
+
+        // Make sure the the last byte to byte was fetched by DMA.
+        while r.sr().read().rxne() {
+            if dma.get_remaining_transfers() == 0 {
+                return Err(RxError::Overrun);
+            }
+        }
+
+        Ok(r.rx_payszr().read().rxpaysz().into())
+    }
+
+    async fn wait_rx_done(&self) -> Result<(), RxError> {
+        poll_fn(|cx| {
+            let r = T::REGS;
+            let sr = r.sr().read();
+            if sr.rxmsgend() {
+                let ret = if sr.rxovr() {
+                    Err(RxError::Overrun)
+                } else if sr.rxerr() {
+                    Err(RxError::Crc)
+                } else {
+                    Ok(())
+                };
+                // Message received, clear interrupt flags.
+                r.icr().write(|w| {
+                    w.set_rxorddetcf(true);
+                    w.set_rxovrcf(true);
+                    w.set_rxmsgendcf(true);
+                });
+                Poll::Ready(ret)
+            } else {
+                // Enable receiver interrupt.
+                T::waker().register(cx.waker());
+                r.imr().modify(|w| w.set_rxmsgendie(true));
+                Poll::Pending
+            }
+        })
+        .await
     }
 }
 
@@ -239,6 +330,10 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
                 w.set_typecevt1cf(true);
                 w.set_typecevt2cf(true);
             });
+        }
+
+        if sr.rxmsgend() {
+            r.imr().modify(|w| w.set_rxmsgendie(false));
         }
 
         // Wake the task to clear and re-enabled interrupts.
--- a/embassy-stm32/src/ucpd.rs
+++ b/embassy-stm32/src/ucpd.rs
@@ -124,6 +124,14 @@ impl<'d, T: Instance> Ucpd<'d, T> {
             // 1.75us * 17 = ~30us
             w.set_ifrgap(17 - 1);
 
+            // TODO: Currently only hard reset and SOP messages can be received.
+            // UNDOCUMENTED: This register can only be written while UCPDEN=0 (found by testing).
+            w.set_rxordseten(0b1001);
+
+            // Enable DMA
+            w.set_txdmaen(true);
+            w.set_rxdmaen(true);
+
             w.set_ucpden(true);
         });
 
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct Ucpd<'d, T: Instance> {
+    _peri: PeripheralRef<'d, T>,
+}
+
+impl<'d, T: Instance> Ucpd<'d, T> {
+    /// Creates a new UCPD driver instance.
+    pub fn new(
+        peri: impl Peripheral<P = T> + 'd,
+        _cc1: impl Peripheral<P = impl Cc1Pin<T>> + 'd,
+        _cc2: impl Peripheral<P = impl Cc2Pin<T>> + 'd,
```
ExplodingWaffle, line 61: best to get the opinion of an actual reviewer, but could it be argued that there should be 4 constructors (cc1, cc2) (cc1, cc2, dbcc1, dbcc2), (cc1), (cc1, dbcc1)  (nb. some packages don't even break out cc2. only cc1 is needed for a "direct attach" device with a plug instead of a receptacle) and then a new_inner like spi? what's the "rule" on passing in these "additonal function" (as opposed to alternate function which need setting as alternate function in gpio MODER + AFR) that don't really need anything doing to them afaik? or maybe they need setting as analog?

ExplodingWaffle (commented):
hope you don't mind the drive by comments :)
had [a go](https://github.com/ExplodingWaffle/embassy/blob/main/embassy-stm32/src/ucpd.rs) at implementing this in the past but got distracted and demotivated by the general cursedness of USB PD and UCPD, so good luck. feel free to pick my brain on this because i've thought a lot about it (some might say too much thinking and not enough coding :D)

timokroeger:
> hope you don't mind the drive by comments :) had [a go](https://github.com/ExplodingWaffle/embassy/blob/main/embassy-stm32/src/ucpd.rs) at implementing this in the past but got distracted and demotivated by the general cursedness of USB PD and UCPD, so good luck. feel free to pick my brain on this because i've thought a lot about it (some might say too much thinking and not enough coding :D)

Every comment is welcome! You have implemented a lot already, gonna take you code as reference going forward.

examples/stm32g4/src/bin/usb_c_pd.rs
```diff
@@ -0,0 +1,62 @@
...
+            (_, CcVState::LOWEST) => CableOrientation::Normal,  // CC1 connected
+            (CcVState::LOWEST, _) => CableOrientation::Flipped, // CC2 connected
+            _ => CableOrientation::DebugAccessoryMode,          // Both connected (special cable)
+        };
+    }
+}
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    // TODO: Disable DBCC pin functionality by default but have flag in the config to keep it enabled when required.
```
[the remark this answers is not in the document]
timokroeger: Unexpected pull downs for a functionality that does not exist. What a footgun! Thanks for the hint.

embassy-stm32/src/ucpd.rs
timokroeger: In my first version I included the dbccX (dead battery input) pins as part of the constructor but decided against that later.
From intuition I would expect the driver to reconfigure the pins I pass it in the constructor as required by the application (e.g. set up pull-up and AF mode).
With CC pins thats what we do: Configure pull resistors as required.
With the DBCC I’m not sure: We actually do not change any behavior on them. The pins are already configured as such before even passing them to the constructor.

timokroeger: > nb. some packages don't even break out cc2. only cc1 is needed for a "direct attach" device with a plug instead of a receptacle
Really? ST never fails to surprise! I would expect the majority of embassy applications to use a usb-c receptacly so would like to restrict the first implementation to that and keep the CC1 only plug for a follow up PR.

embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,581 @@
...
+        // Enable the receiver on one of the two CC lines.
+        r.cr().modify(|w| {
+            w.set_phyccsel(cc_sel);
+            w.set_phyrxen(true);
+        });
+
+        // Enable hard reset receive interrupt.
+        r.imr().modify(|w| w.set_rxhrstdetie(true));
+
+        // Both parts must be dropped before the peripheral can be disabled.
```
ExplodingWaffle, line 136: wrt refcounting: I think it's possible to reuse what USART already does for splitting: not 100% sure how this works, plus it's been recently refactored, but i thiiink: with [this](https://github.com/embassy-rs/embassy/blob/b2d236ee390081ec6aeef1a27da06098f9febbf9/embassy-stm32/build.rs#L392), `fn enable_and_reset()` and `fn disable()` from `trait RccPeripheral` can do all the refcounting for you, so all you'd need to do in `Drop` impls is `T::disable()`

wrt splitting in general: seems like a good idea. hard to wrap my head around what exactly needs to be seperate from what. naively, CC is just one line, the communication is half duplex, so why would anything need to be seperate at all? but then: it's possible (is it desirable? necessary?) to "preempt" a packet transmit with a hard reset (leading to TXMSGABT? though other parts of the RM imply there is no interrupt for this) or an incoming message (TXMSGDISC. RM mentions "concurrent recieve") that you might not want to miss the content of, so wouldn't those need to be separate futures, maybe running in separate tasks/threads/priorities even? a message being recieved could likewise get interrupted by a hard reset reception, but this seems like it maps pretty cleanly onto returning an error from the rx. (maybe this is why the interrupts are called HRSTSENT not TXHRST and RXHRSTDET not HRSTDET?) and obviously you still need to be detecting a type-c detach while doing PD stuff.

food for thought. probably just my overthinking again >:) I imagine the best way to figure out what exactly is needed is actually implementing the PD negotiations, but that's a fair bit of work

embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,581 @@
...
+        }
+
+        // Clear the TX interrupt flags.
+        T::REGS.icr().write(|w| {
+            w.set_txmsgdisccf(true);
+            w.set_txmsgsentcf(true);
+        });
+
+        // Start the DMA and let it do its thing in the background.
+        let _dma = unsafe {
```
ExplodingWaffle, line 391: dropping a Transfer stops it, and here it's dropped immediately because it's unused. it needs to be kept around either by selecting over it like in usart or with an explicit `drop()` after done.

embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,581 @@
...
+        r.cr().modify(|w| {
+            w.set_txmode(Txmode::PACKET);
+            w.set_txsend(true);
+        });
+
+        self.wait_tx_done().await
+    }
+
+    async fn wait_tx_done(&self) -> Result<(), TxError> {
+        let _on_drop = OnDrop::new(|| self.enable_tx_interrupts(false));
```
ExplodingWaffle, line 412: also ought to have an explicit drop

embassy-stm32/src/ucpd.rs
timokroeger: Every variable even if unused will be dropped at the end of the scope, so that should be fine. Maybe you have the special `let _ = ...` pattern in mind where drop executes immediately.

Currently the DMA is started in the constructor not on first poll so it should work. I need to clarify with @Dirbaio if this is a contract we can rely on for DMA. Otherwise a `select` is required.

embassy-stm32/src/ucpd.rs
timokroeger: Should work as expected without explicit drop:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b599bc8e0061bcc73905091ac34d4be1

embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,581 @@
...
+}
+
+impl<'d, T: Instance> PdPhy<'d, T> {
+    /// Receives a PD message into the provided buffer.
+    ///
+    /// Returns the number of received bytes or an error.
+    pub async fn receive(&mut self, buf: &mut [u8]) -> Result<usize, RxError> {
+        let r = T::REGS;
+
+        // Check if a message is already being received. If yes, wait until its
```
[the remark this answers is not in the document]
timokroeger: Is it a use case required for implementing the PD spec? If yes, I fully agree it should be configurable or even made default.

embassy-stm32/src/ucpd.rs
```diff
@@ -0,0 +1,581 @@
...
+
+            // Enable DMA
+            w.set_txdmaen(true);
+            w.set_rxdmaen(true);
+        });
+
+        // Enable the receiver on one of the two CC lines.
+        r.cr().modify(|w| {
+            w.set_phyccsel(cc_sel);
+            w.set_phyrxen(true);
```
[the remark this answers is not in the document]
timokroeger: Good point. If detection of missed receive messages is not importannt (see previous comment) its better to enable the RX phy only on demand. Just like you did in your implementation.

embassy-stm32/src/ucpd.rs
timokroeger: Thanks for the refcounting pointers, I think the code you linked is for refcounting of RCC enable its. The "refcounting" bool here is for tracking enable of the peripheral itself.

Preemting a transmit package is possible with the current API:
1. Drop the ongoing transmit future (it will continue to run in the background, with trash data because the DMA stopped).
2. Send a hard reset: The peripheral interrupts the still running transmission

Regarding splitting you made me overthink this *again* and now I think it might be feasible to implement all of the PD spec without any split.
Pro: No refcounting for drop needed.
Con: Runtime selection of PD comminaciton pin required before TX or RX can be used -> Introduces a new error code. This error condition is currently nicely prevented by the typesystem.

>I imagine the best way to figure out what exactly is needed is actually implementing the PD negotiations, but that's a fair bit of work
Agree!

ExplodingWaffle: > Thanks for the refcounting pointers, I think the code you linked is for refcounting of RCC enable its. The "refcounting" bool here is for tracking enable of the peripheral itself.

hm. Thought these were the same, but now I’ve done even more thinking I’m less sure. I think the reason @Dirbaio generalised the ref counting is because some bits in the RCC registers are shared between several peripherals. Not 100% sure how “disabled” a peripheral is without its clock enabled, but my guess would be  “not fully”. Wonder if this means USART is wrong then, because the UE bit isn’t actually touched anywhere in any Drop impls? UCPD has more asynchronous stuff going on as well e.g. pullups 
If all the “functionality” is disabled, by e.g. CCENABLE and stuff when dropping, or not transmitting on your UART, which you can’t be while dropping it, is it fine? when creating a new one everything should be fixed by the reset, but what about between the clock enable and the reset?

embassy-stm32/src/ucpd.rs
ExplodingWaffle: > Maybe you have the special `let _ = ...` pattern in mind where drop executes immediately.

uh, yeah 😅 woops

embassy-stm32/src/ucpd.rs
timokroeger: Ohh, I misunderstood how the RCC ref-counting works. In any case the imlementation implements at least on call to `T::disable()`.
Let me try to disable as much as possible without internal ref-counting and then hope that turning of the clock shuts down the rest (even when not clearing UCPDEN).

ExplodingWaffle:
some thoughts (some might say overthinking :)) mostly about how dead battery could be handled in the driver I had while writing #2683:

g0 is weird: you need to write a strobe bit in SYSCFG every time you want the CC pin configuation to be applied. don't think this is a big issue but worth keeping in mind. 

according to the non-g0 RM:
image
you should initialise UCPD before disabling DB. makes sense. (wonder if this means DB takes precedence over any other config?) 

g0 only (reserved on others) has DBATTEN bit in UCPD CR:
image
worth noting that since it is in CR that means it can only be changed while UCPD is enabled, unlike the DBDIS bit hanging out in PWR on the other chips. that said, I this isn't much of a problem, since the strobe bit works while UCPD is disabled: 
image
 and if you need to re-enable DB, you can but you need to be careful to do it while UCPD is enabled. this is why I made ucpd_init it's own thing in my PR.

I can't think of many scenarios where you would want to re-enable DB: it isn't needed once UCPD is enabled (since you can set the sink pulldowns just fine with ANAMODE=1 and CCENABLE != 0, which is also required when the DBCC pins aren't wired up). maybe re-enabling DB if you disable UCPD (i.e drop the driver)?? probably keep a flag in the driver struct if DBCC pins are passed in. if they aren't, the user could be using them for something else (this is possible with some care: mentioned in [this forum post](https://community.st.com/t5/stm32-mcus-other-solutions/usb-pd-dead-battery-pin-usage-when-functioning-as-pd-source/m-p/299103/highlight/true#M2230) and the linked [AN5225](https://www.st.com/resource/en/application_note/an5225-introduction-to-usb-typec-power-delivery-for-stm32-mcus-and-mpus-stmicroelectronics.pdf), which I had forgotten about) which we should respect by not enabling them on drop.

Dirbaio (approved):
looking great. Thank you! :rocket:

PR #2658, opened 2024-03-04
stm32: fdcan: merge read impls, add setter for TxBufferMode, rename Queue -> Priority

tcbennun:
This is gonna be a draft for the minute as I want to figure something out.

Does it make sense to possibly return `BusError` when the user requests a read? If so, the current impl isn't quite right, because some of the error states are persistent (particularly the Error Active -> Error Passive -> Bus Off states). If we're gonna return `BusError` upon read, I think it should only be from the LEC (last error code), as its bit field resets upon read.

Other error statuses should be accessible through other means.

Suggestions welcome. @cschuhen @Dirbaio

### DIFF
--- a/embassy-stm32/src/can/fd/config.rs
+++ b/embassy-stm32/src/can/fd/config.rs
@@ -292,14 +292,14 @@ impl Default for GlobalFilter {
 pub enum TxBufferMode {
     /// TX FIFO operation - In this mode CAN frames are trasmitted strictly in write order.
     Fifo,
-    /// TX queue operation - In this mode CAN frames are transmitted according to CAN priority.
-    Queue,
+    /// TX priority queue operation - In this mode CAN frames are transmitted according to CAN priority.
+    Priority,
 }
 
 impl From<TxBufferMode> for crate::pac::can::vals::Tfqm {
     fn from(value: TxBufferMode) -> Self {
         match value {
-            TxBufferMode::Queue => Self::QUEUE,
+            TxBufferMode::Priority => Self::QUEUE,
             TxBufferMode::Fifo => Self::FIFO,
         }
     }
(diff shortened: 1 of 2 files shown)


cschuhen:
The changes look reasonable. Not tested so far. Till now I've kind of been avoiding doing much with the bus error states a bit. I know Thomas/Blaa was playing with those states a bit. I was debating how to incorporate returning error states, particularly with the buffered modes. My other thought was to use an error frame type approach but I guess rust enums(i.e. result) can be low cost enough to not matter too much.

cschuhen:
Here are some comments from Tomasz/bla:
_BusOff and maybe others don't need to be persistent though... If handled in the irq
I was going in other direction: handling errors transparently. But this might be more robust
Question is: what should user do after getting BusError. There should be some kind of way to get a retry
And reset the error state
Without deconstructing all types that might be split across codebase_

tcbennun:
Perhaps it's helpful if I describe what I'd expect as an end user...

- I must have the option to receive, at my discretion, a) individual protocol errors as they occur, and/or b) changes in error state, as they occur
- I must be able to read the error counters at any time
- I must at minimum be able to initiate a bus recovery sequence upon Bus Off
- I should have the option to configure the driver to perform bus recovery automatically upon Bus Off

In any case, if these current changes look reasonable, I'd say let's merge and I'll open a new PR in the coming days to improve the situation a little more. We can stick with the current approach of returning `Result<_, BusError>` in RX. My first port of call will be to ensure that this doesn't result in a persistent error state being constantly sent to the user. The next thing will be bus-off recovery support.

Would be good also to ensure we're not stepping on each others' toes here - is there somewhere we can keep in touch? I'm in the embassy matrix server.

cschuhen:
That's all fine by me. I'm fine to merge what's there now.

Dirbaio:
goddamnit y u no merge

PR #2660, opened 2024-03-04
docs: clarify capabilities of zerocopy channel

lulf:
Even though you can borrow() senders and receivers, they are borrow'ed &mut, so you cannot use more than one. And looking at the fn send implementation, it doesn't look like multiple senders would be safe, as it would hand out the same &mut slice to two senders? I've gone through this a few times in my head now and reached the same conclusion, but happy to be proven wrong!

(Specifically:  https://github.com/embassy-rs/embassy/blob/main/embassy-sync/src/zerocopy_channel.rs#L129)


Dirbaio (approved):
ah yep I think this is copypaste from the non-zeropcopy channel :(

PR #2663, opened 2024-03-05
ci stm32: update U5 H5 WBA target to Armv8-M

eZioPan:
update according to ST info https://www.st.com/content/st_com/en/arm-32-bit-microcontrollers/arm-cortex-m33.html


Dirbaio (approved):
nice find! :rofl:

PR #2664, opened 2024-03-05
ci stm32: build target fix

eZioPan:
- ~cortex-m4 and cortex-m7 of stm32 chips have hardware float~ add a note about hf thing
- stm32f0 should be thumbv6m
- ~[stm32-metapac update need] stm32wl5, stm32wb35 and stm32wb55 are mix dual core chip (cortex-m4 & cortex-m0+)~
- remove some extra whitespace

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -15,7 +15,8 @@ if [ $TARGET = "x86_64-unknown-linux-gnu" ]; then
     BUILD_EXTRA="--- build --release --manifest-path examples/std/Cargo.toml --target $TARGET --out-dir out/examples/std"
 fi
 
-cargo batch  \
+# CI intentionally does not use -eabihf on thumbv7em to minimize dep compile time.
+cargo batch \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features log \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features defmt \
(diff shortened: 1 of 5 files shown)


Dirbaio:
> cortex-m4 and cortex-m7 of stm32 chips have hardware float

CI intentionally does not use -eabihf, because we shouldn't care (nothing in Embassy uses floats) and it makes it faster (shared deps are built once, not twice for -eabi and -eabihf)

other changes look good to me!

eZioPan:
> CI intentionally does not use -eabihf

then what about thumbv8m.main-none-eabihf? should we also remove hf?

Dirbaio:
all thumbv8 chips we support do have hardfloat, so we wouldn't gain speed by switching from `thumbv8m.main-none-eabihf` to `thumbv8m.main-none-eabi` because it's still 1 target. We have to compile for *some* thumbv8 anyway, so it doesn't matter which.

going from `thumbv7em-none-eabi + thumbv7em-none-eabihf` to just `thumbv7em-none-eabi` does give speed gains, because you're reducing the amount of time deps are compiled.

eZioPan:
and what about thumbv7m and thumbv7***e***m?

Should we make them all not have ***e*** extension?

Dirbaio:
`thumbv7m` vs `thumbv7em` does matter for our use case, the `e` has extra instructions that rustc can generate when compiling code.

(fwiw we could literally use thumbv6m for everything since it's all backwards compatible, but the general idea is to keep the target matching the chip, except for floats because they make no difference for us)

eZioPan:
oh no, I got to update svd source for stm32-rs first, and generate svd from it, then update stm32-data, to make this compile...

Dirbaio:
The STM32WB second core only runs the radio blob that is signed+encrypted by ST. You can't run custom code on it. The STM32WB is effectively single-core from the user's point of view. I don't think we should have the `stm32wbxxxx-cm0` targets, they can't be used at all.

Dirbaio (approved):
thanks!

PR #2665, opened 2024-03-06
stm32: impl uart write_all without panic

andresv:
Default `write_all` implementation uses panic: https://github.com/rust-embedded/embedded-hal/blob/master/embedded-io-async/src/lib.rs#L138 and https://github.com/rust-embedded/embedded-hal/blob/master/embedded-io/src/lib.rs#L397.

This version produced 456 bytes smaller binary in my application where embedded-io-async::write_all was used.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -562,6 +562,18 @@ impl<'d, T: BasicInstance> embedded_io_async::Write for BufferedUart<'d, T> {
         self.tx.write(buf).await
     }
 
+    async fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
+        let mut buf = buf;
+        while !buf.is_empty() {
+            match self.tx.write(buf).await {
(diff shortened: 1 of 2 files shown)


Dirbaio:
`write_all` calls into `uart::write` which we *know* never returns `Ok(0)` because it's our own code. Therefore it returning `Ok(0)` is truly a "should never happen" scenario. If it happens it is a bug. It does not make sense to return that to the user, there is nothing they can do about it, there's nothing their code can do to handle it. The only solution is changing the embassy-stm32 code to fix it.

The appropriate way to handle these "should never happen" bugs is a panic.

If the motivation of this change is reducing code size because of the panic bloat, the right fix is enabling `panic_immediate_abort`.  See [the FAQ](https://embassy.dev/book/dev/faq.html#_my_binary_is_still_big_filled_with_stdfmt_stuff). It completely removes the panic bloat. We shouldn't do code changes that wouldn't stand on their own just to reduce the amount of panics.

PR #2668, opened 2024-03-07
Update stm32c0 HSI frequency

tomasbarton-stemcell:
The HSI frequency should be 48 MHz, not 16. I checked the datasheets for both STM32C011x4/x6 and STM32C031x4/x6.


Dirbaio (approved):
huh how did I miss this. Thanks for the find and the PR!

PR #2499, opened 2024-01-30
Update embassy-stm32 after stm32-metapac timer_v2

eZioPan:
## Summary

* Keep BASIC of TIM PAC as `BasicInstance`
* Merge 1CH, 2CH, GP16 of TIM PAC into `CaptureComapre16bitInstance`
* Keep GP32 of TIM PAC as `CaptureComapre32bitInstance`
* Merge 1CH_CMP, 2CH_CMP, ADV of TIM PAC into `ComplementaryCaptureCompare16bitInstace`
* Remove `GeneralPurpose16bitInstance`, `GeneralPurpose32bitInstance` and `AdvancedControlInstance`, since we merge different kinds of PAC TIMs, they are not accurate under new PAC.
* `pin_trait!` has been redesign to match TimerInstance more precise.

Such that this PR didn't do much modify to non-sealed trait.

Since PAC has 10 different blocks, and HAL only got 4 traits. The "sealed" part is a little bit thicker than original.

* All methods other than Channel-count related is put into "smallest" sealed::TimerInstance
  * `sealed::Core` is about PSC, CNT, ARR and Interrupt
  * `sealed::BasicNoCr2` add extra methods to enable DMA. This one is for Timer other than 1CH and 2CH. 1CH and 2CH don't have DMA request.
  * `sealed::GeneralPurpose1ChannelInstance` (1CH of PAC) has all methods that all Channel shared.
  * `sealed::GeneralPurpose1ChannelComplementaryInstance` (1CH_CMP of PAC) has all methods that all Complementary Channel shared.

* Methods for Channel count related is put into "kind-largest" sealed::TimerInstance  
  * `sealed::GeneralPurpose16bitInstance` has all Capture/Compare Channel-count related methods
  * `sealed::AdvancedControlInstance` has all Complementary Capture/Compare Channel-count related methods
* To merge 10 TIM of PAC into 4 TimerInstance of HAL:
  All merges are done by *kinda-unsafe-promotion*
  * 1CH and 2CH are forced implement "`sealed::GP16Instance`"
  * 1CH_CMP and 2CH_CMP are forced implement "`sealed::ADVInstance`"

  mostly just read/write to empty fields of same register.

------

This PR is based on [my fork of stm32-data-generated](https://github.com/eZioPan/stm32-data-generated/tree/timer_v2) for embassy CI pick timer_v2 PAC 

**DO NOT MERGE BEFORE**

- [ ] embassy-rs/stm32-data#364

**MERGED.**

**NEED A** 

[...]

Issue #2409, opened 2024-01-06
STM32F072 TIM15 PWM not working (MOE bit)


martin2250:
The following code should enable the PWM output on pin PA2:
```rust
    let output_green = PwmPin::new_ch1(p.PA2, OutputType::PushPull);
    let mut pwm_green = SimplePwm::new(
        p.TIM15,
        Some(output_green),
        None,
        None,
        None,
        khz(1),
        Default::default(),
    );
    pwm_green.enable(Channel::Ch1);
    pwm_green.set_duty(Channel::Ch1, 10);
```

However, the pin stays in high impedance mode. Looking around I found other people having the same issue and the suggested fix was setting the "master output enable" (MOE) bit in the BDTR register. The following code makes the PWM output work properly on my hardware:
```rust
    const TIM15A: embassy_stm32::pac::timer::TimAdv =
        unsafe { embassy_stm32::pac::timer::TimAdv::from_ptr(0x4001_4000 as usize as _) };
    TIM15A.bdtr().modify(|r| r.set_moe(true))
```

[TIM15 is declared here](https://github.com/embassy-rs/stm32-data-generated/blob/b97d581071b8b403c28fd89b55a80e02ecd3ed16/stm32-metapac/src/chips/stm32f072c8/pac.rs#L190) as TimGp16, which does not have the BDTR register. I'm not sure, however, if is correctly described by TimAdv. Also, other chips are probably affected by this issue as well.

Cheers!

phycrax:
Same issue here with STM32F030C8 TIM16 and TIM17. Setting MOE bit works.

nbars:
This is also the case for TIM15, TIM16 and TIM17 for STM32F030x4/x6/x8/xC and STM32F070x6/xB. I would love to help here but am relatively new to this ecosystem. I guess we need some timer type between general and advanced timers?

jr-oss:
I have a similar problem in my project, that the PWM does not output any signal after updating Embassy.

There used to be a `this.inner.enable_outputs();` in `SimplePwm::new_inner()`, which got removed by PR #2499.
https://github.com/embassy-rs/embassy/pull/2499/files#diff-34e412deccc6f88af9fce995f7c4ef7fc8ce6502ea2df991b562ab205be4a93eL87

Just inserting it again, does not work, because it was also moved to trait `GeneralPurpose1ChannelComplementaryInstance`, which is not used by `SimplePwm`.
[...]

PR #2670, opened 2024-03-07
STM32 SimplePwm: Fix regression and re-enable output pin

jr-oss:
PR #2499 implemented timer hierarchy, but removed enable_outputs() from trait CaptureCompare16bitInstance and from SimplePwm.

This functions is required for advanced timers to set bit BDTR.MOE and to enable the output signal.

### DIFF
--- a/embassy-stm32/src/timer/mod.rs
+++ b/embassy-stm32/src/timer/mod.rs
@@ -439,9 +439,9 @@ pub(crate) mod sealed {
             Self::regs_1ch_cmp().bdtr().modify(|w| w.set_dtg(value));
         }
 
-        /// Enable timer outputs.
-        fn enable_outputs(&self) {
-            Self::regs_1ch_cmp().bdtr().modify(|w| w.set_moe(true));
+        /// Set state of MOE-bit in BDTR register to en-/disable output
+        fn set_moe(&self, enable: bool) {
+            Self::regs_1ch_cmp().bdtr().modify(|w| w.set_moe(enable));
         }
     }
 
@@ -685,6 +685,13 @@ pub trait CaptureCompare16bitInstance:
     + sealed::GeneralPurpose16bitInstance
     + 'static
 {
+    // SimplePwm<'d, T> is implemented for T: CaptureCompare16bitInstance
+    // Advanced timers implement this trait, but the output needs to be
+    // enabled explicitly.
+    // To support general-purpose and advanced timers, this function is added
+    // here defaulting to noop and overwritten for advanced timers.
+    /// Enable timer outputs.
+    fn enable_outputs(&self) {}
 }
 
 #[cfg(not(stm32l0))]
@@ -911,7 +918,13 @@ foreach_interrupt! {
         impl_1ch_cmp_timer!($inst);
         impl_2ch_cmp_timer!($inst);
         impl BasicInstance for crate::peripherals::$inst {}
-        impl CaptureCompare16bitInstance for crate::peripherals::$inst {}
+        impl CaptureCompare16bitInstance for crate::peripherals::$inst {
+            /// Enable timer outputs.
+            fn enable_outputs(&self) {
+                use crate::timer::sealed::GeneralPurpose1ChannelComplementaryInstance;
+                self.set_moe(true);
+            }
+        }
         impl ComplementaryCaptureCompare16bitInstance for crate::peripherals::$inst {}
     };
     ($inst:ident, timer, TIM_1CH_CMP, CC, $irq:ident) => {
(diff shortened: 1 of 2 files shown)


eZioPan:
> I left comments in Issue #2409 with my understanding what happened.
> It's still unclear to me, whether this approach fits the new timer hierarchy.

Ah, I'm terribly sorry for this mess. I didn't notice that `SimplePwm` is implmented on `TIM_GP16` at that time, thus adding `enable_output()` to `TIM_1CH_CMP` is useless for this situation.

You approach is correct on both old and new timer hierarchy. It requires a empty implement `enable_output` for `TIM_GP16`, and enable `MOE` inside the method for `TIM_1CH_CMP` / `TIM_2CH_CMP` / `TIM_ADV`.

embassy-stm32/src/timer/mod.rs
```diff
@@ -438,11 +438,6 @@ pub(crate) mod sealed {
         fn set_dead_time_value(&self, value: u8) {
             Self::regs_1ch_cmp().bdtr().modify(|w| w.set_dtg(value));
         }
-
-        /// Enable timer outputs.
-        fn enable_outputs(&self) {
-            Self::regs_1ch_cmp().bdtr().modify(|w| w.set_moe(true));
-        }
```
eZioPan, line 445: could you restore this method? `MOE` existed on all `TIM_1CH_CMP` `TIM_2CH_CMP` and `TIM_ADV`.

embassy-stm32/src/timer/mod.rs
```diff
@@ -685,6 +680,13 @@ pub trait CaptureCompare16bitInstance:
     + sealed::GeneralPurpose16bitInstance
     + 'static
 {
+    // SimplePwm<'d, T> is implemented for T: CaptureCompare16bitInstance
+    // Advanced timers implement this trait, but the output needs to be
+    // enabled explicitly.
+    // To support basic and advanced timers, this function is added
```
eZioPan, line 686: It better to say

"To support general-purpose and advanced timer, ..."

since there is already another timer called `BasicTimer`

embassy-stm32/src/timer/mod.rs
eZioPan: Or even better, rewrite this method to something like `fn set_moe(&self, enable: bool)`, and call this method in public trait implement.

embassy-stm32/src/timer/mod.rs
```diff
@@ -947,7 +949,14 @@ foreach_interrupt! {
         impl_1ch_cmp_timer!($inst);
         impl_2ch_cmp_timer!($inst);
         impl BasicInstance for crate::peripherals::$inst {}
-        impl CaptureCompare16bitInstance for crate::peripherals::$inst {}
+        impl CaptureCompare16bitInstance for crate::peripherals::$inst {
+            /// Enable timer outputs.
+            fn enable_outputs(&self) {
+                use crate::timer::sealed::AdvancedControlInstance;
+                let r = Self::regs_advanced();
+                r.bdtr().modify(|w| w.set_moe(true));
```
eZioPan, line 957: You might need to copy this code (or `.set_moe()`) to other 2 blocks: `($inst:ident, timer, TIM_1CH_CMP, UP, $irq:ident)`, `($inst:ident, timer, TIM_2CH_CMP, UP, $irq:ident)`, to make sure they also enable `MOE`.

embassy-stm32/src/timer/mod.rs
jr-oss: Replaced `enable_outputs()`by `fn set_moe(&self, enable: bool)`

embassy-stm32/src/timer/mod.rs
jr-oss: Good point. I changed the comment accordingly

embassy-stm32/src/timer/mod.rs
jr-oss: Thanks, I would have missed these locations.
Added `fn enable_outputs(&self)` to these timers, also.

jr-oss:
Thanks for your fast response and don't worry about the mess. 
It's easy to miss the output enable behavior between basic and advanced timers and I only noticed, because I use it in my project.

I can squash the commits, if you agree to the current state.

eZioPan:
> I can squash the commits, if you agree to the current state.

Yeah! It looks good to me!

And I also test PWM output on my dev board with TIM1(ADV), TIM15(2CH_CMP), TIM16(1CH_CMP), they all output correct waveform.

jr-oss:
Thanks. I squashed the commits.

PR #2673, opened 2024-03-08
stm32: can: fd: allow filter setting during normal operation

tcbennun:
I have an application where I need to modify the FDCAN filters dynamically.

As far as I can tell, there's no restriction on writing to the filters in message RAM during normal mode. I still need to test it properly though.

This is somewhat a first pass. I don't really like duplicating the methods like this. Ideas welcome

### DIFF
--- a/embassy-stm32/src/can/fd/peripheral.rs
+++ b/embassy-stm32/src/can/fd/peripheral.rs
@@ -314,15 +314,6 @@ impl Registers {
             self.regs.endn().read().etv() == 0x87654321_u32,
             "Error reading endianness test value from FDCAN core"
         );
-
-        /*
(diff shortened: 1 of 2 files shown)


cschuhen:
Yeah, it does add a bit of bloat. I'm happy to have a play too when I get a chance, see if I can come up with some ideas.

cschuhen:
Something like this could almost work: https://gist.github.com/cschuhen/b6502d54e01f6ded5c7d86ecf3aaa186
Except that I'm not sure how to hide 'registers()' from the user.

tcbennun:
Closing, I'm gonna give this another go, as I want to add yet more functions that should be common to buffered/nonbuffered

PR #2675, opened 2024-03-09
Remove personal preferences regarding editor settings

msrd0:
I personally don't like using _format on save_. I also personally hit ctrl+s after basically every character I type. Other people use editors differently and have different preferences. That's why we let people choose what they prefer and not force it upon them when they open this repository in code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-  "editor.formatOnSave": true,
   "[toml]": {
     "editor.formatOnSave": false
   },


Dirbaio:
Can you remove just the `"editor.formatOnSave": true`, and leave in the `false` for toml and markdown? The reason they're there so people don't send PRs with Cargo.tomls accidentally horribly autoformatted. This way vscode will honor the user setting for Rust code, but never format toml/markdown.

msrd0:
Yep that sounds reasonable, done

Dirbaio (approved):

PR #2677, opened 2024-03-09
stm32/rcc: wait for peripheral clock to be active. also, hold the peripheral reset while enabling the clock.

ExplodingWaffle:
Was looking through the RCC codegen when I was nerdsniped by the mention of an errata. After a bit of digging into it I discovered the situation already well described here: http://efton.sk/STM32/gotcha/g183.html including some forum posts where people have actually had issues: to be clear I haven't, and wasn't able to repro anything after some trying on my STM32G071, but that may be because that chip is too slow and/or simple. The only attempt at an Actually Correct implementation I found was here: https://community.st.com/t5/stm32-mcus-embedded-software/stm32-clock-and-delay-helper-functions/td-p/593082

Currently embassy-stm32 only mitigates this with a DSB on STM32F2: if nothing else it should be doing that on all chips since this is an errata on more chips than just F2 and a feature on everything else. Current RMs say you need to delay for 2 clock cycles (presumably peripheral clock cycles) for the peripheral clock to become active and peripheral read or write accesses to be supported. I'm not too confident about the approach taken in this PR, but I think it is at least overly conservative. I don't have much of an understanding of how exactly wait states and peripheral buses etc work.

I also changed the peripheral reset, starting it before the clock and stopping it after: I think a small "glitch" cooould have been possible before, if a peripheral had been configured and then disabled, the old config would have been enabled briefly between the clock enable and the reset. If nothing else this feels a little bit more robust to me.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -509,25 +509,20 @@ fn main() {
         if let Some(rcc) = &p.rcc {
             let en = rcc.enable.as_ref().unwrap();
 
-            let rst = match &rcc.reset {
+            let (start_rst, end_rst) = match &rcc.reset {
                 Some(rst) => {
                     let rst_reg = format_ident!("{}", rst.register.to_ascii_lowercase());
                     let set_rst_field = format_ident!("set_{}", rst.field.to_ascii_lowercase());
-                    quote! {
-                        crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(true));
-                        crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(false));
-                    }
+                    (
+                        quote! {
+                            crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(true));
+                        },
+                        quote! {
+                            crate::pac::RCC.#rst_reg().modify(|w| w.#set_rst_field(false));
+                        },
+                    )
                 }
-                None => TokenStream::new(),
-            };
-
-            let after_enable = if chip_name.starts_with("stm32f2") {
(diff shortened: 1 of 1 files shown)


Dirbaio:
bender run

Dirbaio:
Can you fix tests? they fail because they build embassy-stm32 for x64 :( 

(the HIL test failure is unrelated, that board is flaky for some reason)

Dirbaio:
Fix looks reasonable. The only thing that makes me a bit uncomfortable is reading from the peripheral while it's in reset. Is that allowed, is there documentation supporting that? I know some peripherals can insert waitstates on accesses, maybe there's a risk the read can hang forever if the peripheral is in reset?

ExplodingWaffle:
Indeed it isn't very clear...
I did find this, in RMs for F0 and F3
image
In F7 onwards (though a few omit this):
image
Changed to remove the peripheral reads. Kept the readback from the RCC register (this is what [the macros in the ST HALs](https://github.com/STMicroelectronics/stm32f7xx_hal_driver/blob/5af4f7192a9331d27e5e90feed5e7cebc3f1356c/Inc/stm32f7xx_ll_bus.h#L314) do) and the status quo DSB. can always be revisited if anyone actually runs into this

Dirbaio (approved):
thanks!

ExplodingWaffle:
Happened to be looking at H7 RCC and it actually describes the procedure with the dummy read. Also provides different guarantees wrt read and writes to a disabled peripheral. Of course, the ST HAL code never actually does the workaround… [despite saying it does](https://github.com/STMicroelectronics/stm32h7xx_hal_driver/blob/fec141ce999da655a48e1a15db83a72d564a1312/Src/stm32h7xx_hal_rcc.c#L50) 🙃
IMG_6410

PR #2680, opened 2024-03-10
Configurable DMA Request Priority

caleb-garrett:
Adds DMA request priority as transfer option, while maintaining the current setting of `VERYHIGH` as the default.

Currently, all DMA transfers are set to the highest priority level and is not configurable. This is acceptable for most use cases but there are cases where priority must be configured for peripheral DMA to work properly, such as the CRYP peripheral. From ST manual RM0410:

> It is important that DMA controller empties the cryptographic processor output FIFO before filling up the cryptographic processor input FIFO. To achieve this, the DMA controller should be configured so that the transfer from the cryptographic peripheral to the memory has a higher priority than the transfer from the memory to the cryptographic peripheral.


Dirbaio (approved):
Thanks!

Crazy that ST designed a peripheral that actually requires this :grimacing:

PR #2683, opened 2024-03-11
stm32: add disable_ucpdx_dead_battery

ExplodingWaffle:
Config option to disable the dead battery behaviour on UCPD parts. Option since if you're using UCPD you need to configure that first before disabling (which will probably be done in the driver). also disables unconditionally on G070 and G0B0 since they aren't supposed to have UCPD but do have this feature.

cc @timokroeger but this doesn't depend on the UCPD driver

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -189,6 +189,18 @@ pub struct Config {
     /// Defaults to P0 (highest).
     #[cfg(gpdma)]
     pub gpdma_interrupt_priority: Priority,
+
+    /// Disables UCPD1 dead battery functionality.
+    ///
+    /// Defaults to true (disabled).
+    #[cfg(ucpd)]
+    pub disable_ucpd1_dead_battery: bool,
+
+    /// Disables UCPD2 dead battery functionality.
+    ///
+    /// Defaults to true (disabled).
+    #[cfg(all(ucpd, stm32g0x1))]
+    pub disable_ucpd2_dead_battery: bool,
 }
 
 impl Default for Config {
@@ -203,6 +215,10 @@ impl Default for Config {
             dma_interrupt_priority: Priority::P0,
             #[cfg(gpdma)]
             gpdma_interrupt_priority: Priority::P0,
+            #[cfg(ucpd)]
+            disable_ucpd1_dead_battery: true,
+            #[cfg(all(ucpd, stm32g0x1))]
+            uisable_ucpd2_dead_battery: true,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


timokroeger:
Great that you had a go at this already!
Can you please wrap the added init code in a `pub(crate) fn disbale_ucpd_dead_battery()` or similar function?
Then the UCPD driver can re-use this function to switch of the dead-battery support if not expliticly disabled through init.

Dirbaio:
- I'd invert the config fields: make them "enable" instead of "disable", and default to false. This avoids awkward double negations where you have to mentally parse "disable: false" as "enable".
- Do we need this in the global `embassy-stm32` config? it's confusing that this is controlled both here and in the UCPD driver. IIUC the reason to disable it on init is it's enabled by default so it might interfere with these pins even if you're not using UCPD at all? in this case, would it make sense to unconditionally disable it in `embassy_stm32::init()` and then have a config setting in UCPD to enable them back? or would that cause a "glitch" at boot where they're disabled for a brief period of time which is unacceptable?

timokroeger:
> would that cause a "glitch" at boot where they're disabled for a brief period of time which is unacceptable?

From "Universal Serial Bus Type-C Cable and Connector Specification Release 2.3":
image
"as quickly as practical" sounds to me like a source can disable power immediately when it sees a glitch. So we cannot disable the dead-battery feature unconditonally unfortunately.

Dirbaio:
oh wow okay. that's unfortunate. LGTM then if we change the config fields to `enable` instead of `disable.

embassy-stm32/src/lib.rs
```diff
@@ -203,6 +215,10 @@ impl Default for Config {
             dma_interrupt_priority: Priority::P0,
             #[cfg(gpdma)]
             gpdma_interrupt_priority: Priority::P0,
+            #[cfg(ucpd)]
+            enable_ucpd1_dead_battery: false,
+            #[cfg(all(ucpd, stm32g0x1))]
```
Dirbaio, line 220: you can do `#[cfg(peri_ucpd1)]`, `#[cfg(peri_ucpd2)]` to query whether the current chip has a peripheral instance.

embassy-stm32/src/lib.rs
```diff
@@ -296,3 +331,41 @@ pub fn init(config: Config) -> Peripherals {
...
 }
+
+#[cfg(ucpd)]
+/// Safety: must only be called when all UCPDs are disabled (e.g. at startup)
+unsafe fn ucpd_init(
+    _cs: critical_section::CriticalSection,
+    ucpd1_db_enable: bool,
+    #[cfg(stm32g0x1)] ucpd2_db_enable: bool,
+) {
+    #[cfg(all(ucpd, stm32g0x1))]
```
Dirbaio, line 342: no need to check for `ucpd` here if the function is `#[cfg(ucpd)]`

embassy-stm32/src/lib.rs
ExplodingWaffle: oh, neat. where do these cfgs actually come from? couldn't figure it out

embassy-stm32/src/lib.rs
ExplodingWaffle: good catch 😅

embassy-stm32/src/lib.rs
Dirbaio: grep for `rustc-cfg` in `build.rs`. This one in particular is here https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/build.rs#L65

embassy-stm32/src/lib.rs
```diff
@@ -189,6 +189,18 @@ pub struct Config {
...
+    /// Enables UCPD1 dead battery functionality.
+    ///
+    /// Defaults to false (disabled).
+    #[cfg(ucpd)]
+    pub enable_ucpd1_dead_battery: bool,
+
+    /// Enables UCPD2 dead battery functionality.
+    ///
+    /// Defaults to false (disabled).
+    #[cfg(all(ucpd, stm32g0x1))]
```
Dirbaio, line 202: you should use `cfg(peri_*)` here too for consistency

Dirbaio (approved):
thanks!

PR #2685, opened 2024-03-11
Improve panic message when requesting frequency higher than clock

msrd0:
Previously it would panic with message "unreachable", which isn't particularly clear about what the problem is and how to fix it.


msrd0:
That CI failure doesn't look related to my changes

Dirbaio:
yeah, nucleo-stm32f091rc is just flaky.

Issue #2162, opened 2023-11-08
stm32f1 ADC example does not work


katyo:
I run ADC example on `stm32f103c8` but seems it does not works as expected.

It just awaits infinitely here:
https://github.com/embassy-rs/embassy/blob/94586576a06cf5869d0cdea6d678a0448bb1fdf8/examples/stm32f1/src/bin/adc.rs#L26

GDB shows me that `wfe` instruction was executed last but seems no events happening.

marsjo:
I run into the same problem but actually only if I use the crates. If I use the dependencies from path it works fine for me.
Did you fix it for you?

pfzetto:
Same issue here on a `stm32f103c6` (Infinite wait on `adc.read()`).

> I run into the same problem but actually only if I use the crates. If I use the dependencies from path it works fine for me. Did you fix it for you?

Using the suggested method with 3638df789e4f498b9058bbeb27bdddab4a80bd49 I can read the adc value one time for each ADC.
On the second invocation it hangs again.
The output of the ADC-example for `stm32f1` is:
```
0.000000 TRACE BDCR ok: 00008200
└─ embassy_stm32::rcc::bd::{impl#2}::init @ /home/paulz/src/embedded/embassy/embassy-stm32/src/fmt.rs:117
0.000000 DEBUG rcc: Clocks { hclk1: Some(Hertz(8000000)), pclk1: Some(Hertz(8000000)), pclk1_tim: Some(Hertz(8000000)), pclk2: Some(Hertz(8000000)), pclk2_tim: Some(Hertz(8000000)), rtc: Some(Hertz(40000)), sys: Some(Hertz(8000000)), usb: None }
└─ embassy_stm32::rcc::set_freqs @ /home/paulz/src/embedded/embassy/embassy-stm32/src/fmt.rs:130
0.000030 INFO  Hello World!
└─ embassy_stm32f1_examples::____embassy_main_task::{async_fn#0} @ src/main.rs:19
```
The code hangs on the first loop iteration because `read` got called for `vref`.

nuuskamummu:
Just spent the evening getting stuck on the same thing. Turns out, for me it was some compiler flags.
If I remove the config below, the ADC works.

 [profile.dev]
 lto = true
 opt-level = 1 //doesnt matter what level, won't work.

45ninjas:
ADC example functions when built without the release profile, or if `opt-level = 0`.

xokdvium:
[...]

PR #2690, opened 2024-03-12
Add small Delay to STM32F1 ADC to avoid lockup

pfzetto:
Added a one tick delay to the `convert` function of the STM32F1 ADC implementation.
Prior the `convert` function didn't return a value causing a infinite await.

I'm only getting started with rust embedded, so please take the following statement cautious: I think this bug exists because the timer registers weren't set fast enough, causing the interrupt registration to be ignored.

related Issues:
- #2162


Dirbaio:
this is a workaround. If we don't understand why the issue happens, the workaround is not guaranteed to work in all cases, and it also makes conversion slower. We should try to find the "real" bug and fix it.

PR #2691, opened 2024-03-12
STM32 CRYP DMA

caleb-garrett:
Adds async CRYP API using DMA for STM32. STM32F7 example is updated to use async API. CRYP test is updated so that encryption occurs in blocking mode, then decryption occurs in async mode.


Dirbaio (approved):
thank you! :rocket:

PR #2693, opened 2024-03-13
Use the same Frame struct for BXCAN and FDCAN.

cschuhen:
This also updates BXCAN to use current embedded_can Id types.

In general this reduces user differences between FDCAN and BXCAN making it easier to share code between the two.


embassy-stm32/src/can/bx/mod.rs
```diff
@@ -148,13 +154,13 @@ impl IdReg {
     /// Sets the remote transmission (RTR) flag. This marks the identifier as
     /// being part of a remote frame.
     #[must_use = "returns a new IdReg without modifying `self`"]
```
Dirbaio, line 156: this `must_use` is now accidentally applying to the below method `to_id`.

Dirbaio (approved):
>  +186 −502 

nice :fire:

PR #2697, opened 2024-03-15
stm32 CORDIC driver

eZioPan:
~need PAC update to embassy-rs/stm32-data-generated@d63f9313acbb7a678631dd41ef82bf76c418b7ca~
(fixed by #2720)


Dirbaio:
> need PAC update to https://github.com/embassy-rs/stm32-data-generated/commit/d63f9313acbb7a678631dd41ef82bf76c418b7ca

to do this, you can change the commit hash in `embassy-stm32/Cargo.toml`

eZioPan:
> you can change the commit hash

yeah, but the PAC also has some change about the timer. I’m considering make a separate PR about update PAC, but I don’t know if it’s a little bit hurry to make an update. Since my CORDIC driver will take long time to finish, and PAC changes seems not that significant and urgent.

eZioPan:
@Dirbaio Would you give this a trusted label?

embassy-stm32/src/cordic/mod.rs
```diff
@@ -0,0 +1,804 @@
...
+
+pub(crate) mod sealed;
+
+/// Low-level CORDIC access.
+#[cfg(feature = "unstable-pac")]
+pub mod low_level {
+    pub use super::sealed::*;
+}
+
+const INPUT_BUF_MAX_LEN: usize = 16;
```
eZioPan, line 24: When using DMA mode, the driver need a buf to merge ARG1 list and ARG2 list into one DMA list.

For current implementation, Cordic driver allocates a fixed size of memeory for it, should we instead ask user to provide a buffer size for Cordic driver? If so, how should we design the API?

Dirbaio: i'd ask the user to pass the data already interleaved, in `[a1, b1, a2, b2, ...]` format. This way avoids copies if they can arrange for the data to be written that way when it's created. If they can't then they have to interleave it themselves, which allows them to choose how the buffer is allocated and sized.

embassy-stm32/src/cordic/mod.rs
```diff
@@ -0,0 +1,804 @@
...
+        T::disable();
+    }
+}
+
+// q1.31 related
+impl<'d, T: Instance> Cordic<'d, T> {
+    /// Run a blocking CORDIC calculation in q1.31 format
+    pub fn blocking_calc_32bit(
+        &mut self,
+        arg1s: &[f64],
```
Dirbaio, line 182: The public APIs take and return f32/f64 and convert to/from fixed internally. IMO it'd be better if they took the fixed numbers directly, and let the user convert if they want to. Maybe the user wants to use fixed for everything, or maybe they want to chain multiple calculations with cordic and only want to pay the overhead of converting to/from float at the beginning and end, not at every calculation.

eZioPan:
hum? test broken on stm32u5a5zj?

eZioPan:
u5z5zj still crash, don’t know why…

eZioPan:
Unfortunatly, I don't have a stm32u5 board. I have to exclude stm32u5a5 and stm32u5a9 from Cordic driver.

Dirbaio (approved):
thank you!

PR #2699, opened 2024-03-15
Impl `MultiwriteNorFlash` for `BlockingAsync`

HaoboGu:
According to https://github.com/rust-embedded-community/embedded-storage/blob/master/src/nor_flash.rs#L191, `MultiwriteNorFlash` trait adds nothing to `NorFlash`, it should be safe to impl `MultiwriteNorFlash` for `BlockingAsync`


Dirbaio (approved):
thanks!

PR #2701, opened 2024-03-15
STM32 UCPD CI Test

timokroeger:
Test sends messages between a sink and a source in either direction.
Also checks if hard resets are detected.


Dirbaio (approved):
thanks for working on this! having tests is great :rocket:

Issue #2702, opened 2024-03-15
STM32H7 FDCAN Improper RAM Access


maiaherringfish:
I am able to configure both FDCAN1 and FDCAN3, but only FDCAN1 can transmit and receive correctly. 

When I try to transmit with FDCAN3 I am not seeing the actual frame I asked it to write (unlike when I transmit with FDCAN1). Instead I'm seeing the same pretty much nonsense frame written (regardless of whether FDCAN1 is enabled or disabled). 

When I receive and both FDCAN1 and FDCAN3 are enabled and on the same bus, FDCAN1 and FDCAN3 receive the same messages.,but if I disable FDCAN1 I get empty frames on FDCAN3

cschuhen:
Yeah, it seems broken for even FDCAN2. I think it may be due to Registers::msg_ram_mut() not taking into account msg_ram_offset.

maiaherringfish:
tested your fix @cschuhen and looks like FDCAN3 is working now. thanks for the quick help on this

PR #2703, opened 2024-03-16
FDCAN: Fix offset issue preventing CAN2 and CAN3 from working.

cschuhen:
Fix for https://github.com/embassy-rs/embassy/issues/2702
Tested for CAN2 but should fix CAN3 on devices that support it also.


cschuhen:
+Trusted please?

Issue #1417, opened 2023-05-01
stm32: I2C "transaction" not supported


ctron:
The "transaction" functions of STM32 still seems to be contain `todo!`.

ctron:
Having the ability to perform two read operations seems to be a requirement for the default PN532 firmware.

Dirbaio:
There's been no work on this afaik, so a PR would be very welcome :) 

maybe I'd avoid the complexity of defining our own `Operation` and converting? since embedded-hal 1.0 is supposed to be "stable forever" maybe we can just reexport embedded-hal's.

Dirbaio:
fixed in the linked PRs.

PR #2713, opened 2024-03-20
stm32/i2c(v1): Implement blocking transactions

sgoll:
This implements blocking transactions for I2C v1 in `embassy-stm32`. The implementation follows the [transaction contract](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction), laid out by `embedded-hal`, i.e. merging consecutive operations of the same type.

The idea of using frame options to re-use the same basic building blocks for reading and writing is inspired by the [STM32CubeF2](https://www.st.com/en/embedded-software/stm32cubef2.html) HAL. This should allow us to extend this PR in a straightforward way to the implementation of asynchronous transactions.

The implementation was tested for all combinations of merged and non-merged read/write operations on an STM32F2 dev board ([NUCLEO-F207ZG](https://www.st.com/en/evaluation-tools/nucleo-f207zg.html)) and logic analyzer/oscilloscope.

This is part of a series of PRs that may eventually solve #1417.

The diff is best viewed with whitespace changes hidden. I tried to touch the existing code as little as possible to make the review easier. There is some more potential for refactoring in a future PR.

### DIFF
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -311,10 +311,10 @@ impl<'d, T: Instance> embedded_hal_1::i2c::I2c for I2c<'d, T, NoDma, NoDma> {
 
     fn transaction(
         &mut self,
-        _address: u8,
-        _operations: &mut [embedded_hal_1::i2c::Operation<'_>],
+        address: u8,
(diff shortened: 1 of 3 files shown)


Dirbaio (approved):
wow, the logic is quite complex. Looks good to me. Thanks for working on this!

PR #2714, opened 2024-03-20
Doc update: signaled does not clear signal

noahbliss:
Burned a few brain cells figuring this one out haha. Signaled does not clear the signal.


Dirbaio (approved):
thanks!

PR #2717, opened 2024-03-20
Add comment warning for new users about changing pins on Ethernet for different devices + example of change

AlejandroFNadal:
Not all STM32H7 devices share the pins of the example. Added a warning and a specific example for STM32H747XIH


Dirbaio (approved):
very helpful, thank you! :)

Dirbaio:
can you rustfmt?

AlejandroFNadal:
You are welcome! I run the formatter (Super newbie in rust, just learnt that it was a thing, thanks <3) and now it looks better.

PR #2721, opened 2024-03-20
usb-logger: avoid data loss at pipe wraparound

flelchuk:
Re-try to write the remainder of a log message to avoid truncation due to short write at pipe wraparound.


Dirbaio (approved):
nice catch, thank you! :)

flelchuk:
@Dirbaio thanks for the quick reply. Lucky Luke is getting jealous!

PR #2727, opened 2024-03-22
Forward the "std" feature to the critical-section crate in embassy-sync.

de-vri-es:
Currently, using `embassy-sync` in unit tests on a non-embedded platform will result in linker errors when using the `CriticalSectionRawMutex`.

Without forwarding this feature, the user needs to understand that `embassy-sync` uses the `critical-section` crate, and add a dependency on it just to enable the feature.

With this change, users can just enable the `"std"` feature in a dev dependency on `embassy-sync`, and the linker error will be resolved.


Dirbaio (approved):

de-vri-es:
Wow, that was quick. Thanks for the review! :D

PR #2730, opened 2024-03-23
stm32: use private_bounds for sealed traits.

Dirbaio:
- **stm32/timer: simplify traits, convert from trait methods to struct.**
- **stm32: use private_bounds for sealed traits.**


chemicstry:
The `unstable-traits` feature and `low_level` modules previously allowed implementing HAL abstractions outside of `embassy-stm32`. This was very useful for drivers that are too specialised or "hacky" to be upstreamed to this repository.

What are the alternatives after this change? I see that `frequency` method was added to fix examples that require it, but what about enable/disable/reset peripheral? Should methods be added for that too?

Dirbaio:
Yes, we can add methods for that too.

I removed the `low_level` modules because in hindsight it was not a great solution. 
- it'd expose *all* of the sealed trait's methods, which we might want to do breaking changes to (this is why it required opting into "unstable" features)
- it'd expose the actual trait. so you could then implement it from outside the crate for other types, making it no longer sealed. That's not something we want! :D

so IMO it's better to have stable low-level public APIs and keep the sealed traits fully private.

PR #2735, opened 2024-03-25
stm32: fix timer api leaking

eZioPan:


embassy-stm32/src/timer/mod.rs
```diff
@@ -69,31 +69,40 @@ pub trait GeneralInstance1Channel: CoreInstance {}
...
+pub trait AdvancedInstance1Channel: BasicNoCr2Instance + GeneralInstance1Channel {}
 /// Advanced 16-bit timer with 2 channels instance.
 
 pub trait AdvancedInstance2Channel: BasicInstance + GeneralInstance2Channel + AdvancedInstance1Channel {}
 
 /// Advanced 16-bit timer with 4 channels instance.
-pub trait AdvancedInstance4Channel: AdvancedInstance2Channel + GeneralInstance4Channel {}
+pub trait AdvancedInstance4Channel: AdvancedInstance2Channel + GeneralInstance4Channel {
+    /// Capture compare interrupt for this timer.
+    type CaptureCompareInterrupt: interrupt::typelevel::Interrupt;
```
Dirbaio, line 104: why move the interrupt from the 1ch to the 4ch trait? 1ch/2ch timers still have the interrupt, right?

(not that it matters now, because we're implementing all 3 traits for all timers ignoring the channel count, but it'll matter after we fix that)

Dirbaio: but they still "logically" have one, it's just that on some timers the CC and UP interrupts go to the same NVIC line? 

it's the same as the I2C EV and ER interrupts. On some chips they go to a single NVIC interrupt I2C1, and in others it's separated into I2C1_EV, I2C1_ER. On the Embassy side we represent it as two `type xxInterrupt` associated types, which will be the same interrupt on chips that don't split it. This works on all chips, since `bind_interrupts!` allows binding the same interrupt to multiple handlers.

though then perhaps with that logic the CC interrupt should go in the "general" instance traits since these can also do CC, hmm.

eZioPan: When it first introduced in #2614, CC is attached to AdvTim (now Adv4Ch).

So for current situation, we should move it to GeneralInstance1Channel?

Dirbaio (approved):
thanks for taking the time for adding all interrupts, it's much better now! :D

PR #2736, opened 2024-03-25
stm32-usb isochronous fix and overall better isochronous support

vDorst:
This pull request fixes the stm32 isochronous issue that we have to set the odd/even frame bit every time.
If we don´t the usb frame is rejected.

Also update `EndpointType` because isochronous endpoint also needs synchronication and usage type information.
`EndpointType` is copied from https://github.com/rust-embedded-community/usb-device

embassy-usb/src/descriptor.rs is not done yet.
For class specific properties like USB AUDIO CLASS two extra bytes have to be added to the endpoint description.
Hacked example: https://github.com/vDorst/embassy/blob/i2s-no-mclk/embassy-usb/src/builder.rs#L467 and https://github.com/vDorst/embassy/blob/i2s-no-mclk/embassy-usb/src/builder.rs#L509.
What could be a good way to provide this information in a generic way?

### DIFF
--- a/embassy-nrf/src/usb/mod.rs
+++ b/embassy-nrf/src/usb/mod.rs
@@ -771,7 +771,7 @@ impl Allocator {
         // Endpoint directions are allocated individually.
 
         let alloc_index = match ep_type {
-            EndpointType::Isochronous => 8,
+            EndpointType::Isochronous(_) => 8,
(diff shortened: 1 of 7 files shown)


vDorst:
I think your are right about that the hardware is only interested in the endpointtype.

Thanks for the information.
I have a better understanding where and how to fix it.

PR #2738, opened 2024-03-26
stm32 H5: LSE low drive mode is not functional

eZioPan:
see ES0565 2.2.17 LSE low drive mode is not functional.

"The LSE oscillator may not start or may stop in low drive mode (LSEDRV = 00). Using this mode is forbidden."


Dirbaio (approved):
:+1:

PR #2739, opened 2024-03-26
Provide embedded_io impls for Uart with and without Dma

### LINKED ISSUE #2072: STM32: Possible to have async uart with only one DMA
Hello again,
I'm using the `stm23g070cb` which only has `DMA1` with 8 DMA channels.

I would like to be able to do `async` receive via DMA on three UART's. While also using DMA channels for other devices.
I dont need `async` transmit in this instance.

Is it possible for the `impl` for `embedded_io_async::Write` and `embedded_io_async::Read` to depend on their respective DMA channel? I believe my current read of this is that both are required for the `impl` to apply.

This would allow for transmit only or receive only dma accelerated async implementations I believe?

adri326:
Weakens the requirement `TxDma: self::TxDma<T>` in the implementation of `embedded_io::Write` for both `Uart` and `UartTx`.

Provides implementations for `embedded_io::Read` for `Uart` and `UartRx`, as well as `embedded_io_async::Read`, bringing it to parity with `BufferedUart`.

These changes seem to be sufficient to make the blocking and async APIs available together in mixed scenarios, like in #2072.

Fixes #2072

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -1260,7 +1260,6 @@ where
 impl<T, TxDma, RxDma> embedded_io::Write for Uart<'_, T, TxDma, RxDma>
 where
     T: BasicInstance,
-    TxDma: crate::usart::TxDma<T>,
 {
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
         self.blocking_write(buf)?;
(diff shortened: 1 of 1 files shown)


Dirbaio:
see [this](https://github.com/embassy-rs/embassy/pull/2185#issuecomment-1810047043)  for why the raw uart shouldn't impl embedded-io Read. 

can you PR just the 1st commit? that one is a nice fix.

adri326:
Ah, fair enough; I'll document that implementation choice instead then

Dirbaio (approved):
very nice and clear explanations, thank you! hopefully they'll help guide users to what they want.

PR #2741, opened 2024-03-26
Added YLab project to "in the wild"

schmettow:
YLab uses Embassy


9names:
https://github.com/schmettow/ystudio either doesn't exist or isn't public

Dirbaio (approved):
very cool! thank you :)

Issue #2372, opened 2023-12-28
embassy-stm32: i2cv1 may be prone to deadlock


jamesmunns:
As of 2023-12-28, the STM32 driver for I2Cv1 (STM32F1, STM32F2, STM32F4, STM32L1) leans heavily on interrupts, as this is how i2cv1 was designed to be used. In multiple places, there's a process that goes roughly like this in thread mode:

* Set waker
* enable interrupts
* poll-fn loop to check events/errors:
    * if error, return Err
    * if done, continue on to the next step
    * if not done or error, return Pending
    
In the interrupts, we do something like:

* Check and store events/errors to a static
* disable interrupts

This makes interrupts essentially a "oneshot". BUT! If we got an interrupt, BUT the event/errors we care about weren't active (due to some race condition? Unclear as of now), *we never re-enable the interrupts*, which means that the pollfn hangs as it is never awoken again.

#2364 fixed one specific case of this, but it's not clear today whether that should be applied to ALL patterns in i2cv1 that use this oneshot interrupt pattern. It might be okay to just add that everywhere! But nobody has had time to dig into that or make a case one way or another.

Until someone digs into this, this issue exists to document "hey this is maybe weird heads up", and will be linked into the code.

# If you notice async I2C is hanging

Congratulations! You are a good candidate to investigate! See where the I2C is hanging, and see if applying a similar "enable interrupts" call like in #2364 helps. If so, please open a pull request.

Maybe related to #560 and #619.

jamesmunns:
If you notice your I2C is hanging, check out the workaround from #2887, you might be able to clear the lockup by reinitializing the I2C peripheral.

fmckeogh:
Would this apply to v2 (STM32F3)? I've been experiencing hanging only when using the async interface (#2884)

bananasmoothii:
Hi, given my high ability to attract bugs, I think I found one. I just don't have the time or knowledge to debug this.

I2C v1 hangs on this:
<img width="635" height="152" alt="Image" src="https://github.com/user-attachments/assets/8f482de3-097c-4e3b-a274-68ac964b3c3b" />

but when using `blocking_` equivalents, everything works.

Problematic code permalink: https://github.com/bananasmoothii/micromouse/blob/b29d6974bd822f4fcfe710796466a073c804efa2/src/sensor/vl53lxx/vl53l0x.rs#L56

Using an STM32F446RE with a Pololu VL53L0X distance sensor.

PR #2742, opened 2024-03-27
stm32/i2c(v1): Implement asynchronous transactions

sgoll:
This is a follow-up to #2713 and implements asynchronous transactions for I2C v1 in `embassy-stm32`. The underlying principles from the first PR have been used (i.e. frame options) and refactored to share identical code between blocking and asynchronous methods where applicable.

The implementation was tested for all combinations of merged and non-merged read/write operations on an STM32F2 dev board ([NUCLEO-F207ZG](https://www.st.com/en/evaluation-tools/nucleo-f207zg.html)) and logic analyzer/oscilloscope.

This is part of a series of PRs that may eventually solve #1417.

Unfortunately, the diff is not as easy to read as the one in the first PR because I had to refactor more of the existing code to make operations work in all combinations of reading/writing, and single-byte/multiple-byte read transfers. There is still more potential for refactoring in a future PR.

I also stumbled upon and fixed several cases where interrupts were not re-enabled after pending futures which could very well be the root cause of #2372 (as hinted at in that issue).


Dirbaio (approved):
thank you! :rocket: 

(the complexity of i2c boggles the mind, lol)

PR #2745, opened 2024-03-27
embassy_stm32: Preseve the RTR flag in messages.

de-vri-es:
This PR ensures that the RTR flag is respected when taking frames out of a mailbox or FIFO of a bxcan peripheral.

Without this PR, all frames are reported as data frames. With this PR, remote transmission requests are reported as such.


cschuhen:
Looks good.

de-vri-es:
Thanks for the review!

PR #2746, opened 2024-03-27
Add multiprio example for stm32h7 inspired by stm32f4

msrd0:
Copied the stm32f4 multiprio example to stm32h7, adjusted the necessary feature flags and the delay to roughly match the 1s/2s timings. I had to switch to `time-driver-tim2` as `time-driver-any` would panic when creating the low level executor because the `allocate_alarm()` call returned `None`.

Still wondering if the `UART4` interrupt was intended to be used that way ...

### DIFF
--- a/examples/stm32h7/Cargo.toml
+++ b/examples/stm32h7/Cargo.toml
@@ -6,9 +6,9 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 # Change stm32h743bi to your chip name, if necessary.
-embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32h743bi", "time-driver-any", "exti", "memory-x", "unstable-pac", "chrono"] }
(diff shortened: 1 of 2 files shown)


examples/stm32h7/Cargo.toml
```diff
@@ -6,9 +6,9 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 # Change stm32h743bi to your chip name, if necessary.
-embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32h743bi", "time-driver-any", "exti", "memory-x", "unstable-pac", "chrono"] }
+embassy-stm32 = { version = "0.1.0", path = "../../embassy-stm32", features = ["defmt", "stm32h723zg", "time-driver-tim2", "exti", "memory-x", "unstable-pac", "chrono"] }
```
Dirbaio, line 9: please don't change the chip name, it makes other example no longer build

PR #2755, opened 2024-03-31
Semaphore synchronization primitive

alexmoon:
This provides both a "greedy" and "fair" async semaphore implementation.

### DIFF
--- a/embassy-sync/src/lib.rs
+++ b/embassy-sync/src/lib.rs
@@ -17,6 +17,7 @@ pub mod once_lock;
 pub mod pipe;
 pub mod priority_channel;
 pub mod pubsub;
+pub mod semaphore;
 pub mod signal;
 pub mod waitqueue;
 pub mod zerocopy_channel;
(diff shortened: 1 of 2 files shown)


lulf (commented):
It looks good! But I'm wondering about the `min` argument to `acquire_all`. What is the use case of passing a `min` lower than "all" permits? That is somewhat unintuitive to me.

alexmoon:
So `acquire_all` is basically equivalent to:

```rust
let n = sema.permits(); // (not actually a method on Semaphore)
let guard = if n < min {
  sema.acquire(min).await
} else {
  sema.acquire(n).await // Never blocks...right?
};
```

The problem is there is a race condition between querying the permit count and acquiring the permits. Another thread could acquire some permits in the interval leading to unexpected blocking. You could do a spin loop on `try_acquire` to catch the case where you fail to acquire the expected number of permits, but that would cause problems for the `FairSemaphore` because you would always be at the back of the queue.

Typical use case for `acquire_all` would be something like `acquire_all(1)` to say "I need at least 1 permit, but I don't want any more permits to be acquired after I get mine until some are released".

lulf:
@alexmoon That makes sense! Thanks! Will start to use this in trouble as well for the HCI data flow control ;)

PR #2759, opened 2024-04-01
Add a (optional) way to measure CPU load on Cortex M executor

spcan:
This PR adds an opt-in way to create an Executor with very basic CPU load measurements. An example for the STM32F7 MCU is included.

Pros:
  - Builtin method that measures CPU load of a poll cycle
  - Each executor has its own independent measurement, no synchronization needed
  - It's opt-in, no changes to the executor if the feature is not used
  - No additional dependencies

Cons:
  - Adds a small delay before the executor starts polling and before the executor goes to sleep
  - The data provided is very simple, the user must implement their own logic to use the data properly
  - Not compatible with `#[embassy_executor::main]` macro

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -38,6 +38,8 @@ embassy-time-driver = { version = "0.1.0", path = "../embassy-time-driver", opti
 embassy-time-queue-driver = { version = "0.1.0", path = "../embassy-time-queue-driver", optional = true }
 critical-section = "1.1"
 
+embassy-time = { version = "0.3.0", path = "../embassy-time", optional = true }
+
 document-features = "0.2.7"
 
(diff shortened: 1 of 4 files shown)


elpiel:
I think this is very useful as an addon for system debuggging and status reports.
Are there any blockers for this PR?

PS: I'm curios if we can somehow check how much load the operation and function call take from the cpu, any idea?

lulf (commented):
Couldn't this be done in a arch-independent way using the trace hooks for executor active and idle? 

https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/raw/trace.rs#L177
https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/raw/trace.rs#L228

Dirbaio:
Hello,

PR #2760, opened 2024-04-01
stm32/can: simplify bxcan api, merging bx::* into the main structs.

Dirbaio:
The bx::* separate structs (Can, Rx, Tx) and separate `Instance` trait
are a relic from the `bxcan` crate. Remove them, move the functionality
into the main structs.

cc @cschuhen


cschuhen:
Hi @Dirbaio,

Thanks for that. Looks good but what's with mod2.rs? I didn't see it referenced anywhere in the diff. I didn't have time to checkout yet, just scanned the diff so I might have missed something.

cschuhen:
I just tested my example on your branch. Seems to work fine.
I also confirmed that I can remove embassy-stm32/src/can/bxcan/mod2.rs that you added and it still works fine. Accidental add?

Dirbaio:
yes, accidental add indeed, oops.

thanks for the testing!

PR #2762, opened 2024-04-02
rp2040: Rename PWM_CH to PWM_SLICE

### LINKED ISSUE #2758: PR2040 PWM naming
We are using embassy for a embedded class in Rust. Students have been confused by the naming convention used by embassy for the PWM. The PR2040's datasheet states that there are 8 PWM slices, each of them having two channels.

Embassy defines the `PWM_CH_X` data type for the slice, not for a channel. Students have been confused by the fact that they cannot use channel 0 (actually 0B) for Pin 1 (https://github.com/UPB-FILS-MA/upb-fils-ma.github.io/issues/84). Our suggestion would be to rename the data type to `PWM_SLICE_X` to reduce confusion. 

If this is something that you might consider, we would be happy to submit a PR. To preserve compatibility with older code, we can define type alias to `PWM_CH_X`.

alexandruradovici:
This pull request renames the `PWM_CH` data type that represents a PWM slice to `PWM_SLICE`. The purpose is to reduce the confusion between the PWM channel and PWM slice.

RP2040 has 8 PWM slices with two channels each (A and B).

It closes #2758 and https://github.com/UPB-FILS-MA/upb-fils-ma.github.io/issues/84.

### DIFF
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -183,14 +183,14 @@ embassy_hal_internal::peripherals! {
     DMA_CH10,
     DMA_CH11,
 
-    PWM_CH0,
-    PWM_CH1,
-    PWM_CH2,
-    PWM_CH3,
-    PWM_CH4,
-    PWM_CH5,
-    PWM_CH6,
-    PWM_CH7,
+    PWM_SLICE0,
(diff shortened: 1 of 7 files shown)


Dirbaio (approved):
thanks!

PR #2763, opened 2024-04-02
rp2040: rename all usages of pwm channel to pwm slices

alexandruradovici:
This pull request renames all the PWM data structures of RP2040 from `Channel` to `Slice`. It modifies the comments to reflect the renames. It follows #2762.

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -129,7 +129,7 @@ impl<'d, T: Channel> Pwm<'d, T> {
     #[inline]
     pub fn new_output_a(
         inner: impl Peripheral<P = T> + 'd,
-        a: impl Peripheral<P = impl PwmPinA<T>> + 'd,
+        a: impl Peripheral<P = impl PwmChannelA<T>> + 'd,
         config: Config,
     ) -> Self {
         into_ref!(a);
@@ -140,7 +140,7 @@ impl<'d, T: Channel> Pwm<'d, T> {
     #[inline]
     pub fn new_output_b(
         inner: impl Peripheral<P = T> + 'd,
-        b: impl Peripheral<P = impl PwmPinB<T>> + 'd,
+        b: impl Peripheral<P = impl PwmChannelB<T>> + 'd,
         config: Config,
     ) -> Self {
         into_ref!(b);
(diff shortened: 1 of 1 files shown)


embassy-rp/src/pwm.rs
```diff
@@ -301,78 +301,78 @@ impl<'d, T: Channel> Drop for Pwm<'d, T> {
...
+slice!(PWM_SLICE1, 1);
+slice!(PWM_SLICE2, 2);
+slice!(PWM_SLICE3, 3);
+slice!(PWM_SLICE4, 4);
+slice!(PWM_SLICE5, 5);
+slice!(PWM_SLICE6, 6);
+slice!(PWM_SLICE7, 7);
+
+/// PWM Channel A.
+pub trait PwmChannelA<T: Slice>: GpioPin {}
```
Dirbaio, line 339: I'd prefer to keep the pin traits as `PwmPin*`, for consistency with other drivers (like TxPin, SdaPin...). These "pin traits" indicate "this pin is usable for this function", it makes sense to call them all "*Pin".

Dirbaio: Or perhaps we can name them `ChannelAPin`, `ChannelBPin`. as in "pin usable for channel A". Consistent with "TxPin" as in "Pin usable for TX".

alexandruradovici: I think `ChannelAPin` and `ChannelBPin` work very nice.

Dirbaio (approved):
thanks!

Dirbaio:
as a side note: it's confusing/inconsistent that the PAC calls the slices "ch"! I think that's where the "wrong" name in embassy-rp originally came from. And that inconsistency is straight from Raspberry Pi... :smiling_face_with_tear:

alexandruradovici:
> as a side note: it's confusing/inconsistent that the PAC calls the slices "ch"! I think that's where the "wrong" name in embassy-rp originally came from. And that inconsistency is straight from Raspberry Pi... 🥲

I figured this out while renaming the pins and verified the data sheet. They call the PWM slices, but the registers are `CH0` ...

Dirbaio:
¯\_(ツ)_/¯

PR #2769, opened 2024-04-04
embassy-net: Implement RAW sockets

GustavToft:


embassy-net/src/raw.rs
```diff
@@ -0,0 +1,178 @@
...
+        f(socket, &s.iface)
+    }
+
+    fn with_mut<R>(&self, f: impl FnOnce(&mut raw::Socket, &mut Interface) -> R) -> R {
+        let s = &mut *self.stack.borrow_mut();
+        let socket = s.sockets.get_mut::<raw::Socket>(self.handle);
+        let res = f(socket, &mut s.iface);
+        s.waker.wake();
+        res
+    } */
```
GustavToft, line 124: Is this just irrelevant for RawSocket?

GustavToft (commented):
I am not sure if there are anything else missing in this implementation, or if it is as simple as this?

embassy-net/Cargo.toml
```diff
@@ -16,11 +16,11 @@ categories = [
 [package.metadata.embassy_docs]
 src_base = "https://github.com/embassy-rs/embassy/blob/embassy-net-v$VERSION/embassy-net/src/"
 src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-net/src/"
-features = ["defmt", "tcp", "udp", "dns", "dhcpv4", "proto-ipv6", "medium-ethernet", "medium-ip", "medium-ieee802154", "igmp", "dhcpv4-hostname"]
+features = ["defmt", "tcp", "udp","raw", "dns", "dhcpv4", "proto-ipv6", "medium-ethernet", "medium-ip", "medium-ieee802154", "igmp", "dhcpv4-hostname"]
```
Dirbaio, line 19: please fix spacing

Dirbaio:
Looking good overall. 

The commented chunks of code indeed don't apply to raw sockets, please remove them. 

Also, can you fix rustfmt?

embassy-net/src/raw.rs
```diff
@@ -0,0 +1,120 @@
...
+                s.register_recv_waker(cx.waker());
+                Poll::Pending
+            }
+        })
+    }
+
+    /// Send a datagram.
+    ///
+    /// This method will wait until the datagram has been sent.`
+    pub async fn send<T>(&self, buf: &[u8]) -> Result<(), raw::SendError> {
```
Dirbaio, line 94: smoltcp's SendError contains only BufferFull. This error can't happen here, because on buffer full we wait, instead of failing. Therefore this function can't fail, so could you remove the Result? Same for `poll_send`.

Dirbaio (approved):
thank you!

PR #2776, opened 2024-04-05
stm32: adc: v3: [h5] set OR.OP0 to 1 when ADCx_INP0 is selected, per RM

tcbennun:
Otherwise, predictably, reading on channel 0 gives nonsense

Tested on the H503


Dirbaio (approved):
nice!

PR #2781, opened 2024-04-06
Remove Generics from `embassy-rp` `Pwm` struct

ismet55555:
This is based on the conversation on Element with @Dirbaio to the problem of being able to add different PWM stucts withing the same array.

https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$fJxGz19y7xLRKlmtc-wOyqP-IXiaZ5IPY5OxoWwwFL4?via=matrix.org&via=tchncs.de&via=grusbv.com

Essentially:

```rust
        let mut pwm_config = Config::default();

        // Initial PWM duty cycle (bightness) - Range 0.0 (0) to 1.0 (65535)
        let duty_cycle = 0.5;
        pwm_config.compare_b = (duty_cycle * u16::MAX as f32) as u16;

        // (<LED ID>, <LED PIN>, <LED KIND>, <HANDLE>)
        let leds = [
            (
                0,
                0,
                "pwm",
                Pwm::new_output_a(slices.slice0, leds.led0, pwm_config),
            ),
            (
                1,
                6,
                "pwm",
                Pwm::new_output_a(slices.slice1, leds.led1, pwm_config),
            ),
            (
                2,
                7,
                "pwm",
                Pwm::new_output_a(slices.slice2, leds.led2, pwm_config),
            ),
            (
                3,
                8,
                "pwm",
                Pwm::new_output_a(slices.slice3, leds.led3, pwm_config),
            ),
        ];


```

This is not tested, I'm not really sure how to test it in the scope of this project. Also, please add any feedback! I kind of struggled through this using various resources to explain things like ChatGPT, so I'm not fully confident in this!

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -81,24 +81,22 @@ impl From<InputMode> for Divmode {
 }
 
 /// PWM driver.
-pub struct Pwm<'d, T: Slice> {
-    inner: PeripheralRef<'d, T>,
+pub struct Pwm<'d> {
     pin_a: Option<PeripheralRef<'d, AnyPin>>,
     pin_b: Option<PeripheralRef<'d, AnyPin>>,
+    channel: usize,
 }
 
-impl<'d, T: Slice> Pwm<'d, T> {
+impl<'d> Pwm<'d> {
     fn new_inner(
-        inner: impl Peripheral<P = T> + 'd,
+        channel: usize,
         a: Option<PeripheralRef<'d, AnyPin>>,
         b: Option<PeripheralRef<'d, AnyPin>>,
         b_pull: Pull,
         config: Config,
         divmode: Divmode,
     ) -> Self {
-        into_ref!(inner);
-
-        let p = inner.regs();
+        let p = pac::PWM.ch(channel);
         p.csr().modify(|w| {
             w.set_divmode(divmode);
             w.set_en(false);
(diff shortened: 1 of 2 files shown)


embassy-rp/src/pwm.rs
```diff
@@ -365,12 +374,6 @@ macro_rules! impl_pin {
     };
 }
 
-impl_pin!(PIN_0, PWM_SLICE0, ChannelAPin);
```
Dirbaio, line 368: why remove these?

embassy-rp/src/pwm.rs
```diff
@@ -81,24 +81,23 @@ impl From<InputMode> for Divmode {
 }
 
 /// PWM driver.
-pub struct Pwm<'d, T: Slice> {
-    inner: PeripheralRef<'d, T>,
+pub struct Pwm<'d> {
+    inner: PeripheralRef<'d, pac::pwm::Channel>,
```
Dirbaio, line 85: I think you shouldn't need htis field anymore, because you can obtain it any time from the PAC if you know the channel number in `self.channel`

Dirbaio:
Thanks for the PR, looks good overall!

For testing I would suggest running some examples, such as [this one](https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/pwm.rs), and check that it works. It'll also check the code builds (it doesn't right now). Or you can try it with your project where you originally found the issue with the generics. 

Also you'll have to rustfmt the code to get CI to pass. You can do this with `rustup install nightly` and `cargo +nightly fmt` in the `embassy-rp` directory.

ismet55555:
wow, you are awesome. I honestly gave up on it. Thank you!

Issue #2768, opened 2024-04-04
[`Sync`]: Add support for `map` methods


c410-f3r:
Hello,

Is it possible to add `map` methods for Mutex guards? For example, tokio has https://durch.github.io/rust-goauth/futures/lock/struct.MappedMutexGuard.html#method.map.

The standard library also has a `map_split` (https://doc.rust-lang.org/std/cell/struct.RefMut.html#method.map_split) for `RefMut` if that is relevant.

Thanks

Dirbaio:
this would be a nice addition, I've needed this myself some time in the past. Would you be willing to implement and PR it?

c410-f3r:
#2782

PR #2782, opened 2024-04-06
Add `map` method

c410-f3r:
As requested in https://github.com/embassy-rs/embassy/issues/2768#issuecomment-2038277704.

Personally, I have never ever created a synchronization primitive so I have no clue if this implementation is sound. Everything was basically a copy and paste from the `futures` crate.

The "sound" part should be double checked for the `Send` and `Sync` implementations.

### DIFF
--- a/embassy-sync/src/mutex.rs
+++ b/embassy-sync/src/mutex.rs
@@ -3,6 +3,7 @@
 //! This module provides a mutex that can be used to synchronize data between asynchronous tasks.
 use core::cell::{RefCell, UnsafeCell};
 use core::future::poll_fn;
+use core::mem;
 use core::ops::{Deref, DerefMut};
 use core::task::Poll;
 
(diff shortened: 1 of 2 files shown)


embassy-sync/src/mutex.rs
```diff
@@ -180,3 +198,122 @@ where
...
+    U: ?Sized,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        // Safety: the MutexGuard represents exclusive access to the contents
+        // of the mutex, so it's OK to get it.
+        unsafe { &mut *self.value }
+    }
+}
+
+unsafe impl<M, T, U> Send for MappedMutexGuard<'_, M, T, U>
```
Dirbaio, line 276: implementing Send/Sync is only sound if `M: Sync`. 

I think if you make MappedMutexGuard contain `&mut U` instead of `*mut U` Rust will be able to automatically infer the right Send/Sync bounds, so no `unsafe` needed.

(embassy-sync has a noop RawMutex that's not Sync, which allows creating async mutexes that can only be shared between tasks in the same thread, not between threads. this is somewhat unique vs other async runtimes)

embassy-sync/src/mutex.rs
```diff
@@ -180,3 +198,122 @@ where
...
+///
+/// This can be used to hold a subfield of the protected data.
+#[clippy::has_significant_drop]
+pub struct MappedMutexGuard<'a, M, T, U>
+where
+    M: RawMutex,
+    T: ?Sized,
+    U: ?Sized,
+{
+    mutex: &'a Mutex<M, T>,
```
Dirbaio, line 213: I think you can avoid one generic param by storing a reference to just the `state` field (of type `&'a BlockingMutex<M, RefCell<State>>`) instead of the whole `Mutex`, which would simplify things a bit.

embassy-sync/src/mutex.rs
c410-f3r: Bounds have been added to `M`.

All implementations I saw use pointers so I am not sure if references are possible.

embassy-sync/src/mutex.rs
c410-f3r: Done. Although I am not sure (again) about soundness

c410-f3r: If fact, I am not sure about anything 👍

embassy-sync/src/mutex.rs
```diff
@@ -180,3 +198,115 @@ where
...
+            let mut s = unwrap!(s.try_borrow_mut());
+            s.locked = false;
+            s.waker.wake();
+        })
+    }
+}
+
+unsafe impl<M, T> Send for MappedMutexGuard<'_, M, T>
+where
+    M: RawMutex + Send,
```
Dirbaio, line 273: this requires `M: Sync`, not `M: Send`.

Dirbaio:
to fix the ci failure for nightly, please rebase on latest main.

embassy-sync/src/mutex.rs
c410-f3r: When working with `Send` types the compiler asks for `T: Sync` when `&T` comes up but here this bound is a bit odd.

Dirbaio (approved):
thank you!

PR #2785, opened 2024-04-07
[embassy-stm32] added comments explaining multiprio interrupts

barnabywalters:


examples/stm32f4/src/bin/multiprio.rs
```diff
@@ -127,6 +127,10 @@ fn main() -> ! {
 
     let _p = embassy_stm32::init(Default::default());
 
+    // STM32s don’t have software-defined interrupts, so just use any free interrupt vectors which aren’t used
+    // by the rest of your application. In this case we’re using UART6 and UART7, but there’s nothing special
```
becothas, line 131:
```suggestion
    // by the rest of your application. In this case we’re using UART4 and UART5, but there’s nothing special
```

examples/stm32h7/src/bin/multiprio.rs
```diff
@@ -127,6 +127,10 @@ fn main() -> ! {
 
     let _p = embassy_stm32::init(Default::default());
 
+    // STM32s don’t have software-defined interrupts, so just use any free interrupt vectors which aren’t used
+    // by the rest of your application. In this case we’re using UART6 and UART7, but there’s nothing special
```
becothas, line 131:
```suggestion
    // by the rest of your application. In this case we’re using UART4 and UART5, but there’s nothing special
```

barnabywalters:
oops, was editing quickly and read the priorities rather than the vectors. Thanks!

examples/stm32f0/src/bin/multiprio.rs
```diff
@@ -126,6 +126,10 @@ fn main() -> ! {
     // Initialize and create handle for devicer peripherals
     let _p = embassy_stm32::init(Default::default());
 
+    // STM32s don’t have software-defined interrupts, so just use any free interrupt vectors which aren’t used
+    // by the rest of your application. In this case we’re using UART1 and UART2, but there’s nothing special
+    // about them. Any otherwise unused interrupt vector would work exactly the same.
```
adamgreig, line 131:
```suggestion
    // STM32s don’t have any interrupts exclusively for software use, but they can all be triggered by software as well as
    // by the peripheral, so we can just use any free interrupt vectors which aren’t used by the rest of your application.
    // In this case we’re using UART1 and UART2, but there’s nothing special about them. Any otherwise unused interrupt
    // vector would work exactly the same.
```

I'm not sure what a "software-defined" interrupt would be, but I think it's clearer to say that it doesn't have any interrupts reserved exclusively for software use (well, except PendSV and SysTick but they're core exceptions...), and note that any of the interrupts can be triggered by software.

barnabywalters: Sounds good, applied to all files

Dirbaio (approved):
tanks!

PR #2519, opened 2024-02-02
net/udp: Add some embedded-nal-async UDP traits

chrysn:
Building on https://github.com/smoltcp-rs/smoltcp/issues/903, this implements some UDP traits.

It notably doesn't implement the core stack trait; as outlined in the docs, that doesn't go well with the user-allocated stack data structures. For the same problem, RIOT OS provides a Stack type that is actually kind of a `Cell<Option<Socket>>` and thus allows the creation of exactly one socket -- that *is* an option here. However, my test applications really already take a bound socket as an argument, so let's see how far this gets us without doing the ugly.

This is a draft PR both because it has waaay too many FIXMEs / todo!()s, and because it hasn't been tested so far (but it builds, so push it :-D )

When done, this closes: https://github.com/embassy-rs/embassy/issues/2516

PR #2790, opened 2024-04-08
net/udp: Relay full UdpMetadata instead of only remote endpoint in poll_ functions

chrysn:
This is a breaking change for users of the poll_ functions. (Some might not notice if they already pass in an IpEndpoint into poll_send_to, or discard that item in poll_recv_from).

It enables embassy users to utilize enhancements in smoltcp that are hopefully coming up in https://github.com/smoltcp-rs/smoltcp/pull/904. Note that those two do not need to lockstep: We can change this now (without gaining many benefits, just access to the packet ID), and once smoltcp updates, the local address will be available as well.

This is the minimal change from https://github.com/embassy-rs/embassy/pull/2519, and enables that to progress more easily -- once this one is in, the larger PR should even be testable without adding too many `[patches.crates-io]` to already complex setups.

[edit: simplified link]


embassy-net/src/udp.rs
```diff
@@ -122,9 +124,13 @@ impl<'a> UdpSocket<'a> {
     ///
     /// When a datagram is received, this method will return `Poll::Ready` with the
     /// number of bytes received and the remote endpoint.
-    pub fn poll_recv_from(&self, buf: &mut [u8], cx: &mut Context<'_>) -> Poll<Result<(usize, IpEndpoint), RecvError>> {
+    pub fn poll_recv_from(
+        &self,
+        buf: &mut [u8],
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(usize, UdpMetadata), RecvError>> {
```
Dirbaio, line 131: `recv_from` and `poll_recv_from` should return the same. can you change `recv_from` too?

(the convention is `poll_*` functions are identical to their non-poll counterpart, except manually polled)

chrysn: Done in a fixup; this even makes the total diff smaller. Shall I squash and force-push now or when reviews are done?

chrysn:
The merge of main into the branch (AIU this is the pattern used in this repo) fixed the nightly builds. The builds are fixed by providing a fixup as requested in https://github.com/embassy-rs/embassy/pull/2790#pullrequestreview-1999747527 also for send_to, so that send_to is now consistent with poll_send_to, and another one that fixes unused import warnings.

Dirbaio (approved):
lgtm, thakns!

PR #2791, opened 2024-04-09
Fix for SPI and CRC reg changes in stm32-data

taunusflieger:
Fix for issue https://github.com/embassy-rs/stm32-data/issues/459

### DIFF
--- a/embassy-stm32/src/adc/g4.rs
+++ b/embassy-stm32/src/adc/g4.rs
@@ -1,0 +1,304 @@
+#[allow(unused)]
+use pac::adc::vals::{Adcaldif, Difsel, Exten};
+use pac::adccommon::vals::Presc;
+
+use super::{blocking_delay_us, Adc, AdcPin, Instance, InternalChannel, Resolution, SampleTime};
+use crate::time::Hertz;
+use crate::{pac, Peripheral};
+
+/// Default VREF voltage used for sample conversion to millivolts.
+pub const VREF_DEFAULT_MV: u32 = 3300;
+/// VREF voltage used for factory calibration of VREFINTCAL register.
+pub const VREF_CALIB_MV: u32 = 3300;
+
+/// Max single ADC operation clock frequency
+#[cfg(stm32g4)]
+const MAX_ADC_CLK_FREQ: Hertz = Hertz::mhz(60);
+#[cfg(stm32h7)]
+const MAX_ADC_CLK_FREQ: Hertz = Hertz::mhz(50);
+
+#[cfg(stm32g4)]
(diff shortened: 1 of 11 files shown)


embassy-stm32/src/crc/v1.rs
```diff
@@ -32,20 +32,32 @@ impl<'d> Crc<'d> {
     /// Feeds a word to the peripheral and returns the current CRC value
     pub fn feed_word(&mut self, word: u32) -> u32 {
         // write a single byte to the device, and return the result
+        #[cfg(not(crc_v1))]
+        PAC_CRC.dr32().write_value(word);
```
dlaw, line 36: At least on STM32G4, any write to dr32 writes _four_ bytes into the CRC (hence the suggestive name). (Don't ask how long it took for me to figure this out... the reference manual hardly explains it at all :tired_face:)

If only a single byte write is intended, I suggest to write `PAC_CRC.dr8()` instead. If a four-byte write is intended, then the comment 1 line above should be updated.

taunusflieger: Good catch will rework - unfortunately there are other breaking changes (inc. tests) which need to be fixed all together to get a successful CI run - I'm not sure if I find the time to address them.

taunusflieger: To me it looks more like a question if that comment is correct - the function is called `feed_word()` why should it write a byte? It takes a `u32` - a word. Shouldn't we have another function to write a byte?

dlaw: Indeed, the comment is not consistent with the name of the function. Looking at crc/v2v3.rs, there are separate functions for `feed_word()` and `feed_byte()`.

embassy-stm32/src/adc/g4.rs
```diff
@@ -0,0 +1,304 @@
...
+            reg.set_deeppwd(false);
+            reg.set_advregen(true);
+        });
+
+        blocking_delay_us(10);
+    }
+
+    fn configure_differential_inputs(&mut self) {
+        T::regs().difsel().modify(|w| {
+            for n in 0..20 {
```
idaniel86, line 173: Hi, there is a bug here, there are max 18 channels, not 20.

It results in panic:
0.000056 ERROR panicked at /Users/xxx/.cargo/git/checkouts/stm32-data-generated-cb34dad5f3150296/687d64d/stm32-metapac/src/chips/stm32g491re/../../peripherals/adc_g4.rs:259:78:
assertion failed: n < 18usize

Dirbaio: can you send a pull request? :pray:

PR #2792, opened 2024-04-09
Support non-secure flash operations on STM32U5

wagcampbell:
Hi, 

This pull request adds support for non-secure flash operations. 

In short, if TrustZone is active (TZEN option bit is 1), it uses secure operations (as before), otherwise it will use non-secure operations. 

As a word of caution, this doesn't handle all flash configurations when TrustZone is active. For instance, one can configure watermark-based flash memory protection to designate areas as secure or non-secure. Per section 7.5.1, Table 58., the default configuration when TZEN =1, is that all flash memory is secure. We could expand on this by checking SECWMx_PSTRT and SECWMx_PEND to determine whether one is operating on secure or non-secure flash memory. Though, I figure this might be worthy a discussion.

I tested the existing stm32u5/flash.rs example on a NUCLEO-U575ZI-Q board with TrustZone disabled (TZEN = 0).

Also, I referenced these STM32 examples. Their README's help provide some additional context. 
https://github.com/STMicroelectronics/STM32CubeU5/tree/main/Projects/NUCLEO-U575ZI-Q/Examples/FLASH

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -115,6 +115,9 @@ low-power-debug-with-sleep = []
 ## Automatically generate `memory.x` file using [`stm32-metapac`](https://docs.rs/stm32-metapac/)
 memory-x = ["stm32-metapac/memory-x"]
 
+## Use secure registers when TrustZone is enabled
+trustzone-secure = []
+
 ## Re-export stm32-metapac at `embassy_stm32::pac`.
 ## This is unstable because semver-minor (non-breaking) releases of embassy-stm32 may major-bump (breaking) the stm32-metapac version.
 ## If this is an issue for you, you're encouraged to directly depend on a fixed version of the PAC.
--- a/embassy-stm32/src/flash/u5.rs
+++ b/embassy-stm32/src/flash/u5.rs
@@ -14,10 +14,19 @@ pub(crate) const fn get_flash_regions() -> &'static [&'static FlashRegion] {
 }
 
 pub(crate) unsafe fn lock() {
+    #[cfg(feature = "trustzone-secure")]
+    pac::FLASH.seccr().modify(|w| w.set_lock(true));
+    #[cfg(not(feature = "trustzone-secure"))]
     pac::FLASH.nscr().modify(|w| w.set_lock(true));
 }
 
 pub(crate) unsafe fn unlock() {
+    #[cfg(feature = "trustzone-secure")]
+    if pac::FLASH.seccr().read().lock() {
+        pac::FLASH.seckeyr().write_value(0x4567_0123);
+        pac::FLASH.seckeyr().write_value(0xCDEF_89AB);
+    }
+    #[cfg(not(feature = "trustzone-secure"))]
     if pac::FLASH.nscr().read().lock() {
         pac::FLASH.nskeyr().write_value(0x4567_0123);
         pac::FLASH.nskeyr().write_value(0xCDEF_89AB);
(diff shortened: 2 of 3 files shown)


Dirbaio:
I'm not sure what's the best way to handle trustzone. There's 3 cases:

- trustzone enabled, running in Secure (S) mode => should use the secure registers
- trustzone enabled, running in Non-secure (NS) mode => should use the nonsecure registers
- trustzone disabled => should use the secure registers (what the current code does. perhaps it doesn't matter?)

so at the very least TZEN=1 doesn't mean we should use the nonsec registers, we might still be running in secure mode.

however, I think instead of detecting this at runtime we should add Cargo features so you can "tell" embassy-stm32 which of the above 3 cases you're building a firmware for. This'd reduce code size, and is unlikely to cause issues because you never want to dynamically run with or without trustzone, the mode is always known at compile time. Perhaps named `trustzone-secure`, `trustzone-nonsecure`.

wagcampbell:
I like the overall approach. That said, I'd like to clarify a few things I've observed: 

This is the case that prompted this PR:
> trustzone disabled => should use the secure registers (what the current code does. perhaps it doesn't matter?)

With TrustZone disabled (`TZEN = 0`), I am not able to erase or write to flash using Secure (S) registers. However, I am able to erase and write using Non-secure (NS) registers. 

With TrustZone enabled (`TZEN = 1`), and using the default values for `SECWMx_PSTRT` and `SECWMx_PEND` (all flash memory is secure), I can can erase and write to flash using the Secure (S) registers.

In short, the current code works so long as I enable TrustZone (`TZEN = 1`). To be clear, I'm using the erase and write operations in the stm32u5 `flash.rs` example for testing. 

Also, I need to study the reference manual closer, but aside from using watermark-based secure flash protection, I'm not sure how to configure the these modes:
> trustzone enabled, running in Secure (S) mode => should use the secure registers
> trustzone enabled, running in Non-secure (NS) mode => should use the nonsecure registers

Specifically, I can designate pages as Secure (S) or Non-secure (NS) using watermark-based protection, but, I'm not familiar with any other way to specify Secure (S) or Non-secure (NS) run modes. I say this mostly because I want to make sure I can test all the cases we're interested in supporting. 

Thanks!

wagcampbell:
This is very helpful context. 

I'll update this PR to support Option 1 in a single commit. Then, add support for Option 2 in a follow-on commit. 

I'll see if I can update or add an example that demonstrates this as well. 

Thanks!

wagcampbell:
This commit should address option 1: 2fa0bb7
I took a stab at at option 2 with this commit: 0ef0eea

I tested this on a [NUCLEO-U575ZI-Q](https://www.st.com/en/evaluation-tools/nucleo-u575zi-q.html).
For future reference, [this](https://community.st.com/t5/stm32-mcus/how-to-disable-trustzone-on-stm32u5-series/ta-p/49793) was helpful for disabling TrustZone.

With `TZEN = 1`:
`cargo run --release --bin flash --features trustzone-secure` // Works (as expected)
`cargo run --release --bin flash` // Fails (as expected)

With `TZEN = 0`:
`cargo run --release --bin flash` // Works (as expected)
`cargo run --release --bin flash --features trustzone-secure` // Fails (as expected)

I could be missing something here, but I didn't see the value in having two separate features.
Instead, having a single feature, `trustzone-secure` and then enabling or disabling that feature.
That said, I'm open to whatever features make the most sense here.

Thanks!

wagcampbell:
Missed adding support secure registers for the `clear_all_err()`.
New commit for option 2: c5119c6

Dirbaio (approved):
very nice, thank you!

> I could be missing something here, but I didn't see the value in having two separate features.

yeah, I agree there's no reason to do it for now. I was just guessing that in the future if we add actual trustzone support, other parts of the HAL might care about "no trustzeon" vs "trustzone, running in NS mode". We can add the other feature when the time comes though.

PR #2794, opened 2024-04-09
USB: fix comments about the config.max_packet_size_0 field

simpkins:
Fix the comment about the default value: this defaults to 64 rather than 8 bytes.

It seems like the max packet size for endpoint 0 should normally be selected automatically, rather than being part of the config.  At best it seems like this setting should just be a hint that gets used if when the bus is operating at full speed.  The contents of the device descriptor should ideally be updated with the correct max packet size after bus enumeration completes.  In practice always using 64 is probably fine if low speed environments never need to be supported. (Super speed requires a max packet size of 512 bytes, which I didn't list in the comments here.)

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -38,11 +38,12 @@ pub struct Config<'a> {
 
     /// Maximum packet size in bytes for the control endpoint 0.
     ///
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

(yes, I agree it should be autodetected. there's not much reason to set it lower if you can do 64...)

PR #2797, opened 2024-04-09
stm32 adc: remove delay trait

andresv:
STM32 ADCs do not need `impl DelayUs` anymore.

### DIFF
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -4,6 +4,7 @@ use core::task::Poll;
 
 use embassy_hal_internal::into_ref;
 
+use super::blocking_delay_us;
 use crate::adc::{Adc, AdcPin, Instance, SampleTime};
 use crate::time::Hertz;
 use crate::{interrupt, Peripheral};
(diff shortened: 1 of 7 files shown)


Dirbaio:
as discussed in Matrix, can you change this code to not require a `Delay` at all? The HAL already knows the clock frequencies of everything and is the source of time for embassy-time (if you're using it), so there's no reason it should ask *the user* how to do a delay.

we're already doing this in other places, for example here

https://github.com/embassy-rs/embassy/blob/2bca875b5f72578cbd20404010d174795d263313/embassy-stm32/src/usb/usb.rs#L270-L273

(perhaps that could be extracted into a `pub(crate) fn blocking_delay_us` utility function)

andresv:
I am using RTIC with [rtic-monotonics](https://lib.rs/crates/rtic-monotonics) as timer provider and sure it compiles now.
Next I am gonna check how USB uses timer and if it would also work nicley with RTIC.

Dirbaio:
`cortex_m::asm::delay` also works fine under RTIC. The reason I'm proposing using that unconditionally is because it simplifies the public API, the user no longer has to hunt around docs to find what delay they can use.

(and please do not add a `&mut impl Delay` to USB! we should request delays from the user in less places, not more)

andresv:
All good, I just haven't checked USB yet when I wrote my comment. I'll add `cortex_m::asm::delay` to ADC.

andresv:
@Dirbaio all green.

Dirbaio:
can you put the delay code in a shared `pub(crate) fn blocking_delay_us()` function instead of copypasting it 14 times?

andresv:
@Dirbaio what would be the most suitable `cfg` for this problem https://ci.embassy.dev/jobs/9417aed042bf?

Dirbaio:
> @Dirbaio what would be the most suitable cfg for stm32wb so it wouldn't be compiled for it https://ci.embassy.dev/jobs/9417aed042bf?

i'd just add `#[allow(dead_code)]` 

> Also should blocking_delay_us be somewhere under time so other drivers like USB could also use it?

maybe some shared module like `crate::util`. We can also leave it in `adc` now and move it later, either is fine by me, up to you.

Dirbaio (approved):
thank you!

Dirbaio:
so many force pushes :P

andresv:
Yes, when I reviewed whole diff threre were some things I did not like and I did not want to put them to a separate commit each time.

PR #2802, opened 2024-04-11
Add linked list for GPDMA on STM32H5

dfuhrmi:
For the transmission of continuous data streams, concepts such as double buffering are often used with the STHM32H7. These manage two separate buffers. The GPDMA unit of the STM32H5 enables linked lists for scheduled DMA transfers.  
My approach offers the possibility to manage several memory areas of the same size and type. The list can have an open end or be linked to the beginning.
A major disadvantage of the linked list concept is that it is not easy to find out which memory areas have already been filled. The asynchronous query with await would also be terminated after the first transfer.
However, the implementation could be used to emulate double buffering, which enables continuous data transfer with hardware support.

### DIFF
--- a/embassy-stm32/src/dma/gpdma.rs
+++ b/embassy-stm32/src/dma/gpdma.rs
@@ -2,6 +2,7 @@
 
 use core::future::Future;
 use core::pin::Pin;
+use core::ptr;
 use core::sync::atomic::{fence, Ordering};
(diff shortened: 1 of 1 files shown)


Dirbaio:
can you fix CI? rebase on latest main should fix it.

Dirbaio:
done in #3923

PR #2803, opened 2024-04-11
Spelling and whitespace fixes

VegaDeftwing:
Fixes some whitespace and spelling issues. I'm sure I missed some too, but this at least gets some of the quick n' dirty ones.

I also noticed a few errors with the word "crap" in them. I think it's funny, but they might be worth changing?

I might add a Github pipeline for https://vale.sh in a future PR if anyone is interested too.


VegaDeftwing:
Based on those ci failures, me thinks those shellcheck changes may not be ideal (or they're pointing out files that really ought to have been checked before weren't?) - I don't have time to investigate at the moment. I'll poke it more later.

Dirbaio (approved):

PR #2808, opened 2024-04-12
stm32: flash: h50: implement bank swapping

tcbennun:
Depends on pac changes: https://github.com/embassy-rs/stm32-data/pull/463

Questions for reviewers:

1. I copied and modified routines in `h50.rs` from other flash impls like `g.rs`. Why, in these, are certain functions labelled `unsafe` despite no obvious memory safety issues? For example, my `perform_bank_swap()` requires an unsafe block for `clear_all_err()` because that function is labelled unsafe, for whatever reason.
2. Should I be peppering `fence(Ordering::SeqCst);` between reg accesses in functions like `perform_bank_swap()`?
3. Any concerns with the approach of having `pub` bank-swap-related functions in this family-specific file? I want these accessible by applications.

### DIFF
--- a/embassy-stm32/src/flash/h50.rs
+++ b/embassy-stm32/src/flash/h50.rs
@@ -61,9 +61,12 @@ pub(crate) unsafe fn blocking_erase_sector(sector: &FlashSector) -> Result<(), E
 
     interrupt::free(|_| {
         pac::FLASH.nscr().modify(|w| {
-            w.set_bksel(match sector.bank {
-                FlashBank::Bank1 => Bksel::BANK1,
-                FlashBank::Bank2 => Bksel::BANK2,
+            // BKSEL ignores SWAP_BANK, so we must take it into account here
+            w.set_bksel(match (sector.bank, banks_swapped()) {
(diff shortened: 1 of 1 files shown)


tcbennun:
Back to draft status while doing tests

tcbennun:
Successfully using this feature on H503 for app-driven DFU! 🥳

Dirbaio (approved):
thank you!

> I copied and modified routines in h50.rs from other flash impls like g.rs. Why, in these, are certain functions labelled unsafe despite no obvious memory safety issues? For example, my perform_bank_swap() requires an unsafe block for clear_all_err() because that function is labelled unsafe, for whatever reason.

historical reasons I think. A while ago all stm32-metapac register access was marked as unsafe.

> Should I be peppering fence(Ordering::SeqCst); between reg accesses in functions like perform_bank_swap()?

no need. that's only needed to ensure no reordering between volatile and regular memory accesses (e.g. when using DMA). Volatile accesses are already guaranteed to not be reordered w.r.t other volatile accesses.

>    Any concerns with the approach of having pub bank-swap-related functions in this family-specific file? I want these accessible by applications.


:+1:

PR #2809, opened 2024-04-13
Add STM32U0 support.

Dirbaio:
- **stm32/spi,crc: update for new PAC**
- **stm32/flash: remove otp flash regions (removed in newer metapacs).**
- **stm32/adc: update g4 for new pac.**
- **Add stm32u0 support.**

### DIFF
--- a/examples/stm32u0/.cargo/config.toml
+++ b/examples/stm32u0/.cargo/config.toml
@@ -1,0 +1,9 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+# replace stm32u083rctx with your chip as listed in `probe-rs chip list`
+runner = "probe-rs run --chip stm32u083rctx"
+
+[build]
+target = "thumbv6m-none-eabi"
+
+[env]
+DEFMT_LOG = "trace"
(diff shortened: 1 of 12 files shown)


Dirbaio:
bender run

Dirbaio:
bender run

Dirbaio:
bender run

examples/stm32u0/.cargo/config.toml
```diff
@@ -0,0 +1,9 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+# replace STM32G071C8Rx with your chip as listed in `probe-rs chip list`
+runner = "probe-rs run --speed 100 --chip STM32u083rctx"
```
diondokter, line 3:
```suggestion
runner = "probe-rs run --chip STM32u083rctx"
```

Not sure if you found you needed to limit the speed, but it works fine for me without (and it's almost 10x the speed)

Dirbaio: ahh tha'ts why it flashed so slow! :rofl:  (it's copypasted from the c0 examples...)

PR #2812, opened 2024-04-14
usb::hid::RequestHandler::set_report(&mut self, ...)

borisfaure:
Goal is to be able to act upon an outside report.
For example a report could show that CapsLock and NumLock are on, but we may only want to do an action if there is a change in those states. This can be done by keeping and comparing with a previous report in the RequestHandler.

I've updated the examples that were using that API.

### DIFF
--- a/embassy-usb/src/class/hid.rs
+++ b/embassy-usb/src/class/hid.rs
@@ -378,7 +378,7 @@ pub trait RequestHandler {
     /// Reads the value of report `id` into `buf` returning the size.
     ///
     /// Returns `None` if `id` is invalid or no data is available.
-    fn get_report(&self, id: ReportId, buf: &mut [u8]) -> Option<usize> {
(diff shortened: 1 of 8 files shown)


HaoboGu:
AFAIK, Capslock and Numslock aren't sent by control endpoints. Have you tried to process Capslock and Numslock in `set_report`? Does it work?

borisfaure:
It can be seen with one of the `usb_hid_keyboard` examples. I've tested it with `stm32f4`.

Dirbaio:
The change makes sense, we should allow `&mut self` if there's nothing preventing it.

Can you do it for all methods in `RequestHandler` instead of just `set_report`, for consistency? thank oyu

HaoboGu:
> It can be seen with one of the `usb_hid_keyboard` examples. I've tested it with `stm32f4`.

that's good. 👍

borisfaure:
I've done the change to all the methods.
I've forced-pushed to benefit from 953e1cd.

PR #2823, opened 2024-04-16
Add critical sections to avoid USB OTG Errata

jamesmunns:
Lifted from this branch which was tested to prevent spurious stall errors.

https://github.com/embassy-rs/embassy/compare/embassy-rs:e38f101...jamesmunns:3302923

### DIFF
--- a/embassy-stm32/src/usb/otg.rs
+++ b/embassy-stm32/src/usb/otg.rs
@@ -1226,12 +1226,10 @@ impl<'d, T: Instance> embassy_usb_driver::EndpointIn for Endpoint<'d, T, In> {
                 w.set_xfrsiz(buf.len() as _);
             });
 
-            critical_section::with(|_| {
-                // Enable endpoint
-                r.diepctl(index).modify(|w| {
-                    w.set_cnak(true);
-                    w.set_epena(true);
-                });
+            // Enable endpoint
+            r.diepctl(index).modify(|w| {
+                w.set_cnak(true);
+                w.set_epena(true);
             });
 
             // Write data to FIFO
--- a/embassy-stm32/src/usb/otg.rs
+++ b/embassy-stm32/src/usb/otg.rs
@@ -672,45 +672,51 @@ impl<'d, T: Instance> Bus<'d, T> {
 
         let r = T::regs();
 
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usb/otg.rs
```diff
@@ -1208,27 +1214,33 @@ impl<'d, T: Instance> embassy_usb_driver::EndpointIn for Endpoint<'d, T, In> {
...
             });
-        });
 
-        // Write data to FIFO
-        for chunk in buf.chunks(4) {
-            let mut tmp = [0u8; 4];
-            tmp[0..chunk.len()].copy_from_slice(chunk);
-            r.fifo(index).write_value(regs::Fifo(u32::from_ne_bytes(tmp)));
-        }
+            critical_section::with(|_| {
```
Dirbaio, line 1229: this nested CS isn't doing anything

jamesmunns: @Dirbaio removed!

Dirbaio (approved):
thank you!

mkj:
@jamesmunns do you remember how to reproduce this? The stm32h7rs doesn't seem to mention that in the errata, I wonder if the critical section can be made conditional there.

Dirbaio:
it was more like "you must access registers in *exactly* the documented order or the peripheral will behave weirdly". Interrupts would cause the order to change, causing endless fun. I don't think this is actually documented as an errata by ST.

jamesmunns:
There was an errata. It shows up in a couple errata docs. If you search for the text "Transmit data FIFO is corrupted when a write sequence to the FIFO is interrupted with accesses to certain OTG_FS registers" and the chip you're interested in, it should show up.

The STM32F405 has it for both USB HS and USB FS:

Screenshot 2025-02-23 at 22 42 21

jamesmunns:
As for how to reproduce, it's very timing based. It would show up every couple days (running 2-3 devices mostly 24/7, talking over postcard-rpc) in my application. You could probably synthetically trigger it, but it would be a pain in the butt.

mkj:
Thanks. I'll see if I can get an answer on the ST forum whether its meant to be fixed.

PR #2828, opened 2024-04-16
Bug: There are at most 18 channels for the STM32G4 ADCs.

idaniel86:
The max number of channels for the ADCs in STM32G4 MCUs is 18 instead of 20, this is causing panic behaviour in smt32-metapac.


Dirbaio (approved):
thanks~!

Issue #2806, opened 2024-04-12
STM32H7 Power Configuration issues leading to general unreliability to start-up


birdistheword96:
This is a placeholder issue to track the knowledge we have on why STM32H7's are so unreliable to start-up. all of my testing has been done on three different STM32H745 boards: a Nucleo, a Discovery board, and a custom in-house board. This is what I have found so far, and will update as I discover more:

- Encountered the exact same issue on all three different STM32H745 boards
- Whenever I try to bring a project to life using Rust/Embassy, it throws a tantrum, failing a few times and getting stuck at while `!PWR.csr1().read().actvosrdy() {}` in the `rcc::init` function.
- Oddly enough, playing musical chairs with various example projects in the embassy repo seems to eventually coax it into working... until the next power cycle, that is.
- Tweaking the voltage scales seems to have no effect on the frequency of the gremlin...
- In a twist, flashing a bare-metal (C) project from STMCube initially follows the same script of failure (post-Embassy project flash). Yet, post a power cycle, it runs as smooth as butter - no hiccups or boot issues whatsoever, power cycle after power cycle.

bsodmike:
I had this issue for a while on the H747xi but managed to resolve it after much testing. Let me dig up the config I'm using. 

I have a spare NUCLEO-H743ZI2 which I can test for you if you like.

birdistheword96:
This is the code I have configured for my custom board, which to be fair has an external oscillator, but I have equivalent C code which I can use t- get it running reliably. so i am trying to work out the difference. This is the config I have:

```rust
    let mut config = Config::default();
    {
        use embassy_stm32::rcc::*;
        // config.rcc.hsi = Some(HSIPrescaler::DIV1);
        config.rcc.hsi = None; // Since we're using HSE
        config.rcc.hse = Some(Hse { freq: mhz(20), mode: HseMode::Bypass});
        config.rcc.csi = false;
        config.rcc.pll1 = Some(Pll {
            source: PllSource::HSE,
            prediv: PllPreDiv::DIV2,
            mul: PllMul::MUL32,
            divp: Some(PllDiv::DIV2),
            divq: Some(PllDiv::DIV8),
            divr: Some(PllDiv::DIV2),
        });
        config.rcc.sys = Sysclk::PLL1_P; // 400 Mhz
        config.rcc.ahb_pre = AHBPrescaler::DIV2; // 200 Mhz
[...]

PR #2829, opened 2024-04-16
stm32: ensure the core runs on HSI clock while setting up rcc

aurelj:
I'm working on a project with a bootloader that does it's things then jump to the main firmware, and I've got issues with the stm32 core locking up when the main firmware starts.
The issue is that both the bootloader and the main firmware are initializing the system clock to use a PLL (based on HSE clock).
When the main firmware starts, the core is already running using the PLL (setup by the bootloader), then the main firmware calls embassy_stm32::init() which setup the PLL by first turning it off (causing the CPU to stop running !!!) before setting it up and turning it on again.

With this patch, rcc::init() will first setup the core to run on the HSI clock, then do the whole clock setup, then switch the core to the requested clock (PLL, HSE...), and then turning off the HSI clock (if requested).
This ensures that the core always have a valid clock source during the whole process of setting up the rcc.

This may also help with config as described in #2806, which use PLL/HSE and disable HSI. The current code was first turning the HSI OFF (while the core is probably running on it), and then setup the requested clock. With this patch, the HSI clock will only be turned off after the requested clock was setup.

On a freshly reset core, this patch won't have any impact (the core is already running on HSI at reset), except for the fact that the HSI will be turned off later in the process, after the rcc was setup as requested.

Note that this patch only applies to F2/F4/F7 series, but if this is accepted, I can make a similar change to other series.

### DIFF
--- a/embassy-stm32/src/rcc/f247.rs
+++ b/embassy-stm32/src/rcc/f247.rs
@@ -146,17 +146,18 @@ pub(crate) unsafe fn init(config: Config) {
         while !PWR.csr1().read().odswrdy() {}
     }
 
+    // Turn on the HSI
+    RCC.cr().modify(|w| w.set_hsion(true));
+    while !RCC.cr().read().hsirdy() {}
+
+    // Use the HSI clock as system clock during the actual clock setup
+    RCC.cfgr().modify(|w| w.set_sw(Sysclk::HSI));
+    while RCC.cfgr().read().sws() != Sysclk::HSI {}
+
     // Configure HSI
     let hsi = match config.hsi {
-        false => {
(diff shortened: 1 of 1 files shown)


bsodmike:
I have a sneaky feeling this is also an issue on the H7x series, https://github.com/embassy-rs/embassy/issues/2806 but let's see.

Dirbaio (approved):
makes sense, thank you!

bsodmike:
@aurelj I believe I've seen this issue on the h747. I'm happy to test if you can port this across.

aurelj:
@bsodmike That's probably the case. You can test PR #3008 which implements the same changes to the h7 (and others).

PR #2832, opened 2024-04-17
stm32: can: fd: implement bus-off recovery

tcbennun:
FDCAN has no automatic bus-off recovery, see RM0492 p. 1600:

image

Also cleaned up the interrupt handler a little


cschuhen:
Changes look good. I did a quick h/w test on STM32G431, example still working fine.

PR #2833, opened 2024-04-17
Allow Uart::new_half_duplex for any usart version

### LINKED ISSUE #2816: STM32: no Uart::new_half_duplex on stm32f411 and stm32f446
The `Uart::new_half_duplex` method requires `usart_v3` or `usart_v4`:
```rust
#[cfg(not(any(usart_v1, usart_v2)))]
    #[doc(alias("HDSEL"))]
    pub fn new_half_duplex(
```

The `stm32f411` and `stm32f446` chips implement `usart_v2`:
```
#[path="../../peripherals/usart_v2.rs"] pub mod usart;
```
https://github.com/embassy-rs/stm32-data-generated/blob/main/stm32-metapac/src/chips/stm32f411ce/pac.rs
https://github.com/embassy-rs/stm32-data-generated/blob/main/stm32-metapac/src/chips/stm32f446re/pac.rs

But STM32CubeMX allows `Half-Duplex` for this chips
Снимок экрана от 2024-04-15 16-42-41

Possible `usart_v2` should allows `new_half_duplex`, I not think that it is STM32CubeMX bug.

qwerty19106:
Any `usart_v1`-`usart_v4` have [set_hdsel](https://github.com/embassy-rs/stm32-data-generated/blob/main/stm32-metapac/src/peripherals/usart_v1.rs#L191) method, and can be used in half-duplex mode.
But now it allowed for usart_v3 and usart_v4 only.
Fixes #2816.


qwerty19106:
Done.

Note: `new_half_duplex_on_rx` and `new_blocking_half_duplex_on_rx` requires `config.swap_rx_tx`. Thus nothing changes for them.

qwerty19106:
I forget fix documentation. For example, It can be changed:
    ...
    /// See [`new_half_duplex_on_rx`][`Self::new_half_duplex_on_rx`] if you would prefer to use an Rx pin **(and it is available for your chip)**.
    /// There is no functional difference between these methods, as both allow bidirectional communication.
    ...
    #[doc(alias("HDSEL"))]
    pub fn new_half_duplex(

PR #2834, opened 2024-04-18
stm32: fdcan: try_write() and try_read() on BufferedCan impl

tcbennun:
Not sure that this makes sense for the non-buffered FDCAN driver, but for buffered, it's nice to have instantaneous `try` variants.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -447,10 +447,24 @@ impl<'c, 'd, T: Instance, const TX_BUF_SIZE: usize, const RX_BUF_SIZE: usize>
         self
     }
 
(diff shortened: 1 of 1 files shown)


tcbennun:
@Dirbaio would you mind checking why the test check is failing? thank you

Dirbaio:
rebase on lastest git main should fix it

cschuhen:
@tcbennun, No worries... I'm OK with this. However, I didn't bother because you can also do it from the Sender i.e. .writer().try_write(...) or Receiver

tcbennun:
> @tcbennun, No worries... I'm OK with this. However, I didn't bother because you can also do it from the Sender i.e. .writer().try_write(...) or Receiver

Okay yeah, I missed that, I'll close to avoid bloat!

PR #2835, opened 2024-04-18
stm32/spi: remove peripheral generic param.

Dirbaio:


honzasp:
Looking at the generated code for `SealedRccPeripheral` in `_generated.rs`, I found that we also maintain a global refcount in `rcc::REFCOUNT_STOP1`; if I understand correctly, this is used to find out whether we can enter low-power state:

```
    fn enable_and_reset_with_cs(_cs: critical_section::CriticalSection) {
        #[cfg(feature = "low-power")]
        unsafe {
            crate::rcc::REFCOUNT_STOP1 += 1
        };
        ...
    }
    fn disable_with_cs(_cs: critical_section::CriticalSection) {
        ...
        #[cfg(feature = "low-power")]
        unsafe {
            crate::rcc::REFCOUNT_STOP1 -= 1
        };
    }
```

It seems that the constructor of the SPI peripheral increments `REFCOUNT_STOP1`, but the destructor doesn't decrement it. Is that OK? If not, perhaps we can extend the `ClockEnableBit` structure to also increment/decrement the appropriate global refcount?

PR #2840, opened 2024-04-19
stm32: can: fd: Properties for common runtime get/set operations

tcbennun:
There are certain operations on the FDCAN that should always be available regardless of the mode currently active, and regardless of FD/non-FD operation, or buffered/non-buffered driver usage:

- setting filters
- retrieving error counters
- retrieving other error statuses
- _more?_

This change proposes a `Properties` struct, associated with the configurator, non-buffered driver, and buffered drivers, which the application can use to perform these actions.

Let me know what you think (esp. @cschuhen).


cschuhen:
Yes, I'm good with that direction. Much better than having to duplicate everything for each of the API's IMO.

embassy-stm32/src/can/fdcan.rs
```diff
@@ -804,6 +828,60 @@ impl TxMode {
...
+
+    /// Set an array of Extended Address CAN filters and overwrite the current set
+    pub fn set_extended_filters(&mut self, filters: &[ExtendedFilter; EXTENDED_FILTER_MAX as usize]) {
+        for (i, f) in filters.iter().enumerate() {
+            T::registers().msg_ram_mut().filters.flesa[i].activate(*f);
+        }
+    }
+
+    /// Get the CAN RX error counter
+    pub fn get_rx_error_count(&self) -> u8 {
```
Dirbaio, line 870: rust convention is to not add `get_` to getters, ie name this just `rx_error_count`.

embassy-stm32/src/can/fdcan.rs
```diff
@@ -199,9 +200,20 @@ impl<'d, T: Instance> CanConfigurator<'d, T> {
...
         }
     }
 
+    /// Get driver properties
+    pub fn properties(&self) -> &Properties<T> {
+        &self.properties
+    }
+
+    /// Get mutable driver properties
+    pub fn properties_mut(&mut self) -> &mut Properties<T> {
```
Dirbaio, line 213: do we really need the `properties/properties_mut` split? I think we can either
- Only make it `&mut`. 
- Only make it `&`, and make all methods in Properties use `&self` (even getters). Should be sound since setters only do volatile reg writes, don't modify data in `self`. And if we don't make it `Sync` you can't call `&self` methods concurrently from different threads.

Dirbaio: this is making the API more complicated for "ideological purity" reasons, not practical reasons. 

If you view `&`/`&mut` references as "shared" / "exclusive", and due to how the hardware works we can allow setting properties with a shared reference "for free" (no refcell/mutex overhead needed), then why not do it? It makes the API simpler for the user, easier to learn.

tcbennun: Made changes according to the above, let me know what you think

cschuhen:
I don't see any issues with this, Thanks Torin! I think this approach will save a lot of duplication down the track as any new features get added.

Dirbaio:
lgtm! please undraft it when ready.

Dirbaio (approved):
neat!

and thanks for adding the test to the h5 too

PR #2841, opened 2024-04-19
Adding ready_to_receive to Channel and Receiver

nerwalt:
Adding ready_to_receive() to enable waiting to receive without actually receiving


lulf (approved):
Nice addition! FYI you can also use `poll_fn` with poll_ready_to_receive, but this is a bit more ergonomic.

PR #2846, opened 2024-04-21
Added PIO pwm examples for rp

BjornTheProgrammer:
## Description
Two additional `rp` examples were added: `pio_pwm.rs`, a bare-metal example of how to do PWM with pio, and `pio_servo.rs`, a more extended example of PWM and pio with servos.

## Motivation
It is pretty common for people to wish to use Servos with their devices, but most people do not wish for the main cores to be kept busy running a servo or other PWM devices. So, these examples utilize the powerful PIO features of the RP2040 to do PWM instead of using the built-in PWM to prevent this from being an issue. These examples give further inspiration to others who wish to implement other things using PIO.


BjornTheProgrammer:
> Nice! Maybe `attach` should be called `new` instead, for consistency?

@kalkyl thank you for the feedback! I have made the requested change.

Dirbaio (approved):
very nice! thank you :D

PR #2848, opened 2024-04-21
Add a comment to the pio rx fifo pull function

jannic:
Accidentally calling `pull` on an empty RX FIFO can lead to difficult to debug errors.

I wonder if the `pull` function should be removed / deprecated, with a suggestion to use `try_pull` instead? As that would obviously be a breaking change, this PR only adds a hint to the doc comment for now.

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -324,6 +324,10 @@ impl<'d, PIO: Instance, const SM: usize> StateMachineRx<'d, PIO, SM> {
     }
 
     /// Pull data from RX FIFO.
+    ///
+    /// This function doesn't check if there is data available to be read.
+    /// If the rx FIFO is empty, an undefined value is returned. If you only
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

perhaps we could rename it to `pull_unchecked`? Breaking changes are fine, next release will be 0.2 anyway. Also I think `push()` has a similar issue, it doesn't check there's space in the fifo?

jannic:
When renaming `pull` to `pull_unchecked`, should `try_pull` also be renamed to `pull`?

I didn't check the `push` functions yet because I think they are not as big a footgun. I may be wrong, but my feeling is that its particularly easy to accidentally call the `pull` function slightly too early.

Dirbaio:
> When renaming pull to pull_unchecked, should try_pull also be renamed to pull?

probably? the `try_` is nice because it reminds it's on you to check if it worked, but the returned `Option` probably alreayd does that job well.

Agree `push` is less of a footgun but it'd still be nice to keep them consistent if we rename/remove.

PR #2849, opened 2024-04-21
Add information about vbus_detection to the FAQ and link it in all USB examples.

joelsa:
This was discussed in the Matrix chat and I deemed it a useful idea, since this is an often-reoccurring question.

It was also discussed to change the default value of `config.vbus_detection` at [embassy-stm32/src/usb/otg.rs#L300](https://github.com/embassy-rs/embassy/blob/a5b006268b7a20ff4ea6ad0d9090c6a6f2bb4602/embassy-stm32/src/usb/otg.rs#L300) to `false`, howver I think `true` is a sensible default, as most boards really should implement it and implementing it should be encouraged. Adding the comments to the examples seems a sensible way to prevent most of the issues regarding this for newcomers.


Dirbaio (approved):
very nice, thank you! :)

Dirbaio:
you'll have to rustfmt to get ci to pass.

PR #2851, opened 2024-04-22
add option to customize stm32 time driver MIN_STOP_PAUSE via a env variable

dragonnn:
On my STM32F401RET6 for some reason the default 250ms MIN_STOP_PAUSE is really unstable, often timed wakeups or wakeups from EXTI GPIO are missed, they is some trace output when those happen but it isn't passed to the task waiting for those.
Changing the MIN_STOP_PAUSE to 2500ms complete fixes those issues.
I know that isn't a solution but more kind of workaround but that is how far I currently got and for my application that is a completely fine trade off. And I think having the ability to adjust that value might be for other user usable too.


dragonnn:
Forgot to add, it works by setting the env variable EMBASSY_STM32_TIME_DRIVER_MIN_STOP_PAUSE_MS to the desired value. Maybe a shorter name might be desired but I didn't want to hide from what crate it comes from.
I based this code by looking at DEFMT_RTT_BUFFER_SIZE from defmt-rtt. I think that is a better way then having a bunch of fixed values via some features. And it still defaults to 250ms if it is not set

Dirbaio:
This is a workaround. Missed wakeups should never happen, I would prefer if we found the root cause and fixed it.

dragonnn:
> This is a workaround. Missed wakeups should never happen, I would prefer if we found the root cause and fixed it.

Yes, I know and I would like that too. I will try dig deeper into it to find what is causing that but I still that workaround can be useful for other people. Not even as a workaround, just something that can be nice to have to adjust, I suspect different MCU and different use cases can have different best value for the MIN_STOP_PAUSE so it can be nice to have an option to fine tune it instead of a fixed value.

Dirbaio:
i'm going to close this for now, hopefully someone can figure out what's the root cause of the issue :cry:

Issue #2784, opened 2024-04-06
rp2040 `SetDutyCycle` is not implemented for pwm


mattiekat:
Found a bit of a rough edge when trying to use a motor controller library and found that [`embedded_hal::pwm::SetDutyCycle`](https://docs.rs/embedded-hal/latest/embedded_hal/pwm/trait.SetDutyCycle.html) is not implemented for any of the pwm types.

Looking at the `rp2040-hal` crate I see it was implemented for [rp2040_hal::pwm::Channel](https://docs.rs/rp2040-hal/latest/rp2040_hal/pwm/struct.Channel.html) maybe I am misunderstanding and this should be wrapped instead?

alxpettit:
Yeah, I'm super confused how to actually use PWM since we don't have a `set_duty` :c

cylewitruk:
Just ran into this as well trying to use an RGB LED module...

alxpettit:
Yeah, it's weird. The docs have this to say about `compare`:

```
    /// The output on channel A goes high when `compare_a` is higher than the
    /// counter. A compare of 0 will produce an always low output, while a
    /// compare of `top + 1` will produce an always high output.
    pub compare_a: u16,
    /// The output on channel B goes high when `compare_b` is higher than the
    /// counter. A compare of 0 will produce an always low output, while a
    /// compare of `top + 1` will produce an always high output.
    pub compare_b: u16,
```

It also says at the top:

```
/// Note the period in clock cycles of a slice can be computed as:
/// `(top + 1) * (phase_correct ? 1 : 2) * divider`
```

This was super confusing to me, as it seems from the latter comment that `compare_a` & `compare_b` ought not to come into the period at all. I think what's happening is we're adjusting one clock, and if that clock value is higher than _another_ clock, the pin will rise? I don't exactly get it, and unfortunately the actual code seems to be making use of low-level MCU features with a lot of magic numbers -- so it's hard to infer from that too.

cylewitruk:
Yeah I would have expected an implementation more like the [NRF](https://github.com/embassy-rs/embassy/blob/a5b006268b7a20ff4ea6ad0d9090c6a6f2bb4602/embassy-nrf/src/pwm.rs#L586) & [STM32](https://github.com/embassy-rs/embassy/blob/a5b006268b7a20ff4ea6ad0d9090c6a6f2bb4602/embassy-stm32/src/timer/simple_pwm.rs#L61) impls...

[...]

PR #2855, opened 2024-04-23
RP2040: Higher-level API for PWM

cylewitruk:
Creating a draft PR for review/discussion as a WIP implementation for  #2784.

These changes try to bring the `Pwm` module a little closer to some of the other implementations we see in `embassy`, though RP2040 is a little different in that it has eight slices with two channels each (A & B), where some registers are set for the slice and others for the channels.

I'm not 100% on the calculations so I'd love an extra eye on them. Based on the [RP2040 Datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf), section `4.5` and some other implementations I looked at.


crabdancing:
Well, the frequency calculations seem to be spot-on from my testing. I only have a cheap Hantek to work with, but it seems to exactly match everything I've tested up to 50Mhz (after which, it strains the capabilities of my oscilloscope and causes a lot of impedence, so it's hard to tell what's going on).

it looks like `set_duty_a` and `set_duty_b` don't actually work independently in my tests.

If I do:

```rs
    pwm.set_duty_a(0).unwrap();
    pwm.set_duty_b(100).unwrap();
```
Both A and B are on, according to my oscilloscope.


But if I do:


```rs
    pwm.set_duty_a(100).unwrap();
    pwm.set_duty_b(0).unwrap();
```

Both A and B are off.

cylewitruk:
@crabdancing awesome you were able to try it out and that it seems to work-"ish" :smile:  I'm waiting for my oscilloscope so I don't have anything to test with yet, so that's a huge help!

I'll see if I can find what might be going wrong with a/b duties tomorrow -- according to the datasheet a/b should be able to operate at different duties and I don't see any issues with the code at a glance..

crabdancing:
Yeah, I've been looking through it and I don't get it either. The mainline version of my program has no trouble independently controlling the PWM signals with `compare_a` and `compare_b`.

crabdancing:
I have a cute little test program that plays an animation on oscilloscopes, alternately moving two channels back and forth past each other by smoothly changing the duty cycle. On the mainline embassy-rs, it seems to work fine, but on the fork it seems completely broken. You can see the test version [here](https://github.com/crabdancing/pwm-dance/blob/testing/src/main.rs), and the original version [here](https://github.com/crabdancing/pwm-dance/blob/main/src/main.rs).

cylewitruk:
@crabdancing thanks for the detailed input! I'll see if I can try to reproduce some of this with tests today and figure out what's going on :smile:

cylewitruk:
> You can see the test version [here](https://github.com/crabdancing/pwm-dance/blob/testing/src/main.rs?rgh-link-date=2024-04-23T23%3A18%3A02Z), and the original version [here](https://github.com/crabdancing/pwm-dance/blob/main/src/main.rs?rgh-link-date=2024-04-23T23%3A18%3A02Z).
@crabdancing 

The second (original) link doesn't seem to work, I get a 404

crabdancing:
Oof, sorry. I accidentally only pushed `testing` branch. Forgot to do `git push --all`. Try now?

cylewitruk:
@crabdancing I'm still looking at this, just so you know -- have been doing some reworking of the API etc. Got my oscilloscope today so I can play with it this weekend and try to test things myself :smile:

crabdancing:
It looks great! :) The alternative syntax seems like it might be better for more easily iterating over slices to avoid duplicate code. That was a frustration I had with a lot of embedded HALs in Rust, when I'm using a lot of pins. I get what they're trying to do with typestate to prevent programmer mistakes, tho!

The builder design pattern seems nice! I generally know what all the options are and what they mean at a glance. I would be totally happy to use this in my code :D

cylewitruk:
> It looks great! :) The alternative syntax seems like it might be better for more easily iterating over slices to avoid duplicate code. That was a frustration I had with a lot of embedded HALs in Rust, when I'm using a lot of pins. I get what they're trying to do with typestate to prevent programmer mistakes, tho!
> 
> The builder design pattern seems nice! I generally know what all the options are and what they mean at a glance. I would be totally happy to use this in my code :D

Thanks for the feedback!

> The alternative syntax seems like it might be better for more easily iterating over slices to avoid duplicate code. That was a frustration I had with a lot of embedded HALs in Rust, when I'm using a lot of pins.

Would you mind elaborating on your use-case a little, maybe a brief example of your desired usage? The `peripherals.slice_x()` methods are just an extension trait on `Peripherals` providing some convenience methods over `Peripherals`, but the "alternative syntax" still uses generics to ensure the correct pins are used. It sounds like you'd like to do something like a `for slice_id in 0..=7`?

cylewitruk:
Branch rename killed this PR, so it's continued over at #2882

crabdancing:
@cylewitruk 

>It sounds like you'd like to do something like a for slice_id in [1u8, 2u8, 5u8, 7u8]? If so, then you'd probably also want a get_slice_pins(n: u8) -> PwmPins or similar so that generics (type safety) can still be used but you can still loop in a slice-agnostic manner? Where PwmPins is something like:

Yes, that's exactly right! This can be resolved with macros, but it is nice to be able to do conventional, easy-to-follow Rust programming to handle lots of pins, instead of being stuck with macros or boilerplate or -- often -- both at once. Generics would also be a solution for this, but the extra heap allocations may be unacceptable for some environments -- which I think just leaves the user with transmuting. Theoretically safe if every type is of the same size, but definitely not ideal to introduce `unsafe` just to iterate over pins.

PR #2857, opened 2024-04-23
Add blocking_write_and_flush to Uart

bsodmike:


Dirbaio:
unfortunately i'm not convinced we should add "do x and y" convenience methods when it's something the user can do themselves so easily. It makes the API surface more complex, and I prefer to avoid "there's 2 ways of doing the same thing" if possible.

PR #2861, opened 2024-04-23
Allow potentially unused import

bsodmike:
Replaces #2860

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -2,6 +2,7 @@ use core::slice;
 use core::sync::atomic::AtomicBool;
 
 use embassy_hal_internal::atomic_ring_buffer::RingBuffer;
+#[allow(unused_imports)]
 use embassy_sync::waitqueue::AtomicWaker;
 
 use super::*;


bsodmike:
@Dirbaio I saw your "?" mark in the channel, just to clarify - when I had vendored a fork of Embassy, I saw 

```rust
warning: the item `AtomicWaker` is imported redundantly
 --> vendor/embassy/embassy-stm32/src/usart/buffered.rs:5:5
  |
5 | use embassy_sync::waitqueue::AtomicWaker;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
6 |
7 | use super::*;
  |     -------- the item `AtomicWaker` is already imported here
  |
  = note: `#[warn(unused_imports)]` on by default
```
but I also saw once CI ran that tests would import this separately without a dependency on `super::*` - so it's needed.

This is just to silence the above warning.

Dirbaio:
under which condition is it unused? afaict it's used unconditionally in `State`.

If it's redundant due to the `use super::*;` maybe we should remove that instead, replacing it with `use super::{list, of, things};`

Dirbaio:
done in #2872

bsodmike:
> under which condition is it unused? afaict it's used unconditionally in `State`.
> 
> 
> 
> If it's redundant due to the `use super::*;` maybe we should remove that instead, replacing it with `use super::{list, of, things};`

Thanks for the suggestion!!

PR #2862, opened 2024-04-23
WebUSB implementation

chmanie:
I created an implementation for the WebUSB spec that can be found here: https://wicg.github.io/webusb/.

This is a rough draft and I just wanted to get a feel for whether you'd be willing to consider this functionality as a part of embassy.
I tried to stay as close to the current class implementations as possible and use existing APIs as much as I could (there's a notable exception - `bos_capability`, see the comment below).

If you feel like this would be a valuable addition, I'll clean it up a bit more and add documentation for it.

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -417,6 +417,11 @@ impl<'a, 'd, D: Driver<'d>> InterfaceAltBuilder<'a, 'd, D> {
         self.builder.config_descriptor.write(descriptor_type, descriptor);
     }
 
+    /// Add a custom Binary Object Store (BOS) descriptor to this alternate setting.
+    pub fn bos_capability(&mut self, capability_type: u8, capability: &[u8]) {
+        self.builder.bos_descriptor.capability(capability_type, capability);
+    }
+
     fn endpoint_in(&mut self, ep_type: EndpointType, max_packet_size: u16, interval_ms: u8) -> D::EndpointIn {
         let ep = self
             .builder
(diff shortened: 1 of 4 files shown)


embassy-usb/src/builder.rs
```diff
@@ -417,6 +417,10 @@ impl<'a, 'd, D: Driver<'d>> InterfaceAltBuilder<'a, 'd, D> {
         self.builder.config_descriptor.write(descriptor_type, descriptor);
     }
 
+    pub fn bos_capability(&mut self, capability_type: u8, capability: &[u8]) {
```
chmanie, line 420: This is the only piece I had to add to the builder. I'm not sure if this is the best spot to add it but I needed a way to add BOS capabilities within the interface API.

chmanie:
Just to keep the conversation going, I unleashed this PR out of draft mode, added all the documentation and improved the API around the URL a bit. Let me know what you think :)

chemicstry:
Please someone correct me if I am wrong, but normally you would use a regular USB class (i.e. CDC-ACM, HID, DFU etc), which defines protocol and USB endpoints that it uses. You then add WebUSB descriptors just for the OS notification and redirection to your URL, no additional endpoints are used for this. When you connect to your USB device from browser, you get a raw USB API, where you can access all device endpoints belonging to all classes that your device has.

chmanie:
I did some more reading and I think you are correct. While most of the implementations I've found _use_ endpoints to initiate transfers, they are definitely not required. I'll try to move the endpoints creation to the example and put them onto another interface.

chmanie:
Ok, I removed all the additional endpoints from the capability implementation itself and moved them to the example. Only change is that they live on interface 1 now which is created in the example itself.

Dirbaio (approved):
This is very cool! thanks

PR #2865, opened 2024-04-24
Cyw43 / Pico W bluetooth, take 2

Dirbaio:
Based on #1820, rebased and cleaned up a bit.

cc issue #1516

Current status is

- Bluetooth init works
- Sending/receiving HCI packets works
- There's an example using the [trouble](https://github.com/embassy-rs/trouble) bluetooth stack that works up to advertising. There's some impedance mismatch in how to glue it with trouble's HCI traits.

Major roadblocks left are:

- Properly integrate bluetooth into the cyw43 event loop, so you can concurrently use it with WiFi. It'd probably need setting up HCI packet queues with ZerocopyChannel similar to how ethernet packets are handled, allowing the user to run Trouble in a separate "bluetooth task"
- Design a good API for constructing the cyw43 driver in either "just wifi", "just bluetooth" or "wifi+bluetooth" modes. The API should be designed so that you don't pay for what you don't use, ie if you initialize just wifi you don't need bluetooth firmware on flash, and the bluetooth-specific code from cyw43 doesn't get compiled-in. Can be done with either "mode" generics, or Cargo features. Cargo features is probably easier.
- Get it properly hooked up to Trouble, so that more than just advertising works.

I don't have the bandwidth to work on this, and probably won't for a while (months). If you want to jump in and carry this over the finish line, ping me on [Embassy's Matrix chat](https://matrix.to/#/#embassy-rs:matrix.org).


lulf:
Quick update: i've update the bluetooth_trouble example to use the latest developments from trouble and bt-hci would should simplify the integration quite a bit.

Dirbaio:
nie :rocket: 

(note: when this is finished, we probably want the example to live in the trouble repo. I think that's where most of the "API evolution" will be, so it'll be easier to keep it up to date there. The cyw43 API is just "hci read, hci write" so probably won't change much once done.

brandonros:
Would you be interested in merging this (so it doesn't go stale) now and let the "combined WiFi mode" related features sit in a backlog? 

I'm down to try this part:

> Get it properly hooked up to Trouble, so that more than just advertising works.

examples/rp/log.txt
brandonros, line 1: did you want to check this in?

Dirbaio: yes oops

cyw43/src/control.rs
```diff
@@ -127,64 +131,70 @@ impl<'a> Control<'a> {
...
 
-        Timer::after_millis(100).await;
+            Timer::after_millis(100).await;
 
-        self.state_ch.set_hardware_address(HardwareAddress::Ethernet(mac_addr));
+            self.state_ch.set_hardware_address(HardwareAddress::Ethernet(mac_addr));
+        }
+
+        if bluetooth_enabled {
+            // TODO: call runner.init_bluetooth somehow and pass it bluetooth_firmware?
```
brandonros, line 194: is this TODO i put still accurate?

cyw43/src/hci_connector.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+impl embedded_io_async::ErrorType for HciConnector {
+    type Error = HciConnectorError;
+}
+
+impl embedded_io_async::Read for HciConnector {
+    async fn read(&mut self, buf: &mut [u8]) -> Result<usize, HciConnectorError> {
+        // TODO: how to get all the way to runner.hci_read()?
```
brandonros, line 26: i feel like in the PR description you said HCI read + write is working so this TODO has to be wrong?

cyw43/src/runner.rs
```diff
@@ -223,7 +275,295 @@ where
...
+        self.bt_set_host_ready().await;
+        self.bt_toggle_intr().await;
+    }
+
+    pub(crate) async fn upload_bluetooth_firmware(&mut self, firmware: &[u8]) {
+        // read version
+        let version_length = firmware[0];
+        let _version = &firmware[1..=version_length as usize];
+        // skip version + 1 extra byte as per cybt_shared_bus_driver.c
+        let firmware = &firmware[version_length as usize + 2..];
```
soypat, line 297: Is comment dated? It seems to me this is skipping 2 bytes, not one

Erhannis:
So, what still needs done, here?  I'm hesitant to commit myself to helping, since I'm not very experienced with rust and have a mediocre track record with volunteer projects, but I'm considering it.

Dirbaio:
Advertising works, connecting works, GATT works.

Things left for the future:
- Interrupts aren't working for some reason. The driver falls back to hot polling if you enable bluetooth.
- No way to initialize bluetooth only yet.

marxin:
Apparently, this PR broke WiFi support for `rp: wifi_tcp_server` example, please see: #3237.

PR #2869, opened 2024-04-25
doc: add RMK to embassy in the wild

HaoboGu:
add [RMK](https://github.com/haobogu/rmk) to embassy in the wild 💓

RMK is a keyboard firmware project which based on embassy.


Dirbaio (approved):
thanks!

PR #2871, opened 2024-04-26
Extract Synopsys USB OTG driver from embassy-stm32

### LINKED ISSUE #2494: Extract sysnopsys part of embassy-stm32/usb_otg into it's own crate
The `usb_otg` code in https://github.com/embassy-rs/embassy/tree/main/embassy-stm32/src/usb_otg can be used for other MCUs besides the STM32, i.e. the ESP32-S2 and the ESP32-S3. (This was the result of #2488)

There are a couple decisions and design choices that need making. I also have questions as someone who is new to USB and PACs.

1. What should this new crate be called? `embassy-usb-synopsys`? Maybe throw `otg` or `driver` in there? idk.
2. STM32 and ESP32 have different register offsets for the peripheral, how should this be handled? Should there be a trait for exposing each register or should the crate have `#[cfg(..)]`s for each mcu? The latter seems like the way.
3. Is it alright to make breaking changes? I don't think there's a choice in the matter.

Some initial question after a scan.

Where is [`crate::pac::otg::Otg`](https://github.com/embassy-rs/embassy/blob/3cd0a54e0e3e7dd4b480034106b8b24144d2385b/embassy-stm32/src/usb_otg/mod.rs#L18) defined?

Any reason [this](https://github.com/embassy-rs/embassy/blob/702007f68b31718d0bea5af403c4c0a299fb6b6e/embassy-stm32/src/usb_otg/usb.rs#L12) isn't `use core::future::poll_fn`?

Why is there a difference in register access between esp32 and stm32? Isn't it all svd2rust?
STM32 is like this.
```Rust
T::regs().gintmsk().write(|w| {
    w.set_usbrst(true);
    w.set_enumdnem(true);
    w.set_usbsuspm(true);
    w.set_wuim(true);
    w.set_iepint(true);
    w.set_oepint(true);
    w.set_rxflvlm(true);
    w.set_srqim(true);
    w.set_otgint(true);
});
```
ESP32 is like this.
```Rust
T::regs().gintmsk().write(|w| {
    w.usbrstmsk().set_bit();
    w.enumdonemsk().set_bit();
    w.usbsuspmsk().set_bit();
    w.wkupintmsk().set_bit();
    w.iepintmsk().set_bit();
    w.oepintmsk().set_bit();
    w.rxflvimsk().set_bit();
    w.sessreqintmsk().set_bit();
    w.otgintmsk().set_bit();
    w
});
```

What is [`<T as RccPeripheral>::enable_and_reset();`](https://github.com/embassy-rs/embassy/blob/702007f68b31718d0bea5af403c4c0a299fb6b6e/embassy-stm32/src/usb_otg/usb.rs#L657) about? This won't be accessible from the synopsys crate so I'll have to put it behind a trait.

Not sure what to do with [this](https://github.com/embassy-rs/embassy/blob/702007f68b31718d0bea5af403c4c0a299fb6b6e/embassy-stm32/src/usb_otg/usb.rs#L659C12-L659C21), I guess that'll go behind a trait too.

[`regs::Fifo`](https://github.com/embassy-rs/embassy/blob/702007f68b31718d0bea5af403c4c0a299fb6b6e/embassy-stm32/src/usb_otg/usb.rs#L1303)?

Besides the above, I think the rest is just a matter of taking a snapshot of the OTG part of the PAC and using it in the crate and expose some of the above via some traits I guess.

bugadani:
Closes #2494


embassy-stm32/src/usb/otg.rs
```diff
@@ -498,72 +170,37 @@ impl<'d, T: Instance> embassy_usb_driver::Driver<'d> for Driver<'d, T> {
...
     fn init(&mut self) {
         super::common_init::<T>();
 
         // Enable ULPI clock if external PHY is used
-        let _ulpien = !self.phy_type.internal();
+        let phy_type = self.inner.phy_type();
+        let _ulpien = !phy_type.internal();
 
         #[cfg(any(stm32f2, stm32f4, stm32f7))]
         if T::HIGH_SPEED {
```
Dirbaio, line 206: maybe you can move all this stm32-specific stuff to `Driver::new_*`, so you don't need to make a wrapper for `Bus` at all

bugadani: I think I'd still need the wrapper because of the Drop implementation.

Dirbaio: ah, that's true :smiling_face_with_tear: 

I wonder if maybe generics is better for this in the end? I know yesterday I said i'd prefer no generics, but I was referring to "Instance" generics. Perhaps we can have a "flavor" trait instead, containing all the chip-specific stuff. embassy-stm32 and esp-hal would provide different impls. This would avoid the issues with the `Instance` generics: both stm32 instances would be `Otg<Stm32>`, not `Otg<OTG_FS>, Otg<OTG_HS>`. The generic param would stay inside the `Driver` wrapper, wouldn't get exposed to the user.

bugadani: We would end up with something like this in the driver, I believe:

```rust
/// Peripheral instance dependent USB IP configuration.
#[derive(Clone, Copy)]
pub struct OtgInstance<'d> {
    /// The USB peripheral.
    pub regs: Otg,
    /// The USB state.
    pub state: &'d State<{ MAX_EP_COUNT }>,
    /// The PHY type.
    pub phy_type: PhyType,
    /// Number of used endpoints.
    pub endpoint_count: usize,
}

pub trait Implementation {
    /// FIFO depth in words.
    const FIFO_DEPTH_WORDS: u16;

    /// Extra RX FIFO words needed by some implementations.
    const EXTRA_RX_FIFO_WORDS: u16;

    /// Whether to set up late cnak
    const QUIRK_SETUP_LATE_CNAK: bool;

    /// Calculates TRDT value based on some internal clock speed.
    fn calculate_trdt(instance: &OtgInstance, speed: vals::Dspd) -> u8;
}
```

I don't mind, I dislike lugging around hardware-specific constants in memory, but it's probably not as neat as you have in mind.

Dirbaio: yeah that's what I was hoping for :+1:

bugadani: Turns out, my predictive powers are terrible 😭 This all looked good on paper but the idea is not exactly feasible IMO.

`init` and `disable` both need the Instance to enable/disable interrupts at least, FIFO_DEPTH_WORDS is also Instance-specific.

What **could** end up in such a trait is `RX_FIFO_EXTRA_SIZE_WORDS` and `calculate_trdt` which doesn't really cry for a new trait.

embassy-stm32/src/usb/otg.rs
```diff
@@ -803,634 +255,52 @@ impl<'d, T: Instance> Bus<'d, T> {
...
-        // TODO: enable the peripheral once enable/disable semantics are cleared up in embassy-usb
+        self.inner.enable().await
     }
 
     async fn disable(&mut self) {
-        trace!("disable");
-
-        // TODO: disable the peripheral once enable/disable semantics are cleared up in embassy-usb
-        //Bus::disable(self);
+        Bus::disable(self);
```
Dirbaio, line 283: there's nothing calling `Bus::enable()` or `Bus::disable()` yet in embassy-usb. Could you put back the `Bus::disable()` in drop?

embassy-stm32/src/usb/otg.rs
Dirbaio: okay! no trait then :+1:

embassy-stm32/src/usb/otg.rs
bugadani: Oops it was never meant to be removed from Drop.

Dirbaio (approved):
thanks!

PR #2876, opened 2024-04-27
stm32 timer: fix 32bit timer off by one ARR error

ftk:
Fixes off-by-one ARR value for 32 bit timers. Now the code is consistent with [16bit timers](https://github.com/ftk/embassy/blob/e7bfd7bac90d254ea6dee408564d5e0f8a80ca19/embassy-stm32/src/timer/low_level.rs#L245)


Dirbaio (approved):
nice find! thank you

PR #2885, opened 2024-04-28
cyw43: Add function to join WPA2 network with precomputed PSK.

CirrusNeptune:
With flags = 0 in PassphraseInfo, CYW firmware skips the PBKDF2 PSK derivation. This makes it possible avoid storing unhashed passwords.

The wpa_passphrase utility may be used to generate this PSK.

### DIFF
--- a/cyw43/src/control.rs
+++ b/cyw43/src/control.rs
@@ -229,11 +229,7 @@ impl<'a> Control<'a> {
     }
 
     /// Join a protected network with the provided ssid and [`PassphraseInfo`].
-    pub async fn join_wpa2_passphrase_info(
-        &mut self,
-        ssid: &str,
-        passphrase_info: &PassphraseInfo,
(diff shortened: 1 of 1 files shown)


cyw43/src/control.rs
```diff
@@ -228,8 +228,12 @@ impl<'a> Control<'a> {
         self.wait_for_join(i).await
     }
 
-    /// Join an protected network with the provided ssid and passphrase.
-    pub async fn join_wpa2(&mut self, ssid: &str, passphrase: &str) -> Result<(), Error> {
+    /// Join a protected network with the provided ssid and [`PassphraseInfo`].
+    pub async fn join_wpa2_passphrase_info(
```
Dirbaio, line 232: PassphraseInfo is not public, so this function probably shouldn't be public.

Dirbaio (approved):
thanks!

PR #2886, opened 2024-04-29
stm32: dac: fix new_internal not setting mode as documented

adamgreig:
The `DacChannel::new` (and `new_internal`) methods set HFSEL, (set the internal mode), and enable their channel, but the `Dac` two-channel version just creates the `DacChannel` without calling `new()`, so we need to set those properties appropriately.


Dirbaio (approved):
thanks!

PR #2887, opened 2024-04-29
stm32: Add workaround for STM32 i2cv1 errata

jamesmunns:
Applies to #2372, and may be ONE of the root causes.

This partially addresses the STM32F405 (and likely all other i2cv1) errata that if the device thinks it hears a STOP at the wrong time, it'll refuse to ever send a START again, causing all i2c transactions to lock up.

Ideally we'd do this at runtime, not to force users to re-configure their devices like I do, but then we'd need to hold onto the config, detect the lockup (timeout on read/write? check the BUSY flag is set when it shouldn't?), wait for the bus to be idle, apply the workaround, reapply config (as swrst resets the peripheral config), and start again.

At least now users can clear the fault by re-initializing the I2C device within their firmware.

### DIFF
--- a/embassy-stm32/src/i2c/v1.rs
+++ b/embassy-stm32/src/i2c/v1.rs
@@ -48,6 +48,32 @@ impl<'d, T: Instance, M: PeriMode> I2c<'d, T, M> {
             //reg.set_anfoff(false);
         });
 
+        // Errata: "Start cannot be generated after a misplaced Stop"
+        //
+        // > If a master generates a misplaced Stop on the bus (bus error)
+        // > while the microcontroller I2C peripheral attempts to switch to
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
oof that errata

govert-overgaauw-ampel:
How should I use this errata work-around runtime? I am using 3 I2C peripherals (v1) with the embassy-hal shared bus. The init function is not accessible so I need to drop and call new again? And how does this work with the i2c shared bus, do i need to recreate all device instances and busses again? I am getting a gordian knot here cause I need access to the original pins again for recreating the instance. Basically this issue triggered on our product after running long enough and is very easy to reproduce by shorting the i2c bus causing it to hang endlessly.

PR #2893, opened 2024-04-30
Make SpiDevice generic over Word size and implement u16 transfer

Ragarnoy:
It's not particularly complicated, this allows to be more in line with other implementations like ExclusiveDevice which also has an u16 word size

### DIFF
--- a/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
+++ b/embassy-embedded-hal/src/shared_bus/blocking/spi.rs
@@ -28,33 +28,30 @@ use crate::shared_bus::SpiDeviceError;
 use crate::SetConfig;
 
 /// SPI device on a shared bus.
-pub struct SpiDevice<'a, M: RawMutex, BUS, CS, Word: Copy + 'static = u8> {
+pub struct SpiDevice<'a, M: RawMutex, BUS, CS> {
     bus: &'a Mutex<M, RefCell<BUS>>,
     cs: CS,
-    _word: core::marker::PhantomData<Word>,
 }
 
-impl<'a, M: RawMutex, BUS, CS, Word: Copy + 'static> SpiDevice<'a, M, BUS, CS, Word> {
+impl<'a, M: RawMutex, BUS, CS> SpiDevice<'a, M, BUS, CS, > {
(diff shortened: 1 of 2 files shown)


Dirbaio:
I think it should be possible to avoid adding the generic param to the struct itself, using it on impls only. This way you don't have to choose a word size when creating the SpiDevice, it'll just transparently implement all transfer sizes the underlying bus implements.

Dirbaio:
Also, SpiDevice should not implement the embedded-hal v0.2 SPI traits. Those are equivalent to the 1.0 SpiBus, ie they're for a bare bus without any CS pin.

Ragarnoy:
> Also, SpiDevice should not implement the embedded-hal v0.2 SPI traits. Those are equivalent to the 1.0 SpiBus, ie they're for a bare bus without any CS pin.

Yeah, I was quite confused about that but went along with it. The traits that were used (Transfer, Write) do not exist anymore, so I removed them, is this correct ?

Dirbaio:
oh, *they were already there*. my bad, I misread the diff and thought you had added them. hmmm :|

embassy-embedded-hal/src/shared_bus/blocking/spi.rs
```diff
@@ -90,62 +91,27 @@ where
...
-pub struct SpiDeviceWithConfig<'a, M: RawMutex, BUS: SetConfig, CS> {
+pub struct SpiDeviceWithConfig<'a, M: RawMutex, BUS: SetConfig, CS, Word: Copy + 'static = u8> {
     bus: &'a Mutex<M, RefCell<BUS>>,
     cs: CS,
     config: BUS::Config,
+    _word: core::marker::PhantomData<Word>,
 }
 
-impl<'a, M: RawMutex, BUS: SetConfig, CS> SpiDeviceWithConfig<'a, M, BUS, CS> {
+impl<'a, M: RawMutex, BUS: SetConfig, CS, Word: Copy + 'static> SpiDeviceWithConfig<'a, M, BUS, CS, Word> {
```
Dirbaio, line 106: can you remove `Word` here too, and leave it in the actual spi impl only?

Issue #2875, opened 2024-04-27
stm32: no fifo field in Uart::Config, and broken FIFO with new_half_duplex mode


qwerty19106:
Now FIFO is always enabled for `usart_v4`.  
https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L1407  
```
        #[cfg(usart_v4)]
        {
            trace!("USART: set_fifoen: true (usart_v4)");
            w.set_fifoen(true);
        }
```

I use `Uart::new_half_duplex` on chip `STM32G030C8` (`usart_v4`). I send ~16 bytes and then read to IDLE:
```
uart.write(&send_buf).await?;
uart.flush().await?;
let res = uart.read_until_idle().await?;
```

But due to `w.set_fifoen(true);` I get ~7 bytes which send over `uart.write`.

1. Availability is there a reason why `Uart::Config` no have `fifo: bool` field?
2. Possible `blocking_flush` should use `TXFIFO Empty` flag if FIFO is enabled.
3. Why `async flush()` use `blocking_flush`? I think we should wait `Transmission complete interrupt` or `TXFIFO empty interrupt`  
https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L1564  
```
    async fn flush(&mut self) -> Result<(), Self::Error> {
        self.blocking_flush()
    }
```  

I am ready to write PR about each point if it is clear how it should work.

Dirbaio:
> Availability is there a reason why Uart::Config no have fifo: bool field?

The reasoning was, there's very little reason to *not* want a FIFO. Having a FIFO is just better: it makes things faster, less likely to lose bytes if you have irq delays, etc. You might think if you don't enable the FIFO you don't have to worry about flushing etc, but that's not true. You still have a 1 word "FIFO" which is the DR itself, so you can still have problems if you don't flush correctly. 

> Possible blocking_flush should use TXFIFO Empty flag if FIFO is enabled.

If `flush()` is returning before the fifo is empty with the current code then yes, definitely.

> Why async flush() use blocking_flush? I think we should wait Transmission complete interrupt or TXFIFO empty interrupt

No one has implemented async flush yet, so it uses blocking flush for now, since it at least works even if not optimally. A PR fixing this would be welcome.

PR #2895, opened 2024-04-30
Fix uart::flush with FIFO at Half-Duplex mode

qwerty19106:
CC #2875
### The problem (`STM32G030C8`, `usart_v4`, FIFO always enabled)
I make Uart with Half-Duplex mode, `write` ~16 bytes, then call `blocking_flush`, then call `read_until_idle`. No other side, only pull up.
As result I get ~7 last bytes which was sent over `uart.write`.

I did a little research and found out that the TC flag works correctly. It this is confirmed by tests also.
> When the last data is written to the USART_TDR register, wait until TC = 1.
–When FIFO mode is disabled, this indicates that the transmission of the last frame
is complete.
–When FIFO mode is enabled, this indicates that both TXFIFO and shift register are
empty.
изображение

Possible it is errate, but I not found it within official errata.

### Solution
Now I clear [cr1.re](https://github.com/embassy-rs/stm32-data-generated/blob/main/stm32-metapac/src/peripherals/usart_v4.rs#L81) flag before `write`, this solves the problem.

The ideal cycle read-write in Half-Duplex master mode:
write - write - ... - flush - read
Then we can clear `cr1.re` before `write`, and set after `flush`.
But we should call `flush` before `read`, but it is impossible with current API.

I propose two variants (only for half-duplex):
1) Add `flush` to start of `read`, clear `cr1.re` before `write`, and set after `flush`.  
It requires the real `async flush`, which I can add some later.
2) Clear `cr1.re` before `write`, and set after `flush`, and set before `read`.
It is some not obvious because after flush the some bytes can be read before we call `read`. But without `flush` the receiver will be disabled until `read`.

It implement variant `1` in this PR, and wait any ideas and discussion.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -359,15 +359,31 @@ impl<'d, T: BasicInstance> UartTx<'d, T, Async> {
 
     /// Initiate an asynchronous UART write
     pub async fn write(&mut self, buffer: &[u8]) -> Result<(), Error> {
+        let r = T::regs();
+
+        // Disable Receiver for Half-Duplex mode
+        if r.cr3().read().hdsel() {
+            r.cr1().modify(|reg| reg.set_re(false));
+        }
+
         let ch = self.tx_dma.as_mut().unwrap();
-        T::regs().cr3().modify(|reg| {
+        r.cr3().modify(|reg| {
             reg.set_dmat(true);
         });
         // If we don't assign future to a variable, the data register pointer
         // is held across an await and makes the future non-Send.
-        let transfer = unsafe { ch.write(buffer, tdr(T::regs()), Default::default()) };
+        let transfer = unsafe { ch.write(buffer, tdr(r), Default::default()) };
         transfer.await;
         Ok(())
+    }
+
+    async fn flush_inner() -> Result<(), Error> {
+        Self::blocking_flush_inner()
+    }
+
+    /// Wait until transmission complete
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usart/mod.rs
```diff
@@ -502,6 +515,10 @@ impl<'d, T: BasicInstance> UartRx<'d, T, Async> {
     ) -> Result<ReadCompletionEvent, Error> {
         let r = T::regs();
 
+        if r.cr3().read().hdsel() {
+            r.cr1().modify(|reg| reg.set_re(false));
```
Dirbaio, line 519: this should be true?

Dirbaio:
I think with Option 2 you can still receive your own bytes if you do `.write(), .read()` without a flush, because .read() will reenable the receiver when still transmitting. Perhaps do option 2, but instead make `.read()` do a flush? so `.write(); .read()` Just Works even if you don't call flush?

Dirbaio:
no. call `flush` at the start of `read`. so that `write + read` works. with the changes in this PR, only `write + flush + read` works.

qwerty19106:
Done.
I think the `Uart::split` method should not be available for Half-Duplex mode. But I not know how do it in current API: it requires particular struct `HalfDuplexUart`, and it broke [BufferedUart](https://docs.embassy.dev/embassy-stm32/git/stm32f446re/usart/struct.BufferedUart.html).

Dirbaio (approved):
thank you!

PR #2897, opened 2024-05-01
Add ch32-hal to HAL list, support WCH's CH32V family

andelf:
ch32-hal is at an early stage of development.

Supported chips are listed at https://github.com/ch32-rs/ch32-hal

An stm32-data clone for all the chips is at https://github.com/ch32-rs/ch32-data


Dirbaio (approved):
thanks!

cool to see more async HALs :star_struck:

PR #2903, opened 2024-05-02
Add stm32 dsihost driver

joelsa:
This PR adds DSI support.

The functionality has been verified using a STM32F469I-DISCO evaluation board. The display driver registers (e.g. for display backlight brightness) can be set and read out via DSI.

An example will be added later.


eZioPan:
please replace the `dsi_phy: None` I added with your entry

eg.

rcc/f247.rs

https://github.com/embassy-rs/embassy/blob/45a2abc392df91ce6963ac0956f48f22bfa1489b/embassy-stm32/src/rcc/f247.rs#L303-L304

I add them at end of each `set_clocks!`, with a `// TODO` comment (or under the `// TODO` block)

JuliDi:
@eZioPan I don't quite understand your comment. So you have already added it on main and want this PR to do what exactly?
Also, I think it would be better to feature-gate it with the `dsihost` cfg instead of `stm32f4, stm32f7` for the sake of conciseness.

Further, it has been discussed on the chat that we don't want to add this additional complexity of the DSI PHY PLL to embassy-stm32, so the current "warning" comment is better imo than only "TODO". Of course, if someone has a good idea how to implement this inside RCC, that would be great.

Regarding the CI failures: Are there any issues on main atm? They seem somewhat unrelated to this PR's changes.

JuliDi:
@eZioPan I see, that makes perfect sense! Thanks for the clarification. I've missed that.

eZioPan:
@JuliDi For CI error, I roughly guess STM32F777ZI might have a slightly different DCKCFGR register name (it should be `dckcfgr1` I guess?). So you might need a feature gate to use the correct register name for that(those) chip(s)?

And CI rustfmt use nightly toolchain, so you might need `cargo +nightly fmt` to make format pass.

joelsa:
> I roughly guess STM32F777ZI might have a slightly different DCKCFGR register name (it should be `dckcfgr1` I guess?). So you might need a feature gate to use the correct register name for that(those) chip(s)?

Yes, the STM32F7xx that have two dedicated clock configuration registers, the PLLSAIDIVR resides in dckcfgr1.

JuliDi:
What a mess ;(
Glad you figured it out!

joelsa:
The U5 series does not have a dedicated clock configuration register at all, which is why I have excluded them there as well.

I don't have a better idea Maybe someone can check whether this is sensible and flexible enough?

The U5 series have a different/newer DSI/LTDC peripheral anyway, so they might need some more love. Unfortunately I don't have a devboard with DSI for them, but the STM32U5G9J-DK2 Discovery Kit looks useful.

JuliDi:
I was able to get the example from ST's BSP running based on the code in this PR. So there should not be any major flaws.

joelsa:
@Dirbaio The current state is now verified working on a 32F469IDISCOVERY kit with an NT35510 display driver and a FRD400B25025 panel, using the dsihost implementation registers can be written to and read from the driver IC and images can be shown with the LTDC:

<details>
  <summary>Image (230kB):</summary>

  IMG_8277

</details>

Dirbaio (approved):
thank you!

(it'd be awesome if you could also PR the example, btw!)

JuliDi:
@Dirbaio the issue with the above example for DSIHOST is that the code has to be very specific to that one particular devboard in that particular revision (don't get me started on the headaches this caused, they mounted a different display controller on some board revisions and that controller requires a different setup sequence...). So the question really is whether you want something that specific in the embassy repo or not.

Dirbaio:
yeah that's fine! any example is better than no example, it's okay if it's specific. If you want to warn users aSuch a requirement cannot be expressed as a Rust type, thus Rust’s subtraction panics (in debug mode) when it is called with bad inputs.bout it, you could write a comment like "this example is for board X revision Y. You'll have to adapt it to your board, make sure you change A, B, C to match your board's docs".

JuliDi:
Sounds good! @joelsa could you add such a comment and make a new PR?

joelsa:
> Sounds good! @joelsa could you add such a comment and make a new PR?

Yes, sure!

@Dirbaio the above Ferris image is stored in the flash as a static u32 array, this would add about 4.5 MB to the repo. The external FMC SDRAM is not set up, so the flash address is passed directly into the LTDC. I used Ferris because I wanted to use something available in the public domain. It would also be inside a different file from the example application logic, in order to not slow down code editors.

Are you comfortable with that? Other options would be to only show a small Ferris in the top left corner, or to set the screen to a constant color.

PR #2904, opened 2024-05-03
Add example for bdma

PegasisForever:
Tested on h723zg

### DIFF
--- a/docs/modules/ROOT/pages/faq.adoc
+++ b/docs/modules/ROOT/pages/faq.adoc
@@ -279,6 +279,14 @@ If you see errors that look like this:
 DMA: error on BDMA@1234ABCD channel 4
 ----
 
-You likely need to set up your linker script to define a special region for this area, and copy data to that region before using with BDMA.
+You need to set up your linker script to define a special region for this area, and copy data to that region before using with BDMA.
 
-TODO: show how to do that
+General steps:
+
+1. Find out which memory region BDMA has access to. You can get this information from the bus matrix and the memory mapping table in the STM32 datasheet.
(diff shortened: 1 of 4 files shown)


jamesmunns (commented):
Just noting that this example is a little unsound, wrt uninitialized linker sections.

I think we should still merge it for now, but making a note so I can open an issue to fix it later to give a better sound example.

Dirbaio (approved):
thanks!

bsodmike:
> Just noting that this example is a little unsound, wrt uninitialized linker sections.

Hi @jamesmunns I'm curious as to what you mean by "uninitlalized" - `static mut RAM_D3: [u8; 64 * 1024] = [0u8; 64 * 1024];` the RAM area here are initialised to `0u8`??

Also - more a general question, do we need to do the same with the linked area for DMA (not BDMA)?

@PegasisForever do you have any working DMA examples for I2c etc? I'm testing on the h747.

Thanks!

bsodmike:
It also looks like we don't have an I2c BDMA transfer, I'm not sure if it's really needed?

jamesmunns:
@bsodmike 

> Hi @jamesmunns I'm curious as to what you mean by "uninitlalized" - static mut RAM_D3: [u8; 64 * 1024] = [0u8; 64 * 1024]; the RAM area here are initialised to 0u8??

cortex-m-rt only initializes some sections in RAM, the rust compiler has no idea what the linker does, and if you place something outside of the normal regions (`.bss`, initialized to zeros; or `.data`, initialized from flash values), then the data is not initialized to what the compiler thinks it is. To fix this, you must either tell Rust that it is uninitialized, e.g. with MaybeUninit, or initialize it somehow before any Rust code runs (this includes pre_init, it's generally only possible in custom assembly or with a bootloader).

PR #2909, opened 2024-05-05
stm32 i2c slave

jrmoulton:
This isn't quite finished. It only has the async slave implementation and the code needs to be cleaned up. 

I could use some help debugging also. 

In the listen method I enable the address match interrupt and register the waker in the poll closure. 

The issue is that the runtime never polls the function again even when an address is matched. 

If I write code that blocks instead of yielding and send and I2C communication the peripheral will still be enabled and will match the address. 

I've tried also adding the wupen bit but that didn't make a difference

```
WUPEN:Wakeup from Stop mode enable 
0: Wakeup from Stop mode disable.
1: Wakeup from Stop mode enable.
```

Any idea why the runtime wouldn't poll the function again?


bsodmike:
Hi Jared, interesting. Will respond soon if I figure anything out.

jrmoulton:
I've figured out why the function wasn't being polled. Now just working on finishing up the implementation

jrmoulton:
@Dirbaio the stm32 i2c supports acting as both a master and a slave. 

Currently in this implementation I have split the slave from the master. This more closely matches the API in the embassy-rp crate. Would it be better to join them and merge the slave functionality into the regular i2c driver or is it better to leave them split?

bsodmike:
I think being part of the regular driver works, if say you introduce a Option<Mode> type config option that is very obvious.

Thoughts?

bsodmike:
@jrmoulton see CI issues please, just a heads up.  keen to test this soon - thanks so much!!

bsodmike:
@Dirbaio I like your CI page title "lol job"?? https://ci.embassy.dev/jobs/586c0e7a5b13

bsodmike:
> @jrmoulton see CI issues please, just a heads up. keen to test this soon - thanks so much!!

`stm32_metapac::i2c::vals::Oamsk` is there in the docs (v`15.0.0`)

```rust
test result: ok. 3 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.06s

    Updating crates.io index
    Updating git repository `https://github.com/embassy-rs/stm32-data-generated`
   Compiling embassy-hal-internal v0.1.0 (/ci/code/embassy-hal-internal)
   Compiling embassy-time v0.3.0 (/ci/code/embassy-time)
   Compiling embassy-usb-synopsys-otg v0.1.0 (/ci/code/embassy-usb-synopsys-otg)
   Compiling embassy-stm32 v0.1.0 (/ci/code/embassy-stm32)
   Compiling embassy-net-driver v0.2.0 (/ci/code/embassy-net-driver)
   Compiling embassy-embedded-hal v0.1.0 (/ci/code/embassy-embedded-hal)
error[E0432]: unresolved import `stm32_metapac::i2c::vals::Oamsk`
  --> src/i2c/mod.rs:19:5
   |
19 | use stm32_metapac::i2c::vals::Oamsk;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `Oamsk` in `i2c::vals`

warning: unused import: `interrupt`
  --> src/time_driver.rs:19:13
   |
19 | use crate::{interrupt, peripherals};
   |             ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `interrupt`
  --> src/exti.rs:14:13
   |
14 | use crate::{interrupt, pac, peripherals, Peripheral};
   |             ^^^^^^^^^

error[E0599]: no method named `init` found for struct `I2cSlave` in the current scope
   --> src/i2c/mod.rs:367:14
    |
299 | pub struct I2cSlave<'d, T: Instance, M: Mode> {
    | --------------------------------------------- method `init` not found for this struct
...
367 |         this.init(freq, config);
    |              ^^^^ method not found in `I2cSlave<'_, T, M>`

warning: unused variable: `waker`
   --> src/dma/ringbuffer.rs:429:33
    |
429 |         fn set_waker(&mut self, waker: &Waker) {}
    |                                 ^^^^^ help: if this is intentional, prefix it with an underscore: `_waker`
    |
    = note: `#[warn(unused_variables)]` on by default
```

embassy-stm32/src/i2c/mod.rs
```diff
@@ -169,6 +167,216 @@ impl<'d, T: Instance, M: Mode> I2c<'d, T, M> {
...
+        let mut this = Self {
+            _peri: peri,
+            tx_dma,
+            rx_dma,
+            #[cfg(feature = "time")]
+            timeout: config.timeout,
+            _phantom: PhantomData,
+        };
+
+        this.init(freq, config);
```
bsodmike, line 367: This is missing.

embassy-stm32/src/i2c/mod.rs
```diff
@@ -13,6 +16,7 @@ use embassy_hal_internal::{into_ref, Peripheral, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
 #[cfg(feature = "time")]
 use embassy_time::{Duration, Instant};
+use stm32_metapac::i2c::vals::Oamsk;
```
bsodmike, line 19: Odd error in CI, needs a deeper look.

jrmoulton:
I'm in the process of reworking the I2C driver to be generic over an I2C mode which can either be Master, Slave, or MultiMaster.

jkorinth:
Just wanted to let you know that I'm also eagerly awaiting slave mode, thanks a lot @jrmoulton!

embassy-stm32/src/i2c/mod.rs
jrmoulton: I haven't been able to figure out why I am getting this error in CI.

bsodmike: May be @Dirbaio can help, he's very familiar with his CI setup...

jrmoulton:
The last remaining ci issues are unrelated to this PR. Finally actually ready for review

jrmoulton:
@Dirbaio is there anything else I need to do in order for this PR to be reviewed?

Dirbaio:
yes, make CI pass. Once CI is green it'll be reviewed

okhsunrog:
any updates on this?

jrmoulton:
I need to undo a few changes that I made and then the CI should pass. It's close. I just haven't spent the last hour or two to get it finished.

okhsunrog:
do you have any example of code I can use with your PR?
I can test on STM32F103C8, STM32F401CE, STM32L071CB, STM32F405RG and maybe on STM32F334C8T6 too

jrmoulton:
I think what I actually want to do is remove my changes that allow 10 bit addressing. It should really be a separate PR. I'll go ahead and work on this for a bit @okhsunrog

okhsunrog:
> remove my changes that allow 10 bit addressing
I think that's the right thing to do, as fixing the CI actually requires fixing 16 examples for different MCU families that use i2c. Using only 7-bit addresses for now will allow this PR to be merged faster

okhsunrog:
Can you please provide an example of using i2c slave for any stm32 mcu? @jrmoulton

jrmoulton:
I'll add one

jrmoulton:
I haven't actually tested the example that I've added. I've tested the slave code but not in this particular example. I've made this example match as closely as I could to the rp2040 i2c slave example.

okhsunrog:
Thanks! I'll test the example with STM32L071CB tomorrow

okhsunrog:
@Dirbaio can this be reviewed? It seem to look ready

okhsunrog:
I tested the PR today, using STM32L071CB as slave and ESP32-C3 as master. I adapted your example, deleting master code, so only slave code it left. I'm looking at the transmission I see that time between read/write bit and ack from slave is about  280 us, so the clock gets stretched for 280 us. for some reason esp32-c3 doesn't like it. I see errors like:
```
E (44590) i2c.master: I2C transaction timeout detected
E (44590) i2c.master: s_i2c_synchronous_transaction(870): I2C transaction failed
```
(I set timeout to -1, I tried making it really big too, no result).
is there a way to decrease the gap? For now, when the slave sends ACK seems like the master is no longer listening, so the SDA stays pulled down and esp32-c3 just freezes. I guess, I should use different hardware as master or decrease the delay before ACK somehow?
image

jrmoulton:
@okhsunrog it looks like the esp32c3 doesn't support clock stretching. When you say you are increasing the timeout, do you mean on the esp32c3 master or on the stm32 slave?

https://docs.esp-rs.org/esp-hal/esp-hal/0.19.0/esp32c3/esp_hal/i2c/struct.I2C.html#method.new_with_timeout_async

jrmoulton:
That timeout is used for the slave as a maximum time to wait for the master when doing a read or write. It isn't directly related. 

The reason that the clock is being stretched for so long is that, the way it is written, the executor has to schedule the task that will respond to the request from the master. This code does not immediately start responding. 

There are reasons that I wrote it this way. The biggest is that in order to respond immediately, the slave needs to have a buffer ready which doesn't really work if the slave can match multiple addresses. Some logic based on the matched address is necessary. I tried to find a way to store a closure to handle this logic but without an allocator I couldn't find a way to do it. So instead the clock is stretched until the task is scheduled and the user decides how to respond once they have seen the matched address. 

I would probably be useful to have functions that work without clock stretching but I haven't written those

okhsunrog:
Yes, one second is very long. And what about `ERROR error while responding Timeout`? Master says that everything is okay, the signal looks fine, why the timeout then?
@jrmoulton

jrmoulton:
Not sure. There could be several reasons. Hard to say without seeing all of the code that you are using

okhsunrog:
I'll send the full code tomorrow. I just got Discovery F4 board, I'll test it both as slave and as master and post the results here too

okhsunrog:
That's the code I used for slave: https://github.com/okhsunrog/stm32l071_i2c_slave_test
That's the code I used for master: https://github.com/espressif/esp-idf/tree/release/v5.3/examples/peripherals/i2c/i2c_tools
The only difference is that in this line: https://github.com/espressif/esp-idf/blob/6568f8c553f89c01c101da4d6c735379b8221858/components/hal/esp32c3/include/hal/i2c_ll.h#L191 I replaced 1 with 0

okhsunrog:
@jrmoulton yesterday I had some free time and I investigated into this issue some more time. The stretching of SCL for 1 second seems to be because of this:
```
impl Default for Config {
    fn default() -> Self {
        Self {
            #[cfg(gpio_v2)]
            sda_pullup: false,
            #[cfg(gpio_v2)]
            scl_pullup: false,
            #[cfg(feature = "time")]
            timeout: embassy_time::Duration::from_millis(1000),
        }
    }
}
```
I tried different timeouts manually and the scl is indeed stretched for the duration I set. When I set it to `Duration::from_micros(100)` the stretching is quite low, around 200-300 us, which is okay. The second issue is that your `pub async fn respond_to_write` function always return with timeout error, no matter which timeout I set, it never returns Ok. and the SCL line stretching time depends on the timeout value in config. could you look into this once again?

okhsunrog:
@jrmoulton you're doing `dma_transfer.await;`, how does it work? when does this feature gets polled?
When I replace the code `timeout.with(self.read_dma_internal_slave(buffer, timeout)).await` with `self.read_dma_internal_slave(buffer, timeout).await` the code gets stuck forever. What could be the issue?

okhsunrog:
Did you test both blocking and async?

bsodmike:
> That's the code I used for slave: https://github.com/okhsunrog/stm32l071_i2c_slave_test
> 
> That's the code I used for master: https://github.com/espressif/esp-idf/tree/release/v5.3/examples/peripherals/i2c/i2c_tools
> 
> The only difference is that in this line: https://github.com/espressif/esp-idf/blob/6568f8c553f89c01c101da4d6c735379b8221858/components/hal/esp32c3/include/hal/i2c_ll.h#L191 I replaced 1 with 0

Can you share your memory.x as well please?

Knaifhogg:
Is there anything other than time that's needed to keep this one moving forward?

jrmoulton:
Just time. At one point I spent quite a bit of time on this getting CI to pass so that it would be reviewed but even once I got CI passing it was never reviewed. 

I would be willing to spend more time fixing it up again/ going over any reported bugs but since I'm not actively using this feature anymore I'd at least need to know that it would be reviewed. 

(not intended as complaint against the maintainers, I'm sure they're busy. That's just my situation).

Knaifhogg:
> @jrmoulton you're doing `dma_transfer.await;`, how does it work? when does this feature gets polled? When I replace the code `timeout.with(self.read_dma_internal_slave(buffer, timeout)).await` with `self.read_dma_internal_slave(buffer, timeout).await` the code gets stuck forever. What could be the issue?

I have had time to implement this for myself now to test it and I appear to get the same issue. `dma_transfer.await` hangs and the timeout wrapper is the only way to end the call. I had ADC set up to use DMA as well and that works fine.

With a timeout set to 2000 micros I can write and read from a i2c memory implementation, so something is written and read, but every call times out.

Knaifhogg:
I had a go at trying to use blocking methods instead, and noticed that they never call slave_start. So after adding it, ADDR is cleared properly. The ADDR -> ADDRCF clear seems to take about 40 - 100 microseconds which is about 3 times what my other MCU code in C seems to take. I'm gonna disable other parts of my code and see if it can be improved.

Knaifhogg:
I fixed some issues I encountered with blocking read/write:
- timeouts when transmitting: I found it would trigger easily once, then become stable, so I just ignore it for now (even at 2000 microseconds). Generally ADDR sequence takes 80 micros and each data write takes 25 micros.
- NACK lingering {from write?), blocking first received data: I noticed that the logic analyzer and the MCU disagreed on what was written to it. The issue was that a NACK would eat up the first RX byte. In the HAL of the MCU code it seemed to wait for NACK, STOP and BUSY, and clear the flags, before leaving the write function. I opted to clear flags before starting read/write, and it improved things.

I also had a bad time when my read buffer was larger than the interface I use, which causes timeouts over and over for each extra byte that is not received (so do not do this).

You can have a look at all my changes here: [commit](https://github.com/Knaifhogg/embassy/commit/766b048928628b3a5cc97e33552a3a2f2a65503b)

With this in place it seems to work for my use case.

Knaifhogg:
As an update I have been using my branch for the last few months without issue for my use case. So I am interested in finding a way forward with this PR so we can get the functionality in and fix any potential bugs later. @Dirbaio Is there a way we can know if this will be actively reviewed so this PR won't become too stale?

trembel:
I would love to see this merged as well!

aaamorgan:
I've been using this successfully on an h7s. As I believe others have noted, I also needed to be sure to provide a buffer that is the exact size as incoming data when responding to a write, which I'm not sure is intended behavior.

What's needed to move this forward? I'm happy to help support or test as needed.

lulf (approved):
I think this looks good from an API standpoint, there's lots of detail I'll just have to rely on others' testimony of using it. But I think that's as good as we can do for now, so once it's rebased, I think it's good to go.

One note though. In embassy-nrf, the listen() takes a buffer which it uses to configure the DMA before the time of reading the command, I'll assume that is just a quirk of nrf and that it's not needed on stm32.

Thanks to those who have reviewed and tested it.

mkj:
> I also needed to be sure to provide a buffer that is the exact size as incoming data when responding to a write

I found the same problem, I was experimenting with doing one byte nack transfers until a stop is received. 
https://github.com/mkj/embassy/blob/d77c1593333de7f4ed78fccc85010754e2b0db21/embassy-stm32/src/i2c/target.rs#L295
(that tree is experimenting with being able to switch between controller/target).

One gotcha to be aware of with the i2c peripheral - using DMA and non-DMA seem to be independent. So after starting with a DMA transfer, you can read from I2C_RXDR but RXNE won't get cleared! That's why I was doing one-byte DMA transfers to nack to completion.

jrmoulton:
Ok I've gone ahead and rebased on top of latest main. If there are corrections/changes/fixes they are probably best done from another PR as I'm not actively using this anymore. @lulf

okhsunrog:
I think we should merge this, yes. It's been waiting for too long :)

PR #2910, opened 2024-05-05
Make embedded-hal v0.2 optional for embassy-time and embassy-embedded-hal

onestacked:
For now it is enabled by default, but could probably be disabled in the next major version.

### DIFF
--- a/embassy-embedded-hal/Cargo.toml
+++ b/embassy-embedded-hal/Cargo.toml
@@ -24,7 +24,7 @@ features = ["std"]
 [features]
 std = []
 time = ["dep:embassy-time"]
-default = ["time"]
+default = ["time", "embedded-hal-02"]
 
 #! Support legacy for embedded hal 0.2 traits 
 embedded-hal-02 = ["dep:embedded-hal-02"]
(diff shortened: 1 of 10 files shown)


Dirbaio:
We shouldn't make e-h0.2 optional because 

- the cost of making it not-optional is quite small. e-h0.2 is tiny, compiles very fast, and adds no bloat to the binary if not used.
- there's many drivers out there still using eh0.2. Disabling it makes them not work, which is extra complexity for the users in something that's already complex. Users already struggle with eh0.2/eh1.0 version mismatches, it'd be worse if they also had to take into account Cargo features.
- We alreadyu have way too many features, each feature increases the complexity of testing (you have to ensure the crate builds with the feature enabled and disabled, multiplied by all the combinations of all other features).

onestacked:
I just find it a bit annoying to have multiple versions of the same package, but I guess it doesn't really matter so I'll close this PR.

onestacked:
Sure I can take a look at rebasing this and reopening this PR (or open a new one if that's preferred).

Should the v02 traits be enabled by default?

Dirbaio:
> I believe currently the BlockingAsync type actually adapts from embedded-hal v0.2 blocking traits to embedded-hal-async v1.0 async traits.

Yes, changing `BlockingAsync` to adapt from 1.0 blocking to 1.0 async sounds good to me. 

I don't think we should drop 0.2 impls elsewhere (or even make them optional with a Cargo feature), they don't hurt users of 1.0.

PR #2914, opened 2024-05-07
embassy_sync::Mutex: Implement traits to match std

wetheredge:
Adds trait implementations to match those in `std`:
- `Mutex<M, T>`
	- [`From<T>`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#486-493)
	- [`Default`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#495-501)
	- [`Debug`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#503-521)
- `MutexGuard`
	- [`Debug`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#556-561)
	- [`Display`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#563-568)
- `MappedMutexGuard`
	- [`Debug`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#672-677)
	- [`Display`](https://doc.rust-lang.org/1.78.0/src/std/sync/mutex.rs.html#679-684)

### DIFF
--- a/embassy-sync/src/mutex.rs
+++ b/embassy-sync/src/mutex.rs
@@ -3,9 +3,9 @@
 //! This module provides a mutex that can be used to synchronize data between asynchronous tasks.
 use core::cell::{RefCell, UnsafeCell};
 use core::future::poll_fn;
-use core::mem;
 use core::ops::{Deref, DerefMut};
 use core::task::Poll;
+use core::{fmt, mem};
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #2920, opened 2024-05-08
stm32: examples: Use unique crate name for stm32h7rs examples

plaes:
Current name clashes with "regular" stm32h7 thus generating cargo warning:
```
warning: skipping duplicate package `embassy-stm32h7-examples` found at ...
```


Dirbaio (approved):
makes sense, thanks!

Issue #2921, opened 2024-05-08
Proposition of example of QSPI usage on STM32


Mirror0:
Hi @Dirbaio 

I was asked to provide example of QSPI usage here https://github.com/embassy-rs/embassy/issues/1293.

I would be happy to merge it into the example repository here, but it uses implementation for specific flash chip to demonstrate communication. It has hard-coded chip commands as constants so it is not very generic.

You can see full code here: https://github.com/Mirror0/qspi-example

And here is my question. Is example in this format acceptable?

Dirbaio:
yes! it's fine if it's for one particular flash chip, it's an example after all. if the user has another flash chip they'll have to edit it, but at least they have a starting point.

i'd collapse it to a single .rs file and remove the need for `#![feature(type_alias_impl_trait)]` so it compiles on stable ,but otherwise it looks great! :+1:

PR #2922, opened 2024-05-09
Example usage of QSPI

Mirror0:
PR for https://github.com/embassy-rs/embassy/issues/2921

### DIFF
--- a/examples/stm32f7/src/bin/qspi.rs
+++ b/examples/stm32f7/src/bin/qspi.rs
@@ -1,0 +1,300 @@
+#![no_std]
+#![no_main]
+#![allow(dead_code)] // Allow dead code as not all commands are used in the example
+
+use defmt::info;
+use embassy_executor::Spawner;
+use embassy_stm32::qspi::enums::{AddressSize, ChipSelectHighTime, FIFOThresholdLevel, MemorySize, *};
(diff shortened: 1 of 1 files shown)


examples/stm32f7/src/bin/qspi.rs
```diff
@@ -0,0 +1,304 @@
...
+        address_size: AddressSize::_24bit,
+        prescaler: 16,
+        cs_high_time: ChipSelectHighTime::_1Cycle,
+        fifo_threshold: FIFOThresholdLevel::_16Bytes,
+    };
+    let driver = Qspi::new_bk1(
+        p.QUADSPI, p.PF8, p.PF9, p.PE2, p.PF6, p.PF10, p.PB10, p.DMA2_CH7, config,
+    );
+    static STATIC_CELL: StaticCell<Qspi<QUADSPI, DMA2_CH7>> = StaticCell::new();
+    let qspi = STATIC_CELL.init(driver);
```
Dirbaio, line 288: I think if you make the FlashMemory take an owned `Qspi` instead of `&'static mut Qspi` you can remove the StaticCell here, which would simplify things a bit.

Mirror0: Done

Dirbaio (approved):
thanks!

PR #2923, opened 2024-05-09
rp: Add raw interrupt handler example

kalkyl:
By request, this example shows how you can use raw interrupt handlers alongside embassy.
The example also showcases some of the options available for sharing resources/data within the application.


Dirbaio (approved):
thanks!

PR #2928, opened 2024-05-10
rp: implement usb stall

bobrippling:
This implements endpoint stalling for `embassy-rp` - the current use-case being that Windows will stall a USB MSD implementation, possibly due to implementation error, or possibly while it's probing the device. This allows us to continue past the stall, rather than panicking.

Tested by running this on a pico implementing a MSD - interested to hear if there are other approaches to testing USB implementations (particularly stalls / error handling).

### DIFF
--- a/embassy-rp/src/usb.rs
+++ b/embassy-rp/src/usb.rs
@@ -412,12 +412,38 @@ impl<'d, T: Instance> driver::Bus for Bus<'d, T> {
         .await
     }
 
-    fn endpoint_set_stalled(&mut self, _ep_addr: EndpointAddress, _stalled: bool) {
-        todo!();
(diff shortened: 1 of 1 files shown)


Dominaezzz:
> Windows will stall a USB MSD implementation

What do you mean by this? Hosts don't stall endpoints, devices/functions do.

Dirbaio (approved):
thanks!

bobrippling:
> > Windows will stall a USB MSD implementation
> 
> What do you mean by this? Hosts don't stall endpoints, devices/functions do.

I suppose I was seeing the unimplemented functions being called from here, so not the device stalling directly, but in response to a message from the host:

https://github.com/embassy-rs/embassy/blob/763b57449fef31b2ca5205542e2c0a52e6b786c1/embassy-usb/src/lib.rs#L601-L611

PR #2929, opened 2024-05-11
rp: Add embedded-sdmmc example

kalkyl:
We've been getting quite a few questions regarding how to use embedded-sdmmc together with embassy-rp lately (there seems to be some confusion around how to make a SpiDevice), so thought i'd make a quick basic example of how to set it up.
Also shows how to reconfigure the underlying SPI bus to increase the SPI clock frequency after sd card initialization.


Dirbaio (approved):
really cool! thank you

Issue #2798, opened 2024-04-09
Configuring pull-up for SPI MISO pin on STM-32


natrow:
For some applications, a pull-up resistor is required on the MISO pin of an SPI bus. Is it possible to use one of the STM32 internal pull-ups for this application? It seems like this would be the relevant line of code:

https://github.com/embassy-rs/embassy/blob/c575c7dc6cab49a99f36eedbdd41e97868e1102b/embassy-stm32/src/spi/mod.rs#L119

If there was an API change to allow passing a `Pull` input to the spi config struct, that would be fantastic.

Dirbaio:
pull requests welcome. ideally it'd work like the pulls in i2c, where you put them in Config. It should be easy to add.

joelsa:
I am using the MCP33131-10-E/MS, which does not use the MOSI pin and needs a pull-up on MISO in order to monitor the status of the recalibration. Up until now, I hacked my way around it, but decided to take a shot at adding this in: https://github.com/embassy-rs/embassy/pull/2943

Issue #2942, opened 2024-05-15
STM32 GPIOv1 input pins being set incorrectly (for SPI)


tobywf:
As far as I can tell, the GPIO MODE bits end up being set incorrectly for the MISO pin (input) when SPI is being set up for devices that use `gpio_v1`. It could also affect other functions that declare "alternate function" input pins.

This behaviour is also present in `embassy_stm32` version `0.1.0` on crates.io. The code is slightly different since the `new_pin!()` macro doesn't exist, but the calls end up being the same.

I don't really understands why this doesn't usually manifest in a bug; I can only assume that for most genuine STM chips if the CNF bits aren't changed, then changing the MODE bits doesn't change the pin's configuration mode.

Please bear with me, this has been hard to trace due to all the indirections. 

---

TL;DR: The issue is that pins are being configured "twice", using `set_as_af_pull()` and `set_speed()`. For input pins, the latter is incorrect.

On page 161 the reference manual [RM0008](https://www.st.com/resource/en/reference_manual/rm0008-stm32f101xx-stm32f102xx-stm32f103xx-stm32f105xx-and-stm32f107xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) Rev21 for STM32F101xx, STM32F102xx, STM32F103xx, STM32F105xx and STM32F107xx, table 20. "Port bit configuration table" indicates that for input pins, the MODE bits ought to be `00`.

![Table 20. Port bit configuration table from RM0008](https://github.com/embassy-rs/embassy/assets/2903454/9a752c45-c327-4abf-bfe2-61ae065d8cd0)

STM32F10X devices all seem to use use `gpio_v1`, e.g. `STM32F103RE` in [`stm32-metapac/src/chips/stm32f103re/pac.rs`](https://github.com/embassy-rs/stm32-data-generated/blob/bedafad200aaaf1c879505d8f6253911f767c9b2/stm32-metapac/src/chips/stm32f103re/pac.rs#L86)

```rust
#[path="../../peripherals/gpio_v1.rs"] pub mod gpio;
```

All SPI pins are set up via the `new_pin!()` macro for both blocking and async, but MISO is the interesting one as it is an input:

https://github.com/embassy-rs/embassy/blob/ea70b440cd1035f28c3f332a2f72d7fa42ac995d/embassy-stm32/src/spi/mod.rs#L412

[...]

PR #2943, opened 2024-05-15
Add miso pullup to spi configuration

joelsa:
Closes https://github.com/embassy-rs/embassy/issues/2798 by adding the MISO-pull-up as an option in the Config struct.

The only interesting bit is that in order for get_current_config to still work, I added a pull() function to the SealedPin trait:

```Rust
    /// Get the pull-up configuration.
    #[inline]
    fn pull(&self) -> Pull {
        critical_section::with(|_| {
            let r = self.block();
            let n = self._pin() as usize;
            #[cfg(gpio_v1)]
            {
                let crlh = if n < 8 { 0 } else { 1 };
                match r.cr(crlh).read().mode(n % 8) {
                    vals::Mode::INPUT => match r.cr(crlh).read().cnf_in(n % 8) {
                        vals::CnfIn::PULL => match r.odr().read().odr(n % 8) {
                            vals::Odr::LOW => Pull::Down,
                            vals::Odr::HIGH => Pull::Up,
                        },
                        _ => Pull::None,
                    },
                    _ => Pull::None,
                }
            }
            #[cfg(gpio_v2)]
            {
                match r.pupdr().read().pupdr(n % 8) {
                    vals::Pupdr::FLOATING => Pull::None,
                    vals::Pupdr::PULLDOWN => Pull::Down,
                    vals::Pupdr::PULLUP => Pull::Up,
                    vals::Pupdr::_RESERVED_3 => Pull::None,
                }
            }
        })
    }
```

I hope that I captured all the cases correctly there. Is it okay for the pull() function to reside inside the SealedPin?

Thje only missing thing (and why this is a draft currently) is that it relies on the MODE bits being set correctly for gpio v1, in order to determine whether the pin is an input or output and this is broken currently because of this bug: https://github.com/embassy-rs/embassy/issues/2942

~~Since it needs changes in the same files, I also tried to mitigate https://github.com/embassy-rs/embassy/issues/2942 in a quick way by introducing the Speed::Input so that the set_speed function does not overwrite the input. Speed::None might be a better name though, and I feel like this solution is not ideal, but it is non-invasive.~~ -> Will be moved into separate PR.


Dirbaio:
Looks good. I'd like to suggest a few additions:

- Add pull fields for mosi, miso, sck, for consistency.
- make the config fields of type `Pull`. So you can set it to None (default), Up, Down. Reason is who knows, there might be other weird SPI chips which require pull-down.
 - note `pullup: bool` makes sense for i2c because you never want pulldown for i2c, but this is not the case for spi. This justifies the inconsistency in config between spi and i2c.

Dirbaio:
> Add pull fields for mosi, miso, sck, for consistency.

uhhh sorry for the confusion, I think this doesn't make sense. MOSI and SCK are always driven by the MCU, so the pull would never affect anything, right?

joelsa:
> uhhh sorry for the confusion, I think this doesn't make sense. MOSI and SCK are always driven by the MCU, so the pull would never affect anything, right?

If we think of the MCU as the Master and don't consider the other way, then yes. For the slave-case, one can imagine a Master that want's a pull-up on MISO in order to mimic some sort of chip and I can imagine adding an internal pull to the pins to avoid floating states, but I think we should think about this after https://github.com/embassy-rs/embassy/pull/2750 is merged.

So far I've only come accross chips that need a **MISO pull-up**. The aforementioned MCP33131-10 needs constant VDIO on MOSI as well (it's read-only), but I always solved this in hardware by hardwiring it until now. I guess it might be good to not think too much about these obscure cases, as they are irrelevant for 99.9% of users.

But changing the bool to Pull even makes the code easier (reduced two match statements), so this seems senible.

Dirbaio:
> but I think we should think about this after https://github.com/embassy-rs/embassy/pull/2750 is merged.

SGTM. The SPI slave will probably have a different Config struct anyway.

embassy-stm32/src/spi/mod.rs
```diff
@@ -50,6 +50,11 @@ pub struct Config {
     pub bit_order: BitOrder,
     /// Clock frequency.
     pub frequency: Hertz,
+    /// Enable internal pullup on MISO.
+    ///
+    /// There are some ICs that require a pull-up on the MISO pin for some applications.
+    /// If you  are unsure, you probably don't need this.
+    pub miso_pullup: Pull,
```
Dirbaio, line 57: this shold be `miso_pull` now.

Dirbaio:
lgtm, is there any reason this is still draft?

joelsa:
Yes, I wanted to wait on https://github.com/embassy-rs/embassy/pull/2996 being merged :)

Dirbaio (approved):
ah cool cool

joelsa:
Wait, can you still abort the merge queue? I might have missed something.

joelsa:
This modulo is wrong there I think:

https://github.com/embassy-rs/embassy/blob/5cba97821f6f3e716452d38bb0c645b51592abc0/embassy-stm32/src/gpio.rs#L702

PR #2945, opened 2024-05-15
Update embedded-hal dependency to 1.0

felipebalbi:


felipebalbi:
Missed the fact that some (all?) crates depend on both e-h 0.2.6 and 1.0. I'll go through the changes more carefully and either close the PR or update it.

felipebalbi:
This PR is not needed.

PR #2950, opened 2024-05-16
stm32g0: add i2c_async example

felipebalbi:
This example will help those having difficulties understanding how to bind interrupts on stm32g0 devices.


Dirbaio (approved):
thanks!

PR #2958, opened 2024-05-18
Embassy book refactor 

lulf:
This reorganizes the book into a basic asciidoc structure, which can be used by `asciidoctor` to generate html, pdf, docbook etc. 

This keeps the existing functionality of including source code from examples, but removes the need to follow the antora documentation layout (simpler to navigate and edit) and allows us to keep the docs build and publish in the main embassy repo to ensure it is in sync. (no separate embassy-book repo).

Example rendered HTML with default styling:  https://lulf.no/example-embassy-book/

In addition to this PR, the publish-to-kubernetes-pod currently done by the GHA in embassy-book must be moved here.


plaes:
Looks good at first glance, though I noticed one small issue in the index in the showcase section, where indentation is off:
image

PR #2960, opened 2024-05-18
Consistent functions for Channel, PriorityChannel and PubSubChannel to return capacity and filling

sourcebox:
Until now, only `Pipe` had all functions to return the capacity and amount of filling. This PR adds the missing functions to Channel, PriorityChannel and PubSubChannel so that all of them have a consistent API.

### DIFF
--- a/embassy-sync/CHANGELOG.md
+++ b/embassy-sync/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - Add `capacity`, `free_capacity`, `len`, `is_empty` and `is_full` functions to `Channel`.
+- Add `capacity`, `free_capacity`, `len`, `is_empty` and `is_full` functions to `PriorityChannel`.
 
 ## 0.5.0 - 2023-12-04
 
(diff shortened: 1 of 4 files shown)


diondokter:
For the pubsub you might want some of these functions on the subscribers and publishers as well

sourcebox:
> For the pubsub you might want some of these functions on the subscribers and publishers as well

Ok, I'll have a look. I never used the pubsub before.

sourcebox:
> > > For the pubsub you might want some of these functions on the subscribers and publishers as well
> > 
> > 
> > Ok, I'll have a look. I never used the pubsub before.
> 
> They've got a reference to the bus, so it should be easy to call the functions on the bus

The publishers already have a `space` function, the subscribers have a `available` function, so adding `len` does only make sense if it's desired to have an alias for them.

diondokter:
The next release is gonna be breaking right? I'd say it's worth changing it to get everything be the same

sourcebox:
> The next release is gonna be breaking right? I'd say it's worth changing it to get everything be the same

Not sure. Maybe @Dirbaio can comment on that.

sourcebox:
I think adding these functions there would require to put them in the `PubSubBehaviour` trait. That would be surely a breaking change because the trait is public.

diondokter:
Ah right, well that trait should probably be sealed. User shouldn't really look at it. It's an implementation detail.

Maybe that should be a different PR then later and leave it for now.

Dirbaio (approved):

PR #2964, opened 2024-05-19
defined binaries to be built

reneherrero:
Hi,

I found that when you open an example with VS Code (not the whole embassy directory, just the sub-folder), you get this weird `error[E0463]: can't find crate for 'test'`.  After some head banging, I found an easy fix: defining the binaries in the `Cargo.toml`.

Hope this helps,

### DIFF
--- a/examples/stm32g4/Cargo.toml
+++ b/examples/stm32g4/Cargo.toml
@@ -27,3 +27,48 @@ static_cell = "2.0.0"
 
 [profile.release]
 debug = 2
+
+[[bin]]
+name = "adc"
+test = false
+bench = false
+
+[[bin]]
+name = "blinky"
+test = false
+bench = false
+
+[[bin]]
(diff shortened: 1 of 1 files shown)


Dirbaio:
> `error[E0463]: can't find crate for 'test'`

this is fixed by these these settings in `.vscode/settings.json`:

https://github.com/embassy-rs/embassy/blob/729f143269a70da426740e141b50016951000059/.vscode/settings.json#L8-L10

If you open a subfolder, the `.vscode/settings.json` at the root doesn't apply, that's why you're getting the error. I suggest you open the root folder instead, and uncomment the crate you want to work on in `rust-analyzer.linkedProjects` instead.

Dirbaio:
(i'd prefer to not explicitly list all binaries since it's a bit of work to keep it up to date)

reneherrero:
Understood

PR #2969, opened 2024-05-20
Pubsub cleanup

diondokter:
Exposes the new length functions on the pubs and subs and seals the pubsubbehavior trait (which users shouldn't (need to) touch).

I guess there's more to clean up or at least things I'd have made different now, but there's nothing that really *needs* more work.

For posterity:

- I think it's weird that `DynSubscriber<T>` and `Subscriber<T>` both deref as `Sub<Channel, T>`.
- Same for publisher
- I'd make the shared functions into a trait and let each of the higher level types implement those traits
- This would be a breaking change, but likely one that doesn't break anyone


lulf (approved):
LGTM!

PR #2974, opened 2024-05-20
stm32/i2c: remove peripheral generic param.

Dirbaio:
- **stm32/rcc: add ClockEnableBit struct.**
- **stm32/spi: remove peripheral generic param.**
- **stm32/spi: check that the RCC enable bit is disabled on drop.**
- **stm32: use funcs for info/state, const for ENABLE_BIT.**
- **stm32/i2c: remove peripheral generic param.**


Dirbaio:
bender run

PR #2976, opened 2024-05-21
Add dsihost example

joelsa:
This adds an example for the DSIHOST that allows to show an Image on the Revision C of the STM32F469NI DISCOVERY board, looking like this:

<details>
  <summary>Image (230kB):</summary>

  IMG_8277

</details>

Adding Ferris the crab as an u32 array increased the total repo size by 1 MB after git compression, alternatives laid down here: https://github.com/embassy-rs/embassy/pull/2903#issuecomment-2120608701


Dirbaio:
very cool! :D

the huge .rs file is a bit intimidating, could you convert it to .bin and use `include_bytes!`?

Dirbaio (approved):
thank you!

Dirbaio:
(squashed so the giant .rs file doesn't appear in repo history)

PR #2983, opened 2024-05-21
Added a CAN example for the stm32l4

reneherrero:
Hi,

Some notes:

1. I did my testing on an stm32l432kc device
2. I had to change the chip and target in the `.cargo/config.toml' to `thumbv7em-none-eabihf` and `STM32L432KCUx` (not checked in as doing so breaks other examples)
3. I had to change the embassy-stm32 feature from `stm32l4s5qi` to `stm32l432kc` in the `Cargo.toml` (same, not checked in as doing so breaks other examples)
4. I don't have a `stm32l4s5qi` device so there's no way for me to ensure this will work, but confidence is reasonably high

Hope this helps,

### DIFF
--- a/examples/stm32l4/src/bin/can.rs
+++ b/examples/stm32l4/src/bin/can.rs
@@ -1,0 +1,68 @@
+#![no_std]
+#![no_main]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_stm32::can::filter::Mask32;
+use embassy_stm32::can::{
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

reneherrero:
Cool.

One more change that *might* be warranted is changing the target form `thumbv7em-none-eabi` to `thumbv7em-none-eabihf`.  While testing with the probe-rs VC Code plugin, I noticed it automatically used the `thumbv7em-none-eabihf` architecture...

PR #2985, opened 2024-05-21
stm32/adc: add AnyAdcChannel type

Univa:
The main motivation for this was to allow for the use case of storing ADC channels in an array, which is difficult to do with the current API.

Summary of changes:
- Adds an `AdcChannel<T>` trait which contains a `fn setup` and `fn channel`. `fn channel` and `fn set_as_analog` is removed from `AdcPin`, and instead uses a supertrait. The logic for `set_as_analog` is moved to `fn setup`. This allows for a single `read` function for the ADC instead of having a separate `read_internal` function (in `g4.rs` and `v4.rs`)
- Places where an "internal channel" (like `Vref` or `Temperature`) implements `AdcPin` have been changed to implement `AdcChannel` instead.
- Adds the `AnyAdcChannel<T>` type, where `T` is an ADC instance. It implements `AdcChannel<T>`, which can be passed to `read`.

Even though `AdcPin` and `InternalChannel` are now empty traits with these changes, I decided to keep them, mainly to let users create abstractions that expect a specific type of ADC channel.

### DIFF
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -4,7 +4,7 @@ use core::task::Poll;
 
 use embassy_hal_internal::into_ref;
 
-use super::{blocking_delay_us, InternalChannel};
+use super::blocking_delay_us;
 use crate::adc::{Adc, AdcChannel, Instance, SampleTime};
 use crate::time::Hertz;
 use crate::{interrupt, Peripheral};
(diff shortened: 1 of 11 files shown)


Dirbaio:
Thanks for the PR. This is definitely a great addition.

> Even though AdcPin and InternalChannel are now empty traits with these changes, I decided to keep them, mainly to let users create abstractions that expect a specific type of ADC channel.

Can you remove them? the use cases for that are a bit limited, and ADC API is complex enough already, so i'd prefer to favor simplicity. Thanks!

Univa:
> Can you remove them? the use cases for that are a bit limited, and ADC API is complex enough already, so i'd prefer to favor simplicity. Thanks!

@Dirbaio Done!

Dirbaio (approved):
thank you! :rocket:

PR #2988, opened 2024-05-22
embassy_stm32: implement optional FIFO scheduling for outgoing frames

de-vri-es:
This PR allows you to enable FIFO scheduling of the transmit mailboxes. This can be quite important to ensure in-order delivery of data.

It also allows to avoid a weird gotcha between `flush_any()` and `try_write()`: when `flush_any()` completes, a call to `try_write()` may still report `TryWriteError::Full`. In one application I'm working on, this is very undesirable. I also updated the documentation, which previously said this could not happen.

The PR also relaxes the restriction for queuing frames in priority based scheduling mode. Previously, you could not queue a frame if another frame with equal **or higher** priority was already queued. This was more strict than necessary to work around the hardware scheduling limitation. The only thing that should be avoided is to have two **equal** priority frames in the mailboxes at the same time.

### DIFF
--- a/embassy-stm32/src/can/bxcan/mod.rs
+++ b/embassy-stm32/src/can/bxcan/mod.rs
@@ -133,7 +133,7 @@ impl<T: Instance> CanConfig<'_, T> {
         self
     }
 
-    /// Enables or disables automatic retransmission of messages.
+    /// Enables or disables automatic retransmission of frames.
     ///
     /// If this is enabled, the CAN peripheral will automatically try to retransmit each frame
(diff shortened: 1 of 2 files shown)


de-vri-es:
Note that I tested this on an STM32L431RC. It works as expected: frames arrive in the order they are queued, regardless of the mailbox they end up in or the priority of the frame.

Dirbaio:
paging the CAN expert: @cschuhen does this look good to you?

cschuhen:
@Dirbaio, Yes, it looks fine. I've not tested it though.

de-vri-es:
Thanks for the review!

Issue #2752, opened 2024-03-30
stm32f2 `RingBufferedUartRx` returns 1 byte and overruns


MabezDev:
I noticed a regression in my application when upgrading from crates.io to current main, it seems the `start()` detection in `read()` has stopped working:

```rust
// Start background receive if it was not already started
if !r.cr3().read().dmar() {
    self.start()?;
}
```

When trying to read I consistently get overruns (because the DMA probably isn't running). If I call `start()` myself it works, until an error occurs and the DMA again doesn't restart.

I tried to bisect but the changes across many APIs made this difficult, but I'm hoping with the specific issue found someone with more domain knowledge can help (I only passively intake what goes on development-wise here).

Will try and add a min repro soon.

MabezDev:
Here is the min repro: https://github.com/embassy-rs/embassy/compare/main...MabezDev:embassy:mvp?expand=1

It seems that actually calling `start()` doesn't help either. Here is the log from running the example:

```rust
0.000122 INFO  Hello World!
└─ uart_dma::____embassy_main_task::{async_fn#0} @ src/bin/uart_dma.rs:19  
0.000946 TRACE USART: presc=1, div=0x0000008b (mantissa = 8, fraction = 11)
└─ embassy_stm32::usart::configure @ /home/mabez/development/rust/embedded/util/embassy/embassy-stm32/src/fmt.rs:117 
0.002441 TRACE Using 16 bit oversampling, desired baudrate: 115200, actual baudrate: 115107
└─ embassy_stm32::usart::configure @ /home/mabez/development/rust/embedded/util/embassy/embassy-stm32/src/fmt.rs:117 
5.294006 INFO  Read 1bytes
└─ uart_dma::____embassy_main_task::{async_fn#0} @ src/bin/uart_dma.rs:51  
5.295166 ERROR panicked at src/bin/uart_dma.rs:50:47:
called `Result::unwrap()` on an `Err` value: Overrun
└─ panic_probe::print_defmt::print @ /home/mabez/.cargo/registry/src/index.crates.io-6f17d22bba15001f/panic-probe-0.3.1/src/lib.rs:104 
Frame 0: "HardFault handler. Cause: Escalated UsageFault (Undefined instruction)." @ 0x08016f4c
Frame 1: __udf @ 0x080072c4 inline
       /home/mabez/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cortex-m-0.7.7/src/../asm/inline.rs:181:5
[...]

PR #2989, opened 2024-05-22
stm32 ringbuffered usart: don't teardown during reconfigure

MabezDev:
Closes https://github.com/embassy-rs/embassy/issues/2752

Unsatisfyingly, I don't know why this works or what has changed since the crates.io release. I have been poking at this for a while and I simply couldn't figure it out; I tried clearing sticky errors in all sorts of places but nothing worked. I would be interested if someone knows though.

It might also be valuable to test the min repro linked in the issue on some other chips as it might be a f2 specific thing.


Dirbaio (approved):
¯\_(ツ)_/¯

PR #2990, opened 2024-05-23
Explicitly license the book prose CC-BY-SA

jamesmunns:
This matches the Rust Embedded books


lulf (approved):
Makes sense I think, especially since some content might be shared between these books.

Dirbaio (approved):

PR #2992, opened 2024-05-23
Add spi v3 & Full-duplex support to the I2S driver.

liarokapisv:
I added PoC support to the I2S driver for the `spi_v3`. This is not fully tested yet, it works for interfacing with the `tlv320aic32x4` codec, and it seems to be working fine from my limited testing. I took care to not change much from the configuration other than the necessary changes to make it work with the v3 versions. I also needed to add Full-Duplex support, so I changed the API, I expect that this will be iterated upon, so I see it mostly as a chance to talk about the interface. Not sure about the typestate pattern here, we could also do panicking read/write, but this seems safer.

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -72,7 +72,7 @@ rand_core = "0.6.3"
 sdio-host = "0.5.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "15" }
-stm32-metapac = { git = "https://github.com/embassy-rs/stm32-data-generated", tag = "stm32-data-34c0188a682b32c32ff147d377e0629b1ebe8318" }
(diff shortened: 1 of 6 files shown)


liarokapisv:
This breaks Rx-only mode because it internally uses the Tx dma stream in master mode.

liarokapisv:
I reverted the type-state design to make the review of the `spi_v3` changes easier to audit. I removed the `Function` parameter from the `Config` since it seems error-prone to expose, especially since we are going to use the `Transmitter` mode to implement the Rx-only mode. Better to express this through the constructors. I cannot currently test on the `spi_v1`, the configuration works for the `spi_v3` if not for the `SPI` driver issues regarding the DMA. I will tackle that in another commit.
I also think we should completely change the `i2s` API so that it uses the `DMA` circular mode similar to the `RingBufferedUsart` variant. Probably worth discussing on a separate issue as well.

Dirbaio (approved):
thank you!

PR #2996, opened 2024-05-24
gpiov1: Do not call set_speed for AFType::Input

joelsa:
Closes https://github.com/embassy-rs/embassy/issues/2942, where this was discussed in greater detail, along with alternatives.

Required for https://github.com/embassy-rs/embassy/pull/2943, so the correct config can be deduced from the registers.


tobywf:
Awesome, thank you!

PR #3002, opened 2024-05-24
stm32/rcc: replace generated enable/disable code with runtime info

honzasp:
#2835 introduced a runtime struct `ClockEnableBit` that allows us to disable a peripheral in RCC using dynamic dispatch, without the need to generate separate code for each peripheral (i.e., without needing code generic in `T: RccPeripheral`).

This PR extends this mechanism (and renames it to `RccInfo`) and completely replaces the polymorphic enable/disable code with a single monomorphic code path parameterized with a 5-byte struct. The code should now also correctly handle refcounts.

This will allow us to remove generics from peripherals that use RCC refcounting (such as ADC or OPAMP), and it also slightly reduces code size for projects that use many peripherals.

### DIFF
--- a/embassy-stm32/src/rcc/mod.rs
+++ b/embassy-stm32/src/rcc/mod.rs
@@ -67,22 +67,184 @@ pub(crate) unsafe fn get_freqs() -> &'static Clocks {
 }
 
 pub(crate) trait SealedRccPeripheral {
-    const ENABLE_BIT: ClockEnableBit;
-
     fn frequency() -> Hertz;
-    fn enable_and_reset_with_cs(cs: CriticalSection);
-    fn disable_with_cs(cs: CriticalSection);
-
-    fn enable_and_reset() {
-        critical_section::with(|cs| Self::enable_and_reset_with_cs(cs))
-    }
-    fn disable() {
-        critical_section::with(|cs| Self::disable_with_cs(cs))
-    }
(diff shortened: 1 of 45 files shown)


embassy-stm32/src/rcc/mod.rs
```diff
@@ -67,23 +67,184 @@ pub(crate) unsafe fn get_freqs() -> &'static Clocks {
...
+            enable_offset,
+            enable_bit,
+            refcount_idx_or_0xff,
+            #[cfg(feature = "low-power")]
+            stop_mode,
+        }
+    }
+
+    // TODO: should this be `unsafe`?
+    pub(crate) fn enable_and_reset_with_cs(&self, _cs: CriticalSection) {
```
honzasp, line 137: It seems that all the enable/disable functions should be `unsafe`, but they were not marked so in the previous version of the code, so I kept them safe to avoid changing too many things at the same time.

Dirbaio (approved):
lgtm, let's do this! thank you :)

(fwiw I have the feeling the longterm solution is going to be removing the refcounts completely and making drivers for ADC, DAC etc handle it themselves. it might be a bit more efficient because only those drivers bear the cost, vs making RccInfo handle refcounting imposes the cost on all drivers. 

It's good to do this change now though, since it unblocks removing generics everywhere now. We can always change refcounting later)

PR #3004, opened 2024-05-25
FAQ: One vs Many tasks

jamesmunns:
Copy and paste from chat


docs/pages/faq.adoc
```diff
@@ -333,3 +333,27 @@ This is not the only possible approach, but if you are looking for where to star
...
+----
+
+There are two main ways to handle concurrency in Embassy:
+
+1. Spawn multiple tasks, e.g. with `#[embassy_executor::task]`
+2. Manage multiple futures inside ONE task using `join()` or `select()` (as shown above)
+
+In general, either of these approaches will work. The main differences of these approaches are:
+
+When using **separate tasks**, each task needs it's own RAM allocation, so there's a little overhead for each task, so one task that does three things will likely be a little bit smaller than three tasks that do one thing (not a lot, probably a couple dozen bytes). In contrast, with **multiple futures in one task**, you don't need multiple task allocations, and it will generally be easier to share data, or use borrowed resources, inside of a single task.
```
jamesmunns, line 355:
```suggestion
When using **separate tasks**, each task needs its own RAM allocation, so there's a little overhead for each task, so one task that does three things will likely be a little bit smaller than three tasks that do one thing (not a lot, probably a couple dozen bytes). In contrast, with **multiple futures in one task**, you don't need multiple task allocations, and it will generally be easier to share data, or use borrowed resources, inside of a single task.
```

lulf (approved):

PR #3005, opened 2024-05-25
Emit cargo:rustc-check-cfg instructions from build.rs

honzasp:
From 1.80 onwards, Rust will [automatically check cfgs at compile time](https://blog.rust-lang.org/2024/05/06/check-cfg.html). This will particularly affect embassy-stm32, because it uses a lot of configs for conditional compilation based on the selected chip.

This PR adds `cargo:rustc-check-cfg` instructions into build.rs, which attempt to cover all possible `cargo:rustc-cfg` instructions that build.rs could produce. The approach that I used is rather simplistic and it requires to hard-code a list of all peripheral versions and list of all supported chips into the build script. I'd be happy for any suggestions how to do this better; perhaps we can read this information from stm32-data, or perhaps it's time to split the build.rs into multiple files?

I'm also not sure whether the list of peripherals and their versions is correct: I simply took the list of [files describing registers in stm32-data](https://github.com/embassy-rs/stm32-data/tree/main/data/registers) and split them at the first underscore :) I'm not 100% sure that's correct, though.

Even with the `cargo:rustc-check-cfg` instructions, rustc still reports some warnings about unknown cfgs. Some of these look like genuine issues (such as embassy-sync referring to `feature = "wasm"`, which is not defined), but for others I'm not sure (such as embassy-stm32 referring to cfgs such as `time`, `rcc_f400` or `flash_g4`).

<details>

<summary>Output of `cargo check` in embassy-stm32</summary>

```
warning: unexpected `cfg` condition value: `wasm`
 --> /home/jan/prog/embassy/embassy-sync/src/lib.rs:1:38
  |
1 | #![cfg_attr(not(any(feature = "std", feature = "wasm")), no_std)]
  |                                      ^^^^^^^^^^^^^^^^
  |
  = note: expected values for `feature` are: `defmt`, `log`, `std`, `turbowakers`
  = help: consider adding `wasm` as a feature in `Cargo.toml`
  = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default

warning: `embassy-sync` (lib) generated 1 warning
warning: unexpected `cfg` condition name: `rcc_f400`
  --> src/rcc/bd.rs:36:59
   |
36 | #[cfg(not(any(rcc_f1, rcc_f1cl, rcc_f100, rcc_f2, rcc_f4, rcc_f400, rcc_f410, rcc_l1)))]
   |                                                           ^^^^^^^^ help: there is a config with a similar name: `rcc_f100`
   |
   = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(rcc_f400)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration
   = note: `#[warn(unexpected_cfgs)]` on by default

warning: unexpected `cfg` condition name: `rcc_f400`
   --> src/rcc/bd.rs:189:67
    |
189 |         #[cfg(not(any(rcc_f1, rcc_f1cl, rcc_f100, rcc_f2, rcc_f4, rcc_f400, rcc_f410, rcc_l1)))]
    |                                                                   ^^^^^^^^ help: there is a config with a similar name: `rcc_f100`
    |
    = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(rcc_f400)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `rcc_f400`
   --> src/rcc/bd.rs:227:75
    |
227 |                 #[cfg(not(any(rcc_f1, rcc_f1cl, rcc_f100, rcc_f2, rcc_f4, rcc_f400, rcc_f410, rcc_l1)))]
    |                                                                           ^^^^^^^^ help: there is a config with a similar name: `rcc_f100`
    |
    = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(rcc_f400)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `time`
  --> src/adc/mod.rs:74:11
   |
74 |     #[cfg(time)]
   |           ^^^^ help: there is a config with a similar name: `timer`
   |
   = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(time)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `time`
  --> src/adc/mod.rs:76:15
   |
76 |     #[cfg(not(time))]
   |               ^^^^ help: there is a config with a similar name: `timer`
   |
   = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(time)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `flash_g4`
  --> src/flash/mod.rs:99:26
   |
99 | #[cfg_attr(any(flash_g0, flash_g4), path = "g.rs")]
   |                          ^^^^^^^^ help: there is a config with a similar name: `flash_g0`
   |
   = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(flash_g4)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `flash_g4`
   --> src/flash/mod.rs:108:9
    |
108 |         flash_g4, flash_h7, flash_h7ab, flash_u5, flash_h50, flash_u0
    |         ^^^^^^^^ help: there is a config with a similar name: `flash_g0`
    |
    = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(flash_g4)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: field `second` is never read
  --> src/rtc/datetime.rs:12:9
   |
10 | pub struct RtcInstant {
   |            ---------- field in this struct
11 |     /// 0..59
12 |     pub second: u8,
   |         ^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `embassy-stm32` (lib) generated 8 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
```

</details>


honzasp:
For rustc older than 1.80, the `cargo:rustc-check-cfg` instruction triggered a warning, so I added rustc version checking using the `rustc_version` crate (which parses the output of `rustc --version`) to the build scripts. This is not very elegant, but afaik it's the only solution :/

Dirbaio:
Thanks for the PR! This is definitely something we'll have to do. Just a few concerns:

- Maintaining the list of chips and versions manually is going to be annoying, they change somewhat often. IMO the best solution would be `stm32-metapac-gen` to generate `static ALL_CHIPS` and `static ALL_VERSIONS` in [`stm32_metapac::metadata`](https://docs.embassy.dev/stm32-metapac/git/stm32h755zi-cm7/metadata/index.html) and read that from build.rs.
- I'd prefer not to add a new dep `rust-version` just for this. You can detect nightly from build.rs [quite easily](https://github.com/Dirbaio/embedded-hal/blob/c9fbac0d80a9aa69ceb4c927fb498682b6120cbc/embedded-hal-async/build.rs) with no extra deps, I propose using that to enable check-cfg for all nightly now (dropping support for older nightlies), then when 1.80 hits stable enable check-cfg unconditionally (dropping support for older stables).

Dirbaio:
i'd do the simplest thing which is just checking for nightly. If you really want to do more complete version parsing, go for it, up to you, but I don't think it's worth the effort because:

in the past we've dropped support for older rusts quite quickly, supporting only latest stable and recent-ish nightly (~1 month), and nobody has complained. We can just do the same here.

> for people who upgrade rustc to 1.80 but don't upgrade embassy

that's still going to happen if they stick to the released embassy-stm32 v0.1, right? I don't think we can do anything about that.

> or a wall of warnings about check-cfg being unstable in rustc < 1.79 (for people who upgrade embassy but don't immediately upgrade rustc

when 1.80 is released, we increase MSRV to 1.80 and emit check-cfg unconditionally. rust <1.79 becomes unsupported.

honzasp:
Ok, that's reasonable, I'll go with just checking for nightly, then. I'm currently blocked with other work, but should be able to make progress by the end of the week.

honzasp:
I made the two changes and marked the PR as draft until the required changes to stm32-data (https://github.com/embassy-rs/stm32-data/pull/483) are merged.

honzasp:
> there's still a few uses of `rustc_version`, could you remove those?

Yeah, I also realized that there are other uses of cfg-s generated from build.rs that I didn't catch before.

honzasp:
The PR should be ready now, but I see that the "build-nightly" CI job produces `warning: cargo:rustc-check-cfg requires -Zcheck-cfg=output flag`; I don't get these errors locally when running `+nightly-2024-05-20`, is it possible that the CI uses a nightly version that's different from what's specified in `rust-toolchain-nightly.toml`? :thinking:

Dirbaio:
i've updated cargo-batch in CI, let's see if that goes away now

bender run

Dirbaio:
it seems the warning did go away. there's new errors though, it seems it wasn't checking cfg's at all before.

honzasp:
> it seems the warning did go away. there's new errors though, it seems it wasn't checking cfg's at all before.

Yes, I found that with the cfg checking enabled, I got some warnings; however, I didn't attempt to fix those to keep the PR size contained, and also because it touches a lot of code that I don't understand. However, with the CI set to fail on warnings, I will try to fix them.

embassy-executor/build.rs
```diff
@@ -3,6 +3,9 @@ use std::fmt::Write;
 use std::path::PathBuf;
 use std::{env, fs};
 
+#[path = "../build_common.rs"]
```
Dirbaio, line 6: unfortunately this breaks `cargo publish`. all files from a crate have to be present under the crate root: 

```
[dirbaio@mars embassy-executor]$ cargo publish --dry-run
    Updating crates.io index
   Packaging embassy-executor v0.5.0 (/home/dirbaio/embassy/embassy/embassy-executor)
   Verifying embassy-executor v0.5.0 (/home/dirbaio/embassy/embassy/embassy-executor)
    Updating crates.io index
   Compiling proc-macro2 v1.0.84
   Compiling unicode-ident v1.0.12
   Compiling ident_case v1.0.1
   Compiling fnv v1.0.7
   Compiling strsim v0.11.1
   Compiling embassy-executor v0.5.0 (/home/dirbaio/embassy/embassy/embassy-executor/target/package/embassy-executor-0.5.0)
   Compiling litrs v0.4.1
   Compiling critical-section v1.1.2
error: couldn't read ../build_common.rs: No such file or directory (os error 2)
 --> build.rs:7:1
  |
7 | mod common;
  | ^^^^^^^^^^^

error: could not compile `embassy-executor` (build script) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: failed to verify package tarball
```

unfortunately I can't see a good solution... i'd just copypaste the stuff. We could make a shared crate but then we have to publish it, version it, etc. I'm not sure if it's worth it.

honzasp: Oh, good catch! What about adding a symlink `embassy-executor/build_common.rs` -> `../build_common.rs` and refering to `./build_common.rs` (the symlink) from `build.rs`? That seems to work with `cargo publish` and doesn't require copy-pasting the same file to multiple directories.

honzasp: I pushed the symlink solution, but I will change it to copy-paste if there is some issue with symlinks that I haven't foreseen.

Dirbaio: do symlinks work if you clone the repo on Windows?

honzasp: Ok, it's copy-pasta, then :)

Dirbaio (approved):
thank you! :rocket:

honzasp:
I expected this to be a quick PR to fix some warnings, but it turned out to be quite involved! But I definitely learned a lot about embassy and stm32-data in the process :) Thank you @Dirbaio for your help and guidance!

PR #3007, opened 2024-05-27
Add proper rxonly support for spi_v3 and force tx dma stream requirem…

liarokapisv:
This commit uses the Receive-only mode of the `spi_v3` to enable receiving without using a transmission DMA channel.
In addition, it forces the user to provide a TX DMA channel for older versions that require transmitting in order to enable the
master clock. It should be straightforward to port to the `spi_v2+` versions as well, I just don't currently have the hardware to properly test the changes, I have only tested the changes on the `spi_v3`.

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -643,36 +643,40 @@ impl<'d> Spi<'d, Async> {
             return Ok(());
         }
 
-        self.info.regs.cr1().modify(|w| {
+        let regs = self.info.regs;
+
+        regs.cr1().modify(|w| {
             w.set_spe(false);
         });
 
-        let comm = self.info.regs.cfg2().modify(|w| {
+        let comm = regs.cfg2().modify(|w| {
             let prev = w.comm();
             w.set_comm(vals::Comm::RECEIVER);
             prev
         });
 
-        let i2scfg = self.info.regs.i2scfgr().modify(|w| {
-            let prev = w.i2scfg();
-            w.set_i2scfg(match prev {
-                vals::I2scfg::SLAVERX | vals::I2scfg::SLAVEFULLDUPLEX => vals::I2scfg::SLAVERX,
-                vals::I2scfg::MASTERRX | vals::I2scfg::MASTERFULLDUPLEX => vals::I2scfg::MASTERRX,
-                _ => panic!("unsupported configuration"),
-            });
-            prev
+        let i2scfg = regs.i2scfgr().modify(|w| {
+            w.i2smod().then(|| {
+                let prev = w.i2scfg();
+                w.set_i2scfg(match prev {
+                    vals::I2scfg::SLAVERX | vals::I2scfg::SLAVEFULLDUPLEX => vals::I2scfg::SLAVERX,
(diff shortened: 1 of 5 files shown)


Dirbaio:
please do enable tsize for all spi versions that have it (v3+). They're all almost the same so it's very likely to work, and also we have hardware-in-the-loop (HIL) tests in CI to verify it.

I've added the `trusted` label which means CI will run HIL tests when you push. Can you extend [this test](https://github.com/embassy-rs/embassy/blob/main/tests/stm32/src/bin/spi.rs) to create a rxonly SPI, and try to do some reads? To check it works, configure MOSI as a GPIO, check if you set it high it reads all 0xFF, if low 0x00.


bender run

embassy-stm32/src/spi/mod.rs
```diff
@@ -623,11 +627,114 @@ impl<'d> Spi<'d, Async> {
 
     /// SPI read, using DMA.
     pub async fn read<W: Word>(&mut self, data: &mut [W]) -> Result<(), Error> {
+        #[cfg(not(spi_v3))]
```
Dirbaio, line 630: it's simpler to cfg the entire `read` function instead of having it call two functions that need cfg'ing anyway

liarokapisv:
It seems a bit tricky to generalize, but I will get to it.

Dirbaio:
bender run

embassy-stm32/src/spi/mod.rs
```diff
@@ -622,12 +626,113 @@ impl<'d> Spi<'d, Async> {
...
+        });
+
+        let tsize = regs.cr2().read().tsize();
+
+        let rx_src = regs.rx_ptr();
+
+        let mut read = 0;
+        let mut remaining = data.len();
+
+        loop {
```
Dirbaio, line 667: you can chunk a slice much easier with `for chunk in read.chunks(65535) {...}`

embassy-stm32/src/spi/mod.rs
```diff
@@ -622,12 +626,113 @@ impl<'d> Spi<'d, Async> {
...
+                w.set_i2scfg(match prev {
+                    vals::I2scfg::SLAVERX | vals::I2scfg::SLAVEFULLDUPLEX => vals::I2scfg::SLAVERX,
+                    vals::I2scfg::MASTERRX | vals::I2scfg::MASTERFULLDUPLEX => vals::I2scfg::MASTERRX,
+                    _ => panic!("unsupported configuration"),
+                });
+                prev
+            })
+        });
+
+        let tsize = regs.cr2().read().tsize();
```
Dirbaio, line 660: why read the old value and restore it? I think this should be always zero since nothing else uses tsize, so you can instead set it to zero at the end?

liarokapisv: I was just naively future-proofing, setting to 0 will work.

embassy-stm32/src/spi/mod.rs
liarokapisv: Thanks, much better this way.

liarokapisv:
I also need to implement the tests, I will get to it asap.

liarokapisv:
@Dirbaio does this look okay?

Dirbaio (approved):
thanks!

PR #3008, opened 2024-05-27
stm32: ensure the core runs on HSI clock while setting up rcc

aurelj:
This is the follow-up to PR #2829 applying the same modification to other stm32 series.

So just as commit 5ecc9b8, this runs the CPU on HSI clock during the whole RCC setup, and disable the HSI clock only after the RCC is fully setup (and only if the HSI clock is not in use).


Dirbaio:
bender run

Dirbaio (approved):
thank you!

PR #3017, opened 2024-05-29
add set_as_analog to stm32 gpio Flex

lucasgranberg:
Simple addition to Flex to allow analog mode to be set. I need this bc the way I work with a comparator. Ideally this would be handled by a comparator driver but I didn't manage to write that.

A case could be made for exposing `set_as_disconnected` too which would do the same thing right now but be semantically different.


Dirbaio (approved):
thanks!

PR #3020, opened 2024-05-30
Remove generics for BXCAN.

cschuhen:
Remove generics from most of the BXCAN structs. Not removed from the initial struct that allows configuration at this stage but removed from the 'split' versions of the struct.

### DIFF
--- a/embassy-stm32/src/can/bxcan/mod.rs
+++ b/embassy-stm32/src/can/bxcan/mod.rs
@@ -5,6 +5,7 @@ use core::future::poll_fn;
 use core::marker::PhantomData;
 use core::task::Poll;
 
+use embassy_hal_internal::interrupt::InterruptExt;
 use embassy_hal_internal::{into_ref, PeripheralRef};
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/can/bxcan/mod.rs
```diff
@@ -154,7 +155,10 @@ impl<T: Instance> Drop for CanConfig<'_, T> {
 
 /// CAN driver
 pub struct Can<'d, T: Instance> {
-    peri: PeripheralRef<'d, T>,
+    _peri: PeripheralRef<'d, T>,
+    instance: &'d crate::pac::can::Can,
```
Dirbaio, line 159: `crate::pac::can::Can` is already a pointer. don't store a pointer to it, store it owned directly. see #3018

cschuhen: Ok, thanks.

Dirbaio (approved):

Issue #3003, opened 2024-05-25
Revise the I2S driver interface.


liarokapisv:
Currently, the `I2S` driver is just a small wrapper over the `SPI`  driver, just enabling the `I2S`-specific configuration and then delegating to the `SPI` read/write functionality. 

I don't think this works good enough for the usual use-cases of the `I2S` driver, which is audio processing/streaming. In this domain, losing any sample results in catastrophic audio glitches. Usual sampling rates are 44.1/48 khz, but it's not unusual to also use 96 khz.
The issue with the `read/write` interface is that it forces the user to re-start the DMA transfer, this means that we need to copy the DMA buffer and restart the DMA in under 20/10us in order for audio samples to not be lost and even that may not be possible due to the peripheral & DMA latencies.

The usual approach is to use DMA in circular mode and process on each half, this adds some configurable latency but allows for processing without losing any audio samples. This is not currently possible with the I2S driver. I think it would be better for the driver to only expose a circular ring-buffered interface similar to the `RingBufferedUsart` one. This avoids the footgun of using the `read/write` methods and leads the user to a more robust approach. Any thoughts are welcome!

Dirbaio:
done in  #3023

PR #3023, opened 2024-05-30
Add i2s support for spi_v3.

liarokapisv:
This PR addresses #3003, revising the interface of the `I2S` driver to a ring-buffered one, similar to the `SAI` driver.


liarokapisv:
I tried to make the changes as minimal as possible. This is "working" with properly configured `tlv320aic32x4` codec. I have only tested with defmt and oscilloscope, but I need to do some more extensive testing because I don't have the audio jacks connected at the moment. However, I think it is ready for a preliminary review.

liarokapisv:
I confirmed that the audio playback works great with these changes.

vDorst:
I am trying this PR. Still working to converting my old audio project to the latest embassy version.
But I am missing the option to use this with a I2S without MCLK pin. Like on the `STM32F401` `SPI3`.
So I added quick added `new_txonly_nomclk()`and `new_inner_nomclk()` to support this mode.

Other think I was missing is a `stop()`to stop the ringbuffer/put the device in low power mode.

liarokapisv:
Thanks, I will add the variants in the PR along with the overflow handling.

vDorst:
I tested it on a stm32F401 (Blackpill) with a PCM5102 Audio DAC.
It works fine.
I would also to see that the ringbuffer `write_immediate()`is also implemented so that you can prefill the buffer before starting the ringbuffer.

liarokapisv:
Added all above suggestions. Does going through the whole `stop().await` dance in the `stop` method seem like a good idea, or should we use `request_stop` instead?
@Dirbaio Want to weight in?

liarokapisv:
@Dirbaio  After the merge this is now very straightforwardly based on the main branch.
With the ringbuffer rework I have succesfully being using this with a 80% DSP processor load at 48k sampling rate without Overruns.

The latest update provides split functionality which enables a pattern for which I was using ping-pong buffers before:

```rust

let mut i2s = ...;
let new_buffer = Signal::<[f32 ; BUFFER_SIZE]>::new();
i2s.start();
let (mut reader, mut writer) = i2s.split();
join!(
    async {
        loop {
            let mut buffer = [0.0 ; BUFFER_SIZE];
            reader.read(&mut buffer).await;
            new_buffer.signal(buffer);
        }
    },
    async {
        loop {
            let buffer = new_buffer.wait().await;
            // process
            writer.write(&buffer).await;
        }
    }
).await;
```

This enables some flexibility in the case of Overruns since clearing one will not affect the timing of the other leading to cascading Overrun failures. It also enables doing parallel reads/writes.

lulf (approved):

PR #3029, opened 2024-06-01
STM32: Generic Futures for Timers

brunob45:
- Move Timer Future to its own file
- Make the interrupt handler capable to handle all interrupts sources
- Add async functions for pwm_input
- Bind interrupt to UpdateInterrupt, as it is defined for all TIM
- Tested on
  - [ ] stm32f103c8
  - [x] stm32f446re


romainreignier:
Thanks @brunob45 for theses changes.
I have tested the input_capture and pwm_input examples STM32F411 and STM32F103 and they seemed to work.
But I have then changed the code to only wait for the falling edge in order to only query the pulse width:
```rust
loop {
    info!("Wait for end of pulse...");
    let width = pwm_input.wait_for_falling_edge().await;
    info!("Pulse width = {} ticks", width);
}
```
In that case, I have an infinite loop like if the interrupt flag is never cleared.

examples/stm32f1/src/bin/pwm_input.rs
```diff
@@ -38,7 +38,7 @@ async fn main(spawner: Spawner) {
 
     unwrap!(spawner.spawn(blinky(p.PC13)));
 
-    let mut pwm_input = PwmInput::new(p.TIM2, p.PA0, Pull::None, khz(10));
+    let mut pwm_input = PwmInput::new(p.TIM2, p.PA0, Pull::None, Irqs, khz(10));
     pwm_input.enable();
 
     loop {
```
romainreignier, line 44: It would be nice to also use the futures in this example too.

brunob45: copied example from stm32f4

brunob45:
> In that case, I have an infinite loop like if the interrupt flag is never cleared.

Found the issue, I was clearing the ch1 interrupt flag instead of ch2 in `wait_for_falling_edge`

showier-drastic:
The comments for CaptureCompareX in the InterruptFlag enum have typos (numbers are all 1).

```
/// All timer interrupts
#[derive(Clone, Copy)]
pub enum InterruptFlag {
    /// Update
    Update = 0,
    /// Capture/compare 1
    CaptureCompare1 = 1,
    /// Capture/compare 1     <==== here
    CaptureCompare2 = 2,
    /// Capture/compare 1     <==== here
    CaptureCompare3 = 3,
    /// Capture/compare 1     <==== here
    CaptureCompare4 = 4,
    /// COM event
    ComEvent = 5,
    /// Trigger
    Trigger = 6,
    /// Break
    Break = 7,
}
```

embassy-stm32/src/timer/ll_async.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Timer future
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct TimerEventFuture<T: GeneralInstance4Channel> {
+    flag: InterruptFlag,
+    phantom: PhantomData<T>,
+}
+
+impl<T: GeneralInstance4Channel> TimerEventFuture<T> {
+    /// Enable the interrupt source and returns a new instance of Future
+    pub fn new(flag: InterruptFlag) -> Self {
```
EliteTK, line 40: I could be wrong. Isn't it the case that if new is pub then anyone can create a future and then drop it causing that interrupt to become disabled.

brunob45:
See #5431

PR #3031, opened 2024-06-01
stm32/gpio: refactor AfType

honzasp:
This PR refactors how pins are configured for alternate function (AF). Instead of configuring a pin with a tuple of `(AFType, Pull)` and optionally with `Speed`, the pin is now configured with the following enum:

```rust
pub enum AfType {
    /// Input with optional pullup or pulldown.
    Input(Pull),
    /// Output with output type and speed and no pull-up or pull-down.
    Output(OutputType, Speed),
    #[cfg(gpio_v2)]
    /// Output with output type, speed and pull-up or pull-down;
    OutputPull(OutputType, Speed, Pull),
}
```

Putting all possible AF settings into a single enum is hopefully more ergonomic, and it also increases type safety because only valid settings can be expressed:

- It's not possible to configure input pin with output speed (which would switch the pin to output on gpio_v1).
- It's always required to explicitly set speed for an output pin (as a follow-up, we might consider allowing the user to configure the output speed for drivers like I2C or SPI).
- It's only possible to configure a pull-up or pull-down resistor with an output pin on gpio_v2, because the hardware doesn't support that on gpio_v1.

### DIFF
--- a/embassy-stm32/src/gpio.rs
+++ b/embassy-stm32/src/gpio.rs
@@ -32,7 +32,9 @@ impl<'d> Flex<'d> {
     }
 
     /// Put the pin into input mode.
-    #[inline]
+    ///
+    /// The internal weak pull-up and pull-down resistors will be enabled according to `pull`.
+    #[inline(never)]
     pub fn set_as_input(&mut self, pull: Pull) {
         critical_section::with(|_| {
             let r = self.pin.block();
@@ -51,35 +53,35 @@ impl<'d> Flex<'d> {
                     Pull::None => vals::CnfIn::FLOATING,
                 };
 
-                let crlh = if n < 8 { 0 } else { 1 };
-                r.cr(crlh).modify(|w| {
+                r.cr(n / 8).modify(|w| {
                     w.set_mode(n % 8, vals::Mode::INPUT);
                     w.set_cnf_in(n % 8, cnf);
                 });
             }
             #[cfg(gpio_v2)]
             {
-                r.pupdr().modify(|w| w.set_pupdr(n, pull.into()));
+                r.pupdr().modify(|w| w.set_pupdr(n, pull.to_pupdr()));
                 r.otyper().modify(|w| w.set_ot(n, vals::Ot::PUSHPULL));
                 r.moder().modify(|w| w.set_moder(n, vals::Moder::INPUT));
             }
         });
     }
 
-    /// Put the pin into output mode.
+    /// Put the pin into push-pull output mode.
     ///
     /// The pin level will be whatever was set before (or low by default). If you want it to begin
     /// at a specific level, call `set_high`/`set_low` on the pin first.
-    #[inline]
+    ///
+    /// The internal weak pull-up and pull-down resistors will be disabled.
+    #[inline(never)]
     pub fn set_as_output(&mut self, speed: Speed) {
         critical_section::with(|_| {
             let r = self.pin.block();
             let n = self.pin.pin() as usize;
             #[cfg(gpio_v1)]
             {
-                let crlh = if n < 8 { 0 } else { 1 };
-                r.cr(crlh).modify(|w| {
-                    w.set_mode(n % 8, speed.into());
+                r.cr(n / 8).modify(|w| {
+                    w.set_mode(n % 8, speed.to_mode());
                     w.set_cnf_out(n % 8, vals::CnfOut::PUSHPULL);
                 });
             }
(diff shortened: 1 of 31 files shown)


embassy-stm32/src/gpio.rs
```diff
@@ -51,8 +53,7 @@ impl<'d> Flex<'d> {
                     Pull::None => vals::CnfIn::FLOATING,
                 };
 
-                let crlh = if n < 8 { 0 } else { 1 };
-                r.cr(crlh).modify(|w| {
+                r.cr(n / 8).modify(|w| {
```
honzasp, line 56: I checked that rustc should be able to infer that `n / 8 < 2` and skip the bounds check in `r.cr(...)`.

honzasp:
@Dirbaio I noticed that the "ci/test" and "ci/test-nightly" jobs passed even when tests/stm32/bin/gpio.rs failed to compile. Looks like the tests in tests/stm32 are not run, is that expected?

honzasp:
Looks like I broke the CI again :(

Dirbaio:
bender run

embassy-stm32/src/gpio.rs
```diff
@@ -521,32 +530,46 @@ impl<'d> OutputOpenDrain<'d> {
...
-    Input,
-    /// Output, drive the pin both high or low.
-    OutputPushPull,
-    /// Output, drive the pin low, or don't drive it at all if the output level is high.
-    OutputOpenDrain,
+pub enum AfType {
+    /// Input with optional pullup or pulldown.
+    Input(Pull),
+    /// Output with output type and speed and no pull-up or pull-down.
+    Output(OutputType, Speed),
```
Dirbaio, line 569: `AfType::Output(_, _)` is the same as `AfType::OutputPull(_, _, Pull::None)`

it's a bit strange to have "redundancy" in enums like this, and i think this will generate extra code in functions that take an `AfType` even if it's not used. It's probably better to:

- make a single `Output` case in all versions.
- The `Output` case contains `OutputType, Speed` on v1, and `OutputType, Speed, Pull` on v2.
- since it's not nice that the struct changes depending on gpio version (making it hard for the user to write code that works on both versions), perhaps we can make the enum private, expose an opaque `pub struct AfType(AfTypeInner)`, and add methods to construct them. so `fn output(type, speed)` is available on all versions and `fn output_pull(type, speed, pull)` only on v2, but they both return the same enum case.

honzasp: The reason why I chose to introduce a conditional `OutputPull` variant instead of a conditional field in the `Output` variant was to minimize the number of `#[cfg]`-s for users while keeping the code as simple as possible (no need for a wrapped private type and/or constructor methods). Note that `AfType` is by design matched exactly in once place, in the `SealedPin::set_as_af()` method, so the cost of extra enum variant in terms of _lines_ of code is quite small.

However, if you are worried about the cost in _bytes_ of code in flash, that's another matter. In the happy case, where `set_as_af()` is called with a statically known pin and a statically known `AfType`, the compiler should be able to reduce everything to a constant a produce the optimal code. But if the compiler is not able to optimize out the `AfType` enum, the code size might suffer, because `set_as_af()` will need a lot of code to convert all the enums to register values. This is potentially made worse by `set_as_af()` being a trait method and marked as `#[inline]`, so we could end up with many copies of the same code in flash.

I will need to inspect the size of generated code to verify my assumptions, but here are some options how we could minimize the size of the generated code:

1. Move the code of `set_as_af()` into a function outside of the `SealedPin` trait and don't mark it as `#[inline]`: this means that we will have just a single copy of this code in flash, so its size would not be that important, and every usage of `set_as_af()` should compile to a function call, which should be quite small (probably smaller than the most optimized inline code, but I need to verify that)
2. Change `AfType` from an enum to a struct with private fields, and store the register values directly in this struct (e.g. for gpio_v2, the struct would contain fields `pupdr`, `ot` and `ospeedr`), and introduce constructor functions like `AfType::input()` and `AfType::output_pull()` to create this struct from user-friendly enums like `Pull` and `Speed`. This would move all the `match`-es out of the `set_as_af()` function into the constructor functions, where the compiler should be able to almost always optimize them out.

I will do some experiments and get back with the results.

honzasp:
I tested five options for `AfType`:

- Option 1 (branch `pin-af`): the original approach from this PR (`AfType` as enum with variants `Input`, `Output` and `OutputPull`)
- Option 2 (branch `pin-af-two-variants`): your proposed apprach (`AfType` as a newtype that wraps `AfTypeInner`, which has two variants `Input` and `Output` and constructor methods on `AfType` which deal with the gpio_v1/gpio_v2 difference
- Option 3 (branch `pin-af-struct`): alternative approach where `AfType` is a struct that directly contains the register values of PUPDR, OT and OSPEEDR.
- Option 4 (branch `pin-af-noinline`): same as Option 1, but the code of `set_as_af()` turned into a function that is `#[inline(never)]`
- Option 5 (branch `pin-af-struct-noinline`): same as Option 3, but the code of `set_as_af()` turned into a function that is `#[inline(never)]`.

I compared compilation sizes of my project, which initializes a two I2Cs, one SPI, three UARTs and some GPIOs and timers and runs on STM32H743ZI. The sizes were as follows:

Option | Size `-Os` | Size `-O3`
---|---|---
Option 1 (`pin-af`) | 79_704 | 102_784
Option 2 (`pin-af-two-variants`) | 79_704 | 102_784
Option 3 (`pin-af-struct`) | 79_704 | 102_784
Option 4 (`pin-af-noinline`) | 79_184 | 102_280
Option 5 (`pin-struct-noinline`) | 79_152 | 102_176

It seems that the compiler is able to completely optimize out the `AfType` in options 1, 2 and 3; this makes sense, because at the time when `set_as_af()` is called in a device driver, both the pin and the `AfType` are constants. In this case, the exact representation of `AfType` is not relevant for code size.

However, code size decreased by almost 1 % if we don't inline `set_as_af()` into the caller, and the struct representation of `AfType` shaved additional bytes off the code size.

---

So my proposal is to use the struct representation of `AfType`, with helper constructor methods that accept user-friendly enums like `Pull` and `Speed`:

```rust
pub struct AfType {
    pupdr: vals::Pupdr,
    ot: vals::Ot,
    ospeedr: vals::Ospeedr,
}
```

My little experiment also suggests that to minimize binary size, we should turn `set_as_af()` and some other initialization functions (such as `Flex::set_as_input()` and `SealedPin::set_as_analog()` (which is used a lot in drivers as `SealedPin::set_as_disconnected()`) into functions that are not inlined and not monomorphised.

@Dirbaio, what are your thoughts?

honzasp:
I went ahead and implemented option 5, turning `AfType` into a struct of register values and marking `set_as_af()` and a few other cold functions into `#[inline(never)]`.

honzasp:
Rebased on main to resolve merge conflicts (from another PR of mine :)). If you are OK with the struct-based `AfType`, we should be ready to go.

Dirbaio (approved):
thank you!

honzasp:
Hmm, are the CI failures due to some bug introduced in this PR, or is it an unrelated failure? :thinking:

honzasp:
We are getting the same error every time. Unfortunately, I don't have the hardware (stm32wl55jc) and I don't even know how could I run the tests myself. I also double-checked the changes to the UART driver and couldn't find anything that looks suspicious. Is there a chance that the test was already broken before this PR?

honzasp:
Rebased on `main` to resolve a merge conflict, perhaps it will fix the CI error?

Dirbaio:
it worked now :man_shrugging: 

I think the failures were unrelated to this PR, there's been some instability in ringbuffered uart in the past as well, the test is disabled in some chips.

honzasp:
Awesome, thanks! :)

PR #3032, opened 2024-06-02
Remove more BXCAN generics.

cschuhen:
Remove the rest of the generics in BXCAN. This covers the non-split Can and CanConfig structs.


Dirbaio (approved):
thanks!

Issue #1411, opened 2023-04-29
`embassy_stm32` incorrectly implements `embedded_hal_async::spi::SpiBus`


cgm616:
According to the [documentation of `embedded-hal-async` as of version 0.2.0-alpha.1](https://docs.rs/embedded-hal-async/0.2.0-alpha.1/embedded_hal_async/spi/trait.SpiBus.html#tymethod.transfer) (the version used by `embassy-stm32`), implementers of the `SpiBus` train are supposed to complete the SPI transfer when `SpiBus::transfer` is called with buffers of different size:

> It is allowed for read and write to have different lengths, even zero length. The transfer runs for max(read.len(), write.len()) words. If read is shorter, incoming words after read has been filled will be discarded. If write is shorter, the value of words sent in MOSI after all write has been sent is implementation-defined, typically 0x00, 0xFF, or configurable.

However, the current implementation of the trait provided by `embassy-stm32` for its SPI devices panics if the buffers are different lengths. Here is the relevant code, in [`embassy-stm32/src/spi/mod.rs`](https://github.com/embassy-rs/embassy/blob/3e730aa8b06401003202bf9e21a9c83ec6b21b0e/embassy-stm32/src/spi/mod.rs#L522): 

```rust
async fn transfer_inner<W: Word>(&mut self, read: *mut [W], write: *const [W]) -> Result<(), Error>
    where
        Tx: TxDma<T>,
        Rx: RxDma<T>,
    {
        let (_, rx_len) = slice_ptr_parts(read);
        let (_, tx_len) = slice_ptr_parts(write);
        assert_eq!(rx_len, tx_len); // this is the offending panic!
       
        ...snip...
}
```

As a workaround, you can use `SpiBus::transfer_in_place` right now if you don't care about keeping a copy of the data sent. However, `SpiBus::transfer` should have the right semantics.

ctron:
I just ran into the same issue.

PR #3034, opened 2024-06-02
Fix STM32 SPI examples

phoracek:
In their current state, these examples fail with:

```
     ERROR panicked at /home/phoracek/.cargo/git/checkouts/embassy-9312dcb0ed774b29/5f8f867/embassy-stm32/src/spi/mod.rs:769:9:
     assertion `left == right` failed
       left: 128
      right: 20
```

It is due to `String::as_bytes` returning slice shrinked to the occupied size of the string.

With this change, these examples are aligned with the BDMA example, using read and write buffers.

### DIFF
--- a/examples/stm32f3/src/bin/spi_dma.rs
+++ b/examples/stm32f3/src/bin/spi_dma.rs
@@ -21,11 +21,17 @@ async fn main(_spawner: Spawner) {
 
     let mut spi = Spi::new(p.SPI1, p.PB3, p.PB5, p.PB4, p.DMA1_CH3, p.DMA1_CH2, spi_config);
 
+    let mut read_buffer = [0; 128];
+    let mut write_buffer = [0; 128];
+
     for n in 0u32.. {
         let mut write: String<128> = String::new();
-        let mut read = [0; 128];
         core::write!(&mut write, "Hello DMA World {}!\r\n", n).unwrap();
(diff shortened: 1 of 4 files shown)


phoracek:
The example does not fail on the size check now, but I have not been able to test it end-to-end yet due to another issue:

```
ERROR panicked at /home/phoracek/.cargo/git/checkouts/embassy-9312dcb0ed774b29/5f8f867/embassy-stm32/src/dma/dma_bdma.rs:285:21:
DMA: error on DMA@40020000 channel 3
```

phoracek:
It required additional tweak to overcome https://community.st.com/t5/stm32-mcus/dma-is-not-working-on-stm32h7-devices/ta-p/49498 and make it work on STM32H7.

I have posted the required change to the H7 example. Although it may be required on other boards too.

I have tested this myself on stm32h750ib and it worked fine.

Dirbaio:
The cause of the crash is https://github.com/embassy-rs/embassy/issues/1411, which we should fix. Once that's done, transfers with different read/write lengths should be allowed, so the examples won't crash anymore.

Other than that, afaict the only broken examples are h7 and h7rs, which do `spi.transfer(&mut read, write.as_bytes())`. The others do `spi.transfer(&mut read[0..write.len()], write.as_bytes())` which shouldn't crash.

phoracek:
I see, gotcha. I will close this then. Thanks for the review

PR #3040, opened 2024-06-04
prepare for embassy-time 0.3.1 release

lulf:
A release a day, keeps the doctor away. 

This adds a convenience script I've been using to bump crate versions. I'd like to improve on this, maybe converting it to a project-specific crate that discover the dependency trees for a crate in the repo and bump it that way, but for now it's better than nothing.


lulf:
bender run

PR #3042, opened 2024-06-05
Adding function to close soft AP.

### LINKED ISSUE #2991: Closing of AP embassy::cyw43
Hello,

I am trying to open an access point on a Pico W to receive WiFi credentials, then join the new WiFi network using these credentials. However, I have encountered some issues:

As far as I can tell, there is no method to close an access point (AP) and the code panics when I attempt to use `.join_wpa2()` without closing it.

Am I missing something? Or is this functionality not yet implemented?

Thank you!

GustavToft:
I am unaware if this is the best way to do it but it seems to work when I test it on a Pico W. 

Let me know if it needs anything else.

fixes #2991


Dirbaio (approved):
thanks!

PR #3044, opened 2024-06-05
rp: Add multichannel ADC

kalkyl:
Adds "round robin" sampling so that multiple channels can be sampled at once.
Including a new example for how to use the ADC with DMA, for both single- and multichannel reads.


jamesmunns (approved):
Looks good to me!

PR #3050, opened 2024-06-05
add send_queue and recv_queue to embassy-net

ant32:
I need to know if queue is almost empty before sending low priority data.


Dirbaio (approved):
thanks!

PR #3052, opened 2024-06-06
embassy-sync: remove T: Send for Signal<M, T>

### LINKED ISSUE #3051: Signal and !Send types?
I'm wondering if there's a reason why the [implementation of `Signal` specifies `T: Send`](https://github.com/embassy-rs/embassy/blob/d06dbf3/embassy-sync/src/signal.rs#L68)? It seems to me that this is handled by the type of mutex used, am I missing something?

jvatic:
The tests all pass with this locally, and I've also tested it in a project using a `NoopRawMutex`.

Closes #3051

### DIFF
--- a/embassy-sync/src/signal.rs
+++ b/embassy-sync/src/signal.rs
@@ -65,7 +65,7 @@ where
     }
 }
 
-impl<M, T: Send> Signal<M, T>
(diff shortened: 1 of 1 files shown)


lulf (approved):
Lgtm, thanks for the PR! :rocket:

PR #3054, opened 2024-06-06
Add Clone and Copy to Error types

dvdsk:
None of them these errors are `non-exaustative`, they are all small enough to be copy (I estimate none are larger than 4 bytes). All of them seem done and its not probable (to me) that large variants will be added in the future.


Dirbaio (approved):
thanks!

dvdsk:
I think this one might need another attempt at the merge queue :)

PR #3057, opened 2024-06-07
Document w5500 State and add w5500 example for stmf4

dvdsk:
This example takes into account the lower memory on the stm32f4. That
should prevent anyone wanting to use the w5500 on any stm from adapting
the w5500 example for the rp which uses a lot more RAM.

On chips with a low amount of ram it's easy to run out of ram. When
looking at the current docs for the State struct it is not easy to
see that these params can cause a lot of ram usage.


Dirbaio (approved):
thanks!

PR #3059, opened 2024-06-09
use nightly waker_getters APIs

zjp-CN:
Since https://github.com/rust-lang/rust/issues/96992 has stalled, to prevent potential unsoundness caused by transmuting to &WakerHack, we can use nightly waker_getters APIs by gating it behind nightly feature in embassy-executor without waiting for it to be stablized.

### DIFF
--- a/embassy-executor/src/raw/waker.rs
+++ b/embassy-executor/src/raw/waker.rs
@@ -31,46 +31,36 @@ pub(crate) unsafe fn from_task(p: TaskRef) -> Waker {
 /// # Panics
 ///
 /// Panics if the waker is not created by the Embassy executor.
-#[cfg(not(feature = "nightly"))]
 pub fn task_from_waker(waker: &Waker) -> TaskRef {
-    // safety: OK because WakerHack has the same layout as Waker.
-    // This is not really guaranteed because the structs are `repr(Rust)`, it is
-    // indeed the case in the current implementation.
-    // TODO use waker_getters when stable. https://github.com/rust-lang/rust/issues/96992
-    let hack: &WakerHack = unsafe { core::mem::transmute(waker) };
-    if hack.vtable != &VTABLE {
-        panic!("Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.")
+    #[cfg(not(feature = "nightly"))]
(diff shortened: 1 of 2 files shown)


embassy-executor/src/raw/waker.rs
```diff
@@ -46,7 +46,31 @@ pub fn task_from_waker(waker: &Waker) -> TaskRef {
     unsafe { TaskRef::from_ptr(hack.data as *const TaskHeader) }
 }
 
+#[cfg(not(feature = "nightly"))]
 struct WakerHack {
     data: *const (),
     vtable: &'static RawWakerVTable,
 }
+
+/// Get a task pointer from a waker.
```
Dirbaio, line 55: can you put the cfgs inside the function, so the function doc comment and the signature aren't duplicated, to ensure they don't go out of sync?

Dirbaio (approved):
thanks!

zjp-CN:
Sorry, I realized I could make the cfg code better

```rust
pub fn task_from_waker(waker: &Waker) -> TaskRef {
    let (vtable, data) = {
        #[cfg(not(feature = "nightly"))]
        {
            struct WakerHack {
                data: *const (),
                vtable: &'static RawWakerVTable,
            }

            // safety: OK because WakerHack has the same layout as Waker.
            // This is not really guaranteed because the structs are `repr(Rust)`, it is
            // indeed the case in the current implementation.
            // TODO use waker_getters when stable. https://github.com/rust-lang/rust/issues/96992
            let hack: &WakerHack = unsafe { core::mem::transmute(waker) };
            (hack.vtable, hack.data)
        }
        #[cfg(feature = "nightly")]
        {
            let raw_waker = waker.as_raw();
            (raw_waker.vtable(), raw_waker.data())
        }
    };

    if vtable != &VTABLE {
        panic!("Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.")
    }
    // safety: our wakers are always created with `TaskRef::as_ptr`
    unsafe { TaskRef::from_ptr(data as *const TaskHeader) }
}
```

But it's now merged...

Dirbaio:
oh, that's true! can you send it in another PR?

PR #3060, opened 2024-06-09
Make BDMA work in polled mode without interrupts

korken89:
During the initialization of an RTIC app it was noticed that the BDMA needed interrupts to operate. 
This PR fixes this which allows it to function in an interrupts-off environment with a polling executor.

Basically to allow the use of `block_on` while initializing an RTIC app.

The condition is used in two places now, however when in interrupt driven mode it uses the original one and in polling mode it uses the new one.
Maybe one should break out the check to not write this twice.

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -495,7 +495,11 @@ impl AnyChannel {
                 let ch = r.ch(info.num);
                 let en = ch.cr().read().en();
                 let circular = ch.cr().read().circ();
-                let tcif = state.complete_count.load(Ordering::Acquire) != 0;
+                let isr = r.isr().read();
+                let cr = r.ch(info.num).cr();
+                let tcif =
+                    state.complete_count.load(Ordering::Acquire) != 0 || (isr.tcif(info.num) && cr.read().tcie());
+
                 en && (circular || !tcif)
             }
         }


jamesmunns:
Noting my comment from chat:

Consistency wise, idk if any of the rest of the embassy drivers assume that they SHOULD work at all with ISRs off. Adding that could be a maintenance headache, as DMA/driver futures will need more complex logic to consider both possible states.

IMO, @korken89 might be better off adding a `select` in the block_on that checks the status flags and manually calls the dma `on_irq` function when appropriate.

I believe this is likely to work, but is at odds architecturally with the rest of the stm32 driver design assumptions.

korken89:
Some reflections from the chat on if one should move it all to the Future:

> Ah yes, I found why it's in the interrupt now. I tried to move it all outside and you can't do the "normal" pattern. That is, on interrupt you disable the interrupt generation bit - but don't clear the flag. And let the Future handle the clearing of the interrupt flag.
> The BDMA does not allow of disabling the interrupt generation bit while the DMA is running, and we can't disable it as that would cancel the transfer.
> So when you want to return to the Future for doing management, you will just get sent back to the IRQ.
That was a quite unfortunate implementation detail of the BDMA.

The final idea is to use the NVIC instead to block the interrupt source.

Dirbaio:
in general it's impossible to make all drivers work with interrupts disabled. Some because they're designed to do the work in the interrupt handler (e.g BufferedUart), others because of (arguably) bad design of the hardware, like in BDMA and USB OTG.

So the question becomes: should we try to make drivers work with interrupts disabled *when it is feasible to do so*? I sort of lean towards no, because:
- It's not something users would be able to rely on consistently, if some drivers work and others don't
- it makes the code more complex and duplicates the testing effort (we have to test with irqs enabled, and disabled)
- Makes it slower/bigger for everyone even if you don't care about masked-irqs operation.

> The final idea is to use the NVIC instead to block the interrupt source.

This is not doable because multiple DMA channels share the same NVIC IRQ line. If you disable it when one channel finishes, the others will stop working.

korken89:
Thanks for the input @Dirbaio !
As you state, I also don't think it's worth it in the general case but if it's trivial to get it to work I would also not be opposed.

What do you think of the small fix I did here for example? It can be made so it's not a code duplication.

Dirbaio:
unfortunately, i'd rather not to since this has a performance impact for all users.

there's many people using embassy for realtime things where extra spent clock cycles matter. We try to make async work for these use cases, we try to fight the "async is fundamentally incompatible with realtime!" misconceptions. Due to this, i'd prefer to prioritize perf.

thanks anyway for the PR!

PR #3062, opened 2024-06-10
fix docu typos, linker addr example

kkrolczyk:
This fixes a couple of typos, and adjusts memory range for bootloader example.

While the bootloader example is unlikely to be used alone - ./examples/boot/application/stm32h7/flash-boot.sh for example overwrites it temporarily with different file altogether, it might be less confusing to anyone wondering why the numbers don't match.


lulf (approved):
Thanks!

PR #3063, opened 2024-06-11
dns: properly handle AddrType::Either in get_host_by_name()

aurelj:
Currently, a DNS query with AddrType::Either will always result to an A query returning a v4 address, even when the stack is only configured in v6 and/or the host name only associated with a v6 address, which makes it pretty useless.

This change allows get_host_by_name() to make up to 2 queries (A and AAAA) for AddrType::Either (depending on wether the first query is successful or not), and it will prioritize the appropriate query depending on which stack version is configured.


Dirbaio (approved):
makes a lot of sense, thank you!

PR #3064, opened 2024-06-11
Clarified comments about the cortex_m::asm::delay functionality in al…

Duesentrieb71:
This clarifies comments about the cortex_m::asm::delay functionality in all multipro.rs examples.

I was confused when I ran multipro.rs on my Pi Pico because I was getting different timings for the threads than what was stated by the comments next to the delay call. It took a while of confused troubleshooting before I was told the answer in the Matrix chat and I want to spare others the wasted time. 
This function only guaranties a minimum number of CPU cycles and these can't simply be translated to seconds via the clock frequency because every chip model has a different conversion factor for this.
My changes look like the following:
Current: `cortex_m::asm::delay(32_000_000); // ~1 second`
Changed: `cortex_m::asm::delay(32_000_000); // At least 32_000_000 CPU cycles (could be ~1 second)`

I left the statements about the (depending on the chip, wrong) seconds in, because I think it is nonetheless useful to know that we are looking for something in the seconds range instead of milliseconds. But this can be changed if desired.


Dirbaio:
thanks for the PR! I agree the current status quo is confusing. iirc I wrote the timings for nRF, then they've been copypasted across many stm32's with wildly different clock speeds without updating.

However, while we're changing it, perhaps it would be better to use `embassy_time::block_for()`? this way the time would be guaranteed correct on all chips without having to adapt the clock cycle count.

Duesentrieb71:
Good idea. I changed it to embassy_time::block_for()

Duesentrieb71:
Sorry, I don't understand.I think I removed them in [this commit](https://github.com/embassy-rs/embassy/commit/6c246f29b107f5ad70d67b62f6b3349ca15b1d65).

Dirbaio:
thank you!

Duesentrieb71:
Gladly. Thank you for this awesome framework and the opportunity to do my first open source contribution!

PR #3066, opened 2024-06-12
Add missing word "pin" in rp pwm documentation

mikkelens:
It is confusing to read without. The corrected line reads like its 'b' pin counterpart (few lines below).


jamesmunns (approved):
Thanks!

PR #3073, opened 2024-06-13
 probe-rs download 43439A0.bin --format bin deprecated

1-rafael-1:
probe-rs download 43439A0.bin --format bin --chip RP2040 --base-address 0x10100000
-->
Error: --format has been renamed to --binary-format. Please use --binary-format bin instead of --format bin


Dirbaio (approved):
thanks!

Issue #3076, opened 2024-06-14
Examples should have a "different chip" checklist


jamesmunns:
It would be good to have a checklist in the source file of every example, to ensure people trying to use examples for different chips/boards know what they need to do. This is a VERY common question in chat, which means likely even more people who never show up in chat have the same issue.

I think having this in-code would be very helpful (from the stm32f4 blinky example):

```rust
// This example is written for the stm32f429-nucleo, with a stm32f429zi chip.
//
// If you are using a different board or chip, make sure you update the following:
//
// * [ ] Update .cargo/config.toml with the correct `probe-rs run --chip STM32F429ZITx` chip name
// * [ ] Update Cargo.toml to have the correct `embassy-stm32` feature, it is currently `stm32f429zi`
// * [ ] If your board has a special clock or power configuration, make sure that it is set up
//       appropriately. If you are unsure, please drop by the Embassy Matrix chat for support.
let p = embassy_stm32::init(Default::default());
info!("Hello World!");
```

This is a bit tedious to write out, but I think it will be helpful to reduce the amount of frustration from new users.

jamesmunns:
If we do want to do this, here's a tracking list of all example kinds that need this:

* [ ] boot
* [ ] nrf-rtos-trace
* [ ] nrf51
* [ ] nrf52810
* [ ] nrf52840
* [ ] nrf52840-rtic
* [ ] nrf5340
* [ ] nrf9160
* [ ] rp
* [ ] std
* [ ] stm32c0
* [ ] stm32f0
* [ ] stm32f1
* [ ] stm32f2
* [ ] stm32f3
* [ ] stm32f334
* [x] stm32f4 - #3077 
* [ ] stm32f469
* [ ] stm32f7
* [ ] stm32g0
* [ ] stm32g4
* [ ] stm32h5
* [ ] stm32h7
* [ ] stm32h7rs
* [ ] stm32l0
* [ ] stm32l1
* [ ] stm32l4
* [ ] stm32l5
* [ ] stm32u0
* [ ] stm32u5
* [ ] stm32wb
* [ ] stm32wba
* [ ] stm32wl
* [ ] wasm

jedkutai:
I’m interested in picking this up. I saw that #3077 was closed during the old PR cleanup. Would you still like this implemented, and is the preferred approach now a short note in each example linking to centralized documentation?

Dirbaio:
[...]

PR #3077, opened 2024-06-14
Add docs to all stm32f4 examples

jamesmunns:
Partial impl of #3076 for stm32f4

### DIFF
--- a/examples/stm32f4/src/bin/adc.rs
+++ b/examples/stm32f4/src/bin/adc.rs
@@ -8,6 +8,24 @@ use embassy_stm32::adc::{Adc, Temperature, VrefInt};
 use embassy_time::{Delay, Timer};
 use {defmt_rtt as _, panic_probe as _};
 
+/// This example is written for the nucleo-stm32f429zi, with a stm32f429zi chip.
+///
+/// If you are using a different board or chip, make sure you update the following:
+///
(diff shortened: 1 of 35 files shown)


jamesmunns:
IMO:

* This is annoyingly copy-paste of the same docs, we COULD have it in a top level readme
* BUT: putting the information as front and center as possible, so it's impossible to miss is likely to be useful

Open to other opinions on this, as the advice is general enough that IMO it is unlikely to change.

joelsa:
A sort of middle-ground would be to have

```
/// This example is written for the nucleo-stm32f429zi, with a stm32f429zi chip.
///
/// If you are using a different board or chip, make sure to follow the steps described here: https://linktoreadmeorfaq.abc/xyz
```
In all the examples and then link to the top-level-readme and/or the FAQs.

lulf:
I agree with the idea of putting information front and center, but I'm a bit worried about the maintainability of it all.

IMHO it only needs to be something that screams 'IF YOU HAVE TROUBLE GO TO THIS LINK'. If that doesn't help, then I doubt having a bigger comment with more info helps, even if it's slightly more convenient not having to open a browser.

michelrandahl:
As a beginner I think this is very helpful.

Putting it in a README might be easier to maintain though, and less cluttered. From a beginners viewpoint I suppose that I would have looked at the readme if there were one in the given example dir.

Dirbaio:
Hello,

PR #3079, opened 2024-06-14
implement with_timeout()/with_deadline() method style call on Future

aurelj:
This allows writing
```rust
rx.read(&mut read_buf).with_timeout(Duration::from_millis(200)).await
```

rather than
```rust
with_timeout(Duration::from_millis(200), rx.read(&mut read_buf)).await
```

This looks nicer to me.

### DIFF
--- a/embassy-time/src/lib.rs
+++ b/embassy-time/src/lib.rs
@@ -32,7 +32,7 @@ pub use delay::{block_for, Delay};
 pub use duration::Duration;
 pub use embassy_time_driver::TICK_HZ;
 pub use instant::Instant;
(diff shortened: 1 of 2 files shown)


dvdsk:
I agree that it looks nicer, I was left wondering why the rest of the rust async ecosystem like tokio did not go for such an approach? Is there some disadvantage I am not seeing. Then I found out tokio does support such an api: https://docs.rs/tokio-util/latest/tokio_util/time/trait.FutureExt.html#method.timeout its just in tokio-util (still part of the tokio project) not tokio itself.

lulf (approved):
I think I'm more a fan of the original approach, since it allows spotting a timeout being applied immediately, but I respect it's a matter of taste, so LGTM.

PR #3084, opened 2024-06-15
Improve stm32 ucpd packet reception

sjoerdsimons:
When using stm32g071b-disco as a PD interceptor i noticed especially GoodCrc packets got missed and sometimes corrupted packets were received. This PR seems to fix both issues :)

### DIFF
--- a/embassy-stm32/src/ucpd.rs
+++ b/embassy-stm32/src/ucpd.rs
@@ -347,15 +347,14 @@ impl<'d, T: Instance> PdPhy<'d, T> {
                 .read(r.rxdr().as_ptr() as *mut u8, buf, TransferOptions::default())
         };
 
-        // Clear interrupt flags (possibly set from last receive).
-        r.icr().write(|w| {
-            w.set_rxorddetcf(true);
-            w.set_rxovrcf(true);
-            w.set_rxmsgendcf(true);
-        });
-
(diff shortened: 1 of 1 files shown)


timokroeger:
Changes look good to me and for the very fast GoodCRC messages that totally makes sense.
Now that the receiver is disabled on drop only, it could be enabled in `split_pd_phy()`.

Trying to remember why I implemented it the way it was before. I think it was to match the (unbuffered) uart behaviour where data can only be received when rx future is active.

sjoerdsimons:
> Changes look good to me and for the very fast GoodCRC messages that totally makes sense. Now that the receiver is disabled on drop only, it could be enabled in `split_pd_phy()`.

Good point; changed that.

> Trying to remember why I implemented it the way it was before. I think it was to match the (unbuffered) uart behaviour where data can only be received when rx future is active.
> 
> In case you are interested: https://github.com/timokroeger/usb-pd-lib-embassy I played around a bit a tried to implement basic PD communication some time ago.

Nice didn't see that before; fwiw i've been contributing to https://github.com/fmckeogh/usb-pd-rs/tree/main/usb-pd for the packet parsing; To put together a bit of tooling to debug/analyse usb-pd communications: https://github.com/sjoerdsimons/ucpd-tools

Dirbaio (approved):
thank you!

PR #3088, opened 2024-06-17
nrf: fix wrong order configuring gpios.

Dirbaio:
Docs say "PSEL.RXD, PSEL.RTS, PSEL.RTS, and PSEL.TXD must only be configured when the UARTE is disabled."
For some reason nrf52 doesn't care but nrf91 does.


Dirbaio:
*of course* this breaks nrf52 :sob:

Dirbaio:
fixed in #3121

PR #3089, opened 2024-06-18
WIP: STM32 Half-Duplex: fix sequential reads and writes

qwerty19106:
After #2895 I discoverer the same problems:
- At read-read-write-write-read sequence MCU can read bytes which was sent over `write`.
- The `read` at start MCU can cause infinity loop because `tc` bit will be set only after `write`.

To fix it I allow either transmitter (`cr1.te()`) or receiver (`cr1.re()`) at `write`, `read` and `flush` calls.
Tested on `STM32F446RE` with two Half-Duplex protocols: Smart Audio and CRSF.


Dirbaio:
bender run

Dirbaio:
bender run

PR #3095, opened 2024-06-19
Add PWM examples for stm32g0

Adancurusul:
Add three `pwm` routines
`pwm_capture` is used to read the duty cycle, width and frequency of the input pwm.
`input_capture` is used to read the input edge frequency
`pwm_complementary` provides two sets of complementary pwm outputs with different duty cycles


lulf (approved):
lgtm, thanks!

PR #3106, opened 2024-06-21
STM32 - Adc V4 - Add asynchronous multi-channel read support

liarokapisv:
This PR adds async support to the `adc v4` driver through a new `read_async` method that takes an `ExactSizedIterator` of `AnyChannel` `mut` references along with their sample times.
This can be used for single-channels as well, and is especially useful for large sampling times of many channels, which can waste important MCU time.
Tested on a `STM32H745VIIx` and compared with separate `read_channel` calls.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1178,6 +1178,11 @@ fn main() {
 
     let signals: HashMap<_, _> = [
         // (kind, signal) => trait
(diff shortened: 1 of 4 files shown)


andresv:
Could you add also usage example to this PR.

liarokapisv:
@andresv done.

PR #3110, opened 2024-06-23
stm32 - Adc v4 - Add averaging support.

liarokapisv:
The Adc v4 peripheral includes a hardware oversampler. This PR adds an averaging interface that keeps most of the current interface backwards compatible while allowing for the common use-case of hardware-averaging. A more comprehensive oversampler interface may be exposed in the future.


lulf (approved):
LGTM, thanks!

PR #3111, opened 2024-06-23
Add async API to TSC

Eekle:
First punt at contributing peripheral code, I hope it's close to where it needs to be.

Based off what I read in `rng.rs`, on the advice of @kkoppul2

Some questions I have:

- What, if anything, should happen to the existing `start_it` and `poll_for_acquisition` methods
- Is it okay that this _requires_ Irqs be linked up to use the TSC? Or should it be optional?

### DIFF
--- a/embassy-stm32/src/tsc/mod.rs
+++ b/embassy-stm32/src/tsc/mod.rs
@@ -75,6 +75,7 @@ pub use enums::*;
 
 use crate::gpio::{AfType, AnyPin, OutputType, Speed};
 use crate::interrupt::typelevel::Interrupt;
+use crate::mode::{Async, Blocking, Mode as PeriMode};
 use crate::rcc::{self, RccPeripheral};
 use crate::{interrupt, peripherals, Peripheral};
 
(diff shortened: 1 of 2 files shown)


kkoppul2:
I'm thinking we can have two constructors here, one with the interrupt, one without. Then users can choose how they want to use it. Make the irq an option.

Eekle:
> I'm thinking we can have two constructors here, one with the interrupt, one without. Then users can choose how they want to use it. Make the irq an option.

Would this be best achieved as an additional generic parameter on the `Tsc` struct?

kkoppul2:
Yeah we can add the interrupt handler as an option and then follow the paradigm of the peripheral mode (Async vs Blocking) and have a constructor for each implementation. The Async takes the interrupts, the blocking doesn't. Then maybe we can combine the non-interrupt start function with the polling function and move that to the Blocking impl and the start_it to the Async impl. Most of the bus implementations already follow the pattern for reference.

Eekle:
@kkoppul2 Updated with async/blocking split.

I have just removed the `_it` functions altogether, since the async wait works with the normal `start` func. I hope that's not problematic?

lulf (approved):
LGTM!

embassy-stm32/src/tsc/mod.rs
```diff
@@ -90,6 +93,35 @@ pub enum Error {
...
+
+impl SealedDriverKind for Async {}
+impl SealedDriverKind for Blocking {}
+
+#[allow(private_bounds)]
+/// Driver variant marker for the TSC peripheral
+pub trait DriverKind: SealedDriverKind {}
+
+impl DriverKind for Async {}
+impl DriverKind for Blocking {}
```
lulf, line 111: Instead of defining your own, use the ones from https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/lib.rs#L21-L41  (see SPI for example).

lulf (changes requested):
Just one comment on redefinition of the modes.

embassy-stm32/src/tsc/mod.rs
Eekle: Thank you, changed!

lulf (approved):

PR #3115, opened 2024-06-24
Add http request example to rp

1-rafael-1:
After myself and 2 others had the same question in 3 days on matrix: Maybe You want an example?

This is not strictly an embassy thing, but it took me (= new to Rust and fairly new to embedded) quite some time & a lot of kind help on matrix to figure out. So my guess is others might benefit.
(thanks @jamesmunns especially!)

Word of caution: Binary size grew > 10xxkb and flashing that somehow corrupts the device and the code fails. The exact same code copied into my own project works just fine. So I added a secction to `cargo.toml` to reduce binary size:

```toml
[profile.dev]
lto = true
opt-level = "z"
```

That way the code now runs without issue, with less than half the original binary size. 

-> if that is not okay, then the entire thing should not go into the repo without some other solution. As a newbie I would then need help to do it differently.

### DIFF
--- a/examples/rp/src/bin/wifi_webrequest.rs
+++ b/examples/rp/src/bin/wifi_webrequest.rs
@@ -13,17 +13,17 @@ use serde_json_core;
 use cyw43_pio::PioSpi;
 use defmt::*;
 use embassy_executor::Spawner;
-use embassy_net::{Config, Stack, StackResources};
-use embassy_net::tcp::client::{TcpClientState, TcpClient};
 use embassy_net::dns::DnsSocket;
+use embassy_net::tcp::client::{TcpClient, TcpClientState};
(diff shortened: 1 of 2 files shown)


1-rafael-1:
I tested some random other examples and it does not look like I broke stuff.

examples/rp/src/bin/wifi_webrequest.rs
jamesmunns, line 1: This file has some commented out code, could you add comments why it needs to be there, or remove it?

1-rafael-1: sure, done :-)

jamesmunns (approved):
Looks good to me, let's see what CI thinks :)

PR #3116, opened 2024-06-25
 ring buffered adc v2

sethkrie:
new PR, taking Dirbao's advice to make the DMA impl in a separate struct that consumes Adc<T> to make RingBufferedAdc<T>. Handling overrun similar to RingBufferedUart.

The API is therefore entirely disjoint from the original ADC API.

### DIFF
--- a/embassy-stm32/src/adc/ringbuffered_v2.rs
+++ b/embassy-stm32/src/adc/ringbuffered_v2.rs
@@ -1,0 +1,376 @@
+use core::marker::PhantomData;
+use core::mem;
+use core::sync::atomic::{compiler_fence, Ordering};
+
+use embassy_hal_internal::{into_ref, Peripheral};
+use stm32_metapac::adc::vals::SampleTime;
+
+use crate::adc::{Adc, AdcChannel, Instance, RxDma};
+use crate::dma::ringbuffer::OverrunError;
+use crate::dma::{Priority, ReadableRingBuffer, TransferOptions};
+use crate::pac::adc::vals;
+use crate::rcc;
+
+fn clear_interrupt_flags(r: crate::pac::adc::Adc) {
+    r.sr().modify(|regs| {
+        regs.set_eoc(false);
+        regs.set_ovr(false);
+    });
+}
+
+#[derive(PartialOrd, PartialEq, Debug, Clone, Copy)]
+pub enum Sequence {
+    One,
+    Two,
+    Three,
+    Four,
+    Five,
+    Six,
+    Seven,
+    Eight,
+    Nine,
+    Ten,
+    Eleven,
+    Twelve,
+    Thirteen,
+    Fourteen,
+    Fifteen,
+    Sixteen,
+}
+
+impl From<Sequence> for u8 {
+    fn from(s: Sequence) -> u8 {
+        match s {
+            Sequence::One => 0,
+            Sequence::Two => 1,
+            Sequence::Three => 2,
+            Sequence::Four => 3,
+            Sequence::Five => 4,
+            Sequence::Six => 5,
+            Sequence::Seven => 6,
+            Sequence::Eight => 7,
+            Sequence::Nine => 8,
+            Sequence::Ten => 9,
+            Sequence::Eleven => 10,
+            Sequence::Twelve => 11,
+            Sequence::Thirteen => 12,
+            Sequence::Fourteen => 13,
+            Sequence::Fifteen => 14,
+            Sequence::Sixteen => 15,
+        }
+    }
+}
+
+impl From<u8> for Sequence {
+    fn from(val: u8) -> Self {
+        match val {
+            0 => Sequence::One,
+            1 => Sequence::Two,
+            2 => Sequence::Three,
+            3 => Sequence::Four,
+            4 => Sequence::Five,
+            5 => Sequence::Six,
+            6 => Sequence::Seven,
+            7 => Sequence::Eight,
+            8 => Sequence::Nine,
+            9 => Sequence::Ten,
+            10 => Sequence::Eleven,
+            11 => Sequence::Twelve,
+            12 => Sequence::Thirteen,
+            13 => Sequence::Fourteen,
+            14 => Sequence::Fifteen,
+            15 => Sequence::Sixteen,
+            _ => panic!("Invalid sequence number"),
+        }
+    }
+}
+
+pub struct RingBufferedAdc<'d, T: Instance> {
+    _phantom: PhantomData<T>,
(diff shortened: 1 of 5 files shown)


sethkrie:
Still have a few things to check regarding overrun. Not ready for a merge/review quite yet.

sethkrie:
Ready for review

embassy-stm32/src/adc/ringbuffered_v2.rs
```diff
@@ -0,0 +1,368 @@
...
+            w.set_dma(true);
+            // Enable continuous conversions
+            w.set_cont(vals::Cont::CONTINUOUS);
+            // DMA requests are issues as long as DMA=1 and data are converted.
+            w.set_dds(vals::Dds::CONTINUOUS);
+            // EOC flag is set at the end of each conversion.
+            w.set_eocs(vals::Eocs::EACHCONVERSION);
+        });
+
+        //Being ADC conversions
```
andresv, line 301: Typo: Begin

Dirbaio (approved):
thank you!

andresv:
@sethkrie so if I have declared for example 4 sequences and my dma buffer size is 16 and then I read out 8 samples I get 
`seq0 seq1 seq2 seq3 seq0 seq1 seq2 seq3` right?

```rust
    let adc_dma_buf = [0u16; 16];
    let mut adc: RingBufferedAdc<embassy_stm32::peripherals::ADC1> = adc.into_ring_buffered(p.DMA2_CH0, adc_dma_buf);

    adc.set_sample_sequence(Sequence::One, &mut p.PA0, SampleTime::CYCLES112);
    adc.set_sample_sequence(Sequence::Two, &mut p.PA1, SampleTime::CYCLES112);
    adc.set_sample_sequence(Sequence::Two, &mut p.PA2, SampleTime::CYCLES112);
    adc.set_sample_sequence(Sequence::Two, &mut p.PA3, SampleTime::CYCLES112);
    
    let mut read_buf = [0u16; 8];
    let _ = adc.start();
    loop {
        match adc.read_exact(&mut read_buf).await {
            Ok(_data) => {
                let toc = Instant::now();
                info!(
                    "\n adc1: {} dt = {}, n = {}",
                    buffer1[0..16],
                    (toc - tic).as_micros(),
                    _data
                );
                tic = toc;
            }
            Err(e) => {
                warn!("Error: {:?}", e);
                read_buf = [0u16; 8];
                let _ = adc.start();
            }
        }
    }
```

If I would have read only 3 readings with one go I would have got `seq0 seq1 seq2` and then `seq3 seq0 seq1` right?

Is currently continous mode working as fast as it can and I cannot define like sample with 100 Hz?

sethkrie:
> @sethkrie if I have declared for example 4 sequences and my dma buffer size is 16 and then I read out 8 samples I get `seq0 seq1 seq2 seq3 seq0 seq1 seq2 seq3` right?
> 
> ```rust
>     let adc_dma_buf = [0u16; 16];
>     let mut adc: RingBufferedAdc<embassy_stm32::peripherals::ADC1> = adc.into_ring_buffered(p.DMA2_CH0, adc_dma_buf);
> 
>     adc.set_sample_sequence(Sequence::One, &mut p.PA0, SampleTime::CYCLES112);
>     adc.set_sample_sequence(Sequence::Two, &mut p.PA1, SampleTime::CYCLES112);
>     adc.set_sample_sequence(Sequence::Two, &mut p.PA2, SampleTime::CYCLES112);
>     adc.set_sample_sequence(Sequence::Two, &mut p.PA3, SampleTime::CYCLES112);
>     
>     let mut read_buf = [0u16; 8];
>     let _ = adc.start();
>     loop {
>         match adc.read_exact(&mut read_buf).await {
>             Ok(_data) => {
>                 let toc = Instant::now();
>                 info!(
>                     "\n adc1: {} dt = {}, n = {}",
>                     buffer1[0..16],
>                     (toc - tic).as_micros(),
>                     _data
>                 );
>                 tic = toc;
>             }
>             Err(e) => {
>                 warn!("Error: {:?}", e);
>                 read_buf = [0u16; 8];
>                 let _ = adc.start();
>             }
>         }
>     }
> ```
> 
> If I would have read only 3 readings with one go I would have got `seq0 seq1 seq2` and then `seq3 seq0 seq1` right?
> 
> Is currently continous mode working as fast as it can and I cannot define like sample with 100 Hz?

Yeah so you're right in that the ADC will always follow it's sequencing as configured, however,  if there is overrun, the context of 'where' you are in the sequence is lost. The driver will stop everything and start from scratch. This is slow and undesirable.

Since async is somewhat non-deterministic in when it polls a task, you can run into trouble here and you'll find that your buffer will always overrun.

So, the idea is that you want to create a buffer that is sufficiently large that your app will most certainly loop back around and wake the read task before it overruns.

For example, I'll give DMA a 512 length buffer and then after starting continuous reads, call read_exact() with a 256 length buffer. The task will sleep until the DMA peripheral flags an interrupt for either a half-transfer or full-transfer event. This means the read_exact() call will have 256 words to read since the previous call, unless one was not read in time, in which case, you'll have overrun.

So two takeaways:

- Use a large buffer (length N), ideally divisible by the number of channels in the sequence. This will prevent weird shifts in the buffer & give the executor enough time to do other stuff besides copying DMA data.
- Read with a buffer that is at the most N/2 in size.

Regarding sampling frequency -- the above issue comes down to the fact that, yes the ADC is running at full-tilt until the conversions/DMA are stopped, however, you may do a one-shot of the scanning continuous mode and use a TIM or EXTI interrupt to call .start() -> .read_exact() -> .stop() to sample your sequence very quickly, but doing so at an overall slower rate.

andresv:
Currently both `read` and `read_exact` are implemented, I wonder if there is an usecase for `read` at all. Like you most likely always want to use `read_exact`.

Also for using `read_exact` there are many footguns and therefore in depth knowledge is needed. Your summary should be somehow condensed and written to `read_exact` documentation.

sethkrie:
I think there's docs covering the space issues (N and N/2) and overrun, but not on the implications of how they interact with async. Note that my .read_exact() is just exposing the ReadableRingBuffef API, which is very well documented. There's definitely some caveats with this implementation though, so I'll open a PR that puts a condensed version of what I wrote above in the adc_ringbuff's read_exact().

One more question, what's the best way to engage more people in the community in conversations around improving the guard rails around this? Maybe someone can shed more light and I can do the leg work, just not sure how to best reach the right people.

Thanks!

andresv:
@sethkrie I added some docs: https://github.com/embassy-rs/embassy/pull/3143

PR #3117, opened 2024-06-25
RP network example fixes

9names:
#3115 pulled the git version of `reqwless` without a rev/tag restriction.
The latest release version of `reqwless` works fine with this example, so this PR switches to that.

Replaced the "guarenteed to be random" constant PRNG seed with entropy data from the Ring Oscillator peripheral.
It was a fun joke but I think it's run it's course.


9names:
Gimme a sec to fix the RNG import, was intending to do cargo fix but forgot

PR #3118, opened 2024-06-25
RP: add option to provide your own boot2

diondokter:
In a project I need a custom boot2 and embassy gets in the way.
This PR adds an option to let embassy not provide it.
This is an active choice a user has to make, so it doesn't get in the way of existing code and ease of setup.


Dirbaio (approved):
thanks!

PR #3123, opened 2024-06-26
stm32/timer: Remove generics and introduce type-safe timer taxonomy

honzasp:
This PR is a refactor of the timer functionality in embassy-stm32 that attempts to address two problems with the current implementation:

- Remove the peripheral types from timer drivers (similar to what we did for UART of I2C)
- Introduce type-safe taxonomy of timers that works with type-erased drivers. This makes sure that drivers (and users) don't attempt to use features that are not supported by the underlying timer peripheral. (Currently, Embassy just gives up on this, pretending that all timers are general-purpose 4-channel timers...).

The approach that I took is [described in the doc comment for the `timer` module](https://github.com/honzasp/embassy/blob/timer-taxonomy/embassy-stm32/src/timer/mod.rs). The foundation for type-safe timers is the `timer::raw::RawTimer` driver, which provides access only to the timer registers and their fields that are supported by the given timer.


honzasp:
Marked as draft because I need to rebase and do more thorough testing before I'm comfortable merging this.

honzasp:
The CI passes and I confirmed that the timer examples work on my STM32H743ZI Nucleo board, so this should be ready for review. (But we need to wait for https://github.com/embassy-rs/stm32-data/pull/495 to land first.)

honzasp:
@Dirbaio, do you think you will have time to look at this PR? :)

Dirbaio:
Hello,

PR #3125, opened 2024-06-27
docs: Fix mention of ExtiInput

neuschaefer:
The struct is called ExtiInput, not ExtiButton.


Dirbaio (approved):

neuschaefer:
@Dirbaio wow, that was quick! :)

PR #3127, opened 2024-06-28
Feat/dds example

shufps:
Adds DDS example. Generates a 1kHz sine wave with 15.625kHz PWM.

Example is demonstrating how to modify registers with the `pac` and how to use `#[interrupt]`.


Dirbaio (approved):
thanks!

PR #3128, opened 2024-06-28
STM32 ADC v3 and V4 DMA support

andresv:
This PR is on top of https://github.com/embassy-rs/embassy/pull/3106 and adds also DMA support to STM32 ADC V3. Original PR added it to V4.

# Needs
- [x]  Unified ADC `DMACFG` enum: https://github.com/embassy-rs/stm32-data/pull/497
- [x]  Unified ADC `CONT` field: https://github.com/embassy-rs/stm32-data/pull/498
- [x] https://github.com/embassy-rs/embassy/pull/3138

# Closes
* https://github.com/embassy-rs/embassy/pull/3106

# Test
Tested on STM32G081.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1182,6 +1182,7 @@ fn main() {
         (("adc", "ADC1"), quote!(crate::adc::RxDma)),
         (("adc", "ADC2"), quote!(crate::adc::RxDma)),
         (("adc", "ADC3"), quote!(crate::adc::RxDma)),
+        (("adc", "ADC4"), quote!(crate::adc::RxDma)),
         (("ucpd", "RX"), quote!(crate::ucpd::RxDma)),
         (("ucpd", "TX"), quote!(crate::ucpd::TxDma)),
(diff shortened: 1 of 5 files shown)


andresv:
It seems stm32 data needs some changes: https://ci.embassy.dev/jobs/8006878f541b.
I could workaround this with `#[cfg`, but it becomes too messy.

andresv:
@Dirbaio this is finished and can be reviewed.

lulf (approved):
LGTM. One minor thing: we seem to have most places following pattern of 'async fn foo()' and 'fn foo_blocking()', whereas this and fdcan is using 'fn foo()' and 'async fn foo_async()'. 

A quick glance suggests that in this case the non-async version is not actually blocking so I'm not sure what's the best name.

andresv:
There are also `rng` and `cordic` which using `async_` prefix.

andresv:
`read` is actually blocking it is calling `convert` which blocks, so I could rename:
```
read -> read_blocking()
read_async -> read()
```

lulf:
@andresv Sounds good to me :+1:

PR #3130, opened 2024-06-28
rp/i2c: add address flexibility and example

kkrolczyk:
Previous i2c examples are using either blocking Embassy API or e-h traits, this example uses Embassy pub API directly, also loosens the requirement for passing specifically u8 or u16 to address a device.


lulf (approved):
LGTM, thanks!

PR #3131, opened 2024-06-28
Less implicit panics

dflemstr:
Cherry-picked commits from #3129 that are strictly about removing implicit panic call-sites

### DIFF
--- a/embassy-stm32/src/rcc/mod.rs
+++ b/embassy-stm32/src/rcc/mod.rs
@@ -138,11 +138,17 @@ impl RccInfo {
     pub(crate) fn enable_and_reset_with_cs(&self, _cs: CriticalSection) {
         if self.refcount_idx_or_0xff != 0xff {
             let refcount_idx = self.refcount_idx_or_0xff as usize;
-            unsafe {
-                crate::_generated::REFCOUNTS[refcount_idx] += 1;
-            }
-            if unsafe { crate::_generated::REFCOUNTS[refcount_idx] } > 1 {
-                return;
+
+            // Use .get_mut instead of []-operator so that we control how bounds checks happen.
+            // Otherwise, core::fmt will be pulled in here in order to format the integer in the
+            // out-of-bounds error.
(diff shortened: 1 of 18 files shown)


Dirbaio (approved):
thanks!

Dirbaio:
the "Remove implicit bounds checking from rcc module" commit breaks fdcan tests. 

the reason is  `unsafe { FOO }.get_mut(0)` copies the entire `FOO` to a temporary, then .get_mut() gets a reference to that. while `unsafe { FOO.get_mut(0) }` does the right thing. see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=56d6c2809a3afeb12c81e1eb6bd1d803). Fun footgun!

embassy-stm32/src/rcc/mod.rs
```diff
@@ -196,11 +202,15 @@ impl RccInfo {
...
-            unsafe {
-                crate::_generated::REFCOUNTS[refcount_idx] -= 1;
-            }
-            if unsafe { crate::_generated::REFCOUNTS[refcount_idx] } > 0 {
-                return;
+
+            // Use .get_mut instead of []-operator so that we control how bounds checks happen.
+            // Otherwise, core::fmt will be pulled in here in order to format the integer in the
+            // out-of-bounds error.
+            if let Some(refcount) = unsafe { crate::_generated::REFCOUNTS }.get_mut(refcount_idx) {
```
Dirbaio, line 209: this is missing the `else`, we don't want to silently ignore wrong indexes.

dflemstr: Hm true, although this didn't show up in my profiling which is why I forgot to add it.  I guess the compiler can prove that the index is valid since `enable_and_reset_with_cs` succeeded, maybe? Or maybe it gets confused since mutations are happening outside `UnsafeCell`, which might be UB in this situation?  I'll restore the panic for sure though!

dflemstr:
Ha interesting footgun indeed!

PR #3132, opened 2024-06-28
wiznet: add version check to initialization sequence

Qix-:
This performs a read from the Wiznet chips to check their respective version registers, returning an error if they mismatch.

Do note that this is largely untested because I'm trying to debug an issue with a chip at the moment, but it was asked that this gets PR'd in anyway since currently the initialization sequence only performs writes and no reads.


Qix-:
Seems to be working! @Dirbaio does the CI have both the W5500 and the W5100s on it?

Dirbaio:
only W5100s.

Dirbaio (approved):
Tested working on a pico-w5500-evb.

thank you!

PR #3136, opened 2024-06-30
Add check for closed connection to `read_ready()`

### LINKED ISSUE #3135: `embedded_io_async::ReadReady` returns incorrect value for a `TcpSocket`
Note - I originally ran into this on embassy-net 0.4.0 where the problem is `read_ready()` returning "true" when a connection is open but has no data. In `main` the code has changed and I think there is a different issue, that `read_ready()` returns false for a closed connection with no data - I've updated the issue to reflect this.

## To replicate

1. Create a `socket: TcpSocket` and connect it to a server
2. Server doesn't send any data
3. Close the socket from the server end
4. Call `socket.read_ready()`
5. If `socket.read_ready()` returned true, call `socket.read()` with a 1-byte buffer, otherwise repeat from 4.

## Observed result

`socket.read_ready()` returns false, and we don't try to read data, code loops forever with a closed socket.

## Expected result

`socket.read_ready()` returns true, we call `socket.read()`, and get an EOF which allows us to see the closed connection and terminate the loop.

## Version

embassy-net, `main` branch

Running on an `esp32c3` using `esp-wifi` for the `TcpSocket`.

## Details

The docs for `read_ready()` say:

```rust
    /// Get whether the reader is ready for immediately reading.
    ///
    /// This usually means that there is either some bytes have been received and are buffered and ready to be read,
    /// or that the reader is at EOF.
    ///
    /// If this returns `true`, it's guaranteed that the next call to [`Read::read`] or [`BufRead::fill_buf`] will not block.
```

However the code in `tcp.rs` is:

```rust
    impl<'d> embedded_io_async::ReadReady for TcpSocket<'d> {
        fn read_ready(&mut self) -> Result<bool, Self::Error> {
            Ok(self.io.with(|s, _| s.can_recv()))
        }
    }
```

This will return false when the connection is closed and the buffer is empty. However this should be true, since the docs say "or that the reader is at EOF" - and in this case we could call `read()` and it would give an EOF and not block.

## Fix

I think what would be required would be to return `true` from `read_ready()` if either:

1. We have data in the receive buffer so `read()` will return this immediately, or
2. The connection is closed and so we `read()` will will return EOF immediately.

I had a look through the logic for `can_recv` and `may_recv` in `smoltcp` and I think `read_ready()` could be implemented as `s.can_recv() || !s.may_recv()`.

While looking at this I drew up a table of the possibilities - if I'm correct, all the functions only depend on whether the receive half of the connection is open, and whether the buffer contains data:

| Receive open | Buffer non-empty | may_recv | can_recv | read_ready |
| ------------ | ---------------- | -------- | -------- | ---------- |
| 0            | 0                | 0        | 0        | 1 (EOF)    |
| 0            | 1                | 1        | 1        | 1 (Buffer) |
| 1            | 0                | 1        | 0        | 0 (Block)  |
| 1            | 1                | 1        | 1        | 1 (Buffer) |

I think this shows `s.can_recv() || !s.may_recv()` should work. 

I also noticed that it looks like `can_recv()` in `smoltcp` could be shortened to `!self.rx_buffer.is_empty()` - I don't think the `!self.may_recv()` part is needed, but I think it should work fine anyway.

trepidacious:
Fix #3135

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -587,7 +587,7 @@ mod embedded_io_impls {
 
     impl<'d> embedded_io_async::ReadReady for TcpSocket<'d> {
         fn read_ready(&mut self) -> Result<bool, Self::Error> {
-            Ok(self.io.with(|s, _| s.can_recv()))
+            Ok(self.io.with(|s, _| s.can_recv() || !s.may_recv()))
         }
     }
 


Dirbaio (approved):
thanks!

PR #3126, opened 2024-06-28
Add support for the stm32 LTDC display peripheral

ninjasource:
This PR adds support for the LTDC display peripheral typically found on high-end graphics oriented stm32 mcus. The included example demonstrates high frame-rate animation on an stm32h735g-dk whilst running another blinky embassy task. The implementation was inspired by how the LTDC is used in the stm32h7xx-hal so if anyone spots any glaring stm32h7-only stuff please let me know. I decided not to use the standardized `embedded_display_controller` library because it was too "blocking focused". Therefore, at the cost of duplicating a few small graphics primitives, this implementation is standalone.

NOTE: The PR requires a small change to the `stm32-data` library for which there is an open [PR](https://github.com/embassy-rs/stm32-data/pull/496). Once this is in the `stm32-metapac` dependencies on `embassy-stm32` should be updated. For convenience, I have linked a pre-built metapac as a temporary measure.

PR #3138, opened 2024-07-01
update stm32-metapac

andresv:
Brings in some STM32 ADC register changes.


andresv:
[Error](https://ci.embassy.dev/jobs/bb053c35300f) comes from recently added LTDC stuff.
It seems `stm32h743bi` does not have such register at all.

Dirbaio:
maybe it works now that #3126  is merged.

PR #3139, opened 2024-07-01
feat(pwm): allow specifying OutputDrive for PWM channels

igiona:
HI everyone,

in an application I'm working on, I need a way to change the driving strength of the GPIO Pins.
Since I did find a `safe` way to do so on my side, I went implementing it here.

For the `SequencePwm` I extended the `Config` with the driving strength, whereas in the `SimplePwm` I added the function to adapt it similarly to the other parameters `set_max_duty`. I omitted the getters because personally I think that it's a "YAGNI".
I did not introduce a `SimpleConfig` because from the code it looks like not to be there intentionally, and I did not want to brake the API.

Happy to change any of this based on your feedback.

Cheers!

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -836,7 +836,7 @@ impl<'d, T: Instance> SimplePwm<'d, T> {
     #[inline(always)]
     pub fn set_ch0_drive(&self, drive: OutputDrive) {
         if let Some(pin) = &self.ch0 {
-            pin.conf().write(|w| w.drive().variant(convert_drive(drive)));
+            pin.conf().modify(|_, w| w.drive().variant(convert_drive(drive)));
         }
     }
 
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/pwm.rs
```diff
@@ -815,6 +831,38 @@ impl<'d, T: Instance> SimplePwm<'d, T> {
         let max_duty = self.max_duty() as u32;
         clk / max_duty
     }
+
+    /// Sets the PWM-Channel0 output drive strength
+    #[inline(always)]
+    pub fn set_ch0_drive(&self, drive: OutputDrive) {
+        if let Some(pin) = &self.ch0 {
+            pin.conf().write(|w| w.drive().variant(convert_drive(drive)));
```
Dirbaio, line 839: these should be .modify() so it leaves the other fields unchanged.

igiona: Oh of course thanks, I wrongly assumed `write()` doing the job for me, I'm a newbie in embedded-rust world as you can tell 😄

Dirbaio (approved):
thanks!

igiona:
@Dirbaio the CI failure looks quite unrelated... possibly a flaky test case?

Dirbaio:
probably

PR #3141, opened 2024-07-01
Update `ReadReady` and `WriteReady` implementations

### LINKED ISSUE #3140: `embedded_io_async::ReadReady` incorrect for `TcpReader`, `embedded_io_async::WriteReady` may be incorrect for `TcpWriter` and `TcpSocket`
Following on from #3135 I realised that while the PR fixed `TcpSocket`, `TcpReader` has the same code, so should have the same change.

This made me think I should also have a quick look at the `WriteReady` implementations, and I think they may have an issue too. For both `TcpSocket` and `TcpWriter` they call through to `s.may_send()`. The docs for this are:

```rust
    /// This function returns true if it's possible to send data and have it arrive
    /// to the remote endpoint. However, it does not make any guarantees about the state
    /// of the transmit buffer, and even if it returns true, [send](#method.send) may
    /// not be able to enqueue any octets.
```

I think this should be `can_send`, with docs:

```rust
    /// Check whether the transmit half of the full-duplex connection is open
    /// (see [may_send](#method.may_send)), and the transmit buffer is not full.
```

This seems to match up well with the docs for `WriteReady`:

```
Get whether the writer is ready for immediately writing.
This usually means that there is free space in the internal transmit buffer.
If this returns true, it's guaranteed that the next call to [Write::write] will not block.
```

trepidacious:
Fix #3140 

Update `ReadReady` for `TcpReader` to match implementation for `TcpSocket`

Update `WriteReady` implementations to use `can_recv()` rather than `may_recv()`, since this will check that the transmit buffer is not full.

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -603,7 +603,7 @@ mod embedded_io_impls {
 
     impl<'d> embedded_io_async::WriteReady for TcpSocket<'d> {
         fn write_ready(&mut self) -> Result<bool, Self::Error> {
-            Ok(self.io.with(|s, _| s.may_send()))
+            Ok(self.io.with(|s, _| s.can_send()))
         }
     }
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #3142, opened 2024-07-02
make some embassy-boot asserts const

lulf:
Uses a new feature in Rust 1.79 to get compile time asserts. It looks like it only works for assert! and not assert_eq!.


Dirbaio (approved):
wow, we have const asserts? til

PR #3144, opened 2024-07-02
Better panic message when peripheral clock is not enabled

Stupremee:
Instead of this misleading message:
image

You now get a much more verbose message indicating what is going wrong
image

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -475,13 +475,21 @@ fn main() {
     }
 
     impl<'a> ClockGen<'a> {
-        fn gen_clock(&mut self, name: &str) -> TokenStream {
+        fn gen_clock(&mut self, peripheral: &str, name: &str) -> TokenStream {
             let clock_name = format_ident!("{}", name.to_ascii_lowercase());
             self.clock_names.insert(name.to_ascii_lowercase());
-            quote!( unsafe { crate::rcc::get_freqs().#clock_name.unwrap() } )
+            quote!(unsafe {
+                unwrap!(
+                    crate::rcc::get_freqs().#clock_name,
+                    "peripheral '{}' is configured to use the '{}' clock, which is not running. \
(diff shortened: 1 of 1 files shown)


embassy-stm32/build.rs
```diff
@@ -475,13 +475,14 @@ fn main() {
     }
 
     impl<'a> ClockGen<'a> {
-        fn gen_clock(&mut self, name: &str) -> TokenStream {
+        fn gen_clock(&mut self, peripheral: &str, name: &str) -> TokenStream {
             let clock_name = format_ident!("{}", name.to_ascii_lowercase());
             self.clock_names.insert(name.to_ascii_lowercase());
-            quote!( unsafe { crate::rcc::get_freqs().#clock_name.unwrap() } )
+            let msg = format!("peripheral '{peripheral}' is configured to use the '{clock_name}' clock, which is not running. Either enable it in 'config.rcc' or change 'config.rcc.mux' to use another clock");
+            quote!( unsafe { crate::rcc::get_freqs().#clock_name.expect(#msg) } )
```
Dirbaio, line 482: i'm a bit worried these strings will get a bit bloaty. There's a few things we can do to mitigate it:

- `.expect()` triggers a standard rust panic, which will do std's string formatting even if using defmt. instead, if you do `unwrap!(some_value, "some_message")` it will use either std::fmt or defmt for the message (thanks to cfgs in fmt.rs), so with defmt it won't put the string in flash.
- when not using defmt, doing `format!` in build.rs means you will get N long strings in flash of the error message customized for each peripheral/clock. Instead, if you let the target do the formatting like `unwrap!(some_value, "peripheral {} blahblahblah clock {} blah check config blah", "I2C1", "hsi")` then there'll be only one copy of the long string and N copies of the small string.

Stupremee: Didn't think of that! it's fixed now

Dirbaio (approved):
thanks!

PR #3148, opened 2024-07-03
stm32 adc: introduce blocking_read

andresv:
Unifies STM32 ADC `read` function naming.
Blocking reads will be named to `blocking_read` and async reads will be just `read`.


andresv:
@lulf all green now.

lulf (approved):
LGTM, thanks!

PR #3151, opened 2024-07-04
[TR3FIR-217] Implemented PKA driver.

piotrkrygier-intent:
Work in progress, there is a problem with exponent. There is a method called "test" on Rsa struct, that tests functionality of modular exponentiation for smallest possible parameters. However, results read from RAM are invalid. This is still under investigation.

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -95,6 +95,8 @@ pub mod ltdc;
 pub mod opamp;
 #[cfg(octospi)]
 pub mod ospi;
+#[cfg(pka)]
+pub mod pka;
 #[cfg(quadspi)]
 pub mod qspi;
 #[cfg(rng)]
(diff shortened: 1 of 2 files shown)


Dirbaio:
Very nice!

Could you add hardware-in-the-loop (HIL) tests, similar to all the other crypto drivers? The steps should be something like:

- Add a `pka` feature, enable it for all chips that have PKA in [Cargo.toml](https://github.com/embassy-rs/embassy/blob/main/tests/stm32/Cargo.toml)
- Add tests [here](https://github.com/embassy-rs/embassy/tree/main/tests/stm32/src/bin), mark with `// required-features: pka`
- run `gen_test.py`

with the `trusted` label HIL tests will run every time you push to this PR. I can also give you a token so you can run the tests on the test farm with `cargo run` from your local machine for faster iteration, if interested please DM me on Matrix.

piotrkrygier-intent:
No. I'll need to rewrite this. Please delete this PR and wait for the updated version, sorry

PR #3152, opened 2024-07-04
add SAI example

Dicklessgreat:


kalkyl:
LGTM, i also tested this on hardware

PR #3154, opened 2024-07-06
Fix BDMA example for soundness

### LINKED ISSUE #2905: Update BDMA example for soundness
Just noting that this example is a little unsound, wrt uninitialized linker sections.

I think we should still merge it for now, but making a note so I can open an issue to fix it later to give a better sound example.

_Originally posted by @jamesmunns in https://github.com/embassy-rs/embassy/pull/2904#pullrequestreview-2038018593_

Dicklessgreat:
Please refer [here](https://github.com/embassy-rs/embassy/pull/2904#issuecomment-2133482150) to see why this change is necessary.
I couldn't test this example because I have a Daisy Seed board(it has STM32H750), but all the SPI6 pins are connected to the SDRAM, and I can't use them. I confirmed it can be build, and run. But not yet tested it can transmit or receive bytes.

I accidentally commit `Merge branch main...`(`e2aa5ef376d0137e92e1f033801cfcaf9084597a`) commit. This commit is not intended to be part of the submission, and I would like to request that they be excluded from the merge.
I can create a new pull request with only the intended changes.
Or you can review the pull request and merge only the necessary changes, excluding the unwanted one.

closes #2905

### DIFF
--- a/examples/stm32h7/src/bin/spi_bdma.rs
+++ b/examples/stm32h7/src/bin/spi_bdma.rs
@@ -10,18 +10,24 @@ use embassy_executor::Executor;
 use embassy_stm32::mode::Async;
 use embassy_stm32::time::mhz;
 use embassy_stm32::{spi, Config};
+use grounded::uninit::GroundedArrayCell;
 use heapless::String;
 use static_cell::StaticCell;
 use {defmt_rtt as _, panic_probe as _};
 
 // Defined in memory.x
 #[link_section = ".ram_d3"]
-static mut RAM_D3: [u8; 64 * 1024] = [0u8; 64 * 1024];
+static mut RAM_D3: GroundedArrayCell<u8, { 64 * 1024 }> = GroundedArrayCell::uninit();
 
 #[embassy_executor::task]
 async fn main_task(mut spi: spi::Spi<'static, Async>) {
-    let read_buffer = unsafe { &mut RAM_D3[0..128] };
-    let write_buffer = unsafe { &mut RAM_D3[128..256] };
+    let (read_buffer, write_buffer) = unsafe {
+        RAM_D3.initialize_all_copied(0);
+        (
+            RAM_D3.get_subslice_mut_unchecked(0, 128),
+            RAM_D3.get_subslice_mut_unchecked(128, 128),
+        )
+    };
 
     for n in 0u32.. {
         let mut write: String<128> = String::new();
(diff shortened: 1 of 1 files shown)


examples/stm32h7/src/bin/spi_bdma.rs
```diff
@@ -10,18 +10,24 @@ use embassy_executor::Executor;
...
 use embassy_stm32::{spi, Config};
+use grounded::uninit::GroundedArrayCell;
 use heapless::String;
 use static_cell::StaticCell;
 use {defmt_rtt as _, panic_probe as _};
 
 // Defined in memory.x
 #[link_section = ".ram_d3"]
-static mut RAM_D3: [u8; 64 * 1024] = [0u8; 64 * 1024];
+static mut RAM_D3: GroundedArrayCell<u8, { 64 * 1024 }> = GroundedArrayCell::uninit();
```
jamesmunns, line 20: Could this be just 256B since you only take 2x128 byte slices?

examples/stm32h7/src/bin/spi_bdma.rs
```diff
@@ -10,18 +10,24 @@ use embassy_executor::Executor;
 use embassy_stm32::mode::Async;
 use embassy_stm32::time::mhz;
 use embassy_stm32::{spi, Config};
+use grounded::uninit::GroundedArrayCell;
```
jamesmunns, line 13: Do you need to add `grounded` to the Cargo.toml as well?

jamesmunns (commented):
Some quick notes

examples/stm32h7/src/bin/spi_bdma.rs
jamesmunns: Ah! I didn't realize it was already there :D

https://github.com/embassy-rs/embassy/blob/main/examples/stm32h7/Cargo.toml#L37

Dicklessgreat: It has already been added:)
https://github.com/embassy-rs/embassy/commit/f91244bae6a4b277f8464c499e50c492a7671039

examples/stm32h7/src/bin/spi_bdma.rs
Dicklessgreat: I also think that.
I will fix it.

Dicklessgreat: It's now 256B :)

jamesmunns (approved):
Thanks!

PR #3157, opened 2024-07-08
Add example for sharing things between tasks

kalkyl:
We get a lot of questions about how to share things between tasks, so i thought an example would be a good idea, showcasing some common scenarios.

### DIFF
--- a/docs/pages/faq.adoc
+++ b/docs/pages/faq.adoc
@@ -352,7 +352,8 @@ There are two main ways to handle concurrency in Embassy:
 
 In general, either of these approaches will work. The main differences of these approaches are:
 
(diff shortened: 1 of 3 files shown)


jamesmunns:
Very happy to have this! Might make sense to do some cross linking with these docs pages for better discoverability:

* https://embassy.dev/book/#_sharing_peripherals_between_tasks
* https://embassy.dev/book/#_multiple_tasks_or_one_task_with_multiple_futures

jamesmunns (approved):
Happy to have this! Added some comments for potentially making this more discoverable.

It might be worth adding a top level doc-comment comparing and contrasting the demonstrated concurrency:

"We demonstrate three different ways of sharing in this example:"

* This is shared via a Cell, this is useful for ...
* This is shared via a RefCell, this is useful for ...
* This is shared via a Mutex, this is useful for ...

kalkyl:
> Happy to have this! Added some comments for potentially making this more discoverable.
> 
> It might be worth adding a top level doc-comment comparing and contrasting the demonstrated concurrency:
> 
> "We demonstrate three different ways of sharing in this example:"
> 
> * This is shared via a Cell, this is useful for ...
> * This is shared via a RefCell, this is useful for ...
> * This is shared via a Mutex, this is useful for ...

Good idea, added some more info and links!

jamesmunns:
Thanks, this looks great!

PR #3158, opened 2024-07-08
Stm dualcore better init

### LINKED ISSUE #3150: STM32 dual core HAL init design
## Problem statement
So embassy-stm32 was made for single-core processors. And so the HAL is made to always initialize things like DMA, GPIO and RCC.
However, the dual core H7 chips share all these peripherals and to use embassy on both cores, both cores must initialize them as if they are single core.

This can lead to problems.

I've got an idea on how to fix this and want some feedback before I implement it.

## Idea outline

So, to initialize the hal, you have to call the [init](https://docs.embassy.dev/embassy-stm32/git/stm32h755zi-cm7/fn.init.html) function and provide a config. From the looks of it there's nothing stopping you from calling it multiple times, which is likely not correct.

I propose we add a second init function for the dual core chips, perhaps called `init_second_core`.
That function would not do any init on its own.

There should be a global flag somewhere that the normal `init` function should then set to true when the init is done.
The `init_second_core` would be fallible and return an error as long as the normal `init` hasn't completed yet.
Bonus is that the normal `init` can also check the flag so it doesn't doesn't allow multiple initializations.

With this the user has the choice which core does the initialization and it's possible to let both cores boot at the same time or let one core boot the other. And say it's a security-critical application, if the designated init core doesn't init, then the other core can still do it as a backup.

### Problem 1

For the flag we do need a linker section so both cores agree on the location.
As long as this location is kept the same and the meaning of it stays the same, then two different versions of embassy can still interact like this.

But the initialization of the flag is a problem. It can't be initialized by both. But if it's left as a `MaybeUninit`, then the flag would still be set after a reboot. I don't know what the best way is to fix that.

### Problem 2

The non-init core also has to know the clock speeds that have been configured.
Currently they are stored in a global variable here: https://github.com/embassy-rs/embassy/blob/e2f9a48457c5d54546a3861fb067071ce93f9742/embassy-stm32/src/rcc/mod.rs#L51-L67

Naively that could also be put in a shared linker section with a known address. But then there are still some problems:

- The clocks struct is generated. So this would only work when they have struct generated. For the chip I'm looking at that's the case, but I'm not sure if we can count on that always being the case. ??
- It's a Rust ABI struct, so (while unlikely) the compiler could generate different layouts for different compiler versions. This can be solved by changing the generated struct to the C ABI (and do something about the options)

### Problem 3

This route uses two global shared variables. Can we actually do that?
Well yes, embassy-rp does something similar. But it forces the user to add some linker flag.
I'm not really sure how that's meant to be done, because I don't see any documentation about it.
So what the best way to do this is unclear to me...

diondokter:
Based on my thoughts in and closes #3150 

Difference is that this lets the user define a shared data region. I couldn't come up with a way to let embassy do it automatically.
This is a bit less user friendly than normal embassy, but I guess you shouldn't buy a dual core chip if you want an easy life.

This is built with the assumption that all stm dual cores share all peripherals that need to be initialized. I think that's true now, but who knows what the future will be. Also, when a 2+ core chip comes out this will need to be refactored.

Shared timer driver doesn't work out of the box, so at least for now the user has to specify different ones. This is true for all peripherals really and this PR doesn't fix that.

With the split init, at least there's the option to return different peripheral structs in the future.

I also introduced a `MaybeHertz`. I needed an `Option<Hertz>` that was repr C (so it can be shared across cores with potential different compiler versions). One pro of it is that it's only 4 bytes per instance instead of the 8 bytes per instance for `Option<Hertz>`.

For now posting this as a draft. It works for my H755.
Please provide feedback!

Things I'm unsure about:

- The names of things, please bikeshed away
- Is Hertz never 0? This is the assumption I've made now.
- Can core 2 be rebooted without core 1 also rebooting? If so, the secondary init can fail to ever initialize.
- I needed the `_dual-core` feature and added it to the dual core chips manually. Not sure if there's a better intended way.
- Atomics are used, but I'm not sure what the status quo for their usage is. I've used the std ones for now. There doesn't seem to be a crate for it in use in the HAL already that works everywhere.


diondokter:
I'll fix the CI after confirmation this route is what we want

liarokapisv:
The general approach makes sense to me, I will need to test on my H745 soon.

lulf (commented):
No objections from me, but could you perhaps write some instructions on how to use the example in a dual-core way, either as comments in the example or a section in the docs? I presume you need to flash a specific example to each core to have it work as intended, or maybe I'm misunderstanding how this works....

diondokter:
> No objections from me, but could you perhaps write some instructions on how to use the example in a dual-core way, either as comments in the example or a section in the docs? I presume you need to flash a specific example to each core to have it work as intended, or maybe I'm misunderstanding how this works....

Yeah, the CM7 example works on its own. The CM4 example needs the CM7 one to be flashed.
I'll write some docs.

What's the 'embassy' way of doing atomics? That's still going wrong in the build

diondokter:
@lulf oof, this was more work still than I expected, but everything passes CI now.

PR #3163, opened 2024-07-10
add TSC examples

michelrandahl:
- Examples of how to use blocking/polling TSC for stm32f303 and stm32l073.
- Example of how to use async TSC for stm32l073, inspired by https://github.com/embassy-rs/embassy/blob/main/examples/stm32u5/src/bin/tsc.rs.

I would have liked to add a corresponding async example for stm32f303, but I am having problems with registering the EXTI2_TSC interrupt with the usual `bind_interrupts!` macro, because I get an ldd conflict. My guess is that I get the conflict because of [this line in the exti.rs](https://github.com/embassy-rs/embassy/blob/6636a5835b27f82abea9000c9f3e93b4455b29c8/embassy-stm32/src/exti.rs#L312). I am quite new to Rust though, so it could be something else.

### DIFF
--- a/embassy-stm32/src/tsc/mod.rs
+++ b/embassy-stm32/src/tsc/mod.rs
@@ -40,7 +40,7 @@
 //! g7.set_io2(context.PE3, PinType::Sample);
 //! g7.set_io3(context.PE4, PinType::Channel);
 //!
-//! let mut touch_controller = tsc::Tsc::new(
+//! let mut touch_controller = tsc::Tsc::new_blocking(
 //!     context.TSC,
 //!     Some(g1),
 //!     Some(g2),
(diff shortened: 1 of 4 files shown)


Dirbaio:
thanks for the examples, looking great!

> problems with registering the EXTI2_TSC interrupt with the usual bind_interrupts! macro, because I get an ldd conflict. My guess is that I get the conflict because of [this line in the exti.rs](https://github.com/embassy-rs/embassy/blob/6636a5835b27f82abea9000c9f3e93b4455b29c8/embassy-stm32/src/exti.rs#L312). I am quite new to Rust though, so it could be something else.

oof yes. this is unfixable with the current design: embassy-stm32 assumes EXTI interrupts are used for EXTI only, so it unconditionally registers them. This means you can't also register it yourself to use it with TSC. 

It should work if you disable the `exti` Cargo feature so it doesn't register the irq, but that's not great because then you can't use exti. 

The ideal solution would be to change EXTI to use `bind_interrupts!`, so you can bind the same irq to both exti and tsc. That's a big refactor tho...

michelrandahl:
thank you @Dirbaio 

> The ideal solution would be to change EXTI to use bind_interrupts!, so you can bind the same irq to both exti and tsc. That's a big refactor tho...

Do you know if anyone is already working on a refac like that?
I am tempted to look into it. Even if I don't succeed and it becomes way over my head, -I will at least gain more embedded Rust experience... If I do look into it, though, it might be quite on and off in between other stuff I am currently working on.

Dirbaio:
no one's working on it, no.

PR #3169, opened 2024-07-11
Add adc oversampling and adc differential for g4

Adancurusul:
Add oversampling in regular mode .
Add adc differential use `pa0 `and `pa1`.
Verify no errors with `stm32g431cbu6`


Dirbaio (approved):
thanks!

PR #3171, opened 2024-07-12
nrf9151 support

nerwalt:
Adding support for the nRF9151 (and nRF9161).


nerwalt:
This is a draft PR.  It fails CI tests as it depends on the nrf9120-pac, which is not yet up-to-date.

embassy-nrf/src/gpiote.rs
```diff
@@ -117,6 +117,7 @@ fn GPIOTE() {
     unsafe { handle_gpiote_interrupt() };
 }
 
+#[cfg(not(any(feature = "_nrf9120")))]
```
Dirbaio, line 120: instead of cfging this out, can you wire up the interrupt like the nrf9160?

embassy-nrf/Cargo.toml
```diff
@@ -107,6 +117,7 @@ _nrf5340-app = ["_nrf5340", "nrf5340-app-pac"]
 _nrf5340-net = ["_nrf5340", "nrf5340-net-pac"]
 _nrf5340 = ["_gpio-p1", "_dppi"]
 _nrf9160 = ["nrf9160-pac", "_dppi"]
+_nrf9120 = ["nrf9120-pac", "_dppi"]
```
Dirbaio, line 120: I think if you define a `_nrf91` feature you could simplify a few of the cfg's in the crate. Like, "if nrf91" instead of "if nrf9160 or nrf9120"

Dirbaio:
could you add embassy-nrf with the new features, and the new examples, to `ci.sh` too? thanks!

embassy-nrf/Cargo.toml
nerwalt: I thought the same.  Let me look into that.  I have my 9160dk back so I can test against both.

nerwalt:
I added secure and non-secure examples for the nrf9151, which differs from what is done for the nrf9160.  Is this okay?

Dirbaio:
yes! that's great

Dirbaio (approved):
thank you! :rocket:

PR #3173, opened 2024-07-13
RP2040: Add DMA transfer abort function

JomerDev:
This PR adds an abort function to RP2040 DMA transfers, which also returns the amount of bytes transferred up until this point.
That can be useful in cases where a frame of dynamic length is read via PIO and the end of a frame is signaled via interrupt, so the code can stop the transfer and immediately know how long the frame was

### DIFF
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -173,7 +173,10 @@ impl<'a, C: Channel> Transfer<'a, C> {
     pub(crate) fn new(channel: impl Peripheral<P = C> + 'a, start_write_addr: u32) -> Self {
         into_ref!(channel);
 
-        Self { channel, start_write_addr }
+        Self {
+            channel,
+            start_write_addr,
+        }
     }
     /// Abort a DMA transfer early
     ///
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -159,36 +159,36 @@ fn copy_inner<'a, C: Channel>(
     });
 
     compiler_fence(Ordering::SeqCst);
-    Transfer::new(ch, to as u32)
+    Transfer::new(ch)
 }
 
 /// DMA transfer driver.
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub struct Transfer<'a, C: Channel> {
     channel: PeripheralRef<'a, C>,
-    start_write_addr: u32,
 }
 
 impl<'a, C: Channel> Transfer<'a, C> {
-    pub(crate) fn new(channel: impl Peripheral<P = C> + 'a, start_write_addr: u32) -> Self {
+    pub(crate) fn new(channel: impl Peripheral<P = C> + 'a) -> Self {
         into_ref!(channel);
 
         Self {
             channel,
-            start_write_addr,
         }
     }
     /// Abort a DMA transfer early
     ///
-    /// Returns the count of bytes transfered up until this point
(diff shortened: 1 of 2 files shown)


Dirbaio:
- this makes the `Transfer` struct bigger which will affect all users even if they don't want this feature. Maybe yo ucan make it return the amount of bytes left instead, which you can read from `trans_count` so you don't need to store anything extra. it's a bit less ergonomic bvecause the user has to subtract but should still work.
- abort should take `self` instead of `&mut self`, otherwise the dma transfer will be aborted again when dropped later.

JomerDev:
Fixed.
I've also changed abort to return the data size as well, because without it the transfer count isn't as useful

Dirbaio:
why? the user knows the data size already because they initiated it themselves.

JomerDev:
While that is technically true, I don't think many people make the connection that the size of the type they created the buffer with equals the data size that the dma transfers. Usually people will pick a `u8` representation for byte buffers

JomerDev:
@Dirbaio Would you prefer it if the abort method always returns the transfer count in bytes and the users can calculate the transfer count for their type if they want to?

Dirbaio:
the "natural" thing is to always use "number of words", where "word" is u8/u16/u32 depending on what the user passed in. For example, if you do a transfer with a `&[u16]` then `.len()` returns a number of u16 words, so it makes sense that `abort()` also returns number of u16 words. 

I wouldn't "helpfully" convert it to bytes for the user.

JomerDev:
Alright, I've changed it so only the transfer count is returned
I don't know why the build has now failed, I didn't change anything else

embassy-rp/src/dma.rs
```diff
@@ -174,6 +174,19 @@ impl<'a, C: Channel> Transfer<'a, C> {
 
         Self { channel }
     }
+    /// Abort a DMA transfer early
+    ///
+    /// Returns the count of transfers still left to do and the data size of the transfers
+    pub fn abort(self) -> u32 {
```
Dirbaio, line 180: one last nit, sorry :sweat_smile: . could you change this to return usize? so the user doesn't have to cast if they want to do index math or slice stuff.

Dirbaio:
to fix ci, can you rebase on latest main?

embassy-rp/src/dma.rs
Dirbaio: also docs need updating.

JomerDev: All done (rebase as well)

embassy-rp/src/dma.rs
```diff
@@ -174,6 +174,19 @@ impl<'a, C: Channel> Transfer<'a, C> {
 
         Self { channel }
     }
+    /// Abort a DMA transfer early
+    ///
+    /// Returns the count of transfers still left to do
+    pub fn abort(self) -> usize {
+        let p = self.channel.regs();
+        let transfer_count = p.trans_count().read();
```
Dirbaio, line 182: maybe you need to read trans_count after stopping? I think it's possible for some more transfers to go through between reading this and stopping?

JomerDev: No, afaik after the abort the transfer count will be reset to zero

Dirbaio: hmm i'm not sure how to do this then, but I'm pretty sure the current code has a race condition.

Dirbaio:
Hello,

PR #3174, opened 2024-07-13
RP2040: Add read_to_break_with_count

JomerDev:
Adds 'read_to_break_with_count' to the RP2040 uart implementation, an extension of 'read_to_break'.
This makes it possible to restart the read immediately after a break has been detected, even if the break is still going on when the read is restarted. Most uart framing where 'read_to_break' is used has some sort of header byte, which can be used to make sure that the read actually read until the next break and doesn't detect the break that was already detected


lulf (approved):
lgtm!

PR #3175, opened 2024-07-13
WIP: Extract Raw Mutex functionality

jamesmunns:
This is an attempt to extract some core functionality to a separate crate.

Background info: https://gist.github.com/jamesmunns/6052ba03e28d2013fc33b718a6416df0

This ALSO currently switches the blocking mutex over to providing a `&mut T`, e.g. making it no longer re-entrant


embassy-sync/src/blocking_mutex/mod.rs
```diff
@@ -3,109 +3,60 @@
...
+// Semver re-export
+pub use raw_mutex_traits::BlockingMutex as Mutex;
 
 /// A mutex that allows borrowing data across executors and interrupts.
 ///
 /// # Safety
 ///
 /// This mutex is safe to share between different executors and interrupts.
-pub type CriticalSectionMutex<T> = Mutex<raw::CriticalSectionRawMutex, T>;
+pub struct CriticalSectionMutex<T> {
```
Dirbaio, line 17: why change these from a type alias to a wrapper struct?

raw-mutex-traits/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// Use [`CriticalSectionMutex`] when data can be shared between threads and interrupts.
+///
+/// Use [`NoopMutex`] when data is only shared between tasks running on the same executor.
+///
+/// Use [`ThreadModeMutex`] when data is shared between tasks running on the same executor but you want a global singleton.
+///
+/// In all cases, the blocking mutex is intended to be short lived and not held across await points.
+/// Use the async [`Mutex`](crate::mutex::Mutex) if you need a lock that is held across await points.
+pub struct BlockingMutex<R, T: ?Sized> {
```
Dirbaio, line 49: perhaps we can do some bikeshedding on the name
- crate should be raw-mutex-trait*s* or raw-mutex-trait?  there's only one trait.
- perhaps "BlockingMutex" only makes sense in the context of embsasy-sync, and it could be just "Mutex" here?

embassy-sync/src/blocking_mutex/mod.rs
jamesmunns: We add some method definitions (namely `borrow`) in this file, since the `BlockingMutex` type is no longer defined in this crate, we can't add methods to it. So I made it a wrapper type so we could keep the same method defs on the named type

raw-mutex-traits/src/lib.rs
jamesmunns: I think making it clear that it is a blocking mutex is a good thing, based on embassy sync and maitake sync, which will each define their own async `Mutex` type.

Basically, I want to save the shorter `Mutex` name for the types we want USERS to reach for, while `BlockingMutex` is mostly for people building other things on top of it.

jamesmunns: re: crate name: yeah, idk. I didn't want to be `embassy-sync-stable-parts` or something, since I want to make it clear that the hope for this is to be useful to other non-embassy libraries, like maitake-sync. Bikeshedding welcome on the crate name tho.

jamesmunns: @Dirbaio I went with `scoped-mutex` as a name, since it fits the reality better!

jamesmunns:
bender run

jamesmunns:
One error, not sure if this is transient or not:

```
 ERROR teleprobe::client > === nucleo-stm32c031c6 out/tests/stm32c031c6/usart_rx_ringbuffered: FAILED: HTTP request failed with status code: 400: Bad Request
 ERROR teleprobe::client > WARN - Matching STM32C031C6Tc based on wildcard. Consider specifying the chip as STM32C031C6Tx instead.
INFO - run_from_ram: false
INFO - flashing program...
WARN - send_jtag_command 242 failed: SwdDpWait
...
WARN - send_jtag_command 242 failed: SwdApWait
WARN - got SwdDpWait/SwdApWait, retrying.
INFO - flashing done!
TRACE - 0.000000 BDCR ok: 00008200
DEBUG - 0.000000 rcc: Clocks { hclk1: Some(Hertz(48000000)), hsiker: Some(Hertz(16000000)), lse: None, pclk1: Some(Hertz(48000000)), pclk1_tim: Some(Hertz(48000000)), rtc: Some(Hertz(32000)), sys: Some(Hertz(48000000)) }
INFO - 0.000007 Hello World!
TRACE - 0.000053 USART: Kind::Uart
TRACE - 0.000091 USART: presc=1, div=0x00000060 (mantissa = 6, fraction = 0)
TRACE - 0.000198 Using 16 bit oversampling, desired baudrate: 500000, actual baudrate: 500000
TRACE - 0.000350 USART: m0: vals::M0::BIT8
TRACE - 0.000389 USART: set_ps: vals::Ps::EVEN
TRACE - 0.000427 USART: set_fifoen: true (usart_v4)
INFO - 0.000488 Spawning tasks
INFO - 0.000633 Ready to receive...
INFO - 0.100952 Starting random transmissions into void...
ERROR - 0.458541 panicked at 'Test fail! read error: Overrun'
ERROR - 0.458602 panicked at /ci/cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/defmt-0.3.8/src/lib.rs:367:5:
explicit panic
INFO -   R0: 00000000   R1: 20002e90   R2: 20001322   R3: 000003ff
INFO -   R4: 00000001   R5: 20002eec   R6: 000000b7   R7: 20002ea0
INFO -   R8: ffffffff   R9: ffffffff  R10: ffffffff  R11: ffffffff
INFO -  R12: ffffffff   SP: 20002e80   LR: fffffff9   PC: 0800473e
INFO - XPSR: 61000003
INFO - 
INFO - Stack:
INFO - 20002e80: 00000000 20002e90 20001322 000003ff
INFO - 20002e90: ffffffff 0800420d 08004224 61000000
INFO - 20002ea0: 20002ea8 08004205 20002ee0 080041e1
INFO - 20002eb0: 20002eec 20002eb0 0800514c 08004c64
INFO - 20002ec0: 00000001 20002ed4 00000001 00000000
INFO - 20002ed0: 20002f24 20002eb4 08002fb5 080030c9
INFO - 20002ee0: 20002f00 0800238d 20000018 00000001
INFO - 20002ef0: 08004a20 20002f08 08004cc4 08000001
INFO - 
INFO - Backtrace:
WARN - UNWIND: Error while checking for exception context. The stack trace will not include the calling frames. : UNWIND: Error reading FrameDescriptorEntry at PC=536882856 : Do not have unwind info for the given address.
INFO - Frame 0: UNWIND: Error while checking for exception context. The stack trace will not include the calling frames. : UNWIND: Error reading FrameDescriptorEntry at PC=536882856 : Do not have unwind info for the given address. @ 0x0800473e
INFO - Hard Fault!
ERROR - Run failed: Firmware crashed
```

jamesmunns:
bender run

jamesmunns:
Test fails consistently, so it's possible I screwed something up with the timers

jamesmunns:
So @Dirbaio,

```
# Main
    Finished `release` profile [optimized + debuginfo] target(s) in 2.67s
   text	   data	    bss	    dec	    hex	filename
  20520	    112	   5508	  26140	   661c	usart_rx_ringbuffered

# My branch
    Finished `release` profile [optimized + debuginfo] target(s) in 2.67s
   text	   data	    bss	    dec	    hex	filename
  21000	    120	   5524	  26644	   6814	usart_rx_ringbuffered
```

But even before, 112 + 5508 is 5620, and we have 3x256 byte buffers in the `usart_rx_ringbuffered` test, which means we are using AT LEAST 6388 bytes of RAM, and if the C0 has 6K of RAM, then we are using 6388/6144 bytes of RAM :p

jamesmunns:
Oh, maybe it's not that:

* all the buffers are either static or in tasks, so it hopefully all ends up in statics?
* shrinking the buffers to 64 bytes didn't make the test pass :(

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// RawMutex implementations must ensure that, while locked, no other thread can lock
+/// the RawMutex concurrently.
+///
+/// Unsafe code is allowed to rely on this fact, so incorrect implementations will cause undefined behavior.
+pub unsafe trait RawMutex {
+    /// Create a new `RawMutex` instance.
+    ///
+    /// This is a const instead of a method to allow creating instances in const context.
+    const INIT: Self;
```
hawkw, line 29: This is a hard blocker for `maitake-sync` , unfortunately. In `maitake-sync`, I use `loom` for testing, and `loom`'s simulated mutex cannot be instantiated in a `const` context, because it must be bound to the _current iteration_ of the Loom test at runtime. I need to be able to implement the trait that has the core `Mutex` functionality for the `loom::sync::Mutex` type, so it can't require a const initializer.

Instead, I think the best approach would be to factor this out into its own trait, like this:
```rust
pub unsafe trait RawMutex {
    /// Lock this `RawMutex`.
    fn lock<R>(&self, f: impl FnOnce() -> R) -> R;
}

pub trait ConstInit {
    const INIT: Self;
}
```

This way, `Mutex` wrapper types like `BlockingMutex` can have `const fn` constructors`where R: RawMutex + ConstInit`, but there can still be a way to construct them with non-const raw mutices (possibly `Default`?).

(as a side note, my preference would be to _not_ make the `ConstInit` trait have a `Self: RawMutex` bound, as I'd like to be able to reuse it for my `RawRwLock` trait, but if it needs to, I can live with that)

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+/// of interior mutability fits your use case.
+///
+/// Use [`CriticalSectionMutex`] when data can be shared between threads and interrupts.
+///
+/// Use [`NoopMutex`] when data is only shared between tasks running on the same executor.
+///
+/// Use [`ThreadModeMutex`] when data is shared between tasks running on the same executor but you want a global singleton.
+///
+/// In all cases, the blocking mutex is intended to be short lived and not held across await points.
+/// Use the async [`Mutex`](crate::mutex::Mutex) if you need a lock that is held across await points.
```
hawkw, line 49: I assume this was copied from docs in `embassy-sync`, since this crate doesn't have an async mutex?
```suggestion
```

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// This is a const instead of a method to allow creating instances in const context.
+    const INIT: Self;
+
+    /// Lock this `RawMutex`.
+    fn lock<R>(&self, f: impl FnOnce() -> R) -> R;
+}
+
+/// Blocking mutex (not async)
+///
+/// Provides a blocking mutual exclusion primitive backed by an implementation of [`raw::RawMutex`].
```
hawkw, line 37: i assume the docs here were copied from `embassy-sync`, since this module path is no longer correct?

```suggestion
/// Provides a blocking mutual exclusion primitive backed by an implementation of [`RawMutex`].
```

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// Which implementation you select depends on the context in which you're using the mutex, and you can choose which kind
+/// of interior mutability fits your use case.
+///
+/// Use [`CriticalSectionMutex`] when data can be shared between threads and interrupts.
+///
+/// Use [`NoopMutex`] when data is only shared between tasks running on the same executor.
+///
+/// Use [`ThreadModeMutex`] when data is shared between tasks running on the same executor but you want a global singleton.
```
hawkw, line 46: these types are also no longer defined in this crate.

hawkw (changes requested):

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// Unsafe code is allowed to rely on this fact, so incorrect implementations will cause undefined behavior.
+pub unsafe trait RawMutex {
+    /// Create a new `RawMutex` instance.
+    ///
+    /// This is a const instead of a method to allow creating instances in const context.
+    const INIT: Self;
+
+    /// Lock this `RawMutex`.
+    fn lock<R>(&self, f: impl FnOnce() -> R) -> R;
```
hawkw, line 32: Oh, one other thing: I'd really like a 
```suggestion
    fn lock<R>(&self, f: impl FnOnce() -> R) -> R;
    
    /// Try to lock this `RawMutex`, without blocking.
    ///
    /// If the lock can be acquired without blocking, this method calls `f` and returns the
    /// result; otherwise, it returns `None`.
    fn try_lock<R>(&self, f: impl FnOnce() -> R) -> Option<R>;
```

I generally use `try_lock` methods in the `fmt::Debug` implementations for my synchronization primitives in `maitake-sync`, and it would be nice to be able to implement `Debug` with scoped raw mutex impls...

hawkw: One other note: what do you think about calling this method either `with_lock` or `locked`, instead of `lock`? Again, I'm mainly thinking about the case of types that provide both scoped _and_ RAII APIs: in that case, the methods on the Mutex type will need to have different names, and I'd like to be able to reserve `Mutex::lock` for a RAII API for consistency with `std`. It would be nice for the mutex type methods to have the same name as the raw mutex API.

In my current `maitake-sync` branch, I called this `with_lock`, but thinking about it some more, I'm starting to wonder if I prefer `locked`. The rationale is that I also have _constructors_ that include `with_` in their names (`Mutex::with_raw_mutex` etc) and it seems slightly unfortunate to use the same prefix in a constructor and a method...though on the other hand, I could change the constructors to `new_with_...`. What do you think?

jamesmunns:
@hawkw implemented most of your suggestions, probably still some thinking to be done on naming wrt whether all the types should have `ScopedRaw`/`BlockingScoped`/etc., as well as `lock()`.

jamesmunns:
bender run

jamesmunns:
Now a bunch of embassy-rp tests are failing 💀

scoped-mutex/src/lib.rs
```diff
@@ -1,57 +1,100 @@
...
+///
+/// This is actually a marker trait for types that implement [`UnconstRawMutex`] and
+/// [`ConstInit`]. This is to allow cases where a mutex cannot be created in const
+/// context, for example some runtime/OS mutexes, as well as testing mutexes like
+/// those from `loom`.
+///
+/// If you are implementing your own RawMutex primitive, you should implement the
+/// [`UnconstRawMutex`] and [`ConstInit`] traits, and rely on the blanket impl
+/// of `impl<T: UnconstRawMutex + ConstInit> RawMutex for T {}`.
+pub trait RawMutex: UnconstRawMutex + ConstInit {}
```
hawkw: not totally sure if i think this alias is pulling its weight, personally --- is it really that much easier for implementations to have `T: RawMutex` bounds instead of just `T: UnconstRawMutex + ConstInit`? IMO it doesn't seem super necessary, and i find the name `UnconstRawMutex` kinda unfortunate...

also, it kind of seems like having this will encourage implementations to just bound their raw mutex with `T: RawMutex` and _also_ require the `ConstInit` trait for _all_ their methods, when they don't actually need it for anything except for const constructors. i kind of feel like just having "UnconstRawMutex" be the "RawMutex" trait would make it more likely for people to put the const constructor in a separate `where T: ConstInit` impl block, instead of requiring it for everything when they don't really need to.

i'm open to being convinced otherwise, if you think this is a major advantage for users.

scoped-mutex/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[must_use]
+    fn try_lock<R>(&self, f: impl FnOnce() -> R) -> Option<R>;
+
+    /// Lock this `ScopedRawMutex`, calling `f()` after the lock has been acquired, and releasing
+    /// the lock after the completion of `f()`.
+    ///
+    /// Panics if the lock is already locked.
+    fn lock<R>(&self, f: impl FnOnce() -> R) -> R {
+        self.try_lock(f).expect("Attempted to take lock re-entrantly")
+    }
```
hawkw, line 74: I don't think this is correct. This method should _block_ if the lock is already locked, instead.  What you've got here is a `RefCell`, not a blocking mutex.

```suggestion
    /// Lock this `ScopedRawMutex`, calling `f()` after the lock has been acquired, and releasing
    /// the lock after the completion of `f()`.
    ///
    /// If the lock is held by another thread or CPU core, this method will block
    /// until the lock is released, and then acquire the lock. The mechanism used
    /// to block depends on the implementation of this trait.
    ///
    /// To attempt to lock the mutex *without* blocking, use
    /// [`ScopedRawMutex::try_lock`].
    fn lock<R>(&self, f: impl FnOnce() -> R) -> R;
```

I also don't really think we should have a default impl for this or not, since it's the _way_ we block while waiting for the lock to become available that's the main thing implementations will be providing. Checking if the lock is locked or not is basically always going to be the same except in the totally-no-op case. If we were going to provide one, though, it should probably look more like:
```rust 
fn lock<R>(&self, f: impl FnOnce() -> R) -> R{
    // work around the fact that `f` is a `FnOnce`; we're only ever going to
    // call it once, but we need to pass it into `try_lock` multiple times regardless.
    let mut f = Some(f);
    loop {
       if let Some(r) = self.try_lock(|| (f.take().unwrap())()) {
           return r;
       }
       core::hint::spin_loop();
    }
}
```
but, this isn't even the most efficient implementation of a spinlock, which would instead probably do something more like [this](https://github.com/hawkw/mycelium/blob/c4992526f87f5c38813e6671baf99089fa24d7f0/maitake-sync/src/spin/mutex.rs#L106-L114), using `Relaxed` loads while waiting for the lock to be unlocked, and backing off in the spin loop.

So, i'm not sure if it's a good idea to provide a default implementation here.

hawkw (changes requested):
I don't think that the current `ScopedRawMutex::lock` method does the right thing on multi-core systems --- it's a `RefCell`, not a `Mutex`. I left some notes about that.

I'm also not super convinced that it's worth having a trait alias for `ScopedRawMutex + ConstInit`, but that one isn't a big deal, and if others like it, it's fine with me.

jamesmunns:
So, CC @Dirbaio and @hawkw:

There's a couple things going on here, but I'm tempted to close this PR. The two main blockers:

* I need to iterate on the `scoped-mutex` API a bit. Eliza makes some good points, but I need to figure out how they mesh with how embassy expects to use this, and *non* threaded targets can implement it. In particular, we DO want to panic if the mutex attempts to re-lock on non-std targets (because we know it means deadlock), but we DON'T on std targets.
* I still have some CI tests failing, and I am somewhat less confident that things will Just Work using non-reentrant scoped locks in embassy. I could really use some discussion with @Dirbaio who has implemented a lot of the time driver impls - particularly for cases like the alarm callbacks which could potentially re-call time-driver methods, which could lead to deadlock.

I won't delete the branch, I think it will be useful in the future if we take this up again, but I think my plan is:

* Work with @hawkw to release a `scoped-mutex` crate, make sure it works for maitake-sync, and ensure it has most of the pieces that it does as of this PR.
* When I (or @Dirbaio) has some more time to discuss and iterate on this, take another stab at moving embassy-sync over to `scoped-mutex`.

hawkw:
@jamesmunns:

> * I need to iterate on the `scoped-mutex` API a bit. Eliza makes some good points, but I need to figure out how they mesh with how embassy expects to use this, and _non_ threaded targets can implement it. In particular, we DO want to panic if the mutex attempts to re-lock on non-std targets (because we know it means deadlock), but we DON'T on std targets.

Correct me if I'm wrong here, but wasn't this kind of the whole point of having a mutex trait? The behavior when already locked can be different for each category of target:

- Implementations for **single-core, no-std** targets can panic instead of blocking
- Implementations for **multi-core, no-std** targets can block by spinning (possibly with backoff, etc)
- Implementations for **std** targets can "just use OS APIs"

I think that the current trait *does* allow all those implementations, and the issue I called out in https://github.com/embassy-rs/embassy/pull/3175#discussion_r1685771688 was just that the documentation and the default impl assumed the single-core behavior, and they probably shouldn't. If we removed the default impl and changed the docs to say that the method should block if there are multiple cores, I think the trait as written is basically fine.

> I won't delete the branch, I think it will be useful in the future if we take this up again, but I think my plan is:
> 
>     * Work with @hawkw to release a `scoped-mutex` crate, make sure it works for maitake-sync, and ensure it has most of the pieces that it does as of this PR.
> 
>     * When I (or @Dirbaio) has some more time to discuss and iterate on this, take another stab at moving embassy-sync over to `scoped-mutex`.

I think that makes sense. It seems like this PR is trying to do two things at the same time:

1. Introduce a raw-mutex abstraction so that `embassy` can use `maitake-sync`'s async primitives but override the default spinlock with a single-core critical-section mutex
2. At the same time, make all of `embassy`'s mutices use that abstraction too.

I don't actually think we *need* to do both of those. Or, at least, we don't need to make the first one depend on the second one. If we just get `maitake-sync` generic over a mutex trait, we can then start adopting `maitake-sync`'s async primitives in Embassy. When `embassy` reaches for a *blocking* mutex, it can continue using its own reentrant mutex without issue. Then, we can look at all the mutices in `embassy`, find the ones where reentrancy is actually needed, and replace the ones that *don't* use reentrancy with the shared abstraction...if we even want to. But that work can be decoupled from the work required to unblock using `maitake-sync`'s async stuff.

One option for now is keep working on a `ScopedRawMutex` trait in `maitake-sync`, which I already have a branch for. I can release a version of the crate with that, which should be sufficient to allow `embassy` to use `maitake-sync` async primitives. If we eventually end up with a crate that _just_ provides a `ScopedRawMutex` trait, I can pretty easily add an implementation of `maitake-sync`'s trait for everything that implements the new trait.

Alternatively, I would be happy to help get a scoped raw mutex crate up first, and make `maitake-sync` use that. The question is whether having a common abstraction is really worth the effort, or if just having _an interface_ that allows embassy to use `maitake-sync` is sufficient?

PR #3185, opened 2024-07-17
(embassy-rp): Add split_ref fn to uart, allowing a mutable reference split into RX & TX handles

MathiasKoch:
- Add `split_ref` fn to `Uart` & `BufferedUart`, allowing a mutable reference split into RX & TX handles
- Change order of RX and TX handles in `BufferedUart::split()` fn, to streamline with other HALs & `Uart`
- Add `set_baudrate` to `BufferedUart` allowing runtime change of configured baudrate

### DIFF
--- a/embassy-rp/src/uart/buffered.rs
+++ b/embassy-rp/src/uart/buffered.rs
@@ -163,9 +163,21 @@ impl<'d, T: Instance> BufferedUart<'d, T> {
         self.tx.send_break(bits).await
     }
 
+    /// sets baudrate on runtime
+    pub fn set_baudrate(&mut self, baudrate: u32) {
+        super::Uart::<'d, T, Async>::set_baudrate_inner(baudrate);
+    }
+
     /// Split into separate RX and TX handles.
-    pub fn split(self) -> (BufferedUartRx<'d, T>, BufferedUartTx<'d, T>) {
-        (self.rx, self.tx)
+    pub fn split(self) -> (BufferedUartTx<'d, T>, BufferedUartRx<'d, T>) {
+        (self.tx, self.rx)
+    }
+
+    /// Split the Uart into a transmitter and receiver by mutable reference,
+    /// which is particularly useful when having two tasks correlating to
+    /// transmitting and receiving.
+    pub fn split_ref(&mut self) -> (&mut BufferedUartTx<'d, T>, &mut BufferedUartRx<'d, T>) {
+        (&mut self.tx, &mut self.rx)
     }
 }
 
--- a/embassy-rp/src/uart/mod.rs
+++ b/embassy-rp/src/uart/mod.rs
@@ -860,6 +860,56 @@ impl<'d, T: Instance + 'd, M: Mode> Uart<'d, T, M> {
         });
     }
 
+    fn lcr_modify<R>(f: impl FnOnce(&mut rp_pac::uart::regs::UartlcrH) -> R) -> R {
+        let r = T::regs();
+
+        // Notes from PL011 reference manual:
+        //
+        // - Before writing the LCR, if the UART is enabled it needs to be
+        //   disabled and any current TX + RX activity has to be completed
+        //
+        // - There is a BUSY flag which waits for the current TX char, but this is
+        //   OR'd with TX FIFO !FULL, so not usable when FIFOs are enabled and
+        //   potentially nonempty
+        //
+        // - FIFOs can't be set to disabled whilst a character is in progress
+        //   (else "FIFO integrity is not guaranteed")
+        //
+        // Combination of these means there is no general way to halt and poll for
+        // end of TX character, if FIFOs may be enabled. Either way, there is no
+        // way to poll for end of RX character.
+        //
+        // So, insert a 15 Baud period delay before changing the settings.
+        // 15 Baud is comfortably higher than start + max data + parity + stop.
+        // Anything else would require API changes to permit a non-enabled UART
+        // state after init() where settings can be changed safely.
+        let clk_base = crate::clocks::clk_peri_freq();
+
+        let cr = r.uartcr().read();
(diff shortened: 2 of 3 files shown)


embassy-rp/src/uart/buffered.rs
```diff
@@ -163,9 +163,21 @@ impl<'d, T: Instance> BufferedUart<'d, T> {
...
-    pub fn split(self) -> (BufferedUartRx<'d, T>, BufferedUartTx<'d, T>) {
-        (self.rx, self.tx)
+    pub fn split(self) -> (BufferedUartTx<'d, T>, BufferedUartRx<'d, T>) {
+        (self.tx, self.rx)
+    }
+
+    /// Split the Uart into a transmitter and receiver by mutable reference,
+    /// which is particularly useful when having two tasks correlating to
+    /// transmitting and receiving.
+    pub fn split_ref(&mut self) -> (&mut BufferedUartTx<'d, T>, &mut BufferedUartRx<'d, T>) {
```
jamesmunns, line 179: Just a check: I think you can usually create the UART with `&mut UART` instead of an owned UART, which then would allow you to split with the two owned RX/TX halves using the borrowed UART.

Is that not necessarily a match for you? You mention using this for two tasks, but tasks usually require `'static`/owned items?

MathiasKoch: Uhh, I didn't think about doing `(&mut my_uart).split()`.. Would that essentially give the same result as this?  

I essentially need it to fit into a signature that looks like this in a trait abstraction:
```
trait Abstraction {
    fn some_fn(&mut self) -> (impl Write, impl Read + BufRead);`
}
```

MathiasKoch: In my case I have an initialization part where I need the two halves split into rx and tx, and after initialization I need to pass the combined `impl Write + BufRead` to some upstream library.

jamesmunns: I think it would - that's generally why other HALs don't have a separate `split_ref` function - all the peripherals take something like `impl PeripheralRef`, which is implemented for both an owned and mutably borrowed peripheral.

Worth trying, I think it would work, but would be worth double checking :)

jamesmunns: It also might be something like:

```rust
let mut buart = BufferedUart::new(&mut uart, ...);
let (tx, rx) = buart.split();
```

instead of:

```rust
let mut buart = BufferedUart::new(uart, ...);
let (tx, rx) = (&mut buart).split();
```

But again - I'm not certain on this.

MathiasKoch: Yeah, my issue here is that I am working primarily in driver land. Let me give an example, and we can see if we can work it out together:

Driver (library) crate:
```rust
pub trait Transport: Write + Read + BufRead {
    fn set_baudrate(&mut self, baudrate: u32);
    fn split_ref(&mut self) -> (impl Write, impl Read + BufRead);
}

async fn run(mut transport: impl Transport) -> ! {

    // `AtClient` comes from an upstream library. 
    // Signature: `fn new<RW: Read + Write>(rw: RW) -> Self`
    let client = AtClient::new(&mut transport);
    client.init().await;
    
    // <snip>

    let (mut tx, mut rx) = transport.split_ref();
    
    // `Mux` is an upstream crate, simplified. 
    // Signature: `async fn run<R: BufRead, W: Write>(mut port_r: R, mut port_w: W) -> !`
    Mux::run(&mut rx, &mut tx).await
}
```

In my bin crate:
```rust

pub struct MyTransport(BufferedUart<....>);

impl Foo::Transport for MyTransport {
    fn set_baudrate(&mut self, baudrate: u32) {
        self.0.set_baudrate(baudrate)
    }

    fn split_ref(&mut self) -> (impl embedded_io_async::Write, impl embedded_io_async::Read) {
        self.0.split_ref()
    }
}

impl embedded_io_async::ErrorType for MyTransport {
    type Error = uart::Error;
}

impl Read for MyTransport {
    async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
        self.0.read(buf).await
    }
}

impl BufRead for MyTransport {
    async fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
        self.0.fill_buf().await
    }

    fn consume(&mut self, amt: usize) {
        self.0.consume(amt)
    }
}

impl Write for MyTransport {
    async fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
        self.0.write(buf).await
    }
}


async fn main() {
    let buart = BufferedUart::new(uart, ...);
    let mut transport = MyTransport(buart);
    
    // Technically it would be possible to change this to `&mut transport`, 
    // if i only implemented `Transport for &mut MyTransport`, 
    // but given that the trait is supposed to be implemented by users, it seems quite fragile?
    Foo::run(transport).await 
}
```

Am I overlooking anything obvious here?

jamesmunns: Not totally certain, but there is some prior art for `impl<T: Trait> Trait for &mut T`:

https://docs.rs/embedded-io-async/latest/src/embedded_io_async/lib.rs.html#166-200

Dirbaio: I think with the `Uart::new(&mut p.UART0)` thing it's possible to implement `split_ref`, but the caveat is you have to reinitialize the uart every time, which might not be what you want.

IMO it's okay to have both "borrowed split" and "create uart driver with borrowed peripheral". there's some overlap between them but not 100%.

jamesmunns: Ah, I think that makes sense, I missed that part, thanks for the explanation @Dirbaio and sorry for missing the point of what you were getting at @MathiasKoch :)

PR #3188, opened 2024-07-17
embassy-stm32: Allow cmd block to be passed in for sdmmc dma transfers

ninjasource:
The sdmmc module currently allows the user to pass in a data_block buffer to perform DMA data transfers. However, cmd_block buffers are created on the stack internally. This poses a problem if the stack RAM does not work with DMA. An example of this is the stm32h735 mcu which can have RAM as ITCM memory (0x20000000) and AXISRAM for use with DMA (0x24000000).

This change allows the user to pass in a statically allocated buffer in a location of their choosing when using the sdmmc module. There are two new functions that mirror existing functions so as to ensure backwards compatibility:

```rust
    /// Initializes card (if present) and sets the bus at the specified frequency.
    pub async fn init_card(&mut self, freq: Hertz) -> Result<(), Error> {
        let mut cmd_buffer = CmdBlock::new();
        self.init_card_with_cmd_buffer(freq, &mut cmd_buffer).await
    }

    /// Initializes card (if present) and sets the bus at the specified frequency.
    /// A cmd_buffer should be passed in if the DMA requirements mean that
    /// stack memory can't be used (the default in `init_card`).
    /// This usually manifests itself as an indefinite wait on a dma transfer because the
    /// dma peripheral cannot access the memory.
    pub async fn init_card_with_cmd_buffer(&mut self, freq: Hertz, cmd_buffer: &mut CmdBlock) -> Result<(), Error> {
      // code here
    }

    /// Write a data block.
    pub async fn write_block(&mut self, block_idx: u32, buffer: &DataBlock) -> Result<(), Error> {
        let mut cmd_buffer = CmdBlock::new();
        self.write_block_with_cmd_buffer(block_idx, buffer, &mut cmd_buffer)
            .await
    }

    /// Write a data block and pass in a cmd buffer rather than using a stack allocated one.
    /// This is required if stack RAM cannot be used with DMA
    pub async fn write_block_with_cmd_buffer(
        &mut self,
        block_idx: u32,
        buffer: &DataBlock,
        cmd_buffer: &mut CmdBlock,
    ) -> Result<(), Error> {
      // code here
    }


```

### DIFF
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -320,6 +320,10 @@ pub struct Sdmmc<'d, T: Instance, Dma: SdmmcDma<T> = NoDma> {
     signalling: Signalling,
     /// Card
     card: Option<Card>,
+
+    /// An optional buffer to be used for commands
+    /// This should be used if there are special memory location requirements for dma
+    cmd_block: Option<&'d mut CmdBlock>,
 }
 
 const CLK_AF: AfType = AfType::output(OutputType::PushPull, Speed::VeryHigh);
@@ -523,6 +527,7 @@ impl<'d, T: Instance, Dma: SdmmcDma<T> + 'd> Sdmmc<'d, T, Dma> {
             clock: SD_INIT_FREQ,
             signalling: Default::default(),
             card: None,
+            cmd_block: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Dirbaio:
thanks for the PR!

I think adding `_with_cmd_buffer` variants for every method might be a bit complex and verbose for the end user. An alternative would be taking the buffer in `new_with_cmd_buffer`, and make `new` default to using some static mut buffer specific to the sdmmc instance?

ninjasource:
Thanks for the idea, that would be better. One small question: There are currently 4 constructors (new_1bit v1, new_1bit v2, new_4bit v1 and new_4bit v2). I would double that to 8 by adding this `cmd_block` input parameter. Alternatively I could add a function called `set_cmd_block` that the user can use to set it. I have added the function for now but can use the constructor option you mentioned easily enough.

> "and make new default to using some static mut buffer specific to the sdmmc instance"
This is not necessary as it can simply be declared on the stack as it was done before resulting in less unsafe code to audit and easier-to-reason-about instance safety!

Dirbaio:
good point. I didn't realize there were so many constructors. thanks!

PR #3192, opened 2024-07-17
Rp example vsys voltage vbus edge

1-rafael-1:
Example how to
- get the voltage applied to vsys
- detect edge on vbus
- do all this while still retaining the ability to use the wifi chip

After this proved more tricky than anticipated on a Pi Pico W because of the wifi chip i decided to hand in an example with mutex and asome explanation around all this - in the hope it may help others.

### DIFF
--- a/examples/rp/src/bin/vsys_voltage_vbus_edge.rs
+++ b/examples/rp/src/bin/vsys_voltage_vbus_edge.rs
@@ -10,6 +10,7 @@
 #![no_main]
 
 use core::borrow::BorrowMut;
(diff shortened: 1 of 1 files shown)


1-rafael-1:
convert to draft -> more testing in my project revelas that a mutex around the wifi peripherals cannot be released due to the tasks required for wifi and net.
Will see if i find a way around that.

1-rafael-1:
this is leading nowhere -> out of the things i learned i made #3216

PR #3196, opened 2024-07-18
Feature i2 slave r2

smeenka:
Hereby a proposal for a i2c slave for the stm32. 

Goal is to support a slave module with 2 addresses (module address and generic address), which can handle a (limited) burst of transactions from the master (in arbitrary order) in real time. 
One can prepare 2 read transactions in embassy context (non real time). The transactions are handled in interrupt context (real-time), and the results are collected in a channel. The transactions can be one by one handled in embassy context, and new transactions can be prepared.
I did choose to make the buffer size NOT generic. Using generics (like  generic_array) for the buffer size makes the code (and error messages) extreme complex.

Functionality is tested with tests i2c_master and i2c_slave in the examples/stm32g0 directory.

Note that this branch is a follow up of branch feature-i2c-slave.


embassy-stm32/src/i2c/v2slave.rs
```diff
@@ -0,0 +1,459 @@
...
+    fn new_read(in_buffer: &[u8]) -> Self {
+        let mut buffer = [0; SLAVE_BUFFER_SIZE];
+        let size = if in_buffer.len() < SLAVE_BUFFER_SIZE {
+            in_buffer.len()
+        } else {
+            SLAVE_BUFFER_SIZE
+        };
+        for i in 0..size {
+            buffer[i] = in_buffer[i];
+        }
```
samueltardieu, line 49:
```suggestion
        let mut buffer = [0; SLAVE_BUFFER];
        let size = in_buffer.len().min(SLAVE_BUFFER_SIZE);
        buffer[..size].copy_from_slice(&in_buffer[..size]);
```

smeenka:
I need some advice how to fix the 3 failing tests..

smeenka:
Need some help for solving the stm32_metapac::i2c issues. Locally the code does build for stable and nightly. So why not on the build server? Is it possible to remove the $HOME/.cargo/bin/git/checkout/* directory on the build server, then try again?

Dirbaio:
Hello,

PR #3200, opened 2024-07-22
Documentation Link Fixes

Dygear:
This fixes some minor documentation link issues. Resolves problems where the links were still referencing the page they used to be on. It also fixes the master branch being renamed to main, so an interstitial is not shown on the Github side of things.


jamesmunns (approved):
Thank you!

PR #3207, opened 2024-07-25
Add extra lifetime to opamp-using structs

### LINKED ISSUE #3184: Opamp output lifetimes slightly wrong
I'm new to embassy and Rust, so forgive me if I make a mistake here. I wasn't able to put an OpAmp(Internal)Output in a struct. I think it's because the lifetimes of the self reference and return types of the following functions are slightly wrong. Below the functions I have written what I think it should be.

https://github.com/embassy-rs/embassy/blob/e54c753537b4b12c3d2fd03ad8e8ba9eaaded06e/embassy-stm32/src/opamp.rs#L82-L87

```Rust
pub fn buffer_ext( 
     &mut self, 
     in_pin: impl Peripheral<P = impl NonInvertingPin<T> + crate::gpio::Pin>, 
     out_pin: impl Peripheral<P = impl OutputPin<T> + crate::gpio::Pin> + 'd, 
     gain: OpAmpGain, 
 ) -> OpAmpOutput<'_, T> {
```

https://github.com/embassy-rs/embassy/blob/e54c753537b4b12c3d2fd03ad8e8ba9eaaded06e/embassy-stm32/src/opamp.rs#L121-L124

```Rust
 pub fn buffer_dac( 
     &mut self, 
     out_pin: impl Peripheral<P = impl OutputPin<T> + crate::gpio::Pin> + 'd, 
 ) -> OpAmpOutput<'_, T> { 
```

https://github.com/embassy-rs/embassy/blob/e54c753537b4b12c3d2fd03ad8e8ba9eaaded06e/embassy-stm32/src/opamp.rs#L149-L153

```Rust
 pub fn buffer_int( 
     &mut self, 
     pin: impl Peripheral<P = impl NonInvertingPin<T> + crate::gpio::Pin>, 
     gain: OpAmpGain, 
 ) -> OpAmpInternalOutput<'_, T> { 
```

diondokter:
Fixes #3184 

Often the `'d` lifetime is `'static`. But with the way the lifetime setup was done, this would force any of the outputs to borrow the opamp for the `'static` lifetime too.

This PR introduces a new lifetime so that the borrowing of the opamp HAL object is separated from the lifetime of the PAC object.

@AlexD9555 could you test if this fixes everything for you?
You can do this by setting the patched git rev in your Cargo.toml to `6ef9e564cad088c246af97f408c5b34dc47301b5`

If you give the ok, then I'll removed the draft status of this PR.

### DIFF
--- a/embassy-stm32/src/opamp.rs
+++ b/embassy-stm32/src/opamp.rs
@@ -38,15 +38,15 @@ impl From<OpAmpSpeed> for crate::pac::opamp::vals::Opahsm {
 /// OpAmp external outputs, wired to a GPIO pad.
 ///
 /// This struct can also be used as an ADC input.
-pub struct OpAmpOutput<'d, T: Instance> {
-    _inner: &'d OpAmp<'d, T>,
+pub struct OpAmpOutput<'a, 'd: 'a, T: Instance> {
+    _inner: &'a OpAmp<'d, T>,
 }
 
 /// OpAmp internal outputs, wired directly to ADC inputs.
 ///
 /// This struct can be used as an ADC input.
-pub struct OpAmpInternalOutput<'d, T: Instance> {
-    _inner: &'d OpAmp<'d, T>,
+pub struct OpAmpInternalOutput<'a, 'd: 'a, T: Instance> {
+    _inner: &'a OpAmp<'d, T>,
 }
 
 /// OpAmp driver.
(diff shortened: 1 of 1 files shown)


AlexD9555:
Thanks, this fixed the problem. I've tested it on a STM32G484 without any errors or warnings.

embassy-stm32/src/opamp.rs
```diff
@@ -38,15 +38,15 @@ impl From<OpAmpSpeed> for crate::pac::opamp::vals::Opahsm {
 /// OpAmp external outputs, wired to a GPIO pad.
 ///
 /// This struct can also be used as an ADC input.
-pub struct OpAmpOutput<'d, T: Instance> {
-    _inner: &'d OpAmp<'d, T>,
+pub struct OpAmpOutput<'a, 'd: 'a, T: Instance> {
```
Dirbaio, line 41: these shouldn't be needed. OpAmp is covariant, meaning you can coerce from `OpAmp<'long>` to `OpAmp<'short>`, so we don't need to keep the original lifetime.

The bug is just the signatures of the functions, that should return `OpAmpOutput<'_>` instead of `OpAmpOutput<`d>`

diondokter: Ah hmmmm, yeah I see. Fixed it!

Dirbaio (approved):
thanks!

PR #3209, opened 2024-07-25
Update to Rust 1.80, make check-cfg unconditional.

Dirbaio:


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.79"
+channel = "1.80"
```
bugadani, line 2: If I can believe my eyes, at least on windows only "1.80.0" works. Though that's from me hacking on probe-rs

PR #3212, opened 2024-07-26
feat(usb): Prepare `embassy-usb` for USB Audio, and add USB Audio Class 1.0 (playback only)

elagil:
Hello,

I implemented a UAC 1.0 device, but this required some extensions and fixes to the implementation of `embassy-usb`.
Some changes are taken from or inspired by https://github.com/embassy-rs/embassy/pull/2736.

Here is an example of where this is in use (still rough): https://github.com/blus-audio/firmware-rs
The feedback endpoint is not yet in use.

Features:
- Added usage type and synchronization type settings for endpoints. This is only relevant for isochronous endpoints.
- Added `extra_fields` that can be appended to an endpoint descriptor. This is required for UAC (e.g. `bRefresh` and `bSynchAddress` fields for UAC1.0)
- Added functionality for allocating endpoints and writing their descriptors separately. This is required, because in UAC1.0, the `synch` endpoint must follow the audio streaming endpoint, but the descriptor of the latter must contain the endpoint index of the former. So I allocate the endpoints first, then write the descriptors with the information about the endpoints. Find an implementation of that approach here: https://github.com/blus-audio/firmware-rs/blob/main/firmware/src/uac1/mod.rs#L223

Fixes:
- Set correct even/odd frame bit for isochronous endpoints. Otherwise, half of the packets are lost.
- Consistent naming of register setters

### DIFF
--- a/embassy-usb/src/class/mod.rs
+++ b/embassy-usb/src/class/mod.rs
@@ -3,4 +3,5 @@ pub mod cdc_acm;
 pub mod cdc_ncm;
 pub mod hid;
 pub mod midi;
+pub mod uac1;
 pub mod web_usb;
--- a/embassy-usb/src/class/uac1/class_codes.rs
+++ b/embassy-usb/src/class/uac1/class_codes.rs
@@ -1,0 +1,151 @@
+//! Audio Device Class Codes as defined in Universal Serial Bus Device Class
+//! Definition for Audio Devices, Release 1.0, Appendix A and Universal Serial
+//! Bus Device Class Definition for Audio Data Formats, Release 1.0, Appendix
+//! A.1.1 (Audio Data Format Type I Codes)
+#![allow(dead_code)]
+
+/// The current version of the ADC specification (1.0)
+pub const ADC_VERSION: u16 = 0x0100;
+
+/// The current version of the USB device (1.0)
+pub const DEVICE_VERSION: u16 = 0x0100;
+
+/// Audio Interface Class Code
+pub const USB_AUDIO_CLASS: u8 = 0x01;
+
+// Audio Interface Subclass Codes
+pub const USB_UNDEFINED_SUBCLASS: u8 = 0x00;
+pub const USB_AUDIOCONTROL_SUBCLASS: u8 = 0x01;
+pub const USB_AUDIOSTREAMING_SUBCLASS: u8 = 0x02;
+pub const USB_MIDISTREAMING_SUBCLASS: u8 = 0x03;
+
+// Audio Protocol Code
+pub const PROTOCOL_NONE: u8 = 0x00;
+
+// Audio Class-Specific Descriptor Types
+pub const CS_UNDEFINED: u8 = 0x20;
+pub const CS_DEVICE: u8 = 0x21;
+pub const CS_CONFIGURATION: u8 = 0x22;
+pub const CS_STRING: u8 = 0x23;
+pub const CS_INTERFACE: u8 = 0x24;
+pub const CS_ENDPOINT: u8 = 0x25;
+
+// Descriptor Subtype
+pub const AC_DESCRIPTOR_UNDEFINED: u8 = 0x00;
+pub const HEADER_SUBTYPE: u8 = 0x01;
+pub const INPUT_TERMINAL: u8 = 0x02;
+pub const OUTPUT_TERMINAL: u8 = 0x03;
+pub const MIXER_UNIT: u8 = 0x04;
+pub const SELECTOR_UNIT: u8 = 0x05;
+pub const FEATURE_UNIT: u8 = 0x06;
+pub const PROCESSING_UNIT: u8 = 0x07;
+pub const EXTENSION_UNIT: u8 = 0x08;
+
+// Audio Class-Specific AS Interface Descriptor Subtypes
+pub const AS_DESCRIPTOR_UNDEFINED: u8 = 0x00;
+pub const AS_GENERAL: u8 = 0x01;
+pub const FORMAT_TYPE: u8 = 0x02;
+pub const FORMAT_SPECIFIC: u8 = 0x03;
+
+// Processing Unit Process Types
+pub const PROCESS_UNDEFINED: u16 = 0x00;
+pub const UP_DOWNMIX_PROCESS: u16 = 0x01;
+pub const DOLBY_PROLOGIC_PROCESS: u16 = 0x02;
+pub const DDD_STEREO_EXTENDER_PROCESS: u16 = 0x03;
+pub const REVERBERATION_PROCESS: u16 = 0x04;
+pub const CHORUS_PROCESS: u16 = 0x05;
+pub const DYN_RANGE_COMP_PROCESS: u16 = 0x06;
+
+// Audio Class-Specific Endpoint Descriptor Subtypes
+pub const EP_DESCRIPTOR_UNDEFINED: u8 = 0x00;
+pub const EP_GENERAL: u8 = 0x01;
+
+// Audio Class-Specific Request Codes
+pub const REQUEST_CODE_UNDEFINED: u8 = 0x00;
+pub const SET_CUR: u8 = 0x01;
+pub const GET_CUR: u8 = 0x81;
+pub const SET_MIN: u8 = 0x02;
+pub const GET_MIN: u8 = 0x82;
+pub const SET_MAX: u8 = 0x03;
+pub const GET_MAX: u8 = 0x83;
+pub const SET_RES: u8 = 0x04;
+pub const GET_RES: u8 = 0x84;
+pub const SET_MEM: u8 = 0x05;
+pub const GET_MEM: u8 = 0x85;
+pub const GET_STAT: u8 = 0xFF;
+
+// Terminal Control Selectors
+pub const TE_CONTROL_UNDEFINED: u8 = 0x00;
+pub const COPY_PROTECT_CONTROL: u8 = 0x01;
+
+// Feature Unit Control Selectors
+pub const FU_CONTROL_UNDEFINED: u8 = 0x00;
+pub const MUTE_CONTROL: u8 = 0x01;
+pub const VOLUME_CONTROL: u8 = 0x02;
+pub const BASS_CONTROL: u8 = 0x03;
+pub const MID_CONTROL: u8 = 0x04;
+pub const TREBLE_CONTROL: u8 = 0x05;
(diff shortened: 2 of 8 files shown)


kalkyl:
Awesome!
Run `cargo +nightly fmt` to fix CI

Do you have a basic usage example to include?

elagil:
Thanks! I just fixed formatting.

I could add an example for the STM32H7, because I can test that.
In my case, I stream USB->SAI->DAC.

However, such an example would still be a bit useless without explicit feedback. There will be jumps in the audio due to the rate discrepancies of SAI output and USB input. Measuring the value that has to be provided over the feedback endpoint is hardware dependent.

I will try to make something useful.

elagil:
I guess I have to do this first, in order to create an example: https://github.com/embassy-rs/stm32-data/issues/509#issuecomment-2247156321

Maybe some details for the reasoning:
The fix allows v2 timers to trigger on the `USB_SOF` signal, which will be used for measuring the feedback value.
The feedback value expresses, how many samples were consumed by the audio device between two `USB_SOF` interrupts. It thus relates the rate of sample consumption to the rate at which input samples are provided. The feedback value is sent to the host, in order for it to adjust the rate at which it sends samples. It can be thought of as a control loop, where the USB audio device is the sensor.

elagil:
This should now be ready for review. You can go through the commits one-by-one:
- Preparation of `embassy-usb` for USB audio
- Implementation of UAC1
- An example, based on STM32F4

The class is meant to be extended by more "applications" in the future. I implemented only a "speaker" application, where the device receives audio from the host. It has the following features:
- Volume control
- Mute control
- Selectable sample rate
- Selectable sample width
- Implementation of the sample rate feedback mechanism (see example for details)
- Separate streaming, control, and feedback instances

**Disclaimer:** I did not test the example on the STM32F4 platform, because I don't have that hardware. I only tested on my STM32H723, which does not exist as an example platform, and I feel like it makes more sense to use the more common F4.

Can someone please test it? For example, print the measured feedback value (in the `usb_feedback_task`) and received sample lengths (in the `audio_receiver_task`).

elagil:
ISO endpoint support itself in the USB drivers was split off to https://github.com/embassy-rs/embassy/pull/3314.
Therefore, this depends on the other PR.

kalkyl:
Is this PR still blocked by something?

elagil:
No, it should be fine. I am only waiting for review and maybe for someone to restart the build job.

I may also rebase it.

kalkyl:
Looks great to me!

vDorst:
@elagil I am trying out your code on my Blackpill (STM32F401 with a PCM5102 DAC without MCLK (I2S3)).

I am the feeling that you are using the SOF capture not in the way it was intended.

Ideally you can run timer2 in normal mode and use channel 1 as a capture channel.
But your have to route the SOF signal to the trigger input.

One crucial part is setting the option registers.
But that is very microcontroller specific, you have to look in the Referance manuals what the options are.
```
// See RM0368 TIM2 option register (TIM2_OR)
    tim2.regs_gp32().or().write(|r| *r = 0b10 << 10);
```
This is how I currently have wire it.
https://github.com/vDorst/stm32_audio/commit/2a8c20c35e405f396b7de428857eda6aa7521381

Other thing I see is that something is flaky.
I mean that I get different errors/panic while the Linux is setting-up the USB/soundcard.
If I press the reset, I may the an other error or it just works fine.

Like issue in USB driver
```
0.848297 INFO  Volume changed to Muted on channel LeftFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
0.848358 INFO  Volume changed to Muted on channel RightFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
0.851623 DEBUG USB set interface number InterfaceNumber(1) to alt setting 0.
└─ embassy_usb::class::uac1::speaker::{impl#10}::set_alternate_setting @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:745 
0.851776 DEBUG Got channel 1 mute state: true.
└─ embassy_usb::class::uac1::speaker::{impl#9}::interface_get_request @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:660 
0.851989 DEBUG Got channel 2 mute state: true.
└─ embassy_usb::class::uac1::speaker::{impl#9}::interface_get_request @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:660 
0.852386 ERROR panicked at /home/r/dev/rust/embassy/embassy-usb-synopsys-otg/src/lib.rs:46:9:
assertion failed: ep_num < ep_count
```

Index out of bound
```
0.681365 INFO  Volume changed to DeciBel(0.0) on channel LeftFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
0.681427 INFO  Volume changed to DeciBel(0.0) on channel RightFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
0.681701 INFO  endpoint_set_request: Request { direction: Out, request_type: Class, recipient: Endpoint, request: 1, value: 256, index: 1, length: 3 } []
└─ embassy_usb::class::uac1::speaker::{impl#9}::endpoint_set_request @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:588 
0.681823 ERROR panicked at /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:606:36:
index out of bounds: the len is 0 but the index is 0
```

BufferOverflow in USB
```
21.306182 WARN  control accept_in failed: BufferOverflow
└─ embassy_usb::{impl#1}::handle_control_in::{async_fn#0} @ /home/r/dev/rust/embassy/embassy-usb/src/lib.rs:383 
```

When it works and I change the volume, then I see the change in the probe-rs output.
again a `assertion failed: ep_num < ep_count`
```
16.083160 DEBUG Set channel 1 volume: -4608
└─ embassy_usb::class::uac1::speaker::{impl#9}::interface_set_volume @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:533 
16.083251 INFO  Volume changed to DeciBel(-18.0) on channel LeftFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
16.083343 INFO  Volume changed to DeciBel(-17.0) on channel RightFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
16.083435 INFO  interface_set_request: Request { direction: Out, request_type: Class, recipient: Interface, request: 1, value: 514, index: 512, length: 2 } [0, 238]
└─ embassy_usb::class::uac1::speaker::{impl#9}::interface_set_request @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:543 
16.083679 DEBUG Set channel 2 volume: -4608
└─ embassy_usb::class::uac1::speaker::{impl#9}::interface_set_volume @ /home/r/dev/rust/embassy/embassy-usb/src/class/uac1/speaker.rs:533 
16.083770 INFO  Volume changed to DeciBel(-18.0) on channel LeftFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
16.083862 INFO  Volume changed to DeciBel(-18.0) on channel RightFront.
└─ stm32_audio::__usb_control_task_task::{async_fn#0} @ src/main.rs:223 
16.130187 ERROR panicked at /home/r/dev/rust/embassy/embassy-usb-synopsys-otg/src/lib.rs:47:13:
ep num: 14 cnt 4
assertion failed: ep_num < ep_count
```



I have the feeling that the stm32-usb driver is buggy.

I am using embassy commit 8eb80c6816a1f6ea7814acd4f48f07e79891f804
Os: OpenSUSE with kernel `6.11.3-2-default`
Rust: rustc 1.82.0 (f6e511eec 2024-10-15)

elagil:
> I am the feeling that you are using the SOF capture not in the way it was intended.
> 
> Ideally you can run timer2 in normal mode and use channel 1 as a capture channel. But your have to route the SOF signal to the trigger input.
> 
> One crucial part is setting the option registers. But that is very microcontroller specific, you have to look in the Referance manuals what the options are.

Yes, you are right. By just reading the counter value in the interrupt, there will be some error due to delays. Using capture should fix that. I will also enable the missing routing of the trigger input for F4.

> I have the feeling that the stm32-usb driver is buggy.

That looks like it, unfortunately. Do you have an idea what might cause it? I think I have some F401 board that I can test this on as well. The H7 (that I tested with) and the F4 use the same synopsis OTG driver, but there are some device-specific config-functions. I will have a look at those. Another idea is that there is some issue with the FIFO sizes.

elagil:
I can reproduce your issues with some custom F401 board that I have.

Some configuration values are not received correctly. For example, in `dmesg` I see

```
[ 1329.474212] usb 3-5.1: current rate 458752 is different from the runtime rate 48000
[ 1344.887822] usb 3-5.1: 2:0: failed to get current value for ch 2 (-110)
```

Some investigation will be required.

vDorst:
@elagil I have no idea what is causes it.

I have seen these dmesged too.

elagil:
@vDorst I changed the timer setup for SOF to capture mode

The example should work on F4, as soon as https://github.com/embassy-rs/embassy/pull/3565 is merged. I don't have an F429 to test, but I did test the H5 example for sample and feedback transfer. I also made sure the feedback value is reasonable.

From my point of view, this is ready to merge.

Dirbaio (approved):
thanks!

(rebased to fix missing ci job)

elagil:
> thanks!
> 
> (rebased to fix missing ci job)

Thanks for merging!

vDorst:
@elagil Great, I had no time to test it before the merge.

I copied the TMR2 setup and interrupt handler from the F4 example in my code.
First the setup so I could see the interrupt working and second time also the interrupt code.
It seems to work fine without a glitch.

elagil:
@vDorst Great, thanks for testing!

elagil:
@vDorst Did you get a valid I2S stream from your F401 with your example code? For me, the bit clock is wrong (around 5.4 MHz instead of expected 3.072 MHz for 48 kHz sample rate at 32 bit stereo).

vDorst:
@elagil Not with a logic-analyser/scoop.
But I did noticed that the frequency was off.
This is the hack to fix that: https://github.com/vDorst/stm32_audio/blob/pr3212/src/main.rs#L434

I ran this code with that hardware for weeks as the main headphone soundcard.
I didn't noticed any sound issues.

PR #3213, opened 2024-07-26
Implement low-power feature for STM32L4 MCUs

DrTobe:
This PR enables the `low-power` feature on STM32L4 devices by implementing the required traits and setting the correct `WakeupInterrupt` and `EXTI_WAKEUP_LINE`.

I have tested this on a NUCLEO-L452RE board with a program inspired by the `stm32l5/stop.rs` example. With `enable_debug_during_sleep = false` and the LED turned off, the current consumption drops down to around 3uA when the delays are large enough to allow STOP2 mode. Waking up works as well (normally, see below).

#### RTC version on STM32L41/42xxx devices
Both the RM0394 reference manual for STM32L4 devices and the AN4759 application note which is also mentioned in embassy-stm32's comments indicate that STM32L41/L42xxx devices use another RTC. Until now, I have not understood how the correct RTC implementation is chosen in embassy's build process, but I can see that the `rtc/v1.rs` file seems to be missing although it is (still) referenced in `rtc/mod.rs`. Altogether, I do not know if the implementation I have done is working for these devices, too. Can someone shed some light upon this?

#### Wake-Up Issues
I have experienced some wake-up issues. Concretely, the device does not wake-up when the first delay is short and prevents STOP2 mode. In a few cases, my `stop.rs` example hung already after a few seconds (so it was definitely not caused by the 30s timeout task). In any case, I do not think that this issue is caused by any changes I did for this PR. Therefore, I am wondering if this also happens on other STM32 families. Unfortunately, I do not have any other device at hand currently. Has anybody experienced something similar on another device type?

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -109,10 +109,10 @@ pub enum StopMode {
     Stop2,
 }
 
-#[cfg(stm32l5)]
+#[cfg(any(stm32l4, stm32l5))]
 use stm32_metapac::pwr::vals::Lpms;
 
-#[cfg(stm32l5)]
+#[cfg(any(stm32l4, stm32l5))]
 impl Into<Lpms> for StopMode {
     fn into(self) -> Lpms {
         match self {
@@ -181,7 +181,7 @@ impl Executor {
 
     #[allow(unused_variables)]
     fn configure_stop(&mut self, stop_mode: StopMode) {
-        #[cfg(stm32l5)]
+        #[cfg(any(stm32l4, stm32l5))]
         crate::pac::PWR.cr1().modify(|m| m.set_lpms(stop_mode.into()));
         #[cfg(stm32h5)]
         crate::pac::PWR.pmcr().modify(|v| {
--- a/embassy-stm32/src/rtc/low_power.rs
+++ b/embassy-stm32/src/rtc/low_power.rs
@@ -65,7 +65,7 @@ pub(crate) enum WakeupPrescaler {
     Div16 = 16,
 }
 
-#[cfg(any(stm32f4, stm32l0, stm32g4, stm32l5, stm32wb, stm32h5, stm32g0))]
+#[cfg(any(stm32f4, stm32l0, stm32g4, stm32l4, stm32l5, stm32wb, stm32h5, stm32g0))]
 impl From<WakeupPrescaler> for crate::pac::rtc::vals::Wucksel {
     fn from(val: WakeupPrescaler) -> Self {
         use crate::pac::rtc::vals::Wucksel;
@@ -79,7 +79,7 @@ impl From<WakeupPrescaler> for crate::pac::rtc::vals::Wucksel {
     }
 }
 
-#[cfg(any(stm32f4, stm32l0, stm32g4, stm32l5, stm32wb, stm32h5, stm32g0))]
+#[cfg(any(stm32f4, stm32l0, stm32g4, stm32l4, stm32l5, stm32wb, stm32h5, stm32g0))]
 impl From<crate::pac::rtc::vals::Wucksel> for WakeupPrescaler {
     fn from(val: crate::pac::rtc::vals::Wucksel) -> Self {
         use crate::pac::rtc::vals::Wucksel;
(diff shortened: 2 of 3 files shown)


dragonnn:
I did had wakeup issues on STM32F4, with is why I made a PR https://github.com/embassy-rs/embassy/pull/2851 with helped in my use case (or at least partly) but I didn't had time to investigate it deeper what is the source of that issue.

DrTobe:
Thanks for pointing out that issue. If this behaviour could also be observed on other MCUs, I would proceed on the assumption that my changes did not introduce that bug so this should not be blocking a merge in my opinion.

So I would only still like to see some feedback about the RTC version question.

Dirbaio:
> Both the RM0394 reference manual for STM32L4 devices and the AN4759 application note which is also mentioned in embassy-stm32's comments indicate that STM32L41/L42xxx devices use another RTC. Until now, I have not understood how the correct RTC implementation is chosen in embassy's build process

This comes from stm32-data. [here](https://github.com/embassy-rs/stm32-data-generated/blob/63dcecb1b412ec6e7f872b652f6bf4809e0a94c2/data/chips/STM32L496ZG.json#L5423) vs [here](https://github.com/embassy-rs/stm32-data-generated/blob/63dcecb1b412ec6e7f872b652f6bf4809e0a94c2/data/chips/STM32L422TB.json#L1819) for example. Then [this](https://github.com/embassy-rs/embassy/blob/afd910016cd185dd121a47c3d481e6148640f071/embassy-stm32/build.rs#L41) converts that to cfgs like `rtc_v3`.

> but I can see that the rtc/v1.rs file seems to be missing although it is (still) referenced in rtc/mod.rs. 

yeah RTCv1 is unimplemented. It still builds because the entire rtc mod is disabled for v1 [here](https://github.com/embassy-rs/embassy/blob/afd910016cd185dd121a47c3d481e6148640f071/embassy-stm32/src/lib.rs#L102).

> Altogether, I do not know if the implementation I have done is working for these devices, too. 

Someone would have to go and test them. However I'll merge anyway since having it working for some devices is already better, no need to aim for perfect.

> I have experienced some wake-up issues. Concretely, the device does not wake-up when the first delay is short and prevents STOP2 mode. In a few cases, my stop.rs example hung already after a few seconds (so it was definitely not caused by the 30s timeout task). In any case, I do not think that this issue is caused by any changes I did for this PR. Therefore, I am wondering if this also happens on other STM32 families. Unfortunately, I do not have any other device at hand currently. Has anybody experienced something similar on another device type?

yeah wakeup is known to be a bit flaky. I don't know why. 

Ultimately I think the better fix will be to make a time driver using LPTIM, instead of using RTC. It has a few advantages, like not needing to dynamically switching between high-frequency TIMx and low-frequency RTC, which causes losing ticks while switching. The current low-power support uses RTC because the person who contributed it was using an F4 which doesn't have LPTIM.

chrenderle:
> This PR enables the `low-power` feature on STM32L4 devices by implementing the required traits and setting the correct `WakeupInterrupt` and `EXTI_WAKEUP_LINE`.
> 
> I have tested this on a NUCLEO-L452RE board with a program inspired by the `stm32l5/stop.rs` example. With `enable_debug_during_sleep = false` and the LED turned off, the current consumption drops down to around 3uA when the delays are large enough to allow STOP2 mode. Waking up works as well (normally, see below).
> 
> ### Issues / Open Questions
> 
> #### RTC version on STM32L41/42xxx devices
> Both the RM0394 reference manual for STM32L4 devices and the AN4759 application note which is also mentioned in embassy-stm32's comments indicate that STM32L41/L42xxx devices use another RTC. Until now, I have not understood how the correct RTC implementation is chosen in embassy's build process, but I can see that the `rtc/v1.rs` file seems to be missing although it is (still) referenced in `rtc/mod.rs`. Altogether, I do not know if the implementation I have done is working for these devices, too. Can someone shed some light upon this?
> 
> #### Wake-Up Issues
> I have experienced some wake-up issues. Concretely, the device does not wake-up when the first delay is short and prevents STOP2 mode. In a few cases, my `stop.rs` example hung already after a few seconds (so it was definitely not caused by the 30s timeout task). In any case, I do not think that this issue is caused by any changes I did for this PR. Therefore, I am wondering if this also happens on other STM32 families. Unfortunately, I do not have any other device at hand currently. Has anybody experienced something similar on another device type?

I was playing around with a L4-Nucleo and the low power feature a year ago or so. I remember having some issues with it not being reliable. Although this could also have been with a L5 Nucleo or L0 Nucleo. Since I have many Nucleos I can first try to reproduce your problems. Then see if the same code has the problems too on a different Nucleo (L5 or U5). Ideally you could share me your code so I can run exactly the code you had problems with.

PR #3214, opened 2024-07-27
Added join_vec function

busslina:
Not tested and missing documentation

### DIFF
--- a/embassy-futures/Cargo.toml
+++ b/embassy-futures/Cargo.toml
@@ -26,3 +26,6 @@ features = ["defmt"]
 [dependencies]
 defmt = { version = "0.3", optional = true }
 log = { version = "0.4.14", optional = true }
+
+[features]
+alloc=[]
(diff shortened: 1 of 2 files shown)


busslina:
Tested

Discussed [here](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$p7VzbndAbQbIy2t33n8yCZBBnjj3a50wlHaMFLGVEp4?via=matrix.org&via=beeper.com&via=fehler-in-der-matrix.de)

lulf:
To me it feels like making embassy-futures alloc is a bit out of scope, and it would be better for you to use something like https://docs.rs/futures/0.3.30/futures/future/fn.join_all.html in those cases. Embassy-futures is 'Utilities for working with futures, compatible with no_std and not using alloc. Optimized for code size, ideal for embedded systems.'.

busslina:
@lulf Thank you, I was looking for that function. I just switched to it and works well. Closing this PR

PR #3216, opened 2024-07-27
add example to rp: orchestrate multiple tasks

1-rafael-1:
I did need some time myself to wrap my head around this and had some kind help from @kalkyl and @jamesmunns on this along the way. There was a question on matrix today around the topic and a few days back also, if I am not mistaken -> so in the hope it might help others :-)


1-rafael-1:
done now,

1-rafael-1:
@kalkyl suggested to add on matrix: 

Nice example rafael ! Maybe would be nice to have that cancelled task be spawned/respawned from the ”orchestrator” too to show this? Also maybe add some consumer task (or in main) that uses the (shared) state? Maybe also add a warning to not send too big things over the channels as a good practice?

1-rafael-1:
@kalkyl have a look if this is what You were aiming for?
@jamesmunns I have added as per input, do You want this?

Dirbaio:
thanks!

PR #3217, opened 2024-07-27
embassy-stm32: wait for BSY flag to clear before flashing

samueltardieu:
While not waiting for the BSY flag to clear works on STM32F103C8T6, it does not on APM32F103C8T6. Only one half-word gets written while the other one gets lost.


Dirbaio (approved):
thank you!

PR #3219, opened 2024-07-28
embassy-net: dns: bound the dns_servers slice

mirkomartn:
`smoltcp::socket::dns::Socket::update_servers()` will panic if a slice exceeding a fixed length is passed to it. Since `embassy-rs` never sets the said const, the default is used. Currently this is 1 (and we can be sure it will never be lower).

If there's an update in `smoltcp` API to expose the maximum length allowed with a `const fn`, the `DNS_MAX_SERVER_COUNT` could be set using it (see https://github.com/smoltcp-rs/smoltcp/pull/962).

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -34,6 +34,8 @@ use embassy_sync::waitqueue::WakerRegistration;
 use embassy_time::{Instant, Timer};
 #[allow(unused_imports)]
 use heapless::Vec;
+#[cfg(feature = "dns")]
+pub use smoltcp::config::DNS_MAX_SERVER_COUNT;
 #[cfg(feature = "igmp")]
 pub use smoltcp::iface::MulticastError;
 #[allow(unused_imports)]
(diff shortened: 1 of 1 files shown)


embassy-net/src/lib.rs
```diff
@@ -823,9 +825,11 @@ impl<D: Driver> Inner<D> {
 
         // Apply DNS servers
         #[cfg(feature = "dns")]
-        s.sockets
-            .get_mut::<smoltcp::socket::dns::Socket>(self.dns_socket)
-            .update_servers(&dns_servers[..]);
+        if !dns_servers.is_empty() {
+            s.sockets
+                .get_mut::<smoltcp::socket::dns::Socket>(self.dns_socket)
+                .update_servers(&dns_servers[..DNS_MAX_SERVER_COUNT]);
```
lulf, line 831: This should probably use https://docs.rs/smoltcp/latest/smoltcp/config/constant.DNS_MAX_SERVER_COUNT.html instead.

lulf (changes requested):

embassy-net/src/lib.rs
mirkomartn: Yes, thank you, I missed this one.

Dirbaio: `dns_servers[..DNS_MAX_SERVER_COUNT]`

this will crash if max is 2 but DHCP gives us 1.

mirkomartn:
Force pushed to use https://docs.rs/smoltcp/latest/smoltcp/config/constant.DNS_MAX_SERVER_COUNT.html instead of a hardcoded value and then once again to fix a ci/fmt fail. Also amended the commit message.

I've also added a `debug!()` statement. This change might impact users that use static config insofar previously a panic would be triggered if a config with multiple DNS servers would be passed, whereas now, some will be simply ignored. However, currently this panic is really inconvenient for users opting for a DHCP config, where a DHCP server might return multiple DNS server addresses, while staying out of users control.

I was thinking about making the `StaticConfigV4` const-generic over the size of the `dns_servers` member (currently hardcoded to 3) and then failing at compile time if that parameter exceeds `DNS_MAX_SERVER_COUNT`, although this is a change in API and it does make things more complicated. I'm not sure if providing a default would make this change backwards compatible or not, didn't have time to try it yet.

Dirbaio (approved):
thanks!

marxin:
I would like to see this fix in the next `embassy-net` crate release. Will you do one any time soon? Thanks!

PR #3221, opened 2024-07-30
embassy-sync: fix the data of LazyLock never drop

wllenyj:
Using `union` can save more space.

And the `MaybeUninit<T>` will never drop the T, when dropping the `MaybeUninit<T>`.
Fixed it.


lulf (approved):
Thanks!

PR #3222, opened 2024-07-31
Fix bootloader-rp's bootloader memory region overflow

bobrippling:
The bootloader region (`FLASH`) mentioned below overflows into the `BOOTLOADER_STATE` region by `0x100` bytes

```
0x10000100 + 24K = 0x10006100
```

https://github.com/embassy-rs/embassy/blob/32adddff9c7fc223853ada7e9ab5b7e06014a47c/examples/boot/bootloader/rp/memory.x#L3-L8


bobrippling:
If this is definitely the case, I'll check the other linker scripts for overlaps and add them to this PR

Dirbaio (approved):
oh, nice find! :+1:

bobrippling:
Thank you - I've checked the rest (only programatically!) and can't find any more overflows

PR #3225, opened 2024-08-02
Prepare some crate releases

lulf:
I would like to release embassy-nrf and some related dependent crates. Changelogs should be pretty up to date from what I can tell.


lulf:
bender run

lulf:
Well, that excalated a bit. I'll leave this hanging for a few days for people to have the change to object before merging and publishing stuff.

Dirbaio (approved):
no objections from me :rocket:

PR #3232, opened 2024-08-05
Misc fixes.

Dirbaio:
- **examples: ensure at least 3 sockets to avoid running out (DHCP, DNS, the user's)**
- **Silence some unused warnings on nightly.**

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -415,8 +415,11 @@ impl<D: Driver> Stack<D> {
     /// ## Example
     /// ```ignore
     /// let config = embassy_net::Config::dhcpv4(Default::default());
-    ///// Init network stack
-    /// static RESOURCES: StaticCell<embassy_net::StackResources<2>> = StaticCell::new();
+    /// // Init network stack
+    /// // NOTE: DHCP and DNS need one socket slot if enabled. This is why we're
+    /// // provisioning space for 3 sockets here: one for DHCP, one for DNS, and one for your code (e.g. TCP).
+    /// // If you use more sockets you must increase this. If you don't enable DHCP or DNS you can decrease it.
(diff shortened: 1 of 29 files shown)


embassy-net/src/lib.rs
```diff
@@ -416,7 +416,7 @@ impl<D: Driver> Stack<D> {
     /// ```ignore
     /// let config = embassy_net::Config::dhcpv4(Default::default());
     ///// Init network stack
-    /// static RESOURCES: StaticCell<embassy_net::StackResources<2>> = StaticCell::new();
+    /// static RESOURCES: StaticCell<embassy_net::StackResources<3>> = StaticCell::new();
```
plaes, line 419: Maybe add comment why 3?

examples/rp/.cargo/config.toml
```diff
@@ -1,5 +1,6 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "probe-rs run --chip RP2040"
+#runner = "probe-rs run --chip RP2040"
+runner = "../../sshprobe.sh"
```
plaes, line 3: Unintended change?

Dirbaio: oops

embassy-net/src/lib.rs
Dirbaio: done

vDorst: What happens if you don't have enough resources allocated, from a user perspective?
Does the application just panic's or do the tasks fight for a free socket?

Dirbaio: it panics.

Note this PR doesn't change the behavior, it just fixes many examples that had socket count set to 2 from before we had DNS (so it was enough, one for DHCP and one for the user). Now that we have DNS they need 3.

PR #3233, opened 2024-08-05
Convert uart half_duplex to use user configurable IO

kkoppul2:
Half duplex UART requires the use of Open drain output with a pull up resistor. This modification has been tested and shown to work between two NUCLEOU545 boards. The verbage on the pin configuration is the same across the reference manuals for all of the chip families that support this behavior.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -1029,8 +1029,9 @@ impl<'d> Uart<'d, Async> {
     /// (when it is available for your chip). There is no functional difference between these methods, as both
     /// allow bidirectional communication.
     ///
-    /// The pin is always released when no data is transmitted. Thus, it acts as a standard
-    /// I/O in idle or in reception.
+    /// The TX pin is always released when no data is transmitted. Thus, it acts as a standard
+    /// I/O in idle or in reception. It means that the I/O must be configured so that TX is
+    /// configured as alternate function open-drain with an external pull-up
     /// Apart from this, the communication protocol is similar to normal USART mode. Any conflict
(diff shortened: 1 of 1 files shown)


Dirbaio:
can you make this configurable? Push-pull half-duplex UART still has use cases, such as allowing for faster baudrates. You just have to ensure you never have conflicts, or add a series resistor to limit the current on conflict.

embassy-stm32/src/usart/mod.rs
```diff
@@ -1051,7 +1066,21 @@ impl<'d> Uart<'d, Async> {
         Self::new_inner(
             peri,
             None,
-            new_pin!(tx, AfType::output(OutputType::PushPull, Speed::Medium)),
+            new_pin!(
+                tx,
+                match half_duplex {
```
Dirbaio, line 1071: this is copypasted 4 times, can you extract it to a method like `HalfDuplexConfig::af_type()`?

embassy-stm32/src/usart/mod.rs
```diff
@@ -211,6 +211,30 @@ impl Default for Config {
...
+    PushPull,
+    /// Open drain output using external pull up resistor
+    OpenDrainExternal,
+    #[cfg(not(gpio_v1))]
+    /// Open drain output using internal pull up resistor
+    OpenDrainInternal,
+}
+
+impl HalfDuplexConfig {
+    pub fn af_type(self) -> gpio::AfType {
```
Dirbaio, line 228: i'd make this non-pub, there's no reason for the user to call it plus this way you don't need to doc it.

kkoppul2: Of course

Dirbaio (approved):
thanks!

PR #3239, opened 2024-08-08
nrf: Fix bunch of low-hanging clippy lint warnings

plaes:
Watchdog and gpio ones are obvious, though maybe we could still keep the uarte warnings so someone has intention to look whether it's possible to refactor the ppi arguments into struct?


Dirbaio:
can you `cargo +nightly fmt` to get CI to pass?

> look whether it's possible to refactor the ppi arguments into struct?

I wouldn't do it. You can't use `impl Trait` syntax in structs, it'd have to use actual generic params. Like 15 of them. It'd make it much harder to read in docs etc.

PR #3240, opened 2024-08-08
Fix ringbuffer hardware race condition bug.

liarokapisv:
When attempting to get a snapshot of the completion count and position, the current ringbuffer implementation uses a critical section. This does indeed prevent the completion_count from updating but does not prevent the position from updating since that is fetched directly from the DMA register.

When the code is called right at the end of the DMA cycle, the position that is read may be that of the next cycle, while the completion_count is still that of the previous cycle. As a result it looks to the code as if we went backwards and an overrun error is emitted.

In order to fix this we remove the critical section and get two readings, comparing the two in order to deduce that a wrap-around occured.

### DIFF
--- a/embassy-stm32/src/dma/ringbuffer.rs
+++ b/embassy-stm32/src/dma/ringbuffer.rs
@@ -40,6 +40,22 @@ pub struct ReadableDmaRingBuffer<'a, W: Word> {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct OverrunError;
 
+// We use this because it is not sufficient to use a critical section
+// since the get_remaining_transfers fetches directly from the DMA register
+// which is not affected by it.
+//
+// The main issue occurs when retrieving right at the cycle boundary
+// where we can't know the cycle of the retrieved position.
+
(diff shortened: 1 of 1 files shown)


liarokapisv:
@Dirbaio The commit breaks the tests due to the mocker requiring more expectations. Any thoughts on adapting the tests ?

Dirbaio:
LGTM to the fix.

I'm not familiar with the tests, I'll be OK with whatever fix you think is best.

liarokapisv:
> LGTM to the fix.
> 
> I'm not familiar with the tests, I'll be OK with whatever fix you think is best.

I think the tests would benefit from the introduction of model based testing. Would you mind adding [`proptest-state-machine`](https://proptest-rs.github.io/proptest/proptest/state-machine.html) as test dependency?

Dirbaio:
that'd require rewriting all tests, which surely is not needed for this PR. Please keep PRs focused, if you want to fix the race condition make a PR that does just that. Then afterwards we can discuss what to do with the tests. Otherwise PRs drag on forever, reviewing big refactors takes much more time than small fixes.

liarokapisv:
Yea definitely would need a separate PR for that, I agree.

liarokapisv:
Closing in favor of #3336

PR #3243, opened 2024-08-09
Initial rp235x support

CBJamo:
There are many places that could reasonably be bikesheded, as I'm the only person who's spent any time with this code and names are important. I especially want input on the feature names and organization.

Features not yet supported:
* hstx
* powman
* risc-v cores
* new pio features
* most bootrom features
* otp
* psram

Examples have been run locally during development for sanity checks, but there is not yet a test suite, as probe-rs does not yet support the rp2350.

Many thanks to @thejpster for doing all the hard parts and for writing the start block and binary info code.

This depends on [rp-pac#5](https://github.com/embassy-rs/rp-pac/pull/5). [rp23-pac](https://github.com/HellbenderInc/rp23-pac) should probably be moved to the embassy-rs org.

Marked draft at least until rp-pac and rp23-pac are updated/moved/released and the related cargo.tomls are updated.


CBJamo:
CI build fails because currently neither the 2040 nor rp235x are selected by default, so there is no pac. I'm not sure what the best way to handle that is. If it's built with -F rp2040 -F rt-2040 or -F rp235x -F rt-235x it should pass.

Tests fail because there are no tests.

I'm actually quite confused about how the nightly build worked. Is embassy-rp not built with nightly?

Dirbaio:
I will review fully soon, just one quick thought for now:

I think we should add rp235x to `rp-pac` (`stm32-metapac` style) instead of making separate crates per chip (`nrfxxx-pac` style). It's much less annoying to release and manage Cargo features. Also it makes sense given the crate is named `rp-pac`.

Chiptool doesn't have builtin support for that. Either we could add something, or we could just make chiptool generate the two sub-pacs and write lib.rs manually to include either one or the other.

CBJamo:
[rp-pac#5](https://github.com/embassy-rs/rp-pac/pull/5) now supports both the 2040 and 2350. I had assumed it'd be hard to do, but it wasn't too bad. I just updated the update.sh to make both and hand wrote a tiny lib.rs. Features did indeed clean up with the single pac.

CBJamo:
ci.sh fails, but I don't understand why. Building embassy-rp with (as far as I know) the same args works.

embassy-rp/src/spi.rs
```diff
@@ -106,15 +106,55 @@ impl<'d, T: Instance, M: Mode> Spi<'d, T, M> {
 
         if let Some(pin) = &clk {
             pin.gpio().ctrl().write(|w| w.set_funcsel(1));
+            pin.pad_ctrl().write(|w| {
+                #[cfg(feature = "rp235x")]
```
8cichy8, line 110: Looks little suspicious. Is cfg intended here or before whole write? (Same for other three same blocks below ...)

CBJamo: This is correct, only the 235x has isolation latches for the gpio pads. The remainder of the pad config is the same.

Dirbaio:
i've done some hacking on the PAC, mostly to arrayify/deduplicate more stuff. The PAC builds ~2x faster now.

Cherry-pick this commit into your branch to fix CI. https://github.com/embassy-rs/embassy/commit/1ff027bdde2da333cab943a9226e458de49dfd4c . BTW I'd request for next PRs you enable the "allow maintainers to push to the branch" thing, it's nice to allow maintainers to do small fixes like that directly on the PR.

embassy-rp/Cargo.toml
```diff
@@ -88,6 +89,17 @@ boot2-w25x10cl = []
 ## ```
 boot2-none = []
 
+rp2040 = ["rp-pac/rp2040"]
+rp235x = ["rp-pac/rp235x"]
+rp235xa = ["rp235x"]
```
Dirbaio, line 94: the `rp235x` feature is "private", we don't want the user to enable it directly, only `rp235xa` or `rp235xb`. to make that clear can you prefix it with a `_`? (take a look at how features are structured in `embassy-nrf` for example)

embassy-rp/Cargo.toml
```diff
@@ -15,6 +15,7 @@ src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-rp/sr
 features = ["defmt", "unstable-pac", "time-driver"]
 flavors = [
     { name = "rp2040", target = "thumbv6m-none-eabi" },
+    { name = "rp235x", target = "thumbv8m.main-none-eabi" },
```
Dirbaio, line 18: you have to enable the chip features for each flavor. Also i'd do one for the `a` and another for the `b` variant since it affects docs (more GPIO pins)

embassy-rp/src/binary_info/mod.rs
```diff
@@ -0,0 +1,174 @@
+//! Code and types for creating Picotool compatible "Binary Info" metadata
+//!
+//! Add something like this to your program, and compile with the "binary-info"
+//! and "rt" features:
+//!
+//! ```
+//! # use rp235x_hal as hal;
```
Dirbaio, line 7: Wrong crate name. This is caught by doctests. Can you make `cargo test --no-default-features --features time-driver,rp235xa` pass, then add it to `.gthub/ci/test.sh`?

embassy-rp/src/clocks.rs
```diff
@@ -855,8 +902,8 @@ pub enum GpoutSrc {
     Usb = ClkGpoutCtrlAuxsrc::CLK_USB as _,
     /// ADC.
     Adc = ClkGpoutCtrlAuxsrc::CLK_ADC as _,
-    /// RTC.
-    Rtc = ClkGpoutCtrlAuxsrc::CLK_RTC as _,
+    // RTC.
+    //Rtc = ClkGpoutCtrlAuxsrc::CLK_RTC as _,
```
Dirbaio, line 906: shouldn't this be cfg'd?

embassy-rp/src/dma.rs
```diff
@@ -297,3 +303,11 @@ channel!(DMA_CH8, 8);
 channel!(DMA_CH9, 9);
 channel!(DMA_CH10, 10);
 channel!(DMA_CH11, 11);
+#[cfg(feature = "rp235x")]
```
Dirbaio, line 306: add a cfg to `CHANNEL_COUNT`, or the higher newer channels will panic due to out of bounds.

embassy-rp/src/flash.rs
```diff
@@ -659,6 +667,11 @@ mod ram_helpers {
         );
     }
 
+    #[inline(never)]
+    #[link_section = ".data.ram_func"]
+    #[cfg(feature = "rp235x")]
+    unsafe fn write_flash_inner(_addr: u32, _len: u32, _data: Option<&[u8]>, _ptrs: *const FlashFunctionPointers) {}
```
Dirbaio, line 673: please don't leave empty functions if unimplemented. Either add a `todo!()` or ideally make the API inaccessible for the user.

embassy-rp/src/flash.rs
```diff
@@ -887,7 +906,7 @@ pub(crate) unsafe fn in_ram(operation: impl FnOnce()) -> Result<(), Error> {
     }
 
     // Make sure CORE1 is paused during the entire duration of the RAM function
-    crate::multicore::pause_core1();
+    //crate::multicore::pause_core1();
```
Dirbaio, line 909: don't disable this, it'll cause crashes on rp2040.

embassy-rp/src/gpio.rs
```diff
@@ -956,6 +991,44 @@ impl_pin!(PIN_27, Bank::Bank0, 27);
 impl_pin!(PIN_28, Bank::Bank0, 28);
 impl_pin!(PIN_29, Bank::Bank0, 29);
 
+#[cfg(feature = "rp235xb")]
```
Dirbaio, line 994: this needs adjusting (probably renaming...) `BANK0_PIN_COUNT` / `BANK0_WAKERS`, or async `wait_for_*` will panic due to OOB.

Dirbaio (changes requested):

embassy-rp/src/flash.rs
CBJamo: Whoops, this was just a debugging thing I forgot to remove.

embassy-rp/src/flash.rs
CBJamo: I've left this as TODO, flash will be my next pr.

embassy-rp/src/binary_info/mod.rs
CBJamo: This turned out to be quite annoying. Cortex-m gets confused so I've hacked around it [here](https://github.com/embassy-rs/embassy/blob/778241fd71b8a0cb541d23b69c3f9e47ec2753ff/embassy-rp/src/lib.rs#L518) but I'm not happy about it. I have to assume there's a smarter way to handle this.

embassy-rp/src/gpio.rs
CBJamo: I've done the cfg bits, but didn't rename anything. Pi still calls all 48 pins on the B bank 0. I think it's because of the gpio co-processor that lets you access all 48 pins in a single cycle. On the other hand, that's not implemented here yet...

CBJamo:
If it's not an issue to use the git dep for the pac then I think this is ready.

embassy-rp/src/gpio.rs
Dirbaio: > Pi still calls all 48 pins on the B bank 0.

Huh, okay.

Dirbaio (approved):
Thanks for the PR! This is awesome. I've got some rp2350's in the mail, can't wait to test it out :upside_down_face:

embassy-rp/src/flash.rs
8cichy8: You should also uncomment crate::multicore::resume_core1() few lines below.

thejpster:
Hello! Just a reminder that under the terms of the licences under which my HAL was placed, you have to include the copyright line from my MIT file along with the paragraphs below it in any copies or substantial portions of the software. As this PR lifts several modules wholesale, I’d say that counts. If I had included a NOTICES file you would also have to carry that, but I didn’t include one.

Dirbaio:
@thejpster  can you send a PR?

thejpster:
Not from my phone. Maybe in a week after my vacation.

PR #3245, opened 2024-08-09
nrf, doc: Don't refer to `Default` impl of `saadc::ChannelConfig`

frankplow:
`saadc::ChannelConfig` does not implement the `Default` trait, so its documentation should not refer to it.  Modify the documentation to instead describe how the struct should be created and configured.


lulf (approved):
Awesome, thanks!

PR #3251, opened 2024-08-13
rp: Fix commented out code in flash.rs

CBJamo:
Missed this in the review, good catch from @8cichy8

PS. I don't have an option to enable maintainer commits? Maybe because I don't own this fork? 
E: Yup, just did a quick test, I can't enable maintainer commits on prs from my work org, but can from my personal fork.


Dirbaio (approved):

Dirbaio:
> Maybe because I don't own this fork?

ah yes that might be it. maybe it's an org setting, i'm not sure.

PR #3258, opened 2024-08-15
embassy-usb: add USB device qualifier descriptor

HaoboGu:
USB device qualifier descriptor is useful when the device supports USB HS. It contains the information of the other speed mode. For example, if the device operates in HS mode, this descriptor contains information about how it could be operated in FS mode, and vice-versa.


HaoboGu:
Code in other PR is here, I'll reopen one

PR #3259, opened 2024-08-15
embassy-usb: add USB device qualifier descriptor

HaoboGu:
USB device qualifier descriptor is useful when the device supports USB HS. It contains the information of the other speed mode. For example, if the device operates in HS mode, this descriptor contains information about how it could be operated in FS mode, and vice-versa.


lulf (approved):
Thanks!

sehicfai:
Question:

According to the USB 2.0 Specification:
"If a full-speed only device (with a device descriptor version number equal to 0200H) receives a
GetDescriptor() request for a device_qualifier, it must respond with a request error. The host must not make
a request for an other_speed_configuration descriptor unless it first successfully retrieves the
device_qualifier descriptor."

Where is this case checked?

PR #3261, opened 2024-08-16
Add rp-rs copyright notice

thejpster:
Updates LICENSE-MIT and some source files to indicate the copyright holder.

The first commit is as required by the MIT license. The second I thought would be useful, but isn't strictly required.


jamesmunns (approved):

LICENSE-MIT
```diff
@@ -1,4 +1,5 @@
 Copyright (c) Embassy project contributors
+Copyright (c) rp-rs organization
```
Dirbaio, line 2: this doesn't apply to the entire repo. Would you consider the copyright line in the .rs files enough?

thejpster: > If you give someone a copy of the software, you need to include the license text and any copyright notice (https://writing.kemitchell.com/2016/09/21/MIT-License-Line-by-Line.html)

Just including the copyright notice in the source files, along with the note that it is MIT licensed, is probably sufficient for source based releases. I don't know the details of how you manage licences in binary releases (or how/if you encourage your users to manage those licenses in their binary releases), but it would be typical to encourage distributors to take a copy of any top-level `LICENSE*` and `NOTICE*` files and to redistribute those along with the binary, and in that case, the copyright notices in the source files would be lost. Perhaps a copy of the `LICENSE-MIT` file at the top-level of every HAL is a reasonable compromise?

thejpster: I should also note that the Rust Project is Quite Bad at this currently. It's a [work in progress](https://github.com/rust-lang/rust/pull/128353).

Dirbaio: ah that's a very good point. `LICENSE*` files at the root seem to be the de-facto standard, copyright lines in loose .rs files won't get picked up by people writing scripts that collect `*LICENSE*` files like yours for their binary releases.

PR #3262, opened 2024-08-16
add one more embassy in the wild example

1-rafael-1:
Want this? 
It is a hobby project, so likely interesting only to beginners,


lulf (approved):
I think its fine to add, thanks! We can add some categories later

1-rafael-1:
@lulf cool, I am happy :-) Thx!
Merge queue is stuck on that transient Bad Request thing, I have seen in another of my PRs here, though :-)

Issue #3256, opened 2024-08-14
STM32 Dual Core Init - Private Peripheral Bus considerations


liarokapisv:
The current dual-core initialization architecture assumes that system initialization is performed fully by the primary core.

Crucially this is not true for Peripherals addressed through the [Private Peripheral Bus](https://developer.arm.com/documentation/100166/0001/Programmers-Model/Processor-memory-model/Private-Peripheral-Bus), which includes the `NVIC`, `MPU` and `SysTick` through the [System Control Space](https://developer.arm.com/documentation/ddi0403/d/System-Level-Architecture/System-Address-Map/System-Control-Space--SCS-?lang=en).

Accesses to the [NVIC](https://github.com/rust-embedded/cortex-m/blob/b0d2e45a9a827f2c229340189cf125a35581f422/cortex-m/src/peripheral/mod.rs#L490) (and similarly for the other PPB peripherals), access a different instance depending on the core on which we are running.

As a result, this affects the [InterruptExt trait](https://github.com/embassy-rs/embassy/blob/7a26e117ccd5f8669548cc8c2424be4691c1c402/embassy-hal-internal/src/interrupt.rs#L153) and transively the [DMA init](https://github.com/embassy-rs/embassy/blob/7a26e117ccd5f8669548cc8c2424be4691c1c402/embassy-stm32/src/dma/dma_bdma.rs#L259) which ends up being called only on the `primary` core. This leaves the secondary core's `NVIC` with masked interrupts. The solution to this is most likely to duplicate the interrupt enablement on the `secondary_init` flow as well.

This interrupt masking is just one instance of the PPB issue and we will probably need to think things through about the rest of the `PPB` peripherals when it comes to dual core usage.

liarokapisv:
Going through the code it seems the biggest culprits are the enabled-at-initialization `dma` and `exti` modules.

wuwbobo2021:
Excuse me, I haven't bought a dual-core STM32 chip, I just feel strange about the current status of dual-core mode support.

Under version 63ade389303ab4f6fe9d34211629b02c16e9bb11 :

In `Peripheral::take_with_cs` found in `embassy-hal-internal/src/macros.rs`, it may be called for once, and the second call would always panic;

`embassy_stm32::init_primary` calls `init_hw` before allowing `init_secondary` to call `init_secondary_hw`, and both `init_hw` and `init_secondary` call `Peripheral::take_with_cs`.

[...]

PR #3263, opened 2024-08-16
stm32: fix dma nvic issues on dual core lines

liarokapisv:
This commit addresses #3256 by disabling dma NVIC interrupt ~~configuration~~ enablement at startup.

Instead, per-channel NVIC interrupt ~~configuration~~ enablement is now done with the rest of the dma channel configuration.
This ensures that each core will only handle the interrupts of the DMA channels that it uses.

Successfuly tested on a complex STM32h745IIx project with ~20 peripherals and ~9 DMA1, DMA2, BDMA channels.

Potential issues:

- ~~Removal of `dma_priority` and `bdma_priority` from `Config`: This is now consistent with the rest of the peripherals. Maybe explicitely deprecate and suggest to manually set the priorities instead?~~ I kept the priority setup on both cores and only removed the enablement.

- build.rs script: I am not confident on the `ChannelInfo` build.rs fill logic. It works on my H7 but is not robust. Any alternative good way to retrieve the `pac::Interrupt` for each dma channel ?

- This PR does not address the EXTI. It should also move the nvic configuration from startup to the `ExtInput` construction.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1494,6 +1494,36 @@ fn main() {
         .flat_map(|p| &p.registers)
         .any(|p| p.kind == "dmamux");
 
+    let mut dma_irqs: BTreeMap<&str, Vec<String>> = BTreeMap::new();
+
+    for p in METADATA.peripherals {
+        if let Some(r) = &p.registers {
+            if r.kind == "dma" || r.kind == "bdma" || r.kind == "gpdma" || r.kind == "lpdma" {
+                for irq in p.interrupts {
+                    let ch_name = format!("{}_{}", p.name, irq.signal);
+                    let ch = METADATA.dma_channels.iter().find(|c| c.name == ch_name).unwrap();
+
+                    // Some H7 chips have BDMA1 hardcoded for DFSDM, ie no DMAMUX. It's unsupported, skip it.
+                    if has_dmamux && ch.dmamux.is_none() {
+                        continue;
+                    }
+
+                    dma_irqs.entry(irq.interrupt).or_default().push(ch_name);
+                }
+            }
+        }
+    }
+
+    #[cfg(feature = "_dual-core")]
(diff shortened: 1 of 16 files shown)


Dirbaio:
- Please don't remove support for setting the irq priorities. You can still implement it with the "new-style" irqs by setting the priorities on init. it's OK to set an irq priority for an irq that you will never enable/use.
- This slightly hurts performance (a few extra clock cycles on `Transfer::new` for enabling the irq) and ram usage (a few extra bytes in `ChannelInfo` for the irqn) a bit for all chips. Could you add cfgs to do "new-style" irqs only on chips that need it (ie multicore. i'd probably add something to build.rs that checks if there's more than 1 core and enables `cfg(multicore)`.)

liarokapisv:
> Please don't remove support for setting the irq priorities. You can still implement it with the "new-style" irqs by setting the priorities on init. it's OK to set an irq priority for an irq that you will never enable/use.

You are right. I will do some changes to pass the config to the secondary core and set the priorities on both cores at init.

> This slightly hurts performance (a few extra clock cycles on Transfer::new for enabling the irq) and ram usage (a few extra bytes in ChannelInfo for the irqn) a bit for all chips. Could you add cfgs to do "new-style" irqs only on chips that need it (ie multicore. i'd probably add something to build.rs that checks if there's more than 1 core and enables cfg(multicore).)

Ideally we would have an extra layer for the Dma channels so that they would need to be statically configured to be passed to the rest of the drivers. This would eliminate having to configure them constantly on both multicore and single core projects but would probably be a very intrusive change. I can always do the `multicore` config, maybe take advantage of the `_dual-core` feature instead?

liarokapisv:
@Dirbaio Does this look ok? I am not sure about the usage of the `_dual-core` feature, is it correct to use this?

Dirbaio (approved):
thanks!

Issue #3177, opened 2024-07-14
how to find ip by host name


PhaestusFox:
I am trying to make sensors that connect to my computer when turned on but because I am using Starlink I don't have access to the DHCP to assign a static IP to my computer, I did find that as long as my computer has its IP i92.168.1.* and * < 20 then Starlink wouldn't auto assign anything to that IP range and break things, but this is unstable as sometimes when my computer restarts it can't set the IP manually and defaults back to the DHCP, this breaks all the sensors as they have a hardcoded IP for the server, I am trying to do mDNS so I can find the server by name rather than a fixed IP but can't seem to get this to work (even after enabling the feature in smoltcp). does anyone know of a way to make this work?

lulf:
What errors are you seeing if you enable the `dns` feature in embassy-net? Make sure you configure enough sockets (most examples use 2) to allow DNS. However, mDNS is not yet exposed in embassy-net so if you cannot use regular DNS you would have to implement support for the mDNS functionality from smoltcp in embassy-net.

PR #3264, opened 2024-08-17
feat(mdns): Enable mdns support through smoltcp

AnthonyGrondin:
Smoltcp supports mDNS but the feature was never used in embassy. Enabling it seems to be all what's required to resolve a `.local` address, in small local testing I did.

Should fix https://github.com/embassy-rs/embassy/issues/3177

cc @PhaestusFox


lulf (approved):
Thanks!

PR #3271, opened 2024-08-18
[UCPD] Add support for non-SOP packets

sjoerdsimons:
Allow capturing (and distinguishing) non-SOP packets as well. The default configuration will just configure SOP packets. For ease of use the default receive function signature is unchanged as for PD sinks (which is likely the common usage) just SOP is enough so no need to differentiate.


Dirbaio (approved):
thank you!

PR #3272, opened 2024-08-20
Add IOConfig struct and pub fn io_config()

beastbytes:
The STM32 TSC needs to do successive acquisitions when using multiple channels in a group. To do this the IOs must be configured for each acquisition. This change reflects the implementations in the STM32 HAL.

See:
* https://community.st.com/t5/stm32-mcus-products/tsc-channel-selection/td-p/164783

Specifically, see the HAL solution provided. It shows that only one acquisition channel must enabled in a group during an acquisition cycle, then the next channel in the next acquisition cycle, and so on.

* https://www.st.com/content/ccc/resource/technical/document/user_manual/group0/d6/4c/20/0d/a1/c1/4c/99/DM00210526/files/DM00210526.pdf/jcr:content/translations/en.DM00210526.pdf

While this is about the STM Touch Library, table 8 (and others) show the acquisition timings required for multiple channels in a group.

### DIFF
--- a/embassy-stm32/src/tsc/mod.rs
+++ b/embassy-stm32/src/tsc/mod.rs
@@ -227,6 +227,15 @@ impl Default for Config {
     }
 }
 
+pub struct IOConfig {
+    /// Channel IO mask
+    pub channel_ios: u32,
+    /// Shield IO mask
+    pub shield_ios: u32,
+    /// Sampling IO mask
+    pub sampling_ios: u32,
+}
+
 /// Pin struct that maintains usage
 #[allow(missing_docs)]
 pub struct TscPin<'d, T, C> {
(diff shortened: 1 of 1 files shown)


michelrandahl:
as someone currently learning STM32 TSC this looks good to me, although I am not sure if we need to write other masks than `ioccr` -if we have already set the other masks (sample, hysterisis, etc) during configuration. I have almost no experience with C so its hard for me to say if they write all these masks each time before an acquisition in the C HAL.

I am also working on something similar https://github.com/embassy-rs/embassy/pull/3274, and didn't see this PR before creating my draft. However, I would not object to getting some small PR like this one merged before mine.

Dirbaio:
is this use case covered by https://github.com/embassy-rs/embassy/pull/3274 ?

Dirbaio:
Hello,

PR #3277, opened 2024-08-23
embassy-net/read document return value Ok(0)

dvdsk:
Ran into this as I forgot to handle the Ok(0) return case and just kept reading. Documenting it here seemed the best way not to run into it again :)

### DIFF
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -79,6 +79,9 @@ impl<'a> TcpReader<'a> {
     ///
     /// Returns how many bytes were read, or an error. If no data is available, it waits
     /// until there is at least one byte available.
+    ///
+    /// A return value of Ok(0) means that the socket was closed and is longer able to
+    /// accept bytes or that the buffer provided is empty.
     pub async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
         self.io.read(buf).await
     }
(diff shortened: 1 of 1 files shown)


embassy-net/src/tcp.rs
```diff
@@ -79,6 +79,9 @@ impl<'a> TcpReader<'a> {
     ///
     /// Returns how many bytes were read, or an error. If no data is available, it waits
     /// until there is at least one byte available.
+    ///
+    /// A return value of Ok(0) means that the socket was closed and is longer able to
```
Dirbaio, line 83: "closed" in TCP is a bit ambiguous. It can be sending a FIN, receiving a FIN, or a RST. 

`Ok(0)` denotes EOF, which in TCP means the remote side has closed the "receive half" of the connection by sending us a FIN, and we've read all the data (ie there's no more data).

This is different than a close by RST (that you do with `.abort()`). In that case `read()` returns `Err(ConnectionReset)`. 

And it's also different than calling `.close()`. That closes the "transmit half" of the connection, ie we can't *send* more bytes, but the receive half stays open.

Can you expand the docs a bit to clarify that? thanks!

dvdsk: thanks for the explanation, ill work it into the docs :+1:

dvdsk:
I have also added some tips to the timeouts

dvdsk:
This should be correct now right?

Dirbaio (approved):
very nice, yes! thank you :)

Issue #3241, opened 2024-08-08
RP2350 support


brandonros:
Do we know enough yet what we need to do to be able to support [RP2350](https://www.raspberrypi.com/news/raspberry-pi-pico-2-our-new-5-microcontroller-board-on-sale-now/) instead of just RP2040?

CBJamo:
Initial support is coming in #3243, though much work remains.

BjornTheProgrammer:
I've preordered the raspberry pi pico 2 and would be willing to test it on actual hardware and write/verify some examples.

ionspin:
I've got a couple of Tiny2350 from pimoroni, got blinky to work without much trouble except that elf2uf2 would work because of hard float in the executable.

```
elf2uf2-rs ./target/thumbv8m.main-none-eabihf/debug/blinky                                                                                                                                                                            
Error: "HARD-FLOAT not supported"
```
Worked around that by using `picotool` and this small script:
`picotool-run.sh`
```bash
picotool uf2 convert -t elf $1 out.uf2
picotool load out.uf2
picotool reboot
```
in `.cargo/config.toml` for the `runner` 

This is a quick workaround until we get support in elf2uf2 for `thumbv8m.main-none-eabihf` or full probe-rs support.

CBJamo:
JP has a pull request for elf2uf2-rs to allow use with the rp2350. https://github.com/JoNil/elf2uf2-rs/pull/33

thejpster:
However, I recommend switching to `picotool` because it's faster and doesn't require mounting the disk drive. Binary releases are available from https://github.com/raspberrypi/pico-sdk-tools/releases

ionspin:
I'm running into a bit of issue, but I'm getting different behavior after `picotool reboot` and resseting the RP2350 (using reset button on Tiny2350, or just unplugging the Pico2). After `picotool reboot` it works fine every time, but the reset get's it into a _partially_ broken state.

This is more or less the current setup, it's reproducible if I remove the USB part and just leave the blinky loop
```rust
#[embassy_executor::main]
async fn main(spawner: Spawner) {
    let peripherals = embassy_rp::init(Default::default());


    let other_side_channel = OTHER_SIDE_CHANNEL.init(Channel::new());

    UsbHandler::new(peripherals.USB, other_side_channel, spawner, true);


    
    let delay = 100;

    let mut counter = 0;
    #[cfg(feature = "tiny2350")]
    {
[...]

PR #3279, opened 2024-08-24
Set up timer0 tick when initializing clocks

ionspin:
RP2350 requires ticks to be started for timer to start counting. `picotool reboot` seems to do this automatically, but after next restart system timer wouldn't count at all.

This addresses my comments in https://github.com/embassy-rs/embassy/issues/3241


jamesmunns:
@ionspin could you please fix the `rustfmt` issues? I can approve this once that is green.

jamesmunns (approved):

PR #3285, opened 2024-08-25
feat(embassy-net): add zero-copy UDP send/recv functions

CirrusNeptune:
Added `recv_from_with` and `send_to_with`. These are conceptually similar to TCP's `read_with` and `write_with` functions.

An application can parse received datagrams directly out of the receive buffer or assemble a datagram of known-length directly into the send buffer.


lulf (approved):
Looks good!

PR #3286, opened 2024-08-25
stm32/usart: sending break character

trnila:
Add function to send break character on STM32 USARTs.
Function is currently sync and first busy-waits until previous break has been sent (should be rare?, usart v3&v4 seems to have interrupt for sent break character).
Then sending break character is requested and it should be sent as soon as the current character is transmitted.


Dirbaio (approved):
thank you!

PR #3297, opened 2024-08-30
rp: 2350 flash write and OTP

CBJamo:
This PR depends on/contains #3287.

Direct mode reads are not supported (ie, reading the flash's jedec id or unique id).

The 2350 doesn't have a boot2 like the 2040, but it does have the concept of a xip setup function that could be customized. By default the bootrom searches for the attached flash chip and provides an xip setup func at the base of the bootram. That bootram is not executable, so it still needs to be copied to ram like boot2 would be.

Currently does not use inline assembly for write_flash_inner.

This pr also imports the OTP functions from rp-hal, and adds a couple helper functions to pull the random numbers written to the OTP by PI to make up for not being able to read the flash's unique id. Again, credit and many thanks to @thejpster for doing the real work for OTP.

Also switch to picotool, as elf2uf2 has not been patched to support the 2350.

### DIFF
--- a/embassy-rp/src/flash.rs
+++ b/embassy-rp/src/flash.rs
@@ -17,9 +17,13 @@ use crate::peripherals::FLASH;
 /// Flash base address.
 pub const FLASH_BASE: *const u32 = 0x10000000 as _;
 
+/// Address for xip setup function set up by the 235x bootrom.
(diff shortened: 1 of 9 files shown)


Dirbaio (approved):
thanks!

PR #3299, opened 2024-08-30
fix(stm32): disable transmitter during during half-duplex flush

badrbouslikhin:
This PR addresses an issue with half-duplex UART communication on STM32H7 (probably valid for all STM32s) where subsequent read cycles unexpectedly return both sent and received bytes, leading to potential data corruption. 
The fix involves disabling the transmitter (before enabling the receiver) after transmission flush in half-duplex mode.


badrbouslikhin:
Is there any blocker to merge this PR?

Dirbaio (approved):
thanks!

PR #3302, opened 2024-09-02
[embassy-time] Feature gate embedded-hal 0.2

GnomedDev:
This avoids pulling in this old version of `embedded-hal` for anyone who doesn't need it, which leads to reduced compile times especially with the removal of the `nb` dep.

### DIFF
--- a/embassy-time/CHANGELOG.md
+++ b/embassy-time/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `embedded-hal 0.2` support is now locked behind the `embedded-hal-02` feature, to be removed in a future breaking version.
+
 ## 0.3.2 - 2024-08-05
 
 - Implement with_timeout()/with_deadline() method style call on Future
--- a/embassy-time/Cargo.toml
+++ b/embassy-time/Cargo.toml
@@ -27,6 +27,9 @@ features = ["defmt", "std"]
 std = ["tick-hz-1_000_000", "critical-section/std"]
 wasm = ["dep:wasm-bindgen", "dep:js-sys", "dep:wasm-timer", "tick-hz-1_000_000"]
 
+## Implement traits from the old `embedded_hal 0.2` library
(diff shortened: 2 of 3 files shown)


Dirbaio:
we intentionally don't feature gate trait impls. They're confusing enough for newcomers, especially with the two embedded-hal versions. Adding feature gates adds an extra layer of complexity on top, where someone can get stuck not working even rustdocs say it should work because they forgot to enable the feature.

> This avoids pulling in this old version of embedded-hal for anyone who doesn't need it, which leads to reduced compile times especially with the removal of the nb dep.

both crates are tiny, they compile in milliseconds so the overall improvement is not measurable:

```
[dirbaio@mars embassy-time]$ cargo clean; cargo build --release
    Finished `release` profile [optimized] target(s) in 0.68s
    Finished `release` profile [optimized] target(s) in 0.60s
    Finished `release` profile [optimized] target(s) in 0.57s
[dirbaio@mars embassy-time]$ cargo clean; cargo build --release --features embedded-hal-02
    Finished `release` profile [optimized] target(s) in 0.61s
    Finished `release` profile [optimized] target(s) in 0.59s
    Finished `release` profile [optimized] target(s) in 0.57s
```

GnomedDev:
Okay, shall I outright remove it then? It's an outdated dependency version for a trait implementation that busy loops.

Dirbaio:
No, there's many drivers still using embedded-hal 0.2, the impls are needed for those to work.

GnomedDev:
These drivers can stay on an old embassy version, like they are staying on an old embedded-hal version, right?

Dirbaio:
The whole point of trait crates like `embedded-hal` is to enable interoperability between crates. If a driver is in e-h0.2 because the author hasn't gotten around to update it or it's unmaintained, it can still work perfectly with the latest embassy, as long as we keep the impls. There's no downside in using an old e-h, it works and that's all they care about.

Removing the impls because "they're old, and old bad" is hurting interoperability for no benefit.

Forcing the user to use an older Embassy is much more disruptive, they're forced to miss out on latest improvements and bug fixes.

GnomedDev:
Okay.

PR #3305, opened 2024-09-03
Add STM32H5 flash driver

Luctins:
This was based on the stm32h7 driver minus the flash bank logic and the fact that the h5 has secure and non secure flash.

The current implementation works, but may have some outstanding issues, like the secure registers and register lockouts, so it needs a bit more testing to be considered "ready". 

I've implemented this so I could do DFU on the stm32h5, and for that it works.

Any review is very much appreciated.

### DIFF
--- a/embassy-stm32/src/flash/h5.rs
+++ b/embassy-stm32/src/flash/h5.rs
@@ -1,0 +1,177 @@
+use core::ptr::write_volatile;
+use core::sync::atomic::{fence, Ordering};
+
+use super::{FlashRegion, FlashSector, FLASH_REGIONS, WRITE_SIZE};
+use crate::flash::Error;
+use crate::pac;
+
+pub(crate) const fn is_default_layout() -> bool {
+    true
+}
+
+// const fn is_dual_bank() -> bool {
+//     FLASH_REGIONS.len() >= 2
+// }
+
+pub(crate) fn get_flash_regions() -> &'static [&'static FlashRegion] {
+    &FLASH_REGIONS
+}
+
+pub(crate) unsafe fn lock() {
+    if !pac::FLASH.nscr().read().lock() {
+        pac::FLASH.nscr().modify(|r| {
+            r.set_lock(true);
+        });
+    }
+}
+
+pub(crate) unsafe fn unlock() {
+    // TODO: check locked first
+    while pac::FLASH.nssr().read().bsy() {
+        #[cfg(feature = "defmt")]
+        defmt::trace!("busy")
+    }
+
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/flash/h5.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    cortex_m::asm::isb();
+    cortex_m::asm::dsb();
+    fence(Ordering::SeqCst);
+
+    clear_all_err();
+
+    pac::FLASH.nscr().write(|w| {
+        w.set_pg(true);
+        // w.set_psize(2); // 32 bits at once
```
Luctins, line 65: check if size can/has to be adjusted

embassy-stm32/src/flash/h5.rs
```diff
@@ -0,0 +1,180 @@
...
+    loop {
+        let sr = pac::FLASH.nssr().read();
+        if !sr.bsy() && !sr.dbne() {
+            break;
+        }
+    }
+    clear_all_err();
+
+    pac::FLASH.nscr().modify(|r| {
+        // TODO: later check bank swap
```
Luctins, line 114: is it relevant to implement the flash bank swapping logic?

embassy-stm32/src/flash/h5.rs
```diff
@@ -0,0 +1,180 @@
...
+        r.set_bksel(match sector.bank {
```
Luctins, line 115: is there a better way to map this? like some implicit conversion. if the bank swap logic is used this has to be inverted too.

embassy-stm32/src/flash/h5.rs
```diff
@@ -0,0 +1,180 @@
...
+    pac::FLASH.nscr().write(|w| w.set_pg(false));
+
+    unwrap!(res)
+}
+
+pub(crate) unsafe fn blocking_erase_sector(sector: &FlashSector) -> Result<(), Error> {
+    // pac::FLASH.wrp2r_cur().read().wrpsg()
+    // TODO: write protection check
+    if pac::FLASH.nscr().read().lock() == true {
+        error!("flash locked");
```
Luctins, line 102: return error here as we cannot write when the flash is locked.

Luctins (commented):
added some of the outstanding issues with the current implementation

Luctins:
Can anyone review this?

Luctins:
Of course.

Luctins:
@Dirbaio I'm done fixing CI, could you review?

Dirbaio (approved):
thanks!

Issue #3295, opened 2024-08-29
USB Host support (with stm32 HAL implementation)


ijager:
I am working on a USB Host implementation, working with the STM32G0B1/C1 family.
Currently I have a basic working example doing enumeration of a HID keyboard and I can receive keystrokes.

I am looking into how to fit this best in embassy (assuming embassy would like to add usbhost functionality), and there are quite some decisions to make about what should go where and about the API surface for each layer.

This is what I have now

- Simple HIDKeyboard implementation. Could serve as an example project
- UsbHost (knows about control requests, enumeration, parsing descriptors). 
- USBHostDriverTrait that is used by USBHostDriver and imlpemented by HAL
- USBHostDriver HAL implementation for stm32 with `usb_v4` IP

## STM32 USBHostDriver

Via `cfg_attr` the HAL choses between `'usb'` and `'otg'` IP. My current usbhost hal implementation is for `usb_v4`. As far as I know only `_v4` supports host. How to fit in this functionality in with the current usb module structure? I guess we don't want to expose this for non-v4 IP.

My current host implementation is independent of the usb driver implementation. For `usb_v4` you could have both, but you cannot use them at the same time since it is the same peripheral. One way is to specify compile time which one you want. 

Which is what I did hardcoded during development:

```rust
//! Universal Serial Bus (USB)

// #[cfg_attr(usb, path = "usb.rs")]
// #[cfg_attr(otg, path = "otg.rs")]
// mod _version;
// pub use _version::*;

#[cfg_attr(usb, path = "usb_host.rs")]
mod _host_version;
pub use _host_version::*;
```

Simplest way is to add a `usb-host` config option to select `usb_host.rs` instead.
In theory one may wish to switch between host and device runtime, that wouldn't be supported like this.
Another advantage is that this approach does not risk breaking any existing usb functionality. Can always combine it later.

Alternatively if the `"usb"` implementation supports both device and host then you have to deal with not allowing that for `_v1` - `_v3`.  And you end up with a lot of `if host { some host logic } else { device logic }`

Any ideas?

## USBHostDriverTrait

`embassy-usb-driver` contains the traits for usb device implementations. Currently I have added a `host` module to this crate for the host traits.

I called it `USBHostDriverTrait` to differentiate with the `USBHostDriver` but better naming is desirable. 

So far I ended up with this API:

```rust
pub trait USBHostDriverTrait {
    type ChannelIn: ChannelIn;
    type ChannelOut: ChannelOut;

[...]

PR #3307, opened 2024-09-04
Add USBHost 

ijager:
This PR adds USBHost in several layers:

- USBHostDriver HAL for STM32 `usb_v4` devices in `embassy-stm32` crate
- USBHostDriverTrait in `embassy-usb-driver` crate
- USBHost implementation in `embassy-usb` crate
- HIDKeyboard example for `stm32g0`

This is the implementation for issue #3295.


i404788:
Hey @ijager,

I'm working on extending this PR to support 'universal' drivers (see wip @ https://github.com/i404788/embassy/tree/usb-asynch). I was wondering if there is a reason to have Channels be static (i.e. non-reconfigurable, and only one of In/Out). On the platform I'm looking to target (USBOTG), the current strategy will quickly consume all channels.

In the Hub handler example I'm working on, it's significantly more efficient to allocate 2 pipes: 1 for interrupt, and one for control IN/OUT, but that requires re-configuring the pipe.

Does the stm32 not support changing attributes? If it does, I think a shared `Channel` trait with a `reconfigure(endpoint: &EndpointDescriptor)` would improve the resource efficiency and convenience for one-shot packets. Only downside is that `write` and `read` may need to check if the direction is correct  (could `ChannelError::InvalidDirection` if it's not supported; or just attempt the transaction on the specified side, I have at least one device which has bulk IN/OUT on the same endpoint_addr  but with `| 0x80` for in).

nikvoid:
I'm also working in this direction. I completely redesigned channel and host API to make it usable from multiple concurrent tasks. On RP2040 you need to reconfigure hardware for each control/bulk/isochronous transfer and only interrupt channels are special. So in my variant channels other from interrupts are basically free, because the only resource they use is stack space. Interrupt channels are limited, but automatically freed on channel drop. Other type of channels should reconfigure hardware and block other tasks that want to issue transfers.

https://github.com/nikvoid/embassy/tree/rp2040-usb-host

https://github.com/nikvoid/embassy/blob/rp2040-usb-host/embassy-usb-driver/src/host.rs#L282
`UsbChannel` trait is parametrised by type and direction, where direction may be either In, Out, or InOut (I'm thinking of improving this part because InOut probably only suitable for Control pipes.

i404788:
@nikvoid Thanks for the notif.

That does look a lot closer to what I can work with. There is one difference that might cause an issue for USBOTG, there channels are limited (even non-interrupt) but each are independent (async-safe) and reconfigurable. Which is why I was looking at a reconfigure API for all channels (can still be type-state by consuming the channels); it seems like your RP2040 would support this (just changing values on the stack), and it would allow for the USBOTG to run without mutexes.

I will rebase to your branch, and if you're up for it we can collaborate on the final traits/interfaces.

ijager:
Good that there's more people working on this. I am happy to change things up. If we have three targets and use cases together that's great for testing. 

I am still on vacation for a week without computer so cannot really checks things right now.

nikvoid:
OK. For now I'm going to write a hub driver and test it on few keyboards.

i404788:
If you have a second RP2040, you could try the CDC-ACM, or MIDI drivers from embassy-usb in device mode (or just put it in usb flash mode).
If you happen to have another Raspberry Pi Zero or similar SBC you can use the [linux gadget for mass storage](https://www.kernel.org/doc/html/latest/usb/mass-storage.html). 

Isochronous isn't used much, mainly audio & video equipment, if you have an android device you might be able to configure it as a webcam (or potentially in audio accessory mode: https://source.android.com/docs/core/audio/usb#accessoryAudio).

nikvoid:
Now I have USB hub working for full speed devices, but not yet for low speed. 
If I simply set `preamble_en` register before transfer to LS device, RxTimeout interrupt will be fired. 
If that interrupt is disabled, it seems to work somehow. However INTERRUPT IN channel for keyboard is in dead silence. 
Maybe I forgot something important to set on hub, or maybe another keyboard is really slow, or both.

i404788:
Could be that you are missing the LS preamble on the interrupt channel, at least in USBOTG you need to set it per channel.

nikvoid:
Yeah, indeed. I overlooked a register for that. Now it works, except for timeouts. There may be another register to configure it, I'll research on that later.

nikvoid:
Timeouts appear to be a problem with specfic hub. Another hub is working fine.

i404788:
@nikvoid I've converted most of your hub driver to a [usb driver trait](https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb/src/handlers/mod.rs): https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb/src/handlers/hub.rs).

If you could review the general structure, especially `wait_for_event` since that's where I'm least sure about the interface.
I've also removed a lot of the dependencies from the driver itself, so the overarching layer can more easily be changed.

nikvoid:
Well, I don't see a reason to create device driver trait. I have updated rp/usb_host_hid_keyboard example with hub usage, as you can see, user can just try all drivers with some priorities, then create task for device. I think this is the very flexible way.

About handler trait: what purpose it serves? I see that it uses raw `UsbHostDriver` and channels. Could you please provide details about its usage? I also spotted several suspicious lines:

https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L115
On RP2040 there is a special register for interval, which is already used by driver. Even if it is not available on other platform, its driver could just use `Timer::after` as polyfill for periodic poll. 

https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L119 
Value returned by interrupt endpoint is a bitmap, where each bit corresponds to port, starting from 1st

https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L53
1st argument is not a device address, it's rather an endpoint address (and yes, this is redundant because control channel are always 0, FIXME)

i404788:
Thank you for the review.

> About handler trait: what purpose it serves?
The main purpose is to create a common interface for drivers such that a user can easily re-use them & create a common initialization sequence. This allows (for example) for automatic enumeration & initialization of drivers from a registry without mapping out what each driver needs to get started; this is effectively similar to how larger OSes do it.

> https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L115
I didn't know you had integrated the HW interrupt into the `request_in`, since it seems like it would do a manual request; I had suspected you got timeouts due to sending interrupt requests too frequently. I think we might want to make a separate interface for the interrupt request to avoid confusing with on-demand requests, but let me know your thoughts on that.

> https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L119

Ah yes I see what you mean, I misinterpreted your impl and tried to optimize; I'll fix that.

> https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/hub.rs#L53

Ditto, my mistake.

nikvoid:
> The main purpose is to create a common interface for drivers such that a user can easily re-use them & create a common initialization sequence. This allows (for example) for automatic enumeration & initialization of drivers from a registry without mapping out what each driver needs to get started; this is effectively similar to how larger OSes do it.

I think the main problem with this is that we are on embedded device without `Box` and `alloc`, and with statically sized... let's call them "device driver structs", we need some way to return these different structs from automatic initialization method. 
Correct me if I misunderstood what you mean.

> I didn't know you had integrated the HW interrupt into the `request_in`, since it seems like it would do a manual request; I had suspected you got timeouts due to sending interrupt requests too frequently. I think we might want to make a separate interface for the interrupt request to avoid confusing with on-demand requests, but let me know your thoughts on that.

Usb is host-centric, and even interrupts are not interrupts, but rather poll-based things, so I though it would be appropriate name. We could just state about this behaviour in docs and/or rename method

i404788:
While I agree we should assume no-alloc for any infra code, I do regularly use alloc in project development, with the traits you now have the options. Functionally, I don't think your struct pattern is much worse; just that it seems to me most drivers don't need to assume it will run on embassy but rather any async rust would have the equivalent code; so a trait would allow for more portability and for 3rd party drivers that neatly integrate (similar to `usbh` with their `Driver` trait, or embedded-hal). 

A few things I was planning to build based on the trait was a faster enumeration filter using the static spec, also with the new structure of the hub you don't need a mutexed device registry (or no registry at all just unique address assignment).

Btw the only reason I'm calling them `handlers` is because driver was already used to describe the `UsbHostDriver`  and gadget-side [`Driver`](https://github.com/nikvoid/embassy/blob/dbbd60c0260f9c85fb12327d8e552762cac8d554/embassy-usb-driver/src/lib.rs#L118) traits; but it seems like [`Handler`](https://github.com/i404788/embassy/blob/175dc5771ead4038029d2406705b3cc2ebd87ec7/embassy-usb/src/lib.rs#L88C11-L88C18) is also already a trait so maybe it should just be `UsbDriver` or `Driver` and accept the potential confusion.

> even interrupts are not interrupts, but rather poll-based things, so I though it would be appropriate name

If you configure HW interrupt polling you may still manually poll, at least on USBOTG (maybe that's a unnecessary use-case to consider). If we only allow hw polling I think just documenting the behavior around this would be good enough. (I do agree that the HostDriver should ensure software polling works the same as hw polling with polyfills)

nikvoid:
So, if I finally understand correctly, handler trait will be something like a middleware between device driver and usb host driver (or just device driver that stands directly over bus driver).

And I have more questions:
https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/mod.rs#L67
This also not the case for rp2040, so handler probably would need to impl Drop.

https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/mod.rs#L69
Currently all UsbHostDriver methods take `&self`, so mut is not necessary. As mentioned in previous question, dropping channels is a handler's responsibility, so it needs to store reference to UsbHostDriver anyway.

https://github.com/i404788/embassy/blob/fb70372f52f3615e92863ef01afcb5d176e5d4c5/embassy-usb/src/handlers/mod.rs#L87
UsbResumableHandler is needed to talk with more devices when channels are limited, right?

i404788:
>  This also not the case for rp2040, so handler probably would need to impl Drop.

Hmm I was hoping this could be done in the Channel itself to avoid complicating the trait & implementation; e.g. give a Ref<AtomicU32>, and a it's own index, which it can use to xor to mark it is cleared & clear the registers needed if it's an hw interrupt pipe.

> Currently all UsbHostDriver methods take &self, so mut is not necessary. As mentioned in previous question, dropping channels is a handler's responsibility, so it needs to store reference to UsbHostDriver anyway.

True I suppose it could be `&self`, then the assumption is that we have internal mutability (in our case usually registers, but if potentially in other drivers separate data needs to be managed) that the driver has to manage.

>  UsbResumableHandler is needed to talk with more devices when channels are limited, right?
Correct, so at a higher layer it can be decided to suspend a device in order to start/resume a different device.

nikvoid:
> Hmm I was hoping this could be done in the Channel itself to avoid complicating the trait & implementation; e.g. give a Ref<AtomicU32>, and a it's own index, which it can use to xor to mark it is cleared & clear the registers needed if it's an hw interrupt pipe.

Or just use static variable, shared with UsbHostDriver, as it is a singleton anyway. Yes, this seems to be simpler approach. I wrapped UsbHostHost::Channel into struct for 2 reasons: first is Drop, second is disconnect detection, because inner channel could only detect root device disconnect.

ijager:
Next week I have some time to look into this. What is the current status? Anything I can do to help move this forward?

i404788:
I think the traits are settled now (at least in concept, feedback is still welcome) found here: https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb-driver/src/host.rs these are slightly changed to from @nikvoid's version based on the discussion. I have also refactored some of the types around embassy-usb* to remove duplicate implementations.

I've made a [keyboard](https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb/src/handlers/kbd.rs) & [hub](https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb/src/handlers/hub.rs) implementation for the [handler/driver trait](https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb/src/handlers/mod.rs)

I'm currently working on implementing and testing an [USBOTG implementation](https://github.com/i404788/embassy/blob/usb-asynch/embassy-usb-synopsys-otg/src/host.rs) using an esp32s3.

The final result for an end-user looks like this (simplified):
https://github.com/i404788/esp-hal/blob/937faa6e0993edbbe9135d1326e3a66074b9f01a/examples/src/bin/embassy_usb_host.rs#L77-L98

ijager:
I am trying to use your branch in my project but run in a few issues.

1. Atomics are not really supported on Cortex M0 (stm32g0)
2. Even though stm32g0 does not use`embassy-usb-synopsys-otg`, it is still compiled. Since it is not optional in Cargo.toml of `embassy-stm32`. Until now this didn't result in problems, but now with the atomics this does ont compile anymore. 

```
error[E0599]: no method named `compare_exchange_weak` found for struct `AtomicU32` in the current scope
  --> /Users/ijager/.cargo/git/checkouts/embassy-f415cad553f7ae41/3cb79bf/embassy-usb-synopsys-otg/src/host.rs:44:18
   |
43 |               if ALLOCATED_PIPES
   |  ________________-
44 | |                 .compare_exchange_weak(
   | |                 -^^^^^^^^^^^^^^^^^^^^^ method not found in `AtomicU32`
   | |_________________|
   |

error[E0599]: no method named `fetch_and` found for struct `AtomicU32` in the current scope
  --> /Users/ijager/.cargo/git/checkouts/embassy-f415cad553f7ae41/3cb79bf/embassy-usb-synopsys-otg/src/host.rs:65:25
   |
65 |         ALLOCATED_PIPES.fetch_and(!(1 << self.channel_idx), core::sync::atomic::Ordering::AcqRel);
   |                         ^^^^^^^^^ method not found in `AtomicU32`

error[E0599]: no method named `fetch_nand` found for struct `AtomicU32` in the current scope
   --> /Users/ijager/.cargo/git/checkouts/embassy-f415cad553f7ae41/3cb79bf/embassy-usb-synopsys-otg/src/host.rs:321:29
    |
321 |             ALLOCATED_PIPES.fetch_nand(!(1 << self.channel_idx), core::sync::atomic::Ordering::AcqRel);
    |                             ^^^^^^^^^^ method not found in `AtomicU32`

error[E0599]: no method named `compare_exchange_weak` found for struct `AtomicU32` in the current scope
   --> /Users/ijager/.cargo/git/checkouts/embassy-f415cad553f7ae41/3cb79bf/embassy-usb-synopsys-otg/src/host.rs:741:22
    |
740 |                   if ALLOCATED_PIPES
    |  ____________________-
741 | |                     .compare_exchange_weak(
    | |                     -^^^^^^^^^^^^^^^^^^^^^ method not found in `AtomicU32`
    | |_____________________|
```

i404788:
@ijager For 2 I added an feature to `embassy-usb-synopsys-otg` since it also currently requires `alloc` (for DMA alignment reasons), but forgot to remove it from default features. It should be fixed for you in the latest commit.

For missing atomics I'm not sure what the best solution is there; I'm assuming you only have a single core then so you might not need atomics. If that's the only obstacle I can take a look once I'm done with validation of the USBOTG, so far everything works as expected but not everything is handled/supported yet.

ijager:
If I understand correctly. I can alloc a `Bulk` or `Interrupt` channel. But I can still use that channel any time to do a control request on `endpoint 0`?

i404788:
No to do a control request you'll need to allocate a `Control` channel, however you can multiplex them under the hood. In both the RP2040 & USBOTG (current impl, in a different mode it could be concurrent) core we share control channels under the hood for control requests and just wait using a mutex-like structure before configuring and transmitting/receiving. 
In USBOTG all non-control channels are dedicated in hardware. In RP2040 (iirc) only interrupt channels are special, all others are shared. 

For example USBOTG control requests look like:
```rust
    async fn control_in(&mut self, setup: &SetupPacket, buf: &mut [u8]) -> Result<usize, ChannelError>
    where
        T: channel::IsControl,
        D: channel::IsIn,
    {
        let _ = SharedChannelGuard::try_claim(0).await; // 'Locks' channel 0
        self.configure_for_endpoint(Some(embassy_usb_driver::Direction::Out));
        ...
  }
```
where `SharedChannelGuard::try_claim` will poll until there are no requests pending on channel 0, since channel 0 is shared across instances. RP2040 has something similar but for all channels shared.

Since each control 'channel' may refer to a different device (with different max_packet_size, speed, etc) we separate them into different structs.

For you can test without locking channels since it only comes up if you start spawning multiple tasks each doing requests or using `select!` on multiple requests.

ijager:
Thanks for the explanation. I totally overlooked the `where T: channel::IsControl` clause. It makes sense now!

In the `Drop` implementation for `OtgChannel<T, D>` it is using `fetch_nand` while in the `drop` of `SharedChannelGuard` `fetch_and` is used. Is this intentional? Seems like all bits will be flipped in the `fetch_nand` case?

```rust
        if self.channel_idx != 0 {
            ALLOCATED_PIPES.fetch_nand(!(1 << self.channel_idx), core::sync::atomic::Ordering::AcqRel);
        }
```

i404788:
That seems to be a typo yes, it's supposed to be `fetch_and`, I'll fix that.

ijager:
Yay I managed to get a device to enumerate with my refactored stm32 HAL! I like how it works with the extended traits that give a lot of the high level logic for free.

I do have an interesting edge case here. The device is an RFID badge reader that implements HIDKeyboard class, but rather poorly, because this device does not respond to SET_PROTOCOL and SET_IDLE commands. It just hangs. 

This is why I implemented `TransferOptions` with a configurable timeout in the host layer before your refactor. I think we should reintroduce something like that to be able to deal with unresponsive devices. Or devices that just don't implement the full spec of a device class.

```
1.003082 DEBUG TCPP03-M20 Value 1: 1C
└─ usb_host_hid_keyboard2::____embassy_main_task::{async_fn#0} @ src/bin/usb_host_hid_keyboard2.rs:115 
1.103668 DEBUG Detecting device
└─ usb_host_hid_keyboard2::____embassy_main_task::{async_fn#0} @ src/bin/usb_host_hid_keyboard2.rs:124 
10.576690 TRACE USB IRQ: device connect/disconnect
└─ embassy_stm32::usb::_version::usb_host::{impl#0}::on_interrupt @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.577148 <lvl> Found device with speed = Full
└─ usb_host_hid_keyboard2::____embassy_main_task::{async_fn#0} @ src/bin/usb_host_hid_keyboard2.rs:133 
10.577636 TRACE Bus reset
└─ embassy_stm32::usb::_version::usb_host::{impl#5}::bus_reset::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.628204 TRACE retarget_channel: addr: 0 ep_type: Control index: 0
└─ embassy_stm32::usb::_version::usb_host::{impl#4}::retarget_channel @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.628875 TRACE retarget_channel: addr: 0 ep_type: Control index: 0
└─ embassy_stm32::usb::_version::usb_host::{impl#4}::retarget_channel @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.629516 TRACE [enum] Attempting to get max_packet_size for new device
└─ embassy_usb::host::ControlChannelExt::enumerate_device::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.630340 TRACE READ DONE, rx_len = 8
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.631042 TRACE Descriptor embassy_usb::host::descriptor::DeviceDescriptorPartial: [18, 1, 16, 1, 0, 0, 0, 64]
└─ embassy_usb::host::ControlChannelExt::request_descriptor::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.633789 TRACE [enum] got max packet size for new device 64, attempting to set address
└─ embassy_usb::host::ControlChannelExt::enumerate_device::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.634552 TRACE READ DONE, rx_len = 0
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.637207 TRACE [enum] Finished setting address
└─ embassy_usb::host::ControlChannelExt::enumerate_device::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.637542 TRACE retarget_channel: addr: 1 ep_type: Control index: 0
└─ embassy_stm32::usb::_version::usb_host::{impl#4}::retarget_channel @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.638549 TRACE READ DONE, rx_len = 18
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.639343 TRACE Descriptor embassy_usb::host::descriptor::DeviceDescriptor: [18, 1, 16, 1, 0, 0, 0, 64, 217, 4, 3, 21, 16, 3, 1, 2, 0, 1]
└─ embassy_usb::host::ControlChannelExt::request_descriptor::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.642242 TRACE Device Descriptor: DeviceDescriptor { len: 18, descriptor_type: 1, bcd_usb: 272, device_class: 0, device_subclass: 0, device_protocol: 0, max_packet_size0: 64, vendor_id: 1241, product_id: 5379, bcd_device: 784, manufacturer: 1, product: 2, serial_number: 0, num_configurations: 1 }
└─ embassy_usb::host::ControlChannelExt::enumerate_device::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.643829 TRACE READ DONE, rx_len = 9
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.644531 TRACE Descriptor embassy_usb::host::descriptor::ConfigurationDescriptor: [9, 2, 41, 0, 1, 1, 0, 128, 80]
└─ embassy_usb::host::ControlChannelExt::request_descriptor::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.648376 TRACE READ DONE, rx_len = 41
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.649414 TRACE Full Configuration Descriptor [41]: [9, 2, 41, 0, 1, 1, 0, 128, 80, 9, 4, 0, 0, 1, 3, 1, 1, 10, 9, 33, 1, 1, 0, 1, 34, 63, 0, 7, 5, 129, 3, 64, 0, 10, 7, 5, 1, 3, 64, 0, 10]
└─ embassy_usb::host::ControlChannelExt::enumerate_device::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:137 
10.652130 TRACE READ DONE, rx_len = 0
└─ embassy_stm32::usb::_version::usb_host::{impl#3}::read_data @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.653137 TRACE retarget_channel: addr: 1 ep_type: Interrupt index: 1
└─ embassy_stm32::usb::_version::usb_host::{impl#4}::retarget_channel @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.653900 TRACE retarget_channel: addr: 1 ep_type: Control index: 0
└─ embassy_stm32::usb::_version::usb_host::{impl#4}::retarget_channel @ /Users/ijager/dev/telecom/embassy/embassy-stm32/src/fmt.rs:124 
10.654541 DEBUG [kbd]: Setting PROTOCOL & idle
└─ embassy_usb::handlers::kbd::{impl#1}::try_register::{async_fn#0} @ /Users/ijager/dev/telecom/embassy/embassy-usb/src/fmt.rs:151 

... hangs here
```

i404788:
Yeah timeouts are definitely needed, in USBOTG it's in the hardware and not configurable (unless I missed a register).

The spec says:
```
The device bus turn-around time is defined by the worst case round trip delay plus the maximum device
response delay (refer to Sections 7.1.18 and 7.1.19 for specific bus turn-around times). If a response is not
received within this worst case timeout, then the transmitter considers that the packet transmission has
failed.
```

So you should be able to define a constant amount of time (per device speed setting) which is the worst-case timeout.

Also does your device require SET_PROTOCOL/SET_IDLE and ignore it, or does it not support it? A device is supposed to give a STALL response to indicate a request is not supported.

nikvoid:
RP2040 also has hardware timeouts, but they are apparently too strict for slow devices (like that usb hub I first encountered this issue with). What about adding method to bus trait and maybe individual channels, like `set_timeout` or so?

i404788:
Hmm I don't think USBOTG can implement a custom timeout at all; any transaction error that gives any feedback will also halt the channel. I think that's because the timeout defined by the spec is already quite generous even for older devices, and the protocol supports notifying of not-ready (NYET/NAK). I've also encountered a few devices that require handling NAKs when going fast but not yet one that actually times out.

For the RP2040 I checked the datasheet, `RX_TIMEOUT` seems like it would be triggered in case of NAK response as well so it's possible the device responded and requested to try again later (busy). @nikvoid could you check if this is the case for your slow hub?

I'm okay adding a fail-safe timeout, but I don't think it would be good to ignore the specs' timeout if it can be avoided.

ijager:
> Also does your device require SET_PROTOCOL/SET_IDLE and ignore it, or does it not support it? A device is supposed to give a STALL response to indicate a request is not supported.

It just doesn't respond at all to the request setup packet. No stall. Continuing with a new, different request will make it respond again. So I guess it is a device with a bad implementation.

Reading the USB 2.0 spec I see the following about timeouts:

- total maximum timeout: 5sec. That is also what i have seen MacOS USB Host doing for this particular device when it does not respond, looking at the logic analyzer.
- SetAddress: maximum `50ms`
- Standard Requests: `50ms` without Data stage, `500ms` with data stage.

I guess we can consider SetAddress a "Standard request without Data Stage". 

I agree with having a default timeout (e.g. 50ms/500ms for low and full speed), and a UsbChannel method `set_timeout` to override it.

Let me know if I am missing anything.

ijager:
@i404788, I have included all latest changes of your branch and have now a working example for STM32G0 based on the new traits.

Regarding the `async fn wait_for_device_event(&self) -> embassy_usb_driver::host::DeviceEvent` API, It might be better to actually pass in the event you're expecting. Because currently it is not fully defined what to wait for. 

If a device is already connected and then I call `wait_for_device_event(&self)`, expecting to get a `Connected` Event, I will never get it. If we pass in the `Connected` event then we can immediately return if the device was already connected.

Perhaps as an optional, so

```rust 

// either 

async fn wait_for_device_event(&self, expected_event: host::DeviceEvent) ->host::DeviceEvent

// or

async fn wait_for_device_event(&self, expected_event: Option<host::DeviceEvent>) ->host::DeviceEvent
```

In passing in `expected_event = None` we could just wait for any future event. But currently only `Connected` and `Disconnected` exist.

ijager:
Hmm also not ideal as we have to pass in the Device Speed, which is dummy:

```rust
    debug!("Detecting device");
    // Wait for root-port to detect device
    let speed = loop {
        match usbhost.wait_for_device_event(Connected(USBSpeed::Full)).await {
            Connected(speed) => break speed,
            _ => {}
        }
    };
```

i404788:
> It just doesn't respond at all to the request setup packet. No stall. Continuing with a new, different request will make it respond again.

Hmm that does sound strange, it could also be a bad request I recently got this with USBOTG as well. For example if the CRC is incorrect the device will just not respond.  

> default timeout (e.g. 50ms/500ms for low and full speed), and a UsbChannel method set_timeout to override it.

Sounds good, I'll see if I can get a decent interface around it for the configuration.

> Regarding the async fn wait_for_device_event(&self) -> embassy_usb_driver::host::DeviceEvent API, It might be better to actually pass in the event you're expecting. Because currently it is not fully defined what to wait for.

I can see your reasoning here, but `wait_for_device_event` needs to serialize the events rather (i.e. Disconnected can only be returned if Connected was returned before, vice-versa), as I think ignoring certain events may cause additional unsoundness. Disconnects always need to be handled since channels and drivers will be invalidated (or at least made non-functional).
I'm open to other suggestions on how to improve it, because I agree it's not as intuitive as it could be.

ijager:
You are right, the driver can be a state machine. Then we could make sure we always get `connect` before `disconnect`.

For unexpected disconnects, those are usually noticed because of failed channel operations/transfers. Is the idea to call `wait_for_device_event()` after detecting a disconnect due to a `ChannelError`?

Currently my USB task looks like this:

```rust
 loop {
        // Wait for device connected
        usbhost.wait_for_device_event().await;
        // do enumeration etc ...
        let Ok(mut kbd) = KbdHandler::try_register(&usbhost, enum_info).await else {
            ...
            break;
        };

        // Read device endpoints until disconnect
        loop {
            match kbd.wait_for_event().await {
                Ok(HandlerEvent::HandlerEvent(KbdEvent::KeyStatusUpdate(ev))) => {
                ...
                }
                Err(HostError::ChannelError(ChannelError::Disconnected)) | Ok(HandlerEvent::HandlerDisconnected) => {
                    info!("Device disconnected");
                    break;
                }
                other => warn!("Unhandled event: {:?}", other),
            }
        }
    // kbd is dropped here
    }
```

So I don't really need to wait_for_disconnect_event, as the `break` will make sure everything is dropped. What is your usecase for disconnect, do you have a different task that handles `DeviceEvent`s?

nikvoid:
> For the RP2040 I checked the datasheet, `RX_TIMEOUT` seems like it would be triggered in case of NAK response as well so it's possible the device responded and requested to try again later (busy). @nikvoid could you check if this is the case for your slow hub?
> 
> **EDIT:** it seems like RP2040 unlike USBOTG does automatic retry given `NAK_POLL` register, but I'm still curious to know if it's getting a NAK or if it's simply not replying since the spec timeouts for NAKs are really long: https://beyondlogic.org/usbnutshell/usb6.shtml#SetupPacket.

I tried to take a closer look into this issue.
I captured communication through faulty hub with logic analyzer, and it seems like keyboard behind the hub doesn't respond with ACKs, or hub drops them. No NAKs, nothing. After 1ms hardware repeats the request and sometimes gets ACK, then software passes the partial descriptor request, but fails in next ones, as they only have single try attempt. 
The average retry count to get partial device descriptor is 5.

embassy-usb/src/host/descriptor.rs
```diff
@@ -0,0 +1,591 @@
...
+        endpoints
+    }
+
+    // Returns the offset to the first matching descriptor in the slice
+    fn identify_descriptor<T: USBDescriptor>(slice: &[u8]) -> Option<usize> {
+        let mut offset = 0;
+        let mut desc_len = slice[offset] as usize;
+        let mut desc_type = slice[offset + 1];
+
+        while desc_type != T::DESC_TYPE || desc_len != T::SIZE {
```
jakerr, line 367: Only accepting descriptors that are the exact size expected as is done in this line would make the descriptor fail to parse if there were existance of any extentions to the descriptors such as  those that exist in USB 2.0+.

For example MIDI 2.0 DeviceEndpoints have can potentially have two extra bytes in positions 7 and 8 and so the DeviceEndpointDescriptor bLength may be 9 rather than 7.

| Offset | Field         | Size | Value | Description |
|--------|---------------|------|-------|-------------|
| 0-6  |     |     |   | (same as in USB 1.1)      |
| 7      | `bRefresh`    | 1    | 0x00  | Unused      |
| 8      | `bSynchAddress` | 1  | 0x00  | Unused      |

What this means is in practice if you assume exact sizes, newer devices will be ignored by this driver if they use any extensions even though they should be backwards compatible.

The assertion that bytes beyond the expected size must be ignored is mentioned briefly in the USB spec in Chapter 9.5 "Descriptors":
> If a descriptor returns with a value in its length field that is less than defined by this specification, the
descriptor is invalid and should be rejected by the host. If the descriptor returns with a value in its length field that is greater than defined by this specification, the extra bytes are ignored by the host, but the next
descriptor is located using the length returned rather than the length expected.

This would also be relevant to `parse_endpoints` (and other descriptor parsing logic above) which assumes the record length when advancing the working_buffer slice `working_buffer = &working_buffer[EndpointDescriptor::SIZE..];` as that could break if some bytes in the extended area happened to collide with a valid start to a descriptor.

I think it would be best to refactor the code to always advance the slice being considered by the amount specified by `bLength` of the current slice, and never assume that it's safe to advance by a static amount.

jakerr: A sent an attempted fix as another PR targeting this PR branch. Please take a look :)

jakerr:
This branch currently has quite a few commits mixed in that are duplicates of things that have already been merged into main that I think make it a bit unwieldy to review.

I tried my hand at rebasing it in my fork's [usb-host](https://github.com/jakerr/embassy/tree/usb-host) that did require a couple conflict resolutions but hopefully I got them right.
Just leaving it here for the authors in case it's helpul.

JYouren:
Would this work support (virtual) serial ports easily?

i404788:
@JYouren Yes, one of the drivers I made for myself is very similar to USB Serial (technically mine is a printer port but works largely the same), and it's quite easy to implement.

JYouren:
Awesome, you wouldn't be able to sign post me to that would you so I can have a look?

i404788:
It isn't public but here are the important parts:

* Make a struct with all desired/required channels
* Implement `UsbHostHandler` to lookup & verify the endpoints of those channels for a new device
* Add your own functions related to the specific function of the device (i.e. `try_write`, `try_read`)

The main difference for the serial version will be the InterfaceDescriptor matching, the rest can likely stay the same.

The you can use it similarly to how I've used it here for kbd (note it does not handle disconnects or use any custom functions): https://github.com/i404788/esp-hal/blob/719d96101174fcfbd718b6bdb0177fe65d46a86f/examples/src/bin/embassy_usb_host.rs#L75-L97

<details>
<summary>Code:</summary>

```rust
pub struct Handler<H: UsbHostDriver> {
    bulk_in_channel: H::Channel<channel::Bulk, channel::In>,
    bulk_out_channel: H::Channel<channel::Bulk, channel::Out>,
    control_channel: H::Channel<channel::Control, channel::InOut>,
}

impl<H: UsbHostDriver> Handler<H> {
    pub async fn try_read<'a>(&mut self, buf: &'a mut [u8]) -> Result<&'a [u8], ChannelError> {
        let actual_read = self.bulk_in_channel.request_in(buf).await?;
        Ok(&buf[..actual_read])
    }

    pub async fn try_write(&mut self, buf: &[u8]) -> Result<(), ChannelError> {
        match select::select(
            Timer::after_millis(200),
            self.bulk_out_channel.request_out(buf),
        )
        .await
        {
            select::Either::First(_) => Err(ChannelError::Timeout),
            select::Either::Second(res) => {
                res?;
                Ok(())
            }
        }
    }
}

impl<H: UsbHostDriver> UsbHostHandler for Handler<H> {
    type PollEvent = ();
    type Driver = H;

    fn static_spec() -> StaticHandlerSpec {
        StaticHandlerSpec::new(None)
    }

    async fn try_register(bus: &H, enum_info: EnumerationInfo) -> Result<Self, RegisterError> {
        println!("Trying to register");
        for iface in enum_info.cfg_desc.iter_interface() {
            debug!("iface: {}", iface);
        }

        let iface = enum_info
            .cfg_desc
            .iter_interface()
            .find(|v| {
                matches!(
                    v,
                    InterfaceDescriptor {
                        interface_class: 0x07,
                        interface_subclass: 0x1,
                        interface_protocol: 0x2,
                        ..
                    }
                )
            })
            .ok_or(RegisterError::NoSupportedInterface)?;

        debug!("Found matching iface!");

        let interrupt_ep = iface
            .iter_endpoints()
            .find(|v| v.ep_type() == EndpointType::Interrupt && v.ep_dir() == Direction::In)
            .ok_or(RegisterError::NoSupportedInterface)?;

        debug!("Found matching intr ep");

        let bulk_in_ep = iface
            .iter_endpoints()
            .find(|v| v.ep_type() == EndpointType::Bulk && v.ep_dir() == Direction::In)
            .ok_or(RegisterError::NoSupportedInterface)?;

        debug!("Found bulk in ep");

        let bulk_out_ep = iface
            .iter_endpoints()
            .find(|v| v.ep_type() == EndpointType::Bulk && v.ep_dir() == Direction::Out)
            .ok_or(RegisterError::NoSupportedInterface)?;

        debug!("Found all endpoints");

        let bulk_in_channel = bus.alloc_channel(
            enum_info.device_address,
            &bulk_in_ep.into(),
            enum_info.ls_over_fs,
        )?;

        let bulk_out_channel = bus.alloc_channel(
            enum_info.device_address,
            &bulk_out_ep.into(),
            enum_info.ls_over_fs,
        )?;

        let control_channel = bus.alloc_channel(
            enum_info.device_address,
            &EndpointInfo::new(
                0.into(),
                EndpointType::Control,
                (enum_info.device_desc.max_packet_size0 as u16)
                    .min(enum_info.speed.max_packet_size()),
            ),
            enum_info.ls_over_fs,
        )?;

        Ok(Handler {
            control_channel,
            bulk_in_channel,
            bulk_out_channel,
        })
    }
}
```

</details>

mikeemoo:
It this still under development? I'd really love USB host support

deluxetiky:
> It this still under development? I'd really love USB host support

Actually we are also waiting for USB host support! If we know latest state, it would be great.

Codetector1374:
@Dummyc0m and I are trying to create an implementation here: https://github.com/vapor-keeb/async-usb-host/

very early stage, still fighting with a lot of design challenge with how to support hot plug and unplug multiple (kinds) of devices.

ijager:
I am also using this branch in a completed project since end of last year. I will probably get back to this as the client has asked for HUB support as well, but not in the next few months. So totally fine to open a new PR and then I can close this one.

AlexCharlton:
I've also spun off [a branch](https://github.com/embassy-rs/embassy/compare/main...AlexCharlton:embassy:usb-host?expand=1) to fix some issues present in this one (but otherwise, I'm very appreciative of the work that was put in here!)

While there are a number of commits in the branch, there are really only three issues to speak of. 

First, `EndpointIterator` is broken if you have more than one endpoint. [This is a fix](https://github.com/embassy-rs/embassy/commit/15344eb588970dfa6b40bf7fcaaedf34d2630a4e), but I would say there's more that could be done for descriptor traversal. `InterfaceIterator` does not find alternate interfaces, nor does it correctly capture its `buffer` beyond the first one (though I'd say the concept of an interface buffer is a bit fraught thanks to the existence of alternate interfaces). Some additional ways of iterating through descriptors would also be useful. Otherwise, handlers need to do their own parsing of the `ConfigurationDescriptor` buffer if they want to access any class-specific descriptors. This additional functionality could readily be done in future PRs, though, so I wouldn't say it's crucial for the first pass.

Second, assuming I haven't misunderstood something about how USB works (which could well be the case) I don't think it's possible for `wait_for_device_event` to accurately return the bus speed of high speed devices upon a connection _without first issuing a reset_. [This commit](https://github.com/embassy-rs/embassy/commit/0a39f18b27202c9cc1d8f142ef4199b50bbf5a64) documents this fact and removes the call to `bus_reset` from `enumerate_root`, since there should be no need to do so at that point. What I have _not_ done in this branch is square up the existing implementations of `bus_reset`.

Third, with only 256 bytes provided, it's not hard to exceed the configuration buffer size, causing a panic. [This commit](https://github.com/embassy-rs/embassy/commit/f5a83658bb4382280a2c9c5688523ece699180f1) bumps the size to 512 bytes, which happens to be large enough for the device I'm testing with, but some configurability would probably be good. I've also added a guard, so that we get an error rather than a panic.

i404788:
@AlexCharlton the first issue seems to have to do with the alignment of the endpoints (if the enpoints are aligned you won't encounter the issue, which likely was in our cases). You're right that this general area could be expanded upon to support more complex device configurations.

For point 2, I believe it depends on the hardware IP used, iirc USBOTG will do the speed detection for you. I could be wrong here as well, but detection worked correctly as far as I've tested with esp-hal, it's possible a redundant `bus_reset` slipped in.

For the configuration buffer, it seems like they can be up to 65535 bytes (`total_len`/`wTotalLength`), so I think we'll have to make some gradations on the size configs to not blow up the stack.

I appreciate your testing, I'll collate & add your improvements into the final PR.

i404788:
I've opened a new PR at https://github.com/embassy-rs/embassy/pull/4032. 

@AlexCharlton I've refactored the Descriptor parsing (also integrating @jakerr's test cases) & SetConfiguration procedure, so now the user or handler can choose the configuration through `EnumerationInfo` and can provide their own arbitrarily long buffer for it. The example handlers they still use the `512` limit to read the active or default config. Let me know what you think.

AlexCharlton:
> I've opened a new PR at #4032.
> 
> @AlexCharlton I've refactored the Descriptor parsing (also integrating @jakerr's test cases) & SetConfiguration procedure, so now the application or handler can choose the configuration through `EnumerationInfo` and can provide their own arbitrarily long buffer for it. The example handlers still use the `512` limit to read the active or default config. Let me know what you think.

Makes sense to me!

ijager:
Closed in favor of #4032

PR #3310, opened 2024-09-04
stm32: Add a first Lptim driver

Ecco:
Does *not* handle all the fields just yet, but does come with a PWM driver that works just like SimplePWM.


Ecco:
Fixed to use the GLOBAL interrupt instead of UP, as per the embassy-rs/stm32-data#520

Dirbaio (approved):
thank you!

PR #3312, opened 2024-09-05
Enable critical-section/std on wasm

9SMTM6:
Without that feature one will find import errors on opening the webpage, that are hard to debug. The feature was indirectly enabled in the wasm example, however the reason wasn't documented and thus it was easy to miss.


lulf (approved):
Thanks!

Issue #2994, opened 2024-05-23
Support sharing USB endpoints between alternative settings of an interface 


Dominaezzz:
At the moment the `embassy-usb` only supports allocating an endpoint per interface alternative setting.

However, in section 9.6.3 of the USB Spec Rev 2.0

> An endpoint is not shared among interfaces within a single configuration unless the endpoint is used by alternate settings of the same interface.

The USB spec supports sharing the endpoint between the alternative settings of an interface, so a device with limited number of endpoints can provide multiple classes where each class only gets exclusive access at a time.

I haven't looked too deeply into how this should be exposed or how users would consume this but I'm just expressing the limitation here as it'll likely involve a breaking change to implement.

What I do wonder though, is whether alternative settings *should* share endpoints or if they just *can*.

sourcebox:
This is also necessary for USB audio when supporting multiple formats (e.g. sample rates) because each requires an alternate setting.

M3gaFr3ak:
@Dominaezzz I couldn't find any comment on that in the USB standard. But it is imperative for Endpoints to be independent of alternate settings. Each alternative setting on an interface is mutually exclusive to each other alternate setting, so an endpoint will be unused and unusable if it is exclusive to an alternate setting which is currently not in use.
It is *possible* to work with this, but it's a waste of endpoints, which are limited in number varying by device.

PR #3314, opened 2024-09-05
Add ISO endpoint support

elagil:
Adds support for ISO endpoints for both the STM32 OTG USB driver and the "regular" (newer?) USB driver.

### DIFF
--- a/embassy-stm32/src/usb/usb.rs
+++ b/embassy-stm32/src/usb/usb.rs
@@ -80,6 +80,8 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
 
         if istr.ctr() {
             let index = istr.ep_id() as usize;
+            CTR_TRIGGERED[index].store(true, Ordering::Relaxed);
+
             let mut epr = regs.epr(index).read();
             if epr.ctr_rx() {
                 if index == 0 && epr.setup() {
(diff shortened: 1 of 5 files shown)


elagil:
For anyone who is reviewing this, find the USB packet buffer layouts that I used. They are relevant for the functions that read/write from/to the `btable`. The values in the `btable` are the addresses of the software packet buffers, and the received/transmitted buffer sizes.

For double-buffered endpoints, both RX and TX count and address fields are used for THE SAME direction (either RX or TX, never both). I hope that is clear from the comments in the code as well.

From RM0481 (STM32H5)
image

From RM0091 (STM32F0x1)
image

M3gaFr3ak:
This is a very important PR to make embassy usable for any audio-related USB stuff. I'm currently implementing USB Audio 2.0 and thanks to this PR, I have a successfully enumerating and streaming device!
When I'm finished, I'll try to PR an example.

elagil:
@M3gaFr3ak I agree, that is what I needed it for. Very curious to see your UAC 2.0 implementation. I made UAC 1.0 in https://github.com/embassy-rs/embassy/pull/3212

Dirbaio (approved):
thank you! looking great

M3gaFr3ak:
@elagil This is the test implementation I'm currently doing: https://github.com/M3gaFr3ak/rp-usb-uac2
It's only using full speed, but that's enough for demoing. I don't actually do proper handling, it currently enumerates and the endpoints receive data. I've also modified embassy-usb to allow reuse of endpoints https://github.com/embassy-rs/embassy/issues/2994
I'll continue working on this until i have a loopback device working (speaker to mic)

PR #3317, opened 2024-09-06
embassy-stm32: Add SimplePwmChannel

GrantM11235:


embassy-stm32/src/timer/simple_pwm.rs
```diff
@@ -89,26 +194,80 @@ impl<'d, T: GeneralInstance4Channel> SimplePwm<'d, T> {
...
+    pub fn ch4(&mut self) -> SimplePwmChannel<'_, T> {
+        self.channel(Channel::Ch4)
+    }
+
+    /// Splits a [`SimplePwm`] into four pwm channels.
+    ///
+    /// This returns all four channels, including channels that
+    /// aren't configured with a [`PwmPin`].
+    // TODO: I hate the name "split"
+    pub fn split(&mut self) -> SimplePwmChannels<'_, T> {
```
Dirbaio, line 248: with this signature, the split channels still borrow from `self`, so you can't easily create a PWM in main then pass ownership of some channels to TaskA and others to TaskB. You'd have to put the PWM in a StaticCell to get the lifetimes right.

Instead the signature could be `fn split(self) -> SimplePwmChannels<'d, T>`. This way the unsplit PWM is simply gone, and the channels can live independently. It should be doable by unsafely duplicating the `Timer`, or making the channels contain a pointer to PAC regs.

GrantM11235: I can't seem to unsafely clone the timer PeripheralRef, there is a missing trait bound somewhere.

Aside from that, it is a bit unfortunate that splitting the channels means that the timer will never be disabled, even if all the channels are dropped.

Ideally, the channels would be refcounted with bit flags, which could also make it easier to hand out channels without accidentally duplicating channels. That sounds pretty complex though, so I am going to leave it as a "future possibility" for now.

Dirbaio: yeah leaving it for future work sounds good to me.

Dirbaio (approved):
lgtm! can you fix CI?

GrantM11235:
I had to add a `Peripheral<P = Self>` bound to the timer `SealedInstance` trait, I hope that is correct

Issue #3313, opened 2024-09-05
Using embassy primitives in webworkers


9SMTM6:
Hey there! Its amazing that you also offer an async runtime with synchronization primitives etc. for the web. And it works well enough, if you use it from the main thread.

However I did notice that most primitives don't seem to work if you use them from a webworker.

That is a rather big ask, and if you think that is out of scope, that's fine, but I thought I'd at least let you know of this usecase.

I made a reproduction [here](https://github.com/9SMTM6/embassy_webworker_issue_reproduction).

I'm very much aware that it might not be possible to make all primitives work in webworkers.
However many WebAPIs are also available in webworkers, so I think that many primitives should be possible if one is a bit careful which concrete API they use.

Dirbaio:
What web APIs is Embassy calling that aren't available in webworkers?

`embassy-sync` uses "just standard async Rust", so I'd say the issue is not there. It's probably in the [executor implementation for wasm](https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/arch/wasm.rs) which does call JS apis to get itself polled from the JS event loop.

9SMTM6:
@Dirbaio This is only the first thing that made errors and in all likelyhood theres going to be more issues when we solve this. But the wasm example breaks because of `embassy-time`, concretely because that is using `wasm_timer` to get a replacement for std::time::Instant (which isn't available on web).

It took me a bit, because of the crazy build system you've got there - understandable but takes getting used to - , and because liker errors with wasm_bindgen are... perhaps more problematic than these on the native platform. But I did just patch `embassy-time(-driver)` to use `web-time` instead, which.... partially worked.

Ticks on the main thread work as before, and on the worker thread I didn't get a panic, but the loop never repeated.

IDK if `web-time` is even supposed to work on web-workers, that would be what I look at now.

Dirbaio:
ah, it's `embassy-time`, makes sense :) 

hope you figure it out, i'll gladly merge PRs that make things work in webworkers.

9SMTM6:
@Dirbaio thanks, I'll look around a bit. I've tried 2 other crates that implement `std::time::Instant` on the web, because that was the easiest check I came up with, all of them failed the same way `web-time` did.

My guess now, as to why they fail that way: They all use `performance.now()`. But that [behaves differently](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) on Web-Workers than it does on the main page. Their start time is different.

[...]

PR #3318, opened 2024-09-06
Use vendored performance API to handle timing on WASM

9SMTM6:
Before that this used wasm_timer, but only for Instant::now. But that panicked on web_workers, because it got there by going over the window, which isn't present in web workers.

This still doesn't work on all browsers though because of browser bugs, thus I added an assertion against usage on web-workers, that can be disabled if one so chooses.

This still needs documentation updates and a bit more testing to be done, but before that I wanted to discuss the implementation. Thus a draft PR.

See #3313.

### DIFF
--- a/embassy-time/Cargo.toml
+++ b/embassy-time/Cargo.toml
@@ -24,8 +24,10 @@ target = "x86_64-unknown-linux-gnu"
 features = ["defmt", "std"]
 
 [features]
+default = ["panic_on_webworker"]
 std = ["tick-hz-1_000_000", "critical-section/std"]
-wasm = ["dep:wasm-bindgen", "dep:js-sys", "dep:wasm-timer", "tick-hz-1_000_000"]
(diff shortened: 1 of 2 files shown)


9SMTM6:
Note that this code, other than I expected (or at least hoped) still doesnt work on Windows, where performance.now isn't broken, which means theres still other things not working as expected.

@Dirbaio Note how I had to wrap the check for whether we're in a webworker into thread_local. I am wondering why the general static in `time_driver_impl` isn't in thread local storage. That's the one thing I can think of as potential issue right now. But then, would that not potentially be an issue elsewhere? And changing this, without branching the macro, would influence change other implementations.

If you don't think this is the issue, I don't know what else to try. As it stands this PR only fixes a panic where the remainder doesn't work anyways, and potentially (but not certainly) makes a complete fix easier for someone else, so I don't know if merging this is worth it.

If you think its worth it, I will add documentation of this particular issue and that there seem to be other issues, otherwise feel free to close this.

Dirbaio:
Hello,

PR #3319, opened 2024-09-06
fix(stm32): fix dma and idle line detection in ringbuffereduartrx

badrbouslikhin:
This change is required to address an issue in the RingBufferedUartRx DMA handling and idle line detection.

When an IRQ wakes the `select` future, both inner futures (`dma` and `uart`) in the select are re-polled. Previously, the `dma` future was polled first. Since it can't check if it was woken for valid reasons (due to API limitations), it could end the select future without clearing the IDLE flag, which is handled in the `uart` future.

The proposed fix reverses the polling order, starting with the `uart` future. This allows proper checking and clearing of the IDLE flag before considering the DMA status, preventing scenarios where the flag remains set across multiple read operations.

Additionally, this PR enables the half-transfer IRQ for the DMA to help prevent data loss at high baud rates. I don't think it introduces any side effects.

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -777,6 +777,7 @@ impl<'a, W: Word> ReadableRingBuffer<'a, W> {
         let dir = Dir::PeripheralToMemory;
         let data_size = W::size();
 
+        options.half_transfer_ir = true;
         options.complete_transfer_ir = true;
         options.circular = true;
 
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
thanks!

note the `dma` future implementation is not very correct, it assumes the 2nd time it gets polled it's because DMA is half-full/full. However rust async allows "spurious wakes/polls". For example if you `select/join` the `wait_for_data_or_idle` future with another future, that another future will cause wakes that will wake both sides of the `select/join`.

I'll merge anyway since this bug was already there before, and at least this PR improves the situation somewhat.

badrbouslikhin:
> However rust async allows "spurious wakes/polls". For example if you select/join the wait_for_data_or_idle future with another future, that another future will cause wakes that will wake both sides of the select/join.

Interesting, I didn't know about "spurious wakes/polls". What would be the correct approach in your opinion?

I tried to implement a check to make sure the dma is half-full/full but all flags have been cleared by the IRQ already.

PR #3329, opened 2024-09-11
net: refactor to simplify lifetimes/generics.

Dirbaio:
Refactor embassy-net api. Advantages:

- Now there's a separate `Runner` struct. `run()` requires `&mut self`, this prevents the user from doing weird stuff like calling `run()` in two tasks concurrently.
- Stack is now Copy, so you can pass it by value everywhere. This means you need only one StaticCell (for the resources) instead of two (resources, stack).
- Stack is no longer generic over the driver, only the runner is. This makes code using embassy-net agnostic to what driver is in use, only the code creating the stack has to care.
- DnsSocket is not generic over the driver anymore.
- No forced `'static` litetime anymore. This means you can now dynamically create/teardown the network stack as many times as you want. If you run everything in a single task with `join` you don't need StaticCell at all.
- `Stack` and the sockets are all covariant over the lifetime now. This helps a bit with avoiding hairy lifetime errors.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -290,8 +290,9 @@ cargo batch \
     $BUILD_EXTRA
 
 
-# temporarily disabled, bluepill board got bricked
+# temporarily disabled, these boards are dead.
 rm -rf out/tests/stm32f103c8
+rm -rf out/tests/nrf52840-dk
 
 rm out/tests/stm32wb55rg/wpan_mac
 rm out/tests/stm32wb55rg/wpan_ble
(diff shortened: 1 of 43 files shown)


Dirbaio:
seems it improves perf a bit, too:

```
embassy-net [stm32 eth perf test](https://github.com/embassy-rs/embassy/blob/main/tests/stm32/src/bin/eth.rs), stm32h755zi, 64k tcp buffer:

smoltcp 0.11.0:
    download:        6273 kB/s  6264 kB/s
    upload:          9619 kB/s  9619 kB/s
    upload+download: 1297 kB/s  1575 kB/s

smoltcp git main 469dccd
    download:        5357 kB/s  5354 kB/s
    upload:          9678 kB/s  9588 kB/s
    upload+download: 1098 kB/s  1122 kB/s

smoltcp https://github.com/smoltcp-rs/smoltcp/pull/991
    download:        5826 kB/s  5822 kB/s
    upload:          9142 kB/s  9145 kB/s
    upload+download: 3043 kB/s  2729 kB/s

smoltcp https://github.com/smoltcp-rs/smoltcp/pull/991 + embassy-net refactor (this PR)
    download:        6148 kB/s  6147 kB/s
    upload:          9678 kB/s  9676 kB/s
    upload+download: 4691 kB/s  4752 kB/s
```

Dirbaio:
bender run

lulf (approved):
Changes look great!

Dirbaio:
bender run

PR #3330, opened 2024-09-11
rp: Fix indexing for pins >31 on rp235xb

CBJamo:
Does not handle using PIO with pins > 31, which is still broken.

### DIFF
--- a/embassy-rp/src/gpio.rs
+++ b/embassy-rp/src/gpio.rs
@@ -846,12 +846,12 @@ pub(crate) trait SealedPin: Sized {
 
     #[inline]
     fn _pin(&self) -> u8 {
-        self.pin_bank() & 0x1f
+        self.pin_bank() & 0x7f
     }
 
     #[inline]
     fn _bank(&self) -> Bank {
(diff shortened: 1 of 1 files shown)


embassy-rp/src/gpio.rs
```diff
@@ -953,7 +965,7 @@ macro_rules! impl_pin {
         impl SealedPin for peripherals::$name {
             #[inline]
             fn pin_bank(&self) -> u8 {
-                ($bank as u8) * 32 + $pin_num
+                ($bank as u8) * 64 + $pin_num
```
Dirbaio, line 968: if bank is bit 7, shouldn't this be 128?

Dirbaio (approved):

embassy-rp/src/gpio.rs
CBJamo: Ah, of course it is. I just didn't hit it because I'm not trying to use the qspi pins. The way pins are selected is a bit fiddly imo. When I get around to trying the gpio co-processor it should improve.

Dirbaio:
spurious failure, forcing merge.

PR #3332, opened 2024-09-12
rp: rp2350 pio pin fixes

CBJamo:
Disable pad isolation on any used pin.
Use GPIOBASE and offset pin bases if all pins are > 16, panic if some pins are < 16 and some are > 31

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -743,22 +743,21 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
             w.set_out_base(config.pins.out_base);
         });
 
-        #[cfg(feature = "_rp235x")]
+        //#[cfg(feature = "_rp235x")]
         {
             let mut low_ok = true;
             let mut high_ok = true;
-            for pin in [
-                config.pins.in_base,
-                config.pins.in_base + config.in_count,
-                config.pins.sideset_base,
-                config.pins.sideset_base + config.pins.sideset_count,
-                config.pins.set_base,
-                config.pins.set_base + config.pins.set_count,
-                config.pins.out_base,
-                config.pins.out_base + config.pins.out_count,
-            ] {
-                low_ok &= pin < 32;
-                high_ok &= pin >= 16;
+
+            let in_pins = config.pins.in_base..config.pins.in_base + config.in_count;
+            let side_pins = config.pins.sideset_base..config.pins.sideset_base + config.pins.sideset_count;
+            let set_pins = config.pins.set_base..config.pins.set_base + config.pins.set_count;
+            let out_pins = config.pins.out_base..config.pins.out_base + config.pins.out_count;
+
+            for pin_range in [in_pins, side_pins, set_pins, out_pins] {
+                for pin in pin_range {
+                    low_ok &= pin < 32;
+                    high_ok &= pin >= 16;
+                }
             }
 
             if !low_ok && !high_ok {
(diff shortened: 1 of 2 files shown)


embassy-rp/src/pio/mod.rs
```diff
@@ -740,6 +764,100 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
...
+                    ),
+                }
+            };
+
+            if shift_gpio_base {
+                sm.pinctrl().write(|w| {
+                    w.set_sideset_count(config.pins.sideset_count);
+                    w.set_set_count(config.pins.set_count);
+                    w.set_out_count(config.pins.out_count);
+                    w.set_in_base(config.pins.in_base - 16);
```
shilga, line 841: this operation breaks if `in_base` is not set at all. Which might happen if one for example only needs sideset pins. It's also the same for all the other pin_bases. I guess there should be an additional check

embassy-rp/src/pio/mod.rs
```diff
@@ -740,6 +764,100 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
...
+                        Some(config.pins.set_base)
+                    } else {
+                        None
+                    },
+                    if config.pins.out_count > 0 {
+                        Some(config.pins.out_base)
+                    } else {
+                        None
+                    },
+                ) {
```
Dirbaio, line 792: this match is crazy. I suggest the following instead:

```rust
let mut low_ok = true;
let mut high_ok = true;
for pin in [...the 4 options].iter().flatten() {
    low_ok &= pin < 32;
    high_ok &= pin >= 16;
}
if !low_ok && !high_ok { panic!("all pins blahblah"); }
let shift_gpio_base = !low_ok;
```

embassy-rp/src/pio/mod.rs
```diff
@@ -740,6 +764,100 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
...
+
+                    (i, side, set, out) => panic!(
+                        "All pins must either be <=31 or >=16, in:{:?}, side:{:?}, set:{:?}, out:{:?}",
+                        i, side, set, out
+                    ),
+                }
+            };
+
+            if shift_gpio_base {
+                sm.pinctrl().write(|w| {
```
Dirbaio, line 837: instead of duplicating this you can make a `shift` variable that's either `0` or `16` and substract it.

embassy-rp/src/pio/mod.rs
CBJamo: I don't think this is a problem. If one of the bases isn't being used, garbage will go into that field in the register, but if that base isn't being used it doesn't matter whats in that field.

shilga: That ends up being a panic. Otherwise I would not have said something.
```0.042804 ERROR panicked at /root/.cargo/git/checkouts/embassy-ae7d843df5da8929/6dbab7b/embassy-rp/src/pio/mod.rs:841:35:
attempt to subtract with overflow```

Comment (CBJamo) [reply embassy-rp/src/pio/mod.rs:841]: Ah, of course. I'm so used to building with release that I didn't think about it.

File: embassy-rp/src/pio/mod.rs [hunk 4]
```diff
@@ -767,95 +767,51 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
...
-                    w.set_in_base(config.pins.in_base);
-                    w.set_sideset_base(config.pins.sideset_base);
-                    w.set_set_base(config.pins.set_base);
-                    w.set_out_base(config.pins.out_base);
-                });
+            let mut low_ok = true;
+            let mut high_ok = true;
+            for pin in [
+                config.pins.in_base,
+                config.pins.in_base + config.in_count,
```
Dirbaio, line 774: if pin_count is 0 you want to ignore the base I think?

CBJamo: Whoops, you sure do.

Dirbaio (approved):

Issue #3289, opened 2024-08-26
USB ID 4000 on STM32H7R/S?


RileyLeff:
Howdy folks, hoping someone can point me in the right direction re: USB core IDs.

I spent the morning fiddling with the [STM32H7 usb_serial example](https://github.com/embassy-rs/embassy/blob/main/examples/stm32h7/src/bin/usb_serial.rs) to get it to run on my STM32H7R/S board (this relatively [new nucleo](https://www.st.com/en/evaluation-tools/nucleo-h7s3l8.html) with a STM32H7S3L8H MCU), as there isn't a USB example for the R/S boards yet. 

Changing some of the prescaler values in the initial config and pin #s got it to compile, but it would panic at runtime with this message: 

> not implemented: Unknown USB core id 4000

I added 0x0000_4000 to the core_id match statement in the relevant file [here](https://github.com/embassy-rs/embassy/blob/a7bfec2467b2fbb9623fb7713ca8f1cca5e78f56/embassy-stm32/src/usb/otg.rs#L286), and just gave it the same config as the 200x and 300x ids. That seems to run the example just fine without any panic. 

Would it be appropriate to submit a PR adding my example and appending 0x_4000 to that match statement with a v2_v3 config? Or are there additional considerations for core ID config beyond "this basic example works" that need to be addressed before adding support for a 4000 usb ID?

Dirbaio:
yes, please send a PR!

treating v4 the same as v2/v3 should be OK. Checking the core version is mostly used to workaround weird quirks in v1, if you tried that and it works then it probably means there's no quirks we should care about in v4.

RileyLeff:
Turns out this was not working as I thought -- while it compiles, runs, and prints a lot of very beautiful INFOs, the usb FS-enabled pins don't actually exist on my board. Whoops. It compiles just fine if I switch it over to the HS pins (which I've confirmed are actually hooked up to the USB port this time, lol) but I'm getting a runtime panic that is completely stumping me: 

> ERROR panicked at 'peripheral 'USB_OTG_HS' is configured to use the 'USB' clock, which is not running. Either enable it in 'config.rcc' or change 'config.rcc.mux' to use another clock'

I don't see usb as a field in rcc or mux. **Any idea how I can set up that clock?** I see usb in the rcc clocks module but no idea how to get it into the config.

[...]

PR #3337, opened 2024-09-15
Add OTG_HS support for STM32H7R/S

doesnotcompete:
This adds support for the embedded high-speed USB peripheral on the STM32H7R/S series. I've confirmed the `usb_serial` example works on a Nucleo-H7S3 board using the internal HS PHY.
These changes build on prior work by @RileyLeff in #3289.

There are some remaining issues:
- I haven't tested this with the OTG_FS peripheral or on other boards with different clock configurations.
- I'm not sure the way I've implemented the clock configuration is ideal. The USBPHYC on the H7R/S series embeds a PLL that can receive a range of different frequencies via the `USBPHYCSEL` clock mux and generates the necessary 48 and 60MHz clocks for the OTG blocks internally (see the clock diagram below). I've simply added `USBPHYC` as a fallback clock for the `OTG_HS` peripheral in [`stm32-data`](https://github.com/embassy-rs/stm32-data/blob/acaf04256034066bd5b3a8426224ccf3e4cb7d19/stm32-data-gen/src/rcc.rs#L308) to get the clock generator to generate a mux definition and then explicitly pick the corresponding reference clock in the H-series RCC driver, but this approach results in the PLL input frequency being registered as the actual OTG kernel clock.
- Possibly related to this, `ahb_freq` in `otg.rs` doesn't represent the AHB bus frequency, but just this USBPHYC PLL input frequency. I've just skipped the check of a sufficient AHB clock on the H7R/S, as I didn't find a straight-forward solution and the USBPHYC PLL input might very well be below 30MHz, but this is likely not ideal.

usb_clk


doesnotcompete:
The CI failure is apparently caused by a missing definition of the `OTG_HS` IP block in the `cubedb` source files of `stm32-data`.
`STM32H7R7A8Ix.xml` and `STM32H7S7Z8Jx.xml` only include the block for `USB_OTG_FS`, although all H7R/S MCUs, including the [H7R7A8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7r7a8.html) and [H7S7Z8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7s7z8.html), should have HS USB with an integrated PHY.
I'm not sure what'd be the best way to fix this.

RileyLeff:
really appreciate your continued work on this! i'm pretty out of my depth here, would not have figured this out myself.

Dirbaio:
> The CI failure is apparently caused by a missing definition of the OTG_HS IP block in the cubedb source files of stm32-data.
STM32H7R7A8Ix.xml and STM32H7S7Z8Jx.xml only include the block for USB_OTG_FS, although all H7R/S MCUs, including the [H7R7A8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7r7a8.html) and [H7S7Z8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7s7z8.html), should have HS USB with an integrated PHY.

wow that's annoying. I've tried updating the xmls from the latest Cube but OTG_HS is still missing. I'd just workaround it for now:  use `#[cfg(peri_otg_hs)]` to conditionally-compile code only if OTG_HS is present.

I'd wait for ST to fix it, perhaps report it. Where do you report cubemx bugs? :thinking:

doesnotcompete:
> > The CI failure is apparently caused by a missing definition of the OTG_HS IP block in the cubedb source files of stm32-data.
> > STM32H7R7A8Ix.xml and STM32H7S7Z8Jx.xml only include the block for USB_OTG_FS, although all H7R/S MCUs, including the [H7R7A8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7r7a8.html) and [H7S7Z8](https://www.st.com/en/microcontrollers-microprocessors/stm32h7s7z8.html), should have HS USB with an integrated PHY.
> 
> wow that's annoying. I've tried updating the xmls from the latest Cube but OTG_HS is still missing. I'd just workaround it for now: use `#[cfg(peri_otg_hs)]` to conditionally-compile code only if OTG_HS is present.
> 
> I'd wait for ST to fix it, perhaps report it. Where do you report cubemx bugs? 🤔

That's a great idea :) I've added a check in the latest commit and also tried reporting the issue to ST here on GitHub.

Dirbaio (approved):
thanks! :rocket:

PR #3338, opened 2024-09-15
RP2350 TRNG support

ionspin:
This is a driver for RP2350 TRNG, I used RNG driver from `embassy-nrf` as a template, but adapted to be more in line with other drivers in `embassy-rp`.

From 2350 datasheet:
```
12.12.1. Overview
RP2350 contains an Arm IP-based True Random Number Generator block. It supports the following features:
• Compliance with FIPS Publication 140-2, BSI AIS-31, and NIST SP 800-90B
• Produces approximately 7.5 kb/s of entropy when the core runs at 150 MHz
On request, the TRNG block generates a block of 192 entropy bits generated by automatically processing a series of
periodic samples from the TRNG block’s internal Ring Oscillator (ROSC).

The TRNG block’s ROSC is a free-running oscillator with no direct connection to the system clocks on RP2350. As a
result, the ROSC generally runs asynchronously to the system clocks.
After a sufficient number of samples have been collected, the TRNG block completes the generation process and
presents the random number in the EHR_DATA[x] result registers.
```

They go on to note that Pico SDK and Pico Bootrom disable all entropy checks and sample as fast as possible from the TRNG ROSC directly, and then pass the data through SHA256 before using it (hardware accelerated in case of Bootrom) or xiroshiro128** (in case of pico-random library in Pico SDK , note that xiroshiro128** is verions 1.0 there) .

I've decided to leave the entropy checks enabled and datasheet recommended config for the TRNG default config as they seem to be sane.

Related to #3241


embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,437 @@
...
+        }
+    }
+
+    fn blocking_read_ehr_registers_into_array(&mut self, buffer: &mut [u8; TRNG_BLOCK_SIZE_BYTES]) {
+        let regs = T::regs();
+
+        let trng_busy_register = regs.trng_busy();
+        let trng_valid_register = regs.trng_valid();
+
+        while trng_busy_register.read().trng_busy() {}
```
ionspin, line 257: This should be replaced with `self.blocking_wait_for_succesfull_generation();`

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,437 @@
...
+    fn enable_irq(&self) {
+        T::Interrupt::unpend();
+        unsafe { T::Interrupt::enable() }
+    }
+
+    fn disable_irq(&self) {
+        T::Interrupt::disable();
+    }
+
+    fn blocking_wait_for_succesfull_generation(&self) {
```
Dirbaio, line 214: `succesfull` -> `successful`

embassy-rp/src/lib.rs
```diff
@@ -402,6 +404,8 @@ embassy_hal_internal::peripherals! {
 
     WATCHDOG,
     BOOTSEL,
+
+    TRNG
```
Dirbaio, line 408: you should cfg this for rp235x only

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,428 @@
...
+/// On request, the TRNG block generates a block of 192 entropy bits generated by automatically processing a series of
+/// periodic samples from the TRNG block’s internal Ring Oscillator (ROSC).
+const TRNG_BLOCK_SIZE_BITS: usize = 192;
+const TRNG_BLOCK_SIZE_BYTES: usize = TRNG_BLOCK_SIZE_BITS / 8;
+
+impl<T: Instance> Trng<T> {
+    /// Create a new TRNG driver.
+    pub fn new(
+        _trng: impl Peripheral<P = T>,
+        _irq: impl Binding<T::Interrupt, InterruptHandler<T>>,
```
Dirbaio, line 163: like the other drivers, add the `'d` lifetime to `Trng` and add `+ 'd` to `trng` and `irq`.

the purpose of the lifetime is to allow creating drivers with borrowed peripherals, so you can drop them and create them again. see this for example: https://github.com/embassy-rs/embassy/blob/main/examples/nrf52840/src/bin/twim_lowpower.rs#L35

Without the `+ 'd` you can create a `Trng` with a borrowed peripheral but hold on to the `Trng` driver forever, which would allow two `Trng` drivers to exist at the same time which would be bad.

embassy-rp/src/lib.rs
ionspin: I think this whole peripherals! macro call is under #[cfg(feature = "_rp235x")]
Is that the right cfg?

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,428 @@
...
+    fn stop_rng(&self) {
+        let regs = T::regs();
+        let source_enable_register = regs.rnd_source_enable();
+        source_enable_register.write(|w| w.set_rnd_src_en(false));
+        let reset_bits_counter_register = regs.rst_bits_counter();
+        reset_bits_counter_register.write(|w| w.set_rst_bits_counter(true));
+    }
+
+    fn enable_irq(&self) {
+        T::Interrupt::unpend();
```
Dirbaio, line 206: this unpend can cause missing events if a lot of time pass between `start_rng` and `enable_irq` (for example if a higher-prio task preempts us)

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,428 @@
...
+            return; // Nothing to fill
+        }
+        self.start_rng();
+
+        let destination_length = destination.len();
+        let mut bytes_transferred = 0usize;
+
+        let mut buffer = [0u8; TRNG_BLOCK_SIZE_BYTES];
+
+        while bytes_transferred < destination_length {
```
Dirbaio, line 334: this code is a bit hard to follow. You're doing chunking manually, you have a  `match remaining` that could be written as an `if`.

I'd suggest chunking with `for chunk in destination.chunks(TRNG_BLOCK_SIZE_BYTES)`, then you just have to deal with the last chunk being shorter than TRNG_BLOCK_SIZE_BYTES.

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,428 @@
...
+        let trng_busy_register = regs.trng_busy();
+        let trng_valid_register = regs.trng_valid();
+
+        let waker = T::waker();
+
+        let destination_length = destination.len();
+
+        let mut counter = 0;
+        poll_fn(|context| {
+            critical_section::with(|_cs| {
```
Dirbaio, line 279: why a critical section?

embassy-rp/src/lib.rs
Dirbaio: ah my bad, didn't notice. sorry!

embassy-rp/src/trng.rs
Dirbaio: with `chunk.copy_from_slice(&buffer[..chunk.len()])`

embassy-rp/src/trng.rs
ionspin: Copy-paste from nrf, with a dose of uncertainty :) 
I presume I don't need it because &mut self guarantees that this execution will be exclusive?

Dirbaio: yep, you have the guarantee no other task will touch TRNG

Dirbaio (approved):
thanks!

embassy-rp/src/trng.rs
```diff
@@ -0,0 +1,405 @@
...
+        let low = regs.ehr_data4().read() as u64;
+        // 12.12.3 After successful generation, read the last result register, EHR_DATA[5] to
+        // clear all of the result registers.
+        let result = (regs.ehr_data5().read() as u64) << 32 | low;
+        self.stop_rng();
+        result
+    }
+}
+
+impl<'d, T: Instance> rand_core::RngCore for Trng<'d, T> {
```
wez, line 357: Sorry to necro, but I have a question; happy to be redirected elsewhere if needed.

The question is: should [rand_core::CryptoRng](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRng.html) also be implemented for Trng?
I'm not enough of a crypo guy to know for sure, but the datasheet says `Compliance with FIPS Publication 140-2, BSI AIS-31, and NIST SP 800-90B` which makes me hope that it is strong enough for this purpose.

Did you already consider and explicitly choose not to implement this trait, or is it an oversight?

On a related topic, in the absence of hardware sources for this, do you have a recommendation for how to satisfy CryptoRng on a pico2?

wez: Never mind: I found that I can init a csprng (eg: `ChaCha20`) from this trng and all is good!

Dirbaio: it should implement it yes. A PR would be appreciated.

Using chacha20 works and might be faster, at the cost of of more code size.

PR #3340, opened 2024-09-16
fix(boot): return signature error when no features

### LINKED ISSUE #3339: Embassy boot update signature verification does not return an error when no signature feature is enabled
In the `embassy-boot` library, the documentation from the `verify_and_mark_updated` method says:

> If no signature feature is set then this method will always return a signature error.

However, the implementation seems to call `mark_updated` without returning any error when no signature feature is enabled.
I think there is either a mistake in the documentation or in the implementation (the later seems to be the case for me).

Note: the code is available in blocking and async methods, both may need to be patched.

https://github.com/embassy-rs/embassy/blob/3d6a270f30c45eaf394c8eb8bf182dd1a7ec2d7b/embassy-boot/src/firmware_updater/asynch.rs#L85

https://github.com/embassy-rs/embassy/blob/3d6a270f30c45eaf394c8eb8bf182dd1a7ec2d7b/embassy-boot/src/firmware_updater/blocking.rs#L120

kingofpayne:
Fixes #3339
Always return signature error in verify_and_mark_updated when no signature features are enabled.

### DIFF
--- a/embassy-boot/src/firmware_updater/asynch.rs
+++ b/embassy-boot/src/firmware_updater/asynch.rs
@@ -107,7 +107,8 @@ impl<'d, DFU: NorFlash, STATE: NorFlash> FirmwareUpdater<'d, DFU, STATE> {
             let mut message = [0; 64];
             self.hash::<Sha512>(_update_len, &mut chunk_buf, &mut message).await?;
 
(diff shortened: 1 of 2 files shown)


lulf (approved):
LGTM, thanks! Please run cargo fmt to fix CI

kingofpayne:
Done!

PR #3341, opened 2024-09-16
rp: add constructor for tx-only blocking UART

shilga:
The same thing exists for rx-only, why not for tx-only?


Dirbaio (approved):
yep, no reason :D

PR #3343, opened 2024-09-16
LPC: embassy-lpc55 hal base with gpio and pint driver

george-cosma:
# Overview
This pull request implements a simple embassy-lpc55 hal based on [`lpc55-pac`](https://github.com/lpc55/lpc55-pac). This PR only implements the GPIO driver and the PINT (pin interrupt) driver, which is, in turn, used internally to detect edge & level changes for input GPIO.

# TODO / Help Wanted
While there are many drivers that are yet to be implemented (most notably a time driver, which I want to tackle soon!), I'm going to limit this PR to GPIO and PINT. Despite the limited scope, there are some design decisions of which I am unsure about.

First, how should the hal be organized in terms of chip-specific details? I thought about making this hal similar to how `embassy-nrf` does it (each nrf family has a `.rs` file that creates the unique configuration, but the drivers are more or less identical). However, I am not that familiar with the other chips in the `lpc` series. I would personally leave it as-is, and deal with this when the problem of another chip arises, but I thought I should get your feedback on this first.

Note:  although I named the hal `embassy-lpc55`, it only functions for a subset of these: the LPC55S6X, LPC55S2x and LPC552X series. There are other series [as shown here](https://ro.mouser.com/c/semiconductors/embedded-processors-controllers/microcontrollers-mcu/arm-microcontrollers-mcu/?q=LPC). However, of these I have tested the hal only on the LPC55S69, as I have the [LPCXpresso55S69 board or sometimes LPC55S69-EVK](https://www.nxp.com/design/design-center/software/development-software/mcuxpresso-software-and-tools-/lpcxpresso-boards/lpcxpresso55s69-development-board:LPC55S69-EVK).

Second, what are  the differences between `Pin` and `SealedPin` inside of the GPIO driver? When creating this hal, I took inspiration from the `embassy-nrf` hal and the `embassy-rp` one. However, I'm unsure of the distinction between these two? My guess would be that `SealedPin` has something to do with the fact that it doesn't need to be implemented on a Peripheral? I'm unsure. If I can understand what the deal with it is, I can document the driver properly, and perhaps cut down on the amount of abstraction.

Thirdly, I noticed in the other hals that pac access is usually done by getting a raw pointer to the relevant register block. This seems fine to me, I've integrated this idea in the `pac_utils.rs` file, but I'm unsure if it is **actually** safe. If the underlying pac handles read/writes atomically, then this isn't an issue. If the opposite is true, registers could end up in an unexpected state if modified from different threads - a race condition. That being said, the current code would not be affected by this due to the fact that each PIO is owned exactly once. The only time the `.modify()` action is  taken on a register, is when the registers are specific to each pin. As such (unless a circumvented by unsafe rust), only one thread could ever modify a pin-specific register. However, putting pac access behind an `Arc<RefCell<...>>` wouldn't be that difficult either.

And a final issue I'd like your feedback on - how should we detect level changes? In this PR I used the Pin Interrupts, but these are designed so they trigger an interrupt only for a specific, user-configured, pin. In total there are only 8 pin interrupts, so if a user wants to wait on more than 8 input changes at a time, they can't.

The LPC55S6X/LPC55S2x/LPC552X user manual also mentions a "Group Interrupt" peripheral (GINT). This allows for multiple pins to contribute to an interrupt, however after the interrupt triggers there is no way to know which pin was the cause. A software implementation that determines the differences between pre-interrupt and post-interrupt would be needed.

So yhea, that's about it. It works, but with a lot of unanswered questions :D - any feedback on this would be appreciated!

# Acknowledgements

The `memory.x` file was based on [`lpc55-hal`'s](https://github.com/lpc55/lpc55-hal) but I'm unsure how to insert that attribution inside of the `.x` file itself.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -171,6 +171,7 @@ cargo batch \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u031r8,defmt,exti,time-driver-any,time \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u073mb,defmt,exti,time-driver-any,time \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u083rc,defmt,exti,time-driver-any,time \
+    --- build --release --manifest-path embassy-nxp/Cargo.toml --target thumbv8m.main-none-eabihf \
     --- build --release --manifest-path cyw43/Cargo.toml --target thumbv6m-none-eabi --features ''\
     --- build --release --manifest-path cyw43/Cargo.toml --target thumbv6m-none-eabi --features 'log' \
     --- build --release --manifest-path cyw43/Cargo.toml --target thumbv6m-none-eabi --features 'defmt' \
@@ -227,6 +228,7 @@ cargo batch \
     --- build --release --manifest-path examples/stm32wb/Cargo.toml --target thumbv7em-none-eabi --out-dir out/examples/stm32wb \
     --- build --release --manifest-path examples/stm32wba/Cargo.toml --target thumbv8m.main-none-eabihf --out-dir out/examples/stm32wba \
     --- build --release --manifest-path examples/stm32wl/Cargo.toml --target thumbv7em-none-eabi --out-dir out/examples/stm32wl \
+    --- build --release --manifest-path examples/lpc55s69/Cargo.toml --target thumbv8m.main-none-eabihf --out-dir out/examples/lpc55s69 \
     --- build --release --manifest-path examples/boot/application/nrf/Cargo.toml --target thumbv7em-none-eabi --features embassy-nrf/nrf52840,skip-include --out-dir out/examples/boot/nrf52840 \
     --- build --release --manifest-path examples/boot/application/nrf/Cargo.toml --target thumbv8m.main-none-eabihf --features embassy-nrf/nrf9160-ns,skip-include --out-dir out/examples/boot/nrf9160 \
     --- build --release --manifest-path examples/boot/application/nrf/Cargo.toml --target thumbv8m.main-none-eabihf --features embassy-nrf/nrf9120-ns,skip-include --out-dir out/examples/boot/nrf9120 \
--- a/embassy-nxp/Cargo.toml
+++ b/embassy-nxp/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "embassy-nxp"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+cortex-m = "0.7.7"
+cortex-m-rt = "0.7.0"
+critical-section = "1.1.2"
+embassy-hal-internal = {version = "0.2.0", path = "../embassy-hal-internal", features = ["cortex-m", "prio-bits-2"] }
+embassy-sync = { version = "0.6.0", path = "../embassy-sync" }
+lpc55-pac = "0.5.0"
+defmt = "0.3.8"
+
+[features]
+default = ["rt"]
+rt = ["lpc55-pac/rt"]
--- a/embassy-nxp/src/gpio.rs
+++ b/embassy-nxp/src/gpio.rs
@@ -1,0 +1,361 @@
+use embassy_hal_internal::impl_peripheral;
+
+use crate::pac_utils::*;
+use crate::{peripherals, Peripheral, PeripheralRef};
+
+pub(crate) fn init() {
+    // Enable clocks for GPIO, PINT, and IOCON
+    syscon_reg()
+        .ahbclkctrl0
+        .modify(|_, w| w.gpio0().enable().gpio1().enable().mux().enable().iocon().enable());
+}
+
+/// The GPIO pin level for pins set on "Digital" mode.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum Level {
+    /// Logical low. Corresponds to 0V.
+    Low,
+    /// Logical high. Corresponds to VDD.
+    High,
+}
+
+/// Pull setting for a GPIO input set on "Digital" mode.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum Pull {
+    /// No pull.
+    None,
+    /// Internal pull-up resistor.
+    Up,
+    /// Internal pull-down resistor.
+    Down,
+}
+
+/// The LPC55 boards have two GPIO banks, each with 32 pins. This enum represents the two banks.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum Bank {
+    Bank0 = 0,
+    Bank1 = 1,
+}
+
+/// GPIO output driver. Internally, this is a specialized [Flex] pin.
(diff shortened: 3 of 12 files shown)


lulf (commented):
Great to see new HALs! Embassy tries to make HALs that cover as widely as possible within a chip family. Is there any particular reason this is limited to lpc55 family? Would it make sense to make this into an embassy-nxp HAL instead? For comparison, the embassy-stm32 HAL covers a _lot_ of chip families in the HAL, so it is certainly possible.

george-cosma:
The simple answer as to why "lpc55" is because this is the only chip family I have at my disposal right now. Of course, I am not married to the idea of having a very specific hal for the lpc55. I would personally favour having an `embassy-lpc`. Other NXP chip families might be substantially different, though that is based more on a hunch. I'm unsure how the `stm32` hal functions to be able to cover so many chip families, but if it is possible for the stm32 chips, it might be possible for nxp chips too.

lulf:
Apologies, I only read the code, not your entire comment :sweat_smile: 

> # Overview
> 
> This pull request implements a simple embassy-lpc55 hal based on [`lpc55-pac`](https://github.com/lpc55/lpc55-pac). This PR only implements the GPIO driver and the PINT (pin interrupt) driver, which is, in turn, used internally to detect edge & level changes for input GPIO.
> # TODO / Help Wanted
> 
> While there are many drivers that are yet to be implemented (most notably a time driver, which I want to tackle soon!), I'm going to limit this PR to GPIO and PINT. Despite the limited scope, there are some design decisions of which I am unsure about.
> 
> First, how should the hal be organized in terms of chip-specific details? I thought about making this hal similar to how `embassy-nrf` does it (each nrf family has a `.rs` file that creates the unique configuration, but the drivers are more or less identical). However, I am not that familiar with the other chips in the `lpc` series. I would personally leave it as-is, and deal with this when the problem of another chip arises, but I thought I should get your feedback on this first.
> 

I think it should go as widely as possible, at least if if it's to be put inside the embassy repo, because we don't want to end up with many hals for very similar chip families.

IMHO the distinction should be on 'peripheral similarity'. If a UART across all NXP MCUs share 90% of the code, they should be in the same HAL.

> Note: although I named the hal `embassy-lpc55`, it only functions for a subset of these: the LPC55S6X, LPC55S2x and LPC552X series. There are other series [as shown here](https://ro.mouser.com/c/semiconductors/embedded-processors-controllers/microcontrollers-mcu/arm-microcontrollers-mcu/?q=LPC). However, of these I have tested the hal only on the LPC55S69, as I have the [LPCXpresso55S69 board or sometimes LPC55S69-EVK](https://www.nxp.com/design/design-center/software/development-software/mcuxpresso-software-and-tools-/lpcxpresso-boards/lpcxpresso55s69-development-board:LPC55S69-EVK).
> 
> Second, what are the differences between `Pin` and `SealedPin` inside of the GPIO driver? When creating this hal, I took inspiration from the `embassy-nrf` hal and the `embassy-rp` one. However, I'm unsure of the distinction between these two? My guess would be that `SealedPin` has something to do with the fact that it doesn't need to be implemented on a Peripheral? I'm unsure. If I can understand what the deal with it is, I can document the driver properly, and perhaps cut down on the amount of abstraction.
> 

You can read more about the reason for sealed traits here https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/

> Thirdly, I noticed in the other hals that pac access is usually done by getting a raw pointer to the relevant register block. This seems fine to me, I've integrated this idea in the `pac_utils.rs` file, but I'm unsure if it is **actually** safe. If the underlying pac handles read/writes atomically, then this isn't an issue. If the opposite is true, registers could end up in an unexpected state if modified from different threads - a race condition. That being said, the current code would not be affected by this due to the fact that each PIO is owned exactly once. The only time the `.modify()` action is taken on a register, is when the registers are specific to each pin. As such (unless a circumvented by unsafe rust), only one thread could ever modify a pin-specific register. However, putting pac access behind an `Arc<RefCell<...>>` wouldn't be that difficult either.
>

It is the 'job' of the HAL to build safe abstractions on top of the PACs. I.e. ensure that registers are written in the correct order, and only by 1 'thread' (the borrow checker will help you with that).

There's always the case that someone writes unsafe code and accesses the registers directly, but that's outside the scope of the HAL.
 
> And a final issue I'd like your feedback on - how should we detect level changes? In this PR I used the Pin Interrupts, but these are designed so they trigger an interrupt only for a specific, user-configured, pin. In total there are only 8 pin interrupts, so if a user wants to wait on more than 8 input changes at a time, they can't.
> 

Often with a HAL, you need to figure out the "sweet spot" of how much flexibility you give in the API vs. how common the use case is. Maybe a limit of 8 is OK, it can be reworked later if someone has the use case.

> The LPC55S6X/LPC55S2x/LPC552X user manual also mentions a "Group Interrupt" peripheral (GINT). This allows for multiple pins to contribute to an interrupt, however after the interrupt triggers there is no way to know which pin was the cause. A software implementation that determines the differences between pre-interrupt and post-interrupt would be needed.
> 
> So yhea, that's about it. It works, but with a lot of unanswered questions :D - any feedback on this would be appreciated!
> # Acknowledgements
> 
> The `memory.x` file was based on [`lpc55-hal`'s](https://github.com/lpc55/lpc55-hal) but I'm unsure how to insert that attribution inside of the `.x` file itself.

You should be able to write comments in linker scripts using '/* */', just stating the attribution.

george-cosma:
> IMHO the distinction should be on 'peripheral similarity'. If a UART across all NXP MCUs share 90% of the code, they should be in the same HAL.
Thinking about it, it should be safer to start off broad (`embassy-nxp`) and if in time we discover if chip families are so different that they require drastically different code, we can split them off much more easily then to merge different hals

jamesmunns:
For what it's worth, NXP has gone through multiple acquisitions, and has a couple of very different chip families active at the moment: https://www.nxp.com/products/processors-and-microcontrollers/arm-microcontrollers/general-purpose-mcus:GENERAL-PURPOSE-MCUS

The biggest three are the Kinetis (formerly freescale), i.mxrt, and LPC.

I would be somewhat surprised if there was a meaningful amount of IP to share across these families, but I also do not use any of them actively (I've used all three in passing in the past, I don't remember them being particularly similar.

embassy-nxp might work, but embassy-lpc, embassy-imxrt, and embassy-kinetis might be better splits.

george-cosma:
Searching around google a bit and speaking with a coworker who worked with more NXP boards, it should be possible to have a single hal. There is a C hal for the zephyr project that groups all of NXP chips in one (https://github.com/zephyrproject-rtos/hal_nxp), and I understood that the Kinetis and i.mxrt ones should be similar enough. LPCs are bit newer and do break away those two, but not completely.

lulf (approved):
I think this is probably a good starting point for an nxp hal, thanks!

lulf (commented):
@george-cosma It just occured to me, could you add the examples to CI as well?

george-cosma:
Sure thing

lulf (approved):

lulf:
bender run

lulf:
bender run

george-cosma:
bender run

george-cosma:
@lulf Fixed it with a dummy force-push 😄

Issue #3342, opened 2024-09-16
boot not compatible with STM32L0x1 device family


kingofpayne:
Erasing a flash memory page in STM32L0x1 devices results in bytes set to 0x00 instead of the usual convention that erasing flash leads to 0xFF. I found this while trying to make embassy-boot work on a STM32L081 chip, and there is no clear indication from ST on this behaviour in the reference manual. The only evidence I can provide is, from [RM0377](https://www.st.com/resource/en/reference_manual/rm0377-ultralowpower-stm32l0x1-advanced-armbased-32bit-mcus-stmicroelectronics.pdf), the presence of the `NOTZEROERR` flag in the `FLASH_SR` register, whose description is:

> This bit is set by hardware when a program in the Flash program or System Memory tries to overwrite a not-zero area.

The issue is that erasing the bootloader state partition results in an invalid state, making the bootloader unable to swap to the DFU as it believes the swap has been attempted and failed, triggering a rollback where it is not possible. To make embassy-boot work on my device, I had to patch the following line and replace the constant with the value 0x00:
https://github.com/embassy-rs/embassy/blob/a23f56b3ddc84a7e87e4b8750df26b91a2cb9637/embassy-boot/src/lib.rs#L17

So it is somewhat expected that the current code is subject to improvement and not ready for all devices yet. Not blaming embassy here as the constant is not yet part of the `NorFlash` trait (see his [issue in embedded-storage](https://github.com/rust-embedded-community/embedded-storage/issues/35)). I even have to note that it was a great experience to see such a magic constant, whose only modifications fixed everything!

There is one example using embassy boot in `examples/boot/application/stm32l0` which I think can be misleading, as I don't expect this to work properly (has anyone tested it?).

While the `NorFlash` trait does not provide enough information yet, an option to define the expected erase value would be helpful, either set at runtime or via a feature?

PR #3344, opened 2024-09-16
boot: flash-erase-zero feature for compatibility with STM32L0x1 devices

kingofpayne:
Proposal for issue #3342, subject for discussion. Any feedback is welcome.
Allow compatibility with devices whose flash erase set bytes to 0x00 instead of 0xFF, using a new flash-erase-zero feature.


lulf (approved):
Thanks for fixing. This used to work using a generic param to embassy-boot, but it appears that has been lost in refactorings. I think using a feature flag is a less intrusive way to handle this, so lgtm!

PR #3350, opened 2024-09-18
net: add all combinations of wait methods for link/config up/down.

Dirbaio:
- **net-esp-hosted: remove useless fn init.**
- **net-esp-hosted: set wpa3_supported=true.**
- **net: add all combinations of wait methods for link/config up/down.**

### DIFF
--- a/embassy-net-esp-hosted/src/control.rs
+++ b/embassy-net-esp-hosted/src/control.rs
@@ -120,7 +120,7 @@ impl<'a> Control<'a> {
             pwd: unwrap!(String::try_from(password)),
             bssid: String::new(),
             listen_interval: 3,
-            is_wpa3_supported: false,
+            is_wpa3_supported: true,
         };
         ioctl!(self, ReqConnectAp, RespConnectAp, req, resp);
         self.state_ch.set_link_state(LinkState::Up);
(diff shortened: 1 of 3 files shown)


embassy-net-esp-hosted/src/control.rs
```diff
@@ -120,7 +120,7 @@ impl<'a> Control<'a> {
             pwd: unwrap!(String::try_from(password)),
             bssid: String::new(),
             listen_interval: 3,
-            is_wpa3_supported: false,
+            is_wpa3_supported: true,
```
plaes, line 123: This commit message should be as TODO in code..

Dirbaio: what's todo? WPA3 works flawlessly now.

plaes: Commit came with following message:
> I've noticed wpa3 still works without this flag, so I'm not sure what this does tbh...

Dirbaio: if WPA3 is working well, the chances of that TODO ever getting looked at are basically zero... I don't think adding it is worth it.

PR #3354, opened 2024-09-19
embassy_sync/pubsub: fix PubSubBehavior visibility

mammothbane:
https://github.com/embassy-rs/embassy/pull/2969 appears to have broken direct `publish_immediate()` on `pubsub::Channel`, as it functionally made `PubSubBehavior` private and didn't delegate this method to the new (private) `SealedPubSubBehavior`.

This change moves `publish_immediate`, `capacity`, and `is_full` from `SealedPubSubBehavior` to `PubSubBehavior` in order to restore them to `pub` visibility.

### DIFF
--- a/embassy-sync/src/pubsub/mod.rs
+++ b/embassy-sync/src/pubsub/mod.rs
@@ -194,6 +194,25 @@ impl<M: RawMutex, T: Clone, const CAP: usize, const SUBS: usize, const PUBS: usi
     }
 }
 
+impl<M: RawMutex, T: Clone, const CAP: usize, const SUBS: usize, const PUBS: usize> crate::pubsub::PubSubBehavior<T>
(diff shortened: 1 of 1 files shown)


diondokter:
Doesn't this make all trait methods public again?

Some are valuable sure, but lots are private and should not be touched by users.
The ones that are allowed public are:

- publish_immediate
- capacity
- free_capacity
- len
- is_empty
- is_full

I'd argue though that only `publish_immediate`, `capacity` and `is_full` should really be public. The others are hard to wrap your head around as they might not act like you'd expect if you don't know how the pubsub is implemented.

mammothbane:
Ok, I'm on board with that motivation &mdash; PR's updated accordingly.

diondokter (approved):
LGTM

But I have no rights to merge :P

PR #3355, opened 2024-09-19
embassy_stm32/eth: support compliance testing

Gerharddc:
This change adds the possibility to perform Ethernet compliance testing with STM32 systems when using a custom PHY struct to generate test patterns.

### DIFF
--- a/embassy-stm32/src/eth/mod.rs
+++ b/embassy-stm32/src/eth/mod.rs
@@ -177,6 +177,20 @@ pub unsafe trait PHY {
     fn poll_link<S: StationManagement>(&mut self, sm: &mut S, cx: &mut Context) -> bool;
 }
 
+impl<'d, T: Instance, P: PHY> Ethernet<'d, T, P> {
+    /// Directly expose the SMI interface used by the Ethernet driver.
+    ///
+    /// This can be used to for example configure special PHY registers for compliance testing.
+    ///
+    /// # Safety
+    ///
+    /// Revert any temporary PHY register changes such as to enable test modes before handing
+    /// the Ethernet device over to the networking stack otherwise things likely won't work.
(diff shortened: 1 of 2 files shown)


Gerharddc:
It seems I've only modified eth v1, I still need to modify eth v2 but it would be great to already get some feedback on the PR if possible.

Gerharddc:
On second thought, it is probably better to add a separate struct called `EthernetTester` (or similar) which consumes the `Ethernet` struct in order to enter test mode. It would then have a function to give back the `Ethernet` struct which would reset the PHY first. This would then leverage the type system to ensure you don't hand over `Ethernet` to the networking stack while the PHY is still in test mode.

Dirbaio:
I'd prefer to avoid extending the `PHY` trait with methods for every single feature one might want from a PHY. 

The trait exists so you can customize the behavior of SMI operations the ethernet driver needs to do. For example, the `embassy-net-driver` trait requires `Eth` to be able to report whether the link is up or down. This means `Eth` has to know how to get this info out of the phy, and since it varies based on the phy it's a trait method so that it can be customized by the user.

This is not the case for compliance test mode. The `Eth` driver never needs to call it, only the end user. So there's no need for including these functions in the trait.

I think exposing access to SMI instead should be enough for the user to do any SMI operation they want, without having to extend the trait for each possible feature:

```rust
impl Eth {
    fn station_management(&mut self) -> &mut impl StationManagement {
        &mut self.station_management
    }
}
```

Gerharddc:
@Dirbaio thanks for the feedback! I agree that it is indeed much cleaner to simply expose SMI directly and have made the suggested change. I opted to mark the function as unsafe though since I believe it could unexpectedly mess things up and most users probably should not be accessing the SMI directly under normal circumstances.

Dirbaio (approved):
thanks!

PR #3356, opened 2024-09-19
stm32: Fix RingBufferedUartRx hard-resetting DMA after initial error

peterkrull:
This fixes the RingBufferedUartRx::read never being able to recover after an initial overrun error, due the the complete reset of the DMA channel, instead of a temporary pause.

Possibly related issues: #1441 #1427


peterkrull:
Actually, this requires one more change within RingBufferedUartRx to be fully fixed

Dirbaio (approved):

peterkrull:
The last commit refers to #3336 where it avoids a panic

PR #3358, opened 2024-09-20
embassy_sync: `Sink` adapter for `pubsub::Pub`

mammothbane:
Corresponding to the `Stream` impl for `pubsub::Sub`.

Notable difference is that we need a separate adapter type to store the pending item, i.e. we can't `impl Sink for Pub` directly. Instead a method `Pub::sink(&self)` is exposed, which constructs a `PubSink`.


mammothbane:
Test failure seems spurious &mdash; not related to anything I changed, and I can't reproduce it locally with `cargo miri test`

lulf:
bender run

Dirbaio:
bender run

PR #3364, opened 2024-09-22
Fix return value of `fdcan` `write`

hansihe:
Currently the return value of `write` is broken, it never returns the previous dequeued frame even when present.

This happens because a slice of length 64 is always passed to Frame::new from within the `abort_pending_mailbox` function, causing `Frame::new` to return None.

The fix is to take a subslice of length `data_length` and pass that to `Frame::new`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 third_party
 /Cargo.toml
 out/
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thanks!

PR #3368, opened 2024-09-23
feat(embassy-net): Implement `wait_recv_ready()` + `wait_send_ready()` for UdpSocket and `wait_read_ready()` + `wait_write_ready()` for TcpSocket

AnthonyGrondin:
This provides a `readable()` implementation for `UdpSocket`, as discussed on Matrix.
This API allows a library maintainer or the end user to wait / poll a socket to see if it is readable, without dequeuing any packet. This is useful for offloading packet processing in a different part of the application.

/cc @ivmarkov

### DIFF
--- a/embassy-net/src/udp.rs
+++ b/embassy-net/src/udp.rs
@@ -103,6 +103,32 @@ impl<'a> UdpSocket<'a> {
         })
     }
 
+    /// Wait until the socket becomes readable.
+    ///
+    /// A socket is readable when a packet has been received, or when there are queued packets in
+    /// the buffer.
+    pub async fn wait_recv_ready(&self) {
+        poll_fn(move |cx| self.poll_recv_ready(cx)).await
+    }
+
+    /// Wait until a datagram can be read.
+    ///
+    /// When no datagram is readable, this method will return `Poll::Pending` and
+    /// register the current task to be notified when a datagram is received.
+    ///
+    /// When a datagram is received, this method will return `Poll::Ready`.
+    pub fn poll_recv_ready(&self, cx: &mut Context<'_>) -> Poll<()> {
+        self.with_mut(|s, _| {
+            if s.can_recv() {
+                Poll::Ready(())
+            } else {
+                // socket buffer is empty wait until at least one byte has arrived
+                s.register_recv_waker(cx.waker());
+                Poll::Pending
+            }
+        })
+    }
+
     /// Receive a datagram.
     ///
     /// This method will wait until a datagram is received.
@@ -162,6 +188,33 @@ impl<'a> UdpSocket<'a> {
             })
         })
         .await
+    }
+
+    /// Wait until the socket becomes writable.
+    ///
+    /// A socket becomes writable when there is space in the buffer, from initial memory or after
+    /// dispatching datagrams on a full buffer.
+    pub async fn wait_send_ready(&self) {
+        poll_fn(move |cx| self.poll_send_ready(cx)).await
+    }
+
+    /// Wait until a datagram can be sent.
+    ///
+    /// When no datagram can be sent (i.e. the buffer is full), this method will return
+    /// `Poll::Pending` and register the current task to be notified when
+    /// space is freed in the buffer after a datagram has been dispatched.
+    ///
+    /// When a datagram can be sent, this method will return `Poll::Ready`.
+    pub fn poll_send_ready(&self, cx: &mut Context<'_>) -> Poll<()> {
+        self.with_mut(|s, _| {
+            if s.can_send() {
+                Poll::Ready(())
+            } else {
+                // socket buffer is full wait until a datagram has been dispatched
+                s.register_send_waker(cx.waker());
+                Poll::Pending
+            }
+        })
     }
 
     /// Send a datagram to the specified remote endpoint.
--- a/embassy-net/src/tcp.rs
+++ b/embassy-net/src/tcp.rs
@@ -376,9 +376,23 @@ impl<'a> TcpSocket<'a> {
         self.io.with_mut(|s, _| s.abort())
     }
 
-    /// Get whether the socket is ready to send data, i.e. whether there is space in the send buffer.
+    /// Return whether the transmit half of the full-duplex connection is open.
+    ///
+    /// This function returns true if it's possible to send data and have it arrive
+    /// to the remote endpoint. However, it does not make any guarantees about the state
+    /// of the transmit buffer, and even if it returns true, [write](#method.write) may
+    /// not be able to enqueue any octets.
+    ///
+    /// In terms of the TCP state machine, the socket must be in the `ESTABLISHED` or
+    /// `CLOSE-WAIT` state.
     pub fn may_send(&self) -> bool {
         self.io.with(|s, _| s.may_send())
+    }
+
+    /// Check whether the transmit half of the full-duplex connection is open
+    /// (see [may_send](#method.may_send)), and the transmit buffer is not full.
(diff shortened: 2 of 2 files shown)


ivmarkov:
I think the implementation is correct.

My only regret is that this PR implements the absolute minimum necessary for `edge-mdns` to work on top of `embassy-net`, and we might need a similar, extra PR in future for `Readable` on top of a TCP socket.

How about extending the PR with a similar impl for the TCP sockets. 
There - I believe -  you would be able to use the `smoltcp` `peek(size: usize)` option with `size = 1`

===========

Implementing a similar `writable` functionality for UDP/TCP sockets:

While we don't strictly need this, it would be nice to have I think if not for anything else, then for symmetry with `readable`.

Unfortunately, `writable` might require extending `smoltcp` itself. 
What we are missing from it, is a method - `can_enqueue` on `smoltcp`'s `PacketBuffer` and then also an extra method - `is_writable` on `smotcp`'s `udp::Socket` and `tcp::Socket`, which just calls `PacketBuffer::can_enqueue`.

AnthonyGrondin:
I'd be more than happy to implement the TCP part and take a look at implementing a `writable()` API. I think we could also add similar APIs  to raw sockets.

The issue is that I don't have any code to test the TCP socket, for readable and writable (same for raw).

In edge-net for example, the `Readable` trait is only ever used for `edge-mdns` with UDP sockets.

ivmarkov:
> So I'm not familiar with `poll_fn`, but smoltcp states the following about registering wakers:
> 
> > Only one waker can be registered at a time. If another waker was previously registered, it is overwritten and will no longer be woken.
> 
> Would calling `readable()` / `writable()` multiple times simultaneously cause the previous call to never resolve?

TLDR: No, but the two tasks, where each is e.g. calling `readable().await` will start to fight each other by each one re-registering its waker, which will result in high CPU usage.

====

Elaboration (hopefully, at least partially correct):

Formally speaking, that's not a problem and the code would still work correctly. This does not mean you should do it (because of the high CPU usage).

However, you usually don't _need_ to do it anyway. 
The cases where you want to share e.g. the read half of a socket across more than one task are rare. I have had use cases where I needed to share the write half BTW - but never the read half.

Also note that as long as you share the read or write half across multiple _futures_ which are however still _within the same task_ (intra-task concurrency, as per Boats), you don't even have the problem from above, as ultimately, these futures do get the same Arc-ed Waker, as wakers are per task.


The usual case is where you want to split the socket into a read and write halves, and then you hand off these halves to two separate tasks (or often two futures in the same task). That's the reason (I suppose) why `smoltcp` has two separate registrations for reading and writing (i.e. full duplex use case).

ivmarkov:
Argh, **I need to take some rest.** :-)

No. My case is NOT about two futures calling `readable` concurrently. It is about two futures (or tasks) sharing a _single data buffer_ - yet - operating on two separate sockets. :-)

So `readable` is just like `read` after all - I don't anticipate there would be use cases where two futures (or tasks) will be interested whether the _same_ socket is readable. But then again - that would work fine without high CPU if the two futures are in the same task (i.e. `select`).

Dirbaio:
For TCP we should implement the official embedded-io [ReadReady](https://docs.rs/embedded-io/latest/embedded_io/trait.ReadReady.html), [WriteReady](https://docs.rs/embedded-io/latest/embedded_io/trait.WriteReady.html).

If we want to extend that with "async wait for", i'd name them `poll_*` and `wait_*`. It's more clear than `async fn readable()`, the name suggests it'd return a bool, as in "is this readable?". So:

```rust
// ReadReady trait impl
fn read_ready(&self) -> bool;
// WriteReady trait impl
fn write_ready(&self) -> bool;
// inherent methods
fn poll_read_ready(&self, cx: &mut Context) -> Poll<()>;
async fn wait_read_ready(&self);
fn poll_write_ready(&self, cx: &mut Context) -> Poll<()>;
async fn wait_write_ready(&self);
```

The API for UDP should be the same for consistency, except:
- it should NOT implement the traits, for the same reason we don't implement the main Read/Write traits: udp Is NOT a byte stream
- It uses "send" and "recv", not "read" and "write", so the equivalent "_ready" methods should be consistent with that.

so i'd suggest:

```rust
fn recv_ready(&self) -> bool;
fn poll_recv_ready(&self, cx: &mut Context) -> Poll<()>;
async fn wait_recv_ready(&self);
fn send_ready(&self) -> bool;
fn poll_send_ready(&self, cx: &mut Context) -> Poll<()>;
async fn wait_send_ready(&self);
```

ivmarkov:
Not that I've been asked, but I'm fine with any names. The `readable` naming came from [here](https://github.com/smol-rs/async-io/blob/master/src/lib.rs#L901).

AnthonyGrondin:
I've renamed the functions to use the them `poll_*` and `wait_*` naming conventions.

It's only missing:
```rust
fn poll_send_ready(&self, cx: &mut Context) -> Poll<()>;
async fn wait_send_ready(&self);
```
for UDP to have the full readable / writable async wait on both TCP and UDP sockets.

I'll have to check what's the best way to check if a UDP connection is ready to send

Dirbaio:
it's `.can_send()` on smoltcp UdpSocket.

embassy-net/src/udp.rs
```diff
@@ -103,6 +103,34 @@ impl<'a> UdpSocket<'a> {
...
+
+    /// Wait until a datagram can be read.
+    ///
+    /// When no datagram is readable, this method will return `Poll::Pending` and
+    /// register the current task to be notified when a datagram is received.
+    ///
+    /// When a datagram is received, this method will return `Poll::Ready`.
+    pub fn poll_recv_ready(&self, cx: &mut Context<'_>) -> Poll<()> {
+        self.with_mut(|s, _| {
+            match s.peek() {
```
Dirbaio, line 122: you can use `.can_recv()` instead of actually peeking, which avoids having to use `unreachable!()`.

AnthonyGrondin:
> it's `.can_send()` on smoltcp UdpSocket.

I can use `udp::Socket::can_send()` provided by smoltcp, but it only checks whether the tx buffer is full or not. It doesn't guarantee that calling `send()` would resolve successfully, as it will return `Err(Error::Unadressable)` if the local or remote port, or remote address are unspecified.

embassy-net/src/udp.rs
ivmarkov: @Dirbaio 

Are we sure that `can_recv()` is an OK replacement for `peek()`?

`peek()` also calls something named [dequeue_adding](https://github.com/smoltcp-rs/smoltcp/blob/a2b92ed6e8d7ae7b1229665c5b3f2e801de111e9/src/storage/packet_buffer.rs#L166), while `can_recv()` doesn't.

Using `can_recv`, can we end up in a situation where `recv_ready` becomes "ready" - yet - a following call to `read`/`recv` still pends, because the only thing in the ringbuffer was padding (whatever "padding" is)?

ivmarkov:
@Dirbaio @AnthonyGrondin Is there anything stopping this from merging? 

I have [this one comment](https://github.com/embassy-rs/embassy/pull/3368#discussion_r1775352088) which might be a concern. Hopefully, I just misunderstand the `smoltcp` API.

Dirbaio:
> I have https://github.com/embassy-rs/embassy/pull/3368#discussion_r1775352088 which might be a concern. Hopefully, I just misunderstand the smoltcp API.

okay i've checked smoltcp's source and this seems to be a bug indeed. Filed https://github.com/smoltcp-rs/smoltcp/issues/1004, will fix later.

PR #3371, opened 2024-09-24
rp23: Disable pad isolation on PWM A pins.

### LINKED ISSUE #3370: RP23 PWM Dirver only works for b pins.
I'm using the Pimorani Pico Plus 2 board (https://shop.pimoroni.com/products/pimoroni-pico-plus-2?variant=42092668289107), running Rust and Embassy.

PWM output appears to work perfectly for B pins, including the LED on PIN 25, however, A pins don't seem to have any output at all, they don't even go high if you set `config::invert_a=true;`

The following code works, i.e. I can see PIN13 toggle at about 3.8kHz with a scope:
```
    let p = init(Default::default());

    let mut c: Config = Default::default();
    c.top = 0x8000;
    c.compare_b = 0x800;
    let mut pwm: Pwm<'_> = Pwm::new_output_b(p.PWM_SLICE6, p.PIN_13, c.clone());
```

However, if I convert it to drive PIN 12  which is an A pin, then there is no output at all:
```
    let p = init(Default::default());

    let mut c: Config = Default::default();
    c.top = 0x8000;
    c.compare_a = 0x800;
    let mut pwm: Pwm<'_> = Pwm::new_output_a(p.PWM_SLICE6, p.PIN_12, c.clone());
```

This is true for other A pins as well, e.g. PIN 28.

I'm very new to Rust and especially Embassy and the RP2350, so I apologise in advance if I've made any stupid mistakes.

Nick

Edit: fixed typo in PIN12 example.

CBJamo:
Also fixes minor bug for 2040 where A pins didn't have their pull up/down enabled.

Fixes #3370

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -106,6 +106,12 @@ impl<'d> Pwm<'d> {
 
         if let Some(pin) = &a {
             pin.gpio().ctrl().write(|w| w.set_funcsel(4));
+            pin.pad_ctrl().modify(|w| {
+                #[cfg(feature = "_rp235x")]
+                w.set_iso(false);
+                w.set_pue(b_pull == Pull::Up);
+                w.set_pde(b_pull == Pull::Down);
+            });
         }
         if let Some(pin) = &b {
             pin.gpio().ctrl().write(|w| w.set_funcsel(4));
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -106,11 +106,9 @@ impl<'d> Pwm<'d> {
 
         if let Some(pin) = &a {
             pin.gpio().ctrl().write(|w| w.set_funcsel(4));
+            #[cfg(feature = "_rp235x")]
             pin.pad_ctrl().modify(|w| {
-                #[cfg(feature = "_rp235x")]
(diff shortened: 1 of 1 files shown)


embassy-rp/src/pwm.rs
```diff
@@ -106,6 +106,12 @@ impl<'d> Pwm<'d> {
 
         if let Some(pin) = &a {
             pin.gpio().ctrl().write(|w| w.set_funcsel(4));
+            pin.pad_ctrl().modify(|w| {
+                #[cfg(feature = "_rp235x")]
+                w.set_iso(false);
+                w.set_pue(b_pull == Pull::Up);
```
Dirbaio, line 112: using `b_pull` is a bit strange for pin A. Maybe pull should always be false for A, since A is always an output?

nimo-labs: I can't comment on the `b_pull` issue, however, I have just tested the fix on my board and it does appear to fix the problem. Thank you for such a quick response.

CBJamo: Ah, I thought it was weird there was only one pull argument, but wasn't smart enough to go check the DS. Fixed.

Dirbaio (approved):

PR #3372, opened 2024-09-25
Add the ability to reuse USB endpoints for multiple alternative functions

M3gaFr3ak:
Solution for https://github.com/embassy-rs/embassy/pull/3212 and https://github.com/embassy-rs/embassy/issues/2994 .

This PR addresses the problem that currently endpoint addresses are automatically increased when building a USB descriptor. It adds a copy of each endpoint-creation function, adding the `_allocated` suffix. These functions accept a &mut reference to the already created endpoint and merely write a new endpoint descriptor. They also grow the buffer of the endpoint if the new descriptor has a higher `max_packet_size` than the already allocated endpoint. The methods handling interface reconfiguration also automatically adjust all endpoints to reflect the settings in the desciptor.
The relevant functions are only implemented for embassy-rp for now, the other packages have the methods implemented with a warning, so it works like before the PR.

Suggestions regarding commit rebasing, or different function names or how it works in general, are welcome :)

### DIFF
--- a/embassy-nrf/src/usb/mod.rs
+++ b/embassy-nrf/src/usb/mod.rs
@@ -14,10 +14,8 @@ use cortex_m::peripheral::NVIC;
 use embassy_hal_internal::{into_ref, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
 use embassy_usb_driver as driver;
-use embassy_usb_driver::{
-    Direction, EndpointAddress, EndpointError, EndpointInfo, EndpointType, Event, SynchronizationType, Unsupported,
-    UsageType,
-};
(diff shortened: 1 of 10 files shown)


M3gaFr3ak:
@Dirbaio Thank you for your thoughts. I'm definitely open to changing the API. Maybe integrate the underlying logic more with the actual drivers and make the allocation an actual request instead of an "order", returning a Success/Fail Enum if the reused endpoint cannot fulfill the requested requirements?

I guess this wont impact the underlying logic that much as most of the heavy lifting then relies on a validated descriptor, after configuration.

M3gaFr3ak:
I'd invite everyone seeing this PR to contribute ideas for a better API. @Dirbaio already presented good ideas and I feel like the Logical vs Physical distinction is a good starting point. Maybe a setting in the builder for selecting allocation schemes should also be added.

Dirbaio:
Hello,

PR #3379, opened 2024-09-27
Stm32: implement async flush for UART

qwerty19106:
1) It implement non-blocking `async flush` for UART by enable Transmission complete interrupt.
2) It consist fix for **half-duplex**: either **tx** or **rx** can be enabled one moment. The #3299 can be closed after merging it, because flush should get non-mutable reference. Flush should wait only, no more.


qwerty19106:
rebased

Dirbaio:
i've added on-hardware tests that test flush (just testing that it doesn't hang, it's hard to test that it actually waits...)

Dirbaio (approved):
tests pass! thanks for the PR :D

PR #3383, opened 2024-09-30
stm32: qei to use timer low-level apis and enable time update interrupt

votrungchi:
Tested on STM32F407VG DISCOVERY KIT


Dirbaio:
what's the goal of enabling the interrupt, if the driver doesn't use it for anything?

votrungchi:
> what's the goal of enabling the interrupt, if the driver doesn't use it for anything?

Hi @Dirbaio, It enabled user can bind their own interrupt handler on demand. In my case, I need the overflow interrupt to handle the encoder multi-rotation.

Dirbaio:
I see, thanks! in that case it should be optional, otherwise it will still fire for users who don't need it, and if they haven't set up the irq handler it will go to DefaultHandler which causes a panic.

PR #3386, opened 2024-10-02
Update embedded-nal-async dependency to 0.8

aurelj:
Note that this requires rust v1.77 because this makes use of core::net.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ please refer to the `.vscode/settings.json` file's `rust-analyzer.linkedProjects
 
 ## Minimum supported Rust version (MSRV)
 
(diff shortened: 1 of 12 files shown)


aurelj:
For the `rp` example to build, we also need an updated `reqwless` that also use embedded-nal-async 0.8.

Patch for `reqwless` already submitted here : https://github.com/drogue-iot/reqwless/pull/91

In the meantime, I guess this PR will have to wait...

Dirbaio:
done in #3397 (it was easier to do it in a single pr, so both smoltcp and embedded-nal switch to `core::net` IP types at the same time. Thanks anyway for your PR!)

PR #3391, opened 2024-10-05
rp: Run RP235x at 150 MHz instead of 125

shilga:
Value taken over from PICO-SDK https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/include/hardware/clocks.h#L163


Dirbaio (approved):
:100:

PR #3393, opened 2024-10-05
embassy-sync: add clear, len, is_empty and is_full functions to zerocopy_channel

sourcebox:
Added these functions for consistency with other modules.

### DIFF
--- a/embassy-sync/src/zerocopy_channel.rs
+++ b/embassy-sync/src/zerocopy_channel.rs
@@ -70,6 +70,28 @@ impl<'a, M: RawMutex, T> Channel<'a, M, T> {
     pub fn split(&mut self) -> (Sender<'_, M, T>, Receiver<'_, M, T>) {
         (Sender { channel: self }, Receiver { channel: self })
     }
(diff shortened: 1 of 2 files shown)


embassy-sync/src/zerocopy_channel.rs
```diff
@@ -70,6 +70,28 @@ impl<'a, M: RawMutex, T> Channel<'a, M, T> {
...
     }
+
+    /// Clears all elements in the channel.
+    pub fn clear(&mut self) {
+        self.state.lock(|s| {
+            s.borrow_mut().clear();
+        });
+    }
+
+    /// Returns the number of elements currently in the channel.
```
Dirbaio, line 81: `len` in the ringbuffer is actually the capacity, ie the max amount of elements in the channel, not the current amount.

(the name is a bit unintuitive, maybe it should be renamed to `capacity`)

To get the current amount you have to calculate it from `front` and `back`.

Dirbaio (approved):
thanks!

PR #3395, opened 2024-10-06
executor/spin: introduce an architecture agnostic executor

Dummyc0m:
Spin polls the raw executor and never sleeps. It is useful for disabling
any power features associated with wfi/wfe-like instructions.

When implementing support for the CH32V30x MCU, the wfi instruction
had issues interacting with the USB OTG peripheral and appeared to be
non-spec-compliant.

1. When sending a USB Data-in packet, the USB peripheral appears to be
unable to read the system main memory while in WFI. This manifests in
the USB peripheral sending all or partially zeroed DATA packets.
Disabling WFI works around this issue.

2. The WFI instruction does not wake up the processor when MIE is
disabled. The MCU provides a WFITOWFE bit to emulate the WFE instruction
on arm, which, when enabled, ignores the MIE and allows the processor to
wake up. This works around the non-compliant WFI implementation.

### DIFF
--- a/embassy-executor-macros/src/lib.rs
+++ b/embassy-executor-macros/src/lib.rs
@@ -94,6 +94,35 @@ pub fn main_cortex_m(args: TokenStream, item: TokenStream) -> TokenStream {
     main::run(&args.meta, f, main::cortex_m()).unwrap_or_else(|x| x).into()
 }
 
+/// Creates a new `executor` instance and declares an architecture agnostic application entry point spawning
+/// the corresponding function body as an async task.
+///
+/// The following restrictions apply:
+///
+/// * The function must accept exactly 1 parameter, an `embassy_executor::Spawner` handle that it can use to spawn additional tasks.
+/// * The function must be declared `async`.
+/// * The function must not use generics.
+/// * Only a single `main` task may be declared.
+///
+/// A user-defined entry macro must provided via the `entry` argument
+///
+/// ## Examples
+/// Spawning a task:
+/// ``` rust
+/// #[embassy_executor::main(entry = "qingke_rt::entry")]
+/// async fn main(_s: embassy_executor::Spawner) {
+///     // Function body
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn main_spin(args: TokenStream, item: TokenStream) -> TokenStream {
+    let args = syn::parse_macro_input!(args as Args);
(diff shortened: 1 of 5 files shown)


ExplodingWaffle:
wow, how’d you figure that out?? was also struggling with the zeroed out packets thing but had no idea what could’ve been causing it. nice one

Codetector1374:
> wow, how’d you figure that out?? was also struggling with the zeroed out packets thing but had no idea what could’ve been causing it. nice one

 we realized that if we do blocking wait after poking send it works, but not if we do async wait. It kind of took a whole day to figure out. Curious are you also working on the usb FS? Or just usbd?

ExplodingWaffle:
Just the usbhs. got the control pipe working and then got stuck on that bug, so I haven’t started on actual endpoints yet. my code is here https://github.com/ExplodingWaffle/ch32-hal if you’re curious

Codetector1374:
> Just the usbhs. got the control pipe working and then got stuck on that bug, so I haven’t started on actual endpoints yet. my code is here https://github.com/ExplodingWaffle/ch32-hal if you’re curious

Oh thats great. We started on fs and planning on doing hs after. We can try to see what code we can cherry pick from your repo.

Dirbaio:
I'm not sure if we should add this as a "quirk" to the riscv executor. There's reasons ARM might want it as well, and once you remove the WFI there's nothing RISCV-specific about the executor anymore.

Instead, it could be a "dummy arch" that does no waiting, the main loop could just be `loop { self.inner.poll() }` and the pender would be a noop. This would work in all archs.

I'm not sure what's the best name. `arch-spin`, `arch-any` ,`arch-agnostic`?

Dummyc0m:
> I'm not sure if we should add this as a "quirk" to the riscv executor. There's reasons ARM might want it as well, and once you remove the WFI there's nothing RISCV-specific about the executor anymore.
> 
> Instead, it could be a "dummy arch" that does no waiting, the main loop could just be `loop { self.inner.poll() }` and the pender would be a noop. This would work in all archs.
> 
> I'm not sure what's the best name. `arch-spin`, `arch-any` ,`arch-agnostic`?

Thanks for the suggestion. That makes a lot of sense. We added a new arch `arch-spin`. Let me know if you have any other suggestions.

Dirbaio (approved):
thanks!

PR #3397, opened 2024-10-06
Update smoltcp, embedded-nal-async to use the `core::net` IP addr types.

Dirbaio:
Depends on
- https://github.com/smoltcp-rs/smoltcp/pull/994
- https://github.com/drogue-iot/reqwless/pull/91


Dirbaio:
merging now that smoltcp is merged. We can use reqwless from git for a bit until it's released (cc @lulf )

PR #3400, opened 2024-10-07
Add capacity, free_capacity, clear, len, is_empty and is_full functions to Channel::{Sender, Receiver}

sourcebox:
These functions are already present in the channel but not implemented in the sender and receiver structs.


Dirbaio (approved):
thanks!

PR #3402, opened 2024-10-07
embassy-sync: documentation and clippy fixes

sourcebox:
This PR does not address all clippy lints, the author of the `Watch` primitive should probably have a closer look on what clippy suggests.

### DIFF
--- a/embassy-sync/CHANGELOG.md
+++ b/embassy-sync/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


peterkrull:
> the author of the Watch primitive should probably have a closer look on what clippy suggests

I'll take a look. Does not seem like there is too much though, at least not with default clippy settings.

sourcebox:
> > the author of the Watch primitive should probably have a closer look on what clippy suggests
> 
> I'll take a look. Does not seem like there is too much though, at least not with default clippy settings.

It's mainly about using `Into` instead of the `From` trait.

lulf (approved):

PR #3404, opened 2024-10-08
stm32: fix low-power EXTI IRQ handler dropped edges

brightbloom:
The STM32 low-power executor's `on-wakeup-irq` function was being called too early and causing the interrupt handler to miss edges.

The root cause is the EXTI IRQ pending bits being wiped before they were checked - specifically `EXTI.pr(0).read().0` did not contain a trigger once `on-wakeup-irq` was called.

Moving `on-wakeup-irq` to be called after checking the EXTI Pending register solved the issue.

Before this patch, the system would not wake up on a single edge:
Screenshot 2024-10-08 at 11 10 00 AM

After the patch, the edge is recognized and the system wakes:
Screenshot 2024-10-08 at 11 07 53 AM

By the way - I'd sometimes found that when I pushed a physical button to trigger the interrupt, the interrupt would wake the system, but when I used a microcontroller to trigger a single edge, it would not. This was because of the button bounce adding additional edges which *re*-asserted the IRQ bit in time for it to be checked.

### DIFF
--- a/embassy-stm32/src/exti.rs
+++ b/embassy-stm32/src/exti.rs
@@ -41,9 +41,6 @@ fn exticr_regs() -> pac::afio::Afio {
 }
 
 unsafe fn on_irq() {
-    #[cfg(feature = "low-power")]
-    crate::low_power::on_wakeup_irq();
-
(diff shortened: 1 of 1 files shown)


Dirbaio (commented):
Thanks for the PR!

this must've been fun to debug :sweat_smile:

PR #3407, opened 2024-10-09
rp: Move pio programs into embassy-rp

CBJamo:
This is something that's been mentioned a few times and so I decided to just go ahead and do it. The main motivation is to get more people using PIO and especially sharing programs so they see more testing and refinement. 

This pr makes one significant change to all the programs, a few minor changes on some, and adds basic documentation. The significant change is to introduce a <thing>Program struct to represent a specific LoadedProgram for the given driver. This way it's easy to re-use the loaded program for multiple state machines.

The HD44780 driver needs one program for setup and one to run, I suspect there is a way to only load the setup program temporarily, but I didn't investigate that. @pennae appears to have written the original, I'd appreciate input on that.

The I made the pwm driver's `set_level` fn private, because I couldn't figure out what it actually does to document it. The example doesn't use it, and neither does the Servo example. @BjornTheProgrammer Do you have a strong opinion on that change?

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -144,6 +144,7 @@ rp2040-boot2 = "0.3"
 document-features = "0.2.7"
 sha2-const-stable = "0.1"
 rp-binary-info = { version = "0.1.0", optional = true }
+smart-leds = "0.4.0"
 
 [dev-dependencies]
 embassy-executor = { version = "0.6.0", path = "../embassy-executor", features = ["arch-std", "executor-thread"] }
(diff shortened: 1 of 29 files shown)


BjornTheProgrammer:
@CBJamo set_level probably should be public, I've documented it better in my own code, but basically, think of set_level as the raw input, and write as a helper function. I've sent this [pr](https://github.com/CBJamo/embassy/pull/1) to fix this issue.

BjornTheProgrammer:
Also I think this is a great idea. I found myself wishing that this was a provided feature and not just an example. Thank you for your work!

CBJamo:
I went ahead and merged your PR. Thanks for the improved docs. 

It's probably worth going through all the pio programs and making it so dropping the struct releases the sm/irq/pin resources, but I'm not 100% sure how to do that. I think it'd probably possible to release the sm instruction space on drop as well.

Dirbaio (approved):
this is :100:

PR #3414, opened 2024-10-11
cyw43: make bluetooth module public

jbuijgers:
The current implementation makes BtDriver and friends private, making it public will improve DX.


lulf:
I'm not sure how this improves DX, as the only public function inside the bluetooth module is for creating the internal BtState type, which is anyway created when creating the driver. Could you demonstrate a use case for why this improves DX?

PR #3419, opened 2024-10-15
Add `set_config` method to RP SPI driver

dstric-aqueduct:
Add a `set_config` method to `Spi` struct to allow reconfiguring SPI mode after creation.

The existing implementation of the `embassy-embedded-hal` trait `SetConfig` is changed to use the new method.

Existing uses of `SetConfig`  trait may need to explicitly call the trait method to maintain current return type.

### DIFF
--- a/embassy-rp/src/spi.rs
+++ b/embassy-rp/src/spi.rs
@@ -84,16 +84,9 @@ impl<'d, T: Instance, M: Mode> Spi<'d, T, M> {
     ) -> Self {
         into_ref!(inner);
 
-        let p = inner.regs();
-        let (presc, postdiv) = calc_prescs(config.frequency);
+        Self::apply_config(&inner, &config);
 
-        p.cpsr().write(|w| w.set_cpsdvsr(presc));
-        p.cr0().write(|w| {
-            w.set_dss(0b0111); // 8bit
-            w.set_spo(config.polarity == Polarity::IdleHigh);
-            w.set_sph(config.phase == Phase::CaptureOnSecondTransition);
-            w.set_scr(postdiv);
-        });
+        let p = inner.regs();
 
         // Always enable DREQ signals -- harmless if DMA is not listening
         p.dmacr().write(|reg| {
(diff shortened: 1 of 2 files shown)


embassy-rp/src/spi.rs
```diff
@@ -244,6 +244,27 @@ impl<'d, T: Instance, M: Mode> Spi<'d, T, M> {
...
+    pub fn set_config(&mut self, config: Config) {
+        let (presc, postdiv) = calc_prescs(config.frequency);
+        
+        let p = self.inner.regs();
+        // disable
+        p.cr1().write(|w| w.set_sse(false));
+
+        // change stuff
+        p.cpsr().write(|w| w.set_cpsdvsr(presc));
+        p.cr0().write(|w| {
```
Dirbaio, line 258: this is copypasted from `new_inner`. could you extract it to a private function `fn apply_config()` instead?

dstric-aqueduct: After the CI failure, I see that the `SetConfig` trait in the [embassy-embedded-hal](https://docs.embassy.dev/embassy-embedded-hal/git/default/trait.SetConfig.html) actually does what I want, so this change may not be needed. Should we consider changing the trait implementation to disable/reenable the peripheral as the `set_frequency` method does?

Dirbaio: whatever's needed for it to work. I don't know if it's required

dstric-aqueduct: The [Pico SDK disables the driver before changing frequencies](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_spi/spi.c), as does the `set_frequency` method. I've updated to add a private `apply_config` method and then reused the `Spi` struct's `set_config` method in the trait implementation.

dstric-aqueduct: This would also be a breaking change as the method's could not conflict depending on Trait in scope or not - I propose we just remove the proposed Spi `set_config` method and update the Trait implementation to disable/reenable the driver at the expense of some code duplication - I'll wait for your thoughts before touching again. Thanks for all your work.

Dirbaio (approved):
thanks!

embassy-rp/src/spi.rs
Dirbaio: don't worry about breaking changes, nextreleasewill be breaking anyway due to other changes.

Dirbaio:
- Can you fix CI?
- Can you squash all commits, and give it a good description of what you changed? "Update spi.rs" is not descriptive at all. (I recommend you do PRs by cloning the repo and using a real desktop git client, the github web ui encourages terrible practices...)

dstric-aqueduct:
> * Can you fix CI?
> * Can you squash all commits, and give it a good description of what you changed? "Update spi.rs" is not descriptive at all. (I recommend you do PRs by cloning the repo and using a real desktop git client, the github web ui encourages terrible practices...)

Squashed all the commits and cleaned up the PR description. 

I think the CI is running cached tests so the modified `spi_sdmmc` example in the PR is not propagating to CI? Not sure how to sync this.

Dirbaio:
fixed+rebased it for you at https://github.com/embassy-rs/embassy/pull/3600

(next time please tick the "allow maintainers to push" box, so I can rebase without creating a new PR. Thanks!)

PR #3429, opened 2024-10-18
blinking red and green leds

ecruzolivera:
in the example the work pins are used, this PR just makes both leds blink

### DIFF
--- a/examples/stm32f4/src/bin/blinky.rs
+++ b/examples/stm32f4/src/bin/blinky.rs
@@ -12,15 +12,18 @@ async fn main(_spawner: Spawner) {
     let p = embassy_stm32::init(Default::default());
     info!("Hello World!");
 
-    let mut led = Output::new(p.PB7, Level::High, Speed::Low);
+    let mut led_green = Output::new(p.PG13, Level::High, Speed::Low);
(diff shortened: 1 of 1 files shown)


ecruzolivera:
I dont understand the CI/build error:

```
+ git clone -n https://github.com/ecruzolivera/embassy.git code
Cloning into 'code'...
error: 6695 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
run failed: exited with code 128
```

Dirbaio:
bender run

ecruzolivera:
> This will likely break blinky for someone else, it depends on which board you are using.

I agree, but I was thinking the only official ST board for the stm32f429I was the [STM32F429I-DISCO](https://www.st.com/en/evaluation-tools/32f429idiscovery.htm), and the PR PINs are for that board's leds.

Is there a recommended way of adding an example per board?

Dirbaio:
ST also has the nucleo-stm32f429zi, the PB7 pin is the blue LED on that one.

In general we don't aim for examples to work out of the box on every single possible board, there's way too many boards. It's expected that users have to do some customization changing the chip name and pin numbers for their board.

PR #3430, opened 2024-10-19
embassy-futures: add select 5 and 6

chrenderle:
Added select with 5 and 6 futures since I need it for my project.


showier-drastic:
I have a question regarding this: why doesn't embassy have a macro like `select!` in tokio?

chrenderle:
> I have a question regarding this: why doesn't embassy have a macro like `select!` in tokio?

Might be a good reason for me to look into macros.

PR #3432, opened 2024-10-20
Add flush for UDP and Raw

### LINKED ISSUE #2193: UdpSocket created and used inside async function doesn't work properly
I wrote a small program (`ESP32`/`no_std`, using `embassy`/`embassy-net`) that sends UDP packets and it work just fine.

When I moved the networking related code into an async function, it stopped working, UDP Packets wouldn't be sent. Everything seemed to progress properly w/o any errors, but the packets weren't sent.

I pinpointed it to the move of the `UdpSocket::new` into the async function. Once it moved into the async function the packets weren't sent. For the program to work, the `UdpSocket` has to be created in the `main()` and all the rest can run inside the async function.

I thought it might have to do with #2175 but the buffers are created in the main function which never ends (so exist before and after the networking code, so no memory issues).
I tried letting the executor run after calling the new (using `Time::after`) , in case something needs to be processed before continuing in the net_task of .

Any ideas what's the reason? Is this maybe by design?

ProfFan:
Resolves #2193 

Waiting on upstream `smoltcp` PR merge

### DIFF
--- a/embassy-net/Cargo.toml
+++ b/embassy-net/Cargo.toml
@@ -68,7 +68,7 @@ multicast = ["smoltcp/multicast"]
 defmt = { version = "0.3", optional = true }
 log = { version = "0.4.14", optional = true }
 
-smoltcp = { git="https://github.com/smoltcp-rs/smoltcp", rev="b65e1b64dc9b66fa984a2ad34e90685cb0b606de", default-features = false, features = [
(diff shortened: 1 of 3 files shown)


Dirbaio:
bender run

Dirbaio:
done in #3437 (had to open another PR because you didn't check "allow maintainers to push" when creating the PR, it's nice to check that in future PRs). Thanks!

PR #3441, opened 2024-10-21
STM32: Add BufferedUart::new permutations usefull for RS485

Luctins:
I've tested this on a STM32H5 and it works without issue.

The option to turn RTS into DE was added because it can be usefull to have this mapping since ST's own tools (like CubeMX) work as such, so the distinction between DE and RTS can an issue if for for example a hardware designer uses tools that do not apply this distinction; Also the fact that the official tooling do not apply this indicates the it may be actually just a naming difference, not a functional difference.

I did not add the same permutations required for the regular Async/Blocking uart but I can if it is required.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -246,6 +246,55 @@ impl<'d> BufferedUart<'d> {
         )
     }
 
+
+    /// Create a new bidirectional buffered UART driver with only RTS pin as the DE pin
+    pub fn new_with_rts_as_de<T: Instance>(
+        peri: impl Peripheral<P = T> + 'd,
+        _irq: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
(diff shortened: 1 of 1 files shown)


Luctins:
@lulf Could you please review this?

embassy-stm32/src/usart/buffered.rs
```diff
@@ -246,6 +246,54 @@ impl<'d> BufferedUart<'d> {
...
+        tx_buffer: &'d mut [u8],
+        rx_buffer: &'d mut [u8],
+        config: Config,
+    ) -> Result<Self, ConfigError> {
+        Self::new_inner(
+            peri,
+            new_pin!(rx, AfType::input(Pull::None)),
+            new_pin!(tx, AfType::output(OutputType::PushPull, Speed::Medium)),
+            new_pin!(rts, AfType::input(Pull::None)),
+            None, // no RTS
```
8cichy8, line 289: This should be "no CTS", right?

lulf (approved):

PR #3451, opened 2024-10-23
stm32/uart impl ReadReady for RingbufferdUart

dvdsk:
I am unsure what `DmaUnsynced` means and need some help with that. For now I have added it as a separate variant for `uart::Error`. Since I do not know what it means it still needs documentation making this a draft.

### DIFF
--- a/embassy-stm32/src/dma/ringbuffer/mod.rs
+++ b/embassy-stm32/src/dma/ringbuffer/mod.rs
@@ -21,6 +21,10 @@ pub trait DmaCtrl {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum Error {
     Overrun,
+    /// the newly read DMA positions don't make sense compared to the previous
+    /// ones. This can usually only occur due to wrong Driver implementation, if
+    /// the driver author (or the user using raw metapac code) directly resets
+    /// the channel for instance.
     DmaUnsynced,
 }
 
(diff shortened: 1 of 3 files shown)


liarokapisv:
DmaUnsynced is an internal error variant of the DMA ringbuffer module. It internally keeps tracks of the previous DMA positions, this is returned when the newly read DMA positions don't make sense compared to the previous ones. This can usually only occur  due to wrong Driver implementation, if the user directly resets the channel for instance. The driver should probably not expose it to end users, better to log an error and notify the user to contact the maintainers or mayb even go nuclear and panic.

dvdsk:
> DmaUnsynced is an internal error variant of the DMA ringbuffer module. It internally keeps tracks of the previous DMA positions, this is returned when the newly read DMA positions don't make sense compared to the previous ones. This can usually only occur due to wrong Driver implementation, if the driver author (or the user using raw metapac code) directly resets the channel for instance. The driver should probably not expose it to end users, better to log an error and notify the user to contact the maintainers or maybe even go nuclear and panic. Other modules just silently convert to an Overrun error since the end results are similar, the ringbuffer can recover from the error in any case.

thank you very much for the clear explanation! I have sprinkled some of it through the source.

Dirbaio:
code LGTM, can you make CI green? thakns!

dvdsk:
should be good now :)

PR #3453, opened 2024-10-24
nrf: Add basic skeleton for AAR and ECB peripherals

plaes:
Adds a simple skeleton drivers which would eventually allow getting rid of "unstable-pacs" dependency for applications utilizing [nrf-sdc](https://github.com/alexmoon/nrf-sdc/issues/32) crate.

Currently one needs to initialize `sdc` by referencing ECB and AAR peripherals via `unstable-pac`:
```rust
{
    let p = embassy_nrf::init(Default::default());
    // Requires `unstable-pac`
    let pac_p = pac::Peripherals::take().unwrap();

    let pac_p = pac::Peripherals::take().unwrap();
    let sdc_p = sdc::Peripherals::new(
        // Peripherals supplied via `unstable-pac`:
        pac_p.ECB, pac_p.AAR,
        // ...
        p.PPI_CH17, p.PPI_CH18, p.PPI_CH20, p.PPI_CH21, p.PPI_CH22, p.PPI_CH23, p.PPI_CH24,
        p.PPI_CH25, p.PPI_CH26, p.PPI_CH27, p.PPI_CH28, p.PPI_CH29,
    );
}
```

How should I deal with shared CCM / AAR resources as these two share the same registers?

Another driver required by nrf-sdc that isn't exposed as peripheralref is CLOCK.

### DIFF
--- a/embassy-nrf/src/aar.rs
+++ b/embassy-nrf/src/aar.rs
@@ -7,7 +7,7 @@ use crate::peripherals::AAR;
 /// AAR (Accelerated Address Resolver)
 ///
 /// Note: This peripheral is current unimplemented!
-/// TODO: AAR shares resources with CCM
+/// TODO: AAR shares resources with CCM and ECB
 pub struct Aar<'d> {
     _p: PeripheralRef<'d, AAR>,
 }
--- a/embassy-nrf/src/aar.rs
+++ b/embassy-nrf/src/aar.rs
@@ -1,0 +1,13 @@
+//! Accelerated address resolver (AAR)
+//!
+use embassy_hal_internal::PeripheralRef;
+
+use crate::peripherals::AAR;
+
+/// AAR (Accelerated Address Resolver)
+///
+/// Note: This peripheral is current unimplemented!
+/// TODO: AAR shares resources with CCM
+pub struct Aar<'d> {
+    _p: PeripheralRef<'d, AAR>,
+}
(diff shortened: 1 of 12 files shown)


Dirbaio:
I'm not convinced we should add singletons for peripherals with no HAL driver yet. 

The reason is: embassy-nrf's position is to pretend PAC singletons don't exist (and they will actually not exist when we switch to [nrf-pac](https://github.com/embassy-rs/nrf-pac)). The goal of the HAL singletons is to prevent conflicts if you stick to using the HAL only. If you use the PAC you're on your own to check you don't conflict with anything. You can already break things if you e.g. use UARTE0 from the HAL with the HAL singleton and the PAC with the PAC singleton at the same time, no unsafe needed. 

So, given embassy-nrf doesn't attempt to prevent conflicts between itself and the PAC, I don't think it should aim to give a tool to nrf-sdc to prevent conflicts with the PAC either.

Adding these singletons has a slight cost, it can be a bit confusing for the user, since it looks like these peripherals are supported by the HAL but are not.

IMO nrf-sdc should just not take any PAC singletons. If a peripheral has no HAL singleton then just don't take any singleton. That's OK since the only way a user can cause a conflict is with the PAC, and it's already on them to ensure no conflicts if they're using the PAC.

> How should I deal with shared CCM / AAR resources as these two share the same registers?

I'd do like in the TWI+SPI (+ UART in nrf53/nrf91) case: make just one peripheral singleton that all the drivers take, to enforce you can only use one at a time.

Dirbaio:
closing in favor of https://github.com/alexmoon/nrf-sdc/pull/35

PR #3454, opened 2024-10-24
Embassy-rp Split Pwm driver to allow separate duty cycle control of each channel.

vivi202:
this PR allow independent controll of duty cycle for each channel in a pwm slice.
```rust 
    let pwm_ab= Pwm::new_output_ab(p.PWM_SLICE3, p.PIN_6, p.PIN_7, pwm_conf.clone());
    let (pwm_a,pwm_b)=pwm_ab.split();
    pwm_a.set_duty_cycle(dutya);
    pwm_b.set_duty_cycle(dutyb);
```

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -407,10 +407,6 @@ pub struct PwmBatch(u32);
 
 impl PwmBatch {
     #[inline]
-    pub fn new() -> Self{
-        Self(0)
-    }
-    #[inline]
     /// Enable a PWM slice in this batch.
     pub fn enable(&mut self, pwm: &Pwm<'_>) {
         self.0 |= pwm.bit();
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -407,6 +407,10 @@ pub struct PwmBatch(u32);
 
 impl PwmBatch {
     #[inline]
+    pub fn new() -> Self{
+        Self(0)
+    }
+    #[inline]
     /// Enable a PWM slice in this batch.
     pub fn enable(&mut self, pwm: &Pwm<'_>) {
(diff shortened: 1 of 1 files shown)


embassy-rp/src/pwm.rs
```diff
@@ -344,6 +344,62 @@ impl<'d> Pwm<'d> {
...
+    }
+}
+
+enum PwmChannel{
+    A,
+    B
+}
+
+/// Single channel of Pwm driver.
+pub struct PwmOutput {
```
Dirbaio, line 361: this should have a `'d` lifetime same as `Pwm`, so you can't make it outlive the peripheral singletons when you create it with borrowed singletons, like `Pwm::new(&mut p.PWM_SLICE3, ...)`.

embassy-rp/src/pwm.rs
```diff
@@ -344,6 +344,62 @@ impl<'d> Pwm<'d> {
     fn bit(&self) -> u32 {
         1 << self.slice as usize
     }
+
+    #[inline]
+    /// Split Pwm driver to allow separate duty cycle control of each channel
+    pub fn split(&self) -> (PwmOutput,PwmOutput){
```
Dirbaio, line 350: we need to ensure only one instance of the pwm channel exists. If this takes `&self` you can call it multiple times and obtain multiple copies of `PwmOutput`.

I suggest adding both an owned `split` that permanently consumes `self`, and a "temporary" split that only borrows it, but mutably so you can't call it again until you drop the `PwmOutput`s.

```rust
pub fn split(self) -> (PwmOutput<'d>, PwmOutput<'d>) {...}
pub fn split_by_ref(&mut self) -> (PwmOutput<'_>, PwmOutput<'_>) {...}
```

Dirbaio (approved):
thanks!

PR #3456, opened 2024-10-25
Add memory mapped mode for stm32 ospi

HaoboGu:
Verified on weact stm32h7b0 board + w25q64 spi flash, code at: https://github.com/HaoboGu/eg-demo/blob/058defff156bab29cc9abeef3db08750d5cc5458/src/w25q.rs#L68-L93


Dirbaio:
could you add an example showing how to use it?

HaoboGu:
sure thing! I'll add next week

HaoboGu:
just got sometime to finish this.. since the default stm32h7 example uses stm32h743, which doesn't have ospi, so I create another stm32h7b0 example folder. @Dirbaio do you have any comments/suggestions on this PR 😄

PR #3458, opened 2024-10-25
Fix LTDC registry settings for STM32U5

### LINKED ISSUE #3457: LTDC register configuration seems to be wrong on STM32U5
When using LTDC driver with a RGB screen on STM32U5 the image is corrupted. Basically, a square is not shown with vertical sides but with diagonal size, since each line is offset by some pixel with respect to the previous one.

The configuration of CFBLR register seems to require some adaptations on U5.
As per section "43.7.23 LTDC layer x color frame buffer length register (LTDC_LxCFBLR)" of Reference manual for STM32U5 RM0456, CFBLL has to be set to the length of one pixel line plus 3 (instead of plus 7 as for H7)

Changing the settings for the registry as described the image was rendered in a more reasonable way.
I have a small patch and an example.  I can submit a PR if required.

Even after this fix, however, the image was not completely correct.  Vertical lines of 1 px were rendered in a fuzzy way (too wide or too thin), as if timings were not correct, even though I put in exactly the same parameters taken from the IOC of the source code generated for the demo board by CubeMX.
In order to get a proper image, I also had to reverse all the polarities for HSYNC, VSYNC, DATA_ENABLE and PIXEL_CLOCK with respect to the settings taken from IOC.

Do you have additional suggestions on what to check and how to additionally troubleshoot the issue with the display parameters?

williams-one:
The configuration of `CFBLR` register seems to require some adaptations on U5.
As per section "43.7.23 LTDC layer x color frame buffer length register (LTDC_LxCFBLR)" of Reference manual for STM32U5 RM0456, CFBLL has to be set to the length of one pixel line plus 3 (instead of plus 7 as for H7)

Closes #3457 

Before the fix when using LTDC driver with a RGB screen on STM32U5 the image is corrupted. Basically, a square is not shown with vertical sides but with diagonal sides, since each line is offset by some additional pixel with respect to the previous one.
Changing the settings for the registry as described the image was rendered in a more reasonable way.

The attached example is derived from `examples\stm32h735\src\bin\ltdc.rs` and was tested on an STM32U5G9J-DK2 demo board

**NOTE**: Even after the fix, the image was not completely correct on the screen.  Vertical lines of 1 px are rendered in a fuzzy way (too wide or too thin), as if timings were not correct, even though I put in exactly the same timing parameters taken from the IOC of the source code generated for the demo board by CubeMX.
In order to get a proper image, I also had to reverse all the polarities for HSYNC, VSYNC, DATA_ENABLE and PIXEL_CLOCK with respect to the settings taken from IOC.
So maybe some additional changes are required to properly support the peripheral


Dirbaio (approved):

williams-one:
@Dirbaio Can you please help me out here?

ci/build is failing because the chip configured in the Cargo.toml (which I didn't change) is `stm32u585ai` which doesn't have LTDC, so the build of `ltdc.rs` fails.

Is it OK if I change it to `stm32u5g9zj` and adjust `i2c.rs` and `usb_serial.rs` pinout accordingly (otherwise they fail)?  Do you have another suggestion?

Dirbaio:
changing it sounds good

williams-one:
Thanks for the help.  I have updated the chip to the one included in the STM32U5G9J-DK2 demo board and updated all the examples with the correct pins for the new chip

Dirbaio (commented):
thanks!

PR #3463, opened 2024-10-25
Update main.rs - Adjust EXTI to match with EXTI in interrupt example

ckrenslehner:


lulf:
bender run

Dirbaio:
I don't think this is right. The interrupt is EXTI15_10, but the EXTI channel is indeed EXTI13.

ckrenslehner:
> I don't think this is right. The interrupt is EXTI15_10, but the EXTI channel is indeed EXTI13.

You are right! I guess somewhere inside the embassy crate there is the mapping from the channel to the IRQ, I did not find where right now, but there has to be. Also it makes at lot of sense to abstract this away!

My bad sorry :-)

PR #1660, opened 2023-07-16
add spi_display with mipidsi example for rp pico

Georges760:
This PR is another example of rp spi display but using the generic mipidsi crate (https://github.com/almindor/mipidsi).

I used the GC9A01 round AMOLED display as I have thi one on hand.

Unfortunately the GC9A01 model has been added recently to mipidsi and is not in v0.7.1 tag (latest). So I use the git version for now.

PR #3467, opened 2024-10-27
Update example spi_display for RP/RP23 to use mipidsi instead of ST7789 crate

Georges760:
I dont have the actual LCD to test and confirm the orientation will be OK. But I have another hardware with GC9A01 and tested it OK on #1660 that I will update once this PR is merged.


lulf (approved):
LGTM, thanks!

PR #3468, opened 2024-10-27
This fixes 2 issues where STM32 BXCAN would hang

korken89:
1. If one received frames under an `interrupt_free` section using `try_read`, in my case `init` in RTIC, the RX IRQ will fire and clear it's enable bit after `interrupt_free` is complete. There is no frame to read so RX is now unconditionally disabled forever.
2. On clearing of RX IRQ, TX stops silently. This happens due to the use of `write` instead of `modify` when modifying IRQ enable bits.

Solution 1: Enable RX IRQs on `try_read` when no frames are available. This solution also solves the issue of very delayed handling of the RX IRQ which would cause the same issue.

Solution 2: Use `modify` instead of `write`.


Dirbaio (approved):
thanks!

PR #3469, opened 2024-10-27
Fix `embassy_rp::spi::new_rxonly` not taking in required `tx_dma`

Meigs2:
Fixes an issue with rxonly spi instances in embassy_rs::spi. DMA reads still (apparently) require a tx_dma channel to function.

Commit simply adds a required tx_dma parameter to `spi::new_rxonly`. I've tested this on an RP2040, using a MAX31855.


Dirbaio (approved):
thanks!

PR #3475, opened 2024-10-29
STM32 Qspi async

### LINKED ISSUE #3473: Stm32 QSPI hangs on command with no-`NONE` awidth or dwidth
I was doing a command like this:

```rust
self.qspi.command(TransferConfig {
    instruction: Command::ChipErase as u8,
    dummy: DummyCycles::_0,
    iwidth: QspiWidth::QUAD,
    dwidth: QspiWidth::QUAD,
    awidth: QspiWidth::QUAD,
    address: None,
});
```

Mostly really because of copy pasting the config from reads and writes.
But this makes the QSPI hang since the code polls the transfer complete flag and the QSPO thinks the address or data isn't complete.

It's fixed by using the appropriate widths:

```rust
self.qspi.command(TransferConfig {
    instruction: Command::ChipErase as u8,
    dummy: DummyCycles::_0,
    iwidth: QspiWidth::QUAD,
    dwidth: QspiWidth::NONE,
    awidth: QspiWidth::NONE,
    address: None,
});
```

Not sure what to do about it.

Maybe an assert? Maybe the command function should ignore it and always pass NONE to the transfer setup function?

### LINKED ISSUE #3474: Stm32 QSPI async incomplete
The async QSPI does already use DMA, but only waits on it blockingly. Is there a reason why the DMA can't be awaited to make it proper async?

diondokter:
Expose an async API and add some sanity checks.

Closes #3474 
Closes #3473 

- [x] Test on actual hardware

### DIFF
--- a/embassy-stm32/src/qspi/mod.rs
+++ b/embassy-stm32/src/qspi/mod.rs
@@ -202,16 +202,19 @@ impl<'d, T: Instance, M: PeriMode> Qspi<'d, T, M> {
     }
 
     fn setup_transaction(&mut self, fmode: QspiMode, transaction: &TransferConfig, data_len: Option<usize>) {
-        if let (Some(_), QspiWidth::NONE) = (transaction.address, transaction.awidth) {
-            panic!("QSPI address can't be sent with an address width of NONE");
+        match (transaction.address, transaction.awidth) {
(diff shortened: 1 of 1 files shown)


diondokter:
Oh, actually I need to add a couple of extra checks...

diondokter:
Alright, I think everything is covered now

Dirbaio (approved):
thanks!

PR #3478, opened 2024-10-29
Reexported some smoltcp types for raw socket.

Frostie314159:
This PR reexports some types from smoltcp for use with raw sockets.

### DIFF
--- a/embassy-net/src/raw.rs
+++ b/embassy-net/src/raw.rs
@@ -5,7 +5,7 @@ use core::mem;
 use core::task::{Context, Poll};
 
 use embassy_net_driver::Driver;
-pub use smoltcp::iface::{Interface, SocketHandle};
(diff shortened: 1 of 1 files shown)


embassy-net/src/raw.rs
```diff
@@ -5,10 +5,10 @@ use core::mem;
 use core::task::{Context, Poll};
 
 use embassy_net_driver::Driver;
-use smoltcp::iface::{Interface, SocketHandle};
+pub use smoltcp::iface::{Interface, SocketHandle};
```
Dirbaio, line 8: why do you need these? they're not used anywherein the RawSocket public API.

Frostie314159: I mistook `RawSocket::with_mut` to be public. So these aren't required, I'll remove them when I get home.

Dirbaio (approved):

Issue #2059, opened 2023-10-14
Add support for Ringbuffers on gpdma


tyler-gilbert:
The DMA on newer devices like the STM32U5 have a GPDMA hardware block. This is distinct from the classic DMA found on, say, STM32F4 devices. The classic DMA has built-in support for circular reads and writes. GPDMA hardware doesn't have built-in support for circular DMA is capable of circular DMA using it's linked-list design structure.

The simplest way to have circular reads/writes on GPDMA is to provide a single linked-list item that resets the write destination for DMA reads and the read source for DMA writes. Considering reads, the LLR is set to update only the source address to point to the start of the buffer.

It looks something like this to set up:

```rust
    if options.circular {
        STATE.circular_address[channel.index()] = mem_addr as u32;
        let lli1 = &STATE.circular_address[channel.index()] as *const u32 as u32;
        if lli1 & 0b11 != 0 {
            panic!("circular address must be 4-byte aligned");
        }
        ch.llr().write(|w| {
            if options.circular {
                w.set_usa(true); //update only the source address
                // lower 16 bites of the address of STATE.circular_address[channel.index()]
                w.set_la((lli1 >> 2usize) & 0x3fff);
            }
        });
        ch.lbar().write(|w| {
            if options.circular {
                // upper 16 bits of the address of STATE.circular_address[channel.index()]
                w.set_lba(lli1 >> 16usize);
            }
        });
    } else {
        ch.llr().write(|_| 0);
    }
```

The GPDMA will need to store a 32-bit memory address in static memory for each channel (STM32U5 has 15-channels, so 60 bytes of RAM): `STATE.circular_address[channel.index()]`. The lower 2 bits must be zero (requires 4-byte alignment) are always zero. Bits 3 to 15 are stored use `w.set_la((lli1 >> 2usize) & 0x3fff)` while the upper 16-bits are stored using `w.set_lba(lli1 >> 16usize)`.

I still need to do some testing to make sure this approach will work but this seems like a good approach to match the circular transfers of classic DMA. Fully supporting linked-list could maybe come later as a cargo feature.

Comments and suggestions are welcome. If I can get this working, I will open a PR.

xoviat:
linked-list dma is on the todo list but has not yet been implemented. the goal is to match the current public api for readable and writable ringbuffer.

> Comments and suggestions are welcome. If I can get this working, I will open a PR.

if you can implement a readable/writable ringbuffer for gpdma, it would be a welcome contribution.

[...]

PR #3480, opened 2024-10-30
feat: stm32 gpdma readable ringbuffer

Luctins:
This is based on previous work: #2059 by @tyler-gilbert and #3162 by @caleb-garrett and mainly aims to make the ringbuffered usart work with gpdma targets. I've only removed the gpdma feature gates around usart since it was what I could test and validate.

The original PR's did contain a writable ringbuffer, but since I won't able to validate if it works, it is left out for a future PR when relevant.

I've also included a test for it and also have tried to not break existing modules depending on the current `RingBuffer` (since there was need to change the dma irq)

Finally, I think it would be interesting to create a new trait for the Gpdma ringbuffers, since they each are used mostly interchangeably and it 'just works' because all of the implementations have the same functions. An example of this issue would be the `request_suspend` method which was missing at first from @tyler-gilbert 's implementation but is used by the ringbuffered usart implementation.

@Dirbaio could you review this?

### DIFF
--- a/embassy-stm32/src/dma/gpdma.rs
+++ b/embassy-stm32/src/dma/gpdma.rs
@@ -1,13 +1,14 @@
 #![macro_use]
 
-use core::future::Future;
+use core::future::{poll_fn, Future};
 use core::pin::Pin;
-use core::sync::atomic::{fence, Ordering};
-use core::task::{Context, Poll};
+use core::sync::atomic::{fence, AtomicUsize, Ordering, AtomicBool};
+use core::task::{Context, Poll, Waker};
 
 use embassy_hal_internal::{into_ref, Peripheral, PeripheralRef};
 use embassy_sync::waitqueue::AtomicWaker;
 
+use super::ringbuffer::{DmaCtrl, ReadableDmaRingBuffer, self};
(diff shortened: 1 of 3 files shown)


osobiehl:
Hi,
I'm also interested in getting this PR merged. I rebased @Luctins branch and fixed compilation issues. I'm also not a great fan of duck typing my way with named methods using GPDMA or normal DMA, but I'm not knowledgeable enough to come up with a good abstraction.

here is a quick branch where I rebased and fixed compilation issues. Also tested manually on my stm32u575 and it seems to work.

Let me know if there's any way to help to get this merged :) 

https://github.com/embassy-rs/embassy/compare/main...osobiehl:embassy:master

DrTobe:
I have used @osobiehl's repo and I can confirm that this works.

"ringbuffers on gpdma" has been around at least since Oct 23, what is missing to get this merged?

tyler-gilbert:
Please see https://github.com/embassy-rs/embassy/pull/4025 which has Readable and Writeable Ring buffers on STM32 GPDMA tested using the SAI unit.

Dirbaio:
done in #3923

PR #3486, opened 2024-11-01
Add ReceiverHandler to logger

BjornTheProgrammer:
This makes it a lot easier for someone to handle incoming data to a serial connection with `embassy-usb-logger`. It adds optional parameter to the `run`, `with_class`, and `with_custom_style` macros to accept a struct that implements RecieverHandler. 

An example of this has been added, but the interface looks like the below.

```Rust
struct Handler;

impl ReceiverHandler for Handler {
    async fn handle_data(&self, data: &[u8]) {
        if let Ok(data) = str::from_utf8(data) {
            let data = data.trim();

            // If you are using elf2uf2-term with the '-t' flag, then when closing the serial monitor,
            // this will automatically put the pico into boot mode
            if data == "q" || data == "elf2uf2-term" {
                reset_to_usb_boot(0, 0); // Restart the chip
            } else if data.eq_ignore_ascii_case("hello") {
                log::info!("World!");
            } else {
                log::info!("Recieved: {:?}", data);
            }
        }
    }

    fn new() -> Self {
        Self
    }
}

#[embassy_executor::task]
async fn logger_task(driver: Driver<'static, USB>) {
    embassy_usb_logger::run!(1024, log::LevelFilter::Info, driver, Handler);
}
```

### DIFF
--- a/embassy-usb-logger/src/lib.rs
+++ b/embassy-usb-logger/src/lib.rs
@@ -3,6 +3,7 @@
 #![warn(missing_docs)]
 
 use core::fmt::Write as _;
+use core::future::Future;
 
 use embassy_futures::join::join;
 use embassy_sync::pipe::Pipe;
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thanks!

PR #3488, opened 2024-11-01
embassy-net: disable multicast filtering

kkrolczyk:
- disable multicast filtering out packets (which seems to be problematic especially in ipv6, which relies heavily on multicast) (edit: updated to be unconditional as requested)
- fix a couple clippy items, make logs slightly cleaner, using hex like for example wireshark

### DIFF
--- a/embassy-net/src/driver_util.rs
+++ b/embassy-net/src/driver_util.rs
@@ -84,7 +84,7 @@ where
     {
         self.0.consume(|buf| {
             #[cfg(feature = "packet-trace")]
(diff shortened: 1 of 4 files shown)


Dirbaio:
can you make it so the multicast filter is unconditionally disabled? ie no Cargo feature. We want things to work out of the box. 

(There's little reason you'd want the multicast filter enabled because there's currently no API to use it and no support in smoltcp to obtain the list of "wanted" multicast adresses either so using it would be hard)

Dirbaio (approved):
thank you!

PR #3489, opened 2024-11-01
Also implement `embedded_io::Write` for `UartTx<'d, T: Instance, Blocking>`

flippette:
I already implemented `embedded_io::Write` for `Uart<'d, T: Instance, Blocking>`, it makes sense to also implement it for this one.


Dirbaio (approved):
thank you!

Issue #3493, opened 2024-11-02
panic! in embassy-usb-synopsys-otg on second serial connection (BUT only if defmt-trace level debugging not enabled)


paulc:
I had a strange issue running the `usb_serial` example on an STM32F401 where the serial device would work fine for the first connection from the host, but if I disconnected the terminal session and then re-connected the driver would panic with the following message.

```
95.600067 [panic_probe::print_defmt::print] lib.rs:104  ERROR panicked at /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/lib.rs:46:9:
assertion failed: ep_num < ep_count
```

Very strangely, if I enabled the `defmt` feature in embassy-stm32 crate and set DEFMT_LOG=trace the problem would go away. As an aside, enabling the `defmt` feature in the embassy-usb-synopsys-otg results in lots of compile errors with `defmt::assert!` (see #3492) (constant function errors) which makes debugging a bit tricky.

It looks like the driver is getting an unexpected response from [let status = r.grxstsp().read()](https://github.com/embassy-rs/embassy/blob/fcbbef01cd3c5292be29b78b674f0593277545e7/embassy-usb-synopsys-otg/src/lib.rs#L41) (status.epnum() was 12 rather than 1 when the error showed up in my testing).

The patch below fixes this by replacing the [assert!(ep_num < ep_count)](https://github.com/embassy-rs/embassy/blob/fcbbef01cd3c5292be29b78b674f0593277545e7/embassy-usb-synopsys-otg/src/lib.rs#L46C17-L46C23) with a test which skips the packet rather than panic-ing.

I suspect that there is a more fundamental issue here but but this does at least fix the panic! issue.

```
diff --git a/embassy-usb-synopsys-otg/src/lib.rs b/embassy-usb-synopsys-otg/src/lib.rs
index f90403936..3197817ee 100644
--- a/embassy-usb-synopsys-otg/src/lib.rs
+++ b/embassy-usb-synopsys-otg/src/lib.rs
@@ -43,7 +43,16 @@ pub unsafe fn on_interrupt<const MAX_EP_COUNT: usize>(
         let ep_num = status.epnum() as usize;
         let len = status.bcnt() as usize;
 
-        assert!(ep_num < ep_count);
+        if ep_num >= ep_count {
+            // Removed `assert!(ep_num < ep_count);` as this fails for second serial connection UNLESS
[...]

Issue #3492, opened 2024-11-02
`defmt` feature doesn't work in embassy-usb-synopsys-otg


paulc:
If you enable the `defmt` feature in the embassy-usb-synopsys-otg this generates a large number of errors when `defmt::assert!` is used instead of `core::assert!`

```
error[E0015]: cannot call non-const fn `defmt::export::acquire` in constant functions
   --> /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/fmt.rs:16:13
    |
16  |             ::defmt::assert!($($x)*);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/otg_v1.rs:212:9
    |
212 |         assert!(n < 7usize);
    |         ------------------- in this macro invocation
    |
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants
    = note: this error originates in the macro `::defmt::assert` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0658]: referencing statics in constant functions is unstable
   --> /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/fmt.rs:16:13
    |
16  |             ::defmt::assert!($($x)*);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/otg_v1.rs:212:9
    |
212 |         assert!(n < 7usize);
    |         ------------------- in this macro invocation
    |
    = note: see issue #119618 <https://github.com/rust-lang/rust/issues/119618> for more information
    = note: `static` and `const` variables can refer to other `const` variables. A `const` variable, however, cannot refer to a `static` variable.
    = help: to fix this, the value can be extracted to a `const` and then used.
    = note: this error originates in the macro `::defmt::assert` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

error: pointers cannot be cast to integers during const eval
   --> /Users/paulc/Sync/Development/embedded-rust/embassy/embassy-usb-synopsys-otg/src/fmt.rs:16:13
    |
16  |             ::defmt::assert!($($x)*);
[...]

PR #3494, opened 2024-11-02
Fix for #3493 - panic! in embassy-usb-synopsys-otg on second serial connection

paulc:
Fix for #3493 - panic! in embassy-usb-synopsys-otg on second serial connection


Dirbaio:
ideally we should try to figure out why the packet is getting corrupted and how to prevent that, instead of just ignoring it.

paulc:
I'd agree that it isn't ideal, I'm happy to do some more digging if someone could point me in the right direction but it's probably getting a bit deep into the USB stack for me (I'm just starting out with Embassy/STM32). It would probably help if #3492 (defmt feature doesn't work in embassy-usb-synopsys-otg) was fixed which would make debugging easier.

Dirbaio:
fixed in #3565

PR #3495, opened 2024-11-02
Enhanced docs for time driver

### LINKED ISSUE #3487: Imprecisions in `embassy-time-driver` API
I am implementing the time-driver for a chip, and I have some questions that are not answered by the docs:
- Once the caller has allocated an alarm, should it assume that it is still allocated after the alarm has rang? This is what I understand from the implementation on stm32 boards.
- Should the caller be expected to set the callback before each call to `set_alarm`? I think the answer is no, from the stm32 driver as well.

I think these cases should be added to the docs of the `Driver` trait as it would benefit implementers. I am ready to open a PR once I have answer to the questions.

carloskiki:
Fixes #3487.


Dirbaio (approved):
thanks!

PR #3496, opened 2024-11-03
low-power: add support for stm32u5

chrenderle:
I can confirm it works on my Nucleo-U575ZI-Q. The power consumption measured is ~16uA.


Dirbaio (approved):
thanks!

trembel:
Thanks for that! Wanted to do it just now myself, and saw after cloning the repo that it's already done :D

PR #3503, opened 2024-11-04
executor: allow overriding `embassy_executor` path in `task` macro

kaspar030:
This PR allows overriding the `embassy_executor` crate path.

We're [wrapping](https://future-proof-iot.github.io/RIOT-rs/dev/docs/api/riot_rs/attr.task.html) the `embassy_executor::task` to provide some convenience.

Without this PR, users of `riot_rs::task` must also explicitly add `embassy_executor` to their application's dependencies, even though riot-rs re-exports it.

### DIFF
--- a/embassy-executor-macros/src/macros/task.rs
+++ b/embassy-executor-macros/src/macros/task.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use darling::export::NestedMeta;
 use darling::FromMeta;
 use proc_macro2::{Span, TokenStream};
@@ -11,6 +13,9 @@ use crate::util::*;
 struct Args {
     #[darling(default)]
     pool_size: Option<syn::Expr>,
+    /// Use this to override the `embassy_executor` crate path. Defaults to `::embassy_executor`.
+    #[darling(default)]
+    embassy_executor: Option<syn::Expr>,
 }
 
 pub fn run(args: TokenStream, item: TokenStream) -> TokenStream {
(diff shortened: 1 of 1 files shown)


lulf:
Interesting.... I added a similar capability to embassy in the past, but it was removed in [bd0aaec6242c86d126fdb3784ae603cbf329957b](https://github.com/embassy-rs/embassy/commit/bd0aaec6242c86d126fdb3784ae603cbf329957b), as it was a bit hard to support the path prefix as the macros evolved.

I think it would be better if you simply copy the needed macros and maintain them independently, which would allow you to adjust them to your needs as well.

CC @Dirbaio since you removed them the previous time around.

kaspar030:
> I think it would be better if you simply copy the needed macros and maintain them independently, which would allow you to adjust them to your needs as well.

Hm, having to maintain a copy would be unfortunate, as apart from the path, the macro does exactly what we need, and it evolves.

Now that there is a user again, maybe we can have this back?

This adds ~5 significant lines for setting up the `#embassy_executor` variable, that doesn't seem unreasonable to support?

lulf:
> > I think it would be better if you simply copy the needed macros and maintain them independently, which would allow you to adjust them to your needs as well.
> 
> Hm, having to maintain a copy would be unfortunate, as apart from the path, the macro does exactly what we need, and it evolves.
> 
> Now that there is a user again, maybe we can have this back?
> 
> This adds ~5 significant lines for setting up the `#embassy_executor` variable, that doesn't seem unreasonable to support?

Yeah, it seems like a pretty small change, though I don't think we can provide any guarantees wrt. not accidentally breaking this. So if that's OK with you, then I guess we can merge it.

kaspar030:
> Yeah, it seems like a pretty small change, though I don't think we can provide any guarantees wrt. not accidentally breaking this. So if that's OK with you to commit to fixing issues, then I guess we can merge it.

I'll happily commit to that. We're here to stay and embassy is at the heart of our project.

kaspar030:
Thanks!

PR #3507, opened 2024-11-06
fix(*): FirmwareUpdater::get_state in cases where WRITE_SIZE != 1

MathiasKoch:
Has previously been fixed for `set_magic`, but `get_state` seems to have been overlooked.


lulf (approved):
Thanks!

PR #3508, opened 2024-11-06
Detect and allow older nightlies

bugadani:
As there is no nightly Xtensa-enabled compiler, we need embassy-executor to support APIs a bit further back.


Dirbaio (approved):
thanks!

PR #3515, opened 2024-11-07
STM32: add support for LPTIM time driver

chrenderle:


Dirbaio:
Most STM32s have LPTIMs with only 1 channel, U5 is the only one with two. The existing time driver needs at least two channels, one for the half-period overflow and another for at least one user's alarm. This means there's no way it's going to work on other STM32s.

I think it'd be best to write a separate driver just for LPTIMs, designed for 1 channel. It should be possible to do without the mid-overflow interrupt if you take it into account in `now()` with a retry loop.

chrenderle:
I will do that. This is just to show that it works. Now that I have a understanding of the time driver and lptim I can implement it with only 1 channel.

chrenderle:
@Dirbaio could you please have a look at it? I implemented the time driver as you said. This is only a first draft. I need to check everything again. I also need to check the whole memory ordering stuff. 

It can confirm it works on my Nucleo U575ZI-Q.

Dirbaio:
Hello,

PR #3518, opened 2024-11-09
Enable user to choose to pass lse clock to peripherals

chrenderle:
This is needed for #3515 when the LPTIM is supposed to be clocked by the LSE.

### DIFF
--- a/embassy-stm32/src/rcc/bd.rs
+++ b/embassy-stm32/src/rcc/bd.rs
@@ -20,6 +20,9 @@ pub enum LseMode {
 pub struct LseConfig {
     pub frequency: Hertz,
     pub mode: LseMode,
+    /// If peripherals other than RTC/TAMP or RCC functions need the lse this bit must be set
+    #[cfg(any(rcc_l5, rcc_u5, rcc_wle, rcc_wl5, rcc_wba))]
+    pub peripherals_clocked: bool,
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


chrenderle:
In the future maybe it can automatically be detected when a peripherals is supposed to be clocked by the LSE.

chrenderle:
Apparently the LSESYSEN bit doesn't exist for every STM32 series. I need to patch this.

Dirbaio:
what would be the downside of unconditionally enabling LSESYS? (if it exists, and if the user requests LSE to be enabled, ofc). Why is it a different bit?

chrenderle:
LSESYS leads to higher power consumption

chrenderle:
I checked all the RMs. The bit exists only for L5, U5 and WL. I changed the code accordingly

chrenderle:
@Dirbaio is there a reason against merging this? If yes I'd like to know what I can change/improve

Dirbaio:
Rebased so it picks up the new xtensa test.

Dirbaio (approved):

Dirbaio (approved):
thanks!

Dirbaio:
Seems it causes the RTC test to hang on u5, wl5, wba.

embassy-stm32/src/rcc/bd.rs
```diff
@@ -235,6 +250,15 @@ impl LsConfig {
...
 
             while !bdcr().read().lserdy() {}
+
+            #[cfg(any(rcc_l5, rcc_u5, rcc_wle, rcc_wl5, rcc_wba))]
+            if let Some(lse_sysen) = lse_sysen {
+                bdcr().modify(|w| {
+                    w.set_lsesysen(lse_sysen);
+                });
+
+                while !bdcr().read().lsesysrdy() {}
```
Dirbaio, line 260: maybe it's this loop that hangs if lse_sysen=false.

chrenderle: Ah yes. I can only wait for lsesysrdy when I set lsesyson=1. When the user chooses lsesys=0 I can't wait for this. Will fix this later

chrenderle:
fixed this in the new commit

PR #3519, opened 2024-11-09
fix(nrf5340-app): added SPIM4 peripheral

igiona:
Added the high-speed SPIM4 for the nrf5340-app.
This peripheral has no slave version, nor it's shared as the other SERIALx peripherals.


Dirbaio (approved):
thanks!

Issue #3417, opened 2024-10-15
embassy_sync::watch default values


Gibbz:
It would be nice to be able to supply default values for the new watch module.
https://docs.embassy.dev/embassy-sync/git/default/watch/index.html

Currently I have something like this to setup default:
```rust
// default values
pub fn init() {
    CLOCK_HOURS_WATCH.dyn_sender().send_if_modified(|value| { *value = Some(0u8); false });
    CLOCK_MINUTES_WATCH.dyn_sender().send_if_modified(|value| { *value = Some(0u8); false });
}

pub static CLOCK_HOURS_WATCH: Watch<WatchMutex, u8, 1> = Watch::new();
pub static CLOCK_MINUTES_WATCH: Watch<WatchMutex, u8, 1> = Watch::new();
```

It could be much neater if we could simply include the default in the new:
```rust
pub static CLOCK_HOURS_WATCH: Watch<WatchMutex, u8, 1> = Watch::new(0u8);
pub static CLOCK_MINUTES_WATCH: Watch<WatchMutex, u8, 1> = Watch::new(0u8);
```

Gibbz:
1. I have some watches that I use for Async and in loops that I'm polling like a mutex. So in that case I don't want to have an uninitialised value.

2. I'm storing a lot of settings, which need an initial state stored in them.

They're my main use cases in my project.

PR #3521, opened 2024-11-10
embassy_sync::watch default values

Gibbz:
Add patch for the following issue: https://github.com/embassy-rs/embassy/issues/3417
Add a new_with() to embassy_sync::watch which accepts default data.

### DIFF
--- a/embassy-sync/src/watch.rs
+++ b/embassy-sync/src/watch.rs
@@ -308,7 +308,7 @@ impl<M: RawMutex, T: Clone, const N: usize> Watch<M, T, N> {
                 receiver_count: 0,
             })),
         }
-}
+    }
 
     /// Create a new `Watch` channel.
     pub const fn new_with(data: Option<T>) -> Self {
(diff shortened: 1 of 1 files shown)


embassy-sync/src/watch.rs
```diff
@@ -310,6 +310,18 @@ impl<M: RawMutex, T: Clone, const N: usize> Watch<M, T, N> {
         }
     }
 
+    /// Create a new `Watch` channel.
+    pub const fn new_with(data: Option<T>) -> Self {
```
lulf, line 314: Rather than taking an option here, it should  pass the value T, since passing None it should use the regular new() instead.

lulf (changes requested):
LGTM but i would prefer it not taking an option in this case

embassy-sync/src/watch.rs
Gibbz: Makes sense to me! I've updated that 👍

lulf (approved):
LGTM, thanks!

Issue #3431, opened 2024-10-20
PriorityChannel need to access the array of items and delete items from the queue


Gibbz:
Im setting up a command queue for my project. I want to delete duplicate items. So I want to get a list of all the items so i can search through them. Basically doing a custom filter.
Either direct access to the mutex which is private.
Or some way to use a predicate to delete items?
Or can I delete items during the order sortings?

```
pub static QUEUE_CHANNEL: PriorityChannel::<CriticalSectionRawMutex, BleCommand, Min, 16> = PriorityChannel::new(); 
```

Gibbz:
Any chance to get this merged?

Gibbz:
I'm having trouble with the CI build on this. Is there something I'm missing?

PR #3523, opened 2024-11-11
Added remove_if to priority channel

Gibbz:
Added remove_if to priority channel. This allows removing items from the queue if there is a match in the predicate.
You can also do this on a sender, receiver and the channel directly itself.

In my sample code I use the following:
```
let send_ble_queue = QUEUE_CHANNEL.sender();
send_ble_queue.remove_if(|i| { i.handle == handle && i.is_single_instance() });
```

This is for my issue here: https://github.com/embassy-rs/embassy/issues/3431

### DIFF
--- a/embassy-sync/src/priority_channel.rs
+++ b/embassy-sync/src/priority_channel.rs
@@ -72,17 +72,6 @@ where
         self.channel.poll_ready_to_send(cx)
     }
 
-    /// Removes the elements from the channel that satisfy the predicate.
-    ///
-    /// See [`PriorityChannel::remove_if()`]
-    pub fn remove_if<F>(&self, predicate: F)
-    where
-        F: Fn(&T) -> bool,
-        T: Clone,
-    {
-        self.channel.remove_if(predicate)
-    }
-
     /// Returns the maximum number of elements the channel can hold.
     ///
     /// See [`PriorityChannel::capacity()`]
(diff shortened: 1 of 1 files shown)


Dirbaio:
- It feels wrong that the sender can remove items from the channel. Ideally the sender should only be able to add, the receiver should only be able to remove.
- I'm not sure about the feature, it feels a bit ad-hoc to your particular use case. If you need to dedup maybe you want global bools or signals denoting "does X need to be done?" instead of a queue, that will dedup "for free"

Gibbz:
Can you give me a bit more information on how you would implement your second point? 

I'm using the que as some items need to sent in order, such as multiple log events. 

Most everything else simply needs priority. And these items really only need to send the latest information, so the older information can be dropped off it had not been sent yet. 

Is this something you want to merge, or should I just implement it specifically for my project?

lulf:
Filtering/selectors are not unusual when using queues, but I think this should only be on the receiver end. One alternative is to use one queue per item you wish to filter on. On the sender side you can anyway filter before you push so there's not really a need for additional APIs there.

Gibbz:
Okay thanks. I'll update the pr soon 👍

Gibbz:
I've updated this removing the sender from using the function. It doesn't seem to be building for some reason. Any idea why this is getting stuck?

Gibbz:
Hey guys, any progress on why this check is not starting?

lulf:
bender run

lulf (approved):

PR #3525, opened 2024-11-12
Allow building executor on v1.82 xtensa toolchains

wetheredge:
This is a followup for #3508 that does not rely on the compiler build date, since the Xtensa toolchain's build dates don't match the official ones. In particular, the Xtensa 1.82 releases were built in late October, but still have the old api.

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -58,7 +58,7 @@ critical-section = { version = "1.1", features = ["std"] }
 trybuild = "1.0"
 
 [build-dependencies]
-rustc_version = "0.4.1"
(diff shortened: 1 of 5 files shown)


bugadani:
Eh yeah I forgot that our compiler is built as a nightly, from the stable branch, so I guess it worked for 1.81 because it was built on the 3rd of September, before the change landed - but 1.82 was already branched off at that point.

I'd prefer this PR over WakerHack as I think `safety: OK because WakerHack has the same layout as Waker` is incidental and not guaranteed to not change randomly.

Dirbaio:
Fixed in #3528

PR #3541, opened 2024-11-16
Disallow SAI start without an initial write

elagil:
Previously, when starting the SAI (and its ring buffer + DMA channel), then writing data to the ring buffer later, the SAI DMA has already been running for some (undeterministic) amount of time. It only makes sense to start the DMA after there is already data in the transmit ring buffer present.

This PR disables `start` for SAI in transmit mode, and starts the DMA just after the transmit ring buffer was filled with an initial amount of data.

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -828,7 +828,6 @@ impl<'a, W: Word> ReadableRingBuffer<'a, W> {
     /// You must call this after creating it for it to work.
     pub fn start(&mut self) {
         self.channel.start();
-        self.clear();
     }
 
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/dma/dma_bdma.rs
```diff
@@ -828,7 +828,6 @@ impl<'a, W: Word> ReadableRingBuffer<'a, W> {
     /// You must call this after creating it for it to work.
     pub fn start(&mut self) {
         self.channel.start();
-        self.clear();
```
elagil, line 831: Clear on start conflicts with `write_immediate`. With `clear` in place, `write_immediate` would have no effect.

Dirbaio (approved):

PR #3545, opened 2024-11-17
Add a function that waits for any SAI/ringbuffer write error

elagil:
Below is a use case, where there is a `channel` that gets supplied with audio sample arrays to play on `sai`.
As soon as there is an overrun error, `sai` shall be dropped and freshly created, so that it doesn't play garbage.

```rust
let mut sai = make_a_new_sai();
let mut renew_sai = false;

loop {
    if renew_sai {
        renew_sai = false;
        drop(sai);
        sai = make_a_new_sai();
    }

    let channel_receive_fut = channel.receive();
    let sai_error_fut = sai.write_error();

    let samples = match select(channel_receive_fut, sai_error_fut).await {
        Either::First(x) => x,
        Either::Second(_) => {
            renew_sai = true;
            continue;
        }
    };

    sai.write(&samples).await.unwrap();
}
```

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -1007,9 +1007,9 @@ impl<'a, W: Word> WritableRingBuffer<'a, W> {
     }
 
     /// Wait for any ring buffer write error.
-    pub async fn write_error(&mut self) -> Result<usize, Error> {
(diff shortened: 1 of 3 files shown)


elagil:
Maybe something to review for @kalkyl

kalkyl:
Thanks! Looks good to me 🙌

PR #3550, opened 2024-11-19
Only access the current time once per polling

bugadani:


bugadani:
This is a net loss when there are no timers running.

PR #3552, opened 2024-11-20
nrf: Add NFCT driver.

Dirbaio:
Based on original work in #2113 by @i404788 and @turbocool3r . Duly credited in `Co-Authored-By`.


embassy-nrf/src/nfct.rs
```diff
@@ -0,0 +1,409 @@
...
+use core::future::poll_fn;
+use core::sync::atomic::{compiler_fence, Ordering};
+use core::task::Poll;
+
+use embassy_hal_internal::{into_ref, PeripheralRef};
+use embassy_sync::waitqueue::AtomicWaker;
+pub use vals::{Bitframesdd as SddPat, Discardmode as DiscardMode};
+
+use crate::interrupt::InterruptExt;
+use crate::pac::nfct::vals;
```
turbocool3r, line 21: This part breaks on nrf5340 (it should be `nfct_s`).

turbocool3r:
Thanks! Been waiting for this to happen for a while 😄

turbocool3r:
I can probably test this thing at some point in the feature (December-January). So far the only problem I've observed is sending an empty frame leads to a hang. Also there is some workaround in nrfx for an issue when no response is sent to an RX'ed frame, likely needs testing as well.

embassy-nrf/src/nfct.rs
Dirbaio: it's not anymore, make sure you're using the latest nrf-pac.

Dirbaio:
is sending an empty frame even allowed?

turbocool3r:
> is sending an empty frame even allowed? are there any card protocols out there that need it?

I don't think so, but I think this needs to be handled with an error or an early exit to prevent the hang.

PR #3555, opened 2024-11-21
Improve CAN Interface

sheepmax:
BRS is off by default for extended frame construction,  so building a frame with BRS enabled is not trivial without adding the `embedded_can` crate.

This PR does the following:
- Add a wrapper for embedded_can::Id, so they can be constructed with Embassy style errors and without external packages.
- Make the interface of the CAN Header struct homogeneous with the rest of the module.


Dirbaio:
I'm not sure what's the advantage of the `Id` wrapper. 

If the issue is you need to depend on `embedded-can` to do something, it should be fixable by reexporting more things (`pub use embedded_can::somethingsomething`)

sheepmax:
The other advantage is turning an option-type error into something more familiar to Embassy errors. This may also be the bases for future refactoring as it provides a more complete API to build on. I don't see anything wrong with a reexport either, but it certainly doesn't feel clean.

Dirbaio:
Hello,

PR #3556, opened 2024-11-21
low-power: add support for stm32u0

chrenderle:
Adding low-power support for the stm32u0. This PR depends on embassy-rs/stm32-data#540.

I can measure ~9uA for a blinky with 0.5Hz LED frequency. During stop the power consumption is ~2uA when the LED is off. The measurements are on my Nucleo-U083RC.
low-power-stop2-u0
low-power-stop2-u0-2

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -70,6 +70,7 @@ use crate::rtc::Rtc;
 
 static mut EXECUTOR: Option<Executor> = None;
 
+#[cfg(not(stm32u0))]
 foreach_interrupt! {
     (RTC, rtc, $block:ident, WKUP, $irq:ident) => {
         #[interrupt]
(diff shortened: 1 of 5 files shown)


chrenderle:
Now all errors should be fixed apart from those depending on the PR.

Dirbaio (approved):
thanks!

Dirbaio:
is the ci failure is due to stm32-metapac needing updating? you can do it by changing the commit hash in Cargo.toml (in two places)

chrenderle:
> is the ci failure is due to stm32-metapac needing updating? you can do it by changing the commit hash in Cargo.toml (in two places)

yes, the errors are because of the stm32-metapac update. I will change the hashes

chrenderle:
@Dirbaio do I just take the latest hash or the one with the changes I need?

Dirbaio:
ideally latest!

Issue #3334, opened 2024-09-14
Support for Dynamically Allocated Executor


PegasisForever:
The current statically allocated task system works well for many projects, but I’ve encountered challenges when trying to implement more complex functionality:

### 1. Static Task Allocation Tightly Couples Code with Embassy

Currently, all tasks must be defined within the root application using Embassy, making it impossible for libraries to create tasks. In my project, the core logic is housed in a separate crate that is hardware-agnostic and independent of any executor, allowing me to test the logic on a PC without hardware. However, due to the limitation of not being able to create tasks within the crate, I’ve had to rely on `join!` macros to handle over 30 futures, which has led to significant performance issues ([reference](https://embassy.dev/book/#_multiple_tasks_or_one_task_with_multiple_futures)).

### 2. Static Task Allocation Cannot Handle Generics (#2454)

From what I understand, static tasks inherently cannot support generics because the task size must be known at compile time. In my case, the crate depends on generic drivers passed in at runtime, so even if Embassy allowed task definitions within a crate, generics would still present a limitation.

### 3. Inefficient Memory Usage When Only One Task Runs at a Time

Consider a device with multiple operating modes, where only one mode is active at any given time. Defining separate tasks for each mode would unnecessarily reserve memory for all tasks, even though only one will run. While combining all modes into a single task and using conditionals can mitigate this (allowing the Rust compiler to optimize task size), this solution becomes unmanageable when dealing with more complex task combinations.

### Downsides of Dynamic Task Allocation

While dynamic allocation introduces common issues such as heap fragmentation and unpredictable performance, the flexibility it provides could solve the problems I’m facing with static tasks.

### Proposal & Implementation

- The task arena in Embassy already functions somewhat like a dynamic executor, albeit with (intentional?) limitations. Technically, adding support for fully dynamic task allocation shouldn’t be too difficult.
- I recognize that part of Embassy's design philosophy is the guarantee that "if it compiles, it works" and a dynamic executor might compromise that simplicity. I also understand that this feature might not align with Embassy's long-term goals.
- Regardless, I’m interested in exploring this feature for my project. Any guidance on where to start in the codebase or potential pitfalls would be greatly appreciated.

wllenyj:
As I know, one of the design goals of Embassy is no dynamic memory allocation. In other words, No alloc crate.
Maybe add `alloc` features?

lulf:
[...]

PR #3558, opened 2024-11-21
Dynamically Allocated Executor

PegasisForever:
Implementation for #3334

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -62,6 +62,8 @@ critical-section = { version = "1.1", features = ["std"] }
 ## Enable nightly-only features
 nightly = ["embassy-executor-macros/nightly"]
 
+alloc = []
+
 # Enables turbo wakers, which requires patching core. Not surfaced in the docs by default due to
 # being an complicated advanced and undocumented feature.
 # See: https://github.com/embassy-rs/embassy/pull/1263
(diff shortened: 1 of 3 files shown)


Dirbaio:
I'm not sure that we should do this. My reasoning:

- You can already use `alloc` to allocate tasks with the current public API with `Box::leak(Box::new(TaskStorage::new()))`. This allows using generics and type inference and everything. Note this allows some things that the API in this PR doesn't allow:
  - You can reuse the storage by spawning it again when a task exits, if you keep the `TaskStorage` around.
  - You can use another allocator, it doesn't have to be `Box`. Anything that'll give you a `&'static TaskStorage` will work.
- Dynamically allocated tasks still have severe downsides, mainly that you can't *free* the memory. This applies to both this PR and the `TaskStorage` API. This is unfixable, it's due to the core executor design (tasks living forever removes the need for refcounting wakers, which is what allows `embassy-executor` to be so fast+small). "Fixing" it would require adding refcounts and vtables so the executor design would basically collapse to the same as `async-task`.

I think no matter what we do `embassy-executor` is going to be a *bad* alloc executor, therefore we shouldn't encourage it. It's fine if it's *possible* to use `TaskStorage` with `alloc`, because it's "advanced" use that you have to go out of your way to do. Adding a more visible API like in this PR is basically telling people "hey go use this". They'll run into the limitation, or maybe not realize task storages are leaked and run into OOMs.

I do agree `alloc` has its use cases. We might want an `embassy-alloc-executor` crate, as a completely separate executor implementation. This way people would decide upfront whether they want alloc, picking between a GOOD alloc executor and a GOOD no-alloc executor. No weird limitations. The "alloc executor" would have the same design as `async-task`, fully refcounting tasks, or perhaps use `async-task` itself.

I do think we should expand docs on how to use `TaskStorage` with `alloc`. Crate docs, examples, FAQ entries would help. It's quite neat and there's little awareness of it.

PegasisForever:
@Dirbaio interesting, I'll experiment with the approaches you mentioned!

Dirbaio:
i'm going to close this for now for the reasons discussed above.

PR #3559, opened 2024-11-22
Generalize AtomicWaker

bugadani:
Multi-core systems might want to use different locks than a critical section. While AtomicWaker isn't too complex, I'm hoping we don't need to duplicate it.


lulf (approved):

bugadani:
```
 ERROR teleprobe::client > WARN - send_jtag_command 242 failed: SwdApWdataError
ERROR - Run failed: An ARM specific error occurred.

Caused by:
    0: The debug probe encountered an error.
    1: An error which is specific to the debug probe in use occurred.
    2: Command failed with status SwdApWdataError
```

Lol is this issue spreading?

PR #3562, opened 2024-11-22
Bump sync version

bugadani:
There were multiple rustc version bumps in the codebase since the last release. At least `collapse_debuginfo` sets MSRV to 1.79 which probably was more relaxed before, so I'm proposing a ~minor~ ~major~ minor release.

### DIFF
--- a/examples/rp/src/bin/bluetooth.rs
+++ b/examples/rp/src/bin/bluetooth.rs
@@ -14,7 +14,7 @@ use embassy_rp::bind_interrupts;
 use embassy_rp::gpio::{Level, Output};
 use embassy_rp::peripherals::{DMA_CH0, PIO0};
 use embassy_rp::pio::{InterruptHandler, Pio};
-use embassy_sync_06::blocking_mutex::raw::NoopRawMutex as NoopRawMutex06;
+use embassy_sync::blocking_mutex::raw::NoopRawMutex;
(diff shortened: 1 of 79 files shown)


Dirbaio:
> so I'm proposing a minor release.

bumping 0.6.0->0.7.0 is a major release though.

bugadani:
Oh this is equally as much fun as we had with an inter-repo dependency cycle. Can I leave the failing examples on old embassy-sync?

examples/rp/src/bin/bluetooth.rs
```diff
@@ -14,7 +14,7 @@ use embassy_rp::bind_interrupts;
 use embassy_rp::gpio::{Level, Output};
 use embassy_rp::peripherals::{DMA_CH0, PIO0};
 use embassy_rp::pio::{InterruptHandler, Pio};
-use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+use embassy_sync_06::blocking_mutex::raw::NoopRawMutex as NoopRawMutex06;
```
bugadani, line 17: Let it be known that I'm not trying to hide this, but I also can't do anything against this :(

Dirbaio: we should remove the example in this repo, to break the circular dep. examples in the trouble repo are more maintained.

PR #3566, opened 2024-11-23
STM32: Implement Asynchronous ADC Reading for G4

IvanLi-CN:
I referenced the implementation of [the G0 series](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/adc/v3.rs).

It currently works correctly on my STM32G431CBU6.


Dirbaio (approved):
thanks!

PR #3574, opened 2024-11-26
feat: Add new feature to enable overclocking

itswenb:
Add new feature `unchecked-overclocking` to disables all assertions during RCC register initialization.

### DIFF
--- a/embassy-stm32/src/fmt.rs
+++ b/embassy-stm32/src/fmt.rs
@@ -10,12 +10,15 @@ compile_error!("You may not enable both `defmt` and `log` features.");
 macro_rules! rcc_assert {
     ($($x:tt)*) => {
         {
-            if cfg!(feature = "unchecked-overclocking") {
+            #[cfg(not(feature = "unchecked-overclocking"))]
+            {
                 #[cfg(not(feature = "defmt"))]
                 ::core::assert!($($x)*);
                 #[cfg(feature = "defmt")]
                 ::defmt::assert!($($x)*);
-            } else {
+            }
+            #[cfg(feature = "unchecked-overclocking")]
+            {
                 #[cfg(feature = "log")]
                 ::log::warn!("`rcc_assert!` skipped: `unchecked-overclocking` feature is enabled.");
                 #[cfg(feature = "defmt")]
--- a/embassy-stm32/src/fmt.rs
+++ b/embassy-stm32/src/fmt.rs
@@ -7,6 +7,25 @@ use core::fmt::{Debug, Display, LowerHex};
 compile_error!("You may not enable both `defmt` and `log` features.");
 
 #[collapse_debuginfo(yes)]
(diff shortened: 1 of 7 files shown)


Dirbaio:
thanks for the PR! this is definitely a nice feature to have.

To reduce verbosity, could you make a `rcc_assert!` macro that either expands to assert or to nothing? this way we don't need a `#[cfg...]` line per assert. 

Also, could you add it to all other RCCs?

itswenb:
> thanks for the PR! this is definitely a nice feature to have.
> 
> To reduce verbosity, could you make a `rcc_assert!` macro that either expands to assert or to nothing? this way we don't need a `#[cfg...]` line per assert.
> 
> Also, could you add it to all other RCCs?

Thank you for the feedback! I've made the changes as suggested. Regarding the other RCCs, I’d be happy to work on them in the future, but at the moment, I’m not familiar with the other chips. I’ll need some time to get up to speed on those. :)

embassy-stm32/src/fmt.rs
```diff
@@ -6,6 +6,25 @@ use core::fmt::{Debug, Display, LowerHex};
 #[cfg(all(feature = "defmt", feature = "log"))]
 compile_error!("You may not enable both `defmt` and `log` features.");
 
+#[collapse_debuginfo(yes)]
+macro_rules! rcc_assert {
+    ($($x:tt)*) => {
+        {
+            if cfg!(feature = "unchecked-overclocking") {
```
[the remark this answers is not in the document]
itswenb: 😱 Sorry, I will change!

itswenb:
## Tested on `stm32f103ze`
> Test

lulf (approved):
LGTM!

embassy-stm32/src/rcc/g4.rs
```diff
@@ -229,16 +229,15 @@ pub(crate) unsafe fn init(config: Config) {
...
+    rcc_assert!(max::SYSCLK.contains(&sys));
 
     // Calculate the AHB frequency (HCLK), among other things so we can calculate the correct flash read latency.
     let hclk = sys / config.ahb_pre;
-    assert!(max::HCLK.contains(&hclk));
+    rcc_assert!(max::HCLK.contains(&hclk));
 
     let (pclk1, pclk1_tim) = super::util::calc_pclk(hclk, config.apb1_pre);
     let (pclk2, pclk2_tim) = super::util::calc_pclk(hclk, config.apb2_pre);
-    assert!(max::PCLK.contains(&pclk2));
```
8cichy8, line 240: Was this meant to be removed, or was it a typo and should have been kept and changed to &pclk1?

itswenb: I think it was a typo, so I removed it because it was duplicated. If needed, I can create another ticket to fix it.

PR #3575, opened 2024-11-26
Use cyw43 FW file size instead of magic numbers

### LINKED ISSUE #3470: rp cyw43 firmware flashing and from_raw_parts len parameter
I'm wondering how you calculate `230_321` and `4_752` for the `len` parameter?

https://github.com/embassy-rs/embassy/blob/80cb2bc4a5b8bfa424126dc7cf7c957c6894628b/tests/rp/src/bin/cyw43-perf.rs#L48-L52

```
  File: cyw43-firmware/43439A0.bin
  Size: 231077    	Blocks: 456        IO Block: 4096   regular file

  File: cyw43-firmware/43439A0_clm.bin
  Size: 984       	Blocks: 8          IO Block: 4096   regular file
```

And maybe would you be so kind and explain the address usage?

Wouldn't it makes more sense to use `0x101c_0000` and `0x101f_b000` depending on the original `len` values?  
And moving it up more if `clm` `len` doesn't need `4752` but based on the size something like `1024`?!

9names:
This should remove the possibility of the firmware slice length being incorrect when the firmware is updated next time.
Would be nice to also have a test that runs the bluetooth firmware so we can have an assert that the firmware blobs don't overlap -  I don't expect this to be a serious concern though.
Resolves #3470

### DIFF
--- a/examples/rp/src/bin/bluetooth.rs
+++ b/examples/rp/src/bin/bluetooth.rs
@@ -44,9 +44,13 @@ async fn main(spawner: Spawner) {
     //     probe-rs download 43439A0.bin --format bin --chip RP2040 --base-address 0x10100000
     //     probe-rs download 43439A0_clm.bin --format bin --chip RP2040 --base-address 0x10140000
     //     probe-rs download 43439A0_btfw.bin --format bin --chip RP2040 --base-address 0x10141400
-    //let fw = unsafe { core::slice::from_raw_parts(0x10100000 as *const u8, 224190) };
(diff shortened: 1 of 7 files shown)


Dirbaio:
bender run

Dirbaio:
cyw43 HIL test is failing for some reason.

Also I don't remember whether the pico w in CI has the latest firmware flashed. if the size doesn't match it's because it has an older one, in which case changing it will break it.

Dirbaio:
Hello,

PR #3576, opened 2024-11-26
Executor: Separate claimed and spawned states

bugadani:
A task is now `CLAIMED` unil its executor is set, and explicitly marked `SPAWNED`. This means wake events no longer enqueue a task - `run_enqueue` will return false if the executor has not yet been updated.


bugadani:
Reasoning about this separately from the other PR is harder than I aimed for.

Issue #3561, opened 2024-11-22
Flash sector erase does not work properly on STM32U5


william-spinelli:
I'm testing flash sector erase on a STM32U5G and it doesn't work properly in all conditions.
I have tried to analyze the issue and here are my findings.  If you want, I'm more than happy to work on a patch 😄 

I think that this could be a summary of all the concurrent causes.

First, all U5 are dual bank but there is no selection of bank register (BKER) in `blocking_erase_sector`

https://github.com/embassy-rs/embassy/blob/fb8e40bdf5c51a8eef00fc5f6f2e8958797ae3f8/embassy-stm32/src/flash/u5.rs#L69-L79

Then there is a problem with page size.  According to ST specs
![image](https://github.com/user-attachments/assets/c07dfa43-4887-4c50-978a-7442edfa9fcb)

But the `erase_size` in the metapac is set to 16384, while I think it should be 8192 to match the size of the page
https://github.com/embassy-rs/stm32-data-generated/blob/5373e57121b8300a3c2f03fb0408d857f6622607/stm32-metapac/src/chips/stm32u5g9zj/metadata.rs#L8-L18

This has been fixed for some variants of the U5 family but not for all
https://github.com/embassy-rs/stm32-data/blob/55b491ee982e52f80a21276a0ccbde4907982b5d/stm32-data-gen/src/memory.rs#L352-L353

Finally there is also another problem in the function for setting the PNB (page number) register here
https://github.com/embassy-rs/stm32-data-generated/blob/5373e57121b8300a3c2f03fb0408d857f6622607/stm32-metapac/src/peripherals/flash_u5.rs#L506-L516

According to ST specs PNB should be 8 bit, so I think the masking is wrong.  It should probably 0xFF instead of 0x7F
![image](https://github.com/user-attachments/assets/dc5955c1-c625-4e29-a57c-1593a9ace9d2)

This is not a problem for devices which has 2MB of flash or less since the maximum page index is 127, but it is a problem for devices with 4MB of flash which have a maximum page index of 255

PR #3583, opened 2024-11-27
STM32U5: Add flash bank selection when erasing a sector

williams-one:
This addresses the problem in https://github.com/embassy-rs/embassy/issues/3561
In the current implementation of `blocking_erase_sector` for U5 there is no selection of the flash bank that contains the page to be erased.
As a consequence, only pages in bank 1 can be erased.  If blocking_erase_sector is called for a page in bank 2, then the corresponding page in bank 1 is cleared.

The implementation mimics what has already implemented for H5.

This same problem has already been raised in the past, and exactly this fix has been proposed, but never implemented.  See here
https://github.com/embassy-rs/stm32-data/pull/464#issuecomment-2050792236

At the moment the bank swap is not supported

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -72,7 +72,7 @@ rand_core = "0.6.3"
 sdio-host = "0.5.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "15" }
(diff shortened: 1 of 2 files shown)


williams-one:
Thanks @Dirbaio!!

PR #3593, opened 2024-11-30
Rework time-driver contract.

bugadani:
This PR simplifies the contract between embassy-time-driver and implementations to a pair of `now` and `schedule_wake` functions. Implementing a timer queue is now the responsibility of the time-driver implementation, the generic-queue or integrated-timer features no longer provide a complete implementation, it's now the responsibility of the time-driver/timer-queue-driver implementor.

If using the executor, integrated-timers are implied, but the users may choose `_generic-queue` if they want to.

Companion changes in esp-hal: https://github.com/esp-rs/esp-hal/pull/2701


bugadani:
Hmm, should I restore the queue size features, yet still expose the const generic? I'd like to not duplicate the features in every HAL, but I also don't need them so keeping them is somewhat confusing.

bugadani:
This PR, as it is, has one known issue besides the build failure:

- It is possible to call `schedule_wake` with the waker of a task that has exited and is just now getting respawned. If it is already in the SPAWNED state, but its executor pointer is not yet set up, it will be enqueued in the timer queue of the old executor. This issue is fixed in #3573

Dirbaio (approved):
:rocket:

PR #3595, opened 2024-12-01
stm32/usart: Implement data bit selection

Sizurka:
Currently the `data_bits` field of STM32 USART configuration is ignored and the USART is always configured for an 8-bit word size.  This implements the actual selection of the word size and adds a 7-bit mode (usually seen in combination with parity and ASCII-only protocols).


Dirbaio (approved):
very nice! thank you :)

PR #3599, opened 2024-12-02
allow custom bcd usb version

LegitCamper:
Allows a custom bcdusb version like 1.1 (0x011) but leaves the default to 2.0/2.1

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -8,6 +8,13 @@ use crate::types::{InterfaceNumber, StringIndex};
 use crate::{Handler, Interface, UsbDevice, MAX_INTERFACE_COUNT, STRING_INDEX_CUSTOM_START};
 
 #[derive(Debug, Copy, Clone)]
+/// Allows Configuring the Bcd USB version below 2.1
+pub enum BcdUsbVersion {
+    Two = 0x0200,
+    TwoOne = 0x0210,
+}
+
+#[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 /// Configuration used when creating [`UsbDevice`].
@@ -18,7 +25,7 @@ pub struct Config<'a> {
     /// Device BCD USB version.
     ///
     /// Default: `0x0210` ("2.1")
-    pub bcd_usb: u16,
+    pub bcd_usb: BcdUsbVersion,
 
     /// Device class code assigned by USB.org. Set to `0xff` for vendor-specific
     /// devices that do not conform to any class.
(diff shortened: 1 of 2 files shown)


embassy-usb/src/descriptor.rs
```diff
@@ -326,11 +326,11 @@ pub(crate) fn device_descriptor(config: &Config) -> [u8; 18] {
         18,   // bLength
         0x01, // bDescriptorType
         0x10,
-        0x02,                     // bcdUSB 2.1
-        config.device_class,      // bDeviceClass
-        config.device_sub_class,  // bDeviceSubClass
-        config.device_protocol,   // bDeviceProtocol
-        config.max_packet_size_0, // bMaxPacketSize0
+        (config.bcd_usb >> 8) as u8, // bcdUSB
```
Dirbaio, line 329: the `0x10` above is also part of `bcd_usb`, it's the lower byte. for example USB 2.1 is `0x0210`.

The config default should also be `0x0210` instead of `0x02`

Dirbaio:
what's the use case for this? in what cases do you want to force the USB version down to 2.0 or 1.x? 

Would these use cases still be satisfied if the API was a "usb version" enum instead of a freeform u16? IMO that'd be more user-friendly because it'd prevent invalid values.

Also, do we need to disable some features that weren't available in older USBs?

LegitCamper:
> what's the use case for this? in what cases do you want to force the USB version down to 2.0 or 1.x?

I think in many cases where you are trying to emulate a specific device. When using 2.1 the host assumes you are using a bos_descriptor since that is what 2.1 is. By forcing the use of 2.0 you are disabling unused features. Maybe the same is true for 1.x, but I'm not sure what other changes might be required for that use case.

> Would these use cases still be satisfied if the API was a "usb version" enum instead of a freeform u16? IMO that'd be more user-friendly because it'd prevent invalid values.

I think so, I can easily create one for 2.1, and 2.0 and others can include others when needed.



> Also, do we need to disable some features that weren't available in older USBs?

In my testing, no everything seems to work as is.

embassy-usb/src/builder.rs
```diff
@@ -15,6 +22,11 @@ pub struct Config<'a> {
     pub(crate) vendor_id: u16,
     pub(crate) product_id: u16,
 
+    /// Device BCD USB version.
+    ///
+    /// Default: `0x0210` ("2.1")
+    pub bcd_usb: BcdUsbVersion,
```
Dirbaio, line 28: I'd name this just `UsbVersion`. 

"BCD" is a reference to Binary Coded Decimal, ie how it's coded in binary. if it's an enum the user no longer cares about that.

Dirbaio (commented):
thanks! :rocket:

PR #3603, opened 2024-12-03
Free tx buffer on ipc send failure

lulf:


lulf:
Converted to draft as I wanna test it first.

Dirbaio (approved):

Issue #2951, opened 2024-05-16
embassy-nrf: Timer HAL broken for extended timers (TIMER3/TIMER4)


peter9477:
The embassy-nrf timer.rs implementation has an impl_timer! macro that allows defining "extended" timers (which is used for TIMER3 and TIMER4 on chips where those are supported).  This creates a constant CCS equal to 4 or 6, which is used in the Timer::new() to initialize the CC registers.  This will panic at run time when TIMER3 or TIMER4 are used to create a Timer, because the impl_timer! macro involves only pac::timer0, which has only 4 CC registers.

As noted in the discussion below https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$dC6N7X1MOSACQwlS0GI9wZsCKI1HS-yGsHlTPTAAEt8?via=matrix.org&via=tchncs.de&via=grusbv.com this will need to reference pac::timer3, possibly via a regs_ext() implementation in the trait (as suggested by @Dirbaio).

peter9477:
We looked at this and couldn't see an obvious and simple way to make the impl_timer macro somehow handle both pac::timer0 and pac::timer3.  It looked like by using a few traits and trait bounds it would be feasible, but we've tried something much simpler which appears to work.

https://github.com/embassy-rs/embassy/commit/aceeeddb907e1fde946c900c6a86122b6b760924

This is done only for 52840 for the moment, as a proof of concept.  Basically since the timer3 register block is a pure superset of the timer0 register block, and since the Timer type is generic on the CCS count so it knows how many compare registers it can access, by always using timer3 when it's defined for a given chip it appears this can resolve the problem.  (We tested and it avoids the panic.)

Is this an acceptable solution?  If so we can apply the changes for the other chips and PR this.

Dirbaio:
yes, it is acceptable! it's not elegant but there's no better solutions I can think of. The STM32 HAL does similar things in some places.

peter9477:
Just added #3605 with our workaround for this.  I'm not sure whether how I made the PR makes sense, since I see there are experimental changes (to use the new PAC) on main which conflict with this. Let me know if I can do this more effectively another way (but don't use big words... I'm a bit clueless about proper use of github and PRs etc.)

peter9477:
So I pulled the aforementioned PR as it wasn't applied against main/HEAD.

That said, we've now migrated our code base to the latest "main" and it appears TIMER3 is no longer causing the panic.  

[...]

PR #3605, opened 2024-12-03
Workaround for #2951 (extended timers were not actually usable)

peter9477:
I don't know if I'm creating this the right way, but the list of changes is correct. (Apologies if the trivial tab -> space changes are undesirable.)

The changes were tested against nrf52840 and work fine. I reviewed the other chips and confirmed that all with "extended" in any impl_timer!() now reference timer3, while the others reference timer0 (which is as they should).

Would be happy to see someone conceive of a cleaner or more obvious way to do this.


peter9477:
Closing as apparently I should do this only against main/HEAD, with the chiptool PAC changes included. Will try again another time when we've been able to migrate our code to use that stuff.

PR #3606, opened 2024-12-04
Implemented MultiwriteNorFlash for stm32

guruprasath-sterna:
This is needed for Sequential Storage crate delete operation.


diondokter:
Hi! This is not possible in the general case. Not all STMs support it.
Maybe your specific part could support it? But then the PR needs to be more fine-grained.

My guess is that only STMs without ECC flash can  support multiwrite

guruprasath-sterna:
Hi! Sorry. I'm new to stm32 and I didn't know this. I will close this PR then. Thanks

Dirbaio:
not all STM32s do, but some do! You can use `cfg`s to add the impl only to those that do.

PR #3607, opened 2024-12-04
derive `defmt::Format` for interrupt enums

de-vri-es:
This PR derives the `defmt::Format` trait for enum interrupts if the `defmt` feature is enabled.

One concern: do all crates the define interrupt enums have a `defmt` feature? I'm hoping CI will tell me 😇


lulf (approved):
LGTM!

PR #3613, opened 2024-12-05
Add support for stm32u595/5a5 OTG_HS in client mode

mubes:
Simple changes to support OTG-HS on the stm32u5 family. New to this so yell up if there's something to be done differently.


embassy-stm32/src/usb/otg.rs
```diff
@@ -1,5 +1,6 @@
 use core::marker::PhantomData;
 
+use crate::pac::rcc::vals::Otghssel;
```
MDr164, line 3: Given that I added this value to the PAC create I think it was behind a feature flag as not all chips actually contain this clock. Looks like CI is also complaining about it on unrelated (non STM32U5) boards.

embassy-stm32/src/usb/otg.rs
```diff
@@ -26,7 +27,6 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
     unsafe fn on_interrupt() {
         let r = T::regs();
         let state = T::state();
-
```
MDr164, line 29: I'd not remove that empty line but that's really just a nitpick

embassy-stm32/src/usb/otg.rs
```diff
@@ -311,6 +314,26 @@ impl<'d, T: Instance> Bus<'d, T> {
             });
         });
 
+        #[cfg(all(stm32u5, peri_usb_otg_hs))]
+        {
+            crate::pac::RCC.ccipr2().modify(|w| {
+                w.set_otghssel(Otghssel::PLL1_P);
```
Dirbaio, line 320: this overrides the user's choice in `config.rcc.mux.otghssel`. you shouldn't write to it explicitly, rcc init already does with the value the user chose in `config.rcc.mux.otghssel`.

embassy-stm32/src/usb/otg.rs
```diff
@@ -103,15 +103,18 @@ impl<'d, T: Instance> Driver<'d, T> {
     pub fn new_hs(
         _peri: impl Peripheral<P = T> + 'd,
         _irq: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
-        dp: impl Peripheral<P = impl DpPin<T>> + 'd,
-        dm: impl Peripheral<P = impl DmPin<T>> + 'd,
+        _dp: impl Peripheral<P = impl DpPin<T>> + 'd,
```
MDr164, line 106: Is there a reason to apply the `_` naming here? Usually that is done for objects that are unused but you call `set_as_ref` on them or am I missing something?

embassy-stm32/src/usb/mod.rs
```diff
@@ -26,13 +26,23 @@ fn common_init<T: Instance>() {
...
     #[cfg(not(stm32h7rs))]
+    #[cfg(not(all(stm32u5, peri_usb_otg_hs)))]
     if freq.0.abs_diff(48_000_000) > 120_000 {
         panic!(
             "USB clock should be 48Mhz but is {} Hz. Please double-check your RCC settings.",
             freq.0
         )
     }
 
+    // For OTG-HS on STM32U5 only the 32MHz clock is fast enough (Table 762, Sect 73.14.4)
```
MDr164, line 37: I recall that in the general datasheet of the U5 it mentioned the HSE clock explicitly needed as well but that might be an implementation detail not relevant here but rather in RCC. Just wanted to write it down so I don't forget about it during testing.

MDr164 (commented):
Looks good so far, I'll get around testing it tomorrow on hardware and let you know how it went.

embassy-stm32/src/usb/otg.rs
```diff
@@ -311,6 +314,26 @@ impl<'d, T: Instance> Bus<'d, T> {
...
 
+        #[cfg(all(stm32u5, peri_usb_otg_hs))]
+        {
+            crate::pac::RCC.ccipr2().modify(|w| {
+                w.set_otghssel(Otghssel::PLL1_P);
+            });
+
+            // Only the 32MHz clock is suitable here, which the magic number represents
+            crate::pac::SYSCFG.otghsphycr().modify(|w| {
+                w.set_clksel(11);
```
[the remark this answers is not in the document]
MDr164: The U5 has a new core clock for the USB subsystem with different clock requirements but I agree that could be done more dynamically.

embassy-stm32/src/usb/otg.rs
mubes: Removed. Not needed now.

embassy-stm32/src/usb/otg.rs
mubes: fixed.

embassy-stm32/src/usb/mod.rs
mubes: That's the reason I put the explicit reference. I thought this tied together with the otgphycr issue above, and the table only gives a value for 30MHz+ entry.  I may have wrapped this around my neck but the PLL1_P was also setting the AHB frequency?

embassy-stm32/src/usb/otg.rs
mubes: Dunno. Rust language limit hit. For the case the pins are analogue these values are not referenced, so added the _. If there's a better way, please advise.

embassy-stm32/src/usb/otg.rs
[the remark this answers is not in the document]
mubes: See comments below.

embassy-stm32/src/usb/otg.rs
mubes: Whoops. That was meant to have gone. Will remove it.

embassy-stm32/src/usb/mod.rs
Dirbaio: USB requires a crystal oscillator (so, HSE) because RC oscillators (like MSI, HSI) are not precise enough to meet the USB spec. This is the case for all chips, and both FS and HS.

(there are exceptions, many stm32 chips have some mechanism to calibrate HSI/MSI against either USB SOF or LSE which does make them precise enough)

Currently Embassy doesn't enforce this, on any chips. Not sure if we should, using USB from HSI does work in practice and if the user really wants to violate the spec we should let them (so it should have an opt-out at least). Also it's not trivial to implement. So IMO it's out of scope for this PR, and if we do it we should do it for all chips.

embassy-stm32/src/usb/otg.rs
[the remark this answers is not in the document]
Dirbaio: which comment?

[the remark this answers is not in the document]
Dirbaio: ah the "For OTG-HS on STM32U5 only the 32MHz clock is fast enough (Table 762, Sect 73.14.4)" thing. 

It's not quite right, the OTG_HS has two *different* clock inputs:
- The APB clock, which must be 30mhz or higher according to Table 762.
- The kernel clock. It's fed from a PLL which takes a clock from the OTGHSSEL mux. The PLL can take as input any of 16 MHz, 19.2 MHz, 20 MHz, 24 MHz, 26 MHz, 32 MHz. 

so all these frequencies are OK to use.

embassy-stm32/src/usb/otg.rs
MDr164: Almost, there is now a newline behind instead of before the function call :D
It was a nit as usually there is an empty line between variable population and and their usage but I don't think there is a styleguide for this for I would have not changed it here.

embassy-stm32/src/usb/otg.rs
MDr164: Oh I totally missed the feature flag there to keep those pins in analog mode which would result in them not be used and rust complaining. My bad.

embassy-stm32/src/usb/otg.rs
```diff
@@ -103,15 +103,18 @@ impl<'d, T: Instance> Driver<'d, T> {
...
     ) -> Self {
-        into_ref!(dp, dm);
-
-        dp.set_as_af(dp.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
-        dm.set_as_af(dm.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
+        // For STM32U5 High speed pins need to be left in analog mode
+        #[cfg(not(all(stm32u5, peri_usb_otg_hs)))]
+        {
+            into_ref!(_dp, _dm);
+            _dp.set_as_af(dp.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
```
MDr164, line 115: You'll also need to change the `dp` inside the method call to `_dp` for non-U5 targets to still compile. Same for the `dm` below.

embassy-stm32/src/usb/otg.rs
Dirbaio: please avoid nitpicking style stuff. we don´t have a styleguide beyond "use rustfmt" (which CI enforces).

MDr164: Commented again as `rustfmt` complains now but I'll refrain from any style nitpicks now. I was working in projects with some strict guidelines so consider my initial comment muscle memory.

embassy-stm32/src/usb/otg.rs
mubes: Fixed

examples/stm32u5/src/bin/usb_hs_serial.rs
```diff
@@ -0,0 +1,129 @@
...
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    info!("Hello World!");
+
+    let mut config = Config::default();
+    {
+        use embassy_stm32::rcc::*;
+	use embassy_stm32::time::Hertz;
```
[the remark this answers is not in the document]
mubes: That's what happens when you use emacs. Fixed.

embassy-stm32/src/usb/otg.rs
mubes: OK, ran the fmt. Should be resolved.

embassy-stm32/src/usb/otg.rs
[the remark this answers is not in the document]
mubes: I'm not seeing the reference to `Usbrefcksel` when I compare with the H7RS code, so I'm leaving this one open for more experienced hands.

embassy-stm32/src/usb/mod.rs
mubes: OK, I think it's time for me to throw this one over the wall now for others with more experience to look at. For me it works. Outstanding issue just really seems to be to bring the PLL mapping over from the H7RS code but I quickly got myself in a mess with that.

embassy-stm32/src/usb/otg.rs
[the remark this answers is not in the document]
MDr164: There is no enum yet in the stm32-data crate for it, I'll add one similarly to how it's done for the H7 series.

embassy-stm32/src/usb/mod.rs
MDr164: There are some enums and defines still missing in the pac for the U5 series that I can add later. Otherwise this looks pretty good imo.

MDr164 (commented):
I tested it on my hardware here and it works just fine. I'll update some of the code according to feedback regarding the clocks if you don't mind.

mubes:
Please go ahead and make whatever changes are appropriate!

embassy-stm32/src/usb/otg.rs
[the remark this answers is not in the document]
MDr164: Added the missing enum for the syscfg reg in this PR: https://github.com/embassy-rs/stm32-data/pull/545

MDr164:
I can't push to your branch as I got no write access for it but this patch is how I intend to fixup some of the clock stuff. It's a bit awkward to call SYSCFG inside the RCC code as the Reg to set the reference clock is part of SYSCFG and not RCC for the U5.

[0001-Update-STM32U5-OTG-HS-clock-handling.txt](https://github.com/user-attachments/files/18040199/0001-Update-STM32U5-OTG-HS-clock-handling.txt)

MDr164:
I updated the PR with an update to the clock selection, not 100% if that is the best way to do it on the U5 but still works on my hardware here. I think the CI failure is due to #3621 ?

Dirbaio:
lgtm! can you fix CI?

MDr164:
> lgtm! can you fix CI?

I was already looking at it but I'm confused why it fails. Am I misunderstanding the conditional compilation flags here? It should either match the first check for STM32H7RS and any STM32U5 with OTG_HS or any other platform so `freq` should be used in any case. I could just rename it to `_freq` but that shouldn't be needed as it wasn't needed before?

embassy-stm32/src/usb/mod.rs
```diff
@@ -16,6 +16,7 @@ fn common_init<T: Instance>() {
 
     // On the H7RS, the USBPHYC embeds a PLL accepting one of the input frequencies listed below and providing 48MHz to OTG_FS and 60MHz to OTG_HS internally
     #[cfg(stm32h7rs)]
+    #[cfg(all(stm32u5, peri_usb_otg_hs))]
```
Dirbaio, line 19: ah it's failing because multiple `cfg`s are ANDed so this code is actually never compiled (no chip is H7RS and U5 at the same time!).

this should work: `#[cfg(any(stm32h7rs, all(stm32u5, peri_usb_otg_hs)))]`

embassy-stm32/src/usb/mod.rs
```diff
@@ -26,6 +27,7 @@ fn common_init<T: Instance>() {
     // Clock might not be exact 48Mhz due to rounding errors in PLL calculation, or if the user
     // has tight clock restrictions due to something else (like audio).
     #[cfg(not(stm32h7rs))]
+    #[cfg(not(all(stm32u5, peri_usb_otg_hs)))]
```
Dirbaio, line 30: could you change this to `#[cfg(not(any(....)))]` ? for consistency with the other `cfg` above, and also because it's easier to read. having two`cfg`s can get confusing.

embassy-stm32/src/usb/mod.rs
MDr164: Oh I always thought multiple `cfg` statements are ORed. Let me quickly fix the `cfg` statement in this PR and then it should be good to go. @mubes please also test on your hardware once to see if everything still works to your liking as well as it's your PR after all.

embassy-stm32/src/usb/mod.rs
MDr164: Ack

MDr164:
> lgtm! can you fix CI?

Done, fixed all the `cfg` statements and CI is happy

mubes:
This is all working fine in my builds. From my side I think it can be committed. Thanks for the work folks!

MDr164:
This closes https://github.com/embassy-rs/stm32-data/issues/519 btw as all registers are there and confirmed working

PR #3616, opened 2024-12-05
Async timer interrupts

hatimbt:
This PR collects together and rebases the changes from #3029. (Credit for work goes to @brunob45).


hatimthayyil:
see #5431

PR #3622, opened 2024-12-07
Allow polling exited tasks

bugadani:
This PR replaces `poll_fn` to a more "inert" one, making it okay to poll `!SPAWNED` tasks. I guess this PR changes the trace output for awoken-exited tasks, but it's probably for the better.

ESP32-S3: cycle count goes from `15749/100` to `14745/100`
ESP32-S2: cycle count goes from `25815/100` to `24809/100`

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -215,8 +215,6 @@ impl<F: Future + 'static> TaskStorage<F> {
         let mut cx = Context::from_waker(&waker);
         match future.poll(&mut cx) {
             Poll::Ready(_) => {
-                waker.wake_by_ref();
-
                 // As the future has finished and this function will not be called
                 // again, we can safely drop the future here.
                 this.future.drop_in_place();
@@ -224,6 +222,10 @@ impl<F: Future + 'static> TaskStorage<F> {
                 // We replace the poll_fn with a despawn function, so that the task is cleaned up
                 // when the executor polls it next.
                 this.raw.poll_fn.set(Some(poll_to_despawn));
+
+                // Make sure we despawn last, so that other threads can only spawn the task
+                // after we're done with it.
+                this.raw.state.despawn();
             }
             Poll::Pending => {}
         }
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -161,6 +161,12 @@ pub struct TaskStorage<F: Future + 'static> {
     future: UninitCell<F>, // Valid if STATE_SPAWNED
 }
 
+unsafe fn poll_to_despawn(p: TaskRef) {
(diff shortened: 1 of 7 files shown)


embassy-executor/src/raw/mod.rs
```diff
@@ -202,16 +201,29 @@ impl<F: Future + 'static> TaskStorage<F> {
...
     unsafe fn poll(p: TaskRef) {
-        let this = &*(p.as_ptr() as *const TaskStorage<F>);
+        let this = &*p.as_ptr().cast::<TaskStorage<F>>();
 
         let future = Pin::new_unchecked(this.future.as_mut());
         let waker = waker::from_task(p);
         let mut cx = Context::from_waker(&waker);
         match future.poll(&mut cx) {
             Poll::Ready(_) => {
+                waker.wake_by_ref();
```
Dirbaio, line 218: this'll force all tasks to do one extra trip through the queue when they exit.

Is it possible to do without? For example, do *both* `state.despawn()` *and* change the `poll_fn` pointer to a noop.

bugadani: Yeah good point it shouldn't really make a difference.

embassy-executor/src/raw/mod.rs
```diff
@@ -203,15 +208,22 @@ impl<F: Future + 'static> TaskStorage<F> {
...
-        let this = &*(p.as_ptr() as *const TaskStorage<F>);
+        let this = &*p.as_ptr().cast::<TaskStorage<F>>();
 
         let future = Pin::new_unchecked(this.future.as_mut());
         let waker = waker::from_task(p);
         let mut cx = Context::from_waker(&waker);
         match future.poll(&mut cx) {
             Poll::Ready(_) => {
-                this.future.drop_in_place();
                 this.raw.state.despawn();
```
Dirbaio, line 218: the despawn call has to be last, because it's what "releases" the task storage.

bugadani: Yeah I noticed immediately, don't worry 😅

embassy-executor/src/raw/mod.rs
```diff
@@ -162,6 +161,12 @@ pub struct TaskStorage<F: Future + 'static> {
     future: UninitCell<F>, // Valid if STATE_SPAWNED
 }
 
+unsafe fn poll_to_despawn(p: TaskRef) {
+    // The task's future has already been dropped, we just mark it as `!SPAWNED`.
+    let this = p.header();
```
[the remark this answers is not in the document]
bugadani: Yeah this can, thanks. I was considering just storing(or swapping in) SPAWNED|RUN_QUEUED in run_enqueue. If that turns out to be a win, this will re-appear.

Dirbaio:
bender run

PR #3626, opened 2024-12-09
rp23: port usb_hid_keyboard example from rp

tommy-gilligan:
I think this is quite a useful example to have.  It's basically just copied verbatim from `rp`.


Dirbaio:
can you `cargo +nightly rustfmt` to get CI green?

PR #3630, opened 2024-12-10
Improve USB DFU support

Gerharddc:
This PR improves USB DFU support by documenting correct MSOS header usage in an example and by using the correct function headers in all cases.


lulf (approved):
LGTM, thanks!

PR #3631, opened 2024-12-10
nrf: Add RESET operations helpers for the nrf5340

Lakier15:
This adds free function helpers for the Reset peripheral to `embassy-nrf` so that it's easier to handle the checking of reset reasons and controlling the network core without having to use the `pac` directly.

It collects the reset reasons under a bitflag so that it can be more easily checked in binaries, instead of constructing a massive if else.

e.g. of reset reasons
```rust
    use embassy_nrf::reset::{self, ResetReason};
    
    for reason in reset::read_reasons().iter() {
        let reas = match reason {
            ResetReason::RESETPIN => "Reset Pin",
            ResetReason::DOG0 => "Application watchdog timer 0",
            ResetReason::CTRLAP => "Application CTRL-AP",
            ResetReason::SREQ => "Application soft reset",
            ResetReason::LOCKUP => "Application CPU lockup",
            ResetReason::OFF => "Wakeup from System OFF when wakeup is triggered by DETECT signal from GPIO",
            ResetReason::LPCOMP => "Wakeup from System OFF when wakeup is triggered by ANADETECT signal from LPCOMP",
            ResetReason::DIF => "Wakeup from System OFF when wakeup is triggered by entering the Debug Interface mode",
            ResetReason::NFC => "Wakeup from System OFF mode due to NFC field being detected",
            ResetReason::DOG1 => "Application watchdog timer 1",
            ResetReason::VBUS => "Wakeup from System OFF mode due to VBUS rising into valid range",
            _ => "unknown",
        };

        defmt::info!("Detected Reset reason: {}", reas);
    }
    reset::clear_reasons();
```

e.g. of starting the network core from the application core
```rust
    // Start network core
    defmt::info!("Releasing network core");
    reset::release_network_core();
```

### DIFF
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -126,6 +126,9 @@ pub mod qspi;
 #[cfg(not(any(feature = "_nrf91", feature = "_nrf5340-app")))]
 pub mod radio;
 #[cfg(not(feature = "_nrf54l"))] // TODO
+#[cfg(feature = "_nrf5340")]
+pub mod reset;
+#[cfg(not(feature = "_nrf54l"))] // TODO
 #[cfg(not(any(feature = "_nrf5340-app", feature = "_nrf91")))]
 pub mod rng;
 #[cfg(not(feature = "_nrf54l"))] // TODO
--- a/embassy-nrf/src/reset.rs
+++ b/embassy-nrf/src/reset.rs
@@ -1,0 +1,82 @@
+//! Reset
+
+#![macro_use]
+
+use bitflags::bitflags;
+use nrf_pac::reset::regs::Resetreas;
+#[cfg(not(feature = "_nrf5340-net"))]
+use nrf_pac::reset::vals::Forceoff;
+
+use crate::chip::pac::RESET;
+
+bitflags! {
+    #[derive(Debug, Copy, Clone, PartialEq)]
+    /// Bitflag representation of the `RESETREAS` register
+    pub struct ResetReason: u32 {
+        /// Reset Pin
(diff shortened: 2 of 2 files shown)


Dirbaio:
thanks for the PR! it's looking good, just one thing:

the RESET peripeheral is quite simple, it's not worth creating a peripheral singleton for it. The purpose of peripheral singletons is to ensure you don'tuse e.g. UARTE0 from two tasks at the same time, but all operations you do with RESET is single register reads or writes, which are already atomic, so it doesn't break if you use it from two tasks. it'd be simpler to make it free-standing functions.

Lakier15:
Makes sense. I just took inspiration from other peripherals.

Are you saying to turn this into free standing function in the same module and just expose the peripheral? In that case, is there already a similar driver in `embassy-nrf` that does that?

Or are you saying it's not worth to have this abstraction in the crate?

Dirbaio:
> Are you saying to turn this into free standing function in the same module and just expose the peripheral? 

yes! just make them free functions, there's no reason to create an owned peripheral singleton in `embassy_nrf::Peripherals`

> In that case, is there already a similar driver in embassy-nrf that does that?

no, not really. RESET is kind of the first peripheral that we add support that is "so simple". Note the equivalent in nrf52 is in POWER for example, which would also make sense to support as freestanding functions, but it's not implemented yet.

Lakier15:
Converted to free functions by getting rid of the unnecessary singleton. Also amended PR  title and description to match the change.

Dirbaio (approved):
looking great! :rocket: 

thank you

PR #3632, opened 2024-12-10
fix(embassy-net): make the `Config` constructors `const`

ROMemories:
This makes the [existing constructors of `embassy_net::Config`](https://docs.embassy.dev/embassy-net/git/default/struct.Config.html#impl-Config) `const`, so they can be used in a `const` context. I don't think this would hinder future evolution of the type or the constructors, and it would make these a little bit easier to use.


lulf (approved):
LGTM!

PR #3637, opened 2024-12-10
Fix ScanType enum export

karimpanacci:
Exports the ScanType enum, which is needed to build the ScanOptions struct


Dirbaio:
can you fix CI?

PR #3646, opened 2024-12-14
Add trait embedded_io_async to uarte

wackazong:


Dirbaio:
The raw Uart can't impl `embedded_io::Read` for a few reasons:

- It is is unbuffered, this means it can only receive while the task is actively waiting on a `.read().await` call. Data that's received between read calls is lost. The `embedded_io::Read` contract requires that data is not so easily lost, drivers using it can assume that.
- It must return early if less than `buf.len()` bytes are received. Drivers can call `read()` with a buffer bigger than the data they're expecting (for example, if they don't know the size of the data in advance) and get back a partially filled buffer with the data that arrived and process that. With this impl, they'd hang instead.

for `embedded_io::Read` there's the [buffered uart](https://docs.embassy.dev/embassy-nrf/git/nrf52833/buffered_uarte/index.html) which does extra fancy things to fulill the embedded-io contract.

Could you add just the Write impls?

wackazong:
Thanks for the explanation. I changed it as requested.

PR #3650, opened 2024-12-14
Add is_x() methods for all EitherN enum variants

avsaase:
This add `is_x()` methods on all `EitherN` enums to more concisely check which future from a `SelectN` completed first, without accessing the value. 

I often use signals for canceling futures "from within" and it's not always needed to access the value of the future, only to know if the signal `wait` future completed first. This can of course also be achieved using `matches!`, `if-let` or `let-else`, but the methods added in this PR make this a bit more concise.

### DIFF
--- a/embassy-futures/src/select.rs
+++ b/embassy-futures/src/select.rs
@@ -14,6 +14,18 @@ pub enum Either<A, B> {
     Second(B),
 }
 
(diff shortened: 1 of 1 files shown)


felipebalbi (approved):

lulf (approved):
LGTM, thanks!

PR #3655, opened 2024-12-16
Fix racy access of TaskHeader::executor

bugadani:


JoHaHu:
b47a631 breaks builds for avr as far as I can tell

PR #3656, opened 2024-12-16
Remove special handling of integrated timer queues and items

bugadani:
This PR places the mapping of a TQI to a queue the responsibility of the timer queue implementation. The existing drivers should be fine, because they use a single queue, multi-queue implementations need to handle some additional tasks.


Dirbaio (approved):
this is :100:

PR #3657, opened 2024-12-17
Remove WakerHack for good.

Dirbaio:
Now that 1.83 xtensa is out, we can remove it unconditionally.


bugadani:
Bye WakerHack, you won't be missed

PR #3663, opened 2024-12-18
Force use 1.83 for Xtensa

bugadani:
For compatibility reasons with the released esp-hal, we're reverting espup to install 1.82 by default. This PR ensures that the embassy CI survives this change.


Dirbaio (approved):
fun! thanks

pinning it is also nice so we don't auto-upgrade to newer releases automatically as well.

PR #3667, opened 2024-12-19
STM32U5: add HSPI support

williams-one:
This PR adds partial support for HSPI on the STM32U5 family.  It also includes an example showing how to use the on-board flash on the STM32U5G9J-DK2 demo board in both OCTA-DTR and memory mapped mode.
The implementation has been modeled after the OSPI interface.

The support is partial since the inner structures define all the 16 lines of the peripheral.  But the public interface only exposes constructors for Single SPI and Octa SPI (both blocking and non blocking), since these are the only two modes that I was actually able to test on my demo board.
In principle I could have added the implementation for all the other modes, but they would have been untested...  So I chose not to :)

I have added this as a reminder at the beginning of the file.
image

Please let me know if this is OK for you


williams-one:
I have updated DummyCycles configuration according to the driver taken from CubeMX

Dirbaio (approved):
thank you!

Dirbaio:
bender run

PR #3669, opened 2024-12-19
embassy-stm32: SPI slew rate configuration

spookyvision:
This PR makes the slew rate/speed for SCK and MOSI configurable and changes the default speed to `Medium`.


Dirbaio (approved):
thank you!

PR #3675, opened 2024-12-21
Expose `rp` bootloader state like `stm32`

korbin:
This PR exposes the `embassy-boot` `BootLoader` `State` in `embassy-boot-rp` [identically to](https://github.com/embassy-rs/embassy/blob/main/embassy-boot-stm32/src/lib.rs#L14-L15) `embassy-boot-stm32`.

This is required to `embassy-usb-dfu` Detach.


lulf (approved):
LGTM, thanks

PR #3678, opened 2024-12-22
STM32h5 UCPD example

elagil:
Also adds USB PD resistor software trimming for H5

### DIFF
--- a/examples/stm32h5/src/bin/usb_c_pd.rs
+++ b/examples/stm32h5/src/bin/usb_c_pd.rs
@@ -1,0 +1,90 @@
+#![no_std]
+#![no_main]
+
+use defmt::{error, info, Format};
+use embassy_executor::Spawner;
+use embassy_stm32::gpio::Output;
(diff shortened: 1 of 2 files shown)


examples/stm32h5/src/bin/usb_c_pd.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let config = Config::default();
+    let p = embassy_stm32::init(config);
+
+    info!("Hello World!");
+
+    // This pin controls the dead-battery mode on the attached TCPP01-M12.
```
9names, line 58: If your example expects a specific board with a specific IC in a particular configuration, it would be nice if you left a comment somewhere listing it so the reader can look up a schematic.

elagil: That's true. I was assuming all H5 examples target the corresponding NUCLEO platform, but it doesn't say that anywhere.

Issue #3278, opened 2024-08-23
BufferedUart + half duplex


BKSalman:
Hello,

I wanted to have uart with `embedded_io_async::Write` & `embedded_io_async::Read` implemented, but also be half duplex

so I'm wondering if that is possible, and just hasn't been implemented yet, or not

thank you for the great project :)

badrbouslikhin:
Hi @BKSalman,
It's already possible. Take a look at [UartRx in embassy_stm32::usart - Rust](https://docs.embassy.dev/embassy-stm32/git/stm32f446re/usart/struct.UartRx.html#method.into_ring_buffered).

Here's a snippet of how to achieve this:
```rust
    const DMA_BUF_SIZE: usize = 256;

    let mut usart = Uart::new_half_duplex(
        p.UART4,
        p.PA0,
        Irqs,
        p.DMA1_CH0,
        p.DMA1_CH1,
        config,
        HalfDuplexConfig::PushPull,
    )
    .unwrap();

    let (mut tx, rx) = usart.split();
    static mut DMA_BUF: [u8; DMA_BUF_SIZE] = [0; DMA_BUF_SIZE];
    let dma_buf = unsafe { DMA_BUF.as_mut() };
    let rx = rx.into_ring_buffered(dma_buf);
```
In this case, `tx` implements `embedded_io_async::Write` and `rx` implements `embedded_io_async::Read`.

BKSalman:
thank you for the help :), this might work for the time being

however I feel like this should be part of the `BufferedUart` API that it just handles the half duplex configuration the same way `Uart` does (using `BufferedUart::new_half_duplex()`), and the user doesn't have to split `tx` and `rx`

PR #3679, opened 2024-12-22
stm32/usart: configurable readback for half-duplex to support 1-wire + ds18b20 example

trnila:
The 1-Wire bus requires the ability to read data while transmission is in progress to detect if a sensor pulls the bus to a low level (logic 0).
This pull request adds a new enum `HalfDuplexReadback` with `Readback` variant that keeps receiver enabled during transmission. Half duplex can be newly created with `BufferedUart` (#3278). A simplified example, based on the RP example, has been added to demonstrate how to use this feature with the DS18B20 temperature sensor.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -12,8 +12,9 @@ use embassy_sync::waitqueue::AtomicWaker;
 #[cfg(not(any(usart_v1, usart_v2)))]
 use super::DePin;
 use super::{
-    clear_interrupt_flags, configure, rdr, reconfigure, send_break, set_baudrate, sr, tdr, Config, ConfigError, CtsPin,
-    Error, Info, Instance, Regs, RtsPin, RxPin, TxPin,
+    clear_interrupt_flags, configure, half_duplex_set_rx_tx_before_write, rdr, reconfigure, send_break, set_baudrate,
+    sr, tdr, Config, ConfigError, CtsPin, Duplex, Error, HalfDuplexConfig, HalfDuplexReadback, Info, Instance, Regs,
+    RtsPin, RxPin, TxPin,
 };
 use crate::gpio::{AfType, AnyPin, OutputType, Pull, SealedPin as _, Speed};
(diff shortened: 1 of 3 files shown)


Dirbaio (approved):
thank you!

wrs:
@trnila Thanks for half-duplex BufferedUart! With readback off, it appears to me that the RE flag is never turned back on after a transmission is complete, so reception doesn't work. If I manually set it after sending, then receiving works. Just wanted to check if I'm missing something before I file an issue and/or try to make a fix.

```rust
        self.tx.write_all(&buf).await?;
        self.tx.flush().await.unwrap();
        // Need to manually enable the receiver after tx is done
        embassy_stm32::pac::USART2.cr1().modify(|w| w.set_re(true));
```

trnila:
Looks like transmitter should be enabled forever (`TE`) after the first write in case of `BufferedUart` (is that even an issue? it could just reduce power consumption a bit?).

Receiver seems to be really disabled forever with readback off after the first transmission. 
I think we could reenable `RE` in [interrupt](https://github.com/embassy-rs/embassy/blob/897d42e01214396b448e5aae06a4186a4c282e7a/embassy-stm32/src/usart/buffered.rs#L90) once all bytes were successfully transmitted (so no bytes are lost between `write` and calling to `read`)?
But there can be issues with FIFO and readback-off: The receiver could be potentially turned on while transmission of last character is in progress and hence receiving that character back. Maybe [TXE and empty tx buffer](https://github.com/embassy-rs/embassy/blob/897d42e01214396b448e5aae06a4186a4c282e7a/embassy-stm32/src/usart/buffered.rs#L120) could be better for enabling receiver. Currently I am without hardware, so I cannot test myself.

wrs:
That makes sense. I am on an F1 so I can't see the effects of the FIFO. My working C driver sets RE in the TC interrupt handler if the tx buffer is empty. I'll give that a try.

I'm thinking it might make sense for the driver to manage an optional output pin for transceiver output enable (similar to CS pin handling in the SPI driver), for when half duplex mode is used with an RS-485 transceiver. It just needs to follow the TE/RE changes.

PR #3682, opened 2024-12-24
Usb composite iads by default

Eekle:
Sets the USB stack to use the "Composite with IADs" profile by default. This is the most "just works" configuration available.

Without this, using the default config and adding - for example - the CDC-ACM class will fail initialisation on Windows due to mismatched class identifiers.

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -34,17 +34,20 @@ pub struct Config<'a> {
     /// Device class code assigned by USB.org. Set to `0xff` for vendor-specific
     /// devices that do not conform to any class.
     ///
-    /// Default: `0x00` (class code specified by interfaces)
+    /// Default: `0xEF`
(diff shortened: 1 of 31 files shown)


embassy-usb/src/builder.rs
```diff
@@ -117,9 +120,9 @@ impl<'a> Config<'a> {
     /// Create default configuration with the provided vid and pid values.
     pub const fn new(vid: u16, pid: u16) -> Self {
         Self {
-            device_class: 0x00,
-            device_sub_class: 0x00,
-            device_protocol: 0x00,
+            device_class: 0xEF,
+            device_sub_class: 0x02,
+            device_protocol: 0x01,
```
Dirbaio, line 125: these need updating the default in the docs as well.

Eekle: Done

Dirbaio (approved):
thakns!

PR #3683, opened 2024-12-25
embassy-net: when std is set enable smoltcp/std

dpeckett:
As explained in the title, this is mostly to solve some clippy issues with ManagedSlice's I've been running into.


Dirbaio:
nothing in embassy-net requires smoltcp apis that are only available under `std`, so it's not possible that this fixes any issues.

The problem with smoltcp+managed you're encountering is typically due to something enabling `managed/std` or `managed/alloc` behind smoltcp's back so they get out of sync. Check which crate it is with `cargo tree --format '{p} {f}'` and remove it, and you'll find everything builds.

PR #3685, opened 2024-12-25
nrf54l: Allow debug access from firmware side

Tiwalun:
This unlocks debugger access to the application core, if it is also unlocked in UICR.

### DIFF
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -507,7 +507,6 @@ pub fn init(config: config::Config) -> Peripherals {
     let mut needs_reset = false;
 
     // Setup debug protection.
-    #[cfg(not(feature = "_nrf54l"))] // TODO
     #[cfg(not(feature = "_nrf51"))]
     match config.debug {
         config::Debug::Allowed => {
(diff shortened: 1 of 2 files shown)


embassy-nrf/src/lib.rs
```diff
@@ -549,18 +548,90 @@ pub fn init(config: config::Config) -> Peripherals {
...
+                debug_enable.set_key(crate::pac::tampc::vals::DomainDbgenCtrlKey::KEY);
+                debug_enable.set_value(DomainDbgenCtrlValue::HIGH);
+
+                p.protect().domain(0).dbgen().ctrl().write_value(writeprotect_clear);
+                p.protect().domain(0).dbgen().ctrl().write_value(debug_enable);
+
+                // Unlock niden
+                //
+                // Reuse values from dbgen disable
+                p.protect()
```
Dirbaio, line 583: this is quite verbose. I see you did it this way to reuse the values across dbgen, niden, spiden but it ends up being more verbose and nothing enforces the bits for these are at the same position. i'd suggest using the plain old closure `write` instead:

```rust
use crate::pac::tampc::vals;

p.protect().domain(0).niden().ctrl().write(|w| {
    w.set_key(vals::DomainDbgenCtrlKey::KEY);
    w.set_writeprotection(vals::DomainDbgenCtrlWriteprotection::CLEAR);
});
```

Dirbaio (approved):
thank!

Issue #3654, opened 2024-12-15
ADC Not Supported On STM32U535 / 545 Series


Spotigus:
The adc1 is currently not supported. Doing a quick comparison to options in cube mx:

- The channel sample times are different.
- Analog watch dogs are not implements.
- Max clock freq is 160 mhz vs 60
- Maybe different DMA differences. (Ran into some issues with the STM23 Hal version at work)

MDr164:
Looks like the closing keyword from #3688 was not picked up here. Should be resolved now with the aforementioned PR (at least on the STM32U595 and STM32U5A5 I got for testing here it works with the new code).

PR #3688, opened 2024-12-27
STM32U5: Add ADC drivers

klownfish:
This pull request adds ADC drivers for the U5 series with both blocking and async reads. One quirk of the U5 series is that some models have two simultaneous revisions of the peripheral. To handle this I've had to touch the internals a bit. The [issue](https://github.com/embassy-rs/stm32-data/pull/525) in stm32-data has some more details. 

Closes issue:
https://github.com/embassy-rs/embassy/issues/3654


Sandvoxel:
I just wanted to add that I’m also looking for ADC driver support for the STM32U5, as I plan to use the ADC in an upcoming project. Thanks for the work on this PR so far, and I’m looking forward to seeing it reviewed!

Dirbaio:
could you fix CI?

klownfish:
Should be fixed now! My first attempt failed so I did a force push to remove it from the history. I'm not entirely sure what's considered good practice with git but I hope it's okay.

Dirbaio (approved):
thanks!

PR #3691, opened 2024-12-28
cyw43-Pio: Add pio clock divider as a param for new

fatfingers23:
I had been having issues with getting the wifi to work on a Pico Plus 2 W and found that changing the pio clock divider to `3` got it working. I am not sure why that was the case or why it's different, but I also found in a handful of other scenarios that the [RM2 breakout](https://shop.pimoroni.com/products/rm2-breakout?variant=53492995719547) also needed the same clock divider. Below is a table of my findings

Notes:

- All connections to the breakout were done with jumper wires to the SP/CE connector on the RM2 breakout
- The new clock divider is `0x0300`
- I have a Pico 2 W on the way, but I have heard it works just fine without this change, so it looks to only affect the Pimoroni line of boards with the RM2
- The Pimoroni Presto is a new device from Pimoroni that uses the RM2 and RP2350B with what looks to be the same as the Pico Plus 2 W. I expect it will need the new clock divider as well

| Model | RM2 | PIO Clk divider | Notes |
| --- | --- | --- | --- |
| Pimoroni Pico Plus 2 W | Onboard | needs new clock divider |  |
| Pimoroni Pico Plus 2  | Breakout | needs new clock divider | Did not work with SP/CE connector to connector (pins 32-35) but did with pins 2-5. May be a RP2350B thing since those pins are not present in rp2040 or rp2350a |
| Pimoroni Presto | Onboard | TBD |  |
| Pico 1 rp2040 | Breakout | Default works, overclock feature does not |  |
| Pico 2 non w | Breakout | needs new clock divider |  |
| Pico 2 W | Onboard(Not rm2, but cyw43) | TBD but others have said no change needed |  |

This PR proposes to add a new parameter to the new method of the PioSpi, allowing users to pass in their own clock divider for the pio. With the RM2 breakout from Pimoroni, as well as the future release of it from Raspberry Pi, this may come up more. I also added some public conts of documented working clock dividers with the original comments and new ones to help users.

Some concerns:
* This is a breaking change when adding a new parameter. I debated if it should be in a new method because I'm not sure if the breaking change is worth it if this is just affecting a small set of users.
* I named the const `RM2_CLOCK_DIVIDER` RM2 because although it mostly looks to be just related to Pimoroni boards I found my Pico 2 with the RM2 breakout module also needed it. But the rp2040 was fine with the default clock divider.
* Left the overclock feature in place. Did not have success with it turned on and passing in the new clock divider
* `0x0300` May not be the best choice, but I have had success with it

I am open to any suggestions or rework needed!


Dirbaio:
thanks!

could you completely remove the `overclock` cargo feature, and use the "overclock" program if `divider < 0x200`? this way there's no possible mismatch.

fatfingers23:
The overclock feature has been removed. I did run into some problems with setting the results of `pio_asm!` to program in an if/else, so I restructured a bit to return the LoadedProgram instead.

Dirbaio (approved):
thank you!

brandonros:
If I try to use `DEFAULT_CLOCK_DIVIDER` on a Rasperry Pi Pico 2 W instead of `RM2_CLOCK_DIVIDER` I get 

```
0.311127 DEBUG init alp
0.311180 DEBUG set f2 watermark
0.311241 DEBUG watermark = 10
0.311256 DEBUG waiting for clock...
0.311294 DEBUG clock ok
0.311622 DEBUG clear request for ALP
0.312271 DEBUG chip ID: 43439
0.315426 DEBUG loading fw
0.468239 DEBUG loading nvram
0.468873 DEBUG starting up core...
0.471290 DEBUG wait for HT clock
0.489786 DEBUG setup interrupt mask
0.489833 DEBUG bluetooth setup interrupt mask
0.489903 DEBUG waiting for F2 to be ready...
0.575367 DEBUG clear pad pulls
0.575437 DEBUG waiting for HT clock...
0.575467 DEBUG clock ok
0.575553 DEBUG shared_addr 00017a7c
0.894459 ERROR panicked at 'assertion failed: success == true'
0.894481 ERROR panicked at /Users/brandon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/defmt-0.3.10/src/lib.rs:380:5:
explicit panic
```

Is this expected? 

The comments make it slightly confusing I think?

```
/// The default clock divider that works for Pico 1 and 2 W. As well as the RM2 on rp2040 devices.
/// same speed as pico-sdk, 62.5Mhz
/// This is actually the fastest we can go without overclocking.
/// According to data sheet, the theoretical maximum is 100Mhz Pio => 50Mhz SPI Freq.
/// However, the PIO uses a fractional divider, which works by introducing jitter when
/// the divider is not an integer. It does some clocks at 125mhz and others at 62.5mhz
/// so that it averages out to the desired frequency of 100mhz. The 125mhz clock cycles
/// violate the maximum from the data sheet.
pub const DEFAULT_CLOCK_DIVIDER: FixedU32<U8> = FixedU32::from_bits(0x0200);
```

> /// The default clock divider that works for Pico 1 and 2 W.

Except it doesn't work?

fatfingers23:
Hey. Sounds like you are running into an issue with using Bluetooth? May have to play around with it and find a clock divider that works better for that, and that may be the one for `RM2_CLOCK_DIVIDER`. The divider was added as a parameter since there are a few different boards with it now and to allow some wiggle room if needed. The comment I wrote there was in the context of Wifi only and could probably use some updating if that's the case.

Dirbaio:
I think the actual behavior is "default" divider works in rp2040, "rm2" divider works on rp235x.

brandonros:
I would say `rm2_divider` of `0x300` and `default` of `0x200` did not work for me on rp235x and I went with a custom one of `0x400` and my issues seemed to go away.

davidarkemp:
@brandonros Thanks for that tip! Using `0x0400` fixed the blinky_wifi on Pico 2 W too. 
I was getting an "inv len mismatch" using either DEFAULT or RM2 values

PR #3696, opened 2024-12-30
[avr] create initial example

tones111:
This commit adds a minimal example demonstrating how to create an embassy executor.  This gets the infrastructure in place for future work to add missing pieces like a timer driver.

### DIFF
--- a/examples/avr/.cargo/config.toml
+++ b/examples/avr/.cargo/config.toml
@@ -1,0 +1,8 @@
+[build]
+target = "avr-unknown-gnu-atmega328"
+
+[target.'cfg(target_arch = "avr")']
+runner = "ravedude uno -cb 57600"
+
+[unstable]
+build-std = ["core"]
--- a/examples/avr/Cargo.toml
+++ b/examples/avr/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.1.0"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-avr-device = { version = "0.5", features = ["atmega328p"] }
+avr-device = { version = "0.6", features = ["atmega328p"] }
 embassy-executor = { version = "0.7.0", path = "../../embassy-executor", features = ["arch-avr", "executor-thread", "nightly"] }
 panic-halt = "1"
 
--- a/examples/avr/Cargo.toml
+++ b/examples/avr/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+edition = "2021"
+name = "embassy-avr-examples"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+avr-device = { version = "0.5", features = ["atmega328p"] }
+embassy-executor = { version = "0.7.0", path = "../../embassy-executor", features = ["arch-avr", "executor-thread", "nightly"] }
+panic-halt = "1"
+
+[profile.dev]
+panic = "abort"
+lto = true
+opt-level = "s"
+
+[profile.release]
+panic = "abort"
+codegen-units = 1
+debug = true
+lto = true
+opt-level = "s"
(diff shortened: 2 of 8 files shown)


examples/avr/Cargo.toml
tones111, line 1: I just discovered the `.vscode/settings.json` file.  This example should be added there to simplify getting rust-analyzer working.

examples/avr/.cargo/config.toml
```diff
@@ -0,0 +1,8 @@
+[build]
+target = "avr-unknown-gnu-atmega328"
```
tones111, line 2: Should the example directory name more closely match the rustc target (avr -> atmega328)?

Currently nightly only supports this target, but avr-hal maintains some [additional spec files](https://github.com/Rahix/avr-hal/tree/main/avr-specs). I'm unsure if these will all end up being their own rustc platforms, if they can drop the gnu dependency, or if they can be consolidated under a more generic name.

 I'm assuming any AVR support here would only use rustc supported targets, otherwise this repo would likely need to duplicate the spec files.

examples/avr/src/bin/blinky.rs
```diff
@@ -0,0 +1,13 @@
...
+#![no_main]
+#![feature(impl_trait_in_assoc_type)]
+
+use embassy_executor::Spawner;
+use panic_halt as _;
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    loop {
+        // TODO
```
tones111, line 11: Toggling a pin using `avr-device` is easy enough, but the big hang up here is lack of an AVR time driver.

Is there interest in implementing/supporting an embassy-specific AVR hal, or should the examples make use of `avr-hal`? Does embassy attempt to have a consistent feel across its HALs?

One challenge is that none of the AVR HAL crates have been published, so git dependencies are needed. I'm also unsure how well it's design aligns with the HALs here.

examples/avr/.cargo/config.toml
tones111: I've learned of some progress regarding rustc avr target support.
[https://github.com/rust-lang/rust/pull/131651](https://github.com/rust-lang/rust/pull/131651)

The proposal consolidates the avr processes into one `avr-unknown-unknown` or `avr-unknown-none` target.

djdisodo: i put "avr-unknown-gnu-atmega328" just because it was there, i think most of the times, you'll be using target jsons from [avr-hal](https://github.com/Rahix/avr-hal)

djdisodo:
i wrote time driver https://github.com/djdisodo/avr_embassy_time/tree/master

it works well enough that i could drive stepper motors with it
and i believe there's some features that should be kept(this implementation has higher resolution for both timer and time)

but there's also lot of features with questionable benefits

for example wake scheduled in goofy linked list to keep it sorted

but since i later found that embassy never drops a timer, if you have a code that may cancel timer, the queue will be always full and it has to check all the timers when inserting new (also it uses a bit more ram)

(also MARGIN_TICKS is random number that i found it working)

examples/avr/.cargo/config.toml
brainstorm: I second @djdisodo, for now the device-specific JSONs are quite a bit more helpful than just arch triplets.

examples/avr/src/bin/blinky.rs
brainstorm: > Does embassy attempt to have a consistent feel across its HALs?

From experience helping out with https://github.com/embassy-rs/embassy/pull/1049, I suspect that the canonical way is to find a "closest hal example" and implement it in embassy "organically". Embassy traits are fairly well laid out, so it doesn't feel chaotic.

I feel you w.r.t crates.io publishing for AVR HALs... I'm unsure why @Rahix hasn't done that yet?

examples/avr/.cargo/config.toml
tones111: Yes, the upstream compiler now uses the `avr-none` target.  Once I get `avr-hal` updated I'll update this MR.

Dirbaio:
Hello,

PR #3697, opened 2024-12-30
fix embassy-time-driver example

tones111:
The ignored example had some compilation issues (when running `cargo test`).


Dirbaio (approved):
thanks!

PR #3698, opened 2024-12-30
Desugar some async fns

bugadani:
Not everything - I've been ignoring HALs and where the desugar would be unreasonably complicated (functions involving generics, for example).


bugadani:
I reran the test locally. It failed for a few times, then I was able to run `gpio` successfully, and after that the `i2c` test passed again. I'm tempted to say this was a transient issue.

PR #3702, opened 2025-01-01
std example cosmetics

BartMassey:
Did some minor cosmetic cleanups to the README and demo in `examples/std`.

### DIFF
--- a/examples/std/src/bin/net.rs
+++ b/examples/std/src/bin/net.rs
@@ -1,3 +1,5 @@
+use core::fmt::Write as _;
+
 use clap::Parser;
 use embassy_executor::{Executor, Spawner};
 use embassy_net::tcp::TcpSocket;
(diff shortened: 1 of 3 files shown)


examples/std/src/bin/net.rs
```diff
@@ -26,6 +28,41 @@ async fn net_task(mut runner: embassy_net::Runner<'static, TunTapDevice>) -> ! {
...
 }
 
+struct WriteBuf<const N: usize> {
+    len: usize,
+    bytes: [u8; N],
+}
+
+/// In a `no_std` situation it is useful to be able to
+/// format into a fixed-size buffer using the
+/// `core::fmt::write` macro. This struct enables that.
```
Dirbaio, line 38: this is a bit too complex for an example, i'd suggest using `heapless::String` instead. you can `write!` into it just like `WriteBuf`.

BartMassey: Doh. Hadn't noticed we were already depending on `heapless` and was trying to avoid it. Thanks for the catch. Fixed.

Dirbaio:
thanks!

PR #3703, opened 2025-01-01
add documentation for the BLE feature of the embassy-stm32-wpan crate

loftyinclination:
A large part of the current implementation has been reverse engineered from Application Note AN5289, and from consultation of the stm32_copro library.
Because of this, there may be a few inaccuracies in the doc comments that have been added.


Dirbaio (approved):
thank you!

it's great to have more docs :)

PR #3705, opened 2025-01-01
Ensure alarm is re-scheduled if timetamp is in the past

### LINKED ISSUE #3672: Executor alarm missed in nrf time driver
At Akiles we have have seen some cases where a custom critical section that unmasks certain interrupts (see https://github.com/alexmoon/nrf-sdc/blob/main/nrf-mpsl/src/critical_section_impl.rs#L8), causing the nrf time driver alarm to be missed. 

The following code makes the assumption that the time between now() and writing the cc register takes less than 3 ticks: https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/time_driver.rs#L212-L252

When using a custom critical section that preempts the time driver in set_alarm, combined with a task that uses a short delay (100 microseconds) wanting to schedule the alarm close to 3 or 4 ticks in the future, it may cause the time driver to miss setting the alarm, causing the executor to not be alarmed of the next task to run.

lulf:
Fixes #3672

Verified that this fixes the reproducer in the reported issue. I considered creating a HIL test for it, but it's a bit quirky since a custom critical section would need to be used, so I was wondering if it's ok to create another 'category' tests/nrf-mpsl' or something that uses the CS from nrf-mpsl instead to verify the fix. 

Wdyt?

### DIFF
--- a/embassy-nrf/src/time_driver.rs
+++ b/embassy-nrf/src/time_driver.rs
@@ -209,47 +209,53 @@ impl RtcDriver {
 
         let r = rtc();
 
-        let t = self.now();
-        if timestamp <= t {
-            // If alarm timestamp has passed the alarm will not fire.
-            // Disarm the alarm and return `false` to indicate that.
-            r.intenclr().write(|w| w.0 = compare_n(n));
+        loop {
+            let t = self.now();
+            if timestamp <= t {
+                // If alarm timestamp has passed the alarm will not fire.
+                // Disarm the alarm and return `false` to indicate that.
+                r.intenclr().write(|w| w.0 = compare_n(n));
 
-            alarm.timestamp.set(u64::MAX);
+                alarm.timestamp.set(u64::MAX);
 
-            return false;
+                return false;
+            }
+
+            // If it hasn't triggered yet, setup it in the compare channel.
+
+            // Write the CC value regardless of whether we're going to enable it now or not.
+            // This way, when we enable it later, the right value is already set.
+
+            // nrf52 docs say:
+            //    If the COUNTER is N, writing N or N+1 to a CC register may not trigger a COMPARE event.
+            // To workaround this, we never write a timestamp smaller than N+3.
+            // N+2 is not safe because rtc can tick from N to N+1 between calling now() and writing cc.
+            //
+            // It is impossible for rtc to tick more than once because
+            //  - this code takes less time than 1 tick
+            //  - it runs with interrupts disabled so nothing else can preempt it.
+            //
+            // This means that an alarm can be delayed for up to 2 ticks (from t+1 to t+3), but this is allowed
+            // by the Alarm trait contract. What's not allowed is triggering alarms *before* their scheduled time,
+            // and we don't do that here.
+            let safe_timestamp = timestamp.max(t + 3);
+            r.cc(n).write(|w| w.set_compare(safe_timestamp as u32 & 0xFFFFFF));
+
+            let diff = timestamp - t;
+            if diff < 0xc00000 {
+                r.intenset().write(|w| w.0 = compare_n(n));
+            } else {
+                // If it's too far in the future, don't setup the compare channel yet.
+                // It will be setup later by `next_period`.
+                r.intenclr().write(|w| w.0 = compare_n(n));
+            }
+
+            // If we have not passed the safe timestamp, we can be sure the alarm will be invoked. Otherwise,
+            // we need to retry setting the alarm.
+            if self.now() <= safe_timestamp {
+                return true;
+            }
         }
-
-        // If it hasn't triggered yet, setup it in the compare channel.
-
-        // Write the CC value regardless of whether we're going to enable it now or not.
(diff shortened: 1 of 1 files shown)


embassy-nrf/src/time_driver.rs
```diff
@@ -209,47 +209,53 @@ impl RtcDriver {
...
+            //
+            // It is impossible for rtc to tick more than once because
+            //  - this code takes less time than 1 tick
+            //  - it runs with interrupts disabled so nothing else can preempt it.
+            //
+            // This means that an alarm can be delayed for up to 2 ticks (from t+1 to t+3), but this is allowed
+            // by the Alarm trait contract. What's not allowed is triggering alarms *before* their scheduled time,
+            // and we don't do that here.
+            let safe_timestamp = timestamp.max(t + 3);
+            r.cc(n).write(|w| w.set_compare(safe_timestamp as u32 & 0xFFFFFF));
```
bugadani, line 242: Let's say an interrupt happens before this line, and when this code regains control `self.now()` equals to `safe_timestamp-1`. That means you'll set an alarm to a too-soon timestamp and you'll not know about it still.

lulf: @bugadani Thanks for the review! I'm not sure if I'm understanding the scenario you describe here fully. I've tried to annotate the code with the situation you're describing (removing the timer intnset/intnclr lines), but it seems to me that the check at the end will catch this situation. Maybe my assumptions around now()/counter incrementing during the irq is wrong...

First is the case I think you are describing where self.now() == safe_timestamp - 1 (i.e. fast irq)

```

            let t = self.now();
timestamp = 10
t = 0
ticks = 0
            if timestamp <= t {
                ...
                return false;
            }

ticks = 1
            let safe_timestamp = timestamp.max(t + 3);
safe_timestamp = 13

IRQ

ticks = 12
            r.cc(n).write(|w| w.set_compare(safe_timestamp as u32 & 0xFFFFFF));


ticks = 12 or 13:

12 <= 13 => alarm will be invoked so we return
            if self.now() <= safe_timestamp {
                return true;
            }

```


Second is the case where it will be retried:

```
            let t = self.now();
timestamp = 10
t = 0
ticks = 0
            if timestamp <= t {
                ...
                return false;
            }

ticks = 1
            let safe_timestamp = timestamp.max(t + 3);
safe_timestamp = 13

IRQ

ticks = 12
            r.cc(n).write(|w| w.set_compare(safe_timestamp as u32 & 0xFFFFFF));

ticks >= 14

14 <= 13 => loop is retried
            if self.now() <= safe_timestamp {
                return true;
            }
```

embassy-nrf/src/time_driver.rs
```diff
@@ -209,47 +209,53 @@ impl RtcDriver {
...
+            // This way, when we enable it later, the right value is already set.
+
+            // nrf52 docs say:
+            //    If the COUNTER is N, writing N or N+1 to a CC register may not trigger a COMPARE event.
+            // To workaround this, we never write a timestamp smaller than N+3.
+            // N+2 is not safe because rtc can tick from N to N+1 between calling now() and writing cc.
+            //
+            // It is impossible for rtc to tick more than once because
+            //  - this code takes less time than 1 tick
+            //  - it runs with interrupts disabled so nothing else can preempt it.
```
Dirbaio, line 236: it'd be great to update this comment. this particular bullet point is a lie (it's what caused this bug in the first place), but the surrounding context needs updating as well.

embassy-nrf/src/time_driver.rs
lulf: Ok, we found the case:

```
            let t = self.now();
timestamp = 10
t = 7
ticks = 7
            if timestamp <= t {
                ...
                return false;
            }

            let safe_timestamp = timestamp.max(t + 3);

safe_timestamp = 10

IRQ

ticks = 9
            r.cc(n).write(|w| w.set_compare(safe_timestamp as u32 & 0xFFFFFF));

9 <= 10 => 
            if self.now() <= safe_timestamp {
                return true;
            }

```

embassy-nrf/src/time_driver.rs
```diff
@@ -209,47 +209,53 @@ impl RtcDriver {
...
+            } else {
+                // If it's too far in the future, don't setup the compare channel yet.
+                // It will be setup later by `next_period`.
+                r.intenclr().write(|w| w.0 = compare_n(n));
+            }
 
-        true
+            // If we have not passed the timestamp, we can be sure the alarm will be invoked. Otherwise,
+            // we need to retry setting the alarm.
+            if self.now() + 3 <= timestamp {
```
Dirbaio, line 255: this could be moved into the `if diff < 0xc00000` branch, the `else` can unconditionally `return true`. saves a bit of work if the alarm is very far in the future.

(in theory you could have an interrupt longer than 0xc00000 ticks, but if you have interrupts that long the calculation in `now()` will break anyway, so it's fine if it also breaks here)

PR #3717, opened 2025-01-03
stm32: Implement reads of DTS peripheral

yodaldevoid:
Only PCLK-driven operation is supported. Similarly, only software triggering is supported.

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -73,7 +73,7 @@ rand_core = "0.6.3"
 sdio-host = "0.5.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "15" }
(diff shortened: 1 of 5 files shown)


yodaldevoid:
Looks like the DTS is not being marked as an RCC peripheral for at least one H7 chip. I'll do some digging to figure out if this is a metapac issue or a hardware difference.

yodaldevoid:
A quick look confirms this is an issue with the metapac. This may take some time to fix.

yodaldevoid:
Fix of metapac in https://github.com/embassy-rs/stm32-data/pull/559.

yodaldevoid:
PR updated with hardware triggers because it was easy enough to add.

Dirbaio (approved):
thank you! :)

PR #3720, opened 2025-01-04
Correctly gate `time` feature of embassy-embedded-hal in embassy-nrf

NZRosto:
Hi there,

This is another small fix to hopefully enable embassy-nrf to be built without a time driver.

This should allow embassy-nrf to not pull in embassy-time if a time driver or time feature is not needed.

As far as I could tell the time feature of embassy-embedded-hal is used only for spi delays, and so those would become panics when embassy-nrf is built without a time feature. This isn't ideal, but a better implementation would probably involve a shared bus that takes a delay implementation, which would be a PR for another day :)

Rust features are hard to reason about so let me know if anything needs changing.

### DIFF
--- a/embassy-nrf/Cargo.toml
+++ b/embassy-nrf/Cargo.toml
@@ -32,7 +32,7 @@ default = ["rt"]
 rt = ["nrf-pac/rt"]
 
 ## Enable features requiring `embassy-time`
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #3724, opened 2025-01-05
FIX: QSPI synchronous read operation hangs when FIFO is not full

CNLHC:
The current QSPI synchronous read function uses tcf!=1 and ftf!=1 as the polling condition. However, this condition fails when the transfer is complete and the FIFO is not full, as both tcf and ftf will be 0. This causes the read loop to block indefinitely, leaving data unread in the FIFO.

This PR resolves the issue by using the correct polling condition: tcf!=1 and flevel==0. This ensures the loop terminates only when the transfer is complete and the FIFO is empty.

Monitoring the flevel register during the read operation confirmed that it remained at 15 (FIFO_SIZE-1) when the issue occurred, validating the hypothesis.


Dirbaio (approved):
thanks!

Issue #3721, opened 2025-01-04
Use internal pull-ups with STM32 Timer/QEI (Quadrature Encoder) Input


leehambley:
More than happy to submit a patch if it's welcome, but as I'm new to the ecosystem I wanted to report an issue and check-in first.

I'm trying to use a simple EC11 rotary quadrature encoder for input and did already find the [`timer::Qei`](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/timer/qei.rs) implementation, however when using that no internal pullups/downs are set. I'm breadboarding right now, so that's fine, but it'd be nice to be able to use the internal pullups since they are there already.

My main task looks something like this:

```
#[embassy_executor::main]
async fn main(spawner: Spawner) {
    let p: embassy_stm32::Peripherals = embassy_stm32::init(Default::default());
    embassy_stm32::pac::GPIOA.bsrr().write(|w| w.set_bs(0, true)); // Set ODR bit high
    embassy_stm32::pac::GPIOA.cr(0).modify(|r| {
        r.set_mode(0, Mode::INPUT);
        r.set_cnf_in(0, CnfIn::PULL);
    });
```

Without spawning any tasks that works just fine, I can see the pin to to about 3V3 essentially immediately. **Note** due to really lack of docs, and lack of familiarity for myself with the ecosystem I am **less than certain** that this is the correct way to set bits on the PUPDR, especially using the ODR to pull the pin high feels incorrect when in my task I'm immediately setting that as an input pin.

The task I have is 

```
#[embassy_executor::task]
async fn encoder_input_task(
    tim: TIM2,
    chan_a: PA0,
    chan_b: PA1, // TODO: make these generic?
    _target_rpm: &'static Mutex<CriticalSectionRawMutex, RefCell<f32>>,
    _direction: &'static DirectionAtomic,
) {
    // Initialize TIM2 for encoder mode
    let qei_chan_a = QeiPin::new_ch1(chan_a);
    let qei_chan_b = QeiPin::new_ch2(chan_b);
    let encoder = Qei::new(tim, qei_chan_a, qei_chan_b);

    loop {
        Timer::after_millis(1000).await;
        let count = encoder.count();
        info!("Count: {}", count);
    }
}
```

I checked the implementation of `Qei::new_inner` and it doesn't look to interact with these registers at all, but I just think something must be going wrong.

[...]

PR #3728, opened 2025-01-05
[stm32] [timer] [qei] allow configuration of pull-up/down internal re…

leehambley:
…sistors on QeiPin

Adjusts the macros for impl<'d, T: GeneralInstance4Channel> QeiPin<'d, T, $channel> to take a 2nd argument which pullup/down should be active for the pins.

Related to https://github.com/embassy-rs/embassy/issues/3721#issuecomment-2571760530

This may not be ready to merge yet. I certainly would appreciate feedback from upstream. 

I verified this works with a Bluepill clone and a Nucleo F031K6.


embassy-stm32/src/timer/qei.rs
```diff
@@ -32,16 +32,29 @@ pub struct QeiPin<'d, T, Channel> {
...
+                #[doc = concat!("Create a new ", stringify!($channel), " QEI pin instance.")]
+                pub fn $new_chx(pin: impl Peripheral<P = impl $pin_trait<T>> + 'd) -> Self {
+                    Self::[<_ $new_chx>](pin, Pull::None)
+                }
+                #[doc = concat!("Create a new ", stringify!($channel), " QEI pin instance with internal pull-up.")]
+                pub fn [<$new_chx _pu>](pin: impl Peripheral<P = impl $pin_trait<T>> + 'd) -> Self {
+                    Self::[<_ $new_chx>](pin, Pull::Up)
+                }
+                #[doc = concat!("Create a new ", stringify!($channel), " QEI pin instance with internal pull-down.")]
+                pub fn [<$new_chx _pd>](pin: impl Peripheral<P = impl $pin_trait<T>> + 'd) -> Self {
```
Dirbaio, line 56: instead of making three functions, could you add a `Config` struct with the pull as a field? This way it's consistent with how other drivers do it (for example [usart](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L208)) and will hopefully let us avoid `paste` for a little longer.

leehambley: That'll break the API I think unless I can figure out how to do optional args. But I'll sure take a look. Thanks for the fast feedback.

Dirbaio: no need, doing breaking changes is fine if needed, and in this case it's what will result in a simpler API and more consistent with the rest of embassy-stm32.

(a nice thing about `#[non_exhaustive] struct Config` is that after we've added it we can add *other* config fields without it being breaking.)

leehambley:
Pull request has been rebased over Embassy main today. How do I get a merge please?

lulf:
Going through old PRs I found this. I can merge it if you rebase.

goodhoko:
Could/should this be closed now that #4722 has been merged? .)

PR #3730, opened 2025-01-06
`embassy-rp`: add `set_pullup()` for `OutputOpenDrain`

Georges760:
In case of OpenDrain Output, a Pull-Up resistor *MUST* be present on the signal. Usually it is a physical resistor on the board, but why not using the RP internal PU capability for it ? Could be handy for board with forgotten PU...

Note: Pull-Down is irrelevant here, so it is not proposed...


Dirbaio (approved):
thakns!

PR #3733, opened 2025-01-07
feat: STM32 Allow to modify the receiver filters from `BufferedCan`, `CanRx`, and `BufferedCanRx`

tdittr:
I want to split up my CAN interface on my STM32 to handle sending and receiving in different tasks. However I also want to be able to change the receive filters while the peripheral is split. Currently the `modify_filters` method is only available on the combined struct. This PR also makes it available on `BufferedCan`, `CanRx`, and `BufferedCanRx`.

From what I could tell from the reference manual and implementation, the filter registers are only used from the `MasterFilter` struct. So I think it should be fine to make them also accessible from the split off `Rx` half, since there is still only one struct that can access them.


Dirbaio (approved):
thanks!

PR #3738, opened 2025-01-07
stm32: generate singletons only for pins that actually exist.

Dirbaio:
Before we'd generate all pins Px0..Px15 for each GPIOx port. This changes
codegen to only generate singletons for actually-existing pins.

(AFs were already previously filtered, so these non-existing pins were already mostly useless)


Dirbaio:
bender run

PR #3742, opened 2025-01-08
[stm32f4] [adc] feat: Add async single adc readings

Christopher-06:
In contrast to the ADC implementations of other series, the STM32F4 series does not yet have an API definition for the asynchronous reading of a single channel. Instead, only the read_blocking() method was available.

In this PR, the API was extended to include an asynchronous read() method. Internal refactorings were also carried out in order to standardize the logic for setting registers and avoid redundant code. The implementation is based on the ADC implementations of other series in embassy_stm32/adc/.

The new method allows asynchronous single conversions to be performed and the ADC to be operated asynchronously in discontinuous mode. This is particularly useful for sporadic queries of an ADC channel, as the read() approach is simpler and more efficient than using a ring buffer DMA transfer.

In addition, the ADC example for the STM32F4 series has been updated. It now demonstrates the use of the read_blocking() method at the beginning and shows the asynchronous use with the new read() method in the infinite loop. This makes it easier to understand the interface and its practical use.

### DIFF
--- a/embassy-stm32/src/adc/v2.rs
+++ b/embassy-stm32/src/adc/v2.rs
@@ -1,3 +1,6 @@
+use core::future::poll_fn;
+use core::task::Poll;
+
 use embassy_hal_internal::into_ref;
 
 use super::blocking_delay_us;
@@ -156,8 +159,7 @@ where
         Vbat {}
     }
 
-    /// Perform a single conversion.
-    fn convert(&mut self) -> u16 {
+    fn start_conversion(&mut self) {
         // clear end of conversion flag
         T::regs().sr().modify(|reg| {
             reg.set_eoc(false);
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/adc/v2.rs
```diff
@@ -167,6 +169,28 @@ where
...
             reg.set_swstart(true);
         });
+    }
+
+    /// Perform a single conversion asynchronously.
+    async fn convert(&mut self) -> u16 {
+        self.start_conversion();
+
+        // wait for actual start and then finish
+        poll_fn(|_| {
```
Dirbaio, line 179: this should register the waker somewhere and wake it from the interrupt. If not, the task will sleep forever if Pending is returned.

Christopher-06:
My mistake - I have now added the waker and activated/deactivated the interrupt accordingly. In my tests with a fast clock and the STM32F446, the single conversion was probably so fast that the ADC reading was already available at this point. However, this should now work universally. Thanks and have a nice day @Dirbaio!

tommasoclini:
Hi, I'm using an stm32f4, I was looking for a way to read a sequence of channels without using continuous mode, and I thought I could implement an `async fn read` like [here](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/adc/v4.rs#L363).
Then I stumbled upon this PR, how should we name the two functions if we want to keep both?

Christopher-06:
My intuition is that the API should follow the core nature of Embassy as an asynchronous framework. Accordingly, the standard functions should also be asynchronous and use simple, intuitive names. I'm not the maintainer or author, so it's not my decision to make, but if I had to choose a direction, my naming strategy would look like this:

- `read(...)`: Asynchronous reading of a single ADC channel
- `read_multi(...)`: Asynchronous reading of one or more ADC channels using DMA
- `read_blocking(...)`: Synchronous blocking read of a single ADC channel from start to result
- `read_multi_blocking(...)`: [if necessary or meaningful at all] Synchronous blocking read of one or more ADC channels using DMA from start to result

Having all functions start with read also improves discoverability with IntelliSense, as all available read operations will appear when typing read.

tommasoclini:
I have opened the [PR](https://github.com/embassy-rs/embassy/pull/4484), you can take a look at it and critique me as much as you want, I'm new to writing low level hal code.

Dirbaio:
Hello,

PR #3744, opened 2025-01-08
stm32: flash: fix flash erase on stm32f3xx series

mrodgers-witekio:
Fix for flash erase not working on STM32F3 series (tested on STM32F303K8 specifically).

STM32F3xx series also needs a wait of at least one clock cycle before reading the BSY bit during a flash erase - previously this was only applied to STM32F1xx series.

This was causing an erase failure due to the `wait_ready_blocking` not actually waiting until the erase operation is complete, and as a result the EOP bit was not set when checked.


Dirbaio (approved):
thanks!

PR #3748, opened 2025-01-09
Fix unsupported trace! call for EndpointAddress

Volkalex28:
Hi. For me embassy-stm32 does not compile for stm32f103 with log feature
It seems EndpointAddress does not implement LowerHex trait. I suggest simply replacing the call with debug.
Or I can create a PR with LowerHex and UpperHex implementations for EndpointAddress


Dirbaio (approved):
thanks!

PR #3779, opened 2025-01-16
Full RCC support for STM32F107

algesten:
STM32F107 has a two more PLLs than F013, which means using `f013.rs` config options doesn't let us configure the entire RCC.

My use case is to use STM32F107 for ethernet with an external HSE. 

This PR adds a special case for F107 with a fuller RCC config.

### DIFF
--- a/embassy-stm32/src/rcc/f013.rs
+++ b/embassy-stm32/src/rcc/f013.rs
@@ -41,6 +41,8 @@ pub enum PllSource {
     HSI,
     #[cfg(rcc_f0v4)]
     HSI48,
+    #[cfg(stm32f107)]
+    PLL2,
 }
 
 #[derive(Clone, Copy)]
@@ -100,9 +102,7 @@ pub struct Config {
     #[cfg(stm32f107)]
     pub pll3: Option<Pll2Or3>,
     #[cfg(stm32f107)]
-    pub prediv1_src: Option<PreDiv1Src>,
-    #[cfg(stm32f107)]
-    pub prediv2: Option<PllPreDiv>,
+    pub prediv2: PllPreDiv,
 
     pub ahb_pre: AHBPrescaler,
     pub apb1_pre: APBPrescaler,
@@ -118,9 +118,9 @@ pub struct Config {
     pub adc34: AdcClockSource,
 
     #[cfg(stm32f107)]
-    pub i2s2_src: Option<I2s2src>,
+    pub i2s2_src: I2s2src,
     #[cfg(stm32f107)]
-    pub i2s3_src: Option<I2s2src>,
+    pub i2s3_src: I2s2src,
 
     /// Per-peripheral kernel clock selection muxes
     pub mux: super::mux::ClockMux,
@@ -143,9 +143,7 @@ impl Default for Config {
             #[cfg(stm32f107)]
             pll3: None,
             #[cfg(stm32f107)]
-            prediv1_src: None,
-            #[cfg(stm32f107)]
-            prediv2: None,
+            prediv2: PllPreDiv::DIV1,
 
             ahb_pre: AHBPrescaler::DIV1,
             apb1_pre: APBPrescaler::DIV1,
@@ -163,9 +161,9 @@ impl Default for Config {
             adc34: AdcClockSource::Hclk(AdcHclkPrescaler::Div1),
 
             #[cfg(stm32f107)]
-            i2s2_src: None,
+            i2s2_src: I2s2src::SYS,
             #[cfg(stm32f107)]
-            i2s3_src: None,
+            i2s3_src: I2s2src::SYS,
 
             mux: Default::default(),
         }
@@ -207,13 +205,6 @@ pub(crate) unsafe fn init(config: Config) {
         }
     };
 
-    #[cfg(stm32f107)]
-    let pll2freq = config.pll2.map(|pll2| {
-        let prediv2 = config.prediv2.unwrap_or(PllPreDiv::DIV1);
-        let in_freq = hse.unwrap() / (prediv2.to_bits() + 1);
-        in_freq * pll2.mul
-    });
-
     // configure HSI48
     #[cfg(crs)]
     let hsi48 = config.hsi48.map(|config| super::init_hsi48(config));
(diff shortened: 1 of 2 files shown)


Dirbaio:
instead of adding a new `f107.rs` file, could you add support for the PLLs in `f013.rs` using cfg's like `cfg(stm32f107)`? 

(We used to have code duplication like this in RCC a while ago and it proved to be an unmaintainable mess. Hard to keep them in sync with fixes and features. the file you've added is already out of sync, for example it does `.unwrap()` instead of `unwrap!(..)` and is missing the asserts for min/max frequencies)

algesten:
@Dirbaio sure. Is there some thinking of how these are bundled together? Like, why is f107 sorted under f013? Like is there some arch concept that certain groups of STM32 belongs together?

Dirbaio:
Yes, they're "more similar". ST had some design for the F1, then F107, F3, F0 are incremental improvements on top of that but the "overall structure" stays very similar. OTOH they designed the PLLs very differently for F2, then F4, F7 are incremental improvements on top of that. Etc etc.

(`f013` means "STM32F0 + STM32F1 + STM32F3", not "STM32F013")

algesten:
@Dirbaio thanks for explaining.

I have pushed an update where I try to merge the changes into f013.rs. Hopefully it's closer to what you want.

embassy-stm32/src/rcc/f013.rs
```diff
@@ -187,21 +234,39 @@ pub(crate) unsafe fn init(config: Config) {
                 }
                 (Pllsrc::HSI_DIV2, unwrap!(hsi))
             }
+            #[cfg(not(stm32f107))]
             PllSource::HSE => (Pllsrc::HSE_DIV_PREDIV, unwrap!(hse)),
+            #[cfg(stm32f107)]
+            PllSource::HSE => (
```
Dirbaio, line 240: this is a bit of a lie, if you set `PllSource::HSE` the source can end up being either HSE or PLL2. I'd suggest adding a new variant `PllSource::PLL2` instead, and removing the `prediv1_src` field from config.

embassy-stm32/src/rcc/f013.rs
```diff
@@ -187,21 +234,39 @@ pub(crate) unsafe fn init(config: Config) {
...
+                },
+            ),
             #[cfg(rcc_f0v4)]
             PllSource::HSI48 => (Pllsrc::HSI48_DIV_PREDIV, unwrap!(hsi48)),
         };
+        #[cfg(not(stm32f107))]
         let in_freq = src_freq / pll.prediv;
+
+        #[cfg(stm32f107)]
+        let in_freq = src_freq / (pll.prediv.to_bits() + 1);
```
Dirbaio, line 254: this shouldn't be needed, `/ pll.prediv` does the right thing automatically (it's a Div impl generated in build.rs, it's *not* dividing by the bit value)

Dirbaio: okay, for this to work add `Prediv1`, `Prediv2` here. https://github.com/embassy-rs/embassy/blob/010d4622f988de813ecddf69b84273f72ab122cf/embassy-stm32/build.rs#L1386 . Then you can directly divide instead of having to do `.to_bits()`.

embassy-stm32/src/rcc/f013.rs
```diff
@@ -169,6 +207,13 @@ pub(crate) unsafe fn init(config: Config) {
         }
     };
 
+    #[cfg(stm32f107)]
+    let pll2freq = config.pll2.map(|pll2| {
+        let prediv2 = config.prediv2.unwrap_or(PllPreDiv::DIV1);
+        let in_freq = hse.unwrap() / (prediv2.to_bits() + 1);
```
Dirbaio, line 213: same as prediv1, divide directly instead of doing .to_bits(). (see below)

embassy-stm32/src/rcc/f013.rs
```diff
@@ -85,6 +95,14 @@ pub struct Config {
...
 
     pub pll: Option<Pll>,
+    #[cfg(stm32f107)]
+    pub pll2: Option<Pll2Or3>,
+    #[cfg(stm32f107)]
+    pub pll3: Option<Pll2Or3>,
+    #[cfg(stm32f107)]
+    pub prediv1_src: Option<PreDiv1Src>,
+    #[cfg(stm32f107)]
+    pub prediv2: Option<PllPreDiv>,
```
Dirbaio, line 105: this sholdn't be Option, because you can't "not enable prediv2".

embassy-stm32/src/rcc/f013.rs
```diff
@@ -99,6 +117,11 @@ pub struct Config {
     #[cfg(all(stm32f3, not(rcc_f37), any(peri_adc3_common, peri_adc34_common)))]
     pub adc34: AdcClockSource,
 
+    #[cfg(stm32f107)]
+    pub i2s2_src: Option<I2s2src>,
+    #[cfg(stm32f107)]
+    pub i2s3_src: Option<I2s2src>,
```
Dirbaio, line 123: same for these.

Dirbaio (approved):
very nice! thank you :rocket:

algesten:
@Dirbaio thanks! Just one thing more to fix. There's an exported `PreDiv1Src` I don't believe I need anymore.

algesten:
Just going to check with my board as well, just to make sure I didn't break anything.

algesten:
Yeah. It's not working. Need to figure out why.

algesten:
@Dirbaio found it. The problem was that `set_prediv1src` must be set _after_ configuring the PLL.

I think we're good to go!

Dirbaio:
oh huh. shouldn't it be the other way around? if pll1 is fed from pll2, init order should be prediv2, pll2, prediv1, pll1. 

With the previous code it'd fail because you were starting pll1 without starting pll2 first, which would hang because it'd have no clock. Now you're leaving pll1 source as default (so, from HSE), then initialize it (which works because it has a clock), then switch it to pll2 while it's already running. I don't think it's allowed to change a pll source clock while it's running.

algesten:
> I don't think it's allowed to change a pll source clock while it's running.

Right. It works, but maybe is dodgy :)

I'll revise the code to ensure the order is correct for: prediv2, pll2, prediv1, pll1

PR #3781, opened 2025-01-17
stm32/rcc: add HSISYS support for g0

markus-k:
adds support to divide HSI clock on g0, which allows running in low power mode on HSI. Basically a port of the same code from the c0 implementation.

Tested on STM32G071KB


Dirbaio (approved):
thanks!

PR #3782, opened 2025-01-17
stm32/can: fix wrong negation of fdcan set_transmit_pause function parameter

Andreychik32:
I'm using dual-core STM32H747BI.

The issue is that this method simply wrongly negates its parameter. 

In RM0399, it's stated that TXP bit _enables_ pause. The function negates boolean argument, so `set_transmit_pause(false)` will inevitably enable pause.

RM0399, page 2679:
CleanShot 2025-01-17 at 6  08 10@2x


Dirbaio (approved):
thanks!

PR #3785, opened 2025-01-20
Fix issue with enabled split-pxy feature

mahu-wm:
With feature embassy-stm32/split-pa0, I got the error ``error[E0425]: cannot find value `pin_name` in this scope``.


Dirbaio:
thanks!

opened #3791 since I wanted to add it to `ci.sh` and you didn't check the "allowed maintainers to push" checkbox when creating the PR.

PR #3792, opened 2025-01-22
stm32: Change unreachable panic to explicit error message when access…

### LINKED ISSUE #2978: SealedRccPeripheral::frequency panic messages could be more helpful
A static message like "clock not enabled for peripheral USB_OTG_HS" is much more helpful than "entered unreachable code".

Since this code is generated, perhaps it wouldn't be too difficult to have an even more helpful panic message like "clock not enabled for peripheral USB_OTG_HS, set `embassy_stm32::Config.rcc.mux.usbsel`".

cc #2973

adamgreig:
…ing clock-muxed peripherals set to invalid clocks

fixes #2978


Dirbaio (approved):
thank you!

PR #3795, opened 2025-01-22
STM32: Option to detect Ethernet PHY address automatically

nikvoid:
Currently `GenericSMI` can't detect ethernet PHY address and will stuck in eternal loop if address is incorrect. This PR adds automatic detection by probing all addresses from 0 to 31 with fixed attempt count and delay. Automatic detection enabled if `GenericSMI` was created with address `0xFF`

### DIFF
--- a/embassy-stm32/src/eth/generic_smi.rs
+++ b/embassy-stm32/src/eth/generic_smi.rs
@@ -3,7 +3,7 @@
 use core::task::Context;
 
 #[cfg(feature = "time")]
-use embassy_time::{Delay, Duration, Timer};
+use embassy_time::{Duration, Timer};
 #[cfg(feature = "time")]
 use futures_util::FutureExt;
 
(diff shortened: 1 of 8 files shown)


embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -62,6 +64,24 @@ impl GenericSMI {
...
+            for addr in 0..32 {
+                sm.smi_write(addr, PHY_REG_BCR, PHY_REG_BCR_RESET);
+                for _ in 0..10 {
+                    if sm.smi_read(addr, PHY_REG_BCR) & PHY_REG_BCR_RESET != PHY_REG_BCR_RESET {
+                        trace!("Found ETH PHY on address {}", addr);
+                        self.phy_addr = addr;
+                        return;
+                    }
+                    embedded_hal_1::delay::DelayNs::delay_us(&mut Delay, 1000);
+                    cortex_m::asm::delay(1000000);
```
8cichy8, line 79: Probably unintended double delay?

nikvoid: Yep

embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -62,6 +64,23 @@ impl GenericSMI {
 
 unsafe impl PHY for GenericSMI {
     fn phy_reset<S: StationManagement>(&mut self, sm: &mut S) {
+        // Detect SMI address
+        #[cfg(feature = "time")]
```
Dirbaio, line 68: it's surprising for the user that this is only available with `time`. Other drivers that need delays fall back to `cortex_m::asm::delay`. For example:

https://github.com/embassy-rs/embassy/blob/9a159a8db0638ea2971eed130a2ae34743da640a/embassy-stm32/src/adc/mod.rs#L83-L93

Maybe we should move that function somewhere shared, there's already a few copypastes of it around...

embassy-stm32/src/eth/generic_smi.rs
```diff
@@ -51,6 +51,8 @@ pub struct GenericSMI {
 
 impl GenericSMI {
     /// Construct the PHY. It assumes the address `phy_addr` in the SMI communication
+    ///
+    /// Set `phy_addr` to `0xFF` for automatic detection (only with `time` feature enabled)
```
Dirbaio, line 55: it's not very "rusty" that there's a special value that has different behavior. I'd suggest either an enum `enum PhyAddr{ Auto, Manual(u8) }` or two different constructors `new` and `new_auto`

nikvoid: I would keep this as special value in internal code, add address bound assert `(addr < 32)` in `new` method and add `new_auto`. This value must be replaced to actual addr. in `phy_reset`, or code will panic.

embassy-stm32/src/eth/generic_smi.rs
nikvoid: This would be better for blocking code instead of dep. on `time`. For now I'll copypaste it, too)

Dirbaio (approved):
thank you!

PR #3797, opened 2025-01-22
Add stream impl for embassy-sync Channels

stargazing-dino:
I recently had a thought to gather all of my channels and do some funky Stream manipulations on them (debouncing mainly).

Would this be a useful addition?


lulf (approved):
LGTM!

PR #3804, opened 2025-01-25
Return Truncate error instead of looping forever on udp send.

Easyoakland:
# Issue
If the udpsocket send methods are given a buffer larger than the udp payload buffer the methods loop await forever trying to send and getting `BufferFull`.

# Fix
Check and early-return if udp::send methods receive a buffer too large to ever be sent.

# Details
smoltcp used to distinguish these but stopped [here](https://github.com/smoltcp-rs/smoltcp/commit/b7871e4e5704963e84b0a5eea04c4601162b4703#diff-528cd95c12007e0dd82ca1181897ed74a8e2dedfd8ca2d151d4f133f785023b2L82). I'm not sure why. You can still the the remnant in the now invalid comment for `smoltcp::socket::udp::Socket::send` mentioning `Truncate`. I don't know if it would be preferred to update smoltcp or leave the fix here in `embassy-net`.

### DIFF
--- a/embassy-net/src/udp.rs
+++ b/embassy-net/src/udp.rs
@@ -21,7 +21,7 @@ pub enum BindError {
     NoRoute,
 }
 
-/// Error returned by [`UdpSocket::recv_from`] and [`UdpSocket::send_to`].
+/// Error returned by [`UdpSocket::send_to`].
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum SendError {
@@ -29,9 +29,11 @@ pub enum SendError {
     NoRoute,
     /// Socket not bound to an outgoing port.
     SocketNotBound,
+    /// There is not enough transmit buffer capacity to ever send this packet.
+    Truncated,
 }
 
-/// Error returned by [`UdpSocket::recv_from`] and [`UdpSocket::send_to`].
+/// Error returned by [`UdpSocket::recv_from`].
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum RecvError {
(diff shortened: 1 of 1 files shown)


embassy-net/src/udp.rs
```diff
@@ -240,11 +244,21 @@ impl<'a> UdpSocket<'a> {
...
+    ///
     /// When the remote endpoint is not reachable, this method will return `Poll::Ready(Err(Error::NoRoute))`.
     pub fn poll_send_to<T>(&self, buf: &[u8], remote_endpoint: T, cx: &mut Context<'_>) -> Poll<Result<(), SendError>>
     where
         T: Into<UdpMetadata>,
     {
+        // Don't need to wake waker in `with_mut` if the buffer will never fit the udp tx_buffer.
+        let send_capacity_too_small = self
+            .stack
+            .with(|i| i.sockets.get::<udp::Socket>(self.handle).payload_send_capacity() < buf.len());
```
Dirbaio, line 257: could you use `self.with()` instead of `self.stack_with()`?

embassy-net/src/udp.rs
```diff
@@ -21,17 +21,19 @@ pub enum BindError {
...
+/// Error returned by [`UdpSocket::send_to`].
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum SendError {
     /// No route to host.
     NoRoute,
     /// Socket not bound to an outgoing port.
     SocketNotBound,
+    /// There is not enough transmit buffer capacity to ever send this packet.
+    Truncated,
```
Dirbaio, line 33: The error name is a bit misleading, it implies the packet was sent anyway but truncated. Could you name it `PacketTooLarge`?

Easyoakland: Fixed: [c0d14a1](https://github.com/embassy-rs/embassy/pull/3804/commits/c0d14a145c901126a548a77026ef2ef0d6e79738)

embassy-net/src/udp.rs
Easyoakland: Fixed: [c0d14a1](https://github.com/embassy-rs/embassy/pull/3804/commits/c0d14a145c901126a548a77026ef2ef0d6e79738)

Easyoakland:
Oh, I forgot to [update](https://github.com/embassy-rs/embassy/compare/main...Easyoakland:embassy:fix-udp-send) the comments to match the name change. Should I make a new Pr for that?

Dirbaio:
oh, yes please! btw could you also remove the `Poll::Ready` mention? it's already implied that when an async function returns it has returned `Poll::Ready`.

PR #3806, opened 2025-01-25
Fix #3805

### LINKED ISSUE #3805: ADC interrupt for STM32F103 doesn't work property
Current ADC implement only works on the example. Adding totally unrelated parts to the program breaks the ADC function.

- The default sample time `SampleTime::CYCLES1_5` works. Set to other cause `Adc::read().await` hang-up forever.
- Then add line `defmt::info!("convert started");` to `embassy-stm32/src/adc/f1.rs` file `convert()` function above line `T::state().waker.register(....` makes it work...
- My project uses USB CDC. When start the USB task, ADC doesn't work, when remove USB stuffs the ADC works again.

, which is quite confusing. And it takes my hours to debug.

It turns out that the ADC End of Convert interrupt doesn't triggered at all. So the `poll_fn` wouldn't wake up once it go pending. It only works if the convert was completed before first time the `poll_fn` is polled.

https://github.com/embassy-rs/embassy/blob/51d87c6603631fda6fb59ca1a65a99c08138b081/embassy-stm32/src/adc/f1.rs#L124-L132

Tnze:
We must register the waker before enabling the EOC interrupt and starting convert

### DIFF
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -118,11 +118,11 @@ impl<'d, T: Instance> Adc<'d, T> {
     async fn convert(&mut self) -> u16 {
         let mut started = false;
         poll_fn(|cx| {
+            T::state().waker.register(cx.waker());
             if !T::regs().cr2().read().swstart() && T::regs().sr().read().eoc() {
                 Poll::Ready(T::regs().dr().read().0 as u16)
             } else {
                 if !started {
-                    T::state().waker.register(cx.waker());
                     T::regs().cr1().modify(|w| w.set_eocie(true)); // End of Convert interrupt enable
                     T::regs().cr2().modify(|reg| reg.set_swstart(true));
                     started = false;
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -125,7 +125,7 @@ impl<'d, T: Instance> Adc<'d, T> {
                 if !started {
                     T::regs().cr1().modify(|w| w.set_eocie(true)); // End of Convert interrupt enable
                     T::regs().cr2().modify(|reg| reg.set_swstart(true));
-                    started = false;
+                    started = true;
                 }
                 Poll::Pending
             }
--- a/embassy-stm32/src/adc/f1.rs
+++ b/embassy-stm32/src/adc/f1.rs
@@ -23,7 +23,7 @@ pub struct InterruptHandler<T: Instance> {
 impl<T: Instance> Handler<T::Interrupt> for InterruptHandler<T> {
     unsafe fn on_interrupt() {
         if T::regs().sr().read().eoc() {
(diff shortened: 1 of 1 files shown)


Tnze:
TODO:
- Apply this to other chips
- Call `unsafe { NVIC::unmask(pac::interrupt::ADC1_2) };` somehow. I don't know

embassy-stm32/src/adc/f1.rs
```diff
@@ -115,24 +116,21 @@ impl<'d, T: Instance> Adc<'d, T> {
 
     /// Perform a single conversion.
     async fn convert(&mut self) -> u16 {
-        T::regs().cr2().modify(|reg| {
-            reg.set_adon(true);
-            reg.set_swstart(true);
-        });
-        T::regs().cr1().modify(|w| w.set_eocie(true));
-
+        let mut started = false;
```
Dirbaio, line 119: I think this shouldn't be needed, it should be OK to start the conversion before registering the waker. Even if the conversion finishes "too soon" the task will see EOC is set and return Ready.

Are you sure these changes are needed to make it work? The "enable interrupt" you added in `new` is not enough?

Tnze: The calling convention for poll should be to apply the operation when the first poll.

Tnze: According to <https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll>

> Futures alone are inert; they must be actively polled to make progress, meaning that each time the current task is woken up, it should actively re-poll pending futures that it still has an interest in.

We should do nothing before the Future is `polled`. (even though the outer asynchronous function will generate such Future, let the internal `poll_fn` follow this is definitely better)

Tnze: Also noticed that
> Note that on multiple calls to poll, only the Waker from the Context passed to the most recent call should be scheduled to receive a wakeup.

I will fix it.

Dirbaio: > The calling convention is that the operation should be apply when the first time poll is called.

`convert` is *already* an `async fn`. All the code in an `async fn` is already executed only after polling the future for the first time (the future autogenerated by the compiler for `async fn convert()`) The `poll_fn` future is just an "inner" future that the user never sees.

Tnze: You are right. But it's nice to make internal Future follow the convention. And I can't see disadvantage there.

Tnze: And I think the change also prevents duplicated trigger the conversion.

Dirbaio: > You are right. But it's nice to make internal Future follow the convention. And I can't see disadvantage there.

It makes the codeuse a bit more flash, it makes it slightly slower, and it is inconsistent with the rest of embassy-stm32.

> And I think the change also prevents duplicated trigger the conversion.

it doesn't. code inside an `async fn` runs only once, the compiler tracks the state of it.

Dirbaio:
#4056

PR #3807, opened 2025-01-26
embassy-sync: fix clear() to wake senders

ibuki2003:
Calling clear() on `Channel`, `PriorityChannel`, `PubSub`, `zerocopy_channel::Channel` should wake the wakers of senders, as receive() do.


Dirbaio (approved):
ooh, nice find! thanks :)

PR #3808, opened 2025-01-26
Set embassy-rp target in docs-rs metadata

jamesmunns:
Avoid docs.rs build failures like https://docs.rs/crate/embassy-rp/0.3.0/builds/1609962


jamesmunns:
(not self-merging because I'm not sure if it's a desirable idea to make this Just Work, and whether we should add a notice for the docs-rs version of the docs that they are only getting rp2040 docs, not rp2350 docs)

Dirbaio (approved):

PR #3810, opened 2025-01-26
remove checks for reserved I2c addresses

### LINKED ISSUE #3809: can't access non-compliant I2c device
Seeed sells water level sensors that expose their readings over I2c. The lower half of the sensor can be read out over address 0x77 and the upper half can be read out over address 0x78. The problem with this is that address 0x78 is not a valid I2c address, it's one of the reserved addresses meant to be used with 10-bit addressing. Nonetheless, the sensors only respond to 7-bit accesses made to 0x78 and don't respond properly to 10-bit accesses. AFAICT this is not spec-compliant behavior.

Currently, the embassy-rp crate checks that the user doesn't use any of the reserved addresses:
https://github.com/embassy-rs/embassy/blob/ef78441ec1721183ed30f9f21e6bd059f582bd2b/embassy-rp/src/i2c.rs#L473-L475
https://github.com/embassy-rs/embassy/blob/ef78441ec1721183ed30f9f21e6bd059f582bd2b/embassy-rp/src/i2c.rs#L778-L781

I'd like to be able to use my sensors with this crate. I already did some testing and it turns out that the rp2040 will behave just fine when the target address is 0x78 and doesn't prevent reserved addresses from being used. One way to solve this would be to just remove the check. Another way would be to add a config option to allow non-compliant addresses. My question to the maintainers is whether this is something that you want to support and if so what solution you'd prefer.

I'd be willing to contribute a PR to fix this.

Freax13:
There are some non-compliant devices that respond to reserved I2c addresses. rp2040 behaves sanely for these addresses, so let's just allow using them.

Closes #3809

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "defmt", allow(deprecated))] // Suppress warnings for defmt::Format using Error::AddressReserved
+
 //! I2C driver.
 use core::future;
 use core::marker::PhantomData;
@@ -40,6 +42,7 @@ pub enum Error {
     /// Target i2c address is out of range
     AddressOutOfRange(u16),
     /// Target i2c address is reserved
+    #[deprecated = "embassy_rp no longer prevents accesses to reserved addresses."]
     AddressReserved(u16),
 }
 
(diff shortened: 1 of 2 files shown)


Freax13:
I marked `Error::AddressReserved` as deprecated. It should be removed in the next breaking release.

Dirbaio (approved):
thanks!

Issue #3205, opened 2024-07-24
STM32H7 recover from SAI `OverrunError`


elagil:
Hello!

I use the `embassy_stm32::sai::SAI4` peripheral driver on the `STM32H723` for writing audio data to a TDM output.

I have to start writing data immediately after calling `start()` on the driver instance - then it works fine. Otherwise, the writable ring buffer gives an `OverrunError`, which I cannot recover from, because subsequent calls to `write()` also give that error immediately. The same happens when there is a pause in writing, and (I guess) that the DMA consumes all of the data in the ring buffer.

I did not find a way to reset the ring buffer (and probably the driver itself). Do I have to make a new driver instance each time? Otherwise, I can also send zeros to the instance, to keep the buffer entertained, but I would rather be able to switch the peripheral off sometimes.

Thanks!

lulf:
I'm not familiar with SAI, but generally all drivers can be passed a &mut instance upon creation, and when the driver goes out of scope, it switches off the peripheral. Then you can reuse the peripheral with the driver next time you want to use it.

elagil:
Thanks, I tried that and it works fine. I think I will stick with it.

I--P:
Could you please elaborate on how you fixed that ? I'm running into the same issue with read() and I don't understand how to get rid of the overrun error

elagil:
@I--P You can have a look at this implementation that I ended up with: https://github.com/blus-audio/firmware-rs/blob/main/blus_mini_mk2/src/audio_routing.rs#L325

When there is an error, a fresh SAI instance is created with `new_sai4()`.

I--P:
@elagil thanks a lot, that's what I ended up with too. It will be very helpful for many people to have a direct link to your code here.

PR #3812, opened 2025-01-26
Allow stm32 H7 and L5 families to use external SAI sync & allow recovering after dropping a DMA channel

noracarmig:
In [sai/mod.rs](../blob/main/embassy-stm32/src/sai/mod.rs), modified `sai_v3_2pdm` and `sai_v3_4pdm` placement in the `#cfg[...]` conditional compilation statements to allow the corresponding L5 and H7 stm32 chips to use external SAI sync, as stated available in the ST reference manuals.

In [dma_bdma.rs](../blob/main/embassy-stm32/src/dma/dma_bdma.rs), added code to reset the complete_count variable defined in the global channel `STATE` defined in [dma/mod.rs](../blob/main/embassy-stm32/src/dma/mod.rs). For some reason re-creating a BDMA channel resets its global state, but re-creating a DMA channel does not, making BDMA peripherals like SAI4 recoverable after dropping their `Sai<'_, SAI4, W>` instance (like in #3205), but not DMA peripherals like SAI1 for example. One consequence of this is `Sai<'_, SAI1, W>` `Sai<'_, SAI2, W>` and `Sai<'_, SAI3, W>` peripherals are stuck in `Overrun` even after dropping and re-creating them.

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -359,11 +359,13 @@ impl AnyChannel {
         match self.info().dma {
             #[cfg(dma)]
             DmaInfo::Dma(r) => {
+                let state: &ChannelState = &STATE[self.id as usize];
                 let ch = r.st(info.num);
 
                 // "Preceding reads and writes cannot be moved past subsequent writes."
                 fence(Ordering::SeqCst);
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
thank you!

PR #3813, opened 2025-01-26
stm32/eth: rename PHY->Phy, GenericSMI -> GenericPhy. Remove unneeded unsafes.

Dirbaio:
We shouldn't use `unsafe` to mark merely "dangerous" actions, only actions that
actually cause UB.


Dirbaio:
bender run

PR #3823, opened 2025-01-29
Correct ADC channels for RP235XB

flippette:
The ADC channel offset and temperature sensor channel were only correct for RP2040 and RP235XA, this PR makes them also correct for RP235XB.

Based on [this table in the RP2350 datasheet](https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf#%5B%7B%22num%22%3A1067%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C115%2C742.498%2Cnull%5D).

### DIFF
--- a/embassy-rp/src/adc.rs
+++ b/embassy-rp/src/adc.rs
@@ -58,11 +58,21 @@ impl<'p> Channel<'p> {
     }
 
     fn channel(&self) -> u8 {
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #3832, opened 2025-02-01
Fix typo in getting started docs

Graicc:
Replaces "hare" with "are" in "If you hare having issues when running `+cargo run --release+`, please check the following:"


Dirbaio (approved):
thanks!

PR #3833, opened 2025-02-02
Add cancel safety notes to Ticker

embediver:
This just adds a section about cancel safety to the embassy-time Ticker struct and the corresponding method to get a Future for the next tick. 

When looking at the source its quite clear that the Ticker is cancel safe, but it would be nice to have a note in the docs, so people don't have to think about it.


Dirbaio (approved):
thanks!

Issue #3681, opened 2024-12-23
I2S write future never completes in `examples/stm32f4/.../i2s_dma.rs`


vinsynth:
I'm using an stm32f411ceu6 chip. Having replaced the necessary chip definitions in `.cargo/config.toml` and `Cargo.toml` (such that the `blinky.rs` program compiles and runs as expected), and remapped pins & DMA streams/channels as necessary, I find that `i2s.write(...)` in `i2s_dma.rs` never completes. Here's my slightly modified code in case anyone can try this on their own chip:
```
#![no_std]
#![no_main]

use defmt::*;
use embassy_executor::Spawner;
use embassy_stm32::i2s::{Config, I2S};
use embassy_stm32::time::Hertz;
use {defmt_rtt as _, panic_probe as _};

#[embassy_executor::main]
async fn main(_spawner: Spawner) {
    let p = embassy_stm32::init(Default::default());
    info!("Hello World!");

    let mut dma_buffer = [0x00_u16; 128];
    let mut i2s = I2S::new_txonly(
        p.SPI3,
        p.PB5,  // sd
        p.PA15, // ws
        p.PB3,  // ck
        p.PB10, // mck
        p.DMA1_CH7,
        &mut dma_buffer,
        Hertz(48_000),
        Config::default(),
    );

    for i in 0_u16.. {
        i2s.write(&mut [i * 2; 64]).await.ok();
        defmt::panic!("if this message is reached, the write function returned!");
        i2s.write(&mut [i * 2 + 1; 64]).await.ok();
    }
}
```
`i2s.write(...)` still doesn't complete when `I2S::new_txonly_nomck(...)` is used instead. I'm connecting the pins to a PCM5102A chip that has worked with other i2s drivers. Is there something I'm missing with my configuration, or is this a library issue?

elagil:
`i2s.start()` is never called, and I think it should.

elagil:
Another thing I noticed is that I2S calculates its clock based on APB1 clock (default 42 MHz) which usually doesn't match the PLLI2S clock. This needs to be fixed at some point.

vinsynth:
[...]

PR #3836, opened 2025-02-02
set PLLI2S M and SRC for f4 chips which support it

vinsynth:
A `TODO` in embassy-stm32/src/rcc/f247.rs notes the need to support PLLM and plli2s_src for f4 chips which support them; this commit aims to rectify that, though there may be more that must be done (e.g., I'm not sure if this applies only to f4 chips or extends to f7, etc.). See #3681.

### DIFF
--- a/embassy-stm32/src/rcc/f247.rs
+++ b/embassy-stm32/src/rcc/f247.rs
@@ -414,6 +414,11 @@ fn init_pll(instance: PllInstance, config: Option<Pll>, input: &PllInput) -> Pll
         }),
         #[cfg(any(all(stm32f4, not(stm32f410)), stm32f7))]
         PllInstance::Plli2s => RCC.plli2scfgr().write(|w| {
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rcc/f247.rs
```diff
@@ -414,6 +414,11 @@ fn init_pll(instance: PllInstance, config: Option<Pll>, input: &PllInput) -> Pll
         }),
         #[cfg(any(all(stm32f4, not(stm32f410)), stm32f7))]
         PllInstance::Plli2s => RCC.plli2scfgr().write(|w| {
+            #[cfg(any(stm32f411, stm32f412, stm32f413, stm32f446))]
+            w.set_pllm(pll.prediv);
+            #[cfg(any(stm32f412, stm32f413))]
```
Dirbaio, line 419: both cfg's are missing stm32f423, that one also has separate PLLM and PLLSRC for i2s.

Dirbaio (approved):
thank you!

PR #3837, opened 2025-02-03
nRF52833: configure internal LDO

klownfish:
The nRF52833 has an internal adjustable voltage regulator. This is very similar to the nRF52840 so this commit is just 3 lines to enable and reuse that code. Tested on a custom board.


Dirbaio (approved):
neat! thanks

PR #3841, opened 2025-02-03
embassy-usb: fix `core::intrinsics` deprecate warning in nightly.

vDorst:
It seems changed in Rust https://github.com/rust-lang/rust/pull/135003.
The compiler suggest to use `core::mem::copy_nonoverlapping`.
But that seems not exists so I used `core::ptr::copy_nonoverlapping` which is used in other parts in Embassy.

Replaced `core::intrinsics::copy_nonoverlapping` with the `core::ptr::copy_nonoverlapping`.

This fixed this warning.

```
   Compiling embassy-usb v0.4.0 (embassy/embassy-usb)
warning: use of deprecated module `core::intrinsics`: import this function via `std::mem` instead
  --> embassy/embassy-usb/src/class/cdc_ncm/mod.rs:17:23
   |
17 | use core::intrinsics::copy_nonoverlapping;
   |                       ^^^^^^^^^^^^^^^^^^^
```


Dirbaio (approved):
oops, we've been a victim of rust-analyzer autocomplete here, I think :) 
thanks!

PR #3844, opened 2025-02-04
SpiDevice cancel safety: always set CS pin to high on drop

showier-drastic:
Currently, if a transfer is dropped, the CS will stay in a low state, which seems to be undesired. This uses `OnDrop` to guarantee that CS will be high regardless what happens.

Note that it is currently still unsafe, at least for STM32, to drop an on-going SPI transfer. More investigation on this is TBD. Some suggestions on this would be nice, e.g., whether it's desired to implement cancel safety for SPI transfers.

### DIFF
--- a/embassy-embedded-hal/src/shared_bus/asynch/spi.rs
+++ b/embassy-embedded-hal/src/shared_bus/asynch/spi.rs
@@ -25,6 +25,7 @@
 //! let display2 = ST7735::new(spi_dev2, dc2, rst2, Default::default(), 160, 128);
 //! ```
 
+use embassy_hal_internal::drop::OnDrop;
 use embassy_sync::blocking_mutex::raw::RawMutex;
 use embassy_sync::mutex::Mutex;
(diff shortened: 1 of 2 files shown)


embassy-embedded-hal/src/shared_bus/asynch/spi.rs
```diff
@@ -97,11 +102,10 @@ where
 
         // On failure, it's important to still flush and deassert CS.
         let flush_res = bus.flush().await;
-        let cs_res = self.cs.set_high();
+        drop(cs_drop);
```
Dirbaio, line 105: this will silently swallow the error.

I'd suggest doing `cs_drop.defuse()` and then do set_high as before so the error can be handled.

showier-drastic: Changed as you advised. Could you please take a look again?

Dirbaio (approved):
thanks!

PR #3845, opened 2025-02-04
update examples/stm32f4/.../i2s_dma.rs

vinsynth:
This example must've been added before plli2s wasn't configured by default; with this commit, the example configures clocks correctly, in addition to demonstrating stereo audio for a more practical demonstration.

I'm not sure what the policy is on which chip(s) ought to be used in these examples; I configured the example for an f411 chip, because that's what I have on hand, but I can change the pins/clocks to those of a different chip as needed.


Dirbaio (approved):
thanks!

PR #3846, opened 2025-02-04
Add saturating_add and saturating_sub to Instant

wackazong:


lulf (approved):
Thanks!

embassy-time/src/instant.rs
```diff
@@ -115,6 +115,18 @@ impl Instant {
     pub fn checked_sub(&self, duration: Duration) -> Option<Instant> {
         self.ticks.checked_sub(duration.ticks).map(|ticks| Instant { ticks })
     }
+
+    /// Adds a Duration to self. In case of overflow, the maximum value is returned.
+    pub fn saturating_add(mut self, duration: Duration) -> Self {
+        self.ticks = self.ticks.saturating_add(duration.ticks);
```
Dirbaio, line 121: Mutating self is a bit strange. could you do `Self::from_ticks(self.ticks.saturating_add(duration.ticks))`? 

same for sub.

Dirbaio: lol already merged

PR #3850, opened 2025-02-04
Disable clippy error around generated main function

vic1707:
The current implementation of `embassy_executor::main` and generated code causes `clippy` to complain even when using an empty function (for demonstration purposes).
image

```
error: future cannot be sent between threads safely
  --> src/main.rs:22:1
   |
22 | #[embassy_executor::main]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^ future returned by `____embassy_main_task` is not `Send`
   |
note: captured value is not `Send`
  --> src/main.rs:23:15
   |
23 | async fn main(_spawner: Spawner) {}
   |               ^^^^^^^^ has type `embassy_executor::Spawner` which is not `Send`
   = note: `*mut ()` doesn't implement `core::marker::Send`
note: captured value is not `Send`
  --> src/main.rs:23:15
   |
23 | async fn main(_spawner: Spawner) {}
   |               ^^^^^^^^ has type `embassy_executor::Spawner` which is not `Send`
   = note: `*mut ()` doesn't implement `core::marker::Sync`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#future_not_send
   = note: `-D clippy::future-not-send` implied by `-D clippy::nursery`
   = help: to override `-D clippy::nursery` add `#[allow(clippy::future_not_send)]`
```

---

Currently the only way to make said error/warning go away is to put
```rs
#![allow(clippy::future_not_send)]
```
at the top of the module calling `embassy_executor::main`. 
Users unfortunately can't put the `allow` attribute like this (attribute gets removed during expansion).
```rs
#[embassy_executor::main]
#[allow(clippy::future_not_send)]
async fn main(_spawner: Spawner) {}
```
nor
```rs
#[allow(clippy::future_not_send)]
#[embassy_executor::main]
async fn main(_spawner: Spawner) {}
```

Adding the attribute directly in the expanded code makes `clippy` happy. 

Note: This might not be the ideal fix (maybe allowing `async fn main(_spawner: SendSpawner) {}` would work but I don't have enough knowledge on embassy to know if that's faceable or wished). 
Adding the attribute seems like a good compromise, being quick and easy to maintain (maybe using `expect` can be better but it requires a relatively new version of rust)

### DIFF
--- a/embassy-executor-macros/src/macros/main.rs
+++ b/embassy-executor-macros/src/macros/main.rs
@@ -161,6 +161,7 @@ pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
 
     let result = quote! {
         #[::embassy_executor::task()]
+        #[allow(clippy::future_not_send)]
         async fn __embassy_main(#fargs) #out {
             #f_body
         }


Dirbaio (approved):
thanks!

(`clippy::future_not_send` is a very dubious lint anyway especially for embedded. Many futures are not Send in embedded, and that's perfectly okay!)

vic1707:
Thank you, I certainly wasn't expecting a reply that quickly!

> (`clippy::future_not_send` is a very dubious lint anyway especially for embedded. Many futures are not Send in embedded, and that's perfectly okay!)

Thank you for the advice! First time doing embedded and async in Rust so I'll keep that in mind 👌

PR #3851, opened 2025-02-04
rp235x: fix atomics, fix SIO spinlock stuck bug, fix missing core1 reset.

### LINKED ISSUE #3839: [RP235x] various hangs when using multicore
I've observed various different hangs when attempting to use core 1 on both the RP2350A and RP2350B. Specifically I've noticed async SPI on core 1 will hang (the core 1 executor continues to run but the task awaiting the SPI future will never be woken up), and channels will hang on the sender side when sending from core 0 to core 1. Neither of these happen 100% of the time but enough that it makes it almost impossible to use the second core reliably.

I've written an [minimal reproducible example](https://github.com/adom-inc/embassy/blob/cee99c9cb6c9ea4a4492616e8fcc2c93a7b590c4/examples/rp23/src/bin/multicore_hang_demo.rs) that works on a bare Pico 2 from my testing. Given the nature of the bug, it's hard to tell if it could be related to the critical section implementation, the executor implementation, or something else entirely.

Dirbaio:
- **rp: Workaround "SIO spinlock stuck bug", reset PROC1 at boot.**
- **rp: make atomics work properly between cores in rp235x.**

closes #3839

### DIFF
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -598,7 +598,7 @@ pub fn init(config: config::Config) -> Peripherals {
     peripherals
 }
 
-#[cfg(all(feature = "rt", feature = "rp2040"))]
+#[cfg(feature = "rt")]
 #[cortex_m_rt::pre_init]
 unsafe fn pre_init() {
     // SIO does not get reset when core0 is reset with either `scb::sys_reset()` or with SWD.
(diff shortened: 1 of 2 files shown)


embassy-rp/src/lib.rs
```diff
@@ -629,17 +629,52 @@ unsafe fn pre_init() {
...
+        // on RP235x, datasheet says "The FRCE_ON register is a development feature that does nothing in production devices."
+        // No idea why they removed it. Removing it means we can't use PSM to reset SIO, because it comes before
+        // PROC0, so we'd need FRCE_ON to prevent resetting ourselves.
+        //
+        // So we just unlock the spinlock manually.
+        pac::SIO.spinlock(31).write_value(1);
+
+        // We can still use PSM to reset PROC1 since it comes after PROC0 in the state machine.
+        pac::PSM.frce_off().write_and_wait(|w| w.set_proc1(true));
+        pac::PSM.frce_off().write_and_wait(|_| {});
```
jannic, line 658: This contains a tiny race condition: Whatever still runs on core 1 may acquire the spinlock after we unlocked it, before core 1 gets reset. So maybe reset core 1 first?

PR #3857, opened 2025-02-07
Add PIO functions.

swork:
Please consider a PR for embassy-rp to add some PIO functions that look to me to be missing:

ConfigPins::set_sideset_pins(), needed like set_set_pins and set_out_pins. (Is there a reason this isn't there that I haven't found yet?)

Several runtime StateMachine manipulations:
 - addr() of running SM, only useful at a stall point but important for that reason
 - tx_threshold() (user code could just save the last-set value I suppose)
 - set_tx_threshold - I've found several PIO programs "out there" that need to manipulate thresholds.
 - rx_threshold()
 - set_rx_threshold()
 - set_thresholds() - both at once, same value

Please review with extreme prejudice!

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -670,6 +670,16 @@ impl<'d, PIO: Instance> Config<'d, PIO> {
         self.exec.jmp_pin = pin.pin();
     }
 
+    /// Sets the range of pins affected by SIDE instructions. The range must be consecutive.
+    /// Set pins must configured as outputs using [`StateMachine::set_pin_dirs`] to be
+    /// effective.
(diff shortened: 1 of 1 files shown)


CBJamo:
The sideset pins are configured in the `use_program` function. Is there some reason this would need to be done at runtime?

The `addr`, `tx_threshold`, and `rx_threshold` functions should probably be prefixed with `get_` to match the other functions.

swork:
Thanks CBJamo. No, I haven't found a reason to change sideset at runtime, just hadn't found it in use_program() so couldn't sort out how to get it into the config. I removed a dash from one instance of the term "side-set" in use_program's docstring, so the spelling "sideset" occurs there as it does elsewhere, like in PinConfig. Again, I'd managed to miss this, though it's certainly prominent enough otherwise. And I added "get_" to the start of the added getters as you suggested.

PR #3867, opened 2025-02-10
timer driver with the features: "time-driver-mtime", "time-driver-timer1" and "time-driver-aot"

mschnell1:
Adding support for 

- the Risk V system Timer by "time-driver-mtime", 

- Timer 1 by "time-driver-timer1" 

- the  always on timer using the low power oscillator by  "time-driver-aot"

The new features acknowledge `TICK_HZ`  which does not uses any CPU, but also does not make too much sense (and only decently works with the AON Timer).  In fact TICK_HZ always should be be `1_000_000 `for  "time-driver-mtime", "time-driver-timer1" and `1_000 ` for "time-driver-aot"


mschnell1:
@Dirbaio: Dis the pull request as you suggested. Don't understand the not successful checks, yet.

Issue #3875, opened 2025-02-12
Task will leak when use with_timeout


wllenyj:
When using `with_timeout` function, the task will be inserted into the time driver queue. If the `with_timeout` returns early with the other Future ready, in other words, it returns before it expires. The task will remain in the time driver queue, even if the task exits.

So we can temporarily set the timer_queue_item's expires_at to u64::MAX. Prevents the task from being re-inserted into the run queue.
```
diff --git a/embassy-executor/src/raw/mod.rs b/embassy-executor/src/raw/mod.rs
index e38a2af..8117b06 100644
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -224,6 +224,9 @@ impl<F: Future + 'static> TaskStorage<F> {
                 // Make sure we despawn last, so that other threads can only spawn the task
                 // after we're done with it.
                 this.raw.state.despawn();
+
+                // Make sure the task is not woken up by the time driver.
+                this.raw.timer_queue_item.expires_at.set(u64::MAX);
             }
             Poll::Pending => {}
         }
```

PR #3876, opened 2025-02-12
embassy-executor: set expires_at to u64::MAX when task exits

wllenyj:
Ensure the task is not woken up by the time driver by setting the expires_at value to u64::MAX. This change prevents potential issues with task scheduling and ensures proper task management.

FIX: #3875

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -224,6 +224,9 @@ impl<F: Future + 'static> TaskStorage<F> {
                 // Make sure we despawn last, so that other threads can only spawn the task
                 // after we're done with it.
                 this.raw.state.despawn();
+
+                // Make sure the task is not woken up by the time driver.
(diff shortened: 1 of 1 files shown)


Dirbaio:
We can't write `expires_at` from the executor thread, it's a race condition. it must only be touched by the time driver, which can run on a different core/thread/interrupt.

It is intended behavior that a task can stay in the timer queue after exiting. If the timer expires when the task is not spawned it does nothing, if it expires after the user has re-spawned the task it just causes a spurious wake (spurious wakes are allowed in Rust async). It's not a "leak" because tasks are statically allocated so the memory wouldn't be reclaimed anyway, and it's not keeping the task marked as "used" so it doesn't prevent re-spawning the task.

Have you run into *actual* issues due to this, or is this a theoretical concern?

wllenyj:
@Dirbaio Thanks for your reply. I will close this.
I am also trying to dynamically alloc tasks. So I want to save some runtime overhead.

PR #3878, opened 2025-02-12
stm32: i2c-v2: Add error handling

showier-drastic:
Currently when error occurs, we have to wait for timeout, which is less than ideal. Enable related interrupts, and return Err when error occurs.


Dirbaio (approved):
thank you!

PR #3880, opened 2025-02-13
STM32: RTC subsecond support

nikvoid:
RTC driver does not read subsecond register. This PR fixes that and also adds microsecond field to `DateTime` struct.


Dirbaio (approved):
thanks!

Issue #3045, opened 2024-06-05
Re-add stm32 flash otp support?


rmja:
stm32 otp support was removed in https://github.com/embassy-rs/embassy/pull/2809 / https://github.com/embassy-rs/stm32-data/pull/460. What was the specific reason for removing this support, and is there any way to get it back?

PR #3881, opened 2025-02-13
feat(stm32): Add OTP flash region (again)

rmja:
This fixes https://github.com/embassy-rs/embassy/issues/3045.

It depends on https://github.com/embassy-rs/stm32-data/pull/567


rmja:
Now that https://github.com/embassy-rs/stm32-data/pull/567 is merged, how do I make this PR depend on it?

Dirbaio:
change the tag of stm32-metapac in Cargo.toml (in 2 places)

rmja:
Got it. It seems as if I get some unrelated compile errors. I do not know if these should be handled here. Currently it cannot compile the adc module for the stm32c0 series, and I am not familiar with that series, and I am therefore not the right person to fix those issues.

rmja:
@Dirbaio is there anything I can do here for this to get merged?

rmja:
We should consider getting this merged, or get the adc compile error fixed in a different pr so that the build error gets resolved. The current main branch of `stm32-data` cannot currently be used before this PR is applied, as the `OTPRegion` struct does not get generated, see https://github.com/embassy-rs/stm32-data/pull/567#issuecomment-2695805589

Dirbaio:
merged in  #3981

PR #3888, opened 2025-02-15
fix: Fix for async I2C v2 driver sequential read/write reads. Fixes #3887

### LINKED ISSUE #3887: Invalid data for sequential async I2C reads/write reads for STM32 devices
For I2C devices that support variable length reads, sending chunks of data as long as the master keeps ACK:ing after each received byte, sequential reads/write reads will sometimes result in data getting out of sync, causing additional reads to return invalid data.

This seems to be caused by a delay between awaiting the DMA read and sending the software STOP signal, which may result in the slave to have time to send a byte of data in between, thus  potentially causing miss-alignment, meaning subsequent reads will read invalid, mixed-up data.

Fredrik-Reinholdsen:
For I2C devices that support variable length reads, sending chunks of data as long as the master keeps ACK:ing after each received byte, sequential reads will sometimes get out of sync, causing additional reads to return invalid data.

This was caused by a delay between awaiting the DMA read and sending the software STOP signal, which may result in the slave to have time to send a byte of data in between, causing potential miss-alignment. This would then cause subsequent reads to return invalid data.

Async write-reads now no longer send STOP after the initial write, matching the behaviour of the blocking version.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -438,6 +438,7 @@ impl<'d> I2c<'d, Async> {
         write: &[u8],
         first_slice: bool,
         last_slice: bool,
+        send_stop: bool,
         timeout: Timeout,
     ) -> Result<(), Error> {
         let total_len = write.len();
@@ -505,10 +506,12 @@ impl<'d> I2c<'d, Async> {
         .await?;
 
         dma_transfer.await;
-
         if last_slice {
             // This should be done already
             self.wait_tc(timeout)?;
+        }
+
+        if last_slice & send_stop {
             self.master_stop();
         }
 
@@ -556,16 +559,16 @@ impl<'d> I2c<'d, Async> {
                     self.info,
                     address,
                     total_len.min(255),
-                    Stop::Software,
+                    Stop::Automatic,
                     total_len > 255,
                     restart,
                     timeout,
                 )?;
+            } else if remaining_len == 0 {
+                return Poll::Ready(Ok(()));
             } else if !(isr.tcr() || isr.tc()) {
                 // poll_fn was woken without an interrupt present
                 return Poll::Pending;
-            } else if remaining_len == 0 {
-                return Poll::Ready(Ok(()));
             } else {
                 let last_piece = remaining_len <= 255;
 
@@ -581,11 +584,6 @@ impl<'d> I2c<'d, Async> {
         .await?;
 
         dma_transfer.await;
-
-        // This should be done already
-        self.wait_tc(timeout)?;
-        self.master_stop();
-
         drop(on_drop);
 
         Ok(())
@@ -601,7 +599,7 @@ impl<'d> I2c<'d, Async> {
             self.write_internal(address, write, true, timeout)
         } else {
             timeout
-                .with(self.write_dma_internal(address, write, true, true, timeout))
+                .with(self.write_dma_internal(address, write, true, true, true, timeout))
                 .await
         }
     }
@@ -623,7 +621,7 @@ impl<'d> I2c<'d, Async> {
             let next = iter.next();
             let is_last = next.is_none();
 
-            let fut = self.write_dma_internal(address, c, first, is_last, timeout);
+            let fut = self.write_dma_internal(address, c, first, is_last, is_last, timeout);
             timeout.with(fut).await?;
             first = false;
             current = next;
@@ -650,7 +648,7 @@ impl<'d> I2c<'d, Async> {
         if write.is_empty() {
             self.write_internal(address, write, false, timeout)?;
         } else {
-            let fut = self.write_dma_internal(address, write, true, true, timeout);
+            let fut = self.write_dma_internal(address, write, true, true, false, timeout);
             timeout.with(fut).await?;
         }
 


embassy-stm32/src/i2c/v2.rs
```diff
@@ -556,16 +559,16 @@ impl<'d> I2c<'d, Async> {
                     self.info,
                     address,
                     total_len.min(255),
-                    Stop::Software,
+                    Stop::Automatic,
                     total_len > 255,
                     restart,
                     timeout,
                 )?;
+            } else if remaining_len == 0 {
```
Fredrik-Reinholdsen, line 567: Swapped the order in this if-statement since transfer-complete interrupt is not raised with `Stop::Automatic`, that would cause an infinite loop.

embassy-stm32/src/i2c/v2.rs
```diff
@@ -581,11 +584,6 @@ impl<'d> I2c<'d, Async> {
         .await?;
 
         dma_transfer.await;
-
```
Fredrik-Reinholdsen, line 584: Removed waiting for transfer complete since this is no longer raised with `Stop::Automatic`, and removed explicit call to send STOP since this is now done automatically after the last read

embassy-stm32/src/i2c/v2.rs
```diff
@@ -556,16 +559,16 @@ impl<'d> I2c<'d, Async> {
...
                     restart,
                     timeout,
                 )?;
+            } else if remaining_len == 0 {
+                return Poll::Ready(Ok(()));
             } else if !(isr.tcr() || isr.tc()) {
                 // poll_fn was woken without an interrupt present
                 return Poll::Pending;
-            } else if remaining_len == 0 {
-                return Poll::Ready(Ok(()));
```
Fredrik-Reinholdsen, line 568: Swapped the order in this if-statement since transfer-complete interrupt is not raised with Stop::Automatic, that would cause an infinite loop.

Dirbaio (approved):
thank you!

agrif:
Hello! I recently moved from `embassy-stm32` 0.2.0 to git, and suddenly my I2C EEPROM reads stopped working. Git bisect blames this PR, and indeed my reads succeed on the parent commit and fail on this commit.

I'm reading the EEPROM with this code, on an STM32F072CB:

````rust
let eeprom_addr = 0;
let mut buffer = [0; 256];
i2c.write_read(EEPROM_I2C_ADDR, &[eeprom_addr], &mut buffer).await?;
````

On previous commits this works, but on this commit it times out. It reads 255 bytes into `buffer` and then times out waiting for the last byte. From [discussion in #3887](https://github.com/embassy-rs/embassy/issues/3887#issuecomment-2660877113) you were a bit concerned about this 255-byte boundary, and it looks like I've run into a problem with it. Using a smaller buffer works fine.

@Fredrik-Reinholdsen, do you have any insight in to what the issue might be, or how to fix it?

Fredrik-Reinholdsen:
Shit, sorry about this.

I may have an idea.  The maximum amount of bytes that can be read without reload is 255,
so probably something is not being handled properly when setting up the reload.

I am a bit strapped for time tonight, so I may not have time to finish.
I will have a look tonight, and if nothing else I can probably finish it tomorrow.

Thanks for your comment

Fredrik-Reinholdsen:
image

The change I made in the code relates to these configuration bits in the I2C peripheral. Previously the auto-end bit was 0 which cause the sync issues I was getting outlines above. It looks to me like the reload bit should be set when transfer exceeds 255 bytes, and based on my read of the datasheet above, that should yield a TCR (Transfer Complete Reload) interrupt.

I will do some more digging and testing.

Fredrik-Reinholdsen:
Hi again, when you have time could you please try out a potential fix on my forked repo here:
https://github.com/Fredrik-Reinholdsen/embassy

Please let me know if it works any better.

agrif:
Thanks for the quick response! This is a subtle bug and it'd be nice to get it fixed for others, but it's also not hard to work around in my case, so it's not time-critical.

Commit 966174b1cf1e5cbcc2f5d146befcecd4ce27b6a2 fails in the same way -- timeout waiting for the 256th byte.

Fredrik-Reinholdsen:
Hi, sorry for the somewhat late reply. I have sick over the past few days.

I have been trying to see if I have any I2C devices lying around were I could try to replicate the issue.
The best I have is NAU88C22YG audio CODEC. This is not a perfect replica though, it has 63 different registers but it does support arbitrarily long reads since the register being read will just wrap, so I can do reads bigger than 256 bytes.

On this device was not able to replicate the issue, but clearly I did break something since it did work before. I will try to do some more thinking and digging to see if I can figure it out. What I2C EEPROM are you using?

The only other change that I made other than the I2C control bits outlined above is the that async `write_read` no longer issues a STOP signal between the write and the read, but I do not see how this could cause the issue.

Any debugging info is greatly appreciated!

agrif:
Commit bfe0dc528526aaf4fe2438efd3207362366623d1 from your repository fixes my issue, and reads out all 256 bytes every time! Thank you for your help.

Every once in a while I run in to an issue where about half of the read buffer is filled with 0xff at the end, but it's not at all consistent and I suspect this is a different issue entirely.

I am reading from a 24C02 EEPROM. I'm not certain of the manufacturer, as it's deep inside of an HP power supply I'm scared to open.

agrif:
@Fredrik-Reinholdsen could you open a PR for these changes?

Fredrik-Reinholdsen:
Sorry for the delay.
I have been travelling for work for the past 20 days.

Posted new PR now with all the same changes bfe0dc5, just with some added comments

agrif:
No worries! I'm very grateful for your help.

PR #3889, opened 2025-02-16
Use probe-rs as default runner for rp23 examples

9names:
Now that probe-rs supports RP235x targets we should use it for rp23 examples


Dirbaio (approved):
hell yeah! thanks

PR #3890, opened 2025-02-16
rp: add ImageDef features

9names:
It's a bit irritating having examples/projects that are basically identical between rp2040 and rp235x except for the ImageDef include and block definition.

This PR puts the most common definitions into embassy-rp in the same way that boot2 is handled.
If you don't select one of these imagedef features on rp235x, you'll get Secure Exe (which all the examples were using anyway).

One issue with the way this is presently implemented:
If you define an Image Definition in your program *and* you have one of these imagedef features enabled, you will end up with two Image Definitions in your binary (bin crate one will be first).
I don't think embassy-rp sets up a `block loop` of these Image Definitions, so I think the bootloader will use the first one, but I need to confirm this.
Maybe it would make more sense to not have an imagedef supplied by default instead, so it's an explicit opt-in for this?

The three features currently implemented are
- imagedef-secure-exe (execute firmware in Secure Mode)
- imagedef-nonsecure-exe (execute firmware in Non-Secure Mode)
- imagedef-none (no Image Definition provided - user must define their own)

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -106,7 +106,7 @@ imagedef-nonsecure-exe = []
 
 ## Have embassy-rp not provide the Image Definition so you can use your own.
 ## Place your own in the ".start_block" section like:
-## ```
+## ```ignore
 ## use embassy_rp::block::ImageDef;
 ##
 ## #[link_section = ".start_block"]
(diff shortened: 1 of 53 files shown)


Dirbaio (approved):
thanks! this makes examples much simpler :)

PR #3891, opened 2025-02-17
stm32: Add standby mode support

rappet:
This is currently work in progress. Following things are missing in my opinion.

- clearer documentation
- RTC configuration (at least the option to disable it until alarms are supported in general)
- WKUP pins
- add and test more chips

Tested only with an STM32F103 and STM32L031

I don't have a good idea how to build a good API for the WKUP pins.
They are quite a bit different than usual hardware as you can just steal the affected pins from other peripherals
as the chip is stopped anyway (and you might have already used them for a button that you also want to use for waking for example).

I currently see an array/struct with a Config for each pin.

Depending on series, there is also some difference between the WKUP pins. I know at least of those variants:

- if enabled, set to pull down and wake on high/rising edge
- if enabled, wake on low/falling edge or high/rising edge depending on wake-up pin polarity flag

### DIFF
--- a/embassy-stm32/src/pwr.rs
+++ b/embassy-stm32/src/pwr.rs
@@ -1,0 +1,155 @@
+//! Power control
+
+/// Configuration what peripherals should be running during standby or shutdown mode.
+/// It is determined by the disabled peripherals, if shutdown mode is used instead of standby
+/// mode if the chip supports it.
+#[derive(Clone, Copy)]
+#[non_exhaustive]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct StandbyConfig {
+    // empty, we might want to disable RTC here, enable WKUP pins, ...
+    #[cfg(any(stm32f1, stm32f4))]
+    pub enable_wkup: bool,
+}
+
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/pwr.rs
```diff
@@ -0,0 +1,100 @@
...
+/// The RTC can wake up the core at a specific time.
+/// If RTC is enabled during standby, more current is used.
+///
+/// ## IWDG
+///
+/// If the Independent Watchdog is enabled before standby,
+/// it will not be disabled and can still reset the core if it was not pet.
+/// This can also be used to wake up after and inaccurate amount of time has passed in standby mode,
+/// which has the advantage that it is in most cases less power hungry than the RTC.
+pub fn standby(_config: &StandbyConfig) {
```
Dirbaio, line 52:
```suggestion
pub fn standby(_config: &StandbyConfig) -> ! {
```

denotes nicely that this doesn't return.

rappet: Done, had to move `unreachable!` out of the critical section.

Dirbaio:
Hello,

PR #3894, opened 2025-02-17
Fix function name

Flamenco:
I searched the repo for other references, and none exist.  Being Rust, I am confident any breaking changes in existing source code clients will be obvious.


Dirbaio (approved):
lol, thanks!

PR #3897, opened 2025-02-17
[embassy-net]: Add `serde:{Deserialize, Serialize}` for `embassy_net::Config`

AnthonyGrondin:
This PR adds the `serde:{Deserialize, Serialize}` traits for `embassy_net::Config` and all of its underlying types.
This depends on:
-  https://github.com/smoltcp-rs/smoltcp/pull/1042 

for `Ipv[4|6]Cidr` types and `RetryConfig`.

My specific use case is to be able to communicate the running configuration from devices to an API, in JSON format, as well as saving the current configuration in NVS, using [Postcard](https://github.com/jamesmunns/postcard).


AnthonyGrondin:
Closing as users should provide their own types if needing to serialize / save / transfer configurations for the following reasons:
- The structs have a lot of "programmer-oriented, internal" stuff that really shouldn't be edited by a user-level config.
- There's no guarantee these structs will stay unchanged in future versions of embassy-net/smoltcp
- Application specific structs provide better control and guarantees over what's configurable with better fowards-compat, and a more lightweight conversion.

PR #3901, opened 2025-02-18
Inital rp23 ci tests

CBJamo:
Some tests need more work.
* The adc test builds, but isn't set up correctly for the 2350 hardware yet.
* The multicore and gpio_multicore tests only work from flash, seems to be a probe-rs issue.
* The i2c and flash tests also only works from flash, these are probably bugs but I don't have time to run them down now.
* The 2350 gpio test skips anything with pull downs. I think these fail because of E9.

The float, bootsel, cyw43, and ethernet tests don't have 2350 equivalents. There's no reason to use the float romfuncs, use the FPU. Bootsel as a button isn't supported on the 2350 yet. The wifi and eth tests don't have appropriate hardware.

The i2c test has also been tweaked to run on one core.

Also, thanks 9names, without #3890 this would have been much uglier.


Dirbaio:
bender run

CBJamo:
bender run

PR #3360, opened 2024-09-21
Bump rand core to 0.9.0

ionspin:
A draft update of rand rand_core and rand_chacha to 0.9.0, it's a draft since 0.9.0 is still at `0.9.0-alpha.2`. Implemented `TryRngCore` to preserve `try_fill_bytes` API.

PR #3902, opened 2025-02-19
Bump rand core to 0.9.1 and rand to 0.9.0

aurelj:
This is kind of a follow-up on #3360.

Note that rand_core now provides a default implementation for TryRngCore so try_fill_bytes() is automatically implemented.

### DIFF
--- a/embassy-nrf/Cargo.toml
+++ b/embassy-nrf/Cargo.toml
@@ -156,7 +156,7 @@ log = { version = "0.4.14", optional = true }
 cortex-m-rt = ">=0.6.15,<0.8"
 cortex-m = "0.7.6"
 critical-section = "1.1"
-rand_core = "0.6.3"
+rand_core = "0.9.1"
 fixed = "1.10.0"
 embedded-storage = "0.3.1"
 embedded-storage-async = "0.4.1"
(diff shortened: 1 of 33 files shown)


aurelj:
Note that the failed test is due to `ed25519_dalek` depending on an old version of rand_core.
See https://github.com/dalek-cryptography/curve25519-dalek/pull/729

Dirbaio:
lgtm, but then we need to wait for an `ed25519_dalek` release

alternatively we could impl *both* the 0.8 and 0.9 trait, similar to how we do with embedded-hal 0.2 and 1.0

Dirbaio:
okay, sorry for taking so long to come back to this.

I don't think we should bump 0.6->0.9 directly, since this is a pretty widely used trait and seems most users have'nt updated yet. For example many rust-crypto crates still use 0.6. If we do the update it'll be impossible to use embassy hals together with rust crypto crates to e.g. generate a private key.

Instead, could you *add* 0.9 without removing 0.6? Add two deps `rand-core-06` and `rand-core-09` and impl both?

Since the `rand-core` crate is smallish and has no deps I don't think we should gate this with Cargo features.

Dirbaio:
https://github.com/embassy-rs/embassy/pull/4217

PR #3907, opened 2025-02-21
avr: Change ci-nightly target

djdisodo:
avr: Change ci-nightly target from avr-unknown-gnu-atmega328 to avr-none

Related to: https://github.com/rust-lang/rust/pull/131651


Dirbaio:
merged with #3963

PR #3910, opened 2025-02-21
Embassy-stm32: use constant array instead of match for PLL div/mul

maor1993:
# TL;DR
Change saves 1KB of code in debug and 1.2KB in release builds

## Why? 
As I was working with an example project I noticed that the `init_pll` function was taking a lot of space
digging deeper using `cargo bloat` I discovered that the abstraction to convert PLL DIV/MUL enums takes **1230 bytes**

Main issue comes from the fact that match statements are not a zero cost abstraction, ergo for large match statements (such as plln which has ~120 values), the size blows up quickly.

## Solution
This pull request replaces the proc macro with a constant array of the same style, and using the enum value as an index to this constant array.
RESERVED values output both 0 on mul and div, causing div/0 fault

## Testing
Using `cargo bloat` and `cargo bloat --release` I can see the following improvements for the following code section

Tested and ran on Nucleo STM32L432

profiles:

```toml
[profile.dev]
codegen-units = 1
debug = 2
debug-assertions = true # <-
incremental = false
opt-level = 2 # <-
overflow-checks = true # <-



[profile.release]
debug = 2
opt-level = "s"
lto = true
codegen-units = 1
```

```rust
...
    {
        use embassy_stm32::rcc::*;
        config.rcc.hsi48 = Some(Hsi48Config {
            sync_from_usb: true,
        }); // needed for USB
        config.rcc.sys = Sysclk::PLL1_R;
        config.rcc.hsi = true;
        config.rcc.pll = Some(Pll {
            source: PllSource::HSI,
            prediv: PllPreDiv::DIV2,
            mul: PllMul::MUL13,
            divp: None,
            divq: None,
            divr: Some(PllRDiv::DIV4), // sysclk 26Mhz (16/2)*13/4
        });
        config.rcc.mux.clk48sel = mux::Clk48sel::HSI48;

        config.rcc.mux.adcsel = mux::Adcsel::SYS;
    }
...
```

### pre modification debug
0.0%   1.8%    728B    embassy_stm32 embassy_stm32::rcc::_version::pll::init_pll
0.0%   2.0%    826B    embassy_stm32 embassy_stm32::_generated::<impl core::ops::arith::Mul<stm32_metapac::rcc::vals::Plln> for embassy_stm32::time::Hertz>::mul
0.0%   1.0%    404B    embassy_stm32 embassy_stm32::_generated::<impl core::ops::arith::Div<stm32_metapac::rcc::vals::Pllp> for embassy_stm32::time::Hertz>::div

Total bytes: **1958B**

### post modification debug
0.0%   2.3%    **992B**    embassy_stm32 embassy_stm32::rcc::_version::pll::init_pll

### pre modification release
0.2%   6.0%  **1.8KiB**            embassy_stm32 embassy_stm32::rcc::_version::pll::init_pll

### post modification release
0.0%   1.8%    **694B**    embassy_stm32 embassy_stm32::rcc::_version::pll::init_pll

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1434,8 +1434,8 @@ fn main() {
 
         if is_rcc_name(e.name) {
             let enum_name = format_ident!("{}", e.name);
-            let muls_ident = format_ident!("{}_MULS", enum_name);
-            let divs_ident = format_ident!("{}_DIVS", enum_name);
+            let muls_ident = format_ident!("{}_MULS", e.name.to_uppercase());
+            let divs_ident = format_ident!("{}_DIVS", e.name.to_uppercase());
 
             let mut mul_values = Vec::new();
             let mut div_values = Vec::new();
@@ -1474,7 +1474,7 @@ fn main() {
                 impl core::ops::Div<crate::pac::rcc::vals::#enum_name> for crate::time::Hertz {
                     type Output = crate::time::Hertz;
                     fn div(self, rhs: crate::pac::rcc::vals::#enum_name) -> Self::Output {
-                        let index = (rhs.to_bits() as usize);
+                        let index = rhs.to_bits() as usize;
                         self * #divs_ident[index] / #muls_ident[index]
                     }
                 }
@@ -1482,7 +1482,7 @@ fn main() {
                 impl core::ops::Mul<crate::pac::rcc::vals::#enum_name> for crate::time::Hertz {
                     type Output = crate::time::Hertz;
                     fn mul(self, rhs: crate::pac::rcc::vals::#enum_name) -> Self::Output {
-                        let index = (rhs.to_bits() as usize);
+                        let index = rhs.to_bits() as usize;
                         self * #muls_ident[index] / #divs_ident[index]
                     }
                 }
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1434,43 +1434,48 @@ fn main() {
 
         if is_rcc_name(e.name) {
             let enum_name = format_ident!("{}", e.name);
-            let mut muls = Vec::new();
-            let mut divs = Vec::new();
+            let muls_ident = format_ident!("{}_MULS", enum_name);
+            let divs_ident = format_ident!("{}_DIVS", enum_name);
+
+            let mut mul_values = Vec::new();
+            let mut div_values = Vec::new();
+            let mut min_val:usize = e.variants[0].value as usize;
             for v in e.variants {
                 let Ok(val) = parse_num(v.name) else {
                     panic!("could not parse mul/div. enum={} variant={}", e.name, v.name)
                 };
-                let variant_name = format_ident!("{}", v.name);
-                let variant = quote!(crate::pac::rcc::vals::#enum_name::#variant_name);
                 let num = val.num;
                 let denom = val.denom;
-                muls.push(quote!(#variant => self * #num / #denom,));
-                divs.push(quote!(#variant => self * #denom / #num,));
-            }
+                // valid_count is our running index for valid variants.
+                mul_values.push(quote!(#num));
+                div_values.push(quote!(#denom));
 
-            g.extend(quote! {
+            }
+        
+            let array_len = mul_values.len(); 
+
+        
+            g.extend(quote! { 
+                const #muls_ident: [u32; #array_len] = [ #(#mul_values),* ];
+                const #divs_ident: [u32; #array_len] = [ #(#div_values),* ];
+        
                 impl core::ops::Div<crate::pac::rcc::vals::#enum_name> for crate::time::Hertz {
                     type Output = crate::time::Hertz;
                     fn div(self, rhs: crate::pac::rcc::vals::#enum_name) -> Self::Output {
-                        match rhs {
-                            #(#divs)*
-                            #[allow(unreachable_patterns)]
-                            _ => unreachable!(),
-                        }
+                        let index = (rhs.to_bits() as usize)-#min_val;
+                        self * #divs_ident[index] / #muls_ident[index]
                     }
                 }
+        
                 impl core::ops::Mul<crate::pac::rcc::vals::#enum_name> for crate::time::Hertz {
                     type Output = crate::time::Hertz;
                     fn mul(self, rhs: crate::pac::rcc::vals::#enum_name) -> Self::Output {
-                        match rhs {
-                            #(#muls)*
-                            #[allow(unreachable_patterns)]
-                            _ => unreachable!(),
-                        }
+                        let index = (rhs.to_bits() as usize)-#min_val;
+                        self * #muls_ident[index] / #divs_ident[index]
                     }
                 }
             });
-        }
+        } 
     }
 
     // ========
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1453,10 +1453,7 @@ fn main() {
                 min_val = v.value;
 
                 let Ok(val) = parse_num(v.name) else {
-                    panic!(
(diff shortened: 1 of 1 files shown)


maor1993:
@Dirbaio any issues?

Dirbaio:
This change actually increases code size. You should compare the output for `cargo size`, that's the true final flash size to be optimized. The reason `cargo bloat` is misleading is the lookup tables create separate symbols that don't count against `init_pll`'s size. What's worse, these go in `.rodata` so they don't even show up in `cargo bloat`'s totals, it only summarizes `.text`.

```
===== BEFORE
[dirbaio@mars stm32l4]$ cargo size --release --bin blinky
  15944      80   34192   50216    c428 blinky

[dirbaio@mars stm32l4]$ cargo objdump --release --bin blinky -- -x | grep -E '(init_pll|_MULS|_DIVS)'
08002578 l     F .text  00000786 embassy_stm32::rcc::_version::pll::init_pll::hf505a4f88f61eb6e

===== AFTER
[dirbaio@mars stm32l4]$ cargo size --release --bin blinky
  16140      80   34192   50412    c4ec blinky
[dirbaio@mars stm32l4]$ cargo objdump --release --bin blinky -- -x | grep -E '(init_pll|_MULS|_DIVS)'
08002558 l     F .text  000001e0 embassy_stm32::rcc::_version::pll::init_pll::hf505a4f88f61eb6e
080038ac l     O .rodata        00000044 embassy_stm32::_generated::HPRE_MULS::h534a7e457edaf32b
080038f0 l     O .rodata        00000044 embassy_stm32::_generated::HPRE_DIVS::h0410e90870358646
08003934 l     O .rodata        00000040 embassy_stm32::_generated::PLLM_MULS::hde42c69d5e93770d
08003b74 l     O .rodata        00000200 embassy_stm32::_generated::PLLN_DIVS::h8cfd928d84544d74
08003974 l     O .rodata        00000200 embassy_stm32::_generated::PLLN_MULS::h9a2a03e94af0ae34
08003d74 l     O .rodata        00000080 embassy_stm32::_generated::PLLP_MULS::h9decc53c09ff253a
08003df4 l     O .rodata        00000080 embassy_stm32::_generated::PLLP_DIVS::ha755cce96fbbff47
08003e74 l     O .rodata        00000010 embassy_stm32::_generated::PLLR_MULS::h2c3d14d53e3d9bac
08003e84 l     O .rodata        00000024 embassy_stm32::_generated::PPRE_MULS::h32de3fe68e85b606
08003ea8 l     O .rodata        00000024 embassy_stm32::_generated::PPRE_DIVS::h49e5da63674d40c5
```

(to get the above to work I had to change them from `const` to `static`, otherwise the symbol would have an unfindable name. This doesn't affect code size).

I agree the codegen for div/mul is terrible and we should improve it. Some ideas to get this to actually decrease size:
- Most of the enums are just "multiply by X" or "divide by X" so the fractions will almost always be `1/x` or `x/1`. The codegen could detect this and make 1 lookup table for `x` instead of 2 for num/denom.
- Codegen could detect the max value and use u8 or u16 for the table instead of u32 if it fits.
- For many of these enums, the value and the 'raw' bit value are related by a constant offset. Codegen could detect it and just generate `lhs / (rhs.to_bits() + offset)`. No lookup table needed.
- If all muls/divs are powers of 2 it could generate `lhs >> table[rhs.to_bits()]`. This saves doing a division (which is slow and generates big code) and could make the table fit into an u8 instead of an u16.

embassy-stm32/build.rs
```diff
@@ -1434,39 +1434,53 @@ fn main() {
...
-            let mut muls = Vec::new();
-            let mut divs = Vec::new();
+            let muls_ident = format_ident!("{}_MULS", e.name.to_uppercase());
+            let divs_ident = format_ident!("{}_DIVS", e.name.to_uppercase());
+
+            let mut mul_values = Vec::new();
+            let mut div_values = Vec::new();
+            let mut min_val = 0;
             for v in e.variants {
+                // check if the diff is greater than 1, if so, we need to fill in rserverd values
```
Dirbaio, line 1444: this is assuming `e.variants` is sorted increasing by value, which doesn't have to be the case.

maor1993:
Argh, My bad.

I think this means we can close this PR, as it might be best to start fresh if we do want to implement your suggestions

regarding each one:

>Most of the enums are just "multiply by X" or "divide by X" so the fractions will almost always be 1/x or x/1. The codegen could detect this and make 1 lookup table for x instead of 2 for num/denom.

In that case the compiler should optimize out the div, rolling back my changes and doing:
`cargo objdump --release --bin blinky -- -S | grep -e '_generated::<impl core::ops::arith::Mul<stm32_metapac::rcc::vals::Plln.*:$' -A 1000`

which has the generated code:
```rust
impl core::ops::Mul<crate::pac::rcc::vals::Plln> for crate::time::Hertz {
    type Output = crate::time::Hertz;
    fn mul(self, rhs: crate::pac::rcc::vals::Plln) -> Self::Output {
        match rhs {
            crate::pac::rcc::vals::Plln::MUL8 => self * 8u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL9 => self * 9u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL10 => self * 10u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL11 => self * 11u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL12 => self * 12u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL13 => self * 13u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL14 => self * 14u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL15 => self * 15u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL16 => self * 16u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL17 => self * 17u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL18 => self * 18u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL19 => self * 19u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL20 => self * 20u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL21 => self * 21u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL22 => self * 22u32 / 1u32,
            crate::pac::rcc::vals::Plln::MUL23 => self * 23u32 / 1u32,
...
```

results in just a muls operation per match and a table in the start of the function:
```assembly
80008c6: e8df f011     tbh     [pc, r1, lsl #1]
 80008ca: 78 00 30 01   .word   0x01300078
 80008ce: 20 01 2a 01   .word   0x012a0120
 80008d2: 04 01 4f 01   .word   0x014f0104
 80008d6: 5e 01 2d 01   .word   0x012d015e
...
 8000c70: 213b          movs    r1, #0x3b
;         Hertz(self.0 * rhs)
 8000c72: 4348          muls    r0, r1, r0
;     }
 8000c74: 4770          bx      lr
```



>Codegen could detect the max value and use u8 or u16 for the table instead of u32 if it fits.

Seems like the compiler already solves this



>For many of these enums, the value and the 'raw' bit value are related by a constant offset. Codegen could detect it and just generate lhs / (rhs.to_bits() + offset). No lookup table needed.

Same here


>If all muls/divs are powers of 2 it could generate lhs >> table[rhs.to_bits()]. This saves doing a division (which is slow and generates big code) and could make the table fit into an u8 instead of an u16.

Unfortunately not all mul/divs are power of twos.
and also it seems that the compiler avoids using the UDIV/SDIV operation and actually does [multiplication by reciprocal](https://stackoverflow.com/questions/30790184/perform-integer-division-using-multiplication) (wow I learned something new today) 

Either way, it seems like current impl is best, I'm closing this one

PR #3913, opened 2025-02-23
Add blinky example for RPi Pico 2W board

U007D:
* also make RPi Pico 2(W) examples easier to find under `rp235x` folder
* also make blinky examples easier to find (sort consistently when all start with "blinky")


U007D:
Verified: the new example code runs on RPi Pico 2 W under macOS `probe-rs` v0.27.0.

9names:
This is a lot of changes for PR that's supposed to add one example.
Are you sure all of these changes are for the better? Like, why does the blink period for the existing blinky example need to change here?

U007D:
The non-W period changed from 1 second for the RPi Pico to 250ms for the RPi Pico 2.  So the example I added for the RPi Pico 2W aligns with the RPi Pico 2 example since that where Pico 2 users will be.

The other changes all make the RP2350 examples easier for users to discover.  Looking by commit you can see the big change is actually just a folder rename.

Dirbaio (approved):
thanks!

Issue #3883, opened 2025-02-14
Broken embedded_hal and embedded_hal_nb reads for stm32 BufferedUartRx


DrTobe:
We (with @embediver) were using the `BufferedUartRx` from `embassy-stm32` via the `embedded_hal_nb` traits but it did not work as expected. But the `embedded_io::Read` implementation works.

Comparing [`embedded_hal_02::serial::Read for BufferedUartRx`](https://github.com/embassy-rs/embassy/blob/29d05328f9b33953b8476ca3a5a7b1e3c9adeaf2/embassy-stm32/src/usart/buffered.rs#L910) (which is also used for `embedded_hal_nb`) with [`BufferedUartRx::blocking_read`](https://github.com/embassy-rs/embassy/blob/29d05328f9b33953b8476ca3a5a7b1e3c9adeaf2/embassy-stm32/src/usart/buffered.rs#L576) (which is used for `embedded_io::Read`) reveals that the implementations are completely different. The working `blocking_read` implementation uses the underlying ring buffer while the broken `embedded_hal_02::serial::Read` implementation seems to access the registers directly.

I assume this is some kind of copy-paste-error which may be relatively easy to fix. Could someone confirm my assumption? If so, should we supply a fix ourselves (which would need some work to understand the code properly first) or is someone able to quickly do it?

Dirbaio:
Oh yes, that's very wrong. The `nb` implementation should also pop bytes from the ringbuffer, not from the hardware directly.

embedded-rust-iml-user:
> If so, should we supply a fix ourselves (which would need some work to understand the code properly first) or is someone able to quickly do it?

If someone wants to step up, feel free. Otherwise, we do our best as soon as we find the time to do it :) Just please let us know then so we don't do the work twice.

DrTobe:
> If so, should we supply a fix ourselves (which would need some work to understand the code properly first) or is someone able to quickly do it?

If someone wants to step up, feel free. Otherwise, we do our best as soon as we find the time to do it :) Just please let us know then so we don't do the work twice.

kpfleming:
This should be closed as #3915 was merged.

PR #3915, opened 2025-02-24
fix: BufferedUartRx embedded_hal_nb::serial::Read impl

CL137:
This fixes the following issue: https://github.com/embassy-rs/embassy/issues/3883

The function was tested and works as expected (in our case at least)

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -908,26 +908,16 @@ impl<'d> embedded_hal_02::serial::Read<u8> for BufferedUartRx<'d> {
     type Error = Error;
 
     fn read(&mut self) -> Result<u8, nb::Error<Self::Error>> {
-        let r = self.info.regs;
-        unsafe {
-            let sr = sr(r).read();
-            if sr.pe() {
-                rdr(r).read_volatile();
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usart/buffered.rs
```diff
@@ -908,26 +908,16 @@ impl<'d> embedded_hal_02::serial::Read<u8> for BufferedUartRx<'d> {
...
-            } else {
-                Err(nb::Error::WouldBlock)
+        let state = self.state;
+        let mut rx_reader = unsafe { state.rx_buf.reader() };
+
+        if let Some(data) = rx_reader.pop_one() {
+            Ok(data)
+        } else {
+            if state.rx_buf.is_full() {
+                self.info.interrupt.pend();
```
Dirbaio, line 918: why pend if full? this would create an interrupt storm.

it should be the other way: in the Ok branch, if the ringbuffer *was* full and thanks to this read it stops being full we have to pend the irq to kickstart reading bytes again. https://github.com/embassy-rs/embassy/blob/c52f56785fbc4ad7feb7a0f61bd40f0d3563a845/embassy-stm32/src/usart/buffered.rs#L617-L621

Dirbaio (approved):
thanks!

PR #3923, opened 2025-02-25
Add GPDMA linked-list + ringbuffer support

elagil:
## Changes
- Removed GPDMA feature gating from `cryp`, `sai`, `spdifrx`, and `i2c_v2` drivers.
- Adjusted DMA/BDMA/GPDMA interfaces -> unified function names and docstrings.
- Added linked-list support.
- Added GPDMA ringbuffer support.
- Added STM32H5 SAI with GPDMA example.

## Tested
Please comment, and I will add wanted/tested combinations (if such a peripheral combination exists).

- [x] SAI with GPDMA (ringbuffered) on STM32H5
- [x] UCPD with GPDMA on STM32H5
- [ ] CRYP with GPDMA?
- [ ] SPDIFRX with GPDMA?
- [ ] I2C_V2 with GPDMA?

### DIFF
--- a/embassy-stm32/src/dma/gpdma/linked_list.rs
+++ b/embassy-stm32/src/dma/gpdma/linked_list.rs
@@ -166,15 +166,27 @@ impl<const ITEM_COUNT: usize> Table<ITEM_COUNT> {
     /// Create a ping-pong linked-list table.
     ///
     /// This uses two linked-list items, one for each half of the buffer.
-    pub unsafe fn new_ping_pong<W: Word>(request: Request, peri_addr: *mut W, buffer: &mut [W]) -> Table<2> {
+    pub unsafe fn new_ping_pong<W: Word>(
+        request: Request,
+        peri_addr: *mut W,
+        buffer: &mut [W],
+        direction: Dir,
+    ) -> Table<2> {
         // Buffer halves should be the same length.
         let half_len = buffer.len() / 2;
         assert_eq!(half_len * 2, buffer.len());
 
-        let items = [
-            LinearItem::new_read(request, peri_addr, &mut buffer[..half_len]),
-            LinearItem::new_read(request, peri_addr, &mut buffer[half_len..]),
-        ];
+        let items = match direction {
+            Dir::MemoryToPeripheral => [
+                LinearItem::new_write(request, &mut buffer[..half_len], peri_addr),
+                LinearItem::new_write(request, &mut buffer[half_len..], peri_addr),
+            ],
+            Dir::PeripheralToMemory => [
+                LinearItem::new_read(request, peri_addr, &mut buffer[..half_len]),
+                LinearItem::new_read(request, peri_addr, &mut buffer[half_len..]),
+            ],
+        };
+
         Table::new(items)
     }
 
--- a/embassy-stm32/src/dma/gpdma/ringbuffered.rs
+++ b/embassy-stm32/src/dma/gpdma/ringbuffered.rs
@@ -12,7 +12,7 @@ use super::{AnyChannel, TransferOptions, STATE};
 use crate::dma::gpdma::linked_list::{RunMode, Table};
 use crate::dma::ringbuffer::{DmaCtrl, Error, ReadableDmaRingBuffer, WritableDmaRingBuffer};
 use crate::dma::word::Word;
-use crate::dma::{Channel, Request};
+use crate::dma::{Channel, Dir, Request};
 
 struct DmaCtrlImpl<'a>(Peri<'a, AnyChannel>);
 
@@ -67,7 +67,7 @@ impl<'a, W: Word> ReadableRingBuffer<'a, W> {
         options: TransferOptions,
     ) -> Self {
         let channel: Peri<'a, AnyChannel> = channel.into();
-        let table = Table::<2>::new_ping_pong::<W>(request, peri_addr, buffer);
+        let table = Table::<2>::new_ping_pong::<W>(request, peri_addr, buffer, Dir::PeripheralToMemory);
 
         Self {
             channel,
@@ -207,7 +207,7 @@ impl<'a, W: Word> WritableRingBuffer<'a, W> {
         options: TransferOptions,
     ) -> Self {
         let channel: Peri<'a, AnyChannel> = channel.into();
-        let table = Table::<2>::new_ping_pong::<W>(request, peri_addr, buffer);
+        let table = Table::<2>::new_ping_pong::<W>(request, peri_addr, buffer, Dir::MemoryToPeripheral);
 
         Self {
             channel,
@@ -222,6 +222,7 @@ impl<'a, W: Word> WritableRingBuffer<'a, W> {
         // Apply the default configuration to the channel.
         unsafe { self.channel.configure_linked_list(&self.table, self.options) };
         self.table.link(RunMode::Circular);
+
         self.channel.start();
     }
 
--- a/embassy-stm32/src/dma/mod.rs
+++ b/embassy-stm32/src/dma/mod.rs
@@ -28,10 +28,13 @@ use embassy_hal_internal::{impl_peripheral, PeripheralType};
 
 use crate::interrupt;
 
+/// The direction of a DMA transfer.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-enum Dir {
+pub enum Dir {
+    /// Transfer from memory to a peripheral.
     MemoryToPeripheral,
+    /// Transfer from a peripheral to memory.
     PeripheralToMemory,
 }
 


etiennecollin:
Hi @elagil,

I just came across this PR from #4025 and I'm really excited about the potential of getting it merged! I was wondering if there's anything still pending that’s blocking the merge.

I'd be happy to help out if there’s anything I can do to move it forward; just let me know!

Thanks you and have a great day.

elagil:
Hi @etiennecollin , that's great to hear! This PR is only pending review by a maintainer, but if you are interested, you can do that too. I am happy to receive some feedback. 

In a later PR, the logic of queuing new linked list items should be improved. This PR implements a linked list GPDMA variant of the well known circular mode of the regular DMA. In the future we only want to extend the linked list queue with new items, if the user application provides new data/read previous data. This way, the GPDMA stops running by itself when the user application stops interacting with it.

etiennecollin:
Hi again @elagil, thank you for your answer!

First off, great work on this implementation! For now, I have one suggestion to give users more control over DMA channel settings. Let me know what you think.

# Issue

Currently, after calling one of the `start()` functions (see [\[1\]](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L90) and [\[2\]](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L236)), the DMA channel’s parameters are fixed and can’t be tweaked, and the DMA is started. Allowing users to adjust things like `trigsel` or `trigpol` after the initial _"default"_ configuration would unlock use cases such as chaining transfers on custom hardware events.

## Example

Imagine scanning a keyboard matrix using the DMA (the algorithm is roughly described [here](https://summivox.wordpress.com/2016/06/03/keyboard-matrix-scanning-and-debouncing/)). Here are the steps

1. Use GPDMA CH0 to set each column GPIO low in sequence through the GPIO ports.
2. Use GPDMA CH1 to read all row pins every time CH0 finishes writing.

> Usually, without the DMA, we would be setting each column GPIO pin low one at a time while keeping the others high, then reading the state of all row GPIO pins to detect which key (intersection of row and column) is pressed.

To do that, you’d want to override CH1’s trigger to event 22 (GPDMA1_CH0_TC) with code like:

```rust
// This is configuring the GPDMA CH1
// This is not tested code, but helps understand the example
ch.tr2().write(|w| {
    w.set_trigsel(22); // Trigger on gpdma1_ch0_tc (based on STM32U5 reference manual p.692).
    w.set_swreq(pac::gpdma::vals::Swreq::HARDWARE);
    w.set_trigm(pac::gpdma::vals::Trigm::LINKED_LIST_ITEM);
    w.set_trigpol(pac::gpdma::vals::Trigpol::RISING_EDGE);
    w.set_tcem(pac::gpdma::vals::Tcem::EACH_LINKED_LIST_ITEM);
});
```

This is just one scenario, but there are many other reasons to override the default channel settings.

# Suggested Solution

I suggest the following solution, but feel free to suggest something else:

- Decouple DMA configuration from the `start()` functions ([\[1\]](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L90) and [\[2\]](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L236)).
- Call the [`configure_linked_list()`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/mod.rs#L285) functions in [`ReadableRingBuffer::new()`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L61) and [`WriteableRingBuffer::new()`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L205).
  - This ensures the DMA is configured before being started.
  - An alternative could be to let the user manually call the configuration functions. There should then be a `dma_configured` flag added into the state that is only set once a configuration method has run. That flag would be checked in `start()`.
- Add a `pub unsafe fn configure()` function to [`ReadableRingBuffer`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L57) and [`WriteableRingBuffer`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/ringbuffered.rs#L203).
  - These are functions which return the `ch` channel handle back to the user for further configuration.
  - Keeping both configure functions unsafe maintains the existing guarantees.
  - The user can decide whether to manually override some settings at their own risk.

> This could also be extended to the simpler [`Transfer`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/mod.rs#L468) struct.

## Example usage

Usage could be as follows:

```rust
// This is not tested code, but helps understand the example
let mut dma_ring_buffer = unsafe {
    ReadableRingBuffer::new(
        channel
        request
        peri_addr
        &mut read_buffer,
        TransferOptions::default(),
    )
};

unsafe {
	let ch = dma_ring_buffer.configure();
	ch.tr2().write(|w| { // or use .modify(|w| ...);
	    w.set_trigsel(22);
	    w.set_swreq(pac::gpdma::vals::Swreq::HARDWARE);
	    w.set_trigm(pac::gpdma::vals::Trigm::LINKED_LIST_ITEM);
	    w.set_trigpol(pac::gpdma::vals::Trigpol::RISING_EDGE);
	    w.set_tcem(pac::gpdma::vals::Tcem::EACH_LINKED_LIST_ITEM);
	});
}

dma_ring_buffer.start();
```

I considered something similar to the [`TransferOptions`](https://github.com/elagil/embassy/blob/d02714cdd1240f853d7bebf1435a6faa942d53cc/embassy-stm32/src/dma/gpdma/mod.rs#L57) struct, but it would be verbose and hard to maintain.

# Conclusion

I have not been working with embassy for long, so there might be a simpler way (or an existing way) to do what I am suggesting. Let me know what you think. I can work on implementing the changes as soon as you answer! Have a great one!

embassy-stm32/src/dma/gpdma/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+
+    fn request_pause(&self) {
+        let info = self.info();
+        let ch = info.dma.ch(info.num);
+
+        // Disable the channel without overwriting the existing configuration
+        ch.cr().modify(|w| {
+            w.set_en(false);
+        });
+    }
```
etiennecollin, line 359: Are `request_stop()` and `request_pause()` functions flipped? I think the right naming should be:
- `request_pause()` -> set suspend bit `true`

embassy-stm32/src/dma/gpdma/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+
+    fn request_pause(&self) {
+        let info = self.info();
+        let ch = info.dma.ch(info.num);
+
+        // Disable the channel without overwriting the existing configuration
+        ch.cr().modify(|w| {
+            w.set_en(false);
+        });
```
etiennecollin, line 358: This should not work. As per page 742 of the STM32U5 reference manual, **writing 0 into the EN bit is ignored**. We should probably only keep the `request_pause()` function modified to use the suspend bit and add a `request_unpause()` function which unsets the suspend bit.

etiennecollin (commented):
Here are a few things that should probably be fixed.

embassy-stm32/src/dma/gpdma/mod.rs
etiennecollin: I suggest removing the `request_stop()` functions and adding

```rust
    fn request_suspend(&self) {
        let info = self.info();
        let ch = info.dma.ch(info.num);

        ch.cr().modify(|w| w.set_susp(true))
    }

    fn request_resume(&self) {
        let info = self.info();
        let ch = info.dma.ch(info.num);

        ch.cr().modify(|w| w.set_susp(false));
    }
```

elagil:
Thanks for your comments, I will check as soon as I can!

etiennecollin:
Great, thanks! The suggested changes were implemented. You can check them out at https://github.com/elagil/embassy/pull/1.

elagil:
> Great, thanks! The suggested changes were implemented. You can check them out at [elagil#1](https://github.com/elagil/embassy/pull/1).

Thanks, I saw that!

elagil:
Regarding your above feature request, here are my first thoughts.

## Split `configure` and `start`
The GPDMA linked-list table itself must not change its location in memory, after its pointer is configured in the GPDMA linked-list address registers. This was my reason for combining `configure` with `start`, as it makes sure that a user cannot move the table by accident between creating it and starting the GPDMA. I think when I implemented ring-buffer support, I did initially configure the table in the ring-buffers `new()` method, but moving the ring buffer structure could also move the table location in memory. There is probably a better solution, but I do want to avoid a non-obvious GPDMA panic in case the table location is invalid.

## Changing GPDMA config on the fly
For changing configurations, I would probably queue a linked-list item that modifies the related register. Your example `GPDMA_CxTR2` can be modified by a linked-list item. I guess we might add support for queuing config-only linked-list items, which have zero transfer length - or we combine the config changes with the transfers themselves. In any case, adding a custom linked-list item is not currently supported.
Direct configuration changes in GPDMA registers are also possible, but they will not be synchronized to the linked-list transfers. I don't think that this is a safe way to do it.

etiennecollin:
Hi,
Thanks for your feedback.

If I understand correctly, moving the ring-buffers might move the linked-list table in memory and break the pointers used to configure the DMA. That's a really good point and I did not think about that. In that case, you are absolutely right; it makes sense to separate `new()` from the configuration as it allows moving the ring-buffer before configuring and starting the DMA.

---

One thing I thought about is that we could have struct that allows collecting custom register-updater functions:

```rust
struct RegisterUpdaters {
    pub tr1: fn(&mut pac::gpdma::regs::ChTr1),
    pub tr2: fn(&mut pac::gpdma::regs::ChTr2),
    pub tr3: fn(&mut pac::gpdma::regs::ChTr3),
}
impl Default for RegisterUpdaters {
    fn default() -> Self {
        Self {
            tr1: |_| {},
            tr2: |_| {},
            tr3: |_| {},
        }
    }
}
``` 

> The transfer registers 1-3 seem to be the most interesting for the user, but we could expose others. See page 700 of the STM32U5 reference manual.

Then, the user can pass this struct to the `WritableRingBuffer::new()` or `ReadableRingBuffer::new()` functions:

```rust
let ring_buffer = WritableRingBuffer::new(
	channel,
	request,
	peri_addr,
	buffer,
	transfer_options,
	register_updaters
);
```

When we create linked-list items, we can use a ref to that `RegisterUpdaters` struct:

```rust
let items = [
    LinearItem::new_read(request, peri_addr, &mut buffer[..half_len], &register_updaters),
    LinearItem::new_read(request, peri_addr, &mut buffer[half_len..], &register_updaters),
];
let table = Table::new(items);
```

and use the updaters as follows in `LinearItem::new_inner()`, for example to update TR2 register:

```rust
let mut tr2 = regs::ChTr2(0);
tr2.set_dreq(match dir {
    Dir::MemoryToPeripheral => Dreq::DESTINATION_PERIPHERAL,
    Dir::PeripheralToMemory => Dreq::SOURCE_PERIPHERAL,
});
tr2.set_reqsel(request);
(register_functions.tr2)(&mut tr2); // Use the closure here
```

where `(register_functions.tr2)()` could be:

```rust
|w: &mut pac::gpdma::regs::ChTr2| {
	w.set_tcem(pac::gpdma::vals::Tcem::EACH_LINKED_LIST_ITEM);
};
```

> The `RegisterUpdaters::default()` will do nothing to the registers, so it's easy to use for people who do not want to manually touch the registers.

---

This should allow some configuration while preventing the undesired memory moves, as the channel configuration can be moved back to the `start()` function.

Anyways, this is not perfect: this method only allows a unique custom configuration for all list items, but I think that could be fine for now. Let me know what you think.

leftger:
Hi everyone!

I didn't realize my PR (#4565) was a far less sophisticated duplicate of this work. Is this PR ready to be checked? If it is I'd strongly recommend you guys re-request a review by one of the maintainers.

elagil:
Hi @leftger!

Yes, it's ready. I pinged @Dirbaio but he was swamped in work, so this is still pending a review.
Also, I assume that reviewing this takes quite some time. If you had a look at it, I am sure that would help!

i509VCB:
I can try to give it a check (I've been following this pull request for a bit), although I'm not normally comfortable with merging stm32 stuff at least (I usually live in mspm0 and nxp hals)

leftger (approved):
Thank you for putting this PR together. I believe it covers functionality that aligns with BDMA implementations. If there's a tiny little nitpick I might add it might be to rename `ringbuffered.rs` into `ringbuffer.rs`. You can totally choose to ignore it 😆

tyler-gilbert:
Excited to see this merge!

elagil:
Thank you all for having a look! I will rebase again and test on hardware - just to be sure that nothing broke.

elagil:
> rename `ringbuffered.rs` into `ringbuffer.rs`

Sounds reasonable!

elagil:
## After rebasing
- Pending changelog.
- My local USB audio test with GPDMA -> SAI doesn't work anymore. I have to investigate.

leftger:
Could you add an example in the examples folder for whatever family/series you're working off of? (could be a subset of your current USB audio test) I'd be more than happy to port it over to the WBA65 and help you debug the issue.

elagil:
Yes, I will add it. It is already a quite small example.

elagil:
@leftger I added an H5 example. It is quite minimal and runs on 8 MHz SAI clock. I get no activity on either data line or left/right clock. Only the clock output works.

leftger:
> @leftger I added an H5 example. It is quite minimal and runs on 8 MHz SAI clock. I get no activity on either data line or left/right clock. Only the clock output works.

Thanks! I will take a look at it 😄

elagil:
@leftger I found the problem, it was caused by refactoring some time ago... See the fix at https://github.com/embassy-rs/embassy/pull/3923/commits/62b4acb3ffc9be5d0d8dae5cfb94686b02daf02a

The driver only made linked-lists for read-mode (peripheral to memory), even for writes. :roll_eyes:

leftger:
Apologies for not giving you a quicker response. I am a bit swamped this work week. I will check it over the weekend. I'm sure with this new fix it should be super straightforward to assure the maintainers that this is something that should be merged. I'll share a video if I get it to work on my board 😀

elagil:
Great, take your time! The fix is very minor.
I will try to add an example for GPDMA ringbuffer UART as well, I think this will help others to test.

> tiny little nitpick I might add it might be to rename `ringbuffered.rs` into `ringbuffer.rs`

Within the dma module, there is already the common `ringbuffer` being used, so renaming it runs into a conflict.

leftger:
That's fine then please ignore my nitpick lol.

elagil:
Tested UCPD on STM32H5 with the DMA changes, using this example:
https://github.com/elagil/usbpd/tree/900326db415c189f216b4c8307b8dad24ac83d6b/examples/embassy-nucleo-h563zi

Still works as before.

See the opening post for a list of features and completed tests.

elagil:
Thanks for merging :heart_eyes:

etiennecollin:
Great work @elagil! Thank you for merging!

elagil:
> Great work @elagil! Thank you for merging!

Thank you for your help!

PR #3927, opened 2025-02-27
fix: stm32h7rs usb_refck panic message now compiles

friedman-ionq:
With the following in Cargo.toml:
```
embassy_stm32 = {"0.2", features = [   "memory-x",
    "stm32h7s3l8",
    "defmt",
    "time-driver-any",
    "exti",
    "unstable-pac",
] }
```
the crate would not compile.  This fixes the issue.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 
 set -eo pipefail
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
Could you add this feature combination to `ci.sh` so it'd have been caught? thanks~

friedman-ionq:
> Could you add this feature combination to `ci.sh` so it'd have been caught? thanks!

Sure.  Turns out the minimum repeatable feature is just "stm32h7s3l8".

friedman-ionq:
> maybe it's due to `defmt`. When it's enabled the `panic!` macro is overridden, so maybe it works with and fails without.

```
$ ./ci.sh
thread 'main' panicked at src/bin/cargo-batch.rs:136:85:
called `Result::unwrap()` on an `Err` value: the package 'embassy-stm32-tests' does not contain this feature: defmt
```

Dirbaio:
the `embassy-stm32-tests` crate doesn't have it, `embassy-stm32` does.

friedman-ionq:
> the `embassy-stm32-tests` crate doesn't have it, `embassy-stm32` does.

``~/work/embassy (main)$ git diff
diff --git a/ci.sh b/ci.sh
index e32105b16..46354f096 100755
--- a/ci.sh
+++ b/ci.sh
@@ -169,6 +169,7 @@ cargo batch \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u031r8,defmt,exti,time-driver-any,time \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u073mb,defmt,exti,time-driver-any,time \
     --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv6m-none-eabi --features stm32u083rc,defmt,exti,time-driver-any,time \
+    --- build --release --manifest-path embassy-stm32/Cargo.toml --target thumbv7em-none-eabihf --features stm32h7s3l8,defmt,time-driver-any,exti,unstable-pac \
     --- build --release --manifest-path embassy-nxp/Cargo.toml --target thumbv8m.main-none-eabihf \
     --- build --release --manifest-path cyw43/Cargo.toml --target thumbv6m-none-eabi --features ''\
     --- build --release --manifest-path cyw43/Cargo.toml --target thumbv6m-none-eabi --features 'log' \
````
does not repro the bug when I execute `./ci.sh` on main.

Dirbaio:
I cannot reproduce the build failure this is supposed to fix. Could you provide the exact build command to reproduce? What `cargo build --features xxxxx` do I have to run from the `embassy-stm32` dir in a fresh repo clone?

`Hertz` values should always be printable with `{}`, if that's not working then we should fix that, not add `.0` as a workaround.

Dirbaio:
can you try again to see if it fails now on latest `main`?

friedman-ionq:
> can you try again to see if it fails now on latest `main`?

Now, it is not only passing on `main`, but also on `0.2`.  Perhaps it is breaking because of a dependency?

Dirbaio:
no idea! Since it works i'm going to close this, feel free to reopen if you do find the compile issue.

PR #3928, opened 2025-02-27
Fix SetConfig implementation for STM32 I2C v2

### LINKED ISSUE #3914: SetConfig does not work on STM32 I2C
I'm using the [`I2cDeviceWithConfig`](https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/i2c/struct.I2cDeviceWithConfig.html) struct in *embassy-embedded-hal*, but it does not appear to be actually changing the I2C bus frequency. I'm using the bus with the *ssd1306* crate to draw to a small display on an STM32F072CB microcontroller.

If I use a plain `I2cDevice` with the underlying I2C bus configured to 100kHz, it takes about 120ms to write to the device. If I configure the underlying I2C bus to 200kHz, it takes about 60ms (as expected).

However, if I use an `I2cDeviceWithConfig` set to 200kHz on top of an underlying I2C bus configured for 100kHz, the flush takes 120ms still. It looks like the configuration override isn't happening at all.

I'm using *embassy-embedded-hal* 0.3.0 and *embassy-stm32* 0.2.0. It looks like the [implementation of this trait](https://github.com/embassy-rs/embassy/blob/embassy-stm32-v0.2.0/embassy-stm32/src/i2c/v2.rs#L803-L817) has not changed between 0.2.0 and the latest git.

The embassy docs link to v2 of the I2C peripheral for my MCU, but I must admit I don't know if the STM32F072CB uses v1 or v2.

agrif:
The I2C peripheral needs to be disabled to change the timings register. This PR fixes the `SetConfig` implementation to disable the peripheral, set the timings, then re-enable it.

I took a glance at the v1 I2C implementation to see if this might be easy to add there, but that peripheral is sufficiently different that I did not feel comfortable changing it without a device to test on, which I don't have.

Fixes #3914.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -855,13 +855,22 @@ impl<'d, M: Mode> SetConfig for I2c<'d, M> {
     type Config = Hertz;
     type ConfigError = ();
     fn set_config(&mut self, config: &Self::Config) -> Result<(), ()> {
+        self.info.regs.cr1().modify(|reg| {
+            reg.set_pe(false);
+        });
+
         let timings = Timings::new(self.kernel_clock, *config);
+
         self.info.regs.timingr().write(|reg| {
             reg.set_presc(timings.prescale);
             reg.set_scll(timings.scll);
             reg.set_sclh(timings.sclh);
             reg.set_sdadel(timings.sdadel);
             reg.set_scldel(timings.scldel);
+        });
+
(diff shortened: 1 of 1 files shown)


mkj (approved):
Looks good. 

Other methods require `&mut self` so I don't think there's any concern about clearing PE affecting other calls.

Dirbaio (approved):

PR #3930, opened 2025-02-28
Fix STM32 UART: Add separate TX waker and only clear idle flag in RingBufferedUartRx

meowcakes:
This resolves an issue whereby one is unable to use a `UartTx` and a `RingBufferedUartRx` in parallel in separate tasks.

Firstly, it changes `clear_idle_flag` to ONLY clear the idle flag as the name suggests it should. Currently, it clears ALL set flags including TC which is checked by `UartTx::flush`. If the TX and RX are being used in parallel, it is possible for the RX to clear the TC flag before the TX can check it, which causes `UartTx::flush` to hang forever.

Secondly, it adds a separate `tx_waker`, otherwise when TX and RX are used in parallel they contend for the currently lone `rx_waker` and can overwrite each others wakers, causing hanging.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -85,6 +85,7 @@ unsafe fn on_interrupt(r: Regs, s: &'static State) {
 
     compiler_fence(Ordering::SeqCst);
     s.rx_waker.wake();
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thanks for the PR, looks good to me.

PR #3935, opened 2025-03-01
embassy-rp: pio: Add access to DMA engine byte swapping

edenbarby:
Changes the dma_pull/push function signatures to give callers access to the byte swap flag for their DMA transfers.


lulf (approved):
LGTM, thanks!

PR #3943, opened 2025-03-03
Update README.md to include a reference to mpfs-hal

AlexCharlton:
Hello, I've been working on a [HAL for the Microchip PolarFire SoC (MPFS](https://github.com/AlexCharlton/mpfs-hal)), and it includes a crate with Embassy support.

While it is not yet complete, I think it has enough features to be useful at this point, so I figured referencing it in the README would be helpful.


lulf:
Could you add this to https://github.com/embassy-rs/embassy/blob/main/docs/pages/overview.adoc as well?

AlexCharlton:
Good catch. Added to `hal.adoc` as well.

PR #3950, opened 2025-03-06
Add ability to create STM32 timer channels with full pin configuration

Matt-Allen-Bose:
Currently there is no way to specify either timer or lptimer pwm channel pin configuration. This PR adds the ability for specifying the pins output type, pull, and speed depending on the gpio version.


Dirbaio (approved):
thanks!

Matt-Allen-Bose:
> thanks!

No problem. Thanks for all the hard work. The embassy project really is excellent.

PR #3951, opened 2025-03-06
Incorrect bit size for FLASH_CR PNB for g4c4

dgehriger:
See https://github.com/embassy-rs/stm32-data/pull/575:

The bitsize for the number of flash pages (PNB) in the FLASH_CR is incorrect for category 4 devices (STM32G491, STM32G4A1).

As a result, erasing a page in bank 2 in single-bank mode will incorrectly erase the corresponding bank 1 page.

image


dgehriger:
Wrong target

PR #3952, opened 2025-03-07
fix(stm32): handle half-duplex in ringbuffered read

badrbouslikhin:
This PR fixes an issue introduced in PR #3379, where the transmitter disable/receiver enable operation in half-duplex mode was moved from the flush method to the read methods. However, that PR did not account for the ringbuffered read method.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -716,7 +716,6 @@ impl<'d> UartRx<'d, Async> {
 
         // make sure USART state is restored to neutral state when this future is dropped
         let on_drop = OnDrop::new(move || {
-            // defmt::trace!("Clear all USART interrupts and DMA Read Request");
             // clear all interrupts and DMA Rx Request
             r.cr1().modify(|w| {
                 // disable RXNE interrupt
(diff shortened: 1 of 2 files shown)


lulf (approved):

Dirbaio:
@badrbouslikhin it seems this is causing the `stm32l152re/usart_rx_ringbuffered` test to fail

badrbouslikhin:
I checked the reference manual (RM0038) and found no obvious reason for the test failure.  
Also, the test does not appear to use a half-duplex USART, so the following condition should fail: `if r.cr3().read().hdsel() && r.cr1().read().te()`.

I am not sure what's going on here, I'll check if I have a board stm32l1xxx board I can test at the office.

badrbouslikhin:
I couldn't find a stm32l1xxx board. I can't debug this further at the moment.

Are you sure it's not an issue with the physical setup of that board? I checked recent PRs, HIL tests didn't run and thus didn't trigger any issue.

Dirbaio:
okay I think the tes tis just flaky. It already randomly fails, this PR just makes it fail more often. Let's just ignore it.

Had to create a new PR #4054 . The next time please enable the "allow maintainers to push" checkbox when creating the PR.

PR #3953, opened 2025-03-07
Add `Instant::try_from_*` constructor functions

Abestanis:
When using `Instant::from_*` (except `from_ticks`), it is possible to overflow on the multiplication, which can cause a panic if debug checks are enabled, which procudes a panic like this:
```
[...]\embassy-time-0.4.0\src\instant.rs:41:20:
attempt to multiply with overflow
```

I've just hit this and for now the only workaround is to do some bounds on the input value checks before passing it to the `Instant::from_*` constructor, but it is a bit cumbersome to find the maximum allowed value. For this reason I'm proposing to add these helper constructors that use `checked_mul` and can fail gracefully on overflow.

### DIFF
--- a/embassy-time/src/duration.rs
+++ b/embassy-time/src/duration.rs
@@ -64,9 +64,9 @@ impl Duration {
 
     /// Creates a duration from the specified number of nanoseconds, rounding up.
     /// NOTE: Delays this small may be inaccurate.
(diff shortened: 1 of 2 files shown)


Dirbaio:
looks good to me! could you add the same to Duration for consitency? thanks

Abestanis:
Good idea, done.

lulf (approved):

PR #3955, opened 2025-03-09
rp: embassy-rp 0.4.0 release prep

9names:
A new embassy-rp release should be ready to ship after these changes.

No code changes, just bumping crate numbers and updating changelogs.
I also updated cyw43-pio and embassy-boot-rp as they depends on the hal version, and I wanted to ensure CI was green


9names:
Okay, maybe this will need to wait until embassy-boot-rp is updated to pass CI :)

lulf (approved):
Thanks!

PR #3964, opened 2025-03-12
Add optional USB driver SOF output

elagil:


embassy-stm32/src/usb/usb.rs
```diff
@@ -287,6 +287,26 @@ pub struct Driver<'d, T: Instance> {
...
+    /// Create a new USB driver with start-of-frame (SOF) output.
+    pub fn new_with_sof(
+        _usb: impl Peripheral<P = T> + 'd,
+        _irq: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
+        dp: impl Peripheral<P = impl DpPin<T>> + 'd,
+        dm: impl Peripheral<P = impl DmPin<T>> + 'd,
+        sof: impl Peripheral<P = impl SofPin<T>> + 'd,
+    ) -> Self {
+        into_ref!(sof);
+        #[cfg(not(stm32l1))]
```
Dirbaio, line 299: what's wrong with L1? 

if SOF output is not supported on L1 for some reason it'd be better to cfg the entire `new_with_sof` method, instead of having it silently not work.

elagil: To be honest, I don't know. I copied how it was done in `new` already:

```rust
        #[cfg(not(stm32l1))]
        {
            use crate::gpio::{AfType, OutputType, Speed};
            dp.set_as_af(dp.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
            dm.set_as_af(dm.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
        }
        #[cfg(stm32l1)]
        let _ = (dp, dm); // suppress "unused" warnings.
```

L1 definitely has a USB peripheral, so I don't understand the existing AF handling.

PR #3965, opened 2025-03-13
Stm32 usart: implement pin config for cts, tx, rts and de

toon23:
Allows pin configuration for cts, tx, rts and de pin in usart::Config. The configuration is applied in all Uart drivers.

Also I removed the now rendundant HalfDuplexConfig which I used as base for the pin config implementation.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -138,6 +138,40 @@ pub enum HalfDuplexReadback {
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Half duplex IO mode
+pub enum OutputConfig {
+    /// Push pull allows for faster baudrates, no internal pullup
+    PushPull,
+    #[cfg(not(gpio_v1))]
+    /// Push pull with internal pull up resistor
+    PushPullPullUp,
+    #[cfg(not(gpio_v1))]
+    /// Push pull with internal pull down resistor
+    PushPullPullDown,
+    /// Open drain output using external pull down resistor
+    OpenDrainExternal,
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/usart/mod.rs
```diff
@@ -136,6 +136,40 @@ pub enum HalfDuplexReadback {
...
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Half duplex IO mode
+pub enum OutputConfig {
+    /// Push pull allows for faster baudrates, no internal pullup
+    PushPull,
+    #[cfg(not(gpio_v1))]
+    /// Push pull with internal pull up resistor
+    PushPullPullUp,
```
Dirbaio, line 147: PushPullPullUp and PushPullPullDown make no sense. If the MCU is always driving the pin the pull resistor has no effect (other than wasting current).

embassy-stm32/src/usart/mod.rs
```diff
@@ -136,6 +136,40 @@ pub enum HalfDuplexReadback {
...
+    #[cfg(not(gpio_v1))]
+    /// Push pull with internal pull down resistor
+    PushPullPullDown,
+    /// Open drain output using external pull down resistor
+    OpenDrainExternal,
+    #[cfg(not(gpio_v1))]
+    /// Open drain output using internal pull up resistor
+    OpenDrainInternal,
```
Dirbaio, line 155: I'd call these OpenDrain and OpenDrainPullUp. This is more consistent with the GPIO driver. It doesn't concern what's on the board, just what the MCU itself is doing (enable pullup, or not enable pullup)

toon23:
Thanks for the review @Dirbaio.  You are right in both cases, I have changed the code accordingly.

PR #3966, opened 2025-03-14
Embassy for MSPM0

i509VCB:
This adds an embassy hal for the Texas Instruments MSPM0 microcontroller series.

So far the GPIO and time drivers have been implemented. I have tested these drivers on the following parts:
- C1104
- L1306
- L2228
- G3507
- G3519

Each of these parts also has two examples: blink and button.

The PAC is generated at https://github.com/mspm0-rs. The PAC is largely based off IDE metadata and the C SDK. The PAC is still work in progress, but that is expected. Otherwise the PAC is largely inspired by stm32-metapac.

Regarding maintenance, I can be the maintainer for the mspm0 hal.

Interestingly the MSPM0C1104 has only 1KB of ram and appears to be able to run some examples. It would be interesting to use that target in the future for optimizing other parts of the embassy ecosystem for code side.

Regarding HAL name (it not being `embassy-ti`), the hardware in other TI ARM chips such as the MSP432, CC13xx/CC26xx is quite different. Those chips also have a different demand for metadata so it would be challenge to adapt for all the parts. As such I made MSPM0 support an individual crate.


examples/mspm0g3519/MSPM0GX51X_Series.yaml
```diff
@@ -0,0 +1,424 @@
+name: MSPM0GX51X Series
+manufacturer:
+  id: 0x17
+  cc: 0x0
+generated_from_pack: true
+pack_file_release: 1.0.0
+variants:
+- name: MSPM0G1518
```
lulf, line 8: This file should probably not be committed

examples/mspm0l2228/MSPM0L122X_L222X_Series.yaml
```diff
@@ -0,0 +1,239 @@
+name: MSPM0L122X_L222X Series
```
lulf, line 1: This file should should not be committed.

lulf (commented):
Thanks, this looks awesome! There are some files I think was not supposed to be committed in here (see comments).

Also, there's a good bunch of todo's, and some cases where there's only todo!()'s in a match statement. Is the plan to fix them, or is it better to leave them out until implemented?

Otherwise this looks consistent with other HALs, so thanks for that!

felipebalbi:
Maybe patch `ci.sh` to run build tests for MSPM0 too?

examples/mspm0g3519/MSPM0GX51X_Series.yaml
i509VCB: The target files are temporary until probe-rs releases a new version after which I'd remove those files.

I can remove these of course, but then you'd have to build probe-rs from git or convince probe-rs to make a new release.

i509VCB:
> Also, there's a good bunch of todo's, and some cases where there's only todo!()'s in a match statement. Is the plan to fix them, or is it better to leave them out until implemented?

The group 0 interrupts are not enabled at the moment so it's probably find to leave it like that. On chips with a group 1 interrupt, the peripherals which CPU raise the other interrupts are not enabled yet.

i509VCB:
I have added the chips with examples to the CI.

Not every chip is run in CI yet because unfortunately IDE metadata does not specify fully interrupt groups, it has to be manually entered from datasheet. So far I have done this for the chips with examples (and boards which I have): https://github.com/mspm0-rs/mspm0-data/tree/master/data/int_group

I'm not sure why C1104 fails to build on CI due to stack issues, but it works fine on my environment (aarch64 Linux on Rust 1.84.1)

lulf (commented):
Looks good! Could you also add the HAL to the following docs:

* https://github.com/embassy-rs/embassy/blob/main/docs/pages/overview.adoc#hardware-abstraction-layers
* https://github.com/embassy-rs/embassy/blob/main/README.md#batteries-included

For the CI build, it uses 1.85. Does it reproduce if you run `ci.sh`?

i509VCB:
Replicating the commands from CI with:

(+nightly isn't necessarily there, but it is needed for manifest path)
```
cargo +nightly build -Z unstable-options --release --manifest-path examples/mspm0c1104/Cargo.toml --target thumbv6m-none-eabi --artifact-dir target/examples/mspm0c1104
```

That does result in a linker error. However a regular run, so:
```
cd examples/mspm0c1104
cargo build
```

does work when using Rust 1.85.

i509VCB:
Huh apparently cargo batch is not considering my setting `DEFMT_RTT_BUFFER_SIZE` in `examples/mspm0c1104/.cargo/config.toml`

Explictly setting that does fix the issue.

lulf (approved):
Awesome!

Issue #2793, opened 2024-04-09
Question: Possibility to determine task name from ptr


Murmele:
When using `rtos-trace` as id the task pointer as u32 is used. How do I find out, which task it is?

PR #3969, opened 2025-03-14
Add possibility to get the id of a task to be able to use it with rtos-trace

Murmele:
This can be used in combination with rtos-trace to match task names with id's
fixes: https://github.com/embassy-rs/embassy/issues/2793

### DIFF
--- a/embassy-executor/src/spawner.rs
+++ b/embassy-executor/src/spawner.rs
@@ -34,6 +34,15 @@ impl<S> SpawnToken<S> {
         }
     }
 
+    /// Returns the task if available, otherwise 0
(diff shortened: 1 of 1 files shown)


embassy-executor/src/spawner.rs
```diff
@@ -34,6 +34,15 @@ impl<S> SpawnToken<S> {
         }
     }
 
+    /// Returns the task if available, otherwise 0
```
Murmele, line 37:
```suggestion
    /// Returns the task id if available, otherwise 0
```

Murmele:
Thanks for merging!

PR #3972, opened 2025-03-19
embassy-net-esp-hosted: Add basic softap and scanning support 

mjwells2002:
Currently a draft because of the modifications to proto.rs, unsure if directly modifying this is correct


mjwells2002:
Not sure if this is the best way to do this but it has all the features i wanted to add, and it works

embassy-net-esp-hosted/src/control.rs
```diff
@@ -160,7 +257,8 @@ impl<'a> Control<'a> {
     async fn ioctl(&mut self, msg: &mut CtrlMsg) -> Result<(), Error> {
         debug!("ioctl req: {:?}", &msg);
 
-        let mut buf = [0u8; 128];
+        //Network scan result is 67 bytes max of 16 can be returned
+        let mut buf = [0u8; 1072];
```
lulf, line 261: I'm wondering if this should be passed in to the ioctl... it seems that going from 128 -> 1072 would increase RAM usage quite a bit.

lulf (commented):
General approach looking good, just a question about the ioctl buf.

embassy-net-esp-hosted/src/control.rs
mjwells2002: yes this probably should be, its only networking scanning that requires so much memory, also turns out the max of 16 i assumed here is wrong, i ran into a situation where i get a runtime panic for running out of memory on the ioctl buffer

mjwells2002:
took a while to get back to this but managed to today, i updated the ioctl macro to take in the length of the buffer  and set the size increase for only scanning, also changed it to prevent a runtime panic if more scan results are returned then can be handled, im not quite sure the correct way to deal with that situation, so its just ignored for now

bugadani:
Scanning has been implemented separately in #6344.

Dirbaio:
Hello,

PR #3975, opened 2025-03-19
[RP235x] Fix wrong funcsel on RP2350 GPOUT/GPIN

JuliDi:
The FUNCSEL value for GPOUT/GPIN has changed from RP2340 to RP2350. Embassy has been using 0x08 for both chips, therefore the GPOUT example did not work on the Pico 2.
This PR now distinguishes the two chips and sets FUNCSEL to 0x09 for the RP2350.

See RP2340 (left) and RP2350 (right) datasheets:
RP_funcsel


CBJamo:
Nice catch, LGTM.

PR #3976, opened 2025-03-19
define stm32h7rs ethernet pin traits

friedman-ionq:
In order to fit in the internal flash, needed to disable dns and dhcp as well as switch from TCP to UDP.


Dirbaio:
the issue here is the pin function names for h7rs are inconsistent with other chips (they have a redundant `MII_`/`RMII_` prefix. This should be fixed in stm32-data, making the names consistent.

friedman-ionq:
I looked at stm32-data, and it wasn't obvious what to change.  I'll leave the changes to the examples directory as a test case for whoever updates stm32-data, but I'm not going to work this issue further.

Dirbaio:
can you rebase to get CI to pass?

Dirbaio:
thanks~

PR #3981, opened 2025-03-20
update stm32-data, add C0 adc, add OTP memory regions.

Dirbaio:
- **Update stm32-data.**
- **Added ADC support for STM32C0.**
- **feat(stm32): Add OTP flash region (again)**
- **Fix compile error in ospi**


Dirbaio:
bender run

PR #3982, opened 2025-03-20
embassy-net: Add std and alloc features

mkj:
This was necessary when I enabled `std` feature for `managed` used directly in my own project.
Not certain if this is the best way or there is some other approach I could take.
(including `smoltcp` on my own crate would work, but then it needs matching versions etc)

---

These are passed through to smoltcp, which otherwise won't handle some match cases when managed/std feature is enabled externally.

```
  --> .../index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/iface/socket_set.rs:82:15
   |
82 |         match &mut self.sockets {
   |               ^^^^^^^^^^^^^^^^^ pattern `&mut ManagedSlice::Owned(_)` not covered
```


mkj:
On further thought, maybe it'd be better to just fix it in smoltcp, I'll do that.

lulf (approved):

lulf:
Thanks!

Issue #1304, opened 2023-03-28
BufferedUart split in embarry_rp returns tx/rx in reverse order than Uart


uzytkownik:
BufferedUart returns it in rx/tx order while Uart returns it in tx/rx oder:

```rust

    /// Split the Uart into a transmitter and receiver, which is particuarly
    /// useful when having two tasks correlating to transmitting and receiving.
    pub fn split(self) -> (UartTx<'d, T, M>, UartRx<'d, T, M>) {
        (self.tx, self.rx)
    }

    pub fn split(self) -> (BufferedUartRx<'d, T>, BufferedUartTx<'d, T>) {
        (self.rx, self.tx)
    }
```

It should probably be consistent.

wetheredge:
Looks like this was fixed in https://github.com/embassy-rs/embassy/commit/f733071908ced05bc006b43e1383336290faafdc

wetheredge:
[`BufferedUart::new`](https://github.com/embassy-rs/embassy/blob/f733071908ced05bc006b43e1383336290faafdc/embassy-rp/src/uart/buffered.rs#L93-L96) and [`BufferedUart::new_with_rtscts`](https://github.com/embassy-rs/embassy/blob/f733071908ced05bc006b43e1383336290faafdc/embassy-rp/src/uart/buffered.rs#L114-L119) still take interrupts before pins, while [`Uart::new`](https://github.com/embassy-rs/embassy/blob/f733071908ced05bc006b43e1383336290faafdc/embassy-rp/src/uart/mod.rs#L705-L708) and [`Uart::new_with_rtscts`](https://github.com/embassy-rs/embassy/blob/f733071908ced05bc006b43e1383336290faafdc/embassy-rp/src/uart/mod.rs#L729-L734) take pins then interrupts. Seems like pins then interrupts is more common in other peripherals.

JonathanPlasse:
Would a PR be welcome to take pins before interrupts on `BufferedUart`?

Dirbaio:
yes~

mgomez0:
Is this as trivial as it seems? I'm new to open source and the project in general. Also not sure what the scope of testing at my bench should look like, or if CI and inspection is enough here.

mgomez0:
@Dirbaio I think this can be closed now.

PR #3983, opened 2025-03-20
BufferedUart initialization

mgomez0:
Modifies BufferedUart new() and new_with_rtscts() to take pins before interrupts at initialization.
Related to #1304.


lulf (commented):
Looks good, though we should probably also ensure so that embassy-nrf and embassy-stm32 are consistent with embassy-rp. Would you like to make modifications to them as well while you're at it?

lulf:
Thanks! Could you rebase this please?

mgomez0:
@lulf good to go.

lulf (approved):
Thanks!

lulf:
Could you also squash these commits? 

I guess this will break the API for lots of people, so it's worth adding a note to the changelogs for each HAL as well I think.

mgomez0:
@lulf squashed commits and added notes to changelogs.

lulf:
Thanks!

PR #3992, opened 2025-03-22
Small USB-related fixes

elagil:
- Fixes errata, found for STM32H5: "During OUT transfers, the correct transfer interrupt (CTR) is triggered a little before the last USB SRAM accesses have completed. If the software responds quickly to the interrupt, the full buffer contents may not be correct."
A delay of 800 ns is recommended as a workaround.
- Fixes default mute state of the UAC1 class. For hosts that don't use the control interface (e.g. Android) there is no way to get the device out of mute.

### DIFF
--- a/embassy-stm32/src/usb/usb.rs
+++ b/embassy-stm32/src/usb/usb.rs
@@ -7,6 +7,7 @@ use core::task::Poll;
 
 use embassy_hal_internal::into_ref;
 use embassy_sync::waitqueue::AtomicWaker;
+use embassy_time::Timer;
 use embassy_usb_driver as driver;
 use embassy_usb_driver::{
     Direction, EndpointAddress, EndpointAllocError, EndpointError, EndpointInfo, EndpointType, Event, Unsupported,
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/usb/usb.rs
```diff
@@ -887,6 +887,16 @@ impl<'d, T: Instance> driver::EndpointOut for Endpoint<'d, T, Out> {
...
 
+        // Errata for STM32H5, 2.20.1:
+        // During OUT transfers, the correct transfer interrupt (CTR) is triggered a little before the last USB SRAM accesses
+        // have completed. If the software responds quickly to the interrupt, the full buffer contents may not be correct.
+        //
+        // Workaround:
+        // Software should ensure that a small delay is included before accessing the SRAM contents. This delay should be
+        // 800 ns in Full Speed mode and 6.4 μs in Low Speed mode.
+        #[cfg(stm32h5)]
+        embassy_time::Timer::after_nanos(800).await;
```
Dirbaio, line 898: 800ns is such a short delay that i'd make it blocking. It'll have better perf.

Issue #3997, opened 2025-03-24
Peripheral trait unsound: Mutexes can clone peripherals in safe code.


agrif:
I'm pretty sure I found a soundness issue with the Peripheral trait. Mutex guards implement DerefMut, and so guards that contain a peripheral implement the Peripheral trait. You can then get a PeripheralRef out of the guard, which drops the lock on the mutex, allowing you to borrow it again immediately.

`````rust
static PERI_MUTEX: Mutex<Option<PERI>> = Mutex::new(None);

fn main() {
    // initialize our peripheral mutex, e.g. via hal::init()
    {
        *PERI_MUTEX.lock().unwrap() = Some(PERI);
    }

    // MutexGuards implement DerefMut, and therefore Peripheral
    // but turning it into a PeripheralRef drops the guard...
    let ref1: PeripheralRef<'static, PERI> =
        { MutexGuard::map(PERI_MUTEX.lock().unwrap(), |o| o.as_mut().unwrap()).into_ref() };

    // so we can just borrow it again immediately
    let ref2: PeripheralRef<'static, PERI> =
        { MutexGuard::map(PERI_MUTEX.lock().unwrap(), |o| o.as_mut().unwrap()).into_ref() };

    println!("ref1 {:?} ref2 {:?}", *ref1, *ref2);
}
`````

[(playground with supporting code from embassy_hal_internal::peripheral)](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=2c9e444a1064133e391db8e7648eb27e)

My understanding is that holding two active `PeripheralRef<'static, PERI>` tokens for the same peripheral is a no-no, and breaks the assumption that only one thing has mutable access to a given peripheral at a time.

The most obvious fix would be to restrict the Peripheral trait implementation specifically to `&mut P` rather than `DerefMut<Target = P>`, but I don't know how much damage that would cause or what other legitimate uses for the DerefMut implementation there are.

Morgane55440:
i would like to add that not only does `MutexGuard` break here, but basically any type which has any Drop glue is likely to break this.

for example one might want to make a wrapper that holds a peripheral, and perform cleanup on drop. this would cause the cleanup to happen whenever `into_ref` is called

using [`core::mem::forget`](https://doc.rust-lang.org/nightly/core/mem/fn.forget.html) has a decent chance to fix the issue, but seems likely make things quite complicated for end users.

i would recommend just making it for `&mut P`

[...]

PR #3999, opened 2025-03-24
Remove Peripheral trait, rename PeripheralRef->Peri.

Dirbaio:


agrif:
Linking back to #3997.

Is there a reason why you can't just change the Peripheral implementation to `&mut P` specifically? That seems like a smaller change overall. Then mark the trait itself as unsafe and document that implementors need to be careful about soundness for some types (like MutexGuard).

Dirbaio:
The motivation for this is not just fixing the soundness issue. I've never been 100% happy with the design of `Peripheral`. Currently HAL drivers take `Impl Peripheral<P = impl UsartPin> + 'd`, this is confusing to the user:

- when you see it in rustdoc you have to look up what it is, to learn what can you pass.
- the main point of it was to allow passing `&mut p.USART1`, but it turns out it's not discoverable at all. People don't figure it out on their own, because it's "hidden" by the `impl Peripheral for T where T: DerefMut` which is gnarly rust generics garbage
- it has _nested `impl Trait`s_ 😱

With this new design, the `Peripherals` struct has `Peri<...>` and HAL drivers take `Peri<...>`. It's much more straightforward. The only downside is it's a bit more verbose.

agrif:
I see, thank you for the explanation.

docs/pages/layer_by_layer.adoc
```diff
@@ -4,7 +4,7 @@ If you're new to Embassy, it can be overwhelming to grasp all the terminology an
 
 This guide uses the STM32 IOT01A board, but should be easy to translate to any STM32 chip. For nRF, the PAC itself is not maintained within the Embassy project, but the concepts and the layers are similar.
 
-The application we'll write is a simple 'push button, blink led' application, which is great for illustrating input and output handling for each of the examples we'll go through. We'll start at the Peripheral Access Crate (PAC) example and end at the async example.
+The application we'll write is a simple 'push button, blink led' application, which is great for illustrating input and output handling for each of the examples we'll go through. We'll start at the Peri Access Crate (PAC) example and end at the async example.
```
i509VCB, line 7: Regex find and replace oops

lulf (approved):
Looks good to me, the only thing I personally dislike is the Peripheral -> Peri rename (which I can totally live with!), which might not be super obvious unless you're used to embassy and embedded terminology (Is it a periscope? :laughing:) .

embassy-nrf/src/qdec.rs
```diff
@@ -252,7 +250,7 @@ pub enum LedPolarity {
     ActiveLow,
 }
 
-/// Peripheral static state
+/// Peri static state
```
timokroeger, line 253: another regex oops

embassy-nxp/src/gpio.rs
```diff
@@ -218,9 +216,9 @@ pub(crate) trait SealedPin: Sized {
 /// [AnyPin]. By default, this trait is sealed and cannot be implemented outside of the
 /// `embassy-nxp` crate due to the [SealedPin] trait.
 #[allow(private_bounds)]
-pub trait Pin: Peripheral<P = Self> + Into<AnyPin> + SealedPin + Sized + 'static {
-    /// Degrade to a generic pin struct
-    fn degrade(self) -> AnyPin {
+pub trait Pin: PeripheralType + Into<AnyPin> + SealedPin + Sized + 'static {
+    ///.into() to a generic pin struct
```
timokroeger, line 220: another regex oops

embassy-rp/src/clocks.rs
```diff
@@ -51,7 +50,7 @@ static CLOCKS: Clocks = Clocks {
     rtc: AtomicU16::new(0),
 };
 
-/// Peripheral clock sources.
+/// Peri clock sources.
```
timokroeger, line 53: another regex oops

embassy-rp/src/clocks.rs
```diff
@@ -81,7 +80,7 @@ pub struct ClockConfig {
     pub ref_clk: RefClkConfig,
     /// System clock configuration.
     pub sys_clk: SysClkConfig,
-    /// Peripheral clock source configuration.
+    /// Peri clock source configuration.
```
timokroeger, line 83: another regex oops

embassy-rp/src/clocks.rs
```diff
@@ -681,7 +680,7 @@ pub(crate) unsafe fn init(config: ClockConfig) {
         peris.set_hstx(false);
     }
 
-    // Peripheral clocks should now all be running
+    // Peri clocks should now all be running
```
timokroeger, line 683: another regex oops

embassy-rp/src/clocks.rs
```diff
@@ -753,7 +752,7 @@ pub fn clk_ref_freq() -> u32 {
     CLOCKS.reference.load(Ordering::Relaxed)
 }
 
-/// Peripheral clock frequency.
+/// Peri clock frequency.
```
timokroeger, line 755: another regex oops

embassy-rp/src/gpio.rs
```diff
@@ -919,9 +917,9 @@ pub(crate) trait SealedPin: Sized {
 
 /// Interface for a Pin that can be configured by an [Input] or [Output] driver, or converted to an [AnyPin].
 #[allow(private_bounds)]
-pub trait Pin: Peripheral<P = Self> + Into<AnyPin> + SealedPin + Sized + 'static {
-    /// Degrade to a generic pin struct
-    fn degrade(self) -> AnyPin {
+pub trait Pin: PeripheralType + Into<AnyPin> + SealedPin + Sized + 'static {
+    ///.into() to a generic pin struct
```
timokroeger, line 921: another regex oops

timokroeger (commented):
Regex replaced a lot of stuff in comments. This is just what I found when quickly going over the changes.

bugadani:
Just something I noticed and I'm not sure if you're just not there yet or forgot: `AnyPin::steal` exists, returns `AnyPin` and `AnyPin` is now `Copy`.

Dirbaio:
> AnyPin is now Copy.

yes, this is intended. AnyPin (and all the other peripheral types) aren't singletons anymore, they're just "info on which peripheral it is", just plain old data. The actual owned thing is `Peri`. You can't use a bare `AnyPin`, you have to do `unsafe { Peri::new_unchecked(AnyPin::steal(4)) }`. It's `new_unchecked` where you declare that you know that pin isn't being used anywhere.

Given this, `steal` doesn't need to be unsafe, hm.. Also perhaps it's not the best name. And it's definitely inconsistent with how steal works with the regular peripherals, so at least this needs fixing.

Which option is best? Make all steal methods return Peri?

```rust
let a /* : Peri<'static, PA5>*/ = PA5::steal();
let b /* : Peri<'static, AnyPin>*/= AnyPin::steal(5);
```

or not have any steal methods, make the user use Peri::new_unchecked?

```rust
let a /* : Peri<'static, PA5>*/ = Peri::new_unchecked(PA5); // PA5 could be a unit struct.
let b /* : Peri<'static, AnyPin>*/= Peri::new_unchecked(AnyPin::new(5));
```

alexmoon:
> Looks good to me, the only thing I personally dislike is the Peripheral -> Peri rename (which I can totally live with!), which might not be super obvious unless you're used to embassy and embedded terminology (Is it a periscope? 😆) .

Would `Periph` be a better abbreviation?

alexmoon:
> or not have any steal methods, make the user use Peri::new_unchecked?

This option seems better to me. You could also spell `Peri::new_unchecked` as `Peri::steal`

cschuhen:
Decided to test the current state of this PR. No code user changes required. All seems to be working fine: FDCAN, SPI, I2C, Exti, GPIO.  on STM32G431.

Dirbaio:
After some discussion in Matrix, we're finally going with option 1 "Make all steal methods return Peri". Advantages are it's a bit simpler, avoids the user having to handle bare `PA5`, `AnyPin` types, allows marking `Peri::new_unchecked` as `doc(hidden)` because users never need it, requires no changes to existing code.

Thank you guys for noticing the regex oopses and for testing :)

mwerezak:
Yeah the name is unfortunate because Peri is a brand of hot sauce here and I can never not think about it when I read that name.

PR #4001, opened 2025-03-24
Start adding support for MIMXRT600 series

felipebalbi:
Still need to add relevant drivers, but at least we can compile using the various PACs.


felipebalbi:
@i509VCB, FYI. Something like this could be done for your work on iMXRT1011. What do you think? After porting GPIO over to all platforms, maybe we look into integrating the other SVDs into `imxrt-ral`?

embassy-nxp/src/pac_utils.rs
```diff
@@ -4,44 +4,44 @@
 /// Due to the type system of peripherals, access to the settings of a single pin is possible only
 /// by a single thread at a time. Read/Write operations on a single registers are NOT atomic. You
 /// must ensure that the GPIO registers are not accessed concurrently by multiple threads.
-pub(crate) fn gpio_reg() -> &'static lpc55_pac::gpio::RegisterBlock {
-    unsafe { &*lpc55_pac::GPIO::ptr() }
+pub(crate) fn gpio_reg() -> &'static crate::pac::gpio::RegisterBlock {
+    unsafe { &*crate::pac::GPIO::ptr() }
 }
```
i509VCB, line 9: I gave both reference manuals a skim and it appears that the GPIO peripheral on the RT1010 and RT600 are different. So this whole file is probably just `lpc55_rt600` stuff. It's probably going to end up being cfg spam anyways because lpc55 exists.

Let alone that imxrt-ral has a different API structure here.

Namely the RT1010 has no PINT driver.

Really this is just a note because this is not enabled for the rt600 anyways.

i509VCB (commented):
This seems reasonable so far. There are some slight peripheral differences like you mentioned but those can be handled at the HAL level.

Maybe you should hop in the imxrt matrix chat regarding migrating the SVDs to ral in the near future.

embassy-nxp/src/pac_utils.rs
felipebalbi: I think for this we could introduce `gpio/mod.rs` and equivalent `gpio/v1.rs`, `gpio/v2.rs`, ..., `gpio/vN.rs` for each variant whenever applicable.

Ultimately, having all families supported on the RAL would be much, much better.

i509VCB:
Regarding drivers did you intend to add them all here in this pull request? I would probably advise doing each in a step by pull request.

GPIO would probably make sense to do first in this pull request.

felipebalbi:
After discussing with @korbin and @i509VCB on embassy matrix, abandoning this PR.

PR #4013, opened 2025-03-26
rp: expose critical section internals

wowkster:
This PR exposes an API for working with the critical_section implementation in embassy-rp. This is useful in the context of panic handlers because if a core panics from within a critical section, there is currently no mechanism to release it manually.

This commit also introduces a `CoreId` enum, similar to the one from embassy-stm32. This allows user code to query which core currently holds the critical section; a necessary function for safely calling `manually_release` from within a panic handler.

### DIFF
--- a/embassy-rp/src/critical_section_impl.rs
+++ b/embassy-rp/src/critical_section_impl.rs
@@ -1,6 +1,8 @@
+//! Implementation of critical_section which is safe to use from both cores
+
 use core::sync::atomic::{AtomicU8, Ordering};
 
-use crate::pac;
+use crate::{multicore::CoreId, pac};
 
 struct RpSpinlockCs;
 critical_section::set_impl!(RpSpinlockCs);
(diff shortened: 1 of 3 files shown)


wowkster:
Previously the `embassy_rp::critical_section_impl` module was private, but the Spinlock struct was marked as pub. Do we want to expose this as well? If so, we should make sure to document the behavior cased by `RP2350-E2`.

wowkster:
I think there are legitimate uses for this, although yes you do need to be very careful to be able to use it safely. Since the CS implementation can't itself panic, it shouldn't be UB unless you release it from the non-owning core or from a context where it would be released again.

Additionally, you actually can't just release the spinlock through the pac because you need access to the static LOCK_OWNER variable. Otherwise you can't determine if the current core holds the lock. 

I think writing to a defined region of RAM might be a good solution for my particular problem, but I can imagine that there are more sophisticated use cases where you may want to do some work within a panic handler without touching external state.

Lastly, an annoying side effect of not being able to do this is that defmt_rtt, which relies on critical section, does not work from the other core while a panic happens on the core which holds the lock.

Could we consider putting this API behind a feature gate?

Dirbaio:
> can't just release the spinlock through the pac because you need access to the static LOCK_OWNER variable. Otherwise you can't determine if the current core holds the lock.

`manually_release()` is not doing any checking either, it's force-unlocking no matter which core has it. So you *can* get the equivalent of `manually_release()` with just the PAC. 

What you can't get is `current_lock_owner()`. However i'm not sure how that's useful. If you wait in user code for the other core to release the lock you can get deadlocks (if you got a panic the firmware is crashed anyway, it's possible the other core also hangs forever).

> Could we consider putting this API behind a feature gate?

I would prefer not, sorry.

wowkster:
Alright no worries. For now I'm going to do what you suggested with writing to RAM as I think that solves my use case anyway. I appreciate the attention to this though.

PR #4016, opened 2025-03-27
Support eMMC devices via SDMMC (STM32)

antonok-edm:
I've added support for eMMC memory devices in `embassy-stm32` using the SDMMC peripheral.

I've confirmed that this code works correctly for basic initialization, reads, and writes on a device that I've previously been using under `stm32f4xx-hal`, using 1, 4, and 8 data lanes, each with several different operating frequencies up to 24MHz.

Note that I don't have any appropriate hardware to test:
- `gpio_v1` config
- `sdmmc_v2` config
- SD card regressions

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -70,7 +70,7 @@ cortex-m-rt = ">=0.6.15,<0.8"
 cortex-m = "0.7.6"
 futures-util = { version = "0.3.30", default-features = false }
 rand_core = "0.6.3"
-sdio-host = "0.6.0"
+sdio-host = "0.9.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "16" }
 stm32-metapac = { git = "https://github.com/embassy-rs/stm32-data-generated", tag = "stm32-data-4a964af03b298de30ff9f84fcfa890bcab4ce609" }
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -70,7 +70,7 @@ cortex-m-rt = ">=0.6.15,<0.8"
 cortex-m = "0.7.6"
 futures-util = { version = "0.3.30", default-features = false }
 rand_core = "0.6.3"
-sdio-host = "0.5.0"
+sdio-host = "0.6.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "16" }
 stm32-metapac = { git = "https://github.com/embassy-rs/stm32-data-generated", tag = "stm32-data-4a964af03b298de30ff9f84fcfa890bcab4ce609" }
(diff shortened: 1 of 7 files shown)


Dirbaio:
Thanks for the PR!

Could you not add a new trait + generic param for the mode, track it at runtime instead (with e.g. an enum?). Reasons are:
- There's code duplication in `init_emmc`/`init_card` and `read_sd_status`/`read_ext_csd`. With an enum you can easily "if emmc do this, if sd do that" in the few things that do change without having to copypaste the entire methods. 
- Less generics makes the API simpler for the user.

antonok-edm:
@Dirbaio sure, I'll try that. Poor timing on my part - right after I submitted this PR, I already had to rebase it because of #4005 doing the same thing for DMA :sweat_smile:

antonok-edm:
@Dirbaio I just pushed a new enum-based implementation in case you'd like to generally check over it. There are a couple of things that I think are slightly less clean than with traits but it's not too bad overall. I didn't attempt to address the code duplication you mentioned yet; will get to that tomorrow hopefully.

(previous implementation is still in [`antonok-edm/embassy#emmc-trait`](https://github.com/antonok-edm/embassy/tree/emmc-trait) if needed)

antonok-edm:
@Dirbaio ok, should be ready for review now - I've deduplicated the `init` methods and pulled common code out from all the block read/write functions into `complete_datapath_transfer`.

Dirbaio (approved):
thank you!

Dirbaio:
Seems this breaks the [sdcard test](https://github.com/embassy-rs/embassy/blob/main/tests/stm32/src/bin/sdmmc.rs)

bender run

Dirbaio:
hm I think the "task arena is full" error is a red herring. Task arena is gone since #4020. Try rebasing on latest main, you'll likely get the same error in PR CI as in the merge queue CI, which is "waiting for card: NoCard".

antonok-edm:
@Dirbaio SD card CI is now fixed, but there is still a `usart_rx_ringbuffered` error which seems unrelated, even after rebasing on `main`. Not sure if you have any ideas there.

Dirbaio:
yeah it's just flaky, i've removed it in #4054 

bender run

chemicstry:
Hey, thank you for implementing this right when I need it!

I, however, am having problems and would appreciate any help. I'm using STM32F429ZIT6 with EMMC16G-MW28-03022 flash on 8 bit bus and it just freezes on `init_emmc()`. Seems like an infinite loop as no other async tasks are being executed. What is strange, is that when I set breakpoints anywhere in the code it manages to initialize the eMMC and reports correct capacity. Any read after that fails with `panicked at embassy-stm32\src\dma\dma_bdma.rs:289:21: DMA: error on DMA@40026400 channel 6`. I tried using 1, 4 and 8 bit buses with no luck.

Have you experienced anything similar and could nudge me where to look?

PR #4017, opened 2025-03-27
embassy-rp: Spinlock mutex implementation

shilga:


Dirbaio:
seems test is failing

Dirbaio:
the other multicore tests work reliably, i've not seen them be flaky. Have you tried running the test locally?

shilga:
I have not run them locally. It was already my plan to do that as soon as I'm back home. Need to figure out how to do it first actually...

The stacktrace which is in the logs of the run is the moment the error happens? Or am I just reading it wrong? Is it possible that the test is not waiting long enough before trying to connect?

Is it possible there was some other hickup? There is also the perf test failing in that specific run. But it's suspicious of course that exactly the new test was failing.

Dirbaio:
> how to do it

- `cargo install --git https://github.com/embassy-rs/teleprobe --locked`
- uncomment right runner here https://github.com/embassy-rs/embassy/blob/main/tests/rp/.cargo/config.toml#L9-L10
- `cargo run --release --bin whatever`

> Is it possible there was some other hickup? 

let's rerun to be sure. (note the `trusted` label makes tests run every time you push to the PR now)

bender run

shilga:
okay, so the same two tests failed again with the same stacktrace. The new multicore mutix test seems to only fail on the Pimoroni Pico Plus 2, not on the normal PICO. The perf test the other way around.

Dirbaio:
the cyw43 test is known flaky btw, just ignore it.

Dirbaio:
> I assume the multicore tests are deactivated for a reason on the RP2350B board

oh :rofl: yeah that explains it.

shilga:
that was the cyw43 again

shilga:
Thank you for the support

PR #4007, opened 2025-03-25
Statically allocated task pools on stable

0e4ef622:
Not sure if this has been explored before, but had to try myself. Using `size_of` and `align_of` we can create a static variable with the same size and alignment as the task pool that we then transmute to use.

Example:

```rust
#[task]
async fn task1(a: u32, b: u32, #[cfg(any())] c: u32) {
    let (_, _) = (a, b);
}
```

Expanded:
```rust
#[doc(hidden)]
async fn __task1_task(a: u32, b: u32, #[cfg(any())] c: u32) {
    let (_, _) = (a, b);
}
fn task1(a: u32, b: u32, #[cfg(any())] c: u32) -> ::embassy_executor::SpawnToken<impl Sized> {
    trait _EmbassyInternalTaskFn<Args>: Copy {
        type Fut: ::core::future::Future + 'static;
    }
    impl<F, Fut, T0, T1> _EmbassyInternalTaskFn<(T0, T1)> for F
    where
        F: Copy + FnOnce(T0, T1) -> Fut,
        Fut: ::core::future::Future + 'static,
    {
        type Fut = Fut;
    }
    impl<F, Fut, T0, T1, T2> _EmbassyInternalTaskFn<(T0, T1, T2)> for F
    where
        F: Copy + FnOnce(T0, T1, T2) -> Fut,
        Fut: ::core::future::Future + 'static,
    {
        type Fut = Fut;
    }
    const fn __task_pool_size<F, Args>(_: F) -> usize
    where
        F: _EmbassyInternalTaskFn<Args>,
    {
        ::core::mem::size_of::<::embassy_executor::raw::TaskPool<F::Fut, POOL_SIZE>>()
    }
    const fn __task_pool_align<F, Args>(_: F) -> usize
    where
        F: _EmbassyInternalTaskFn<Args>,
    {
        ::core::mem::align_of::<::embassy_executor::raw::TaskPool<F::Fut, POOL_SIZE>>()
    }
    const fn __task_pool_new<F, Args>(_: F) -> ::embassy_executor::raw::TaskPool<F::Fut, POOL_SIZE>
    where
        F: _EmbassyInternalTaskFn<Args>,
    {
        ::embassy_executor::raw::TaskPool::new()
    }
    const fn __get_pool<F, Args>(
        _: F,
    ) -> &'static ::embassy_executor::raw::TaskPool<F::Fut, POOL_SIZE>
    where
        F: _EmbassyInternalTaskFn<Args>,
    {
[...]

PR #4020, opened 2025-03-28
Statically allocate task pools on stable Rust.

Dirbaio:
Revised version of #4007 with the following changes:

- Move things from the macro to lib.rs. Makes editing easier and reduces amount of code generated by the macro.
- Instead of adding a dep on `elain`, make our own mini simplified version.
- Use `repr(C)` struct instead of tuple for the alignment (tuples are `repr(Rust)`)

Thanks @0e4ef622 for the awesome idea of how to do it and the first implementation.


Dirbaio:
bender run

Dirbaio:
bender run

PR #4022, opened 2025-03-29
mspm0: blocking uart driver

i509VCB:
And examples.

G3519 can't execute this currently due to an unrelated WWDT0 issue (no idea why it's interrupt goes off, I am not even touching it).

### DIFF
--- a/embassy-mspm0/src/uart.rs
+++ b/embassy-mspm0/src/uart.rs
@@ -1,0 +1,1085 @@
+#![macro_use]
+
+use core::marker::PhantomData;
+use core::sync::atomic::{compiler_fence, AtomicU32, Ordering};
+
+use embassy_embedded_hal::SetConfig;
+use embassy_hal_internal::PeripheralType;
+
+use crate::gpio::{AnyPin, PfType, Pull, SealedPin};
+use crate::interrupt::{Interrupt, InterruptExt};
+use crate::mode::{Blocking, Mode};
+use crate::pac::uart::{vals, Uart as Regs};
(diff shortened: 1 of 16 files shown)


i509VCB:
Some bits of cleanup I need to do. Will undraft when ready.

embassy-mspm0/src/uart.rs
```diff
@@ -0,0 +1,1085 @@
...
+    let half = U26F6::from_num(1) / U26F6::from_num(2);
+    // fbrd = INT(((FRAC(BRD) * 64) + 0.5))
+    let fbrd = brd
+        .frac()
+        .checked_mul(fbrd_scale)?
+        .checked_add(half)?
+        .checked_to_num::<u8>()?;
+
+    Some((ibrd, fbrd))
+}
```
i509VCB, line 978: The last thing before I merge. I would like someone to double check the calculations here.

[18.2.3.4 Baud Rate Generation (page 1351)](https://www.ti.com/lit/ug/slau846/slau846.pdf) describes the calculations for `IBRD` and `FBRD`. There is an example at the bottom of this file which does verify the example in the datasheet is produced. This also does work since the examples at 115200 baud for the virtual com port do work.

i509VCB: @korbin I gave this an attempt to make it const and it would involve manual `match`/`if let`/`let` guard statements in a lot of places. I feel that makes it far harder to read and understand. Until `impl const` is stable I doubt this function will be easier as const.

PR #4025, opened 2025-03-30
Gpdma circular operation

tyler-gilbert:
This PR adds support for circular DMA operation using the GPDMA. The GPDMA requires a pointer to a 32-bit address location which tells the DMA where the start of the next transaction. This value is added to the DMA channel state.

This PR also enables ring buffer operation on the SAI driver to prove the GPDMA code is working.

I have tested this using an STM32U545 Nucleo dev board with an NAU8315 speaker driver in I2S mode and an SPH0645 I2S microphone.

This PR also upates the SAI driver to accept a u16 for the frame length. This allows setting the frame length to 256 clock cycles. The underlying register is only 8 bits but the register is the number of clock cycles minus one. So a `u16` is need to convey 256 cycles.

This is the SAI configuration used to test.

```rs
let (sub_block_receiver, sub_block_transmitter) = sai::split_subblocks(tx_peripherals.sai);

//The NAU8315 has some odd bring up requirements. The EN may need to be toggled after cycling the power.
//The EN or power may also need to be cycles after the SCLK/FS signals are available.

let (tx_config, transmitter) = {
    let tx_config = {
        config.mode = Mode::Master;
        config.tx_rx = TxRx::Transmitter;
        config.mute_detection_counter = embassy_stm32::dma::word::U5(0);
        config.master_clock_divider = MasterClockDivider::Div6;
        config.bit_order = BitOrder::MsbFirst;
        config.fifo_threshold = FifoThreshold::Empty;
        config.sync_output = true;
        config.stereo_mono = StereoMono::Mono;

        // NAU8315 datasheet claims min bitrate clock with 8KHz FS clock is 2048KHz
        // On the STM32U545, this works with and without the following lines of code
        // On the STM32F429, without the following lines of code, there are a lot of pops and snaps
        config.frame_length = 256;
        config.frame_sync_active_level_length = embassy_stm32::dma::word::U7(128);

        config
    };

    (
        tx_config,
        sai::Sai::new_asynchronous(
            sub_block_transmitter,
            tx_peripherals.sck,
            tx_peripherals.sd,
            tx_peripherals.fs,
            tx_peripherals.dma,
            tx_buffer,
            tx_config,
        ),
    )
};

let receiver = {
    let rx_config = {
        let mut config = tx_config.clone();
        config.tx_rx = TxRx::Receiver;

        config.data_size = DataSize::Data32;
        config.frame_length = 64;
        config.frame_sync_active_level_length = embassy_stm32::dma::word::U7(32);
        config
    };

    sai::Sai::new_asynchronous(
        sub_block_receiver,
        rx_peripherals.sck,
        rx_peripherals.sd,
        rx_peripherals.fs,
        rx_peripherals.dma,
        rx_buffer,
        rx_config,
    )
};
```

This is the pin configuration:

```rs
pub type SpiTransmitChannel = Peri<'static, peripherals::GPDMA1_CH5>;
pub type SpiReceiveChannel = Peri<'static, peripherals::GPDMA1_CH2>;
pub struct SaiTransmitterPeripherals {
    pub sai: Peri<'static, peripherals::SAI1>,
    pub sck: Peri<'static, peripherals::PB3>, //D3
    pub sd: Peri<'static, peripherals::PB5>,  //D4
    pub fs: Peri<'static, peripherals::PB6>,  //D15
    pub dma: SaiTransmitChannel,
}

pub struct SaiReceiverPeripherals {
    pub sck: Peri<'static, peripherals::PA8>, //D7 OR CN10 - 23
    pub sd: Peri<'static, peripherals::PA10>, //CN10 - 4
    pub fs: Peri<'static, peripherals::PA9>,  //CN10 - 2
    pub dma: SaiReceiveChannel,
}
```

### DIFF
--- a/embassy-stm32/src/dma/gpdma.rs
+++ b/embassy-stm32/src/dma/gpdma.rs
@@ -1,13 +1,14 @@
 #![macro_use]
 
-use core::future::Future;
+use core::future::{poll_fn, Future};
 use core::pin::Pin;
-use core::sync::atomic::{fence, Ordering};
-use core::task::{Context, Poll};
+use core::sync::atomic::{fence, AtomicUsize, Ordering};
+use core::task::{Context, Poll, Waker};
 
 use embassy_hal_internal::Peri;
 use embassy_sync::waitqueue::AtomicWaker;
 
+use super::ringbuffer::{DmaCtrl, Error, ReadableDmaRingBuffer, WritableDmaRingBuffer};
 use super::word::{Word, WordSize};
 use super::{AnyChannel, Channel, Dir, Request, STATE};
 use crate::interrupt::typelevel::Interrupt;
@@ -46,11 +47,15 @@ impl From<WordSize> for vals::Dw {
 
 pub(crate) struct ChannelState {
     waker: AtomicWaker,
+    circular_address: AtomicUsize,
+    complete_count: AtomicUsize,
 }
 
 impl ChannelState {
     pub(crate) const NEW: Self = Self {
         waker: AtomicWaker::new(),
+        circular_address: AtomicUsize::new(0),
+        complete_count: AtomicUsize::new(0),
     };
 }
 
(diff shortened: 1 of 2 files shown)


tyler-gilbert:
For those who have expressed interest in a circular buffer on the GPDMA, I finally got around to getting this working with the latest version of embassy.

cc: @elagil @Luctins @osobiehl @DrTobe

elagil:
Very nice!

I don't know if you have seen it, but I had also implemented a version of circular GPDMA here: https://github.com/embassy-rs/embassy/pull/3923
Maybe we can consolidate the efforts?

My final goal is to only queue a new linked-list item, when the buffer can be safely used by the DMA. For example, in write mode, the next linked list item is only queued, when the user called `write()` before the DMA finished with the current LLI. Otherwise the DMA stops (current LLI has empty address for the next item) and the next `write()` returns an error to the user.

tyler-gilbert:
@elagil 

#3923 looks nice. It does the job for me. I tested your branch on my hardware setup and both the audio output and input worked as expected. I am happy to consolidate the effort. Is there anything you need?

I think merging your branch makes more sense because it expands the feature-set to better match the hardware features (GPDMA linked-list support).

elagil:
@tyler-gilbert 
Thank you for testing!

Actually, it doesn't work correctly for me yet. I assume that I have an error in the buffer switching logic. My tests are done with USB audio streaming, which also has non-constant buffer write lengths.

I don't know yet how to finalize it, but your SAI fixes for supporting longer frames seem like something that can be merged already.

Dirbaio:
done in #3923

PR #4027, opened 2025-03-31
Remove task arena FAQ

bugadani:


Dirbaio:
thanks! perhaps we should make an embassy-executor release first?

bugadani:
Fine with me, this will be here once you are ready for it :)

PR #4028, opened 2025-03-31
stm32: Add non-async EXTI configuration for RTIC and custom interrupt handlers

michelrandahl:
# Add blocking EXTI support for RTIC and custom interrupt handlers

## Motivation

This PR adds support for using Embassy's EXTI functionality with RTIC and custom interrupt handling in general. The current Embassy EXTI implementation registers its own interrupt handlers, which creates ldd errors when trying to use it with RTIC or custom applications that also define the same interrupts.

## Changes

- Add a new feature flag `exti-with-custom-handlers` that makes it possible to configure EXTI functionality while allowing applications to define their own interrupt handlers (required for RTIC integration)
- Refactor EXTI code to separate common low-level functionality into a reusable module
- Create a non-async, blocking version of the EXTI API for use with interrupt-driven frameworks (eg. RTIC)
- Conditionally compile Embassy's interrupt handlers only when not using custom handlers
- Add examples demonstrating:
  - Custom interrupt handlers (without RTIC)
  - Integration with RTIC

## Objectives

- Enable using Embassy with RTIC without compiler ldd errors
- Maintain backward compatibility with existing Embassy applications

## Testing

I've tested this implementation on an STM32L4 series MCU with both custom interrupt handlers, RTIC as well as the old stm32l4 examples to verify that old logic is still working.

### DIFF
--- a/embassy-stm32/src/exti/low_level.rs
+++ b/embassy-stm32/src/exti/low_level.rs
@@ -129,6 +129,7 @@ pub(super) fn clear_exti_pending(pin: u8) {
 ///
 /// # Returns
 /// `true` if an interrupt is pending, `false` otherwise
+#[cfg(feature = "exti-with-custom-handlers")]
 pub(super) fn is_exti_pending(pin: u8) -> bool {
     let pin = pin as usize;
 


Dirbaio:
This has been brought before, my idea for how to solve it was to make async EXTI use `bind_interrupts!` just like the other drivers.

`bind_interrupts!` exists precisely to solve this problem: it avoids the HAL registering every single interrupt for you, to leave you the option of handling the interrupt yourself (with cortex-m-rt, or with RTIC, or whatever). 

A while ago all drivers (exti, uart, spi, etc) would outright reigster all the interrupts. When `bind_interrupts!` was introduced most drivers were ported to it, but EXTI wasn't. I think it can.

The way it'd work is to use async EXTI you'd `bind_interrupts!` it and pass the `Irqs` token to `ExtiInput::new()` to "prove" you did. If you don't then you can't create an `ExtiInput`, but you're free to register the irq handler yourself and use it with either the PAC or with a "low-level" exti API like you've added in this PR.

embassy-stm32/src/exti/blocking.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /// Configures the EXTI line to detect rising edges
+    ///
+    /// After calling this method, the EXTI line will generate interrupts
+    /// on rising edges of the input signal.
+    ///
+    /// # Returns
+    /// A mutable reference to self for method chaining
+    pub fn rising_edge(&mut self) -> &mut Self {
```
Dirbaio, line 61: don't make "builder-like" APIs that do chaining, it's inconisnstent with how the rest of the Embassy APIs look like.

michelrandahl:
> This has been brought up before, my idea for how to solve it was to make async EXTI use `bind_interrupts!` just like the other drivers.
> 
> `bind_interrupts!` exists precisely to solve this problem: it avoids the HAL registering every single interrupt for you, to leave you the option of handling the interrupt yourself (with cortex-m-rt, or with RTIC, or whatever).
> 
> A while ago all drivers (exti, uart, spi, etc) would outright reigster all the interrupts. When `bind_interrupts!` was introduced most drivers were ported to it, but EXTI wasn't. I think it can.
> 
> The way it'd work is to use async EXTI you'd `bind_interrupts!` it and pass the `Irqs` token to `ExtiInput::new()` to "prove" you did. If you don't then you can't create an `ExtiInput`, but you're free to register the irq handler yourself and use it with either the PAC or with a "low-level" exti API like you've added in this PR.

Thats an interesting insight, it definitely makes sense to keep this consistent across the drivers. I'll give it a stab and see if I can implement it for exti.

WillaWillNot:
@michelrandahl in this time since this PR, there are now blocking versions of the EXTI futures (`poll_for_rising_edge` etc) in the driver, and I just added custom EXTI interrupt binding in PR #4922 . Does that cover everything needed for this RTIC use case? If so, perhaps you could resubmit this as just your RTIC usage example!

michelrandahl:
@WillaWillNot While it doesn't immediately enable RTIC usage, it makes it significantly easier to introduce a non-async alternative to the async ExtiInput without needing feature flags and workarounds. Your hard work is much appreciated, thank you! :pray: 

I was not too happy about my draft PR here, so I'm closing it and creating a new PR with relevant changes.

PR #4032, opened 2025-04-01
Add USBHost v2 

i404788:
Follow up to #3307 (see for discussion)

This add `UsbHostDriver` & `UsbChannel` traits allowing HAL's to implement Usb Host support.

Example handlers are in `embassy-usb/src/handlers/`, and there is additionally a UAC handler made by @AlexCharlton [available](https://github.com/embassy-rs/embassy/tree/22cb302a4e65598d79d95457ba69166666bbfcda/embassy-usb/src/handlers/uac).

Changes since #3307:
* `UsbChannel`s now have a `set_timeout` to enable configurable timeouts
* Descriptor parsing is refactored to use iterators, this is more intuitive since parsing needs to be done sequentially
* Configuration of Usb Devices now happens *after* enumeration, and can be done either by the handler (usually deferring to the active config, example handlers) or the application.

Pending tasks:
* [ ] Update embassy-rp to use the latest change
* [ ] Update stm32g0 to add `set_timeout`
* [ ] Update esp-hal (out-of-tree) to add `set_timeout`
* [ ] Documentation & configs clean up


embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn retarget_channel(&mut self, addr: u8, endpoint: &EndpointInfo, pre: bool) -> Result<(), HostError>;
+
+    /// Send IN request of type other from control
+    /// For interrupt channels this will return the result of the next succesful interrupt poll
+    async fn request_in(&mut self, buf: &mut [u8]) -> Result<usize, ChannelError>
+    where
+        D: channel::IsIn;
+
+    /// Send OUT request of type other from control
+    async fn request_out(&mut self, buf: &[u8]) -> Result<usize, ChannelError>
```
Dominaezzz, line 303: Why does this return `usize`? Would the channel ever write less data that what was given and not call it some kind of error?

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+    where
+        T: channel::IsControl,
+        D: channel::IsOut;
+
+    /// Retargets channel to a new endpoint, may error if the underlying driver runs out of resources
+    fn retarget_channel(&mut self, addr: u8, endpoint: &EndpointInfo, pre: bool) -> Result<(), HostError>;
+
+    /// Send IN request of type other from control
+    /// For interrupt channels this will return the result of the next succesful interrupt poll
+    async fn request_in(&mut self, buf: &mut [u8]) -> Result<usize, ChannelError>
```
Dominaezzz, line 298: Is `buf` a single packet or a whole transfer?
If it's the latter, this is missing ZLP handling

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/// ## Virtual USB Channels
+/// These contain the required information to send a packet correctly to a device endpoint.
+/// The information is carried with the channel on creation (see [`UsbHostDriver::alloc_channel`]) and can be changed with [`UsbChannel::retarget_channel`].
+///
+/// It is up to the hal's driver how to implement concurrent requests, some hardware IP may allow for multiple hardware channels
+///  while others may only have a single channel which needs to be multiplexed in software, while others still use DMA request linked-lists.
+/// Any of these are compatibile with the UsbChannel with varying degrees of sync primitives required.
+pub trait UsbChannel<T: channel::Type, D: channel::Direction> {
```
Dominaezzz, line 280: Are the generics really needed? These could be replaced with runtime checks or more traits.
```Rust
trait UsbChannel;
trait UsbChannelIn;
trait UsbChannelOut;
trait UsbChannelControl;
```

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+        D: channel::IsIn;
+
+    /// Send OUT control request
+    async fn control_out(&mut self, setup: &SetupPacket, buf: &[u8]) -> Result<usize, ChannelError>
+    where
+        T: channel::IsControl,
+        D: channel::IsOut;
+
+    /// Retargets channel to a new endpoint, may error if the underlying driver runs out of resources
+    fn retarget_channel(&mut self, addr: u8, endpoint: &EndpointInfo, pre: bool) -> Result<(), HostError>;
```
Dominaezzz, line 294: When should this be chosen over `drop`ing and calling `alloc_channel` again?

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+pub enum ChannelError {
+    /// Either the packet to be written is too long to fit in the transmission
+    /// buffer or the received packet is too long to fit in `buf`.
+    BufferOverflow,
+
+    /// Response from device/bus was not interpretable (Crc, Babble)
+    BadResponse,
+
+    /// Transaction was canceled
+    Canceled,
```
Dominaezzz, line 19:
```suggestion
    Cancelled,
```

embassy-usb-driver/src/host.rs
i404788: This is for the whole transfer, I'm not sure why you'd need to handle ZLP at the interface level though. The ZLP handling is determined by the driver implementation by the channel type.

embassy-usb-driver/src/host.rs
Dominaezzz, line 294:
```suggestion
    fn retarget(&mut self, addr: u8, endpoint: &EndpointInfo, pre: bool) -> Result<(), HostError>;
```

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,309 @@
...
+    impl IsOut for InOut {}
+}
+
+/// Specify the timeout of a channel
+pub struct TimeoutConfig {
+    /// Maximum response timeout for transactions with a Data Stage
+    pub data_timeout: Duration,
+    /// Maximum response timeout for transactions without a data stage
+    pub standard_timeout: Duration,
+}
```
Dominaezzz, line 262: Could these be substituted with `with_timeout` on the user side? These don't seem granular enough to be useful.

Dominaezzz (commented):
Thanks for working on this

embassy-usb-driver/src/host.rs
i404788: True, had to check the other drivers (I'm most familiar with synposys), but they all do full writes, so the usize is unnecessary.  I also don't imagine a driver would fall back to a higher layer rather than looping & yielding. So I'll change this.

embassy-usb-driver/src/host.rs
Dominaezzz: > The ZLP handling is determined by the driver implementation by the channel type.

The USB Class implementation decides where a ZLP is needed.
The host implementation needs to know whether to expect/send one at the end of the transfer, particularly when `buf.len()` is a multiple of the endpoint's max packet size.

i404788: Unless I'm mistaken, this is already provided in the `EndpointInfo` that is provided to a channel.

embassy-usb-driver/src/host.rs
i404788: The main issue with `with_timeout` is that it has to be done for each request, if forgotten the request may stall infinitely. And since the standard already provides timeout defaults/recommendations, it makes more sense to configure it up-front (with the spec's recommended values) rather than wrap each call that is a transfer.

Dominaezzz: That's true. Do those timeouts only apply to control transfers?

embassy-usb-driver/src/host.rs
i404788: I've considered removing it before, the reason against it is for fields in structs, and the somewhat counterintuitive nature of dropping mid-function to reclaim channels.

embassy-usb-driver/src/host.rs
Dominaezzz: > Unless I'm mistaken, this is already provided in the `EndpointInfo` that is provided to a channel.

That's impossible, it's not something that is decided once, it's decided per transfer as a ZLP isn't always needed.

I'll go about this a different way. Let's say I have a channel and I want to send three packets `[[0, 0, 0, 0, 0, 0, 0, 0], [1, 1, 1, 1, 1, 1, 1, 1], []]`. (The endpoint has a max packet size of 8)

How would I do this using `request_out`?

embassy-usb-driver/src/host.rs
i404788: I believe the spec doesn't differentiate between types for timeouts just if it has a data stage or not (with the exception of SetAddress command). So the timeouts apply to all transfers on a specific channel.

embassy-usb-driver/src/host.rs
i404788: `request_out(&[0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1])` should give you the requested output for an mps of 8. If you *only* want to a ZLP you would do `request_out(&[])`. As I understand it ZLP is mostly about flushing when there is ambiguity, which in many cases is already handled in hw rather than the driver.

Dominaezzz: Right, what if I want to send `[[0, 0, 0, 0, 0, 0, 0, 0], [1, 1, 1, 1, 1, 1, 1, 1]]` instead? (I've remove the last empty packet)
What would the `request_out` call look like?

> As I understand it ZLP is mostly about flushing when there is ambiguity, which in many cases is already handled in hw rather than the driver.

Yes it usually about that but it's still up to the usb class to decide this. Sure, the hw handles it but the it still needs to be told when to do it. (I'm also mostly familiar with Synopsis)

i404788: If there is a need to send a k*mps transfer without zlp, then I might be mistaken on how dwc2 handles zlp (although I can't find anything about it); otherwise there would be no way physical way to send a packet without zlp.

Dominaezzz: DWC2 asks for a transfer_size and packet count for this reason. You can simply do +1 on the packet count if you want to send the ZLP for a k*mps transfer.

For the examples above:
- `[[0, 0, 0, 0, 0, 0, 0, 0], [1, 1, 1, 1, 1, 1, 1, 1], []]` would be transfer size of 16 and packet count of 3.
- `[[0, 0, 0, 0, 0, 0, 0, 0], [1, 1, 1, 1, 1, 1, 1, 1]]` would be transfer size of 16 and packet count of 2.

i404788: Interesting, Is there any reason not to do a ZLP in this kind of scenario other than working around badly implemented usb devices? I don't see this being implemented in any of my references.

Dominaezzz: Ehh, to support all USB classes. Sending a ZLP when you shouldn't would simply break the class protocol.

For a concrete example see https://github.com/embassy-rs/embassy/pull/2799#discussion_r1900469805 .

i404788: Hmm definitely an interesting read from a device perspective. But your thread reminded me that I recently worked on a user-space (minimal) msc driver which uses `nusb` and `nusb` also doesn't have any explicit zlp handling at all, so I'm not sure that it is needed in practice.

I would rather stick to 'standard' usb transfers for the >99% of cases rather than adding zlp exceptions that I currently have no example for on the host-side. If it is found that it's needed it can be refactored fairly easily to `request_out_raw(data); if even {request_out_raw(&[])}` or similar with a small extension to the trait.
I can image it might be useful for streaming as mentioned in the link, but I'm not sure if we should handle that in the current version.

embassy-usb-driver/src/host.rs
i404788: Interrupt transfers should have a timeout (for the same reasons as before), as they *need* to be nak'ed for it to be a valid poll (they're not actually interrupts). Isochronous makes sense to omit.

You could specify that the timeout is specifically for the data stage of a request as is in the standard (which is valid for interrupt & bulk), however looking at the dwc driver it's already not possible to make this distinction with the hw.

We could do `set_timeout(Duration)` & `set_timeout_control(Duration, Duration`  & missing for isochronous; which doesn't seem as nice. If you have any ideas for this I'm open to suggestions.

embassy-usb-driver/src/host.rs
Dominaezzz: > I would rather stick to 'standard' usb transfers for the >99% of cases rather than adding zlp exceptions that I currently have no example for on the host-side.

What do you mean by "standard"? ZLP handling isn't an exception, it's just part of how USB transfers work.
Sometime the device you're talking knows exactly how much data it's getting and sometimes it doesn't, ZLP is typically used in the latter case.

> If it is found that it's needed it can be refactored fairly easily to `request_out_raw(data); if even {request_out_raw(&[])}`

Why leave this problem to software when most USB hardware can handle this?

> I can image it might be useful for streaming as mentioned in the link, but I'm not sure if we should handle that in the current version.

If the current `request_in` takes whole transfers (as suppose to just singular packets) then it's streaming as mentioned in the link. It must be handled to be compliant.

i404788: > Why leave this problem to software when most USB hardware can handle this?

It may be overridden by the host driver, but since it's equivalent there is not reason not to provide a default implementation.

> What do you mean by "standard"? 

*Not* handling ZLP is the exception, having to specify you want to deviating from ZLP seems to barely be used or implemented in any library, driver or hal I can find. A 'standard' usb transfer is ZLP if k*mps bytes otherwise a short packet; so the device always know when a transfer has ended.

While the spec is ambiguous, in every case I've looked so far issues with ZLP are either quirks of the device (i.e. inconsistent handling between devices), or issues with the host controller driver. For example Zypher has the `USB_TRANS_NO_ZLP` for device-side, it's only used 3 times, only one of which was due to an incompatibility issue; and even then they could not identify the root-cause (https://github.com/zephyrproject-rtos/zephyr/issues/20250).

In another case where a fix was "disable ZLP" (https://github.com/zephyrproject-rtos/zephyr/issues/21713), it ended up being a race condition in the usb stack. Asserting `Related to this, I understand ZLP's are essential to the USB spec. So, it seems counter intuitive to me to suppress them.`

And for host-side the always-ZLP is stronger: 

https://github.com/libusb/libusb/issues/1254#issuecomment-145814098

> This one is quite clear. For Bulk Out, the host code needs to send out the ZLP as the device side may not know the end of the transfer.

Again I have not found a case where omitting ZLP is more correct then sticking to bullet point 2 of the spec:
```
An endpoint must always transmit data payloads with a data field less than or equal to the endpoint’s
reported wMaxPacketSize value. When a bulk IRP involves more data than can fit in one maximum-sized
data payload, all data payloads are required to be maximum size except for the last data payload, which will
contain the remaining data. A bulk transfer is complete when the endpoint does one of the following:
• Has transferred exactly the amount of data expected
• Transfers a packet with a payload size less than wMaxPacketSize or transfers a zero-length packet
```

Dominaezzz: > While the spec is ambiguous

The spec is ambiguous because this decision is up to the usb class.

That quote from the spec perfectly states what I'm trying to say.
You need a way for the usb class to choose which bullet point they want to use.
Just add a `with_zlp: bool` parameter to handle both bullet points.

You don't need to find a case, the spec is all you need.

AlexCharlton:
Just looking into the refactored code now. Off the top, here's a commit that allows this branch to compile with the `log` feature: https://github.com/AlexCharlton/embassy/commit/9e6299c77638c9fca10e6ee5dd93e82324052cf5

embassy-usb/src/handlers/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /// Retrieve a device configuration by index up to a max of [`DeviceDescriptor::num_configurations`]
+    pub async fn get_configuration<'a, D: channel::IsIn, C: UsbChannel<channel::Control, D>>(
+        &self,
+        index: u8,
+        channel: &mut C,
+        cfg_desc_buf: &'a mut [u8],
+    ) -> Result<ConfigurationDescriptor<'a>, HostError> {
+        if self.device_desc.num_configurations > index {
```
AlexCharlton, line 122: Should be
```rust
        if index >= self.device_desc.num_configurations {
```

AlexCharlton:
https://github.com/AlexCharlton/embassy/commit/7eec4d897599723a3b2ab19f10a1fe258d601afd contains some additions to `descriptors.rs` that I've found useful for my handler.

There are three changes here, only one substantive:
- A `iter_descriptors` method is added to `ConfigurationDescriptor` (and a `DescriptorIterator` is defined). This is similar to `iter_interface` but is geared to being generally useful by handlers that need to be able to parse some set of class-specific descriptors. It returns byte slices of all descriptors found in the configuration.
- `StringIndex` is made pub, since I think this is a useful type alias worth using by handlers if they have their own descriptor types
- `EndpointDescriptor` is given `PartialEq` since have a test that tries to compare them

i404788:
I've already implemented a raw descriptor iterator (which includes the offset to the original buffer that's sometimes needed) at
https://github.com/i404788/embassy/blob/c463b80eb25ad680ff852a1eccefd4e39109e038/embassy-usb/src/host/descriptor.rs#L304

But it's only internally exposed, I'll add the `iter_descriptors` function seems very useful. `PartialEq` & `StringIndex` also seems like a good idea.

embassy-usb/src/host/mod.rs
```diff
@@ -64,7 +64,7 @@ pub trait ControlChannelExt<D: channel::Direction>: UsbChannel<channel::Control,
         };
 
         self.control_in(&packet, &mut buf).await?;
-        trace!("Descriptor {}: {=[u8]}", core::any::type_name::<T>(), buf);
+        trace!("Descriptor {}: {:x?}", core::any::type_name::<T>(), buf);
```
lynxD, line 67: Introduces error when ```cargo build```-ing in ```examples/rp```:
```
error: unknown display hint: "x?"
  --> /home/user/d/embassy/embassy-usb/src/host/mod.rs:67:16
   |
67 |         trace!("Descriptor {}: {:x?}", core::any::type_name::<T>(), buf);
   |                ^^^^^^^^^^^^^^^^^^^^^^
```
It only gets visible after fixing some other errors.

examples/stm32g0/src/bin/usb_host_hid_keyboard.rs
```diff
@@ -0,0 +1,144 @@
...
+use embassy_stm32::gpio::{AfType, Level, Output, OutputType, Speed};
+use embassy_stm32::i2c::{self, I2c};
+use embassy_stm32::time::{mhz, Hertz};
+use embassy_stm32::usb::UsbHost;
+use embassy_stm32::{bind_interrupts, pac, peripherals, usb, Config};
+use embassy_time::Timer;
+use embassy_usb::handlers::kbd::KbdHandler;
+use embassy_usb::handlers::UsbHostHandler;
+use embassy_usb::host::UsbHostBusExt;
+use embassy_usb_driver::host::DeviceEvent::Connected;
```
lynxD, line 15: I'm new to rust and embassy so I don't know if my idea is any good.

Wouldn't it be cleaner to make the API so that only the HAL create(host driver implementation) and stack implementation(embassy-usb) would use embassy_usb_driver crate? And users would just import two things: HAL crate for their platform and the USB stack crate. 
I'm not sure how to accomplish that. Move common types to new crate? Re-export the symbols in the USB stack crate?

i404788: Yes you are right, re-exporting is the correct way to do this, I'll make sure to change this

examples/rp/src/bin/usb_host_keyboard.rs
```diff
@@ -0,0 +1,45 @@
...
+        match usbhost.wait_for_device_event().await {
+            Connected(speed) => break speed,
+            _ => {}
+        }
+    };
+
+    println!("Found device with speed = {:?}", speed);
+
+    let enum_info = usbhost.enumerate_root_bare(speed, 1).await.unwrap();
+    let mut kbd = KbdHandler::try_register(&usbhost, &enum_info)
```
lynxD, line 37: How would I go about supporting the keyboard being unplugged and plugged in again?
Might be worth it to add that to the example.

lynxD: There is ``` HandlerEvent::HandlerDisconnected ``` but it's not triggered on disconnect. I don't know if it's a bug or it's for something else.

I've managed to get re-plugging devices to partially work. Seems channels get exhausted after some time (probably due to lack of drop_channel). All in all it's reliable only the first time the device is plugged in.
```
#![no_std]
#![no_main]

use defmt::*;
use embassy_rp::{bind_interrupts, peripherals, usb};
use embassy_usb_driver::host::{UsbHostDriver, DeviceEvent};
use embassy_usb::host::UsbHostBusExt;
use embassy_usb::handlers::{UsbHostHandler, kbd};
use embassy_futures::select::{select, Either};
use embassy_usb_driver::Speed;

    USBCTRL_IRQ => usb::host::InterruptHandler<peripherals::USB>;
});

#[embassy_executor::main]
async fn main(_spawner: Spawner) -> ! {

    let mut usbhost = usb::host::Driver::new(p.USB, Irqs);

    loop {
        match usbhost.wait_for_device_event().await {
            DeviceEvent::Connected(speed) => {
                handle_device(&mut usbhost, speed).await;
            }
            DeviceEvent::Disconnected => {
                warn!("Disconnected when there's no device?");
            }
        }
    };
}

async fn handle_device(
    usbhost: &mut embassy_rp::usb::host::Driver<'_, peripherals::USB>,
    speed: Speed) {
    let enum_info = usbhost.enumerate_root_bare(speed, 1).await.unwrap();
    info!("enum_info.device_desc = {}", enum_info.device_desc);
    let mut kbd = kbd::KbdHandler::try_register(usbhost, &enum_info).await.expect("Couldn't register keyboard");
    handle_kbd_events(usbhost, &mut kbd).await;
}

async fn handle_kbd_events(
    usbhost: &mut embassy_rp::usb::host::Driver<'_, peripherals::USB>,
    kbd: &mut kbd::KbdHandler<embassy_rp::usb::host::Driver<'_, peripherals::USB>>){
    loop {
        let result = select(usbhost.wait_for_device_event(), kbd.wait_for_event()).await;
        match result {
            Either::First(dev_event) => match dev_event {
                DeviceEvent::Connected(_) => {
                    defmt::panic!("More than one device?");
                }
                DeviceEvent::Disconnected => {
                    info!("Disconnected!");
                    break;
                }
            }
            Either::Second(kbd_event) => {
                info!("keyboard event = {}", kbd_event);
            }
        }
    };
}
```

embassy-usb/src/handlers/kbd.rs
```diff
@@ -0,0 +1,195 @@
...
+            enum_info.device_address,
+            &interrupt_ep.into(),
+            enum_info.ls_over_fs,
+        )?;
+
+        debug!("[kbd]: Setting PROTOCOL & idle");
+        const SET_PROTOCOL: u8 = 0x0B;
+        const BOOT_PROTOCOL: u16 = 0x0000;
+        if let Err(err) = control_channel
+            .class_request_out(SET_PROTOCOL, BOOT_PROTOCOL, iface.interface_number as u16, &[])
```
lynxD, line 98: This is a bit unfortunate as according to the HID spec the simplified, boot protocol is optional an the full report protocol is mandatory. So we should probably just use the report protocol for everything. I'd think we want to support something like game pads/joysticks anyway and the boot protocol is defined for keyboard and mice only.
I'll trying to write report protocol support atm.

embassy-usb/src/host/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+            // TODO: Log error or make descriptor error not generic
+            // error!("Device [{}]: Descriptor parse failed: {}", addr, e);
+            HostError::InvalidDescriptor
+        })
+    }
+
+    /// Request the underlying bytes for a descriptor of a specific type.
+    /// bytes.len() determines how many bytes are read at maximum.
+    /// This can be used for descriptors of varying length, which are parsed by the caller.
+    async fn request_descriptor_bytes<T: USBDescriptor>(
```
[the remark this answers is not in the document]
i404788: Yeah, we're kind of deprecating the USBDescriptor trait as it's only useful for fetching the headers of descriptors. Since we are restricted on memory buffer the typical flow of a descriptor request is `fetch header -> check if static buffer can support descriptor -> fetch buffer as size of descriptor`. For HID specifically you fetch the entire `ConfigurationDescriptor` and then use the `RawDescriptorIterator` as shown here https://github.com/embassy-rs/embassy/pull/4032/commits/39c9d167668d50c0a5b00cafb2de361175febfdb

Could definitely rename it, maybe `USBDescriptorHeader` or something similar would be more apt?

embassy-usb/src/handlers/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        channel::{self, IsIn, IsOut},
+        HostError, UsbChannel, UsbHostDriver,
+    },
+    Speed,
+};
+
+use crate::host::descriptor::{ConfigurationDescriptor, DeviceDescriptor, USBDescriptor};
+use crate::host::ControlChannelExt;
+
+pub mod hub;
```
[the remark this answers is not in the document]
i404788: The `handlers` module is current for host drivers but we can move it to host. Also I think it would be interesting to have an emulated hub for supporting multiple virtual devices, but that seems better for another PR.

embassy-usb/src/host/mod.rs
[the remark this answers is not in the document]
lynxD: Report descriptors don't even contain their own length(it's in corresponding HID descriptor). It's more akin to a string.
So I'm not sure it would be useful for that.
I don't have better idea for the name.

lynxD:
I've fixed some bugs:
https://github.com/lynxD/embassy/tree/usb-asynch-v2

Qix-:
@lynxD you can PR into the author's branch on their fork by the way. If merged, the changes will show up here automatically.

embassy-usb/src/host/mod.rs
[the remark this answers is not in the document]
i404788: Hmm it seems like the length of the report should be in the `wItemLength` in `HIDDescriptor` (EndpointDescriptor eqv.). So yeah `USBDescriptorHeader` wouldn't apply for ReportDescriptors

embassy-usb/src/handlers/kbd.rs
lynxD: I've been tinkering with hid last few days. I've came to conclusion it would deserve a crate of it's own. From what I see Bluetooth reuses the same report descriptors and parsing those is not insignificant amount of code.
I'd think we would be better served by a different example than a keyboard. I haven't checked but it seems to me that mass storage class would should be simple, useful and short. I might get around to writing later.

I think it would be good to have another example tat relies on vendor_id and product_id. So that we get the driver choice right. I don't have idea of a device that would not have a class, Not be a device that should have been a class device. Would have a  simple and short driver. Would be useful to embassy users. Anyone has an idea? If we don't get any we can make a dummy driver.

embassy-usb/src/host/descriptor.rs
```diff
@@ -0,0 +1,730 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index >= self.cfg_desc.num_interfaces {
+            None
+        } else {
+            // Assuming spec compliant descriptors the next descriptor should always be at our offset
+            let remaining_buf = &self.cfg_desc.buffer[self.offset..];
+            // FIXME: Fallible, propegate errors?
+            let iface = InterfaceDescriptor::try_from_bytes(remaining_buf).ok()?;
+            self.offset += iface.len as usize + iface.buffer.len();
+            self.index += 1;
```
lynxD, line 197: There is a bug here. It assumes that there is one interface descriptor per interface.

A DS4 gamepad gives me following configuration descriptor decoded with (http://eleccelerator.com/usbdescreqparser/):
```
0x09,        // bLength
0x02,        // bDescriptorType (Configuration)
0xE1, 0x00,  // wTotalLength 225
0x04,        // bNumInterfaces 4
0x01,        // bConfigurationValue
0x00,        // iConfiguration (String Index)
0xC0,        // bmAttributes Self Powered
0xFA,        // bMaxPower 500mA

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x00,        // bInterfaceNumber 0
0x00,        // bAlternateSetting
0x00,        // bNumEndpoints 0
0x01,        // bInterfaceClass (Audio)
0x01,        // bInterfaceSubClass (Audio Control)
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x0A,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x01,        // bDescriptorSubtype (CS_INTERFACE -> HEADER)
0x00, 0x01,  // bcdADC 1.00
0x47, 0x00,  // wTotalLength 71
0x02,        // binCollection 0x02
0x01,        // baInterfaceNr 1
0x02,        // baInterfaceNr 2

0x0C,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x02,        // bDescriptorSubtype (CS_INTERFACE -> INPUT_TERMINAL)
0x01,        // bTerminalID
0x01, 0x01,  // wTerminalType (USB Streaming)
0x06,        // bAssocTerminal
0x02,        // bNrChannels 2
0x03, 0x00,  // wChannelConfig (Left and Right Front)
0x00,        // iChannelNames
0x00,        // iTerminal

0x0A,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x06,        // bDescriptorSubtype (CS_INTERFACE -> FEATURE_UNIT)
0x02,        // bUnitID
0x01,        // bSourceID
0x01,        // bControlSize 1
0x03, 0x00,  // bmaControls[0] (Mute,Volume)
0x00, 0x00,  // bmaControls[1] (None)

0x09,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x03,        // bDescriptorSubtype (CS_INTERFACE -> OUTPUT_TERMINAL)
0x03,        // bTerminalID
0x02, 0x04,  // wTerminalType (Headset)
0x04,        // bAssocTerminal
0x02,        // bSourceID
0x00,        // iTerminal

0x0C,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x02,        // bDescriptorSubtype (CS_INTERFACE -> INPUT_TERMINAL)
0x04,        // bTerminalID
0x02, 0x04,  // wTerminalType (Headset)
0x03,        // bAssocTerminal
0x01,        // bNrChannels 1
0x00, 0x00,  // wChannelConfig
0x00,        // iChannelNames
0x00,        // iTerminal

0x09,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x06,        // bDescriptorSubtype (CS_INTERFACE -> FEATURE_UNIT)
0x05,        // bUnitID
0x04,        // bSourceID
0x01,        // bControlSize 1
0x03, 0x00,  // bmaControls[0] (Mute,Volume)
0x00,        // iFeature

0x09,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x03,        // bDescriptorSubtype (CS_INTERFACE -> OUTPUT_TERMINAL)
0x06,        // bTerminalID
0x01, 0x01,  // wTerminalType (USB Streaming)
0x01,        // bAssocTerminal
0x05,        // bSourceID
0x00,        // iTerminal

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x01,        // bInterfaceNumber 1
0x00,        // bAlternateSetting
0x00,        // bNumEndpoints 0
0x01,        // bInterfaceClass (Audio)
0x02,        // bInterfaceSubClass (Audio Streaming)
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x01,        // bInterfaceNumber 1
0x01,        // bAlternateSetting
0x01,        // bNumEndpoints 1
0x01,        // bInterfaceClass (Audio)
0x02,        // bInterfaceSubClass (Audio Streaming)
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x07,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x01,        // bDescriptorSubtype (CS_INTERFACE -> AS_GENERAL)
0x01,        // bTerminalLink
0x01,        // bDelay 1
0x01, 0x00,  // wFormatTag (PCM)

0x0B,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x02,        // bDescriptorSubtype (CS_INTERFACE -> FORMAT_TYPE)
0x01,        // bFormatType 1
0x02,        // bNrChannels (Stereo)
0x02,        // bSubFrameSize 2
0x10,        // bBitResolution 16
0x01,        // bSamFreqType 1
0x00, 0x7D, 0x00,  // tSamFreq[1] 32000 Hz

0x09,        // bLength
0x05,        // bDescriptorType (See Next Line)
0x01,        // bEndpointAddress (OUT/H2D)
0x09,        // bmAttributes (Isochronous, Adaptive, Data EP)
0x84, 0x00,  // wMaxPacketSize 132
0x01,        // bInterval 1 (unit depends on device speed)
0x00,        // bRefresh
0x00,        // bSyncAddress

0x07,        // bLength
0x25,        // bDescriptorType (See Next Line)
0x01,        // bDescriptorSubtype (CS_ENDPOINT -> EP_GENERAL)
0x00,        // bmAttributes (None)
0x00,        // bLockDelayUnits
0x00, 0x00,  // wLockDelay 0

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x02,        // bInterfaceNumber 2
0x00,        // bAlternateSetting
0x00,        // bNumEndpoints 0
0x01,        // bInterfaceClass (Audio)
0x02,        // bInterfaceSubClass (Audio Streaming)
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x02,        // bInterfaceNumber 2
0x01,        // bAlternateSetting
0x01,        // bNumEndpoints 1
0x01,        // bInterfaceClass (Audio)
0x02,        // bInterfaceSubClass (Audio Streaming)
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x07,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x01,        // bDescriptorSubtype (CS_INTERFACE -> AS_GENERAL)
0x06,        // bTerminalLink
0x01,        // bDelay 1
0x01, 0x00,  // wFormatTag (PCM)

0x0B,        // bLength
0x24,        // bDescriptorType (See Next Line)
0x02,        // bDescriptorSubtype (CS_INTERFACE -> FORMAT_TYPE)
0x01,        // bFormatType 1
0x01,        // bNrChannels (Mono)
0x02,        // bSubFrameSize 2
0x10,        // bBitResolution 16
0x01,        // bSamFreqType 1
0x80, 0x3E, 0x00,  // tSamFreq[1] 16000 Hz

0x09,        // bLength
0x05,        // bDescriptorType (See Next Line)
0x82,        // bEndpointAddress (IN/D2H)
0x05,        // bmAttributes (Isochronous, Async, Data EP)
0x22, 0x00,  // wMaxPacketSize 34
0x01,        // bInterval 1 (unit depends on device speed)
0x00,        // bRefresh
0x00,        // bSyncAddress

0x07,        // bLength
0x25,        // bDescriptorType (See Next Line)
0x01,        // bDescriptorSubtype (CS_ENDPOINT -> EP_GENERAL)
0x00,        // bmAttributes (None)
0x00,        // bLockDelayUnits
0x00, 0x00,  // wLockDelay 0

0x09,        // bLength
0x04,        // bDescriptorType (Interface)
0x03,        // bInterfaceNumber 3
0x00,        // bAlternateSetting
0x02,        // bNumEndpoints 2
0x03,        // bInterfaceClass
0x00,        // bInterfaceSubClass
0x00,        // bInterfaceProtocol
0x00,        // iInterface (String Index)

0x09,        // bLength
0x21,        // bDescriptorType (HID)
0x11, 0x01,  // bcdHID 1.11
0x00,        // bCountryCode
0x01,        // bNumDescriptors
0x22,        // bDescriptorType[0] (HID)
0xFB, 0x01,  // wDescriptorLength[0] 507

0x07,        // bLength
0x05,        // bDescriptorType (Endpoint)
0x84,        // bEndpointAddress (IN/D2H)
0x03,        // bmAttributes (Interrupt)
0x40, 0x00,  // wMaxPacketSize 64
0x05,        // bInterval 5 (unit depends on device speed)

0x07,        // bLength
0x05,        // bDescriptorType (Endpoint)
0x03,        // bEndpointAddress (OUT/H2D)
0x03,        // bmAttributes (Interrupt)
0x40, 0x00,  // wMaxPacketSize 64
0x05,        // bInterval 5 (unit depends on device speed)

// 225 bytes
```
Notice how there are multiple descriptors with `bInterfaceNumber` = 1.
This iterator would never get to interface with `bInterfaceNumber` = 3 (the hid class one).

I'll fix it so that the user will get all of them. At minimum this needs to be mentioned in doc string so the user knows to expect more then `num_interfaces` members.

i404788: Hmm yeah seems like we're not accounting for `bAlternateSetting` variations, we could try to parse all descriptors (should account for most edge-cases) but it would be a bit slower. Would definitely be good to mention in docs as reminder of this behaviour.

lynxD: I've made the fix to this and some other problems:
https://github.com/i404788/embassy/pull/2

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,287 @@
+//! USB host driver traits and data types.
+
+use core::time::Duration;
+
+use crate::{EndpointInfo, EndpointType, Speed};
+
+/// Errors returned by [`ChannelOut::write`] and [`ChannelIn::read`]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ChannelError {
```
lynxD, line 10: Am I wrong or what we call channel is called pipe in USB spec?
If so I'd think we should align ourselves with the spec.

embassy-usb/src/host/descriptor.rs
lynxD: @i404788 kind reminder, Could you take a look?

Qix-:
From what I gather this doesn't cover ULPI hosts in e.g. stm32, does it?

embassy-usb/src/host/descriptor.rs
dobrowolski-lukasz: @i404788 Ping

dobrowolski-lukasz:
It's been more than a month since last activity of @i404788 (this PR's author) here. I've created new PR to push this forward.
https://github.com/embassy-rs/embassy/pull/4321

i404788:
@lynxD Thanks for picking it up, I currently (and probably won't anytime soon) have time to work on it, so I'll close my PR in favor of https://github.com/embassy-rs/embassy/pull/4321.

If I do end up with some spare time somehow, I'll help review/resolve pending issues.

Qix-:
Just an update for everyone subscribed to this: https://github.com/embassy-rs/embassy/pull/4321.https://github.com/embassy-rs/embassy/pull/4321 was just merged! :D

PR #4033, opened 2025-04-01
[embassy-executor] Add two new `trace` hooks, improve docs

jamesmunns:
This is an extension of the `trace` interface that adds two new hooks:

1. `_embassy_trace_poll_start` - used to mark the time AFTER executor polling starts, but BEFORE task polling starts. This is usable for measuring overhead of the scheduler algorithm itself.
2. `_embassy_trace_task_end` - used to mark the end of a task, at the time of dropping the contained future and freeing the associated space.

I've also added quite a bit of docs.

EDIT: this is a breaking change, but fine because https://github.com/embassy-rs/embassy/pull/4033#issuecomment-2769199765

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -219,6 +219,9 @@ impl<F: Future + 'static> TaskStorage<F> {
         let mut cx = Context::from_waker(&waker);
         match future.poll(&mut cx) {
             Poll::Ready(_) => {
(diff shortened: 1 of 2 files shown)


jamesmunns:
@Dirbaio removed that section!

lulf (approved):

jamesmunns:
@Dirbaio any objections to me hitting merge?

Dirbaio:
yolo

PR #4037, opened 2025-04-02
mspm0: Fix `set_pf_unchecked()` assertion

optlink:
The current assertion triggers when `pf` is 0 which is required for analog functions.


i509VCB (approved):
Yep little mistake on my behalf.

PR #4039, opened 2025-04-02
stm32/can: remove bus error checking when can bus protocol handling is disabled

Andreychik32:
The `TODO` comment is indeed right.
If protocol exception handling is disabled, it makes no sense to check protocol status register for errors.

Tested on `STM32H503CB`, where without fix simple `read_fd().await` call in a loop would hang MCU indefinitely in error checking loop.

Still can't imagine why FDCAN controller sets bits into PSR register when it's specifically told not to handle bus errors by setting PXHD bit in CCCR, but whatever.

Maybe more complete overhaul for fdcan driver is needed.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -698,7 +698,9 @@ impl RxMode {
             let ts = T::calc_timestamp(T::state().ns_per_timer_tick, ts);
             Some(Ok(Envelope { ts, frame }))
         } else if let Some(err) = T::registers().curr_error() {
-            // TODO: this is probably wrong
+            if T::registers().regs.cccr().read().pxhd() {
+                return None;
+            }
             Some(Err(err))
         } else {
             None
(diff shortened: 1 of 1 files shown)


cschuhen:
@Andreychik32, I'm not convinced about your patch. My my understanding is that the PXHD bit of CCCR only stops you getting Protocol exceptions which is bit 14 of the PSR. I would expect other errors to still take place. This is further explained in section "CAN FD operation" page 1957 of RM 0440 Rev 8. This bit seems to only relate to a specific combination of the FD can bits that is reserved for future use. 

Admittedly, I am looking at the STM32G4 RM but I believe the core of the controller is the same as H5, both being based on Bosch M_CAN.

Generally, I've not know CAN controllers that can let you disable all errors. CAN controllers are compelled to go into an error state when things don't stack up.

What bits exactly are you getting as set in PSR.

Andreychik32:
After all, I guess that my solution is not ideal and different approach is needed.

PR #4042, opened 2025-04-03
Add embassy-imxrt

felipebalbi:
Adds initial support for MIMXRT600 series MCUs from NXP. Subsequent PRs will add more drivers.


embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1056 @@
...
+    /// Trigger on edge.
+    Edge,
+}
+
+#[cfg(feature = "rt")]
+#[interrupt]
+#[allow(non_snake_case)]
+fn GPIO_INTA() {
+    irq_handler(&GPIO_WAKERS);
+}
```
i509VCB, line 64: UM11147 says there is also an `INTB`. Is that extra interrupt not needed?

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1056 @@
...
+
+#[cfg(feature = "rt")]
+fn irq_handler(port_wakers: &[Option<&PortWaker>]) {
+    let reg = unsafe { crate::pac::Gpio::steal() };
+
+    for (port, port_waker) in port_wakers.iter().enumerate() {
+        if port_waker.is_none() {
+            continue;
+        }
```
i509VCB, line 91: Could be the following to eliminate the future `port_waker.unwrap()` lower in this function:
```rust
let Some(port_waker) = port_waker else {
    continue;
};
```

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1056 @@
...
+
+        let stat = reg.intstata(port).read().bits();
+        for pin in BitIter(stat) {
+            // Clear the interrupt from this pin
+            reg.intstata(port).write(|w| unsafe { w.status().bits(1 << pin) });
+            // Disable interrupt from this pin
+            reg.intena(port)
+                .modify(|r, w| unsafe { w.int_en().bits(r.int_en().bits() & !(1 << pin)) });
```
i509VCB, line 99: Is it possible that between clearing the interrupt status and disabling the interupt for the pin, another edge/level event happens which would set the status register to a high value?

It does appear that the interrupt is cleared in the constructor of `InputFuture`, so that would prevent a new instance of InputFuture seeing this event.

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1056 @@
...
+    //
+    // At this point, all GPIO interrupts are masked. No interrupts
+    // will trigger until a pin is configured as Input, which can only
+    // happen after initialization of the HAL
+    unsafe { interrupt::GPIO_INTA.enable() };
+}
+
+mod sealed {
+    pub trait Sealed {}
+}
```
i509VCB, line 134: I imagine more stuff is going to be sealed, probably belongs in `lib.rs`

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1056 @@
...
+        let new_pin = unsafe { self.pin.clone_unchecked() };
+        drop(self);
+        Flex::<SenseEnabled>::new(new_pin)
+    }
+}
+
+/// Input pin
+pub struct Input<'d> {
+    pin: Flex<'d, SenseEnabled>,
+}
```
i509VCB, line 363: Should this implement `Drop` to undo the `set_as_input`? In other hals to preserve that after drop you need to `mem::forget` it

i509VCB (commented):
Initial comments on GPIO, still need to read through the rest.

embassy-imxrt/src/gpio.rs
i509VCB: Same with Output as well, I guess I forgot to click save on that bit of review as well.

embassy-imxrt/src/gpio.rs
felipebalbi: We can route any GPIO to any of those two interrupts. We haven't found a reasonable use for the other one. Guess it will be left for interrupt executors.

embassy-imxrt/src/gpio.rs
felipebalbi: Yup, I'll fix it.

embassy-imxrt/src/gpio.rs
felipebalbi: I'm not sure what would be a good policy here. Keep the last settings or undo everything back to reset state? I guess I'll `mem::forget` it for now.

i509VCB: Yes the pins would return to a floating state. stm32 for example: https://docs.embassy.dev/embassy-stm32/git/stm32u595vj/gpio/struct.Output.html

felipebalbi: Wait a minute, when `Input` dropped, the inner `Flex` will be dropped, which already has a `Drop` impl. This matches what stm32 is doing.

i509VCB: Ah right I didn't see that.

embassy-imxrt/src/gpio.rs
felipebalbi: What is the concern here? That the same event could retrigger due to e.g. bounces? The future is consumed and creating a new one clears any potential left-over events. I don't such opportunity. If I missed the point, let me know.

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1058 @@
...
+    pub fn new(pin: Peri<'d, impl GpioPin>) -> Self {
+        pin.set_function(Function::F0)
+            .disable_analog_multiplex()
+            .disable_input_buffer();
+
+        Self {
+            pin: pin.into(),
+            _sense_mode: PhantomData::<SenseDisabled>,
+        }
+    }
```
i509VCB, line 348: From what I can tell, since both `Flex<SenseDisabled>` and `Flex<SenseEnabled>` both use `new` for the constructor, you will need to specify with a turbofish operator which once you want.

I'd prefer to make it explicit in the function name which mode you have seleted. So a `new_with_sense` for `SenseEnabled` and `new` for `SenseDisabled`.

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1058 @@
...
+
+        pin.block().intpol(pin.port()).modify(|r, w| match level {
+            Level::High => unsafe { w.bits(r.bits() & !(1 << pin.pin())) },
+            Level::Low => unsafe { w.bits(r.bits() | (1 << pin.pin())) },
+        });
+
+        // Enable pin interrupt on GPIO INT A
+        pin.block()
+            .intena(pin.port())
+            .modify(|r, w| unsafe { w.int_en().bits(r.int_en().bits() | (1 << pin.pin())) });
```
i509VCB, line 449: I think these `modify` accesses to be guarded by a `critical_section`.

Say we have 2 instances for different pins in the same port. The first lives on the main "thread" and the other lives in an `InterruptExecutor` (interrupt "thread").

The first pin does a `read` to get the current status of the interrupt on the main thread. An interrupt comes in and execution starts on the interrupt thread. The pin on the interrupt does the RMW operation to enable it's own interrupt.

Then execution returns to the main thread. Since the modify on the first pin was interrupted, the read has the other pin's interrupt disabled. Then first pin writes to enable the interrupt which has cleared the interrupt bit of the pin on the interrupt executor.

embassy-imxrt/src/iopctl.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// The caller MUST ensure valid port and pin numbers are provided,
+    /// and that multiple instances of [`AnyPin`] with the same port
+    /// and pin combination are not being used simultaneously.
+    ///
+    /// Failure to uphold these requirements will result in undefined behavior.
+    ///
+    /// See Table 297 in reference manual for a list of valid
+    /// pin and port number combinations.
+    #[must_use]
+    pub unsafe fn new(port: u8, pin: u8) -> Self {
```
i509VCB, line 212: In other hals we use `steal` for this purpose.

embassy-imxrt/src/iopctl.rs
```diff
@@ -0,0 +1,715 @@
...
+// GPIO pins back into an Output or Input pin specifically.
+embassy_hal_internal::impl_peripheral!(AnyPin);
+
+impl SealedPin for AnyPin {}
+
+embassy_hal_internal::impl_peripheral!(FC15Pin);
+
+impl SealedPin for FC15Pin {}
+
+macro_rules! impl_iopctlpin {
```
i509VCB, line 283: Like the issue mentioned with enabling interrupts, the modify calls should be guarded by a critical section.

embassy-imxrt/src/gpio.rs
i509VCB: Probably not a concern since the InputFuture clears it anyways and interrupt is disabled.

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1058 @@
...
+}
+
+/// Input Sense mode.
+pub trait Sense: Sealed {}
+
+/// Sense Enabled Flex pin.
+///
+/// This is a true flex pin as the input buffer is enabled.
+/// It can sense its own level when even when configured as an output pin.
+pub struct SenseEnabled;
```
i509VCB, line 141:
```suggestion
pub enum SenseEnabled {}
```

So that this type can't be constructed.

embassy-imxrt/src/gpio.rs
```diff
@@ -0,0 +1,1058 @@
...
+impl Sealed for SenseEnabled {}
+impl Sense for SenseEnabled {}
+
+/// Sense Enabled Flex pin.
+///
+/// This is **not** a true flex pin as the input buffer is disabled.
+/// It cannot be turned into an input and it cannot see its own state, but it consumes less power.
+/// Consider using a sense enabled flex pin if you need to read the pin's state or turn this into an input,
+/// however note that **power consumption may be increased**.
+pub struct SenseDisabled;
```
i509VCB, line 151:
```suggestion
pub enum SenseDisabled {}
```

same as above

embassy-imxrt/src/gpio.rs
felipebalbi: argh, makes sense. Fixing.

embassy-imxrt/src/gpio.rs
felipebalbi: Fixed

embassy-imxrt/src/gpio.rs
felipebalbi: Fixed

embassy-imxrt/src/gpio.rs
felipebalbi: Fixed

embassy-imxrt/src/gpio.rs
felipebalbi: Fixed

embassy-imxrt/src/iopctl.rs
felipebalbi: I'll change, sure :)

i509VCB (approved):
LGTM

I would like one other person from embassy to give this a quick skim for anything else.

felipebalbi:
> Looks good, but this also deserves some mention in the book, https://github.com/embassy-rs/embassy/tree/main/docs

Thanks for the reminder. Added it to the docs now :)

lulf (approved):
Thanks!

PR #4044, opened 2025-04-03
STM32: One pulse mode timer

yodaldevoid:
This adds a one pulse mode timer driver. This allows creating an output pulse and/or interrupt after a trigger and a configurable delay.

I have only implemented external triggers as that is what I required. Similarly, I have not implemented output pins.

This was tested on an STM32H533RE.


Dirbaio (approved):
thanks!

PR #4045, opened 2025-04-04
stm32h7rs xspi 

mkj:
This is based on an initial patch by @rickrogers-ionq . It works fine in single SPI mode in my (fairly brief) amount of testing, I haven't got octospi working yet (waiting for a datasheet for the flash part on the nucleo board).

I'm not sure if there's a better way to handle the NCS1/NCS2 switching in https://github.com/embassy-rs/embassy/commit/f094e34efd6c7d083937b11d388270325e4cc00c . When using a single chip, either NCS1 or NCS2 Pin can be provided, with CSSEL register set to the one being used. In future someone could add another interface to allow switching on the fly between two chips (the unused `ncs_alt`).

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -73,7 +73,7 @@ rand_core = "0.6.3"
 sdio-host = "0.5.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "16" }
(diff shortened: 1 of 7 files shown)


Dirbaio (approved):
thanks!

Issue #4030, opened 2025-03-31
executor: WFI in critical section never resumes (qingke)


outfoxxed:
I've been writing a time driver for the WCH CH583 chip, which uses WCH's QingKe V4A (riscv32imac) architecture, and the systick interrupts aren't working correctly under embassy, while they work fine in a standalone project. I've traced this to the WFI instruction inside a critical section linked below.

https://github.com/embassy-rs/embassy/blob/a44abaf7e4562fa5393087fd845bf0d02141325b/embassy-executor/src/arch/riscv32.rs#L64-L74

The critical section implementation in the qingke crate disables all interrupts, which means the WFI will never be interrupted.

https://github.com/ch32-rs/qingke/blob/73a21d1b527327c17bd80b8722d66e10a4fcd46e/src/critical_section_impl.rs#L7-L18

Excerpt from the processor manual:
![Image](https://github.com/user-attachments/assets/4f035b88-a013-4b17-b570-fa554248debe)

Changing the embassy executor snippet above to WFI outside of the critical section results in interrupts working as expected.
```rs
let dowfi = critical_section::with(|_| {
    // if there is work to do, loop back to polling
    // TODO can we relax this?
    if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
        SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
        false
    }
    // if not, wait for interrupt
    else {
        true
    }
});

if dowfi {
    core::arch::asm!("wfi");
}
```

I'm opening an issue for this instead of a PR as I assume there's some reason the WFI is happening inside a critical section, but I don't know what that reason could be.

Dirbaio:
riscv spec says an interrupt while interrupts disabled must still cause WFI to exit (but it's not handled right away, it's handled after exiting the critical section).

Maybe this is a bug in the ch583?

outfoxxed:
`mstatus` behaves exactly the same as `gintenr` here, so its likely a bug in the chip.

I've tested it in depth and found that WFI *does* respond to interrupts, but only if they happen in a short (couple ms or less) window from WFI being called inside a critical section.

Is there a way to work around this?

ExplodingWaffle:
Have you tried the WFITOWFE and SEVONPEND bits in the PFIC? They should make the interrupt behaviour of these cores more well behaved (specifically SEVONPEND states that disabled interrupts should wake the core)

[...]

PR #4046, opened 2025-04-04
executor: add executor selection to #[embassy_executor::main]

outfoxxed:
I ended up making a custom executor for the wch ch58x HAL. (ref https://github.com/embassy-rs/embassy/issues/4030)

When an architecture isn't selected, the embassy_executor::main macro is currently undefined, requiring the user or the HAL to define their own main function. While that is acceptable, I'd like to propose the ability to specify an executor in the macro.

Currently this PR does the bare minimum to allow an executor to be specified, but does not re-export the macro through embassy-executor as I'm not sure what the best way to handle architecture here is. I'm hoping someone can weigh in with a good direction to take this, or reject it outright before I give it too much time. Thanks!

### DIFF
--- a/embassy-executor-macros/src/macros/main.rs
+++ b/embassy-executor-macros/src/macros/main.rs
@@ -16,42 +16,57 @@ enum Flavor {
 pub(crate) struct Arch {
     default_entry: Option<&'static str>,
     flavor: Flavor,
+    executor_required: bool,
 }
 
 pub static ARCH_AVR: Arch = Arch {
     default_entry: Some("avr_device::entry"),
     flavor: Flavor::Standard,
+    executor_required: false,
 };
 
 pub static ARCH_RISCV: Arch = Arch {
     default_entry: Some("riscv_rt::entry"),
     flavor: Flavor::Standard,
+    executor_required: false,
 };
 
 pub static ARCH_CORTEX_M: Arch = Arch {
     default_entry: Some("cortex_m_rt::entry"),
     flavor: Flavor::Standard,
+    executor_required: false,
 };
 
 pub static ARCH_SPIN: Arch = Arch {
     default_entry: None,
     flavor: Flavor::Standard,
+    executor_required: false,
 };
 
 pub static ARCH_STD: Arch = Arch {
     default_entry: None,
     flavor: Flavor::Standard,
+    executor_required: false,
 };
 
 pub static ARCH_WASM: Arch = Arch {
     default_entry: Some("wasm_bindgen::prelude::wasm_bindgen(start)"),
     flavor: Flavor::Wasm,
+    executor_required: false,
+};
+
+pub static ARCH_UNSPECIFIED: Arch = Arch {
+    default_entry: None,
+    flavor: Flavor::Standard,
+    executor_required: true,
 };
 
 #[derive(Debug, FromMeta, Default)]
 struct Args {
     #[darling(default)]
     entry: Option<String>,
+    #[darling(default)]
+    executor: Option<String>,
 }
 
 pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
@@ -112,9 +127,10 @@ pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
         error(&mut errors, &f.sig, "main function must have 1 argument: the spawner.");
     }
 
-    let entry = match args.entry.as_deref().or(arch.default_entry) {
-        None => TokenStream::new(),
-        Some(x) => match TokenStream::from_str(x) {
+    let entry = match (args.entry.as_deref(), arch.default_entry.as_deref()) {
+        (None, None) => TokenStream::new(),
+        (Some(x), _) | (None, Some(x)) if x == "" => TokenStream::new(),
+        (Some(x), _) | (None, Some(x)) => match TokenStream::from_str(x) {
             Ok(x) => quote!(#[#x]),
             Err(e) => {
                 error(&mut errors, &f.sig, e);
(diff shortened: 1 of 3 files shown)


Dirbaio:
looks good! any reason it's still draft?

outfoxxed:
> looks good! any reason it's still draft?

Yes. When arch isn't specified as a crate feature, embassy_executor doesn't re-export the main macro. That means usage currently looks like this, and requires a direct dependency on embassy_executor_macros.
```rs
#[embassy_executor_macros::main_riscv(entry = "qingke_rt::entry", executor = "ch58x_hal::embassy::executor::Executor")]
```
Are you OK with users having a direct dependency on the macros crate, or would you prefer it to be exported in some other way?

Dirbaio:
maybe it's better to always export the macro, then make it emit a compiler error if `arch-xx` is not set *and* `executor = "..."` is not set either.

outfoxxed:
The issue with that is if you set arch-xx, __pender gets defined twice (once in the HAL, once as embassy's default for the arch) and the program can't be linked.

embassy-executor-macros/src/macros/main.rs
```diff
@@ -112,17 +125,44 @@ pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
...
+            TokenStream::new()
+        }
     };
 
+    let executor = match (args.executor.as_deref(), &arch.default_entry) {
+        (None, Entry::Required) => {
+            error(
+                &mut errors,
+                &f.sig,
+                "An executor must be specified when no architecture features are selected",
```
Dirbaio, line 153: It's better if the error specifies *how* the user can fix the issue. Something like:

> No architecture selected for `embassy-executor`. Make sure you've enabled one of the `arch-*` features in your `Cargo.toml`.
> 
> Alternatively, if you want to use a custom executor implementation specify it with the `executor` argument. For example: `#[embassy_executor::main(executor = "some_crate::Executor")]`.

embassy-executor-macros/src/macros/main.rs
```diff
@@ -112,17 +125,44 @@ pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
         error(&mut errors, &f.sig, "main function must have 1 argument: the spawner.");
     }
 
-    let entry = match args.entry.as_deref().or(arch.default_entry) {
-        None => TokenStream::new(),
-        Some(x) => match TokenStream::from_str(x) {
+    let entry = match (&args.entry.as_deref(), &arch.default_entry) {
+        (None, Entry::NotRequired) => TokenStream::new(),
+        (Some(x), _) | (None, Entry::Some(x)) if *x == "" => TokenStream::new(),
```
Dirbaio, line 130: what's the point of allowing `""`?

outfoxxed: A function called directly by assembly or some other initialization code that isn't a full HAL with proc macros

embassy-executor-macros/src/macros/main.rs
```diff
@@ -13,45 +13,58 @@ enum Flavor {
     Wasm,
 }
 
+pub(crate) enum Entry {
+    NotRequired,
+    Required,
+    Some(&'static str),
```
Dirbaio, line 19: this is somewhat strange. What do the 3 cases mean?

Could you instead leave `default_entry` unchanged and add another field `requires_executor` to `Arch`?

embassy-executor-macros/src/macros/main.rs
Dirbaio: yes, but can't the user simply not specify `entry` in that case?

embassy-executor-macros/src/macros/main.rs
outfoxxed: I wasn't aware multi-line errors from proc macros work as well as they do, nice.

embassy-executor-macros/src/macros/main.rs
outfoxxed: I could but it would complicate the pattern match significantly.

outfoxxed:
Improved the error messages

embassy-executor-macros/src/macros/main.rs
outfoxxed: Renamed to `None`, `Required` and `Default`, which might be a little clearer. I can redo it with a boolean if you'd like though.

outfoxxed: Actually I'll just try it and see if it seems ok

Dirbaio: why? there's no need to touch the `entry` code at all. 

If no `arch-*` feature is enabled and `entry="...."` is not specified in the macro, just don't apply any entry macro. We shouldn't error in this case.

embassy-executor-macros/src/macros/main.rs
outfoxxed: The reason I made specifying it a requirement is to show an error message that is easier to understand than a link error from a missing entrypoint, which is what would happen if entry wasn't supplied or defaulted by an arch feature.

outfoxxed:
Removed the entry specification requirement.

embassy-executor-macros/src/macros/main.rs
```diff
@@ -123,6 +139,32 @@ pub fn run(args: TokenStream, item: TokenStream, arch: &Arch) -> TokenStream {
         },
     };
 
+    let executor = match (
+        args.executor.as_deref(),
+        arch.default_entry.as_deref(),
+        arch.executor_required,
+    ) {
+        (None, None, true) => {
```
Dirbaio, line 147: this means on arch unknown if the user doesn't specify the executor but does specify the entry it'll silently use the default instead of erroring. is this correct?

it seems to me it should error.

outfoxxed: Missed that one, thanks

Dirbaio (approved):
thank you!

PR #4057, opened 2025-04-07
mspm0: add uart tests

i509VCB:
This also fixes a bug in the uart clock calculation where it could select an oversampling faster than what the hardware is providing.

Depends on https://github.com/embassy-rs/teleprobe/pull/29


i509VCB:
Merging so I can get per model metapac generation under way.

PR #4058, opened 2025-04-07
chore: update cyw43 bt-hci version

lulf:
The new version contains some fixes to hci serialization.


rhaoio:
Any chance this can get pushed to crates.io too? looks like it's still on the previous version

PR #4063, opened 2025-04-08
Frequency display is now consistent

friedman-ionq:
Some frequency variables were being logged with `variable.0` and some as `variable`.  This PR makes the style consistent as `frequency`.


bugadani:
Given that the [Display impl](https://github.com/embassy-rs/embassy/blob/a23e971d313e8098b8d3a13f3ddf7629aa79a98c/embassy-stm32/src/time.rs#L13) suffixes ` Hz`, this PR makes defmt and log produce different outputs. Would you mind fixing that, please?

bugadani:
Now it'll not display Hz with defmt at all 😅

Dirbaio:
oh oops I was too trigger-happy

PR #4070, opened 2025-04-10
[RP] Fix missing config.composite_with_iads = false; in webusb example

### LINKED ISSUE #4050: rpi2040 webusb example fails
When I run `cargo run --bin usb_webusb` from `examples/rp` I get

```
examples/rp - (main) > cargo run --bin usb_webusb
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.18s
     Running `probe-rs run --chip RP2040 target/thumbv6m-none-eabi/debug/usb_webusb`
      Erasing ✔ [00:00:00] [############################################################################################################################################################################] 24.00 KiB/24.00 KiB @ 64.60 KiB/s (eta 0s )
  Programming ✔ [00:00:00] [############################################################################################################################################################################] 24.00 KiB/24.00 KiB @ 39.48 KiB/s (eta 0s )    Finished in 1.004s
ERROR panicked at 'if composite_with_iads is set, you must set device_class = 0xEF, device_sub_class = 0x02, device_protocol = 0x01'
└─ embassy_usb::builder::{impl#1}::new @ /Users/derenrich/oss/embassy/embassy-usb/src/builder.rs:179
ERROR panicked at /Users/derenrich/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/defmt-1.0.1/src/lib.rs:385:5:
explicit panic
└─ panic_probe::print_defmt::print @ /Users/derenrich/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/panic-probe-0.3.2/src/lib.rs:104
```

but other examples work. For example:

```
examples/rp - (main) > cargo run --bin blinky
   Compiling embassy-rp-examples v0.1.0 (/Users/derenrich/oss/embassy/examples/rp)
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.64s
     Running `probe-rs run --chip RP2040 target/thumbv6m-none-eabi/debug/blinky`
      Erasing ✔ [00:00:00] [############################################################################################################################################################################] 12.00 KiB/12.00 KiB @ 64.24 KiB/s (eta 0s )
  Programming ✔ [00:00:00] [############################################################################################################################################################################] 12.00 KiB/12.00 KiB @ 37.88 KiB/s (eta 0s )    Finished in 0.529s
INFO  led on!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:21
INFO  led off!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:25
INFO  led on!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:21
INFO  led off!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:25
INFO  led on!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:21
INFO  led off!
```

I'm using a bog standard pico 2040 purchased from a reputable dealer. I believe this was caused by the change described here: https://github.com/embassy-rs/embassy/blob/b1179c5090a945d8b15919c99144f157e11d4654/embassy-usb/CHANGELOG.md?plain=1#L12

Manually setting `composite_with_iads` to false in https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/usb_webusb.rs#L52 does eliminate the panic but it still won't enumerate via webusb on my system (but this could be user error).

JuliDi:
Running the usb-webusb example for the rp235x causes the following panic:
```
0.000395 ERROR panicked at /home/user/.cargo/git/checkouts/embassy-c08a80187403f815/6919732/embassy-usb/src/builder.rs:179:13:
if composite_with_iads is set, you must set device_class = 0xEF, device_sub_class = 0x02, device_protocol = 0x01
└─ panic_probe::print_defmt::print @ /home/jdickert/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/panic-probe-1.0.0/src/lib.rs:104 
```
Since the classes and protocol are not set to the necessary values, I think the `composite_with_iads` should simply be set to false in this case.

Also made the change in the RP2340 example.

fixes #4050

### DIFF
--- a/examples/rp/src/bin/usb_webusb.rs
+++ b/examples/rp/src/bin/usb_webusb.rs
@@ -51,12 +51,6 @@ async fn main(_spawner: Spawner) {
     config.max_power = 100;
     config.max_packet_size_0 = 64;
 
-    // Required for windows compatibility.
-    // https://developer.nordicsemi.com/nRF_Connect_SDK/doc/1.9.1/kconfig/CONFIG_CDC_ACM_IAD.html#help
-    config.device_class = 0xff;
-    config.device_sub_class = 0x00;
-    config.device_protocol = 0x00;
-
     // Create embassy-usb DeviceBuilder using the driver and config.
     // It needs some buffers for building the descriptors.
     let mut config_descriptor = [0; 256];
--- a/examples/rp235x/src/bin/usb_webusb.rs
+++ b/examples/rp235x/src/bin/usb_webusb.rs
@@ -51,12 +51,6 @@ async fn main(_spawner: Spawner) {
     config.max_power = 100;
     config.max_packet_size_0 = 64;
 
-    // Required for windows compatibility.
-    // https://developer.nordicsemi.com/nRF_Connect_SDK/doc/1.9.1/kconfig/CONFIG_CDC_ACM_IAD.html#help
-    config.device_class = 0xff;
-    config.device_sub_class = 0x00;
-    config.device_protocol = 0x00;
-
     // Create embassy-usb DeviceBuilder using the driver and config.
     // It needs some buffers for building the descriptors.
     let mut config_descriptor = [0; 256];


Dirbaio:
Does it also work if you remove these instead? 

```rust
    config.device_class = 0xff;
    config.device_sub_class = 0x00;
    config.device_protocol = 0x00;
```

that'd be a simpler fix I think

JuliDi:
That works, yes. No idea about the Windows compatibility stuff. I can change the PR either way.

Dirbaio:
that's great, yes please change it to do that.

> No idea about the Windows compatibility stuff

the comment is not really accurate. The default class/subclass/protocol indicate the device is composite (1 device, N functions), then the webusb code ensures the *function* gets the right class/subclass/protocol for webusb. https://github.com/embassy-rs/embassy/blob/main/embassy-usb/src/class/web_usb.rs#L12-L14

JuliDi:
> that's great, yes please change it to do that.
Done. I replaced it with a single commit removing the three lines in question and the comment.

Issue #4075, opened 2025-04-10
Data spoofing in a CAN frame


RichardWGNR:
Hello,

I am developing a CAN filter on STM32. It is a device that takes transmissions from CAN1 and sends them to CAN2 and vice versa. Some transmissions need to be changed on the fly to activate some functions in the car.

The current implementation of the embassy-stm32 crate does not provide a way to do **byte replacement** in a received transmission. Instead I have to copy data from an existing transmission to a new one.

![Image](https://github.com/user-attachments/assets/91c85185-6847-4329-90dd-32983e556aa3)

This is a very common use case when designing CAN filters, so I propose to give the ability to take a modifiable reference to a slice of transmission data to avoid copy operations and swap bytes via zero-copy abstraction.

For example, it would look like this:

![Image](https://github.com/user-attachments/assets/e165adb2-eb6a-484e-8713-3be6369e40c7)

RichardWGNR:
Consider PR #4076 as irrelevant.
Watch PR #4077 instead.

RichardWGNR:
Resolved in #4077.

PR #4077, opened 2025-04-10
[embassy-stm32] Capability to modify CAN frame data without copying.

RichardWGNR:
Implemented the ability to take a changeable reference to data in a CAN frame.
There are no breaking changes.

Read more in issue #4075.


cschuhen:
Seems reasonable to me. Personally, I would have merged the formatting changeset.

RichardWGNR:
@Dirbaio Thanks for the prompt response! Working with embassy is a pleasure.

PR #4085, opened 2025-04-13
update documentation and examples to mention RP235x

1-rafael-1:
I noticed some places, where I believe RP235x should be mentioned or in examples comments were copypasted referring to RP2040

Not sure about the bootloader thing, only RP2040 mentioned but also I have no idea if this works on RP235x, so I left that alone.


examples/rp235x/src/bin/adc.rs
```diff
@@ -40,8 +40,8 @@ async fn main(_spawner: Spawner) {
 }
 
 fn convert_to_celsius(raw_temp: u16) -> f32 {
-    // According to chapter 4.9.5. Temperature Sensor in RP2040 datasheet
-    let temp = 27.0 - (raw_temp as f32 * 3.3 / 4096.0 - 0.706) / 0.001721;
+    // According to chapter 12.4.6 Temperature Sensor in RP235x datasheet
+    let temp = 27.0 - (raw_temp as f32 * 3.3 / 4096.0 - ..0.706) / 0.0..01721;
```
Dirbaio, line 44: acciedntal change?

1-rafael-1: Oh... whoops... you meant the code not the chapter. No idea how that change happened but it is of course accidental. Will update that!

lulf (approved):

lulf (approved):
Looks good, needs a formatting change according to rustfmt

1-rafael-1:
> Looks good, needs a formatting change according to rustfmt

Editing on smartphone.... should be good now 😅

PR #4089, opened 2025-04-14
feat(embassy-stm32/opamp): Add some STM32G4 opamp usage

IvanLi-CN:
1. The current `buffer_ext` includes a `gain` parameter, which feels a bit odd to me, so I split it into two functions: `buffer_ext` without the `gain` parameter and `pga_ext` with the `gain` parameter.  
2. Added missing gain multipliers for `x32` and `x64`.  
3. Added support for `standalone_dac_int`, `standalone_dac_ext`, `standalone_int`, and `standalone_ext`.  
4. Introduced a calibration method.


Dirbaio (approved):
thank you

PR #4093, opened 2025-04-14
Enable input mode for PWM pins on RP235x and disable it on drop

1-rafael-1:
PWM input does not work on rp235xa. This PR fixes this by enabling input mode if using pwm in input mode

Test result with fix: Can confirm reading correct rate from generated function. Before fix reads 0Hz.

image

Have a good look before merging, the code is based on AI agent suggestion. I am especially unsure if the drop implementation is as it ought to be, we will be setting IE to false even if the PWM was never initialized as input. But I found no convincing way that i liked to read the way it is set up without introducing more code so I left it like it is proposed here.

Here is an excerpt from the rp2350 datasheet backing up the viability of the fix itself: Chapter 9.3, outlining changes wrt rp2040:
image

### DIFF
--- a/embassy-rp/src/pwm.rs
+++ b/embassy-rp/src/pwm.rs
@@ -157,6 +157,11 @@ impl<'d> Pwm<'d> {
             pin.pad_ctrl().modify(|w| {
                 #[cfg(feature = "_rp235x")]
                 w.set_iso(false);
+                #[cfg(feature = "_rp235x")]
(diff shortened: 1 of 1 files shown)


Dirbaio:
thanks! :D

1-rafael-1:
@Dirbaio Merge Build broke, since CI had worked here I assume it is the known flaky behavior thing.

PR #4094, opened 2025-04-14
Canfd improvements

kevswims:
Fixes the error handling in the canfd peripheral to match the bxcan. An error will only be returned once even if it remains present instead of returning the error every time read is called. The error state is tracked using the interrupt flags. If the flag is set then the error is new and is reported.

Also adds a configuration function to set the TX buffer mode to allow usage in FIFO mode.


Dirbaio:
Hello,

PR #4095, opened 2025-04-15
Implement MII for STM32 V1 ethernet peripheral

G33KatWork:
Much like PR #2465 already did for the STM32 V2 Ethernet peripheral, this patch implements the MII interface for the STM32 V1 Ethernet peripheral.

I tested the code on a custom STM32F207 board and I get a link and traffic flows from A to B. So at least for that version of the peripheral, it works. I don't have hardware for the `eth_v1a` version, but I consulted the appropriate datasheet of the SoC (I forgot which one it was - some F1xx I think) which contains this version of the MAC and set the pin modes as described, so that should work as well.


Dirbaio:
thanks! :)

PR #4098, opened 2025-04-15
[embassy-net-nrf91] patch greater rx_data_len size

wieerwill:
Switching from a NRF9160 MCU to the newer NRF9151 there were some changes regarding LTE connections. I used the same code on both boards.

1. the NRF9151 switched it's logic to use the external oscillator by default; this can be changed by setting HFXOSRC to `0` and reboot. In the same step it's also recommended and verified to use `hfclk_source = HfclkSource::Internal` in your init config, e.g.:
    ```
    let mut peripheral_config = embassy_nrf::config::Config::default();
    peripheral_config.hfclk_source = HfclkSource::Internal;
    peripheral_config.lfclk_source = LfclkSource::InternalRC;
    let p = embassy_nrf::init(peripheral_config);
    ```
Except of course if you got an explicit external clock in use. 

2. starting up the modem with correct configuration leads to a endless wait state while the modem starts up. Issue covered up here was a change in the `rx_control_len` from uint `16` to `32`. Changing this leads to usage as usual. 

Tested on the NRF9160-DK and a NRF9151 testboard. 

Done:
- describe errors and solution
- add nrf9151 feature to crate; crate can be used as is in older projects and patched for newer MCU
- patch rx_data_length
- assert_eq got unique error messages
- add feature note to Readme

TODO: examples to test and add to examples folder (possibly other PR in future)

### DIFF
--- a/embassy-net-nrf91/src/lib.rs
+++ b/embassy-net-nrf91/src/lib.rs
@@ -346,8 +346,11 @@ impl StateInner {
                 self.rx_data_list = desc.data_list_ptr;
                 let rx_control_len = unsafe { addr_of!((*self.rx_control_list).len).read_volatile() };
                 let rx_data_len = unsafe { addr_of!((*self.rx_data_list).len).read_volatile() };
-                assert_eq!(rx_control_len, LIST_LEN);
-                assert_eq!(rx_data_len, LIST_LEN);
+                assert_eq!(rx_control_len, LIST_LEN, "control list length mismatch");
+                #[cfg(feature = "defmt")]
+                assert_eq!(rx_data_len, 32, "data list length mismatch");
+                #[cfg(not(feature = "defmt"))]
+                assert_eq!(rx_data_len, LIST_LEN, "data list length mismatch");
                 self.init = true;
 
                 debug!("IPC initialized OK!");
--- a/embassy-net-nrf91/Cargo.toml
+++ b/embassy-net-nrf91/Cargo.toml
@@ -10,8 +10,9 @@ repository = "https://github.com/embassy-rs/embassy"
 documentation = "https://docs.embassy.dev/embassy-net-nrf91"
 
 [features]
-defmt = [ "dep:defmt", "heapless/defmt-03" ]
-log = [ "dep:log" ]
+defmt = ["dep:defmt", "heapless/defmt-03"]
+log = ["dep:log"]
+nrf9151 = []
 
 [dependencies]
 defmt = { version = "0.3", optional = true }
@@ -21,9 +22,9 @@ nrf-pac = "0.1.0"
 cortex-m = "0.7.7"
 
 embassy-time = { version = "0.4.0", path = "../embassy-time" }
-embassy-sync = { version = "0.6.2", path = "../embassy-sync"}
-embassy-futures = { version = "0.1.0", path = "../embassy-futures"}
-embassy-net-driver-channel = { version = "0.3.0", path = "../embassy-net-driver-channel"}
+embassy-sync = { version = "0.6.2", path = "../embassy-sync" }
+embassy-futures = { version = "0.1.0", path = "../embassy-futures" }
+embassy-net-driver-channel = { version = "0.3.0", path = "../embassy-net-driver-channel" }
 
 heapless = "0.8"
 embedded-io = "0.6.1"
--- a/embassy-net-nrf91/README.md
+++ b/embassy-net-nrf91/README.md
@@ -7,3 +7,7 @@ See the [`examples`](https://github.com/embassy-rs/embassy/tree/main/examples/nr
 ## Interoperability
 
 This crate can run on any executor.
+
+## Features
+
+By default the nrf91 crate supports NRF9160 and similar. With newer generation of chips use feature `nrf9151` to support the new data sizes.


Dirbaio:
The modem firmware tells us the list length for rx. It would be better to just use it, and not validating it equals 16 or 32. 

This is better because 
- it doesn't require a Cargo feature
- It'll work if in the future Nordic decides to change the list length to e.g. 42.

It'll need a small refactor to not use a fixed-length array for the rx list. Could you do that?

wieerwill:
@Dirbaio thanks for your fast reply. 

I too first thought that in the code we could just remove the assert_eq but it comes with faults: a change or defect in the modem would go unnoticed; e.g. a faulty value, wrong read address or error in the unsafe access could cause greater errors later on. As the LIST_LEN is used at other parts of the code too, it could cause problems if a length is read that is not divide-able by 8/16/32.

With the feature the check is still in place. 
With the assert error messages any modem change at this part is easy fixable and searchable.
Future Changes of any nRF MCU can be detected with that with ease. And if Nordic decides to change the list length again it can be added with a clear visible and separated feature too. 

If you still suggest to just use the the given list length from the modem, add this comment a 🚀 and i refactor it to not use a fixed-length array.

Dirbaio:
Yes, we really should change it.

There's two lists:

- For TX (ie application -> modem), we decide the length, we allocate it, we tell the modem the length and address. Code for this can use LIST_LEN and fixed-size arrays.
- For RX (ie modem -> application), *the modem* decides the length and allocates it. It tells us the length and address. 

Nordic's lib uses the len provided by the modem. When I initially implemented embassy-net-nrf91, I saw the RX list length seemed to be always 16 and decided to just hardcode it because it was easier.

With modem firmware 2.0.0 (available for nrf9151, not for nrf9160) they changed the list length to 32.

wieerwill:
cleaned and squashed the PR 

with those changes i can access and communicate with both the nrf9160 and nrf9151 modem (verified trough AT commands). Sadly i only get the nrf9160 to establish a connection but not the nrf9151. the nrf9151 loops endless at `stack.wait_config_up()`, to make up for other faults i added the timeout as the on-board modem shuts itself down after 60 seconds.

currently i can not establish a LTE/NB-IoT connection with the nrf9151 and opened a question regarding this in the Nordic DevZone: https://devzone.nordicsemi.com/f/nordic-q-a/122021/nrf9151-replaces-nrf9160-lte-not-working-any-more

.github/ci/doc.sh
```diff
@@ -42,6 +42,7 @@ docserver-builder -i ./embassy-usb-logger -o webroot/crates/embassy-usb-logger/g
 docserver-builder -i ./embassy-usb-synopsys-otg -o webroot/crates/embassy-usb-synopsys-otg/git.zup
 
 docserver-builder -i ./embassy-net -o webroot/crates/embassy-net/git.zup
+docserver-builder -i ./embassy-net-nrf91 -o webroot/crates/embassy-net-nrf91/git.zup
```
wieerwill, line 45: embassy net nrf91 is missing in the embassy docs. this should add it so we have a nice and up to date documentation available online

embassy-net-nrf91/src/context.rs
```diff
@@ -48,6 +48,8 @@ pub enum Error {
     AtParseError,
     /// Error parsing IP addresses.
     AddrParseError,
+    /// Timed out waiting for network attach.
+    AttachTimeout,
```
wieerwill: additionally add a optional Timeout, that if the modem connection timer (usually 60 seconds) clocks out, the modem connection process is handled gracefully and returns a readable error

embassy-net-nrf91/src/context.rs
```diff
@@ -66,7 +68,7 @@ pub struct Status {
     /// Gateway if assigned.
     pub gateway: Option<IpAddr>,
     /// DNS servers if assigned.
-    pub dns: Vec<IpAddr, 2>,
+    pub dns: Vec<IpAddr, 3>,
```
wieerwill: according to docs and comments, three DNS addresses were supposed

embassy-net-nrf91/src/context.rs
```diff
@@ -143,13 +145,17 @@ impl<'a> Control<'a> {
                 .map_err(|_| Error::BufferTooSmall)?;
             let _ = self.control.at_command(op, &mut buf).await;
             // Ignore ERROR which means no pin required
+            trace!("SIM PIN configured");
+        } else {
+            trace!("No SIM PIN to send");
```
wieerwill: more versatile log output if trace is active

embassy-net-nrf91/src/context.rs
```diff
@@ -292,15 +300,31 @@ impl<'a> Control<'a> {
...
 
     async fn wait_attached(&self) -> Result<Status, Error> {
-        while !self.attached().await? {
-            Timer::after(Duration::from_secs(1)).await;
+        // Poll every second, up to 60 seconds, then error out
+        const POLL_INTERVAL: Duration = Duration::from_secs(1);
+        const MAX_TIMEOUT: Duration = Duration::from_secs(60);
+
+        let mut elapsed = Duration::from_secs(0);
+        loop {
```
wieerwill: the modem attach timeout logic

embassy-net-nrf91/src/lib.rs
```diff
@@ -119,14 +117,16 @@ async fn new_internal<'a>(
     let shmem_ptr = shmem.as_mut_ptr() as *mut u8;
 
     const SPU_REGION_SIZE: usize = 8192; // 8kb
-    assert!(shmem_len != 0);
+    trace!("  shmem_ptr = {}, shmem_len = {}", shmem_ptr, shmem_len);
+
+    assert!(shmem_len != 0, "shmem length must not be zero");
```
wieerwill, line 124: return users readable error feedback

embassy-net-nrf91/src/lib.rs
```diff
@@ -202,6 +217,8 @@ async fn new_internal<'a>(
 
         rx_control_list: ptr::null_mut(),
         rx_data_list: ptr::null_mut(),
+        rx_control_len: 0,
+        rx_data_len: 0,
```
wieerwill, line 223: added new variable RX lengths

embassy-net-nrf91/src/lib.rs
```diff
@@ -344,10 +367,14 @@ impl StateInner {
...
+
                 let rx_control_len = unsafe { addr_of!((*self.rx_control_list).len).read_volatile() };
                 let rx_data_len = unsafe { addr_of!((*self.rx_data_list).len).read_volatile() };
-                assert_eq!(rx_control_len, LIST_LEN);
-                assert_eq!(rx_data_len, LIST_LEN);
+
+                trace!("modem control list length: {}", rx_control_len);
+                trace!("modem data    list length: {}", rx_data_len);
+                self.rx_control_len = rx_control_len;
+                self.rx_data_len = rx_data_len;
```
wieerwill, line 376: get and set the RX Lengths we got from the modem

embassy-net-nrf91/src/lib.rs
```diff
@@ -463,14 +490,19 @@ impl StateInner {
 
     fn process(&mut self, list: *mut List, is_control: bool, ch: &mut ch::Runner<MTU>) -> bool {
         let mut did_work = false;
-        for i in 0..LIST_LEN {
+        let max = if is_control {
+            self.rx_control_len
+        } else {
+            self.rx_data_len
+        };
+        for i in 0..max {
```
wieerwill, line 497: either who is in control gets to tell the iteration length

embassy-net-nrf91/src/lib.rs
```diff
@@ -605,11 +635,35 @@ impl StateInner {
...
-                    2 => false,
+                    2 => {
+                        trace!("AT request ack");
+                        false
+                    }
                     // AT response
                     3 => self.handle_resp(msg),
                     // AT notification
-                    4 => false,
+                    4 => {
```
wieerwill: get some more useful tracing logs about AT notifications

wieerwill:
Hi @diondokter you were recommended to me and i saw you wrote the nrf-modem crate and did some embassy work here too. could you please take a look into this?

diondokter:
Hi, I've never looked at the internals of libmodem properly, so I don't have a lot of info on that front. Not sure what I can help with...

wieerwill:
> Hi, I've never looked at the internals of libmodem properly, so I don't have a lot of info on that front. Not sure what I can help with...

thanks for your lightning fast reply. could you verify the changes of this PR work on a NRF9151? i got a hard time testing and figuring out the internals of the modem and NRF and only got a custom PCB instead of the DevBoard therefore i want to make sure this works and does not throw up elsewhere. especially connecting to LTE/NB-IoT Networks

diondokter:
I only have an nRF9160 :(
All the support for other 91-series chips in nrf-modem has been submitted by other people (but libmodem smoothes over them a lot)

lulf:
fyi, I got hold of an nrf9151, so I will give it a spin in the coming days.

embassy-net-nrf91/src/context.rs
```diff
@@ -292,15 +305,34 @@ impl<'a> Control<'a> {
     }
 
     async fn wait_attached(&self) -> Result<Status, Error> {
-        while !self.attached().await? {
-            Timer::after(Duration::from_secs(1)).await;
+        trace!("waiting for attached");
+        // Poll every second, up to 60 seconds, then error out
+        const POLL_INTERVAL: Duration = Duration::from_secs(1);
+        const MAX_TIMEOUT: Duration = Duration::from_secs(60);
```
lulf: This should be configurable, maybe via some kind of config type passed when creating the context. Because in my case it can take longer than 60 secs, and I've seen some cases where it could take up to 20 minutes to attach in testing (probably some kind of rate limiting if you attach too frequently).

lulf (commented):
I finally got around to test it, and it works nicely on my nrf9151-dk. Nice work! I also verified it still works on my nrf9160 DK.

I left one comment regarding the timeout that I think needs to be configurable, as I had to adjust that for the attach to succeed.

Dirbaio:
Thanks for the PR! I've rebased it and removed unrelated things to the main point of the PR (nrf9151)

- Removed the attach timeout. In async Rust it's generally idiomatic to not build timeouts into things, because the user can add their own from outside with `with_timeout` or `select`. 
- Removed URC printing. It's nice to have, feel free to send it again in a separate PR. IMO we should ideally add a proper public API for it insetad of just logging it.
- Kept only some of the extra tracing

PR #4100, opened 2025-04-15
fix: stm32wb55 low power feature did not compile

ckrenslehner:
The low power feature did not compile for the stm32wb55.
Entering sleep is working, but I still have to check if there needs to be added some special handling for the CPU2.


ckrenslehner:
I added a line to the CI file. I hope it's fine like this.

Thanks :-)

Dirbaio:
thanks!

PR #4101, opened 2025-04-16
embassy-boot: remove default features from embassy-embedded-hal

tomasbarton-stemcell:
embassy-embedded-hal enables the time feature by default, which will require you to include the time driver if you use any of the functions that use it (for example spi::blocking), otherwise the binary fails to link.


Dirbaio:
I'd prefer to make the `time` feature not default. Default features are a footgun precisely because this reason, too easy to forgot to disable them.

Dirbaio:
https://github.com/embassy-rs/embassy/pull/4455

PR #4104, opened 2025-04-17
[embassy-usb-dfu] Add verification support

Gerharddc:
This PR makes the ed25519-based verification support offered by `embassy-boot` available for use in `embassy-usb-dfu`. This will allow devices offering update support via USB DFU to verify that updates are correctly signed before installing them. Additionally, the PR adds an example showing how to use this new feature.


Gerharddc:
I still need to verify that this works on real hardware and update the README.md for the stm32wb-dfu-verify example but would appreciate to already get feedback on the implementation.

PR #4106, opened 2025-04-17
fix: core was not recognized and 1-step lock did not work

ckrenslehner:
The Core was not recognized properly, and the 1-step lock did not work, because the match was wrong.

@Dirbaio had the idea of using the build feature flags instead of reading the register for identifying the Core. 
How would I specify such features for CM7 and CM4 for e.g. stm32h745?

### DIFF
--- a/embassy-stm32/src/hsem/mod.rs
+++ b/embassy-stm32/src/hsem/mod.rs
@@ -4,17 +4,96 @@ use embassy_hal_internal::PeripheralType;
 
 use crate::pac;
 use crate::rcc::RccPeripheral;
-// TODO: This code works for all HSEM implemenations except for the STM32WBA52/4/5xx MCUs.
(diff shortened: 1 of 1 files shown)


Dirbaio:
you can add here another line https://github.com/embassy-rs/embassy/blob/443ffd8b6df67844208ceec58d5443e01b99a159/embassy-stm32/build.rs#L1922

like this

```rust
cfgs.push(format!("core_{}", core));
```

so you can do `cfg(core_cm4)`

ckrenslehner:
Ok thanks I will add the cfgs

Dirbaio:
Have you seen #4092 ? seems you guys are working on the same thing.

ckrenslehner:
I just updated this, as this is a stepping stone to #4116

Dirbaio:
Hello,

PR #4107, opened 2025-04-17
Add CMSIS-DAP driver

bugadani:
Because why not. I managed to extract this from eprobe, and since I have two copies in my toys, I thought we might as well try and add it to embassy-usb.

### DIFF
--- a/embassy-usb/src/class/cmsis_dap_v2.rs
+++ b/embassy-usb/src/class/cmsis_dap_v2.rs
@@ -1,0 +1,101 @@
+//! CMSIS-DAP V2 class implementation.
+
+use core::mem::MaybeUninit;
+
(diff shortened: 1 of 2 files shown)


embassy-usb/src/class/cmsis_dap_v2.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub struct CmsisDapV2Class<'d, D: Driver<'d>> {
+    _interface: InterfaceNumber,
+    _name: StringIndex,
+    read_ep: D::EndpointOut,
+    write_ep: D::EndpointIn,
+}
+
+impl<'d, D: Driver<'d>> CmsisDapV2Class<'d, D> {
+    /// Creates a new CdcAcmClass with the provided UsbBus and `max_packet_size` in bytes. For
```
Dirbaio, line 45: CdcAcmClass

PR #4109, opened 2025-04-18
STM32 CAN: Make `data()` return correct length slice

adamgreig:
Currently `data()` always returns an 8- or 64-byte slice, even though the actual CAN frame contains less data. The remaining bytes will be zero, so the user must check the frame DLC to see how many bytes were actually sent by the remote end. With this change the slice is the same length as was transmitted, with no extra bytes.

Additionally removed the `ClassicData::MAX_DATA_LEN` constant because almost all instances were using a hardcoded `8` anyway, and the `FdData` only ever uses `64` and doesn't have an equivalent `MAX_DATA_LEN`. Happy to make everything use `MAX_DATA_LEN` instead if that's preferred.

Finally fixed a missing cfg gate on a defmt call that broke the build on G4 without defmt feature.

### DIFF
--- a/embassy-stm32/src/can/bxcan/registers.rs
+++ b/embassy-stm32/src/can/bxcan/registers.rs
@@ -299,9 +299,9 @@ impl Registers {
         mb.tdtr().write(|w| w.set_dlc(frame.header().len() as u8));
 
         mb.tdlr()
-            .write(|w| w.0 = u32::from_ne_bytes(unwrap!(frame.data()[0..4].try_into())));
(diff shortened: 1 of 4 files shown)


Dirbaio:
seems HIL tests are failing

adamgreig:
Oops, missed that `data()` was being used internally in a way that assumed it was always 8 bytes long. Fixed.

cschuhen:
Hi guys, I saw this PR was stuck for a week. I've pushed a fix  as https://github.com/embassy-rs/embassy/pull/4137 . In it I just make raw_data public.

PR #4113, opened 2025-04-18
Make the nrf Twim RAM buffer a instance variable instead of stack allocated

alexmoon:
This prevents the ram buffer from being duplicated in every future that calls a Twim transaction method. In my firmware this ended up saving about 4kb of ram.

I can implement this for the other peripherals that implement copying into ram buffers if this looks good to you.

### DIFF
--- a/embassy-nrf/src/twim.rs
+++ b/embassy-nrf/src/twim.rs
@@ -4,7 +4,6 @@
 
 use core::future::{poll_fn, Future};
 use core::marker::PhantomData;
-use core::mem::MaybeUninit;
 use core::sync::atomic::compiler_fence;
 use core::sync::atomic::Ordering::SeqCst;
 use core::task::Poll;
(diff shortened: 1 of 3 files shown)


lulf (approved):
Looks good to me. It's a bit clunky to size this buffer correctly perhaps, is there a way to simplify that?

alexmoon:
It just needs to be large enough for the largest write operation your firmware sends from flash. I don't think there's any way for `embassy-nrf` to help with that. We could add a `usize` to the `RAMBufferTooSmall` error variant so the error message could say what size buffer was needed? Though that would increase the size of the error enum.

lulf:
> It just needs to be large enough for the largest write operation your firmware sends from flash. I don't think there's any way for `embassy-nrf` to help with that. We could add a `usize` to the `RAMBufferTooSmall` error variant so the error message could say what size buffer was needed? Though that would increase the size of the error enum.

Yeah, I don't like that option either... But maybe at least some guidance in the rustdoc for the new() on the sizing of the buffer?

lulf (approved):
Thanks!

PR #4114, opened 2025-04-18
Add thread-safe reference-counting pointer (Arc) implementation

AlixANNERAUD:


lulf:
IMHO this is outside the scope of embassy-sync which deals with no_std synchronization primitives. 

Besides, if you have alloc, why not use the Arc from std instead? I think you only need alloc for that anyway

AlixANNERAUD:
> IMHO this is outside the scope of embassy-sync which deals with no_std synchronization primitives.
> 
> Besides, if you have alloc, why not use the Arc from std instead? I think you only need alloc for that anyway


This is quite within the scope of embassy-sync, as it’s a synchronization primitive designed specifically for no_std environments. The standard Arc you’re referring to is part of std, not just alloc, because it relies on atomic operations for reference counting. These atomic operations are only available on targets that support atomics, which isn’t always the case in embedded or no_std contexts.

So even if you have alloc, you can’t use std::sync::Arc unless your target supports atomics and you’re linking against std. embassy-sync provides alternatives that are safe to use in those constrained environments.

lulf:
In that case, 

> > IMHO this is outside the scope of embassy-sync which deals with no_std synchronization primitives.
> > Besides, if you have alloc, why not use the Arc from std instead? I think you only need alloc for that anyway
> 
> This is quite within the scope of embassy-sync, as it’s a synchronization primitive designed specifically for no_std environments. The standard Arc you’re referring to is part of std, not just alloc, because it relies on atomic operations for reference counting. These atomic operations are only available on targets that support atomics, which isn’t always the case in embedded or no_std contexts.
> 

The reason I don't think it belongs there is that generally primitives that require alloc are not in embassy-sync, it provides no-std and primitives that doesn't require alloc.

> So even if you have alloc, you can’t use std::sync::Arc unless your target supports atomics and you’re linking against std. embassy-sync provides alternatives that are safe to use in those constrained environments. 

~~Why not base that on alloc + https://crates.io/crates/portable-atomic then? Then combine that with https://crates.io/crates/embedded-alloc to get alloc with no_std. That way you'd be able to provide a free standing implementation that'd work without embassy-sync.~~ I found this https://docs.rs/portable-atomic-util/latest/portable_atomic_util/struct.Arc.html

AlixANNERAUD:
Ok that totally makes sense.

Issue #3615, opened 2024-12-05
[embassy-rp] pwm::Pwm drop leaves gpios in incorrect state


Jojo-1000:
Hi, I am using a Pwm to control some motors on Pico 2.

I noticed that when the Pwm goes out of scope, it leaves them running. I looked at the drop implementation and there seems to be a difference between `Pwm` and `PwmOutput`. Using the two channels split as `PwmOutput` works as expected. Here is the implementation for reference:

This one works as expected:
https://github.com/embassy-rs/embassy/blob/af02310edc60c568368d2e8805bb07a0797723a8/embassy-rp/src/pwm.rs#L398-L427

This one leaves the pins in an incorrect state, the part about pull down is missing:
https://github.com/embassy-rs/embassy/blob/af02310edc60c568368d2e8805bb07a0797723a8/embassy-rp/src/pwm.rs#L485-L495

## Reproduction
Unexpected:
```rs
{
    let mut c = embassy_rp::pwm::Config::default();
    let pwm_freq = 20_000;
    let clock_freq = embassy_rp::clocks::clk_sys_freq();
    c.top = (clock_freq / pwm_freq) as u16 - 1;
    
    let mut pwm = embassy_rp::pwm::Pwm::new_output_ab(pwm_slice6, pin_11, pin_10, c.clone());
    pwm.set_duty_cycle_fully_off().unwrap();
}
// There is an output signal
```
Expected:
```rs
{
    let mut c = embassy_rp::pwm::Config::default();
    let pwm_freq = 20_000;
    let clock_freq = embassy_rp::clocks::clk_sys_freq();
    c.top = (clock_freq / pwm_freq) as u16 - 1;
    
    let pwm = embassy_rp::pwm::Pwm::new_output_ab(pwm_slice6, pin_11, pin_10, c.clone());
    let (pwm_a,pwm_b) = pwm.split();
    pwm_a.unwrap().set_duty_cycle_fully_off().unwrap();
    pwm_b.unwrap().set_duty_cycle_fully_off().unwrap();
}
// There is no output signal
```

1-rafael-1:
@vivi202 since this was probably my fault i today tried to reproduce this - without success. I used gpio25 for convenience, but no way could i see the led still lighting up after leaving scope or calling drop. Can You still repro?

vivi202:
hi i'm not the autor of this issue i implemented `pwm.split()`. 
i wrote a fix for this issue but i did not have the time to  test.

[...]

PR #4115, opened 2025-04-18
pwm: enable pull-down resistors for pins in Drop implementation

1-rafael-1:
Adresses #3615, appling a fix written by @vivi202 (I can take no credits!)

I could not reproduce the issue, but the reasoning behind the fix seems plausible. I tested the change by setting up and dropping Pwm, which to me behaves as expected.

After setting up and 50% duty cycle:
image

After dropping:
image


lulf (approved):
Thanks!

1-rafael-1:
@lulf could I ask you to run the queue again?

PR #4116, opened 2025-04-19
DRAFT: stm32wb55 - sync cpu2 with hsem when entering sleep

ckrenslehner:
I did not find where to mark this as draft, but here is a basic implementation of using the HSEM for syncing the with the CPU2 when entering sleep.
The stm32wb55 requires this when accessing some peripherals

Among these are: 
- RNG
- Crypto
- Stop Mode
- Flash

The concrete flow charts for entering and exiting STOP mode are in this [document](https://www.st.com/content/ccc/resource/technical/document/application_note/group1/43/ea/2f/dc/10/a3/46/e6/DM00598033/files/DM00598033.pdf/jcr:content/translations/en.DM00598033.pdf)
Page 23f.

I have some general questions:
- How should I handle the HSEM creation
- Now its created via the init and then passed to the executor again similar to the RTC
- The HSEM is needed in other places too. But with other underlying HSEM IDs. Should I split the HSEM into separate ID-Structs or how could I model such usecase in rust/embassy the best way?

I would greatly appreciate feedback on how to model this "correctly" :-)

### DIFF
--- a/embassy-stm32/src/hsem/mod.rs
+++ b/embassy-stm32/src/hsem/mod.rs
@@ -4,17 +4,96 @@ use embassy_hal_internal::PeripheralType;
 
 use crate::pac;
 use crate::rcc::RccPeripheral;
-// TODO: This code works for all HSEM implemenations except for the STM32WBA52/4/5xx MCUs.
+// TODO: This code works for all HSEM implementations except for the STM32WBA52/4/5xx MCUs.
 // Those MCUs have a different HSEM implementation (Secure semaphore lock support,
 // Privileged / unprivileged semaphore lock support, Semaphore lock protection via semaphore attribute),
 // which is not yet supported by this code.
 use crate::Peri;
 
 /// HSEM error.
-#[derive(Debug)]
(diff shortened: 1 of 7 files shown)


ckrenslehner:
Okay I also realized that keeping the HSEM initialized all the time needs to be avoided. Otherwise, entering the stop mode is not possible because, of the RCC init from the HSEM Module.

So to enter sleep the low power executor needs to init the HSEM peripheral on the inside to sync with CPU 2.

donghoonpark:
Is there any progress with #4092? I'm the one who is desperately waiting this feature and if you don't mind I have willing to test it in my hardware. Is it working stop mode implemenation with cpu2 network feature?

ckrenslehner:
Hi! I did not get any feedback anymore. No there is no progress. If you could test it on your hardware this would be great! :-)

xoviat:
Are you still open to working on this?

ckrenslehner:
I do not have access to the hardware anymore, unfortunately!

Dirbaio:
Hello,

PR #4119, opened 2025-04-21
nrf53: add WDT support

### LINKED ISSUE #480: nrf53: add WDT support
There's two watchdogs WDT0, WDT1, while the existing driver assumes only one named WDT. Needs some refactor, should be easy.

0e4ef622:
Closes #480.

### DIFF
--- a/embassy-nrf/src/wdt.rs
+++ b/embassy-nrf/src/wdt.rs
@@ -120,7 +120,7 @@ impl<T: Instance> Watchdog<T> {
 
         let mut handles = [const { WatchdogHandle { index: 0 } }; N];
         for i in 0..N {
-            handles[i] = unsafe { WatchdogHandle::steal(T::INDEX, i as u8) };
+            handles[i] = unsafe { WatchdogHandle::steal::<T>(i as u8) };
             handles[i].pet();
         }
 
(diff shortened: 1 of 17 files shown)


embassy-nrf/src/wdt.rs
```diff
@@ -178,25 +189,50 @@ impl WatchdogHandle {
...
+    const REGS: pac::wdt::Wdt;
+}
+
+/// WDT instance.
+#[allow(private_bounds)]
+pub trait Instance: SealedInstance + PeripheralType + 'static + Send {
+    /// Interrupt for this peripheral.
+    type Interrupt: interrupt::typelevel::Interrupt;
+    /// Index of the watchdog instance.
+    const INDEX: u8;
```
Dirbaio, line 225: can you put `INDEX` in `SealedInstance`, so we don't make it public?

0e4ef622: I had it public because `WatchdogHandle::steal` is also public and it takes the instance index. Actually I could make `steal` take the instance as a type parameter instead.

Dirbaio: ah right . yeah that works.

Dirbaio (approved):
thanks!

PR #4120, opened 2025-04-22
[embassy-usb-dfu] Add generic DFU marking interface

Gerharddc:
This commit adds an interface that allows users to customise how the bootloader is informed that DFU mode is needed on the next boot.


lulf (approved):
Thanks!

PR #4123, opened 2025-04-23
Manually implement the future for `with_timeout`

alexmoon:
Using `async fn` for these functions causes the memory for `fut` to be double-allocated due to rust-lang/rust#62958. Switching to a manual implementation of the future prevents the double-allocation which may be significant if the future is large.

### DIFF
--- a/embassy-time/src/timer.rs
+++ b/embassy-time/src/timer.rs
@@ -1,8 +1,7 @@
 use core::future::{poll_fn, Future};
-use core::pin::{pin, Pin};
+use core::pin::Pin;
 use core::task::{Context, Poll};
 
-use futures_util::future::{select, Either};
 use futures_util::stream::FusedStream;
 use futures_util::Stream;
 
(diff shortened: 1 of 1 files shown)


embassy-time/src/timer.rs
```diff
@@ -46,24 +43,50 @@ pub trait WithTimeout {
...
-    async fn with_deadline(self, at: Instant) -> Result<Self::Output, TimeoutError> {
-        with_deadline(at, self).await
+    fn with_deadline(self, at: Instant) -> UntilTimer<Self> {
+        with_deadline(at, self)
+    }
+}
+
+/// Future for the [`with_timeout`] and [`with_deadline`] functions.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct UntilTimer<F> {
```
Dirbaio, line 69: could you rename this to `TimeoutFuture`? 

It makes sense to call `xxxFuture` things that implement Future

Dirbaio:
Thanks!

ArXen42:
Noticed that after updating embassy with this PR merged my firmware size dropped by a few % for both flash and RAM.
Neat!

PR #4124, opened 2025-04-24
Add function to allow re-init rcc config for stm32

mickem:
I had a need to configure an external clock generator and then switch to it which means I needed to call init and then configure RCC.
After som input from the matrix chat this is what I came up with.

Not sure if this makes sense, and I did not find any test so I have not added tests.
Might make sense to add an example but as a meaningful example would require a specific external clock generator I am unsure how to proceed but it would be possible to add:

```rust
#[embassy_executor::main]

    // Configure the external clock here

    {
        use embassy_stm32::rcc::*;
        config.rcc.hse = Some(Hse {
            freq: mhz(8),
            mode: HseMode::Bypass,
        });
        config.rcc.pll = Some(Pll {
            src: PllSource::HSE,
            prediv: PllPreDiv::DIV1,
            mul: PllMul::MUL9,
        });
        config.rcc.sys = Sysclk::PLL1_P;
        config.rcc.ahb_pre = AHBPrescaler::DIV1;
        config.rcc.apb1_pre = APBPrescaler::DIV2;
        config.rcc.apb2_pre = APBPrescaler::DIV1;
    }
    embassy_stm32::reinitialize_rcc(config);

    info!("end program");
}
```

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -616,3 +616,31 @@ fn init_hw(config: Config) -> Peripherals {
         p
     })
 }
+
+
+/// Re-initialize the `embassy-stm32` clock configuration with the provided configuration.
+///
+/// This is useful when you need to alter the CPU clock after configuring peripherals.
+/// For instance, configure an external clock via spi or i2c.
+///
+/// Please not this only re-configures the rcc and the time driver (not GPIO, EXTI, etc).
+///
+/// This should only be called after `init`.
+#[cfg(not(feature = "_dual-core"))]
+pub fn reinitialize_rcc(config: Config) {
+    critical_section::with(|cs| {
+        unsafe {
+            rcc::init(config.rcc);
+
+            // must be after rcc init
+            #[cfg(feature = "_time-driver")]
+            time_driver::init(cs);
+
+            #[cfg(feature = "low-power")]
+            {
+                crate::rcc::REFCOUNT_STOP2 = 0;
+                crate::rcc::REFCOUNT_STOP1 = 0;
+            }
+        }
+    })
+}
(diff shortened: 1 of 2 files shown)


mickem:
Build fail due to the critical section variable is not used in (I assume here) som configurations. Not sure of a good way to handle this I could rename the variable _cs or I could split the code but since it is a closure it is hard to add feature toggles on it...

Dirbaio:
Thanks!

- Could you make the function `embassy_stm32::rcc:reinit()` instead of `embassy_stm32::reinitialize_rcc()`?
- It should take only the RCC config, not the whole embassy-stm32 config.
- Could you deduplicate the code inside the CS? This way if something gets added there in the future we ensure both init and reinit stay in sync.

embassy-stm32/src/lib.rs
```diff
@@ -600,19 +601,38 @@ fn init_hw(config: Config) -> Peripherals {
...
+/// Re-initialize the `embassy-stm32` clock configuration with the provided configuration.
+///
+/// This is useful when you need to alter the CPU clock after configuring peripherals.
+/// For instance, configure an external clock via spi or i2c.
+///
+/// Please not this only re-configures the rcc and the time driver (not GPIO, EXTI, etc).
+///
+/// This should only be called after `init`.
+#[cfg(not(feature = "_dual-core"))]
+pub fn reinit(config: rcc::Config) {
```
Dirbaio, line 620: could you move this to the RCC module so it's called as `embassy_stm32::rcc:reinit()` as I said in my last comment and not `embassy_stm32::reinit()`?

mickem: Sorry, I did not see the rcc, please beware that we now have two `rcc::init*` functions which is somewhat awkward.
Also not 100% sure if the `crate::time_driver::init(_cs)` should also be moved but given your feedback I assume this is what we want.

One option for the two `rcc:init` function would be to wrap the old `rcc::init` inside the new `rcc::init_rcc` but unsure if this is what we want.

Dirbaio (approved):
thank you!

PR #4125, opened 2025-04-24
STM32: Configurable bank support

### LINKED ISSUE #2309: STMG4 Dual Bank Flash support
I have noticed that STMG4 support was added a few days ago:

https://github.com/embassy-rs/embassy/pull/2294

However, this PR is not going to work correctly on dual bank flash MCUs such as STM32G474VE that I work with. The reason is that dual bank mode uses 2KB page size while single bank uses 4K pages.

image

I worked around this issue by deleting two pages on each call to `blocking_erase_sector` when I was implementing embassy bootloader several weeks ago. Additionally errata 2.2.2 should be implemented (I reused the same code from `f4.rs/save_data_cache_state()` as the workaround procedure is the same). Other functions beside erasing the sector work without changes.

(btw I have not tested the PR, however I took the same approach as the author and reused the code from `g0.rs` so I assume that there are the same issues I faced.)

diondokter:
It builds locally, but not sure how I can make the metapac work here without it being merged there since the stm32-data-generated only exists for main...

https://github.com/embassy-rs/stm32-data/pull/596

To make this work, I added two features 'single-bank' and 'dual-bank'. They are optional. If a chip can't configure this, the user doesn't have to select one.
If a chip does have multiple flash options, selecting one of the features is required. There's also a cfg being set so the flash drivers knows it needs to check the configured settings.

The config system is generic, ~but it's only implemented for the G4 in this PR~.

An alternative approach is going the same route as the dual core settings where you get a STM32* feature for single bank and one for dual bank. Let me know if that's more desirable.

closes #2309


diondokter:
Alright, I've implemented the configurable flash support for all STM chips I could find that have that feature in one way or another

diondokter:
bender run

jamesmunns:
bender run

bugadani:
bender run

jamesmunns (approved):
Fake approval to see if that makes bender happy, don't actually merge

jamesmunns:
bender run

diondokter:
Alright, this should be finished.

I found there was some dual bank implementation for the F4 series. So I could rip that out. This makes the PR net negative LOC! :D

The F4 & F7 series are weird with their banks and sectors, but they should work.
For anything that runs on the HIL I've added dual/single-bank feature with which they should be factory programmed, so should all remain working.

diondokter:
Hmmm well, I'm sure I didn't break the uart and that it's flaky somehow

lulf (commented):
Looks good to me, I've merged https://github.com/embassy-rs/stm32-data/pull/596 so please update this PR with the correct ref.

diondokter:
Alright, that should be it!

PR #4126, opened 2025-04-24
[embassy-rp] Add disable and enable tx and rx functions for UART

marcemmers:
When writing some Modbus code on top of the RP2040 UART I encountered break and framing errors. I traced this back to the fact that my hardware pulls down the RX pin whenever I am transmitting data.

The UART peripheral has support for disabling RX or TX without influencing the other. It seemed like a good addition to make this available.

I thought about creating a type that freezes the UartRx or UartTx so you cannot have it disabled and try to receive or send respectively but could not figure it out since I am still learning rust.

Also, it is my first PR here so some guidance might be necessary, thanks for that!

### DIFF
--- a/examples/rp/src/bin/uart_rs485.rs
+++ b/examples/rp/src/bin/uart_rs485.rs
@@ -1,0 +1,62 @@
+//! This example shows how to use UART (Universal asynchronous receiver-transmitter) in the RP2040 chip.
+//!
+//! No specific hardware is specified in this example. Only output on pin 0 and pin 2, and input on pin 1 is tested.
+//! The Raspberry Pi Debug Probe (https://www.raspberrypi.com/products/debug-probe/) could be used
+//! with its UART port.
+
+#![no_std]
+#![no_main]
+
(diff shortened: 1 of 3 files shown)


examples/rp/src/bin/uart_rs485.rs
```diff
@@ -0,0 +1,62 @@
+//! This example shows how to use UART (Universal asynchronous receiver-transmitter) in the RP2040 chip.
+//!
+//! No specific hardware is specified in this example. Only output on pin 0 and pin 2, and input on pin 1 is tested.
+//! The Raspberry Pi Debug Probe (https://www.raspberrypi.com/products/debug-probe/) could be used
+//! with its UART port.
```
9names, line 5: Please update this file comment, it's not correct - I assume you copied this example and forgot to update it

9names (changes requested):
Please add an entry to the embassy-rp changelog for this change.

Dirbaio:
Hello,

PR #4127, opened 2025-04-24
Update opamp code to current stm32-metapac

skoe:
Some trivial enums have been removed from the OpAmp API in stm32-metapac, this PR updates the HAL accordingly.


Dirbaio:
thank you!

PR #4128, opened 2025-04-24
[embassy-rp] Rewrite PIO onewire implementation

marcemmers:
I encountered some issues testing the current implementation that led me to spent some time rewriting it.

This new implementation has a couple of improvements over the current one:
- It does not block if there is no device present on the bus
- It allows for better control when to perform a reset. The reset also performs the presence detection.
- There is a implementation of the search algorithm, this allows you to find addresses of the connected devices.
- The clock is now actually calculated based on the sys clock frequency

The only downside in my opinion is the unsafe calls to make the pio jump to the right instruction. But they should be safe since the state machine is disabled when performing the jumps.

Tested this with 0, 1 and 2 devices connected using the example program.

### DIFF
--- a/embassy-rp/src/pio_programs/onewire.rs
+++ b/embassy-rp/src/pio_programs/onewire.rs
@@ -12,7 +12,6 @@ pub struct PioOneWireProgram<'a, PIO: Instance> {
     prg: LoadedProgram<'a, PIO>,
     reset_addr: u8,
     next_bit_addr: u8,
-    search_addr: u8,
 }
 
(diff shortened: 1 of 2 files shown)


jamesmunns:
CC @kalkyl because I think you wrote the original version :)

kalkyl:
Nice! Yes my original implementation was MVP, only what i needed for a project, so these improvements are very welcome! Looks all good to me (i have no easy way to test this on my hardware since that project is deployed in the wild since long)

jamesmunns:
bender run

PR #4129, opened 2025-04-25
Added access to internal registers and setup of channels to allow working with triggered ADC conversions

mickem:
I tried to model this after how the low level timer works.
But I found no way to convert the internal regs to pac regs so instead I opted for allowing access to the internal regs.

I guess this might also be considered controversial as it allows external access to internal regs. But it makes extensions easier and there is a lot of modes in the ADCs in the stm so I think it might be warranted. I for instance this to configure triggering as well as reading from multiple ADC:s

Like so:
```rust
    adc.regs_cr2().modify(|r| r.set_exten(Exten::RISING_EDGE));
    adc.regs_cr2().modify(|r| r.set_extsel(0b1110));
```

It also splits the blocking_read function in two as I want to setup the channel but not convert as I convert based on triggers.

Not sure if this is a good or bad idea so please don't hesitate to shoot it down or suggest better alternatives :)

### DIFF
--- a/embassy-stm32/src/adc/v2.rs
+++ b/embassy-stm32/src/adc/v2.rs
@@ -1,10 +1,11 @@
+use stm32_metapac::adc::regs::{Cr1, Cr2, Dr, Sr};
+use stm32_metapac::common::{Reg, R, RW};
+
 use super::blocking_delay_us;
 use crate::adc::{Adc, AdcChannel, Instance, Resolution, SampleTime};
 use crate::peripherals::ADC1;
 use crate::time::Hertz;
 use crate::{rcc, Peri};
-use stm32_metapac::adc::regs::{Cr1, Cr2, Dr, Sr};
-use stm32_metapac::common::{Reg, R, RW};
 
 mod ringbuffered_v2;
 pub use ringbuffered_v2::{RingBufferedAdc, Sequence};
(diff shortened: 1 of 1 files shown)


Dirbaio:
I don't think we should do this. Reasons:

- I would rather have proper support for the different ADC modes to the HAL. Encouraging raw register access incentivizes people to not do that.
- This adds raw register access to the main, high-level ADC driver. The user is expected to mess with the hardware state behind the driver's back and still use the high-level driver to acquire readings. This is very brittle because it makes users rely on implementation detalis. For example if the user sets a special bit to enable some behvaior then does a read, and later the read function gets changed to accidentally reset that bit (e.g. does a write instead of modify) the user's code stops working. Note with timers you use *only* low-level register access, or *only* high-level drivers like PWM, you can't mix due to this reason.
- Officially the PAC is "unstable", ie a minor embassy-stm32 upgrade can do a major stm32-metapac upgrade. Reexporting regs breaks that. They should at least be gated with the `unstable-pac` feature.

Also note raw register access is already possible with `embassy_stm32::pac::ADC3.whatever().write(|w| ...)`.

mickem:
While I think it is noble to wait for proper support. Currently many objects in the HAL lacks "proper support" for many things. So waiting makes embassy hard to use in the near term.

If you look at how stm32 (which I am coming from) have solved this they have "user code" where the user can, and is expected yo, customize things.

And while I can access them separately via PAC that looses the oo approach one is used to in an oo language.

I.e. 
```rust
// Init code
let mut adc = Adc::new(p.ADC1);

// Somewhere else in the code
adc.regs_cr2().modify(|r| r.set_exten(Exten::RISING_EDGE));

// versus via pac where I need to know which instance is tied to which channel
embassy_stm32::pac::ADC1.cr2().write(|w| w.set_exten(Exten::RISING_EDGE));
```

But I hear you and I will close this PR. That said I think lack of access to registers in embassy makes embassy very hard to use in real world scenarios **now** and unfortunately less viable of an option for the company I work for.

PR #4132, opened 2025-04-25
Dead-time computation in complementary PWM fixed

GreatKet:
Required DTG[7:5] bits added, tests also corrected


Dirbaio (approved):
thanks!

PR #4134, opened 2025-04-25
fix: atomic waker does not call wake on the existing waker when registering a new one

ckrenslehner:
There seems to be a difference in the behavior of `AtomicWaker` and `WakeRegistration`. In particular, `AtomicWaker` does not wake the existing waker if a new one is registered. Therefore, the waker from before has no chance to re-register itself.

- No call of `wake` [here](https://github.com/embassy-rs/embassy/blob/main/embassy-sync/src/waitqueue/atomic_waker.rs#L24)
- Call of `wake` [here](https://github.com/embassy-rs/embassy/blob/main/embassy-sync/src/waitqueue/waker_registration.rs#L35)

If this is intended, we could add some docs.

### DIFF
--- a/embassy-sync/src/waitqueue/atomic_waker.rs
+++ b/embassy-sync/src/waitqueue/atomic_waker.rs
@@ -22,6 +22,10 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
         self.waker.lock(|cell| {
             cell.set(match cell.replace(None) {
                 Some(w2) if (w2.will_wake(w)) => Some(w2),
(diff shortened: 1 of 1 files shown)


Dirbaio:
Yes, this is intended. AtomicWaker is only used in drivers for waking tasks from interrupts, in that case you never should have two tasks waiting on the same thing, so waking the old waker is not useful.

ckrenslehner:
Okay, thanks. 
I will open another PR adding this to the docs.

PR #4137, opened 2025-04-27
Stm32 can data len

cschuhen:
This PR is a clone of https://github.com/embassy-rs/embassy/pull/4109 but it fixes the remaining CI failure by just making raw_data function public. The alternative would be to feature select based on bxcan vs canfd however I don't see any great risk in making that function available. It might be more optimal in some use cases.

From 4109:

Currently data() always returns an 8- or 64-byte slice, even though the actual CAN frame contains less data. The remaining bytes will be zero, so the user must check the frame DLC to see how many bytes were actually sent by the remote end. With this change the slice is the same length as was transmitted, with no extra bytes.

Additionally removed the ClassicData::MAX_DATA_LEN constant because almost all instances were using a hardcoded 8 anyway, and the FdData only ever uses 64 and doesn't have an equivalent MAX_DATA_LEN. Happy to make everything use MAX_DATA_LEN instead if that's preferred.

Finally fixed a missing cfg gate on a defmt call that broke the build on G4 without defmt feature.

### DIFF
--- a/embassy-stm32/src/can/bxcan/registers.rs
+++ b/embassy-stm32/src/can/bxcan/registers.rs
@@ -299,9 +299,9 @@ impl Registers {
         mb.tdtr().write(|w| w.set_dlc(frame.header().len() as u8));
 
         mb.tdlr()
-            .write(|w| w.0 = u32::from_ne_bytes(unwrap!(frame.data()[0..4].try_into())));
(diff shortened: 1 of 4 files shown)


cschuhen:
I have no idea why AG was made the author of d2183ab97814b8140fdad356543f84855a0eac67.

Dirbaio:
thanks!

PR #4146, opened 2025-04-29
[embassy-nrf] Add `InputRef`, `OutputRef`, and `FlexRef`

alexmoon:
Adds reference types for GPIOs so that they can be borrowed by drivers without being torn down when the driver is dropped. This allows, for example, an `Output` to be used in a `SimplePwm`, then the `SimplePwm` can be dropped and a `gpiote::OutputChannel` created using the same `Output` without its configuration being reset (and the pin disconnected) in between.

### DIFF
--- a/embassy-nrf/src/gpio.rs
+++ b/embassy-nrf/src/gpio.rs
@@ -3,6 +3,7 @@
 
 use core::convert::Infallible;
 use core::hint::unreachable_unchecked;
+use core::ops::{Deref, DerefMut};
 
 use cfg_if::cfg_if;
 use embassy_hal_internal::{impl_peripheral, Peri, PeripheralType};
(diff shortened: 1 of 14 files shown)


lulf:
I don't really like the changes to the API here where you can't just pass the peripheral anymore, but I do understand the need for this. I thought passing &mut p.P0_10 etc. would work for this use case, but seems that no longer the case. Is that due to the Peri changes?

alexmoon:
Yeah, I agree. I considered making a `From<Peri<'_, ...>>` impl for `OutputRef` and `InputRef` which would use the default level and pull options. That would make it `p.P0_10.into()` instead of `Output::new(p.P0_10, Level::Low).into()`. Note that `InputChannel` and `OutputChannel` were already taking `Input` and `Output` parameters instead of raw peripherals. This is sort of bringing `Pwm` in line with those.

0e4ef622:
I'm not sure what the purpose of InputRef/OutputRef are, it seems like adding a reborrow method to Input/Output would suffice.

alexmoon:
> EDIT: oh if its borrowed it won't drop.

Yep

Dirbaio:
Hello,

PR #4147, opened 2025-04-29
[trace]: Add task naming capability to tracing infrastructure

kat-perez:
This change adds a TaskTracker to the tracing infrastructure. The name and ID of a task are now stored in TaskHeader and a new function called `spawn_named` is added to assign a name to the task for viewing in SystemView.


embassy-executor/src/raw/trace.rs
```diff
@@ -83,6 +83,129 @@
...
+use core::sync::atomic::{AtomicUsize, Ordering};
+use rtos_trace::TaskInfo;
+
+const MAX_TASKS: usize = 1000;
+
+/// Represents a task being tracked in the task registry.
+///
+/// Contains the task's unique identifier and optional name.
+#[derive(Clone)]
+pub struct TrackedTask {
```
bugadani, line 96: You'll have a better chance with this PR if you store this data in the task headers, instead of wasting 3-4K RAM.

kat-perez: @bugadani Thanks for the review. I've updated my implementation to move the task information to TaskHeader and have used a LinkedList instead to keep track of all active tasks.

embassy-executor/src/spawner.rs
```diff
@@ -154,6 +154,39 @@ impl Spawner {
...
 
+    /// Spawns a new task with a specified name.
+    ///
+    /// # Arguments
+    /// * `name` - Static string name to associate with the task
+    /// * `token` - Token representing the task to spawn
+    ///
+    /// # Returns
+    /// Result indicating whether the spawn was successful
+    #[cfg(feature = "rtos-trace")]
```
lulf, line 165: Seems like this should be feature = "trace" ?

embassy-executor/src/spawner.rs
```diff
@@ -154,6 +154,39 @@ impl Spawner {
...
+                unsafe { self.executor.spawn(task) };
+                Ok(())
+            }
+            None => Err(SpawnError::Busy),
+        }
+    }
+
+    /// When rtos-trace is disabled, spawn_named falls back to regular spawn.
+    /// This maintains API compatibility while optimizing out the name parameter.
+    #[cfg(not(feature = "rtos-trace"))]
```
lulf, line 185: Seems like this should be feature = "trace" ?

embassy-executor/src/raw/trace.rs
```diff
@@ -210,10 +230,93 @@ pub(crate) fn executor_idle(executor: &SyncExecutor) {
...
+#[cfg(feature = "trace")]
+pub fn get_tasks_by_name(name: &'static str) -> impl Iterator<Item = TaskRef> + 'static {
+    filter_active_tasks(move |task| task.name() == Some(name))
+}
+
+/// Get tasks by ID
+#[cfg(feature = "trace")]
+pub fn get_task_by_id(id: u32) -> Option<TaskRef> {
+    filter_active_tasks(move |task| task.id() == id).next()
+}
```
lulf, line 305: Does these need to be public, or can they be internal? I'm a bit confused by the 'scattered' API, which makes it a bit hard to verify that it's all safe.

lulf (commented):
Looks good in general, I've left some comments on the API.

 Also, would it make sense to move more of this into the 'trace.rs'?

jamesmunns:
I do want to take a look at this, if we end up merging https://github.com/embassy-rs/embassy/pull/4035, it might make sense to re-use cordyceps for the intrusive linked list of (named) tasks

kat-perez:
@lulf 
> Also, would it make sense to move more of this into the 'trace.rs'?

Yeah, I can define a trait instead in trace.rs which extends Spawner since the spawn_named function is used exclusively for SystemView at the moment

kat-perez:
@lulf @jamesmunns Updated PR to move more of the functionality into trace.rs, please review. Thanks!

kat-perez:
@lulf Should I wait until #4035 is merged and rework cordyceps for the intrusive linked list of named task into this PR or will this be accepted as-is?

lulf (approved):
I think the current variant looks good!

PR #4150, opened 2025-04-30
RP: rp2040 overclocking

### LINKED ISSUE #3919: embassy-rp: RP2040@200MHz
The RP2040 has been certified for 200MHz system clock a few days ago.

According to the datasheet only an increased core voltage (1.15V) is required. As far as I can see, switching the core supply is not possible thru the HAL (until now this had very limited relevance).

Should we support this or even set as default?

1-rafael-1:
This PR introduces the ability to initialize the rp2040 with other frequencies than 125Mhz. A user can either use a higher level API to achieve a setting or set the PLL manually. See the two new examples for more details.

Should close #3919 

Disclosure for Review:
I have used AI Agents to help me code this and my level of experience is limited, most of what I touched I learned as I went. 
The agent I fed with the Pi Pico SDK as well as the datasheet, in the datasheet I tried to make sure I read up on all the things touched to see if i spot discrepancies. Those I found, I corrected.

I have tested:
- compile for rp2040
- compile for rp235x
- run blinky on both
- look for hardcoded 125Mhz throughout the HAL, fix that
- run my examples
- run a few of the other rp examples to see if i immediately spot a thing not working

EDIT:
- added tests for the calculations into the clocks module, found overflow issues and fixed them
- added a test to tests/rp

### DIFF
--- a/embassy-rp/src/clocks.rs
+++ b/embassy-rp/src/clocks.rs
@@ -391,7 +391,7 @@ impl ClockConfig {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```rust,ignore
     /// // Overclock to 200MHz
     /// let config = ClockConfig::at_sys_frequency_mhz(200);
     /// ```
@@ -424,7 +424,7 @@ impl ClockConfig {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```rust,ignore
     /// // Use a non-standard 16MHz crystal to achieve 250MHz
     /// let config = ClockConfig::with_custom_crystal(16_000_000, 250_000_000);
     /// ```
@@ -875,7 +875,7 @@ pub struct RtcClkConfig {
 ///
 /// # Example
 ///
-/// ```
+/// ```rust,ignore
 /// // Find parameters for 133MHz system clock from 12MHz crystal
 /// let pll_params = find_pll_params(12_000_000, 133_000_000).unwrap();
 /// ```
@@ -885,7 +885,7 @@ fn find_pll_params(input_hz: u32, target_hz: u32) -> Option<PllConfig> {
     const PLL_SYS_REFDIV: u8 = 1;
 
     // Calculate reference frequency
-    let reference_freq = input_hz / PLL_SYS_REFDIV as u32;
+    let reference_freq = input_hz as u64 / PLL_SYS_REFDIV as u64;
 
     // Start from highest fbdiv for better stability (like SDK does)
     for fbdiv in (16..=320).rev() {
@@ -900,10 +900,10 @@ fn find_pll_params(input_hz: u32, target_hz: u32) -> Option<PllConfig> {
         // (more conservative/stable approach)
         for post_div1 in (1..=7).rev() {
             for post_div2 in (1..=post_div1).rev() {
-                let out_freq = vco_freq / (post_div1 * post_div2) as u32;
+                let out_freq = vco_freq / (post_div1 * post_div2);
 
                 // Check if we get the exact target frequency without remainder
-                if out_freq == target_hz && (vco_freq % (post_div1 * post_div2) as u32 == 0) {
+                if out_freq == target_hz as u64 && (vco_freq % (post_div1 * post_div2) == 0) {
                     return Some(PllConfig {
                         refdiv: PLL_SYS_REFDIV,
                         fbdiv: fbdiv as u16,
@@ -928,7 +928,7 @@ fn find_pll_params(input_hz: u32, target_hz: u32) -> Option<PllConfig> {
 
         for post_div1 in (1..=7).rev() {
             for post_div2 in (1..=post_div1).rev() {
-                let out_freq = vco_freq / (post_div1 * post_div2) as u32;
+                let out_freq = (vco_freq / (post_div1 * post_div2) as u64) as u32;
                 let diff = if out_freq > target_hz {
                     out_freq - target_hz
                 } else {
@@ -1018,7 +1018,10 @@ pub(crate) unsafe fn init(config: ClockConfig) {
             cortex_m::asm::delay(delay_cycles);
 
             // Only now set the BOD level after voltage has stabilized
-            vreg.bod().write(|w| w.set_vsel(voltage.recommended_bod()));
+            vreg.bod().write(|w| {
+                w.set_vsel(voltage.recommended_bod());
+                w.set_en(true); // Enable brownout detection
+            });
         }
     }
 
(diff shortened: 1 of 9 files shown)


1-rafael-1:
Ready for review

embassy-rp/src/clocks.rs
```diff
@@ -440,81 +988,135 @@ pub(crate) unsafe fn init(config: ClockConfig) {
...
-            RefClkSrc::Xosc => (Src::XOSC_CLKSRC, Aux::CLKSRC_PLL_USB, xosc_freq / div),
-            RefClkSrc::Rosc => (Src::ROSC_CLKSRC_PH, Aux::CLKSRC_PLL_USB, rosc_freq / div),
-            RefClkSrc::PllUsb => (Src::CLKSRC_CLK_REF_AUX, Aux::CLKSRC_PLL_USB, pll_usb_freq / div),
-            // RefClkSrc::Gpin0 => (Src::CLKSRC_CLK_REF_AUX, Aux::CLKSRC_GPIN0, gpin0_freq / div),
-            // RefClkSrc::Gpin1 => (Src::CLKSRC_CLK_REF_AUX, Aux::CLKSRC_GPIN1, gpin1_freq / div),
-        }
+
+    // SETUP TEMPORARY STABLE CLOCKS FIRST
+    // Configure USB PLL for our stable temporary clock
+    // This follows the SDK's approach of using USB PLL as a stable intermediate clock
```
9names, line 1047: Which SDK?
Can you link to this code?

embassy-rp/src/clocks.rs
```diff
@@ -783,45 +1409,99 @@ fn start_xosc(crystal_hz: u32, delay_multiplier: u32) {
...
 
-    // Wait for PLL to lock
-    while !p.cs().read().lock() {}
+    // 5. Wait for PLL to lock with a timeout
+    let mut timeout = 1_000_000; // Reasonable timeout value
+    while !p.cs().read().lock() {
+        timeout -= 1;
+        if timeout == 0 {
+            // PLL failed to lock, return 0 to indicate failure
+            return 0;
```
9names, line 1485: You might want to return an Err() here instead of expecting a user to know that 0 is an error.

embassy-rp/src/clocks.rs
```diff
@@ -1197,3 +1878,206 @@ pub fn dormant_sleep() {
...
+            assert_eq!(VoltageScale::V0_85.recommended_bod(), 0b0111); // ~0.774V (91% of 0.85V)
+            assert_eq!(VoltageScale::V0_90.recommended_bod(), 0b1000); // ~0.817V (91% of 0.90V)
+            assert_eq!(VoltageScale::V0_95.recommended_bod(), 0b1001); // ~0.860V (91% of 0.95V)
+            assert_eq!(VoltageScale::V1_00.recommended_bod(), 0b1010); // ~0.903V (90% of 1.00V)
+            assert_eq!(VoltageScale::V1_05.recommended_bod(), 0b1011); // ~0.946V (90% of 1.05V)
+            assert_eq!(VoltageScale::V1_10.recommended_bod(), 0b1100); // ~0.989V (90% of 1.10V)
+            assert_eq!(VoltageScale::V1_15.recommended_bod(), 0b1101); // ~1.032V (90% of 1.15V)
+            assert_eq!(VoltageScale::V1_20.recommended_bod(), 0b1110); // ~1.075V (90% of 1.20V)
+            assert_eq!(VoltageScale::V1_25.recommended_bod(), 0b1111); // ~1.118V (89% of 1.25V)
+            assert_eq!(VoltageScale::V1_30.recommended_bod(), 0b1111); // ~1.118V (86% of 1.30V) - using max available
```
9names, line 1901: Isn't this test just a copy-paste of the original code? Does this prove anything?

examples/rp/src/bin/overclock.rs
```diff
@@ -0,0 +1,65 @@
...
+const COUNT_TO: i64 = 10_000_000;
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) -> ! {
+    // Set up for clock frequency of 200 MHz
+    // This will set all the necessary defaults including slightly raised voltage
+    let config = Config::new(ClockConfig::at_sys_frequency_mhz(200));
+
+    // Show the voltage scale for verification
+    info!("System core voltage: {}", Debug2Format(&config.clocks.voltage_scale));
```
9names, line 25: A bit surprising to users if their reported system core voltage is None if using the default core voltage.

9names (changes requested):

examples/rp/src/bin/overclock.rs
```diff
@@ -0,0 +1,65 @@
...
+use embassy_rp::gpio::{Level, Output};
+use embassy_time::{Duration, Instant, Timer};
+use {defmt_rtt as _, panic_probe as _};
+
+const COUNT_TO: i64 = 10_000_000;
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) -> ! {
+    // Set up for clock frequency of 200 MHz
+    // This will set all the necessary defaults including slightly raised voltage
```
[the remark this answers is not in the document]
1-rafael-1: Agreed. I will remove this here and find a better place and be more thorough in the actual module.

examples/rp/src/bin/overclock.rs
1-rafael-1: You are quite right. I will have a look at how to improve the whole thing.

embassy-rp/src/clocks.rs
```diff
@@ -1,4 +1,78 @@
...
+//! This module provides both high-level convenience functions and low-level manual
+//! configuration options for the RP2040 clock system.
+//!
+//! ## High-Level Convenience Functions
+//!
+//! For most users, these functions provide an easy way to configure clocks:
+//!
+//! - `ClockConfig::crystal(12_000_000)` - Default configuration with 12MHz crystal giving 125MHz system clock
+//! - `ClockConfig::at_sys_frequency_mhz(200)` - Set system clock to a specific frequency with automatic voltage scaling
+//! - `ClockConfig::with_external_crystal(16_000_000)` - Configure with a non-standard crystal frequency
```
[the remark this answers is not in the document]
1-rafael-1: Looking at this further I believe I overshot my mark somewhat. I will try and remove none-too-useful convenience from the external custom oscillator path, after all this can still be achieved by manual configuration and will be super-rare anyway.
And You are right of course, 15Mhz is the fastest supported Oscillator. Will make note of that, too.

embassy-rp/src/clocks.rs
```diff
@@ -69,6 +144,56 @@ pub enum PeriClkSrc {
...
+/// Core voltage scaling options for RP2040.
+///
+/// The RP2040 voltage regulator can be configured for different output voltages.
+/// Higher voltages allow for higher clock frequencies but increase power consumption and heat.
+#[cfg(feature = "rp2040")]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u8)]
+pub enum VoltageScale {
+    /// 0.85V
+    V0_85 = 0b0110,
```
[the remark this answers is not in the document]
1-rafael-1: yeah, hastyness on my side, sorry. The 0.8V was optically different from the higher ones because it sports more possible register values and that way i overlooked it. Will fix.

embassy-rp/src/clocks.rs
```diff
@@ -1,4 +1,78 @@
...
+//! config.voltage_scale = Some(VoltageScale::V1_15);
+//! ```
+//!
+//! ## Voltage Scaling for Overclocking (RP2040 only)
+//!
+//! When overclocking beyond 133MHz, higher core voltages are needed:
+//!
+//! - Up to 133MHz: `VoltageScale::V1_10` (default)
+//! - 133-200MHz: `VoltageScale::V1_15`
+//! - Above 200MHz: `VoltageScale::V1_20` or higher
```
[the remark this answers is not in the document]
1-rafael-1: I read that online or watched that on Youtube... can't find the source now and anyway You are of course right to question that. I have removed the doc at this point and moved a refined version to crystal_freq, only citing the datasheet and nothing else.

embassy-rp/src/clocks.rs
```diff
@@ -1,4 +1,78 @@
...
+//! ```
+//!
+//! Or using the default configuration:
+//! ```rust,ignore
+//! let config = ClockConfig::default();
+//! ```
+//!
+//! ### Overclock to 200MHz
+//! ```rust,ignore
+//! let config = ClockConfig::at_sys_frequency_mhz(200);
```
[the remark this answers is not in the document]
1-rafael-1: It sounded like "The clock config at sys frequency (of whatever)"... I t sounded better in my head :-) 
I removed this fn anyway, so luckily I have no need to come up with a nicer name.

embassy-rp/src/clocks.rs
```diff
@@ -69,6 +144,56 @@ pub enum PeriClkSrc {
...
 }
 
+/// Core voltage scaling options for RP2040.
+///
+/// The RP2040 voltage regulator can be configured for different output voltages.
+/// Higher voltages allow for higher clock frequencies but increase power consumption and heat.
+#[cfg(feature = "rp2040")]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u8)]
+pub enum VoltageScale {
```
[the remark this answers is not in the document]
1-rafael-1: I am bad at names, that is why. Is now CoreVoltage.

embassy-rp/src/clocks.rs
```diff
@@ -192,20 +370,248 @@ impl ClockConfig {
...
+        }
+
+        // For other frequencies, provide appropriate voltage scaling and PLL configuration
+        // Standard crystal on Raspberry Pi Pico boards is 12MHz
+        const DEFAULT_CRYSTAL_HZ: u32 = 12_000_000;
+
+        let sys_freq_hz = sys_freq_mhz * 1_000_000;
+        let config = Self::crystal_freq(DEFAULT_CRYSTAL_HZ, sys_freq_hz);
+
+        config
```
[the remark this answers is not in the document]
1-rafael-1: removed the entire fn.

embassy-rp/src/clocks.rs
1-rafael-1: it does prove:
- That I can be dense sometimes
- That You have a keen eye
...removed...

examples/rp/src/bin/overclock.rs
1-rafael-1: I now no longer use Option here and also default to 1_10, that should work better in that regard

examples/rp/src/bin/overclock.rs
```diff
@@ -0,0 +1,65 @@
+//! # Overclocking the RP2040 to 200 MHz
+//!
+//! This example demonstrates how to configure the RP2040 to run at 200 MHz using a higher level API.
```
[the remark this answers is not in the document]
1-rafael-1: Yes, okay. Done.

embassy-rp/src/clocks.rs
1-rafael-1: Here: https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/clocks.c in the official Pi Pico SDK. 

This part is the one I was most unsure of, so i fed an agent the entire SDK as well as the embassy-rp code to have it determine what they do in what order and based on that where we need changes. Then I came back to have it confirm back if we do as they do. Did that once more just now based on the changes I am making after this review.  

Since I have not coded a single line of C in my life i find it confusing to read, so to confirm viability I set out to add a test to embassy and also of course tested on my boards.  So far it all seems to work out.

embassy-rp/src/clocks.rs
1-rafael-1: Not sure if I agree, or more precisely: What good can we do here? I changed this to return Err(), but the only consumer is the init() fn, so my best idea was to use the newly minted result, but then just panic on Err. Is that what You had in mind? Are we better off that way? Could we not just panic here instead of retunrning 0? After all, nothing in the firmware will work anyway, we can just as well abort early?

1-rafael-1:
@9names  Thank You very much for taking the time to review this, much appreciated! I pushed the first batch of changes, not done yet.

embassy-rp/src/pio_programs/hd44780.rs
```diff
@@ -134,7 +135,12 @@ impl<'l, P: Instance, const S: usize> PioHD44780<'l, P, S> {
 
         let mut cfg = Config::default();
         cfg.use_program(&word_prg.prg, &[&e]);
-        cfg.clock_divider = 125u8.into();
+
+        // Scale the divider based on system clock frequency
+        // Original: 125 at 125 MHz (1 MHz PIO clock)
```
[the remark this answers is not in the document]
1-rafael-1: Agree, removing the comment

embassy-rp/src/pio_programs/rotary_encoder.rs
```diff
@@ -48,7 +49,12 @@ impl<'d, T: Instance, const SM: usize> PioEncoder<'d, T, SM> {
         cfg.set_in_pins(&[&pin_a, &pin_b]);
         cfg.fifo_join = FifoJoin::RxOnly;
         cfg.shift_in.direction = ShiftDirection::Left;
-        cfg.clock_divider = 10_000.to_fixed();
+
+        // Original: 10_000 at 125 MHz (12.5 KHz PIO clock)
+        // Scale divider to maintain same PIO clock frequency at different system clocks
+        let divider = (clk_sys_freq() as f32 / 12_500.0).to_fixed();
```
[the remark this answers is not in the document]
1-rafael-1: refactored accordingly

1-rafael-1:
@9names I think I covered all Your points. Thanks again!

There seems to be an issue with CI, it reports all clear, but looking at its log I do not think that is necessarily true.

It says: 

```
No teleprobe token found, skipping running HIL tests
```

So I believe something about running code on @Dirbaio s test rig is not working. Even if You are happy with the PR I woiud suggest to not merge before we saw a proper CI run on it.

embassy-rp/src/clocks.rs
9names: Sorry, it's not very clear in the above so I'm going to ask bluntly:
Did you write this code, is this LLM generated and human edited, or is this entirely LLM generated?
Did you review this code, did you do review plus have an LLM do it, or have you relied entirely on LLM's for that?

I need to know how much human verification has been involved to understand if you're asking me verify that your agent has done what it says it did.

9names:
We can always ask someone to kick off the HIL tests once this PR is ready for merge.
I did test it out locally to verify that it was functional.

embassy-rp/src/clocks.rs
9names: Returning Err instead of a magic number than means error allows you to follow standard Rust error handling patterns.

As to "should we just panic": no. This isn't an unrecoverable error.
Communicating that information is important, even if you're just panicking at the next level at the moment.

embassy-rp/src/clocks.rs
9names: After reading the pico-sdk sources I think the original code closely resembles [runtime_init_clocks()](https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/src/rp2_common/pico_runtime_init/runtime_init_clocks.c#L40). But with your changes, this function is now an amalgam of  `runtime_init_clocks()` and [set_sys_clock_pll()](https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/src/rp2_common/hardware_clocks/clocks.c#L380).

The concept of using a "temporary stable clock" doesn't make sense here. The purpose of `set_sys_clock_pll()` is to migrate from one PLL config across to another one, keeping peripherals working while you go. During init, no peripherals should be running yet!
So there's no value in setting up and switching to `usb_clk` first.

1-rafael-1: I do not mind you asking bluntly, really. The code would classify as LLM generated and human edited, or that comes closest for this part. As for review: Did it and had LLM assist in that, but as a mere hobbyist my relevant skill is limited.
 
Before starting out i identified possible sources of truth on the matter, the official SDK being the most digestible to an agent, the datasheet coming only second because such a large pdf is usually a bad source the search operations involved. That data was fed into the agent tasked to identify where we may diverge from the way the SDK does it today and propose changes, with the goal of enabling clocking.
 
All that i fed into an new session to double-check backwards if we diverge from what the SDK does. Such reverse-checks sometimes reveal discrepancies. Sadly, this was as much as I could realistically do at this point, given my level of knowledge.
The level of change in a central initialization sequence made me uneasy, so I tried to assure quality by testing -> Adding tests and doing manual tests to confirm functionality of the overclocking itself but more importantly to the continued functionality of the existing system.

Probably not Your most favorite answer, but to my defense I tried to disclose that in the PR upfront.

So next I will rework this part, manually this time, to edit out the unnecessary part and produce a version that is close to the original code. 

Oh, and thank You again for Your time!

embassy-rp/src/clocks.rs
1-rafael-1: Okay, got it i think. So basically returning Err makes the fn more useful in case the surrounding system ever needs it for something else or can someday recover from an error here.

embassy-rp/src/clocks.rs
1-rafael-1: Okay, I reworked the `init()` fn to more closely resemble the original code.
It now has these differences re the original:

+ setting core voltage, if that has been adjusted. And if it has, and only then, change brownout detection thresholds (only needed then, i believe, because the datasheet says it has defaults if nothing is set)

+ the `configure_pll()` returns Err(), if the time to lock exceeds a threshold. That was AI suggesed but I do think that makes sense, else we may wait forever. So in `init()`we need to panic if that happens, because what else could we do. So that code block changed slightly. I can change that back, too, I You object.

The `configure_pll()` has a power down sequence that it originally did not have. I looked at the C SDKs `pll_init()` and it looks like it performs extensive resets initially, too (see resets.h). So I kind of believe this is still valid to do. That is a thing that warrants a look from You though.

I tested manually, we can still overclock and we can - not surprisingly - still run at defaults. So I am hoping You like this better.

1-rafael-1:
@9names are there more things You would want changed?

embassy-rp/src/clocks.rs
```diff
@@ -192,20 +362,156 @@ impl ClockConfig {
...
+    /// the usual 12Mhz crystal, or panic if no valid parameters can be found.
+    ///
+    /// # Note on core voltage:
+    /// To date the only officially documented core voltages (see Datasheet section 2.15.3.1. Instances) are:
+    /// - Up to 133MHz: V1_10 (default)
+    /// - Above 133MHz: V1_15, but in the context of the datasheet covering reaching up to 200Mhz
+    /// That way all other frequencies below 133MHz or above 200MHz are not explicitly documented and not covered here.
+    /// In case You want to go below 133MHz or above 200MHz and want a different voltage, You will have to set that manually and with caution.
+    #[cfg(feature = "rp2040")]
+    pub fn crystal_freq(sys_freq_hz: u32) -> Self {
```
9names, line 396: This function name is confusing.
It looks like you're going to set the frequency of the crystal, not the system clock frequency.

I think `system_freq(hz: u32) -> Self` would make it a lot more clear what it is you're going to set here.

9names (changes requested):

embassy-rp/src/clocks.rs
```diff
@@ -914,125 +934,79 @@ pub(crate) unsafe fn init(config: ClockConfig) {
...
-            cortex_m::asm::delay(delay_cycles);
+            if settling_time_us != 0 {
+                // Delay in microseconds, using the ROSC frequency to calculate cycles
+                let cycles_per_us = rosc_freq / 1_000_000;
+                let delay_cycles = settling_time_us * cycles_per_us;
+                cortex_m::asm::delay(delay_cycles);
+            }
 
-            // Only now set the BOD level after voltage has stabilized
+            // Only now set the BOD level. At htis point the voltage is considered stable.
```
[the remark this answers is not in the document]
1-rafael-1: :-)

embassy-rp/src/clocks.rs
```diff
@@ -893,6 +888,30 @@ pub(crate) unsafe fn init(config: ClockConfig) {
...
+    // let gpin0_freq = config.gpin0.map_or(0, |p| {
+    //     core::mem::forget(p.1);
+    //     p.0
+    // });
+    // CLOCKS.gpin0.store(gpin0_freq, Ordering::Relaxed);
+    // let gpin1_freq = config.gpin1.map_or(0, |p| {
+    //     core::mem::forget(p.1);
+    //     p.0
+    // });
+    // CLOCKS.gpin1.store(gpin1_freq, Ordering::Relaxed);
```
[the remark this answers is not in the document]
1-rafael-1: ugh, correction: so many of these places.... I will keep the comment after the imports and refer users to that.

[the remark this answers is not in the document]
1-rafael-1: ...22 places...

embassy-rp/src/clocks.rs
```diff
@@ -914,125 +934,79 @@ pub(crate) unsafe fn init(config: ClockConfig) {
...
-    };
-    CLOCKS.rosc.store(rosc_freq, Ordering::Relaxed);
-
-    // Configure XOSC - we'll need this for our temporary stable clock
-    let xosc_freq = match &config.xosc {
+    let (xosc_freq, pll_sys_freq, pll_usb_freq) = match config.xosc {
         Some(config) => {
+            // start XOSC
+            // datasheet mentions support for clock inputs into XIN, but doesn't go into
+            // how this is achieved. pico-sdk doesn't support this at all.
```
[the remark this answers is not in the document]
1-rafael-1: I see Your point, that comment is not from me though. It is 2 years old. 

I searched the SDK and I read what there is to read (4 hits) in the mentioned document.... You are right it sounds that there is no software requirement here. The SDK seems to do nothing related.

Comment deleted.

embassy-rp/src/clocks.rs
1-rafael-1: Agree, changed.

9names (approved):
Looks good to me

lulf (approved):
Thanks for reviewing @9names

1-rafael-1:
@lulf i do not think I broke this, could you run this again?

PR #4151, opened 2025-05-01
AtomicWaker: take and wake

bugadani:
While in embassy tasks are static, in other environments it's usually not the norm. There, keeping the Waker may force the task that registered it to remain allocated, potentially indefinitely. Given that AtomicWaker is used in peripheral drivers that aren't tied to a particular async framework, letting go of the Waker once it's not needed is probably better than keeping it around.

I expect this change to surface some issues around Futures that only register the waker once and expect it to stick around.

### DIFF
--- a/embassy-executor/tests/test.rs
+++ b/embassy-executor/tests/test.rs
@@ -156,8 +156,7 @@ fn waking_after_completion_does_not_poll() {
     let (executor, trace) = setup();
     executor.spawner().spawn(task1(trace.clone(), waker)).unwrap();
 
-    unsafe { executor.poll() };
-    waker.wake();
+    // Task registers waker, then exits
     unsafe { executor.poll() };
 
     // Exited task may be waken but is not polled
@@ -176,7 +175,6 @@ fn waking_after_completion_does_not_poll() {
             "pend",       // spawning a task pends the executor
             "poll task1", //
             "pend",       // manual wake, gets cleared by poll
-            "pend",       // manual wake, single pend for two wakes
             "pend",       // respawning a task pends the executor
             "poll task1", //
         ]
--- a/embassy-sync/src/waitqueue/atomic_waker.rs
+++ b/embassy-sync/src/waitqueue/atomic_waker.rs
@@ -32,11 +32,9 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
 
     /// Wake the registered waker, if any.
     pub fn wake(&self) {
-        self.waker.lock(|cell| {
-            if let Some(w) = cell.take() {
-                w.wake();
-            }
-        })
+        if let Some(waker) = self.waker.lock(|cell| cell.take()) {
+            waker.wake();
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


embassy-executor/tests/test.rs
```diff
@@ -156,8 +156,7 @@ fn waking_after_completion_does_not_poll() {
     let (executor, trace) = setup();
     executor.spawner().spawn(task1(trace.clone(), waker)).unwrap();
 
-    unsafe { executor.poll() };
```
bugadani, line 159: I'm not sure what I was thinking here, the task is single shot and exits by the time this poll returns.

embassy-sync/src/waitqueue/atomic_waker.rs
```diff
@@ -33,9 +33,8 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
     /// Wake the registered waker, if any.
     pub fn wake(&self) {
         self.waker.lock(|cell| {
-            if let Some(w) = cell.replace(None) {
-                w.wake_by_ref();
-                cell.set(Some(w));
+            if let Some(w) = cell.take() {
+                w.wake();
             }
         })
```
Dominaezzz, line 39: Would I be able to sell you on this now?
```Rust
        if let Some(w) = self.waker.lock(|cell| cell.take())  {
            w.wake();
        }
```

I'm thinking the lock should be released earlier. You never know how expensive/slow `wake()` might be.

Dirbaio:
bender run

Dirbaio:
I'm wondering, this alone doesn't fix the whole problem right? we also have to ensure all drivers deregister all their wakers on drop, otherwise they'll stay there forever keeping a reference to the task.

And the extra code to do that would be "waste" when running with embassy-executor.

bugadani:
No, not the whole problem, but with the follow-up PR the bits are there for HAL authors to do it if they want to, without their own AtomicWaker.

Dirbaio:
Is this something you're considering doing in esp-hal? 

How're you planning of dealing with "the extra code to do that would be "waste" when running with embassy-executor"? just accept the bloat?

Dominaezzz:
> I'm wondering, this alone doesn't fix the whole problem right? we also have to ensure all drivers deregister all their wakers on drop, otherwise they'll stay there forever keeping a reference to the task.

Once this (and the follow up PR lands), I want to make PRs to `embassy-sync` to fix more of the problem.
For example [the `Future` returned by `Signal::wait`](https://github.com/embassy-rs/embassy/blob/03122737c1f7d13339d36a19e234a472edd82c93/embassy-sync/src/signal.rs#L111) doesn't free the waker when it drops/cancels.
Of course, unlike this PR, everyone would be paying for the extra code to do this. Would you be against this? (I don't want waste the elbow grease)

bugadani:
So, this PR should either be neutral, or size/perf-positive because there's no need to write the Waker back into the Cell. However, the next PR will not be - `take` or `deregister` or whatever we end up with is a net loss for embassy. So what if we added a `non-static-tasks` feature to let other ecosystems work? The implication is that the next PR would have to be reverted from `take` to `deregister` so that the function could be no-op if the feature is not set.

PR #4152, opened 2025-05-01
AtomicWaker: add `take` method

bugadani:
This PR adds a function to extract a waker from an AtomicWaker, which may be useful e.g. when dropping a driver whose AtomicWaker lives in static memory.

Alternatively this can be done by a "wake" call.

### DIFF
--- a/embassy-sync/src/waitqueue/atomic_waker.rs
+++ b/embassy-sync/src/waitqueue/atomic_waker.rs
@@ -30,9 +30,14 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
         })
     }
 
+    /// Extracts the previously registered waker, leaving `None` in its place.
+    pub fn take(&self) -> Option<Waker> {
+        self.waker.lock(|cell| cell.take())
+    }
+
     /// Wake the registered waker, if any.
     pub fn wake(&self) {
-        if let Some(waker) = self.waker.lock(|cell| cell.take()) {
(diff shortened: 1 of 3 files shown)


embassy-sync/src/waitqueue/atomic_waker.rs
```diff
@@ -30,12 +30,19 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
         })
     }
 
+    /// Removes previously registered waker.
+    pub fn deregister(&self) {
```
Dominaezzz, line 34: This could be
```suggestion
    pub fn take(&self) -> Option<Waker> {
```
which pretty much does the same thing but is much more useful.

paolobarbolini:
What's the status on this? This PR sounds like a good idea, as it aligns the API and the behavior of the atomic waker to other popular crates like `atomic-waker` and `futures`. The current behavior of never taking the waker out is quite surprising and I'd immagine that in some cases leads to many useless consecutive wakes to the async runtime or even mask cases in which the user forgets to re-register the waker before returning `Poll::Pending`.

bugadani:
This is blocked on https://github.com/embassy-rs/embassy/pull/4151 - or rather, on this question: should embassy-specific types/HALs care about compatibility with other async runtimes? On our end, we can just duplicate AtomicWaker in esp-hal in whichever way we see fit, and then our motivator for these PRs disappears.

PR #4157, opened 2025-05-02
add `timer_src_clk_sys` feature to support running RP2350x timer at s…

U007D:
…ystem clock frequency

NOTE:
@Dirbaio, 

The changes seem to work. But I'm concerned that removing the `tick-hz-1_000_000` from the `time-driver` feature may break existing users.  What do you think?

I'll submit the MR, with this concern expressed in the description.

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -42,7 +42,8 @@ critical-section-impl = ["critical-section/restore-state-u8"]
 unstable-pac = []
 
 ## Enable the timer for use with `embassy-time` with a 1MHz tick rate.
-time-driver = ["dep:embassy-time-driver", "embassy-time-driver?/tick-hz-1_000_000", "dep:embassy-time-queue-utils"]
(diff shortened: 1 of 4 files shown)


lulf:
> …ystem clock frequency
> 
> NOTE: @Dirbaio,
> 
> The changes seem to work. But I'm concerned that removing the `tick-hz-1_000_000` from the `time-driver` feature may break existing users. What do you think?
> 

The default for not setting this feature is 1 mhz (https://docs.embassy.dev/embassy-time-driver/git/default/index.html#tick-rate), so I think this should be OK.


> I'll submit the MR, with this concern expressed in the description.

U007D:
This MR now also properly sets the RP2350's `TIMER0.SOURCE` to `CLK_SYS` on `init()` when the `timer-src-clk-sys` feature is enabled.

U007D:
Is there anything else you'd like to see me chang or include in this MR, @Dirbaio?

Dirbaio:
Hello,

PR #4162, opened 2025-05-03
 chore: update all non-example defmt dependencies

nyurik:


jamesmunns:
@nyurik this looks interesting, but CI is failing. It's possible that there are some you missed, or some of the the 0.3.x deps need to be updated to the one that got semver-tricked into 1.0:

```
thread 'main' panicked at src/bin/cargo-batch.rs:136:85:
called `Result::unwrap()` on an `Err` value: failed to select a version for `defmt`.
    ... required by package `embassy-executor v0.7.0 (/ci/code/embassy-executor)`
    ... which satisfies path dependency `embassy-executor` (locked to 0.7.0) of package `embassy-basic-example v0.1.0 (/ci/code/docs/examples/basic)`
versions that meet the requirements `^1.0.1` are: 1.0.1

the package `defmt` links to the native library `defmt`, but it conflicts with a previous package which links to `defmt` as well:
package `defmt v0.3.10`
    ... which satisfies dependency `defmt = "^0.3"` (locked to 0.3.10) of package `embassy-basic-example v0.1.0 (/ci/code/docs/examples/basic)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "defmt"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `defmt` which could resolve this conflict
run failed: exited with code 101
```

So `embassy-basic-example` needs to be updated somehow?

Dirbaio:
Yeah this is because CI caches lockfiles. 

I think updating defmt everywhere to 1.0 should fix it.

kpfleming:
Updating the examples will require heapless==0.9.x, and that hasn't been released yet.

PR #4167, opened 2025-05-04
Clarify docs for signal and watch

JuliDi:
This is an attempt at clarifying the documentation around `Signal` and `Watch` after searching for a bug where `Signal` had been used instead of `Watch` with multiple consumers.

`Watch` was lacking any description of the generic parameter `N`, which is the number of receivers. I have added that at two places.

`Signal` was lacking the description from the top-level docs that it is intended for a _single_ consumer. The `wait` function was not explicit about taking the value from the signal (though it could be inferred since the Future returns  `T`).

Let me know if there is anything I can change to improve this PR. 

Also, maybe we could add some clarification on why Signal does not implement the Receiver like Watch does? Watch does seem to check how many consumers there are, while Signal lets you add multiple consumers, leading to "unexpected" behavior.

### DIFF
--- a/embassy-sync/src/signal.rs
+++ b/embassy-sync/src/signal.rs
@@ -6,7 +6,7 @@ use core::task::{Context, Poll, Waker};
 use crate::blocking_mutex::raw::RawMutex;
 use crate::blocking_mutex::Mutex;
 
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thanks!

Issue #4165, opened 2025-05-04
SAI read future never completes in `examples/stm32h7/.../sai.rs`


vinsynth:
I'm using a Daisy Seed, rev. 7 as required by the example. I replaced the necessary chip definitions in `.cargo/config.toml` and `Cargo.toml`, in addition to adding the `memory.x` file equivalent linked [here](https://github.com/electro-smith/libDaisy/blob/master/core/STM32H750IB_flash.lds) from the manufacturer. As far as I can tell, with this setup, neither `sai_receiver.read(...)` nor `sai_transmitter.write(...)` in `sai.rs` ever complete. I don't think it's a hardware issue; running a passthrough example from the manufacturer on the chip works as expected. Perhaps some defaults have changed for clock configuration since the example's writing, and now the SAI peripheral isn't being initialized as it should be? Or something is wrong with my `memory.x`?

finnh-dev:
I am currently investigating a similar problem for the [daisy-embassy](https://github.com/Dicklessgreat/daisy-embassy) BSP. 
The interesting bit is, that our code still works for the Daisy Seed 1.1 version (WM8731 Codec) but encounters the same issue with the 1.2 version (PCM3060 Codec).
From what i've gathered when looking at the differences in SAI configuration between the codecs the main difference is the Master/Slave mode of the transmitter and reciever:
| Codec | SAI Transmitter Mode | SAI Receiver Mode |
|:---|:---|:---|
|WM8731| Slave | Master |
|PCM3060| Master | Slave |

Switching the Master/Slave Configuration for the PCM3060 implementation seems to fix the issue where the future never completes but doesnt produce the expected audio output (In my case just some noticable noise instead of the expected audio passthrough).

We recently switched from using `embassy-stm32` at rev `ca3d091faf3ef76ea45f88eea5c2a6700378daff` to version `0.2.0` (rev ` 6034b17728b3528e42c8499a2893dc35d51d559`) and this seems to cause the issue as using older versions of `daisy-embassy` still works as expected.

Dicklessgreat:
Just to add a bit of context,, the Seed 1.1 uses I2C to initialize the WM8731, while the Seed 1.2 does not—it's configured in ”hardware mode” (using its config pins), so the settings are fixed and the i2c interface is not connected like the 1.1. I'm not yet sure how this difference affects the current SAI implementation, but I can at least confirm that the combination of WM8731 and I2C works correctly.

vinsynth:
[...]

PR #4169, opened 2025-05-05
Fix h7 SAI example start sequence

vinsynth:
This commit fixes the start sequence of the stm32h7 SAI example, ensuring the master subblock is started via `transmitter.write()` before first calling `receiver.read()` for the slave subblock. See #4165.


kalkyl:
Thanks!

PR #4170, opened 2025-05-06
chore: Wrap link_section attribute with unsafe

yutannihilation:
Rust 2024 edition requires marking several attributes as `unsafe`, and `link_section` is one of these.

https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html

As Embassy doesn't migrate to Rust 2024 edition, so this isn't a problem yet. However, if a user create a new Embassy project with the latest Rust toolchain, its edition is Rust 2024, so the compilation fails with this error:

```
error: unsafe attribute used without unsafe
  --> src/bin/tmp.rs:21:3
   |
21 | #[link_section = ".bi_entries"]
   |   ^^^^^^^^^^^^ usage of unsafe attribute
   |
help: wrap the attribute in `unsafe(...)`
   |
21 | #[unsafe(link_section = ".bi_entries")]
   |   +++++++                            +

error: could not compile `rp2350-blink` (bin "tmp") due to 1 previous error
```

In order to mark attributes as unsafe, Rust >=1.82 is needed. As Embassy uses Rust 1.85, it won't be a problem to introduce these `unsafe` without waiting for Rust 2024.

### DIFF
--- a/docs/pages/faq.adoc
+++ b/docs/pages/faq.adoc
@@ -268,7 +268,7 @@ General steps:
 1. Find out which memory region BDMA has access to. You can get this information from the bus matrix and the memory mapping table in the STM32 datasheet.
 2. Add the memory region to `memory.x`, you can modify the generated one from https://github.com/embassy-rs/stm32-data-generated/tree/main/data/chips.
 3. You might need to modify `build.rs` to make cargo pick up the modified `memory.x`.
-4. In your code, access the defined memory region using `#[link_section = ".xxx"]`
+4. In your code, access the defined memory region using `#[unsafe(link_section = ".xxx")]`
 5. Copy data to that region before using BDMA.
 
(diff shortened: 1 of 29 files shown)


yutannihilation:
Hmm, it seems I was a bit too optimistic. Probably I need to fix cortex-m-rt first.

```
   Compiling embassy-nrf v0.3.1 (/ci/code/embassy-nrf)
error: this attribute is not allowed on an interrupt handler controlled by cortex-m-rt
  --> /ci/code/embassy-rp/src/multicore.rs:93:1
   |
93 | #[unsafe(link_section = ".data.ram_func")]
   | ^

error: this attribute is not allowed on an interrupt handler controlled by cortex-m-rt
   --> /ci/code/embassy-rp/src/multicore.rs:118:1
    |
118 | #[unsafe(link_section = ".data.ram_func")]
    | ^

   Compiling embassy-stm32 v0.2.0 (/ci/code/embassy-stm32)
```

lulf (approved):
Thanks for fixing these!

PR #4175, opened 2025-05-07
iMXRT OS timer

felipebalbi:
Let the user select the OS Timer0 as a time source.


felipebalbi:
Question here: this PR removes `time-driver` feature in favor of `time-driver-rtc` and `time-driver-os-timer`. Should I keep `time-driver` for compatibility and just make sure that it enables `time-driver-rtc` or `time-driver-os-timer`?

lulf (approved):
There's been no release yet, so I don't think you need to keep it backwards compatible.

PR #4176, opened 2025-05-07
Implement embedded-io-async traits for USB CDC ACM

matt-rodgers:
Implement embedded-io-async Read and Write traits for USB CDC ACM.

For the Receiver side, this requires being able to buffer a received packet if the caller doesn't read the entire length in one go, so a separate `BufferedReceiver` is created for this.


lulf (approved):
Thanks! :rocket:

PR #4178, opened 2025-05-08
Stm32: Fix opamp copy pasta mistake

diondokter:
The opamp was refactored in #4089 but I think there was a copy paste error.

The old code:
https://github.com/embassy-rs/embassy/blob/d41eeeae79388f219bf6a84e2f7bde9f6b532516/embassy-stm32/src/opamp.rs#L149-L156

When a gain was selected (which is now always true in the new function), the VmSel is 0x10 (PGA). But in the new code it was left on OUTPUT.

I was seeing weird ADC values and this fixes it.


skoe:
Quick review comment: I’ve cross-checked the change against the original code and RM0440. The change is correct: in PGA mode with gain ≥ 2, VM_SEL should be set to 0b10 (VmSel::PGA). +1

PR #4184, opened 2025-05-10
Add intercore communication example for STM32H755CM4 and CM7

Ragarnoy:
This pull request introduces intercore communication examples for the STM32H755CM4 and CM7. These examples are functional in debug mode but currently not operational in release mode. The reason for this limitation is under investigation.

### DIFF
--- a/examples/stm32h755cm4/src/bin/intercore.rs
+++ b/examples/stm32h755cm4/src/bin/intercore.rs
@@ -1,0 +1,181 @@
+#![no_std]
+#![no_main]
+
+// IMPORTANT: This must match EXACTLY the definition in CM7!
+mod shared {
+    use core::sync::atomic::{AtomicU32, Ordering};
+
+    /// Shared LED state between CM7 and CM4 cores
+    #[repr(C, align(4))]
+    pub struct SharedLedState {
+        // Magic number for validation
+        pub magic: AtomicU32,
+        // Counter for synchronization testing
+        pub counter: AtomicU32,
+        // LED states packed into a single atomic
+        pub led_states: AtomicU32,
+    }
+
+    // Bit positions in led_states
+    pub const GREEN_LED_BIT: u32 = 0;
+    pub const YELLOW_LED_BIT: u32 = 1;
+
+    impl SharedLedState {
+        pub const fn new() -> Self {
+            Self {
+                magic: AtomicU32::new(0xDEADBEEF), // Magic number
+                counter: AtomicU32::new(0),
+                led_states: AtomicU32::new(0),
+            }
+        }
+
+        /// Set LED state using safe bit operations
+        #[inline(never)]
+        pub fn set_led(&self, is_green: bool, state: bool) {
+            let bit = if is_green { GREEN_LED_BIT } else { YELLOW_LED_BIT };
+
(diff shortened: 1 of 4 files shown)


Ragarnoy:
@diondokter Asking for your point of view on this since you did all of the work for the dual core H7 on embassy

examples/stm32h755cm4/src/bin/intercore.rs
```diff
@@ -0,0 +1,188 @@
...
+            let current = self.led_states.load(Ordering::SeqCst);
+
+            let new_value = if state {
+                current | (1 << bit) // Set bit
+            } else {
+                current & !(1 << bit) // Clear bit
+            };
+
+            self.led_states.store(new_value, Ordering::SeqCst);
+            core::sync::atomic::fence(Ordering::SeqCst);
```
diondokter, line 60: In a bunch of places you do this fence after an atomic operation. I don't think this is necessary (though I could be wrong)

noppej:
> A lot of inlining has to be disabled for this to work in release

@Ragarnoy Are you referring inlining in the embassy libs?

Ragarnoy:
> > A lot of inlining has to be disabled for this to work in release
> 
> @Ragarnoy Are you referring inlining in the embassy libs?

nope, inlining the functions in the example

examples/stm32h755cm4/src/bin/intercore.rs
```diff
@@ -0,0 +1,182 @@
...
+        pub fn set_led(&self, is_green: bool, state: bool) {
+            let bit = if is_green { GREEN_LED_BIT } else { YELLOW_LED_BIT };
+            let current = self.led_states.load(Ordering::SeqCst);
+
+            let new_value = if state {
+                current | (1 << bit) // Set bit
+            } else {
+                current & !(1 << bit) // Clear bit
+            };
+            self.led_states.store(new_value, Ordering::SeqCst);
```
noppej, line 58: @Ragarnoy I suspect this is a gap in my knowledge / understanding ... can you please help me understand why it is OK to do this `Ordering::SeqCst`, even though our Cargo.toml references `cortex-m` with feature `critical-section-single-core`?

diondokter: Only emulated atomics use critical sections.
Real ones, like the H7 has, don't. So these are real atomics that work fine across cores

noppej: @diondokter - thanks for the pointer about the H7 atomics.  With this in mind, I do find it confusing that the `Cargo.toml` requires that `critical-section-single-core` feature flag. Is it just silently ignored by `cortex-m`, or?

Dirbaio: in H7 multicore you build two separate firmwares one for each core. So each firmware is effectively "single core", and the other core is considered to be "outside" the Rust abstract machine. 

When communicating between cores you have to manually ensure that you only use primitives suitable for that, such as atomics, or shared memory with fences placed properly. You can only use repr(C) structs since the Rust abi is not stable.

Using critical-section-single-core is fine because you're never going to use critical sections for shared data (how would you, you can't share repr(Rust) things between cores)

PR #4186, opened 2025-05-12
nrf5340: Add IPC peripheral

0e4ef622:
This is a heavily reworked version of #3658.

I added the PPI task/event, and made each IPC event its own struct that can be handled independently.


lulf (approved):
Looks like a great way to use it, thanks!

PR #4187, opened 2025-05-12
RP235x overclocking

1-rafael-1:
After overclocking for rp2040 got enabled, this adds convenient overclocking for rp235x.

**Maintainer Note**:

Unlike for the rp2040 there is no manufacturer support for running on higher frequency or higher core voltage. Meaning 
 this is out of specs for now and in case this is considered too risky, please do not merge this!

**Reviewer Note**:

+ AI usage only for syntax help on setting registers, rest is hand-rolled
+ Add Core Voltage options for rp235x, add brownout detection values for rp235x
+ Add rp235x branch for this to `init()`
+ Add fn `core_voltage()` to retrieve the value from register instead of reading it from the config and update examples accordingly
+ Add one rp235x example
+ Add test branch for rp235x
+ some documentation changes

### DIFF
--- a/embassy-rp/src/clocks.rs
+++ b/embassy-rp/src/clocks.rs
@@ -140,6 +140,10 @@ pub enum PeriClkSrc {
 ///
 /// The voltage regulator can be configured for different output voltages.
 /// Higher voltages allow for higher clock frequencies but increase power consumption and heat.
+///
+/// **Note**: For RP235x the maximum voltage is 1.30V, unless unlocked by setting unless the voltage limit
+/// is disabled using the disable_voltage_limit field in the vreg_ctrl register. For lack of practical use at this
+/// point in time, this is not implemented here. So the maximum voltage in this enum is 1.30V for now.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[repr(u8)]
 pub enum CoreVoltage {
@@ -227,54 +231,6 @@ pub enum CoreVoltage {
     #[cfg(feature = "_rp235x")]
     /// RP235x: 1.30V
     V1_30 = 0b01111,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.35V
-    V1_35 = 0b10000,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.40V
-    V1_40 = 0b10001,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.50V
-    V1_50 = 0b10010,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.60V
-    V1_60 = 0b10011,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.65V
-    V1_65 = 0b10100,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.70V
-    V1_70 = 0b10101,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.80V
-    V1_80 = 0b10110,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 1.90V
-    V1_90 = 0b10111,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 2.00V
-    V2_00 = 0b11000,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 2.35V
-    V2_35 = 0b11001,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 2.50V
-    V2_50 = 0b11010,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 2.65V
-    V2_65 = 0b11011,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 2.80V
-    V2_80 = 0b11100,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 3.00V
-    V3_00 = 0b11101,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 3.15V
-    V3_15 = 0b11110,
-    #[cfg(feature = "_rp235x")]
-    /// RP235x: 3.30V
-    V3_30 = 0b11111,
 }
 
 impl CoreVoltage {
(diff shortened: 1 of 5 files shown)


1-rafael-1:
@9names could I kindly ask You to review, if You have the time? You did the last one and so I am hoping you are on topic without needing to do too much fresh mental work. 
NO issue at all, if You do not find the time! None of this is urgent.

embassy-rp/src/clocks.rs
```diff
@@ -1283,6 +1397,58 @@ pub fn clk_rtc_freq() -> u16 {
     CLOCKS.rtc.load(Ordering::Relaxed)
 }
 
+/// The core voltage of the chip.
+///
+/// Returns the current core voltage or an error if the voltage register
+/// contains an unknown value.
+pub fn core_voltage() -> Result<CoreVoltage, &'static str> {
```
CBJamo, line 1404: Use an error enum rather than a str for the result error type. In embedded we want to avoid strings whenever we can.

embassy-rp/src/clocks.rs
```diff
@@ -136,56 +136,141 @@ pub enum PeriClkSrc {
...
+/// is disabled using the disable_voltage_limit field in the vreg_ctrl register. For lack of practical use at this
+/// point in time, this is not implemented here. So the maximum voltage in this enum is 1.30V for now.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[repr(u8)]
 pub enum CoreVoltage {
-    /// 0.80V - Suitable for lower frequencies
+    // RP2040 voltage levels
+    #[cfg(feature = "rp2040")]
+    /// RP2040: 0.80V
     V0_80 = 0b0000,
```
CBJamo, line 153: Since the two chips don't share any values in this enum, I would do:
```rust
#[cfg(feature = "rp2040")]
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
#[repr(u8)]
pub enum CoreVoltage {
//2040 stuff
}

#[cfg(feature = "_rp235x")]
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
#[repr(u8)]
pub enum CoreVoltage {
//2350 stuff
}
```

I would also add `#[cfg_attr(feature = "defmt", derive(defmt::Format))]` to both enums. In fact, I'd add that to any enum or struct that derives debug. Then remove the calls to Debug2Format when printing this enum.

CBJamo (commented):
I was watching the last PR, but never had the time to actually review. This mostly looks good to me, just a couple notes.

embassy-rp/src/clocks.rs
1-rafael-1: I adapted from other places returning string. I can change that, gladly. I would then rather change all error strings to the new error enum, okay?

embassy-rp/src/clocks.rs
1-rafael-1: Will do

embassy-rp/src/clocks.rs
1-rafael-1: ah strike that... i will just change this to return error enum.

1-rafael-1: I did it and I generally like it. There is one thing though that I found no nice way for. `Init()`does not return result, so there we must panic. Tha means I did this ugly thing here:

```Rust
let (xosc_freq, pll_sys_freq, pll_usb_freq) = match config.xosc {
        Some(config) => {
            // start XOSC
            start_xosc(config.hz, config.delay_multiplier);

            let pll_sys_freq = match config.sys_pll {
                Some(sys_pll_config) => match configure_pll(pac::PLL_SYS, config.hz, sys_pll_config) {
                    Ok(freq) => freq,
                    #[cfg(feature = "defmt")]
                    Err(e) => panic!("Failed to configure PLL_SYS: {}", e),
                    #[cfg(not(feature = "defmt"))]
                    Err(_e) => panic!("Failed to configure PLL_SYS"),
                },
                None => 0,
            };
            let pll_usb_freq = match config.usb_pll {
                Some(usb_pll_config) => match configure_pll(pac::PLL_USB, config.hz, usb_pll_config) {
                    Ok(freq) => freq,
                    #[cfg(feature = "defmt")]
                    Err(e) => panic!("Failed to configure PLL_USB: {}", e),
                    #[cfg(not(feature = "defmt"))]
                    Err(_e) => panic!("Failed to configure PLL_USB"),
                },
                None => 0,
            };

            (config.hz, pll_sys_freq, pll_usb_freq)
        }
        None => (0, 0, 0),
    };
```

Is there a nicer way? And if no, should I just ditch the defmt branch? After all, a panic is a panic and we will not be reading the message anyway, since we do not get the chip initialized.

embassy-rp/src/clocks.rs
1-rafael-1: Done, and You were right, this looks more readable now.
Also added `#[cfg_attr(feature = "defmt", derive(defmt::Format))]` to the enums and structs that derive debug.

1-rafael-1:
> I was watching the last PR, but never had the time to actually review. This mostly looks good to me, just a couple notes.

Thank You for taking the time!

embassy-rp/src/clocks.rs
CBJamo: You don't need the cfg gates there. There are some helper macros in the fmt.rs file. If the defmt feature isn't enabled, then this will do core::panic, and if defmt is enabled it'll use defmt::panic.

You do need to use the debug formatter `{:?}` like this though:
```rust
Err(e) => panic!("Failed to configure PLL_USB: {:?}", e),
```
Or core::panic won't know how to print the enum.

embassy-rp/src/clocks.rs
```diff
@@ -136,15 +148,20 @@ pub enum PeriClkSrc {
...
-/// Core voltage regulator settings for RP2040.
+/// Core voltage regulator settings.
 ///
-/// The RP2040 voltage regulator can be configured for different output voltages.
+/// The voltage regulator can be configured for different output voltages.
 /// Higher voltages allow for higher clock frequencies but increase power consumption and heat.
+///
+/// **Note**: For RP235x the maximum voltage is 1.30V, unless unlocked by setting unless the voltage limit
+/// is disabled using the disable_voltage_limit field in the vreg_ctrl register. For lack of practical use at this
+/// point in time, this is not implemented here. So the maximum voltage in this enum is 1.30V for now.
```
CBJamo, line 158: This comment needs to be cleaned up, looks like some copy-paste errors.

embassy-rp/src/clocks.rs
```diff
@@ -168,24 +185,91 @@ pub enum CoreVoltage {
...
 
-#[cfg(feature = "rp2040")]
+/// Core voltage regulator settings.
+///
+/// The voltage regulator can be configured for different output voltages.
+/// Higher voltages allow for higher clock frequencies but increase power consumption and heat.
+///
+/// **Note**: For RP235x the maximum voltage is 1.30V, unless unlocked by setting unless the voltage limit
+/// is disabled using the disable_voltage_limit field in the vreg_ctrl register. For lack of practical use at this
+/// point in time, this is not implemented here. So the maximum voltage in this enum is 1.30V for now.
```
CBJamo, line 195: Same here

embassy-rp/src/clocks.rs
1-rafael-1: sorry :(

embassy-rp/src/clocks.rs
1-rafael-1: Thanks! I was sure I tried this... but since this now worked on first try I must have messed it up somehow.

Dirbaio (approved):

PR #4188, opened 2025-05-12
[embassy-usb-dfu] support function level WinUSB GUIDs

Gerharddc:
This PR makes it possible to provide function level msos GUIDs to usb_dfu. This helps to ensure that composite DFU devices automatically get assigned the WinUSB driver on Windows.

### DIFF
--- a/embassy-usb-dfu/src/application.rs
+++ b/embassy-usb-dfu/src/application.rs
@@ -2,7 +2,7 @@ use embassy_boot::BlockingFirmwareState;
 use embassy_time::{Duration, Instant};
 use embassy_usb::control::{InResponse, OutResponse, Recipient, RequestType};
 use embassy_usb::driver::Driver;
-use embassy_usb::{msos, Builder, Handler};
+use embassy_usb::{Builder, FunctionBuilder, Handler};
 use embedded_storage::nor_flash::NorFlash;
 
 use crate::consts::{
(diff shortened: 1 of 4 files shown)


Gerharddc:
This PR is ready for review now. After more extensive physical testing it turned out the comments I left were not entirely accurate, but that has been corrected now.

embassy-usb-dfu/src/application.rs
```diff
@@ -130,8 +130,23 @@ pub fn usb_dfu<'d, D: Driver<'d>, MARK: DfuMarker, RST: Reset>(
...
+        // Adding them here on the function level appears to only be needed for compositive devices.
+        // In addition to being on the function level, they should also be added to the device level.
+        //
+        func.msos_feature(msos::CompatibleIdFeatureDescriptor::new("WINUSB", ""));
+        func.msos_feature(msos::RegistryPropertyFeatureDescriptor::new(
+            "DeviceInterfaceGUIDs",
+            msos::PropertyData::RegMultiSz(winusb_guids),
+        ));
+    }
+
```
lulf, line 149: Could this have been a closure that gets the func passed, instead of an option parameter that seems very specific to winusb?

Gerharddc: Great suggestion, thanks! I can't see use-case for customizing `func` further right now, but it's definitely a more future proof approach. I'll give that approach a shot and see if it works out

Gerharddc: Ok, I have now updated this to use a closure instead. Unfortunately, I could not manage to put the closure inside an `Option` though since that results in all sorts of lifetime hell...

lulf (approved):
Thanks!

Gerharddc:
@lulf thanks for approving! Are more changes needed before you will be able to merge?

Issue #4110, opened 2025-04-18
Cannot use RTC functionality alongside low_power module


tomhampshire:
Using `low_power` requires you to relinquish control of the RTC peripheral to the executor. e.g:
```rust

    // give the RTC to the executor...
    let mut rtc = Rtc::new(p.RTC, RtcConfig::default());
    static RTC: StaticCell<Rtc> = StaticCell::new();
    let rtc = RTC.init(rtc);
    embassy_stm32::low_power::stop_with_rtc(rtc);
```
[source](https://docs.embassy.dev/embassy-stm32/git/stm32f412re/low_power/index.html)

This means that the RTC can not be subsequently used for the purposes of getting/setting the datetime, as this requires mutable references.

tomhampshire:
A suggestion from the embassy chat room was to implement a `split` function for the RTC - one part can be passed to the executor, and the other used for general datetime functionality (get/set)

tomhampshire:
An excerpt from the chat:

dirbaio
i'm not sure what the best fix would be
the lowpower executor does require the rtc to live forever
so it should prevent the user from dropping or reconfiguring it

danielb
Split it in two

dirbaio
yeah
so much boilerplate though
it'd be nice if it was "more automatic" like the other time drivers. you enable it in Cargo features then embassy_stm32::init does all the necessary init
so you don't have to explicitly configure rtc, split it, give one half to the executor

tomhampshire:
@Dirbaio what do you think of the following?

Proposal: Refactor embassy_stm32::rtc using split for Low-Power and Application Access

Problem:

The current `rtc::Rtc` struct combines low-power wakeup management and application-level time access (`now()`, `set_datetime()`). This creates an ownership conflict when `embassy_stm32::low_power::stop_with_rtc` requires exclusive control of the `Rtc` instance, preventing the application from simultaneously accessing time functions.

Proposed Solution:

Adopt the split pattern, common in embassy-stm32, to separate RTC functionalities:

1. Initialisation: `Rtc::new(p.RTC, config)` initialises the peripheral (clocks, prescalers) and returns a single, unified `Rtc` instance.
2. Splitting: Introduce a `rtc.split() -> (RtcControl, RtcClock)` method. This method consumes the `Rtc` instance.
   - `RtcControl`: This handle manages low-power wakeup functions and holds related state (e.g., `stop_time`). It's designed to be passed to `stop_with_rtc`.
[...]

PR #4191, opened 2025-05-13
embassy-stm32: split rtc low-power wakeup and time functions

tomhampshire:
Ref: #4110 

The current `rtc::Rtc` struct combines low-power wakeup management and application-level time access (`now()`, `set_datetime()`). This creates an ownership conflict when `embassy_stm32::low_power::stop_with_rtc` requires exclusive control of the `Rtc` instance, preventing the application from simultaneously accessing time functions.

Changes to existing API have been minimised.

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -230,6 +230,9 @@ pub struct Config {
     /// RCC config.
     pub rcc: rcc::Config,
 
+    /// RTC config.
+    pub rtc: Option<rtc::RtcConfig>,
+
     /// Enable debug during sleep and stop.
     ///
     /// May increase power consumption. Defaults to true.
@@ -283,6 +286,7 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             rcc: Default::default(),
+            rtc: None,
             #[cfg(dbgmcu)]
             enable_debug_during_sleep: true,
             #[cfg(any(stm32l4, stm32l5, stm32u5))]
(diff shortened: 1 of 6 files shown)


tomhampshire:
Submitted for initial review - I don't yet have the hardware to test if the device is entering low-power states, so this still needs testing

tomhampshire:
I have made an assumption that, if no RtcConfig is specified, the Rtc isn't configured. The other option is, if no RtcConfig is specified, the default config is used...

xoviat:
Unclear why this is needed--we already have time provider.

tomhampshire:
If you use embassy low-power, you cannot get/set the time from the RTC
> Unclear why this is needed--we already have time provider.

xoviat:
you absolutely can get the time with time provider, but you cannot set it. this is intentional because it may interfere with the wakeup timer timing. we can add an api to set it, but we may need to set the time in a critical section. in any case, it would require more thought than just splitting it.

tomhampshire:
My use case is: a standalone device which updates its RTC datetime with a host machine over USB. Therefore, the device would initialise with an unset (datetime) clock, and this would be set when "docking" with a host machine over USB. I'm unable to use the low-power features for this current use case, due to the need to get/set the clock values from/to the RTC. The alternatives are to not use the low-power features (my current solution), or to use a standalone i2c/SPI RTC module.


> you absolutely can get the time with time provider, but you cannot set it. this is intentional because it may interfere with the wakeup timer timing. we can add an api to set it, but we may need to set the time in a critical section. in any case, it would require more thought than just splitting it.

xoviat:
I will submit a pull request that addresses this.

xoviat:
actually, looking back at what was said in chat, this seems like this is somewhat of a valid approach. I will try to accomplish what  the idea was there.

Dirbaio:
Hello,

PR #4193, opened 2025-05-13
embassy-rp: Make bit-depth of I2S PIO program configurable

embediver:
The `PioI2sOutProgram` had a fixed 16bit sample depth, but that isn't reflected in the types name or the documentation.

The `PioI2sOut` output driver on the other hand accepts a `bit_depth` argument.
This was previously only used to calculate the clock frequency, while no other values than 16 made any sense.

This MR implements setting the sample bit-depth by setting it to scratch register `Y`, from where it is read by the PIO program. 

Also the channel argument is removed, since only 2 channels are supported.

### DIFF
--- a/embassy-rp/src/pio_programs/i2s.rs
+++ b/embassy-rp/src/pio_programs/i2s.rs
@@ -9,6 +9,10 @@ use crate::pio::{
 use crate::Peri;
 
 /// This struct represents an i2s output driver program
+///
+/// The sample bit-depth is set through scratch register `Y`.
+/// `Y` has to be set to sample bit-depth - 2.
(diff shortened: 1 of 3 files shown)


embediver:
Channel count wasn't really supported before and I don't think I2S could support it.
Atleast I don't know how I2S with only one channel would have to look, since a word clock (which is also the left/right clock) is mandatory.
And more than two channels is definitely not supported by standard I2S.

The channel argument was only used to calculate the clock frequency by calculating `sample_rate * bit_depth * channel_count`.
So now its just a fixed x2 in the calculation.

Dirbaio:
sounds good then. Thanks :)

PR #4197, opened 2025-05-14
Make futures-util and futures-sink optional

thetek42:
This PR makes futures-util and futures-sink within embassy-time and embassy-sync optional dependencies that can be enabled with feature flags. The flags are enabled by default as to hopefully not break to many people's code after updating their dependencies :)

### DIFF
--- a/embassy-sync/Cargo.toml
+++ b/embassy-sync/Cargo.toml
@@ -20,15 +20,18 @@ src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-sync/
 target = "thumbv7em-none-eabi"
 
 [features]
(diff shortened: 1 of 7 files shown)


Dirbaio:
I don't think we should add Cargo features unless the compile time improvement is very big. Adding features makes libs harder to use.

If you make them non-default, users will find something that the docs says should be there is not there, then they have to hunt down what feature enables it and add it.

If you make them default all it takes is *one* crate to forget `default-features=false` (like all crates using embassy-sync in this very repo :) ) and the compile time improvement is gone.

thetek42:
Good point. I would still be in favour of it, but do with this what you want ^^

PR #4199, opened 2025-05-14
Update doc comment for adc read to match current implementation

Jacke-debug:
The code in the doc comment did not compile due to updates to the function. After this change it now matches the code in the example folder.


Dirbaio (approved):
thanks!

PR #4201, opened 2025-05-14
#[test] Attribute Macro for embassy-executor

### LINKED ISSUE #4192: feature request: `#[test]` Attribute Macro for `embassy-executor` (std-only)
It would be great to have a `#[test]` attribute macro in `embassy-executor`, similar to [#tokio::test](https://docs.rs/tokio/latest/tokio/attr.test.html). This would enable running async unit tests within an Embassy runtime, without the need for boilerplate setup.

The feature would target `std` environments (e.g., desktop), and be especially useful for quickly testing logic-only components in isolation.

If this sounds like a useful addition, I'd be happy to work on an implementation.

AlixANNERAUD:
Resolves #4192

### DIFF
--- a/embassy-executor-macros/src/macros/test.rs
+++ b/embassy-executor-macros/src/macros/test.rs
@@ -1,0 +1,19 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, ItemFn};
+
+pub fn test(_args: TokenStream, input: TokenStream) -> TokenStream {
+    let input_fn = parse_macro_input!(input as ItemFn);
+    let fn_name = &input_fn.sig.ident;
+
+    let gen = quote! {
+        #[test]
+        fn #fn_name() {
+            embassy_executor::run(|spawner| async move {
+                #input_fn
+            });
+        }
+    };
+
+    gen.into()
+}
--- a/embassy-executor-macros/src/macros/test.rs
+++ b/embassy-executor-macros/src/macros/test.rs
@@ -19,6 +19,7 @@ pub fn test(_args: TokenStream, input: TokenStream) -> TokenStream {
         #[test]
         fn #fn_name() {
             let mut executor = ::embassy_executor::Executor::new();
+            let executor = unsafe { ::core::mem::transmute::<_, &'static mut ::embassy_executor::Executor>(&mut executor) };
             executor.run(|spawner| {
                 spawner.spawn(#task_name()).unwrap();
             });
--- a/embassy-executor-macros/src/macros/test.rs
+++ b/embassy-executor-macros/src/macros/test.rs
@@ -1,16 +1,26 @@
 use proc_macro::TokenStream;
-use quote::quote;
+use quote::{format_ident, quote};
 use syn::{parse_macro_input, ItemFn};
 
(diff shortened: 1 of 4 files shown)


embassy-executor-macros/src/macros/test.rs
```diff
@@ -0,0 +1,30 @@
...
+        async fn #task_name() {
+            #input_fn
+        }
+
+
+        #[test]
+        fn #fn_name() {
+            let mut executor = ::embassy_executor::Executor::new();
+            let executor = unsafe { ::core::mem::transmute::<_, &'static mut ::embassy_executor::Executor>(&mut executor) };
+            executor.run(|spawner| {
```
Dirbaio, line 23: this will run forever, the test won't complete even if the task returns.

embassy-executor-macros/src/macros/test.rs
```diff
@@ -0,0 +1,30 @@
...
+        #[allow(clippy::future_not_send)]
+        async fn #task_name() {
+            #input_fn
+        }
+
+
+        #[test]
+        fn #fn_name() {
+            let mut executor = ::embassy_executor::Executor::new();
+            let executor = unsafe { ::core::mem::transmute::<_, &'static mut ::embassy_executor::Executor>(&mut executor) };
```
Dirbaio, line 22: this is unsound, the executor really needs `&'static mut self`. Even if all tasks stop running, they keep pointers to the executors and anything can keep a pointer to a task through a waker, so you can't ensure the executor won't be use-after-free'd.

embassy-executor-macros/src/macros/test.rs
AlixANNERAUD: Wow, you're so fast! I didn't even have time to make the formatter happy.

You're right, that's actually one of my concerns.

Currently, `run` doesn't return even when there are no more pending tasks in the queue.
I think this could be fixed by changing how the executor behaves — maybe by implementing a custom `fn run_and_return`, which would, in addition to polling and waiting, also check if the queue is empty.

From what I can see, `raw::Executor` and its inner components (`SyncExecutor`, `RunQueue`, etc.) seem to have some mechanism to check whether the queue is empty, so it looks feasible.

That said, I'll need to implement all of that myself.

What do you think?

Dirbaio:
What is the use case for this? 

On std, you can test async logic that uses embassy-sync, embassy-time etc just fine using `#[tokio::test]`. All the embassy crates work just fine on non-embassy executors.

On no-std have you seen https://crates.io/crates/embedded-test ?

I'm not sure that we should have this in embassy-executor. It doesn't work on no-std, but embassy-executor is designed for no-std so people will inevitably try to use it on no-std and fail and then try to extend it until it does, at which point we'd have reinvented embedded-test. Doing it properly is hard, I'd rather people use embedded-test.

embassy-executor-macros/src/macros/test.rs
AlixANNERAUD: A little unsound indeed.
What would be the best approach then? Should I make it static inside the function body using a `StaticCell`?
That would require the end user to either import `static_cell` themselves, or for `static_cell` to be re-exported by Embassy.

PR #4212, opened 2025-05-16
fix: add bank confdiguration doc build features

lulf:


lulf:
Needs a new version of https://github.com/embassy-rs/docserver builder on the host

Dirbaio:
> Needs a new version of https://github.com/embassy-rs/docserver builder on the host

done, let's se eif it works

PR #4214, opened 2025-05-16
Impl Drop for stm32 Rng

DrTobe:
This disables the RNG in the CR::RNGEN register and calls `rcc::disable` to possibly disable the corresponding clock if the refcount goes to zero. Using the "low-power" feature, I have verified that this allows to enter STOP2 mode again after the RNG has been dropped.


lulf (approved):
Thanks!

PR #4216, opened 2025-05-16
embassy-rp (rp2040): Rtc wait_for_alarm

1-rafael-1:
This PR adds the ability to wait for an RTC alarm. 

## Reviewer note: 

+ **!!!Breaking change!!!** due to the fact that Rtc struct `new()` fn now expects Irqs
+ Implementation of binding interupt and handling waker inspired from adc.rs, this is my first time doing this, so I hope I got it reasonably right.
+ I feel like I cheated because I used a flag to mark that we had an interrupt occur - i found no good documentation ad no good place in the Rtc registers to read the same fact from there, and a bunch of tests with that got me nowhere. In case this was rather unenlightened... let me know please!
+ AI was used for syntax help and to answer my 100 questions on what stuff like compiler_fence does and how interrupts, wakers, ... work. Or in other words: Newbie mistakes are not unlikely.

I tested on a Pico board with the provided new example, so this works:
- `wait_for_alarm` no matter if one is yet scheduled yet or not
- alarm_scheduled` can determine if we have an alarm sceduled or not and if we have we get the DateTimeFilter out of it.

### DIFF
--- a/examples/rp/src/bin/rtc_alarm.rs
+++ b/examples/rp/src/bin/rtc_alarm.rs
@@ -1,0 +1,69 @@
+//! This example shows how to use RTC (Real Time Clock) for scheduling alarms and reacting to them.
+
+#![no_std]
+#![no_main]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_futures::select::{select, Either};
+use embassy_rp::bind_interrupts;
+use embassy_rp::rtc::{DateTime, DateTimeFilter, DayOfWeek, Rtc};
+use embassy_time::Timer;
+use {defmt_rtt as _, panic_probe as _};
+
+// Bind the RTC interrupt to the handler
+bind_interrupts!(struct Irqs {
+    RTC_IRQ => embassy_rp::rtc::InterruptHandler;
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_rp::init(Default::default());
+    let mut rtc = Rtc::new(p.RTC, Irqs);
+
+    if !rtc.is_running() {
+        info!("Start RTC");
+        let now = DateTime {
+            year: 2000,
+            month: 1,
+            day: 1,
+            day_of_week: DayOfWeek::Saturday,
+            hour: 0,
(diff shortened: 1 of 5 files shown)


examples/rp/src/bin/rtc_alarm.rs
```diff
@@ -0,0 +1,69 @@
...
+                info!(
+                    "Now: {}-{:02}-{:02} {}:{:02}:{:02}",
+                    dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
+                );
+
+                // See if the alarm is already scheduled, if not, schedule it
+                match rtc.alarm_scheduled() {
+                    None => {
+                        info!("Scheduling alarm for 30 seconds from now");
+                        let next_30s = if dt.second == 59 { 0 } else { dt.second + 30 };
```
8cichy8, line 53: Why is there check for 59? Shouldn't this be: dt.second > 29?

1-rafael-1:
@Dirbaio the failure in the merge queue was my fault. I thought these tests would also run in CI but they don't and the test had a timing issue. I extracted the test into an example an re-ran that, then after fixing & successful test introduced that back into the test.

Sorry for wasting Your time! :(

Is there a way one can run the teleprobe rig from a developers end?

Dirbaio:
you can run it locally like this:
- cargo install --git https://github.com/embassy-rs/teleprobe
- cd tests/rp
- uncomment this https://github.com/embassy-rs/embassy/blob/main/tests/rp/.cargo/config.toml#L9
- cargo run --release --bin rtc --features rp2040

if you don't have a rpi pico ping me on Matrix and i'll give you a token you can use to remotely run tests on the HIL test farm.

1-rafael-1:
@Dirbaio 
cool, learned a new thing  -thx!
Picos I have plenty.... :-)

I had to use 
```bash
cargo run --release --bin rtc --features rp2040 --target thumbv6m-none-eabi
```
to run it, but besides that it worked first try.

Test passes. So merge queue should be happy now. 
As a bonus I now know how to run all the tests next time around and can avoid a roundtrip.

PR #4223, opened 2025-05-20
Cleanup Pins and RCC for FDCAN on STM32

DrTobe:
This PR uses the reference counting logic already implemented in the `fdcan` module to also call `set_as_disconnected` on the pins and to possibly disable the RCC clock. Before, only the "buffered" readers and writers have been counted to set the modes back to `NonBuffered` after the last one had been dropped. Now, all readers and writers are counted, a TX _and_ RX type like `Can` or `BufferedCan` counts as both.

This works because there is only one direction to initialize the types: `CanConfigurator` -> `Can` (unbuffered) -> `BufferedCan`/`BufferedCanFd` (-> drop). In other words: Since there is no way to go back from buffered to unbuffered without stealing the peripheral, users should definitely drop the old instance before. So when the last active instance is dropped, the modes are set back to `NonBuffered` and pins and RCC are cleaned up. When a new "peripheral steal" happens, the peripheral should be in the expected (initialization) state.

Additionally, the `CanConfigurator` must already be considered as a writer+reader because it already enables the pins and RCC clock, so it also needs to trigger the cleanup.

But to be honest, I am not 100% satisfied with this implementation because it is not very elegant in several ways:
1. The `NotifyCreated` events must be triggered manually whenever one of the several CAN types ist created, and for every CAN type, a separate `Drop` impl must be supplied (all being nearly identical). And there are many CAN types: `CanConfigurator`, `Can`, `BufferedCan`, `BufferedCanFd`, and the separate `Tx` and `Rx` types (unbuffered and buffered, FD and classic). Instead, I would rather like to implement an RAII guard with a corresponding constructor, `Clone` impl and `Drop` impl and make all CAN types contain the required guard(s). This would be a less error prone implementation.
2. There seems to be some amount of duplication: First, the `SealedInstance` contains `Info`, `Registers` and `internal_operation`, but the `Info` struct contains `Registers` and `internal_operation`, too. Second, `calc_timestamp` is implemented for the `Info` type and for the `SealedInstance`, too. Some of this duplication is caused by the need to work around type erasures, but I guess this could be cleaned up. Specifically, I think we could split all static information into the `Info` and `State` types and make all CAN types contain one `&'static Info` and one `State` reference/pointer. Furthermore, I guess that the `internal_operation` and the `calc_timestamp` implementation can be moved out of or be removed from the `SealedInstance`, respectively.
3. At some places, we reference a `&'static State` which is sometimes mutably accessed via `self.state as *const State as *mut State` followed by a dereference. This may work and seems to be ok if we think about it in terms of pointers and memory addresses, but according to [the Rustonomicon](https://doc.rust-lang.org/nomicon/transmutes.html), transmuting a shared reference to a mutable reference is absolutely forbidden, and that's what we are effectively doing here. So I would suggest to be honest about what we do here and _at least_ use a `*mut State` directly instead of a `&'static State`. The accesses are and need to be unsafe anyways. But even better, why don't we use a `critical_section::Mutex` here?

For now, I have only implemented the naive approach because I would like to get some feedback on the three intended changes first. This is because these seem to be larger changes and I would not like to do all the work without it being merged eventually. So, what do you think about the suggested refactoring? Otherwise, I am also fine with merging this as is, or, we could merge this for now and do the refactoring afterwards. This last approach would be nice because the cleanup feature gets available already and the following refactoring could be split in small(er) parts.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -10,7 +10,7 @@ use embassy_sync::channel::Channel;
 use embassy_sync::waitqueue::AtomicWaker;
 
 use crate::can::fd::peripheral::Registers;
-use crate::gpio::{AfType, OutputType, Pull, Speed};
+use crate::gpio::{AfType, OutputType, Pull, SealedPin as _, Speed};
 use crate::interrupt::typelevel::Interrupt;
 use crate::rcc::{self, RccPeripheral};
 use crate::{interrupt, peripherals, Peri};
@@ -187,12 +187,16 @@ impl<'d> CanConfigurator<'d> {
 
         rcc::enable_and_reset::<T>();
 
+        let info = T::info();
+        let state = unsafe { T::mut_state() };
+        state.tx_pin_port = Some(tx.pin_port());
+        state.rx_pin_port = Some(rx.pin_port());
+        (info.internal_operation)(InternalOperation::NotifySenderCreated);
+        (info.internal_operation)(InternalOperation::NotifyReceiverCreated);
+
         let mut config = crate::can::fd::config::FdCanConfig::default();
         config.timestamp_source = TimestampSource::Prescaler(TimestampPrescaler::_1);
         T::registers().into_config_mode(config);
-
-        rx.set_as_af(rx.af_num(), AfType::input(Pull::None));
-        tx.set_as_af(tx.af_num(), AfType::output(OutputType::PushPull, Speed::VeryHigh));
 
         unsafe {
             T::IT0Interrupt::unpend(); // Not unsafe
@@ -204,8 +208,8 @@ impl<'d> CanConfigurator<'d> {
         Self {
             _phantom: PhantomData,
             config,
-            info: T::info(),
-            state: T::state(),
+            info,
+            state,
             properties: Properties::new(T::info()),
             periph_clock: T::frequency(),
         }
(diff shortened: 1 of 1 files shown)


cschuhen:
Hi, @DrTobe , @Dirbaio,
The intent of the patch seems reasonable. I can test it out in on my own projects in the next couple of days(and review the patch more thoroughly) however I implemented the reference counting in response to how others were using the driver, not my own use-case.

It seems reasonable to me to follow up on the separate points as independent patches. Especially points 2 & 3 which are quite unrelated.

As for your points:

1. Sounds like a good idea. I'm generally a fan of trying to incorporate RAII. If it is just literally based on what is there now, wouldn't the guard need a copy of info for this to work? Would we incorporate this into info?
2. Yes, agreed. I've never been really happy with the amount of duplication that I put into the driver. Especially between the Classic and Fd variants (which is not what you are talking about). 
3. Yes, fair point. I guess with my lack of rust experience, I didn't think of using a critical section mutex. I've always considered locking mutexes as forbidden from interrupts but I guess a critical section mutex isn't really a mutex in that sense, it's just a critical section with semantics to explain what is being protected. My attempt to at least make things not be totally broken was to still use a critical section when accessing state as mut. The biggest real gap that I can think of is if some task operating on the Can classes was higher priority than the CAN interrupt itself. This would only be an issue if using the buffered modes, which makes totally no sense to have the priorities that way.... Anyway, my biggest concern would be how much overhead is in the mutex, especially when it would need to be locked for every interrupt.

lulf (approved):
LGTM, but I'll await review/testing from @cschuhen before merging.

cschuhen:
I've tested the branch with my steup now and it's working fine. Though I don't really exercise the use case that was the initial trigger for the reference counting. I've gone over the PR and LGTM.

DrTobe:
I will test the newest `main` with our application code, but I expect this to work.

Additionally, I will have a look at the others things I suggested to clean that up a bit now.

Thanks for merging.

cschuhen:
@DrTobe, I started looking at (3) in your points. Actually, it prompted me to create:  https://github.com/embassy-rs/embassy/pull/4262#pullrequestreview-2875109855

PR #4227, opened 2025-05-20
mspm0: generate interrupt group handlers

i509VCB:
No more ugly `#[path]` attributes or manually written code. The only annoying part is that the user must tell rustc to link the `interrupt_group.x` file.

This does require a little linker *magic*, inspired by `cortex-m-rt`. In the future for other things in interrupt groups I'll write an abstraction to allow binding these. When `#[linkage = "extern_weak"]` is promoted to stable then we can have a far nicer setup than this.

GPIO examples work correctly from my testing. (I should write some tests for these soon eh?).


felipebalbi (approved):
Looks much nicer :)

PR #4219, opened 2025-05-19
Add support for rtc_v3h7rs

aaamorgan:
Updates to be compatible with https://github.com/embassy-rs/stm32-data/pull/609

With these changes, I could successfully run the rtc stm32h7rs example.

PR #4228, opened 2025-05-20
Adding EEPROM support to embassy-stm32

okhsunrog:
Adding EEPROM support for STM32L0 and STM32L1 series


okhsunrog:
CI is failing for now, https://github.com/embassy-rs/embassy/pull/4219/commits/9343200e442d94e5e4e55a3a42a5d20ce04b5abc fixes it for me.
Waiting for https://github.com/embassy-rs/embassy/pull/4219 to get merged so I can fix the CI

okhsunrog:
Added examples, tested on STM32L071C8T6, EEPROM works for me

embassy-stm32/src/flash/eeprom.rs
```diff
@@ -0,0 +1,226 @@
...
+        let start = offset;
+        let misalign = (start % 4) as usize;
+        let prefix_len = if misalign == 0 {
+            0
+        } else {
+            (4 - misalign).min(data.len())
+        };
+        let (prefix, rest) = data.split_at(prefix_len);
+        let aligned_len = (rest.len() / 4) * 4;
+        let (aligned, suffix) = rest.split_at(aligned_len);
```
felipebalbi, line 166: isn't this somewhat equivalent to `let (prefix, shorts, suffix) = data.align_to::<u32>();`

okhsunrog: Not quite, `data.align_to::<u32>() ` would split based on alignment in data, my code is splitting based in alignment in EEPROM offset. Actually my code has an UB,  I shouldn't  just do `let u32_data = core::slice::from_raw_parts(aligned.as_ptr() as *const u32, aligned.len() / 4);`, it can cause an UB, as I don't check data alignment. I'll fix it in the next commit

okhsunrog:
@felipebalbi now it looks better, I believe

okhsunrog:
bender run

okhsunrog:
I think this is ready to be merged, waiting for approval

okhsunrog:
@lulf actually the methods are on the same Flash type, otherwise we can't use Flash and Eeprom at the same time. Dario suggested that I just add methods for eeprom to the existing type

lulf:
Ah, that makes sense, thanks!

okhsunrog:
@lulf what happened, why it was removed from merge queue?

okhsunrog:
looks like rpi-pico hil tests are failing for some reason?

lulf:
Yeah, this is an issue with the CI rig I believe. It's been  like this for a while, and I think it's just a hardware issue (CC @Dirbaio)

PR #4231, opened 2025-05-21
feat(stm32): provide a `const` constructor on `rcc::Config`

### LINKED ISSUE #4213: STM32: introducing a `const` constructor for `rcc::Config`
Currently, the only way of instantiating an `embassy_stm32::rcc::Config` is through the `Default` trait, which of course makes it not instantiable from `const` contexts.

Would you consider a PR that introduces a `const` `new()` constructor method on that type (and on the children types required for that), so that it is possible to instantiate it in `const` contexts?

ROMemories:
## Description

Currently, the only way of instantiating an `embassy_stm32::rcc::Config` is through the `Default` trait, which makes it not instantiable from `const` contexts.

This PR introduces a `const` `embassy_stm32::rcc::Config::new()` constructor.

## How to review this PR

- This PR is better reviewed commit by commit: each commit targets one STM32 (sub-)family, except for the first one which *also* introduces a `const` constructor for `Hsi48Config` and `LsConfig`. I can easily split that first commit or, on the opposite, squash all this PR's commits into a single one if you prefer.
- Please do check that this PR is complete and that I didn't forget an STM32 family or some feature-gated field. I did do a basic sanity check by grepping for `::default(` in `embassy-stm32/src/rcc` (the remaining seems irrelevant to this PR).

## Issues references

Closes #4213.

### DIFF
--- a/embassy-stm32/src/rcc/bd.rs
+++ b/embassy-stm32/src/rcc/bd.rs
@@ -92,6 +92,16 @@ pub struct LsConfig {
 }
 
 impl LsConfig {
+    pub const fn new() -> Self {
+        // on L5, just the fact that LSI is enabled makes things crash.
+        // TODO: investigate.
+
+        #[cfg(not(stm32l5))]
+        return Self::default_lsi();
+        #[cfg(stm32l5)]
+        return Self::off();
+    }
+
     pub const fn default_lse() -> Self {
         Self {
             rtc: RtcClockSource::LSE,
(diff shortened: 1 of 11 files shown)


embassy-stm32/src/rcc/bd.rs
```diff
@@ -92,6 +92,16 @@ pub struct LsConfig {
 }
 
 impl LsConfig {
+    pub const fn new() -> Self {
```
ROMemories, line 95: `new()` may not the best name for this constructor given its behavior?

embassy-stm32/src/rcc/c0.rs
```diff
@@ -59,9 +59,9 @@ pub struct Config {
     pub mux: super::mux::ClockMux,
 }
 
-impl Default for Config {
+impl Config {
     #[inline]
```
ROMemories, line 63: I added an `#[inline]` attribute on `new()` constructors when the `Default::default()` function had one. Should I remove it, or should I instead also add that attribute to every `Default::default()` as they only call `new()`?

ROMemories: I removed all the `#[inline]` attributes as they were not reducing the code size for me when compiling for the C0/WBA/L lines. They're easy to get back later if needed.

PR #4232, opened 2025-05-21
Add PWM waveform generation using DMA burst mode

RaulIQ:
This PR adds the waveform_up_multi_channel method, enabling PWM waveform generation using DMA burst transfers triggered by timer update events (UEV).

DMA operates through the DMAR register using configured DBA/DBL to update multiple CCRx registers in one transfer.

image


lulf (approved):
Thanks!

PR #4233, opened 2025-05-22
Hide embedded-hal 0.2 implementations behind a feature

lynxD:
Motivation:
    Make it easy to ensure that one does not use embedded-hal 0.2 API by mistake.
    Show intent to deprecate embedded-hal 0.2 compatibility eventually.

I've cargo check-ed all modified crates and examples. All of those that pass on master also pass with this change.
Do tell me what else needs to be done. Entries in changelogs?

I've also fixed some warnings that I've stumbled upon.

### DIFF
--- a/embassy-embedded-hal/src/adapter/blocking_async.rs
+++ b/embassy-embedded-hal/src/adapter/blocking_async.rs
@@ -1,5 +1,3 @@
-use embedded_hal_02::blocking;
-
 /// Wrapper that implements async traits using blocking implementations.
 ///
 /// This allows driver writers to depend on the async traits while still supporting embedded-hal peripheral implementations.
@@ -21,18 +19,24 @@ impl<T> BlockingAsync<T> {
 //
 // I2C implementations
 //
+#[cfg(feature = "embedded-hal-02")]
 impl<T, E> embedded_hal_1::i2c::ErrorType for BlockingAsync<T>
 where
     E: embedded_hal_1::i2c::Error + 'static,
-    T: blocking::i2c::WriteRead<Error = E> + blocking::i2c::Read<Error = E> + blocking::i2c::Write<Error = E>,
+    T: embedded_hal_02::blocking::i2c::WriteRead<Error = E>
+        + embedded_hal_02::blocking::i2c::Read<Error = E>
+        + embedded_hal_02::blocking::i2c::Write<Error = E>,
 {
     type Error = E;
 }
 
+#[cfg(feature = "embedded-hal-02")]
 impl<T, E> embedded_hal_async::i2c::I2c for BlockingAsync<T>
 where
     E: embedded_hal_1::i2c::Error + 'static,
-    T: blocking::i2c::WriteRead<Error = E> + blocking::i2c::Read<Error = E> + blocking::i2c::Write<Error = E>,
+    T: embedded_hal_02::blocking::i2c::WriteRead<Error = E>
+        + embedded_hal_02::blocking::i2c::Read<Error = E>
+        + embedded_hal_02::blocking::i2c::Write<Error = E>,
 {
     async fn read(&mut self, address: u8, read: &mut [u8]) -> Result<(), Self::Error> {
         self.wrapped.read(address, read)
@@ -61,18 +65,20 @@ where
 // SPI implementatinos
 //
 
+#[cfg(feature = "embedded-hal-02")]
 impl<T, E> embedded_hal_async::spi::ErrorType for BlockingAsync<T>
 where
     E: embedded_hal_1::spi::Error,
-    T: blocking::spi::Transfer<u8, Error = E> + blocking::spi::Write<u8, Error = E>,
+    T: embedded_hal_02::blocking::spi::Transfer<u8, Error = E> + embedded_hal_02::blocking::spi::Write<u8, Error = E>,
 {
     type Error = E;
 }
 
+#[cfg(feature = "embedded-hal-02")]
 impl<T, E> embedded_hal_async::spi::SpiBus<u8> for BlockingAsync<T>
 where
     E: embedded_hal_1::spi::Error + 'static,
-    T: blocking::spi::Transfer<u8, Error = E> + blocking::spi::Write<u8, Error = E>,
+    T: embedded_hal_02::blocking::spi::Transfer<u8, Error = E> + embedded_hal_02::blocking::spi::Write<u8, Error = E>,
 {
     async fn flush(&mut self) -> Result<(), Self::Error> {
         Ok(())
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -6,7 +6,7 @@ use core::ptr;
 
 use embassy_embedded_hal::SetConfig;
 use embassy_futures::join::join;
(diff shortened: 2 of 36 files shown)


lynxD:
I'm yet to check anything with an embedded-hal 0.2 device driver.

There are things that I'm not sure about. I'll mark them with a comments

embassy-stm32/src/spi/mod.rs
```diff
@@ -6,7 +6,7 @@ use core::ptr;
 
 use embassy_embedded_hal::SetConfig;
 use embassy_futures::join::join;
-pub use embedded_hal_02::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_3};
+pub use embedded_hal_1::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_3};
```
lynxD, line 9: Need to make sure if this won't break embedded 0.2 based device drivers.

embassy-embedded-hal/src/adapter/blocking_async.rs
```diff
@@ -21,18 +19,24 @@ impl<T> BlockingAsync<T> {
 //
 // I2C implementations
 //
+#[cfg(feature = "embedded-hal-02")]
 impl<T, E> embedded_hal_1::i2c::ErrorType for BlockingAsync<T>
 where
     E: embedded_hal_1::i2c::Error + 'static,
-    T: blocking::i2c::WriteRead<Error = E> + blocking::i2c::Read<Error = E> + blocking::i2c::Write<Error = E>,
+    T: embedded_hal_02::blocking::i2c::WriteRead<Error = E>
```
lynxD, line 26: Need to change to embedded_hal_1

lynxD:
TODO: Update code that requires embedded-hal 0.2 from examples. // We shouldn't be promoting use of legacy APIs.

lulf:
We'd like to avoid this for [these reasons](https://github.com/embassy-rs/embassy/pull/3302)

lynxD:
Ha ha, I haven't thought this was tried before.

Could you help clarify a few points for me?

> we intentionally don't feature gate trait impls. They're confusing enough for newcomers, especially with the two embedded-hal versions. Adding feature gates adds an extra layer of complexity on top, where someone can get stuck not working even rustdocs say it should work because they forgot to enable the feature.

This could be addressed by adding reminders to docstrings of all gated features and even making the embeded-hal-02 feature default. Is that not good enough?


> The whole point of trait crates like embedded-hal is to enable interoperability between crates. If a driver is in e-h0.2 because the author hasn't gotten around to update it or it's unmaintained, it can still work perfectly with the latest embassy, as long as we keep the impls. There's no downside in using an old e-h, it works and that's all the user cares about.
> 
> Removing the impls because "they're old, and old bad" is hurting interoperability for no benefit.

I can see some downsides:
1. Unless we actively get rid of e-h0.2 code from our examples users will copy it to their new projects. There are some present now. Gating the e-h0.2 would allow us to see any back slips with our examples at a glance. Currently we have examples using e-h0.2 api in stm32f4, stm32h7 and stm32l4.
2. It's not "old therefore bad". It's "unstable therefore bad". Do we have any guarantee that there won't be any breaking changes to e-h0.2 in the future?
3. Is it really a good idea to take on technical debt while embassy is 0.x for a interface that is 0.x so the users can use old and unmaintained drivers that themselves are (mostly) 0.x? When somebody wants to use an old driver they should just update it(fork it if it's unmaintained). This way everyone will benefit.
4. It just looks bad when your project depends on unstable crates. I can imagine managers in a corporate setting using that against proposal to use embassy for a new project.
5. What about if(when) e-h2.0 comes out? defmt ran into similar problem: https://github.com/knurling-rs/defmt/issues/877
I'd add feature flags for all major versions. This way projects that actively get rid of legacy could avoid accidental back-slipping.

Backwards compatibility is good but I don't believe it's a good idea to force it on all users. I'd think it should be opt-in. At very least  it should be opt-out.

lulf:
I've tried to address most of your points under, but it feels to me like many of the arguments carry a lot of anecdotal evidence.

The summary is that:

1. It really isn't a big deal to maintain 0.2 impls at this point
2. We'd like to make it as friction-less as possible to use existing drivers.
3. 0.2 implementations doesn't get in the way of anyone not using them.

> Ha ha, I haven't thought this was tried before.
> 
> Could you help clarify a few points for me?
> 
> > we intentionally don't feature gate trait impls. They're confusing enough for newcomers, especially with the two embedded-hal versions. Adding feature gates adds an extra layer of complexity on top, where someone can get stuck not working even rustdocs say it should work because they forgot to enable the feature.
> 
> This could be addressed by adding reminders to docstrings of all gated features and even making the embeded-hal-02 feature default. Is that not good enough?
> 
> > The whole point of trait crates like embedded-hal is to enable interoperability between crates. If a driver is in e-h0.2 because the author hasn't gotten around to update it or it's unmaintained, it can still work perfectly with the latest embassy, as long as we keep the impls. There's no downside in using an old e-h, it works and that's all the user cares about.
> > Removing the impls because "they're old, and old bad" is hurting interoperability for no benefit.
> 
> I can see some downsides:
> 
>     1. Unless we actively get rid of e-h0.2 code from our examples users will copy it to their new projects. There are some present now. Gating the e-h0.2 would allow us to see any back slips with our examples at a glance. Currently we have examples using e-h0.2 api in stm32f4, stm32h7 and stm32l4.
> 

That's a fair point, and something to fix, but its fixable without having to make e-h 0.2 opt-in.

>     2. It's not "old therefore bad". It's "unstable therefore bad". Do we have any guarantee that there won't be any breaking changes to e-h0.2 in the future?
> 

I'm not sure what you mean by unstable here. It's been around for a long time, and I doubt e-h 0.2 will see any changes in the future, so I don't think that's a concern. 

>     3. Is it really a good idea to take on technical debt while embassy is 0.x for a interface that is 0.x so the users can use old and unmaintained drivers that themselves are (mostly) 0.x? When somebody wants to use an old driver they should just update it(fork it if it's unmaintained). This way everyone will benefit.
> 

The technical debt for embassy in this case is very small and I generally see very few changes to this part of the code, because it usually shares the implementation with embedded-hal 1.0.

>     5. What about if(when) e-h2.0 comes out? defmt ran into similar problem: [Allow bitflags `v1` and `v2` to coexist knurling-rs/defmt#877](https://github.com/knurling-rs/defmt/issues/877)
>        I'd add feature flags for all major versions. This way projects that actively get rid of legacy could avoid accidental back-slipping.
> 

I imagine that if e-h 2.0 comes out, we'd still support both 1.0 and 0.2 if it's not a maintenance burden and it still re-uses the implementations as today.

> 
> Backwards compatibility is good but I don't believe it's a good idea to force it on all users. I'd think it should be opt-in. At very least it should be opt-out.

For users that don't want to use 0.2, they won't be affected by embassy implementing the 0.2 traits, so I don't understand this concern.

Dirbaio:
Two of the the Embassy project goals are:
- Good interoperability with the ecosystem.
- Easy to get started, especially for Rust newcomers.

Removing embedded-hal 0.2 goes against these goals. Some of the concerns you've shared are valid, but the project goals take priority.

PR #4234, opened 2025-05-22
Prepare embassy-sync 0.7

mkj:
I've updated the changelog for embassy-sync (I'd like to see a crates.io release with the zerocopy-channel fix).
I think this needs a major bump because of the critical-section std change.

If there are other planned pending embassy-sync changes please don't let this rush a release.


lulf (approved):
Thanks for doing that!

PR #4236, opened 2025-05-22
Support RGBW on PioWs2812

MatrixSenpai:
Some ws2812 (aka neopixel) devices have an additional white led. The current pio program does not support these and can cause weird shifting issues due to using only 24 bits on the FIFO config threshold. Add an optional program that supports this white led, and supports using `smart_led::RGBW<u8>`

See [this repository](https://github.com/MatrixSenpai/ripple/blob/6dcf3a26aa6b17bc549177ae20b9e43c298fc95e/binary/src/ws2812/pio.rs#L50-L94) for the original working implementation.

### DIFF
--- a/embassy-rp/src/pio_programs/ws2812.rs
+++ b/embassy-rp/src/pio_programs/ws2812.rs
@@ -2,7 +2,7 @@
 
 use embassy_time::Timer;
 use fixed::types::U24F8;
-use smart_leds::RGB8;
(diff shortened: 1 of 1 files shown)


MatrixSenpai:
There are multiple potential solutions for this problem, the approach i took was to make no changes to the existing api, adding additional compatibility if the user prefers. If another approach is preferred, I can make updates!

PR #4237, opened 2025-05-22
embassy-rp: implement input/output inversion

felipebalbi:
RP2040/RP23xx support inversion in HW of the inputs and outputs. Implement minimal support for that.


Dirbaio:
thanks!

PR #4244, opened 2025-05-23
executor: Make state implementations and their conditions match

janderholm:
Hi!

I am playing around with embassy on AVR (an ATmega328p) and stumbled into some oddities. There are three implementations of the executor state that are conditionally selected based on the existance of 8-bit atomics. All of them however use 32-bit integers.

Since only 2 bits are actually used, I have modified the non ARM specific ones to use a u8 instead of a u32.  In the case of AVR, there are no 8-bit atomics but using a u8 with a critical section instead of u32 still save around 60-bytes or so in a minimal test.

I might be missing something obvious so feel free to reject the pull request.


Dirbaio (approved):
this is a very nice find. thank you :)

PR #4245, opened 2025-05-24
nrf52840: example ieee 802.15.4 packet send/receive

bobsrac:


bobsrac:
candidate examples for nrf52840 ieee802.15.4 send/receive

PR #4249, opened 2025-05-24
embassy-stm32: add support for flash async mode

iamadahn:
This adds ability to configure flash to work in Async mode, which let's you use https://docs.rs/sequential-storage for example, I tested it with example and everything went fine.

### DIFF
--- a/embassy-stm32/src/flash/f1f3.rs
+++ b/embassy-stm32/src/flash/f1f3.rs
@@ -1,9 +1,24 @@
 use core::ptr::write_volatile;
 use core::sync::atomic::{fence, Ordering};
 
+use embassy_sync::waitqueue::AtomicWaker;
+use pac::flash::regs::Sr;
+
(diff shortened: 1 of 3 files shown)


iamadahn:
fixed ci, was failing since I forgot to add back flash_f3 feature after tinkering. Tested provided examples with STM32F103C8T6 on custom board and STM32F303RE on nucleo board. It's my first ever pr, so please guide me anything else must be done before merge

RAnders00:
This looks great, why did you close the PR?

iamadahn:
Since nobody answered here, I joined their matrix chat and asked what should I do next and I don't clearly remember the exact answer, but I was told smth like it still won't be really async (I don't remember the exact why) and there is a reason why async flash is implemented only for stm32f4 and that I should use embassy_embedded_hal::adapter::BlockingAsync wrapper to add needed traits, which I did and it was working as expected

PR #4257, opened 2025-05-27
start referencing examples in embassy-stm32

robamu:
I think some references in the docs itself to the examples would be extremely useful. This is a first attempt for embassy-stm32. There are a lot of examples, so I always used STM32F4 for now, which seems to contain the most examples.

### DIFF
--- a/embassy-stm32/src/adc/mod.rs
+++ b/embassy-stm32/src/adc/mod.rs
@@ -1,4 +1,8 @@
 //! Analog to Digital Converter (ADC)
+//!
+//! ## Examples
+//!
+//! - [STM32F4](https://github.com/embassy-rs/embassy/blob/main/examples/stm32f4/src/bin/adc.rs)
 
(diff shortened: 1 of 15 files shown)


lulf:
This is the sort of stuff that gets out of date very quickly though. Also it's referencing the main branch, which won't be for a released crate.

robamu:
Hmm, introducing a link to a tag would introduce overhead to update the links on every release though..
As it is right now, users could at least click on the link, and select the appropriate tag for a specific version.

I definitely agree that that can get out of date quickly. If some link checker was able to verify that the link does not point to 404, the links becoming invalid could be avoided through tooling.

robamu:
Alternatively: Centralize documentation by peripheral module inside `examples/README.md` (maybe some other tool can check the links more easily) and only refer to that document. I think markdown can also reference local files/paths..

Dirbaio:
Hello,

Issue #4226, opened 2025-05-20
Docs: clarify how PAC usage is "unsafe"


natevw:
In the Embassy book here https://github.com/embassy-rs/embassy/blob/ef32187ed7349f3883d997b6f1590e11dbc8db81/docs/pages/layer_by_layer.adoc?plain=1#L11 the claim is made that a PAC:

> does not prevent you from writing unsafe code.

It goes on to give an example blinky app using PAC, which at no point uses any actual `unsafe` blocks. So I do not understand what this introduction is trying to demonstrate. As a beginner I am wondering/guessing that it may be one of these:

* that perhaps in Rust embedded cultural practice, the entire PAC crate as a whole is simply considered unsafe and just does not mark its functions as such
  * but I don't find any mention of this in the docs for the crate itself https://docs.embassy.dev/stm32-metapac/git/stm32c011d6/index.html
  * and I see examples to the contrary in e.g. https://docs.rust-embedded.org/book/start/registers.html where *some* of the closures do need to deal with `unsafe` while most do not
* that perhaps the example is really more pseudo-code and in practice would need to be wrapped in various `unsafe` caveats
  * I didn't try compile/run it but the code otherwise looks like it's supposed to work as-is so I'd be a little surprised if it were broken?
  * especially since in the context I would expect the example to deliberately *emphasize* just how unsafe using a PAC is given the intro…
* or perhaps the introductory paragraph is using "unsafe" not in a technical Safe Rust vs. Unsafe Rust way, but more of just a colloquial e.g. "you might configure a register wrong and then your code/board won't work right"?
  * but you could still make those sorts of logic/hardware mistakes with a HAL crate to some extent too
  * and honestly it seems unlikely given the marketing around Rust that such a core term as "unsafe code" would get used generically/casually when talking about low-level API options here

So none of these guesses seems quite right. Could the wording of this section be updated to clarify what it's trying to say?

PR #4259, opened 2025-05-27
Update layer_by_layer.adoc to clarify PAC limitations

natevw:
proposed fix for #4226, re. https://github.com/embassy-rs/embassy/issues/4226#issuecomment-2908772500

### DIFF
--- a/docs/pages/layer_by_layer.adoc
+++ b/docs/pages/layer_by_layer.adoc
@@ -8,7 +8,7 @@ The application we'll write is a simple 'push button, blink led' application, wh
 
 == PAC version
 
-The PAC is the lowest API for accessing peripherals and registers, if you don't count reading/writing directly to memory addresses. It provides distinct types to make accessing peripheral registers easier, but it does not prevent you from writing unsafe code.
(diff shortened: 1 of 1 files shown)


i509VCB:
I still think it is valuable to emphasize a pac *may* result in unsafe code if you program the hardware incorrectly (DMA is the primary example of this). lulf did mention this in the linked issue.

Technically I think pac crates should be 100% unsafe, but that would only result in a lot of extra `unsafe {}` blocks with no real benefit unless you explain every single instance of unsafe access.

workingjubilee:
> Technically I think pac crates should be 100% unsafe, but that would only result in a lot of extra `unsafe {}` blocks with no real benefit unless you explain every single instance of unsafe access.

Constructing a type `unsafe`ly... even a ZST that can be created anywhere... that then exposes the API of the crate using safe functions would allow you to contain the `unsafe` to a single invocation. The ergonomics would still take a small hit from occasionally having to will that ZST into existence using a line of code, but I think "you have to use `unsafe` code to cause memory corruption" is a worthwhile property to preserve in Rust programs.

PR #4262, opened 2025-05-28
Enable Sync and Send for DynamicSender and DynamicReceiver.

cschuhen:
DynamicSender and DynamicReceiver, just seem to be a fat pointer to a Channel which is already protected by it's own Mutex already. In fact, you can share the Channel already betwen threads and create Dynamic*er's in the target threads. It should be safe to share the Dynamic*er's directly.

### DIFF
--- a/embassy-sync/src/channel.rs
+++ b/embassy-sync/src/channel.rs
@@ -164,6 +164,57 @@ impl<'ch, T> DynamicSender<'ch, T> {
     }
 }
 
+/// Send-only access to a [`Channel`] without knowing channel size.
+/// This version can be sent between threads but can only be created if the underlying mutex is Sync.
+pub struct SendDynamicSender<'ch, T> {
+    pub(crate) channel: &'ch dyn DynamicChannel<T>,
+}
+
+impl<'ch, T> Clone for SendDynamicSender<'ch, T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<'ch, T> Copy for SendDynamicSender<'ch, T> {}
+unsafe impl<'ch, T: Send> Send for SendDynamicSender<'ch, T> {}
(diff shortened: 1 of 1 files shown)


cschuhen:
Oh, C++ brain. Keep forgetting about Mutexes that don't. So, what are the options here? Add SendableDynamicSender and SendableDynamicReceiver that will only get issued if the right type of Mutex is used?

cschuhen:
@lulf, does the attached patch seem safe? These new variants should only be creatable if a Sync Mutex type is used.

embassy-sync/src/channel.rs
```diff
@@ -317,6 +368,61 @@ where
...
+    ///
+    /// See [`Channel::poll_receive()`]
+    pub fn poll_receive(&self, cx: &mut Context<'_>) -> Poll<T> {
+        self.channel.poll_receive(cx)
+    }
+}
+
+impl<'ch, M, T, const N: usize> From<Receiver<'ch, M, T, N>> for SendableDynamicReceiver<'ch, T>
+where
+    M: RawMutex + Sync,
```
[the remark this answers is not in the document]
cschuhen: Oh, probably. Didn't realize something can be Sync and not Send.

lulf (approved):

embassy-sync/src/channel.rs
```diff
@@ -317,6 +368,61 @@ where
     }
 }
 
+/// Receive-only access to a [`Channel`] without knowing channel size.
+/// This version can be sent between threads but can only be created if the underlying mutex is Sync.
+pub struct SendableDynamicReceiver<'ch, T> {
```
lulf, line 373: Oops, this one should've been SendDynamicReceiver. Could you fix the name @cschuhen ?

cschuhen: Oops, went back over my bash history.... Yes, I stuffed up running my rename script. #4265

PR #4266, opened 2025-05-29
Allow `-> impl Future<Output = ()>` in #[task]

0e4ef622:
I tried to keep errors the same. They are a bit ugly on stable in the worst case, although they look ok in rust-analyzer.

image


embassy-executor/tests/test.rs
```diff
@@ -73,6 +73,28 @@ fn executor_task() {
     )
 }
 
+#[test]
+fn executor_task_rpit() {
+    #[task]
+    fn task1(trace: Trace) -> impl Future<Output = ()> {
```
Dirbaio, line 79: can you also add a test for ` -> impl Future<Output = !>` ?

embassy-executor/src/lib.rs
```diff
@@ -65,16 +65,25 @@ pub mod _export {
 
     use crate::raw::TaskPool;
 
+    trait TaskReturnValue {}
+    impl TaskReturnValue for () {}
+    impl TaskReturnValue for Never {}
+
+    #[diagnostic::on_unimplemented(
+        message = "task function futures must resolve to `()`",
```
Dirbaio, line 73: why not also mention `!` ?

embassy-executor/src/lib.rs
```diff
@@ -205,4 +214,38 @@ pub mod _export {
...
+    #[allow(dead_code)]
+    type Never = <fn() -> ! as HasOutput>::Output;
+}
+
+/// Implementation details for embassy macros.
+/// Do not use. Used for macros and HALs only. Not covered by semver guarantees.
+#[doc(hidden)]
+#[cfg(feature = "nightly")]
+pub mod _export {
+    pub trait TaskReturnValue {}
```
Dirbaio, line 240: maybe adding `on_unimplemented` here like for TaskFn on stable would improve the error messages?

embassy-executor/tests/ui.rs
```diff
@@ -17,6 +17,10 @@ fn ui() {
     t.compile_fail("tests/ui/nonstatic_struct_elided.rs");
     t.compile_fail("tests/ui/nonstatic_struct_generic.rs");
     t.compile_fail("tests/ui/not_async.rs");
+    // #[cfg(not(feature = "nightly"))] // output differs on stable and nightly
```
Dirbaio: why is this commented? is it because we're running the tests on nightly both with and without the `nightly` feature enabled? 

Maybe the test could check for nightly dynamically?  run `rustc --version` and check if it contains `nightly`.

embassy-executor/tests/ui.rs
```diff
@@ -17,6 +17,10 @@ fn ui() {
...
+    // t.compile_fail("tests/ui/bad_return_impl_trait.rs");
+    #[cfg(feature = "nightly")]
+    t.compile_fail("tests/ui/bad_return_impl_trait_nightly.rs");
```
Dirbaio: can you add a test for what happens if the user uses the wrong trait? `#[task] fn task() -> impl Send`

embassy-executor/tests/ui.rs
0e4ef622: Do you want 3 copies of the same test (stable, nightly, nightly feature flag) or would (stable, nightly feature flag) be enough? Another option could be to move it to a doctest so where the compiler output isn't tested.

For now I'll go with stable and nightly feature flag, so this test won't run on nightly without the feature flag since I don't see the value of adding and maintaining a 3rd copy.

Dirbaio:
thank you! :rocket:

PR #4267, opened 2025-05-29
convert embassy-stm32::qspi::enums `Into` impls into equivalant `From` impls

melvdlin:
Convert manual `Into` impls in `embassy-stm32::qspi::enums` into equivalent `From` impls


Dirbaio:
bender run

PR #4271, opened 2025-05-30
Use a critical section mutex in CAN Fdcan driver.

cschuhen:
The FDCAN driver was using a static function to hand out &mut State internally. Generally, care was taken to only access this from within a critical section or Interrupt. The interrupt case is an issue, if some code decided to [de]configure the buffering mode from a higher priority interrupt level than the CAN interrupt. This seems unlikely, anyway... 

This change was proposed by @DrTobe in: https://github.com/embassy-rs/embassy/pull/4223

This required: https://github.com/embassy-rs/embassy/pull/4265 as a dependency. Which also points us to the fact that buffered mode was unsound if the Channel was constructed with a Raw mutex. The first commit in this series blocks that.

This patch has been tested on STM32G431 boards, mainly with the buffered mode.

### DIFF
--- a/embassy-stm32/src/can/common.rs
+++ b/embassy-stm32/src/can/common.rs
@@ -1,29 +1,29 @@
-use embassy_sync::channel::{DynamicReceiver, DynamicSender};
+use embassy_sync::channel::{SendDynamicReceiver, SendDynamicSender};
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
bender run

PR #4279, opened 2025-06-03
Added Option to enable HW Oversampling in STM32 V3 ADCs

benjaminschlegel87:
Added HW oversampling support like in v4.rs for v3.rs with adjusting the oversampling and bit shifting options to the V3 register definition (2 to 256x oversampling and 0 to 8 Bit shifts)

This is my first ever PR so let me know if something is not right


benjaminschlegel87:
As I am working with a STM32L4 I did not see at first these three function implementing the same feature for g0 and u0. Now I am thinking if removing these three functions is the right way as it is covered by the approach I implemented and is also used in the V4 ADC.
Or if I should expand these three functions for the other variants?

```    
    #[cfg(any(adc_g0, adc_u0))]
    pub fn set_oversampling_shift(&mut self, shift: u8) {
        T::regs().cfgr2().modify(|reg| reg.set_ovss(shift));
    }

    #[cfg(any(adc_g0, adc_u0))]
    pub fn set_oversampling_ratio(&mut self, ratio: u8) {
        T::regs().cfgr2().modify(|reg| reg.set_ovsr(ratio));
    }

    #[cfg(any(adc_g0, adc_u0))]
    pub fn oversampling_enable(&mut self, enable: bool) {
        T::regs().cfgr2().modify(|reg| reg.set_ovse(enable));
    }
```

Dirbaio:
could you rebase then fix the build issue?

benjaminschlegel87:
build issue is fixed

Issue #3909, opened 2025-02-21
No way for user to get system clock frequency from embassy-stm32


dlaw:
After configuring clocks with embassy-stm32, peripheral clock frequencies are available through the `.frequency()` methods of the corresponding peripherals. However, there seems to be no way for user code to find out the system clock frequency.

Internal embassy-stm32 code is able to use the private `get_freqs()` method for this:
https://github.com/search?q=repo%3Aembassy-rs%2Fembassy%20unsafe%20%7B%20crate%3A%3Arcc%3A%3Aget_freqs()%20%7D.sys&type=code

Would it be possible to provide a public RCC method which returns the system clock frequency?

I would be happy to attempt the implementation if this is deemed desirable.

diondokter:
Ah yes, I've run into this a couple of times as well. Would be nice to get access to the clocks (also more than just the system clock).
I guess the reason it's not public is because it's not considered a stable API.

PR #4287, opened 2025-06-06
Add public function to get current clock configuration

dlaw:
This PR provides a means for user code to access the current clock configuration, addressing #3909.

We must ensure that the new `rcc::clocks()` function does not allow the user to hold a reference to `CLOCK_FREQS`: (1) before `embassy_stm32::init()`, (2) during `rcc::reinit()`. This is done using references to the `Peri<RCC>` singleton.

Change requests welcome.... I am still wrapping my head around the Peri struct


lulf (approved):
Thanks!

PR #4295, opened 2025-06-09
Generate pins for all opamp pin naming schemes

skoe:
The new code implements the corresponding traits for the opamp pin naming schemes of all families, which are
VP/VM/VINP/VINM/VPx/VMx/VINPx/VINMx.

The same pin must not be used for multiple channels for the same opamp. For example, if VINM0 and VINM1 of the same opamp were assigned to the same pin, the channel would not be unique, meaning that the traits would be implemented in a conflicting manner. This also applies to VP/VM/VINP/VINM, which are interpreted as channel 0.

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -81,7 +81,7 @@ futures-util = { version = "0.3.30", default-features = false }
 sdio-host = "0.9.0"
 critical-section = "1.1"
 #stm32-metapac = { version = "16" }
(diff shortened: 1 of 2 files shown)


Dirbaio:
IMO we should change stm32-data to generate one consistent naming instead. I'd suggest VINPx, VINMx.

Dirbaio:
closing in favor of https://github.com/embassy-rs/stm32-data/pull/620

PR #4298, opened 2025-06-09
mspm0: add license files

i509VCB:
This was already declared in Cargo.toml, just adding the files


lulf (commented):
We don't generally add them per module, they are in the top-level folder already.

i509VCB:
> We don't generally add them per module, they are in the top-level folder already.

Yes that is correct. embassy-rp seems to be the exception because of the rp-rs attribution. We can't symlink either because that causes cargo to explode during publish.

Issue #2322, opened 2023-12-19
embassy_sync::zerocopy_channel isn't MPSC


darkwater:
```rust
impl<'a, M: RawMutex, T> Sender<'a, M, T> {
  pub fn borrow(&mut self) -> Sender<'_, M, T>
```

The additional Sender takes the lifetime of the borrow call, making the original Sender unusable until it's dropped. Is this channel type destined to be SPSC, or is this just a mistake in the API? And if it's really MPSC, couldn't it implement Clone and perhaps Copy?

Dirbaio:
It's intended to be SPSC, yes. The way "zero-copy" is done makes it hard to do MPSC, you'd have to keep track which "slots" are reserved to which senders, vs with SPSC it's trivial, it's only the "next" empty slot that can ever be reserved.

mdrssv:
> It's intended to be SPSC, yes. The way "zero-copy" is done makes it hard to do MPSC, you'd have to keep track which "slots" are reserved to which senders, vs with SPSC it's trivial, it's only the "next" empty slot that can ever be reserved.

Wouldn't it be sufficient to turn `front` & `back` into an array of `[Option<usize>; N]` where `N` represents the max number of sender & receiver instances?

PR #4299, opened 2025-06-10
`zerocopy_channel` fixed channel and borrowed sender/receiver

mkj:
The first commit implements a borrowed sender/receiver for `Channel`. This allows sending from the `Channel` itself, and also could be an alternative to MPSC such as in #2322. Currently there isn't any async `sender()`/`receiver()` equivalent, since that'd need a bounded number of wakers and seems like complication. 
The extra bools in the struct don't add any size, the only overhead will be some bit flips on construction/drop.

The second commit allows having a `FixedChannel` with local storage. I _think_ the use of `UnsafeCell` is safe, but it would be good if someone with more `unsafe` experience could take a look over it. `new_cloned()` name is a bit arbitrary.


Dirbaio:
Hello,

PR #4300, opened 2025-06-10
feat(embassy-executor): add rtos-trace-all-tasks feature flag

kat-perez:
Add a new feature flag `rtos-trace-all-tasks` that enables tracing of all tasks when using rtos-trace, not just named tasks. By default, rtos-trace only traces tasks that have been given explicit names. This new flag allows developers to trace anonymous tasks as well, which can be useful for debugging and profiling.

The implementation adds a `should_trace_task()` helper function that checks whether a task should be traced based on the enabled features. When `rtos-trace-all-tasks` is enabled, all tasks are traced. Otherwise, only named tasks are traced (preserving the existing behavior).

This change maintains backward compatibility - the default behavior remains unchanged unless the new feature flag is explicitly enabled.

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -95,8 +95,10 @@ executor-thread = []
 executor-interrupt = []
 ## Enable tracing support (adds some overhead)
 trace = []
-## Enable support for rtos-trace framework
+## Enable support for rtos-trace framework (named tasks only by default)
 rtos-trace = ["dep:rtos-trace", "trace", "dep:embassy-time-driver"]
+## Include all tasks in tracing (not just the named ones)
+rtos-trace-all-tasks = ["rtos-trace"]
 
 #! ### Timer Item Payload Size
 #! Sets the size of the payload for timer items, allowing integrated timer implementors to store
--- a/embassy-executor/src/raw/trace.rs
+++ b/embassy-executor/src/raw/trace.rs
@@ -141,7 +141,6 @@ impl TaskTracker {
             }
         }
     }
-
     /// Performs an operation on each task in the tracker
     ///
     /// This method traverses the entire list of tasks and calls the provided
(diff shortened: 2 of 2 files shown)


kat-perez:
cc: @lulf

Dirbaio:
What was the rationale for tracing only named tasks? IMO we should just trace all tasks always.

kat-perez:
@Dirbaio I have a system that spawns a lot of tasks, generates a lot of overhead and then SystemView starts to produce Buffer Overflow events and then crashes. I've tried increasing the buffer size and increasing my J-Link interface rate, but I still would like the option to separate out a specific set of tasks only after diagnosing the entire system with all tasks traced.

Dirbaio:
I see. It's a bit strange to control which tasks get traced by giving them a name or not. There's other possible uses for task names, such as logging or showing a task list in a debugger. Maybe it should be a separate attribute?

kat-perez:
Sure, that makes sense to want to not activate based on naming. I can add this attribute to TaskHeader:
```
    #[cfg(feature = "trace")]
    pub(crate) trace_excluded: bool,
```
Usage:
With `rtos-trace` feature enabled:
```
use embassy_executor::SpawnerTraceExt;

// These are traced by default
spawner.spawn(my_task())?;
spawner.spawn_named("main-task", main_task())?;

// Exclude noisy tasks you don't care about
spawner.spawn_no_trace(background_heartbeat())?;
spawner.spawn_named_no_trace("usb-handler", usb_task())?;
```

Selective measurement with `rtos-trace-selective` feature
```
use embassy_executor::SpawnerTraceExt;

// These are NOT traced by default
spawner.spawn(background_task())?;
spawner.spawn_named("helper", helper_task())?;

// Only trace the tasks you want to measure
spawner.spawn_with_trace(critical_task())?;
spawner.spawn_named_with_trace("perf-critical", performance_task())?;
```

Please let me know if you have any further suggestions for improvement. Thanks!

kat-perez:
@Dirbaio The above design is in a new PR here: #4301

PR #4301, opened 2025-06-10
add selective tracing support to SpawnerTraceExt

kat-perez:
Add spawn_with_trace(), spawn_no_trace() and named variants to provide fine-grained control over task tracing. Supports both default tracing (rtos-trace) and selective tracing (rtos-trace-selective) modes while maintaining backward compatibility.

### DIFF
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -95,8 +95,10 @@ executor-thread = []
 executor-interrupt = []
 ## Enable tracing support (adds some overhead)
 trace = []
-## Enable support for rtos-trace framework
+## Enable support for rtos-trace framework (traces all tasks by default)
 rtos-trace = ["dep:rtos-trace", "trace", "dep:embassy-time-driver"]
+## Only trace tasks that are explicitly marked for tracing (selective mode)
+rtos-trace-selective = ["rtos-trace"]
 
 #! ### Timer Item Payload Size
 #! Sets the size of the payload for timer items, allowing integrated timer implementors to store
--- a/embassy-executor/Cargo.toml
+++ b/embassy-executor/Cargo.toml
@@ -95,10 +95,12 @@ executor-thread = []
 executor-interrupt = []
 ## Enable tracing support (adds some overhead)
 trace = []
+## Enable selective tracing support
+trace-mode-selective = ["trace"]
 ## Enable support for rtos-trace framework (traces all tasks by default)
 rtos-trace = ["dep:rtos-trace", "trace", "dep:embassy-time-driver"]
 ## Only trace tasks that are explicitly marked for tracing (selective mode)
-rtos-trace-selective = ["rtos-trace"]
+rtos-trace-selective = ["rtos-trace", "trace-mode-selective"]
 
 #! ### Timer Item Payload Size
 #! Sets the size of the payload for timer items, allowing integrated timer implementors to store
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -94,6 +94,8 @@ pub(crate) struct TaskHeader {
     #[cfg(feature = "trace")]
     pub(crate) id: u32,
     #[cfg(feature = "trace")]
+    pub(crate) trace_excluded: bool,
+    #[cfg(feature = "trace")]
     all_tasks_next: AtomicPtr<TaskHeader>,
 }
 
@@ -194,6 +196,8 @@ impl<F: Future + 'static> TaskStorage<F> {
                 name: None,
                 #[cfg(feature = "trace")]
                 id: 0,
+                #[cfg(feature = "trace")]
+                trace_excluded: cfg!(feature = "rtos-trace-selective"),
                 #[cfg(feature = "trace")]
                 all_tasks_next: AtomicPtr::new(core::ptr::null_mut()),
             },
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -93,7 +93,7 @@ pub(crate) struct TaskHeader {
     pub(crate) name: Option<&'static str>,
     #[cfg(feature = "trace")]
     pub(crate) id: u32,
-    #[cfg(feature = "trace")]
+    #[cfg(feature = "trace-mode-selective")]
     pub(crate) trace_excluded: bool,
     #[cfg(feature = "trace")]
     all_tasks_next: AtomicPtr<TaskHeader>,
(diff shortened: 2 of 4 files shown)


kat-perez:
This PR was added as a design alternative to #4300

kat-perez:
@Dirbaio @lulf

embassy-executor/src/raw/mod.rs
```diff
@@ -94,6 +94,8 @@ pub(crate) struct TaskHeader {
     #[cfg(feature = "trace")]
     pub(crate) id: u32,
     #[cfg(feature = "trace")]
+    pub(crate) trace_excluded: bool,
```
lulf, line 97: This should probably be guarded by a trace-mode-selective flag to save some space for programs not using it.

kat-perez: thanks, added the trace-mode-selective flag in the latest update

kat-perez:
@lulf the ci/build job seemed to fail to error 137. Could I get a re-run for this job?

embassy-executor/src/raw/mod.rs
kat-perez: @lulf (bump)

lulf:
bender run

embassy-executor/Cargo.toml
```diff
@@ -95,8 +95,12 @@ executor-thread = []
 executor-interrupt = []
 ## Enable tracing support (adds some overhead)
 trace = []
-## Enable support for rtos-trace framework
+## Enable selective tracing support
+trace-mode-selective = ["trace"]
```
ghost, line 99: I propose that this feature flag be orthogonal to the trace and rtos-trace features, i.e. you do --features "trace trace-selective" and --features "rtos-trace trace-selective". Generating the cross product of features doesn't scale well.

You could implement a static assertion that warns if "trace-selective" is used without selecting a trace backend, too.

This change will also simplify some of your cfg logic below.

embassy-executor/src/raw/mod.rs
ghost: Nit: I propose we call this trace_enabled and reverse the logic. Positive logic is more ergonomic.

embassy-executor/src/raw/trace.rs
```diff
@@ -181,6 +181,12 @@ pub trait TaskRefTrace {
 
     /// Set the ID for a task
     fn set_id(&self, id: u32);
+
+    /// Check if a task is excluded from tracing
+    fn is_trace_excluded(&self) -> bool;
```
ghost, line 186: Either remove the is_ prefix in the method or add it to the variable, too ([C-Getter](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter))

embassy-executor/src/raw/trace.rs
```diff
@@ -205,6 +211,49 @@ impl TaskRefTrace for TaskRef {
...
+    }
+}
+
+/// Helper function to determine if a task should be traced
+///
+/// This function supports two tracing modes based on feature flags:
+/// - `rtos-trace-selective`: Only trace tasks that are explicitly enabled for tracing
+/// - Default `rtos-trace`: Trace all tasks by default
+#[inline]
+fn should_trace_task(task: &TaskRef) -> bool {
```
ghost, line 241: I don't like the distinction between rtos-trace and embassy trace on the API level. IMO this is a leaky abstraction.

If the distinction should be held up then at least this rtos-trace specific function should be named as such: should_rtos_trace_task().

embassy-executor/src/spawner.rs
```diff
@@ -196,6 +201,50 @@ pub trait SpawnerTraceExt {
...
+
+    /// Spawns a new task with a name and explicitly enables it for tracing.
+    ///
+    /// # Arguments
+    /// * `name` - Static string name to associate with the task
+    /// * `token` - Token representing the task to spawn
+    ///
+    /// # Returns
+    /// Result indicating whether the spawn was successful
+    fn spawn_named_with_trace<S>(&self, name: &'static str, token: SpawnToken<S>) -> Result<(), SpawnError>;
```
ghost, line 225: Same argument as for the feature flags: Cross product of features doesn't scale. See @Dirbaio 's argument on Matrix, too. I think we need a different approach here, e.g. task metadata as proposed by Dario.

embassy-executor/src/spawner.rs
```diff
@@ -212,6 +261,87 @@ impl SpawnerTraceExt for Spawner {
                 let task_id = task.as_ptr() as u32;
                 task.set_id(task_id);
 
+                // In selective mode, tasks are excluded by default (must use spawn_named_with_trace)
+                // In default mode, all tasks are traced
+                #[cfg(feature = "rtos-trace-selective")]
+                task.set_trace_excluded(true);
+                #[cfg(not(feature = "rtos-trace-selective"))]
+                task.set_trace_excluded(false);
```
ghost, line 269: I propose that methods related to this feature should not be present at all when not needed (i.e. hidden behind the same feature flag, too).

embassy-executor/src/raw/trace.rs
```diff
@@ -384,14 +439,17 @@ where
...
-            let name = task.name().unwrap_or("unnamed task\0");
-            let info = rtos_trace::TaskInfo {
-                name,
-                priority: 0,
-                stack_base: 0,
-                stack_size: 0,
-            };
-            rtos_trace::trace::task_send_info(task.id(), info);
+            // Only send task info for tasks that should be traced
+            if should_trace_task(&task) {
```
ghost, line 443: This method should not be generated when the feature is not present.

fg-cfh (commented):
I'm new to this community, so I won't block the PR but according to my personal standards it is not ready to be merged:
- The newly introduced feature should work for all kinds of tracing (i.e. embassy _and_ rtos_trace) alike.
- The feature flags are confusing and unnecessarily complex/verbose. Can and should be simplified.
- Runtime and config-time logic is mixed up. Runtime logic should not be constant. If something becomes contant then it should be made constant by config, not by having runtime methods return constant values IMHO.

embassy-executor/src/raw/trace.rs
ghost: See a similar additive solution in #4383.

ghost:
Blocked by #4383

Dirbaio:
@kat-perez now that #4383 is merged, could you change this to be a "trace" bool field in Metadata?

Dirbaio:
Hello,

PR #4306, opened 2025-06-13
stm32/adc/v3: allow DMA reads to loop through enabled channels

yodaldevoid:
Tested on an STM32H533RE. Documentation of other chips has been reviewed to ensure compatibility, but not extensively.

I can make a similar change for other ADC versions, but I don't have the hardware to test it.


yodaldevoid:
@Dirbaio I hate to be a bother, and I understand being busy, but it has been a month. Is there anything that I should change or do for this to be merged?

yodaldevoid:
I've added an example modeled after the ring-buffered ADC examples.

yodaldevoid:
@lulf @Dirbaio Am I missing anything from this PR before it can be merged?

lulf (approved):

lulf (approved):
Sorry for the late reply!

Issue #4312, opened 2025-06-17
U5: MSI auto calibration not enabled when LSE is present


snakehand:
MSI auto calibration is not enabled. It is possible to manually enable it, but then the derived clock frequencies are not precise.

The PLL feature should be automatically enabled for MSIK when LSE is within range, and correct MSIS / MSIK frequencies should be computed.

snakehand:
Closed with #4313

PR #4313, opened 2025-06-17
U5: Enable MSI auto calibration and compute frequencies

snakehand:
This adresses #4312

### DIFF
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -5,7 +5,7 @@ pub use crate::pac::rcc::vals::{
     Hpre as AHBPrescaler, Msirange, Msirange as MSIRange, Plldiv as PllDiv, Pllm as PllPreDiv, Plln as PllMul,
     Pllsrc as PllSource, Ppre as APBPrescaler, Sw as Sysclk,
 };
-use crate::pac::rcc::vals::{Hseext, Msirgsel, Pllmboost, Pllrge};
+use crate::pac::rcc::vals::{Hseext, Msipllsel, Msirgsel, Pllmboost, Pllrge};
 #[cfg(all(peri_usb_otg_hs))]
 pub use crate::pac::{syscfg::vals::Usbrefcksel, SYSCFG};
 use crate::pac::{FLASH, PWR, RCC};
@@ -131,7 +131,18 @@ pub(crate) unsafe fn init(config: Config) {
     PWR.vosr().modify(|w| w.set_vos(config.voltage_range));
     while !PWR.vosr().read().vosrdy() {}
 
-    let msis = config.msis.map(|range| {
+    let lse_calibration_freq = match config.ls.lse {
+        Some(lse_config) => {
+            if lse_config.peripherals_clocked && (31_000..=34_000).contains(&lse_config.frequency.0) {
+                Some(lse_config.frequency)
+            } else {
+                None
+            }
+        }
+        _ => None,
+    };
+
+    let mut msis = config.msis.map(|range| {
         // Check MSI output per RM0456 § 11.4.10
         match config.voltage_range {
             VoltageScale::RANGE4 => {
@@ -184,8 +195,25 @@ pub(crate) unsafe fn init(config: Config) {
         RCC.cr().modify(|w| {
             w.set_msikon(true);
         });
+        if lse_calibration_freq.is_some() {
+            // Enable the MSIK auto-calibration feature
+            RCC.cr().modify(|w| w.set_msipllsel(Msipllsel::MSIK));
+            RCC.cr().modify(|w| w.set_msipllen(true));
+        }
         while !RCC.cr().read().msikrdy() {}
-        msirange_to_hertz(range)
+        if let Some(freq) = lse_calibration_freq {
+            // Estimate frequency based on it being a fixed fractional multiple of LSE
+            let base = msirange_to_hertz(range).0;
+            let multiplier = (base + 8096) / 16384;
+            let msik_freq = Hertz(freq.0 * multiplier / 2);
+            if config.msis == config.msik {
+                // If MSIS and MSIK are the same range both will be auto calibrated to the same frequency
+                msis = Some(msik_freq)
+            }
+            msik_freq
+        } else {
+            msirange_to_hertz(range)
+        }
     });
 
     let hsi = config.hsi.then(|| {
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -173,31 +173,39 @@ pub(crate) unsafe fn init(config: Config) {
     PWR.vosr().modify(|w| w.set_vos(config.voltage_range));
     while !PWR.vosr().read().vosrdy() {}
 
-    let lse_calibration_freq = match config.ls.lse {
-        Some(lse_config) => {
-            // Allow +/- 5% tolerance for LSE frequency
-            if lse_config.peripherals_clocked && (31_100..=34_400).contains(&lse_config.frequency.0) {
-                // Check that the calibration is applied to an active clock
-                match (
-                    config.auto_calibration.base_mode(),
-                    config.msis.is_some(),
-                    config.msik.is_some(),
-                ) {
-                    (MsiAutoCalibration::MSIS, true, _) => {
-                        // MSIS is active and using LSE for auto-calibration
-                        Some(lse_config.frequency)
-                    }
-                    (MsiAutoCalibration::MSIK, _, true) => {
-                        // MSIK is active and using LSE for auto-calibration
-                        Some(lse_config.frequency)
-                    }
-                    // improper configuration
-                    _ => panic!("MSIx auto-calibration is enabled for a source that has not been configured.")                }
-            } else {
-                None
+    let lse_calibration_freq = if config.auto_calibration != MsiAutoCalibration::Disabled {
+        // LSE must be configured and peripherals clocked for MSI auto-calibration
+        let lse_config = config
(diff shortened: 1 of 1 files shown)


Dirbaio:
Thanks for the PR!

From looking at the RM, it seems to me MSI PLL is only usable when LSE is exactly 32768hz, and therefore the MSI freq will always be exactly what's specified by MSIRANGE, so we can always use `msirange_to_hertz` and `calculate_calibrated_msi_frequency` is not actually needed. Am I missing something?

screenshot-2025-06-22_23-30-42

snakehand:
We have tested this in the lab with a 32000 Hz source, over wide temperature ranges, and have found that the "PLL" works over a wider range of temperatures. It is not described as a PLL in the documentation, but as "auto-calibration feature" , and I believe it is implemented by a couple of counters that are in a control loop with the calibration function. The calibration function has a wide offset range (+/- several % ) , and it is reasonable to assume that a similar range of input frequencies must work for the reference.

Dirbaio:
"when a 32.768khz external oscillator is present" strongly implies other frequencies are not supported. 

Also, the section for the LSE also says it must be a 32768hz crystal. Other frequencies are only allowed in bypass mode. And "when a 32.768khz external oscillator is present" also implies it must be an oscillator, not bypass mode.

STM32CubeMX only allows 32768hz if LSE is a crystal. It doesn't seem to complain if you enable MSI PLL and set LSE to another frequency, but it doesn't change the MSI frequency based on it like this PR does.

Do you have a source from ST stating this is supported? I'm a bit hesitant of adding support for running out of spec if it increases complexity.

embassy-stm32/src/rcc/u5.rs
```diff
@@ -131,7 +131,18 @@ pub(crate) unsafe fn init(config: Config) {
     PWR.vosr().modify(|w| w.set_vos(config.voltage_range));
     while !PWR.vosr().read().vosrdy() {}
 
-    let msis = config.msis.map(|range| {
+    let lse_calibration_freq = match config.ls.lse {
```
Dirbaio: instead of automatically deciding to enable msipll, could you add it to the config so the user explicitly chooses to. It could be `None`, `Some(Msik)` or `Some(Msis)`. 

If ther'es no LSE or the freq is too off just panic, this way the user notices if they do something wrong instead of it silently doing no calibration.

embassy-stm32/src/rcc/u5.rs
```diff
@@ -514,3 +539,51 @@ fn init_pll(instance: PllInstance, config: Option<Pll>, input: &PllInput, voltag
...
+    fn calculate_frequency(&self, lse_freq: Hertz) -> Hertz {
+        Hertz(lse_freq.0 * self.numerator / self.denominator)
+    }
+}
+
+/// Get the calibration fraction for a given MSI range
+/// Based on STM32U5 datasheet table for LSE = 32.768 kHz
+fn get_msi_calibration_fraction(range: Msirange) -> MsiFraction {
+    match range {
+        Msirange::RANGE_48MHZ => MsiFraction::new(1465, 1), // Range 0: 48.005 MHz
```
Dirbaio: for each "group" of 4 the numerator is the same and the denominator is 1, 2, 3, 4. Maybe this can help compactify the table?

```
Msirange::RANGE_48MHZ     => MsiFraction::new(1465, 1), // Range 0: 48.005 MHz
Msirange::RANGE_24MHZ     => MsiFraction::new(1465, 2), // Range 1: 24.003 MHz
Msirange::RANGE_16MHZ     => MsiFraction::new(1465, 3), // Range 2: 16.002 MHz
Msirange::RANGE_12MHZ     => MsiFraction::new(1465, 4), // Range 3: 12.001 MHz
Msirange::RANGE_4MHZ      => MsiFraction::new(122, 1),  // Range 4: 3.998 MHz
Msirange::RANGE_2MHZ      => MsiFraction::new(122, 2),  // Range 5: 1.999 MHz
Msirange::RANGE_1_33MHZ   => MsiFraction::new(122, 3),  // Range 6: 1.333 MHz
Msirange::RANGE_1MHZ      => MsiFraction::new(122, 4),  // Range 7: 0.999 MHz
Msirange::RANGE_3_072MHZ  => MsiFraction::new(94, 1),   // Range 8: 3.08 MHz
Msirange::RANGE_1_536MHZ  => MsiFraction::new(94, 1),   // Range 9: 1.54 MHz
Msirange::RANGE_1_024MHZ  => MsiFraction::new(94, 3),   // Range 10: 1.027 MHz
Msirange::RANGE_768KHZ    => MsiFraction::new(94, 4),   // Range 11: 0.77 MHz
Msirange::RANGE_400KHZ    => MsiFraction::new(12, 1),   // Range 12: 393 kHz
Msirange::RANGE_200KHZ    => MsiFraction::new(12, 2),   // Range 13: 196.6 kHz
Msirange::RANGE_133KHZ    => MsiFraction::new(12, 3),   // Range 14: 131 kHz
Msirange::RANGE_100KHZ    => MsiFraction::new(12, 4),   // Range 15: 98.3 kHz
```

embassy-stm32/src/rcc/u5.rs
```diff
@@ -184,8 +195,22 @@ pub(crate) unsafe fn init(config: Config) {
...
+            // Enable the MSIK auto-calibration feature
+            RCC.cr().modify(|w| w.set_msipllsel(Msipllsel::MSIK));
+            RCC.cr().modify(|w| w.set_msipllen(true));
+        }
         while !RCC.cr().read().msikrdy() {}
-        msirange_to_hertz(range)
+        if let Some(freq) = lse_calibration_freq {
+            let msik_freq = calculate_calibrated_msi_frequency(range, freq);
+            if config.msis == config.msik {
+                // If MSIS and MSIK are the same range both will be auto calibrated to the same frequency
```
Dirbaio: they don't need to be the exact same, it's enough if they're in the same RC (group of 4)

embassy-stm32/src/rcc/u5.rs
snakehand: Config has been included in latest commit.

embassy-stm32/src/rcc/u5.rs
snakehand: Large match statement has been compacted to some calculations and a small table lookup.

embassy-stm32/src/rcc/u5.rs
snakehand: Logic has been updated to detect if the same clock source has been used, in which case dependent frequency is the recomputed taking care of cases where the output divider is different.

snakehand:
The MSIPLLFAST can not be configred in the current code. I am thinking about how this should be included in the configuration, and it would be good to have that figured out now to avoid future breaking changes. @Dirbaio

embassy-stm32/src/rcc/u5.rs
```diff
@@ -195,24 +228,38 @@ pub(crate) unsafe fn init(config: Config) {
...
             msirange_to_hertz(range)
-        }
+        };
+        while !RCC.cr().read().msikrdy() {}
+        msik
     });
 
+    // If both MSIS and MSIK are enabled, we need to check if they are using the same internal source.
+    if let Some(lse_freq) = lse_calibration_freq {
+        if let (Some(msis_range), Some(msik_range)) = (config.msis, config.msik) {
```
Dirbaio, line 290: <del>this makes autocalibration only be applied when both msis and msik are enabled, which I think is too restrictive?</del>

<del>for example if config has `msis=Some(..), msik=None, autocalbration=MSIS`, autocalibration will not be applied, which would be a bit surprising.</del>

EDIT my bad, it's handled above, this is just for the case when both can be calibrated at the same time.

We should also probably panic if the user has set autocalbration=MSIx and the corresponding MSIx is None, that's an invalid config and we shouldn't silently ignore it.

snakehand: Auto-calibration will be applied when the configuration matches and active clock. This code block detects when the clock source is shared and the other oscillator needs to have the frequency recomputed. Line 195 detects an improper configuration.

embassy-stm32/src/rcc/u5.rs
```diff
@@ -133,7 +157,8 @@ pub(crate) unsafe fn init(config: Config) {
 
     let lse_calibration_freq = match config.ls.lse {
         Some(lse_config) => {
-            if lse_config.peripherals_clocked && (31_000..=34_000).contains(&lse_config.frequency.0) {
+            // Allow +/- 5% tolerance for LSE frequency
+            if lse_config.peripherals_clocked && (31_100..=34_400).contains(&lse_config.frequency.0) {
```
Dirbaio: cases that are invalid config should panic instead of silently not do calibration:

- Requested calibrating MSIx but that MSIx is off
- Lse not running
- Lse freq out of range
- Others..?

embassy-stm32/src/rcc/u5.rs
Dirbaio: yeah my bad, I misunderstood the code. Loooks good.

 There's still cases where an invalid config is sliently ignored, see my other comment.

snakehand: I can add a panic!() - I was just hesitant to do so...

embassy-stm32/src/rcc/u5.rs
snakehand: I added a basic panic, will make the configuration checks more comprehensive in a later commit.

snakehand: Also added an assert that LSE peripherals were clocked

Dirbaio (approved):
🚀 

thanks!

PR #4316, opened 2025-06-18
fix: stm32 i2c slave blocking r/w

Knaifhogg:
This fixes an issue where the slave interface would time out when the master goes from a short write to a read (e.g. when accessing memory registers) with a START signal between. The previous implementation would expect the full buffer length to be written before starting to listen to new commands.

This also adds debug trace printing which helped during implemention and testing.

Places error checking into a function inspired from a C implementation of HAL.

Tested on a STM32G070.

Note: I noticed that sometimes the `wait_rxne` still times out (like before this PR) but it was rather rare and the result in that case is that it works as before, which requires the application to handle the Err(Timeout).

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -129,6 +129,7 @@ defmt = [
     "embassy-net-driver/defmt",
     "embassy-time?/defmt",
     "embassy-usb-synopsys-otg/defmt",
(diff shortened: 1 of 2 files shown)


Knaifhogg:
Found some differences ~~in the behaviour as Master~~, will investigate.
EDIT: Solved, was an issue with looping and checking busy flag.

PR #4320, opened 2025-06-18
Fix sync bounds on LazyLock and OnceLock

### LINKED ISSUE #3904: embassy-net: Thread safety issue?
Hi,

I'm occasionally getting this error when trying to send UDP packets from an interrupt (by directly calling `poll_send_to` of `UdpSocket`):

```
18.739376 ERROR panic-probe-0.3.2/src/lib.rs:104  panicked at embassy/embassy-net/src/lib.rs:388:27:
already borrowed: BorrowMutError
```

The fact that you can create such crash without any unsafe code, IMHO, means that there's a potential unsoundness:

The `Stack` of embassy-net struct directly use `RefCell` rather than `Mutex` to achieve shared access. As such, you can only operate all network-related things within one executor.

Is this correct?

### LINKED ISSUE #4307: LazyLock implements Sync on all T
Embassy's LazyLock implements `Sync` for all `T` even if `T` itself is not `Sync`: 
https://github.com/embassy-rs/embassy/blob/56572ef0adffd6258adc10fb424e37a8b4ddc19c/embassy-sync/src/lazy_lock.rs#L34

But the `std` version of `LazyLock` has additional bounds: 
https://doc.rust-lang.org/std/sync/struct.LazyLock.html#impl-Sync-for-LazyLock%3CT,+F%3E

Is it sound for embassy to do that or is it a bug?

wmmc88:
This pull request introduces changes to ensure stricter `Sync` trait implementations for `LazyLock` and `OnceLock` types in the `embassy-sync` crate. It also adds regression tests to verify these changes and prevent incorrect usage of these types with non-`Sync` data. Additionally, a new dependency (`trybuild`) is added for compile-time testing.

### Stricter `Sync` Implementations:

* [`embassy-sync/src/lazy_lock.rs`](diffhunk://#diff-cd31bb3d341607967bd756da8f3d44ab3106dc08d38c7b4ec235230b584c2723L34-R39): Updated the `Sync` implementation for `LazyLock` to require both `T` and `F` to implement `Sync`.
* [`embassy-sync/src/once_lock.rs`](diffhunk://#diff-e8478b2fa15a8e88ebb2d2ea0faecdfc7223d2c3b8c5ccc0ce631aa432d74939L45-R45): Updated the `Sync` implementation for `OnceLock` to require `T` to implement `Sync`.

### Regression Tests for Compile-Time Errors:

* [`embassy-sync/tests/ui.rs`](diffhunk://#diff-b93a33ee71aaf155cb7b6f03478c2859084d60d09bb29b83e108f1b086a32e1fR1-R13): Added compile-fail tests using `trybuild` to ensure `LazyLock` and `OnceLock` do not unconditionally implement `Sync` for non-`Sync` types.
* [`embassy-sync/tests/ui/sync_impl/lazy_lock_function.rs`](diffhunk://#diff-8718c0de6b942640ffb134d6add1b5e573e2841c472cd6be7e5deece2e29a1ecR1-R15): Added a test case where a `LazyLock` is initialized with a closure capturing a non-`Sync` variable, which should fail to compile.
* [`embassy-sync/tests/ui/sync_impl/lazy_lock_type.rs`](diffhunk://#diff-294d2eb1b6946e68e6c4fd888f878c778449d58b7eddaf65770d4838422f484eR1-R6): Added a test case where `LazyLock` is used with a non-`Sync` type (`*mut u8`), which should fail to compile.
* [`embassy-sync/tests/ui/sync_impl/once_lock.rs`](diffhunk://#diff-fd48543eaaf2518651235c829ec4a185716da86bedf3e6f6a3ece260cb9164b9R1-R6): Added a test case where `OnceLock` is used with a non-`Sync` type (`*mut u8`), which should fail to compile.

### Dependency Addition:

* [`embassy-sync/Cargo.toml`](diffhunk://#diff-8ef39bee7253569f3e25944bb7297e3b70d36e6e5308c17c8b0bce9c2cc903f4R46): Added `trybuild` dependency for compile-time testing.

fixes #4307
fixes #3904

### DIFF
--- a/embassy-sync/Cargo.toml
+++ b/embassy-sync/Cargo.toml
@@ -43,3 +43,4 @@ futures-util = { version = "0.3.17", features = [ "channel", "sink" ] }
 # Enable critical-section implementation for std, for tests
 critical-section = { version = "1.1", features = ["std"] }
 static_cell = { version = "2" }
+trybuild = "1.0.105"
--- a/embassy-sync/src/lazy_lock.rs
+++ b/embassy-sync/src/lazy_lock.rs
@@ -31,7 +31,12 @@ union Data<T, F> {
     f: ManuallyDrop<F>,
 }
 
-unsafe impl<T, F> Sync for LazyLock<T, F> {}
+unsafe impl<T, F> Sync for LazyLock<T, F>
+where
+    T: Sync,
+    F: Sync,
+{
+}
 
 impl<T, F: FnOnce() -> T> LazyLock<T, F> {
     /// Create a new uninitialized `StaticLock`.
--- a/embassy-sync/src/once_lock.rs
+++ b/embassy-sync/src/once_lock.rs
@@ -42,7 +42,7 @@ pub struct OnceLock<T> {
     data: Cell<MaybeUninit<T>>,
 }
 
-unsafe impl<T> Sync for OnceLock<T> {}
+unsafe impl<T> Sync for OnceLock<T> where T: Sync {}
 
 impl<T> OnceLock<T> {
     /// Create a new uninitialized `OnceLock`.
--- a/embassy-sync/tests/ui.rs
+++ b/embassy-sync/tests/ui.rs
@@ -1,0 +1,13 @@
+// #[cfg(not(miri))]
+#[test]
+fn ui() {
+    let t = trybuild::TestCases::new();
+
+    // These test cases should fail to compile since OnceLock and LazyLock should not unconditionally implement sync
+    // for all types. These tests are regression tests against the following issues:
+    // * https://github.com/embassy-rs/embassy/issues/4307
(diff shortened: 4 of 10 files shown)


lulf (approved):
TIL trybuild. Thanks for the fixes! 🚀

lulf (commented):
A few questions:

* Does the stderr files need to be committed?
* If the compiler output changes some wording, does the test fail?

wmmc88:
> A few questions:
> 
> * Does the stderr files need to be committed?
> * If the compiler output changes some wording, does the test fail?

The files need to be committed because the error output is compared with what's committed
 
If the compiler output changes,the test also fails. (Ex. If it starts falling for a different reason)

Embassy executor also has trybuild tests

PR #4323, opened 2025-06-20
embassy-stm32: Rename PWM Input constructors, add warning on usable timer peripherals

phycrax:
Some timers don't support PWM Input mode. 
See below for STM32G4 RM0440:
{430C5AA0-6125-4A27-A9D7-6312D2219B93}

And PWM Input is usable only with CH1 and CH2 on all families.
{D1555EE1-1492-4F49-BBEC-B349992188F9}


Dirbaio:
looks good. can you update examples so that CI passes?

phycrax:
> looks good. can you update examples so that CI passes?

Done. Also merged timer generic channel changes.

PR #4337, opened 2025-06-24
STM32H7RS Examples: Add OPI functionality to the XSPI example

kevswims:
- Cleans up the SPI and OPI commands to match the datasheet for the flash used on the Nucleo board
- Creates a separate impl for OPI operations

This has been tested on an STM32H7S3 nucleo board.

Unfortunately, the datasheet for the flash chip used on the nucleo board is not public however I was able to get access to it in order to get this example fixed up to match the datasheet.


mkj:
Thanks, looks good and I've confirmed it runs here.

PR #4338, opened 2025-06-24
mspm0: add dma driver

i509VCB:
Add a DMA driver for mspm0. So far only implement features available on all (basic) channels. copying slices is also not implemented yet nor anything related to source/destination stride. 

~~A few things pending before this is ready to review~~
- [x] merge https://github.com/mspm0-rs/mspm0-data/pull/6
- [x] add tests for u16, u32 and u64 transfers
- [x] ~~Figure out why `TransferMode::Single` with `dst.len() > 2` performs 2 transfers instead of 1 (as expected in datasheet)~~ (likely hardware bug)
- [x] ~~Test multiple triggers (`resume()`)~~
- [x] Remove the settings.json changes (after I am done)

### DIFF
--- a/embassy-mspm0/src/dma.rs
+++ b/embassy-mspm0/src/dma.rs
@@ -1,0 +1,626 @@
+//! Direct Memory Access (DMA)
+
+#![macro_use]
+
+use core::future::Future;
+use core::mem;
+use core::pin::Pin;
+use core::sync::atomic::{compiler_fence, Ordering};
+use core::task::{Context, Poll};
+
+use critical_section::CriticalSection;
+use embassy_hal_internal::interrupt::InterruptExt;
+use embassy_hal_internal::{impl_peripheral, PeripheralType};
+use embassy_sync::waitqueue::AtomicWaker;
+use mspm0_metapac::common::{Reg, RW};
+use mspm0_metapac::dma::regs;
+use mspm0_metapac::dma::vals::{self, Autoen, Em, Incr, Preirq, Wdth};
+
+use crate::{interrupt, pac, Peri};
+
+/// The burst size of a DMA transfer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BurstSize {
+    /// The whole block transfer is completed in one transfer without interruption.
+    Complete,
+
+    /// The burst size is 8, after 9 transfers the block transfer is interrupted and the priority
+    /// is reevaluated.
+    _8,
+
+    /// The burst size is 16, after 17 transfers the block transfer is interrupted and the priority
+    /// is reevaluated.
+    _16,
+
+    /// The burst size is 32, after 32 transfers the block transfer is interrupted and the priority
+    /// is reevaluated.
+    _32,
+}
+
+/// DMA channel.
+#[allow(private_bounds)]
+pub trait Channel: Into<AnyChannel> + PeripheralType {}
+
+/// Full DMA channel.
+#[allow(private_bounds)]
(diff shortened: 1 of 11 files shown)


optlink:
I do have some experience with this DMA controller in a real application and I think your issue might be the software trigger. With an external trigger from the ADCs the single transfer mode works as expected. I have not verified this myself but I suspect that the software trigger just holds the trigger active and makes the transfer mode redundant given that I see no actual way to control the software trigger.

i509VCB:
Yeah it does feel like software triggering the dma doesn't behave right. If I set the destination to 4 bytes in single transfer mode 2 bytes are written as well. That doesn't match the idea of the software trigger letting the DMA run to completion.

This feels like a hardware bug or the datasheet is too ambigious to answer?

optlink:
Hmm that is strange. Based on the software trigger behavior of the ADC I expected it to be similar. It does seem like a pathological case though. When would you use such a configuration in a real application? Perhaps it’s considered undefined behavior and just not documented very well like most of this series.

i509VCB:
> When would you use such a configuration in a real application?

Probably not practical to do single software triggered single transfers for an array of reads (basically using the CPU for memcpy with extra steps). Exhaustively testing the driver options is why that test case came to exist.

optlink:
Fair enough and that’s what I guessed was happening. The rest of the driver matches what I know of this module so I think it’s sound except for this one test case.

embassy-mspm0/src/dma.rs
```diff
@@ -0,0 +1,607 @@
...
+
+impl Default for TransferOptions {
+    fn default() -> Self {
+        Self {
+            mode: TransferMode::Single,
+        }
+    }
+}
+
+// TODO: u128 (LONGLONG) support is available but not on every part number (do any currently released
```
Iooon: Doesn't it say in the G3507 reference that it supports u128? Maybe I am wrong

embassy-mspm0/src/dma.rs
```diff
@@ -0,0 +1,607 @@
...
+
+#[cfg(feature = "rt")]
+#[interrupt]
+fn DMA() {
+    use crate::BitIter;
+
+    let events = pac::DMA.int_event(0);
+    let mis = events.mis().read();
+
+    // TODO: Handle DATAERR and ADDRERR? However we do not know which channel causes an error.
```
Iooon, line 608: Am I correct that we need to handle this on the channel level, because of the way the chip raises the error?
So each channel would need to check the IIDX[j].STAT register individually?

i509VCB: So the array access for `int_event` is selecting the "CPU event" interrupts. Index 1 for `int_event` is the GEN_EVENT group.

Looking at the STAT values (and it matches with INT):
image

We see that the DMA address and data errors do not specify what channel caused the error. We can be told of course when a channel finishes, but if an error happens we don't know which channel failed.

embassy-mspm0/src/dma.rs
i509VCB: Okay so there is a target. I probably need to find some metadata info so that I can feature gate u128 transfers to parts that support it.

embassy-mspm0/src/dma.rs
i509VCB: Well to match what embassy-stm32 does, panic on data or address error.

i509VCB:
I think we're good at this point to merge. I'll address the u128 transfers when the metadata for determining if that is possible is dealt with.

PR #4341, opened 2025-06-25
Adapt to new opamp pin naming scheme and opamp IP version numbers

skoe:
The new code implements the corresponding traits for the common opamp pin naming scheme of all families, which is VINPx/VINMx and adapts the opamp driver to new opamp IP version numbers.

The same pin must not be used for multiple channels for the same opamp. For example, if VINM0 and VINM1 of the same opamp were assigned to the same pin, the channel would not be unique, meaning that the traits would be implemented in a conflicting manner.


skoe:
Ups, forgot to commit a file. Will update it tonight.

Dirbaio:
:ok_hand: :pray: :heart:

PR #4342, opened 2025-06-25
Allow simultaneous use of ADC1 and ADC2 on STM32F3

dingari:
ADC1 and ADC2 share an interrupt handler on the STM32F3x. This new type allows us to create instance of each one using the shared interrupt. Tested and working on an STM32F303RCTx.


Dirbaio:
You can bind one interrupt to two handlers, so the `Irqs` token is valid for both ADCs.

```rust
    ADC1_2 => adc::InterruptHandler<ADC1>, adc::InterruptHandler<ADC2>;
});
```

With this I think the changes in this PR aren't needed?

dingari:
Interesting! I spent a  while trying to figure this out awhile ago, but I think  you're right!

I'll close this PR for now.

PR #4343, opened 2025-06-26
Added STM32WBA6XXX devices as features

leftger:
Added the following
- new devices (a.k.a. features) from `stm32-data`
- within `ci.sh`: attempt to build 2 `stm32wba6[2|5]` devices
- handled a default rtc implementation if the generated peripherals didn't match the mapping.
- handled a new generated peripheral `rtc_v3c0` recently introduced to `stm32-data-generated`

### DIFF
--- a/embassy-stm32/src/rtc/mod.rs
+++ b/embassy-stm32/src/rtc/mod.rs
@@ -25,7 +25,7 @@ use crate::time::Hertz;
     ),
     path = "v2.rs"
 )]
-#[cfg_attr(any(rtc_v3, rtc_v3u5, rtc_v3l5, rtc_v3h7rs), path = "v3.rs")]
+#[cfg_attr(any(rtc_v3, rtc_v3u5, rtc_v3l5, rtc_v3h7rs, rtc_v3c0), path = "v3.rs")]
 mod _version;
 #[allow(unused_imports)]
 pub use _version::*;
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/rtc/mod.rs
```diff
@@ -17,15 +17,21 @@ pub use self::datetime::{DateTime, DayOfWeek, Error as DateTimeError};
...
+    rtc_v2f0, rtc_v2f2, rtc_v2f3, rtc_v2f4, rtc_v2f7, rtc_v2h7, rtc_v2l0, rtc_v2l1, rtc_v2l4, rtc_v2wb
+))]
+#[path = "v2.rs"]
+mod _version;
+
+#[cfg(any(rtc_v3, rtc_v3u5, rtc_v3l5, rtc_v3h7rs))]
+#[path = "v3.rs"]
+mod _version;
+
+#[cfg(not(any(
```
Dirbaio: we shouldn't have "defaults" like this. It's better to have compilation fail when adding new versions than to sliently use the wrong version, so you can notice and add the new version.

leftger: You're right. I was cracking my head for the past couple of hours. I think there's something deeper that doesn't want to build/generate the rest of the RTC modules. ~I only got about 2-3 to generate on the `embassy` repo but I get a ton more on `stm32-data`~

leftger (commented):
I'll set this PR as draft until I get this fixed.

embassy-stm32/src/rtc/mod.rs
```diff
@@ -25,7 +25,7 @@ use crate::time::Hertz;
     ),
     path = "v2.rs"
 )]
-#[cfg_attr(any(rtc_v3, rtc_v3u5, rtc_v3l5, rtc_v3h7rs), path = "v3.rs")]
+#[cfg_attr(any(rtc_v3, rtc_v3u5, rtc_v3l5, rtc_v3h7rs, rtc_v3c0), path = "v3.rs")]
```
leftger, line 28: A new peripheral was generated between the last `stm32-data-generated` hash in this codebase and the most recent one.

`rtc_v3c0` was recently added so this is simply incorporating that addition to `embassy`

leftger:
OK I'm pretty sure the fall through issue on rtc/mod.rs has been addressed now.

PR #4344, opened 2025-06-26
nxp: add feature for lpc55

i509VCB:
this is needed since I will be working on adding support for the MCX and iMXRT families to embassy-nxp


i509VCB:
@george-cosma any comments?

Eventually the plan is to put lpc55 stuff into nxp-pac, but need to make room for the new pac to exist (hence doing this).

i509VCB:
George said things looked fine over Matrix.

PR #4348, opened 2025-06-26
[stm32]: Replace per-channel traits in `timer` module with generic versions

melvdlin:
This changes the `embassy_stm32::timer` API to use traits that are generic over the timer channel instead of having a separate trait for every channel.

For instance, the traits `Channel1Pin`, `Channel2Pin`, `Channel3Pin` and `Channel4Pin` are replaced with a singular trait `TimerChannel<C>`, where `C` can be one of `Ch1`, `Ch2`, `Ch3` or `Ch4`.

Additionally, methods like
```rs
impl<'d, T: ...> CapturePin<'d, T, Ch1> {
    pub fn new_ch1(pin: Peri<'d, impl Channel1Pin<T>>, pull: Pull) -> Self { ... }
}

impl<'d, T: ...> CapturePin<'d, T, Ch2> {
    pub fn new_ch2(pin: Peri<'d, impl Channel1Pin<T>>, pull: Pull) -> Self { ... }
}

...
```
are replaced with a single, generic equivalent:
```rs
impl<'d, T: ..., C: TimerChannel> CapturePin<'d, T, C> {
    pub fn new(pin: Peri<'d, impl TimerPin<T, C>>, pull: Pull) -> Self { ... }
}
```


Dirbaio (approved):
very nice! thanks

PR #4349, opened 2025-06-26
[stm32h723]: fix spdifrx example

melvdlin:
The STM32H723 SPDIF RX example was not yet adapted to the new `Peri` API and thus failed to compile. This PR remedies that.


Dirbaio:
this should've been caught in CI but wasn't probably because `ci.sh` is missing building these examples. Can you add it so it doesn't happen again?

melvdlin:
done

PR #4350, opened 2025-06-27
Minimise build profile tweaking in rp/rp235x examples

9names:
To avoid confusion around build profile tweaks (see #4346), remove optimisation/lto changes and just enable generation of debug symbols for release profiles.
Also add a comment explaining that it is intentional so it is less likely to be seen as a mistake in the examples.


Dirbaio:
bender run

PR #4351, opened 2025-06-27
mspm0: Fix inverted GPIO logic

optlink:
`get_output_level()` was attempting to read the input buffer, which is currently disabled by this HAL when a pin is configured as an `Output`, resulting in a constant logical low reading. After fixing this I noticed that the equivalent input reading logic was inverted. This PR fixes both issues.


i509VCB:
Well that's embarrassing...

PR #4358, opened 2025-06-30
Use proper RCC clock enable for opamps

skoe:
OpAmp::new() now resets the opamp and enables its clock. The clock disabled when the opamp is dropped. On families that use SYSCFGEN (F3 and G4), this is not done because clock is always on in Embassy.
    
This change makes use of the RCC driver, which uses a reference counter to prevent conflicts. The opamp itself is still disabled when its output is dropped.

The PR intentionally does not add opamp features for more families yet to make reviewing easier.

The following table shows the opamp clock enable bits for all families, taken from RMs and/or CubeMX generated code:

| Family  | Example     | RCC       |
| ------- | ----------- | --------- |
| L1 (*)  | STM32L152RE | COMPEN    |
| F3      | STM32F303K8 | SYSCFGEN  |
| L4      | STM32L432KC | OPAMPEN   |
| L4+     | STM32L4P5ZG | OPAMPEN   |
| H7      | STM32H755ZI | OPAMPEN   |
| L5      | STM32L552ZE | OPAMPEN   |
| G4      | STM32G474RE | SYSCFGEN  |
| U5      | STM32U545RE | OPAMPEN   |
| H5      | STM32H503RB | OPAMPEN   |
| U0      | STM32U031R8 | OPAMPEN   |
| U3      | STM32U385RG | OPAMPEN   |

(*) L1 OPAMP not yet in stm32-data

### DIFF
--- a/embassy-stm32/src/opamp.rs
+++ b/embassy-stm32/src/opamp.rs
@@ -4,6 +4,8 @@
 use embassy_hal_internal::PeripheralType;
 
 use crate::pac::opamp::vals::*;
+#[cfg(not(any(stm32g4, stm32f3)))]
+use crate::rcc::RccInfo;
 use crate::Peri;
 
 /// Performs a busy-wait delay for a specified number of microseconds.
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/opamp.rs
```diff
@@ -600,6 +616,15 @@ foreach_peripheral!(
 foreach_peripheral! {
     (opamp, $inst:ident) => {
         impl SealedInstance for crate::peripherals::$inst {
+            fn info() -> &'static Info {
+                static INFO: Info = Info {
+                    #[cfg(any(stm32g4, stm32f3))]
+                    rcc: crate::peripherals::SYSCFG::RCC_INFO,
```
Dirbaio: many other things use syscfg, we shouldn't disable its clock on drop. We shouldn't refcount it either, syscfg is handled specially: it's enabled directly at boot and never disabled.

skoe: Thanks for the hint. :beer:

PR #4360, opened 2025-07-02
Link to esp-generate

bugadani:
Some people are trying to use cargo-embassy for our chips and end up disappointed. Let's link to the vendor-supported generator.


lulf (approved):
:rocket:

PR #4361, opened 2025-07-02
embassy-nrf: fix PWM loop count

jue89:
The PWM sequencer offers to run a sequence for a limited time by specifying `SingleSequenceMode::Times(n)` with `n > 0`.

Currently this value is ignored and the loop register is set to `1`.

This PR fixes the behaviour. It has been tested using a NRF9160 in secure mode.


lulf (approved):
Thanks!

PR #4363, opened 2025-07-03
rp: add new pio spi program

wowkster:
# Summary

This PR adds a new PIO based SPI driver to ``embassy-rp``. It supports both blocking and async operation using DMA transfers just like the hardware SPI driver, but currently can not be reconfigured at runtime without making breaking changes to the PIO APIs. 

# API Differences

I attempted to make the interface as similar as possible to the current hardware driver, but some changes were necessary to make configuration simpler. The biggest difference is that the PIO program has to be configured with the clock phase, since it isn't possible to use the same PIO program for CPHA0 and CPHA1 modes. Because of this limitation, I decided to create a separate ``Config`` structure for the PIO SPI driver which omits the phase.

# Unfinished Business

I think it would be possible to reconfigure the clock polarity and frequency at runtime by changing the state machine configuration, but changing the clock phase would be challenging. Currently there is a mechanism to free PIO instruction memory, but ownership of that memory is not tracked through the type system since ``LoadedProgram`` references are not kept after configuration.

# Open Questions

To make the async implementation work, I had to do what is akin to ``dma::read_repeated`` if it existed. Is it worth creating this function for situations where data needs to be discarded through DMA?

Additionally, The PIO implementation currently sets up its own DMA transfers since it needs access to the DMA ``CTRL_TRIG.BSWAP`` bit. Could these be refactored to use the DMA helper functions? The BSWAP bit could be set to the required value and then restored after the transfer is complete.

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -983,7 +983,7 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
         self.with_paused(|sm| {
             for pin in pins {
                 Self::this_sm().pinctrl().write(|w| {
-                    w.set_set_base(pin.pin());
+                    w.set_set_base(pin.pin() - if PIO::PIO.gpiobase().read().gpiobase() { 16 } else { 0 });
                     w.set_set_count(1);
                 });
                 // SET PINDIRS, (dir)
@@ -998,7 +998,7 @@ impl<'d, PIO: Instance + 'd, const SM: usize> StateMachine<'d, PIO, SM> {
         self.with_paused(|sm| {
             for pin in pins {
                 Self::this_sm().pinctrl().write(|w| {
-                    w.set_set_base(pin.pin());
+                    w.set_set_base(pin.pin() - if PIO::PIO.gpiobase().read().gpiobase() { 16 } else { 0 });
                     w.set_set_count(1);
                 });
                 // SET PINS, (dir)
@@ -1361,6 +1361,7 @@ impl<'d, PIO: Instance> Pio<'d, PIO> {
         PIO::state().users.store(5, Ordering::Release);
         PIO::state().used_pins.store(0, Ordering::Release);
         PIO::Interrupt::unpend();
+
         unsafe { PIO::Interrupt::enable() };
         Self {
             common: Common {
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -12,7 +12,7 @@ use fixed::types::extra::U8;
 use fixed::FixedU32;
 use pio::{Program, SideSet, Wrap};
 
-use crate::dma::{Channel, Transfer, Word};
+use crate::dma::{self, Channel, Transfer, Word};
 use crate::gpio::{self, AnyPin, Drive, Level, Pull, SealedPin, SlewRate};
 use crate::interrupt::typelevel::{Binding, Handler, Interrupt};
 use crate::relocate::RelocatedProgram;
@@ -281,6 +281,18 @@ impl<'l, PIO: Instance> Pin<'l, PIO> {
         });
     }
 
+    /// Configure the output logic inversion of this pin.
+    #[inline]
+    pub fn set_output_inversion(&mut self, invert: bool) {
+        self.pin.gpio().ctrl().modify(|w| {
+            w.set_outover(if invert {
+                pac::io::vals::Outover::INVERT
+            } else {
+                pac::io::vals::Outover::NORMAL
+            })
+        });
+    }
+
     /// Set the pin's input sync bypass.
     pub fn set_input_sync_bypass(&mut self, bypass: bool) {
         let mask = 1 << self.pin();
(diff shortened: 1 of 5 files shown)


CBJamo:
This is great! I'm super happy to see more pio programs getting added. I am curious about what you need more than 2 spi ports for, if you can talk about it publicly.

Feel free to change/refine LoadedProgram, it was just the most obvious way allow configuring multiple SMs without duplicating the program. FYI, I think you're the first person to implement any new pio programs in embassy-rp itself after I moved/normalized the examples.

I feel the same about the broader pio api. If changes are needed to expose more complete functionality then make them. I don't think anyone did super intensive design here, so we got here by adding features as new programs need them. Please correct me if that design did happen somewhere (probably matrix?) and I missed it.

wowkster:
Thank you! I found it interesting to implement and would love to answer your questions. As to why we needed more than 2 SPI ports, we are developing safety critical motion control hardware using the RP235xy and require high SPI bandwidth to simultaneously interact with a CAN FD controller, read motor status and encoder data from multiple TMC2240 stepper motor drivers, and read sensor feedback data from an SPI compatible high precision ADC, all with low latency. Given the different mode and frequency requirements of these devices and our ADC's specific quirks which make it difficult to use through the SpiDevice API, we found it simpler to allocate extra GPIO pins to a dedicated PIO driven SPI bus.

I think the LoadedProgram API is good since it allows sharing an installation across multiple state machines but it breaks down slightly when needing to reconfigure the state machines. Through an exclusive reference to the LoadedProgram, it would be safe to unload it and attempt to load a new altered version. However, without references being stored by the drivers after initialization, designing an API that isn't cumbersome to use would present some challenges. 

I propose for some of the PIO drivers to take ownership of their programs if the user wants to reconfigure them at runtime. It would cause some duplication, but in practice I think it would be minimal. Additionally, just like many peripherals use a type state parameter to track whether or not they own a DMA channel, maybe certain PIO drivers could use a type state parameter to keep track of whether they own the PIO instruction memory and can be reconfigured, or if they only hold a shared reference to it. This allows the same memory to be shared while also solving the issue of freeing instruction memory while the state machine is still running. If access to the LoadedProgram is not actually required when holding a reference, drivers could hold a phantom reference instead to make the abstraction zero-cost.

With the above changes, the PIO SPI driver could optionally support initial configuration as well as runtime reconfiguration in the same way that the SPI peripheral driver does.

My questions about the DMA APIs still hold but I recognize that this may not be the best place to address them.

Dirbaio:
bender run

CBJamo:
> ADC's specific quirks 

Why is it always an ADC where you end up needing dedicated or non-standard spi? Sounds like a cool project though!

> I propose for some of the PIO drivers to take ownership of their programs if the user wants to reconfigure them at runtime. 

In this case, since the program is either 2 or 3 instructions, I think we should skip the LoadedProgram step entirely. Just give the `Spi` struct exclusive ownership of 3 instruction slots and install the appropriate program for the polarity you need at runtime. In the small number of cases where a user needs the "wasted" slot, all the pio stuff is still exposed and they can do it by hand.

I also like your PhantomData idea, but it just doesn't feel like it's worth the complexity here with such small pio programs.

wowkster:
Yes it always seems to be the ADCs haha. 

I think this assessment is fair until we can demonstrate a need for a more general approach, so I would be happy to update my implementation to match your proposed design. Cases where you would need such control over the PIO memory seem like they would be pretty rare in any real application, but it might be worth revisiting this if more PIO programs are contributed.

I will update this PR as soon as I get the chance. Thanks for your input!

wowkster:
Do these changes look sufficient? I added runtime reconfiguration APIs similar to the builtin SPI peripheral. The only thing that I couldn't implement was the `SetConfig` trait from `embassy_embedded_hal` since reloading the program requires exclusive access to the PIO handle.

evalldor:
I was just about to open a PR for my PIO SPI implementation when I saw this. I solved the `SetConfig` by loading both programs into SM memory at init. I also did not use DMA pull_repeated and instead waited for the SM to stall before returning from a write. You can see my branch here: https://github.com/evalldor/embassy/blob/pio-spi/embassy-rp/src/pio_programs/spi.rs

wowkster:
At slower SPI speeds, that may block for a sufficiently long period which could be undesirable.  I think the configuration change is sensible and I don't know that one approach is necessary better than the other. Personally I would err on the side of using less memory if it's possible.

CBJamo:
If we loaded both programs and put them in a `SpiPioProgram` struct, then that struct could be shared between multiple SMs. Compared to the previous plan we're only down 2 instructions, and it would come out ahead as soon as there was a second SM configured for SPI. Since it would also allow for `SetConfig` that seems like a good tradeoff to me. 

I agree that DMA is essential for SPI. In some applications (sd cards being the most common) there are large and fast transfers. In those cases the SM's fifo is too small, and it's going fast enough you don't want the cpu to get involved.

wowkster:
I like that idea and I think it's a good balance. Should I go ahead and implement that?

CBJamo:
I think it's the best design we've discussed so far. If we ever come up with a better one we can always have a follow on PR.

wowkster:
Ok I will get that in ASAP

9names:
This PR should be closed since #4649 incorporates this and it has already been merged

PR #4366, opened 2025-07-03
Add note about UART line breaks being different from ASCII

mattwigway:
I confused myself for a while because I'm working with a GPS module that sends an ASCII line break character (\n) to indicate the end of a transmission. I initially thought `read_to_break` would read until it saw a line break character, but it actually does the (much more sensible) reading until a UART break condition is encountered. This adds that distinction to the docs.


jamesmunns (approved):

Dirbaio:
I've never seen uart break called "line break". I think it'd be better to completely remove the name "line break" from docs, and call it just "break", or "break condition"

mattwigway:
@Dirbaio done!

Dirbaio:
thanks!

PR #4373, opened 2025-07-06
embassy-executor: rtos-trace: fix task naming for new tasks

fg-cfh:
Tasks that are spawned after starting SystemViewer were not named. This change ensures that tasks spawned while SystemViewer is running will be properly named, too.


ghost:
@Dirbaio This one is independent of #4383. It's a simple bug fix and could be reviewed/merged already.

PR #4372, opened 2025-07-06
embassy-executor: trace: name main task

fg-cfh:
This change ensures that the main task will be named when tracing is active. This improves tracing usability.

PR #4383, opened 2025-07-08
executor: add task metadata, misc trace improvements.

Dirbaio:
- **executor: do not store task IDs in RAM, we can get it from the pointer every time.**
- **executor: allow trace and rtos-trace to coexist additively.**
- **executor: add "task metadata" concept, make name a task metadata.**
- **executor: do not deref a mut ptr to the entire taskheader.**

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -31,7 +31,9 @@ cargo batch \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt,arch-cortex-m,executor-thread,executor-interrupt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,trace \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,rtos-trace \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,trace,rtos-trace \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,executor-thread \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,executor-interrupt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,executor-thread,executor-interrupt \
--- a/ci.sh
+++ b/ci.sh
@@ -31,6 +31,7 @@ cargo batch \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt,arch-cortex-m,executor-thread,executor-interrupt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,metadata-name \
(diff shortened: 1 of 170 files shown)


embassy-executor/src/metadata.rs
```diff
@@ -0,0 +1 @@
+pub struct Metadata {}
```
ghost: Seems to be in the wrong commit. Should be introduced in the 'add "task metadata" concept.' commit.

embassy-executor/src/raw/trace.rs
```diff
@@ -84,6 +84,7 @@
 use core::cell::UnsafeCell;
 use core::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
 
+#[cfg(feature = "rtos-trace")]
```
ghost: Unrelated change? Should probably be moved to the next commit ( allow trace and rtos-trace to coexist additively).

ci.sh
```diff
@@ -31,7 +31,9 @@ cargo batch \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt,arch-cortex-m,executor-thread,executor-interrupt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,trace \
```
ghost, line 35: These changes do not obviously match the commit message. Wrong commit?

embassy-executor/src/raw/mod.rs
```diff
@@ -368,7 +368,7 @@ impl<F: Future + 'static, const N: usize> TaskPool<F, N> {
     /// This will loop over the pool and spawn the task in the first storage that
     /// is currently free. If none is free, a "poisoned" SpawnToken is returned,
     /// which will cause [`Spawner::spawn()`](super::Spawner::spawn) to return the error.
-    pub fn spawn(&'static self, future: impl FnOnce() -> F) -> SpawnToken<impl Sized> {
+    pub fn spawn(&'static self, future: impl FnOnce() -> F) -> Result<SpawnToken<impl Sized>, SpawnError> {
```
ghost, line 369: This seems to be a rather important breaking change requiring general visibility in changelogs, migration docs, etc.

I propose that we introduce a non-breaking additive change in stable embassy with a separate spawn_with_metadata() function, at the same time deprecating must_spawn() and spawn_named(). Then have a separate PR with the breaking changes.

The existing spawn function may call spawn_with_metadata() with Default::default() as proposed by @jamesmunns on Matrix.

Please disregard if this project doesn't maintain separate branches as proposed. I'm new to embassy so might get this wrong.

embassy-executor/src/spawner.rs
```diff
@@ -198,51 +203,6 @@ impl Spawner {
...
-pub trait SpawnerTraceExt {
-    /// Spawns a new task with a specified name.
-    ///
-    /// # Arguments
-    /// * `name` - Static string name to associate with the task
-    /// * `token` - Token representing the task to spawn
-    ///
-    /// # Returns
-    /// Result indicating whether the spawn was successful
-    fn spawn_named<S>(&self, name: &'static str, token: SpawnToken<S>) -> Result<(), SpawnError>;
```
ghost, line 214: This is a breaking change. I think embassy might want to provide a backwards compat implementation as a minor release. I'm not so familiar with embassy's versioning approach so please disregard if you don't have such a concept in this project.

fg-cfh (approved):
Like this change. :-) Just a few organisational nits.

Marked #4372 as depending on this change and made it a draft. Will update #4372 as soon as this change lands.

embassy-executor/src/raw/mod.rs
Dirbaio: Next embassy-executor release is going to be a major bump anyway (0.8.0). 

the `spawn_with_metadata()` idea is interesting though, hmm.

lulf (approved):
Making the spawn token a result is nice, makes the API cleaner IMHO :rocket:

ghost:
@Dirbaio Is it possible that the build problem was just temporary? Maybe restarting the job will fix it...

Dirbaio:
bender run

jamesmunns:
Noting that #4035 also has additional metadata in the form of `Deadline`s, whichever lands first should be refactored correctly with whichver lands second.

Dirbaio:
#4606

PR #4389, opened 2025-07-10
fix(executor): pin the executor in `raw::Executor::id()`

ROMemories:
This is a follow-up to #4049 which introduced `Executor::id()`. This PR pins the executor inside that method, so that two executors ID being different is a sufficient condition for the executors they originate from be different.

It additionally updates that method's documentation to clarify this additional guarantee.

### DIFF
--- a/embassy-executor/src/raw/mod.rs
+++ b/embassy-executor/src/raw/mod.rs
@@ -568,8 +568,13 @@ impl Executor {
     }
 
     /// Get a unique ID for this Executor.
+    ///
+    /// Two executors are the same if and only if their IDs are the same.
     pub fn id(&'static self) -> usize {
-        &self.inner as *const SyncExecutor as usize
+        // Pinning is required to be sure the executor stays at the same address, which is directly
+        // used to build the ID. This is required to be able to conclude from different executors
+        // IDs that they indeed originate from different executors.
+        &Pin::static_ref(self).inner as *const SyncExecutor as usize
     }
 }
 


embassy-executor/src/raw/mod.rs
```diff
@@ -568,8 +568,13 @@ impl Executor {
...
 
     /// Get a unique ID for this Executor.
+    ///
+    /// Two executors are the same if and only if their IDs are the same.
     pub fn id(&'static self) -> usize {
-        &self.inner as *const SyncExecutor as usize
+        // Pinning is required to be sure the executor stays at the same address, which is directly
+        // used to build the ID. This is required to be able to conclude from different executors
+        // IDs that they indeed originate from different executors.
+        &Pin::static_ref(self).inner as *const SyncExecutor as usize
```
ROMemories, line 577: Because there already is a `&'static` reference to the executor, I believe pinning is free.

Dirbaio:
`&'static` already guarantees the address won't change, creating a Pin or not makes no difference.

ROMemories:
> `&'static` already guarantees the address won't change, creating a Pin or not makes no difference.

That's what allows [`Pin::static_ref()`](https://doc.rust-lang.org/stable/std/pin/struct.Pin.html#method.static_ref) to even exist in the first place, but [that wasn't so obvious](https://github.com/rust-lang/rust/issues/78186) and pinning makes it clear that the implementation directly relies on the executor's location in memory.

Dirbaio:
`&'static` already makes that clear. I don't think we should merge this, it's not needed for correctness and makes the code more complicated.

PR #4393, opened 2025-07-11
nrf-nrf54l: add basic nrf54l support

dmitry-tarnyagin:
Here we have implemented basic support for the following peripheral blocks of the nRF54L:

 * uarte, buffered_uarte
 * twim, twis
 * spim, spis
 * saadc
 * pwm
 * dppic, instantiated for dppic 00, 20, 30
 * gpiote, instantiated for gpiote 20, 30
 * timer

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -12,7 +12,7 @@ use core::cmp::min;
 use core::future::{poll_fn, Future};
 use core::marker::PhantomData;
 use core::slice;
-use core::sync::atomic::{compiler_fence, AtomicBool, AtomicPtr, AtomicU8, AtomicUsize, Ordering};
+use core::sync::atomic::{compiler_fence, AtomicBool, AtomicU8, AtomicUsize, Ordering};
 use core::task::Poll;
 
 use embassy_hal_internal::atomic_ring_buffer::RingBuffer;
@@ -39,7 +39,6 @@ pub(crate) struct State {
     rx_started_count: AtomicU8,
     rx_ended_count: AtomicU8,
     rx_ppi_ch: AtomicU8,
-    rx_ppi_regs: AtomicPtr<()>,
 }
 
 /// UART error.
@@ -61,7 +60,6 @@ impl State {
             rx_started_count: AtomicU8::new(0),
             rx_ended_count: AtomicU8::new(0),
             rx_ppi_ch: AtomicU8::new(0),
-            rx_ppi_regs: AtomicPtr::new(core::ptr::null_mut()),
         }
     }
 }
@@ -119,18 +117,11 @@ impl<U: UarteInstance> interrupt::typelevel::Handler<U::Interrupt> for Interrupt
                     r.rxd().ptr().write_value(ptr as u32);
                     r.rxd().maxcnt().write(|w| w.set_maxcnt(half_len as _));
 
-                    #[cfg(feature = "_ppi")]
-                    type Ppic = pac::ppi::Ppi;
-                    #[cfg(feature = "_dppi")]
-                    type Ppic = pac::dppic::Dppic;
-
                     let chn = s.rx_ppi_ch.load(Ordering::Relaxed);
-                    let regs_ptr = s.rx_ppi_regs.load(Ordering::Relaxed);
-                    let regs = Ppic::from_ptr(regs_ptr);
 
                     // Enable endrx -> startrx PPI channel.
                     // From this point on, if endrx happens, startrx is automatically fired.
-                    regs.chenset().write(|w| w.0 = 1 << chn);
+                    ppi::regs(()).chenset().write(|w| w.0 = 1 << chn);
 
                     // It is possible that endrx happened BEFORE enabling the PPI. In this case
                     // the PPI channel doesn't trigger, and we'd hang. We have to detect this
@@ -154,7 +145,7 @@ impl<U: UarteInstance> interrupt::typelevel::Handler<U::Interrupt> for Interrupt
 
                     // Check if the PPI channel is still enabled. The PPI channel disables itself
                     // when it fires, so if it's still enabled it hasn't fired.
-                    let ppi_ch_enabled = regs.chen().read().ch(chn as _);
+                    let ppi_ch_enabled = ppi::regs(()).chen().read().ch(chn as _);
 
                     // if rxend happened, and the ppi channel hasn't fired yet, the rxend got missed.
                     // this condition also naturally matches if `!started`, needed to kickstart the DMA.
@@ -162,7 +153,7 @@ impl<U: UarteInstance> interrupt::typelevel::Handler<U::Interrupt> for Interrupt
                         //trace!("manually starting.");
 
                         // disable the ppi ch, it's of no use anymore.
-                        regs.chenclr().write(|w| w.set_ch(chn as _, true));
+                        ppi::regs(()).chenclr().write(|w| w.set_ch(chn as _, true));
 
                         // manually start
                         r.tasks_startrx().write_value(1);
@@ -693,9 +684,6 @@ impl<'d, U: UarteInstance, T: TimerInstance> BufferedUarteRx<'d, U, T> {
         ppi_ch1.enable();
 
         s.rx_ppi_ch.store(ppi_ch2.number() as u8, Ordering::Relaxed);
-        s.rx_ppi_regs
-            .store(ppi::regs(ppi_ch2.inst()).as_ptr(), Ordering::Relaxed);
-
         let mut ppi_group = PpiGroup::new(ppi_group);
         let mut ppi_ch2 = Ppi::new_one_to_two(
             ppi_ch2,


dmitry-tarnyagin:
Checks passed, ready for review.

embassy-nrf/src/compat_nrf54l.rs
```diff
@@ -0,0 +1,353 @@
+#![allow(missing_docs)]
+use extend::ext;
+
+use crate::pac::common::{Reg, RW, W};
+use crate::pac::{gpio, pwm, spim, spis, twim, twis, uarte};
+
+#[repr(transparent)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct SpiFrequencyVals(pub u32);
```
lulf, line 9: Why define the values here instead of using them from nrf-pac for nrf54l? Is that because they're missing from the pac? If so, maybe it can be fixed in the pac.

embassy-nrf/src/buffered_uarte_nrf54l.rs
```diff
@@ -0,0 +1,688 @@
...
+                }
+            }
+        }
+
+        //trace!("irq: end");
+    }
+}
+
+/// Buffered UARTE driver.
+pub struct BufferedUarte<'d, U: UarteInstance> {
```
lulf, line 157: What is the reason for adding a separate BufferedUarte implementation for nrf54l?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) struct State {
     rx_started_count: AtomicU8,
     rx_ended_count: AtomicU8,
     rx_ppi_ch: AtomicU8,
+    rx_ppi_regs: AtomicPtr<()>,
```
lulf: Could this be some kind of Ppi instance type (or AnyPpi if you want to erase the type) instead of a raw pointer?

embassy-nrf/src/ppi/dppi.rs
```diff
@@ -4,8 +4,48 @@ use crate::{pac, Peri};
...
+#[derive(Copy, Clone)]
+/// DPPI instance
+pub enum DppiInstance {
+    /// DPPI_00 instance
+    Dppi00,
+    /// DPPI_20 instance
+    Dppi20,
+    /// DPPI_30 instance
+    Dppi30,
+}
```
lulf, line 17: Could this instead be built into some type erased AnyPpi type (similar to AnyPin)?

lulf (commented):
Thanks for the PR! A lot of stuff, so this will take some time to review in detail. After a first look I have several questions in the review comments. Overall, there's more code duplication than I expected, and I'm not sure about using the extend crate, which makes the nrf54 very 'bolted on' instead of doing the necessary rewrite to make the HAL flexible to support nrf54. The other thing is the use of enums for instance types, where the usual pattern has been to have an Instance trait that gets implemented for `FooX`, and potentially type-reased when converting to `AnyFoo`.

embassy-nrf/src/compat_nrf54l.rs
dmitry-tarnyagin: They are quite incompatible, unfortunately. The nRF54L has a prescaler register, and the integer value is interpreted as the core-clock-to-SCK divisor. Other nRFs use magic constants (reflected in the PAC) and a frequency register instead of a prescaler. The SpiFrequencyVals here is an attempt to maintain a common API for all the chips.

embassy-nrf/src/buffered_uarte_nrf54l.rs
dmitry-tarnyagin: UARTE blocks are implemented quite differently in nRF54L. In other nRFs, the number of transferred bytes can be counted as the number of RXRDY events. It’s a bit racy, since RXRDY means “received,” not “written to memory,” but it usually works—the inconsistency window is only a few 16 MHz cycles.

In nRF54, the UARTE block (or rather the EasyDMA block) implements buffering. Bytes are collected into u32 words before being sent to memory. The number of RXRDY events no longer matches the number of transferred bytes, and bytes in the internal buffer aren’t transferred to memory until the DMA END event, which can never happen if a packed size is not multiple-of-four.

The only way to flush the RX buffer is to trigger a STOP task and wait for DMA END. Fortunately, Nordic has implemented an RX timeout as a separate event, so the strategy is: wait for timeout → force stop → restart RX.

The common buffered_uarte implementation uses a PPI network to count transferred bytes via RXRDY events. nRF54L uses the timeout‐stop‐restart strategy and doesn’t need external PPI machinery at all.

An attempt to keep both implementations in the same code failed under a tangle of #[cfg]s and parallel code, so there are now two separate files.

embassy-nrf/src/buffered_uarte.rs
dmitry-tarnyagin: It can be completely removed as it’s not relevant to the nRF54. It was part of a naive attempt to just tune the buffered_uarte for the nRF54.

embassy-nrf/src/ppi/dppi.rs
dmitry-tarnyagin: I can try..

dmitry-tarnyagin:
@lulf 
Thanks for the comments!

I’ll clean up the original buffered_uarte.rs, removing rx_ppi_regs and/if any other leftover development code.

I think it’s fine to bolt on nRF54L for now. The massive change in the nRF54 register binary interface is probably due to the new DMA.RX.MATCH functionality, which is useless because of errata 44. Right now, I don’t want to ruin the existing embassy-nrf code with tons of

if #[cfg(feature = "_nrf54l")] {
    r.dma().rx()…
} else {
    r.txd()…
}, 
it's too ugly for no benefit.

If a dependency on the extend crate is problematic, the compatibility layer can be implemented manually using traits / impl traits for.

Again, thanks for the feedback! Please share any further comments and suggestions. Much appreciated!

Regards, Dmitry

embassy-nrf/src/ppi/dppi.rs
dmitry-tarnyagin: ..Update: The PPI implementation closely resembles the existing implementation of GPIO. Here, the DPPI instance is a direct analogue of a GPIO port. I think it’s quite in-style :).
Existing AnyChannel and AnyGroup are already extended with controller instances.

dmitry-tarnyagin:
non-54l buffered_uarte.rs is reverted.

dmitry-tarnyagin:
Hi @lulf , I'm just wondering what the way forward is here :)

Regards, Dmitry

Dirbaio:
I think it might be possible to use chiptool transforms in `nrf-pac` to change the old-style `r.txd()` in the old chips to the new-style `r.dma().rx()` so that they're all compatible.

This'd allow supporting all chips with no cfg's and no strange extension traits.

lulf:
> Hi @lulf , I'm just wondering what the way forward is here :)
> 
> Regards, Dmitry

Sorry for the slow reply, thanks for your comments. I agree with @Dirbaio here, if transforms can be added to nrf-pac and simplify the HAL code, that would be ideal.

dmitry-tarnyagin:
Hi, @lulf, @Dirbaio;

Thanks for your comments!

Unfortunately, the situation isn't as simple as it might look alike. Replacing "r.txd()-style" with "r.dma().tx()-style" in older chips could cause existing projects based on widely used nRF52, nRF91, and nRF53 series to fail to compile.

The main issue here isn't with rxd() and txd(), but rather with the mass renaming of tasks and events. The rxd and txd registers are primarily internal to the driver and rarely named directly in client code. However, tasks and events are naturally and widely used in creating PPI networks, and renaming them would break existing user implementations. For reference, more than 50% of our commercial software utilizes quite sophisticated PPI networks.

In addition, the periphery binary interface changes introduced in the nRF54 series appear experimental, and it's unclear whether Nordic will retain this approach or shift to something even newer in future microcontroller families. If we proceed with adopting this new register organization despite user concerns (me screaming:)), there's no guarantee we won't be forced to repeatedly adapt to further such changes in the future.

I believe it would be better to adopt an opposite strategy: preserve (existing) interface compatibility across all Nordic chips as much as possible, unless there's an absolutely clear reason for change and fork (buffered UART being an exception, for example).

Cheers, Dmitry

lulf:
> Hi, @lulf, @Dirbaio;
> 
> Thanks for your comments!
> 
> Unfortunately, the situation isn't as simple as it might look alike. Replacing "r.txd()-style" with "r.dma().tx()-style" in older chips could cause existing projects based on widely used nRF52, nRF91, and nRF53 series to fail to compile.
> 
> The main issue here isn't with rxd() and txd(), but rather with the mass renaming of tasks and events. The rxd and txd registers are primarily internal to the driver and rarely named directly in client code. However, tasks and events are naturally and widely used in creating PPI networks, and renaming them would break existing user implementations. For reference, more than 50% of our commercial software utilizes quite sophisticated PPI networks.
> 
> In addition, the periphery binary interface changes introduced in the nRF54 series appear experimental, and it's unclear whether Nordic will retain this approach or shift to something even newer in future microcontroller families. If we proceed with adopting this new register organization despite user concerns (me screaming:)), there's no guarantee we won't be forced to repeatedly adapt to further such changes in the future.
> 
> I believe it would be better to adopt an opposite strategy: preserve (existing) interface compatibility across all Nordic chips as much as possible, unless there's an absolutely clear reason for change and fork (buffered UART being an exception, for example).
> 

Thanks for explaining your situation, so if I understand you correctly, you already depend on nrf-pac API for other parts of your firmware and would like to avoid rewriting that.

However, I don't think there is anything that would require you to rewrite your product to use newer nrf-pac. It could be using the existing version '0.1.0' with the existing API alongside embassy-nrf using the new one '0.2.0'.

dmitry-tarnyagin:
> It could be using the existing version '0.1.0' with the existing API alongside embassy-nrf using the new one '0.2.0'.

It's not optimal for heterogeneous systems (typical in our case: nRF91+nRF5x), where we otherwise could share code between nRF91 and nRF54.

Just to clarify: It's not an issue for us at the moment, as we _currently_ don't have any non-54 embassy-nrf projects. However, I'm concerned that other embassy-nrf users may be unhappy. Curiously, if the problem with a sudden pac api change does arise, we would likely solve it using exactly the proposed extend trait strategy to tame the new API :).

Dirbaio:
Changing the new layout to the old layout sounds good to me as well.

I think you'd need to add some "move regs from block to parent, delete block" transform to chiptool for this, while the reverse could work with MakeBlock.

dmitry-tarnyagin:
> Changing the new layout to the old layout sounds good to me as well.

Can we combine both the old and the new layout? Keeping the dma blocks, but exposing also as legacy aliases? What do you think?

Dirbaio:
I'd prefer there to be only one way to do things.

dmitry-tarnyagin:
> I'd prefer there to be only one way to do things.

Then an open question remains: how should _new_ registers of the nRF54 be mapped from DMA blocks if the registers do not exist in older MCUs?

For more or less all blocks:
 - events_dma.tx/rx.busterror
 - events_dma.rx.match(n)
 - tasks_dma.enable_match(n)
 - tasks_dma.disable_match(n)
 (match is quite useless at the moment due to an errata, men can be used in some cases in ppi) 
 For PWM, there is no equivalent for DMA events at all
 
 Variants: 
  - keep the existing dma blocks, but move "classic" registers to the parent
  - totally ignore the non-classic registers
  - keep the existing dma blocks, generate "classic" alias registers at the parent level (in chiptool, no extension traits)
  - invent new "registers" at the parent level, i.e. events_dma.rx.match(n) -> events_dma_rx_match(n)
  
  What's your opinion?

Dirbaio:
honestly I think that's an argument in favor of changing old chips to the new layout.

dmitry-tarnyagin:
Sorry for late replay, I'm on vacation.
Unfortunately we do not have a budget for massive redesign of classic nrfs at the moment. It might be we will be able to do it later.
With best regards,
Dmitry

lulf:
I understand completely, and thanks for your efforts so far. Luckily I suppose there is no rush to merge anything, since you have something that works for you. I also think as nrf54s will get more popular, more people will step up to work on this as well.

oxabz:
Hi @dmitry-tarnyagin!

I've been working on [nrf54l support](https://github.com/oxabz/embassy/tree/nrf54l15-support) too.

On the matter of the registers being renamed, I've settled on a similar solution as yours. 

I'm curious about your implementation of PPI. 
From what I understand from the code, nothing forbids the API consumer from creating channels with events from two different DPPI peripherals. It looks like the DPPI could fail silently. This could maybe be fixed by automatically creating DPPI bridges as needed, but it would still leave the possibility of a silent fail if all the bridges are used.

In my implementation I type-encoded the peripheral's domain to enforce DPPI constraints, but it's a bit clunky and makes type-erased peripheral generic over "domains", which is not ideal for the size of the binary.

What's your opinion on the domain separations?
Regards,
oxabz

lulf:
Just a heads up, I'm looking into doing more transforms in nrf-pac now to avoid too much cfg in embassy-nrf. Aiming to get it into a state where most existing peripherals can be enabled without too much cfg's

dmitry-tarnyagin:
nrf54 support is implemented in the https://github.com/embassy-rs/embassy/pull/4793
Closing this PR :)

PR #4396, opened 2025-07-13
Fix CDC ACM BufferedReceiver buffer calculation

PegasisForever:


embassy-usb/src/class/cdc_acm.rs
```diff
@@ -501,7 +501,7 @@ impl<'d, D: Driver<'d>> BufferedReceiver<'d, D> {
     fn read_from_buffer(&mut self, buf: &mut [u8]) -> usize {
         let available = &self.buffer[self.start..self.end];
         let len = core::cmp::min(available.len(), buf.len());
-        buf[..len].copy_from_slice(&self.buffer[..len]);
+        buf[..len].copy_from_slice(&self.buffer[self.start..(self.start + len)]);
```
rursprung: couldn't you instead read directly from `available` which already uses the `self.start..self.end` slice?


```suggestion
        buf[..len].copy_from_slice(available[..len]);
```

matt-rodgers: Copying from `available[..len]` is what I intended to write when I added the `BufferedReceiver` support - thanks for catching the silly mistake 🤦

PegasisForever:
@rursprung thanks for the suggestion!

Dirbaio:
can you fix ci?

PR #4397, opened 2025-07-14
Make USB endpoint allocator methods accept an optional `EndpointAddress`

korbin:
This PR adds an optional `ep_addr` to all USB endpoint allocation methods. If a method receives`None`, the old incremental allocation behavior applies.

This enables developers to select a specific endpoint index, enabling implementation and emulation of devices where endpoints are hardcoded (e.g. FTDI USB-to-serial adapters).

I am still testing this on the various platforms that I have on hand, just wanted to get this open for discussion re: modifying these interfaces; there are a few places where this parameter doesn't necessarily need to be exposed (we can default to `None`).

### DIFF
--- a/embassy-stm32/src/usb/usb.rs
+++ b/embassy-stm32/src/usb/usb.rs
@@ -362,6 +362,7 @@ impl<'d, T: Instance> Driver<'d, T> {
     fn alloc_endpoint<D: Dir>(
         &mut self,
         ep_type: EndpointType,
+        ep_addr: Option<EndpointAddress>,
         max_packet_size: u16,
         interval_ms: u8,
     ) -> Result<Endpoint<'d, T, D>, driver::EndpointAllocError> {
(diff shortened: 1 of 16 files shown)


embassy-stm32/src/usb/usb.rs
```diff
@@ -373,25 +374,56 @@ impl<'d, T: Instance> Driver<'d, T> {
...
+                    // Do not reuse this index.
+                    // FIXME: Bulk endpoints can be double buffered, but are not in the current implementation.
+                    return false;
+                }
+
+                let used_dir = match D::dir() {
+                    Direction::Out => ep.used_out,
+                    Direction::In => ep.used_in,
+                };
+                !used || (ep.ep_type == ep_type && !used_dir)
```
mkj: Maybe this logic could go into a `is_valid()` function shared between the specified/find() cases for readability?

korbin:
> Looks useful, thanks. Will the compiler be smart enough to discard the `ep_addr: Some` code paths when `None` is passed?

I can't get linking to work on Godbolt to test for sure, but it seems like these branches are only a handful of extra instructions on ARM - much of the logic is shared; I feel like tree shaking or link time optimization should be able to prune them for binary size optimization. These functions are only called tens of times in the worst case scenario - so there's no runtime performance penalty to consider.

okhsunrog:
@korbin you can use cargo-show-asm, works for me

PR #4402, opened 2025-07-15
feat: add embassy-release tool

lulf:
* Print dependencies of a crate
* Bump version in dependent crates
* Release using cargo release
* Config file to control features and target

The idea is that the tool can be used to simplify the tasks of releasing, and gradually performing more and more of the tasks required to do a release. It's not intended as yet another release tool for rust, but for calling out to other tools to do the tasks required for a release.

### DIFF
--- a/embassy-release/src/main.rs
+++ b/embassy-release/src/main.rs
@@ -1,0 +1,305 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fs,
+    path::{Path, PathBuf},
+};
+
+use clap::{Parser, Subcommand, ValueEnum};
+use regex::Regex;
+use serde::Deserialize;
+use std::process::Command as ProcessCommand;
+use toml_edit::{DocumentMut, Item, Value};
+use walkdir::WalkDir;
+
+/// Tool to traverse and operate on intra-repo Rust crate dependencies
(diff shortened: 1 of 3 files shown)


embassy-release/src/main.rs
```diff
@@ -0,0 +1,305 @@
...
+    //    .status()
+    //    .map_err(|e| format!("Failed to run cargo release --execute: {}", e))?;
+
+    //if !status.success() {
+    //    return Err(format!(
+    //        "`cargo release {kind_str} --execute` failed in crate {crate_name}"
+    //    ));
+    //}
+
+    Ok(())
```
lulf, line 259: Intentionally commented to avoid accidental releasing while developing :)

rursprung: how about adding a `--execute` to this tool as well and only call it if that is passed along?

lulf: Somehow adding cargo-semver-checks would be nice to suggest/warn whether to do a minor/patch release.

i509VCB:
I guess with this xtask like stuff, would be nice to have a way to run cargo +nightly fmt on the whole repo. Of course doesn't need to be done now.

PR #4405, opened 2025-07-16
Add `derive(Debug)` for `embassy_rp::i2c::I2c`

rursprung:
i've split this in two commits since i'm not sure if you want to take the second one (enabling the warning without adding all the derives - on the other hand, if i'd add them then i'd turn it into a `deny`)


rursprung:
for now i've just removed the 2nd commit which added `#![warn(missing_debug_implementations)]` to `embassy-rp/src/lib.rs`
probably it makes more sense to add it as `deny` and just clean up all findings if CI anyway trips up on it (though i don't understand why it doesn't trip up on all the other warnings which are also being issued...)

PR #4408, opened 2025-07-17
Add STM32WBA USB_OTG_HS support

leftger:
This PR attempts to align the USB_OTG_HS support present in the STM32U5 series now to the STM32WBA family of devices.

- Some register names were slightly altered so those had to be addressed.
- An enum had to be handled.
- Thank you to Dario for fixing the build issues I was having 😄


Dirbaio:
lgtm :rocket:

PR #4409, opened 2025-07-17
mspm0: add buffered uart driver

i509VCB:


embassy-mspm0/src/uart/mod.rs
```diff
@@ -1021,9 +1060,29 @@ fn read_with_error(r: Regs) -> Result<u8, Error> {
...
 
+/// This function assumes CTL0.ENABLE is set (for errata cases).
+fn busy(r: Regs) -> bool {
+    // Errata UART_ERR_08
+    if cfg!(any(
+        mspm0g151x, mspm0g351x, mspm0l110x, mspm0l130x, mspm0l134x, mspm0c110x,
+    )) {
+        let stat = r.stat().read();
+        // "Poll TXFIFO status and the CTL0.ENABLE register bit to identify BUSY status."
+        !stat.txfe() || !stat.rxfe()
```
Iooon: I don't really understand this comment. Don't we poll TXFIFO and RXFIFO status?

i509VCB: So that's a verbatim quote from the errata. In order to match the behavior of busy, which is whether TX or RX are active, we need to check if anything is in either of those fifos. 

(Busy implies both, but the errata only says you check one? Yes that is a contradiction)

i509VCB: Actually 

> This bit is set as soon as the transmit FIFO or TXDATA register 
becomes non-empty (regardless of whether UART is enabled) or if a 
receive data is currently ongoing (after the start edge have been 
detected until a complete byte, including all stop bits, has been 
received by the shift register).
In IDLE_Line mode the Busy signal also stay

So it's rising edge of RX or anything in TX fifo, so yes I did make a mistake there.

i509VCB:
When CI returns I will merge this.

PR #4412, opened 2025-07-17
Removed reference to ADC3 for STM32G4x1 if the peripheral does not exist

Wassasin:
Fixes regression introduced in https://github.com/embassy-rs/embassy/pull/4390 for STM32G431 and STM32G441, which do not have an ADC3. Results in the build failing.

STM32G491 and STM32G4A1 do have the instance, so this is the easiest way to gate these `impl`s.


Dirbaio:
thanks!

PR #4413, opened 2025-07-17
STM32F107: Fix inadvertent re-configuration of the SWJ/JTAG pins when activating the (R)MII interface

fwolter:
AFIO_MAPR.SWJ_CFG must be written as 0b111 to leave its value unchanged.


Dirbaio:
thanks!

Issue #3612, opened 2024-12-05
cyw43 on rp pico 2 w


rztz:
The same code runs on rp2040 with cyw43, but dies on rp2350 with cyw43.
On rp2350 it panics at bp_read addr is not dividable by 4.

Disabling firmware-logs skips this code, but the module also does not work.
It reports something like 'TX stalled' or 'len inv mismatch'

Compare the following trace dumps side by side:

rp2350:
0.001596 WARN  main()
0.001224 TRACE WL_REG off/on
0.271306 TRACE read REG_BUS_TEST_RO
0.271393 TRACE 0x3030303
0.271438 TRACE 0xfeedbead
0.271459 TRACE write REG_BUS_TEST_RW
0.271587 TRACE 0x12345678
0.271610 TRACE read REG_BUS_CTRL
0.271643 TRACE 0b00110000
0.271662 TRACE write REG_BUS_CTRL
0.271692 TRACE read REG_BUS_CTRL
0.271756 TRACE 0b10110001
0.271776 TRACE read REG_BUS_TEST_RO
0.271818 TRACE 0xfeedbead
0.271841 TRACE read REG_BUS_TEST_RW
0.271877 TRACE 0x12345678
0.271901 TRACE write SPI_RESP_DELAY_F1 CYW43_BACKPLANE_READ_PAD_LEN_BYTES
0.271944 TRACE Make sure error interrupt bits are clear
0.271972 TRACE enable a selection of interrupts
0.272008 DEBUG init alp
0.272038 DEBUG set f2 watermark
0.272093 DEBUG watermark = 10
0.272112 DEBUG waiting for clock...
0.272152 DEBUG clock ok
0.272164 DEBUG clear request for ALP
0.272200 TRACE backplane_readn addr = 18000000 len = 2
0.272301 TRACE backplane_readn addr = 18000000 len = 2 val = 2d85a9af
0.272340 DEBUG chip ID: 43439
0.272376 TRACE backplane_readn addr = 18103800 len = 1
0.272436 TRACE backplane_readn addr = 18103800 len = 1 val = 2d850001
0.272479 TRACE backplane_readn addr = 18103800 len = 1
0.272523 TRACE backplane_readn addr = 18103800 len = 1 val = 2d850001
0.272566 TRACE backplane_readn addr = 18104800 len = 1
0.272608 TRACE backplane_readn addr = 18104800 len = 1 val = 2d850001
[...]

Issue #3960, opened 2025-03-11
Cyw43 read address sometimes misaligned


Abestanis:
I was experiencing random crashes after 24+ hours of running a project, and the panic points to an assert in the `cyw43` crate:
```
cyw43-0.3.0\src\bus.rs:163:9:
assertion failed: addr % 4 == 0
```
I've since then disabled the `firmware-logs` feature and not experienced that panic since.

My guess is that the initial `bp_read` in `log_read` reads some garbage which is why the second one gets a misaligned address, but I'm not too sure.

In any case though I don't think a misaligned address there is an unrecoverable error and justifies panicking. Could we either not panic and log instead or not panic and just try to read the misaligned address?

Dirbaio:
I mean this divider https://github.com/embassy-rs/embassy/blob/38f26137fc67beb874aa73c9a7ab2150d9f3d372/examples/rp/src/bin/wifi_tcp_server.rs#L67

Abestanis:
Ah, I misunderstood, sorry. I'm also using `DEFAULT_CLOCK_DIVIDER`, so I still would assume it would work without problems with this default clock divider.

nui:
I got same issue on Raspberry pi pico 2w when using `DEFAULT_CLOCK_DIVIDER`.

Using `RM2_CLOCK_DIVIDER` work without problems.

i509VCB:
I appear to see this issue immediately after flashing `blinky_wifi` with a Pico 2W. Using `RM2_CLOCK_DIVIDER` does fix it. Should we just set that divider by default so that the examples work (although this does slow stuff down)?

iwanders:
I'm just chiming in with another sample point, Pico 2W without any external components, flashed over USB with `picotool 2.1.1`, compiled with `rustc 1.88.0`.
To get a blinking LED I needed `RM2_CLOCK_DIVIDER`, as well as the `task-arena-size-40960` executor feature that was mentioned [here](https://github.com/embassy-rs/embassy/issues/2274#issuecomment-1868147748). Without that feature it stopped just before [this line](https://github.com/embassy-rs/embassy/blob/9651cfca51a273ba46d34ce8197fc0e63389b09e/examples/rp235x/src/bin/blinky_wifi.rs#L72) (at least, as identified with printing to a usb serial port).

Switching the divider sounds reasonable, with a comment that it may be possible to increase it but the example uses this for better compatibility?
[...]

PR #4414, opened 2025-07-17
RP2350: Fix PIO clock divider in the blinky Wi-Fi example

quarters78:
I was unable to initialize CYW43 chip on my Raspberry Pi Pico 2 W with the `DEFAULT_CLOCK_DIVIDER`. The solution comes from here: https://github.com/embassy-rs/embassy/issues/3612.

### DIFF
--- a/examples/rp235x/src/bin/blinky_wifi.rs
+++ b/examples/rp235x/src/bin/blinky_wifi.rs
@@ -5,7 +5,7 @@
 #![no_std]
 #![no_main]
 
-use cyw43_pio::{PioSpi, DEFAULT_CLOCK_DIVIDER};
+use cyw43_pio::{PioSpi, RM2_CLOCK_DIVIDER};
 use defmt::*;
 use embassy_executor::Spawner;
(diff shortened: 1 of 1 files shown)


i509VCB:
Also https://github.com/embassy-rs/embassy/issues/3960

examples/rp235x/src/bin/blinky_wifi.rs
```diff
@@ -58,7 +58,7 @@ async fn main(spawner: Spawner) {
     let spi = PioSpi::new(
         &mut pio.common,
         pio.sm0,
-        DEFAULT_CLOCK_DIVIDER,
+        RM2_CLOCK_DIVIDER,
```
i509VCB, line 63: Might be a good idea to explain why we are using the RM2 divider on the Pico 2W in a comment.

Could also link to https://github.com/embassy-rs/embassy/issues/3960 in that note.

quarters78: Added a comment with the explanation and the link. Thanks!

PR #4415, opened 2025-07-18
embassy-sync fix for std

AlexCharlton:
Enable the critical-section std feature when the embassy-sync std feature is selected. 

Without this fix, embassy-sync will not compile on std, due to missing (critical-section) symbols.


Dirbaio:
Thanks for the PR :) 

Unfortunately this is intentional, see #4048 #4051 #4052

To fix it you have to add `critical-section = { version = "1", features = ["std"] }` to your Cargo.toml instead.

AlexCharlton:
Ah ha, I see. Sorry for the noise!

PR #4417, opened 2025-07-18
feat(embassy-nrf): add uicr hfxo workaround

nponsard:
In the nRF MDK, there is a workaround "UICR_HFXO_WORKAROUND"  that resests the HFXO values if they are erased. This is necessary on my nrf9151 to have the modem working properly.

I found this thanks to these two messages:
https://github.com/diondokter/nrf-modem/issues/32#issuecomment-2704598018  
https://devzone.nordicsemi.com/f/nordic-q-a/96093/nrf9160-porting-the-modem-library-to-work-with-bare-metal-application/435351?focus=true

I'm not sure I placed this workaround in the correct place, or if this should be handled by embassy-nrf.

I tested this on the nrf9151 of a Thingy91 X, without this workaround the modem gets stuck in initialization, with the workaround it works as intended.

### DIFF
--- a/embassy-nrf/src/lib.rs
+++ b/embassy-nrf/src/lib.rs
@@ -584,6 +584,8 @@ pub mod config {
 #[allow(unused)]
 mod consts {
     pub const UICR_APPROTECT: *mut u32 = 0x00FF8000 as *mut u32;
+    pub const UICR_HFXOSRC: *mut u32 = 0x00FF801C as *mut u32;
+    pub const UICR_HFXOCNT: *mut u32 = 0x00FF8020 as *mut u32;
     pub const UICR_SECUREAPPROTECT: *mut u32 = 0x00FF802C as *mut u32;
     pub const APPROTECT_ENABLED: u32 = 0x0000_0000;
     #[cfg(feature = "_nrf9120")]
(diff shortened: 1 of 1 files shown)


embassy-nrf/src/lib.rs
```diff
@@ -674,6 +674,50 @@ pub fn init(config: config::Config) -> Peripherals {
...
+
+        if hfxocnt == 255 && hfxosrc == 1 {
+            let irq_disabled = cortex_m::register::primask::read().is_inactive();
+            if !irq_disabled {
+                cortex_m::interrupt::disable();
+            }
+            cortex_m::asm::dsb();
+            while !nvmc.ready().read().ready() {}
+
+            nvmc.config().write(|w| w.set_wen(pac::nvmc::vals::Wen::WEN));
```
Dirbaio: could you use `uicr_write()` instead of manually doing it?

Also, the disabling irqs + dsb is a workaround for [this errata](https://docs.nordicsemi.com/bundle/errata_nRF9151_Rev2/page/ERR/nRF9151/Rev2/latest/anomaly_151_7.html)
- it should probably go in uicr_write so that it applies to all uicr writes, not just this one.
- the disable+enable should use `cortex_m::interrupt::free()` instead of reimplementing it from scratch.

nponsard: I changed it to use uicr_write() and implemented the errata in it, should this errata fix be behind a compile-time condition ?

PR #4427, opened 2025-07-20
stm32: Use `unsafe` block in IRQ handler

clubby789:
When developing against `embassy-stm32` locally with a cargo path dependency, this produces a slightly annoying `unsafe_op_in_unsafe_fn` warning - add an unsafe block to resolve this.


i509VCB:
Could you also copy this change to the other embassy hals? (Unfortunately due to $crate, we need to copy paste this macro).

i509VCB:
Probably flaky cyw43?

PR #4429, opened 2025-07-21
Updates to support stm32c071

obe1line:
Modifications required to support the STMC071 as used in the Nucleo-C071RB development board.
Untested changes, but building locally


embassy-stm32/Cargo.toml
```diff
@@ -237,6 +237,47 @@ stm32c031g4 = [ "stm32-metapac/stm32c031g4" ]
...
+stm32c092cc = [ "stm32-metapac/stm32c092cc" ]
+stm32c092ec = [ "stm32-metapac/stm32c092ec" ]
+stm32c092fb = [ "stm32-metapac/stm32c092fb" ]
+stm32c092fc = [ "stm32-metapac/stm32c092fc" ]
+stm32c092gb = [ "stm32-metapac/stm32c092gb" ]
+stm32c092gc = [ "stm32-metapac/stm32c092gc" ]
+stm32c092kb = [ "stm32-metapac/stm32c092kb" ]
+stm32c092kc = [ "stm32-metapac/stm32c092kc" ]
+stm32c092rb = [ "stm32-metapac/stm32c092rb" ]
+stm32c092rc = [ "stm32-metapac/stm32c092rc" ]
```
Dirbaio, line 280: Can you add a few of the new chips to `ci.sh`?

embassy-stm32/src/rcc/c0.rs
```diff
@@ -186,14 +186,21 @@ pub(crate) unsafe fn init(config: Config) {
         hsiker: hsiker,
         hse: hse,
         rtc: rtc,
-
+        
+        #[cfg(any(stm32c071))]
```
Dirbaio: use `cfg(crs)` instead. HSI48 is always present if and only if CRS is present, so this is more "portable" than hardcoding chip names

embassy-stm32/src/rcc/c0.rs
```diff
@@ -186,14 +186,21 @@ pub(crate) unsafe fn init(config: Config) {
...
         lsi: None,
         lse: None,
     );
 
-    RCC.ccipr()
-        .modify(|w| w.set_adc1sel(stm32_metapac::rcc::vals::Adcsel::SYS));
+    #[cfg(not(any(stm32c071)))]
+    let r = RCC.ccipr();
+    #[cfg(any(stm32c071))]
+    let r = RCC.ccipr1();
```
Dirbaio: ideally this should be fixed in the PAC. Rename CCIPR -> CCIPR1 in rcc_c0.yaml so it's the same in both chips.

obe1line: Change made in the PAC. Does it also require a transform RCC_C0v2.yaml creating?
https://github.com/embassy-rs/stm32-data/pull/650

Dirbaio: > Does it also require a transform RCC_C0v2.yaml creating?

no need, transforms are only used when initially extracting from SVD. Once already extracted we just do changes to the registers yaml directly.

embassy-stm32/src/rcc/c0.rs
obe1line: updated - and set to None

embassy-stm32/Cargo.toml
obe1line: Added and now building without errors.

Dirbaio:
thank you!

PR #4430, opened 2025-07-21
Add STM32F1 AFIO remap

fwolter:
This sets the AFIO_MAPR/AFIO_MAPR2 registers on STM32F1 automatically when configuring a peripheral, which supports remapping.

The AFIO clock is already enabled on all AFIO devices here: https://github.com/embassy-rs/embassy/blob/2cae242f51cd1553ca9deb1cfd73daed606c5476/embassy-stm32/src/gpio.rs#L868

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Modify BufferedUart initialization to take pins before interrupts ([#3983](https://github.com/embassy-rs/embassy/pull/3983))
 - Added a 'single-bank' and a 'dual-bank' feature so chips with configurable flash bank setups are be supported in embassy ([#4125](https://github.com/embassy-rs/embassy/pull/4125))
 - Add automatic setting of remap bits when using alternate DMA channels on STM32F0 and STM32F3 devices ([#3653](https://github.com/embassy-rs/embassy/pull/3653))
+- Add support for STM32F1 GPIO remapping (AFIO) ([#4430](https://github.com/embassy-rs/embassy/pull/4430))
 
 ## 0.2.0 - 2025-01-10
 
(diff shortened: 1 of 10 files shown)


leftger:
@fwolter The build failed. Try running ./ci.sh before making the PR to make sure it passes CI. Personally I comment out line 13 to make the script work on my Mac but don't include that change in my PRs

fwolter:
That was to be expected. I added a "depends" in the description.

Dirbaio:
looking great :rocket: 

there's one missing thing: there's no check for whether the pin combination the user requested is valid or not. For example USART3 has this table:

screenshot-2025-08-04_22-34-45

It's not possible to use TX=PB10 and RX=PD9 at the same time since one requires USART3_REMAP=0 and the other USART3_REMAP=3. However if the user tries it, it'll compile fine, it'll run, but the value of USART3_REMAP will be whatever pin the hal calls `.afio_remap()` last, making the other pin not work.

Ideally this'd be a compile-time error just like the AF checking for the rest of stm32 families, but it'd require fairly complicated where clauses just for one family, so I'd be OK with it being a panic when you create the USART driver. It *has* to error somehow, silently not working is not great.

fwolter:
I implemented compile-time errors and AFIO value resolution by type inference. I think it's not too complicated. 

However, timers have ambiguities for instance when created only with one channel. There can be multiple remap values leading to the same result in that case. Therefore, I implemented runtime resolution for all timers. It takes an additional 44 bytes of code size for a timer configured with one channel.

The tests are primarily supposed to show that the code compiles, i.e., all AFIO values can be inferred and there are no type ambiguities.

Dirbaio:
this is great, thanks for all your work :rocket: 

I'm surprised the compiler can infer `A` like that, I thought getting it to infer the right remap mode was going to be impossible, huh.

I've done a few changes:

- Changed timers to do type inference, no remap at runtime. Reasons are: i'm not a fan of the extra complexity, IMO it's better if all peripherals work the same way, for maintainability and for the user to learn the API. It's not that bad that you have to specify the remap manually in cases of ambiguity.
- Cfg'd out `af_num()` on AFIO chips.
- Cfg'd out the `A` generic param on non-AFIO chips.

fwolter:
Thanks for picking this up! Turned out a pretty nice solution, I think! I wasn't aware cfg attributes and macros can be used inline and even inside generics.

PR #4433, opened 2025-07-21
[embassy-time] Enable arch-std when std is selected

diondokter:
We've got a library that uses embassy-time. To run the tests, we enable the std feature of it.

On linux this works fine, but on windows this causes a linker error:
```
libembassy_executor-e28bd8c8eaa8888b.rlib(embassy_executor-e28bd8c8eaa8888b.embassy_executor.52b44d93bbaf442c-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol __pender referenced in function _ZN16embassy_executor3raw6Pender4pend17h826e692e54cdf92eE
```

I've found that windows is little more strict about linking things. The pender is also missing on linux, but for some reason the linker doesn't mind there.

This PR activates the arch-std feature on the imported embassy-executor. This solves the issue.

A reason not to merge this is if we expect there are people who use embassy-time with the std feature activated, but then wants a non-std executor arch. I wouldn't know who would do that...

### DIFF
--- a/embassy-time-queue-utils/Cargo.toml
+++ b/embassy-time-queue-utils/Cargo.toml
@@ -52,6 +52,8 @@ generic-queue-128 = ["_generic-queue"]
 
 _generic-queue = []
 
+std = ["embassy-executor/arch-std"]
+
 [package.metadata.embassy_docs]
 src_base = "https://github.com/embassy-rs/embassy/blob/embassy-time-queue-utils-v$VERSION/embassy-time-queue-utils/src/"
 src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-time-queue-utils/src/"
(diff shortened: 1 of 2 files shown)


Dirbaio:
this is intentional, see #4415  #4048 #4051 #4052

diondokter:
Ah hmmm... Then I'll have to fix it in my project.
It's just a little meh to import embassy-executor only to get it to compile on windows...

Maybe this could be fixed in the executor? When the `_arch` feature is not active, don't call the pender anywhere?

Dirbaio:
this is a slightly different case though. I know esp-idf has (or would want to have) a custom executor arch using task notifications. Not sure if it still might want to use use the std embassy-time driver? Not sure. 

> When the _arch feature is not active, don't call the pender anywhere?

`_arch` not being active means the user is using the raw executor directly and supplying the pender themselves, we still have to call it.

diondokter:
Hmmm, ok, then this is just a sucky situation. I'll solve it at my end then.
Thanks for the responses!

PR #4435, opened 2025-07-21
MSPM0: Add I2C Controller (blocking & async) + examples for mspm0l1306, mspm0g3507 (tested MCUs)

bespsm:
- lib: add i2c mod to lib
- lib: add `bind_interrupts` macros for async workflow
- config: add I2C SDA,SCA pin traits code generation
- config: add clock source for the I2C
- config: add clock divider for the I2C
- config: add I2C BusSpeed configuration
- I2C: add blocking API: blocking_read, blocking_write, blocking_write_read
- I2C: add async API: async_write, async_read, async_write_read
- I2C: add embedded-hal (v0.2) API for blocking & async impl.
- I2C: add embedded-hal (v1.0) API for blocking & async impl.
- I2C-tests: checks for timer_period & check_clock_rate fn's
- mspm0l1306 examples: add I2C blocking & async examples
- mspm0l1306 examples: add -O2 optimization due to Flash limitations
- mspm0g3507 examples: add I2C blocking & async examples

Looking forward for a feedback: @i509VCB


embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            clock_source: ClockSel::BusClk,
+            clock_div: ClockDiv::DivBy1,
+            invert_sda: false,
+            invert_scl: false,
+            sda_pull: Pull::None,
+            scl_pull: Pull::None,
+            bus_speed: BusSpeed::FastMode,
```
i509VCB: I'd tend to be conservative here and default to standard mode.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+}
+
+impl Config {
+    pub fn sda_pf(&self) -> PfType {
+        PfType::input(self.sda_pull, self.invert_sda)
+    }
+    pub fn scl_pf(&self) -> PfType {
+        PfType::input(self.scl_pull, self.invert_scl)
+    }
+    fn timer_period(&self, clock_speed: u32) -> u8 {
```
i509VCB: I need to give the datasheet a read and see if this is possible with fixed point math.

Sure calculations could be done with a floats, but that's going to result in rustc pulling in softfloat and put quite a bit more code into flash. Some parts like C110x might not be able to cope.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+
+    /// Fast mode plus.
+    ///
+    /// The fast mode plus runs at 1 MHz.
+    FastModePlus,
+
+    /// Custom mode.
+    ///
+    /// The custom mode frequency (in Hz) can be set manually.
+    Custom(u32),
```
i509VCB, line 108: Use Hertz type.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+use crate::Peri;
+
+/// The clock source for the I2C.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ClockSel {
+    /// Use the bus clock.
+    ///
+    /// By default the BusClk runs at 32 MHz.
+    BusClk,
```
i509VCB, line 28: 32MHz is probably correct on L130x? But G350x iirc has a PLL drive the bus clock?

I do want to support using the bus clock in the future. Although given the user could set the PLL in G350x to practically anything, maybe the BusClk variant would also pass the clock rate in Hertz? 

Before we do that we need to have a init do the proper clock configuration. For now we can  just support MfClk.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+
+    /// Use the middle frequency clock.
+    ///
+    /// The MCLK runs at 4 MHz.
+    MfClk,
+    // BusClk,
+    // BusClk depends on the timer's power domain.
```
i509VCB: I believe I2C cannot run using LFCLK, so this comment can be removed.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+    DivBy5,
+    // "Divide clock source by 6.
+    DivBy6,
+    // "Divide clock source by 7.
+    DivBy7,
+    // "Divide clock source by 8.
+    DivBy8,
+}
+
+impl Into<vals::Ratio> for ClockDiv {
```
i509VCB: This Into implementation makes the PAC public API (although it is unnameable without unstable-pac feature). Could you use a module private function for the mapping instead?

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+        };
+
+        write!(f, "{}", message)
+    }
+}
+
+impl core::error::Error for Error {}
+
+/// mspm0g, mspm0c, mspm0l, msps00 have 8-bytes FIFO
+pub const FIFO_SIZE: usize = 8;
```
i509VCB: I need to double check the datasheets. Although it should be possible for the metapac to know this information?

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1156 @@
...
+            w.set_agfsel(vals::Agfsel::AGLIT_50);
+            w.set_chain(true);
+        });
+
+        // Reset controller transfer, follow TI example
+        self.info
+            .regs
+            .controller(0)
+            .cctr()
+            .write_value(i2c::regs::Cctr::default());
```
i509VCB: Do reserved bits need to be kept at boot values?

embassy-mspm0/src/lib.rs
```diff
@@ -179,6 +180,10 @@ pub fn init(config: Config) -> Peripherals {
             w.set_mfpclken(true);
         });
 
+        pac::SYSCTL.sysosccfg().modify(|w| {
+            w.set_freq(pac::sysctl::vals::SysosccfgFreq::SYSOSCBASE);
+        });
```
i509VCB: This is going to be chip dependent, and should be something the user specifies in the `Config`. We can manage this in a later pull request.

examples/mspm0g3507/src/bin/i2c.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#![no_std]
+#![no_main]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_mspm0::i2c::{BusSpeed, ClockSel, Config, I2c};
+use {defmt_rtt as _, panic_halt as _};
+
+const ADDRESS: u8 = 0x6a;
```
i509VCB, line 15: What device is this by chance? (I don't think the BSL connection for the launchpad boards since the XDS110 runs its own I2C in controller mode).

embassy-mspm0/src/i2c.rs
bespsm: The formula of calculating time period (TPR) is in the chapter "Clock Select and I2C Speed".
From there: "TPR is the programmed value of the TPR bits in the I2Cx.MTPR register", so by design it should be "u8"

embassy-mspm0/src/i2c.rs
Iooon: Yeah, I also think that we maybe have another location that handles the BusClk configuration and where we could read the frequency from. I think if we remove the BusClk for now we cannot support some I2C modes (FastMode). Is it preferred to reduce to MfClk for now and maybe some modes are then not supported for now?

embassy-mspm0/src/i2c.rs
Iooon: Yeah, we probably can parse it from the header files:
image

Iooon: And we need it since MSPM0C110x is different.
image

embassy-mspm0/src/i2c.rs
bespsm: Sorry, didn't get the comment. I'll try to come up with  fixed point implementation

embassy-mspm0/src/i2c.rs
bespsm: there is a Hertz type but it's not in `embassy-time`  yet, so I'll add M0-internal one, okay?

embassy-mspm0/src/i2c.rs
bespsm: That's why I added TODOs to `calculate_clock_rate` function. So it could be extended in the future. I can also create an issue about it, to don't forget to track it.

As @Iooon said, with MfClk we allow only a `Standard Mode`, which is not much

examples/mspm0g3507/src/bin/i2c.rs
bespsm: this is arduino board address, that I was using for the testing. Feel free to propose another one

embassy-mspm0/src/i2c.rs
bespsm: done

embassy-mspm0/src/i2c.rs
bespsm: done

embassy-mspm0/src/i2c.rs
bespsm: added Herz type

embassy-mspm0/src/i2c.rs
bespsm: removed

embassy-mspm0/src/i2c.rs
i509VCB: Yep feel free to open an issue about clock config. I'll tag that issue when you do so.

embassy-mspm0/src/i2c.rs
bespsm: @Iooon weird about MSPM0C110x. In the reference (slau893b) it's written about I2C: `Independent 8-byte FIFOs for reception and transmission`. 
I check the rest of M0 chips, there are more that have 4 bytes: mspm0c1105, mspm0c1106, mspm0c110x, mspm0h321x, mspm0l111x

It's possible have get fifo size for reading and writing dynamically. I will work on it

embassy-mspm0/src/i2c.rs
bespsm: could you rephrase the question please? didn't get it

i509VCB: When you use write, every single bit of the register is initially set to 0 (closure will set things then). If the hardware keeps the reserved bits at a value of 1, then you may get some weird behavior, maybe not now but potentially in the future.

modify will preserve bits you didn't explicitly edit.

embassy-mspm0/src/lib.rs
bespsm: I see, make sense

embassy-mspm0/src/i2c.rs
bespsm: here you go: https://github.com/embassy-rs/embassy/issues/4445

bespsm: @i509VCB I changed the logic, clock source is defined by analyzing `bus_speed`  field of the config

embassy-mspm0/src/i2c.rs
bespsm: thanks, fixed

embassy-mspm0/src/i2c.rs
Iooon: Hmm yeah, seems they are pretty sure in the reference manual. I thought maybe they talk about the upper limit for the C-series.
We could also check the TI examples how they use the c110x.

But when you try to parse it dynamically, please do it in the msp-data project (in a separate PR). See [here](https://github.com/mspm0-rs/mspm0-data/blob/master/mspm0-data-gen/src/header.rs).

embassy-mspm0/src/i2c.rs
bespsm: moved as private function

embassy-mspm0/src/lib.rs
bespsm: removed

bespsm:
After switching to I2C standard mode, driver start working incorrectly (write-read test of 17 bytes test failed). I will do the research

Iooon:
> After switching to I2C standard mode, driver start working incorrectly (write-read test of 17 bytes test failed). I will do the research

Seems very likely that this is because of the wrong timer period. When executing the tests, a specific case with FastMode seems to work but all other cases fail. Probably the new calculation is a little bit wrong. Make sure you really work periods instead of frequencies (or account for that). That was my error at the beginning.

bespsm:
> > After switching to I2C standard mode, driver start working incorrectly (write-read test of 17 bytes test failed). I will do the research
> 
> Seems very likely that this is because of the wrong timer period. When executing the tests, a specific case with FastMode seems to work but all other cases fail. Probably the new calculation is a little bit wrong. Make sure you really work periods instead of frequencies (or account for that). That was my error at the beginning.

This bug is reproducible even with old timer period calculation, I will dig into it more.

Added: this is expected behavior according TI. Blocking API is capable to transfer max 8 bytes in one transaction. It's explicitly mentioned in the same TI example and it shows the same faulty behavior. Async (with interrupts) API works well with different i2c speeds

embassy-mspm0/src/i2c.rs
bespsm: @Iooon Thanks, I'll follow your advice

i509VCB:
Checking in, I assume the driver is still being worked on? (and whether it is working?)

examples/mspm0g3507/src/bin/i2c.rs
i509VCB: While it would be ideal to use a common device for the example, none of the launchpad boards have a built-in i2c device to play with...

Is this a specific board or is it some internal thing you can't release (if that is the case, we can always change the example in the future to something publicly accessible).

embassy-mspm0/src/i2c.rs
i509VCB: I had actually given other drivers a look and they don't use a Hertz type (embassy-rp). Sorry for the confusion. u32 should suffice then.

bespsm:
> Checking in, I assume the driver is still being worked on? (and whether it is working?)

yep, still working on.


Added:  @i509VCB I guess I addressed all issues

embassy-mspm0/src/i2c.rs
bespsm: fine, reverted

embassy-mspm0/src/i2c.rs
bespsm: I pulled fifo size from "mspm0-data"

examples/mspm0g3507/src/bin/i2c.rs
bespsm: I borrowed i2c example from arduino studio with AD5171 digital potentiometer

i509VCB:
Implementation looks fine. I'll give the errata sheet a look just in case there is something possibly sneaky there we need to know about.

embassy-mspm0/src/i2c.rs
```diff
@@ -0,0 +1,1208 @@
...
+            .await;
+
+            if res.is_err() {
+                self.master_stop();
+                return res;
+            }
+
+            for byte in chunk {
+                *byte = self.info.regs.controller(0).crxdata().read().value();
+            }
```
i509VCB, line 819: We should verify this does not cause problems with `I2C_ERR_08` errata.

i509VCB:
Giving the errata a check:
- I2C_ERR_01 - SMBUS related issue. We don't implement SMBUS mode yet so we can deal with that later.
- I2C_ERR_02 - Quick command related issue. Only applicable in target mode, so ignore it.
- I2C_ERR_03 - I2C peripheral mode cannot wake up device when sourced from MFCLK. Maybe something to keep in mind when we allow entering STOP/STANDBY. Ignore for now.
- I2C_ERR_04  - Only relevant in target mode
- I2C_ERR_05 - We don't touch active bit during transmission, not applicable.
- I2C_ERR_06 - SMBus timeout again, so ignore.
- I2C_ERR_07 - We do single CCTR writes. So this is worked around indirectly.
- I2C_ERR_08 - Maybe an issue, see comment
- I2C_ERR_09 - Target mode again
- I2C_ERR_10 - Low power, doesn't need to be done yet.

embassy-mspm0/src/i2c.rs
bespsm: @i509VCB  I couldn't add errata to the project due to compiling issues. could you point, what exactly should I verify here?

i509VCB: https://www.ti.com/lit/pdf/slaz742, `I2C_ERR_08` states:
> FIFO Read directly after RXDONE interrupt will cause erroneous data to be read

The workaround listed is:
> Wait 2 I2C CLK cycles for the FIFO to guarantee to have the latest data. I2C CLK will
based on the CLKSEL register in the I2C registers.

Steps:
1. IRQ handler wakes the tasking which is pending
2. poll_fn closure is run, when RXDONE is set then return from poll_fn.
3. Check a branch and maybe stop the controller transmission.
4. Read from the fifo

The time gap between step 1 and 4 is non-deterministic. To properly address the errata we would need to insert a wait for 2 I2C clock cycles. Although giving that a thought we would need to know the core clock speed to issue the correct number of cycles to delay. But that is going to depend on #4445.

I'll open an issue about that errata, as we will need to deal with it later.

i509VCB:
Flaky tests... Lets hope another run works.

i509VCB:
@Dirbaio when you can please merge. 5340dk tests are flaking.

bespsm:
@Dirbaio kind reminder to check this one

> @Dirbaio when you can please merge. 5340dk tests are flaking.

bespsm:
@i509VCB  @Dirbaio  I see that more PR get merged recently, Could we try again?

bespsm:
> Sorry to throws this in now, but I noticed that the tests in the i2c.rs file don't work. If the implementation seems to be correct we should at least change the tests. Failing tests are: i2c::tests::ti_calculate_timer_period i2c::tests::ti_calculate_timer_period_2 i2c::tests::ti_calculate_timer_period_3 i2c::tests::ti_calculate_timer_period_4

fixed and tested with oscillator

bespsm:
> @i509VCB @Dirbaio I see that more PR get merged recently, Could we try again?

kind reminder for another attempt 🥲

i509VCB:
I'll open a pull request to disable the 5340 tests, as those seem to have broken. I can only use the merge queue myself.

PR #4419, opened 2025-07-19
nxp: Add MIMXRT1011 GPIO and time driver

i509VCB:
The start of #4345, this time the IMXRT parts.

So far a GPIO and time driver.

The PIT driver is a bit unfortunate for the moment with the potential drift. Will see in the future if the general purpose timer can do a better job.

PR #4436, opened 2025-07-22
nxp: Add mimxrt1062 support 

i509VCB:
Depends on #4419

The examples in this case are designed for the IMXRT1060-EVK.

The same chip is used in the Teensy 4.0/1, but that will probably get its own set of examples due to some differences such as the FCB and how it is flashed.


i509VCB:
Pretty much uncontroversial to merge.

PR #4443, opened 2025-07-22
executor: mark unsafe tasks as unsafe

Brezak:
Currently creating a task creating from a unsafe function doesn't make the newly created task unsafe. e.g.
```rust
#[embassy_executor::task]
async unsafe fn oops() {}

let token = oops();
```
compiles even if it shouldn't. Fixing this is a breaking change.


embassy-executor/tests/ui.rs
```diff
@@ -32,4 +32,5 @@ fn ui() {
     t.compile_fail("tests/ui/self.rs");
     t.compile_fail("tests/ui/type_error.rs");
     t.compile_fail("tests/ui/where_clause.rs");
+    t.pass("tests/ui/task_safety_attribute.rs");
```
Dirbaio, line 37: Can you add a test that checks launching an unsafe task with no `unsafe{}` actually fails?

Brezak: That's what https://github.com/embassy-rs/embassy/pull/4443/files#diff-d25105085e1cf9be4acef7145735f6eef7e355cb24b06dc7653e958d6ffb41bdR2 is for. If `not_safe` doesn't end up being unsafe it will trigger the lint.

Dirbaio: oh okay. Nice! 👍

embassy-executor-macros/src/macros/task.rs
```diff
@@ -120,6 +120,18 @@ pub fn run(args: TokenStream, item: TokenStream) -> TokenStream {
     task_inner.vis = syn::Visibility::Inherited;
     task_inner.sig.ident = task_inner_ident.clone();
 
+    // Forcefully mark the inner task as safe.
+    // SAFETY: We only ever call task_inner in functions
+    //         with the same safety preconditions as task_inner
+    task_inner.sig.unsafety = None;
+    let task_body = task_inner.body;
+    task_inner.body = quote! {
+        #[allow(unused_unsafe, reason = "Not all function bodies may require being in an unsafe block")]
```
Dirbaio: this `unsafe{}` will propagate to user code, the user will be able to do unsafe things without explicitly writing `unsafe{}` which is not good. Same for the `#allow(unused_unsafe)`.

I think it might be better to not force `task_inner.sig.unsafety = None`, and add `unsafe{}` in `task_outer_body` in the right place if needed.

Brezak: Problem is that if task inner is made unsafe it can't implement `FnMut` (implementing Fn traits for unsafe functions would let you call them in safe generic code) and therefore can't implement `TaskFn`. I may have a solution. If we wrap the user code in a conditionally safe closure we instantly invoke it should work. I'm just worried about the codegen impact.

Brezak: Update: won't work. There are no unsafe closures. I've also tried throwing the code into a inner inner function and just calling that means we need to await it and create a new future that just calls into the original future. Still have stuff left to try.

lulf:
@Brezak This looks good to me, were you able to try other approaches after your last [comment](https://github.com/embassy-rs/embassy/pull/4443#discussion_r2223796031)?

Brezak:
> @Brezak This looks good to me, were you able to try other approaches after your last [comment](https://github.com/embassy-rs/embassy/pull/4443#discussion_r2223796031)?

Yeah, I did. The only solution that worked is what we see here. I'm not happy about the whole mess with parsing the user's return type and manipulating it, or the slight diagnostic regression, but nothing else worked.

PR #4444, opened 2025-07-22
nxp: make docs generate

i509VCB:


Dirbaio:
you need to add it here as well https://github.com/embassy-rs/embassy/blob/main/.github/ci/doc.sh

maybe it'd be great to have it automatically walk all Cargo.tomls in the repo but it doesn't do that today...

i509VCB:
fixed

PR #4450, opened 2025-07-23
stm32/adc/v3: Add method to sample ADC without reconfiguring every time

clubby789:
This moves the ADC and channel configuration code to a new public function `prepare_read`, then adds a `read_without_preparing` function which just starts a DMA transfer and starts conversion. This reduces wasted time reconfiguring the same channels repeatedly.

Tested on my STM32WB55RG

### DIFF
--- a/embassy-stm32/src/adc/v3.rs
+++ b/embassy-stm32/src/adc/v3.rs
@@ -266,8 +266,8 @@ impl<'d, T: Instance> Adc<'d, T> {
     /// channel number and not the pin order in `sequence`.
     ///
     /// Example
-    /// ```rust,ignore
-    /// use embassy_stm32::adc::{Adc, AdcChannel}
+    /// ```rust,no_run
+    /// use embassy_stm32::adc::{Adc, AdcChannel, SampleTime}
     ///
     /// let mut adc = Adc::new(p.ADC1);
     /// let mut adc_pin0 = p.PA0.degrade_adc();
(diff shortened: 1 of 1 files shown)


clubby789:
By the way, I didn't touch it because I was unsure if there was a good reason, but why is continuous conversion mode used if we stop the ADC after receiving the first conversion? Wouldn't it be simpler to just use single conversion mode?

embassy-stm32/src/adc/v3.rs
```diff
@@ -292,11 +292,112 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+        if channel.channel() == 0 {
+            T::regs().or().modify(|reg| reg.set_op0(true));
+        }
+
+        // Configure channel
+        Self::set_channel_sample_time(channel.channel(), sample_time);
+    }
+
+    /// Configure the ADC peripheral with the given channels for reading with DMA
+    pub async fn prepare_read(&mut self, sequence: impl ExactSizeIterator<Item = (&mut AnyAdcChannel<T>, SampleTime)>) {
```
Jacke-debug, line 399: I suggest to rename this method to configure_regular_sequence to make the distinction between regular and injected more clear.

embassy-stm32/src/adc/v3.rs
```diff
@@ -292,11 +292,112 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+    /// ).await;
+    /// for _ in 0..5 {
+    ///     adc.read_without_preparing(
+    ///         p.DMA1_CH2.reborrow(),
+    ///         &mut measurements,
+    ///     ).await;
+    ///     defmt::info!("measurements: {}", measurements);
+    /// }
+    /// ```
+    pub async fn read_without_preparing(&mut self, rx_dma: Peri<'_, impl RxDma<T>>, readings: &mut [u16]) {
```
Jacke-debug, line 362: I suggest to rename this to something like "trigger_regular_conversion_and_await_result" (maybe a bit long but I think you get the point). 

We may also want to think about how we can add support for non-software triggered ADC measurements in the future by setting the EXTEN bit.

Dirbaio:
Hello,

PR #4459, opened 2025-07-25
feat: add RTC time driver

IrinaCh524:
I implemented the embassy time diver using the RTC peripheral.
For now, we are using the lpc55-pac for RTC. Later that will be switched out for the nxp-pac one.  :)
@george-cosma please review. :D

### DIFF
--- a/embassy-nxp/Cargo.toml
+++ b/embassy-nxp/Cargo.toml
@@ -50,6 +50,9 @@ log = ["dep:log"]
 ## Use Periodic Interrupt Timer (PIT) as the time driver for `embassy-time`, with a tick rate of 1 MHz
 time-driver-pit = ["_time_driver", "embassy-time?/tick-hz-1_000_000"]
 
+## Use Real Time Clock (RTC) as the time driver for `embassy-time`, with a tick rate of 32768 Hz
+time-driver-rtc = ["_time_driver", "embassy-time?/tick-hz-32_768"]
+
 ## Reexport the PAC for the currently enabled chip at `embassy_nxp::pac` (unstable)
 unstable-pac = []
 # This is unstable because semver-minor (non-breaking) releases of embassy-nxp may major-bump (breaking) the PAC version.
--- a/embassy-nxp/src/time_driver/rtc.rs
+++ b/embassy-nxp/src/time_driver/rtc.rs
@@ -1,0 +1,172 @@
+use core::cell::{Cell, RefCell};
+use core::task::Waker;
+
+use critical_section::CriticalSection;
+use embassy_hal_internal::interrupt::{InterruptExt, Priority};
+use embassy_sync::blocking_mutex::CriticalSectionMutex as Mutex;
+use embassy_time_driver::{time_driver_impl, Driver};
+use embassy_time_queue_utils::Queue;
+use lpc55_pac::{interrupt, PMC, RTC, SYSCON};
+struct AlarmState {
+    timestamp: Cell<u64>,
+}
+
+unsafe impl Send for AlarmState {}
+
+impl AlarmState {
+    const fn new() -> Self {
+        Self {
+            timestamp: Cell::new(u64::MAX),
+        }
+    }
+}
+
+pub struct RtcDriver {
+    alarms: Mutex<AlarmState>,
+    queue: Mutex<RefCell<Queue>>,
+}
+
+time_driver_impl!(static DRIVER: RtcDriver = RtcDriver {
+    alarms: Mutex::new(AlarmState::new()),
+    queue: Mutex::new(RefCell::new(Queue::new())),
+});
+impl RtcDriver {
+    fn init(&'static self) {
+        let syscon = unsafe { &*SYSCON::ptr() };
+        let pmc = unsafe { &*PMC::ptr() };
+        let rtc = unsafe { &*RTC::ptr() };
+
+        syscon.ahbclkctrl0.modify(|_, w| w.rtc().enable());
+
+        // By default the RTC enters software reset. If for some reason it is
+        // not in reset, we enter and them promptly leave.q
+        rtc.ctrl.modify(|_, w| w.swreset().set_bit());
+        rtc.ctrl.modify(|_, w| w.swreset().clear_bit());
+
+        // Select clock source - either XTAL or FRO
+        // pmc.rtcosc32k.write(|w| w.sel().xtal32k());
+        pmc.rtcosc32k.write(|w| w.sel().fro32k());
+
+        // Start the RTC peripheral
+        rtc.ctrl.modify(|_, w| w.rtc_osc_pd().power_up());
+
+        // rtc.ctrl.modify(|_, w| w.rtc_en().clear_bit()); // EXTRA
+
+        //reset/clear(?) counter
+        rtc.count.reset();
+        //en rtc main counter
+        rtc.ctrl.modify(|_, w| w.rtc_en().set_bit());
+        rtc.ctrl.modify(|_, w| w.rtc1khz_en().set_bit());
+        // subsec counter enable
+        rtc.ctrl.modify(|_, w| w.rtc_subsec_ena().set_bit());
+
+        // enable irq
+        unsafe {
+            interrupt::RTC.set_priority(Priority::from(3));
+            interrupt::RTC.enable();
+        }
+    }
+
+    fn set_alarm(&self, cs: CriticalSection, timestamp: u64) -> bool {
+        let rtc = unsafe { &*RTC::ptr() };
+        let alarm = &self.alarms.borrow(cs);
+        alarm.timestamp.set(timestamp);
+        let now = self.now();
+
+        if timestamp <= now {
+            alarm.timestamp.set(u64::MAX);
+            return false;
+        }
+
+        //time diff in sub-sec not ticks (32kHz)
+        let diff = timestamp - now;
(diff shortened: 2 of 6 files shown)


embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -0,0 +1,173 @@
...
+        rtc.count.reset();
+        //en rtc main counter
+        rtc.ctrl.modify(|_, w| w.rtc_en().set_bit());
+        rtc.ctrl.modify(|_, w| w.rtc1khz_en().set_bit());
+        // subsec counter enable
+        rtc.ctrl.modify(|_, w| w.rtc_subsec_ena().set_bit());
+
+        let mut cp = cortex_m::peripheral::Peripherals::take().unwrap();
+        unsafe { cp.NVIC.set_priority(interrupt::RTC, 3) };
+        unsafe { cortex_m::peripheral::NVIC::unmask(interrupt::RTC) };
```
i509VCB: Assuming lpc55-pac implements `cortex_m::InterruptNumber` for `RTC`, you should be able to use the functions in `InterruptExt` from `embassy-hal-internal` to unmask interrupts/set priority.

embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -0,0 +1,173 @@
...
+        let rtc = unsafe { &*RTC::ptr() };
+
+        loop {
+            let sec = rtc.count.read().val().bits() as u64;
+            let sub = rtc.subsec.read().subsec().bits() as u64;
+            let total_ticks_1 = sec * 32768 + sub;
+
+            let sec = rtc.count.read().val().bits() as u64;
+            let sub = rtc.subsec.read().subsec().bits() as u64;
+            let total_ticks_2 = sec * 32768 + sub;
```
i509VCB: Do we need to calculate the number of ticks since boot twice in `now()`?

The following should work inside the loop:
```rust
let sec = rtc.count.read().val().bits() as u64;
let sub = rtc.subsec.read().subsec().bits() as u64;
let sec2 = rtc.count.read().val().bits() as u64;

if sec == sec2 {
    return sec * 32768 + sub;
}

// Otherwise the loop will try again
```

This change would only require calculating the number of ticks since boot once instead of twice.

embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -0,0 +1,173 @@
+use core::cell::{Cell, RefCell};
+use core::task::Waker;
+use critical_section::CriticalSection;
+use embassy_sync::blocking_mutex::CriticalSectionMutex as Mutex;
+use embassy_time_driver::time_driver_impl;
+use embassy_time_driver::Driver;
+use embassy_time_queue_utils::Queue;
+use lpc55_pac::{interrupt, NVIC, PMC, RTC, SYSCON};
+struct AlarmState {
```
eva-cosma:
```suggestion
use core::cell::{Cell, RefCell};
use core::task::Waker;

use critical_section::CriticalSection;
use embassy_sync::blocking_mutex::CriticalSectionMutex as Mutex;
use embassy_time_driver::{time_driver_impl, Driver};
use embassy_time_queue_utils::Queue;
use lpc55_pac::{interrupt, PMC, RTC, SYSCON};
struct AlarmState {
```

To pass CI

embassy-nxp/src/time_driver/rtc.rs
eva-cosma: I think we do. `sub` is more likely to be changed compared to `sec` or `sec2`, since that register tracks the time between the seconds. IMHO, it's not enough to check only one or the other. The sub-second counter and second counters must match. We might be able to change the check to not calculate the total ticks until after we verify

```rust
let sec = ...;
let sub = ...;

let sec2 = ...;
let sub2 = ...;

if sec == sec2 && sub == sub2 {
    return sec * 32768 + sub;
}
```

embassy-nxp/src/time_driver/rtc.rs
eva-cosma: Macro: https://github.com/embassy-rs/embassy/blob/main/embassy-hal-internal/src/interrupt.rs
Example Usage in `embassy-rp` (1): https://github.com/embassy-rs/embassy/blob/c45b6d4a5109330b6980b0a6843b06c65c8ca04c/embassy-rp/src/lib.rs#L106
Example Usage in `embassy-rp` (2): https://github.com/embassy-rs/embassy/blob/c45b6d4a5109330b6980b0a6843b06c65c8ca04c/embassy-rp/src/time_driver.rs#L126

embassy-nxp/Cargo.toml
```diff
@@ -50,6 +50,8 @@ log = ["dep:log"]
 ## Use Periodic Interrupt Timer (PIT) as the time driver for `embassy-time`, with a tick rate of 1 MHz
 time-driver-pit = ["_time_driver", "embassy-time?/tick-hz-1_000_000"]
 
+time-driver-rtc = ["_time_driver", "embassy-time?/tick-hz-32_768"]
```
eva-cosma, line 54: nitpick: Add a comment similar to the above `time-driver-pit` to explain this feature

embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -0,0 +1,173 @@
...
+            if next == u64::MAX {
+                //no next event found after retry
+                return;
+            }
+        }
+    }
+}
+
+impl Driver for RtcDriver {
+    //read time:sec+wake count in ms
```
eva-cosma: Is it in miliseconds? I think each tick is 1/32768 of a second for this peripheral

embassy-nxp/src/time_driver/rtc.rs
i509VCB: Giving it a thought both of our suggestions have the same behavior. One just contains more instructions.

Assume it's 0 ticks since boot and we get a call to now(). Sec is going to be 0. But between loading sec and loading sub the thread of execution is preempted. Execution does not return until 1000 ticks since boot (sub did not carry over). If Timer::after is 1000 ticks then your await is now 1000 ticks late. 

In the case of my suggestion you would still have the alarm trigger at tick 2000. Yours would likely also be tick 2000, but could be tick 2001 if you happened to need another timer tick to read the new values. 

I mention this behavior because it might seem a little underdefined and undocumented. I wonder if @Dirbaio could clarify whether that is okay behavior or is intended (`now()` can't have zero overhead because mmio must happen). 

Scheduling a wake will take a critical section though, so that will prevent preemption but you need to compute the timestamp at which to wake first. 

The nuclear option is to always take a critical section inside of `now()` but that doesn't solve the problem. In the case of lpc55 the hardware doesn't latch sub when sec is read so you would have to do the comparison of sec and sec2 anyways (rtc runs asynchronously). And Timer::after_xyz needs to branch to embassy_time_now and then RtcDriver::now(). So you could still encounter a late timestamp calculation.

Anyways I'll wait for Dario before I claim I have the right answer.

embassy-nxp/src/time_driver/rtc.rs
i509VCB: Definitely not milliseconds. Because the feature will enable tick-hz-32_768 if embassy-time is present in the dependency graph. That would be ticks. `Timer::after(Duration)` and friends will convert from a time unit to ticks.

embassy-nxp/src/time_driver/rtc.rs
eva-cosma: I think I understand why we saw things differently.

Let's take the simplified code
```rust
let sec = ...;
// [P1]
let sub = ...;
// [P2]

return sec * 32768 + sub;
```

Your concern is that if an interrupt were to happen at `[P1]`, and then return, the value of `sec` could have been modified (it was just about to switch over, but the interrupt pushed it over the edge). In this case, I agree, re-checking only sec makes sense.

My concern was an interrupt happening at `[P2]`, not returning the actual `now`, but a delayed `now`. However, now that I think about it, my solution wouldn't solve this. More so that an interrupt can still happen between the total tick calculation and the function return. Putting this behind a critical section seems overkill though only way to solve this.

I'm not as familiar to the constraints of the `now` function -- how strict do we need to be?

embassy-nxp/Cargo.toml
eva-cosma: No idea why this isn't showing up as resolved. but it is resolved.

embassy-nxp/src/time_driver/rtc.rs
eva-cosma: @i509VCB I am more than happy to accept your code suggestion and move this line of questioning into an issue, as it is not super-critical.

eva-cosma:
@i509VCB @Dirbaio Please re-review

embassy-nxp/src/time_driver/rtc.rs
i509VCB: Given Dario is quite busy, I'll leave the further debate to later. But yes I would like an issue to make sure this doesn't all off the edge of the earth.

i509VCB (approved):
Just need to resolve merge conflicts. The discussion about now() we will address at a later date.

embassy-nxp/src/time_driver/rtc.rs
eva-cosma: #4498

i509VCB (approved):

PR #4460, opened 2025-07-25
feat: add divs/divt for stm32h7rs

rickrogers-ionq:
* stm32h7rs includes 2 additional pll divisors: divs and divt
* these changes make those divisors available in embassy-stm32/src/rcc/h.rs


Dirbaio:
lgtm! though just adding it to the struct is not enough, you'll have to update `init_pll()`

rickrogers-ionq:
thanks @Dirbaio !  sorry I missed the init fn.  I'll look into that now.

rickrogers-ionq:
I had the init mods in another local working tree and forgot to pull them in.  should be good now.

PR #4463, opened 2025-07-26
Added support for PLL as a clock source on STM32WBA

leftger:
The RCC HAL for WBA devices had support for HSE and HSI but not for the PLL. That support has been added due to the recent additions in `stm32-data` that added the PLL enums. With that in place a new addition is the inclusion of sanity checks for the PLL VCO generated frequency (it'll panic just like in the U5 HAL). That aside there's also the addition of the PLLFRACN fractional multiplier. That has also been added. The frequency asserts also take this PLLFRACN into account.

A timer example has also been added showcasing this new support for the PLL.

### DIFF
--- a/embassy-stm32/src/rcc/wba.rs
+++ b/embassy-stm32/src/rcc/wba.rs
@@ -1,19 +1,17 @@
 pub use crate::pac::pwr::vals::Vos as VoltageScale;
 use crate::pac::rcc::regs::Cfgr1;
-pub use crate::pac::rcc::vals::{
-    Hpre as AHBPrescaler, Hsepre as HsePrescaler, Ppre as APBPrescaler, Sw as Sysclk, Pllsrc as PllSource,
-    Plldiv as PllDiv, Pllm as PllPreDiv, Plln as PllMul, Hpre5 as AHB5Prescaler, Hdiv5,
-};
+#[cfg(all(peri_usb_otg_hs))]
+pub use crate::pac::rcc::vals::Otghssel;
 use crate::pac::rcc::vals::Pllrge;
+pub use crate::pac::rcc::vals::{
+    Hdiv5, Hpre as AHBPrescaler, Hpre5 as AHB5Prescaler, Hsepre as HsePrescaler, Plldiv as PllDiv, Pllm as PllPreDiv,
+    Plln as PllMul, Pllsrc as PllSource, Ppre as APBPrescaler, Sw as Sysclk,
+};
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/rcc/wba.rs
```diff
@@ -1,35 +1,102 @@
...
+#[cfg(all(peri_usb_otg_hs))]
+pub use crate::pac::{syscfg::vals::Usbrefcksel, SYSCFG};
+
 /// HSI speed
 pub const HSI_FREQ: Hertz = Hertz(16_000_000);
 // HSE speed
 pub const HSE_FREQ: Hertz = Hertz(32_000_000);
 
+// Allow dividing a Hertz value by an AHB5 prescaler directly
+impl Div<AHB5Prescaler> for Hertz {
```
Dirbaio: These impls should be autogenerated. You can do it by extending this so it matches "hpre5" https://github.com/embassy-rs/embassy/blob/996974e313fa5ec2c7c2d9dd0998fab244c0a180/embassy-stm32/build.rs#L1600

embassy-stm32/src/rcc/wba.rs
```diff
@@ -40,14 +107,17 @@ pub struct Config {
...
+            hse: None,
+            pll1: None,
             sys: Sysclk::HSI,
             ahb_pre: AHBPrescaler::DIV1,
+            ahb5_pre: AHB5Prescaler::DIV1,
             apb1_pre: APBPrescaler::DIV1,
             apb2_pre: APBPrescaler::DIV1,
             apb7_pre: APBPrescaler::DIV1,
-            ls: crate::rcc::LsConfig::new(),
+            lsi: crate::rcc::LsConfig::new(),
```
Dirbaio: I don't think we should rename this, "LS" applies to all low-speed clocks (LSI, LSE), not just LSI.

embassy-stm32/src/rcc/wba.rs
leftger: Thanks for that insight. I have taken this out.

embassy-stm32/src/rcc/wba.rs
leftger: That makes sense. I didn't fully grasp the naming reasoning before.

Dirbaio:
lgtm, can you fix ci?

leftger:
Sorry I was working on my dev board (WBA65RI). I fixed it to work on the current example config (WBA55CG)

PR #4475, opened 2025-07-28
Make it possible to do configuration changes prior to enabling the ADC.

snakehand:
Some configurations can only be done prior to enabling the ADC. Make it possible to provide a closure for the new() method that performs such initialisation.

See RM0456 33.7.2 ADC system control register (ADC12_CCR)

### DIFF
--- a/embassy-stm32/src/adc/v4.rs
+++ b/embassy-stm32/src/adc/v4.rs
@@ -159,7 +159,7 @@ pub enum Averaging {
 
 impl<'d, T: Instance> Adc<'d, T> {
     /// Create a new ADC driver.
-    pub fn new(adc: Peri<'d, T>) -> Self {
+    pub fn new_with(adc: Peri<'d, T>, init: impl FnOnce(&mut Self)) -> Self {
         rcc::enable_and_reset::<T>();
 
         let prescaler = Prescaler::from_ker_ck(T::frequency());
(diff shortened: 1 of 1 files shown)


Dirbaio:
this is a bit of a workaround, the user shouldn't have to worry about these details.

Options that would work are:

- Pass these settings as a config struct to new() like many other drivers. This means to change them you have to drop and recreate the ADC.
- Make changing these settings do "disable, change, reenable"
- Do not enable ADC in new(), enable it on the fly when actually using the adc.

snakehand:
@Dirbaio enable was private, so are you suggesting that public enable & disable functions in the API is the most straightforward fix ?

Dirbaio:
No, the user should not need micromanage the enable bit either. By "enable it on the fly when actually using the adc" I meant the *HAL* does it when you do a read.

The HAL is a Hardware *Abstraction* Layer, it should abstract things. As a user you should be able to "create ADC, set some config, do a read" and have it Just Work, without having to worry about low-level details like "some bits can't be set while enabled" or "you must enable before doing a read". The HAL should take care of the proper order of operations.

snakehand:
@Dirbaio the problem with such a design is that the affected settings ( Enable vrefint, temperature, vbat ) are in the common registers, and can be shared by 2 ADCs, so it could then be possible to initialise 1 ADC without these features, and if the other ADC is attempted to be initialised with these features later, that will not work since the first ADC is enabled and prevents modification. Adding a configuration to new() is also a breaking change, but that is maybe not an issue ?

Dirbaio:
Breaking changes are fine.

The ADC_COMMON is a painful long-standing issue. I think the way ADC drivers are modeled today is wrong, we should have  an "ADC group" driver that owns e.g. ADC1, ADC2, ADC3, ADC123_COMMON all at once, which lets you set common settings, split it into parts for each ADCx and then use the parts individually.

snakehand:
I will close this - and take the input into consideration and perhaps come back with a better solution.

PR #4478, opened 2025-07-29
STM32WBA usb-dfu example

leftger:
With the addition of USB_OTG_HS support on the WBA65 devices I took the liberty of adapting the current examples on the STM32WB devices to my Nucleo board. The main differences are:

- The memory.x files for both the application and the bootloader have to account for the larger flash and memory (1/2MM and 256/512KB)
- The new examples run off the HSI (in case the HSE isn't connected to a crystal)
- The examples are proven to work on the STM32WBA65RIVx Nucleotide board.


examples/boot/application/stm32wba-dfu/fw.bin
Dirbaio, line 1: unintentionally added?

leftger: Sorry 🤦‍♂️ . I'll take it out.

PR #4481, opened 2025-07-29
STM32 CAN: Add mutable access for ID in Header and Frame structs

mbrieske:
In communication protocols like CANopen, the message ID is often a combination of a base arbitration ID and a specific node ID. To support a similar mechanism I need to modify the arbitration ID by adding the node ID before the message is transmitted.

### DIFF
--- a/embassy-stm32/src/can/frame.rs
+++ b/embassy-stm32/src/can/frame.rs
@@ -59,6 +59,11 @@ impl Header {
         &self.id
     }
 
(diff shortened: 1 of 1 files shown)


cschuhen:
Seems fine, I guess if you have a use case where you want to create the frame before knowing where to send it.

lulf (approved):
Thanks! And thanks for the review @cschuhen

PR #4488, opened 2025-08-01
Embassy futures 0.2.0

lulf:


bugadani:
I'll put it here, too, to be visible: the breaking change is a soundness fix. IMHO this could be v0.1.2

lulf:
Thanks @bugadani . I also think perhaps a 0.1.2 would be OK for this.

HaoboGu:
It breaks user code, an extra `pin!` is required after updating, so I do think it's a breaking change.

lulf:
> It breaks user code, an extra `pin!` is required after updating, so I do think it's a breaking change.

Yeah, I think theres different opinions on how strict this breakage is in the event of a bug/bad API or for instance a security issue. You could argue that it's OK to break semver in those cases.

PR #4490, opened 2025-08-01
embassy-sync: Update `MultiWakerRegistration::register` docs

Brezak:
In 3081ecf301a54f8ed3d0f72350dd21f8ac9e1b18 `register` was changed to clear the buffer when it's full, but the docs weren't updated.


lulf (approved):
Thanks!

PR #4496, opened 2025-08-02
feat(embassy-rp): RP2350 - Add support for QMI CS1, and for APS6404L PSRAM

magnusnordlander:
This pull request adds support for the QMI CS1 peripheral, and then uses that support to implement further support for the APS6404L PSRAM as QMI Memory 1.

I have tried to keep the code as general as foreseeable for other forms of PSRAM. The code is presumably general enough to handle using a second serial NOR flash instead of a PSRAM, but I imagine a PSRAM is a much more common use case.

### DIFF
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -43,6 +43,10 @@ pub mod rtc;
 pub mod spi;
 mod spinlock;
 pub mod spinlock_mutex;
+#[cfg(feature = "_rp235x")]
+pub mod qmi_cs1;
+#[cfg(feature = "_rp235x")]
+pub mod psram;
 #[cfg(feature = "time-driver")]
 pub mod time_driver;
 #[cfg(feature = "_rp235x")]
@@ -380,6 +384,8 @@ embassy_hal_internal::peripherals! {
 
     SPI0,
     SPI1,
+
+    QMI_CS1,
 
     I2C0,
     I2C1,
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -35,7 +35,11 @@ pub mod multicore;
 #[cfg(feature = "_rp235x")]
 pub mod otp;
 pub mod pio_programs;
+#[cfg(feature = "_rp235x")]
+pub mod psram;
 pub mod pwm;
+#[cfg(feature = "_rp235x")]
+pub mod qmi_cs1;
 mod reset;
 pub mod rom_data;
 #[cfg(feature = "rp2040")]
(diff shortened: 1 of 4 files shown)


lulf:
Thanks for the PR! I'm not sure if this makes sense, but I'm wondering if integrating a specific psram driver in embassy-rp is a scalable way to do it. Generally we try to avoid having drivers that should'nt be HAL-specific in the tree. 

Could it make sense to move the aps6404l driver into a separate crate, and expose the necessary APIs in embassy-rp publicly for it to be able to use it? Taking this a step further, maybe there is some generic way that the aps6404l could work with any HAL that implements a certain set of traits in the future?

magnusnordlander:
So I'll answer the easy question first:

> Taking this a step further, maybe there is some generic way that the aps6404l could work with any HAL that implements a certain set of traits in the future?

Probably not. *Most* of what this PR does is just configure the RP2350 QMI peripheral. Other MCUs may have QMI peripherals, but that doesn't mean they share configuration options in a meaningful way. It could possibly be used with other HALs for the RP2350 specifically, but typically HALs don't share configuration structs.

Even the APS6404L specific verification function uses the QMI peripheral to directly communicate with the APS6404L, but it's essentially just sending a single command and reading a response. It's just that doing it over the QMI interface using Direct CSR is very convoluted. *That* could absolutely be shared with other RP2350 HALs, but that's because it's all inline assembly. 

> Could it make sense to move the aps6404l driver into a separate crate, and expose the necessary APIs in embassy-rp publicly for it to be able to use it?

It could. Most of this PR is already just general QMI configuration code. The APS6404L specific code comes in two categories: Configuration settings for the APS6404L (which absolutely could be moved) - and the verification routine, which could be moved too. 

There is a third part that *might* be APS6404L specific, and that's the `enter_quad_cmd`, which (to the APS6404L at least) needs to be sent using regular SPI, not QSPI. What command to send is just configuration (and it's an Option, so if your IC doesn't need an enter_quad_cmd, you can just pass None), but of course, there *may* be other PSRAM ICs that *could* work with the RP2350 QMI controller, but require some other initialization routine. Sending that command could also be moved to a separate crate (since it's essentially just inline assembly), but other PSRAM ICs may or may not need something similar.

It is however worth noting that at least right now, the APS6404L is (to my knowledge) the *only* PSRAM IC compatible with the RP2350, so it's kind of difficult to know what other PSRAM ICs would or wouldn't require.

If this is more just my reasoning: *Let me know if you'd like I can split off the APS6404L specific code into a separate crate. It's not a big problem to do it.*

CBJamo:
> It is however worth noting that at least right now, the APS6404L is (to my knowledge) the only PSRAM IC compatible with the RP2350, so it's kind of difficult to know what other PSRAM ICs would or wouldn't require.

Why would this be true? The 2350's QMI appears to be fairly flexible, such that any (Q/D)SPI psram should work.

magnusnordlander:
> Why would this be true? The 2350's QMI appears to be fairly flexible, such that any (Q/D)SPI psram should work.

It is fairly flexible, there just aren't that many SPI PSRAM ICs out there. That being said, after doing some research I did find a couple of other ICs that also should work. 

I'll do a refactor to make this a more general QMI peripheral, and then release a separate crate with the APS6404L specific code.

PR #4497, opened 2025-08-02
feat: lpc55 blocking usart added

frihetselsker:
Hello! For LPC55 USART with blocking mode was implemented.

- Implemented structure `Usart` which can be split into `UsartRx` and `UsartTx`
- Implemented `Blocking` mode
- Added `Config` and `Error`
- All the sufficient IO-functions for `Blocking` mode were implemented
- `Usart` was made generic for the sake of flexibility
- All the interactions with PAC are hidden macro called `impl_instance!()`
- Added USART peripherals for LPC55
- Added example `usart_blocking.rs` for demonstration

@george-cosma, please review this PR

### DIFF
--- a/embassy-nxp/Cargo.toml
+++ b/embassy-nxp/Cargo.toml
@@ -24,7 +24,8 @@ log = { version = "0.4.27", optional = true }
 embassy-time = { version = "0.4.0", path = "../embassy-time", optional = true }
 embassy-time-driver = { version = "0.2", path = "../embassy-time-driver", optional = true }
 embassy-time-queue-utils = { version = "0.2", path = "../embassy-time-queue-utils", optional = true }
-
+embedded-io = "0.6.1"
+embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["unproven"] }
 ## Chip dependencies
 lpc55-pac = { version = "0.5.0", optional = true }
 nxp-pac = { version = "0.1.0", optional = true, git = "https://github.com/i509VCB/nxp-pac", rev = "be4dd0936c20d5897364a381b1d95a99514c1e7e" }
--- a/embassy-nxp/src/usart.rs
+++ b/embassy-nxp/src/usart.rs
@@ -1,0 +1,6 @@
+//! Universal Synchronous/Asynchronous Receiver/Transmitter (USART) driver.
+#![macro_use]
+
+#[cfg_attr(feature = "lpc55", path = "./usart/lpc55.rs")]
+mod inner;
+pub use inner::*;
--- a/embassy-nxp/src/usart/lpc55.rs
+++ b/embassy-nxp/src/usart/lpc55.rs
@@ -1,0 +1,885 @@
+use core::marker::PhantomData;
+
+use embassy_hal_internal::{Peri, PeripheralType};
+use embedded_io::{self, ErrorKind};
+pub use sealed::SealedInstance;
+
+use crate::gpio::AnyPin;
+use crate::{Blocking, Mode};
+
+/// Serial error
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum Error {
+    /// Triggered when the FIFO (or shift-register) is overflowed.
+    Overrun,
+    /// Triggered when there is a parity mismatch between what's received and
+    /// our settings.
+    Parity,
+    /// Triggered when the received character didn't have a valid stop bit.
+    Framing,
+    /// Triggered when the receiver detects noise
+    Noise,
+    /// Triggered when the receiver gets a break
+    Break,
+}
+
+impl embedded_io::Error for Error {
+    fn kind(&self) -> ErrorKind {
+        match self {
+            Error::Overrun => ErrorKind::Other,
+            Error::Parity => ErrorKind::InvalidData,
+            Error::Framing => ErrorKind::InvalidData,
+            Error::Noise => ErrorKind::Other,
+            Error::Break => ErrorKind::Interrupted,
+        }
+    }
+}
+/// Word length.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum DataBits {
+    /// 7 bits data length.
+    DataBits7,
+    /// 8 bits data length.
+    DataBits8,
+    /// 9 bits data length. The 9th bit is commonly used for addressing in multidrop mode.
+    DataBits9,
+}
+
+impl DataBits {
+    fn bits(&self) -> u8 {
+        match self {
+            Self::DataBits7 => 0b00,
+            Self::DataBits8 => 0b01,
+            Self::DataBits9 => 0b10,
+        }
+    }
+}
+
+/// Parity bit.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Parity {
+    /// No parity.
+    ParityNone,
+    /// Even parity.
+    ParityEven,
+    /// Odd parity.
+    ParityOdd,
+}
+
+impl Parity {
+    fn bits(&self) -> u8 {
+        match self {
+            Self::ParityNone => 0b00,
+            Self::ParityEven => 0b10,
+            Self::ParityOdd => 0b11,
+        }
+    }
+}
+
+/// Stop bits.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum StopBits {
+    /// 1 stop bit.
+    Stop1,
+    /// 2 stop bits. This setting should only be used for asynchronous communication.
+    Stop2,
+}
+
+impl StopBits {
+    fn bits(&self) -> bool {
+        return match self {
+            Self::Stop1 => false,
+            Self::Stop2 => true,
+        };
+    }
+}
+
+/// UART config.
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// Baud rate.
+    pub baudrate: u32,
+    /// Word length.
+    pub data_bits: DataBits,
+    /// Stop bits.
+    pub stop_bits: StopBits,
+    /// Parity bit.
+    pub parity: Parity,
+    /// Invert the tx pin output
+    pub invert_tx: bool,
+    /// Invert the rx pin input
+    pub invert_rx: bool,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            baudrate: 9600,
+            data_bits: DataBits::DataBits8,
+            stop_bits: StopBits::Stop1,
+            parity: Parity::ParityNone,
+            invert_rx: false,
+            invert_tx: false,
+        }
+    }
+}
+
+/// # Type parameters
+/// 'd: the lifetime marker ensuring correct borrow checking for peripherals used at compile time
+/// T: the peripheral instance type allowing usage of peripheral specific registers
+/// M: the operating mode of USART peripheral
+pub struct Usart<'d, T: Instance, M: Mode> {
+    tx: UsartTx<'d, T, M>,
+    rx: UsartRx<'d, T, M>,
+}
+
+pub struct UsartTx<'d, T: Instance, M: Mode> {
+    phantom: PhantomData<(&'d (), T, M)>,
+}
+
+pub struct UsartRx<'d, T: Instance, M: Mode> {
+    phantom: PhantomData<(&'d (), T, M)>,
+}
+
+impl<'d, T: Instance, M: Mode> UsartTx<'d, T, M> {
+    pub fn new(_usart: Peri<'d, T>, tx: Peri<'d, impl TxPin<T>>, config: Config) -> Self {
+        Usart::<T, M>::init(Some(tx.into()), None, config);
+        Self::new_inner()
+    }
+
+    #[inline]
+    fn new_inner() -> Self {
+        Self { phantom: PhantomData }
+    }
+
+    pub fn blocking_write(&mut self, buffer: &[u8]) -> Result<(), Error> {
+        T::blocking_write(buffer)
+    }
+
+    pub fn blocking_flush(&mut self) -> Result<(), Error> {
+        T::blocking_flush()
+    }
+
+    pub fn tx_busy(&self) -> bool {
+        T::tx_busy()
+    }
+}
+
+impl<'d, T: Instance> UsartTx<'d, T, Blocking> {
+    pub fn new_blocking(_usart: Peri<'d, T>, tx: Peri<'d, impl TxPin<T>>, config: Config) -> Self {
+        Usart::<T, Blocking>::init(Some(tx.into()), None, config);
+        Self::new_inner()
+    }
+}
+
+impl<'d, T: Instance, M: Mode> UsartRx<'d, T, M> {
+    pub fn new(_usart: Peri<'d, T>, rx: Peri<'d, impl RxPin<T>>, config: Config) -> Self {
+        Usart::<T, M>::init(None, Some(rx.into()), config);
+        Self::new_inner()
+    }
+
+    #[inline]
+    fn new_inner() -> Self {
+        Self { phantom: PhantomData }
+    }
+
+    pub fn blocking_read(&mut self, mut buffer: &mut [u8]) -> Result<(), Error> {
+        while !buffer.is_empty() {
+            match Self::drain_fifo(self, buffer) {
+                Ok(0) => continue, // Wait for more data
+                Ok(n) => buffer = &mut buffer[n..],
+                Err((_, err)) => return Err(err),
+            }
+        }
+        Ok(())
+    }
+
+    /// Returns:
+    /// - Ok(n) -> read n bytes
+    /// - Err(n, Error) -> read n-1 bytes, but encountered an error while reading nth byte
+    fn drain_fifo(&mut self, buffer: &mut [u8]) -> Result<usize, (usize, Error)> {
+        T::drain_fifo(buffer)
+    }
+}
+
+impl<'d, T: Instance> UsartRx<'d, T, Blocking> {
+    pub fn new_blocking(_usart: Peri<'d, T>, rx: Peri<'d, impl RxPin<T>>, config: Config) -> Self {
+        Usart::<T, Blocking>::init(None, Some(rx.into()), config);
+        Self::new_inner()
+    }
+}
+
+impl<'d, T: Instance> Usart<'d, T, Blocking> {
+    pub fn new_blocking(
+        usart: Peri<'d, T>,
+        tx: Peri<'d, impl TxPin<T>>,
+        rx: Peri<'d, impl RxPin<T>>,
+        config: Config,
+    ) -> Self {
+        Self::new_inner(usart, tx.into(), rx.into(), config)
+    }
+}
+
+impl<'d, T: Instance, M: Mode> Usart<'d, T, M> {
+    fn new_inner(_usart: Peri<'d, T>, mut tx: Peri<'d, AnyPin>, mut rx: Peri<'d, AnyPin>, config: Config) -> Self {
+        Self::init(Some(tx.reborrow()), Some(rx.reborrow()), config);
+        Self {
+            tx: UsartTx::new_inner(),
+            rx: UsartRx::new_inner(),
+        }
+    }
+
+    fn init(_tx: Option<Peri<'_, AnyPin>>, _rx: Option<Peri<'_, AnyPin>>, config: Config) {
+        T::enable_clock();
+        T::reset_flexcomm();
+        let source_clock: u32 = T::select_clock(config.baudrate);
+        T::configure_flexcomm();
+        T::tx_pin_config();
+        T::rx_pin_config();
+        Self::set_baudrate(source_clock, config.baudrate);
+        T::configure_usart(config);
+        T::disable_dma();
+        T::enable_usart();
+    }
+
+    fn set_baudrate(source_clock: u32, baudrate: u32) {
+        // There are two types of dividers: integer divider (baud rate generator register and oversample selection value)
+        // and fractional divider (fractional rate divider).
+        // For oversampling, the default is industry standard 16x oversampling, i.e. OSRVAL = 15
+
+        // The formulas are:
+
+        // FLCK = (clock selected via FCCLKSEL) / (1 + MULT / DIV)
+        // DIV is always 256, then:
+        // FLCK = (clock selected via FCCLKSEL) / (1 + MULT / 256)
+
+        // Baud rate = [FCLK / (OSRVAL+1)] / (BRGVAL + 1) =>
+        // Baud rate = [FCLK / 16] / (BRGVAL + 1)
+
+        // There are 2 unknowns: MULT and BRGVAL.
+        // MULT is responsible for fractional division
+        // BRGVAL is responsible for integer division
+
+        //  The Fractional Rate Generator can be used to obtain more precise baud rates when the
+        //  function clock is not a good multiple of standard (or otherwise desirable) baud rates.
+        //  The FRG is typically set up to produce an integer multiple of the highest required baud
+        //  rate, or a very close approximation. The BRG is then used to obtain the actual baud rate
+        //  needed.
+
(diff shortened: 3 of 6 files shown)


embassy-nxp/Cargo.toml
```diff
@@ -24,7 +24,8 @@ log = { version = "0.4.27", optional = true }
 embassy-time = { version = "0.4.0", path = "../embassy-time", optional = true }
 embassy-time-driver = { version = "0.2", path = "../embassy-time-driver", optional = true }
 embassy-time-queue-utils = { version = "0.2", path = "../embassy-time-queue-utils", optional = true }
-
+embedded-io = "0.6.1"
+embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["unproven"] }
```
eva-cosma, line 28: I think we can let the name just be `embedded-hal`, no need for `-02` in my eyes. Or is there a reason I am missing?

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+
+/// Serial error
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum Error {
+    /// Triggered when the FIFO (or shift-register) is overflowed.
+    Overrun,
+    /// Triggered when a break is received
+    Break,
```
eva-cosma: What does a break mean in this context? A loss of connection between devices, or something else?

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    /// 7 bits.
+    #[doc = "7 bit Data length."]
+    DataBits7,
+    /// 8 bits.
+    #[doc = "8 bit Data length."]
+    DataBits8,
+    /// 9 bits.
+    #[doc = "9 bit data length. The 9th bit is commonly used for addressing in multidrop mode."]
+    DataBits9,
+}
```
eva-cosma: I think we could remove the `#[doc = ...]` attribute and let the longer messages be the actual doc comment. It should be the same thing more or less. Plus the longer, more explanation-filled docs aren't that big.

Same comment for the other enums as well

```suggestion
pub enum DataBits {
    /// 7 bits data length
    DataBits7,
    /// 8 bits data length
    DataBits8,
    /// 9 bit data length. The 9th bit is commonly used for addressing in multidrop mode
    DataBits9,
}
```

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+        return match self {
+            Self::STOP1 => false,
+            Self::STOP2 => true,
+        };
+    }
+}
+
+/// UART config.
+#[non_exhaustive]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
eva-cosma: nitpick: IMHO, `Copy`, `PartialEq` and `Eq` doesn't make sense to me. While we can implement them on it, it might bloat the debug builds with extra code. And we already know how painfully slow flashing the LPC is :(

```suggestion
#[derive(Clone,  Debug)]
```

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    rx: UsartRx<'d, T, M>,
+}
+
+pub struct UsartTx<'d, T: Instance, M: Mode> {
+    phantom: PhantomData<(&'d (), T, M)>,
+}
+
+pub struct UsartRx<'d, T: Instance, M: Mode> {
+    phantom: PhantomData<(&'d(), T, M)>,
+}
```
eva-cosma: It's not immediately obvious to me what `'d`, `'T`, and `M` represent. I would actually appreciate a small doc comment explaining what they are/ what they are used for. Especially since this is public API

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    pub fn new(_usart: Peri<'d, T>, tx: Peri<'d, impl TxPin<T>>, config: Config) -> Self {
+        Usart::<T, M>::init( Some(tx.into()), None, config);
+        Self::new_inner()
+    }
+
+    fn new_inner() -> Self {
+        Self {
+            phantom: PhantomData,
+        }
+    }
```
eva-cosma: Unless `new_inner` is used somewhere else, or could be useful for us in the future, I would just put it directly into `new`
```suggestion
    pub fn new(_usart: Peri<'d, T>, tx: Peri<'d, impl TxPin<T>>, config: Config) -> Self {
        Usart::<T, M>::init( Some(tx.into()), None, config);
        
        Self {
            phantom: PhantomData,
        }
```

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    pub fn new(_usart: Peri<'d, T>, rx: Peri<'d, impl RxPin<T>>, config: Config) -> Self {
+        Usart::<T, M>::init(None, Some(rx.into()), config);
+        Self::new_inner()
+    }
+
+    fn new_inner() -> Self {
+        Self {
+            phantom: PhantomData,
+        }
+    }
```
eva-cosma: Same comment as for TX, unless there is a reason, I would inline `new_inner`

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+                Ok(n) => buffer = &mut buffer[n..],
+                Err((_, err)) => return Err(err),
+            }
+        }
+        Ok(())    
+    }
+
+    fn drain_fifo(&mut self, buffer: &mut [u8]) -> Result<usize, (usize, Error)> {
+        T::drain_fifo(buffer)
+    }
```
eva-cosma, line 205: It's not immediately obvious to me what the return type represents. My guess is that `Ok(n) -> read n bytes` or `Err((n, e)) -> read n bytes, but then encountered error`. Is that right? I would appreciate a doc comment on this.

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    fn set_baudrate(source_clock: u32, baudrate: u32) {
+        let brg_value = source_clock / (16 * baudrate); 
+        let raw_clock = source_clock / (16 * brg_value);
+        let mult_value: u32 = (raw_clock * 256 / baudrate) - 256;
+    
+        assert!(mult_value < 256);
+        assert!(brg_value < 256);
+        
+        T::set_baudrate(mult_value as u8, brg_value as u8);   
+    }
```
eva-cosma: There are a lot of magic numbers and formulas here, some references to where they come from, or even some explanation for them would be nice :)

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+
+    /// Read from USART RX blocking execution until done.
+    pub fn blocking_read(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+        self.rx.blocking_read(buffer)
+    }
+
+    /// Check if UART is busy transmitting.
+    pub fn busy(&self) -> bool {
+        self.tx.busy()
+    }
```
eva-cosma: nitpick. Although documented and generally obvious this refers to the `tx`, I could see a world where at a glance this could be interpreted as `RX or TX is currently being used by another task` when non-blocking USART gets made/implemented.

```suggestion
    /// Check if UART is busy transmitting.
    pub fn tx_busy(&self) -> bool {
        self.tx.busy()
    }
```

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+            }
+
+
+            // Adaptive clock choice based on baudrate
+            // By default, oversampling rate is set to be 16x
+            // Minimal and maximum values were computed taking these formulas into account:
+            // Flexcomm Interface function clock = (clock selected via FCCLKSEL) / (1 + MULT / DIV)
+            // baud_rate = source_clock / (16 * BRG). By default, oversampling rate is 16x.
+            
+            fn select_clock(baudrate: u32) -> u32 {
```
eva-cosma: nitpick: I appreciate the comments, it explains cleanly what is happening, but not WHY this is happening. I would appreciate some insight into why we must oversample, what is oversampling rate, why the clock must be higher than the baudrate.

Oh, and extra nitpick, and also maybe move the comments into the function, they can rest at the start, but iniside. Or make them doc comments. One of the two.

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+    ) => {
+        impl $crate::usart::SealedInstance for $crate::peripherals::$inst {
+            type UsartRegBlock = $crate::pac::$usartX::RegisterBlock;
+
+            fn usart_reg() -> &'static Self::UsartRegBlock {
+                unsafe { &*$crate::pac::$USARTX::ptr() }
+            }
+
+            fn enable_clock() {
+                syscon_reg().ahbclkctrl0.modify(|_, w| w.iocon().enable());
```
eva-cosma: Are there side-effects of writing to this register? For example if the iocon is already enabled, would re-enabling it reset anything? I'm thinking of the scenario in which we already have iocon enabled from another USART (or other peripheral) and we re-enable it, and I am worried of potential side-effects.

eva-cosma: If there might be, we could add a critical section, check the bit, and then write it if not enabled. It's not ideal if we get interrupted between reading/checking/writing, hence the suggestion for critical section

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+                syscon_reg().ahbclkctrl1.modify(|_, w| w.$FLEXCOMM_CLK_FIELD().enable());
```
eva-cosma: Similar to above, though I'm 90% sure this shouldn't have consequences/side-effects for double-enable. Though please double check

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+        T::reset_flexcomm();
+        
+        let source_clock: u32 = T::select_clock(config.baudrate);
+        T::configure_flexcomm();
+        T::tx_pin_config();
+        T::rx_pin_config();
+        Self::set_baudrate(source_clock, config.baudrate);
+        T::configure_usart(config);
+        T::configure_dma();
+        T::enable_usart();
```
eva-cosma: I would like to highlight how beautiful this code looks. It's very clear what is happening. Love it!

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+                    .normal()
+                    .oeta()
+                    .disabled()
+                    .oesel()
+                    .standard()
+                    .autoaddr()
+                    .disabled()
+                    .oepol()
+                    .low()
+                });
```
eva-cosma: I would appreciate it if there was a comment explainig all the default configuration we are doing to this. Or at least a page reference number to look it up

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+                    .stoplen()
+                    .bit(config.stop_bits.bits())
+                    .rxpol()
+                    .bit(config.invert_rx)
+                    .txpol()
+                    .bit(config.invert_tx)
+                });            
+            }
+
+            fn configure_dma() {
```
eva-cosma: Do we use DMA for blocking read/write?

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+                }
+            }
+
+            fn set_baudrate(mult_value: u8, brg_value: u8) {
+                syscon_reg()
+                    .$FRG_FIELD()
+                        .modify(|_, w| unsafe { w.div().bits(0xFF).mult().bits(mult_value as u8) });
+                Self::usart_reg()
+                    .brg
+                        .modify(|_, w| unsafe { w.brgval().bits((brg_value - 1) as u16) }); // Baud rate = Flexcomm Interface fucntion clock / (BRGVAL + 1)
```
eva-cosma: Nitpick: move the comment above the line. Easier to read
```suggestion
                // Baud rate = Flexcomm Interface fucntion clock / (BRGVAL + 1)
                Self::usart_reg()
                    .brg
                        .modify(|_, w| unsafe { w.brgval().bits((brg_value - 1) as u16) });
```

examples/lpc55s69/src/bin/usart_blocking.rs
```diff
@@ -0,0 +1,39 @@
...
+use embassy_executor::Spawner;
+use embassy_nxp::usart::{Config, Usart};
+use {defmt_rtt as _, panic_halt as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_nxp::init(Default::default());
+    let mut usart = Usart::new_blocking(p.USART2, p.PIO0_27, p.PIO1_24, Config::default());
+    let tx_buf: &[u8; 14] = b"Hello, Ferris!";
+    let mut rx_buf: [u8; 14] = [0u8; 14];
```
eva-cosma: Are the type annotation required? Or can they be inferred by rust?

examples/lpc55s69/src/bin/usart_blocking.rs
```diff
@@ -0,0 +1,39 @@
...
+    loop {
```
eva-cosma:
```suggestion
    let mut rx_buf: [u8; 14] = [0u8; 14];
    
    loop {
```

examples/lpc55s69/src/bin/usart_blocking.rs
```diff
@@ -0,0 +1,39 @@
...
+        info!("Write a message");
+        usart.blocking_write(tx_buf).unwrap();
+        usart.blocking_flush().unwrap();
+        info!("Read a message");
```
eva-cosma:
```suggestion
        usart.blocking_flush().unwrap();
        
        info!("Read a message");
```

examples/lpc55s69/src/bin/usart_blocking.rs
```diff
@@ -0,0 +1,39 @@
...
+        match from_utf8_mut(&mut rx_buf) {
+            Ok(str) => {
+                info!("The message is: {}", str);
+            }
+            Err(_) => {
+                error!("Error in converting to UTF8");
+            }
+        }
+
+        // Time driver should be used here instead
```
eva-cosma:
```suggestion
        // TODO(frihetselsker): Replace nop delay-loop with time driver, when implemented
```

embassy-nxp/src/usart/lpc55.rs
eva-cosma, line 1: I would personally like this file to be split up. My suggestion:

```
|- lpc55.rs -> contains `mod` declaration and `pub use`/`pub(crate) use` directives. To the user, it should only look like all of the API was implemented in this file.
\- lpc55/
    |- error.rs
    |- config.rs
    |- api.rs  - the Usart/UsartRx/UsartTx structures and their impl blocks
    |- hal_embedded.rs - the hal_embedded implementations
    |- embedded_io.rs - the embedded_io implementationts
    |- instance.rs - everything related to the internal driver code that, basically the macro rule
    \- instance_impl.rs -- implementation of pins/usarts. We could argue it should be part of `instance`, but it makes more sense to me to separate these since they are more config than business logic
```

I would also like some input from the maintainers if they are okay with this (@i509VCB ), since this is mostly my personal preference for slim self-contained files.

george-cosma (changes requested):
First of all - good job! I really like it and it shows you've put great effort into not only making it work but also making it look pretty and carefully engineered. I've left some nitpicks and some suggestions, more as finishing touches.

Also, don't forget to run `cargo fmt`, the CI is failing because of that. If you don't have differences locally when running cargo fmt it might be because of versioning issues/ not using stable/nightly correctly. I think I had some troubles with @IrinaCh524 about this at some point.

embassy-nxp/Cargo.toml
i509VCB: We can't have both 0.2 and 1.0 in the same dependency graph without them being named different.

embassy-nxp/src/usart/lpc55.rs
i509VCB: I think that's maybe too much splitting. If the file were 2000-3000+ lines long then maybe some spitting would make sense.

embassy-nxp/src/usart/lpc55.rs
i509VCB: ahbclkctrl0 could probably be written by multiple threads.

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,854 @@
...
+        fn configure_usart(config: Config);
+        fn configure_dma();
+        fn enable_usart();
+        fn blocking_write(buffer: &[u8]) -> Result<(), Error>;        
+        fn blocking_flush() -> Result<(), Error>;
+        fn tx_busy() -> bool;
+        fn drain_fifo(buffer: &mut [u8]) -> Result<usize, (usize, Error)>;
+        fn reset_flexcomm();
+        fn tx_pin_config();
+        fn rx_pin_config();
```
i509VCB: How many of these could be default impl on the trait definition. The size of the `impl_instance` macro does not look fun to maintain.

embassy-nxp/src/usart/lpc55.rs
i509VCB: My guess is that it is being disabled? Which you probably want to do in case you bind blocking and async interrupt handlers at the same time.

embassy-nxp/src/usart/lpc55.rs
frihetselsker: @i509VCB, hello! While you're still here, I want to ask you a question. I agree with you absolutely that `impl_instance` is extremely huge, and I can make a default implementation of the following functions:

- `configure_usart`
- `configure_dma` , which I'll fix as well
- `enable_usart`
- `blocking_write` , `blocking_flush`, `tx_busy` and `drain_fifo`

Basically, every function that depends solely on `usart` register, others depend more on macro implementation, because for every USART its own registers or fields are used. My question is: is it possible to make this block much simpler than that? While I was implementing it, I felt like I still could simplify this block, but I didn't see how.

i509VCB: An example (typing code on mobile is not enjoyable): https://github.com/embassy-rs/embassy/blob/main/embassy-nxp/src/gpio/rt1xxx.rs#L683

Your impl_instance would do things like "what block is it. For anything that depends on that you just define the implementation in the trait (if a function in a trait has a function body, then that is the default implemention).

examples/lpc55s69/src/bin/usart_blocking.rs
frihetselsker: I've just checked: yes, they can. Fixing right now.

embassy-nxp/src/usart/lpc55.rs
frihetselsker: Yes, for now, we can just put `#[inline]`. It'll become evident whether we need it or on when async version is being developed.

embassy-nxp/Cargo.toml
eva-cosma: Ah okay.  Makes sense.

embassy-nxp/src/usart/lpc55.rs
eva-cosma: I'll defer to your judgement here. I think we can leave it as-is for now and revisit it later when async and dma access gets implemented

embassy-nxp/src/usart/lpc55.rs
i509VCB: You hold the line low for one uart "frame". Sometimes used for signalling an end of message.

frihetselsker: Yes. Sorry, I've forgotten to remove it from my code for now. I haven't tried it yet, it can be implemented in the future.

embassy-nxp/src/usart/lpc55.rs
frihetselsker: Yes, it is being disabled

embassy-nxp/src/usart/lpc55.rs
frihetselsker: @i509VCB, hello again. I've tried to solve the issue so that it will be easier to maintain the code. However, `RegisterBlock` is also generic and Rust cannot infer the type, `lpc55-pac` limits me. We can leave it as-is, once `nxp-pac` gets lpc55 support, it will be fixed. I'll create a new issue when I get an approval from you.

george-cosma (approved):

frihetselsker:
I've fixed a bug in `set_baudrate()` function, now it should not fail.

embassy-nxp/src/usart/lpc55.rs
i509VCB: > ```
> type UsartRegBlock = $crate::pac::$usartX::RegisterBlock;
> ```

Is there a reason why all usart instances couldn't share the same block type? I'd think just pick usart1 as the type for all instances? We can deal with merging these in nxp-pac so its just a generic `usart`.

We would ideally like to avoid that so we can erase the type of `T` in the `Usart` types.

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,916 @@
...
+    ($name:ident) => {
+        impl SealedMode for $name {}
+        impl Mode for $name {}
+    };
+}
+
+/// Blocking mode.
+pub struct Blocking;
+
+impl_mode!(Blocking);
```
i509VCB: This should go in `lib.rs`. It will apply to more than just uart

frihetselsker: @i509VCB I've fixed what you asked.

- I moved those lines to `lib.rs`
- I finally found the actual type of pointers to `USARTx` and wrote the default implementation
- I updated the example because RTC driver had been merged

Waiting for your review. Also:

> We can deal with merging these in nxp-pac so its just a generic `usart`.
>
>We would ideally like to avoid that so we can erase the type of `T` in the `Usart` types.

I agree with you absolutely. Ideally, there should be an `Info` structure that will contain all the necessary pointers for every USART instance, but let's wait for the `nxp-pac`.

i509VCB: Yeah sure we can wait to fix that later, just don't want to forget about it.

embassy-nxp/src/usart.rs
```diff
@@ -0,0 +1,24 @@
...
+    ($name:ident) => {
+        impl SealedMode for $name {}
+        impl Mode for $name {}
+    };
+}
+
+/// Blocking mode.
+pub struct Blocking;
+
+impl_mode!(Blocking);
```
i509VCB: I guess I should clarify I meant `src/lib.rs` for the location of this.

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -0,0 +1,885 @@
...
+    }
+}
+
+/// Stop bits.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum StopBits {
+    /// 1 stop bit.
+    STOP1,
+    /// 2 stop bits. This setting should only be used for asynchronous communication.
+    STOP2,
```
i509VCB: Probably should be `StopX`. `STOP` isn't an abbreviation for anything and the Rust conventions for names suggest even naming abbreviations as if its a word (`Http`, not `HTTP`)

embassy-nxp/src/usart.rs
frihetselsker: I'm sorry, will fix it right now.

embassy-nxp/src/usart/lpc55.rs
frihetselsker: Agree, already renamed.

frihetselsker:
@i509VCB I've made final changes. Sorry again for not moving code to `lib.rs` right away.

PR #4510, opened 2025-08-05
Light up maths functions for RP2040

JYouren:
There is some useful math functions available in the rp codebase that are not publicly accessible. This allows them to be used.

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -61,8 +61,6 @@ use crate::time_driver::get_driver;
 
 const THREAD_PENDER: usize = usize::MAX;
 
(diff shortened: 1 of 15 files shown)


embassy-rp/src/float/add_sub.rs
```diff
@@ -1,3 +1,5 @@
+//! Rust implementation of floating point addition and subtraction.
```
mkj: Would it be clearer for this to be "ROM implementation"? There doesn't seem much rust-specific.

JYouren: Sure, updated

lulf (approved):

Issue #4513, opened 2025-08-07
stm32/i2c: clock rate issues above 400 kHz


tomaz-suller:
I tried looking at the signals generated by the I2C peripheral on a logic analyser and couldn't get the clock to go over 400 kHz, whereas fast mode plus can go up to 1 MHz. I've been doing my tests on the NUCLEO-WB55 with the STM32WB55RG, GPIOv2 and I2Cv2.

A possible culprit seems to be the GPIO settings for SDA and SCK:

https://github.com/embassy-rs/embassy/blob/3a197df07c14873a52e42394ef350152eef60a41/embassy-stm32/src/i2c/config.rs#L142-L170

Medium speed means high frequency signals would fail to be generated properly.

phycrax:
I see two possible solutions for this.
1. Add speed to Config and let users to decide the speed.
2. Add frequency to Config and derive speed from given frequency.

I would vote for the first option as it gives more control to users. Maybe both frequency and speed should be moved into the Config struct.

phycrax:
Yeah both UART and SPI Configs contain the frequency and baud, it's better to move freq into config.

PR #4517, opened 2025-08-08
embassy-stm32: Add frequency and gpio speed to Config for i2c and i2s

phycrax:
Fixes [#4510](https://github.com/embassy-rs/embassy/issues/4513)
Also did a little bit cleanup to unify naming for config params.

### DIFF
--- a/embassy-stm32/src/i2c/config.rs
+++ b/embassy-stm32/src/i2c/config.rs
@@ -1,6 +1,7 @@
 #[cfg(gpio_v2)]
 use crate::gpio::Pull;
 use crate::gpio::{AfType, OutputType, Speed};
+use crate::time::Hertz;
 
 #[repr(u8)]
 #[derive(Copy, Clone)]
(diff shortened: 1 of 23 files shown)


Dirbaio (approved):
thanks!

tomaz-suller:
I know this has been approved, but do we really want to break the public-facing API so carelessly? We could leave the frequency arguments as they are in the existing `I::new` methods and just add the GPIO speed to the I2C and I2S

tomaz-suller:
The merge error seems to have to do not with the changes but with an issue in the CI

```
 ERROR teleprobe::client > === nrf5340-dk out/tests/nrf5340-dk/buffered_uart: FAILED: HTTP request failed with status code: 400: Bad Request
 ERROR teleprobe::client > ERROR - Run failed: Probe did not appear after the max settle time.
```

Dirbaio:
> I know this has been approved, but do we really want to break the public-facing API so carelessly?

next release is going to be a major bump anyway, there are many breaking changes in git already.

PR #4525, opened 2025-08-11
book: Suggest Picotool for RP chips instead of elf2uf2-rs

robjwells:
This amends the FAQ section of the book to suggest the use of [Picotool](https://github.com/raspberrypi/picotool) to upload binaries to RP2* chips without using a debug probe, instead of [`elf2uf2-rs`](https://github.com/JoNil/elf2uf2-rs/).

As of Rust 1.89, the stable compiler emits ELF binaries that contain a header that [fails an overly strict check in elf2uf2-rs](https://github.com/JoNil/elf2uf2-rs/issues/40). In addition, elf2uf2-rs appears to be unmaintained, while Picotool is maintained by the vendor.

The link in the amended documentation is to the [pico-sdk-tools repo](https://github.com/raspberrypi/pico-sdk-tools), not the Picotool repo, as it is where the pre-built binaries are published.

### DIFF
--- a/docs/pages/faq.adoc
+++ b/docs/pages/faq.adoc
@@ -6,16 +6,16 @@ Please feel free to add items to link:https://github.com/embassy-rs/embassy/edit
 
 == How to deploy to RP2040 or RP235x without a debugging probe.
 
-Install link:https://github.com/JoNil/elf2uf2-rs[elf2uf2-rs] for converting the generated elf binary into a uf2 file.
+Install link:https://github.com/raspberrypi/pico-sdk-tools/releases[Picotool] for uploading the binary.
 
 Configure the runner to use this tool, add this to `.cargo/config.toml`:
 [source,toml]
 ----
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "elf2uf2-rs --deploy --serial --verbose"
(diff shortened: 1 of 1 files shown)


robjwells:
Force-push updated the Picotool arguments to include `--update` (skip writing identical sectors) and `--verify`.

Edit: Actually the _following_ force-push makes that update as the one above did nothing. Sorry, it's late here.

docs/pages/faq.adoc
```diff
@@ -6,16 +6,16 @@ Please feel free to add items to link:https://github.com/embassy-rs/embassy/edit
 
 == How to deploy to RP2040 or RP235x without a debugging probe.
 
-Install link:https://github.com/JoNil/elf2uf2-rs[elf2uf2-rs] for converting the generated elf binary into a uf2 file.
+Install link:https://github.com/raspberrypi/pico-sdk-tools/releases[Picotool] for converting the generated elf binary into a uf2 file.
```
Dirbaio: this is no longer accurate: it doen't convert to uf2, just flashes directly.

robjwells: Interesting, thanks. I've reworded this to just "Install Picotool for uploading the binary".

PR #4536, opened 2025-08-12
chore: add changelog enforcement

lulf:
Based on similar workflow for esp-hal.

This requires pull requests to add changelog entries to modules that are touched, unless the skip-changelog label is set.


lulf:
Did a quick test, and it works. I think this will make changelog maintenance scale with the number of contributors.

PR #4538, opened 2025-08-13
Added VREFBUF-TRIM manual write. Known errata

leftger:
VREFBUF has TRIM values at known memory locations.
Screenshot 2025-08-12 at 7 05 53 PM

This PR updates the new VREFBUF implementation to provide proper support for its initialization.
Screenshot 2025-08-12 at 7 05 09 PM

Any feedback is welcome 😄

### DIFF
--- a/embassy-stm32/src/vrefbuf/mod.rs
+++ b/embassy-stm32/src/vrefbuf/mod.rs
@@ -1,5 +1,4 @@
 //! Voltage Reference Buffer (VREFBUF)
-// use core::ptr::{read_volatile, write_volatile};
 use core::marker::PhantomData;
 
 use embassy_hal_internal::PeripheralType;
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/vrefbuf/mod.rs
```diff
@@ -12,6 +11,26 @@ pub struct VoltageReferenceBuffer<'d, T: Instance> {
     vrefbuf: PhantomData<&'d mut T>,
 }
 
+#[cfg(rcc_wba)]
+#[repr(usize)]
+enum VRefBufTrim {
```
Dirbaio: it's very strange to create an enum like this just for the values. I'd suggest making `get_refbuf_trim` return `*const u32` and putting the addresses there.

leftger: I have removed the enum. Thanks for the feedback.

PR #4541, opened 2025-08-13
fix stm32 i2c: wait for STOP flag in blocking read to terminate read transaction

### LINKED ISSUE #4540: embassy-stm32: Bug in blocking I2C
Hi,

I am using the eeprom24x crate to read and write to an eeprom via I2C. When reading single bytes from it, all works. When I read multiple bytes, the I2C read operation returns a timeout error the second time I use it.
When I switch to using the async i2c implementation though, everything works as expected.

Here is a bit of example code:

```rust
#[embassy_executor::main]
async fn app(_spawner: Spawner) {
    configure_clock(&mut config);

    let mut config: embassy_stm32::i2c::Config = Default::default();
    config.frequency = Hertz(100_000); // 400 kHz
    let i2c = I2c::new_blocking(p.I2C4, p.PB8, p.PB7, config);
    let mut eeprom = Eeprom24x::new_24x64(i2c, Default::default());

    // let i2c: I2c<'_, embassy_stm32::mode::Async, Master> =
    //     I2c::new(p.I2C4, p.PB8, p.PB7, Irqs, p.DMA1_CH5, p.DMA1_CH1, config);
    // let mut eeprom = Eeprom24x::new_24x64_async(i2c, Default::default());

    defmt::info!("EEPROM read byte: {}", eeprom.read_byte(0x00).unwrap());
    Timer::after(Duration::from_micros(100)).await;

    let mut data = [0u8; 4];
    for _ in 0..3 {
        if let Ok(()) = eeprom.read_data(0x00, &mut data) {
            defmt::info!("EEPROM read data: {:?}", data);
        } else {
            defmt::error!("Failed to read EEPROM data");
        };
    }
    Timer::after(Duration::from_micros(100)).await;
    loop {}
}
```

The defmt output for running this is consistently:
```
INFO  EEPROM read byte: 255
INFO  EEPROM read data: [255, 255, 255, 255]
ERROR Failed to read EEPROM data
ERROR Failed to read EEPROM data
```

I connected to the bus with my logic analyzer, this is what I found:

For the blocking I2C:

Image

And for the async I2C:

Image

We can see that the byte transfer succeeds for both variants as expected, but in the blocking operation there are no ACK sent anymore after the first data byte is read, and there also is no STOP / NAK at the end of the transfer causing the SDA to stay low and all following operations to time out.
At this point I don't know how to debug this any further. Any help would be appreciated!

mbrieske:
This PR fixes a bug where subsequent blocking I2C reads would fail with a timeout.

The read_internal function, which uses Stop::Automatic, was returning before the STOP condition was actually complete on the bus. This created a race condition where the next transaction would start while the bus was still busy.

The fix is to call wait_stop() at the end of the read. This ensures the transaction is fully terminated and the bus is idle before the function returns.

Fixes #4540

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -408,6 +408,7 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
                 *byte = self.info.regs.rxdr().read().rxdata();
             }
         }
+        self.wait_stop(timeout)?;
         Ok(())
     }
 
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -464,11 +464,13 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
             }
         }
         // Wait until the write finishes
-        let result = self.wait_tc(timeout);
+        self.wait_tc(timeout)?;
         if send_stop {
             self.master_stop();
+            self.wait_stop(timeout)?;
         }
-        result
+
+        Ok(())
     }
 
     // =========================


mbrieske:
fixes #4542 now too

PR #4544, opened 2025-08-13
STM32F1: Add I2S

fwolter:
The SPI cubedb XML files are shared across F1 devices that have and do not have I2S. Therefore, I didn't make the change in perimap or via the peripheral version cfg.
STM32F100 and F101 do not have I2S.
grafik


Dirbaio:
closing in favor of #4620. I ended up adding versions in stm32-data for chips with and without i2s so we can match on those in embassy-stm32 instead of having to encode which chips have it with cfgs.

PR #4550, opened 2025-08-14
Make timer queue item opaque, move to new crate

### LINKED ISSUE #4516: Decouple executor and integrated timers
Updating embassy-executor is a breaking change for embassy-time-queue-utils, which is in turn a breaking change for HALs, embassy-time, and crates using either of those. This is because embassy-executor and embassy-time-queue-utils are coupled by the integrated timer queue feature. Currently, embassy-time-queue-utils needs the TimerQueueItem struct, a way to access that struct given a Waker, and a way to wake up a task when a timer queue item is due.

This dependency needs to be broken somehow. An attempt was made [here](https://github.com/embassy-rs/embassy/pull/4503), but the end result is replaces explicit dependency with implicit expectations (the dependency also exists, just backwards, embassy-executor depending on the [much less often changing] queue utils crate), so I don't consider it shippable.

bugadani:
Take ~~2~~ 3 to fix #4516

Using words isn't exactly space efficient, but necessary for 64-bit targets. The ETQU queue item contains 3 pointers, so we can't simply use bytes to define the item size - unless we want to specify the dependency separately for different `target_pointer_width` cfgs.


lulf (commented):
I don't have much useful to contribute, but the timer item-size is generally not something any regular joe developer would concern himself about right? Or should there be some guidance on the impact of these values/why they are set the way they are?

bugadani:
> but the timer item-size is generally not something any regular joe developer would concern himself about right

This whole crate will ideally not be visible to end users. The intended audience is HAL developers who need a *custom* integrated queue implementation.

lulf (approved):
No objections for me. If this approach been discussed and agreed upon, its fine by me to merge it.

embassy-time-queue-utils/CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- Removed the embassy-executor dependency
+
+## 0.2.0 - 2025-08-04
+
+Bumpep embassy-executor
```
ivmarkov: Typo: "bumped" maybe?

PR #4555, opened 2025-08-15
feat(embassy-time): add Ticker::now_and_every

darkwater:
I use Ticker to run a process X seconds, but I'd like that process' first run to be immediately rather than X seconds after boot. In my situation, I can't just put the `ticker.next().await;` after the process, because I `select()` it with something else.

### DIFF
--- a/embassy-time/CHANGELOG.md
+++ b/embassy-time/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `saturating_add` and `saturating_sub` to `Instant`
 - Add `Instant::try_from_*` constructor functions
 - Add `Duration::try_from_*` constructor functions
+- Add `Ticker::now_and_every` constructor function
 - Don't select `critical-section` impl for `std`
 - Manually implement the future for `with_timeout`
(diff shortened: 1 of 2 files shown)


fwolter:
Wouldn't an `expire()` method be more versatile?

darkwater:
I do think an additional method to change the deadline would be useful too yeah, what exactly do you have in mind?

The main limitation with the current API is that every methods adds `+ self.duration` to the deadline.

Edit: Maybe the `expires_at` and `duration` fields should just be public. The only reason I see to not do this would to be able to change the internals later, but I think Ticker is pretty solid as-is.

amchugh:
I was about to write a similar method, `fn at_and_then(first_tick: Instant, duration: Duration) -> Self` which accomplishes almost the same thing as you've written (`at_and_then(Instant::now(), duration)` is the same as `now_and_every(duration)`). I can get around this with the following:
```rs
let mut ticker = Ticker::every(duration);
ticker.reset_at(first_tick - duration);
```
This feels clunky. You could get your `now_and_every` by doing:
```rs
let mut ticker = Ticker::every(duration);
ticker.reset_at(Instant::now() - duration);
```
That is gross to me. These both also run the risk of underflow... so a proper fix would be even longer.

I think one of these two APIs should exist, for convenience. I don't think making the fields public is the way to go, I think the API should provide the functions people want to use instead :)

Happy to write that commit too. I think my name could do with some work.

Dirbaio:
Hello,

Issue #4558, opened 2025-08-17
Order of ADC and Channel creation impacts results


cconstantine:
If you move [this line](https://github.com/embassy-rs/embassy/blob/bfe4395b3b9792bb79363e2f32a9ab7bf69bb78d/examples/rp/src/bin/adc.rs#L25) to [here](https://github.com/embassy-rs/embassy/blob/bfe4395b3b9792bb79363e2f32a9ab7bf69bb78d/examples/rp/src/bin/adc.rs#L31), the readings from the ADC will be wrong.

It appears that you need to initialize the ADC before creating the channels, but that isn't obvious from the code and the docs don't say anything about it.  Would it make sense to add something to the docs?

9names:
So you're saying that if you do
```rust
let mut p26 = Channel::new_pin(p.PIN_26, Pull::None);
let mut adc = Adc::new(p.ADC, Irqs, Config::default());
```
instead of

```rust
let mut adc = Adc::new(p.ADC, Irqs, Config::default());
let mut p26 = Channel::new_pin(p.PIN_26, Pull::None);
```
it doesn't work properly?

I think just documenting this isn't going to save people from the footgun here.
Probably best if we require a reference to an `Adc` instance to set up Channels so that you can't do it wrong.

9names:
This seems to only affect the temperature sensor, regular ADC measurements are still correct regardless of order

PR #4559, opened 2025-08-17
fix: add adc ref to rp channel creation funcs

9names:
Potential fix for #4558.
I've applied the change for both new_pin() and new_temp_sensor(), but in my testing it seems that it's only necessary for new_temp_sensor(), at least on rp2040.

It's really an API question whether we should include it for symmetry between the different calls, or only include it for the function that it's necessary for.

Will convert it from draft once this has an answer, but I thought it was worth showing what effect it will have on examples.


Dirbaio:

leftger:
It's been a year. If someone wants to resubmit this in the future you're more than welcome to do so.

PR #4560, opened 2025-08-17
Optimize embedded_hal_async::delay::DelayNs impl

823984418:
Modify `async fn` to return `impl Future`.

`async { xx.await }` is not zero cost, so such invalid packaging should be avoided without affecting correctness.

### DIFF
--- a/embassy-time/src/delay.rs
+++ b/embassy-time/src/delay.rs
@@ -1,3 +1,5 @@
+use core::future::Future;
+
 use super::{Duration, Instant};
 use crate::Timer;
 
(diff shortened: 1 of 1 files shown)


lulf:
Both variants should produce the same code by the compiler, are you seeing otherwise?

823984418:
> Both variants should produce the same code by the compiler, are you seeing otherwise?

Not so. Simple polling code gets optimized purely because LLVM sees through the scheduling logic via inlining. Once the scheduling becomes slightly more complex, each nested async block introduces an additional match discriminant.

[Playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=5038a1394f5c2715b18e211137d8b047)

In fact, each async block generates 3 + number of `.await` states, where the 3 states include: 'not yet polled', 'completed', and 'panicked'.

Dirbaio (approved):
yeah it's very unfortunate. the compiler SHOULD be smart enough to realize the extra states are useless, but currently it doesn't.

PR #4564, opened 2025-08-18
Add configurable internal pullups for rp i2c

erwin-ps:
- Example updated to demonstrate enabling internal pullups
- Add `sda_pullup` and `scl_pullup` fields to I2C Config

### DIFF
--- a/embassy-rp/CHANGELOG.md
+++ b/embassy-rp/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- next-header -->
 ## Unreleased - ReleaseDate
+- add `i2c` internal pullup options ([#4564](https://github.com/embassy-rs/embassy/pull/4564))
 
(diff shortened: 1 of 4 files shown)


embassy-rp/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- next-header -->
 ## Unreleased - ReleaseDate
+- add `i2c` internal pullup options ([#4564](https://github.com/embassy-rs/embassy/pull/4564))
```
jamesmunns, line 10: Maybe note that this also now makes pullups *off* by default, whereas before they were *on* by default.

TBH I think off is a better default, but it is a change, which might affect some people.

erwin-ps: yes true, i can imagine it is troublesome if you update and your device stops working.

what shall i do mention the default state change in the changelog or change the default to on again?

PR #4569, opened 2025-08-19
rp: fix blocking I2C example regarding pull-up resistors

robjwells:
This amends the [blocking I2C example](https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/i2c_blocking.rs) for embassy-rp.

Commit bbc3e49 added a pull-up configuration and a comment that pull-ups were not enabled by default. This was made out-of-date by badcdcc, which ensured pull-ups were enabled to reduce the disruption of the larger I2C configuration change (see #4564).

This PR removes the (now-unnecessary) pull-up configuration, and adds a comment to clarify that the default I2C configuration enables pull-ups.

### DIFF
--- a/examples/rp/src/bin/i2c_blocking.rs
+++ b/examples/rp/src/bin/i2c_blocking.rs
@@ -49,10 +49,8 @@ async fn main(_spawner: Spawner) {
     let scl = p.PIN_15;
 
     info!("set up i2c ");
-    let mut config = Config::default();
-    // by default internal pullup resitors are disabled
-    config.sda_pullup = true;
-    config.scl_pullup = true;
+    // Default I2C config enables internal pull-up resistors.
+    let config = Config::default();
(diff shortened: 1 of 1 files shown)


jamesmunns (approved):
Good catch, thanks!

It's gunna ask for a changelog item :)

robjwells:
Whoops, I'll sort that now.

jamesmunns:
Nope, you were right!

robjwells:
Config doesn't need to be mutable now, I'll fix and force-push.

mkj:
Is there any plan to eventually have a breaking change for pullups off by default? If so maybe the example should keep the explicit config and comment that?

robjwells:
I’m happy to revise as is deemed best.

jamesmunns:
I'm going to merge this now, because it corrects something that is definitely wrong now.

I don't know how to remind us to re-visit this before the next breaking change, but we'll do our best I suppose!

PR #4572, opened 2025-08-20
nrf54l - Add SPI support

nerwalt:
Adding SPI (Spim and Spis) support to the nrf54l

### DIFF
--- a/embassy-nrf/src/spim.rs
+++ b/embassy-nrf/src/spim.rs
@@ -13,6 +13,9 @@ use embassy_embedded_hal::SetConfig;
 use embassy_hal_internal::{Peri, PeripheralType};
 use embassy_sync::waitqueue::AtomicWaker;
 pub use embedded_hal_02::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_3};
+#[cfg(feature = "_nrf54l")]
+pub use pac::spim::vals::Order as BitOrder;
+#[cfg(not(feature = "_nrf54l"))]
 pub use pac::spim::vals::{Frequency, Order as BitOrder};
 
 use crate::chip::{EASY_DMA_SIZE, FORCE_COPY_BUFFER_SIZE};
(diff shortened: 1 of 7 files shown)


nerwalt:
This PR is inspired by [4393](https://github.com/embassy-rs/embassy/pull/4393) but takes a different approach to the SPI frequency. The nrf54l introduces a frequency prescaler that can take a relatively wide range of values compared to other NRF chips. Instead of introducing a new Frequency type, I left the prescaler as a u8 for simplicity. The prescaler value is checked in `set_config()`, and an error is returned if it doesn't meet its constraints.

embassy-nrf/src/spim.rs
```diff
@@ -230,13 +258,29 @@ impl<'d, T: Instance> Spim<'d, T> {
...
+        #[cfg(not(feature = "_nrf54l"))]
+        {
+            r.rxd().ptr().write_value(rx_ptr);
+            r.rxd().maxcnt().write(|w| w.set_maxcnt(rx_len as _));
+        }
+        #[cfg(feature = "_nrf54l")]
+        {
+            r.dma().rx().ptr().write_value(rx_ptr);
+            r.dma().rx().maxcnt().write(|w| w.set_maxcnt(rx_len as _));
+        }
```
dmitry-tarnyagin, line 271: Here and below you can see something I reeeeeally would like to avoid (and definitely don't want to support :-) ): #[cfg]'s in the drive code that address the register programming interface change introduced by the nrf54l. 
It looks like there is no good way out of the problem. My approach was to hide the change and emulate classic interface, but the project owners want to move the whole codebase to the new interface at the pac level instead. 

My opinion: I don't like this and I don't like that [(c) Muzzy from Gondoland], but do not want to block nrf54l development either :)

lulf (commented):
As mentioned in the comments, and on the earlier nrf54 PR, it would be great if changes were made to nrf-pac to expose a more generic interface that both nrf54 and nrf52 can support

embassy-nrf/src/spim.rs
nerwalt: Agreed that it is not an easy problem. I gave this approach a shot to see what it looks like. I am not a big fan of all the `#[cfg()]s`. I think I understand why you went the `extend` route after looking at the `uarte` module. The number `#[cfg()]`s required in that module is a lot.

nerwalt:
> As mentioned in the comments, and on the earlier nrf54 PR, it would be great if changes were made to nrf-pac to expose a more generic interface that both nrf54 and nrf52 can support

I will look more into the PAC changes. I wanted to try this approach to see how it looked. The same approach to the `uarte` module makes it unmaintainable, so I understand the pushback.

PR #4574, opened 2025-08-21
MSPM0: add window watchdog implementation (WWDT0, WWDT1)

bespsm:
- bump revision for mspm0-data-generated
- add main implementation based on WWDT0, WWDT1 peripherals
- add watchdog examples for all chips, that tests "too early" and "too late" cases


bespsm:
@i509VCB as promised, PR with implementation

embassy-mspm0/src/watchdog.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        // init delay, 16 cycles
+        cortex_m::asm::delay(16);
+
+        //init watchdog
+        T::REGS.wwdtctl0().write(|w| {
+            w.set_clkdiv(config.timeout.get_clkdiv());
+            w.set_per(config.timeout.get_period());
+            w.set_mode(vals::Mode::WINDOW);
+            match config.closed_window {
```
i509VCB: I think a conversion function for closed_window would make the function a little less long.

embassy-mspm0/src/watchdog.rs
```diff
@@ -0,0 +1,330 @@
...
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            timeout: Timeout::Sec1,
+            closed_window: ClosedWindowPercentage::Zero,
+        }
+    }
+}
+
+pub struct Watchdog<'d, T: Instance> {
```
i509VCB: Given some chips have multiple watchdog instances, do we want to consider erasing the generic type here?

You would only need to store a `wwdt::Wwdt` inside which is 4 bytes total (its a pointer pretty much).

embassy-mspm0/src/watchdog.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        // Set Window0 as active window
+        T::REGS.wwdtctl1().write(|w| {
+            w.set_winsel(vals::Winsel::WIN0);
+            w.set_key(vals::Wwdtctl1Key::KEY);
+        });
+
+        // make sure watchdog triggers BOOTRST
+        pac::SYSCTL.systemcfg().write(|w| {
+            w.set_wwdtlp0rstdis(false);
```
i509VCB: What about `WWDTLP1RSTDIS`?

i509VCB (commented):
I assume this pull request isn't doing the NMI mode for the watchdog timers? (Likely a future one?)

embassy-mspm0/src/watchdog.rs
```diff
@@ -0,0 +1,330 @@
...
+        });
+
+        // Set Window0 as active window
+        T::REGS.wwdtctl1().write(|w| {
+            w.set_winsel(vals::Winsel::WIN0);
+            w.set_key(vals::Wwdtctl1Key::KEY);
+        });
+
+        // make sure watchdog triggers BOOTRST
+        pac::SYSCTL.systemcfg().write(|w| {
```
i509VCB: If WWDTLP1RSTDIS is needed then this sysctl access needs to be guarded by a critical section in the case where thread a configures wwdt0 and thread b preempts and configures wwdt1.

And it would need to be a modify because both watchdogs are independent.

bespsm:
> I assume this pull request isn't doing the NMI mode for the watchdog timers? (Likely a future one?)

exactly, NMI could be provided later, if someone needs it

embassy-mspm0/src/watchdog.rs
bespsm: fixed

embassy-mspm0/src/watchdog.rs
bespsm: ohh, sure thanks!

embassy-mspm0/src/watchdog.rs
bespsm: fixed

embassy-mspm0/src/watchdog.rs
bespsm: just replacing  generic type with `wwdt::Wwdt ` complains about missing trait `PeripheralType` for `Wwdt`. 

I guess it would require some some macro magic, any hints to fix it?

i509VCB: So you would just replace `T::REGS` with `self.regs` at usage sites. (wwdt::Wwdt is just Regs)

embassy-mspm0/src/watchdog.rs
```diff
@@ -0,0 +1,341 @@
...
+            w.set_key(vals::Wwdtctl1Key::KEY);
+        });
+
+        critical_section::with(|_| {
+            // make sure watchdog triggers BOOTRST
+            pac::SYSCTL.systemcfg().write(|w| {
+                w.set_wwdtlp0rstdis(false);
+
+                #[cfg(any(mspm0g110x, mspm0g150x, mspm0g151x, mspm0g310x, mspm0g350x, mspm0g351x))]
+                w.set_wwdtlp1rstdis(false);
```
i509VCB: We should not unconditionally write this the wwdt 1 field. If we aren't using WWDT1 in this instance of Watchdog, we technically do not own its fields or this bit of sysctl. This also lets the user do raw pac access on wwdt if they wanted. 

To determine if we have WWDT0, check if `self.regs == pac::WWDT0`

If we somehow get a 3rd wwdt in the future then we can change that later.

embassy-mspm0/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub mod gpio;
 pub mod i2c;
 pub mod timer;
 pub mod uart;
+pub mod watchdog;
```
i509VCB: We should match the datasheet/reference manual and use `wwdt` for the module name. All the other embassy-hals do this. Even nRF with its unusual `twim` module name. 

The type names inside the module can stay as is.

bespsm: fine, fixed

embassy-mspm0/src/watchdog.rs
bespsm: what we receive in `new` function, is an instance of generated structs  (generated by `embassy-hal-internal/src/macros.rs` ) with `PeripheralType` trait, not `wwdt::Wwdt`. And in `impl_wwdt_instance!` we map one with another to get to access to the registers.

But going back to your comment about "multiple watchdog instances", I adapted the example with WWDT0 & WWDT1 on launchpad g3507. it works with the current implementation well (both watchdogs restarts the system)

embassy-mspm0/src/watchdog.rs
bespsm: fixed

embassy-mspm0/src/watchdog.rs
bespsm: Okay, I interpreted wrongly your comment and tried to exclude `T` from `new` function as well. in the recent push I fixed it

bespsm:
@i509VCB please check

embassy-mspm0/src/wwdt.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+pub struct Watchdog {
+    regs: &'static Regs,
+}
+
+impl Watchdog {
+    /// Watchdog initialization.
+    pub fn new<T: Instance>(_instance: Peri<T>, config: Config) -> Self {
```
i509VCB, line 269: I was confused why you were having `new` take ownership of the peripheral (and not mutably borrow), but this is apparently correct per reference manual:

> The first write (with a key match) to the WWDTCTL0 register enables the WWDT. Once the WWDT is enabled, the WWDTCTL0 register becomes write protected.

embassy-mspm0/src/wwdt.rs
```diff
@@ -0,0 +1,345 @@
...
+                if *T::regs() == pac::WWDT0 {
+                    w.set_wwdtlp0rstdis(false);
+                }
+
+                #[cfg(any(mspm0g110x, mspm0g150x, mspm0g151x, mspm0g310x, mspm0g350x, mspm0g351x))]
+                if *T::regs() == pac::WWDT1 {
+                    w.set_wwdtlp1rstdis(false);
+                }
+            });
+        });
```
i509VCB: Given that a write to `WWDTCTL0` starts the watchdog, do we want to move this before the write to `WWDTCTL0`?

Also this needs to be modify, not write.

embassy-mspm0/src/wwdt.rs
```diff
@@ -0,0 +1,345 @@
...
+            w.set_window0(config.closed_window.get_native_size());
+            w.set_window1(vals::Window::SIZE_0);
+            w.set_key(vals::Wwdtctl0Key::KEY);
+        });
+
+        // Set Window0 as active window
+        T::regs().wwdtctl1().write(|w| {
+            w.set_winsel(vals::Winsel::WIN0);
+            w.set_key(vals::Wwdtctl1Key::KEY);
+        });
```
i509VCB, line 314: This is already set at reset, but I feel we should do it before wwdtctl0 write so that the correct window is always set (and it will make a window 1 setup if added in the future less error prone)..

i509VCB (approved):
Other than comments I think its ready.

embassy-mspm0/src/wwdt.rs
bespsm: in the example from TI, they set it in this order. In the description it stated:
```
 *  The Window Watchdog Timer module allows you to switch between two different
 *  configured windows. This allows you to configure the watchdog with two
 *  different window sizes and switch between them during your application.
 ```
so it could be changed dynamically during the watchdog work

embassy-mspm0/src/wwdt.rs
bespsm: I see, fixed

bespsm: also rechecked it on HW, works :smiling_face_with_tear:

PR #4575, opened 2025-08-21
Bump `heapless` to v0.9

HaoboGu:
`heapless` just released v0.9.1, this PR bumps all `heapless` dependency to 0.9-compatible version


HaoboGu:
Need to wait for other dependencies like `smoltcp` updates their `heapless` version

PR #4581, opened 2025-08-23
OSPI RAM Support

jake-taf:
This PR adds support for more OctoSPI configurations (tested with APS6408 RAM).

## Changes

1. Makes DQSE and SIOO configurable instead of hardcoded

2. Makes the memory-mapped write instruction configurable (otherwise it's never set)

3. Fixes a bug where the address DTR was using the config for the instruction DTR instead of its own

4. Adds new functions for creating an OctoSPI with a DQS pin

## Backwards compatibility

DQSE for memory-mapped read / write are opposites, but default can only ever be one of them. so memory-mapped writes default to DQS being false unless configured. However, since the write instruction register and the DQS pin were never set, I don't think anyone was using memory-mapped writes or the DQS feature.


i509VCB:
~~#4496?~~

Nevermind confused me

PR #4583, opened 2025-08-25
stm32/adc/v3: added support for DMA based adc sampling

maor1993:
A lot of the code is based / copied from `ringbuffered_v2.rs`

I did make several changes I hope are accepted 

~1. V3 adc dma is not seperated to its own file to lower the amount of code duplication~
2. ADC dma will not halt/kill DMA on overrun, allowing the user the freedom to handle the error as they wish
3. User has access to `disable_adc` , which allows to pause ADC and DMA 

tested on stm32l412kb6u and stm32l432kc

### DIFF
--- a/embassy-stm32/src/adc/ringbuffered_v3.rs
+++ b/embassy-stm32/src/adc/ringbuffered_v3.rs
@@ -16,7 +16,7 @@ pub struct RingBufferedAdc<'d, T: Instance> {
 }
 
 impl<'d, T: Instance> RingBufferedAdc<'d, T> {
-    pub fn new(dma: Peri<'d, impl RxDma<T>>, dma_buf: &'d mut [u16]) -> Self {
+    pub(crate) fn new(dma: Peri<'d, impl RxDma<T>>, dma_buf: &'d mut [u16]) -> Self {
         //dma side setup
         let opts = TransferOptions {
             half_transfer_ir: true,


maor1993:
@Dirbaio , it seems the CI is stuck for a couple of hours, is this a known issue?

maor1993:
@Dirbaio Okay to merge?

maor1993:
hey @lulf , any reason not to merge this?

embassy-stm32/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: Updated stm32-metapac and stm32-data dependencies
 - feat: stm32/adc/v3: allow DMA reads to loop through enable channels
 - fix: Fix XSPI not disabling alternate bytes when they were previously enabled
+- feat: stm32/adc/v3: added support for Continous DMA configuration
```
jduranf: All continous should be changed to continuous

embassy-stm32/src/adc/v3.rs
```diff
@@ -12,6 +12,13 @@ pub use pac::adc::vals::{Ovsr, Ovss, Presc};
 use super::{
     blocking_delay_us, Adc, AdcChannel, AnyAdcChannel, Instance, Resolution, RxDma, SampleTime, SealedAdcChannel,
 };
+
+#[cfg(adc_v3)]
```
jduranf: To add g0 compatibility, use #[cfg(any(adc_v3, adc_g0))]

embassy-stm32/src/adc/v3.rs
```diff
@@ -12,6 +12,13 @@ pub use pac::adc::vals::{Ovsr, Ovss, Presc};
...
+mod ringbuffered_v3;
+
+#[cfg(adc_v3)]
```
jduranf: To add g0 compatibility, use #[cfg(any(adc_v3, adc_g0))]

embassy-stm32/src/adc/v3.rs
```diff
@@ -559,6 +566,107 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+    ///
+    /// The `dma_buf` should be large enough to prevent DMA buffer overrun.
+    /// The length of the `dma_buf` should be a multiple of the ADC channel count.
+    /// For example, if 3 channels are measured, its length can be 3 * 40 = 120 measurements.
+    ///
+    /// `read` method is used to read out measurements from the DMA ring buffer, and its buffer should be exactly half of the `dma_buf` length.
+    /// It is critical to call `read` frequently to prevent DMA buffer overrun.
+    ///
+    /// [`read`]: #method.read
+    #[cfg(adc_v3)]
```
jduranf: To add g0 compatibility, use #[cfg(any(adc_v3, adc_g0))]

embassy-stm32/src/adc/v3.rs
```diff
@@ -559,6 +566,107 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+                        w.set_sq(_i - 9, channel.channel());
+                    });
+                }
+                14..=15 => {
+                    T::regs().sqr4().modify(|w| {
+                        w.set_sq(_i - 14, channel.channel());
+                    });
+                }
+                _ => unreachable!(),
+            }
```
jduranf, line 666: Add those lines to add g0 compatibility:

#[cfg(any(adc_g0, adc_u0))]
{
    channel_mask |= 1 << channel.channel();
}

jduranf (commented):
Hi @maor1993,

To add compatibility in a STM32G0, some minor changes should be done. I tested those modifications in a STM32G071CBUx, and seems to work fine.

embassy-stm32/src/adc/v3.rs
maor1993: Done, but is the adc_u0 cfg mixed with adc_g0? since we're not adding the adc_u0 to the other parts?

jduranf: In this function, you are using #[cfg(any(adc_g0, adc_u0))] in other parts. So lets be coherent.

And if we want to enable this feature for the u0, I think we just have to change #[cfg(any(adc_v3, adc_g0, adc_u0))] for #[cfg(any(adc_v3, adc_g0, adc_u0))] in the beginning of this function.

maor1993: Okay, I see the why I'm confused. 

this PR #4674 
changed a lot of the surrounding functions in regards to configuration in adc_g0, so what used to be `#[cfg(any(adc_g0, adc_u0))]` is now completely different. 

do you suggest I implement the old style? or this new merged one? I don't have any hardware here with adc_g0 nor adc_u0 so I'm uncertain how to test that these configurations actually work...

jduranf: Sorry, you are right, I tested it without taking in account PR #4674

The same changes done in pub `async fn read` should be done in `pub fn into_ring_buffered`. This is the result:

```
    #[cfg(any(adc_v3, adc_g0))]
    pub fn into_ring_buffered<'a>(
        &mut self,
        dma: Peri<'a, impl RxDma<T>>,
        dma_buf: &'a mut [u16],
        sequence: impl ExactSizeIterator<Item = (&'a mut AnyAdcChannel<T>, SampleTime)>,
    ) -> RingBufferedAdc<'a, T> {
        assert!(!dma_buf.is_empty() && dma_buf.len() <= 0xFFFF);
        assert!(sequence.len() != 0, "Asynchronous read sequence cannot be empty");
        assert!(
            sequence.len() <= 16,
            "Asynchronous read sequence cannot be more than 16 in length"
        );
        // reset conversions and enable the adc
        Self::cancel_conversions();
        self.enable();

        //adc side setup

        // Set sequence length
        #[cfg(not(any(adc_g0, adc_u0)))]
        T::regs().sqr1().modify(|w| {
            w.set_l(sequence.len() as u8 - 1);
        });

        #[cfg(adc_g0)]
        {
            let mut sample_times = Vec::<SampleTime, SAMPLE_TIMES_CAPACITY>::new();

            T::regs().chselr().write(|chselr| {
                T::regs().smpr().write(|smpr| {
                    for (channel, sample_time) in sequence {
                        chselr.set_chsel(channel.channel.into(), true);
                        if let Some(i) = sample_times.iter().position(|&t| t == sample_time) {
                            smpr.set_smpsel(channel.channel.into(), (i as u8).into());
                        } else {
                            smpr.set_sample_time(sample_times.len(), sample_time);
                            if let Err(_) = sample_times.push(sample_time) {
                                panic!(
                                    "Implementation is limited to {} unique sample times among all channels.",
                                    SAMPLE_TIMES_CAPACITY
                                );
                            }
                        }
                    }
                })
            });
        }
        #[cfg(not(adc_g0))]
        {
            #[cfg(adc_u0)]
            let mut channel_mask = 0;

            // Configure channels and ranks
            for (_i, (channel, sample_time)) in sequence.enumerate() {
                Self::configure_channel(channel, sample_time);

                // Each channel is sampled according to sequence
                #[cfg(not(any(adc_g0, adc_u0)))]
                match _i {
                    0..=3 => {
                        T::regs().sqr1().modify(|w| {
                            w.set_sq(_i, channel.channel());
                        });
                    }
                    4..=8 => {
                        T::regs().sqr2().modify(|w| {
                            w.set_sq(_i - 4, channel.channel());
                        });
                    }
                    9..=13 => {
                        T::regs().sqr3().modify(|w| {
                            w.set_sq(_i - 9, channel.channel());
                        });
                    }
                    14..=15 => {
                        T::regs().sqr4().modify(|w| {
                            w.set_sq(_i - 14, channel.channel());
                        });
                    }
                    _ => unreachable!(),
                }

                #[cfg(adc_u0)]
                {
                    channel_mask |= 1 << channel.channel();
                }
            }

            // On G0 and U0 enabled channels are sampled from 0 to last channel.
            // It is possible to add up to 8 sequences if CHSELRMOD = 1.
            // However for supporting more than 8 channels alternative CHSELRMOD = 0 approach is used.
            #[cfg(adc_u0)]
            T::regs().chselr().modify(|reg| {
                reg.set_chsel(channel_mask);
            });
        }
        // Set continuous mode with Circular dma.
        // Clear overrun flag before starting transfer.
        T::regs().isr().modify(|reg| {
            reg.set_ovr(true);
        });

        #[cfg(not(any(adc_g0, adc_u0)))]
        T::regs().cfgr().modify(|reg| {
            reg.set_discen(false);
            reg.set_cont(true);
            reg.set_dmacfg(Dmacfg::CIRCULAR);
            reg.set_dmaen(true);
        });
        #[cfg(any(adc_g0, adc_u0))]
        T::regs().cfgr1().modify(|reg| {
            reg.set_discen(false);
            reg.set_cont(true);
            reg.set_dmacfg(Dmacfg::CIRCULAR);
            reg.set_dmaen(true);
        });

        RingBufferedAdc::new(dma, dma_buf)
    }
```

Tested in my STM32G0 and works. I think enabling this feature for STM32U0 should be safe, but I haven't any hardware to check it.

maor1993: done + some fixes

maor1993:
@jduranf can you merge?

jduranf:
> @jduranf can you merge?

Sorry, no. I'm like you, just wanting this PR to be merged

maor1993:
@Dirbaio , this PR is open for over a month, passes all checks and required by other users, is there a reason not merge this?

embassy-stm32/src/adc/v3.rs
```diff
@@ -559,6 +566,137 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+        // Set continuous mode with Circular dma.
+        // Clear overrun flag before starting transfer.
+        T::regs().isr().modify(|reg| {
+            reg.set_ovr(true);
+        });
+
+        #[cfg(not(any(adc_g0, adc_u0)))]
+        T::regs().cfgr().modify(|reg| {
+            reg.set_discen(false);
+            reg.set_cont(true);
```
Jacke-debug, line 685: Using continuous and circular mode means you sample as fast as possible for the ADC. This means that the DMA buffer fills up really quickly and care must be taken when setting up the task that reads from the buffer. 

As an alternative approach I think you should consider using triggered measurements to control the rate at which new samples are taken. I have started working on this for the g4 series #4786 but it would be good to develop similar APIs for all versions in the long run.

maor1993: Yes, getting a fast as possible buffer was the point.
this is intended for audio applications where you want a constant stream of samples

in my case I needed to sample 4 analog inputs as fast as possible and only pass to a different task samples that pass some threshold.

as such, by setting the sequence to the highest sample time I was able to get super fast samples without needing to retrigger the ADC, essentially acting as a clock source for the audio pipeline 

while the device I used, **stm32l412kb** does not support changing the clock to the ADC, other devices do.

alternatively you can also drive the conversion starts from a timer, as such this will ensure the ADC is always ready to provide samples and the relevant task will be awaken every time the DMA buffer is half full.

xoviat:
I've not been around for a while, but barring objections from @lulf and @Dirbaio I'm going to try to get these ADC pull requests merged. For me, ideally I would like to see a consistent interface across all versions. So let's see what we have here:

This version of adc ringbuffer:

- pub stop_continuous_sampling
- pub disable_adc
- pub teardown_adc
- pub async read
- pub blocking_read
- pub adc::into_ring_buffered
- pub new
- start_continuous_sampling

Existing version of adc ringbuffer:

- pub set_sample_sequence
- pub start
- pub teardown_adc
- pub blocking_read
- pub async read
- pub adc::into_ring_buffered
- is_on
- stop_adc
- start_adc
- set_channel_sample_time
- set_channels_sample_time
- stop
- setup_adc

Just from a glance, I prefer what you've done, allowing the sequence to be passed to the constructor, rather than set as a method. To be honest, the existing ringbuffer seems like an attempt to work around the under-developed v2 mod.

The largest objection that I have is the public new constructor for the ringbuffer. There should be one way to obtain a ringbuffer, `into_ring_buffer`. The other objections are not really your fault, because they are a product of the existing ringbuffer_v2.

If you remove the `pub new` and one one objects, then in a few days I will merge this. I will then need to work to merge the ringbuffered.rs files, mostly by working on the existing v2.

maor1993:
> I've not been around for a while, but barring objections from @lulf and @Dirbaio I'm going to try to get these ADC pull requests merged. For me, ideally I would like to see a consistent interface across all versions. So let's see what we have here:
> 
> This version of adc ringbuffer:
> 
>     * pub stop_continuous_sampling
> 
>     * pub disable_adc
> 
>     * pub teardown_adc
> 
>     * pub async read
> 
>     * pub blocking_read
> 
>     * pub adc::into_ring_buffered
> 
>     * pub new
> 
>     * start_continuous_sampling
> 
> 
> Existing version of adc ringbuffer:
> 
>     * pub set_sample_sequence
> 
>     * pub start
> 
>     * pub teardown_adc
> 
>     * pub blocking_read
> 
>     * pub async read
> 
>     * pub adc::into_ring_buffered
> 
>     * is_on
> 
>     * stop_adc
> 
>     * start_adc
> 
>     * set_channel_sample_time
> 
>     * set_channels_sample_time
> 
>     * stop
> 
>     * setup_adc
> 
> 
> Just from a glance, I prefer what you've done, allowing the sequence to be passed to the constructor, rather than set as a method. To be honest, the existing ringbuffer seems like an attempt to work around the under-developed v2 mod.
> 
> The largest objection that I have is the public new constructor for the ringbuffer. There should be one way to obtain a ringbuffer, `into_ring_buffer`. The other objections are not really your fault, because they are a product of the existing ringbuffer_v2.
> 
> If you remove the `pub new` and one one objects, then in a few days I will merge this. I will then need to work to merge the ringbuffered.rs files, mostly by working on the existing v2.

Thanks for the effort, 

will `pub(crate) new` suffice? I'm uncertain how to implement `adc::into_ring_buffer` otherwise

xoviat:
> will pub(crate) new suffice? I'm uncertain how to implement adc::into_ring_buffer otherwise

yes, that's fine.

maor1993:
> > will pub(crate) new suffice? I'm uncertain how to implement adc::into_ring_buffer otherwise
> 
> yes, that's fine.

👍🏾  done

lulf:
@xoviat Thanks for reviewing and merging, feel free to pick up the stm32 PRs that have been around for a while, I don't have much bandwidth or technical insight into stm32 to  follow up on the more complex ones.

xoviat:
@maor1993 @jduranf can you discuss on matrix? 

https://app.element.io/#/room/#embassy-rs:matrix.org

PR #4584, opened 2025-08-25
Feat: embassy-usb mass storage class

elpiel:
Hello,
I've decided to take a stab at the Mass storage class and open a work-in-progress PR to collect some feedback.
I still haven't tested it, just finished implementing this part and since I haven't implemented any other usb class before, feedback is much appreciated!


Murmele:
I currently use a fork of [this](https://github.com/embassy-rs/embassy/pull/1242) version. It was good for my purpose. I just has to extend `scsi/block_device.rs` to support multiblock read/write. It is quite hacky and I don't like it but it works.

The reason was that sending multiple blocks at once is faster than sending block by block. The problem is that you might not have enough memory to store multiple blocks in the ram (Scenario: USB Mass storage + SD Card reading).
1) So I am starting a multiblock write.
2) Read a complete block (512Byte) from the sd card.
3) Send it out over usb (always 64byte chunks)
4) Start again from 2 until all desired blocks are send

My changes can be found at https://github.com/Murmele/embassy branch `wip`

elpiel:
> I currently use a fork of [this](https://github.com/embassy-rs/embassy/pull/1242) version. It was good for my purpose. I just has to extend `scsi/block_device.rs` to support multiblock read/write. It is quite hacky and I don't like it but it works.
> 
> The reason was that sending multiple blocks at once is faster than sending block by block. The problem is that you might not have enough memory to store multiple blocks in the ram (Scenario: USB Mass storage + SD Card reading).
> 
>     1. So I am starting a multiblock write.
> 
>     2. Read a complete block (512Byte) from the sd card.
> 
>     3. Send it out over usb (always 64byte chunks)
> 
>     4. Start again from 2 until all desired blocks are send
> 
> 
> My changes can be found at https://github.com/Murmele/embassy branch `wip`

Thank you for informing me about the PR.
I've also found a good compromise of 4 KB (8 blocks) to be a very good buffer size for multi-block writes.

I need to check how it's implemented in your branch but I'd love to try it out. I've been wanting this feature for so long... especially coupled with custom "filesystem" of sorts for sequential and very fast log writing.

I'll close this PR for now as the other one seems much more complete and this one hasn't been tested in any way.

PR #4585, opened 2025-08-25
Replace nrf5340's SAADC input pins with correct mappings

dempfi:
Updated the SAADC pins to correct ones per [nordic's docs](https://docs.nordicsemi.com/bundle/ps_nrf5340/page/chapters/pin.html)


Dirbaio:
Can you add a changelog entry?

dempfi:
@Dirbaio updated

PR #4586, opened 2025-08-25
Introduce trait to configure clock when creating ADC

per42:
Required on STM32WL with default HAL initialization.

The trait is only implemented for adc_g0, but all
that have clock config should add implementations.


embassy-stm32/src/adc/mod.rs
```diff
@@ -51,6 +51,41 @@ pub struct Adc<'d, T: Instance> {
...
+pub enum CkModePclk {
+    DIV1 = 3,
+    DIV2 = 1,
+    DIV4 = 2,
+}
+
+/// Asynchronous ADCCLK prescaler
+/// * ADC_CCR:PRESC in STM32WL5x
+#[repr(u8)]
+pub enum Presc {
```
Dirbaio: do not duplicate enums, just reexport them from the PAC

embassy-stm32/src/adc/mod.rs
```diff
@@ -51,6 +51,41 @@ pub struct Adc<'d, T: Instance> {
...
+    DIV128,
+    DIV256,
+}
+
+/// Create ADC with explicit clock for the analog ADC
+pub trait AdcWithClock<'d, T, U>
+where
+    T: Instance,
+{
+    fn new(adc: crate::Peri<'d, T>, clock: U) -> Self;
```
Dirbaio: this doesn't need a trait, this `fn new` could just be a regular function in the Adc struct.

The reason a trait is worse is it's harder to use: it's more hidden in the docs, users need to import the trait to call the function.

per42: The purpose was to have a shared interface with all ADC peripheral versions. `new()` may be the least valuable - some non-v3 ADCs may have the same clock source and prescaler pair - but I aim to extend the interface with methods such as `read()` in other traits. Perhaps only using traits for methods is the best compromise?

Dirbaio: There's advantages in using traits as you mention, and there's disadvantages as I mention. However the decision to not use traits *has already been made*: the entire Embassy API surface uses regular methods, it never uses traits this way to extract common methods across versions. There's no reason ADC is special to warrant doing something different, and it'd hurt consistency.

embassy-stm32/src/adc/mod.rs
per42: Can you help me find it? I thought that register values' types would be according to the register's methods:

stm32-metapac/src/peripherals/adc_g0.rs:
```rust
impl Ccr {
    ...
    pub const fn presc(&self) -> u8 {...}
```
which in this case is `u8`. I only got the enumeration from the reference manual. I browsed the `pac::adc*::vals` and found nothing for presc. Perhaps it is missing in the stm32-data-sources?

PR #4589, opened 2025-08-26
stm32/sai: make NODIV independent of MCKDIV

tomaz-suller:
These values should clearly be made independent, at least in the WB55 microcontroller (and so I suspect in all the others) since NODIV implies no divider between the master clock and the bit clock, but it does NOT imply no divider between the SAI clock and the master clock. Quoting from the WB55 reference manual (RM0434 Rev 14, p. 1219):

> #### Clock generator programming when NODIV = 1
> When MCKDIV is different from 0, the frequency of the bit clock (SCK\_x) is given in the
> formula below:
> $`F_{SCK\_x} = F_{MCLK\_x} = \frac{F_{sai\_x\_ker\_ck}}{MCKDIV}`$

Tests I have performed with the WB55 show you can use `NODIV = 1` with `MCKDIV != disabled` and still get valid configurations and correct clock frequencies.
I'm not entirely sure about integrating this change though since CubeMX follows the same behaviour as Embassy currently (i.e. NODIV == MCKDIV disabled).

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- feat: stm32/sai: make NODIV independent of MCKDIV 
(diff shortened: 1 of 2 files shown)


tomaz-suller:
@lulf I had to merge the changelog and the auto-merge got disabled again

PR #4593, opened 2025-08-26
chore: prepare embassy crate releases

lulf:


lulf:
Could I get a quick look over this @bugadani ?

cyw43-pio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cyw43-pio"
-version = "0.6.0"
+version = "0.6.1"
```
bugadani: Shouldn't this be 0.7? If someone is using `embassy-rp = 0.7` and `cyw43-pio = 0.6`, they will be auto-updated to this version and will need to update embassy-rp, too.

lulf: I think the root cause of this is a variant of https://github.com/obi1kenobi/cargo-semver-checks/issues/638. The rustdoc and types generated for the cyw43-pio crate depending on embassy-rp 0.7 and embassy-rp 0.8 is identical.

So I think for now, the safest approach we can take in a releaser tool is to propagate minor version bumps in all dependencies, even if it's not strictly necessary. Another alternative would be to for each dependent version, manually parse the rustdoc and look for types from the 'current' crate in the rustdoc for the dependents, and in that case decide to do a minor bump if the 'current' is minor bumped. It would probably reduce the number of false bumps a bit.

(cc @Dirbaio)

PR #4595, opened 2025-08-26
Add I2S support for ST chips with spi_v2 hardware

pthien:
[Problem]
STM32 chips with spi_v2 hardware support I2S, but it was not previously enabled.

[Solution]
- Enabled is2 support for spi_v2 hardware. Note: this depends on register fixes in stm32-data.
- Added an example based on embassy\examples\stm32f4\src\bin\i2s_dma.rs

[Testing]
I ran the example on a NUCLEO-G0B1RE with an I2S dac and heard the correct tones.

Note: This depends on [this PR](https://github.com/embassy-rs/stm32-data/pull/663) in stm32-data


Dirbaio:
closing in favor of #4620. I ended up adding versions in stm32-data for chips with and without i2s so we can match on those in embassy-stm32 instead of having to encode which chips have it with cfgs.

PR #4598, opened 2025-08-28
embassy-sync: Don't drop wakers in Signal::reset

0e4ef622:
Dropping wakers is a recipe for hangs, as was discovered on [matrix](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$l_xv_ruxW4G3yWKAC6g2cVQYZiCBqcfY_9204R9lo4c?via=matrix.org&via=beeper.com&via=mozilla.org).

Might be worth deprecating and removing the method altogether since `try_take` has the same effect.


lulf (approved):
Nice, thank you!

PR #4599, opened 2025-08-28
fix(embassy-stm32): Prevent dropped DacChannel from disabling Dac peripheral if another DacChannel is still in scope

### LINKED ISSUE #4577: embassy-stm32 bug: splitting the DAC in main renders individual channels unusable in tasks
Originally [reported in matrix](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$mBmZqJXjr_lEm1IVs30xEDHfM9bE6LJihrT3FUF0xfs?via=matrix.org&via=beeper.com&via=tchncs.de) as follows:

> Hello. I'm getting different behavior from the DAC on an STM32 F-series depending on whether the code runs inside main or inside a task. I created a simple app to demonstrate. In [this commit](https://github.com/universalhandle/embassy_potential_dac_bug/commit/375bdca2c90b5da4ee008e34c5fbd339d79129d0), I move the DAC logic out of main; it is at this point that the info logs begin to defy expectations (e.g., "DAC should have been set to 3025; reading 0"). Would someone mind having a look? Not sure if I've missed something about how to properly spawn tasks, or if I've found a bug.

@adamgreig identified the problem: "each channel has a Drop impl that disables the DAC peripheral, but that means in your case when main exits it drops DAC1CH2 which disables DAC1 entirely, then the dac task tries to use DAC1CH1 without luck." Reading between the lines, it seems my example app may have uncovered an edge case. I'm inferring that this problem only occurs when the DAC is split in main and one of the channels is allowed to drop, but that if the other channel is assigned to a task the DAC remains usable. I'll spend a little time exploring under exactly what conditions the breakage occurs and report back here.

He also suggested a workaround: "if you move the Dac object to the dac task before you split it you should be fine."

universalhandle:
Fix #4577 by counting references to DacChannel. Modeled after similar code in the `can` module.

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -12,6 +12,7 @@ use crate::{peripherals, Peri};
 
 mod tsel;
 use embassy_hal_internal::PeripheralType;
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
 pub use tsel::TriggerSel;
 
 /// Operating mode for DAC channel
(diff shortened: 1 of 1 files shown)


universalhandle:
As I'm re-reviewing these changes, [these lines](https://github.com/embassy-rs/embassy/blob/76d4a567466d25c296173a47aa067d44247f6642/embassy-stm32/src/dac/mod.rs#L578-L580) have me wondering if I've introduced a second reference counter rather than fixing a bug in one I hadn't previously noticed. Perhaps someone could help me understand what's going on there? I'm New Here™ and welcome close reviews (I don't want to contribute bad code!) and constructive criticism (I've already learned a lot looking into this bug and welcome the opportunity to learn more). Thanks!

Dirbaio:
we used to have builtin refcounts in the rcc module for uart split but it was removed here https://github.com/embassy-rs/embassy/commit/183f2f6913032600d74ea058b50a1fcedbebe719

It seems that was never properly wired up for DAC.

Anyway now that it's removed it's easier to add refcounting in the DAC driver itself like USART does.

universalhandle:
> we used to have builtin refcounts in the rcc module for uart split but it was removed here [183f2f6](https://github.com/embassy-rs/embassy/commit/183f2f6913032600d74ea058b50a1fcedbebe719)

I see. In that case, does it make sense to remove [these two lines](https://github.com/embassy-rs/embassy/blob/76d4a567466d25c296173a47aa067d44247f6642/embassy-stm32/src/dac/mod.rs#L578-L579) to tidy things up? Happy to make the PR if you think so.

Dirbaio:
ah yes please do. the comment is a lie now

Issue #3741, opened 2025-01-08
Don't reset SYSCFG peripherals or switch system clock in embassy_stm32::init


meowcakes:
If the bootloader uses OSPI to memory map external flash and/or RAM (e.g. if the application binary is too large for embedded flash), calling `embassy_stm32::init` breaks it due to two reasons (that I am able to tell):
- SYSCFG is reset here https://github.com/embassy-rs/embassy/blob/06869e2e85695f4528d1148dfd81700c905771c5/embassy-stm32/src/lib.rs#L518
- The system clock is switched here https://github.com/embassy-rs/embassy/blob/06869e2e85695f4528d1148dfd81700c905771c5/embassy-stm32/src/rcc/h.rs#L419

PR #4602, opened 2025-08-29
Fix stm32h7rs init when using external flash via XSPI

meowcakes:
Addresses https://github.com/embassy-rs/embassy/issues/3741

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -431,13 +431,10 @@ pub(crate) unsafe fn init(config: Config) {
     while !RCC.cr().read().hsirdy() {}
 
     #[cfg(stm32h7rs)]
-    let (xspi1sel, xspi2sel) = {
-        // Save XSPI clock source settings and switch the clock source so it will use HSI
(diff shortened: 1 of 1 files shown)


lulf (commented):
Could you squash and add a more descriptive commit message? 😅

meowcakes:
> Could you squash and add a more descriptive commit message? 😅

Done

embassy-stm32/src/rcc/h.rs
```diff
@@ -678,6 +688,13 @@ pub(crate) unsafe fn init(config: Config) {
 
     config.mux.init();
 
+    #[cfg(stm32h7rs)]
+    {
+        // Set the XSPI clock source back to what it was originally
+        RCC.ahbperckselr().modify(|w| w.set_xspi1sel(xspi1sel));
```
Dirbaio: this will make it so the user's settings in `config.rcc.mux.xspi[12]sel` are ignored, is this what we really want?

meowcakes: I wasn't even aware of that config setting. You're right, it would be better to let it get set to whatever is set in the config. Fixed.

PR #4605, opened 2025-08-29
MSPM0: add i2c target implementation

Iooon:
Implement the i2c target interface from https://github.com/embassy-rs/embassy/tree/main/embassy-rp for mspm0.
In the interface we define a listen function that is used to receive the commands from the i2c controller asynchronously so that the user can handle them accordingly. 
Currently the implementation heavily favors async because of the easier implementation with interrupts.

The example was tested on the dev. board: MSPM0G3507
Sadly my MSPM0L1306 board is broken at the moment, but the example should work exactly the same. So if needed I can add the example to the L1306 , but someone else would need to test it.

### DIFF
--- a/embassy-mspm0/src/i2c.rs
+++ b/embassy-mspm0/src/i2c.rs
@@ -56,19 +56,6 @@ pub enum ClockDiv {
 }
 
 impl ClockDiv {
-    fn into(self) -> vals::Ratio {
-        match self {
-            Self::DivBy1 => vals::Ratio::DIV_BY_1,
-            Self::DivBy2 => vals::Ratio::DIV_BY_2,
-            Self::DivBy3 => vals::Ratio::DIV_BY_3,
-            Self::DivBy4 => vals::Ratio::DIV_BY_4,
-            Self::DivBy5 => vals::Ratio::DIV_BY_5,
-            Self::DivBy6 => vals::Ratio::DIV_BY_6,
-            Self::DivBy7 => vals::Ratio::DIV_BY_7,
-            Self::DivBy8 => vals::Ratio::DIV_BY_8,
-        }
-    }
-
     fn divider(self) -> u32 {
         match self {
             Self::DivBy1 => 1,
@@ -79,6 +66,21 @@ impl ClockDiv {
             Self::DivBy6 => 6,
             Self::DivBy7 => 7,
             Self::DivBy8 => 8,
+        }
+    }
+}
+
+impl From<ClockDiv> for vals::Ratio {
+    fn from(value: ClockDiv) -> Self {
+        match value {
+            ClockDiv::DivBy1 => Self::DIV_BY_1,
+            ClockDiv::DivBy2 => Self::DIV_BY_2,
+            ClockDiv::DivBy3 => Self::DIV_BY_3,
+            ClockDiv::DivBy4 => Self::DIV_BY_4,
+            ClockDiv::DivBy5 => Self::DIV_BY_5,
+            ClockDiv::DivBy6 => Self::DIV_BY_6,
+            ClockDiv::DivBy7 => Self::DIV_BY_7,
+            ClockDiv::DivBy8 => Self::DIV_BY_8,
         }
     }
 }
@@ -133,6 +135,11 @@ pub enum ConfigError {
     ///
     /// The clock soure is not enabled is SYSCTL.
     ClockSourceNotEnabled,
+
+    /// Invalid target address.
+    ///
+    /// The target address is not 7-bit.
+    InvalidTargetAddress,
 }
 
 #[non_exhaustive]
@@ -140,7 +147,7 @@ pub enum ConfigError {
 /// Config
 pub struct Config {
     /// I2C clock source.
-    clock_source: ClockSel,
+    pub(crate) clock_source: ClockSel,
 
     /// I2C clock divider.
     pub clock_div: ClockDiv,
@@ -159,6 +166,12 @@ pub struct Config {
 
     /// Set the pull configuration for the SCL pin.
     pub bus_speed: BusSpeed,
+
+    /// 7-bit Target Address
+    pub target_addr: u8,
+
+    /// Control if the target should ack to and report general calls.
+    pub general_call: bool,
 }
 
 impl Default for Config {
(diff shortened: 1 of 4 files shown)


embassy-mspm0/src/i2c_target.rs
```diff
@@ -0,0 +1,435 @@
...
+/// Status after responding to a controller read request.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ReadStatus {
+    /// Transaction completed successfully. The controller either NACKed the last byte
+    /// or sent a STOP condition.
+    Done,
+    /// Transaction Incomplete, controller trying to read more bytes than were provided
+    NeedMoreBytes,
+    /// Transaction Complere, but controller stopped reading bytes before we ran out
```
i509VCB:
```suggestion
    /// Transaction complete, but controller stopped reading bytes before we ran out
```

embassy-mspm0/src/i2c_target.rs
```diff
@@ -0,0 +1,435 @@
...
+}
+
+/// Status after responding to a controller read request.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ReadStatus {
+    /// Transaction completed successfully. The controller either NACKed the last byte
+    /// or sent a STOP condition.
+    Done,
+    /// Transaction Incomplete, controller trying to read more bytes than were provided
```
i509VCB:
```suggestion
    /// Transaction incomplete, controller trying to read more bytes than were provided
```

embassy-mspm0/src/i2c.rs
```diff
@@ -159,6 +166,12 @@ pub struct Config {
 
     /// Set the pull configuration for the SCL pin.
     pub bus_speed: BusSpeed,
+
+    /// 7-bit Target Address
+    pub target_addr: u8,
```
i509VCB: Do the other hals add `target_addr` as a field for target mode? Or do the other hals make a new config type?

embassy-mspm0/src/i2c.rs
```diff
@@ -83,6 +70,21 @@ impl ClockDiv {
...
+            ClockDiv::DivBy3 => Self::DIV_BY_3,
+            ClockDiv::DivBy4 => Self::DIV_BY_4,
+            ClockDiv::DivBy5 => Self::DIV_BY_5,
+            ClockDiv::DivBy6 => Self::DIV_BY_6,
+            ClockDiv::DivBy7 => Self::DIV_BY_7,
+            ClockDiv::DivBy8 => Self::DIV_BY_8,
+        }
+    }
+}
+
```
i509VCB: I think we did this as an associated function rather than a trait impl to avoid making `From<ClockDiv> for vals::Ratio` public API. Probably should revert this part.

examples/mspm0g3507/src/bin/i2c_target.rs
```diff
@@ -0,0 +1,60 @@
...
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) -> ! {
+    let p = embassy_mspm0::init(Default::default());
+
+    let instance = p.I2C1;
+    let scl = p.PB2;
+    let sda = p.PB3;
+
+    let mut config = i2c::Config::default();
+    config.general_call = true;
```
i509VCB: I think we should explicitly set the target address here, even if redundant for the sake of making the examples helpful.

charlesbmi:
Just dropping in to say thanks for writing this; since there was a comment about not yet testing on MSPM0L1306, which I'm using, I tried it out on an LP-MSPM0L1306 and it worked after making basic pin-revisions:
* Peripheral: I2C1 → I2C0
* Pins: SCL PB2 → PA1, SDA PB3 → PA0
* Interrupt: I2C1 => InterruptHandler<I2C1> → I2C0 => InterruptHandler<I2C0>

Posting screenshot of aardvark + logs in case it's helpful
image

Iooon:
> Just dropping in to say thanks for writing this; since there was a comment about not yet testing on MSPM0L1306, which I'm using, I tried it out on an LP-MSPM0L1306 and it worked after making basic pin-revisions:
> 
> * Peripheral: I2C1 → I2C0
> * Pins: SCL PB2 → PA1, SDA PB3 → PA0
> * Interrupt: I2C1 => InterruptHandler → I2C0 => InterruptHandler
> 
> Posting screenshot of aardvark + logs in case it's helpful image

Perfect, thank you. I have added the example for the l130 and changed it accordingly.

examples/mspm0g3507/src/bin/i2c_target.rs
Iooon: Done

embassy-mspm0/src/i2c.rs
Iooon: They make a new config type. If we do this we need to think how we handle the clock setup for i2c target. Because it is still very similar to i2c controller and uses the implemented helper functions to set it up.

i509VCB: The one argument I could see for sharing config types is the loopback mode, but having to specify clock info twice would be annoying there.

We can address this later of course.

Iooon: We could do a target config with a new function and create an i2c controller config to configure the clock.
```rust
pub struct Config {
    controller_config: i2c::Config,
    pub target_addr: u8,
    ...
}

impl Config {
    pub fn new(clock_source: ClockSel, clock_div: ClockDiv, target_addr: u8, ...) -> Self {
    }
}
```

i509VCB: Giving the fields a look, what about making the controller config an extra type?

`i2c::Config` only involves clock setup related things at the moment. `i2c_target::Config` would only additionally need to be specified for target setups.

So something like:
```rust
impl I2cTarget<'_> {
    pub fn new(..., config: &i2c::Config, target_config: &i2c_target::Config) -> Result<Self> { ... } 
    
    pub fn set_config(&mut self, config: &i2c::Config, target_config: &i2c_target::Config) -> Result<()>
}
```

This does result in a funny case like uart where the tx/rx ends (controller and target in this case) are independently configurable but share clock state, that is probably fine.

Although that makes the `SetConfig` trait a bit fun... As you'd have a tuple type to set config.

Iooon: Sounds good for now to me. Maybe we could even split the i2c controller config into a clock config and controller specific stuff later.

PR #4606, opened 2025-08-29
Taskmeta update

diondokter:
Now rebased instead of merged (so this PR is in favor of #4604)

I've updated the branch to be mergeable again.
This is essentially an updated #4383 but due to the rebase, I don't know how to cleanly create a PR to get it in that PR.

### DIFF
--- a/.github/ci/test.sh
+++ b/.github/ci/test.sh
@@ -12,7 +12,7 @@ export CARGO_TARGET_DIR=/ci/cache/target
 # used when pointing stm32-metapac to a CI-built one.
 export CARGO_NET_GIT_FETCH_WITH_CLI=true
 
-cargo test --manifest-path ./embassy-executor/Cargo.toml
+cargo test --manifest-path ./embassy-executor/Cargo.toml --features metadata-name
 cargo test --manifest-path ./embassy-futures/Cargo.toml
 cargo test --manifest-path ./embassy-sync/Cargo.toml
 cargo test --manifest-path ./embassy-embedded-hal/Cargo.toml
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ async fn main(spawner: Spawner) {
     let p = embassy_nrf::init(Default::default());
 
     // Spawned tasks run in the background, concurrently.
-    spawner.spawn(blink(p.P0_13.into())).unwrap();
+    spawner.spawn(blink(p.P0_13.into()).unwrap());
 
     let mut button = Input::new(p.P0_11, Pull::Up);
     loop {
--- a/ci.sh
+++ b/ci.sh
@@ -31,7 +31,9 @@ cargo batch \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv6m-none-eabi --features defmt,arch-cortex-m,executor-thread,executor-interrupt \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,trace \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,rtos-trace \
+    --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,trace,rtos-trace \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,executor-thread \
     --- build --release --manifest-path embassy-executor/Cargo.toml --target thumbv7em-none-eabi --features arch-cortex-m,executor-interrupt \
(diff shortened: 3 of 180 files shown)


diondokter:
So, I want to see this merged because I want #4035 merged.

I have time to do the work. I just need to know what needs to be done.
The branch this PR is based on doesn't mention any blockers. So hopefully we can get it in?
Then I can update the EDF scheduler to use the new metadata to put the deadlines in.

broose-goose:
Just wondering, why was must_spawn removed?

lulf:
> Just wondering, why was must_spawn removed?

The spawn token is now a Result which means that instead of having 2 APIs like before, you only need 1:

```
s.must_spawn(foo());
s.spawn(foo()).unwrap();
```

Now becomes

```
s.spawn(foo().unwrap());
```

0x53A:
This PR broke compatibility between latest master of embassy and latest master of esp-hal, because esp-hal is using ``must_spawn`` inside the ``#[esp_hal_embassy::main]`` procmacro: https://github.com/esp-rs/esp-hal/blob/fedc9fc5dfd6b50b3fab2b02eb43ff375c02d3e5/esp-hal-procmacros/src/embassy.rs#L171

Do you think it could be re-introduced, until esp-hal is updated?

diondokter:
@0x53A Is there a reason why you're on the git version of the executor? The esp-hal uses the released 0.9 version and that should still work fine

0x53A:
Not a good one, I'm trying to use [embassy/trouble](https://github.com/embassy-rs/trouble), which is still heavily work in progress, and _thought_ using the latest version of everything would be the easiest way to get going.

diondokter:
Looks like trouble uses the released versions of the crates, so you should be good with using those

0x53A:
yes that works, thank you.

~~One last question, I did see a ``LoadProhibited`` exception when using embassy @ df10e8a6bc407544d29c234ed00bdec3e9caf837, is that something I should investigate a little bit, or is it kind of expected that it just randomly crashes with potentially mismatching versions? I was using that commit, because it's the last version with the old ``must_spawn``.~~

It works with the latest master, so please ignore that.

<details>

```
[rprintln] Hello, world!
21:40:50.926: [INFO] esp32_bluetooth_test: [log::info] Hello, world!
21:40:50.926: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: Our address = FF:E4:05:1A:8F:FF
21:40:50.926: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: Starting advertising and GATT service
21:40:50.926: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: [adv] advertising
21:41:03.026: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: [adv] connection established
21:41:03.026: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: [custom_task] notifying connection of tick 1
21:41:03.026: [INFO] esp32_bluetooth_test::ble_bas_peripheral_sec: [custom_task] RSSI: 0
21:41:03.166: panicked at C:\Users\Lukas\.cargo\git\checkouts\esp-hal-29ef8683203c9b81\fedc9fc\esp-hal\src\exception_handler\mod.rs:10:5:
21:41:03.166: 
21:41:03.166: 
21:41:03.166: Exception occurred on ProCpu 'LoadProhibited'
21:41:03.166: Context { PC: 1107389965, PS: 395824, A0: 2181131511, A1: 1070446160, A2: 4294967293, A3: 255, A4: 17421496, A5: 3, A6: 4294967295, A7: 1070446160, A8: 2181131863, A9: 1070437248, A10: 17421496, A11: 255, A12: 4294967040, A13: 4294967293, A14: 1070162844, A15: 1070437248, SAR: 4, EXCCAUSE: 28, EXCVADDR: 17421496, LBEG: 1074097928, LEND: 1074097938, LCOUNT: 0, THREADPTR: 0, SCOMPARE1: 256, BR: 0, ACCLO: 0, ACCHI: 0, M0: 0, M1: 0, M2: 0, M3: 0, F64R_LO_CPENABLE: 0, F64R_HI: 0, F64S: 0, FCR: 0, FSR: 0, F0: 0, F1: 0, F2: 0, F3: 0, F4: 0, F5: 0, F6: 0, F7: 0, F8: 0, F9: 0, F10: 0, F11: 0, F12: 0, F13: 0, F14: 0, F15: 0 }
```

image

image

</details>

diondokter:
That's a trouble or esp-hal problem and I don't know which one. Probably best to ask there or in one of the chat rooms

bugadani:
Just FYI you're not obligated to use `esp_hal_embassy::main`, you can write code that uses the correct APIs. esp-hal will only ever support released versions of embassy-executor. If 0.10 drops soon, we'll update. Otherwise you'll have to live with this incompatibility.

hsteinhaus:
I see, that this change has happened long ago, but I just got surprised by the removal of `must_spawn` while updating my application. Obviously the removal happened without a notice in changelog and no prior deprecation.

I actually considered the notation of `must_spawn` very clean and idiomatic, as there is rarely an alternative choice in an embedded application, if an important task fails spawn. Now, a hell of `.unwrap()` calls broke lose in my main, as well in embassy's own examples. Could you help me to understand this decision, e.g. with a link to more context?

lulf:
@hsteinhaus https://github.com/embassy-rs/embassy/issues/5766

Agreed changelog entry was a bit vague and could be clearer on the API change.

PR #4608, opened 2025-08-29
[embassy-executor]: Upstream "Earliest Deadline First" Scheduler (version 2)

diondokter:
An updated #4035 that makes use of the new metadata API.

The example output is still good.

I've also retested with #4374 and these are the numbers:
```txt
cycles per poll
              nrf52832
master         78
edf            78
edf enabled    91
```

I don't have a nice rp2040 to test with...


examples/nrf52840-edf/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+edition = "2021"
+name = "embassy-nrf52840-edf-examples"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
```
lulf, line 5: Please add a publish = false here to prevent it from being picked up by embassy-devtool

lulf (commented):
Could you add a build combination for edf-scheduler in [package.metadata.embassy] as well? This will make the release tool check it for semver break as well.

diondokter:
@lulf like this?

lulf:
@diondokter Yup, that looks good for the example. Also add this to the same section in embassy-executor/Cargo.toml:

```
[package.metadata.embassy]
build = [
...
    {target = "thumbv7em-none-eabi", features = ["arch-cortex-m", "executor-interrupt", "executor-thread", "edf-scheduler"]},
...
]
```

jamesmunns:
Updated numbers using the `bench2` test from #4374:

<details><summary>Details</summary>
<p>

## Main

```
Test OK, count=109655, cycles=15300
  R0: 00000001   R1: 200200e0   R2: 00000000   R3: 20020249
  R4: 0001ac57   R5: 00000004   R6: 00000000   R7: 2003ff60
  R8: ffffffff   R9: ffffffff  R10: ffffffff  R11: ffffffff
 R12: ffffffff   SP: 2003ff58   LR: 200002ab   PC: 20001094
XPSR: 6100000f
```

## `upstream-drs-2`, SHA `1736fd055`

### EDF NOT active (default)

```
Test OK, count=106185, cycles=15799
  R0: 00000001   R1: 200200e0   R2: 00000000   R3: 20020249
  R4: 00019ec9   R5: 00000004   R6: 00000000   R7: 2003ff58
  R8: ffffffff   R9: ffffffff  R10: ffffffff  R11: ffffffff
 R12: ffffffff   SP: 2003ff50   LR: 200002af   PC: 20001098
XPSR: 6100000f
```

### EDF active (feat=`embassy-executor/edf-scheduler`)

```
Test OK, count=95870, cycles=17499
  R0: 00000001   R1: 200200e8   R2: 00000000   R3: 20020251
  R4: 0001767e   R5: 00000004   R6: 20020088   R7: 2003ff40
  R8: ffffffff   R9: ffffffff  R10: ffffffff  R11: ffffffff
 R12: ffffffff   SP: 2003ff38   LR: 20000313   PC: 20001124
XPSR: 6100000f
```

</p>
</details> 

```
cycles per poll
               nrf52832  rp2040
master         78        153
edf            78        158
edf enabled    91        175
```

edit: this matches the previous test run (a bit faster with edf enabled): https://github.com/embassy-rs/embassy/pull/4035#issuecomment-3073300137

diondokter:
Alright, so perf looks good and I just added the extra metadata and fixed the merge conflict

Dirbaio:
Implementation looks good, and it's great that perf doesn't regress now :rocket: 

1. The public API is a bit strange though:

    - Metadata gives you a `&Deadline`. 
    - `Deadline` has `pub fn instant_ticks(&self) -> u64`, to call that you must get the `&Deadline` from the `Metadata`.
    - Then it has a few `pub fn *_current_task_deadline() -> impl Future` methods, to call those you do *not* go through `Metadata`.

    Maybe the API should be just methods on `Metadata`? More consistent with the current `name`/`set_name`.

    - `fn deadline(&self) -> u64`
    - `fn set_deadline(&self) -> u64`
    - `fn clear_deadline(&self)`
    - etc

2.  `set_current_task_deadline_after`: it requires `embassy-time-driver`, which should be optional.
3. about cargo features:
    - I'd name the feature `scheduler-deadline`. 
    - I'm not sure if `metadata-deadline` is useful as a separate feature. What it'd be used for? I'd just make it all controlled by 'scheduler-deadline'.

diondokter:
Alright! Made the changes.

You're right about the weird API. I didn't critically review it after the metadata refactoring.

> * I'm not sure if `metadata-deadline` is useful as a separate feature. What it'd be used for? I'd just make it all controlled by 'scheduler-deadline'.

Idk, I just mirrored the existing metadata feature which looked nice to me. But I've removed it as per your preference.

embassy-executor/src/metadata.rs
```diff
@@ -7,18 +7,26 @@ use core::task::Poll;
...
 
 impl Metadata {
     pub(crate) const fn new() -> Self {
         Self {
             #[cfg(feature = "metadata-name")]
             name: Mutex::new(Cell::new(None)),
+            // NOTE: The deadline is set to zero to allow the initializer to reside in `.bss`. This
+            // will be lazily initalized in `initialize_impl`
+            #[cfg(feature = "scheduler-deadline")]
+            deadline: raw::Deadline::new_unset(),
```
bugadani, line 29: > NOTE: The deadline is set to zero

but it's really set to u64::MAX, isn't it? 👀

diondokter: Huh, seems like it yeah. Someone (me?) must've messed something up in a refactor

diondokter: Maybe deadline should be a nonzero type so we can put it in an option for free. When somebody sets the deadline to 0, we just secretly set it to 1 and nobody realistically would notice.

But it'd make the instruction count higher for doing unwrap_or in the scheduler, so maybe we shouldn't.

Other idea would be to NOT the deadline (or do MAX - deadline as the stored value). That would make it count down instead of up. So 0 would become the highest deadline... Would only cost one instruction outside of the hot loop

PR #4613, opened 2025-08-29
feat: add helper to read/write full transfer blocks

felipebalbi:
instead of always transferring only USB packets, add a provided method to transmit an entire data block by using a simple loop construct.

### DIFF
--- a/embassy-usb-driver/src/lib.rs
+++ b/embassy-usb-driver/src/lib.rs
@@ -242,7 +242,7 @@ pub trait EndpointOut: Endpoint {
     ///
     /// This should also clear any NAK flags and prepare the endpoint to receive the next packet or
     /// data block.
-    async fn read_data(&mut self, buf: &mut [u8]) -> Result<usize, EndpointError> {
+    async fn read_transfer(&mut self, buf: &mut [u8]) -> Result<usize, EndpointError> {
         let mut n = 0;
         loop {
             let i = self.read(&mut buf[n..]).await?;
@@ -370,11 +370,11 @@ pub trait EndpointIn: Endpoint {
     ///
     /// If the buffer size is evenly divisible by wMaxPacketSize, this will also ensure the
     /// terminating zero-length-packet is transmitted.
-    async fn write_data(&mut self, buf: &[u8]) -> Result<(), EndpointError> {
+    async fn write_transfer(&mut self, buf: &[u8], needs_zlp: bool) -> Result<(), EndpointError> {
         for chunk in buf.chunks(self.info().max_packet_size as usize) {
             self.write(chunk).await?;
         }
-        if buf.len() % self.info().max_packet_size as usize == 0 {
+        if needs_zlp && buf.len() % self.info().max_packet_size as usize == 0 {
             self.write(&[]).await?;
         }
         Ok(())
--- a/embassy-usb-driver/src/lib.rs
+++ b/embassy-usb-driver/src/lib.rs
@@ -236,6 +236,22 @@ pub trait EndpointOut: Endpoint {
     ///
     /// This should also clear any NAK flags and prepare the endpoint to receive the next packet.
     async fn read(&mut self, buf: &mut [u8]) -> Result<usize, EndpointError>;
+
+    /// Read until the buffer is full or we receive a short packet from the USB host returning the
+    /// actual length of the entire data block.
+    ///
+    /// This should also clear any NAK flags and prepare the endpoint to receive the next packet or
+    /// data block.
+    async fn read_data(&mut self, buf: &mut [u8]) -> Result<usize, EndpointError> {
+        let mut n = 0;
+        loop {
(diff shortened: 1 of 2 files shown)


felipebalbi:
@jamesmunns FYI

jamesmunns:
For what it's worth, ergot's handlers are a bit more extensive than this:

* RX: https://github.com/jamesmunns/ergot/blob/main/crates/ergot/src/interface_manager/profiles/direct_edge/eusb_0_5.rs#L112-L157
* TX: https://github.com/jamesmunns/ergot/blob/main/crates/ergot/src/interface_manager/interface_impls/embassy_usb.rs#L160-L192

The PR currently does not handle the following edge cases:

* If the RX buffer is too small, it will currently resume receiving the in-progress transfer as if it was a second transfer, instead of consuming + discarding the oversize transfer (e.g. if it's a 256 byte transfer and you have a 128 byte buffer, it'll return an error on the first 128 bytes, and return the back 128 bytes as a correct frame)
* There's no timeout on TX, I think I've observed some impls hanging if the host is connected but is not actively receiving/draining frames. The user COULD do their own cancellation/timeout, but may need to manually send a `&[]` IF cancellation occurs.

embassy-usb-driver/src/lib.rs
```diff
@@ -349,6 +365,20 @@ pub trait ControlPipe {
...
+    ///
+    /// If the buffer size is evenly divisible by wMaxPacketSize, this will also ensure the
+    /// terminating zero-length-packet is transmitted.
+    async fn write_data(&mut self, buf: &[u8]) -> Result<(), EndpointError> {
+        for chunk in buf.chunks(self.info().max_packet_size as usize) {
+            self.write(chunk).await?;
+        }
+        if buf.len() % self.info().max_packet_size as usize == 0 {
+            self.write(&[]).await?;
+        }
```
0e4ef622: Whether you send the ZLP or not depends on whether the host is expecting one or not, so there should be a param for it.

0e4ef622:
I think `read_transfer` and `write_transfer` would be better, it matches the USB terminology. The docs should also include this "transfer" term.

jamesmunns:
Screenshot 2025-08-30 at 13 16 46

The USB 2.0 spec (see above, section "5.3.2 Pipes") has some vague standardization for this pattern, I don't know if there is a more specific name than "I/O Request Packet (IRP) of a variable-sized unit of data".

nusb (at least v0.1, I haven't checked v0.2 yet) does just generally call these "Transfers" or "Transfer Requests".

In general, I think `read_transfer` and `write_transfer` might be better names, and we might want to name drop "I/O Request Packet"/"IRP" in the docs.

0e4ef622:
image


It's pretty clear here (section "5.8.3 Bulk Transfer Packet Size Constraints"), although this is about bulk transfers there's similar language for interrupt and control transfers too.

embassy-usb-driver/src/lib.rs
felipebalbi: makes sense, I'll do that.

felipebalbi:
> I think `read_transfer` and `write_transfer` would be better, it matches the USB terminology. The docs should also include this "transfer" term.

I agree, transfer reads better.

felipebalbi:
> For what it's worth, ergot's handlers are a bit more extensive than this:
> 
> * RX: https://github.com/jamesmunns/ergot/blob/main/crates/ergot/src/interface_manager/profiles/direct_edge/eusb_0_5.rs#L112-L157
> * TX: https://github.com/jamesmunns/ergot/blob/main/crates/ergot/src/interface_manager/interface_impls/embassy_usb.rs#L160-L192
> 
> The PR currently does not handle the following edge cases:
> 
> * If the RX buffer is too small, it will currently resume receiving the in-progress transfer as if it was a second transfer, instead of consuming + discarding the oversize transfer (e.g. if it's a 256 byte transfer and you have a 128 byte buffer, it'll return an error on the first 128 bytes, and return the back 128 bytes as a correct frame)

Sorry, just to be clear; you want me to drop the extra bytes in the transfer? I don't think that's a good idea. Say host wants to send 1MiB of data, embassy side can receive it in partial chunks of e.g. 4096 bytes and process it sequentially. Maybe the generic code should not make that decision and just receive until either a short packet, a ZLP, or the buffer is filled up.

> * There's no timeout on TX, I think I've observed some impls hanging if the host is connected but is not actively receiving/draining frames. The user COULD do their own cancellation/timeout, but may need to manually send a `&[]` IF cancellation occurs.

That's another thing I'm not sure should be done generically. The application can employ `with_timeout()` for that, no?

embassy-usb-driver/src/lib.rs
felipebalbi: done

felipebalbi:
I pushed a new commit fixing comments, unclear what `embassy` prefers. If it's preferrable to have a single commit, let me know and I can quickly squash them all.

PR #4616, opened 2025-08-31
nrf54l: Disable glitch detection and enable DC/DC in init.

OBauck:
Reduce current consumption by turning off glitch detection and turn on DC/DC converter.
Reduction in current consumption from 700uA to <10uA in sleep and from 6.13mA/3.75mA to 2.61mA/1.56mA with cpu running (128MHz/64MHz).
The device checks if an inductor is connected to the dcc pin. If no inductor is detected, then the device remains in ldo mode. So there is no need to have dc/dc option with nrf54l.
Glitch detection: https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/chapters/power-and-clock/glitchdet/doc/glitchdet.html
Regulator: https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/regulators.html#ariaid-title2


Dirbaio:
lgtm, can you fix CI?

OBauck:
Had to add "_s" feature to pass CI, so I split glitchdetection and dcdc into 2 blocks.

PR #4624, opened 2025-09-01
embassy-usb-dfu: fix: do net reset in GetStatus request

paroga:
Only reset the device after a USB reset request. This avoids error messages with update tools, which expect a response to a GetStatus request (like dfu-util).


korbin (approved):
I've run into this one a few times, this would be nice to have.

lulf (approved):

PR #4625, opened 2025-09-01
embassy-usb-dfu: Add a trait for the update logic

paroga:
I'd like to introduce a `trait` use `embassy-usb-dfu` without `embassy-boot` and implement my own update logic.

Before finishing that / cleaning it up, I'd like to get some feedback on the API and how to upstream that, since it's a very intrusive change and will break all downstream implementations.

### DIFF
--- a/embassy-boot/Cargo.toml
+++ b/embassy-boot/Cargo.toml
@@ -30,6 +30,8 @@ log = { version = "0.4", optional = true }
 ed25519-dalek = { version = "2", default-features = false, features = ["digest"], optional = true }
 embassy-embedded-hal = { version = "0.5.0", path = "../embassy-embedded-hal" }
 embassy-sync = { version = "0.7.2", path = "../embassy-sync" }
+embassy-time = { version = "0.5.0", path = "../embassy-time", optional = true }
+embassy-usb = { version = "0.5.1", path = "../embassy-usb", optional = true  }
 embedded-storage = "0.3.1"
 embedded-storage-async = { version = "0.4.1" }
 salty = { version = "0.3", optional = true }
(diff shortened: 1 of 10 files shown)


ExplodingWaffle:
Have you seen https://github.com/embassy-rs/embassy/pull/2407? It seems to me that the preferred way would be to make DFU just another class in embassy-usb (unless dirbaio changed his mind)

paroga:
> Have you seen #2407? It seems to me that the preferred way would be to make DFU just another class in embassy-usb (unless dirbaio changed his mind)

no, I didn't. thanks for the information. any idea how i can help in getting that merged?

@Dirbaio is there something like a list of required things in #2407? can i help somehow to finish it?

Dirbaio:
Yes I'd prefer to add the "unopinionated" class implementation to `embassy-usb`, then the `embassy-usb-dfu` crate could just use that and wrap it into an "opinionated" api that uses embassy-boot.

The missing thing in #2407 was to change `embassy-usb-dfu` to use the class so we don't have two implementations. Anyway, it's now quite old, it's probably easier to start from scratch.

paroga:
> Yes I'd prefer to add the "unopinionated" class implementation to `embassy-usb`, then the `embassy-usb-dfu` crate could just use that and wrap it into an "opinionated" api that uses embassy-boot.
> 
> The missing thing in #2407 was to change `embassy-usb-dfu` to use the class so we don't have two implementations. Anyway, it's now quite old, it's probably easier to start from scratch.

@Dirbaio if you're ok, i'll try to rebase it to latest `main` and adopt `embassy-usb-dfu` to use it in the next days/weeks and update this PR with it.

paroga:
@Dirbaio Since the remaining logic is minimal I'd move the `embassy-boot` adaptation directly into that create and deprecate `embassy-usb-dfu`. Does that sound reasonable?

If the basic structure looks good, I'll add the missing documentation and adopt the examples too.

Dirbaio:
sounds good yes.

Dirbaio:
closing in favor of #5216 (it ports embassy-usb-dfu to use the standalone class impl so there's no duplication)

PR #4630, opened 2025-09-01
Add a wait_for_update function to low_level timers

EliteTK:
I am using this to implement a OPM pulse with a wait-for-completion. This is to reliably clock a HX711 without risking holding the clock line for too long. I also think exposing this interrupt as a future is more flexible than trying to have a module per conceptual timer use-case.

Let me know if there's something wrong with this approach, I think this code is correct but I also wrote it about half a year ago... I'm successfully using this code on an STM32F411CE at the moment.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - fix: stm32/i2c in master mode (blocking): subsequent transmissions failed after a NACK was received
 - feat: stm32/timer: add set_polarity functions for main and complementary outputs in complementary_pwm
 - Add I2S support for STM32F1, STM32C0, STM32F0, STM32F3, STM32F7, STM32G0, STM32WL, STM32H5, STM32H7RS
+- Added `low_level::Timer::new_with_interrupt` as well as a mode generic
+  parameter to `low_level::Timer` to distinguish between timers with and without
+  a update interrupt binding. Existing code should add
+  `timer::low_level::mode::NoIrq`.
+- Added `low_level::Timer::wait_for_update` to asynchronously wait for an update
+  interrupt.
+
 ### Fixed
 
 - STM32: Prevent dropped DacChannel from disabling Dac peripheral if another DacChannel is still in scope ([#4577](https://github.com/embassy-rs/embassy/pull/4577))
--- a/embassy-stm32/src/timer/complementary_pwm.rs
+++ b/embassy-stm32/src/timer/complementary_pwm.rs
@@ -4,7 +4,7 @@ use core::marker::PhantomData;
 
 pub use stm32_metapac::timer::vals::{Ckd, Ossi, Ossr};
 
-use super::low_level::{CountingMode, OutputPolarity, Timer};
+use super::low_level::{CountingMode, NoIrq, OutputPolarity, Timer};
 use super::simple_pwm::PwmPin;
 use super::{AdvancedInstance4Channel, Ch1, Ch2, Ch3, Ch4, Channel, TimerComplementaryPin};
 use crate::gpio::{AnyPin, OutputType};
@@ -40,7 +40,7 @@ impl<'d, T: AdvancedInstance4Channel, C: TimerChannel> ComplementaryPwmPin<'d, T
 
 /// PWM driver with support for standard and complementary outputs.
 pub struct ComplementaryPwm<'d, T: AdvancedInstance4Channel> {
-    inner: Timer<'d, T>,
+    inner: Timer<'d, T, NoIrq>,
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
--- a/embassy-stm32/src/timer/input_capture.rs
+++ b/embassy-stm32/src/timer/input_capture.rs
@@ -5,7 +5,7 @@ use core::marker::PhantomData;
 use core::pin::Pin;
 use core::task::{Context, Poll};
 
-use super::low_level::{CountingMode, FilterValue, InputCaptureMode, InputTISelection, Timer};
+use super::low_level::{CountingMode, FilterValue, InputCaptureMode, InputTISelection, NoIrq, Timer};
 use super::{CaptureCompareInterruptHandler, Channel, GeneralInstance4Channel, TimerPin};
 pub use super::{Ch1, Ch2, Ch3, Ch4};
 use crate::gpio::{AfType, AnyPin, Pull};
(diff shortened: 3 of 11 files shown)


EliteTK:
I've added a changelog entry and made it not unsafe in exchange for making it take a mutable reference (It looks like this is how everything else handles this case).

Dirbaio:
for this to work the user has to bind the interrupt handler. There's no `Irqs` param in `new()` to enforce that, the user could forget to do it and then the future won't work for no apparent reason.

We can't simply add `Irqs` to `new()` because the low-level timer is supposed to allow the user to write their own irq handler if they want to. There's two possible solutions:

- Add a "mode" generic param to `struct Timer` so it can be in "without interrupts" mode or "with interrupts" mode. Only the "with interrupts" constuctor needs `Irqs`, and `wait_for_update` is only available in the "with interrupts" mode.  Similar to the async/blocking modes for uart/spi/i2c/etc. 
- No extra generic param, just require `Irqs` as a param to `fn wait_for_update()`. 

Perhaps the latter is simpler.

EliteTK:
@Dirbaio, thank you for looking at this.

The IRQs needing to be bound might be another reason why I had it marked unsafe. I couldn't remember the reasoning but had a feeling it wasn't just for the exclusivity of the UIE flag. In my code, my wrapper handles the enforcement of having the IRQ bound.

Regarding the options, the latter also seems simpler to me but, what would you prefer to see?

I'd much rather get this upstream than keep having to maintain my own patches, so I am completely flexible on the approach.

Dirbaio:
latter sounds good to me yes.

EliteTK:
So I just spent a few minutes on this.

It's a pretty easy change on the `embassy-stm32` side.

But the problem appears on the use-side.

So I have the aforementioned OPM timer pulser which pulses a timer once to bit bang this not-actually-a-normal-serial protocol that the HX711 load cell amplifiers use: https://paste.rs/MdYf2.rs (Yes, I know, the problem case I am avoiding by writing this code is basically impossible unless something goes seriously wrong first.)

In that code, only new takes the irqs. Now I have this diff:

```patch
diff --git a/controller/src/pulse.rs b/controller/src/pulse.rs
index 60d9d4d..8cd89b6 100644
--- a/controller/src/pulse.rs
+++ b/controller/src/pulse.rs
@@ -26,13 +26,14 @@ impl From<CountingMode> for low_level::CountingMode {
     }
 }
 
-pub struct Pulse<'d, T: GeneralInstance4Channel> {
+pub struct Pulse<'d, T: GeneralInstance4Channel, I: Binding<T::UpdateInterrupt, UpdateInterruptHandler<T>> + Copy + 'd> {
     inner: Timer<'d, T>,
+    irq: I,
 }
 
-impl<'d, T: GeneralInstance4Channel> Pulse<'d, T> {
+impl<'d, T: GeneralInstance4Channel, I: Binding<T::UpdateInterrupt, UpdateInterruptHandler<T>> + Copy + 'd> Pulse<'d, T, I> {
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub fn new(
         tim: Peri<'d, T>,
         ch1: Option<PwmPin<'d, T, Ch1>>,
         ch2: Option<PwmPin<'d, T, Ch2>>,
@@ -40,10 +41,11 @@ impl<'d, T: GeneralInstance4Channel> Pulse<'d, T> {
         ch4: Option<PwmPin<'d, T, Ch4>>,
         period: Hertz,
         counting_mode: CountingMode,
-        _irq: impl Binding<T::UpdateInterrupt, UpdateInterruptHandler<T>> + 'd,
+        irq: I,
     ) -> Self {
         let this = Self {
             inner: Timer::new(tim),
+            irq,
         };
 
         this.inner.set_counting_mode(counting_mode.into());
@@ -80,7 +82,7 @@ impl<'d, T: GeneralInstance4Channel> Pulse<'d, T> {
     pub async fn trigger_and_wait(&mut self) {
         self.inner.clear_update_interrupt();
         self.inner.start();
-        self.inner.wait_for_update().await;
+        self.inner.wait_for_update(self.irq).await;
     }
 }
```

And this isn't _too_ bad, except that now every single thing up the stack which holds a Pulse has to be generic over the IRQs and have them as a generic parameter. Needless to say, this adds a lot of noise.

I'm split between the other suggestion, or maybe wait_for_update could just be made unsafe again. With proper documentation on ensuring that its use is wrapped in a wrapper where new takes the IRQs like the one I pasted.

I guess it's also an option that I only make my `trigger_and_wait` function take the IRQs but that's only slightly less noisy up the stack.

What are your thoughts?

Dirbaio:
ugh yeah that's annoying. Perhaps adding a "mode" generic param is best then.

EliteTK:
Alright...

Mode it is.

I've tested that it works on real hardware at least but that doesn't mean it's definitely 100% correct.

Let me know if you would like any more major changes.

EliteTK:
Oh... right... the examples

EliteTK:
As a side note, I was thinking about: "then the future won't work for no apparent reason". There are things you still need to remember to do (for example, with my Pulse*):

* Timer::regs_core().egr().write(|r| r.set_ug(true)) otherwise the registers don't get initialised
* T::UpdateInterrupt::enable()
* Timer::start()

That being said, this is low_level::Timer not sunshine_and_roses::Timer, and we are in the land of embedded where if you get many things subtly wrong nothing works for no clear and obvious reason. So, honestly I don't think it's worth trying to fix the cases I listed above.

(* For reference https://paste.rs/4NkLq.rs - this is the actually working version not the old one I pasted earlier by accident)

EliteTK:
Except maybe UpdateInterrupt enable/disable?

I tried adding that just now, but you can't specialise Drop implementations, so T::UpdateInterrupt::disable() would always have to be called for both NoIrq and WithIrq Timer drivers.

EliteTK:
On that note, is there something else you'd like me to try with this?

I think the `UpdateInterrupt enable/disable` thing could be a justification for making this unsafe and documenting all the pre-requisites. I still think it's very valuable for the low level timer driver to expose this stuff just because it makes things about as ultimately flexible as they can be. If, for example, the capture compare interrupt got the same treatment, then _all_ the higher level timer drivers that currently exist could be implemented on top of low level without needing to be part of embassy-stm32. STM32 timers at the end of the day are incredibly versatile and it's unlikely that embassy-stm32 will ever fully capture the breadth of drivers you could write on top of them.

It could also be argued that with it being unsafe, and with documentation of the additional requirements, that the `WithIrq/NoIrq` stuff could also be removed. This would reduce the noise and avoid the potential future issue of dealing with two mode parameters (one for update interrupt and one for capture compare).

EliteTK:
~~Is there now some way to achieve this functionality without needing supporting code in embassy-stm32? I can't find anything obvious.~~

Never mind: I saw on matrix that this is just a cleanup pass and that I'm free to rebase and create a new PR.

Dirbaio:
Hello,

PR #4640, opened 2025-09-05
fix(embassy-stm32): Remove duplicate fn call, vestigial from earlier refcount implementation

universalhandle:
Refs: #4577


Dirbaio:
thanks!

Issue #3408, opened 2024-10-09
Imprecise PLL frequency calculation


t123yh:
Consider this example for STM32F407:

```rust
    let mut config = Config::default();
    {
        config.rcc.hse = Some(Hse {
            freq: Hertz(25_000_000),
            mode: HseMode::Oscillator,
        });
        config.rcc.pll_src = PllSource::HSE;
        config.rcc.pll = Some(Pll {
            prediv: PllPreDiv::DIV15,
            mul: PllMul::MUL192,
            divp: Some(PllPDiv::DIV2),
            divq: None,
            divr: None,
        });
        config.rcc.ahb_pre = AHBPrescaler::DIV1;
        config.rcc.apb1_pre = APBPrescaler::DIV4;
        config.rcc.apb2_pre = APBPrescaler::DIV2;
        config.rcc.sys = Sysclk::PLL1_P;
    }

    let p = embassy_stm32::init(config);
```

This first divides input frequency (25MHz) by 15, multipliy it by 192, then divide it by 2. This would result in an actual sysclk of `25000000/15*192/2=160000000` (160MHz), which is a nice-looking frequency.

However, the PLL code in [f4's rcc](https://github.com/embassy-rs/embassy/blob/456c226b29799f7db56ab60b0ae3d95cc7d6a32a/embassy-stm32/src/rcc/f247.rs#L372) first divides the frequency, then multiplies. This would make sysclk = `Hertz(159999936)`, rather than 160MHz, due to rounding issue (because 25 is not divisible by 15).

We cannot simply change the code to multiply first and then divide, because `25000000*192=4800000000`, which is out of bound for u32 (Hertz).

The impact of this is somewhat huge. If you do something like this:

```rust
    let mut can1 = Can::new(p.CAN1, p.PD0, p.PD1, Irqs);

    can1.modify_filters()
        .enable_bank(0, Fifo::Fifo0, Mask32::accept_all());

    can1.modify_config()
        .set_loopback(false) // Receive own frames
        .set_silent(false)
        .set_bitrate(500_000);
```

`set_bitrate` will panic, because it considers APB1 to have `39999984Hz`, thus cannot have 500000Hz baud rate.

Dirbaio:
I'd try using u64 and check if code size increases too much or not. I think it shouldn't, if you use LTO then RCC code usually gets optimized very well because the input is all compile-time-known constants.

[...]

PR #4642, opened 2025-09-06
Improve f247 PLL output frequency calculation precision

EliteTK:
It was really bugging me, okay?

The PLL frequency calculation will now produce the correct result for all integer output frequencies and a truncated result (e.g. 1234.5 -> 1234) for all non-integer output frequencies instead of accumulating an error.

I've tested this on an STM32F411CE.

I'm pretty confident that the "truncated or correct" guarantee is right, but I was a bit paranoid so I also brute forced all configurations (including many which would be unstable) of my F411CE's RCC to check: https://paste.rs/ZaZ3k.py

This addresses #3408 .

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add I2S support for STM32F1, STM32C0, STM32F0, STM32F3, STM32F7, STM32G0, STM32WL, STM32H5, STM32H7RS
 - fix: STM32: Prevent dropped DacChannel from disabling Dac peripheral if another DacChannel is still in scope ([#4577](https://github.com/embassy-rs/embassy/pull/4577))
 - feat: Added support for more OctoSPI configurations (e.g. APS6408 RAM) ([#4581](https://github.com/embassy-rs/embassy/pull/4581))
+- Improve the PLL output frequency calculation precision for the F2, F4, and F7
+  series.
 
 ## 0.4.0 - 2025-08-26
 
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1682,8 +1682,7 @@ fn main() {
 
         if is_rcc_name(e.name) {
             let enum_name = format_ident!("{}", e.name);
-            let mut muls = Vec::new();
-            let mut divs = Vec::new();
+            let mut nds = Vec::new();
             for v in e.variants {
                 let Ok(val) = parse_num(v.name) else {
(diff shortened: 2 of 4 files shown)


EliteTK:
Hmm, nevermind, I forgot about the u32 problem. Although I have an alternative idea.

EliteTK:
Okay, ended up with a different solution which uses u64 for the calculations. There's also now a new internal `time::Scale` type to which an RCC prescaler / divider / multiplier can be converted (and the Div/Mul impls for those now use time::Scale just to simplify things).

This is also working on my STM32 but should now work for the whole range of high speed external clock inputs (of which the HSI clock is a member).

Worth noting that the division by the peripheral prescalers is still done in the "imprecise" way. But the saving grace is that if you're aiming for an integer clock post prescale, then you will need to start with an integer clock anyway, so someone targeting a very precise peripheral clock shouldn't have any issue.

Dirbaio:
FIx looks good in essence (doing `src as u64 * mul / (div * prediv)` instead of `src / prediv * mul / div`). 

The implementation is complex however. I'd prefer if it was done without introducing a new `struct Scale` and impls. I think it should be possible to just do the `src as u64 * mul / (div * prediv)` calculation in the RCC code directly. Perhaps change the build.rs impls to impl `From<xxx> for u32` instead of `Div/Mul`.

EliteTK:
Can't do `From<xxx> for u32` as neither `xxx` nor `u32` are from embassy-stm32.

Also, while working on this, I realised there's a numerator and denominator meaning that `From<xxx> for u32` might itself result in imprecision (although I'm not aware of a case where this would actually happen).

EliteTK:
Yeah, just checked, a bunch of f1 and f3 chips have Usbpre which has a 1.5 divisor, and Pllmul which has a 6.5 multiplier. So implementing `From<Usbpre> for u32` or `From <Pllmul> for u32` in these cases would lead to imprecise results. They're not the right chip family but the current Scale solution could be applied for more precise clocks for those families too.

(Edit: Although it is extremely rare honestly...)

Dirbaio:
> Usbpre which has a 1.5 divisor, and Pllmul which has a 6.5 multiplier

oh no you're absolutely right. I forgot about those. Making a fraction struct makes sense then, hmm.

EliteTK:
I can split this into multiple commits if you like. One to add the rational type and one to utilise it in the timer code.

Would also be happy to further discuss how to proceed with this change or something else entirely, either here or on matrix.

EliteTK:
~~Hey, can you let me know why this was closed?~~

Never mind: I saw on matrix that this is just a cleanup pass and that I'm free to rebase and create a new PR.

Dirbaio:
Yes! :(

PR #4646, opened 2025-09-08
MSPM0: add adc implementation

Iooon:
Adding the adc implementation:
- Reading a single channel, blocking and asynchronously
- Reading a sequence of channels, only asynchronously (the interface from [here](https://github.com/embassy-rs/embassy/blob/a6cd24907aa43a8178a16b0db3d6b376f67f7540/embassy-stm32/src/adc/v3.rs#L411) )

I tested the implementation before my L1306 board broke with the dev boards for MSPM0G3507 and MSPM0L1306.

### DIFF
--- a/embassy-mspm0/build.rs
+++ b/embassy-mspm0/build.rs
@@ -68,6 +68,7 @@ fn generate_code() {
     g.extend(generate_pin_trait_impls());
     g.extend(generate_groups());
     g.extend(generate_dma_channel_count());
+    g.extend(generate_adc_constants());
 
     let out_dir = &PathBuf::from(env::var_os("OUT_DIR").unwrap());
     let out_file = out_dir.join("_generated.rs").to_string_lossy().to_string();
@@ -218,6 +219,59 @@ fn generate_dma_channel_count() -> TokenStream {
     let count = METADATA.dma_channels.len();
 
     quote! { pub const DMA_CHANNELS: usize = #count; }
+}
+
+fn generate_adc_constants() -> TokenStream {
+    let vrsel = METADATA.adc_vrsel;
+    let memctl = METADATA.adc_memctl;
+
+    if vrsel == 3 {
+        quote! {
+            pub const ADC_VRSEL: u8 = #vrsel;
+            pub const ADC_MEMCTL: u8 = #memctl;
+
+            #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+            #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+            /// Reference voltage (Vref) selection for the ADC channels.
+            pub enum Vrsel {
+                /// VDDA reference
+                VddaVssa = 0,
+
+                /// External reference from pin
+                ExtrefVrefm = 1,
+
+                /// Internal reference
+                IntrefVssa = 2,
+            }
+        }
+    } else if vrsel == 5 {
+        quote! {
+            pub const ADC_VRSEL: u8 = #vrsel;
+            pub const ADC_MEMCTL: u8 = #memctl;
+
+            #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+            #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+            /// Reference voltage (Vref) selection for the ADC channels.
+            pub enum Vrsel {
+                /// VDDA reference
+                VddaVssa = 0,
+
+                /// External reference from pin
+                ExtrefVrefm = 1,
+
+                /// Internal reference
+                IntrefVssa = 2,
+
+                /// VDDA and VREFM connected to VREF+ and VREF- of ADC
+                VddaVrefm = 3,
+
+                /// INTREF and VREFM connected to VREF+ and VREF- of ADC
+                IntrefVrefm = 4,
+            }
+        }
+    } else {
+        panic!("Unsupported ADC VRSEL value: {vrsel}");
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -561,6 +615,7 @@ fn generate_peripheral_instances() -> TokenStream {
             "uart" => Some(quote! { impl_uart_instance!(#peri); }),
             "i2c" => Some(quote! { impl_i2c_instance!(#peri, #fifo_size); }),
             "wwdt" => Some(quote! { impl_wwdt_instance!(#peri); }),
+            "adc" => Some(quote! { impl_adc_instance!(#peri); }),
             _ => None,
         };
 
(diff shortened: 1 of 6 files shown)


embassy-mspm0/build.rs
```diff
@@ -220,6 +221,59 @@ fn generate_dma_channel_count() -> TokenStream {
...
+                /// VDDA reference
+                VddaVssa = 0,
+
+                /// External reference from pin
+                ExtrefVrefm = 1,
+
+                /// Internal reference
+                IntrefVssa = 2,
+            }
+        }
```
i509VCB: We could cfg guard variants depending on the vrsel capabilities. I don't think we need to generate this type dynamically.

embassy-mspm0/build.rs
```diff
@@ -220,6 +221,59 @@ fn generate_dma_channel_count() -> TokenStream {
...
+
+                /// VDDA and VREFM connected to VREF+ and VREF- of ADC
+                VddaVrefm = 3,
+
+                /// INTREF and VREFM connected to VREF+ and VREF- of ADC
+                IntrefVrefm = 4,
+            }
+        }
+    } else {
+        panic!("Unsupported ADC VRSEL value: {vrsel}");
```
i509VCB: But do keep this assertion in case there is a mistake in metapac.

embassy-mspm0/src/adc.rs
```diff
@@ -0,0 +1,483 @@
...
+            Resolution::BIT10 => 10,
+            Resolution::BIT8 => 8,
+        }
+    }
+}
+
+pub use crate::_generated::Vrsel;
+
+/// ADC configuration.
+pub struct AdcConfig {
```
i509VCB: I don't think we need to prefix this with `Adc`. If you need to refer to multiple config types in code, `adc::Config` is usually more clear.

i509VCB: `#[non_exhaustive]`?

Also would Clone and Copy be an issue to implement here?

embassy-mspm0/src/adc.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+impl<'d, T: Instance> Adc<'d, T, Async> {
+    /// A new asynchronous ADC driver instance.
+    pub fn new_async(
+        peri: Peri<'d, T>,
+        config: AdcConfig,
+        _irqs: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>>
+            + interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>>
```
i509VCB: Is this not the same trait bound twice?

embassy-mspm0/src/adc.rs
```diff
@@ -0,0 +1,483 @@
...
+    {
+        assert!(sequence.len() != 0, "Asynchronous read sequence cannot be empty");
+        assert!(
+            sequence.len() == readings.len(),
+            "Sequence length must be equal to readings length"
+        );
+        assert!(
+            sequence.len() <= ADC_MEMCTL as usize,
+            "Asynchronous read sequence cannot be more than {} in length",
+            ADC_MEMCTL
```
i509VCB: Would it be helpful to have some associated constant to tell you how long your sequence can be at max (on `Adc`)?

embassy-mspm0/build.rs
Iooon: Would this be a better solution?
```rust
fn generate_adc_constants() -> TokenStream {
    let vrsel = METADATA.adc_vrsel;
    let memctl = METADATA.adc_memctl;

    println!("cargo::rustc-check-cfg=cfg(adc_neg_vref)");
    match vrsel {
        3 => (),
        5 => println!("cargo:rustc-cfg=adc_neg_vref"),
        _ => panic!("Unsupported ADC VRSEL value: {vrsel}"),
    }
    quote! {
        pub const ADC_VRSEL: u8 = #vrsel;
        pub const ADC_MEMCTL: u8 = #memctl;
    }
}
``` 
and then in adc.rs
```rust
#[derive(Clone, Copy, PartialEq, Eq, Debug)]
/// Reference voltage (Vref) selection for the ADC channels.
pub enum Vrsel {
    /// VDDA reference
    VddaVssa = 0,

    /// External reference from pin
    ExtrefVrefm = 1,

    /// Internal reference
    IntrefVssa = 2,

    /// VDDA and VREFM connected to VREF+ and VREF- of ADC
    #[cfg(adc_neg_vref)]
    VddaVrefm = 3,

    /// INTREF and VREFM connected to VREF+ and VREF- of ADC
    #[cfg(adc_neg_vref)]
    IntrefVrefm = 4,
}
```

embassy-mspm0/src/adc.rs
Iooon: Sounds good.

Iooon: Yes, then I will also implement the Default trait

embassy-mspm0/src/adc.rs
Iooon: Yeah you are right, makes it way cleaner.

embassy-mspm0/src/adc.rs
Iooon: Yeah that's a nice idea.

embassy-mspm0/build.rs
i509VCB: Yes. You can use the functions on cfg_set as well to manage defining whatever so rustc doesn't complain about unknown cfg names

Iooon: The rustc-check-cfg should handle this (that rustc does not complain about unkown cfg names). But is this a good practice to do in embassy?

i509VCB: So `cfgs.declare` and `cfgs.enable` wrap rustc-check-cfg in a way that is easier to use.

Iooon: Got it :+1:

embassy-mspm0/src/adc.rs
```diff
@@ -0,0 +1,516 @@
...
+        self.channel
+    }
+}
+
+/// ADC channel.
+#[allow(private_bounds)]
+pub trait AdcChannel<T>: SealedAdcChannel<T> + Sized {
+    /// Allows an ADC channel to be converted into a type-erased [`AnyAdcChannel`].
+    #[allow(unused_mut)]
+    fn degrade_adc(mut self) -> AnyAdcChannel<T> {
```
i509VCB: AnyPin doesn't appear to need degrade?

Might be worth copying what we do for AnyPin and make AdcChannel impl `Into<AnyAdcChannel<T>>`? https://github.com/embassy-rs/embassy/blob/main/embassy-mspm0/src/gpio.rs#L626

Iooon: Done

i509VCB (approved):

PR #4647, opened 2025-09-08
lpc55: move to nxp-pac

frihetselsker:
Hello! @IrinaCh524 and I finished rewriting the existing codebase for LPC55 board to use `nxp-pac`.

- Rewrote GPIO with simplifications
- Rewrote PINT and RTC driver
- Rewrote USART code with optimizations and simplifications

### DIFF
--- a/embassy-nxp/Cargo.toml
+++ b/embassy-nxp/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 
 [package.metadata.embassy]
 build = [
-    {target = "thumbv8m.main-none-eabihf", features = ["defmt", "lpc55"]},
+    {target = "thumbv8m.main-none-eabihf", features = ["defmt", "lpc55-core0"]},
     {target = "thumbv7em-none-eabihf", features = ["defmt", "mimxrt1011", "rt", "time-driver-pit"]},
     {target = "thumbv7em-none-eabihf", features = ["defmt", "mimxrt1062", "rt", "time-driver-pit"]},
 ]
@@ -18,7 +18,7 @@ src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-nxp/s
 features = ["defmt", "unstable-pac" ] # TODO: Add time-driver-any, as both lpc55 and mimxrt1xxx use different drivers.
 
 flavors = [
-    { regex_feature = "lpc55", target = "thumbv8m.main-none-eabihf" },
+    { regex_feature = "lpc55-core0", target = "thumbv8m.main-none-eabihf" },
     { regex_feature = "mimxrt.*", target = "thumbv7em-none-eabihf" },
 ]
 
@@ -78,6 +78,6 @@ _rt1xxx = []
 _time_driver = ["dep:embassy-time-driver", "dep:embassy-time-queue-utils"]
 
 #! ### Chip selection features
-lpc55 = ["nxp-pac/lpc55s69_cm33_core0"]
+lpc55-core0 = ["nxp-pac/lpc55s69_cm33_core0"]
 mimxrt1011 = ["nxp-pac/mimxrt1011", "_rt1xxx", "dep:imxrt-rt"]
 mimxrt1062 = ["nxp-pac/mimxrt1062", "_rt1xxx", "dep:imxrt-rt"]
--- a/embassy-nxp/src/gpio.rs
+++ b/embassy-nxp/src/gpio.rs
@@ -1,7 +1,7 @@
 //! General purpose input/output (GPIO) driver.
 #![macro_use]
 
-#[cfg_attr(feature = "lpc55", path = "./gpio/lpc55.rs")]
+#[cfg_attr(feature = "lpc55-core0", path = "./gpio/lpc55.rs")]
 #[cfg_attr(rt1xxx, path = "./gpio/rt1xxx.rs")]
 mod inner;
 pub use inner::*;
--- a/embassy-nxp/src/lib.rs
+++ b/embassy-nxp/src/lib.rs
@@ -4,9 +4,9 @@
 pub(crate) mod fmt;
 
 pub mod gpio;
-#[cfg(feature = "lpc55")]
+#[cfg(feature = "lpc55-core0")]
 pub mod pint;
-#[cfg(feature = "lpc55")]
+#[cfg(feature = "lpc55-core0")]
 pub mod usart;
 
 #[cfg(feature = "_time_driver")]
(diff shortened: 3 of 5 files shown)


embassy-nxp/Cargo.toml
```diff
@@ -37,13 +37,13 @@ embedded-io = "0.6.1"
 embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["unproven"] }
 ## Chip dependencies
 lpc55-pac = { version = "0.5.0", optional = true }
```
i509VCB: Since we don't use lpc55-pac any more we should remove the dependency.

embassy-nxp/Cargo.toml
```diff
@@ -79,6 +79,6 @@ _rt1xxx = []
 _time_driver = ["dep:embassy-time-driver", "dep:embassy-time-queue-utils"]
 
 #! ### Chip selection features
-lpc55 = ["dep:lpc55-pac"]
+lpc55 = ["nxp-pac/lpc55s69_cm33_core0"]
```
i509VCB: Do we want to rename the feature to match the fact that it is core0? (core1 won't happen for a bit I suppose).

embassy-nxp/src/pint.rs
```diff
@@ -88,18 +88,18 @@ enum InterruptOn {
...
-        crate::pac::NVIC::unmask(crate::pac::Interrupt::PIN_INT6);
-        crate::pac::NVIC::unmask(crate::pac::Interrupt::PIN_INT7);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT0);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT1);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT2);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT3);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT4);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT5);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT6);
+        cortex_m::peripheral::NVIC::unmask(crate::pac::interrupt::PIN_INT7);
```
i509VCB: You should be able to write all of these in terms of `InterruptExt`.

e.g.
```
interrupt::PIN_INT0.unmask()
```

Edit: or it might be `enable()`

embassy-nxp/src/pint.rs
```diff
@@ -294,49 +262,49 @@ fn handle_interrupt(interrupt_number: usize) {
...
 #[allow(non_snake_case)]
-#[interrupt]
+#[cortex_m_rt::interrupt]
 fn PIN_INT6() {
     handle_interrupt(6);
 }
 
 #[allow(non_snake_case)]
-#[interrupt]
+#[cortex_m_rt::interrupt]
```
i509VCB: You probably don't need to fully qualify the path these usages of `#[interrupt]`

embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -83,33 +85,36 @@ impl RtcDriver {
...
         critical_section::with(|cs| {
-            let rtc = unsafe { &*RTC::ptr() };
-            let flags = rtc.ctrl.read();
-            if flags.alarm1hz().bit_is_clear() {
-                rtc.ctrl.modify(|_, w| w.alarm1hz().set_bit());
+            let rtc = RTC;
+            let flags = rtc.ctrl().read();
+            //i feel like there might be something wrong here but i cant put my finger on it
+            if flags.alarm1hz() == false {
+                rtc.ctrl().modify(|w| w.set_alarm1hz(true));
```
i509VCB: What specifically came up here?

embassy-nxp/src/time_driver/rtc.rs
```diff
@@ -83,33 +85,36 @@ impl RtcDriver {
...
                 self.trigger_alarm(cs);
             }
 
-            if flags.wake1khz().bit_is_clear() {
-                rtc.ctrl.modify(|_, w| w.wake1khz().set_bit());
+            if flags.wake1khz() == nxp_pac::rtc::vals::Wake1khz::TIMEOUT {
+                rtc.ctrl().modify(|w| w.set_wake1khz(nxp_pac::rtc::vals::Wake1khz::RUN));
```
i509VCB: Redundant long paths that can be shortened with an import (`vals::Wake1khz`) would be enough

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -223,29 +226,70 @@ impl<'d, T: Instance> Usart<'d, T, Blocking> {
...
+            121..=1500 => {
+                SYSCON
+                    .fcclksel(T::instance_number())
+                    .modify(|w| w.set_sel(crate::pac::syscon::vals::FcclkselSel::ENUM_0X4)); // 1 MHz
+                1_000_000
+            }
+            _ => {
+                panic!("{} baudrate is not permitted in this mode", config.baudrate);
+            }
+        };
```
i509VCB: Don't need to address this now, but we'd need a way to resolve this after erasing `T` when doing something like changing baud rate/reconfiguring in the future.

i509VCB (commented):
A few little comments. All hardware code looks correct compared to original.

embassy-nxp/src/time_driver/rtc.rs
IrinaCh524: While migrating from `lpc55-pac` to `nxp-pac`, I wasn’t entirely confident about this part and left a comment as a reminder to revisit it if the test failed. Since the test passed, I forgot to come back and update the comment. My bad :)).

embassy-nxp/src/usart/lpc55.rs
frihetselsker: Should we then create a trait with an instance number or add this to the struct `Info`?

embassy-nxp/src/time_driver/rtc.rs
IrinaCh524: I was right. something was wrong there: `TIMEOUT` and `RUN` should've been switched here and in the on_interrupt function.

embassy-nxp/src/usart/lpc55.rs
i509VCB: You could probably add that to Info. But not needed today (reconfigure isn't implemented yet?)

i509VCB:
Would it be possible to squash this down? Distinct commit barriers don't make much sense with how new things are.

eva-cosma:
Hello @i509VCB ! We would prefer to keep the commits separate mainly to distinguish authorship. This was a joint effort yet the work was discrete - breaking the work down into drivers. Would this be fine?

i509VCB:
> Hello @i509VCB ! We would prefer to keep the commits separate mainly to distinguish authorship. This was a joint effort yet the work was discrete - breaking the work down into drivers. Would this be fine?

Well it's more that I'd prefer to not have 50 commits added in a single pull request. But doesn't really matter currently.

embassy-nxp/src/usart/lpc55.rs
frihetselsker: Yes, let's leave it for further improvements.

PR #4650, opened 2025-09-08
Add Wiznet W6100 driver

cjdell:
Tested with this device:
https://docs.wiznet.io/Product/iEthernet/W6100/w6100-evb-pico2

I've introduced a new register `SOCKET_INTR_CLR` as the W6100 (and presumably later devices) use a _different_ register to clear the socket interrupt. See `Sn_IRCLR` on page 61:
https://docs.wiznet.io/img/products/w6100/w6100_ds_v105e.pdf

On previous driver implementations (W5100/W5500) this register address is simply a duplicate of `SOCKET_INTR`.

Weirdly the MAC filter does not operate like the W5500 as having it turned on will stop DHCP working. I've left the MAC filter disabled but this does not affect the operation of this device (although it will cause more noise on the SPI bus).

For this reason I would still consider this driver experimental however I would like to start the conversation of getting this upstreamed.

### DIFF
--- a/embassy-net-wiznet/CHANGELOG.md
+++ b/embassy-net-wiznet/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Added experimental W6100 driver with disabled MAC filter (does not currently work with it enabled)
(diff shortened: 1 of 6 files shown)


mkj:
Looks good, thanks

lulf (approved):
Looks good, thanks! Please add an entry to the changelog as well

Dirbaio:
please don't increase the version nor add a section to the changelog, that's done automatically at release time. Could you just add it to the "Unreleased" section?

Dirbaio (approved):
thanks!

PR #4655, opened 2025-09-11
Remove `Sized` bound from `MutexGuard::map`

RobertZ2011:
Since `MutexGuard` has `T: ?Sized`, `U` does not need to be restricted to `Sized` types. This now allows using `map` to cast from `MutexGuard<'_, M, ImplsTrait>` to `MutexGuard<'_, M, dyn Trait>`.


Dirbaio:
lgtm, can you add a changelog?

RobertZ2011:
> lgtm, can you add a changelog?

Done, thanks for taking a look.

PR #4656, opened 2025-09-12
embassy-rp: allow injection of FIFO handler

kaspar030:
Add `fifo-handler` feature for injecting a custom handler of FIFO messages.
Also enable FIFO queue in core1->core0 direction.

In Ariel OS, we need a way to trigger the scheduler on the other core. The FIFO peripheral is made for this, but also claimed by embassy-rp, for *stopping* the second core from the first. Inside embassy, I think this is used to stop execution from flash while messing with it.

This PR (optionally, based on `fifo-handler` feature) makes this a bit more flexible:

1. It is now also possible to stop the first core from the second
2. With `fifo-handler` enabled, an `fn handle_fifo(u32) -> bool` is called for each FIFO message arriving from the other core that is not used for stopping or resuming.

Signed-off-by: Kaspar Schleiser <kaspar@schleiser.de>

### DIFF
--- a/embassy-rp/src/multicore.rs
+++ b/embassy-rp/src/multicore.rs
@@ -108,54 +108,63 @@ impl<const SIZE: usize> Stack<SIZE> {
     }
 }
 
-#[cfg(all(feature = "rt", feature = "rp2040"))]
-#[interrupt]
+#[cfg(feature = "fifo-handler")]
+extern "Rust" {
+    fn handle_fifo_token(token: u32) -> bool;
+}
+
+#[cfg(feature = "rt")]
 #[link_section = ".data.ram_func"]
-unsafe fn SIO_IRQ_PROC1() {
+#[inline]
+unsafe fn sio_handler() {
     let sio = pac::SIO;
     // Clear IRQ
     sio.fifo().st().write(|w| w.set_wof(false));
 
     while sio.fifo().st().read().vld() {
-        // Pause CORE1 execution and disable interrupts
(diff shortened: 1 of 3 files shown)


CBJamo:
I like this change.

Separately, on the 2350, pausing/resuming the cores should probably be switched to use a pair of doorbells rather than the fifo. Possibly that would work for your scheduler case as well. Then the fifos could be left free for users.

embassy-rp/src/multicore.rs
```diff
@@ -285,23 +295,55 @@ where
...
+    };
 }
 
 /// Pause execution on CORE1.
 pub fn pause_core1() {
     if IS_CORE1_INIT.load(Ordering::Acquire) {
-        fifo_write(PAUSE_TOKEN);
-        // Wait for CORE1 to signal it has paused execution.
-        while fifo_read() != PAUSE_TOKEN {}
+        cortex_m::interrupt::free(|_| {
```
9names: Is it necessary to disable interrupts while waiting for core1 to respond to a pause request?

Dirbaio:
Sorry for the mega long delay. I'm working on clearing the PR queue.

PR looks good, we definitely should allow the user to use the FIFO. However this PR only implements part of it.

- Could you add a function to send something on the FIFO?
- Requiring the user to match an `extern "Rust"` is a bit dangerous. What about a `set_fifo_handler!(my_func)` macro?
- Could you add an example?

Also could you rebase on latest main?

PR #4658, opened 2025-09-12
Add example for W55RP20-EVB-Pico

Riceman2000:
With the addition of the SPI PIO program, a new Wiznet Ethernet controller(ish) can be added to the examples. The W55RP20 is just a RP2040 and a W5500 in the same package. While the other examples could also be adapted to the W55RP20 it is not really necessary because the only thing that changes is the creation of the SPI bus.

Note that the operating frequency is significantly reduced from the theoretical PIO max of 31.25MHz. I did some testing to figure out where instability starts and that point is above 10MHz. A comment was added to the example to warn about this.

Testing was done with a harness that sent TCP traffic and monitored for panics, nothing obvious was found.

Thanks


KingCol13:
Thanks, it's good to get more examples for these boards.

It's a bit surprising that the SPI frequency has to be so low. The [comments](https://github.com/embassy-rs/embassy/blob/be794533d3929e316c65b4296de47292ae0eae67/cyw43-pio/src/lib.rs#L39) in `cyw43-pio` suggest 62.5MHz half-duplex PIO SPI works fine, and [section 5.5.4](https://docs.wiznet.io/img/products/w5500/W5500_ds_v110e.pdf#%5B%7B%22num%22%3A147%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C82%2C753%2C0%5D) of the W5500 datasheet suggests a "theoretical guaranteed" frequency of 33.3MHz. I wonder what's going wrong for 31.25MHz to be unstable.

Have you tried 12.5MHz too? I think it would make sense for the highest stable frequency to be an integer fraction of the system clock frequency.

Riceman2000:
Just did some more testing and 12.5MHz was stable while 12.6MHz was not. I upped the frequency used to 12.5MHz.
Interestingly, 12500001Hz worked for chip initialization before having some kind of numerical error. There is definitely something going on for it to have such a well defined wall. I wish I could logic analyzer these comms but they are in-package.

From the W5500 data sheet as you mentioned we should have no problem running this as fast as the 31.25MHz limit of the PIO clock dividers. Any ideas on validating the PIO SPI side of this more thoroughly? 

Here is a good blog post about this chip: https://www.raspi.jp/2024/10/wiznet-w55rp20-evb-pico-report/

Thanks for the reply

KingCol13:
> There is definitely something going on for it to have such a well defined wall.

I think what's going on is explained by [this comment](https://github.com/embassy-rs/embassy/blob/be794533d3929e316c65b4296de47292ae0eae67/cyw43-pio/src/lib.rs#L26) - when the PIO clock divider has a fractional part rather than just being an integer then some PIO clocks are done at double frequency and some at half. At 12500001Hz maybe you were able to initialise before the counter triggered the double frequency. Maybe we'd have some luck with other larger integer PIO clock dividers?

Looking at the WIZnet examples, [they use](https://github.com/WIZnet-ioNIC/WIZnet-PICO-C/blob/524de46c2764541077e62c575f96319617030184/port/ioLibrary_Driver/inc/wizchip_spi.h#L18) a clock divider of 2, which I think corresponds to 31.25MHz, but [their SPI PIO program](https://github.com/WIZnet-ioNIC/WIZnet-PICO-C/blob/524de46c2764541077e62c575f96319617030184/port/ioLibrary_Driver/src/wizchip_qspi_pio.pio#L7) looks different to the embassy/datasheet one [here](https://github.com/embassy-rs/embassy/blob/be794533d3929e316c65b4296de47292ae0eae67/embassy-rp/src/pio_programs/spi.rs#L40).

I don't have good ideas for validating anything but I'm not that good at electronics. I don't have the board myself but I've been thinking about picking one up to play with.

Thanks for looking at this!

examples/rp/src/bin/pio_spi.rs
```diff
@@ -27,9 +27,9 @@ async fn main(_spawner: Spawner) {
 
     // These pins are routed to different hardware SPI peripherals, but we can
     // use them together regardless
-    let mosi = p.PIN_6; // SPI0 SCLK
-    let miso = p.PIN_7; // SPI0 MOSI
-    let clk = p.PIN_8; // SPI1 MISO
+    let mosi = p.PIN_6;
+    let miso = p.PIN_7;
+    let clk = p.PIN_8;
```
KingCol13: Did you mean to remove these comments?

examples/rp/src/bin/pio_spi_async.rs
```diff
@@ -27,9 +27,9 @@ async fn main(_spawner: Spawner) {
 
     // These pins are routed to different hardware SPI peripherals, but we can
     // use them together regardless
-    let mosi = p.PIN_6; // SPI0 SCLK
-    let miso = p.PIN_7; // SPI0 MOSI
-    let clk = p.PIN_8; // SPI1 MISO
+    let mosi = p.PIN_6;
+    let miso = p.PIN_7;
+    let clk = p.PIN_8;
```
KingCol13: Ditto here

examples/rp/src/bin/pio_spi.rs
Riceman2000: I did but I realize that was a mistake, I thought they were a typo mislabeling the pins but now I see they were meant to point out the use of a PIO for non-standard SPI pins. 

I fixed their removal.

Riceman2000:
The PIO fractional clock is probably a source of our problems. 
I'd imagine that when the clock jitters we are getting a malformed packet which causes the panic.

I am going to spend a bit of time to make a test harness than can step through a range of frequencies. I'll report back on how it goes.

Riceman2000:
I am seeing that the first time we get the correct chip version number (should be 4) is with a SPI clock frequency of around 14.4MHz with a good amount of variation between runs. I do not know why this is and I really do not have the equipment to look into it very deeply. 

However, given that this is working at high speed in the normal W5500 dev boards and not this one I have to assume it is some kind of weirdness with the PIO clocks or modules themselves. 

@KingCol13 what are your thoughts for this PR? Should it be pushed through with the lower SPI frequency or should it wait until there is more certainty with the SPI PIO program? I am open to either.

Dirbaio:
Have you tried using XOSC? maybe it has less jitter than ROSC.

```rust
let mut config = embassy_rp::Config::default();
config.clocks = ClockConfig::crystal(12_000_000);
let p = embassy_rp::init(config);
```

Riceman2000:
With the switch over to XOSC we still get a valid chip number right around 14.4MHz with some fluctuation between runs. Does not seem to have affected it.

Dirbaio:
Strange!

Do you want to merge the example as-is even if it's not the fastest it could theoretically go? you can always send a PR later to update it if you figure out how to make it faster.

Riceman2000:
It is not ideal but I also don't think this platform is the correct choice to debugging this issue. It should be done in an isolated environment. So Id say let's merge this if everyone is ok with it and I'll open another issue for further discussion on the PIO SPI issues.

If anyone has any ideas on how to lay that out I'd love to hear them. I think it could be done with a single Pi Pico board with one core using a PIO SPI and the other core using a standard SPI. This would eliminate some variables like clock drift and isolate the jitter. Open to suggestions and I'll rig something up in my spare time.

Thanks for the input everyone, it is appreciated

KingCol13:
I think merging this now and treating the speed as a different issue was the right call.

I think the next steps would be to test the other implementations:
- https://github.com/tnishinaga/w55rp20_tcp_server_example/tree/main this one from the above linked blog
- https://github.com/WIZnet-ioNIC/WIZnet-PICO-C this one from WIZnet

If one of those works at a higher frequency then that would indicate a deeper problem to find with the code. If they don't work at higher frequencies it could be interesting to try and rule out other nearby integer clock dividers:
- `125MHz/8 = 15.625MHz`
- `125MHz/7 = 17.857142857142858MHz`
- `125MHz/6 = 20.83333333333333MHz`
- `125MHz/5 = 25MHz`

PR #4660, opened 2025-09-13
rp/pio: clear irq separately after waiting

djdisodo:
benefits: allow you to delay irq clear, this could be useful if you are using irq status to signal back to state machine

this wouldn't break most of the code unless there were a code that stores return value of IrqFuture which is ()

draft as not tested


djdisodo:
however, even with this api
holding IRQ set state longer or setting it manually seems impossible

or we can just make IRQ not clear automatically
and allow user to set/clear irq manually

but i think that way it'll break a lot of codes where it assumes irq clears automatically(without even showing a warning or any)

Dirbaio:

PR #4663, opened 2025-09-14
WIP - Added backup sram

usbalbin:
I do not have any experience with the structure of embassy-stm32, its build script, linker script etc. Hopefully at the very least this can serve as something to discuss around for how to actually solve something like this :)

With that said, here is the beginning of my attempt at adding support for the battery backed sram available for several of the stm32 families.

### Example usage:
```rust
#![no_std]
#![no_main]

use defmt::*;
use embassy_stm32::pac::pwr::vals::Retention;
use embassy_stm32::{backup_sram, Config};
use embassy_time::Timer;

#[embassy_executor::main]
    config.rcc.ls.backup_ram_retention = Retention::PRESERVED;

    info!("Started!");

    let (bytes, status) = backup_sram::init(p.BKPSRAM);

    match status {
        backup_sram::Status::BackupRamDisabled => info!("BKPSRAM just enabled"),
        backup_sram::Status::AlreadyActive => info!("BKPSRAM already enabled"),
    }

    loop {
        info!("byte0: {}", bytes[0]);
        bytes[0] = bytes[0].wrapping_add(1);
        Timer::after_millis(500).await;
    }
}
```

<details>
    <summary>First (no longer representative) attempt:</summary>
    ```rust
    #[repr(C)]
    struct MyStruct {
        foo: u32,
    }

    // This will be run the first time to set the defaults and on struct layout missmatch
    let fn_to_setup_defaults = |_reason: ReasonForDataReset| MyStruct { foo: 42 };

    // Here we get a mutable reference to the value stored in battery backed ram
    let x: &'static mut MyStruct = backup_sram::init::<MyStruct>(p.BACKUP_RAM, fn_to_setup_defaults);

    loop {
        x.foo += 1; // These writes will persist resets and loss of vcc
        defmt::info!("I have ran {} iterations since I was programmed/battery was inserted", x.foo);
    }
    ```
    **see [examples/stm32h5/src/bin/backup_sram.rs](https://github.com/embassy-rs/embassy/pull/4663/files#diff-f103015bff8cb3773f1339f98ea4d0e2f99385e9486dfdfafa3a6e0cb97d7bef) instead**
</details>

### DIFF
--- a/embassy-stm32/src/rcc/bd.rs
+++ b/embassy-stm32/src/rcc/bd.rs
@@ -1,8 +1,8 @@
-#[cfg(rcc_h5)]
+#[cfg(backup_sram)]
 use core::sync::atomic::AtomicBool;
 use core::sync::atomic::{compiler_fence, Ordering};
 
-#[cfg(rcc_h5)]
+#[cfg(backup_sram)]
 use stm32_metapac::pwr::vals::Retention;
 
 use crate::pac::common::{Reg, RW};
@@ -14,7 +14,7 @@ pub const LSI_FREQ: Hertz = Hertz(40_000);
 #[cfg(not(any(stm32f0, stm32f1, stm32f3)))]
 pub const LSI_FREQ: Hertz = Hertz(32_000);
 
-#[cfg(rcc_h5)]
+#[cfg(backup_sram)]
 pub(crate) static WAS_BKPSRAM_ALREADY_POWERED_BY_BATTERY: AtomicBool = AtomicBool::new(false);
 
 #[allow(dead_code)]
@@ -97,7 +97,7 @@ pub struct LsConfig {
     pub rtc: RtcClockSource,
     pub lsi: bool,
     pub lse: Option<LseConfig>,
-    #[cfg(rcc_h5)]
+    #[cfg(backup_sram)]
     pub backup_ram_retention: Retention,
 }
 
@@ -123,7 +123,7 @@ impl LsConfig {
                 peripherals_clocked: false,
             }),
             lsi: false,
-            #[cfg(rcc_h5)]
+            #[cfg(backup_sram)]
             backup_ram_retention: Retention::LOST,
         }
     }
@@ -133,7 +133,7 @@ impl LsConfig {
             rtc: RtcClockSource::LSI,
             lsi: true,
             lse: None,
-            #[cfg(rcc_h5)]
+            #[cfg(backup_sram)]
             backup_ram_retention: Retention::LOST,
         }
     }
@@ -143,7 +143,7 @@ impl LsConfig {
             rtc: RtcClockSource::DISABLE,
             lsi: false,
             lse: None,
-            #[cfg(rcc_h5)]
+            #[cfg(backup_sram)]
             backup_ram_retention: Retention::LOST,
         }
     }


Dirbaio:
nice! a few tips:

- it'd be better to add the the address+size  of the backup sram to stm32-data here, so we can then use it from embassy-stm32 without massive cfg spam. https://github.com/embassy-rs/stm32-data/blob/main/stm32-data-gen/src/memory.rs
- it's simpler to just offer raw byte access. Instead of `unsafe trait Storable` and checksums I'd just offer the user `&mut [u8]` (if the hardware allows it, otherwise read/write fns could work), and let the user take care of serialization/checksumming themselves. This way is less opinionated, it allows the user to do more things that wouldn't be possible with this "store struct, load struct" api.

usbalbin:
> it'd be better to add the the address+size of the backup sram to stm32-data here, so we can then use it from embassy-stm32 without massive cfg spam. https://github.com/embassy-rs/stm32-data/blob/main/stm32-data-gen/src/memory.rs

Thanks! Something like this? https://github.com/embassy-rs/stm32-data/pull/671

> it's simpler to just offer raw byte access [...]

Fair enough :)

embassy-stm32/src/backup_sram.rs
```diff
@@ -0,0 +1,62 @@
...
+    static mut BKPSRAM: [MaybeUninit<u8>; BKPSRAM_SIZE];
+}
+
+fn is_bkpsram_powered_by_battery() -> bool {
+    todo!()
+}
+
+/// Setup battery backed sram
+pub fn init(_backup_sram: Peri<'static, BKPSRAM>) -> (&'static mut [u8], Status) {
+    // TODO: Enable RCC bits(or is this done by the RTC? If so require a reference to an RTC instance to ensure that it has been setup)
```
usbalbin, line 32: @denenilsson, I believe you had some code for this and the `Set RCC bits to enable battery power to BKPSRAM`? :)

embassy-stm32/src/rcc/bd.rs
```diff
@@ -193,6 +204,16 @@ impl LsConfig {
...
 
+        // Enable backup regulator for peristent battery backed sram
+        #[cfg(rcc_h5)]
+        crate::pac::PWR.bdcr().modify(|w| w.set_bren(self.backup_ram_retention));
+
+        #[cfg(rcc_h5)]
+        if self.backup_ram_retention == Retention::PRESERVED {
+            // Wait for backup regulator voltage to stabilize
+            while !crate::pac::PWR.bdsr().read().brrdy() {}
+        }
```
usbalbin, line 225: Do you want this in this file or should it be in backup_sram.rs?

Also, related question, I think it would be nice to let the user know if the retention was already enabled(SRAM contains data since last run) or if not. I think that would probably be easier if this was done as part of `backup_sram::init()`?

usbalbin: >  think it would be nice to let the user know if the retention was already enabled

Done using `WAS_BKPSRAM_ALREADY_POWERED_BY_BATTERY`

embassy-stm32/build.rs
```diff
@@ -122,6 +124,11 @@ fn main() {
         singletons.push(p.name.to_string());
     }
 
+    if has_bkpsram {
+        singletons.push("BKPSRAM".to_string());
+        cfgs.declare("backup_sram");
```
usbalbin: Am I doing this wrong? The `#[cfg(backup_sram)]` in [embassy-stm32/src/lib.rs](https://github.com/embassy-rs/embassy/pull/4663/files#diff-ea44d5093ee80a6c0f485f764d7a6be5eabae9c398408544ea07725938d61858) does not seem to activate even if I remove the `if has_bkpsram` check here.

usbalbin:
I will try to test in the near future.

In the mean time, what do you think of this now?

usbalbin:
This has now been tested on a custom board with an stm32h523ret with a clock battery

usbalbin:
@Dirbaio, is this better? Also let me know and I'll rebase and clean up the history

embassy-stm32/src/rcc/bd.rs
```diff
@@ -1,5 +1,10 @@
+#[cfg(backup_sram)]
+use core::sync::atomic::AtomicBool;
 use core::sync::atomic::{compiler_fence, Ordering};
 
+#[cfg(backup_sram)]
+use stm32_metapac::pwr::vals::Retention;
```
usbalbin, line 6: Since this is exposed in the config, do you want this reexported or exposed as a new type?

usbalbin:
Closed by #4813

PR #4667, opened 2025-09-15
Unify derivation of Clone, Copy and defmt::format for all *SPI configs

goodhoko:
Hi!

I'm in the middle of improving the error handling in OSPI and noticed this detail that makes the API tiny bit more cumbersome than it needs to be.

One can't use the same config for more than one transfer because read/write methods take it by ownership and `TransferConfig` doesn't implement `Clone` neither `Copy`.

This PR unifies what we derive for all config structs in all SPI implementations in `embassy-stm32` to `Clone`, `Copy` and `defmt::format` if the `defmt` feature is enabled.

It'd be best if could also change the read/write methods to merely borrow `TransferConfig`. None of them really need the ownership so it's just a question of changing the function signatures. However that would be a breaking change. Let me know if that's a desirable change and I'll be happy to submit a PR.


goodhoko:
@Dirbaio Thanks for the quick approve. I had to do my homework and propagate some of the `derive`s down to the other enums. Please take another look. The CI is all green now!

PR #4669, opened 2025-09-16
embassy-stm32: Optional pins for QSPI

phycrax:
This lets users save some pins where QSPI is being used for other purposes.


Dirbaio:
Have you tried passing a `None`? I think it'll fail with "type annotations needed" because the compiler still needs to know the type of the `impl XxxPin`s.

This is the reason drivers usually have one constructor for each possible set of pins you might want to use, instead of using `Option`.

phycrax:
Wow TIL, I should have tried. Thanks!

PR #4673, opened 2025-09-16
Add embassy-microchip HAL

felipebalbi:
A real embedded controller MCU.

### DIFF
--- a/docs/pages/microchip.adoc
+++ b/docs/pages/microchip.adoc
@@ -11,3 +11,5 @@ The following peripherals have a HAL implementation at present
 * GPIO
 * I2C
 * PWM
+* TACH
+* Time driver (using the CCT)


felipebalbi:
Any suggestions as to why build fails with this PR?

felipebalbi:
> Any suggestions as to why build fails with this PR?

caused by missing embassy package metadata.

embassy-microchip/src/i2c.rs
```diff
@@ -0,0 +1,1183 @@
...
+    /// Address register and General Call Match to be disabled.
+    pub addr2: Option<u8>,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            speed: BusSpeed::Standard,
+            addr1: Some(0x7f),
+            addr2: Some(0x7f),
```
mkj: Would `None` be a better default, and set `gc_dis=1` by default? (see below). I guess not too important now given multi-master isn't supported.

embassy-microchip/src/i2c.rs
```diff
@@ -0,0 +1,1183 @@
...
+
+            r.wctrl().write(|w| w.set_pin(true));
+            r.own_addr().write(|w| {
+                w.set_addr1(addr1);
+                w.set_addr2(addr2);
+            });
+
+            r.cfg().write(|w| {
+                if addr1 == 0 || addr2 == 0 {
+                    w.set_gc_dis(false);
```
mkj, line 492: Only if `config.addr1 == Some(0) || config.addr2 == Some(0)`?

examples/microchip/src/bin/i2c.rs
```diff
@@ -0,0 +1,36 @@
...
+use embassy_executor::Spawner;
+use embassy_microchip::i2c::I2c;
+use embassy_time::Timer;
+use {defmt_rtt as _, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    unsafe {
+        let cp = cortex_m::Peripherals::steal();
+        cp.SCB.vtor.write(0x000c_0000);
```
mkj: Could use `cortex-m-rt` `set-vtor` feature?

embassy-microchip/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+///
+/// This returns the peripheral singletons that can be used for creating drivers.
+///
+/// This should only be called once at startup, otherwise it panis.
+pub fn init(_config: config::Config) -> Peripherals {
+    let peripherals = Peripherals::take();
+
+    unsafe {
+        // ROM code leaves interrupts globally disabled, d'oh.
+        cortex_m::interrupt::enable();
```
mkj, line 518: :+1:

mkj (commented):
Looks good. 
Do the mec/cec parts share many peripherals with any other Microchip parts (atsamd, pic32c, polarfire)? From what I could tell they are fairly standalone. I wonder if the crate should be named something more specific.

felipebalbi:
> Looks good. Do the mec/cec parts share many peripherals with any other Microchip parts (atsamd, pic32c, polarfire)? From what I could tell they are fairly standalone. I wonder if the crate should be named something more specific.

They seem super different from anything else, but the HAL can handle the differences internally, much like the various versions of IP blocks in STM32 family. If folks prefer a different name, I'll rename, no problem :)

examples/microchip/src/bin/i2c.rs
felipebalbi: nice, done

embassy-microchip/src/i2c.rs
felipebalbi: Yeah, this is an interesting part of microchip's i2c controller. The thing is always acting as both slave and master. When working as a slave, it can respond to two different addresses. If any of those addresses is 0, then it can't respond to a General Call.

I will default to `None` and disable General Call by default, but that means `own_addr` above will write `0` to both addresses.

That being said, this condition here will likely remain the same because we only have to disable General Call matching when any of the slave addresses match the general call address (0)

embassy-microchip/src/i2c.rs
felipebalbi: done

embassy-microchip/src/i2c.rs
mkj: I thought it was the other way around, you set an address to 0 (and have gc_dis=0) for responding to general call?

> The address ‘0000000b’ is the General Call Address. An Own_Ad-
dress of 0h will match the General Call Address unless this match is
disabled by setting GC_DIS in the Configuration Register

https://ww1.microchip.com/downloads/en/DeviceDoc/00002379A.pdf

felipebalbi: The way I read this is:

If you set address to 0, your device will respond to general call. If you disable general call (gc_dis=1), then you can respond to address 0 as if it were a regular slave address.

No?

9names:
The reason why all the stm32's are in 1 hal is there *is* a ton of overlap between the chips..
It makes sense to keep them together.

Does dealing with the differences between all the families microchip currently own (avr, sam, kinetis, PIC32, mec17 etc) in a single HAL make things easier or harder?

embassy-microchip/src/i2c.rs
mkj: That's possible, but then there wouldn't be any way to configure it to avoid acking on address 0 (either with GC or regular slave interrupt status)? I'll check on hardware (cec1736) when I get a chance. It's not a blocker for merging this, can be adjusted later.

embassy-microchip/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+/// Macro to bind interrupts to handlers.
+///
+/// This defines the right interrupt handlers, and creates a unit struct (like `struct Irqs;`)
+/// and implements the right [`Binding`]s for it. You can pass this struct to drivers to
+/// prove at compile-time that the right interrupts have been bound.
+///
+/// Example of how to bind one interrupt:
+///
+/// ```rust,ignore
+/// use embassy_rp::{bind_interrupts, usb, peripherals};
```
i509VCB:
```suggestion
/// use embassy_microchip::{bind_interrupts, usb, peripherals};
```

For the peripheral to bind in the example, can't really use USB for now. Maybe point it at I2C?

embassy-microchip/src/pwm.rs
```diff
@@ -0,0 +1,226 @@
...
+        match self {
+            PwmError::InvalidDutyCycle => ErrorKind::Other,
+        }
+    }
+}
+
+/// PWM driver.
+pub struct Pwm<'d, T: Instance> {
+    _peri: Peri<'d, T>,
+    _pin: Peri<'d, AnyPin>,
```
i509VCB, line 74: You probably need to deconfigure the pin on drop. Probably throw the equivalent of `set_as_disconnected` on SealedPin?

embassy-microchip/src/pwm.rs
```diff
@@ -0,0 +1,226 @@
...
+	$(
+	    impl SealedInstance for peripherals::$peri {
+		#[inline(always)]
+		fn regs() -> Pwm0 {
+		    pac::$peri
+		}
+	    }
+
+	    impl Instance for peripherals::$peri {}
+	)*
```
i509VCB, line 193: Unfortunately rustfmt doesn't know how to deal with macros.

docs/pages/microchip.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+* link:https://github.com/OpenDevicePartnership/mec17xx-pac[mec17xx-pac]
+
+== Peripherals
+
+The following peripherals have a HAL implementation at present
+
+* GPIO
+* I2C
+* PWM
```
i509VCB, line 13: time driver?

embassy-microchip/src/lib.rs
felipebalbi: argh, finally sorted this one out. :)

embassy-microchip/src/pwm.rs
felipebalbi: solved

docs/pages/microchip.adoc
felipebalbi: fixed

embassy-microchip/src/i2c.rs
felipebalbi: I think this should be suitable the way it is now. Just finished rebasing to latest HEAD, vll go through the build failures and try to get this one merged.

embassy-microchip/src/pwm.rs
felipebalbi: I definitely need to improve that. I'll add it to my TODO list as something I need to fix with a certain priority. I think I need to change how I define pins and where some methods exist, otherwise I'll always have some duplication. If it's okay, I'll merge the HAL as is and fix it in a follow-up PR.

kurtjd (approved):

PR #4679, opened 2025-09-18
Adding IO drive strength constructor to Mco.

denenilsson:
adding rcc::Mco::new_with_speed() constructor to control to allow lower drive strength on generated master clock out signal.

Simple usage example added.

### DIFF
--- a/embassy-stm32/src/rcc/mco.rs
+++ b/embassy-stm32/src/rcc/mco.rs
@@ -91,16 +91,28 @@ pub struct Mco<'d, T: McoInstance> {
 
 impl<'d, T: McoInstance> Mco<'d, T> {
     /// Create a new MCO instance.
-    pub fn new(peri: Peri<'d, T>, pin: Peri<'d, impl McoPin<T>>, source: T::Source, prescaler: McoPrescaler) -> Self {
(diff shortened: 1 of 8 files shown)


Dirbaio:
I think it'd be better to change `new` to take a non_exhaustive Config struct similar to other drivers. It'd have "prescaler" and "speed" as fields.

denenilsson:
Maybe like this?

embassy-stm32/src/rcc/mco.rs
```diff
@@ -91,16 +91,28 @@ pub struct Mco<'d, T: McoInstance> {
...
-            set_as_af!(pin, AfType::output(OutputType::PushPull, speed));
+            T::_apply_clock_settings(source, config.prescaler);
+            set_as_af!(pin, AfType::output(OutputType::PushPull, config.speed));
         });
 
         Self { phantom: PhantomData }
     }
 }
+
+pub struct McoConfig {
```
Dirbaio, line 105: can you mark this with `#[non_exhaustive]`? This makes it so we can add more fields in the future without it being a breaking change.

denenilsson:
Sure can :)

Dirbaio:
can you rebase to fix conflicts and get CI to pass?

denenilsson:
Should be ok now?

PR #4685, opened 2025-09-19
feat(nrf/spim): erase Instance type from Spim

Abrahamh08:
It was between this and adding an AnySpi (which would be theoretically more memory efficient but effectively the same), but it seems that embassy is going this type erasure route in general.

This can be made a few bytes more efficient by doing a long match on Interrupt to retrieve the instance, but that seems annoying to me.


Dirbaio:
very cool! Can you fix CI?

PR #4686, opened 2025-09-20
add basic RTC driver for nRF

robamu:
My first driver contribution, so I was not sure about the policy of API like `steal` or `regs`, but I really like exposing those for maximum flexbility.

I could introduce an additional error which denies CC[3] manipulation for RTC0 (it does not seem to be implemented for all nRF families I checked), but I was not sure about that.

I tested the example app on a nRF52840 board.


embassy-nrf/src/rtc.rs
```diff
@@ -0,0 +1,272 @@
...
+            return Err(PrescalerOutOfRangeError(prescaler));
+        }
+
+        T::regs().prescaler().write(|w| w.set_prescaler(prescaler as u16));
+        Ok(Self(rtc))
+    }
+
+    /// Create a new `Rtc` driver, configuring it to run at the given frequency.
+    pub fn new_for_freq(rtc: Peri<'d, T>, freq_hz: u32) -> Result<Self, PrescalerOutOfRangeError> {
+        let prescaler = (32_768_000 / freq_hz).saturating_sub(1);
```
robamu: this is actually wrong.. need to fix this.

Dirbaio:
looks good!

We should make the RTC1 singleton disappear if the feature `time-driver-rtc1` is enabled, so you can't try to use it at the same time.

robamu:
Oh, that is a good idea.

Does the register access have to be featute gated behind "unstable-pac"?

Dirbaio:
> Does the register access have to be featute gated behind "unstable-pac"?

yes

robamu:
Also, additional note: I could have just stolen the RTC peripheral in the interrupt, circumventing the whole mutex shenanigans, which should be safe because the RTC is not used in the main thread anymore. Do you think this is okay for the example, or would you prefer the Mutex way, to avoid showcasing something which can be possibly unsound under other circumstances?

lulf (approved):
I think the current example is preferrable, thanks you for contributing this!

PR #4687, opened 2025-09-20
nrf: add persist() method for gpio and ppi

0e4ef622:
Would've also added to Timer but Timer doesn't impl Drop, is that a bug or a feature?


Dirbaio:
looks good, thanks!

Yes, timers should have a drop.

PR #4688, opened 2025-09-21
Use msplim for RP2350 stackguard

9names:
The arm cores on RP2350 are cortex-M33, which has a hardware stack check function.
Use msplim to set the limit for the main stack pointer instead of using the MPU.

Also added an example to check that the stack protection is working.

Note: my confidence that this works around the limits is pretty low.
If you set the example array length as so:
```
let t = [0u8; CORE1_STACK_LENGTH - 200];
```
core0 stops responding via defmt, which shouldn't happen. The existing stack guard is also not perfect so maybe not the end of the world, but i'd feel more condfident if we had a better test.


9names:
I switched this from
```rust
cortex_m::register::msplim::write(stack_bottom as u32);
```
to
```rust
cortex_m::register::msplim::write(stack_bottom.addr() as u32);
```

I think that's more idiomatic? We don't need provenance because it's never becoming a pointer again.

jamesmunns:
> core0 stops responding via defmt, which shouldn't happen

I wonder if core1 is holding/thrashing the mutex used by defmt? Though if you still see the LED blinking in the example, then that's definitely odd.

Issue #2997, opened 2024-05-24
USB HID Keyboard not working in BIOS (before Linux start)


dariogoetz:
MCU: STM32F401CC

I have written an embassy-based keyboard firmware for a keyboard. The firmware works well in Linux. However, the keyboard does not produce anything when the laptop asks for a system password when booting (before Linux starts).

I could reproduce the issue with the code from the examples directory `examples/stm32f4/src/bin/usb_hid_keyboard.rs` (adjusted for clocks of the STM32F401CC, otherwise vanilla): Within my Linux system, it produces an `a` when pulling PC13 up. In the password prompt before starting Linux, however, nothing happens.

I have another firmware based on `rtic`, `stm32f4xx-hal`, and `usb-device` that does not have this issue, so I would exclude it being a hardware issue.

Any ideas, what the issue could be?

dariogoetz:
Thank you very much for the feedback. I tried various different combinations, also setting different values for `config.device_class`  (the default value and `0x03`) and `config.composite_with_iads` while fixing `config.device_sub_class=0x01`. None worked, unfortunately.

When you wrote "f your device supports the boot protocol", what is the "device" you speak of? I am using an STM32F401CCU6 black pill board. Is that what you mean? Apart from that, the various requirements described in the F and G sections of the document that you linked refer to software-related things that are (or should be) implemented in `embassy-usb`. Maybe I got that wrong, though.

PR #4691, opened 2025-09-21
Add support for changing hid protocol mode

matteo-meluzzi:


matteo-meluzzi:
https://github.com/embassy-rs/embassy/issues/2997

matteo-meluzzi:
Mouse example running in the BIOS


https://github.com/user-attachments/assets/59f18804-eebe-480f-a85d-e2fd62cff854

embassy-usb/src/class/hid.rs
```diff
@@ -31,6 +38,30 @@ const HID_REQ_SET_REPORT: u8 = 0x09;
 const HID_REQ_GET_PROTOCOL: u8 = 0x03;
 const HID_REQ_SET_PROTOCOL: u8 = 0x0b;
 
+/// Get/Set Protocol mapping
+/// See (7.2.5 and 7.2.6): <https://www.usb.org/sites/default/files/hid1_11.pdf>
+#[cfg(not(feature = "usbd-hid"))]
```
HaoboGu: I think just defining `HidProtocolMode` in `embassy-usb` is fine, no need to import it from `usbd-hid`.

embassy-usb/src/class/hid.rs
```diff
@@ -106,13 +137,15 @@ fn build<'d, D: Driver<'d>>(
     state: &'d mut State<'d>,
     config: Config<'d>,
     with_out_endpoint: bool,
+    usb_subclass: u8,
+    usb_protocol: u8,
 ) -> (Option<D::EndpointOut>, D::EndpointIn, &'d AtomicUsize) {
     let len = config.report_descriptor.len();
 
-    let mut func = builder.function(USB_CLASS_HID, USB_SUBCLASS_NONE, USB_PROTOCOL_NONE);
+    let mut func = builder.function(USB_CLASS_HID, usb_subclass, usb_protocol);
```
HaoboGu: Is there any particular reason that `config.device_sub_class` and `config.device_protocol` are not used here?

embassy-usb/src/class/hid.rs
matteo-meluzzi: fair point, i removed the import

embassy-usb/src/class/hid.rs
matteo-meluzzi: @HaoboGu, I believe that the `config.device_sub_class` and `config.device_protocol` you're referring to belong to the USB device config (` embassy_usb::Config`), while the build() function has a HID class config (`embassy_usb::class::hid::Config`) as parameter.

Now that I read your comment, I realize that it would be better if the `usb_subclass` and `usb_protocol` were specified in `embassy_usb::class::hid::Config`. I made the necessary enum definitions and updated `embassy_usb::class::hid::Config`. The issue with this is that it's not backwards compatible. I hope that's fine.

embassy-usb/src/class/hid.rs
```diff
@@ -31,6 +29,52 @@ const HID_REQ_SET_REPORT: u8 = 0x09;
...
+    }
+}
+
+/// USB HID interface subclass values.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u8)]
+pub enum HidSubclass {
+    /// Only report mode is supported.
+    ReportOnly = 0,
```
HaoboGu: I suggest to use the name from [hid spec](https://www.usb.org/sites/default/files/documents/hid1_11.pdf), aka `No` and `Boot`

matteo-meluzzi: @HaoboGu , that's a nice idea. I updated the names as you suggested. I also updated all `usb_hid_mouse` and `usb_hid_keyboard` examples.

HaoboGu:
lgtm!

matteo-meluzzi:
> lgtm!

@HaoboGu i synced with main, we should be good to go

HaoboGu (commented):
It looks really good to me, but I'm not the person that has the power to merge PRs. Maybe Dirbaio wants to have a look too.

matteo-meluzzi:
Ah I see. @Dirbaio would you mind having a look as well?

examples/rp/src/bin/usb_hid_keyboard.rs
```diff
@@ -135,6 +161,18 @@ impl RequestHandler for MyRequestHandler {
         OutResponse::Accepted
     }
 
+    fn get_protocol(&self) -> HidProtocolMode {
+        let protocol = HidProtocolMode::from(HID_PROTOCOL_MODE.load(Ordering::Relaxed));
+        info!("The current HID protocol mode is: {}", protocol);
+        protocol
+    }
+
+    fn set_protocol(&mut self, protocol: HidProtocolMode) -> OutResponse {
```
lulf, line 170: Should these log statements perhaps be removed or ~~less verbose~~ at the trace or debug log level ?

matteo-meluzzi: I think that having verbose logging in the examples is not necessarily a bad thing. A newcomer might benefit from slightly verbose logs. The other functions in `impl RequestHandler for MyRequestHandler` also have `info!` messages:
```
struct MyRequestHandler {}

impl RequestHandler for MyRequestHandler {
    fn get_report(&mut self, id: ReportId, _buf: &mut [u8]) -> Option<usize> {
        info!("Get report for {:?}", id);
        None
    }

    fn set_report(&mut self, id: ReportId, data: &[u8]) -> OutResponse {
        info!("Set report for {:?}: {=[u8]}", id, data);
        OutResponse::Accepted
    }

    fn get_protocol(&self) -> HidProtocolMode {
        let protocol = HidProtocolMode::from(HID_PROTOCOL_MODE.load(Ordering::Relaxed));
        info!("The current HID protocol mode is: {}", protocol);
        protocol
    }
...
```
Plus the `get_protocol` function is called only once during startup and the `set_protocol` is called only when plugging the USB device in a computer while in the BIOS, so they are not spammy logs.

matteo-meluzzi: If you feel strongly against the logs I will remove them though

PR #4697, opened 2025-09-22
lpc55: spi master example for nxp-pac

LithDefaux:
- Made a working example for initializing and transferring data using spi for lpc55 with nxp-pac.


LithDefaux:
@george-cosma

i509VCB:
I assume this is still being worked on since its rather PAC heavy in the example?

LithDefaux:
Yes, I am currently working on the proper SPI driver

Dirbaio:
i'm going to close this since it's not an example of using the HAL's SPI driver. Ideally we'd have the SPI driver then write an example for it. There's not much point in merging an example using the PAC and later changing it.

PR #4702, opened 2025-09-23
Add read_from_break method to STM32 UartRx

gdyr:
For the purposes of receiving DMX (ANSI E1.11), or other protocols that may signal the start of a message with a break character:

This PR adds a new read method to `UartRx`, called `read_from_break`, with the following logic:
1. DMA transaction is set up but DMAR bit is not set
2. Overrun errors are ignored.
3. In UART interrupt handler, if a framing error (break) is detected and DMAR bit is not set, clear the error and set DMAR

Other logic remains the same, i.e. if a second break occurs before the read is complete, this is still a framing error.

I've validated this on an STM32L476RG (usart_v3) but no other chips or peripheral versions so far.
I did attempt to implement this outside of embassy code, but found the first few bytes were consistently missed unless handling this in the interrupt.

Please let me know if you think there's a better approach, or would like any changes!


Dirbaio:
Hello,

PR #4717, opened 2025-09-28
stm32/timer: Support 32-bit timers in SimplePwm waveform_up method

KanoPeng233:
Add TimerBits matching following waveform method pattern to handle both 16-bit and 32-bit timer DMA transfers with appropriate pointer types.


xoviat:
@chasingRs can you add a changelog entry?

KangjianPeng:
@xoviat Of course! I've already added it. Thank you.

PR #4718, opened 2025-09-28
Reset SAADC for nrf52, workaround for anomaly 241.

mike-kfed:
I ran into [nrf52 Anomaly 241](https://docs.nordicsemi.com/bundle/errata_nRF52810_Rev1/page/ERR/nRF52810/Rev1/latest/anomaly_810_241.html) writing a low power BLE firmware. I added an example for nrf52810 to illustrate the problem. I used the nrf52 DK and the Power Profiler Kit II (PPK2) to confirm the current usage.

It only happens when more than one analogue pin is sampled. Therefore a reset should be optional. My initial idea was to reset in the Drop impl, but that would not be optional. Then I came up with the following: add a reset method that consumes the SAADC instance. This forces the programmer to deal with setting up the SAADC again if needed, conserving power if needed (on a constant power source device one likely doesn't care).


Dirbaio:
Thanks for the PR!

I'd prefer to integrate the workaround in drop, so it Just Works for everyone without them having to be aware of the errata and the workaround. It's true it makes it non-optional but it's just a handful of register writes that take a few clock cycles each. Also, doing the reset when not needed shouldn't hurt.

mike-kfed:
Over night I realised that a Drop impl is not optimal for other reasons too, imagine the following legitimate pseudo code

```rust
let saadc = Saadc::new();
loop {
    saadc.sample();
    sleep(from_secs(60));
}
```

drop will never be called because saadc is never dropped. And if we wanted to reset using drop, it probably causes the end-user to write a comment that drop resets the saadc anyways, might as well call the method reset() then?

```rust
loop {
    let saadc = Saadc::new();
    saadc.sample();
    core::mem::drop(saadc); // doing a reset to conserve energy
    sleep(from_secs(60));
}
```

in the above example without the explicit drop or an explicit limited scope the reset wont happen before the sleep() either, again wasting energy unbeknownst to the user.

The only way, I can think of, to hand-hold the end-user and force the reset of the saadc at compile time would be to change the signature of sample() to consume the instance. But then I'd think there are legitimate use cases where you want to sample multiple times before resetting the saadc.

I'd love for something that somehow guides the end user at compile time, but haven't found the optimal solution. Let me know what you think. I can change it to a Drop impl, but please consider the fact that either way the end-user has to know about it and figure out the API and manually apply it in both cases to avoid energy consumption.

Dirbaio:
We're already using the "to reduce power consumption, drop the driver" convention in other drivers (uart, twi, spi, qspi). See [example](https://github.com/embassy-rs/embassy/blob/main/examples/nrf52840/src/bin/twim_lowpower.rs), [example](https://github.com/embassy-rs/embassy/blob/main/examples/nrf52840/src/bin/qspi_lowpower.rs). SAADC should work the same way.

mike-kfed:
true, I rebased and added it to the Drop impl

Dirbaio:
thanks!

Issue #2948, opened 2024-05-16
Expose peripheral registers when enabling the unstable-pac feature.


liarokapisv:
Foremost, thanks a ton for your great work!

The HAL cannot cover all use-cases, and at some point users may need to access the registers directly.
This can currently be achieved by enabling the `unstable-pac` feature and accessing the registers.
Unfortunately, there is currently no way for the user to associate the HAL peripherals with their registers
meaning that they cannot abstract over the peripherals themselves, since the registers are exposed only through the Sealed traits.

I propose exposing the `regs` values from the Sealed traits when enabling the `unstable-pac` feature. Just the registers
should be enough, there is no reason to enforce any extra public API requirements like exposing the helper Sealed trait methods 
(although I personally would find that very useful as long as the HAL does not make any promises about the compatibility between versions). Note that this is already the case with some peripherals, like `AnyPin` that allow exposing their `GPIO` port through the `block` method.

This allows users to bypass the HAL locally when needed without having to change the external interface of the code.

As an example, I demonstrate one of my recent use cases:

I am using the following DACs through `SPI`: https://www.rohm.com/products/data-converter/d-a-converters/10bit-d-a/bu22210muv-product.
I need to run 3 `SPI` daisy-chains with x2 of them in each one and approach ~10khz update rate for each channel.

This is implemented on the `STM32H745IIx` by using  `DMA` Streams triggered by a timer and chained `DMAMUX` requests to trigger `SPI DR` register writes and synchronized `GPIO BSRR` register writes for the `CS` handling.

The implementation, at a high-level, looks like this:

```
TIM CH1 -> DMA1_CH1 -> SPI1 DR (also triggers DMAMUX's GEN1)
TIM CH2 -> DMA1_CH2 -> SPI1_CS BSRR with proper PIN values to set/reset (also triggers DMAMUX's GEN2)

DMAMUX_GEN1 -> DMA1_CH3 -> SPI2 DR
DMAMUX_GEN1 -> DMA1_CH5 -> SPI3 DR

DMAMUX_GEN2 -> DMA1_CH4 -> SPI2_GPIO BSRR
DMAMUX_GEN2 -> DMA1_CH6 -> SPI3_GPIO BSRR.

```

This is set up to run in circular mode without interrupts, writing to the DACs is achieved by writing to proper buffer positions
associated with the SPI DR writes.

[...]

PR #4719, opened 2025-09-28
nrf: expose regs on all peripherals

0e4ef622:
This makes it easier to implement generic drivers externally. Mainly interested in timer but I did it for all of them.

Related: #2948

### DIFF
--- a/embassy-nrf/CHANGELOG.md
+++ b/embassy-nrf/CHANGELOG.md
@@ -11,12 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - changed: Remove `T: Instance` generic params in all drivers.
 - changed: nrf54l: Disable glitch detection and enable DC/DC in init.
 - changed: Add embassy-net-driver-channel implementation for IEEE 802.15.4
-- changed: add persist() method for gpio and ppi
 - added: basic RTC driver
 - changed: add persist() method for gpio, gpiote, timer and ppi
 - changed: impl Drop for Timer
 - added: expose `regs` for timer driver
 - added: timer driver CC `clear_events` method
+- added: expose `regs` for all peripherals
 
 ## 0.7.0 - 2025-08-26
 
--- a/embassy-nrf/src/egu.rs
+++ b/embassy-nrf/src/egu.rs
@@ -38,6 +38,7 @@ impl<'d> Egu<'d> {
 }
 
 pub(crate) trait SealedInstance {
+    #[cfg(not(feature = "unstable-pac"))]
     fn regs() -> pac::egu::Egu;
 }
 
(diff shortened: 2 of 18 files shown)


i509VCB:
I do think exposing `regs` on types is a good idea. Although I'm not sure if we want to make this require `unstable-pac` (not just for nRF, but all HALs).

An application may pin some version of nrf-pac/whatever-pac and just needs to know the address for some peripheral singleton. That could be exposed regardless of the `unstable-pac` feature.

This makes me lean more towards just giving the user a unit pointer `*mut ()`. And let the application explicitly cast as needed to the pinned version or the reexport of the pac if `unstable-pac` is enabled.

0e4ef622:
Hmm, that's an interesting point. Are there any downsides other than ergonomics? Most I can think of is if someone writes a driver against one chip and someone else tries to use it with another chip that has registers at different offsets, that wouldn't become a compiler error.  That doesn't seem to be an issue with nRF at least, not sure about other families.

lulf:
I think any API from the underlying should be guarded behind the unstable-pac signalling that we reserve the right to break the API at any time.

Even so, I think leaking these internals defeats the purpose of having a HAL in the first place. Ideally most functionality should be covered within the HAL, and for the few cases it isn't, you can use the pac directly. Exposing the regs could easily promote some bad patterns. It makes it easier to footgun yourself where the HAL assumes ownership of the peripheral but some other part of your code could be manipulating the registers as well. It's always possible to do that, but the point is lets not make it too easy?

0e4ef622:
I can understand that. The main thing this makes possible (or at least not inconvenient) is accessing the regs from a generic interrupt handler that is usable with the `bind_interrupts!` macro.

```rust
pub struct InterruptHandler<T: Instance> {
    _phantom: PhantomData<T>,
}

impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandler<T> {
    unsafe fn on_interrupt() {
        // how to get regs?
    }
}
```

i509VCB:
> Exposing the regs could easily promote some bad patterns. It makes it easier to footgun yourself where the HAL assumes ownership of the peripheral but some other part of your code could be manipulating the registers as well. It's always possible to do that, but the point is lets not make it too easy?

I would agree that using the pointers to a peripheral correctly is the user's responsibility while considering ownership. If the HAL owns a peripheral and you need to do something PAC related that the HAL doesn't do (although a pull request would make sense here) or shouldn't expose (XBAR triggering is practically impossible for embassy-nxp to expose without unsafe, or purpose built drivers). 

The return type as a `*mut ()` makes sense because in user code it makes it blindingly clear that you are trying to access PAC related functionality because converting to the PAC types would be unsafe. While it's made easier, the `unsafe` blocks you will need to write as a user will shame you into asking "am I going to cause awful bugs".

PR #4720, opened 2025-09-29
QEI Pin Pull Up/Down Constructor

bschwind:
Related to #3721, and contains an alternate solution to part of #3728

This breaks from the convention of other drivers a bit, where they would specify the pull value in a `Config` struct for the peripheral. That's how it's being done in #3728, though that PR needs a rebase on `main`, and the change is a bit more complicated.

If we prefer #3728, then we could rebase it (I gave it a quick try just now though it was a bit non-trivial), or we could re-implement it on the latest `main`, as things have changed quite a lot in the 6 months since that PR was opened.

I also removed the call to `pin.set_low()` in the `QeiPin` constructor. I'm a bit confused on why the `pin.set_low();` call was needed. I tested a rotary encoder without it, and the count doesn't seem to be affected. Is there some other reason the pins need to be set low on construction?


Dirbaio:
We shouldn't break from the convention unless there's a good reason.

bschwind:
@Dirbaio fair enough! The diff was small enough that I thought it was worth a try to get the functionality in.

I'll give a harder try at rebasing #3728 or re-implementing the functionality on the current `main`.

PR #4721, opened 2025-09-29
fix: remove instance generic for multiwrite implementation

lulf:


Dirbaio:
heh a gap in CI coverage, nice

PR #4728, opened 2025-10-01
rp: add support for double-buffered DMA transfers

laptou:
I was trying to set up a Quectel LC86 GNSS receiver. It outputs a near-constant stream of UART data at 115200 bps, which is fast enough that other techniques for reading the data weren't fast enough (they resulted in constant overrun errors), so I needed a double-buffered DMA transfer. `embassy` doesn't exactly expose all of the APIs needed to do this in user code, so I forked it and added it myself. What do y'all think?


Dirbaio:
Hello!

This is the same as the DMA-powered buffered uart in stm32, rp. Can you make it consistent with them? The main difference is the buffered uart has a single large ringbuffer that DMA is constantly writing to, and the user calls `read()` to pop data off it. It's much more ergonomic because the user doesn't have to deal with any "buf views".

PR #4730, opened 2025-10-02
bump cortex-ar to v0.3

robamu:


embassy-executor/Cargo.toml
```diff
@@ -98,7 +98,7 @@ portable-atomic = { version = "1.5", optional = true }
 cortex-m = { version = "0.7.6", optional = true }
 
 # arch-cortex-ar dependencies
-cortex-ar = { version = "0.2", optional = true }
+cortex-ar = { version = ">=0.2, <=0.3", optional = true }
```
bugadani: This will always pick 0.3

robamu: even if some other crates in the dependency tree only allow 0.2?

bugadani: Yes, if there are multiple incompatible major versions in the dependency tree, cargo will use both.

robamu: I've skimmed over https://doc.rust-lang.org/cargo/reference/resolver.html#dependency-resolution again. Wouldn't this variant help the resolver with unification in that case? 

In any case, I can simply set this to 0.3

robamu: Okay, I can simply set it to 0.3

robamu: I've read https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility again

> The application of other constraints or heuristics may cause these to be unified, picking one version (e.g. 0.6.5). 

for

```toml
# Package A
[dependencies]
rand = ">=0.6,<0.8.0"

# Package B
[dependencies]
rand = "0.6"  # meaning `>=0.6.0,<0.7.0`
```

Can this be advantageous in some situations?

PR #4733, opened 2025-10-02
mspm0: add mspm0h321x support

i509VCB:
This also changes selection of GPIO interrupt type to be calculated by the buildscript. H321x is yet another exception with GPIOB being a physical interrupt rather than part of an interrupt group.

H3215/6 are true 5V parts. I wonder if that is a first for an embassy hal


embassy-mspm0/Cargo.toml
```diff
@@ -243,6 +245,7 @@ mspm0g3519pn = ["mspm0-metapac/mspm0g3519pn"]
 mspm0g3519pz = ["mspm0-metapac/mspm0g3519pz"]
 mspm0g3519rgz = ["mspm0-metapac/mspm0g3519rgz"]
 mspm0g3519rhb = ["mspm0-metapac/mspm0g3519rhb"]
+mspm0h3216pt = ["mspm0-metapac/mspm0h3216pt"]
```
i509VCB, line 249: There will probably be more packages in the future (including H3215). TI just hasn't announced them yet.

https://github.com/mspm0-rs/mspm0-data/issues/14 is tracking this

embassy-mspm0/Cargo.toml
```diff
@@ -15,17 +15,18 @@ publish = false
 [package.metadata.embassy]
 build = [
     {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0c1104dgs20", "time-driver-any"]},
+    {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g1107ycj", "time-driver-any"]},
+    {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g1505pt", "time-driver-any"]},
+    {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g1519rhb", "time-driver-any"]},
+    {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g3105rhb", "time-driver-any"]},
     {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g3507pm", "time-driver-any"]},
     {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0g3519pz", "time-driver-any"]},
+    {target = "thumbv6m-none-eabi", features = ["defmt", "mspm0h3216pt", "time-driver-any"]},
```
Iooon, line 25: Any specific reason to change the order?

Iooon (approved):
Looks good

embassy-mspm0/Cargo.toml
i509VCB: I prefer to match the order that features use, which is alphabetically sorted.

PR #4736, opened 2025-10-03
fix incomplete timer unqueuing that prevent any further timer in the same task

aurelj:
This fixes an issue when using never ending timers such as: `Timer::at(Instant::MAX)`

When such a timer is canceled and a another timer is initiated in the same task,
the new timer won't actually be added to the queue and thus will never wake.

The issue is that when the never ending timer is added to the queue, `next_expiration()`
is called and will simply remove the timer from the queue (it will never be reached
so there is no point in keeping it in the queue and checking it every time), but it
will not reset the waker to None.
So when the next timer is scheduled with `schedule_wake()`, a waker will already be
present and `schedule_wake()` will consider this item is already part of the queue
and won't add it in the queue anymore, so it will also never wake !

Here is an example the exhibit this issue. First time the `select()` will be called with
a never ending timer, but after the first edge change on input, `select()` is
called with a 500ms timer that should trigger a break of the loop if no edge happens
during the 500ms.
Currently, this loop will never end. The PR fixes this issue.

```
#[embassy_executor::task]
async fn stable_input(mut input: ExtiInput<'static>) {
    let mut deadline = None;

    loop {
        match select(
            input.wait_for_any_edge(),
            Timer::at(deadline.unwrap_or(Instant::MAX)),
        )
        .await
        {
            Either::First(()) => {
                deadline = Some(Instant::now() + Duration::from_millis(500));
            }
            Either::Second(_) => break,
        }
    }

    info!("stable input");
}
```

For the record, this bug was introduced in 74037f04933f4ec9a678e0b47fd6819e7c0489a9


embassy-time-queue-utils/src/queue_integrated.rs
```diff
@@ -130,6 +130,9 @@ impl Queue {
                 // Remove it
                 prev.set(item.next.get());
                 item.next.set(None);
+                // Ensure there is no waker left in items that are not part of the timer queue
+                // even when the waker was not called (for timer that never expires, ie expires_at = u64::MAX)
+                item.waker = None;
```
0e4ef622, line 136: This comment isn't very clear, the problem was `schedule_wake` uses the presence of a waker to determine if an item is part of the queue or not, and that invariant wasn't maintained.

aurelj: Indeed, reading this comment again now, I realize it wasn't very clear.
Updated it now. It is hopefully clearer.

embassy-time-queue-utils/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Fix usage of never ending timers
```
bugadani:
```suggestion
- Fixed an issue where never-ending timers were not correctly removed from the timer queue
```

aurelj:
> I would like a test in e.g. `tests\stm32\src\bin\timer.rs`.

Ok, I added a test that only pass with this fix. (The last Timer actually freeze without this fix)

>  but the changelog entry is a bit meh.

Yep, I included your improved changelog, thanks.

Dirbaio:
embassy-time-queue-utils v0.3.1 released with the fix.

PR #4738, opened 2025-10-04
lpc55: pwm simple

frihetselsker:
Hello! I've implemented the simple version of PWM for NXP LPC55.

For now, in `Cargo.toml` the commit from my fork of `nxp-pac` is used. Once [PR](https://github.com/i509VCB/nxp-pac/pull/11) will be merged, the appropriate commit will be added.


embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+            invert: false,
+            phase_correct: true,
+            enable: true,
+            divider: 255,
+            prescale_factor: 255,
+            compare: 1_000_000_000,
+            top: 2_000_000_000,
+        }
+    }
+}
```
eva-cosma: I am not super sure having a Default implementation makes sense for the user? There is no obvious reason why we would have top at `2_000_000_000` for example.

I do recognize that this gives us the advantage to have some reasonable values (i.e. `inverted: false`), on top of which we can modify the values. How do other embassy crates handle the creation of big structs?

Related question:  Would we need `divider` and `prescale_factor` at the same time?

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        // Reset SCTimer => Reset counter and halt it.
+        // Cannot reset it because it will reset all the other instances and will break the whole functionality
+        // when a new output is added.
+        // SYSCON
+        //     .presetctrl1()
+        //     .modify(|w| w.set_sct_rst(crate::pac::syscon::vals::SctRst::ASSERTED));
+        // SYSCON
+        //     .presetctrl1()
+        //     .modify(|w| w.set_sct_rst(crate::pac::syscon::vals::SctRst::RELEASED));
```
eva-cosma: Could you explain in more detail (or give an example) what the issue is here?

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+        // Stop and reset the counter
+        SCT0.ctrl().modify(|w| {
+            w.set_bidir(
+                0,
+                if config.phase_correct {
+                    crate::pac::sct0::vals::Bidir::UP_DOWN
+                } else {
+                    crate::pac::sct0::vals::Bidir::UP
+                },
+            );
```
eva-cosma:
```suggestion
            if config.phase_correct {
                w.set_bidir(0, crate::pac::sct0::vals::Bidir::UP_DOWN);
            } else {
                w.set_bidir(0, crate::pac::sct0::vals::Bidir::UP);
            }
```

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+        SYSCON.sctclkdiv().modify(|w| w.set_div(config.divider));
+
+        SCT0.config().modify(|w| {
+            w.set_unify(crate::pac::sct0::vals::Unify::UNIFIED_COUNTER);
+            w.set_clkmode(crate::pac::sct0::vals::Clkmode::SYSTEM_CLOCK_MODE);
+            w.set_noreload(0, true);
+            w.set_autolimit(0, true);
+        });
+
+        // Before settign the mathces, it has to be assured that compare value is than the top value
```
eva-cosma:
```suggestion
        // Before setting the match registers, we have to make sure that `compare` is lower or equal to `top`, otherwise <insert reason as to what would happen if we don't do this/ why it doesn't make sense> 
```

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+        });
+
+        // Event configuration
+        critical_section::with(|_cs| {
+            // If it is already set, don't change
+            if SCT0.ev(0).ev_ctrl().read().matchsel() != 15 {
+                SCT0.ev(0).ev_ctrl().modify(|w| {
+                    w.set_matchsel(15);
+                    w.set_combmode(crate::pac::sct0::vals::Combmode::MATCH);
+                    w.set_stateld(crate::pac::sct0::vals::Stateld::ADD); // STATE + statev, where STATE is a on-board variable.
```
eva-cosma:
```suggestion
                     // STATE + statev, where STATE is a on-board variable.
                    w.set_stateld(crate::pac::sct0::vals::Stateld::ADD);
```

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+                    w.set_statev(0); // = statev
+                });
+            }
+        });
+        SCT0.ev(output_number + 1).ev_ctrl().modify(|w| {
+            w.set_matchsel((output_number + 1) as u8);
+            w.set_combmode(crate::pac::sct0::vals::Combmode::MATCH);
+            w.set_stateld(crate::pac::sct0::vals::Stateld::ADD); // STATE + statev, where STATE is a on-board variable.
```
eva-cosma:
```suggestion
            // STATE + statev, where STATE is a on-board variable.
            w.set_stateld(crate::pac::sct0::vals::Stateld::ADD);
```

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,314 @@
...
+            w.set_statev(0);
+        });
+
```
eva-cosma, line 190: Okay, you get the idea, comment on their own lines please

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: > There is no obvious reason why we would have top at 2_000_000_000 for example.
 
Yes, it does make sense. I left it for a more convenient testing. However, when the PR is close to its merge, I'll put another value, for instance, the maximum value.
 
> How do other embassy crates handle the creation of big structs?

`embassy-rp`:

```rust
/// The configuration of a PWM slice.
/// Note the period in clock cycles of a slice can be computed as:
/// `(top + 1) * (phase_correct ? 1 : 2) * divider`
#[non_exhaustive]
#[derive(Clone)]
pub struct Config {
    /// Inverts the PWM output signal on channel A.
    pub invert_a: bool,
    /// Inverts the PWM output signal on channel B.
    pub invert_b: bool,
    /// Enables phase-correct mode for PWM operation.
    /// In phase-correct mode, the PWM signal is generated in such a way that
    /// the pulse is always centered regardless of the duty cycle.
    /// The output frequency is halved when phase-correct mode is enabled.
    pub phase_correct: bool,
    /// Enables the PWM slice, allowing it to generate an output.
    pub enable: bool,
    /// A fractional clock divider, represented as a fixed-point number with
    /// 8 integer bits and 4 fractional bits. It allows precise control over
    /// the PWM output frequency by gating the PWM counter increment.
    /// A higher value will result in a slower output frequency.
    pub divider: fixed::FixedU16<fixed::types::extra::U4>,
    /// The output on channel A goes high when `compare_a` is higher than the
    /// counter. A compare of 0 will produce an always low output, while a
    /// compare of `top + 1` will produce an always high output.
    pub compare_a: u16,
    /// The output on channel B goes high when `compare_b` is higher than the
    /// counter. A compare of 0 will produce an always low output, while a
    /// compare of `top + 1` will produce an always high output.
    pub compare_b: u16,
    /// The point at which the counter wraps, representing the maximum possible
    /// period. The counter will either wrap to 0 or reverse depending on the
    /// setting of `phase_correct`.
    pub top: u16,
}

impl Default for Config {
    fn default() -> Self {
        Self {
            invert_a: false,
            invert_b: false,
            phase_correct: false,
            enable: true, // differs from reset value
            divider: 1.to_fixed(),
            compare_a: 0,
            compare_b: 0,
            top: 0xffff,
        }
    }
}
```
`embassy-stm32` doesn't have default configuration for both `simple_pwm` and `complementary_pwm`.

`embassy-nrf`:

```rust
/// Configuration for the PWM as a whole.
#[non_exhaustive]
pub struct Config {
    /// Selects up mode or up-and-down mode for the counter
    pub counter_mode: CounterMode,
    /// Top value to be compared against buffer values
    pub max_duty: u16,
    /// Configuration for PWM_CLK
    pub prescaler: Prescaler,
    /// How a sequence is read from RAM and is spread to the compare register
    pub sequence_load: SequenceLoad,
    /// Drive strength for the channel 0 line.
    pub ch0_drive: OutputDrive,
    /// Drive strength for the channel 1 line.
    pub ch1_drive: OutputDrive,
    /// Drive strength for the channel 2 line.
    pub ch2_drive: OutputDrive,
    /// Drive strength for the channel 3 line.
    pub ch3_drive: OutputDrive,
}

impl Default for Config {
    fn default() -> Config {
        Config {
            counter_mode: CounterMode::Up,
            max_duty: 1000,
            prescaler: Prescaler::Div16,
            sequence_load: SequenceLoad::Common,
            ch0_drive: OutputDrive::Standard,
            ch1_drive: OutputDrive::Standard,
            ch2_drive: OutputDrive::Standard,
            ch3_drive: OutputDrive::Standard,
        }
    }
}
```
 
 I would stick to the `embassy-rp` default configuration: maximum value at top, minimal settings to dividers, no inversion and phase correction.
 
> Related question: Would we need divider and prescale_factor at the same time?

Divider does the same thing as prescaler, but from the SYSCON side. The clock can be divided from 1 to 255 twice, SYSCON first, SCTimer second. It affects the number of ticks per second.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Regarding the reset, starting from the creation of the second output, if we reset the device that means that the whole device will be reset and the match that was set for the first output will disappear. The device is cleared and after adding new outputs the logic of previous output will be thrown away and forgotten. In this case it should be put in the init of the whole board as I did in the new push.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Yes, already fixed.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Fixed.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Fixed.

embassy-nxp/src/pwm/lpc55.rs
eva-cosma: I'm not a fan of arbitrary (yes, even max is arbitrary) values, but I don't have any better suggestion for this.

embassy-nxp/src/pwm/lpc55.rs
eva-cosma: :+1:  <thread closed>

(I can't actually close my own comments, but I'll write that I want them closed :laughing: )

embassy-nxp/src/pwm/lpc55.rs
eva-cosma: :+1:

embassy-nxp/src/pwm/lpc55.rs
eva-cosma: :+1:

embassy-nxp/src/pwm/lpc55.rs
eva-cosma: :+1:

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Let's wait for @i509VCB 's opinion.

i509VCB: One proposal could be a "partial default". I have an example here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=56360b2b603495a68963ef057a86f876

You use the constructor for fields that have no good defaults and can override individual fields as declared.

embassy-nxp/src/pwm/lpc55.rs
i509VCB: End of line comments can sometimes be useful, but I tend to avoid them if they go beyond 110 columns (embassy has no strict policy on this iirc).

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// Set the PWM config.
+    pub fn set_config(&mut self, config: &Config) {
+        Self::configure(self.output, config);
+    }
+
+    fn configure(output_number: usize, config: &Config) {
+        // Stop and reset the counter
+        SCT0.ctrl().modify(|w| {
+            if config.phase_correct {
+                w.set_bidir(0, crate::pac::sct0::vals::Bidir::UP_DOWN);
```
i509VCB: lot of full import paths that could be shortened. where you do vals::<Type> or directly import the type is up to you.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: I think it's a good trade-off. I've just modified the code.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Fixed.

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,329 @@
...
+//             invert: false,
+//             phase_correct: false,
+//             enable: true,
+//             divider: 255,
+//             prescale_factor: 255,
+//             compare: 1_000_000_000,
+//             top: 2_000_000_000,
+//         }
+//     }
+// }
```
i509VCB: Remove unused impl

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,329 @@
...
+        SCT0.config().modify(|w| {
+            w.set_unify(vals::Unify::UNIFIED_COUNTER);
+            w.set_clkmode(vals::Clkmode::SYSTEM_CLOCK_MODE);
+            w.set_noreload_l(true);
+            w.set_autolimit_l(true);
+        });
+
+        // Before setting the match registers, we have to make sure that `compare` is lower or equal to `top`,
+        // otherwise the counter will not reach the match and, therefore, no events will happen.
+        core::assert!(config.compare <= config.top);
```
i509VCB: You will want to remove the `core::` qualification here. We want to use https://github.com/embassy-rs/embassy/blob/main/embassy-nxp/src/fmt.rs for those macros so that these messages appear in defmt/log.

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,329 @@
...
+            SCT0.out(output_number)
+                .out_set()
+                .modify(|w| w.set_set(1 << (output_number + 1)));
+        } else {
+            // High when event 0 is active
+            SCT0.out(output_number).out_set().modify(|w| w.set_set(1 << 0));
+            // Low when event `output_number + 1` is active
+            SCT0.out(output_number)
+                .out_clr()
+                .modify(|w| w.set_clr(1 << (output_number + 1)));
```
i509VCB, line 210: This will require another nxp-pac pull request, but we could probably turn `set_clr`, `set_set` into a bit array rather than manually having to type out the bit shifts.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: FIxed

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: I've added a comment about it so that I don't forget about it.

embassy-nxp/src/pwm/lpc55.rs
frihetselsker: Fixed

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        if TOP_VALUE.load(Ordering::Relaxed) == 0 {
+            // Match 0 will reset the timer using TOP value
+            SCT0.match_(0).modify(|w| {
+                w.set_matchn_l((config.top & 0xFFFF) as u16);
+                w.set_matchn_h((config.top >> 16) as u16);
+            });
+        } else {
+            panic!("The top value cannot be changed after the initialization.");
+        }
```
i509VCB, line 158: So if you drop a pwm, and then construct it again later, would the top value cause a panic?

frihetselsker: Yes, good catch. I've just pushed a new commit with a fix.

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,325 @@
...
+        SCT0.count().read().0
+    }
+}
+
+impl<'d> Drop for Pwm<'d> {
+    fn drop(&mut self) {
+        INSTANCE_NUMBER.fetch_sub(1, Ordering::Relaxed);
+        if INSTANCE_NUMBER.load(Ordering::Relaxed) == 0 {
+            TOP_VALUE.store(0, Ordering::Relaxed);
+        }
```
i509VCB: I need to think if we should have stronger memory ordering here.

Dropping a PWM on two different threads is a possibility, in which you would want Acquire and Release semantics at minimum?

frihetselsker:
@i509VCB Fixed the issue #4771 in this push.

embassy-nxp/src/pwm/lpc55.rs
```diff
@@ -0,0 +1,325 @@
...
+use crate::gpio::AnyPin;
+use crate::pac::iocon::vals::{PioDigimode, PioFunc, PioMode, PioOd, PioSlew};
+use crate::pac::sct0::vals;
+use crate::pac::syscon::vals::{SctRst, SctclkselSel};
+use crate::pac::{SCT0, SYSCON};
+
+// Since for now the counter is shared, the TOP value has to be kept.
+static TOP_VALUE: AtomicU32 = AtomicU32::new(0);
+// To check if there are still active instances.
+static INSTANCE_NUMBER: AtomicU8 = AtomicU8::new(0);
```
i509VCB: We should call this a reference count. That makes it more clear what is intended.

frihetselsker: Fixed.

i509VCB (approved):

PR #4752, opened 2025-10-09
[embassy nrf] remove panic on uarte overrun, return error on read instead

### LINKED ISSUE #4750: [embassy-nrf] Remove "BufferedUarte overrun" panic
Please do not panic the RTOS on something like this, give developers a way of handling the error or just output to the logs, do not kill the whole system for a lost bit in the UART

https://github.com/embassy-rs/embassy/blob/35b0ba4ce0fed7588febe504e16bbf1788384f5a/embassy-nrf/src/buffered_uarte.rs#L89-L91

Kezii:
fixes #4750 

probably not the optimal solution? let me know what to do instead

### DIFF
--- a/embassy-nrf/src/buffered_uarte.rs
+++ b/embassy-nrf/src/buffered_uarte.rs
@@ -40,6 +40,7 @@ pub(crate) struct State {
     rx_started_count: AtomicU8,
     rx_ended_count: AtomicU8,
     rx_ppi_ch: AtomicU8,
+    rx_overrun: AtomicBool,
 }
 
 /// UART error.
@@ -47,7 +48,8 @@ pub(crate) struct State {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum Error {
-    // No errors for now
+    /// Buffer Overrun
+    Overrun,
 }
 
 impl State {
@@ -61,6 +63,7 @@ impl State {
             rx_started_count: AtomicU8::new(0),
             rx_ended_count: AtomicU8::new(0),
             rx_ppi_ch: AtomicU8::new(0),
+            rx_overrun: AtomicBool::new(false),
         }
     }
 }
@@ -87,8 +90,8 @@ impl<U: UarteInstance> interrupt::typelevel::Handler<U::Interrupt> for Interrupt
                 r.errorsrc().write_value(errs);
 
                 if errs.overrun() {
-                    #[cfg(feature = "defmt")]
-                    defmt::warn!("BufferedUarte overrun");
+                    s.rx_overrun.store(true, Ordering::Release);
+                    ss.rx_waker.wake();
                 }
             }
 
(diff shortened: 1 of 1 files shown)


i509VCB:
In an ideal world the reader would be told that an overrun happened, rather than an error message. `read` would return an error at the point where an overrun happened.

Edit: An example of this is what is done in the  mspm0 HAL: https://github.com/embassy-rs/embassy/blob/35b0ba4ce0fed7588febe504e16bbf1788384f5a/embassy-mspm0/src/uart/buffered.rs#L950-L965

Kezii:
Hello,
Thank you for your comment, I tested this patch, based on your suggestion, to work correctly. this is ready for review

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -762,6 +767,11 @@ impl<'d> BufferedUarteRx<'d> {
             compiler_fence(Ordering::SeqCst);
             //trace!("poll_read");
 
+            if s.rx_overrun.load(Ordering::Relaxed) {
+                s.rx_overrun.store(false, Ordering::Relaxed);
```
lulf: I think you can convert these load + store into a swap:

```
if s.rx_overrun.swap(false, ...) {
}
```

https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU32.html#method.swap

Kezii: thank you, done

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -87,7 +90,8 @@ impl<U: UarteInstance> interrupt::typelevel::Handler<U::Interrupt> for Interrupt
                 r.errorsrc().write_value(errs);
 
                 if errs.overrun() {
-                    panic!("BufferedUarte overrun");
+                    s.rx_overrun.store(true, Ordering::Relaxed);
```
lulf: Should this be Ordering::Acquire since the write is in an interrupt and the read is not?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -762,6 +767,10 @@ impl<'d> BufferedUarteRx<'d> {
             compiler_fence(Ordering::SeqCst);
             //trace!("poll_read");
 
+            if s.rx_overrun.swap(false, Ordering::Relaxed) {
```
lulf: Should this be Ordering::AcqRel?

embassy-nrf/src/buffered_uarte.rs
```diff
@@ -820,6 +829,9 @@ impl<'d> BufferedUarteRx<'d> {
     /// we are ready to read if there is data in the buffer
     fn read_ready(&self) -> Result<bool, Error> {
         let state = self.buffered_state;
+        if state.rx_overrun.swap(false, Ordering::Relaxed) {
```
lulf: Also Ordering::AcqRel?

embassy-nrf/src/buffered_uarte.rs
Kezii: apparently that breaks the build



error: atomic stores cannot have `Acquire` or `AcqRel` ordering
  --> src/buffered_uarte.rs:93:46
   |
93 |                     s.rx_overrun.store(true, Ordering::Acquire);
   |                                              ^^^^^^^^^^^^^^^^^
   |
   = help: consider using ordering modes `Release`, `SeqCst` or `Relaxed`
   = note: `#[deny(invalid_atomic_ordering)]` on by default

Kezii: should I use Release for the store and AcqRel for the swap?

lulf: Ah, yes, my bad. This one should be Release.

Kezii:
I used Release for the interrupt handler, and Acquire for the user code

In my understanding, there's no need for AcqRel since we are not polling on the false value ?

lulf:
> I used Release for the interrupt handler, and Acquire for the user code
> 
> In my understanding, there's no need for AcqRel since we are not polling on the false value ?

Hmm, yeah. You're right

PR #4758, opened 2025-10-14
mspm0/gpio: use maitake-sync for gpio wakers

### LINKED ISSUE #4015: mspm0/gpio: reduce memory used by wakers
For some of the MSPM0 parts, especially those with little RAM, the GPIO wakers take up a considerable amount of RAM. The numbers are based off the current size of `AtomicWaker` being 8 bytes at the time of writing.

For example with C110x, objdump is telling me that ***25%*** of RAM (C110x has 1KB of RAM) is used just for GPIO wakers:

```
i509vcb@verdigris:~/Dev/embassy/examples/mspm0c1104$ objdump -C -t ./target/thumbv6m-none-eabi/release/button | grep '.bss'
200000d8 l     O .bss   00000100 embassy_mspm0::gpio::PORTA_WAKERS
20000050 l     O .bss   00000004 button::__embassy_main::POOL
20000054 l     O .bss   00000084 embassy_executor::_export::ARENA
200001d9 l     O .bss   00000001 defmt_rtt::TAKEN
200001da l     O .bss   00000001 defmt_rtt::CS_RESTORE
200001db l     O .bss   00000001 defmt_rtt::ENCODER
200001dc l     O .bss   00000001 defmt_rtt::ENCODER
200001dd l     O .bss   00000001 defmt_rtt::ENCODER
200001d8 g     O .bss   00000001 _EMBASSY_DEVICE_PERIPHERALS
20000050 g       .bss   00000000 __sbss
200001e0 g       .bss   00000000 __ebss
```

In addition, C1104 will only ever have up to 18 GPIO pins. This means that 14 slots, or 112 bytes of RAM used for GPIO wakers is just wasted space. For L1306 the GPIO wakers still take up 6.25% of RAM. The waste is less on L1306, but its still 32 bytes of RAM that is wasted by the 4 unused slots.

For both of these parts, the numbers come from the `button` example which only ever needs a single waker. This means there is 248 bytes of memory wasted in these examples.

What would be desirable is a way to only allocate the waker in RAM, if and only if the pin is actually used as an async input. Somehow we would need to know what pins are available and then only actually use the RAM for those wakers.

I was told about https://github.com/embassy-rs/embassy/pull/1016 but that approach falls apart when applied to `AnyPin`.

Enabling turbo wakers would halve the memory usage, but unused pins still waste space.

i509VCB:
Fixes https://github.com/embassy-rs/embassy/issues/4015
~~https://github.com/hawkw/mycelium/pull/555 must be merged first.~~

I have verified this new setup works correctly on C1104. I will test on G3507 and maybe L2228 tomorrow. Keeping this a draft till the dependent pull request is merged.

I also consider this pull request to be a testbed for whether other embassy-hals may benefit from using maitake-sync, specifically `WakeMap`.

As discussed in https://github.com/embassy-rs/embassy/issues/4015, maitake-sync allows the cost of GPIO wakers to be paid by usage rather than statically, even if most pins are unused. Currently this disables the old method of statically allocated wakers. I would like some discussion whether this would be a good idea to mark as a choice (although I want to spin up an example using many pins to test this).

Here I have some numbers for RAM and flash usage on MSPM0C1104, for which statically allocated wakers waste 25% of RAM. For C1104, the blinky example is freed up nearly 25% of RAM. Unfortunately this increases the `.text` size. I suspect some optimization of what is inlined may help here.

Before waitmap (blinky):
```
   3944      80     408    4432    1150 blinky
```

i509vcb@verdigris:~/Dev/embassy/examples/mspm0c1104$ cargo objdump --release --bin blinky -- -C -t | grep '.bss'
    Finished `release` profile [optimized + debuginfo] target(s) in 0.05s
```
20000050 l     O .bss   00000040 blinky::__embassy_main::POOL::hdf4890576116d2f8
20000098 l     O .bss   00000100 embassy_mspm0::gpio::PORTA_WAKERS::h3e46d7edcc235649
20000090 l     O .bss   00000008 embassy_mspm0::dma::STATE::hc0d1f2780f4fe38c
20000199 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.0)
2000019a l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.1)
2000019b l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.2)
2000019c l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.3)
2000019d l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.4)
20000198 g     O .bss   00000001 _EMBASSY_DEVICE_PERIPHERALS
20000050 g       .bss   00000000 __sbss
200001a0 g       .bss   00000000 __ebss
```

After waitmap (blinky):
```
   4204      80     168    4452    1164 blinky
```

i509vcb@verdigris:~/Dev/embassy/examples/mspm0c1104$ cargo objdump --release --bin blinky -- -C -t | grep '.bss'
    Finished `release` profile [optimized + debuginfo] target(s) in 0.05s
```
20000050 l     O .bss   00000040 blinky::__embassy_main::POOL::h1d50e2a13d16c60d
20000094 l     O .bss   00000014 embassy_mspm0::gpio::GPIO_WAIT_MAP::h2c627e5d7bbe0548
200000a8 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.0)
200000a9 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.1)
200000aa l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.2)
200000ab l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.3)
200000ac l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.4)
20000090 g     O .bss   00000001 _EMBASSY_DEVICE_PERIPHERALS
20000050 g       .bss   00000000 __sbss
200000b0 g       .bss   00000000 __ebss
```

Before waitmap (button):
```
   4100      80     408    4588    11ec button
```

i509vcb@verdigris:~/Dev/embassy/examples/mspm0c1104$ cargo objdump --release --bin button -- -C -t | grep '.bss'
    Finished `release` profile [optimized + debuginfo] target(s) in 0.05s
```
20000098 l     O .bss   00000100 embassy_mspm0::gpio::PORTA_WAKERS::h3e46d7edcc235649
20000050 l     O .bss   00000040 button::__embassy_main::POOL::ha91ef408a93b6f8f
20000090 l     O .bss   00000008 embassy_mspm0::dma::STATE::hc0d1f2780f4fe38c
20000199 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.0)
2000019a l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.1)
2000019b l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.2)
2000019c l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.3)
2000019d l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h5651992cc6ed4618 (.4)
20000198 g     O .bss   00000001 _EMBASSY_DEVICE_PERIPHERALS
20000050 g       .bss   00000000 __sbss
200001a0 g       .bss   00000000 __ebss
```

After waitmap (button):
```
   5172      80     228    5480    1568 button
```

i509vcb@verdigris:~/Dev/embassy/examples/mspm0c1104$ cargo objdump --release --bin button -- -C -t | grep '.bss'
    Finished `release` profile [optimized + debuginfo] target(s) in 0.05s
```
20000050 l     O .bss   00000080 button::__embassy_main::POOL::h992edb0ed5a67f9b
200000d0 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.0)
200000d1 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.1)
200000d2 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.2)
200000d3 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.3)
200000d4 l     O .bss   00000001 defmt_rtt::RTT_ENCODER::h6f41292e6a6b2194 (.4)
200000d8 l     O .bss   00000014 embassy_mspm0::gpio::GPIO_WAIT_MAP::h2c627e5d7bbe0548
20000050 g       .bss   00000000 __sbss
200000ec g       .bss   00000000 __ebss
200000d5 g     O .bss   00000001 _EMBASSY_DEVICE_PERIPHERALS
```

@jamesmunns would appreciate review from you

### DIFF
--- a/embassy-mspm0/CHANGELOG.md
+++ b/embassy-mspm0/CHANGELOG.md
@@ -24,3 +24,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat: add CPU accelerated division function (#4966)
 - feat: Add trng implementation (#5172)
 - fix: feature guard pins used for NRST and SWD (#5257)
+- feat: Move from GPIO waker arrays to maitake-sync wait map
--- a/embassy-mspm0/Cargo.toml
+++ b/embassy-mspm0/Cargo.toml
@@ -72,6 +72,11 @@ micromath = "2.0.0"
 # mspm0-metapac = { version = "" }
 mspm0-metapac = { git = "https://github.com/mspm0-rs/mspm0-data-generated/", tag = "mspm0-data-d9f54366c65cec47957065f42fe04542b852a8cd" }
 rand_core = "0.9"
+# Force no cache padding or the types are too large.
+maitake-sync = { version = "0.3.0", default-features = false, features = ["critical-section", "no-cache-pad"]}
+# maitake sync requires portable atomic. For Cortex-M0 it is fine to force a dependency on the critical section feature.
+# however on MSPM33 parts it may be desired to use hardware atomics instead.
+portable-atomic = { version = "1.13.1", features = ["critical-section"] }
 
 [build-dependencies]
 proc-macro2 = "1.0.94"
--- a/embassy-mspm0/src/gpio.rs
+++ b/embassy-mspm0/src/gpio.rs
@@ -2,11 +2,9 @@
 
 use core::convert::Infallible;
 use core::future::Future;
-use core::pin::Pin as FuturePin;
-use core::task::{Context, Poll};
 
 use embassy_hal_internal::{Peri, PeripheralType, impl_peripheral};
-use embassy_sync::waitqueue::AtomicWaker;
+use maitake_sync::WaitMap;
 
 use crate::pac::gpio::vals::*;
 use crate::pac::gpio::{self};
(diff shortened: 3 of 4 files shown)


Dirbaio:
stm32 has kind of the same problem. My idea to fix it was to make EXTI also use `bind_interrupts!` so the user chooses which pins they want to use with async gpio. Each pin would get a static "state" with the waker similar to the other drivers, so the states for unused pins would get completely optimized out. 

This has other advantages, for example the HAL doesn't force registering all interrupts so you can use one pin with async and another pin with manual interrupts, which is a recurring ask from users.

Have you considered whether this would work for mpsm0 as well?

i509VCB:
> stm32 has kind of the same problem. My idea to fix it was to make EXTI also use `bind_interrupts!` so the user chooses which pins they want to use with async gpio. Each pin would get a static "state" with the waker similar to the other drivers, so the states for unused pins would get completely optimized out.
> 
> This has other advantages, for example the HAL doesn't force registering all interrupts so you can use one pin with async and another pin with manual interrupts, which is a recurring ask from users.
> 
> Have you considered whether this would work for mpsm0 as well?

Yes longer term it should probably be possible to let the gpio interrupts get bound individually so that the user can pick a handler depending on the pin. Where this is kind of complicated for TI is that some chips use an interrupt group for GPIO. This might mean the mspm0 hal would need some hacky bits to handle that. 

At least so far it is consistent whether a chip has dedicated interrupts or uses an interrupt group (no mix of both).

Even with your suggestion, I still see good reason to consider using maitake-sync: I consider letting the handler being configurable and how wakers are implemented internally to be two different problems. An example is waiting for a pair of pins, one at a time in a loop. With maitake-sync you only need space for one waiter in the future state machine. Statically allocated wakers would appear more consistent, but only using 1 waker at a time means 8 bytes get wasted, as both pin wakers are disjoint in usage.

i509VCB:
I have updated this and it no longer relies on the git dependency for maitake-sync. While this isn't completely optimal per some of Dirbaio's comments, it is a good improvement to the current RAM usage.

Since embassy-mcxa does something similar it is good to merge in my opinion.

PR #4761, opened 2025-10-15
Implement the dual external flash feature for stm32 qspi

SimonMNielsen:
Hi All!

I am trying to implement the stm32 dual flash feature for use with the qspi module. I got some pointers from @Dirbaio on Matrix for what was initially needed, essentially some new constructors for the memory banks. It has resulted in this branch. I am having some problems getting it to work though. As soon as I set the dual flash flag to true(this flag was already implemented), the quad spi keeps getting stuck at busy after the first qspi command: 
image

Im not sure if its the qspi setup or how the external memflash is implemented. My test code is based on the embassy example here:
https://github.com/embassy-rs/embassy/blob/main/examples/stm32h742/src/bin/qspi.rs

I came across the following in the stm32 reference manual regarding reading and writing in dual flash mode: 
image

I am not that well versed on the embassy qspi code. Will the above require a change to how the transactions are setup before the read/write commands are executed?

(PS: if I set the dual flash flag to false, I can read and write to the (first) external flash just fine, so Im pretty sure the flash specific commands are all set correct.)

Anyone got some input?

### DIFF
--- a/embassy-stm32/src/qspi/mod.rs
+++ b/embassy-stm32/src/qspi/mod.rs
@@ -69,9 +69,8 @@ pub struct Config {
     pub sample_shifting: SampleShifting,
     /// GPIO Speed
     pub gpio_speed: Speed,
-    /// Dual flash mode 
+    /// Dual flash mode
     pub dual_flash: bool,
-
 }
 
 impl Default for Config {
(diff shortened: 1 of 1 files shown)


SimonMNielsen:
I finally had some time to look into this again and it was as expected: status registers etc. were not being read/written twice. It tricked me a bit because in my App I was memory mapping the external flash chips almost at the beginning, but just prior of doing this I had some code that did some reading of chip id etc. And this is what was failing.

I got it working now and and im even using it to run my entire App from the 2 external flashchips in dual-mode :)

SimonMNielsen:
rebased this and fixed the formatting, anything more needed from my end in order to get this in?

PR #4762, opened 2025-10-15
embassy-nrf: Make PWM drivers more flexible

de-vri-es:
This PR:

* Derives more traits for public structs in the PWM module.
* Allows you to set the idle state of pins for when the PWM is stopped.

  From the nRF9160 manual:
  image
* Allows you to configure the pins in the constructor of SimplePwm to avoid glitches on the output levels during initialization (in fact, you can now configure the whole PWM peripherals, same as with the `Sequencer` struct).
* Exposes bit 16 of the duty cycle value in `SimplePwm` in a `DutyCycle` struct so you can use inverted polarity.
* Adds a function to set all channels of a `SimplePwm` at once, to avoid doing four DMA transfers when one would have been enough.

Tested with an nRF9160.

### DIFF
--- a/embassy-nrf/CHANGELOG.md
+++ b/embassy-nrf/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added: Add basic RTC support for nRF54L
 - changed: apply trimming values from FICR.TRIMCNF on nrf53/54l
 - changed: do not panic on BufferedUarte overrun
+- added: allow direct access to the input pin of `gpiote::InputChannel`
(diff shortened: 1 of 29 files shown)


de-vri-es:
Is this a good way forward? It's a breaking change as it is now. I think it's nicer this way since it will be a more consistent API. But if desired I can avoid breaking the API by adding new functions instead of modifying existing ones.

lulf (approved):
I like the API, nice work! :rocket:  It'll make my nrf54 PR harder to merge :smiling_face_with_tear: but i'll rebase once this is merge :)

de-vri-es:
Haha, thanks :)

PR #4765, opened 2025-10-16
add STSPHSRX & STSPHST handler

blueskycoco:
in case we get stsphst int with OUT_DATA_DONE, need clear doepint stsphsrx bit, else will block subsequent setup packet receive.

[TRACE] irq (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:29) 
[TRACE] === status 01040400 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:45) 
[TRACE] OUT_DATA_RX ep=0 len=64 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:71) 
[TRACE] === status 01060000 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:45) 
[TRACE] OUT_DATA_DONE ep=0 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:99) 
[TRACE] === status 09060000 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:45) 
[TRACE] OUT_DATA_DONE ep=0 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:99) 
[TRACE] control: data_out (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:1313) 
[TRACE] read start len=64 (embassy_usb_synopsys_otg embassy-usb-synopsys-otg/src/lib.rs:1083)

### DIFF
--- a/embassy-usb-synopsys-otg/CHANGELOG.md
+++ b/embassy-usb-synopsys-otg/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+## 0.3.2 - 2025-10-16
+
+- add STSPHSRX & STSPHST handler
+
 ## 0.3.1 - 2025-08-26
 
 - Improve receive performance, more efficient copy from FIFO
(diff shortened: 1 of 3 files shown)


bugadani:
I wouldn't mind merging this, but esp-idf documents this bit as "applicable only in Case of Scatter Gather DMA mode." 🤔

Dirbaio:
i'm going to close this since AFAICT that bit does nothing for our use case (no DMA), so I don't see how can this make a difference.

If you indeed are seeing failures without the fix that get fixed by this please tell us which chip it is, which core version number, and logs before+after the fix.

PR #4770, opened 2025-10-19
nxp/lpc55: move usart ALT pin definitions to impl_xx_pin macros

i509VCB:
This should make a future move to the buildscript generating pin impls easier.

Also split impl_pin into specific TX and RX macros to make it clearer to read.

@eva-cosma @frihetselsker please verify I did not accidentally change mappings. I did verify both usart examples build.

Longer term some clean up could be nice since passing around `(Peri<'d, AnyPin>, PioFunc)` can get a little messy.

### DIFF
--- a/embassy-nxp/CHANGELOG.md
+++ b/embassy-nxp/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- next-header -->
 ## Unreleased - ReleaseDate
+- LPC55: Move ALT definitions for USART to TX/RX pin impls. 
 - LPC55: Remove internal match_iocon macro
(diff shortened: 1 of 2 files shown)


frihetselsker (approved):
It's absolutely perfect from my point of view. `PioFunc` is inside the pin trait as it was desired.

eva-cosma:
`new_inner` is getting quite beefy, but this seems like it makes the compiler happy, so I am happy. Maybe some docs on `new_inner` wouldn't hurt, but I'll leave that up to you

eva-cosma (approved):

i509VCB:
> `new_inner` is getting quite beefy, but this seems like it makes the compiler happy, so I am happy. Maybe some docs on `new_inner` wouldn't hurt, but I'll leave that up to you

Longer term I may want to look at making new_inner less messy. Although it would probably look closer in style to the mspm0 uart.

PR #4778, opened 2025-10-22
allow setting stm32 SAI frame_length to 256

everdrone:
[This line](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/sai/mod.rs#L699) prevents users from setting the frame_length parameter to 256.

According to RM0433 rev.8 51.4.6, the frame length is set with FRL+1.
If embassy expects an u8 and subtracts 1 in that line, the maximum value of FRL is 254, resulting in a frame length of 255 which is illegal in some cases (when the MCLK is enabled.

The cleanest solution would be to use `u16` to retain compatibility.


embassy-stm32/src/sai/mod.rs
```diff
@@ -696,7 +696,12 @@ impl<'d, T: Instance, W: word::Word> Sai<'d, T, W> {
             w.set_fspol(config.frame_sync_polarity.fspol());
             w.set_fsdef(config.frame_sync_definition.fsdef());
             w.set_fsall(config.frame_sync_active_level_length.0 as u8 - 1);
-            w.set_frl(config.frame_length - 1);
+
+            if config.frame_length > 256 {
+                panic!("Frame length cannot be greater than 256");
+            }
+
+            w.set_frl((config.frame_length - 1) as u8);
```
xoviat: What about `(config.frame_length - 1).try_into().unwrap()`?

PR #4781, opened 2025-10-24
add a sample for the lcd of the stm32h745i-disco

0x53A:
I have that exact code running on a board in front of me:
Untitled

As mentioned in the README of the new folder, it was adapted from the STM32H735 sample.


0x53A:
(I rebased on main but do not currently have access to the board; i can test it in a few days)

PR #1276, opened 2023-03-14
Remove initialiser generic from saadc

huntc:
The initialiser generic is not required when using the `run_timer_sampler` method, and there is no example. This commit solves both.

PR #4782, opened 2025-10-24
Remove unrequired generic from run_time_sampler

Hellymaw:
Generic `I` isn't required by the method. Same change as proposed in #1276, just without the added example


Dirbaio:
merged in #6620

PR #4784, opened 2025-10-24
RP2040/235x: Push DMA data to PIO TX FIFO through ping-pong

### LINKED ISSUE #4190: [rp2040] DMA: Enhance PIO RX DMA support for continuous hardware-chained ping-pong buffering via CHAIN_TO
Current high-level transfer API is not designed to support ping-pong buffering using CHAIN_TO.

It is crucial when dealing for example with camera data, where it is needed to fill a DMA buffer with a few lines and processing it very quickly while the second buffer is filled meanwhile.

I've tried double buffer approach without chaining:

```rust
let transfer_a = rx_fifo.dma_pull(dma_channel.reborrow(), buffer_a, false);
transfer_a.await;
  
loop {        
      {
          let transfer_b = rx_fifo.dma_pull(dma_channel.reborrow(), buffer_b, false);
          process_data(buffer_a);
          transfer_b.await;
      }
      {
          let transfer_a = rx_fifo.dma_pull(dma_channel.reborrow(), buffer_a, false);
          process_data(buffer_b);
          transfer_a.await;
      }
}
```
However it is not reliable and I observe missing data quite quickly (it probably depends on other running tasks).

holly-hacker:
This PR adds a function on `embassy_rp::pio::StateMachineTx` to allow pushing continuous data to a PIO state machine through a ping-pong/double buffering mechanism, allowing one buffer to be filled while the other is being sent. This method uses 2 DMA channels with `CHAIN_TO` allowing no downtime between transfers (assuming the user fills their buffers in time using the given callback).

Example usage to generate a 440Hz sine wave over 8 digital pins at a desired sample rate:
```rs
const SAMPLE_RATE: u32 = 16_000; // 16kHz

let tx = pio.sm0.tx();

let mut data_1 = [0x0u8; 128];
let mut data_2 = [0x0u8; 128];

let mut sample_index = 0usize;
tx.dma_push_ping_pong(
    p.DMA_CH0.reborrow(),
    p.DMA_CH1.reborrow(),
    &mut data_1,
    &mut data_2,
    |buf| {
        info!("In start of fill callback, index={}", sample_index);
        if sample_index > 100_000 {
            return core::ops::ControlFlow::Break(());
        }

        for b in buf.iter_mut() {
            let time = sample_index as f32 / SAMPLE_RATE as f32;
            let wave = fast_sin(time * 440. * f32::consts::PI * 2.);
            *b = ((wave + 1.) / 2. * 256.) as u8;

            sample_index += 1;
        }

        core::ops::ControlFlow::Continue(())
    },
)
.await;

// push a zero to reset the pin state
tx.dma_push(p.DMA_CH0, &[0u8; 1], false).await;
```

At 16kHz, this results in a buffer of ~8ms which looks clean (or as clean as you can get with a messy 8bit R-2R "DAC" :smile:) on the oscilloscope:
image

---

Some open questions:
- I haven't added any compiler fences which are used in noticed in `dma_push`. I'm not knowledgeable enough about compiler internal to know how/when the compiler will re-order instructions to know where to insert them. Feedback is appreciated, but the current implementation seems to work
- I haven't looked at task priority yet so in a program with many concurrent tasks, another task may take the cpu time needed to fill the buffer. I'm not sure if this is something that should be handled on embassy's side.
- Users can't send partial buffers which may be desired when ending a transfer. I could fix this by changing `ControlFlow` to `ControlFlow::<(), Option<u32>>` to let the user specify how much of their buffer they want to send, but that makes the API a bit ugly. Not sure what to do here, open to feedback.
- I haven't touched any code in `embassy_rp::dma`, which seems to contain some duplicate code to `embassy_rp::dma::StateMachineTx`. I'm not sure how the maintainers want to approach this.

I haven't added any examples yet. If desired, I can do that when the general design of this function is approved.

I'm pretty new to embedded dev in general, so I may have missed something obvious. Thorough code review would be appreciated.

Fixes #4190


felipebalbi:
Maybe add your example above as an actual example in the repository (i.e. as part of examples/rp235x).

holly-hacker:
I've added the rp235x example and tested it locally to ensure it works, but I haven't tested with the rp2040 yet because I'd prefer to get some feedback on this PR first.

embassy-rp/src/pio/mod.rs
```diff
@@ -530,6 +531,95 @@ impl<'d, PIO: Instance, const SM: usize> StateMachineTx<'d, PIO, SM> {
...
+            trace!("Waiting for ping transfer to finish");
+            Transfer::new(ch1.reborrow()).await;
+
+            // re-init DMA 1 (without triggering it)
+            ch1.regs().read_addr().write_value(data1.as_ptr() as u32);
+
+            trace!("Fill ping buffer");
+            if let ControlFlow::Break(()) = fill_buffer_callback(data1) {
+                break;
+            }
```
enbyted, line 596: I'm not sure this is sound. If the callback is too slow and the `ch1` is triggered while it's still executing we're holding onto a mutable borrow while another immutable access (by the DMA) is happening.

Perhaps the channel should be disabled for the duration of this callback? Or the function should be unsafe?

If you went with marking the function as unsafe, I'm not sure how could you even guarantee soundness of this without analyzing the whole system, as other executing tasks could block the CPU for long enough to make the callback unable to meet its deadline.

embassy-rp/src/pio/mod.rs
```diff
@@ -530,6 +531,95 @@ impl<'d, PIO: Instance, const SM: usize> StateMachineTx<'d, PIO, SM> {
...
+
+            trace!("Waiting for pong transfer");
+            Transfer::new(ch2.reborrow()).await;
```
enbyted, line 599: How does this behave in case the callback runs for too long? Can it potentially hand forever?
Can it immediately fall-through even though ch2 is currently running (because we missed it finishing)?

embassy-rp/src/pio/mod.rs
```diff
@@ -530,6 +531,95 @@ impl<'d, PIO: Instance, const SM: usize> StateMachineTx<'d, PIO, SM> {
...
+
+            // re-init DMA 1 (without triggering it)
+            ch1.regs().read_addr().write_value(data1.as_ptr() as u32);
+
+            trace!("Fill ping buffer");
+            if let ControlFlow::Break(()) = fill_buffer_callback(data1) {
+                break;
+            }
+
+            trace!("Waiting for pong transfer");
```
enbyted, line 598: Technically you should have a fence here, before the other transfer is started. Though I'm quite sure that an await point makes for a fence anyway, so perhaps it's not an issue.

If you disabled the chanel for the duration of the callback then you do need a fence before re-enabling the channel. Otherwise the compiler could reorder some writes from inside the callback to after the re-enable.

embassy-rp/src/pio/mod.rs
holly-hacker: We could only set the `chain_to` register when the callback finishes and manually trigger the next DMA transfer if we notice the callback finished too late, but I'm not sure how to implement this in a way that is free of race conditions. Open to feedback here.

dobrowolski-lukasz:
Project I'm working on would benefit from this PR. It uses PIO to output audio and setting up "normal" DMA results in pauses in output. With my use case in mind I have two concerns:

1. It would be great to have a way to tell when we didn't make it on time to fill in the second buffer. A callback maybe? I'd use it to save a timestamp to measure how much late the buffer filling is and print it.
2. I'm unsure how to get data to fill the buffer cleanly with this implementation. In my app I await on embassy-sync::zerocopy_channel  to get data form another task. The callback is non-async.

I'm not proficient in Rust yet, so It could just be me not seeing how to do those things.

Do let me know how I could help this PR move forward.

holly-hacker:
I'm open to feedback and ideas on this PR. I haven't been able to put this code into my "production" project yet so I haven't been able to explore it too much, and I don't quite know what the best API design here is.

mvirkkunen:
Hi!

I was actually working on a similar update a while back for the exact same purpose of glitchless audio output but never got around to finishing it. I'm fairly sure in the process I determined that it's quite difficult to implement this soundly in the way the current implementation is, that is, having two channels that write to the peripheral which are `chain_to`'d to each other. This is the way I tried first! I believe the problem was that either the transfer initialization is always racy because you might end up writing into a buffer the DMA is using, or alternatively you can get in a situation where the channels will keep ping-ponging forever if the code running on the CPU crashes for instance.

What I eventually arrived at was using one DMA channel as a control channel that writes values to another DMA channel's control registers. This is the reason there are multiple permutations of the DMA control registers in different orders, so that you can set up another channel that writes to them and the last write can be the trigger (you'll notice I finally PR'd the change that fixes the register types in `rp-pac` but that update is going to be hairy due to how long ago it was last updated) This enabled me to handle underflow gracefully without race conditions or fast IRQs.

I had a preliminary working implementation of this but then I started thinking about making a nicer API that doesn't require a separate method on every peripheral for normal vs. ping-pong TX and eventually got bored of it.

dobrowolski-lukasz:
> I had a preliminary working implementation of this but then I started thinking about making a nicer API that doesn't require a separate method on every peripheral for normal vs. ping-pong TX and eventually got bored of it.

I'd appreciate it if you could share that code even if it's work in progress.

dobrowolski-lukasz:
@mvirkkunen  Sorry to nag You but would you be willing to publish your incomplete work? I'd rather try to improve existing code than start from scratch. And repeat work that has already been done.

mvirkkunen:
@dobrowolski-lukasz My memory must be failing me, after taking a closer look I don't have an implementation of this after all.

There's a C example of the "control channel" idea in general here: https://github.com/raspberrypi/pico-examples/blob/master/dma/control_blocks/control_blocks.c - however this doesn't do ping-pong DMA but more like gather DMA from multiple buffers set up beforehand. It is possible to modify this so that there is only one control block and it is modified on the fly, but getting the write timing to be exactly correct (to avoid writing into things that might be in use) can possibly be a bit tricky.

enbyted:
A different approach - one that I have implemented for a different system - is to have the ping-pong action with 2 channels, something like @mvirkkunen has described. You have one channel that sends data and another that writes the source address trigger register od the first channel with whatever the current „busy” buffer is.

it operates under different constraints - it ensures constant data flow with no glitches, however you will get repeated data putput if you don’t make the deadline. For my system it is exactly what is needed sicne I’m using it for a video framebuffer.
It is relatively easy to make it stop sending instead, though.

It all depends on how you want to handle software that doesn’t meet the deadline.

At any rate, if you’re interested I can prepare that code to be shared somewhere. It requires pinning and is not made for embassy though.

dobrowolski-lukasz:
> It all depends on how you want to handle software that doesn’t meet the deadline.

I don't really care about what strategy is used so I can keep data flowing without interruptions (assuming source prepares that data on time).
Ideally DMA would underflow on missed deadline. I've modified PioI2sOutProgram to insert empty samples when it's starved anyway.

What I need is ability to measure and log missed deadlines.

> At any rate, if you’re interested I can prepare that code to be shared somewhere. It requires pinning and is not made for embassy though.

I'd be grateful! Adapting it to embassy should be faster and easier than starting from scratch.

enbyted:
@dobrowolski-lukasz sorry it took a couple days, I had to jump through a couple hoops to release this under MIT/apache dual license rather than GPL 2.

Here you go: https://gist.github.com/enbyted/7ae1bfc317607ac8c5f8d9c1ebc3811b

enbyted:
> Ideally DMA would underflow on missed deadline.

The solution that I posted should be adaptable to that constraint. If you don't chain the data channel to the repeat channel by default then the transfers will stop. Then in the swap routine you can set the chaining (and detect if deadline was missed).

It will be tricky with timing, though if you assume that the recipient peripheral is much slower than the core, you should be able to pause the transfer (set EN to 0), verify the state, and resume it afterwards.

This should also be adaptable to making the swap an asynchronous operation - you just need to wait for the repeat channel to finish via interrupt instead of synchronously waiting.

holly-hacker:
Small update: the project I needed this functionality for has fallen through so if anyone wants to take over, let me know and I'll close this PR in favor of yours. Otherwise, I'll leave it open for now but may close it in the future and delete my fork, as I don't have the time right now to find a better approach for this PR.

#5122 may also be of interest to people that need this functionality.

Boring legal disclaimer: I dedicate the changes in this PR to the public domain, so anyone may feel free to reuse it in any form without attribution or otherwise having to follow the MIT or CC-BY-SA license. This only applies to my changes and not the Embassy project as a whole.

dobrowolski-lukasz:
I'm giving up on this.

I've implemented restarting DMA channel in IRQ handler and it's enough to prevent any underflows in my use case so I don't have a need for complex multi-channel setup. I've opened some PRs with changes to the project as a result.

Thanks to holly-hacker, mvirkkunen  and enbyted! Your code and comments helped me a lot with getting the thing to work properly.

leftger:
bender run

PR #4786, opened 2025-10-25
Rework regular conversion to allow use with external trigger and avoid reconfiguration for each sample

Jacke-debug:
This PR is based on the changes in https://github.com/embassy-rs/embassy/pull/4243 which introduces configuration of injected ADC measurements. This PR expands the ADC module for STM32 g4 processors to support both injected and regular ADC conversions.

* Add methods for configuring ADC Regular conversions with external trigger
* Add example to showcase how both regular and injected ADC conversions can be used simultaneously
* Left the old ADC `read()` method as it was for backwards compatibility

### DIFF
--- a/examples/stm32g4/src/bin/adc_injected_and_regular.rs
+++ b/examples/stm32g4/src/bin/adc_injected_and_regular.rs
@@ -114,6 +114,9 @@ async fn main(_spawner: embassy_executor::Spawner) {
     adc1.start_regular_conversion();
     adc1.start_injected_conversion();
 
+    // Enable interrupt at end of injected ADC conversion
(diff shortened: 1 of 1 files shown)


xoviat:
The first question I have is how the regular conversion differs from the readable ringbuffer that I have implemented. If injected conversions were not a factor, would that be sufficient?

Jacke-debug:
> The first question I have is how the regular conversion differs from the readable ringbuffer that I have implemented. If injected conversions were not a factor, would that be sufficient?

Yes, thanks for highlighting. It seems to cover my needs for the regular channels. I will scrap this PR and open a new one based on the RingBuffer module instead: https://github.com/embassy-rs/embassy/pull/4840

Issue #4785, opened 2025-10-24
"attempt to subtract with overflow" sometimes on GPIO interrupt


doxxx:
On an MSPM0L1306, when toggling a GPIO for which I have a task using `wait_for_any_edge()`, I sometimes get a panic `attempt to subtract with overflow` at:

https://github.com/embassy-rs/embassy/blob/4bff7cea1a26267ec3671250e954d9d4242fabde/embassy-mspm0/build.rs#L198

Let me know if you need example code.

i509VCB:
That is probably a bug. We should probably check that value is not 0 before subtracting to figure out which group interrupt was triggered.

I am certainly curious why we are getting spurious interrupts there.

doxxx:
It was interrupt group 1, which I believe is only for GPIO and the comparator. In my firmware I wasn't using the comparator. However, the GPIO pin which was triggering the interrupt is connected to a push button on a breadboard with bad contacts. Pushing the button would sometimes toggle the GPIO rapidly a few times.

PR #4794, opened 2025-10-27
mspm0: group irq handlers must check for NO_INTR (#4785)

doxxx:
In the case of spurious interrupts, the interrupt group's STAT register may be set to NO_INTR, which must be checked before attempting to calculate the interrupt index from the STAT value.

Fixes: #4785

### DIFF
--- a/embassy-mspm0/build.rs
+++ b/embassy-mspm0/build.rs
@@ -201,7 +201,7 @@ fn generate_groups() -> TokenStream {
                     return;
                 }
 
-                // MUST subtract by 1 because NO_INTR offsets IIDX values. 
+                // MUST subtract by 1 because NO_INTR offsets IIDX values.
                 let iidx = stat.to_bits() - 1;
 
                 let Ok(group) = #group_enum::try_from(iidx as u8) else {


doxxx:
Not yet tested on actual hardware where the problem was originally occurring.

doxxx:
I've tested this with the hardware and code where I originally found the problem, and I am no longer able to trigger the panic.

embassy-mspm0/build.rs
```diff
@@ -194,8 +194,15 @@ fn generate_groups() -> TokenStream {
...
-                // MUST subtract by 1 since 0 is NO_INTR
-                let iidx = group.iidx().read().stat().to_bits() - 1;
+                let stat = group.iidx().read().stat();
+
+                // check for spurious interrupts
+                if stat == crate::pac::cpuss::vals::Iidx::NO_INTR {
+                    return;
+                }
+
+                // MUST subtract by 1 because Iidx::INT0=1
```
i509VCB:
```suggestion
                // MUST subtract by 1 because NO_INTR offsets IIDX values. 
```

i509VCB (approved):
Small clarity doc change. Implementation looks good.

PR #4799, opened 2025-10-28
stm32: timer sync proposal

xoviat:
proposal related to synchronizing timers. will discuss on matrix.


xoviat:
I am rethinking this and may come back to it later.

PR #4807, opened 2025-10-29
stm32/uart: fix rb uart race

### LINKED ISSUE #4682: stm32/usart RingBufferedUartRx can race hardware when clearing interrupt flags
In `check_idle_and_errors()`, the status register is read, then all possible flags are cleared, then whatever was read is processed.

The operation happens inside a critical section "so that the flags aren't set after read and before clear", but the critical section doesn't stop the USART hardware from setting the flags between the read and the clear. Instead, only flags that were read as set should be cleared, which is what's done in `UartRx` and `BufferedUartRx`.

I'll try and get a fix together for this later if no one gets to it first, but right now I'm not using RingBufferedUartRx except for testing, so it's not as urgent for me.

https://github.com/embassy-rs/embassy/blob/4b5707b2563c725ce9bcd5537e2637d9bcaae7c6/embassy-stm32/src/usart/ringbuffered.rs#L310-L323

xoviat:
closes #4682

### DIFF
--- a/embassy-stm32/src/usart/ringbuffered.rs
+++ b/embassy-stm32/src/usart/ringbuffered.rs
@@ -340,19 +340,16 @@ impl Drop for RingBufferedUartRx<'_> {
 /// For usart_v1 and usart_v2, all status flags must be handled together anyway because all flags
 /// are cleared by a single read to the RDR register.
 fn check_idle_and_errors(r: Regs) -> Result<bool, Error> {
-    // Critical section is required so that the flags aren't set after read and before clear
-    let sr = critical_section::with(|_| {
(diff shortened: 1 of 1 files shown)


xoviat:
cc @adamgreig

adamgreig:
Thanks, I forgot about this! Do we need the critical section at all?

xoviat:
The other areas of the uart mod where this is present have removed the critical section. Shall I remove it here?

xoviat:
critical section removed! anything else?

xoviat:
bender run

adamgreig (approved):
LGTM!

Issue #4723, opened 2025-09-29
DBCKEND flag not set on STM32H753, causing init_sd_card to hang


KOBA789:
I'm encountering an issue when using SDMMC on STM32H753 with Embassy.

* When calling `init_sd_card`, the function hangs.
* After investigation, I found that it gets stuck in `complete_datapath_transfer`, waiting for the DBCKEND flag.
* However, checking the registers showed that the DBCKEND flag is never set.

According to the STM32H753 reference manual, the DBCKEND flag should be set in this situation, but in my experiments it never happens.

I have a hypothesis (not fully confident): when using IDMA, the DBCKEND flag might not be set at all. This could potentially be an erratum.

As a workaround, I tried calling `complete_datapath_transfer(false)`, and then everything worked correctly.

I also found this related commit: [https://github.com/embassy-rs/embassy/commit/c8a4a4995844be1b61d1a1479a6009eeb69b1117](https://github.com/embassy-rs/embassy/commit/c8a4a4995844be1b61d1a1479a6009eeb69b1117)

From this commit, I can see that the code was intentionally changed to wait for DBCKEND instead of DATAEND. It seems to reflect a deliberate design choice, but I couldn’t find any documentation or reasoning behind this change.

Could you explain the rationale for preferring DBCKEND over DATAEND here?
Also, has anyone observed similar behavior with IDMA and DBCKEND on STM32H7?

KOBA789:
@Dirbaio I'd like to understand the background behind this change. Could you explain a bit? https://github.com/embassy-rs/embassy/commit/c8a4a4995844be1b61d1a1479a6009eeb69b1117

szechyjs:
I'm experiencing the same issue with `init_sd_card` hanging on STM32L5, although I can't confirm if its related to the above `DBCKEND` change.

Edit: seeing same issue on STM32U5 as well.

szechyjs:
I just worked my way through my pile of Nucleo board and discovered that all the `sdmmc_v2` boards that I tried (L5, U5, H7) experienced this issue, the `sdmmc_v1` board (F4) worked fine. Like mentioned above setting `complete_datapath_transfer(false)` resolves the issue.

Looking at the reference manual there are some possible preconditions for the `DBCKEND` flag to fire, I'm not sure how to verify if these are being set.

<img width="619" height="101" alt="Image" src="https://github.com/user-attachments/assets/2ba44053-4dae-42d7-9fd0-097f8c4d822d" />

[...]

PR #4809, opened 2025-10-30
fix: stm32/sdmmc: don't wait for DBCKEND flag on sdmmc_v2 devices

szechyjs:
Possible fix for #4723. This seems to resolve the problem, however as pointed out in the original issue, we don't know the intent of https://github.com/embassy-rs/embassy/commit/c8a4a4995844be1b61d1a1479a6009eeb69b1117. Should we actually be waiting for `DBCKEND` or is `DATAEND` sufficient on sdmmc_v2/IDMA devices?

Update: this resolves the hanging issue, however it results in CRC errors when compiled in release mode (debug works fine though).

### DIFF
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -1035,6 +1035,8 @@ impl<'d, T: Instance> Sdmmc<'d, T> {
     #[allow(unused)]
     async fn complete_datapath_transfer(block: bool) -> Result<(), Error> {
         let res = poll_fn(|cx| {
+            // Compiler might not be sufficiently constrained here
+            // https://github.com/embassy-rs/embassy/issues/4723
             T::state().register(cx.waker());
             let status = T::regs().star().read();
 


xoviat:
Do you know the cause of the CRC errors?

szechyjs:
> Do you know the cause of the CRC errors?

I just pushed what seemed to fix the DCRC issue. There was definitely some kind of race condition/timing issue, as a simple defmt::debug! line in the right place would fix things. But I believe the root issue was that the status register data was stale/incorrect during the poll_fn, as it would have the `dcrcfail` bit set on first pass, but the interrupt would never fire for it.

The current state of this PR is working pretty reliably for me now in both debug/release mode. I'd appreciate it if some others can test it as well.

embassy-stm32/src/sdmmc/mod.rs
```diff
@@ -1032,12 +1032,11 @@ impl<'d, T: Instance> Sdmmc<'d, T> {
...
     /// Wait for a previously started datapath transfer to complete from an interrupt.
     #[inline]
+    #[allow(unused)]
     async fn complete_datapath_transfer(block: bool) -> Result<(), Error> {
-        let regs = T::regs();
-
         let res = poll_fn(|cx| {
             T::state().register(cx.waker());
-            let status = regs.star().read();
+            let status = T::regs().star().read();
```
xoviat, line 1041: If you were to revert this, and then put `compiler_fence(Ordering::SeqCst)` after registering the waker, is the problem resolved?

xoviat:
ea77131 should not impact the behavior of the logic, and only works because of compiler implementation. maybe this can be merged, but the issue cannot be closed until the compiler is constrained appropriately.

embassy-stm32/src/sdmmc/mod.rs
szechyjs: I tried the following and it didn't work. It resulted in the same crc errors.
```patch
diff --git a/embassy-stm32/src/sdmmc/mod.rs b/embassy-stm32/src/sdmmc/mod.rs
index b698c475f..e220466c5 100644
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -5,6 +5,7 @@ use core::default::Default;
 use core::future::poll_fn;
 use core::ops::{Deref, DerefMut};
+use core::sync::atomic::{Ordering, compiler_fence};
 use core::task::Poll;

 use embassy_hal_internal::drop::OnDrop;
@@ -1033,9 +1034,12 @@ impl<'d, T: Instance> Sdmmc<'d, T> {
     /// Wait for a previously started datapath transfer to complete from an interrupt.
     #[inline]
     async fn complete_datapath_transfer(block: bool) -> Result<(), Error> {
+
         let res = poll_fn(|cx| {
             T::state().register(cx.waker());
-            let status = T::regs().star().read();
+            compiler_fence(Ordering::SeqCst);
+            let status = regs.star().read();

             if status.dcrcfail() {
                 return Poll::Ready(Err(Error::Crc));
```

szechyjs: For what its worth, searching around the codebase, most `poll_fn` usage calls `T::regs()` within the wrapped function, rather than using a previously stored variable.

xoviat: I tried, but could not figure it out either. Add a comment above the line pointing to #4723, and noting that the compiler may not be sufficiently constrained here. Then, I will merge this as an improvement over the current situation, if nothing else comes up.

szechyjs:
Go figure, now I can't get debug mode to work without a DCRCFAIL running on latest commit (release mode fine). I have a feeling we got lucky before with the status register fix.

PR #4810, opened 2025-10-30
embassy-nrf: add `gpiote::InputChannel::wait_for_high/low()`

de-vri-es:
This PR turns `gpiote::InputChannel::wait()` into a `fn wait() -> impl Future` instead of an `async fn`. This guarantees that the interrupt is enabled and the event flag is cleared directly when the function is called, and not delayed until the future is polled.

This enables use cases like this:
```rust
let wait = input_channel.wait();
if input.pin().is_low() {
  return;
}
wait.await;
```

Which should give race-free waiting for a pin to become low.

Note: I also think this function should actually take `&mut self`. It clears the event flag, so multiple concurrent calls will interfere with each-other.

However, making it `&mut self` will actually break the use case above :p 

So.. do we :see_no_evil: about the `&self` or should I maybe directly implement functions like `wait_for_high()` and `wait_for_low()` to avoid races and still support the use case above?

Those functions might also actually want to change the event polarity though, or they would need to report an error if the polarity is wrong. Or we would need separate types for the different polarities, but that feels a bit overkill.

### DIFF
--- a/embassy-nrf/CHANGELOG.md
+++ b/embassy-nrf/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - bugfix: Do not write to UICR from non-secure code on nrf53
 - bugfix: Add delay to uart init anomaly fix
 - changed: `BufferedUarte::read_ready` now uses the same definition for 'empty' so following read calls will not block when true is returned
+- added: add `gpiote::InputChannel::wait_for_high()` and `wait_for_low()` to wait for specific signal level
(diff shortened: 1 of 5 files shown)


lulf:
I think my preference would be to add wait_for_low, wait_for_high to the API and require &mut. The example code both using wait and input seems like something we'd want to avoid.

de-vri-es:
Updated and tested with the `nrf9160` :)

de-vri-es:
Thanks for the review!

s-ol:
should `InputChannel` not implement `embedded_hal_async::digital::Wait`?

PR #4813, opened 2025-10-31
add backup sram

### LINKED ISSUE #4659: Battery backed RAM
What do you think about adding some nice API for accessing the battery backed RAM found in, for example, the stm32h5 series?

---

I suppose we could always offer some simple API with read/write methods.

However, do you think it would be possible to figure out how to achieve something like this in a safe way?

```rust
enum ReasonForDataReset {
    /// This is the first time we run or the first time after battery power was lost
    BackupRamDisabled,

    /// The struct layout does not seem to match
    ///
    /// `all_bytes` may be used to try to recover if the firmware was updated on purpose
    StructureMissmatch{ all_bytes: &mut[u8] }
}

...

#[repr(C)]
struct MyStruct {
    foo: u32,
    bar: MyEnum, // <-- Not all bit patterns are valid for this type
}

// This will be run the first time to set the defaults and on struct layout missmatch
let fn_to_setup_defaults = |_reason: ReasonForDataReset| MyStruct { foo: 42 };

// Here we get a mutable reference to the value stored in battery backed ram
let x: &'static mut MyStruct = BackupRam::init::<MyStruct>(p.BACKUP_RAM, fn_to_setup_defaults);

loop {
    x.foo += 1; // These writes will persist resets and loss of vcc
    defmt::info!("I have ran {} iterations since I was programmed/battery was inserted", x.foo);
}
```

How do we ensure
* The user uses the same type with the same layout every time
    * Would it be possible to create a checksum for the types layout and run fn_to_setup_defaults on missmatch or something?
* The type does not contain pointers/references out of the battery backed memory
    * Create a trait for types that are safe to store?
* Large structures does not overflow the stack in fn_to_setup_defaults.
* Ensure the compiler does not think it is reading uninitialized  garbage.

xoviat:
taken forward from a previous pr.

closes #4659.

### DIFF
--- a/embassy-stm32/src/rcc/bd.rs
+++ b/embassy-stm32/src/rcc/bd.rs
@@ -1,6 +1,8 @@
 use core::sync::atomic::{Ordering, compiler_fence};
 
 use crate::pac::common::{RW, Reg};
+#[cfg(backup_sram)]
+use crate::pac::pwr::vals::Retention;
 pub use crate::pac::rcc::vals::Rtcsel as RtcClockSource;
 use crate::time::Hertz;
 
@@ -89,6 +91,8 @@ pub struct LsConfig {
     pub rtc: RtcClockSource,
     pub lsi: bool,
     pub lse: Option<LseConfig>,
+    #[cfg(backup_sram)]
+    pub enable_backup_sram: bool,
 }
 
 impl LsConfig {
@@ -113,6 +117,8 @@ impl LsConfig {
                 peripherals_clocked: false,
             }),
             lsi: false,
+            #[cfg(backup_sram)]
+            enable_backup_sram: false,
         }
     }
 
@@ -121,6 +127,8 @@ impl LsConfig {
             rtc: RtcClockSource::LSI,
             lsi: true,
             lse: None,
+            #[cfg(backup_sram)]
+            enable_backup_sram: false,
         }
     }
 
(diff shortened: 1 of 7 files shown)


xoviat:
confirmed on hardware.

xoviat:
bender run

embassy-stm32/src/rcc/bd.rs
```diff
@@ -193,6 +203,22 @@ impl LsConfig {
...
+                w.set_bren(match self.enable_backup_sram {
+                    true => Retention::PRESERVED,
+                    false => Retention::LOST,
+                });
+            });
+
+            // Wait for backup regulator voltage to stabilize
+            while self.enable_backup_sram && !crate::pac::PWR.bdsr().read().brrdy() {}
+        }
+
```
usbalbin, line 221: I mean it has to be some part of this that breaks things, since this is the only code change that affect all H5 tests, right?

> WARN - Deadline exceeded!

Is that the host giving up waiting for the target to finish?

xoviat:
bender run

xoviat:
bender run

xoviat:
bender run

usbalbin:
It finally worked? :)

What was the problem in the end?

xoviat:
The teleprobe server runs on a separate vm from the ci, and if it crashes while a usb port is disabled (usb port power is cycled before each test), the usb port is left disabled and not recognized when teleprobe is restarted. Then, all subsequent jobs will be unable to access that device.

I'm working on a fix, but only the project leader has access to the teleprobe vm, so I have to be careful not to update the pac (which is triggering this) before the issue is resolved, so it doesn't take down the ci again.

PR #4819, opened 2025-11-02
Add SDRAM constructor for 16 data bits, 13 address bits

northernpaws:
Adds an FMC SDRAM constructor for SDRAM that uses 13 address bits and 16 data bits. Tested working with an ISSI `IS42S16160J-7TLI`.


northernpaws:
Oops, just realized I didn't put it in a branch on my fork which I meant to do. Will make a new PR with that fixed.

PR #4821, opened 2025-11-02
embassy-stm32: Automatic FDCAN bus off recovery

### LINKED ISSUE #4816: embassy-stm32: FDCAN bus off recovery option
Hi!

Currently, when an IT0 interrupt occurs and the BO flag is checked, the FDCAN driver automatically clears the INIT bit in the CCCR register.

https://github.com/embassy-rs/embassy/blob/9b1add3d83e98c6c9ae9230bf35cd89bba530a20/embassy-stm32/src/can/fdcan.rs#L90-L96

I believe that the driver does not have the right to decide on a monopoly recover from bus off.

RichardWGNR:
Hi!

With this PR, we give the developer the right to decide whether to use automatic FDCAN peripheral recovery in case of a bus error or not (for example manual error handling: some delay from last bus off). This PR grew out of Issue #4816.

Example
```rust
let mut cfg = CanConfigurator::new(p.FDCAN1, p.PB8, p.PB9, Irqs);
cfg.set_config(FdCanConfig {
    automatic_bus_off_recovery: false,
    ..Default::default()
});

cfg.start(OperatingMode::NormalOperationMode)
```

At present, the periphery is always recovered automatically, without the possibility of choice.

There are no breaking changes here.

Closes #4816.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -37,8 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - change: stm32/uart: BufferedUartRx now returns all available bytes from the internal buffer
 - change: timer: added output compare values
 - feat: timer: add ability to set master mode
+- fix: usart: fix race condition in ringbuffered usart
 - feat: stm32/fdcan: add ability to control automatic recovery from bus off ([#4821](https://github.com/embassy-rs/embassy/pull/4821))
-- 
+
 ## 0.4.0 - 2025-08-26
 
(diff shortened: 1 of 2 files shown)


MRWGNRRR:
Ready to merge

MRWGNRRR:
Should we use the term “**automatic_bus_off_recovery**”? Or should we use “**automatic_bus_off_management**” (ABOM) as the name of the bit in the bxCAN register? Or does it make no difference?

xoviat:
Naming on this is up to you since you requested the feature, and I have no other concerns. If someone else wants something different, it can be discussed.

PR #4823, opened 2025-11-02
Migrate from cortex-ar to aarch32-cpu

robamu:
- Feature name `arch-cortex-ar` remains the same.
- Legacy ARM architectures are not supported.


lulf (approved):
Thanks!

PR #4824, opened 2025-11-02
Disable PIO SM before setting Y register

### LINKED ISSUE #4822: embassy-rp OneWire intermittent failures after commit 8f10e3638d7
In diagnosing the issue, I was seeing an extra byte of 0xFF between the reset and the 0x55 I expected see for the match ROM command.  I tried a number of permutations of the code.  It turned out that just adding this line of code to write_bytes caused the issue.

    unsafe { self.sm.set_y(u32::MAX as u32) };

I then analyzed the PIO code:

        self.tx().push(value);
        self.exec_instr(OUT);

and the push() is:

    /// Force push data to TX FIFO.
    pub fn push(&mut self, v: u32) {
        PIO::PIO.txf(SM).write_value(v);
    }

And where is the state machine? Well, it is blocked on reading the FIFO at next_bit waiting to read/write the next block of data. Yes, the state machine is picking up the 0xff bits from setting the y register to 0xffffffff.

To fix the problem for me, I added a disable and enable the state machine around the existing set_y commands.

         unsafe {
            self.sm.set_enable(false);
            self.sm.set_y(u32::MAX as u32);
            self.sm.set_enable(true);
        }

mikeytdisco:
The command to set the Y register uses the FIFO to send the value.  Unfortunately, the state machine idle state is a blocking read on the FIFO and sometimes gets the data from the FIFO causing a corrupted data stream on the one wire bus. Fixes #4822 .

### DIFF
--- a/embassy-rp/CHANGELOG.md
+++ b/embassy-rp/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - add `wait_for_alarm` and `alarm_scheduled` methods to rtc module ([#4216](https://github.com/embassy-rs/embassy/pull/4216))
 - rp235x: use msplim for stack guard instead of MPU
 - Add reset_to_usb_boot for rp235x ([#4705](https://github.com/embassy-rs/embassy/pull/4705))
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #4830, opened 2025-11-03
low_power:  update api to allow reconfig

### LINKED ISSUE #4110: Cannot use RTC functionality alongside low_power module

xoviat:
closes #4110 

update the low-power api to allow reconfiguration. also update the write function to require a mutable reference, as it should have.

### DIFF
--- a/examples/stm32h5/src/bin/stop.rs
+++ b/examples/stm32h5/src/bin/stop.rs
@@ -12,7 +12,6 @@ use embassy_stm32::rcc::{HSIPrescaler, LsConfig};
 use embassy_stm32::rtc::{Rtc, RtcConfig};
 use embassy_stm32::{Config, Peri};
 use embassy_time::Timer;
-use static_cell::StaticCell;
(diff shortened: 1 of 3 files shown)


xoviat:
This addresses the immediate issue, but eventually the rtc should be moved into init if configured for low-power. The rtc cannot be split because the write method is mutable. For read-only methods, the time provider can be used.

PR #4831, opened 2025-11-03
mspm0: read_reset_cause()

doxxx:
This adds a `read_reset_cause` function that returns an enum value derived from the SYSCTL.RSTCAUSE register.

Based on the work by @charlesbmi in #4732.

Example usage code tested on-device:

```rs
match embassy_mspm0::read_reset_cause() {
    Ok(cause) => {
        use embassy_mspm0::ResetCause::*;
        let desc = match cause {
            NoReset => "No reset since last read.",
            PorHwFailure => "POR- violation, SHUTDNSTOREx or PMU trim parity fault.",
            PorExternalNrst => "NRST triggered POR (>1s hold).",
            PorSwTriggered => "Software triggered POR.",
            BorSupplyFailure => "BOR0- violation.",
            BorWakeFromShutdown => "SHUTDOWN mode exit.",
            BootrstNonPmuParityFault => "Non-PMU trim parity fault.",
            BootrstClockFault => "Fatal clock failure.",
            BootrstSwTriggered => "Software triggered BOOTRST.",
            BootrstExternalNrst => "NRST triggered BOOTRST (<1s hold).",
            BootWwdt0Violation => "WWDT0 violation.",
            SysrstBslExit => "BSL exit.",
            SysrstBslEntry => "BSL entry.",
            SysrstWwdt1Violation => "WWDT1 violation.",
            SysrstFlashEccError => "Flash uncorrectable ECC error.",
            SysrstCpuLockupViolation => "CPULOCK violation.",
            SysrstDebugTriggered => "Debug triggered SYSRST.",
            SysrstSwTriggered => "Software triggered SYSRST.",
            CpurstDebugTriggered => "Debug triggered CPURST.",
            CpurstSwTriggered => "Software triggered CPURST.",
        };
        uart_log!(logger, 256, "Reset cause: {}", desc);
    },
    Err(code) => {
        uart_log!(logger, 256, "Reset cause: Unknown ({})", code);
    }
}
```

### DIFF
--- a/embassy-mspm0/src/lib.rs
+++ b/embassy-mspm0/src/lib.rs
@@ -234,3 +234,114 @@ impl Iterator for BitIter {
         }
     }
 }
+
+/// Reset cause values from SYSCTL.RSTCAUSE register.
+/// Based on MSPM0 L-series Technical Reference Manual Table 2-9 and
+/// MSPM0 G-series Technical Reference Manual Table 2-12.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ResetCause {
+    /// No reset since last read
+    NoReset,
+    /// VDD < POR- violation, PMU trim parity fault, or SHUTDNSTOREx parity fault
+    PorHwFailure,
+    /// NRST pin reset (>1s)
+    PorExternalNrst,
+    /// Software-triggered POR
+    PorSwTriggered,
+    /// VDD < BOR- violation
+    BorSupplyFailure,
+    /// Wake from SHUTDOWN
+    BorWakeFromShutdown,
+    /// Non-PMU trim parity fault
+    #[cfg(not(any(
+        mspm0c110x,
+        mspm0c1105_c1106,
+        mspm0g110x,
+        mspm0g150x,
+        mspm0g151x,
+        mspm0g310x,
+        mspm0g350x,
+        mspm0g351x
+    )))]
+    BootrstNonPmuParityFault,
+    /// Fatal clock fault
+    BootrstClockFault,
+    /// Software-triggered BOOTRST
+    BootrstSwTriggered,
+    /// NRST pin reset (<1s)
+    BootrstExternalNrst,
+    /// WWDT0 violation
+    BootrstWwdt0Violation,
+    /// WWDT1 violation (G-series only)
+    #[cfg(any(mspm0g110x, mspm0g150x, mspm0g151x, mspm0g310x, mspm0g350x, mspm0g351x))]
+    SysrstWwdt1Violation,
+    /// BSL exit (if present)
+    SysrstBslExit,
+    /// BSL entry (if present)
+    SysrstBslEntry,
+    /// Uncorrectable flash ECC error (if present)
+    #[cfg(not(any(mspm0c110x, mspm0c1105_c1106, mspm0g351x, mspm0g151x)))]
+    SysrstFlashEccError,
+    /// CPU lockup violation
+    SysrstCpuLockupViolation,
+    /// Debug-triggered SYSRST
+    SysrstDebugTriggered,
+    /// Software-triggered SYSRST
+    SysrstSwTriggered,
+    /// Debug-triggered CPURST
+    CpurstDebugTriggered,
+    /// Software-triggered CPURST
+    CpurstSwTriggered,
+}
+
+/// Read the reset cause from the SYSCTL.RSTCAUSE register.
+///
+/// This function reads the reset cause register which indicates why the last
+/// system reset occurred. The register is automatically cleared after being read,
+/// so this should be called only once per application startup.
+///
+/// If the reset cause is not recognized, an `Err` containing the raw value is returned.
+#[must_use = "Reading reset cause will clear it"]
+pub fn read_reset_cause() -> Result<ResetCause, u8> {
+    let cause_raw = pac::SYSCTL.rstcause().read().id();
+
+    use ResetCause::*;
+    use pac::sysctl::vals::Id;
+
+    match cause_raw {
+        Id::NORST => Ok(NoReset),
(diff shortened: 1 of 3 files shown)


doxxx:
I don't understand those ci/build failures. This builds just fine locally.

doxxx:
Ah, I see the problem. For some MSPM0 chips the `WWDT0` reset cause is a different level, so the generated PAC enum is named differently: `BOOTWWDT0` vs `SYSWWDT0`.

felipebalbi:
> Ah, I see the problem. For some MSPM0 chips the `WWDT0` reset cause is a different level, so the generated PAC enum is named differently: `BOOTWWDT0` vs `SYSWWDT0`.

Maybe better to patch the PAC in that case?

doxxx:
There's other funny stuff going on with the MSPM0L13xx PAC -- it has WWDT0 and WWDT1 reset causes but there's only one WWDT0 defined for e.g MSPM0L1306... And some of the PAC reset cause values don't line up with the MSPM0 tech reference either.

I'll investigate some more.

embassy-mspm0/src/lib.rs
```diff
@@ -234,3 +234,94 @@ impl Iterator for BitIter {
...
+/// Based on MSPM0 Technical Reference Manual Table 2-9.
+/// Includes additional WWDT1 violation cause from MSPM0G series.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u8)]
+pub enum ResetCause {
+    /// No reset since last read
+    NoReset = 0x00,
+    /// VDD < POR- violation, PMU trim parity fault, or SHUTDNSTOREx parity fault
+    PorHwFailure = 0x01,
```
i509VCB: Now thats an omnibus reset cause lol

doxxx: image

Yep.

doxxx:
I couldn't think of a better way to do this. Even if I fixed the reset cause Id enum in the PAC, it still would have a different set of values for the G-series, which would make for a horrific `cfg` condition to get it right. I thought it would be simpler to hard-code the values according to the tech reference.

i509VCB:
> I couldn't think of a better way to do this. Even if I fixed the reset cause Id enum in the PAC, it still would have a different set of values for the G-series, which would make for a horrific `cfg` condition to get it right. I thought it would be simpler to hard-code the values according to the tech reference.

In this case if there are enough differences I would just define 2 enums and pick based on chip family with cfg.

doxxx:
> In this case if there are enough differences I would just define 2 enums and pick based on chip family with cfg.

It's one extra enum variant for G-series chips. Everything else is the same as far as I can tell. I can use cfg to make that one enum variant only preset for G series.

Are you okay with using the hard-coded values instead of the PAC enum values?

i509VCB:
> > In this case if there are enough differences I would just define 2 enums and pick based on chip family with cfg.
> 
> It's one extra enum variant for G-series chips. Everything else is the same as far as I can tell. I can use cfg to make that one enum variant only preset for G series.
> 
> Are you okay with using the hard-coded values instead of the PAC enum values?

Something like
```
PorHwFailure = pac::ResetCause::PorHwFailure
```
is fine for enum definitions. But I'm not sure we want to make it repr(u8)

doxxx:
> > > In this case if there are enough differences I would just define 2 enums and pick based on chip family with cfg.
> > 
> > 
> > It's one extra enum variant for G-series chips. Everything else is the same as far as I can tell. I can use cfg to make that one enum variant only preset for G series.
> > Are you okay with using the hard-coded values instead of the PAC enum values?
> 
> Something like
> 
> ```
> PorHwFailure = pac::ResetCause::PorHwFailure
> ```
> 
> is fine for enum definitions. But I'm not sure we want to make it repr(u8)

Alternatively, we don't define values for `ResetCause` and just let the compiler generate values. It doesn't look like other enums maintain the underlying register value.

i509VCB:
> Alternatively, we don't define values for `ResetCause` and just let the compiler generate values. It doesn't look like other enums maintain the underlying register value.

That's fine and probably preferable (we don't guarantee ABI stability of values)

doxxx:
Spurious job failure?

i509VCB:
> Spurious job failure?

C1104 failed to build. Probably that missing Id field you mentioned?

doxxx:
> > Spurious job failure?
> 
> C1104 failed to build. Probably that missing Id field you mentioned?

Ah crap, yes. I need to figure out how to do a build against every variant to catch that.

doxxx:
This depends on https://github.com/mspm0-rs/mspm0-data/pull/18 to add the missing PAC enum for the C-series but will still need further fixes since there are subtle differences in the PAC enum between series.

doxxx:
Still have to edit this to make two of the variants optional on c110x.

embassy-mspm0/src/lib.rs
```diff
@@ -234,3 +234,96 @@ impl Iterator for BitIter {
...
+
+/// Read the reset cause from the SYSCTL.RSTCAUSE register.
+///
+/// This function reads the reset cause register which indicates why the last
+/// system reset occurred. The register is automatically cleared after being read,
+/// so this should be called only once per application startup.
+///
+/// If the reset cause is not recognized, an `Err` containing the raw value is returned.
+#[must_use = "Reading reset cause will clear it"]
+pub fn read_reset_cause() -> Result<ResetCause, u8> {
```
i509VCB, line 306: Last comment is relating to return value. We do exhaustively measure each value hardware can produce, so we could mark other values as unreachable.

unreachable_unchecked is technically valid but would cause undefined behavior when new chips are added and this code isn't updated.

doxxx: I'd rather err (haha) on the side of caution and keep the Err return to avoid undefined behavior. A runtime panic is better than undefined behavior.

i509VCB (approved):
If possible can you squash the commits?

PR #4832, opened 2025-11-03
stm32/adc: merge v2 and v3 ringbuffers

xoviat:
So #4583 has teardown but not setup. It seems that teardown should not be pub, since once it's called, it's not possible to use the dma again. Then there is dma stop, which clears the dma buffer. My proposed interface is:

- start: controls `aden`; does not affect dma
- clear: clears the dma buffer
- stop: controls `aden`; does not affect dma but waits until adc is read.
- teardown: called on drop


xoviat:
ping @Jacke-debug

Jacke-debug:
Good that you started working on this @xoviat! I'm mainly waiting for feedback on my own two PRs: #4243 and #4786. The second one includes all changes from the first also but it may be preferable to merge in parts. The second one also has an example that highlights how to use the new contributions from both of them so that example may be a good staring point.

PR #4833, opened 2025-11-03
stm32/timer/simplepwm: Fix docs formatting and clarify timer usage

plaes:
Firstly, fix code formatting inside docstring and then hopefully clarify the timer availability a bit more.

### DIFF
--- a/embassy-stm32/src/timer/simple_pwm.rs
+++ b/embassy-stm32/src/timer/simple_pwm.rs
@@ -309,7 +309,9 @@ impl<'d, T: GeneralInstance4Channel> SimplePwm<'d, T> {
     /// Generate a sequence of PWM waveform
     ///
     /// Note:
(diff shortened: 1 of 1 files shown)


lulf:
You should add the `ignore` attribute as well.

plaes:
> You should add the `ignore` attribute as well.

Thanks, added. Also updated the same note for `waveform_up` docstring.

Would it make sense to add a method that would automatically infer correct `DMAx_CHy` based on the used timer/channel? For example currently we do:

```rust
let mut foo = SimplePwm::new(d.TIM4, ...);
foo.waveform_up(dp.DMA1_CH7.reborrow(), pwm_channel, color).await;
```

Or can't we somehow infer the DMA instance and channel based on the timer and fill in the DMA1_CH7 automatically?

PR #4836, opened 2025-11-04
further nRF54 improvements 

lulf:
* Enable some missing ppi channels and groups
* Initial support for random number generation using the cracen peripheral
* Ability to set clock speed for nrf54


Dirbaio:
Unleash the CRACEN :squid: :ocean:

PR #4838, opened 2025-11-04
stm32/low-power: cleanup and improve

xoviat:


xoviat:
bender run

xoviat:
bender run

PR #4840, opened 2025-11-04
Implement into_ring_buffered for g4. Add methods for injected ADC sampling.

Jacke-debug:
This PR expands the ADC module for STM32 g4 processors to support both injected and regular ADC conversions using the RingBuffered module. 

* Add methods for configuring ADC triggers for both regular and injected conversions
* Add example to showcase how both regular and injected ADC conversions can be used simultaneously

### DIFF
--- a/examples/stm32g4/.cargo/config.toml
+++ b/examples/stm32g4/.cargo/config.toml
@@ -1,21 +1,9 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-# Change this runner as required for your MCU.
-runner = [
-    "probe-rs",
-    "run",
-    "--chip",
-    "STM32G431VBTx",
-    "--speed",
-    "5000",
-    "--preverify",
-    "--log-format",
-    "{t} [{L}] {s}",
-]
-
+# replace STM32G071C8Rx with your chip as listed in `probe-rs chip list`
+runner = "probe-rs run --chip STM32G484VETx"
 
 [build]
-target = "thumbv7em-none-eabihf"
+target = "thumbv7em-none-eabi"
 
 [env]
-DEFMT_LOG = "info"
-DEFMT_RTT_BUFFER_SIZE = "4096"
+DEFMT_LOG = "trace"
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/adc/g4.rs
```diff
@@ -477,6 +507,191 @@ impl<'d, T: Instance> Adc<'d, T> {
...
+    /// should be exactly half of the `dma_buf` length. It is critical to call `read` frequently to
+    /// prevent DMA buffer overrun. Or configure a software trigger for the regular ADC conversions
+    /// at a fixed interval.
+    ///
+    /// [`read`]: #method.read
+    pub fn into_ring_buffered<'a>(
+        &mut self,
+        dma: Peri<'a, impl RxDma<T>>,
+        dma_buf: &'a mut [u16],
+        sequence: impl ExactSizeIterator<Item = (&'a mut AnyAdcChannel<T>, SampleTime)>,
```
Jacke-debug: @xoviat, this mut gives me a bit of a headache in my end-application as I need to store the pins in a static mut. It seems to me like these AnyAdcChannel does not need to be mut. Do you think its possible to change all the ADC signatures in the future and remove this mut?

xoviat:
What needs to happen is we need to have a split function (like in uart where we split uart rx and uart tx), where we split the injected conversions and regular conversions. So make `into_ring_buffered` like the other versions, and then create a new `into_ring_buffered_and_injected` method that returns `(ring_buffer, injected_struct)`, where the injected struct has the functions to control injected conversions.

xoviat:
you can discuss on https://app.element.io/#/room/#embassy-rs:matrix.org, to answer any questions you may have.

xoviat:
@Jacke-debug if you fix the rustfmt issue, I will merge this and then work on more cleanup.

Jacke-debug:
> @Jacke-debug if you fix the rustfmt issue, I will merge this and then work on more cleanup.

I did some more updates, hope that's ok! :)

PR #4843, opened 2025-11-05
nrf/i2s: allow master mode without MCK pin

kdrag0n:
Outputting the generated master clock is optional as per Nordic docs: "The MCK signal can be routed to an output pin (specified in PSEL.MCK) to supply external I2S devices that require the MCK to be supplied from the outside." [1]

Some I2S DACs, such as MAX98357A, don't require a master clock input so it can be left disconnected. nRF Connect SDK supports this.

[1] https://docs.nordicsemi.com/bundle/ps_nrf52833/page/i2s.html
[2] https://www.analog.com/media/en/technical-documentation/data-sheets/max98357a-max98357b.pdf

---

Not sure whether this API is ideal as it requires the caller to write `None::<Peri<'static, AnyPin>>`. Maybe these functions should be changed to take AnyPin like some other APIs?


lulf (commented):
I think what I'd prefer is that you rename the current to `fn new_master_with_mck()`, then remove mck from `new_master()`, and have both call into a `new_inner` or something that is not public.

Dirbaio:

kdrag0n:
Updated!

PR #4846, opened 2025-11-05
Light up Raspberry PI pico maths functions that were disabled

JYouren:
Light up Raspberry PI pico maths functions that were disabled


Dirbaio:
The float functions define intrinsics like [this](https://github.com/embassy-rs/embassy/blob/33f256ad01a1997b31e062d11fc45d1579609068/embassy-rp/src/float/add_sub.rs#L57) so they get called when you do normal Rust float math, like `a + b`. This is why they're not public, you're not supposed to call them directly.

JYouren:
ok, thanks. How do I get access to functions like ATAN?

Dirbaio:
the standard float methods should work I think 
https://doc.rust-lang.org/std/primitive.f32.html#method.atan

JYouren:
Looks like that is std only

PR #4847, opened 2025-11-05
stm32/dsi: support short write without parameters

yannick818:
Also allow a short DSI command without parameter.

This enables commands like this:
```rust
const SLPOUT: u8 = 0x11;
const SLPOUT_DATA: [u8; 0] = [];
dsi.write_cmd(VCID, SLPOUT, &SLPOUT_DATA).unwrap();
```


xoviat:
can you add a changelog entry?

PR #4851, opened 2025-11-06
update metapac

xoviat:


lulf:
Note: you can apply the label 'skip-changelog' to make it skip the changelog requirement

xoviat:
bender run

PR #4871, opened 2025-11-12
Standalone ethernet station management

datdenkikniet:
Instead of "locking" ethernet station management into the `Ethernet` peripheral, a new pseudo-peripheral will control access to this functionality.

This also removes the requirement for the `Ethernet` peripheral to know any details about the underlying PHY (such as the communication interface that the STM32 uses to communicate with it, lifetime bounds, etc.), making it more generic and in some sense shareable.

The `eth::StationManagement` trait is moved into the new `sma` module, and the `eth::Phy` trait no longer requires any specific PHY interaction pattern.

Some open questions:

1. `Sma` does not have a specific enable bit, so it cannot be an `RccPeripheral`. However, it must know the value of `hclk1` to determine the correct divider. I'm not sure if this is the "right" way to do it.

Follow-up work to this PR could involve making the `StationManagement` API `async` (the V2 peripheral has support for interrupts on MDIO transactions).

Most of the code is copy-pasted, but I have not tested this as of right now. Will comment & edit this post once I've gotten around to that.


Dirbaio:
looks good overall!

- I'd name the singletons ETH_MAC, ETH_SMA so it's obvious they're part of ETH
- I'd keep the SMA stuff inside the eth mod, for the same reason
- Maybe we should keep `fn new()` being the convenience constructor that constructs everything since that's what most people will use, and add the "advanced" constructor with a new name `fn new_with_phy()`.
   - fn new() can probably hardcode constructing a `GenericPhy` to simplify things, if that's not OK for you, you can use the advanced one.

embassy-stm32/src/eth/sma/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use stm32_metapac::common::{RW, Reg};
+
+pub use self::_version::*;
+
+/// Station Management Interface (SMI) on an ethernet PHY
+pub trait StationManagement {
+    /// Read a register over SMI.
+    fn smi_read(&mut self, phy_addr: u8, reg: u8) -> u16;
+    /// Write a register over SMI.
+    fn smi_write(&mut self, phy_addr: u8, reg: u8, val: u16);
```
xoviat, line 19: If you were to add `poll_link(cx)` with a default implementation that uses `embassy_time` with the fallback above, then it would be possible to implement `poll_link` with an interrupt.

datdenkikniet: I think this is an excellent idea, but not sure if putting it in this PR is the right move. There's quite a bit of movement already.

embassy-stm32/build.rs
```diff
@@ -1028,8 +1033,6 @@ fn main() {
         (("eth", "REF_CLK"), quote!(crate::eth::RefClkPin)),
         (("eth", "RX_CLK"), quote!(crate::eth::RXClkPin)),
         (("eth", "TX_CLK"), quote!(crate::eth::TXClkPin)),
-        (("eth", "MDIO"), quote!(crate::eth::MDIOPin)),
```
datdenkikniet, line 1031: I don't quite understand what this does. Removed it for now, but would be interested to hear if that is the correct move.

datdenkikniet:
> I'd name the singletons ETH_MAC, ETH_SMA so it's obvious they're part of ETH

I'd like to do this, but renaming `ETH` to `ETH_MAC` seems to break a bunch of stuff, and I'm not particularly well-versed in embassy's build.rs :P I would need some help/pointers if to get this working.
 
> I'd keep the SMA stuff inside the eth mod, for the same reason

Seems sensible. Done.

> Maybe we should keep `fn new()` being the convenience constructor that constructs everything since that's what most people will use, and add the "advanced" constructor with a new name `fn new_with_phy()`.

Yeah, I think this is a good idea. To decrease the "constructors with a billion arguments" problem I'll try to add a new `Pins` that can be constructed externally, and only have `new` and `new_mii` take all of the pins individually. Will do something about that later.

embassy-stm32/build.rs
xoviat: allows you to use this: `rx: Peri<'d, if_afio!(impl MDIOPin<T, A>)>,` in the SMI constructor.

datdenkikniet: Ah okay. Have tried my best to adapt to this now, which I think works.

embassy-stm32/src/eth/sma/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+trait SealedInstance {
+    fn regs() -> (Reg<AddressRegister, RW>, Reg<DataRegister, RW>);
+}
+
+/// MDIO instance.
+#[allow(private_bounds)]
+pub trait Instance: SealedInstance + PeripheralType + Send + 'static {}
```
xoviat, line 28: You need a pin trait on this and the pin trait impl needs to be changed in build.rs.

datdenkikniet: The `pin_trait_impl` macro seems to only accept paths of a very specific form (`crate::$mod:ident::$trait:ident`). 

Instead of adding the `pin_trait` here and fighting `pin_trait_impl` I've left the pin trait in `crate::eth` and made it require `sma::Instance` instead of just `Instance`.

Is that (more) acceptable?

xoviat: Yeah, sorry I missed that. Can you get the examples to build?

Dirbaio:
> I'd like to do this, but renaming ETH to ETH_MAC seems to break a bunch of stuff, and I'm not particularly well-versed in embassy's build.rs :P I would need some help/pointers if to get this working.

okay, don't worry about it then. ETH, ETH_SMA is also fine.

embassy-stm32/src/eth/sma/mod.rs
datdenkikniet: The CI examples: yes. 

I'm going to hold off on all the other examples until I've managed to do the "basic constructor signature" suggestion that dirbaio had above since that will be equally breaking again and I can't be bothered to rewrite it twice. 

Will probably find some time to do so tomorrow or on the weekend :)

Dirbaio:
>'ll try to add a new Pins that can be constructed externally

it's better to do the dumb thing and make N constructors even if it's verbose internally. It's the simplest thing for the user. Doing smart tricks with a "struct Pins" makes the embassy-stm32 code terser but makes it harder to use for the users.

Look at FMC, it has a bajillion constructors but it's fine.

datdenkikniet:
Alright, it all seems to compile now.

Open questions/things I need to investigate:

1. What is the minimal subset of things that are necessary to enable for the SMA to work? Is the current "just enable eth in ahb(1)enr" suffice?
2. Should `ETH_SMA` be an `RccPeripheral` It doesn't have a dedicated enable/disable bit, and IIRC it's only affected by the `ahb(1)enr` bit. Will investigate, but a question to reviewers: what exactly does being an `RccPeripheral` entail? Does `ETH_SMA` qualify? Does the current "just read RCC to get hclk1" suffice, or is there a better way to guarantee that clocks work as expected?

xoviat:
> What is the minimal subset of things that are necessary to enable for the SMA to work? Is the current "just enable eth in ahb(1)enr" suffice?

Probably, yes. You'll have to test that, though.

> Should ETH_SMA be an RccPeripheral It doesn't have a dedicated enable/disable bit, and IIRC it's only affected by the ahb(1)enr bit. Will investigate, but a question to reviewers: what exactly does being an RccPeripheral entail? Does ETH_SMA qualify? Does the current "just read RCC to get hclk1" suffice, or is there a better way to guarantee that clocks work as expected?

Ideally in this case where they both share a bit you would have a refcount and enable/disable the rcc when the refcount transitions between 0 and 1 (see usart for an example of this). An RccPeripheral is a peripheral that requires a bus clock to operate. Of course, this implies that nearly all peripherals are rcc peripherals, but the critical distinction here is that the peripheral must access an rcc register, rather than its own register. So you can't do `T::regs().enable_some_bit` to enable it because the rcc bit is not part of its own registers, it's part of the rcc registers. So in that case, if you want to enable it, you need the rccperipheral trait.

xoviat:
Actually, it doesn't appear that the rccperipheral trait is used for eth, even though it is specified.

datdenkikniet:
Alright, build is green now!

> Actually, it doesn't appear that the rccperipheral trait is used for eth, even though it is specified.

Yeah, this was previously only used for determining the `hclk1` frequency to calculate the SMA scaler. 

The implementation requires three bits: `ethen`,  `ethrxen` and `ethtxen`, so there's not one single bit defining "ethernet functioning".

I'm not sure what to do with this: the whole ethernet block "obviously" requires some clocks to be active, same goes for SMA and/or some pins to be connected. Does keeping the RCC trait as marker help?

Additionally, I noticed that the `Drop` impl for `Ethernet` never unsets `ethen`, it only unsets `ethrxen` and `ethtxen`. So it's a bit wooly. Would greatly appreciate some guidance :P

xoviat:
> The implementation requires three bits: ethen, ethrxen and ethtxen, so there's not one single bit defining "ethernet functioning".

You can try turning off ethtxen and ethrxen and test whether sma still works. If it does, then you can refcount ethen and enable the other two bits only for the mac.

datdenkikniet:
I've checked now, and on an F767 and H723 setting `ethen` suffices for the SMA to do its job.

I've updated the CHANGELOG and am hesitant to start poking at the enable bits: the current implementation is quite "fire and forget" and I would like to keep it like that for now. Naturally am willing to change that, but if I can get away with not doing that it makes my life a little easier ;)

PR #4872, opened 2025-11-12
Fix flash erase on dualbank STM32Gxxx

Wassasin:
Erasure on a dual-bank page would use the wrong page ID. The erase would silently fail, but writing the same page would result in a program error.

See #4400.

Tested on STM32G0B1RC.

### DIFF
--- a/embassy-stm32/src/flash/g.rs
+++ b/embassy-stm32/src/flash/g.rs
@@ -44,7 +44,6 @@ pub(crate) unsafe fn blocking_write(start_address: u32, buf: &[u8; WRITE_SIZE]) 
 }
 
 pub(crate) unsafe fn blocking_erase_sector(sector: &FlashSector) -> Result<(), Error> {
-    let idx = (sector.start - super::FLASH_BASE as u32) / super::BANK1_REGION.erase_size as u32;
     wait_busy();
     clear_all_err();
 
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/flash/g.rs
```diff
@@ -54,9 +53,9 @@ pub(crate) unsafe fn blocking_erase_sector(sector: &FlashSector) -> Result<(), E
             #[cfg(any(flash_g0x0, flash_g0x1, flash_g4c3))]
             w.set_bker(sector.bank == crate::flash::FlashBank::Bank2);
             #[cfg(flash_g0x0)]
-            w.set_pnb(idx as u16);
+            w.set_pnb(sector.index_in_bank as u16);
             #[cfg(not(flash_g0x0))]
-            w.set_pnb(idx as u8);
+            w.set_pnb(sector.index_in_bank as u8);
```
xoviat, line 58: How do you know this impacts others, like g4?

Wassasin: Should work in the same manner, with the exception that the G4 series can support both modes for all SKUs, and for the G0 some SKU's only support `single-bank`, only support `dual-bank`, and some support both.

The stm32_metapac defines which banks exist depending on the `single-bank` and `dual-bank` features. The code in https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/flash/common.rs#L174-L197 then determines which bank and page should be addressed depending on the requested address.

Swapping between `single-bank` and `dual-bank` on embassy is a manual process, and I am not sure if I will find the time to test soon. However I can test this change on a STM32G431KB and STM32G474RB with the default modes easily.

Wassasin: I can confirm that this patch works fine for STM32G431KB.

PR #4876, opened 2025-11-13
Fix UpDma not implemented on STM32H5

Wratox:


Dirbaio:
This is a data inconsistency, could you fix it in https://github.com/embassy-rs/stm32-data/ instead ?

usbalbin:
Something like this https://github.com/embassy-rs/stm32-data/pull/684?

PR #4884, opened 2025-11-13
Fix async adc reads on h5 and others

xoviat:
@yodaldevoid Does this resolve your issue?


yodaldevoid:
Yes, this seems to resolve the issue.

PR #4887, opened 2025-11-14
adc: fix g4 injected sequence

xoviat:
Fix g4 injected conversions and incorrect differential sequence logic.

@Jacke-debug does this work for you?


Jacke-debug:
Yes the example does work for me with these chanes, thank you!

PR #4890, opened 2025-11-15
Avoid generating update events when changing timer period

Jacke-debug:
When using the pre-load registers, for example with complementary PWM it is important that the ARR and CCR registers are updated simultaneously at the update event. It therefore does not make sense to generate an Update event in Software as part of the set_frequency method. Instead the function needs to return the new ARR such that new CCR values can be set before the update event occurs.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased - ReleaseDate
 
 - fix: Avoid generating timer update events when updating the frequency ([#4890](https://github.com/embassy-rs/embassy/pull/4890))
+- chore: cleanup low-power add time
+- fix: Allow setting SAI peripheral `frame_length` to `256`
 - fix: flash erase on dual-bank STM32Gxxx
 - feat: Add support for STM32N657X0
 - feat: timer: Add 32-bit timer support to SimplePwm waveform_up method following waveform pattern ([#4717](https://github.com/embassy-rs/embassy/pull/4717))
--- a/embassy-stm32/src/adc/g4.rs
+++ b/embassy-stm32/src/adc/g4.rs
@@ -1,14 +1,12 @@
 #[cfg(stm32g4)]
 use pac::adc::regs::Difsel as DifselReg;
 #[allow(unused)]
+#[cfg(stm32g4)]
+pub use pac::adc::vals::{Adcaldif, Adstp, Difsel, Dmacfg, Dmaen, Exten, Rovsm, Trovs};
+#[allow(unused)]
 #[cfg(stm32h7)]
 use pac::adc::vals::{Adcaldif, Difsel, Exten};
-#[allow(unused)]
-#[cfg(stm32g4)]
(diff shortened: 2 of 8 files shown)


xoviat:
Is the prescaler usually updated on timer overflow with this method? Or is it only the ARR that is being updated?

Jacke-debug:
> Is the prescaler usually updated on timer overflow with this method? Or is it only the ARR that is being updated?

From the STM32H7 reference manual we can see that:
>The prescaler can divide the counter clock frequency by any factor between 1 and 65536. It 
is based on a 16-bit counter controlled through a 16-bit register (in the TIMx_PSC register). 
It can be changed on the fly as this control register is buffered. The new prescaler ratio is 
taken into account at the next update event.

For the ARR value Auto-reload needs to be enabled by setting the ARPE bit. Otherwise (my guess is) it is not possible to change the ARR while running the timer(?). Not sure if we need to handle that in some way also. For the Complementary PWM implementation ARPE is enabled as part of the new method.

usbalbin:
I have mostly worked with the HRTIM timer found on some G4, H7 and F334 devices. Those have a bit `tuXdis` which may be used to temporarily disable the transfer from preload to active registers.

https://github.com/usbalbin/stm32-hrtim/blob/main/src/timer.rs#L305

RM0440 28.5.56
> **TAUDIS**: Timer A update disable
>
>This bit is set and cleared by software to enable/disable an update event generation temporarily on
timer A.
> * 0: Update enabled. The update occurs upon generation of the selected source.
> * 1: Update disabled. The updates are temporarily disabled to allow the software to write multiple
> registers that have to be simultaneously taken into account.

However I am unsure if there is something similar for the non-HRTIM timers.

I will take a look

usbalbin:
What do you think about UDIS in TIMx_CR1?

RM0481 40.4.1
> **UDIS**: Update disable
>This bit is set and cleared by software to enable/disable UEV event generation.
> * 0: UEV enabled. The Update (UEV) event is generated by one of the following events:
>  - Counter overflow/underflow
>  - Setting the UG bit
>  - Update generation through the slave mode controller
>  Buffered registers are then loaded with their preload values.
> * 1: UEV disabled. The Update event is not generated, shadow registers keep their value
> (ARR, PSC). However the counter and the prescaler are reinitialized if the UG bit is set or if
> a hardware reset is received from the slave mode controller.

RM0481
> 38.3.13 PWM mode
[...] The PWM mode can be selected independently on each channel (one PWM per tim_ocx
output) by writing 0110 (PWM mode 1) or 0111 (PWM mode 2) in the OCxM bits in the
TIMx_CCMRx register. The corresponding preload register must be enabled by setting the
OCxPE bit in the TIMx_CCMRx register, and eventually the autoreload preload register (in
upcounting or center-aligned modes) by setting the ARPE bit in the TIMx_CR1 register.
As the **preload registers are transferred to the shadow registers only when an update event
occurs**, before starting the counter, all registers must be initialized by setting the UG bit in
the TIMx_EGR register.

Jacke-debug:
I'm not sure I understand why UDIS would be relevant here? I want the update event to occur as normal, not triggered manually by the method for setting the timer frequency.

xoviat:
@Jacke-debug can you discuss on matrix?

xoviat:
can you resolve the merge conflict?

Issue #4863, opened 2025-11-10
embassy-stm32 - Unsplit timer types for more control


usbalbin:
Hi!

We are looking to use embassy-stm32 hal for a couple of projects, all which will use the hardware timers for different things. One of the things is triac control for soft starting an AC-motor. For that, we are thinking "one pulse mode", restart timer on ETR input, set output on compare match, reset on period etc. Perhaps even load the next pulse width from the DMA which is given the ramp profile as an array. 

However that does not really map to any of the timer types available in [embassy-stm32/src/timer](https://github.com/embassy-rs/embassy/tree/main/embassy-stm32/src/timer).

From what I can tell most of that is available but not on the same type, I would need some combination of for example
* one_pulse.rs (trigger input and single shot mode)
* simple_pwm.rs (output to pin and DMA support)

In my mind "one pulse", "pwm output" and "input capture" does not need to be mutually exclusive. I understand if the majority of users are happy with the different types and the discoverability is better that way. However it would be nice to have something slightly higher than the pac level as exposed in the low level timer type but still having the control to do what I want.

What do you think about moving down a lot of logic from the different types and putting it directly on the low level type so as to make that more useful(and delegate to the inner type in the higher level types so as to keep the same API) and/or perhaps add new timer type (insert too-many-standards-meme here) which can do all the things.

Coming from stm32g4xx-hal, and to some extent stm32h5xx-hal, where there is quite extensive use of builder types for setting up peripherals with lots of options. From what I have seen so far, embassy does not seem to do that quite as often. Is there any particular reason for this or would you think something like this could work?

For example

```rust
let mut triac_ctrl = MySwissArmyKnifeTimer::new(p.TIM1)
    //.frequency()
    .prescaler(...)
    .period(...)
    .ch1_out(output_pin, initial_duty)
    //.ch1_out_compl(...)
    .trigger(zero_detect_pin)
    .enable_oneshot_mode()
    .output_compare_mode(OutputCompareMode::PwmMode2)
    .init();

triac_ctrl.waveform_up(dma, MY_SOFT_START_RAMP_PROFILE).await;
```

Or do you have any other suggestions? :) I suppose some of the the things above could be made into methods on the final type to mutate it after it has been initialized and the things that have to be done before could be put into some `Config` struct if you prefer?

cc @denenilsson

xoviat:
[...]

Issue #4829, opened 2025-11-03
decide on a better changelog format


xoviat:
The principal issue to be avoided is merge conflicts that arise out of the currrent changelog format. A decision on this is pending.

szechyjs:
At the very least, can we start squashing PRs on merge? The commit history here is atrocious, it's a giant wall of merge commits, fmts, and changelog updates; and the actual changes scattered somewhere in between them all.

<img width="799" height="450" alt="Image" src="https://github.com/user-attachments/assets/b1189a46-16f2-4e30-8b91-7dda0de5ff81" />

i509VCB:
> At the very least, can we start squashing PRs on merge? The commit history here is atrocious, it's a giant wall of merge commits, fmts, and changelog updates; and the actual changes scattered somewhere in between them all.

I would certainly appreciate a way to squash pull requests without asking every contributor where I review to squash/fixup where everything fits in a single commit.

marwin1991:
I don't know if you are looking for such a  tool, but you can check out https://logchange.dev/tools/logchange/getting-started/

Jinja templates should meet your needs.

Apache Solr using it exactly because of wasting time for pull requests conflicts

https://lists.apache.org/thread/4dotf4qx4ss3qr3xonv2y63v7wdg40nt

PR #4892, opened 2025-11-15
stm32: Cleanup timers

usbalbin:
Try to clean up some duplicated logic in the different timer types by moving common methods down to the low level timer.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased - ReleaseDate
 
+- feat: Add waveform methods to ComplementaryPwm
 - chore: cleanup low-power add time
 - fix: Allow setting SAI peripheral `frame_length` to `256`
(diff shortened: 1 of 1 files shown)


usbalbin:
I have not yet had the opportunity to test this on real hardware

usbalbin:
cc @denenilsson and #4863

xoviat:
What do you think of #4890?

xoviat:
Can you resolve the merge conflict? #4829 is not done yet.

embassy-stm32/src/timer/complementary_pwm.rs
```diff
@@ -218,60 +218,57 @@ impl<'d, T: AdvancedInstance4Channel> ComplementaryPwm<'d, T> {
     ///
     /// Note:
     /// you will need to provide corresponding TIMx_UP DMA channel to use this method.
+    #[inline(always)]
```
xoviat, line 221: I'll merge this anyway, but in general, you shouldn't use this unless you have a good reason to think that you're smarter than the compiler (e.g. you are using inline assembly or something).

usbalbin: Ok, I'll avoid that in the future

PR #4894, opened 2025-11-15
Avoid generating update events when changing timer period, #2.

xoviat:
Continuation of #4890 to resolve merge conflict.

### DIFF
--- a/embassy-stm32/src/timer/low_level.rs
+++ b/embassy-stm32/src/timer/low_level.rs
@@ -275,7 +275,8 @@ impl<'d, T: CoreInstance> Timer<'d, T> {
     /// Generate timer update event from software.
     ///
     /// Set URS to avoid generating interrupt or DMA request. This update event is only
-    /// used to load value from pre-load registers.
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/timer/low_level.rs
```diff
@@ -272,6 +272,16 @@ impl<'d, T: CoreInstance> Timer<'d, T> {
         self.regs_core().cr1().modify(|r| r.set_cen(true));
     }
 
+    /// Generate timer update event from software.
+    ///
+    /// Set URS to avoid generating interrupt or DMA request. This update event is only
+    /// used to load value from pre-load registers.
+    pub fn generate_update_event(&self) {
```
usbalbin, line 280: Not sure how to phrase it, but should we add some note about this forcing an update which may produce glitches in the output waveform if the timer is already running?

xoviat: "If called when the timer is running, it may disrupt the output waveform."

PR #4897, opened 2025-11-16
MSPSM0: module MATHACL

bespsm:
- [mspm0: bump mspm0-metapac rev. & add minimath lib](https://github.com/embassy-rs/embassy/commit/80a8d7d9798514cd1cfa7f4188c51f25334952dc)
- [mspm0: add MATHACL module and following operations: sin,cos](https://github.com/embassy-rs/embassy/commit/88b3bc3f63dd00512298fc08d3f6b66cc1db72e8)
- [mspm0: add example of MATHACL based & tested on mspm0g3507](https://github.com/embassy-rs/embassy/commit/c237c2d0e19d7504778c63e2c780c1dd278da6fc)

### DIFF
--- a/embassy-mspm0/src/lib.rs
+++ b/embassy-mspm0/src/lib.rs
@@ -19,6 +19,7 @@ pub mod dma;
 pub mod gpio;
 pub mod i2c;
 pub mod i2c_target;
+#[cfg(any(mspm0g150x, mspm0g151x, mspm0g350x, mspm0g351x))]
 pub mod mathacl;
 pub mod timer;
 pub mod uart;


i509VCB:
The hardware is natively fixed point from what I can tell. I'm not sure as to why we would be converting to `f32` for the user?

bespsm:
> The hardware is natively fixed point from what I can tell. I'm not sure as to why we would be converting to `f32` for the user?

Do you suggest, to make end-user to convert fixed-point data?

embassy-mspm0/src/mathacl.rs
```diff
@@ -0,0 +1,257 @@
...
+pub enum Error {
+    ValueInWrongRange,
+    NBitsTooBig,
+}
+
+pub struct Mathacl<'d, T: Instance> {
+    _peri: Peri<'d, T>,
+    regs: &'static Regs,
+    _phantom: PhantomData<T>,
+}
```
i509VCB:
```suggestion
pub struct Mathacl<'d> {
    regs: &'static Regs,
    _phantom: PhantomData<&'d mut ()>,
}
```

Only need phantom data for lifetimes to be handled right. We don't need to store the instance of Peri here.

i509VCB: `regs` could also just be just the `Regs` type. Its a pointer in size anyways.

bespsm: done

bespsm:
@i509VCB I fixed the pipeline, please check

bespsm:
@i509VCB more concerns or shall we merge?

i509VCB:
Probably good to merge for now. We can address fixed point operations later.

PR #4900, opened 2025-11-16
[embassy-rp] Remove instance from rp spi and i2c

marcemmers:
A while back I made this (https://github.com/embassy-rs/embassy/pull/4155) change for the uart peripherals. The spi and i2c peripherals could still use this same change, so here is it.

The only thing I had to remove that somebody recently added (https://github.com/embassy-rs/embassy/pull/4405) was the `Debug` derive on the `I2c` type. I don't really understand the use-case of having it so I removed it since there is no easy way of keeping it. If there is a use-case for it I can try to find a way to manually implement it to get similar output.

### DIFF
--- a/embassy-rp/src/i2c_slave.rs
+++ b/embassy-rp/src/i2c_slave.rs
@@ -5,8 +5,9 @@ use core::task::Poll;
 
 use pac::i2c;
 
-use crate::i2c::{AbortReason, FIFO_SIZE, Instance, InterruptHandler, SclPin, SdaPin, set_up_i2c_pin};
-use crate::interrupt::typelevel::{Binding, Interrupt};
+use crate::i2c::{AbortReason, FIFO_SIZE, Info, Instance, InterruptHandler, SclPin, SdaPin, set_up_i2c_pin};
+use crate::interrupt::InterruptExt;
+use crate::interrupt::typelevel::Binding;
 use crate::{Peri, pac};
 
 /// I2C error
@@ -89,15 +90,16 @@ impl Default for Config {
 }
 
 /// I2CSlave driver.
-pub struct I2cSlave<'d, T: Instance> {
-    phantom: PhantomData<&'d mut T>,
+pub struct I2cSlave<'d> {
+    info: &'static Info,
     pending_byte: Option<u8>,
     config: Config,
+    phantom: PhantomData<&'d mut ()>,
 }
 
-impl<'d, T: Instance> I2cSlave<'d, T> {
+impl<'d> I2cSlave<'d> {
     /// Create a new instance.
-    pub fn new(
+    pub fn new<T: Instance>(
         _peri: Peri<'d, T>,
         scl: Peri<'d, impl SclPin<T>>,
         sda: Peri<'d, impl SdaPin<T>>,
@@ -111,9 +113,10 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
         set_up_i2c_pin(&sda, config.sda_pullup);
 
         let mut ret = Self {
-            phantom: PhantomData,
+            info: T::info(),
             pending_byte: None,
             config,
+            phantom: PhantomData,
         };
 
         ret.reset();
@@ -125,9 +128,10 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
     /// You can recover the bus by calling this function, but doing so will almost certainly cause
     /// an i/o error in the master.
     pub fn reset(&mut self) {
-        let p = T::regs();
+        let info = self.info;
+        let p = info.regs;
 
-        let reset = T::reset();
+        let reset = (info.reset)();
         crate::reset::reset(reset);
         crate::reset::unreset_wait(reset);
 
@@ -166,8 +170,8 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
 
         // mask everything initially
         p.ic_intr_mask().write_value(i2c::regs::IcIntrMask(0));
-        T::Interrupt::unpend();
-        unsafe { T::Interrupt::enable() };
+        info.interrupt.unpend();
+        unsafe { info.interrupt.enable() };
     }
 
     /// Calls `f` to check if we are ready or not.
@@ -181,7 +185,7 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
     {
         future::poll_fn(|cx| {
             // Register prior to checking the condition
-            T::waker().register(cx.waker());
+            self.info.waker.register(cx.waker());
             let r = f(self);
 
             if r.is_pending() {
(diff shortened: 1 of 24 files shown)


embassy-rp/src/i2c_slave.rs
```diff
@@ -89,15 +89,15 @@ impl Default for Config {
...
-pub struct I2cSlave<'d, T: Instance> {
-    phantom: PhantomData<&'d mut T>,
+pub struct I2cSlave {
+    info: &'static Info,
     pending_byte: Option<u8>,
     config: Config,
 }
 
-impl<'d, T: Instance> I2cSlave<'d, T> {
+impl I2cSlave {
```
ivmarkov: Hm, is removing the `'d` lifetime OK?

Wouldn't that mean that the peripherals will not be borrowed for the lifetime of the driver but rather - immediately available for reuse which is problematic?

marcemmers: I did think about that and I don't think its really an issue.

All functions creating an instance take the peripherals by move instead of borrowing them, so the borrow checker should take care of it anyway.

If you would `reborrow` the peripherals it's still possible, but that would be the case with the lifetime still there too.

ivmarkov: > I did think about that and I don't think its really an issue.
> 
> All functions creating an instance take the peripherals by move instead of borrowing them, so the borrow checker should take care of it anyway.
> 
> If you would `reborrow` the peripherals it's still possible, but that would be the case with the lifetime still there too.

I don't think that's correct. Perhaps you could try it by yourself? Reborrow a periph, push it down the driver's `new` function and then observe how the original periph is available for reuse _right after_ the `new` function had been called and well _before_ the driver is dropped.

IMO you should keep the `'d` lifetime into the driver as it used to be.

marcemmers: Seems that you are right. I did test it but only after my fix and of course then it was possible. I will add back the lifetime, thanks!

Learn something everyday with the borrow checker 😄

embassy-rp/src/i2c_slave.rs
```diff
@@ -89,15 +90,16 @@ impl Default for Config {
 }
 
 /// I2CSlave driver.
-pub struct I2cSlave<'d, T: Instance> {
-    phantom: PhantomData<&'d mut T>,
+pub struct I2cSlave<'d> {
+    info: &'static Info,
     pending_byte: Option<u8>,
     config: Config,
+    phantom: PhantomData<&'d ()>,
```
ivmarkov: Nit: shouldn't this be `PhantomData<&'d mut ()>`?

marcemmers: This goes a little bit over my head on what the reason would be for this.
I did find this: https://doc.rust-lang.org/nomicon/phantom-data.html, but this is talking about `PhantomData` with a `T` in there. Since `T` is `()` here I expect it does not really matter that much as we care mostly about the lifetime which is handled the same.

But I would rather get an answer from somebody who knows this better than me.

ivmarkov: >  Since T is ()

The "phantom" data you are modeling is NOT `()` but a **reference** (to `()`). To be more precise, a **mutable** (= exclusive) reference to the data.

And `()` is used simply because it does not really matter what exactly that reference points to, but rather that it _is_ a reference, with a certain lifetime.

And the reason I say you need to faithfully model a `&'d mut` rather than `&'d` is because there are differences between (exclusive) mutable references, and shared (immutable) references. Whether these differences are material for the concrete case of `Peri<'d>` I'm not sure (hence the "nit"), but it is till a good idea I think to try to faithfully model the reference. And for the case of `Peri::borrow()` it _is_ a mutable (exclusive) reference in that you can't borrow the same peripheral twice and pass the two borrows to two separate drivers.

Then - and if nothing else - the original reference in the `PhantomData` before your changes _was_ `&'d mut ()` so I guess if it wasn't broken we should not change it. :)

ivmarkov: Oh yeah, if it is not clear, "pointers" in Rust (i.e. references) are also "data".

ivmarkov: [Here's the definition of `Peri`](https://github.com/embassy-rs/embassy/blob/26f651ca3ad0c78e676c8b77293e70f92e4b9bba/embassy-hal-internal/src/peripheral.rs#L21). Since you are "erasing" the `Peri` type (as it does not matter) all that's left is the **reference** to the erased `T` peripheral, which reference - inside `Peri` - is `&'d mut`.

marcemmers: Ah I see my link was not complete, I was referencing this table at the bottom of the page:
https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns

Looking at that table it seems that `mut` or not is only influencing the variance of `T` and if the type will be Send/Sync. I am probably oversimplifying it though.

I agree with you that changing it back to `mut` is better, even if it is because it was that way like you said. Just also trying to learn a little bit ;)

Thanks for the explanations so far!

ivmarkov:
@marcemmers In case you are wondering what is going on - I don't have commit privileges for `embassy-rp`. You need to ping some of the maintainers for a more thoughtful code review and merge approval.

Probably (@lulf ?)

lulf (approved):
The change looks good to me, but perhaps you can create an explicit impl Debug that instead of removing it? Even though the instance is erased, it could still print the mode.

marcemmers:
> The change looks good to me, but perhaps you can create an explicit impl Debug that instead of removing it? Even though the instance is erased, it could still print the mode.

Sure I can add something, could be something like this: 
```
impl<M: Mode> core::fmt::Debug for I2c<'_, M> {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        let instance = match self.info.regs {
            pac::I2C0 => "I2C0",
            pac::I2C1 => "I2C1",
            _ => "N/A",
        };
        f.debug_struct("I2c")
            .field("Instance", &instance)
            .field("phantom", &self.phantom)
            .finish_non_exhaustive()
    }
}
```

But I kind of have the feeling that it is a bit over-engineered, especially as most other peripherals don't derive Debug either. So I am still wondering if it is useful to keep it. I saw another I2c PR for the raspberry crate also removing it.

ivmarkov:
> > The change looks good to me, but perhaps you can create an explicit impl Debug that instead of removing it? Even though the instance is erased, it could still print the mode.
> 
> Sure I can add something, could be something like this:
> 
> ```
> impl<M: Mode> core::fmt::Debug for I2c<'_, M> {
>     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
>         let instance = match self.info.regs {
>             pac::I2C0 => "I2C0",
>             pac::I2C1 => "I2C1",
>             _ => "N/A",
>         };
>         f.debug_struct("I2c")
>             .field("Instance", &instance)
>             .field("phantom", &self.phantom)
>             .finish_non_exhaustive()
>     }
> }
> ```
> 
> But I kind of have the feeling that it is a bit over-engineered, especially as most other peripherals don't derive Debug either. So I am still wondering if it is useful to keep it. I saw another I2c PR for the raspberry crate also removing it.

I think Ulf is asking you to just print the `M` so to say, while you are going a bit out of your way to print what was just lost/erased - the instance.

PR #4903, opened 2025-11-17
RP2040/235x: Fixed significant issue with i2c slave respond_to_read

computer-whisperer:
I2c slave reads larger than a single chunk are completely broken without this, as a full 16 bytes was being written to the FIFO on every poll rather than waiting for space. This just limits FIFO writes based on free space.

Tested on rp2040.

### DIFF
--- a/embassy-rp/src/i2c_slave.rs
+++ b/embassy-rp/src/i2c_slave.rs
@@ -5,9 +5,7 @@ use core::task::Poll;
 
 use pac::i2c;
 
-use crate::i2c::{
-    AbortReason, FIFO_SIZE, Instance, InterruptHandler, SclPin, SdaPin, set_up_i2c_pin,
-};
+use crate::i2c::{AbortReason, FIFO_SIZE, Instance, InterruptHandler, SclPin, SdaPin, set_up_i2c_pin};
 use crate::interrupt::typelevel::{Binding, Interrupt};
 use crate::{Peri, pac};
 
@@ -322,9 +320,7 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
                 }
 
                 if bytes_written < buffer.len() {
-                    for _ in 0..((FIFO_SIZE - p.ic_txflr().read().txflr()) as usize)
-                        .min(buffer.len() - bytes_written)
-                    {
+                    for _ in 0..((FIFO_SIZE - p.ic_txflr().read().txflr()) as usize).min(buffer.len() - bytes_written) {
                         p.ic_clr_rd_req().read();
                         p.ic_data_cmd().write(|w| w.set_dat(buffer[bytes_written]));
                         bytes_written += 1;
--- a/embassy-rp/src/i2c_slave.rs
+++ b/embassy-rp/src/i2c_slave.rs
@@ -301,7 +301,7 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
             return Err(Error::InvalidResponseBufferLength);
         }
 
-        let mut chunks = buffer.chunks(FIFO_SIZE as usize);
+        let mut bytes_written = 0;
 
         self.wait_on(
             |me| {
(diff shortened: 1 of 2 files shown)


embassy-rp/src/i2c_slave.rs
```diff
@@ -319,10 +319,11 @@ impl<'d, T: Instance> I2cSlave<'d, T> {
...
 
-                if let Some(chunk) = chunks.next() {
-                    for byte in chunk {
+                if bytes_written < buffer.len() {
+                    for _ in 0..((FIFO_SIZE - p.ic_txflr().read().txflr()) as usize).min(buffer.len() - bytes_written) {
                         p.ic_clr_rd_req().read();
-                        p.ic_data_cmd().write(|w| w.set_dat(*byte));
+                        p.ic_data_cmd().write(|w| w.set_dat(buffer[bytes_written]));
+                        bytes_written += 1;
                     }
```
9names, line 327: Yep, old logic was wrong, and this looks correct to me.
Do you think this logic would be more clear if we use bytes_written as the start of the range so that we could still have data access from the iterator instead of array indexing?

```rust
let bytes_remaining = buffer.len() - bytes_written;

if bytes_remaining > 0 {
    let fifo_free = (FIFO_SIZE - p.ic_txflr().read().txflr()) as usize;
    let write_len = fifo_free.min(bytes_remaining);
    for byte in &buffer[bytes_written..bytes_written + write_len] {
        p.ic_clr_rd_req().read();
        p.ic_data_cmd().write(|w| w.set_dat(*byte));
    }
    bytes_written += write_len;
```

9names: I tested both yours and the above with https://github.com/embassy-rs/embassy/blob/main/tests/rp/src/bin/i2c.rs, both pass.
the iterator style version is a little larger (not unexpected).

I couldn't get the test to fail with the old code though, even though there's a test for 32 byte slave read. Any ideas?

CBJamo: The old logic works because (after the first) each iteration of wait_on only starts when the tx_empty (actually that the tx fifo is < a threshold value, which we set to 0), done, or abort interrupt fires. So the either the fifo is empty, or it doesn't matter. 

@computer-whisperer must have hit some case where respond_to_read is called with bytes in the fifo. I'm not sure how that can happen though. It would be bad news though, because it means there's junk that'll go out before your real message, and probably you'll generate an abort interrupt with the last chunk.

computer-whisperer: I was seeing read transactions of hundreds of bytes in a buffer abbreviated down to only a few tens of bytes on the wire, with the first ten or so making it out intact followed by a scattered sampling of the values in the buffer.

Your logic only holds if nothing else can cause a wait_on hit. What if this async is participating in a join() or a select(), which can trigger extra poll()s? That is the scenario that was triggering issues for me.

CBJamo: Ah, of course. That does explain why the test passes though, since they're not in a join or similar. If it's feasible, it would be good to add a test case that covers this.

FWIW, I would recommend against putting respond_to_read, or in fact any async i2c_slave fn, in a select. Canceling these futures is very likely to stall the bus, and might lock it.

Dirbaio:
i'm going to close this, since the real bug it's fixing is "i2c doesn't behave well under cancelation".

The right fix is somehow ensure canceling method calls somewhat properly aborts the transaction, so it doesn't leave the i2c in a weird state that breaks the *next* transaction.

computer-whisperer:
> i'm going to close this, since the real bug it's fixing is "i2c doesn't behave well under cancelation".
> 
> The right fix is somehow ensure canceling method calls somewhat properly aborts the transaction, so it doesn't leave the i2c in a weird state that breaks the _next_ transaction.

That's not what this is doing though? The problem is that the existing version's correctness is gated to it only returning when it's interrupt triggers, and any poll from other sources (join or select) corrupts the fifo drive.

Dirbaio:
my bad, I see it now. old code unconditionally pushes bytes on poll, new code checks fifo level
Sorry for the confusion.

PR #4907, opened 2025-11-17
stm32: fix buffered flush method

xoviat:
I think I found the right fix after looking at again
15:25
the important thing is that tc can be set without the interrupt firing (tcie = false simply won't propagate to the nvic, but it will still set tc). 

and it can be set if txe completes without data
even if in the irq.

15:26
so to fix that, I clear tc at the end of the irq, and I check whether the irq is enabled before waking the flush waker
cc @jamesmunns

### DIFF
--- a/docs/pages/bootloader.adoc
+++ b/docs/pages/bootloader.adoc
@@ -43,14 +43,14 @@ Partition Size~dfu~= Partition Size~active~+ Page Size~active~
 +
 All values are specified in bytes.
 
-* BOOTLOADER STATE - Where the bootloader stores the current state describing if the active and dfu partitions need to be swapped. When the new firmware has been written to the DFU partition, a magic field is written to instruct the bootloader that the partitions should be swapped. This partition must be able to store a magic field as well as the partition swap progress. The partition size given by:
+* BOOTLOADER STATE - Where the bootloader stores the current state describing if the active and dfu partitions need to be swapped. When the new firmware has been written to the DFU partition, a magic field is written to instruct the bootloader that the partitions should be swapped. This partition must be able to store a magic field as well as the partition swap progress. The partition size is given by:
 +
-Partition Size~state~ = Write Size~state~ + (2 × Partition Size~active~ / Page Size~active~)
+Partition Size~state~ = (2 × Write Size~state~) + (4 × Write Size~state~ × Partition Size~active~ / Page Size~active~)
 +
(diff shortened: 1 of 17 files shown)


embassy-stm32/src/usart/buffered.rs
```diff
@@ -129,6 +125,10 @@ unsafe fn on_interrupt(r: Regs, state: &'static State) {
             });
         }
     }
+
+    // For others it is cleared above with `clear_interrupt_flags`.
+    #[cfg(any(usart_v1, usart_v2))]
+    sr(r).modify(|w| w.set_tc(false));
```
xoviat: This may not be necessary because the bit seems to be cleared by reading from the status register and writing to the data register. So maybe this can be moved back, and only `tcie` can be checked.

jamesmunns:
## Before

Screenshot 2025-11-18 at 18 12 41

## After

Screenshot 2025-11-18 at 18 12 22

## Code

```rust
// Mark sending
device.de.set_high();
Timer::after_micros(100).await;

defmt::info!("Sending uart data ({=usize})...", payload.len());

let do_send = async {
    device.usart.write_all(payload).await?;
    device.usart.flush().await?;
    Ok(())
};
let res: Result<(), embassy_stm32::usart::Error> = do_send.await;

// Note that the flush is "done" once the last character is pushed into the FIFO, this
// may be up to two chars from when the last bit is on the wire!
Timer::after_micros(device.two_char_delay_us.into()).await;

// Now wait another 100us or so to ensure the pin is set low (this is overkill)
device.de.set_low();
Timer::after_micros(100).await;
```

@xoviat I tested this change, cherry picked into https://github.com/jamesmunns/embassy/commit/ef821b0f0d017ccc60ced91c9f0bf581119c0c14 (I can't super easily update all the other deps right now), and this change does seem to improve things!

I did run it a couple of dozen times and seems to be consistent.

embassy-stm32/src/usart/buffered.rs
jamesmunns: Is there something specific you want me to test with this?

xoviat:
bender run

jamesmunns:
@xoviat yep, still looks good!

Screenshot 2025-11-20 at 15 28 33

updated my cherry pick here: https://github.com/embassy-rs/embassy/compare/main...jamesmunns:embassy:james/f405-test

PR #4908, opened 2025-11-18
RP2040/235x: Add I2C bus lockup detection and async recovery implementation.

computer-whisperer:
Various possibilities for bus-lock exist with the rp2040 and rp2350 i2c peripherals. The synopsis i2c peripheral is not included with the bus-clear option, meaning detection and recovery has to be implemented in software. 

This change adds detection of these conditions to both blocking and async usage, and bit-banged automatic recovery for the async variants.

### DIFF
--- a/embassy-rp/src/i2c.rs
+++ b/embassy-rp/src/i2c.rs
@@ -5,11 +5,13 @@ use core::future;
 use core::marker::PhantomData;
 use core::task::Poll;
 
+use embassy_hal_internal::drop::OnDrop;
 use embassy_hal_internal::{Peri, PeripheralType};
 use embassy_sync::waitqueue::AtomicWaker;
+use embassy_time::Timer;
 use pac::i2c;
 
-use crate::gpio::AnyPin;
+use crate::gpio::{AnyPin, Pin, SealedPin};
 use crate::interrupt::Interrupt;
 use crate::interrupt::typelevel::{Binding, Interrupt as _};
 use crate::{interrupt, pac, peripherals};
@@ -42,6 +44,10 @@ pub enum Error {
     InvalidWriteBufferLength,
     /// Target i2c address is out of range
     AddressOutOfRange(u16),
+    /// SDA is being held low
+    SDAHeldLow,
+    /// SCL is being held low
+    SCLHeldLow,
     /// Target i2c address is reserved
     #[deprecated = "embassy_rp no longer prevents accesses to reserved addresses."]
     AddressReserved(u16),
@@ -92,6 +98,8 @@ pub const FIFO_SIZE: u8 = 16;
 pub struct I2c<'d, M: Mode> {
     info: &'static Info,
     phantom: PhantomData<(&'d mut (), M)>,
+    sda: Peri<'d, AnyPin>,
+    scl: Peri<'d, AnyPin>,
 }
 
 impl<M: Mode> core::fmt::Debug for I2c<'_, M> {
@@ -169,6 +177,15 @@ impl<'d> I2c<'d, Async> {
 
         let mut abort_reason = Ok(());
 
+        // Cancel-safety
+        let on_drop = OnDrop::new(|| {
+            // Abort
+            p.ic_enable().write(|w| {
+                w.set_abort(true);
+                w.set_enable(true);
+            });
+        });
+
         while remaining > 0 {
             // Waggle SCK - basically the same as write
             let tx_fifo_space = self.tx_fifo_capacity();
@@ -236,7 +253,9 @@ impl<'d> I2c<'d, Async> {
             };
         }
 
-        self.wait_stop_det(abort_reason, send_stop).await
+        let res = self.wait_stop_det(abort_reason, send_stop).await;
+        on_drop.defuse();
+        res
     }
 
     async fn write_async_internal(
@@ -247,6 +266,15 @@ impl<'d> I2c<'d, Async> {
         let p = self.info.regs;
 
         let mut bytes = bytes.into_iter().peekable();
+
+        // Cancel-safety
+        let on_drop = OnDrop::new(|| {
+            // Abort
+            p.ic_enable().write(|w| {
+                w.set_abort(true);
+                w.set_enable(true);
+            });
+        });
 
         let res = 'xmit: loop {
             let tx_fifo_space = self.tx_fifo_capacity();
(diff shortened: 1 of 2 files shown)


embassy-rp/src/i2c.rs
```diff
@@ -473,18 +506,88 @@ impl<'d, T: Instance + 'd, M: Mode> I2c<'d, T, M> {
...
         p.ic_enable().write(|w| w.set_enable(false));
+
+        if !ignore_lockup {
+            if self.is_sda_low() {
+                return Err(Error::SDAHeldLow);
+            }
+            if self.is_scl_low() {
+                return Err(Error::SCLHeldLow);
+            }
+        }
```
CBJamo, line 547: What's the use case for this/why would you ever want to ignore a locked bus?

computer-whisperer: I encountered scenarios in testing where back-to-back blocking i2c reads could trigger a false positive on this. My main theory is that the last few bits of the previous transaction were still in-flight, and the previous behavior of ignoring it and letting the hardware run is the safer option if there is no automatic recovery system.

I am open to ideas for other ways to handle this.

CBJamo: Hmm, I think this is on me. When I added the SetConfig impl, I removed the [configuration of the ic_con register](https://github.com/computer-whisperer/embassy/commit/0c6d3ea05116c2798529ae68136f0e29f04f92e1#diff-17f91584e05b6990ce6a084fa9c1b86cbd6929f97741bc661690e2c4467cc512L380). Most of the register defaults to what we want, but the TX_EMPTY_CTRL bit defaults to 0, [while we want it to be 1](https://github.com/computer-whisperer/embassy/blob/624d9ac4381cf2c187ac6761274385149ede2dcf/embassy-rp/src/i2c.rs#L680). 

I think if that bit is set, the blocking fns will block until the transaction is complete, rather than until the fifo is empty, and eliminate the false positives.

Dirbaio:
ping! @computer-whisperer @CBJamo I don't know enough to know whether we want to do this fix or not:

> Hmm, I think this is on me. When I added the SetConfig impl, I removed the [configuration of the ic_con register](https://github.com/computer-whisperer/embassy/commit/0c6d3ea05116c2798529ae68136f0e29f04f92e1#diff-17f91584e05b6990ce6a084fa9c1b86cbd6929f97741bc661690e2c4467cc512L380). Most of the register defaults to what we want, but the TX_EMPTY_CTRL bit defaults to 0, [while we want it to be 1](https://github.com/computer-whisperer/embassy/blob/624d9ac4381cf2c187ac6761274385149ede2dcf/embassy-rp/src/i2c.rs#L680).
> 
> I think if that bit is set, the blocking fns will block until the transaction is complete, rather than until the fifo is empty, and eliminate the false positives.

About building bitbanged i2c recovery into embassy-rp directly: IME this is a common problem to i2c in all chips. A reset at the wrong time can leave things stuck. For example in our product we do a bitbanged dummy cycles + stop condition thing at boot to avoid this. So I do think it's useful. However, i'm not sure if it should be builtin to embassy-rp:
- This is something the user can do themselves in user code. You can bitbang the pins then drop the gpio and create the i2c.
- The problem it's solving is somewhat "universal", it's not specific to rp.

computer-whisperer:
> About building bitbanged i2c recovery into embassy-rp directly: IME this is a common problem to i2c in all chips. A reset at the wrong time can leave things stuck. For example in our product we do a bitbanged dummy cycles + stop condition thing at boot to avoid this. So I do think it's useful. However, i'm not sure if it should be builtin to embassy-rp:
> 
>     * This is something the user can do themselves in user code. You can bitbang the pins then drop the gpio and create the i2c.
> 
>     * The problem it's solving is somewhat "universal", it's not specific to rp.

The problem I ran into was not startup but intermediate -- mid runtime corruption and recovery. In that situation destroying the i2c peripheral, bit-banging a recovery, and returning control is nontrivial when a handful of mutexed handles exist in different subsystems sharing the bus. Would you expect a user that wants a robust i2c bus to create a wrapper type that handles recreating the driver while maintaining the share split?

Dirbaio:
I would expect the bus to never lock up after the MCU has booted. If all the i2c chips on the bus are compliant and our master implementation has no bugs it should be impossible for lockup to happen. The master should always do a stop condition at the end, the slave chip should always release the bus.

I'd look hard into finding and fixing possible bugs, instead of adding the lockup bitbang reset which is sort of a workaround. Do you have oscilloscope traces of the lockups?

Bootup is the only place where a lockup can legitimately happen, if the MCU resets in the middle of a transaction.

CBJamo:
> ping! @computer-whisperer @CBJamo I don't know enough to know whether we want to do this fix or not:
> 
> > Hmm, I think this is on me. When I added the SetConfig impl, I removed the [configuration of the ic_con register](https://github.com/computer-whisperer/embassy/commit/0c6d3ea05116c2798529ae68136f0e29f04f92e1#diff-17f91584e05b6990ce6a084fa9c1b86cbd6929f97741bc661690e2c4467cc512L380). Most of the register defaults to what we want, but the TX_EMPTY_CTRL bit defaults to 0, [while we want it to be 1](https://github.com/computer-whisperer/embassy/blob/624d9ac4381cf2c187ac6761274385149ede2dcf/embassy-rp/src/i2c.rs#L680).
> > I think if that bit is set, the blocking fns will block until the transaction is complete, rather than until the fifo is empty, and eliminate the false positives.

The change I made there was a regression, and should be fixed.

> I would expect the bus to never lock up after the MCU has booted. If all the i2c chips on the bus are compliant and our master implementation has no bugs it should be impossible for lockup to happen. The master should always do a stop condition at the end, the slave chip should always release the bus.

Unfortunately, I have also seen mid-operation lockups. In my case, the issue appears (we've never 100% proved this) to be a noisy slave causing other parts on the bus to miss the stop and lock the bus. So far we've "fixed" these problems by disconnecting the noisy slave (TI's bq25792 battery charger) and just missing out on the information it could give us. It's worth noting that the bit-banged recovery isn't a silver bullet, in the most broken case I've seen, neither a manual stop nor manual clocking recovers the bus. Without more insight into the i2c implementations in the various parts on the bus, it's very difficult to pin down why this is happening, but the only sure way to recover the bus in these cases is to reset everything.

For mid-operation, building the i2c recovery into embassy-rp isn't strictly necessary, but as computer-whisperer said, it will give better ergonomics. I'd be happy to see it included.

Dirbaio:
after the conversation here and on Matrix i'm on the fence. If this is such a common problem maybe it's worth including the fix even if it's ugly from the "purity" standpoint :) 

@CBJamo if you think we should merge this then it shall get merged.

Dirbaio:
bender run

White-Rabbit-Scientific:
I used a German i2c device and it locked up during use. This was a known "feature" of this particular part.

PR #4909, opened 2025-11-18
Remove deprecated atomic-polyfill crate for portable-atomic

BjornTheProgrammer:
This prevents a warning like the following from appearing when running `cargo audit -q` in another project.

```
➜  burn git:(embedded-fix) ✗ cargo audit -q --color always
Crate:     atomic-polyfill
Version:   1.0.3
Warning:   unmaintained
Title:     atomic-polyfill is unmaintained
Date:      2023-07-11
ID:        RUSTSEC-2023-0089
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0089
Dependency tree:
atomic-polyfill 1.0.3
└── embassy-rp 0.8.0
```


BjornTheProgrammer:
I would appreciate some help on this to figure out how to get the builds working properly for the tests.

Dirbaio:
the reason we aren't using `portable-atomic` there is to avoid the user having to specify these ugly cfgs. 

Instead of using `portable-atomic` could you rewrite the affected code to not require atomic CAS? for example using critical sections. Embassy-stm32 does this since it has the same problem with cortex-M0's, see this for example:

https://github.com/embassy-rs/embassy/blob/1f9c436afe6b0bcb306803d916e28df9e910479c/embassy-stm32/src/usart/mod.rs#L1138-L1144

BjornTheProgrammer:
Hmm you're probably right that we should just rewrite the code to not need portable atomics at all. I'll look into it on a new PR.

BjornTheProgrammer:
Closing in favor of PR #4948

PR #4913, opened 2025-11-18
stm32/eth: allow external poll irq

xoviat:
This allows using an exti pin to trigger link state changes, which is supported by some phys.

### DIFF
--- a/embassy-stm32/src/eth/generic_phy.rs
+++ b/embassy-stm32/src/eth/generic_phy.rs
@@ -1,14 +1,23 @@
 //! Generic SMI Ethernet PHY
 
+#[cfg(feature = "exti")]
+use core::mem;
 use core::task::Context;
 
+#[cfg(feature = "exti")]
+use embassy_hal_internal::Peri;
 #[cfg(feature = "time")]
 use embassy_time::{Duration, Timer};
-#[cfg(feature = "time")]
+#[cfg(any(feature = "exti", feature = "time"))]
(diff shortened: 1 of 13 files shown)


xoviat:
bender run

Dirbaio:
I think it'd make more sense to do this at the PHY level, not at the SMA level?

- SMA is just "smi read, smi write"
- GenericPhy has SMA + poll interval, poll_link does the "reschedule myself in poll_interval"
- GenericPhyWithPin has SMA + exti pin, poll_link does the "reschedule myself on interrupt"

xoviat:
okay, I will move it.

embassy-stm32/src/eth/generic_phy.rs
```diff
@@ -43,15 +52,24 @@ mod phy_consts {
...
 use self::phy_consts::*;
 
+/// Link State Poll Type.
+enum LinkStatePollType<'d> {
+    #[cfg(feature = "time")]
+    Interval(Duration),
+    #[cfg(not(feature = "time"))]
+    Continuous,
+    #[allow(dead_code)]
+    External((Input<'d>, bool, bool)),
```
Dirbaio, line 62: it's hard to know what these bools mean, i'd suggest changing this to have field names, or make a separate struct.

xoviat: This is a private enum.

PR #4915, opened 2025-11-19
Improvements to embassy-boot documentation

schult:
I've added documentation for all embassy-boot package features. This is particularly important for the `flash-erase-zero` feature, which is required to use embassy-boot on certain devices.

I've also corrected the stated size requirements for the STATE partition. The existing documentation did not account for the space used by the revert operation. The corresponding assert was also updated.


lulf (approved):
Thank you!

PR #4917, opened 2025-11-19
Add generation of a continuous/ ringbuffered waveforms for PWM

KorribanMaster:
This feature adds the generation of a continuous waveforms in the `SimplePWM`
In the future this may be expanded for waveform_up and `ComplementaryPWM` 

The goal is to generate a continous PWM signal that will run continuosly without any gaps using a circular DMA stream. This will block forever when awaited


xoviat:
Can you make this into_ring_buffered and add a ring buffered file like in adc or usart?

KorribanMaster:
> Can you make this into_ring_buffered and add a ring buffered file like in adc or usart?

Yes i think that is doable although there is a little difference here as we are "sending" data to the peripheral here instead of receiving. I would suggest the name should then be along the waveform_ringbuffered ? Or are you thinking in the direction of `RingBufferedPWM` as a seperate struct alongside `SimplePWM` and `ComplementaryPWM`

Another option would be to solve it like the embassy-nrf crate, where this is called [SequencePWM](https://docs.embassy.dev/embassy-nrf/git/nrf52840/pwm/struct.SequencePwm.html) so it would be more idiomatic i think to call it `SequencePWM`. This would also align with my final goal to actually implement double buffering with the dma where possible similiar to the discussion in this [stm32h7_hal](https://github.com/stm32-rs/stm32h7xx-hal/blob/5166da8a5485d51e60a42d3a564d1edae0c8e164/src/dma/mod.rs#L682) leveraging the DMA double buffer support as detailed in Idea 1 in this [issue](https://github.com/embassy-rs/embassy/issues/702) but i wanted to deliver it in smaller steps

xoviat:
`into_ring_buffered(self) -> RingBufferedWaveform<T: Instance>` consumes the channel and returns `PWMRingBuffer` struct, which contains `WritableDmaRingBuffer` (already exists) and the channel number. Then you can either write continuously to the buffer to keep it fed, or do nothing and dma will keep going on its own. I think you can get the same struct from both the simple and complementary pwm.

Let me know if you have further questions on this.

KorribanMaster:
Okay one more question to the function signature
```rust
//simple_pwm.rs
/// Simple PWM driver.
pub struct SimplePwm<'d, T: GeneralInstance4Channel> {
    inner: Timer<'d, T>,
}
....
impl<'d, T: GeneralInstance4Channel> SimplePwm<'d, T> {
    /// Create a new simple PWM driver.
  pub fn into_ring_buffered<C: TimerChannel>(&mut self, dma: Peri<'_, impl super::Dma<T, C>>, duty: &[u16]) -> RingBufferedPWMChannel{
    }
    ...
}

// ringbuffered.rs

/// A single ring buffered channel of a pwm, obtained from [`SimplePwm::into_ring_buffered`],
/// It is not possible to change the pwm frequency because
/// the frequency configuration is shared with all four channels.
pub struct RingBufferedPwmChannel<'d, T: GeneralInstance4Channel> {
    timer: ManuallyDrop<Timer<'d, T>>,
    channel: Channel,
    duty_buffer: WritableDmaRingBuffer,
}

impl<'d, T: GeneralInstance4Channel> RingBufferedPwmChannel<'d, T> {
    /// Construct a ringbuffer filled with the given buffer data.
   
    pub fn reset(&mut self){ }

    /// Get the remaining writable dma samples.
    pub fn len(&mut self) -> Result<usize, Error> { }

    /// Get the full ringbuffer capacity.
    pub const fn cap(&self) -> usize { }

    pub fn write(&mut self, buf: &[W]) -> Result<(usize, usize), Error> {
        }

    pub fn write_immediate(&mut self, buf: &[W]) -> Result<(usize, usize), Error> {
    }

    /// Wait for any ring buffer write error.
    pub async fn wait_write_error(&mut self, dma: &mut impl DmaCtrl) -> Result<usize, Error> {
    }

    /// Write an exact number of elements to the ringbuffer.
    ///
    /// Returns the remaining write capacity in the buffer.
    #[allow(dead_code)]
    pub async fn write_exact(&mut self, buffer: &[W]) -> Result<usize, Error> {
        
    }

    fn write_raw(&mut self, dma: &mut impl DmaCtrl, buf: &[W]) -> Result<(usize, usize), Error> {
    }

    fn write_buf(&mut self, offset: usize, value: W) {
    }
}

```
Does this go into the right direction?

xoviat:
That's basically okay. There is a problem with the current static typing that prevents what I really want to do, but it requires some thought and I can play around with the typing after the feature is functional. Unless there is an objection, I will merge this and clean it up myself, if I think it needs some adjustment, if it is functional.

KorribanMaster:
Let me check whether this is fully functional when I will be back home on the Weekend

KorribanMaster:
@xoviat  I think I got it quite a bit further however in the example i added the pwm is not working anymore...
```
cargo run -r --bin pwm_ringbuffer
   Compiling embassy-stm32f7-examples v0.1.0 (/home/eicke/git/embassy/examples/stm32f7)
    Finished `release` profile [optimized + debuginfo] target(s) in 0.32s
     Running `probe-rs run --chip STM32F767ZITx target/thumbv7em-none-eabihf/release/pwm_ringbuffer`
      Erasing ✔ 100% [####################]  32.00 KiB @  77.59 KiB/s (took 0s)
  Programming ✔ 100% [####################]  26.00 KiB @  38.62 KiB/s (took 1s)                                                                                                  Finished in 1.19s
0.000000 [INFO ] PWM Ring Buffer Example (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:20)
0.000000 [TRACE] BDCR ok: 00008200 (embassy_stm32 src/rcc/bd.rs:291)
0.000000 [DEBUG] flash: latency=6 (embassy_stm32 src/rcc/f247.rs:264)
0.000000 [DEBUG] rcc: Clocks { afif: MaybeHertz(0), hclk1: MaybeHertz(200000000), hclk2: MaybeHertz(200000000), hclk3: MaybeHertz(200000000), hse: MaybeHertz(8000000), hsi: MaybeHertz(16000000), hsi_hse: MaybeHertz(0), lse: MaybeHertz(0), lsi: MaybeHertz(0), pclk1: MaybeHertz(50000000), pclk1_tim: MaybeHertz(100000000), pclk2: MaybeHertz(100000000), pclk2_tim: MaybeHertz(200000000), pll1_q: MaybeHertz(100000000), plli2s1_q: MaybeHertz(0), pllsai1_p: MaybeHertz(0), pllsai1_q: MaybeHertz(0), rtc: MaybeHertz(32000), sys: MaybeHertz(200000000) } (embassy_stm32 src/rcc/mod.rs:78)
0.000000 [INFO ] Channel 1 (PE9/D6): Static 50% duty cycle (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:62)
0.000000 [INFO ] PWM max duty: 200 (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:66)
0.000030 [INFO ] Ring buffer capacity: 128 (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:88)
0.000061 [INFO ] Pre-writing initial waveform data... (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:94)
0.000091 [INFO ] Channel 2 (PE11/D5): Ring buffered sine wave started (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:98)
0.060119 [INFO ] Ring buffer getting full, wrote 1 of 32 (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:129)
0.110168 [INFO ] Ring buffer getting full, wrote 0 of 32 (pwm_ringbuffer src/bin/pwm_ringbuffer.rs:129)
``` 
It seems like the DMA and timer stop working after the first byte is transferred any Idea what could be wrong?

PR #4919, opened 2025-11-19
small tweak for USB HID reader

robamu:
Allows running from a mutable reference, which was useful for stuffing the reader into local RTICv2 resources. I could not send it to a task by value because the reader is not sendable.

### DIFF
--- a/embassy-usb/CHANGELOG.md
+++ b/embassy-usb/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased - ReleaseDate
 
 - Add support for USB HID Boot Protocol Mode 
+- `HidReader::run` does not consume itself anymore and only requires `&mut self`
 
 ## 0.5.1 - 2025-08-26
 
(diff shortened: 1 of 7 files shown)


Dirbaio:
this is a limitation in RTIC that doesn't affect embassy-executor. RTIC folks are working on it, see https://github.com/rtic-rs/rtic/pull/1116 https://github.com/rtic-rs/rtic/pull/1108

We shouldn't add workarounds in the public API for this limitation. The problem with allowing `&mut` is you can cancel the runner then restart it, which can break it if you do it halfway through a transfer.

robamu:
Oh, that is good to know. I have not thought of this case. So this becomes potentially problematic if you just save the future, drop it and then re-run the `run` method?

Just out of curiosity: Is this not problematic for all runners with the `run(&mut self) -> !` style that already exist, or are those runners specifically designed to allow this? I have seen several of those in the code base.

leftger:
Based on the age of this PR and no commits after dirbaios comments this PR should be closed

leftger:
I'll wait a week for any more changes but if none occur I'll close it then

robamu:
this can be closed

PR #4922, opened 2025-11-19
Bindable EXTI interrupts, and removal of AnyChannel support for ExtiInput

WillaWillNot:
Following up on #4028, getting this working sure was a puzzle!

Unlike other peripheral drivers, Exti needs to cope with the need to assign a single handler at the code level to a vector which represents interrupts for several drivers. Also unlike other drivers that look for bindings, the parameter that tells Exti which binding to look for... can be type-erased!

The first issue isn't that hard, the Exti's in-library handler just has to be generic over the interrupt rather than the instance, and the user will need to deal with the implementation detail of sometimes needing to bind to EXTI0_1 or similar in their bind_interrupts! invocation, rather than intuitively binding to EXTI0, EXTI1, etc.

For the second issue, after trying many different approaches, the best I could come up with was a type-erased AnyBinding, which is unfortunately kind of awkward to cast into. Binding has to be implementable by the user, which means you can't seal it with a sealed supertrait. Which means you can't blanket implement `From<Binding<I, H<I>>> for AnyBinding`, because Binding is user-implementable, and you clash with the blanket `From<T> for T` implementation. This also means the binding has to be checked at runtime, not compile time.

This also required representing the type of the handler in the binding somehow, so I had to add a HandlerType enum. (Interrupts already have an exhaustive enum for this.) This is implemented in a defaulted constant in the Handler trait, with a privacy guard to keep user code using the User type. Rather than trying to make an exhaustive change across the entire codebase to assign types, I just left everything but Exti using the default, since nothing else is expecting AnyBindings.

Not thrilled that I have to propose a brand new type erased interface for something used so widely across the codebase just to get bindable EXTI working, but I couldn't find another way. I know it's a bit inelegant, but I'm trying to build the minimum viable horse to put in front of my cart.

This is a draft because I'm aware there's a giant pile of examples I need to edit before it's ready to merge, but I wanted to get eyes on a breaking API change sooner rather than later. It's only breaking for `embassy_stm32::exti::Exti::new(),` everything else I was able to make backwards compatible.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -77,7 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat: stm32/flash: add async support for h7 family
 - feat: exti brought in line with other drivers' interrupt rebinding system ([#4922](https://github.com/embassy-rs/embassy/pull/4922))
 - removal: ExtiInput no longer accepts AnyPin/AnyChannel; AnyChannel removed entirely
-- change: build script now generates `unimpl_tsc` cfg option when a chip has a TSC peripheral but no driver
+- fix: build script ensures EXTI2_TSC is listed as the IRQ of EXTI2 even if the PAC doesn't
 - feat: stm32/lcd: added implementation
 
 ## 0.4.0 - 2025-08-26
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -56,16 +56,12 @@ fn main() {
 
     eprintln!("chip: {chip_name}");
 
-    cfgs.declare("unimpl_tsc");
     for p in METADATA.peripherals {
         if let Some(r) = &p.registers {
             cfgs.enable(r.kind);
             foreach_version_cfg(&mut cfgs, r.kind, r.version, |cfgs, cfg_name| {
                 cfgs.enable(cfg_name);
             });
-        } else if p.name == "TSC" {
-            //Even if the registers are missing, EXTI needs to know if TSC is present in silicon to know whether the EXTI2 interrupt is shadowed by EXTI2_TSC
-            cfgs.enable("unimpl_tsc")
         }
     }
 
@@ -357,8 +353,13 @@ fn main() {
     // ========
     // Generate interrupt declarations
 
+    let mut exti2_tsc_shared_int_present: Option<stm32_metapac::metadata::Interrupt> = None;
     let mut irqs = Vec::new();
     for irq in METADATA.interrupts {
+        // The PAC doesn't ensure this is listed as the IRQ of EXTI2, so we must do so
+        if irq.name == "EXTI2_TSC" {
+            exti2_tsc_shared_int_present = Some(irq.clone())
+        }
         irqs.push(format_ident!("{}", irq.name));
     }
 
@@ -1816,7 +1817,19 @@ fn main() {
     for p in METADATA.peripherals {
         let mut pt = TokenStream::new();
 
+        let mut exti2_tsc_injected = false;
+        if let Some(ref irq) = exti2_tsc_shared_int_present
+            && p.name == "EXTI"
+        {
+            exti2_tsc_injected = true;
+            let iname = format_ident!("{}", irq.name);
+            let sname = format_ident!("{}", "EXTI2");
+            pt.extend(quote!(pub type #sname = crate::interrupt::typelevel::#iname;));
+        }
         for irq in p.interrupts {
+            if exti2_tsc_injected && irq.signal == "EXTI2" {
+                continue;
+            }
             let iname = format_ident!("{}", irq.interrupt);
             let sname = format_ident!("{}", irq.signal);
             pt.extend(quote!(pub type #sname = crate::interrupt::typelevel::#iname;));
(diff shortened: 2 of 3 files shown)


WillaWillNot:
Alright, I think this is ready now. There were a few other issues to iron out, including handling the shared EXTI2_TSC interrupt on chips with TSC peripherals; I actually needed to add a cfg option for when the metapac contains a TSC peripheral but no registers for it, signaling that TSC is present in the silicon even though embassy can't yet implement the driver. (Example: STM32F398VE.) This involved adding a check to the build script, as well as a conditional branch to the impl_exti macro.

This previously worked due to the hardcoded EXTI handlers being defined through the for_each_interrupt macro, but now the EXTI peripherals themselves need to reference their IRQ, so for_each_interrupt won't work anymore.

In addition to affecting dozens of examples, this change affected the example used in the layer-by-layer section of the embassy book, so I had to update the language there too. While I was at it, I expanded the answer in the FAQ about manual interrupts to recommend using `Handler` and `bind_interrupts` (it previously only mentioned defining the raw ISR in place.)

(Also, I added a commented-out line to .vscode/settings.json to run rustfmt with +nightly, as a non-breaking hint to anyone else who gets bitten by there being nightly-only options set in the rustfmt 🙂)

Let me know if you have any thoughts on the approach here - I know I'm really diving into the deep end (all because I wanted my EXTI dependent code to run a little faster), and I hope my solution seems in line with the project philosophy.

xoviat:
So I don't want to waste your time with all of the work that you've already done, but I want to propose my idea. Keep in mind that @Dirbaio  will have the final say on this, so what I think may not be entirely accurate.

The first question that I asked myself is why anybinding was needed. Of course the answer is anychannel. Now that I know that, I have a solution that would be easier for the user and I think require less modification. If you decide to go this route, I would make sure to save this work so that you can bring it back later if you need to.

The first change will be typing the interrupt to the channel:

```
pub trait Channel: PeripheralType + SealedChannel + Sized {
    /// Interrupt for this peripheral.
    type Interrupt: interrupt::typelevel::Interrupt;
    /// Get the EXTI channel number.
    fn number(&self) -> PinNumber;
}
```

This will allow you to bind the interrupts using the standard approach. The next problem is that anychannel does not have an interrupt, so let's make one:

```
impl Channel for AnyChannel {
    type Interrupt = crate::interrupt::typelevel::AnyExtiInterrupt;
    fn number(&self) -> PinNumber {
        self.number
    }
}
```

Then we need to create the interrupt in interrupt.rs, right above the handler trait:

```
                    pub enum AnyExtiInterrupt {}
                    impl SealedInterrupt for AnyExtiInterrupt {}
                    impl Interrupt for AnyExtiInterrupt {
                        const IRQ: super::Interrupt = unsafe { core::mem::transmute(0) };
                    }
```

This leaves us with two problems: the first is that you cannot use the interrupt inside of the exti new function because it is potentially fake and the AnyExtiInterrupt is not bound. The latter problem can be addressed by modification of the bind_interrupts macro to bind AnyExtiInterrupt if one of the exti interrupts is bound. This modification may be nontrivial as a proc-macro may be required (I am not a macro expert).

The former problem could be addressed by adding a new get_irq() function to the Interrupt or SealedInterrupt trait, which would return the pac irq, and then returning that from the type exti peripheral and also using it to construct anychannel,  which could then be used to construct a new typelevel irq or to enable interrupts directly using the pac. This irq would also be used at runtime to confirm correct interrupt binding.

The advantage of this approach compared to yours is that the user would be blind to the complexity of this scheme. It still feels icky though, and it may be better to just remove anychannel.

xoviat:
@WillaWillNot I deleted my previous comment, the decision is that we are going to remove `AnyChannel` if you want to implement this, and you will need to remove `AnyBinding` as well.

WillaWillNot:
@xoviat  sounds good - I had considered proposing removing AnyChannel myself, I was just unsure about which side of the tradeoff "more exposed complexity, less edits in preexisting complex codebases" (AnyBinding) vs "less exposed complexity, more edits in preexisting complex codebases" (remove AnyChannel) would be better. Eliminating AnyChannel type erasure might be a bit of a pain in the butt in many codebases, but it's not *that* bad, you just might need some unrolled structs and matches instead of the convenience of iterating.

I'll have a new version within the next few days that does that.

I think this *might* also allow me to hide the complexity of the user needing to know whether to use e.g. EXTI13 or EXTI15_10 on their chip when writing their bind_interrupts. Maybe. Because of the fact that bind_interrupts needs to define both the binding and the precisely named C linkage ISR, it may still require either a proc_macro or a dependency on the `paste` crate, and I'm not sure if you'd want to introduce that dependency.

WillaWillNot:
Done! Wasn't able to find a way to hide the weird multipurpose interrupt mapping from the user, but I noticed a similar situation is already shown in the layer-by-layer example so I just added some text there for anyone who gets surprised by it.

Also, the TSC thing was even stickier than I initially thought - there's some chips that don't shadow EXTI2 with the TSC, it's on its own independent line. So, now I just have the build script search for EXTI2_TSC in the interrupt list, and if present, inject it into the EXTI interrupt mapping. It's ugly but it works. At least I'm not adding a cfg option anymore, and impl_exti! doesn't need the branching logic.

Really, that should be fixed in the PAC, and I might look into submitting a PR over there. It should list EXTI2_TSC as the line for EXTI2. Until then though, this ugly bandaid logic is needed to make sure the mapping happens.

(Making ExtiInput check the bindings at compile time rather than runtime surfaced that issue via a failing build - one good reason to try to stick with compile time checks when possible!)

xoviat:
Updating the PAC is not a big deal, since the git version is used. So feel free to fix the tsc issue, if you want.

PR #4923, opened 2025-11-20
Added KillableExecutor

cryptographix:
`KillableExecutor`, as its name suggests, is an `arch-std` `Executor` that can be cleanly killed. The main noticeable difference is that the `init` closure passed to `run()` has a second parameter of type `Killer`, which wraps a `Signaler` (so as to be `Copy` and `Clone`). This `killer` is checked during the `Executor`'s polling loop, and when signalled, causes the `Executor` to exit.

Rationale: Although it makes little sense to terminate a device-based Executor, `arch-std` is running on a conventional OS, and thus there are use-cases where it makes sense to terminate. Specifically, when testing 'embassy'-based code on `std-arch` with h/w mocks, having a way to terminate the `Executor` allows a `test_harness`to be created such that `#[test]` and `cargo test` can be used. Example: 
```
#[test]
fn test_clock() {
  test_harness(|spawner, killer| {
    // this is the task under test
    spawner.spawn(clock_task()).unwrap();

    // and here is the test-code, running as a separate task.
    spawner.spawn(clock_test_task(spawner, killer)).unwrap();
  });
}
```
In the example, the #task-under-test# `clock_task` is first spawned, followed by a separate test task `clock_test_task` which receives the killer object and signals the `Executor` to shut down once the tests have finished. In this way, `cargo test` continues and runs the next `#[test]`.

Example `test_harness`:
```
pub fn test_harness<F: FnOnce(Spawner, Killer)>(fun: F) {
  static EXECUTOR: StaticCell<KillableExecutor> = StaticCell::new();
  let executor = EXECUTOR.init(KillableExecutor::new());

  // Setup logging
  env_logger::builder()
    .filter_level(log::LevelFilter::Info)
    .format_timestamp(None)
    .format_target(false)
    .init();

  // and run the test
  executor.run(fun);
}
```
I'm pretty sure there's some cleanup missing .. and I'm not yet certain what happens if a test panics, but I'm submitting this initial PR as a way to solicit comments from the more knowledgeable, as to whether this is viable.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased - ReleaseDate
 
+- fix: Fixed ADC4 enable() for WBA
+- feat: allow use of anyadcchannel for adc4
+- fix: fix incorrect logic for buffered usart transmission complete.
 - feat: add poll_for methods to exti
 - feat: implement stop for stm32wb.
 - change: rework hsem and add HIL test for some chips.
--- a/embassy-stm32/src/adc/adc4.rs
+++ b/embassy-stm32/src/adc/adc4.rs
@@ -113,10 +113,26 @@ foreach_adc!(
             }
 
             fn enable() {
+                let cr_initial = ADC4::regs().cr().read();
+                let isr_initial = ADC4::regs().isr().read();
+
+                if cr_initial.aden() && isr_initial.adrdy() {
+                    return;
+                }
+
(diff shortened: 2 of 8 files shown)


Gerharddc:
How about a simpler approach doing this with an AtomicBool such as https://github.com/Gerharddc/embassy/commit/d7b1f33482501b476fa689d4854c4843da8c20e0 perhaps?

Gerharddc:
@cryptographix it seems https://github.com/embassy-rs/embassy/pull/5047 has now been merged as an alternative solution. does this work for your use case?

cryptographix:
> @cryptographix it seems #5047 has now been merged as an alternative solution. does this work for your use case?

Yes. Thank you! Simpler solution, love it.

PR #4925, opened 2025-11-20
[STM32] LCD

diondokter:
I mean, why not finish it almost 1.5 years later :P

### DIFF
--- a/embassy-stm32/src/lcd.rs
+++ b/embassy-stm32/src/lcd.rs
@@ -152,6 +152,7 @@ pub enum Drive {
 pub struct Lcd<'d, T: Instance> {
     _peri: PhantomData<&'d mut T>,
     duty: Duty,
+    ck_div: u32,
 }
 
 impl<'d, T: Instance> Lcd<'d, T> {
@@ -227,9 +228,11 @@ impl<'d, T: Instance> Lcd<'d, T> {
             }
         }
 
+        let ck_div = lcd_clk.0 / ((1 << ps) * (div + 16));
+
         trace!(
-            "lcd_clk: {}, fps: {}, ps: {}, div: {}",
-            lcd_clk, best_fps_match, ps, div
+            "lcd_clk: {}, fps: {}, ps: {}, div: {}, ck_div: {}",
+            lcd_clk, best_fps_match, ps, div, ck_div
         );
 
         if best_fps_match == u32::MAX || ps > 0xF {
@@ -271,6 +274,7 @@ impl<'d, T: Instance> Lcd<'d, T> {
         Self {
             _peri: PhantomData,
             duty: config.duty,
+            ck_div,
         }
     }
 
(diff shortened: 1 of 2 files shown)


embassy-stm32/build.rs
```diff
@@ -1352,16 +1352,32 @@ fn main() {
...
+
+                // LCD is special
+                if regs.kind == "lcd" {
+                    key.1 = pin.signal.trim_end_matches(char::is_numeric);
+
+                    // Some lcd pins have multiple lcd functions
+                    // Dedup so they don't get the trait implemented twice
+                    if !seen_lcd_pins.insert(pin.pin) {
+                        continue;
+                    }
```
diondokter: I should investigate this. There's probably a way to not see duplicates by filtering more.
I think the reason is that some segment pins can also function as com pins.

examples/stm32u0/.cargo/config.toml
```diff
@@ -1,6 +1,6 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 # replace stm32u083rctx with your chip as listed in `probe-rs chip list`
-runner = "probe-rs run --chip stm32u083rctx"
+runner = "probe-rs run --chip stm32u083rctx --catch-hardfault"
```
diondokter: Whoops, should revert this

examples/stm32u0/Cargo.toml
```diff
@@ -7,7 +7,7 @@ publish = false
 
 [dependencies]
 # Change stm32u083rc to your chip name, if necessary.
-embassy-stm32 = { version = "0.4.0", path = "../../embassy-stm32", features = [ "defmt", "time-driver-any", "stm32u083rc", "memory-x", "unstable-pac", "exti", "chrono"]  }
+embassy-stm32 = { version = "0.4.0", path = "../../embassy-stm32", features = [ "defmt", "time-driver-any", "stm32u083mc", "memory-x", "unstable-pac", "exti", "chrono"]  }
```
diondokter, line 10: I think the HIL, if it has the U0, is the DK board. That uses the MC, not the RC.

embassy-stm32/build.rs
diondokter: Ok, I don't think there's a good way to get rid of this.
Some pins have multiple SEG alternate functions. This works because the peripheral itself can multiplex them.
You'd think they have a nice general AF feature in their GPIO peripheral, but apparently they needed to duplicate it in the LCD peripheral for some reason...

diondokter:
Alright, I've found some L0's support LCD too and it looks like it's the same peripheral. Cool!

But for some reason the L0 HAL doesn't impl `RccPeripheral` for `LCD`...
Anybody know why?

Edit: Confirmed this info is missing in the metapack

diondokter:
Ok, that should be it :D
Support for all STM families with LCD. Can't test them though since I only have a U0-DK with an LCD screen.
But they should all work since it's all the same IP block for all.

Like anything with ST, this took more time than I estimated lol

PR #4927, opened 2025-11-20
stm32: Fixed ADC4 enable() for WBA

leftger:
The example for the STM32WBA65 device had a configuration issue. That has been addressed. At the same time there was an initialization issue with the ADC4 peripheral that caused the same example to hang on the WBA65 devices. The fix boils down to the following:
- If the device is enabled and ready just return
- If it's undergoing a conversion then stop the ADC4 peripheral
- Otherwise do the normal initialization procedure
- DON"T clear the ADRDY flag


leftger:
Thanks! 🚀

PR #4934, opened 2025-11-23
Expose timer queue impls unconditionally, support timer queue items stored outside of tasks (e.g. RTOS-aware `block_on`)

bugadani:
This PR is a backwards-compatible implementation that allows HAL implementers to mix queue implementations, which allows users to mix executors. This is mainly targeted at RTOS implementations which can use embassy-executor, and also provide their own `block_on` implementations simultaneously.

### DIFF
--- a/embassy-time-queue-utils/src/queue_integrated.rs
+++ b/embassy-time-queue-utils/src/queue_integrated.rs
@@ -53,7 +53,7 @@ impl Queue {
 
     /// Schedules a task to run at a specific time.
     ///
-    /// If this function returns `true`, the called should find the next expiration time and set
+    /// If this function returns `true`, the caller should find the next expiration time and set
     /// a new alarm for that time.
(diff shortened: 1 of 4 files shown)


embassy-time-queue-utils/src/queue_integrated.rs
```diff
@@ -55,12 +56,22 @@ impl Queue {
...
         };
+        self.schedule_wake_queue_item(at, item, waker)
+    }
+
+    /// Schedules a task to run at a specific time, using its integrated queue item.
+    ///
+    /// If this function returns `true`, the called should find the next expiration time and set
+    /// a new alarm for that time.
+    #[inline]
+    pub fn schedule_wake_queue_item(&mut self, at: u64, item: &mut TimerQueueItem, waker: &Waker) -> bool {
```
bugadani, line 76: Should this be `unsafe` and require that `item` comes from `waker`?

lulf (approved):
No objection from me :)

PR #4944, opened 2025-11-25
stm32: remove waveform method

xoviat:
it seems to be a duplicate of waveform_up

### DIFF
--- a/embassy-stm32/src/timer/low_level.rs
+++ b/embassy-stm32/src/timer/low_level.rs
@@ -13,9 +13,10 @@ use embassy_hal_internal::Peri;
 pub use stm32_metapac::timer::vals::{FilterValue, Mms as MasterMode, Sms as SlaveMode, Ts as TriggerSource};
 
 use super::*;
+use crate::dma::Transfer;
 use crate::pac::timer::vals;
+use crate::rcc;
 use crate::time::Hertz;
-use crate::{dma, rcc};
 
 /// Input capture mode.
 #[derive(Clone, Copy)]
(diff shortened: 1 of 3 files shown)


xoviat:
@KorribanMaster any thoughts on this?

embassy-stm32/src/timer/low_level.rs
```diff
@@ -663,25 +664,51 @@ impl<'d, T: GeneralInstance4Channel> Timer<'d, T> {
...
-        // optional TODO:
-        // clean FEIF after disable UDE
-        if !original_update_dma_state {
-            self.enable_update_dma(false);
+                    #[cfg(any(bdma, gpdma))]
+                    Transfer::new_write(
+                        dma,
+                        req,
+                        duty,
+                        self.regs_1ch().ccr(channel.index()).as_ptr() as *mut u32,
```
KorribanMaster, line 707: Use the regs_gp16() here otherwise it wont work for channels >1 i fixed this in my pr already

xoviat:
@KorribanMaster can you submit a new pr with the fix?

KorribanMaster:
Sure

quarthex:
So, is there a way to generate a waveform with a `dma` which implements `Dma<T>` like before, or was it an error?

For instance, on an STM32F042G6, can I generate a waveform on `TIM3_CH1`?

See _§10.3.2 DMA request mapping_, [Reference manual](https://www.st.com/resource/en/reference_manual/rm0091-stm32f0x1stm32f0x2stm32f0x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)

xoviat:
That doesn't have dma up?

xoviat:
Looks like it does

quarthex:
> That doesn't have dma up?

`TIM3_UP` exists, on `DMA1_CH3`. 
Unfortunately, in my use case, `DMA1_CH2` and `DMA1_CH3` are already used by `I2C1`.
That’s why I was trying to find a GPIO pin drivable by a PWM on one of the other DMA channels (1, 4 or 5).

image

quarthex:
brought back by PR #4980 (for cross-reference)

PR #4947, opened 2025-11-25
stm32: allow granular stop for uart

xoviat:
allows granular stop without having to drop usart, and allows path to implementation for other peripherals.

### DIFF
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -8,8 +8,9 @@ use embassy_sync::waitqueue::AtomicWaker;
 
 use super::ringbuffer::{DmaCtrl, Error, ReadableDmaRingBuffer, WritableDmaRingBuffer};
 use super::word::{Word, WordSize};
-use super::{AnyChannel, BusyChannel, Channel, Dir, Request, STATE};
+use super::{AnyChannel, Channel, Dir, Request, STATE};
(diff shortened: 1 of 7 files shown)


liebman:
Is this preferred over `crate::low_power::DeviceBusy`?

xoviat:
DeviceBusy is a user-facing API. This is an internal API that is used because it does not have to store the stop enum. Before this PR, any uart struct held would block stop. After this PR, the uart struct itself doesn't block stop, only use of the write and read methods.

This is implemented in such a way so that the current behavior is maintained (holding a struct will block stop), but each peripheral can be converted to this over time. Generally, this method is preferred internally because the stop mode doesn't need to be stored.

liebman:
You may want to hit up i2c/v2.rs where I used DeviceBusy for `stm32wlex`

xoviat:
I already removed that. I assume that was there because of dma, yes?

liebman:
Yes, if there is an active transaction we can't STOP as DMA is lost.

xoviat:
Yes, so I refcounted the DMA is well. The reasons not to stop is because the peripheral is in use and the DMA is in use. I'm still playing around with the API here, but in the final version, with each call to write the refcount will increase by two when write is called (one for the peripheral, one for the dma), unless the peripheral is lp and unless the dma is lp. If both the peripheral and the dma are lp, then stop will happen.

xoviat:
See #4931.

PR #4950, opened 2025-11-25
Implement AON Timer driver for the RP2350

Ragarnoy:
I recently needed an RTC, thought that the RP2350 didn't have one, but it actually does and has something that's kind of better (depending on your use case). This is my shot at implementing it.

Changes

AON Timer:
  - 64-bit millisecond counter (1 kHz resolution)
  - Configurable clock sources (XOSC/LPOSC)
  - Alarm wake modes: WFI/WFE, DORMANT, or both
  - Async alarm waiting with interrupts

DateTime module:
  - Moved datetime types from RTC to shared embassy-rp/src/datetime/ module
  - Works with both RTC (RP2040) and AON Timer (RP2350)
  - Supports chrono or standalone implementation
  - Added Unix epoch conversion (ms since 1970-01-01)

AON Timer DateTime API:
  - set_datetime(), now_as_datetime(), set_alarm_at_datetime()

I did not implement the time driver because the 1ms resolution is too low for a main timer. This works better as a simple "global" clock for waking from dormant (feel free to change my mind on this i'm not 100% convinced)

I also was too lazy to have an example that wakes from dormant state. I'm not sure there is a way to enable dormant state through embassy right now

(tested on actual RP2350 hardware)

### DIFF
--- a/embassy-rp/src/aon_timer/mod.rs
+++ b/embassy-rp/src/aon_timer/mod.rs
@@ -685,13 +685,7 @@ impl<'d> AonTimer<'d> {
         poll_fn(|cx| {
             WAKER.register(cx.waker());
 
-            if critical_section::with(|_| {
-                let occurred = ALARM_OCCURRED.load(Ordering::SeqCst);
-                if occurred {
-                    ALARM_OCCURRED.store(false, Ordering::SeqCst);
-                }
-                occurred
-            }) {
+            if ALARM_OCCURRED.swap(false, Ordering::SeqCst) {
                 self.clear_alarm();
 
                 compiler_fence(Ordering::SeqCst);


embassy-rp/src/datetime/datetime_no_deps.rs
```diff
@@ -0,0 +1,78 @@
...
+    Monday = 1,
+    Tuesday = 2,
+    Wednesday = 3,
+    Thursday = 4,
+    Friday = 5,
+    Saturday = 6,
+}
+
+impl DateTime {
+    /// Convert to Unix timestamp (milliseconds since 1970-01-01 00:00:00 UTC)
```
9names, line 60: If it's in milliseconds it's not really a Unix timestamp, is it? This is more like a javascript timestamp.

embassy-rp/src/datetime/datetime_no_deps.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// The [DateTime] contains an invalid year value. Must be between `0..=4095`.
+    InvalidYear,
+    /// The [DateTime] contains an invalid month value. Must be between `1..=12`.
+    InvalidMonth,
+    /// The [DateTime] contains an invalid day value. Must be between `1..=31`.
+    InvalidDay,
+    /// The [DateTime] contains an invalid day of week. Must be between `0..=6` where 0 is Sunday.
+    InvalidDayOfWeek(
+        /// The value of the DayOfWeek that was given.
+        u8,
```
9names: Why does this error return the value when the others don't?

embassy-rp/src/datetime/datetime_no_deps.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Structure containing date and time information
+#[derive(Clone, Debug)]
+pub struct DateTime {
+    /// 0..4095
+    pub year: u16,
+    /// 1..12, 1 is January
+    pub month: u8,
+    /// 1..28,29,30,31 depending on month
+    pub day: u8,
+    ///
```
9names: Weird to have a blank comment on DayOfWeek when the others are all populated

embassy-rp/src/aon_timer/mod.rs
```diff
@@ -0,0 +1,730 @@
...
+
+use embassy_hal_internal::Peri;
+use embassy_hal_internal::interrupt::InterruptExt;
+use embassy_sync::waitqueue::AtomicWaker;
+use embassy_time::Duration;
+
+pub use crate::datetime::{DateTime, DayOfWeek, Error as DateTimeError};
+use crate::{interrupt, pac};
+
+const POWMAN_PASSWORD: u32 = 0x5AFE << 16;
```
9names, line 184: Why did you write it this way, do you find this easier to read than 0x5afe0000?

embassy-rp/src/aon_timer/mod.rs
```diff
@@ -0,0 +1,730 @@
...
+    /// Controls which low-power wake mechanisms are enabled for alarms.
+    /// See [`AlarmWakeMode`] for details on each mode.
+    pub alarm_wake_mode: AlarmWakeMode,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            clock_source: ClockSource::Xosc,
+            clock_freq_khz: 12000, // 12 MHz XOSC
```
9names, line 243: Do you think this is going to be the most common configuration?

I would have expected most users would want to use this with LPOSC so that they can use it to wake from SLEEP or DORMANT state.

embassy-rp/src/datetime/epoch.rs
```diff
@@ -0,0 +1,146 @@
...
+
+fn days_from_epoch_to_year(year: u16) -> u64 {
+    let mut days = 0u64;
+    for y in EPOCH_YEAR..year {
+        days += if is_leap_year(y) { 366 } else { 365 };
+    }
+    days
+}
+
+fn day_of_week_from_days(days_since_epoch: u64) -> u8 {
```
9names: Any particular reason you think folks might want to pass in more than u32::MAX value days since epoch?

embassy-rp/src/aon_timer/mod.rs
```diff
@@ -0,0 +1,730 @@
+//! AON (Always-On) Timer driver for RP2350
+//!
+//! The AON Timer is a 64-bit counter that runs at 1 kHz (1ms resolution) and can operate during
```
9names: The tick rate is configurable, see `12.10.8. Using a tick faster than 1ms` in the datasheet
```suggestion
//! The AON Timer is a 64-bit counter that usually runs at 1 kHz (1ms resolution) and can operate during
```

embassy-rp/src/datetime/epoch.rs
```diff
@@ -0,0 +1,146 @@
...
+    millis += dt.hour as u64 * MS_PER_HOUR;
+    millis += dt.minute as u64 * MS_PER_MINUTE;
+    millis += dt.second as u64 * MS_PER_SECOND;
+
+    Ok(millis)
+}
+
+#[cfg(not(feature = "chrono"))]
+pub(crate) fn millis_to_datetime(millis: u64) -> Result<DateTime, Error> {
+    let total_days = millis / MS_PER_DAY;
```
9names: total days doesn't have to be u64 since you'll return Error further down if the year is > 4095.
and your `remaining_ms` is going to be less than 86400000 after the modulo, it could also drop back to u32.
maybe add an assert for `total_days <= u32::max` and then `let total_days = total_days as u32;` so you don't have to do so much 64bit math?

embassy-rp/src/datetime/epoch.rs
```diff
@@ -0,0 +1,146 @@
...
+
+fn days_in_month(year: u16, month: u8) -> u8 {
+    if month == 2 && is_leap_year(year) {
+        29
+    } else {
+        DAYS_IN_MONTH[(month - 1) as usize]
+    }
+}
+
+fn days_from_epoch_to_year(year: u16) -> u64 {
```
9names: u64 for number of days when you're only giving u16 years?

embassy-rp/src/aon_timer/mod.rs
Ragarnoy: I think it's clearer since spec says it's the 16 upper bits

embassy-rp/src/datetime/datetime_no_deps.rs
Ragarnoy: True, I clarified it (milliseconds since epoch)

embassy-rp/src/datetime/datetime_no_deps.rs
Ragarnoy: True, I removed the u8 because for years this would induce an u16 and it's really not worth it

embassy-rp/src/datetime/datetime_no_deps.rs
Ragarnoy: True! No idea how that happened

embassy-rp/src/aon_timer/mod.rs
Ragarnoy: I'm not totally sure how I feel about it. Using LPOSC as default might break things for people who need higher resolution and didn't bother choosing a non-default clock.                                                        
                                                                                                                                                                                                                                   Plus, DORMANT mode isn't implemented in Embassy yet, so anyone wanting to use it would have to look into how to do it first anyway - at which point they'd see the LPOSC requirement.                                            
                                                                                                                                                                                                                                   
Maybe this can be up for debate with the rest of the community, but imo it's probably better to leave XOSC for now.

embassy-rp/src/datetime/epoch.rs
Ragarnoy: True, idk why I did that

embassy-rp/src/aon_timer/mod.rs
Ragarnoy: fixed, added typically

embassy-rp/src/datetime/epoch.rs
Ragarnoy: fixed!

embassy-rp/src/datetime/epoch.rs
Ragarnoy: fixed as well

embassy-rp/src/aon_timer/mod.rs
```diff
@@ -0,0 +1,730 @@
...
+        poll_fn(|cx| {
+            WAKER.register(cx.waker());
+
+            if critical_section::with(|_| {
+                let occurred = ALARM_OCCURRED.load(Ordering::SeqCst);
+                if occurred {
+                    ALARM_OCCURRED.store(false, Ordering::SeqCst);
+                }
+                occurred
+            }) {
```
9names: This looks like you're trying to manually implement `ALARM_OCCURRED.swap(false, Ordering::SeqCst)`.
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.swap

embassy-rp/src/aon_timer/mod.rs
Ragarnoy: Then i suppose we keep XOSC for now

embassy-rp/src/aon_timer/mod.rs
Ragarnoy: Fixed!

9names (approved):
Looks good to me

PR #4951, opened 2025-11-26
Add custom idle-hook feature to thread mode executor

MichaelHopfengaertner:
### Scope
This PR adds a backwards-compatible feature, that allows the registration of a custom idle-hook function for the thread mode executor.

By enabling the `idle-hook` feature for `embassy-executor` an additional executor method is provided to the application:
```rust
#[cfg(feature = "idle-hook")]
pub fn with_idle_hook(mut self, idle_hook: fn(&Executor)) -> Self
```

This feature is also added to `embassy-executor-macros` for convenient custom idle-hook registration:
```rust
#[embassy_executor::main(idle_hook = "my_idle_hook")]
```

The feature is supported on the following architectures: `arch-cortex-m`, `arch-cortex-ar`, `arch-riscv32`, `arch-avr`, `arch-std`.

The executor's default idle implementation is provided to the application `Executor::default_idle(&self)`, e.g. if the custom idle-hook wants/needs to use it.

### Goal
Previously, the idle behavior for each `Executor` implementation was pre-defined, e.g. `WFE`/`WFI`/`SLEEP` instruction.
If an application or chip requires a different/custom idle implementation or pre-idle/post-idle steps (as in [embassy-stm32/src/low_power.rs](https://github.com/embassy-rs/embassy/blob/45da8625392d59da59b607b6fd510f7610ee58a1/embassy-stm32/src/low_power.rs#L340-L354)), a custom `Executor` has to be implemented.

The **goal** is to enable developers to **use** the **existing executor implementations** **while** being able to **customize** the **idle** **behavior** very **easily** (if necessary).
This **reduces** the **overhead** and **duplication** of implementing a custom `Executor`.

#### Example embassy-nrf
```rust
use embassy_executor::{Executor, Spawner};
use embassy_time::Timer;

fn my_custom_idle_hook(executor: &Executor) {
    // TODO: Optionally, execute custom pre-idle steps, e.g. power-management, stop mode configuration, etc.

    // Put chip into idle state, e.g. by using the exposed executor's default idle implementation
    executor.default_idle();

    // TODO: Optionally, execute custom post-idle steps, e.g. clock re-initialization, etc.
}

#[embassy_executor::main(idle_hook = "my_custom_idle_hook")]
    let _p = embassy_nrf::init(embassy_nrf::config::Config::default());

    loop {
        Timer::after_millis(1000).await;
    }
}
```
#### Example: embassy-nrf + nrf-softdevice
When using `nrf-softdevice` the arch-cortex-m executor's default idle implementation is not able to put the system into a low power mode.

```rust
use embassy_executor::{Executor, Spawner};
use embassy_nrf::interrupt::Priority;
use embassy_time::Timer;
use nrf_softdevice::{raw, Softdevice};

fn my_custom_idle_hook(_executor: &Executor) {
    // ~3uA consumption (when using nrf-softdevice's idle implementation)
    unsafe { raw::sd_app_evt_wait() };

    // ~830uA consumption (when using executor's default idle implementation)
    // _executor.default_idle();
}

#[embassy_executor::main(idle_hook = "my_custom_idle_hook")]
    let mut config = embassy_nrf::config::Config::default();
    config.time_interrupt_priority = Priority::P2;
    let _p = embassy_nrf::init(config);
    let _sd = Softdevice::enable(&nrf_softdevice::Config::default());

    loop {
        Timer::after_millis(1000).await;
    }
}
```


ivmarkov:
Actually, I would need to use this with... `arch-std`. Is there any reason why this can't be supported with that platform?

Christopher-06:
That would be a cool and practical addition.
It would also make it much easier to measure sleep time for estimating CPU usage. So far, you have to copy the existing executor and just save a timer value briefly before going to sleep and after waking up. With this PR, this would also be possible directly with the existing executor!

MichaelHopfengaertner:
> Actually, I would need to use this with... arch-std. Is there any reason why this can't be supported with that platform?

Yes, we could also add the feature to `arch-std`. I wanted to keep it simple in the first place by just exposing `Executor::default_idle()` as function, without passing a reference to the executor instance around (which would be required to access the `arch-std` signaler within the `default_idle` implementation).

But I think we could easily change this from 

```rust
fn my_custom_idle_hook() {
    Executor::default_idle();
}
```

to something like this

```rust
fn my_custom_idle_hook(executor: &Executor) {
    executor.default_idle();
}
```

That would probably even be cleaner and more flexible.

I will change it.

MichaelHopfengaertner:
Added `idle-hook` feature for `arch-std` and updated PR description.

Dirbaio:
I'm not convinced we should do this.

- It uses a function pointer, which adds a slight overhead.
- There's a lot of overlap with the executo "platform" concept (previously named "arch", was just renamed). The reason the raw executor API exists is to let you implement your own "platform" outside embassy-executor, allowing customizing the main loop. Also, if you need to customize the sleep you often also need to customize the pender. For example, the low-power platform we now have in `embassy-stm32` needs it. A custom idle hook doesn't allow this, a custom platform does.

MichaelHopfengaertner:
> It uses a function pointer, which adds a slight overhead.

Yep, it should be around 5 cycles on Cortex-M.

> The reason the raw executor API exists is to let you implement your own "platform" outside embassy-executor, allowing customizing the main loop.

I agree.
However, I felt that implementing a custom executor was a significant amount of code duplication.
I'm not sure why the pender had to be changed for the `embassy-stm32` low-power platform. Couldn't the standard `waker` (and `__pender`) concept be used here as well?
Then the low-power platform could also be based on the standard Cortex-M (arch) executor with idle hook feature, with an additional abstraction layer containing the platform specifics (low-power setup, time-driver resume, etc.).

In general, the idea was to minimize code duplication when starting a new project and only adapt the pre/post sleep steps to the respective platform.

xoviat:
> I'm not sure why the pender had to be changed for the embassy-stm32 low-power platform. Couldn't the standard waker (and __pender) concept be used here as well?

Because the executor would try to enter sleep when there were tasks pending. There's no way to query ER on cortex-m directly, so an atomicbool was required.

MichaelHopfengaertnerTado:
> Because the executor would try to enter sleep when there were tasks pending. There's no way to query ER on cortex-m directly, so an atomicbool was required.

I see.

This is probably another unrelated discussion:
```rust
        // ...
        #[cfg(feature = "low-power-pender")]
        if TASKS_PENDING.load(Ordering::Acquire) {
            TASKS_PENDING.store(false, Ordering::Release);

            return;
        }

        compiler_fence(Ordering::Acquire);

        critical_section::with(|cs| {
        // ...
```
Inside [low_power.rs::configure_pwr()](https://github.com/embassy-rs/embassy/blob/cc185218671dd13ce0dfaec45f14792812e9fdce/embassy-stm32/src/low_power.rs#L391-L400) it seems for me the `TASKS_PENDING` acts as an **optimization** to avoid the "expensive" power-down sequence, when the task pender IRQ happens while polling other threads (`self.inner.poll()`).
However, there is probably still a race in the instructions in between checking `TASKS_PENDING` and entering the critical section afterwards.

Practically, this shouldn't be an issue (since WFE should not go asleep afterwards), just wanted to mention it.

However, we could also slightly change my proposal: Adding a reference to the `AtomicBool` flag inside the `Executor` instance. Inside the custom idle hook we could then access and interact with this flag (e.g. like `configure_pwr()` is doing it).

Dirbaio:
I'm going to close this for the reasons in my above comment. 

The executor already has an "extension point" that can do everything (custom platform). Having two extension points complicates it more, it's not obvious which one you should use, and you still need a custom platform in some cases (like stm32). Making the idle hook work for stm32 would complicate it even more and expose implementation details (the AtomicBool).

PR #4953, opened 2025-11-26
feat: support nrf54 GRTC as time-driver

lulf:
* Refactor GRTC peripheral splitting it into multiple channels
* Reserve channel 1 for time-driver if enabled
* Implement time-driver using GRTC (RTC peripheral is now removed).
* Add timer example to nrf54l15
* Update saadc after API changes

### DIFF
--- a/embassy-net-nrf91/Cargo.toml
+++ b/embassy-net-nrf91/Cargo.toml
@@ -18,7 +18,7 @@ log = ["dep:log"]
 defmt = { version = "1.0.1", optional = true }
 log = { version = "0.4.14", optional = true }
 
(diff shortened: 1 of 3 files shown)


lulf:
I need to investigate a bit what happened to the CLOCK_POWER interrupt.

Dirbaio:
> what happened to the CLOCK_POWER interrupt.

yeah I saw that too. I think it's an SVD mistake.

lulf:
Filed https://github.com/NordicSemiconductor/nrfx/issues/142 - adding it back in the mean time. All their docs, at least regarding SDC for nRF54 seems to indicate it should be there.

lulf:
Merging this now, hopefully the SVDs get fixed and we can add l05,l10 etc.

PR #4955, opened 2025-11-26
i2c slave fix (early termination)

beeelias:
this fixes a bug where we have a bug where the i2c write is shorter than our buffer, causing the transmission to never terminate properly. Which causes all consecutive writes/reads to fail

```
    let mut buf = [0u8; 3];
    loop {
        match dev.listen().await {
            Ok(SlaveCommand {
                kind: SlaveCommandKind::Write,
                address: Address::SevenBit(DEV_ADDR),
            }) => {
                // 2. Respond to Write
                match dev.blocking_respond_to_write(&mut buf) {}
....
```


xoviat:
Changelog

PR #4957, opened 2025-11-27
Upgrade heapless to 0.9, embedded-io-async to 0.7 and defmt to 1.x

brainstorm:


Dirbaio:
See #4745 for embedded-io

We can't upgade heapless until smoltcp upgrades it.

About defmt: we're already using 1.0 everywhere, there's nothing to upgrade.

brainstorm:
> See #4745 for embedded-io
> 
> We can't upgade heapless until smoltcp upgrades it.
> 
> About defmt: we're already using 1.0 everywhere, there's nothing to upgrade.

Thanks for the feedback @Dirbaio! ... what's the blocker on heapless 0.9 for smol? I see it merged already in https://github.com/smoltcp-rs/smoltcp/pull/1083 ?

Dirbaio:
A release.

brainstorm:
> About defmt: we're already using 1.0 everywhere, there's nothing to upgrade.

If I understood correctly from the last comment in https://github.com/rust-embedded/embedded-hal/pull/716#issuecomment-3585062255 the removal of all `defmt-03` references is a good cleanup to do/change (upgrade) regardless? ... and if it's still breaking, would it make sense to semver-bump it accordingly or you want to keep `defmt-03` for a while longer?

> A [smol] release.

Pre-xmas communitly present? 🙏🏻 Let me know if I can help in any way with this :)

I'll now get acquainted with https://github.com/embassy-rs/cargo-embassy-devtool to compile all embassy-rs locally avoid CI failed builds spam (sorry about them!).

Dirbaio:
> the removal of all defmt-03 references is a good cleanup to do/change (upgrade) regardless?

Embassy crates don't define any `defmt-03` features, they just use them from dependencies. If a new version of a dep adds the `defmt` feature we should upgrade to it, of course.

Dirbaio:
https://github.com/embassy-rs/embassy/pull/5342

PR #4962, opened 2025-11-27
embassy-rp: Add PioClk [and mclk to I2S]

dobrowolski-lukasz:
Add a simple PIO program for outputting a square wave of given frequency.
Switch to pio_programs::clock_divider::calculate_pio_clock_divider_value for clock divider calculation for I2S PIO programs.
Change PioBatch to enable user to start two PIO based drivers at the same time.

PioBatch could be further extended so that the drivers could register callbacks to be executed after state machine is started. Current code does not need it so I haven't added this.

I'm not sure with some of names I've used. If anybody has better ideas do let me know.

Test: I've run pio_i2s and pio_clk examples and checked all outputs with an oscilloscope.

### DIFF
--- a/embassy-rp/CHANGELOG.md
+++ b/embassy-rp/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 - Add output enable inversion API (gpio, pio)
+- Add PIO clock generator
+- Change PioBatch interface
 
 ## 0.9.0 - 2025-11-27
 


dobrowolski-lukasz:
@lulf: Sorry to be pinging you but could You take a look?

examples/rp/src/bin/pio_i2s.rs
9names, line 1: I think it makes sense to have this as a new example, not to replace the old one, so that folks who want i2s without mclk can easily see how to do that (and I can easily test that you haven't broken that configuration).
You should also update the file header to reflect the new pinout.

examples/rp/src/bin/pio_clk.rs
```diff
@@ -0,0 +1,32 @@
+//! This example shows how to create a clock using the PIO module in the RP2040 chip.
```
9names:
```suggestion
//! This example shows how to output a clock signal on an output pin using the PIO module in the RP2040 chip.
```

embassy-rp/src/pio_programs/clk.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/// Pio backed clock generator
+pub struct PioClk<'d, T: Instance, const SM: usize> {
+    sm: StateMachine<'d, T, SM>,
+    pin: Pin<'d, T>,
+}
+
+impl<'d, T: Instance, const SM: usize> PioClk<'d, T, SM> {
+    /// Configure state mashine for clock generation
```
9names:
```suggestion
    /// Configure state machine for clock generation
```

embassy-rp/src/pio_programs/clk.rs
```diff
@@ -0,0 +1,78 @@
+//! Clock generator using PIO
```
9names:
```suggestion
//! Clock output signal generator using PIO
```

dobrowolski-lukasz:
> I think long-term we should probably have an PIO program that incorporates MCLK generation (to avoid requiring 2 SM's for I2S). (...)

I don't think that is feasible, at least not with configurable mclk/bclk ratio. PioI2sOut already uses X and Y scratch registers and I have an unpublished change that reuses ISR as another scratch(last possible one) so no way to parametrize that at PIO assembly level. Another option would be dynamic PIO assembly generation but AFAICT it would make the program at least 4 times as long and that would defeat the purpose of saving on an SM as there would be not much space left to load any program.

PIOs are very flexible but also very limited. :/

Thank You for reviewing!

PR #4966, opened 2025-11-30
embassy-mspm0:  add CPU accelerated float division function

bespsm:
- mspm0-mathacl: add division example
- mspm0-mathacl: add CPU accelerated float division function
- using internal Mathacl Qval type for providing length of fractional part

### DIFF
--- a/embassy-mspm0/CHANGELOG.md
+++ b/embassy-mspm0/CHANGELOG.md
@@ -21,3 +21,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat: Add read_reset_cause function
 - feat: Add module Mathacl & example for mspm0g3507 (#4897)
 - feat: Add MSPM0G5187 supportt
+- feat: add CPU accelerated division function (#4966)
--- a/embassy-mspm0/src/mathacl.rs
+++ b/embassy-mspm0/src/mathacl.rs
@@ -13,19 +13,23 @@ use micromath::F32Ext;
 use crate::Peri;
 use crate::pac::mathacl::{Mathacl as Regs, vals};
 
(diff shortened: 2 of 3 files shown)


bespsm:
@i509VCB  I added division function and used it it in SINCOS

i509VCB:
I'm not sure how I feel about doing all the calculations and then convert to I16F16 internally, which can't match extreme ends of real f32 math. The error tolerance makes me question this more. 

I do think we should probably make mathacl assume types from fixed point for calculations. TI implemented iqmathlib using the accelerator, so it is clearly intended for that.

It makes more sense that the user understands what is going on by making them explicitly convert to/from f32.

I'm not going to say no given you are doing the work here, but maybe I'll ask a bit on matrix because the cordic on stm32 is similar in some ways.

i509VCB:
From what I can tell, the MathACL does:
- i32
- u32
- u16f16
- i16f16

For DIV anything from i1f31 to i32 and vice versa for signed?

bespsm:
@i509VCB  I hope I understood your comment correctly. I added IQType, where the integer and the fractional parts are set explicitly, also added division for i32 and u32. The examples are adopted

bespsm:
@i509VCB do you have any more concerns?

bespsm:
@xoviat please check PR

xoviat:
I review stm32 only

i509VCB (approved):
I'm no expert with regards to this part of the hardware, but otherwise it looks fine.

Edit: Please squash if possible so I can merge.

bespsm:
> I'm no expert with regards to this part of the hardware, but otherwise it looks fine.
> 
> Edit: Please squash if possible so I can merge.

done

PR #4969, opened 2025-12-01
stm32: Add blocking_listen for blocking I2C driver

TrAyZeN:
Currently calling `listen` on blocking I2C drivers (created with `I2c::new_blocking`) blocks the program as the interrupt handlers are not binded. This PR restricts the `listen` implementation to async I2C drivers and adds `blocking_listen` implementation for blocking I2C drivers.

Here is a snippet of the previously mentioned issue:
```rust
#[embassy_executor::main]
    let mut i2c = I2c::new_blocking(p.I2C1, p.PB6, p.PB9, i2c::Config::default())
        .into_slave_multimaster(SlaveAddrConfig::basic(MCU2_I2C_ADDR));

    // The following call blocks the program
    let res = i2c.listen().await;

    debug!("Listen completed");
    loop {}
}
```


xoviat:
Is the concern about code size or that you can't use blocking without an executor?

TrAyZeN:
Sorry I am not sure to get your question. My issue is that `listen` is implemented for both blocking and async I2C drivers, but the implementation depends on the interrupt handlers being binded. It works fine for async I2C drivers that [requires these interrupt handlers to be binded](https://github.com/TrAyZeN/embassy/blob/52a9b08f0ca13d23bfb039c884a9101997c10567/embassy-stm32/src/i2c/mod.rs#L156-L158). However, [this is not ensured](https://github.com/TrAyZeN/embassy/blob/52a9b08f0ca13d23bfb039c884a9101997c10567/embassy-stm32/src/i2c/mod.rs#L176-L181) when creating a new blocking I2C driver. From what I understand, blocking drivers are not using interrupts, so I am proposing to restrict the `listen` implementation only to async I2C drivers and to add a `blocking_listen` methods for blocking drivers.

PR #4970, opened 2025-12-01
Feature SPI Bidirectional Mode

beeelias:
- Single-wire mode constructor: Spi::new_bidi accepts SCK plus one SDIO pin (wired to MOSI) and both TX/RX DMA channels. It configures SDIO as push‑pull and enables CR1.BIDIMODE,
    so the pin can alternate between driving and listening without reconfiguring GPIO.
  - Direction control: call spi.set_direction(Some(Direction::Transmit)) before sending data and spi.set_direction(Some(Direction::Receive)) before capturing data. Passing None
    disables bidirectional mode and reverts to normal MOSI/MISO if you need it.
  - Compatible APIs: all existing blocking and async write, read, transfer, and DMA helpers work unchanged. Writes pull from your TX slice, reads fill the provided RX slice, and
    transfer variants still perform simultaneous full‑duplex exchanges.
  - FIFO/word-size handling stays in the shared helpers, so all generations (v1–v6) automatically flush RX FIFOs, match data width, and respect errata quirks.
  - Usage sketch:

```rs
    let mut spi = Spi::new_bidi(p.SPI1, p.PA5, p.PA7, dma_tx, dma_rx, config);
    spi.set_direction(Some(Direction::Transmit));
    spi.blocking_write(&tx)?;
    spi.set_direction(Some(Direction::Receive));
    spi.blocking_read(&mut rx)?;
```

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -359,6 +359,7 @@ impl<'d, M: PeriMode, CM: CommunicationMode> Spi<'d, M, CM> {
     }
 
     /// Set SPI direction
+    #[cfg(any(spi_v1, spi_v2, spi_v3))]
     pub fn set_direction(&mut self, dir: Option<Direction>) {
         let (bidimode, bidioe) = match dir {
             Some(Direction::Transmit) => (vals::Bidimode::BIDIRECTIONAL, vals::Bidioe::TRANSMIT),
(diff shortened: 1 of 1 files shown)


xoviat:
Make sure this passed the build, then convert it to ready.

beeelias:
> Make sure this passed the build, then convert it to ready.

Interesting, updating the stm32-metapac removed some flags... making my code uncompilable

beeelias:
Oh never mind I am sorry, I am rel new to this project, so I am not that familiar with the versioning system.

PR #4972, opened 2025-12-02
feat: support OTA commands in esp-hosted driver

lulf:
* Expose OTA functionality in control
* Handle OTA swap wait in runner

### DIFF
--- a/embassy-net-esp-hosted/src/control.rs
+++ b/embassy-net-esp-hosted/src/control.rs
@@ -177,7 +177,6 @@ impl<'a> Control<'a> {
     pub async fn ota_end(&mut self) -> Result<(), Error> {
         let req = proto::CtrlMsg_Req_OTAEnd {};
         ioctl!(self, ReqOtaEnd, RespOtaEnd, req, resp);
-        // Ensures that run loop awaits reset
         self.shared.ota_done();
         // Wait for re-init
         self.init().await?;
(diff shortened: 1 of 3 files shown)


embassy-net-esp-hosted/src/control.rs
```diff
@@ -146,6 +151,13 @@ impl<'a> Control<'a> {
         Ok(())
     }
 
+    /// Initiate a firmware update.
+    pub async fn update(&mut self) -> Result<UpdateControl<'_, 'a>, Error> {
```
Dirbaio: this signature means you can't do any wifi stuff (connect, disconnect, etc) while you're doing an OTA. Is this required by the esp-hosted firmware? if not maybe it'd be better to do plain old methods in Control?

lulf:
bender run

embassy-net-esp-hosted/src/control.rs
lulf: Ah... good point. My original plan was to ensure a correct order of the OTA commands, and that END was not forgotten having to consume it in finish(). There's not a lot of docs on the requirements, but looks like as long as each command is not interrupted, it should be ok to interleave https://github.com/espressif/esp-hosted/blob/master/esp_hosted_fg/docs/Linux_based_host/ota_update.md 

I will change it to explicit methods then.

PR #4713, opened 2025-09-26
Fixes #4709. Creating an open AP after joining a WPA network doesn't …

nathansamson:
…work

This fixes the issue by always resetting the security value when creating a new open AP.

Issue #4709, opened 2025-09-25
cyw43 can not properly create a open AP after joining another AP


nathansamson:
Full code below but basically the code does this

* Creates an open AP
* Waits for 2 minutes so you have time to try to join
* Closes the AP
* Joins your pre configured network (change line 23-24 in my example code)
* Waits a minute
* Creates the same open AP again

What happens:
* When I join the AP with my Android phone during the first 2 minute period it connects waits a bit then gives an error (presumably because it will not get an IP as there is no DHCP server running). So perfectly normal

When I join I see these logs

~~~
10.106228 [DEBUG] === EVENT AUTH_IND: EventMessage { version: 2, flags: 0, event_type: 4, status: 0, reason: 0, auth_type: 0, datalen: 0, addr: [6a, 12, 7e, ba, c5, 86], ifname: [77, 6c, 30, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], ifidx: 0, bsscfgidx: 0, }  [] (cyw43 cyw43/src/runner.rs:492)
10.123288 [DEBUG] === EVENT ASSOC_IND: EventMessage { version: 2, flags: 0, event_type: 8, status: 0, reason: 0, auth_type: 0, datalen: 105, addr: [6a, 12, 7e, ba, c5, 86], ifname: [77, 6c, 30, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], ifidx: 0, bsscfgidx: 0, }  [00, 05, 63, 79, 77, 34, 33, 01, 08, 82, 84, 8b, 96, 24, 30, 48, 6c, 32, 04, 0c, 12, 18, 60, 3b, 16, 51, 51, 53, 54, 73, 74, 75, 76, 77, 78, 79, 7a, 7b, 7c, 7d, 7e, 7f, 80, 81, 83, 84, 82, 2d, 1a, af, 09, 17, ff, ff, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 01, 00, 7f, 01, 04, dd, 07, 00, 50, f2, 02, 00, 01, 00, dd, 08, 8c, fd, f0, 01, 01, 02, 01, 00, dd, 06, 50, 6f, 9a, 23, 01, 01] (cyw43 cyw43/src/runner.rs:492)
[...]

PR #4976, opened 2025-12-02
Always disable authentication before setting up AP

simonborje:
I ran into an issue similar to #4709 where I am unable to join the secure AP created by `control.start_ap_wpa2()` if I have previously tried to join a WPA2 network. The logs would show
```
[DEBUG] 181.513942: === EVENT PRUNE: EventMessage { version: 2, flags: 0, event_type: 23, status: 0, reason: 1, auth_type: 0, datalen: 0, addr: [6e, 7b, 85, 29, 17, 30], ifname: [77, 6c, 30, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], ifidx: 0, bsscfgidx: 0, }  []
```
I found that the issue went away if I extended the fix of resetting the security to "Open" from #4713 to also apply when creating secure APs.

### DIFF
--- a/cyw43/CHANGELOG.md
+++ b/cyw43/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Reset WPA security before creating secure AP
+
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #4980, opened 2025-12-03
timer: restore waveform method

xoviat:
requested for limited dma channels.


xoviat:
bender run

xoviat:
bender run

xoviat:
cc @quarthex

PR #4981, opened 2025-12-03
GPDMA suspend channel before reset if already enabled

### LINKED ISSUE #3900: STM32U5: uart.read_until_idle() every second transfer also returns the prevoius data + new data.
I make a repo to easy demostrate the issue: https://github.com/vDorst/uartdmabu

Setup: STM32 Development board (STM32H573I-DK) with STM32H573IIK3Q.
Connect USB cable to STLINK port. Used for debug (probe-rs) and VCP (virtual comport).
I used putty to connect VCP port with settings: 115200-8-n-1.
UART7 can also be used. In my project that was I first noticed this behaviour.

After reset, you see 'Hello!' in putty terminal.

If you slowly (1 key/sec) type `1` to `0` in putty you can observe this behavior

Putty ouput
```
Hello!

Got: 1
Got: 2
Got: 23
Got: 4
Got: 45
Got: 6
Got: 67
Got: 8
Got: 89
Got: 0
```

Probe-rs output:
```
# cargo run --release
   Compiling uartdmabug v0.1.0 (<snip>\\uartdmabug)
    Finished `release` profile [optimized + debuginfo] target(s) in 2.45s
     Running `probe-rs run --chip STM32H573IIKx --connect-under-reset target\thumbv8m.main-none-eabihf\release\uartdmabug`
      Erasing ✔ 100% [####################]  32.00 KiB @ 471.92 KiB/s (took 0s)
  Programming ✔ 100% [####################]  28.00 KiB @  82.02 KiB/s (took 0s)                                                                                                                                                                                        
     Finished in 0.41s
Start Init
0.000091 ERROR UART7: Zero size!
└─ uartdmabug::__mb_link_task_task::{async_fn#0} @ src\main.rs:43
0.000091 ERROR UART1: Zero size!
└─ uartdmabug::__mb_link_task_task::{async_fn#0} @ src\main.rs:43
UART1: Got: n_bytes 1 data: [31]
UART1: Got: n_bytes 1 data: [32]
UART1: Got: n_bytes 2 data: [32, 33]
UART1: Got: n_bytes 1 data: [34]
UART1: Got: n_bytes 2 data: [34, 35]
UART1: Got: n_bytes 1 data: [36]
UART1: Got: n_bytes 2 data: [36, 37]
UART1: Got: n_bytes 1 data: [38]
UART1: Got: n_bytes 2 data: [38, 39]
UART1: Got: n_bytes 1 data: [30]
```

It is not only a single character but also has the same behavior with larger data.

If I parse `ACBD`, `EFGH`, `IJLK` as seperet groups into putty I got this output

Putty ouput
```
Got: ACBD
Got: ACBDEFGH
Got: IJLK
```

Probe-rs output:
```
UART1: Got: n_bytes 4 data: [41, 43, 42, 44]
UART1: Got: n_bytes 8 data: [41, 43, 42, 44, 45, 46, 47, 48]
UART1: Got: n_bytes 4 data: [49, 4a, 4c, 4b]
```

I also tested this code on stm32u0 (arm thumbv6) and work fine.
But stm32u0 has a different DMA block.
So I think there is something wrong with the DMA setup.

I also noticed that:
- the first `read_until_idle()` always return Ok(0).
- First two transfers seems to be fine.

@Dirbaio you seems the initial author of the stm32u5 dma code. 
Any idea?

### LINKED ISSUE #4452: UART DMA Issue on STM32U5 Series
I recently noticed some odd behavior coming from the Async UART interface using a STM32U5-series chip (specifically the STM32U575ZIT6Q). If the DMA transfer is interrupted (for example, when using idle line detection as in `read_until_idle`), then the data from the peripheral becomes corrupted in often unpredictable ways.

My main concern here is that this issue may indicate a larger issue with the DMA implementation not limited to the UART peripheral which could cause state/memory corruption issues. However, I have not had the time to investigate whether this issue is reproducible on any other kind of peripheral.

I catalogue this behavior and provide a minimum reproducible example in [this repository](https://github.com/gmarino2048/embassy-stm32u5-uart-dma-error), so feel free to take a look and ask any questions. I'll try to answer them as well as I can. To repeat it here: multiple reads to the same buffer will introduce leading zeroes equal to the incoming message length. For example, the first call to `read_until_idle` yields `Read Amount: 2, Data: [0xCA, 0xFE]` while the second yields `Read Amount: 4, Data: [0x00, 0x00, 0xCA, 0xFE]` despite the same values `[0xCA, 0xFE]` being sent both times. If the memory address changes at all, the behavior gets even more unpredictable.

To simulate both `read_until_idle` and changing addresses, I provide the `read-until-idle` and `change-destination` features in the sample project. The default featureset just proves that the peripheral reads and writes work properly when using the standard `read` function.

This seems heavily related to https://github.com/embassy-rs/embassy/issues/3900, except the leading values are zeroes instead of previous data.

usbalbin:
I think this

Fixes #3900
Fixes #4452

[This](https://github.com/embassy-rs/embassy/issues/3900#issuecomment-3607440027) tested on a custom H523 board

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased - ReleaseDate
 
+- fix: stm32: GPDMA driver reset ignored during channel configuration
 - fix: stm32: SPI driver SSOE and SSM manegment, add `nss_output_disable` to SPI Config
 - change: stm32: use typelevel timer type to allow dma for 32 bit timers
 - fix: fix incorrect handling of split interrupts in timer driver


xoviat:
cc @fan2nd

xoviat:
bender run

fan2nd:
It‘s ok on u545 too. Good!

PR #4992, opened 2025-12-05
MCXA CRC

felipebalbi:


embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,756 @@
...
+            self.feed_word(*w);
+        }
+
+        for b in suffix {
+            self.feed_byte(*b);
+        }
+
+        // read back result.
+        self.read_crc()
+    }
```
felipebalbi: I guess I could make a generic version of this and share it with both 16- and 32 bits.

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,756 @@
...
+        // if transposition is enabled, result sits in the upper 16 bits
+        if ctrl.totr().is_byts_trnps() || ctrl.totr().is_byts_bts_trnps() {
+            (Self::regs().data32().read().bits() >> 16) as u16
+        } else {
+            Self::regs().data16().read().bits()
+        }
+    }
+
+    /// Feeds a slice of bytes into the CRC peripheral. Returns the computed checksum.
+    pub fn feed_bytes(&mut self, bytes: &[u8]) -> u16 {
```
jamesmunns: Might be worth calling "feed" instead of "feed bytes", and mention you do the fancy striding logic already so people aren't tempted to do it on their own too.

felipebalbi: done

embassy-mcxa/src/crc.rs
felipebalbi: okay, implemented an option for this. Let me know what you think.

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,806 @@
...
+        // Reference manual states:
+        //
+        // "After writing all the data, you must wait for at least two
+        // clock cycles to read the data from CRC Data (DATA)
+        // register."
+        cortex_m::asm::delay(2);
+
+        let ctrl = Self::regs().ctrl().read();
+
+        if W::is_16bit() {
```
jamesmunns: Misc note: instead of `is_Nbit() -> bool`, you could have a `write(Self)` (for `feed_word`) and `read() -> Self` method (for read_crc), and have it be monomorphized instead of branching at runtime.

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    /// Instantiates a new CRC peripheral for the `Xmodem` algorithm.
+    pub fn new_xmodem(peri: Peri<'d, CRC0>) -> Self {
+        Self::new_algorithm16(peri, Algorithm16::Xmodem)
+    }
+
+    /// Feeds a slice of bytes into the CRC peripheral. Returns the computed checksum.
+    ///
+    /// The input is split using `align_to::<u32>` into:
```
jamesmunns: nit: you can just say "data is strided efficiently, e.g. doing as many 32-bit writes as possible".

examples/mcxa/src/bin/crc.rs
```diff
@@ -0,0 +1,91 @@
+#![no_std]
+#![no_main]
+
+use embassy_executor::Spawner;
+use hal::config::Config;
+use hal::crc::Crc;
+use {defmt_rtt as _, embassy_mcxa as hal, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
```
jamesmunns, line 32: nit: you could maybe pull in the `crc` crate as a dep here and validate that the software and hw impls agree with each other as a smoke test?

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,806 @@
...
+        Self::regs().data32().write(|w| unsafe { w.bits(config.seed) });
+        Self::regs().ctrl().modify(|_, w| w.was().data());
+    }
+
+    fn regs() -> &'static crate::pac::crc0::RegisterBlock {
+        unsafe { &*crate::pac::Crc0::ptr() }
+    }
+
+    /// Read the computed CRC value
+    fn read_crc<W: Word>(&mut self) -> W {
```
jamesmunns: pedantic nit: if you wanted to make this a bit harder to ever hold wrong, you could have a "Transaction" API that takes a config to init the CRC, allow you to `feed` as much as you want, then have a `finalize` method that consumes the transaction and reads back the CRC, sort of like the `crc` crate does: https://docs.rs/crc/latest/crc/

This would prevent folks forgetting to (re-init).

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,806 @@
...
+    fn new_inner(_peri: Peri<'d, CRC0>) -> Self {
+        _ = unsafe { enable_and_reset::<CRC0>(&NoConfig) };
+
+        Crc {
+            _peri,
+            _phantom: PhantomData,
+        }
+    }
+
+    // Configure the underlying peripheral. `f` is expected to set the
```
jamesmunns: Why do we use a closure for bit width but use plain config for the rest of the items? It feels like you could just pass in the appropriate variant instead of a closure.

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,806 @@
+//! Cyclic Redundandy Check (CRC)
```
jamesmunns:
```suggestion
//! Cyclic Redundancy Check (CRC)
```

Colemak!!!

embassy-mcxa/src/crc.rs
felipebalbi: done.

embassy-mcxa/src/crc.rs
felipebalbi: I like that, let me give it a go.

embassy-mcxa/src/crc.rs
felipebalbi: done

embassy-mcxa/src/crc.rs
felipebalbi: done

examples/mcxa/src/bin/crc.rs
felipebalbi: done

embassy-mcxa/src/crc.rs
felipebalbi: done

embassy-mcxa/src/crc.rs
felipebalbi: done

embassy-mcxa/src/crc.rs
jamesmunns: Code comment needs to be updated

embassy-mcxa/src/crc.rs
```diff
@@ -0,0 +1,758 @@
...
+        }
+
+        impl Word for $t {}
+    };
+}
+
+impl_word!(
+    u8,
+    8,
+    |regs: &'static crate::pac::crc0::RegisterBlock, word| {
```
jamesmunns: you could probably make these free functions instead of closures, it might simplify the macro a bit, but the optimizer will probably inline it anyway.

embassy-mcxa/src/crc.rs
felipebalbi: done

embassy-mcxa/src/crc.rs
felipebalbi: done (I think that's what you meant)

PR #4995, opened 2025-12-06
Add nrf54lm20 support

HaoboGu:
Add blinky, gpio, gpiote and timer examples


lulf:
This PR seems to contain some other commits, could you  squash and rebase?

HaoboGu:
Does "squash and merge" work?

HaoboGu:
Squashed the commits

examples/nrf54lm20/run.sh
```diff
@@ -0,0 +1,15 @@
...
+Device nrf54lm20a_m33
+SelectInterface SWD
+Speed 4000
+LoadFile ${ELF_FILE}
+r
+g
+q
+EOF
+
+defmt-print -e $1 tcp
```
lulf: I think this file should probably not be committed.

HaoboGu: probe-rs doesn't have nRF54LM20A support officially, so I use this jlink script to flash and monitor the defmt-rtt output.

HaoboGu: Oh I just found [this PR](https://github.com/probe-rs/probe-rs/pull/3656), so the git version of probe-rs should work, I'll test it asap and remove the jlink script if probe-rs works

HaoboGu: yeah the latest version works! I've removed the jlink script

PR #5001, opened 2025-12-07
Use multiple IP addresses in static configurations

ekatwired:
Not as much in IPv4 networks, but IPv6 interfaces often, I dare to say except edge cases, always have multiple addresses assigned, starting with link-local address, which is used then to gain IP address using SLAAC and (i did not research this one yet) for DHCPv6 functioning. Also it is very common to have ULA and GUA prefices available for use.

Neither SLAAC or DHCPv6 is implemented in neither smalltcp or embassy network stack, therefore I myself wanted to implement SLAAC using ICMPv6 manually, but lack of exposed interface for having multiple addresses makes it impossible.

Commit 5aaa8078f6a15b4750c4d2507e094699d5140ce3 just changes single prefix entry for a vector. There are currently no overflow handling except throwing a panic and as we could have multiple addresses this may not be the best API. Also it is currently IPv6-only.

One of options could be just having a shared pool of assigned addresses and for dynamic configurations just merge with these? Would probably need some feedback for that :D

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -39,6 +39,7 @@ use embassy_time::{Instant, Timer};
 use heapless::Vec;
 #[cfg(feature = "dns")]
 pub use smoltcp::config::DNS_MAX_SERVER_COUNT;
+pub use smoltcp::config::IFACE_MAX_ADDR_COUNT;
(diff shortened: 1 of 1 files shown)


Dirbaio:

PR #5002, opened 2025-12-07
[stm32] use enable_and_reset_without_stop for select peripherals + block_stop

hjeldin:
Start peripherals with enable_and_reset_without_stop to prevent call to `increment_stop_refcount`
Use block_stop to increment refcount in the method scope and prevent stop1/2 when an async method is being awaited
Implemented on:
- [x] adc v3
- [x] i2c v1/v2
- [x] qspi
- [x] spi
- [x] sdmmc
- [x] simple_pwm

Only select async methods use block_stop, hence the draft.


hjeldin:
Superseded by #5008

PR #5004, opened 2025-12-08
wba: fix no_mangle attrs and enable mac bindings

leftger:
- Added an implementation of the UTIL sequencer (`wba/util_seq.rs`) that mirrors ST’s UTIL_SEQ API.
- Reworked `wba/mac_sys_if.rs` and `wba/ll_sys_if.rs` so the Mac and Link Layer register with the new sequencer
- Enabled the `stm32-bindings/wba_wpan_mac` feature in `Cargo.toml` within the `wba_mac` flag so that the `mac` bindings are present whenever the MAC stack is built.
- Tidied up the module tree by re-exporting `util_seq` from `wba/mod.rs`, letting other WBA code import the sequencer without digging into filesystem paths.

### DIFF
--- a/embassy-stm32-wpan/Cargo.toml
+++ b/embassy-stm32-wpan/Cargo.toml
@@ -59,7 +59,7 @@ wb55_mac = ["dep:bitflags", "dep:embassy-net-driver", "dep:smoltcp", "smoltcp/me
 
 wba = [ "dep:stm32-bindings" ]
 wba_ble = [ "stm32-bindings/wba_wpan_mac" , "stm32-bindings/wba_wpan" ]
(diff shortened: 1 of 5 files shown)


xoviat:
`mac_sys_if` and `ll_sys_if` are perfect. util_seq, however, should not be extern "C" because it is not referred to by any C symbols present in  the libraries. How do I know? You can uncomment `ST_MAC_init` in the mac_ffd example and see the list of missing symbols present. We could expand our "C interface" to include the sequencer of course, but it is a separate component, and therefore we should only include it in the "C" interface if the libraries reference it, which AFAIK, they don't.

Regarding the implementation of the scheduler, what you have is not enough because there needs to be a coupling between the scheduler and the embassy executor so that both the scheduler tasks and the executor tasks can run cooperatively. I think we should try to include an executor in wpan that accomplishes this. However, this can wait until later, and I can take care of this if you want (I have an idea for this, but I'm not sure whether it will work).

Would you like to merge this for now, or do you have any other questions?

PR #5012, opened 2025-12-08
[MCXA]: Adc driver v2

MathisDeroo:
ADC driver update for MCX-A :
- Implement all the pins.
- Manage pin configuration within the driver and removed pin.rs file.
- Add missing ADC instances.
- Support interrupt as per embassy framework.
- Fully support set_conv_command_config.
- Add async function to read ADC value.

ADC2 and ADC3 seem to be missing in the PAC. 

Please @felipebalbi @jamesmunns provide your feedback as usual.
Thanks !

### DIFF
--- a/embassy-mcxa/src/adc.rs
+++ b/embassy-mcxa/src/adc.rs
@@ -2,16 +2,14 @@
 use core::marker::PhantomData;
 
 use embassy_hal_internal::{Peri, PeripheralType};
-
-use crate::gpio::{GpioPin, SealedPin};
-use crate::interrupt::typelevel::{Handler, Interrupt};
-use crate::pac;
 use maitake_sync::WaitCell;
 use paste::paste;
 
 use crate::clocks::periph_helpers::{AdcClockSel, AdcConfig, Div4};
 use crate::clocks::{ClockError, Gate, PoweredClock, enable_and_reset};
-
+use crate::gpio::{GpioPin, SealedPin};
+use crate::interrupt::typelevel::{Handler, Interrupt};
+use crate::pac;
 use crate::pac::adc1::cfg::{HptExdi, Pwrsel, Refsel, Tcmdres, Tprictrl, Tres};
 use crate::pac::adc1::cmdh1::{Avgs, Cmpen, Next, Sts};
 use crate::pac::adc1::cmdl1::{Adch, Mode};


embassy-mcxa/src/interrupt.rs
```diff
@@ -271,11 +274,48 @@ impl InterruptExt for Rtc {
...
+    unsafe fn enable(&self) {
+        cortex_m::peripheral::NVIC::unmask(Interrupt::ADC0);
+    }
+
+    /// Disable ADC0 in NVIC.
+    #[inline]
+    unsafe fn disable(&self) {
+        cortex_m::peripheral::NVIC::mask(Interrupt::ADC0);
+    }
+
```
felipebalbi: none of these are needed. `embassy_hal_internal::interrupt_mod!()` generates a lot of these for us. A considerable amount of this file needs to be deleted. Could no add any new ones meanwhile? It you just use `embassy_hal_internal::InterruptExt` and move a few things around, everything should still work as expected without duplicating the trait on our HAL.

embassy-mcxa/src/adc.rs
```diff
@@ -1,42 +1,29 @@
...
-/// Token for ADC1
-pub type Adc1 = crate::peripherals::ADC1;
-impl Instance for crate::peripherals::ADC1 {
-    #[inline(always)]
-    fn ptr() -> *const Regs {
-        pac::Adc1::ptr()
-    }
-}
+/// Global wait cell for alarm notifications
+static WAKER: WaitCell = WaitCell::new();
```
felipebalbi: This driver works with multiple ADC instances. You probably want a per-instance `WaitCell`, See [i2c](https://github.com/embassy-rs/embassy/blob/main/embassy-mcxa/src/i2c/mod.rs#L111-L114) for an example of how you could do this.

embassy-mcxa/src/adc.rs
```diff
@@ -89,22 +96,28 @@ impl Default for LpadcConfig {
...
 
+/// Configuration for a conversion command.
+///
+/// Defines the parameters for a single ADC conversion operation.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct ConvCommandConfig {
-    pub sample_channel_mode: Ctype,
-    pub channel_number: Adch,
-    pub chained_next_command_number: Next,
+    pub sample_channel_mode: u8,
```
felipebalbi: are all 256 options valid?

embassy-mcxa/src/adc.rs
```diff
@@ -89,22 +96,28 @@ impl Default for LpadcConfig {
...
+    pub channel_number: u8,
```
felipebalbi: 256 channels?

embassy-mcxa/src/adc.rs
```diff
@@ -89,22 +96,28 @@ impl Default for LpadcConfig {
...
+    pub chained_next_command_number: u8,
```
felipebalbi: to avoid repetition, if there are only a handful of valid options, we can use an `enum` to prevent invalid configuration parameters.

embassy-mcxa/src/adc.rs
```diff
@@ -121,14 +137,51 @@ pub struct ConvResult {
     pub conv_value: u16,
 }
```
felipebalbi, line 147: This looks like it should be a `Result<T, Error>`.

embassy-mcxa/src/adc.rs
```diff
@@ -121,14 +137,51 @@ pub struct ConvResult {
...
+    /// # Arguments
+    /// * `_inst` - ADC peripheral instance
+    /// * `pin` - GPIO pin to use for ADC
+    /// * `_irq` - Interrupt binding for this ADC instance
+    /// * `config` - ADC configuration
+    pub fn new(
+        _inst: Peri<'a, I>,
+        pin: Peri<'a, impl AdcPin<I>>,
+        _irq: impl crate::interrupt::typelevel::Binding<I::Interrupt, InterruptHandler<I>> + 'a,
+        config: LpadcConfig,
```
felipebalbi, line 183: Because the config is defined within the module, its name is already namespaced to `hal::lpadc::Config`. The extra `Lpadc` prefix ends up being just duplication.

embassy-mcxa/src/adc.rs
```diff
@@ -121,14 +137,51 @@ pub struct ConvResult {
...
+
+    /// Initialize ADC without interrupt support (for polling mode).
+    ///
+    /// # Arguments
+    /// * `_inst` - ADC peripheral instance
+    /// * `pin` - GPIO pin to use for ADC input
+    /// * `config` - ADC configuration
+    pub fn new_polling(_inst: Peri<'a, I>, pin: Peri<'a, impl AdcPin<I>>, config: LpadcConfig) -> Self {
+        Self::new_inner(_inst, pin, config)
+    }
```
felipebalbi: seems like this driver could benefit from a `Mode` trait encoding `Async` vs `Blocking` behavior.

embassy-mcxa/src/adc.rs
```diff
@@ -280,64 +344,94 @@ impl<'a, I: Instance> Adc<'a, I> {
...
     pub fn set_conv_command_config(&self, index: u32, config: &ConvCommandConfig) {
-        let adc = unsafe { &*I::ptr() };
+        let adc = &*I::ptr();
+
+        macro_rules! write_cmd {
+            ($idx:expr) => {{
+                paste! {
+                    adc.[<cmdl $idx>]().write(|w| unsafe {
+                        w.adch()
+                            .bits(config.channel_number)
```
felipebalbi: you can use `variant()` instead of `bits()`.

embassy-mcxa/src/adc.rs
```diff
@@ -280,64 +344,94 @@ impl<'a, I: Instance> Adc<'a, I> {
...
+    ///
+    /// Configures a conversion command slot with the specified parameters.
+    /// Commands define how conversions are performed (channel, resolution, etc.).
+    ///
+    /// # Arguments
+    /// * `index` - Command index
+    /// * `config` - Command configuration
     pub fn set_conv_command_config(&self, index: u32, config: &ConvCommandConfig) {
-        let adc = unsafe { &*I::ptr() };
+        let adc = &*I::ptr();
```
felipebalbi: does this actually compile? I thought dereferencing raw pointers can only be done within `unsafe` blocks.

UPDATE: checked the `Instance` impl block. `I::ptr()` already produces a regular rust reference. You don't need to deref here. Just `I::ptr()` will do.

embassy-mcxa/src/adc.rs
```diff
@@ -280,64 +344,94 @@ impl<'a, I: Instance> Adc<'a, I> {
...
-                });
-            }
+            1 => write_cmd!(1),
+            2 => write_cmd!(2),
+            3 => write_cmd!(3),
+            4 => write_cmd!(4),
+            5 => write_cmd!(5),
+            6 => write_cmd!(6),
+            7 => write_cmd!(7),
             _ => panic!("Invalid command index: must be between 1 and 7"),
```
felipebalbi: let's no `panic!()` here. Maybe produce a `Result` variant.

embassy-mcxa/src/adc.rs
```diff
@@ -363,47 +465,244 @@ impl<'a, I: Instance> Adc<'a, I> {
...
+    ///
+    /// Reads and returns the next conversion result from the FIFO.
+    /// Returns `None` if the FIFO is empty.
+    ///
+    /// # Returns
+    /// - `Some(ConvResult)` if a result is available
+    /// - `None` if the FIFO is empty
+    pub fn get_conv_result(&self) -> Option<ConvResult> {
+        let adc = &*I::ptr();
+        let fifo = adc.resfifo0().read().bits();
```
felipebalbi: the pac gives you safe field accessors, perhaps use those?

embassy-mcxa/src/adc.rs
MathisDeroo: I don't get your point, can you explain a bit more ?

embassy-mcxa/src/adc.rs
MathisDeroo: I tried and can't as it refers to the wrong cdl::Adch.
I'm currently looking for a way to "dynamically" refer to the proper register module, as depending on the index value passed in parameter cmdl3::Adch or cmdl7::Adch (for instance) is expected by I specify only cmdl1::Adch from use crate::pac::adc1::cmdl1::{Adch, Mode};

jamesmunns: You can PR a patch to the SVD to make all the registers derived from the same instances, like: https://github.com/OpenDevicePartnership/mcxa-pac/blob/main/patch/mrcc.yaml#L15-L27

Then all the types will be the the same, and you can use them across the different peripheral instances.

MathisDeroo: Nice !! 😃 Let me have a look to it, I'll mention you and Felipe for the PR review

felipebalbi:
@MathisDeroo, CI is still failing with your PR, could make `rustfmt` happy?

MathisDeroo:
> @MathisDeroo, CI is still failing with your PR, could make `rustfmt` happy?

Let's see 😄

MathisDeroo:
> > @MathisDeroo, CI is still failing with your PR, could make `rustfmt` happy?
> 
> Let's see 😄

Looks like the job still fails after rustfmt:

```
+ git clone -n https://github.com/MathisDeroo/embassy.git code
Cloning into 'code'...
+ cd code
+ git checkout 4f0eb421de9e08bbbf7f9a58f8b29c451de59894
HEAD is now at 4f0eb421d run rustfmt
+ exec .github/ci/rustfmt.sh
Diff in /ci/code/embassy-mcxa/src/adc.rs:2:
 
 use embassy_hal_internal::{Peri, PeripheralType};
-
-use crate::gpio::{GpioPin, SealedPin};
-use crate::interrupt::typelevel::{Handler, Interrupt};
-use crate::pac;
 use maitake_sync::WaitCell;
 use paste::paste;
 
Diff in /ci/code/embassy-mcxa/src/adc.rs:12:
 use crate::clocks::periph_helpers::{AdcClockSel, AdcConfig, Div4};
 use crate::clocks::{ClockError, Gate, PoweredClock, enable_and_reset};
-
+use crate::gpio::{GpioPin, SealedPin};
+use crate::interrupt::typelevel::{Handler, Interrupt};
+use crate::pac;
 use crate::pac::adc1::cfg::{HptExdi, Pwrsel, Refsel, Tcmdres, Tprictrl, Tres};
 use crate::pac::adc1::cmdh1::{Avgs, Cmpen, Next, Sts};
 use crate::pac::adc1::cmdl1::{Adch, Mode};
run failed: exited with code 123

```

embassy-mcxa/src/adc.rs
jamesmunns: The suggestion is to just name the type `Config`, instead of `LpadcConfig`. Since people import it from the `lpadc` module, having the prefix on the type doesn't bring a lot of value.

If people DO need to bring in multiple configs, they can rename on import, e.g.

```rust
use hal::lpadc::Config as AdcConfig;
use hal::lpuart::Config as LpuartConfig;
```

jamesmunns:
@MathisDeroo I have some follow-up comments, but the CI on this is now green. I'll open up a follow-up PR and tag you in it!

embassy-mcxa/src/adc.rs
MathisDeroo: I see ! Make sense thanks

PR #5015, opened 2025-12-09
[MCXA]: Extend DMA interface

jamesmunns:
CC @bogdan-petru 

This is a rebased + reworked version of https://github.com/OpenDevicePartnership/embassy-mcxa/pull/52.

You can diff JUST my changes with this comparison: https://github.com/jamesmunns/embassy/compare/james/rebase-dma...jamesmunns:embassy:james/dma-suggestions, https://github.com/jamesmunns/embassy/tree/james/rebase-dma has JUST the rebased original PR, with no additional changes.

### DIFF
--- a/embassy-mcxa/src/dma.rs
+++ b/embassy-mcxa/src/dma.rs
@@ -773,7 +773,12 @@ impl<C: Channel> DmaChannel<C> {
     ///
     /// The source and destination buffers must remain valid for the
     /// duration of the transfer.
-    pub fn mem_to_mem<W: Word>(&self, src: &[W], dst: &mut [W], options: TransferOptions) -> Result<Transfer<'_>, Error> {
+    pub fn mem_to_mem<W: Word>(
+        &self,
+        src: &[W],
+        dst: &mut [W],
+        options: TransferOptions,
+    ) -> Result<Transfer<'_>, Error> {
         let mut invalid = false;
         invalid |= src.is_empty();
         invalid |= src.len() > dst.len();
@@ -1789,7 +1794,7 @@ impl<'a> Transfer<'a> {
             if es.err().is_error() {
                 // Currently, all error fields are in the lowest 8 bits, as-casting truncates
                 let errs = es.bits() as u8;
-                return Poll::Ready(Err(TransferErrorRaw(errs)))
+                return Poll::Ready(Err(TransferErrorRaw(errs)));
             }
 
             // Check if we're past the half-way point
@@ -1907,7 +1912,6 @@ impl TransferErrorRaw {
     pub fn has_source_address_err(&self) -> bool {
         (self.0 & (1 << 7)) != 0
     }
-
 }
 
 impl Iterator for TransferErrorRawIter {


jamesmunns:
Also worth noting, I've renamed some examples out to `raw_dma_*`, these are for examples where we don't have a fully-safe API for yet.

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -692,10 +696,37 @@ pub struct LpuartRx<'a, M: Mode> {
...
     _rts_pin: Option<Peri<'a, AnyPin>>,
-    _rx_dma: Option<Channel<'a>>,
     mode: PhantomData<(&'a (), M)>,
 }
 
+/// Lpuart TX driver with DMA support.
+pub struct LpuartTxDma<'a, T: Instance, C: DmaChannelTrait> {
+    info: Info,
+    _tx_pin: Peri<'a, AnyPin>,
+    tx_dma: DmaChannel<C>,
```
felipebalbi, line 706: Shouldn't this be something like a `Peri<'a, AnyChannel>`? That is, be a peripheral and a type-erased DMA channel type?

embassy-mcxa/src/dma.rs
```diff
@@ -0,0 +1,2445 @@
...
+    /// * `options` - Transfer configuration options
+    ///
+    /// # Safety
+    ///
+    /// The source and destination buffers must remain valid for the
+    /// duration of the transfer.
+    pub fn mem_to_mem<W: Word>(&self, src: &[W], dst: &mut [W], options: TransferOptions) -> Transfer<'_> {
+        assert!(!src.is_empty());
+        assert!(dst.len() >= src.len());
+        assert!(src.len() <= 0x7fff);
```
felipebalbi: how about producing a `Result` instead?

embassy-mcxa/src/dma.rs
```diff
@@ -0,0 +1,2445 @@
...
+        t.ch_csr().modify(|_, w| w.done().clear_bit_by_one());
+
+        fence(Ordering::SeqCst);
+    }
+}
+
+impl<'a> Unpin for Transfer<'a> {}
+
+impl<'a> Future for Transfer<'a> {
+    type Output = ();
```
felipebalbi: Looks like the output could encode potential errors and perhaps tell the caller how many bytes were transferred.

embassy-mcxa/src/lpuart/mod.rs
jamesmunns: I think it's reasonable, DmaChannel is also a zst, and has useful helper methods.

embassy-mcxa/src/dma.rs
jamesmunns: Done

embassy-mcxa/src/dma.rs
jamesmunns: Added the ability to return any errors, if present.

Getting the number of bytes transfered is a bit harder if it wasn't a simple transfer (e.g. scatter gather, etc.)

PR #5019, opened 2025-12-09
Feature: USB SOF support

dragonmux:
For some firmware we're building we need access to Start of Frame interrupts to run certain actions and state transitions. This PR aims to build out the machinery to support this and provide the hooks to a Handler for these bus events.

Changes tested working on a STM32U585.


embassy-usb-synopsys-otg/src/lib.rs
```diff
@@ -30,12 +30,20 @@ pub unsafe fn on_interrupt<const MAX_EP_COUNT: usize>(r: Otg, state: &State<MAX_
...
+        || ints.otgint()
+        || ints.srqint()
+        || ints.sof()
+    {
         // Mask interrupts and notify `Bus` to process them
         r.gintmsk().write(|w| {
             w.set_iepint(true);
             w.set_oepint(true);
             w.set_rxflvlm(true);
+            w.set_sofm(true);
```
felipebalbi: you should probably make this optional. One interrupt every 125 us is quite a lot.

dragonmux: Good point - need to think on how to do this then in a way that makes a reasonable amount of sense.. perhaps having the interrupt enable being a call user code can make during setup and being a toggle - `set_enable_sof()` to match the naming scheme used?

dragonmux: Working on the assumption that naming is okay, we have implemented the suggestion and documented accordingly.

felipebalbi: sorry for the delay. Why don't you add a `enable_sof_interrupts` to the `Config` struct passed to the `Driver` impl?
Then just make sure to update the docs to reflect the idea that it will cause a lot of interrupts and should only be enabled if you really it.

dragonmux: Ok, we'll change the approach to that then 👍🏼 thought of using a function so the user would be able to turn them on and off on demand, but you're correct that it wants to be from the Config as the stack is currently written on account of it not supporting multi-config devices.

dragonmux: That should now be done - please let us know what you think.

embassy-usb/src/builder.rs
```diff
@@ -117,6 +117,16 @@ pub struct Config<'a> {
...
     pub max_power: u16,
+
+    /// Whether to enableSOF interrupts
+    ///
+    /// Start of Frame can be a useful event on the bus, happening either every
+    /// 1ms for USB LS and FS, or every 125µs for USB HS. This can be used as a
+    /// free extra timer in a lot of cases, and as a way to process events at
+    /// the end of the frame if things needed deferring. However, this can be
+    /// a significant amount of interrupt traffic, so only enable this if you
+    /// really need this style of handling.
```
felipebalbi, line 145: like the other fields, please add "Default: `false`".

embassy-usb-synopsys-otg/src/lib.rs
```diff
@@ -500,6 +507,7 @@ impl<'d, const MAX_EP_COUNT: usize> embassy_usb_driver::Driver<'d> for Driver<'d
                 ep_out: self.ep_out,
                 inited: false,
                 instance: self.instance,
+                enable_sof_interrupts: enable_sof_interrupts,
```
felipebalbi: You can use the shortcut initialization:

```suggestion
                enable_sof_interrupts,
```

dragonmux: That looks like a refactoring left-over from when the field didn't match the parameter name 👍🏼

felipebalbi (approved):
LGTM, but I'm not the owner of the USB driver, would prefer someone if else would review and merge.

CBJamo:
The 2040/2350 can support this as well. If you don't have hardware to test with I'll give it a go after this merges.

dragonmux:
We don't have a good way to test that, no - we only touched the hardware we're set up to test right now and could prove out.

leftger (approved):
LGTM

embassy-usb-driver/src/lib.rs
```diff
@@ -168,7 +168,7 @@ pub trait Driver<'a> {
     ///
     /// This consumes the `Driver` instance, so it's no longer possible to allocate more
     /// endpoints.
-    fn start(self, control_max_packet_size: u16) -> (Self::Bus, Self::ControlPipe);
+    fn start(self, control_max_packet_size: u16, enable_sof_interrupts: bool) -> (Self::Bus, Self::ControlPipe);
```
bugadani, line 196: I wonder if this could be reformulated in a way that would itself be a breaking change now, but future additions wouldn't. If `start` had a non-exhaustive Config struct with setter methods as its argument, adding new options to that Config struct could be a minor change.

But is this even necessary, couldn't the driver just enable the interrupt unconditionally?

Dirbaio:
I think the structure of this is wrong.  `embassy-usb-driver` crate is a portable, hardware-independent trait. Therefore you should be able to use it in a fully hardware-independent way. 

If you set `enable_sof_interrupts = true` to the hardware-independent you're requesting to enable SOF interrupts because you want to do something with them. But ... how? How can you handle this interrupt in a hardware-independent way? You can't.

Adding a flag to a hardware-independent trait to enable a hardware-dependent feature makes no sense. Same with the `embassy-usb` changes, all they do is forward the flag, but `embassy-usb` itself has no way to use the SOF interrupt for anything. The user has to.

The flag should be in the hardware-specific API, in the constructors for stm32 usb and otg.

Dirbaio:
To clarify, I think we should do either of these:

A) Actually wire up everything so that you can use SOF fully from the hardware-independent embassy-usb API (for example, methods like `set_sof_handler()` or `get_sof_count()` or `wait_for_sof()`, whatever is needed to use SOF)
B) Keep the flag in hardware-dependent code.

Threading the flag through hardware-independent code just to enable hardware-dependent behavior is a weird middle.

felipebalbi:
> I think the structure of this is wrong. `embassy-usb-driver` crate is a portable, hardware-independent trait. Therefore you should be able to use it in a fully hardware-independent way.
> 
> If you set `enable_sof_interrupts = true` to the hardware-independent you're requesting to enable SOF interrupts because you want to do something with them. But ... how? How can you handle this interrupt in a hardware-independent way? You can't.
> 
> Adding a flag to a hardware-independent trait to enable a hardware-dependent feature makes no sense. Same with the `embassy-usb` changes, all they do is forward the flag, but `embassy-usb` itself has no way to use the SOF interrupt for anything. The user has to.
> 
> The flag should be in the hardware-specific API, in the constructors for stm32 usb and otg.

yeah, this makes sense, actually. Is there a case where SOF should be enabled temporarily? Otherwise, enabling it only at construction time should work just fine.

dragonmux:
> Is there a case where SOF should be enabled temporarily?

SOF creates a phenominal number of interrupts particularly under HS USB. If you're implementing something that only needs SOF handling under certain conditions such as DFU when actually actively downloading data to the device, then that's fine.. but during the idle state that's wholy inappropriate and needlessly causes high load and thus high power usage. It makes perfect sense to only enable SOF interrupts on reception of the first DFU_DNLOAD packet, and turn them back off during manifestation - for example.

If the firmware being written never uses SOF, then you're subjecting people to a permanent significantly high interrupt load as a penalty for just using USB, trashing power usage and potentially even the timing characteristics they may need.

Not having the ability to control when SOF is enabled is just silly, and articially limits the kinds of thing you can express with embassy's USB stack. Artificial limitaiton is a problem that already plagues the stack - you can't have more than one configuration, for example, which limits some of the things that can be done on more USB endpoint resource constrained devices and if you wish to do more complex multi-function things.

We've had our hands rather full with both a not amazing personal situation, and work things. We'll try and address the review point about using a config struct, however the change as it stands is written to ballance the performance hit and power usage load against the needs of things like DFU when needing to do the Flash manipulation out of band to the DFU polling machinary. If you don't need SOF, you should not pay for it. If you need it you should.

dragonmux:
> If you set enable_sof_interrupts = true to the hardware-independent you're requesting to enable SOF interrupts because you want to do something with them. But ... how? How can you handle this interrupt in a hardware-independent way? You can't.

SOF interrupts are used to hang stuff on that you want to do independent of request traffic and on what is essentially a free timer. SOF itself isn't the event you're necessarily handling.. this misunderstands what the marker and the hardware provided support for alerting about the marker.. is for.. The handling goes in a Handler same as "you have a packet on an endpoint", "host has asked to switch alt-modes on an interface", and "there's a control packet here for you".. see https://github.com/dragonmux/dragonBoot-rs/blob/aca8e0f72a4f8e72f23ead67b24b02f8434725c7/src/dfu.rs#L397 for an example and how incredibly hardware-independant it is.

Dirbaio:
that link is 404

dragonmux:
Ah, right, sorry - private repo.. it does exist, it's just not public right now. 'sec while we change that as there's no reason now for it to be private.

flaviojs:
SOF already exists in the USB 1 spec, and is mentioned for Isochronous transactions and error handling.
It seems like SOF events and SOF handling is a generic usb concept and should be part of the generic API.

The versatility and extra processing means it should not be a configuration thing but instead a enable/disable-at-runtime thing.
It probably should be part of the generic API, I'm not aware of any hardware that does not allow SOF enabling/disabling, but the function trait could just return a `Result` and allow the driver to error if it decides not to implement it.

Dirbaio:
I'm going to close this since I think it's not the right design. 

The intended use for SOF is *not* "a free 1khz timer I can use to tick my state machines". Sure it's convenient but you have alternatives. You can use true async instead of polling things. You can poll in a background task on a ticker, or on a `loop { poll(); yield_now().await }` if you don't want to drag in `embassy_time`.

The use for SOF - the case where you *really* need to handle SOF - is Audio, Video and other isochronous use cases. It's critical for timing, so you don't want it to go through the executor and the USB task's poll loop, you want to run your own code in the actual SOF interrupt. The SOF callback as implemented in this PR goes through the USB task, so it can have unbounded latency between the SOF arriving and the user code in the callback running, so it's not usable for timing.

I would accept a PR that adds support for embassy-usb-synopsys-otg (and other USB driver impls) to specify a callback that runs directly from the SOF IRQ. Since this is hardware-dependent there should be no standard API in embassy-usb to enable it, it should be in embassy-usb-synopsys-otg (and whatever driver)

PR #5022, opened 2025-12-09
MCXA TRNG driver

felipebalbi:


embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+            run_length4_limit_min: 0,
+            run_length5_limit_max: 0,
+            run_length5_limit_min: 0,
+            run_length6_limit_max: 0,
+            run_length6_limit_min: 0,
+            retry_count: 2,
+            long_run_limit_max: 32,
+            sparse_bit_limit: 0,
+            poker_limit_max: 0,
+            osc_mode: OscMode::DualOscs,
```
felipebalbi: Should I add a comment here stating these were directly taken from NXP's sdk sample?

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -949,4 +949,6 @@ pub(crate) mod gate {
 
     // DMA0 peripheral - uses NoConfig since it has no selectable clock source
     impl_cc_gate!(DMA0, mrcc_glb_cc0, mrcc_glb_rst0, dma0, NoConfig);
+    // TRNG peripheral - uses NoConfig since it has no selectaple clock source
```
jamesmunns:
```suggestion
    // TRNG peripheral - uses NoConfig since it has no selectable clock source
```

embassy-mcxa/Cargo.toml
```diff
@@ -48,6 +48,9 @@ maitake-sync = { version = "0.2.2", default-features = false, features = ["criti
 embassy-time = { version = "0.5.0", optional = true }
 embassy-time-driver = { version = "0.2.1", optional = true }
 
+rand-core-06 = { package = "rand_core", version = "0.6" }
```
jamesmunns, line 51: Any reason to support 0.6? It seems like the bulk of the world is already on v0.9.x: https://crates.io/crates/rand-core

embassy-mcxa/Cargo.toml
```diff
@@ -39,7 +39,7 @@ embedded-hal-nb = { version = "1.0" }
 embedded-io = "0.6"
 embedded-io-async = { version = "0.6.1" }
 heapless = "0.8"
-mcxa-pac = { git = "https://github.com/OpenDevicePartnership/mcxa-pac", features = ["rt", "critical-section"], version = "0.1.0", rev = "e7dfed8740b449b6ac646bab8ac6776a3c099267" }
+mcxa-pac = { git = "https://github.com/OpenDevicePartnership/mcxa-pac", features = ["rt", "critical-section"], version = "0.1.0", rev = "02bd04a21ef8f8f67f88239ff5df765bb7e60fd8" }
```
jamesmunns, line 42: Do we want to get `mcxa-pac` released? Not a blocker for this PR, but might be worth doing it to avoid needing to keep the version pinned.

embassy-mcxa/src/trng.rs
jamesmunns: Could do that, or make this a method like `Config::nxp_sdk_defaults()`, document it there, and call that func inside of `Default::default`.

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+        regs().mctl().write(|w| w.rst_def().set_bit());
+    }
+}
+
+/// TRNG Async Driver
+pub struct AsyncTrng<'d> {
+    _peri: Peri<'d, TRNG0>,
+}
+
+impl<'d> AsyncTrng<'d> {
```
jamesmunns: Any reason you went for `AsyncTrng` instead of `Trng<'d, M: Mode>`? It's a little inconsistent with other peripherals.

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+    /// Return a random u64, blocking version.
+    pub fn blocking_next_u64(&mut self) -> u64 {
+        Self::start();
+        Self::blocking_wait_for_generation();
+
+        let mut result = u64::from(regs().ent(0).read().bits()) << 32;
+        result |= u64::from(regs().ent(1).read().bits());
+
+        // New random bytes are generated only after reading ENT7
+        let _ = regs().ent(7).read().bits();
```
jamesmunns: Maybe use ent7 for the lower bits then so you don't need the extra read?

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    /// Return a random u32, blocking version.
+    pub fn blocking_next_u32(&mut self) -> u32 {
+        Self::start();
+        Self::blocking_wait_for_generation();
+        let result = regs().ent(0).read().bits();
+
+        // New random bytes are generated only after reading ENT7
+        let _ = regs().ent(7).read().bits();
```
jamesmunns: Maybe just use ent(7) for the value then instead of needing a second read?

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+    fn fill_chunk(chunk: &mut [u8]) {
+        let mut entropy = [0u32; 8];
+
+        for (i, item) in entropy.iter_mut().enumerate() {
+            *item = regs().ent(i).read().bits();
+        }
+
+        let entropy: [u8; 32] = unsafe { core::mem::transmute(entropy) };
+
+        chunk.copy_from_slice(&entropy[..chunk.len()]);
```
jamesmunns, line 177: This assumes `chunk` is <= 32. We could probably avoid pre-chunking before calling this function, and just handle reasonable stride here instead.

If not, would be good to document the requirements here. There will also probably be a panicking branch from copy_from_slice.

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    /// Return a random u32, blocking version.
+    pub fn blocking_next_u32(&mut self) -> u32 {
+        Self::start();
+        Self::blocking_wait_for_generation();
+        let result = regs().ent(0).read().bits();
+
+        // New random bytes are generated only after reading ENT7
+        let _ = regs().ent(7).read().bits();
+        Self::stop();
```
jamesmunns: Is there a benefit to stopping the TRNG? it seems like leaving it running would be a good thing to keep accumulating entropy between calls?

I don't see a TRNG section in the ref manual I have, so I might need to get that from you.

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+impl<'d> Trng<'d> {
+    /// Instantiates a new TRNG peripheral driver.
+    pub fn new(_peri: Peri<'d, TRNG0>, config: Config) -> Self {
+        _ = unsafe { enable_and_reset::<TRNG0>(&NoConfig) };
+
+        Self::configure(config);
+        Self { _peri }
+    }
+
+    fn configure(config: Config) {
```
jamesmunns, line 53: In general: missing any checks that max >= min for config

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,620 @@
...
+    InvalidBuffer,
+}
+
+/// I2C interrupt handler.
+pub struct InterruptHandler;
+
+impl Handler<typelevel::TRNG0> for InterruptHandler {
+    unsafe fn on_interrupt() {
+        if regs().int_status().read().bits() != 0 {
+            regs().int_ctrl().write(|w| {
```
jamesmunns, line 457: if you're doing a full clear, and all bits are zero on reset, you can just do `reg().int_ctrl().write(|w| w);`

embassy-mcxa/src/trng.rs
felipebalbi: I added three dedicated constructors following the manual and left a `new_with_custom_config` constructor which takes the full `Config` argument and allows the user to configure it any way they want. Let me know what you think? Happy to update.

embassy-mcxa/Cargo.toml
felipebalbi: Are you happy with the current state of the PAC? My concern is that we release it now and soon, as we get more parts of the MCU lit up we keep making releases because of more SVD issues we find along the way. No strong feelings from me, just wondering.

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,760 @@
...
+    }
+
+    /// Instantiates a new TRNG peripheral driver with 512 samples of entropy.
+    pub fn new_512(_peri: Peri<'d, TRNG0>) -> Self {
+        Self::new_inner(_peri, Default::default())
+    }
+
+    /// Instantiates a new TRNG peripheral driver.
+    ///
+    /// NOTE: this constructor makes not attempt at validating the
```
jamesmunns:
```suggestion
    /// NOTE: this constructor makes no attempt at validating the
```

embassy-mcxa/src/trng.rs
```diff
@@ -0,0 +1,760 @@
...
+                run_length2_limit_max: 24,
+                run_length2_limit_range: 25,
+                run_length3_limit_max: 17,
+                run_length3_limit_range: 18,
+                ..Default::default()
+            },
+        )
+    }
+
+    /// Instantiates a new TRNG peripheral driver with  256 samples of entropy.
```
jamesmunns:
```suggestion
    /// Instantiates a new TRNG peripheral driver with 256 samples of entropy.
```

embassy-mcxa/src/trng.rs
felipebalbi: I can certainly change :) In fact, I had the mode typestate before. No real reason to remove it, honestly.

embassy-mcxa/Cargo.toml
jamesmunns: I don't have any particular preference between the two, but I also don't know how heavyweight a process it is for an ODP crate release.

version numbers are free, might as well use them :)

embassy-mcxa/src/trng.rs
felipebalbi: yeah, I changed that to `max` and `range` since that's what the HW wants. Additionally, `min` can be larger than `max` (well, in `unsigned` arithmetic anyway) because some min values can go negative. Switching to `range` instead of `min` "solves" that too.

embassy-mcxa/src/trng.rs
felipebalbi: https://www.nxp.com/docs/en/reference-manual/MCXAP144M240F61SRM.pdf

Chapter 17

embassy-mcxa/src/trng.rs
felipebalbi: Let me know if you'd prefer the `Mode` typestate.

embassy-mcxa/src/clocks/mod.rs
felipebalbi: argh, done :)

embassy-mcxa/Cargo.toml
felipebalbi: I'd defer to @Dirbaio here. `embassy-rp` supports both. Do we want to drop support for 0.6?

embassy-mcxa/src/trng.rs
felipebalbi: private method only called from `*fill_bytes()` and that definitely splits into 32-byte chunks.

embassy-mcxa/src/trng.rs
felipebalbi: I guess I could. Technically the randomness should be uniformly distributed.

jamesmunns:
@felipebalbi you can now run `./fmtall.sh` at the repo root to fix those pesky fmt errors :)

embassy-mcxa/src/trng.rs
felipebalbi: done

embassy-mcxa/src/trng.rs
felipebalbi: that, for some reason, didn't work. Maybe the register mask is incorrect in the SVD? I'll spend some more time trying to understand it but, for now, left it as is.

embassy-mcxa/src/trng.rs
jamesmunns: It would be consistent with the other drivers. I don't feel particularly strongly about it though. Probably depends on if there's anything worth sharing between the impls.

embassy-mcxa/src/trng.rs
felipebalbi: done.

embassy-mcxa/Cargo.toml
felipebalbi: I'll create a release shortly :)

embassy-mcxa/src/trng.rs
felipebalbi: Sounds good, let me update it.

felipebalbi: done

PR #5030, opened 2025-12-10
MCXA: Add a function to read Reset Reason

felipebalbi:


embassy-mcxa/src/reset_reason.rs
```diff
@@ -0,0 +1,108 @@
+//! Reset reason
+//!
+//! MCXA families keep the most recent reset reason in the SRS
+//! register of the CMC block. This lets users understand why the MCU
+//! has reset and take appropriate corrective actions if required.
+
+/// Reads the most recent reset reason from the Core Mode Controller
+/// (CMC).
+pub fn reset_reason() -> ResetReason {
+    critical_section::with(|_| {
```
jamesmunns: We don't need a critical section here, the read is atomic, and the checks made after `read()` are on the stored `u32` value. The value also doesn't seem to be clear-able (only SSRS is clear-able), so probably no need to worry about races either.

embassy-mcxa/src/reset_reason.rs
```diff
@@ -0,0 +1,108 @@
...
+        let regs = unsafe { &*crate::pac::Cmc::steal() };
+
+        let srs = regs.srs().read();
+
+        if srs.wakeup().is_enabled() {
```
jamesmunns: I'm not sure if more than one item can be set, but if so, we might want something like we have for DMA errors: https://github.com/embassy-rs/embassy/blob/32a1d0ef7ea52d3cee9959ff52d47fd13fc6b4b9/embassy-mcxa/src/dma.rs#L1836-L1965

felipebalbi: I had the same question but then I decided that being reset by both Watchdog and Jtag at the same time is probably not possible.

embassy-mcxa/src/reset_reason.rs
felipebalbi: done

PR #5035, opened 2025-12-11
Can bus fifo selection

### LINKED ISSUE #5034: STM32: Fifo selection for CAN bus
The current CAN bus driver for STM32 doesn't support fifo selection when reading incoming messages. It's reasonable to conceive of a scenario wherein message destined for fifo0 starve out messages for fifo1.
Although it seems to be abandoned now, the fdcan crate from Richard Meadows, which seems to serve as the inspiration for this implementation, had split up reads into `receive0` and `receive1` which I think gives the user much better control.
I've put together a quick PR that restores the granularity but prefers to receive an enum to the read function specifying the desired fifo instead of having separate functions.

Please let me know if this feature is desired and and I can finish fixing the examples.

bryanniwa:
Closes #5034

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -340,9 +340,9 @@ impl<'d> Can<'d> {
         TxMode::write(&self.info, frame).await
     }
 
-    /// Returns the next received message frame
-    pub async fn read(&mut self) -> Result<Envelope, BusError> {
-        RxMode::read_classic(&self.info).await
+    /// Returns the next received message frame from the desired FIFO
+    pub async fn read(&mut self, fifo: Fifo) -> Result<Envelope, BusError> {
+        RxMode::read_classic(&self.info, fifo).await
     }
 
     /// Queues the message to be sent but exerts backpressure.  If a lower-priority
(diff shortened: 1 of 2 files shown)


xoviat:
build needs to pass

bryanniwa:
Yeah sorry, I created an issue with more information. Probably shouldn't have bothered with that and just put it all in the PR description. I was looking for some feedback on the idea before I nailed down all the tests and examples.
If you want, I can fix the build before you take a look anyway

xoviat:
It seems reasonable to me, so I will probably merge it if you fix the build. The only possible is concern is that some users may want to read from any fifo--the old behavior?

cschuhen:
Personally I'm not a fan of this change as is. Sure it is possible to starve fifo1, but in that situation, things are kind of broken anyway. You will get overruns. This change is also quite unfriendly for many users who just want the next message without having to worry about the two FIFO's. Many users will just try to set up filters to split traffic between two to make the best use of the FIFO memory. Also, it becomes a breaking change.

If this change is really required, I'd prefer to not change the existing functions. We could add a family of *_fifo or *_from_fifo variants. In this case, one approach could be to add 'Any' to the enum and then the 'old' functions just wrap the new.

bryanniwa:
@cschuhen that seems reasonable to me; making it a breaking change is definitely not required. I still think it's useful to have access to the individual FIFOs so I'll probably do as you suggested and wrap the original commands while exposing a more specific method in case the user desires finer control.

I haven't had time to work on this recently so I'll finish it up this upcoming weekend.

leftger:
@bryanniwa This PR is quite old. Do you still want to submit this PR? If you can address the changes within the next month I'm sure we can get it merged in, otherwise it should get deleted by then.

PR #5037, opened 2025-12-11
[MCXA]: Refactor the ADC API

jamesmunns:
The current ADC API exposes quite a bit of low level functionality, but is not yet particularly "misuse resistant". It allowed (and still somewhat allows) the user to make configuration combinations that could fail to work at runtime.

I've restructured the current API a bit, particularly for the async variant, to make it easier to get right. I've left the blocking API still mostly allow whatever you'd like.

I think there's more we COULD do here, but speaking with @felipebalbi this is probably not a high priority, as the currently planned usage is pretty simple.

Some notable changes:

* Some APIs were made blocking-only, and removed from the async variant
* The async variant can no longer manually set command and trigger details, some setting have been made public methods
* I've changed the `impl` macro to tie the correct channel to the correct pin, and removed that from being configurable

CC @MathisDeroo - let me know if you have any questions WHY I made any changes!

### DIFF
--- a/embassy-mcxa/src/adc.rs
+++ b/embassy-mcxa/src/adc.rs
@@ -439,9 +439,7 @@ impl<'a, M: ModeAdc> Adc<'a, M> {
             .modify(|_, w| w.cal_avgs().variant(config.conversion_average_mode));
 
         adc.cfg().write(|w| unsafe {
-            if config.enable_analog_preliminary {
-                w.pwren().pre_enabled();
-            }
+            w.pwren().bit(config.enable_analog_preliminary);
 
             w.pudly()
                 .bits(config.power_up_delay)


jamesmunns:
I also went ahead and remove the instance generic from the ADC as well.

embassy-mcxa/src/adc.rs
```diff
@@ -241,22 +432,16 @@ impl<'a, I: Instance, M: ModeAdc> Adc<'a, I, M> {
...
 
         adc.cfg().write(|w| unsafe {
-            let w = if config.enable_analog_preliminary {
-                w.pwren().pre_enabled()
-            } else {
-                w
-            };
+            if config.enable_analog_preliminary {
+                w.pwren().pre_enabled();
+            }
```
felipebalbi: I think you can use something like `variant(config.enable_analog_preliminary)`

embassy-mcxa/src/adc.rs
```diff
@@ -363,185 +543,87 @@ impl<'a, I: Instance, M: ModeAdc> Adc<'a, I, M> {
...
+    fn set_conv_command_config_inner(&self, index: usize, config: &ConvCommandConfig) -> Result<()> {
+        let (cmdl, cmdh) = match index {
+            1 => (self.info.cmdl1(), self.info.cmdh1()),
+            2 => (self.info.cmdl2(), self.info.cmdh2()),
+            3 => (self.info.cmdl3(), self.info.cmdh3()),
+            4 => (self.info.cmdl4(), self.info.cmdh4()),
+            5 => (self.info.cmdl5(), self.info.cmdh5()),
+            6 => (self.info.cmdl6(), self.info.cmdh6()),
+            7 => (self.info.cmdl7(), self.info.cmdh7()),
+            _ => return Err(Error::InvalidConfig),
```
felipebalbi, line 577: we should probably make these into arrays in the PAC

felipebalbi (approved):
LGTM, two very minor nits.

embassy-mcxa/src/adc.rs
jamesmunns: I'm not sure if would be a win, the numbering is 1-7, so we'd have to do an offset by one anyway.

MathisDeroo:
Hi @jamesmunns thanks for the changes ! I totally agree with you based on our discussion yesterday. I'll go over the changes to see what's the best way to do it for future drivers

PR #5047, opened 2025-12-12
Add `run_until` function to std Executor as to support grafeul shutdown.

Gerharddc:
This is an alternative approach to https://github.com/embassy-rs/embassy/pull/4923 that strives to be simpler

### DIFF
--- a/embassy-executor/src/arch/std.rs
+++ b/embassy-executor/src/arch/std.rs
@@ -55,11 +55,24 @@ mod thread {
         ///
         /// This function never returns.
         pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+            self.run_until(init, || false);
+            unreachable!()
+        }
+
+        /// Run the executor until a flag is raised.
+        ///
+        /// This function is identical to `Executor::run()` apart from offering a `done` flag to stop execution.
+        pub fn run_until(&'static mut self, init: impl FnOnce(Spawner), mut done: impl FnMut() -> bool) {
             init(self.inner.spawner());
 
             loop {
                 unsafe { self.inner.poll() };
-                self.signaler.wait()
+
+                if done() {
+                    break;
+                }
+
+                self.signaler.wait();
             }
         }
     }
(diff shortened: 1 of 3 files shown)


embassy-executor/src/arch/std.rs
```diff
@@ -62,6 +63,23 @@ mod thread {
                 self.signaler.wait()
             }
         }
+
+        /// Run the executor until a flag is raised.
+        ///
+        /// This function is identical to `Executor::run()` apart from offering a `done` flag to stop execution.
+        pub fn run_until(&'static mut self, done: &'static AtomicBool, init: impl FnOnce(Spawner)) {
```
Dirbaio: it'd be simpler to do `done: impl FnMut() -> bool`. This way the user can use any synchronization method, not necessarily an AtomicBool.

Dirbaio: also i'd order the params as `run_until(self, init, done)`. it's more consistent if we have `run`, `run_until` that the extra params are added to the right always.

embassy-executor/src/arch/std.rs
Dirbaio, line 57: you can implement `run` by calling `run_until` to avoid duplicate code.

Gerharddc: indeed, that makes it much cleaner

embassy-executor/src/arch/std.rs
```diff
@@ -55,11 +55,24 @@ mod thread {
         ///
         /// This function never returns.
         pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+            self.run_until(init, || false);
+            panic!("It should not be possible to get here.")
```
Dirbaio: the compiler can check this for you. remove this line and the `;` in the previous line and it'll compile.

Gerharddc: not for me at least, it complains about the return type being `()` instead of `!`. it does not seem to realise this causes an endless loop

embassy-executor/src/arch/std.rs
Gerharddc: implemented, thanks!

embassy-executor/src/arch/std.rs
Gerharddc: the alternative seems to be to use the unsafe `std::hint::unreachable_unchecked` function but I'm not sure if that is better than panic

Dirbaio: ah right `run_until()` is not `-> !`. Sorry, my bad, melted brain today.

so just one minor nit left: The idiomatic way to mark spots that should never be reached is `unreachable!()` then, not a panic with a message.

Gerharddc: ah yes, I completely forgot that one exists :)

PR #5049, opened 2025-12-12
Spi::new, apply IO driver strength

denenilsson:
instantiation with Spi::new did not apply IO driver strength from config (speed)


Dirbaio:
are you sure it's not working? it's supposed to be applied by these lines

https://github.com/embassy-rs/embassy/blob/640ca72c0e33b252026191ac9bc64fa6b0c9b166/embassy-stm32/src/spi/mod.rs#L754-L755

if it's indeed not getting applied we should look into why these lines aren't applying it, not simply apply it again somewhere else.

denenilsson:
Oh, I missed these lines when I dove into it today. I was prepping my board for a run in the EMC lab in the morning, and the scope gave me sharp edges with ringing on the receiving end. When i patched it went away, strange.. I need to verify on Monday when I am back at the bench.

Dirbaio:
haven't heard back, closing.

PR #5059, opened 2025-12-13
stm32: set pins to ANALOG on init

liebman:


examples/stm32wle5/src/bin/i2c.rs
```diff
@@ -79,14 +61,15 @@ async fn async_main(_spawner: Spawner) {
     let mut i2c = I2c::new(p.I2C2, p.PB15, p.PA15, IrqsI2C, p.DMA1_CH6, p.DMA1_CH7, {
         let mut config = i2c::Config::default();
         config.frequency = Hertz::khz(100);
-        config.timeout = Duration::from_millis(500);
+        config.timeout = Duration::from_millis(1000);
         config
     });
 
+    let _device_busy = low_power::DeviceBusy::new_stop2();
```
xoviat: Why is this needed?

xoviat:
I would like to have a per-port gpio refcount that disables/enables the port entirely and impl Drop for the pins so that they are set as analog on drop (the peripherals can store AnyPin). However, this is a good intermediate compromise.

examples/stm32wle5/src/bin/i2c.rs
liebman: nice catch - I'll remove that. However it looks like for I2C2 is not properly setting its low-power availability, this test reads i2c the first time then fails with a timeout for every attemt after that. I was going to debug that after I finished this PR.

liebman: So `embassy-data` has  `(r"^STM32WLE5.*:I2C2", StopMode::Stop2),`. which is correct but when I run the i2c example its going into STOP2, which should be blocked.

liebman: ok - its this in I2c: 
```
    fn enable_and_init(&mut self, config: Config) {
        self.info.rcc.enable_and_reset_without_stop();
        self.init(config);
    }
```
It should be with_stop its registers loose their config in stop2

liebman: PR: https://github.com/embassy-rs/embassy/pull/5060

xoviat:
bender run

xoviat:
Not perfect, but good enough.

PR #5060, opened 2025-12-13
stm32: i2c should enable with stop

liebman:
An I2C peripheral will loose its register values at some stop level.  We need to prevent that level if its used.

### DIFF
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -226,7 +226,7 @@ impl<'d, M: Mode> I2c<'d, M, Master> {
     }
 
     fn enable_and_init(&mut self, config: Config) {
-        self.info.rcc.enable_and_reset_without_stop();
+        self.info.rcc.enable_and_reset();
         self.init(config);
     }
 }


xoviat:
This will not be merged as is because some (most?) chips don't lose their registers on stop2 and the eventual goal is that users won't have to drop their peripherals to enter stop, everything will Just Work (tm). Now, to actually discuss the way forward (sorry, I keep breaking things):

First, the current implementation by me has a bug (and possibly others as well): it calls disable instead of disable_without_stop. This will need to be changed to disable_without_stop.

Second, I'll accept two approaches for wle (and other chips that lose their registers in stop2):
- enable/disable with cfgs for those chips that lose registers and enable_without_stop/disable_without_stop for all other chips.
- (ideal) store config in the struct and check whether initialization is required at the top of each async public method, right after block_stop.

liebman:
> * enable/disable with cfgs for those chips that lose registers and enable_without_stop/disable_without_stop for all other chips.

I'd prefer this one - how do you envision selecting  enable/disable vs enable_without_stop/disable_without_stop?  

> * (ideal) store config in the struct and check whether initialization is required at the top of each async public method, right after block_stop.

not sure I'm a fan of using more ram for something that only saves about 3µA (stm32wle5 difference between stop1 and stop2 listed in the datasheet), well at least for my silly project,  and my only stm32 experience is stm32wle5.
Also stm32wle5 (based on the datasheet and not my use case) I2C2 slave can interrupt when its address is seen in STOP1, but not in STOP2, the reconfigure on wake would not help here, same with UARTS. (yea they could use I2C3, stop2 safe, or LPUART1) 

That said I read some STM32 app note that suggested that interrupts be disabled before wfe and re-enabled after.  This would allow things like recovering RCC (stm32wle) or to add pre/post hooks where the user could re-configure devices that lost configuration before the interrupt is serviced.  Not sure how feasible that would be for embassy-stm32.

xoviat:
> I'd prefer this one - how do you envision selecting enable/disable vs enable_without_stop/disable_without_stop?

For example, you can just use cfg(wle) for now.

> not sure I'm a fan of using more ram for something that only saves about 3µA 

Is stop2 needed on wle at all? If it's too much trouble we can just disable it on wle until we have a better solution.

liebman:
I'm not against blocking stop2 for now (stm32wle), at least for my current usage, just sad to see the < 1µA power usage for the button_exti example go away ;-).

xoviat:
have you seen #5066? I think it offers the best solution.

lucasgranberg:
I opted for doing low power mode outside of Embassy but I used Standby with RTC and SRAM2 retention. Take all I say with a grain of salt as I am not too familiar with STOP2. I reset the device after standby and started from main again after.

I don't know how different wle is but it seems like it might be an oddball. I would not volounteer to maintain that mess. It seems like there would need to be a lot of exceptions for wle.

I think it is fine to drop the peripherals and reinitialize them after. There is no low power i2c iirc so it makes sense to me to do the initialization "by hand".

Even relying on exti for wakeup is not the most efficient as there is wkup pins that don't need exti. 

Ofc it would be nice to drop the consumption during shorter delays but it seems like an hard problem to solve.

liebman:
closing in favor of https://github.com/embassy-rs/embassy/pull/5066

liebman:
@lucasgranberg 
Some stm32 familys do or can be configured to retain all peripheral registers in STOP2, the stm32wle5 can retain SOME registers, these are all listed in the datasheet/rm. Having it automatically drop to using just a few microamps then when timer/signal happens continuing on from exactly where it was takes a huge burden off of the developer.

Using stop modes means you don't waste cycles initializing peripherals at each wake, this can make an app use less power then using standby (depending on how often you need to wake).

PR #5061, opened 2025-12-13
embassy_rp: improve calculate_pio_clock_divider

### LINKED ISSUE #5048: embassy-rp: calculate_pio_clock_divider returns invalid clock_divider
Testing code and expected behaviour:

```rust
use embassy_rp::pio_programs::clock_divider::calculate_pio_clock_divider;
use fixed::FixedU32;
use fixed::types::extra::U8;

type Div = FixedU32<U8>;

let target_hz: u32 = 96_000_000;
let sys_hz: u32 = embassy_rp::clocks::clk_sys_freq();

// testing with 240MHz system clock
assert_eq!(sys_hz, 240_000_000);

// current:
let divider_buggy: Div = calculate_pio_clock_divider(target_hz);

// fixed:
let divider_q8 = ((sys_hz as u64) << 8) / (target_hz as u64);
let divider_correct = Div::from_bits(divider_q8 as u32);

// should be 2.5 in Q8 (clk_sys / 2.5 = 96 MHz)
assert_eq!(divider_correct, Div::from_num(2.5f32));

// fails: 3 != 2.5
assert_eq!(divider_buggy, divider_correct);
```

This breaks behaviour in time-sensitive PIO state machines for protocols like USB.

ericseppanen:
The current implementation of `calculate_pio_clock_divider` only computes an integer divider, which is a missed opportunity since the return type explicitly represents an 8-bit fractional part.

This implementation ensures that the fractional part of the clock divider is accurately calculated. It uses additional u32 division/mod operations; I think this is preferable to one u64 division, but that would work as well. I don't have a great understanding of how division works on rp2040, so any advice is welcome.

 Other changes:
- Add additional asserts to catch out-of-bounds results.
- Make the version that accepts the system clock as a parameter `const` and public, so if anyone wants to hardcode the system frequency they can avoid any runtime computation.
- Remove the `inline` attributes because the function has grown quite a bit.
- Add unit tests to verify that the math works.

Fixes #5048.

### DIFF
--- a/embassy-rp/src/pio_programs/clock_divider.rs
+++ b/embassy-rp/src/pio_programs/clock_divider.rs
@@ -29,7 +29,7 @@ pub fn calculate_pio_clock_divider(target_hz: u32) -> fixed::FixedU32<U8> {
 /// A fixed-point divider value suitable for use in a PIO state machine configuration
 pub const fn calculate_pio_clock_divider_value(sys_hz: u32, target_hz: u32) -> fixed::FixedU32<U8> {
     // Requires a non-zero frequency
-    assert!(target_hz > 0, "PIO clock frequency cannot be zero");
+    core::assert!(target_hz > 0);
 
     // Compute the integer and fractional part of the divider.
     // Doing it this way allows us to avoid u64 division while
@@ -41,9 +41,9 @@ pub const fn calculate_pio_clock_divider_value(sys_hz: u32, target_hz: u32) -> f
     let result = integer << 8 | frac;
 
     // Ensure the result will fit in 16+8 bits.
-    assert!(result <= 0xffff_ff, "pio clock divider too big");
+    core::assert!(result <= 0xffff_ff);
     // The clock divider can't be used to go faster than the system clock.
-    assert!(result >= 0x0001_00, "pio clock divider cannot be less than 1");
+    core::assert!(result >= 0x0001_00);
 
     fixed::FixedU32::from_bits(result)
 }


ericseppanen:
I hit compile errors with `defmt::assert!` in the `const fn`. Resolved by using `core::assert` without a string literal instead; not sure if that's the right approach.  I could add an error enum, or split this into three functions (runtime-with-defmt, const-without-defmt, and math-inner), but maybe that would be overkill?

Or I could just drop the public const fn idea entirely, and this will always be calculated at runtime.

katazina0:
> I hit compile errors with `defmt::assert!` in the `const fn`. Result by using `core::assert` without a string literal instead; not sure if that's the right approach. I could add an error enum, or split this into three functions (runtime-with-defmt, const-without-defmt, and math-inner), but maybe that would be overkill?
> 
> Or I could just drop the public const fn idea entirely, and this will always be calculated at runtime.

Could two functions work, one const and one not? That way things could be asserted at compile time and if needed, at runtime too. Would probably be a good idea to gate defmt asserts behind the "defmt" feature flag as well.

ericseppanen:
> Could two functions work, one const and one not? 

Perhaps, with some code repetition. There are many possible permutations, so I'm hoping to hear what style the maintainers prefer.

> Would probably be a good idea to gate defmt asserts behind the "defmt" feature flag as well.

This is already handled by `embassy_rp`'s `assert` macro: https://github.com/embassy-rs/embassy/blob/c33a86190c3602ba23a6d5757caaa86b9bcea626/embassy-rp/src/fmt.rs#L10-L19

embassy-rp/src/pio_programs/clock_divider.rs
```diff
@@ -14,12 +13,62 @@ use crate::clocks::clk_sys_freq;
...
+/// * `sys_hz` - The system clock frequency in Hz
+/// * `target_hz` - The desired PIO clock frequency in Hz
+///
+/// # Returns
+///
+/// A fixed-point divider value suitable for use in a PIO state machine configuration
+pub const fn calculate_pio_clock_divider_value(sys_hz: u32, target_hz: u32) -> fixed::FixedU32<U8> {
     // Requires a non-zero frequency
-    assert!(target_hz > 0, "PIO clock frequency cannot be zero");
+    core::assert!(target_hz > 0);
```
Dirbaio, line 32: use `assert!` so panic messages are properly logged with defmt if enabled.

ericseppanen: Thanks for having a look!

Unfortunately, `assert!` will not compile in a `const fn` if the `defmt` feature is enabled. This is what I meant in a previous comment:
> I hit compile errors with defmt::assert! in the const fn. Resolved by using core::assert without a string literal instead; not sure if that's the right approach. I could add an error enum, or split this into three functions (runtime-with-defmt, const-without-defmt, and math-inner), but maybe that would be overkill?
> 
> Or I could just drop the public const fn idea entirely, and this will always be calculated at runtime.

Dirbaio: ah my bad, this is what i get for not reading. Thanks!

ericseppanen: I'm happy to code it whatever way you like, I'm just not sure what the expected workaround should be. If I can't have `feature=defmt` + `const fn`, then I guess I need to either not write `const fn`, or remove support for `defmt` in these asserts.

Or I can search for a way to make it use `core::assert` in const context but `defmt::assert` at runtime. Not sure that's possible though.

PR #5066, opened 2025-12-14
stm32: block stop2 for wle and misc fixes

xoviat:
alternative to #5060 and some other fixes. 

unconditionally block stop2 for wle if any peripheral is held.

### DIFF
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -129,7 +129,7 @@ impl<'d> Drop for I2CDropGuard<'d> {
             x.set_as_disconnected()
         }
 
-        self.info.rcc.disable();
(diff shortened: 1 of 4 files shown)


xoviat:
bender run

liebman:
What are we going to do with peripheral specified stop mode limits? https://github.com/embassy-rs/stm32-data/blob/main/stm32-data-gen/src/low_power.rs#L5C1-L42C1

xoviat:
Sorry for the delays replying, keep getting pulled away. Anyway, the logic is now that for wle, if enable_without_stop is called it will block stop2 while the peripheral is held if the stop mode is stop2 or stop1, but stop1 for example will not be blocked for i2c until a method is called.

xoviat:
@liebman can you discuss on matrix?

PR #5067, opened 2025-12-14
embassy-sync: Update `Pipe::try_write` docs

MJE10:
The description was probably copied from `Pipe::try_read`. However, when writing, an error is produced if the `Pipe` is full, not if it is empty.


lulf (approved):
Thanks!

PR #5068, opened 2025-12-14
WPAN: 1st draft of linklayer_plat

leftger:
Added embassy-time for delay.

- Also added NVIC for task management
- Need to solve for RNG

### DIFF
--- a/embassy-stm32-wpan/src/wba/linklayer_plat.rs
+++ b/embassy-stm32-wpan/src/wba/linklayer_plat.rs
@@ -82,7 +82,10 @@ use core::sync::atomic::{AtomicBool, AtomicI32, AtomicPtr, AtomicU32, Ordering};
 use cortex_m::asm::{dsb, isb};
 use cortex_m::interrupt::InterruptNumber;
 use cortex_m::peripheral::NVIC;
-use cortex_m::register::{basepri, primask};
+use cortex_m::register::basepri;
+use critical_section;
+#[cfg(feature = "defmt")]
+use defmt::trace;
 use embassy_stm32::NVIC_PRIO_BITS;
 use embassy_time::{Duration, block_for};
 
@@ -123,7 +126,6 @@ static RADIO_CALLBACK: AtomicPtr<()> = AtomicPtr::new(ptr::null_mut());
 static LOW_ISR_CALLBACK: AtomicPtr<()> = AtomicPtr::new(ptr::null_mut());
 
 static IRQ_COUNTER: AtomicI32 = AtomicI32::new(0);
-static PRIMASK_SNAPSHOT: AtomicU32 = AtomicU32::new(0);
 
 static PRIO_HIGH_ISR_COUNTER: AtomicI32 = AtomicI32::new(0);
 static PRIO_LOW_ISR_COUNTER: AtomicI32 = AtomicI32::new(0);
@@ -136,6 +138,10 @@ static RADIO_SLEEP_TIMER_VAL: AtomicU32 = AtomicU32::new(0);
 
 static PRNG_STATE: AtomicU32 = AtomicU32::new(0);
 static PRNG_INIT: AtomicBool = AtomicBool::new(false);
+
+// Critical-section restore token for IRQ enable/disable pairing.
+// Only written when the IRQ disable counter transitions 0->1, and consumed when it transitions 1->0.
+static mut CS_RESTORE_STATE: Option<critical_section::RestoreState> = None;
 
 unsafe extern "C" {
     static SystemCoreClock: u32;
@@ -281,6 +287,19 @@ pub unsafe fn run_radio_sw_low_isr() {
 //   */
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn LINKLAYER_PLAT_ClockInit() {
+    //   uint32_t linklayer_slp_clk_src = LL_RCC_RADIOSLEEPSOURCE_NONE;
+    //
+    //   /* Get the Link Layer sleep timer clock source */
+    //   linklayer_slp_clk_src = LL_RCC_RADIO_GetSleepTimerClockSource();
+    //   if(linklayer_slp_clk_src == LL_RCC_RADIOSLEEPSOURCE_NONE)
+    //   {
+    //     /* If there is no clock source defined, should be selected before */
+    //     assert_param(0);
+    //   }
+    //
+    //   /* Enable AHB5ENR peripheral clock (bus CLK) */
+    //   __HAL_RCC_RADIO_CLK_ENABLE();
+    trace!("LINKLAYER_PLAT_ClockInit: get_slptmr_value");
     let _ = link_layer::ll_intf_cmn_get_slptmr_value();
 }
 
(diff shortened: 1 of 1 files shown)


leftger:
@xoviat Here's some draft code. Let me know if it aligns to what you're looking for otherwise we can iterate upon this 😄

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -1,645 +1,417 @@
...
-//   /* Exit critical section */
-//  __set_PRIMASK(primask_bit);
-//
-// }
-//
-// /**
-//   * @brief  Link Layer assertion API
-//   * @param  condition: conditional statement to be checked.
-//   * @retval None
-//   */
```
xoviat, line 158: Can you copy the documentation over to the new functions as well?

leftger:
@xoviat apologies for the delay. Let me know if you need further changes.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
 //   * @retval None
 //   */
-// void LINKLAYER_PLAT_NotifyWFIEnter(void)
-// {
-//   /* Check if Radio state will allow the AHB5 clock to be cut */
-//
-//   /* AHB5 clock will be cut in the following cases:
-//    * - 2.4GHz radio is not in ACTIVE mode (in SLEEP or DEEPSLEEP mode).
-//    * - RADIOSMEN and STRADIOCLKON bits are at 0.
-//    */
```
xoviat, line 192: The inline comments should also be preserved.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
-//   if((LL_PWR_GetRadioMode() != LL_PWR_RADIO_ACTIVE_MODE) ||
-//      ((__HAL_RCC_RADIO_IS_CLK_SLEEP_ENABLED() == 0) && (LL_RCC_RADIO_IsEnabledSleepTimerClock() == 0)))
```
xoviat, line 194: These statements seem like they would use the pac to query these values. Have you looked into what this checks?

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
-//     while ( LL_RCC_HSE_IsReady() == 0);
-// #endif /* CFG_SCM_SUPPORTED */
-//   }
-//   else
-//   {
-//     /* Disable RADIO baseband clock (active CLK) */
-//     HAL_RCCEx_DisableRadioBBClock();
-//   }
-// }
-//
```
xoviat, line 246: This seems to be something like `<T as RccPerpheral>::enable_and_reset()`. We can add a new radio peripheral to embassy-stm32 that allows wpan to enable/disable these clocks, but there should at least be an assertion that the radio bb clock is enabled here.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn LINKLAYER_PLAT_GetRNG(ptr_rnd: *mut u8, len: u32) {
+    if ptr_rnd.is_null() || len == 0 {
+        return;
+    }
+
+    for i in 0..len {
+        let byte = (prng_next() >> ((i & 3) * 8)) as u8;
+        ptr::write_volatile(ptr_rnd.add(i as usize), byte);
+    }
```
xoviat, line 500: Is this a software rng that you wrote? Does the hardware rng currently work on wba? Why not use that?

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn LINKLAYER_PLAT_EnableIRQ() {
+    if counter_release(&IRQ_COUNTER) {
+        let snapshot = PRIMASK_SNAPSHOT.swap(0, Ordering::Relaxed);
+        if snapshot != 0 {
+            cortex_m::interrupt::disable();
+        } else {
+            cortex_m::interrupt::enable();
+        }
+    }
```
xoviat, line 597: This should be done with `critical_section::release(restore_state)`. If there is no restore state, then it should be a no-op.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
```diff
@@ -72,304 +69,398 @@ fn test_fn() {}
...
-//   irq_counter ++;
-// }
-//
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn LINKLAYER_PLAT_DisableIRQ() {
+    if counter_acquire(&IRQ_COUNTER) {
+        let snapshot = if primask::read().is_active() { 1 } else { 0 };
+        PRIMASK_SNAPSHOT.store(snapshot, Ordering::Relaxed);
+    }
+    cortex_m::interrupt::disable();
```
xoviat: This should be done with `critical_section::acquire()`, and the restore_state struct should be saved to a static mut.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
leftger: Excellent question: It has to do with the ownership of the RNG. We don't have any assurances of where in the code flow the RNG would be initialized. I was thinking about this too. I don't have a mental model of how we can momentarily take ownership of the RNG and then immediately relinquish it. If you have a suggestion of how to do this please let me know.

leftger: This code was meant to be more of a placeholder TBH.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
leftger: I think it has to do with the power management status of the WBA. I'll investigate further.

embassy-stm32-wpan/src/wba/linklayer_plat.rs
leftger: I agree but I think this isn't currently possible until a radio peripheral has been added to `embassy-stm32`

leftger:
OK I addressed most things except:

- Radio clock assertion: I think it's gonna need a radio peripheral added to `embassy-stm32`. I don't know how useful an assertion is going to be.
- Power radio modes still need to be resolved (right now I'm unclear on how to pull this)
- RNG: How do I take temporary ownership of the RNG?

As always comments are welcome @xoviat

leftger:
Right now the RNG doesn't fully work on the WBA6X series:

https://github.com/leftger/embassy/blob/c33a86190c3602ba23a6d5757caaa86b9bcea626/embassy-stm32/src/rng.rs#L103

PR #5075, opened 2025-12-15
stm32wb stop2 low power fix

Avtopolet:
I encountered a problem with the stm32wb55cgu6 controller. When using a low-power executor, the controller wouldn't enter stop2, and the consumption was 2 mA. After some investigation, I found that the semaphores were still enabled. Now the consumption has dropped to 250 µA (with debug enabled). Perhaps I need to revert the configure_stop_stm32wb function back to using semaphores.


xoviat:
don't worry about the failures, I can fix them

PR #5076, opened 2025-12-15
embassy-executor: add pend_notify hook

### LINKED ISSUE #5063: Feature request: RunQueue::dequeue_all should perform the loop operation until the queue is truly empty
I wrote a lowpowerExecutor according to embassy-stm32 `mod low-power`. Before WFE, I configure the stop mode as needed, and after WFE, I reconfigure the clock tree. However, many times the queue is not empty, resulting in many meaningless stop entries and clock restoration operations.

xoviat:
closes #5063


xoviat:
@fan2nd does this solve the problem?

fan2nd:
This is exactly what I want. But I tend to make the modifications only in embassy-executor, use `pub(crate)` instead of `pub`.

PR #5078, opened 2025-12-16
[embassy-nrf] reset peripheral pins on Drop for TWIM and UARTE

### LINKED ISSUE #5051: [NRF52832] BufferedUarte does not go back to zero power consumption after drop
Instantiating a BufferedUarte even once ruins the low power of the NRF52832

Image

Example constructor:

```rust

Some(BufferedUvarte::new(
                    p.UARTE0,
                    p.TIMER1,
                    p.PPI_CH0,
                    p.PPI_CH1,
                    p.PPI_GROUP0,
                    p.P0_18,
                    p.P0_17,
                    Irqs,
                    uart_config,
                    unsafe { &mut CELLULAR_TX_BUF },
                    unsafe { &mut CELLULAR_RX_BUF },
                ))
```

After dropping the instance the power consumption remains high

Something is broken in the driver

Kezii:
fixes #5051 

This is probably more aggressive than what's strictly needed to address the mentioned bug, but this ensures that the coredump of the peripheral is the same after drop as it was before initialization (at least for what's concerned in the bug's code)

it should be harmless anyway

### DIFF
--- a/embassy-nrf/src/twim.rs
+++ b/embassy-nrf/src/twim.rs
@@ -742,8 +742,6 @@ impl<'a> Drop for Twim<'a> {
             w.set_connect(nrf_pac::shared::vals::Connect::DISCONNECTED);
         });
 
-        r.frequency().write(|w| w.set_frequency(Frequency::K250));
-
         trace!("twim drop: done");
     }
 }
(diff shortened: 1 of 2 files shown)


Dirbaio:
It's almost harmless, but it does increase code size a bit. Could you identify which register writes are needed for the low-power fix and do just those? Thanks!

Kezii:
this is the minimal change that fixes the power consumption bug

Kezii:
apparently not, the bug is fixed in the test but not in production, hold on

PR #5082, opened 2025-12-16
embassy-net-tuntap: update async-io

Gerharddc:
This is to a fix a transitive dependency issue with RUSTSEC-2024-0384


Gerharddc:
Hmmm, the CI failure in `embassy-stm32-wpan/src/wb55/sub/ble.rs` seems unrelated to this change and does not reproduce on my machine either. I'm not sure what is causing it.

PR #5091, opened 2025-12-16
[iMXRT] dma: fix potential underflow bug

felipebalbi:
Fix a potential underflow when calculating transfer count for small buffers. While at that, also add a constant for maximum DMA transfer size.


felipebalbi:
wrong branch

Issue #5079, opened 2025-12-16
WebUSB example is broken on windows.


rukai:
Here is a timeline of events that I've recreated by bisecting the issue on my RP pico:

https://github.com/embassy-rs/embassy/pull/2894 - April 2024 - initial webusb support (example doesnt work on windows)
https://github.com/embassy-rs/embassy/pull/2894 - May 2024 - fixes webusb example for windows
https://github.com/embassy-rs/embassy/pull/3682 - Dec 2024 - breaks webusb example for all OS's
https://github.com/embassy-rs/embassy/pull/4070 - April 2025 - fixes webusb example for non-windows OS's
today - webusb example still broken on windows. Using the webusb API from a webpage, the claimInterface method blocks forever. Also, the 2 interfaces show up in device manager under "Other Devices", complaining that there are no drivers installed for this device.

I'm still learning how all this usb stuff works, so if anyone knows what the correct fix is please jump in, otherwise I'll keep poking at this problem.

rukai:
I've found that disabling composite_iad resolves the issue, but I don't think that's a good solution since webusb should still work with composite devices.

rukai:
I've found that setting up the MS OS descriptor as:
```rust
    builder.msos_descriptor(windows_version::WIN8_1, 0);
    builder.msos_writer().configuration(0);
    builder.msos_writer().function(InterfaceNumber(0));
    builder.msos_feature(msos::CompatibleIdFeatureDescriptor::new("WINUSB", ""));
    builder.msos_feature(msos::RegistryPropertyFeatureDescriptor::new(
        "DeviceInterfaceGUIDs",
        msos::PropertyData::RegMultiSz(DEVICE_INTERFACE_GUIDS),
    ));
    builder.msos_writer().function(InterfaceNumber(1));
    builder.msos_feature(msos::CompatibleIdFeatureDescriptor::new("WINUSB", ""));
    builder.msos_feature(msos::RegistryPropertyFeatureDescriptor::new(
        "DeviceInterfaceGUIDs",
        msos::PropertyData::RegMultiSz(DEVICE_INTERFACE_GUIDS),
    ));
```
Makes the device show up correctly in device manager instead of being listed under "Other Devices"

This change matches the MS OS docs for composite devices:

<img width="788" height="569" alt="Image" src="https://github.com/user-attachments/assets/b7ebc252-4849-4fdf-bc83-59050904d64a" />

[...]

PR #5093, opened 2025-12-17
Fix rp webusb example on windows

rukai:
closes https://github.com/embassy-rs/embassy/issues/5079

Refer to the ticket for more details.

Fix is to follow the composite device MS OS structure [documented officially here](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/microsoft-os-2-0-descriptors-specification), but I also found [this more concrete example](https://github.com/hathach/tinyusb/discussions/823#discussioncomment-7473503) really helpful.

That is, we use the configuration and function subheaders to define just a specific interface/function and then tell windows to use the WINUSB driver for just those interface/functions.

To enable this structure I needed to remove the `"device features must be added before the first configuration subset"` assertion, but according to the MS OS specs, that assertion is incorrect.

I also renamed the vendor_code field and setter to ms_vendor_code (**breaking change**) as that is closer to the upstream field name `bMS_VendorCode` AND it helps to make it clear that the meaning is entirely unrelated to the vendor code used by the webusb BOS and the vendor code as in PID/VID.
But I'm happy to revert the setter change and/or the vendor_code field change if needed.

And finally, I needed to clear out windows' cache of the old broken descriptors from the example.
What I did was:
1. Download https://www.nirsoft.net/utils/usb_devices_view.html
2. Find the relevant entries
3. right click "uninstall selected devices"

Please perform these steps if you are testing both the before and after of this change, or have previously run the example.

### DIFF
--- a/embassy-usb/src/msos.rs
+++ b/embassy-usb/src/msos.rs
@@ -110,10 +110,6 @@ impl<'d> MsOsDescriptorWriter<'d> {
             !self.is_empty(),
             "device features may only be added after the header is written"
         );
-        assert!(
-            self.config_mark.is_none(),
-            "device features must be added before the first configuration subset"
-        );
         self.write(desc);
     }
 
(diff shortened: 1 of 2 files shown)


Dirbaio:
Thanks for the PR. Yes, please remove the "ms_vendor_code" naming change.

rukai:
> Thanks for the PR. Yes, please remove the "ms_vendor_code" naming change.

Reverted.

Dirbaio:
thanks :rocket:

PR #5095, opened 2025-12-17
mspm0: only cfg guard time driver in peripherals struct, but generate singleton types always

i509VCB:
Also remove some of the useless singletons which are generated.

### DIFF
--- a/embassy-mspm0/build.rs
+++ b/embassy-mspm0/build.rs
@@ -1,5 +1,5 @@
 use std::cmp::Ordering;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


embassy-mspm0/build.rs
```diff
@@ -323,26 +324,18 @@ fn get_singletons(cfgs: &mut common::CfgSet) -> Vec<Singleton> {
             });
         }
 
-        let mut signals = BTreeSet::new();
+        // TODO: Why does C1104 generate duplicate pins?
+        let mut gpio_singletons = BTreeSet::new();
```
i509VCB: Need to look at this

PR #5096, opened 2025-12-17
[MCXA] Add watchdog support with interrupt and reset feature

MathisDeroo:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/75

### DIFF
--- a/embassy-mcxa/src/wwdt.rs
+++ b/embassy-mcxa/src/wwdt.rs
@@ -66,9 +66,6 @@ impl<'d> Watchdog<'d> {
 
         let watchdog = Self { _peri, info };
 
-        // ensure watchdog is disabled
-        watchdog.disable();
-
         let base_frequency = crate::clocks::with_clocks(|clocks| {
             // Ensure clk_1m is active at the required power level
             clocks.ensure_clk_1m_active(&crate::clocks::PoweredClock::NormalEnabledDeepSleepDisabled)
@@ -82,11 +79,6 @@ impl<'d> Watchdog<'d> {
         // Can't use enable_and_reset API here because WWDT doesn't have a reset signal.
         let mrcc = unsafe { pac::Mrcc0::steal() };
         mrcc.mrcc_glb_cc0().modify(|_, w| w.wwdt0().enabled());
-
-        crate::pac::Interrupt::WWDT0.unpend();
-        unsafe {
-            crate::pac::Interrupt::WWDT0.enable();
-        }
 
         let timeout_cycles = (frequency as u64 * config.timeout.as_micros()) / 1_000_000;
 
@@ -120,6 +112,11 @@ impl<'d> Watchdog<'d> {
 
         watchdog.lock_oscillator();
 
+        crate::pac::Interrupt::WWDT0.unpend();
+
+        // Safety: `_irq` ensures an Interrupt Handler exists.
+        unsafe { crate::pac::Interrupt::WWDT0.enable() };
+
         Ok(watchdog)
     }
 
@@ -147,13 +144,6 @@ impl<'d> Watchdog<'d> {
     /// Function is blocking until the watchdog is actually started.
     fn enable(&self) {
         self.info.mod_().modify(|_, w| w.wden().run());
-        while self.info.tc().read().count() == 0xFF {}
-    }
-
-    /// Disable the watchdog timer.
-    /// Function is blocking until the watchdog is actually stopped.
-    fn disable(&self) {
-        self.info.mod_().modify(|_, w| w.wden().stop());
         while self.info.tc().read().count() == 0xFF {}
     }
 
@@ -207,10 +197,16 @@ impl Handler<typelevel::WWDT0> for InterruptHandler {
         let wwdt = unsafe { &*pac::Wwdt0::ptr() };
 
         if wwdt.mod_().read().wdtof().bit_is_set() {
+            #[cfg(feature = "defmt")]
+            defmt::trace!("WWDT0: Timeout occurred");
+
             wwdt.mod_().modify(|_, w| w.wdtof().set_bit());
         }
 
         if wwdt.mod_().read().wdint().bit_is_set() {
+            #[cfg(feature = "defmt")]
+            defmt::trace!("WWDT0: Warning interrupt");
+
             wwdt.mod_().modify(|_, w| w.wdint().clear_bit_by_one());
         }
     }


MathisDeroo:
Hi @felipebalbi @jamesmunns, FYI here is the WWDT driver support. Please provide your feedback, thanks !

jamesmunns:
@MathisDeroo you can run `./fmtall.sh` in the root of the embassy repo to fix the formatting issues!

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
+//! Windowed Watchdog Timer (WWDT) driver for MCXA microcontrollers.
+//!
+//! The WWDT is a hardware timer that can reset the system or generate an interrupt if the software fails to
+//! periodically "feed" the watchdog within a specified time window. This helps detect
+//! and recover from software failures or system hangs.
+//! The FRO12M provides a 1 MHz clock (clk_1m) used as WWDT0 independant clock source. This clock is / 4 by an internal fixed divider.
+//!
```
felipebalbi: nit: these two lines should be swapped.

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+        // Set the WDPROTECT bit to false after the Feed Sequence (0xAA, 0x55)
+        self.set_protect_mode(false);
+    }
+
+    /// Feed the watchdog to prevent timeout.
+    ///
+    /// This must be called periodically before the timeout period expires to prevent
+    /// the watchdog from triggering a reset or interrupt.
+    pub fn feed(&self) {
+        //TBD To avoid an unintended interrupt, it is a good practice to disable interrupts around a feed sequence
```
jamesmunns: you can do this:

```rust
critical_section::with(|_cs| {
  self.info.feed().write(|w| unsafe { w.feed().bits(0xAA) });
  self.info.feed().write(|w| unsafe { w.feed().bits(0x55) });
});
```

MathisDeroo:
> @MathisDeroo you can run `./fmtall.sh` in the root of the embassy repo to fix the formatting issues!

Why there is always this issue even when I run cargo fmt ? ❓ ❓

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    ///               will be generated at this time before the actual timeout. If `None`, the
+    ///               watchdog will directly reset the system on timeout.
+    ///
+    pub fn start(&mut self, period: Duration, warning: Option<Duration>) {
+        let timeout_cycles = (self.frequency as u64 * period.as_micros()) / 1_000_000;
+
+        // Ensure the value fits in u32 and is within valid range
+        // Writing a value below FFh causes 00_00FFh to load into the register. Therefore, the minimum timeout interval is TWDCLK X 256 X 4.
+        assert!(timeout_cycles <= 0xFFFFFF, "Timeout too large for watchdog counter");
+        assert!(timeout_cycles > 0xFF, "Timeout too small for watchdog counter");
```
felipebalbi: looks like `start` should return a `Result`

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+        self.set_timeout_value(self.load_value);
+
+        //Windows value is set to max at reset for no effect.
+
+        if let Some(warning_value) = warning {
+            let warning_cycles = (self.frequency as u64 * warning_value.as_micros()) / 1_000_000;
+            assert!(
+                warning_cycles <= 0x3FF,
+                "Warning value too large. At 1MHz, warning value must be <= 4.092ms."
+            );
```
felipebalbi: in general terms, these asserts should be returned to the caller as an `Err` variant of the `Result` type. `assert` will cause a `panic!()`.

jamesmunns:
@MathisDeroo embassy has a somewhat quirky formatting setup, since there are so many different projects and examples. For now, using the script I mentioned manually will do it the Right™️ way.

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// Feed the watchdog to prevent timeout.
+    ///
+    /// This must be called periodically before the timeout period expires to prevent
+    /// the watchdog from triggering a reset or interrupt.
+    pub fn feed(&self) {
+        //TBD To avoid an unintended interrupt, it is a good practice to disable interrupts around a feed sequence
+        self.info.feed().write(|w| unsafe { w.feed().bits(0xAA) });
+        self.info.feed().write(|w| unsafe { w.feed().bits(0x55) });
```
felipebalbi: how about wrapping with a `critical_section`, then?

```suggestion
        critical_section::with(|_| {
            self.info.feed().write(|w| unsafe { w.feed().bits(0xAA) });
            self.info.feed().write(|w| unsafe { w.feed().bits(0x55) });
        });
```

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// Enable or disable the watchdog timer.
+    ///
+    /// # Arguments
+    ///
+    /// * `bit` - `true` to enable (run), `false` to disable (stop).
```
felipebalbi: perhaps mention here the function blocks until done?

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn enable(&self, bit: bool) {
+        self.info
+            .mod_()
+            .modify(|_, w| if bit { w.wden().run() } else { w.wden().stop() });
+        while self.info.tc().read().count() == 0xFF {}
+    }
```
felipebalbi: I'm not sure I like this raw boolean. Maybe define an enum or split into `enable()` and `disable()` APIs?

Additionally, you can get rid of the `if` by using the `bit()` method:

```rust
.modify(|_, w| w.wden().bit(bit));
```

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// * `bit` - `true` for threshold mode (protected), `false` for flexible mode.
+    fn set_protect_mode(&self, bit: bool) {
+        self.info.mod_().modify(|_, w| {
+            if bit {
+                w.wdprotect().threshold()
+            } else {
+                w.wdprotect().flexible()
+            }
+        });
+    }
```
felipebalbi: same comment as before. Either add an enum or split into protect and unprotect methods. Bit comment is valid too.

examples/mcxa/src/bin/wwdt_interrupt.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    let mut watchdog = Watchdog::new(p.WWDT0, Irqs);
+    let mut led = Output::new(p.P3_18, Level::High, DriveStrength::Normal, SlewRate::Fast);
+
+    // Set the LED high for 2 seconds so we know when we're about to start the watchdog
+    led.toggle();
+    Timer::after_secs(2).await;
+
+    // Set to watchdog to generate interrupt if it's not fed within 1.05 seconds, and start it.
+    // The warning interrupt will trigger 4us before the timeout.
```
felipebalbi: should it be `4ms`? You're passing 4000 micros.

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+        } else {
+            self.enable_reset();
+        }
+
+        self.lock_oscillator();
+        self.enable(true);
+        self.feed();
+
+        // Set the WDPROTECT bit to false after the Feed Sequence (0xAA, 0x55)
+        self.set_protect_mode(false);
```
felipebalbi: Why is the watchdog left unprotected? Would like to understand your reasoning. It seems to me that it would be best to keep it protected and ensure that `TC[COUNT]` is changed only at appropriate times.

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+    // Start the watchdog timer with the specified timeout period.
+    ///
+    /// # Arguments
+    ///
+    /// * `period` - The timeout period after which the watchdog will trigger
+    /// * `warning` - Optional warning interrupt period before timeout. If `Some`, an interrupt
+    ///               will be generated at this time before the actual timeout. If `None`, the
+    ///               watchdog will directly reset the system on timeout.
+    ///
+    pub fn start(&mut self, period: Duration, warning: Option<Duration>) {
```
felipebalbi: Another way to do this would be to pass both `period` and `warning` as part of a `Config` struct to the constructor. Then pre-program the watchdog with everything you need, during construction and leave start with the single responsibility of starting it, that is: `enable(); feed();`

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,220 @@
...
+            } else {
+                w.wdprotect().flexible()
+            }
+        });
+    }
+
+    /// Enable interrupt mode.
+    fn enable_interrupt(&self) {
+        self.info.mod_().modify(|_, w| w.wdreset().interrupt());
+    }
```
felipebalbi, line 158: when you split the various fields of a register to different methods, they lose their atomicity. That is, if you were more conservative with how you're splitting these methods, we could write to `mod_()` with one atomic operation. It looks like passing desired configuration to the constructor through `Config` would easily allow that to be true.

embassy-mcxa/src/wwdt.rs
MathisDeroo: I found that setting the watchdog to threshold mode (protected) has a big constraint. Indeed per the RM "In addition to the above two conditions, when MOD[WDPROTECT] = 1, the watchdog counter must not be greater than the value of WARNINT[WARNINT] for a correct feed to occur".

I had quite few trouble to find the root cause of my interrupt/reset event when protected mode was used, I imagine it would be the same for a customer using it. Also I don't think it add lot of value to the driver.

embassy-mcxa/src/wwdt.rs
MathisDeroo: Added the config struct with period (Duration) and warning (Option<Duration>). As all the configuration is now done in the constructor, I removed the timeout value and frequency from the WWDT structure as it is not used afterward. Same as before with the start() function, the watchdog constructor returns Result

felipebalbi:
@MathisDeroo, I have updated your PR to today's `main` branch and also fixed formatting and some warnings that were left over.

MathisDeroo:
> @MathisDeroo, I have updated your PR to today's `main` branch and also fixed formatting and some warnings that were left over.

Thanks @felipebalbi ! Do you see anything else missing preventing merging the PR ?

felipebalbi:
> > @MathisDeroo, I have updated your PR to today's `main` branch and also fixed formatting and some warnings that were left over.
> 
> Thanks @felipebalbi ! Do you see anything else missing preventing merging the PR ?

from my point of view, it looks okay. @jamesmunns, any comments on WWDT driver?

felipebalbi (approved):
LGTM

embassy-mcxa/src/wwdt.rs
```diff
@@ -0,0 +1,213 @@
...
+    TimeoutTooLarge,
+    WarningTooLarge,
+}
+
+/// WWDT configuration
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Config {
+    /// The timeout period after which the watchdog will trigger
+    pub timeout: Duration,
+    pub warning: Option<Duration>,
```
jamesmunns, line 32: Currently, `embassy-time` is an optional dependency, which means that technically if it was omitted, we would get compilation errors because Instant and Duration rely on it. The types also only fully make sense with an `embassy-time-driver` impl, because that defines the tick rate.

In my opinion, we should make embassy-time mandatory (since we will practically always have it), and if someone wants to allow skipping it in the future, we can figure out how to gate things properly then.

I won't block the review on this, but I'll file a follow-up issue to discuss.

examples/mcxa/src/bin/wwdt_interrupt.rs
```diff
@@ -0,0 +1,61 @@
...
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let config = Config::default();
+    let p = hal::init(config);
+
+    defmt::info!("Watchdog example");
+
+    let wwdt_config = hal::wwdt::Config {
+        timeout: Duration::from_millis(1050),
+        warning: Some(Duration::from_micros(4000)),
```
jamesmunns, line 27: ~~Should we be checking if `warning <= timeout`? It seems like it might be worth at least warning about, maybe returning an error to catch accidental user error where they THINK they will get a warning but don't.~~

Just kidding, I realized it's `millis` and `micros` :)

embassy-mcxa/src/wwdt.rs
jamesmunns: Opened https://github.com/embassy-rs/embassy/issues/5189 to track, will fix shortly.

PR #5097, opened 2025-12-17
Adding Basic RTC API 

cesardtamayo:
Add Hardware Agnostic RTC API with unit testing.
The API is being used by examples/smt32wba6/src/bin/rtc.rs

Unit tests added:
* test_rtc_defaults_to_correct_date
* test_rtc_instantiates_correctly
* test_rtc_updates_date_time_correctly


Dirbaio:
Unfortunately I don't think we should do this. It's not embassy-stm32's job to define traits for abstracting over different RTC hardware. 

An RTC trait is something that could go in the `embedded-hal` crate, then `embassy-stm32` could implement that.

cesardtamayo:
> Unfortunately I don't think we should do this. It's not embassy-stm32's job to define traits for abstracting over different RTC hardware.
> 
> An RTC trait is something that could go in the `embedded-hal` crate, then `embassy-stm32` could implement that.

Do you mean the embassy-hal-internal directory? I could implement it the way you are suggesting

Dirbaio:
No, [embedded-hal](https://github.com/rust-embedded/embedded-hal/)

Dirbaio:
I meant `rust-embedded/embedded-hal`, that's where the link in my previous comment goes. This is maintained by the Rust Embedded Working Group, not Embassy.

In general the Embassy project doesn't define traits for abstracting things because they're an "ecosystem-wide" concern. If Embassy defines its own trait and implements it on its HALs, it's not very useful because you still can use it with Embassy only, other non-Embassy HALs won't implement it.

If the trait is in `embedded-hal`, which is from the Rust Embedded WG, a wider part of the ecosystem tends to implement it, which makes it more useful.

I'm not sure how easy it'll be to add an RTC trait to `embedded-hal` though. The datetime from `embassy-stm32` is stm32-specific, other RTCs probably do it differently or are missing features (like day of week, or microseconds)

cesardtamayo:
> I meant `rust-embedded/embedded-hal`, that's where the link in my previous comment goes. This is maintained by the Rust Embedded Working Group, not Embassy.
> 
> In general the Embassy project doesn't define traits for abstracting things because they're an "ecosystem-wide" concern. If Embassy defines its own trait and implements it on its HALs, it's not very useful because you still can use it with Embassy only, other non-Embassy HALs won't implement it.
> 
> If the trait is in `embedded-hal`, which is from the Rust Embedded WG, a wider part of the ecosystem tends to implement it, which makes it more useful.
> 
> I'm not sure how easy it'll be to add an RTC trait to `embedded-hal` though. The datetime from `embassy-stm32` is stm32-specific, other RTCs probably do it differently or are missing features (like day of week, or microseconds)

I'll talk to the team to see if that's worth pursuing. We are trying to add some APIs that we could use in the future, but what you are saying makes sense to me. 
Thank you for the feedback :)

leftger:
This PR is orthogonal to the purposes of Embassy. I'll work with cesardtamayo to work on something more appropriate in the future.

PR #5098, opened 2025-12-17
make stm32wb examples use low-power-pender

xoviat:
allows using the low-power-pender


Dirbaio:
i'd prefer to keep the executor not coupled to the HALs. If all the feature does is set the default executor it's better to tell the user to set it themselves.

xoviat:
the feature sets the executor and the entry. without this, the user has to set the executor and entry every time they use low power.

PR #5107, opened 2025-12-18
[MCXA]: Add SPLL/PLL1 support

jamesmunns:
Adds support for basic "Mode 1" usage of the SPLL/PLL1. This allows for the use of `pll1_clk` or `pll1_clk_div` as a clock source.

This does not impl some of the fancier things (clock monitoring, spread spectrum), nor does it let you swap the core/CPU clock to the SPLL yet, but those can be implemented next.

### DIFF
--- a/embassy-mcxa/src/clocks/mod.rs
+++ b/embassy-mcxa/src/clocks/mod.rs
@@ -978,6 +978,9 @@ impl ClockOperator<'_> {
             return Ok(());
         };
 
+        // Ensure the LDO is active
+        self.ensure_ldo_active();
+
         // match on the source, ensure it is active already
         let res = match cfg.source {
             config::SpllSource::Sosc => self


embassy-mcxa/src/clocks/config.rs
```diff
@@ -143,6 +147,114 @@ pub struct SoscConfig {
...
+#[non_exhaustive]
+pub enum SpllMode {
+    /// Mode 1a does not use the Pre/Post dividers.
+    ///
+    /// `Fout = m_mult x SpllSource`
+    ///
+    /// Both of the following constraints must be met:
+    ///
+    /// * Fout: 275MHz to 550MHz
+    /// * Fout: 4.3MHz to 2x Max CPU Frequency
```
felipebalbi, line 196: is there a typo here somewhere?

jamesmunns:
Screenshot 2025-12-18 at 19 43 24

SPLL wiggles for proof! (from `clkout.rs` example)

embassy-mcxa/src/clocks/config.rs
jamesmunns: There isn't here, but maybe there's room for confusion. When you DON'T use `p_div`~/the pre-divider~, then `Fcco = Fout`. ~I do think 1b and 1c's docs might be wrong tho, lemme check~.

jamesmunns: ah, `Fcco = Fout` when you don't use the *post-divider*, which is `p_div`.

PR #5115, opened 2025-12-19
stm32/sdmmc: call on_drop on sd drop

### LINKED ISSUE #5112: STM32 sdmmc_v2 freezing during reconnect
Working on a device with an STM32H563 and an SD card socket. I have connected the full 4-bit interface between the STM and the card and started to write code based on the new interface introduced by @xoviat in commit [d9c90df](https://github.com/embassy-rs/embassy/commit/d9c90df7578dda10d888941b8df5a2295373ca1f).

Basic code structure:
```rust
async fn main() {
        // ...
        let mut sdmmc = Sdmmc::new_4bit(..., Default::default());
        loop {
            // wait for CD pin (microswitch in SD socket) to go low
            let _ = run_inner(&mut sdmmc).await;
        }
}

async fn run_inner(sdmmc: Sdmmc<'static>) -> Result<...> {
        let mut storage = StorageDevice::new_sd_card(sdmmc, &mut cmd_block, mhz(25))
                .await
                .map_err(|_| "new_sd_card")?;
        // read data from SD card in endless loop
}
```

The code runs perfectly fine when the SD card is present at startup or inserted for the first time. When I unplug the card, I get the following error on 'read()': 'Error::Crc' or `Error::Timeout` and the outer loop starts waiting for the card again. When I insert the card another time, the executor freezes in the following place: `StorageDevice::new_sd_card -> StorageDevice::acquire -> Sdmmc::init_idle -> Sdmmc::clkcr_set_clkdiv -> Sdmmc::wait_idle`.

Setting the interface speed to khz(400) does not change the outcome, neither does using `Sdmmc::new_1bit` instead.

Is there a way to reset the SDMMC interface before reconnecting to prevent this error?

xoviat:
fixes #5112. 

@martin2250 does this solve your issue?

### DIFF
--- a/embassy-stm32-wpan/Cargo.toml
+++ b/embassy-stm32-wpan/Cargo.toml
@@ -41,7 +41,7 @@ log = { version = "0.4.17", optional = true }
 
 cortex-m = "0.7.6"
 heapless = "0.8"
-aligned = "0.4.2"
+aligned = "0.4.3"
 critical-section = "1.2"
 
 bit_field = "0.10.2"
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -196,7 +196,7 @@ volatile-register = { version = "0.2.1" }
 bitflags = "2.10.0"
 
 block-device-driver = { version = "0.2" }
-aligned = "0.4.1"
+aligned = "0.4.3"
 heapless = "0.9.1"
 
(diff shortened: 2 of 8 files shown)


martin2250:
Yes! I switched back and forth a few times and tested different configurations. With this git rev, the card reconnects every time.

Thank you for the quick fix!

PR #5117, opened 2025-12-19
[embassy-nrf] add task shutdown for timer, shut it down on BufferedUarte, fix power consumption

### LINKED ISSUE #5051: [NRF52832] BufferedUarte does not go back to zero power consumption after drop

Kezii:
Apparently my NRF52832 QFAAE chip was affected by errata 78, leading to #5051

https://docs.nordicsemi.com/bundle/errata_nRF52832_Rev2/page/ERR/nRF52832/Rev2/latest/anomaly_832_78.html

The fix is to add support for the (deprecated) shutdown task for the timer and call it on Drop for the BufferedUarte

fixes #5051

### DIFF
--- a/embassy-nrf/src/buffered_uarte/v1.rs
+++ b/embassy-nrf/src/buffered_uarte/v1.rs
@@ -856,6 +856,8 @@ impl<'a> Drop for BufferedUarteRx<'a> {
         let r = self.r;
 
         self.timer.stop();
+        // workaround errata 78
+        self.timer.shutdown();
 
         r.intenclr().write(|w| {
             w.set_rxdrdy(true);
--- a/embassy-nrf/src/timer.rs
+++ b/embassy-nrf/src/timer.rs
@@ -173,6 +173,12 @@ impl<'d> Timer<'d> {
         self.r.tasks_clear().write_value(1)
     }
 
+    /// Shuts down the timer.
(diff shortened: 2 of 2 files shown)


Kezii:
one thing tho: starting the timer in https://github.com/Kezii/embassy/blob/62a00767d6411b497105e5228e62cc7a9e18fac1/embassy-nrf/src/buffered_uarte/v1.rs#L721 might not be necessary

the timer is used as a counter

the datasheet is not clear on that (though it says that it can be started) but nrfx source code doesn't test for the timer to be started for counter mode

not starting the timer fixes the power consumption problem for the buffereduarte and doesn't seem to impact functionality for my testing, but I don't trust pushing this type of change

Dirbaio:

PR #5143, opened 2025-12-27
cache spi status once

xoviat:


JonathanLindsey:
I don't have the context of what this does, but this fixes the issue I was having where the rp235x `blinky_wifi` example was freezing after printing:
```
[DEBUG] waiting for F2 to be ready...
```

PR #4948, opened 2025-11-25
Remove atomic-polyfill with critical-section instead

BjornTheProgrammer:
This prevents a warning like the following from appearing when running `cargo audit -q` in another project.

```
➜  burn git:(embedded-fix) ✗ cargo audit -q --color always
Crate:     atomic-polyfill
Version:   1.0.3
Warning:   unmaintained
Title:     atomic-polyfill is unmaintained
Date:      2023-07-11
ID:        RUSTSEC-2023-0089
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0089
Dependency tree:
atomic-polyfill 1.0.3
└── embassy-rp 0.8.0
```

PR #5147, opened 2025-12-28
[embassy-rp] Fix PIO freeze

### LINKED ISSUE #5146: [embassy-rp] blinky_wifi example freezes or prints TX stalled, bug in PIO
## How to reproduce

Making the additions below to `examples/rp235x/src/bin/blinky_wifi.rs` causes it to freeze. The comments in this diff explain what can be changed to make it rapidly print "TX stalled" instead.

Image

## Versions effected

This bug was introduced by PR #4948 and affects `embassy-rp v0.9.0`.
This is NOT related to issue #5144 which only effected the main branch for a few days 6792854f9c1931a086b0362fc8a1e98943d0c63f...f64ee7a7227f977707d547d5099b8d2831481263 (Dec 24th - Dec 27 2025).

## How I came across this in real life

I came across this bug because I had put a lot of the wifi initialization code in a separate function to abstract it away in my program. I now realize I can can pass `&mut pio` to this function instead of moving `pio` into it. But this is a bug regardless. The added block in the example above is a visually simpler way to describe the bug vs. using a function.

I'll have a PR posted soon.

JonathanLindsey:
Fixes #5146.

The regression was introduced by #4948 which replaced the `atomic-polyfill` dependency with custom code. When the [fetch_sub](https://docs.rs/atomic-polyfill/1.0.3/atomic_polyfill/struct.AtomicU8.html#method.fetch_sub) function was replaced, the subtle detail that that function returned the *previous* value instead of the new one after the subtraction was missed resulting in an off-by-one bug in the `on_pio_drop` function.

An alternate PR or follow up could use the `portable-atomic` crate instead of `atomic-polyfill`.

This PR fixes the bug and renames some variables to make the distinction clearer. I did NOT change to returning the new value in case this code is ever replaced by `portable-atomic` or similar.

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -1232,10 +1232,10 @@ impl<'d, PIO: Instance> Common<'d, PIO> {
         });
         // we can be relaxed about this because we're &mut here and nothing is cached
         critical_section::with(|_| {
-            let val = PIO::state().used_pins.load(Ordering::Relaxed);
+            let old_val = PIO::state().used_pins.load(Ordering::Relaxed);
             PIO::state()
                 .used_pins
-                .store(val | 1 << pin.pin_bank(), Ordering::Relaxed);
+                .store(old_val | 1 << pin.pin_bank(), Ordering::Relaxed);
         });
 
         Pin {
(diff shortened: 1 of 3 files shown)


JonathanLindsey:
@Dirbaio Not sure if this got missed from being posted in the middle of the holidays or something. This is ready for review. No rush though if you're in the middle of other things.

BjornTheProgrammer:
Ahh @JonathanLindsey, I really apologize for the mess up. Regarding on the idea of using `portable-atomic`, that was my original idea as well. Although I made it work with `portable-atomic`, doing so required cfg flags to be put in the entire code base. If you want to continue that approach, you can refer to my original [PR](https://github.com/embassy-rs/embassy/pull/4909).

lulf:
Agreed with review. I've cherry-picked the first commit and pushed.

PR #5149, opened 2025-12-29
cyw: cleanup sdio

xoviat:
more cleanup...

### DIFF
--- a/cyw43/src/sdio.rs
+++ b/cyw43/src/sdio.rs
@@ -1,13 +1,12 @@
 use core::slice;
 
 use aligned::{A4, Aligned};
-use embassy_futures::yield_now;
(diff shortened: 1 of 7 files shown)


cyw43/src/sdio.rs
```diff
@@ -228,144 +218,71 @@ where
...
-            debug!("cmd53 byte write failed");
-        }
-    }
-
-    async fn cmd53_write(&mut self, func: u32, addr: u32, buf: &[u32]) {
-        let mode = if buf.len() >= BLOCK_SIZE {
-            Mode::Block
-        } else {
-            Mode::Byte
-        };
```
xoviat, line 250: This was a bug for certain... the buf.len() is not the block size.

PR #5150, opened 2025-12-29
embassy-rp: change watchdog interface

dobrowolski-lukasz:
Add duration argument to watchdog.feed function.
watchdog.start function's delay keeps duration argument but it's meaning changes to initial watchdog feeding.

This change makes the watchdog more flexible. It allows one to set different initial timeout duration then during normal operation. This allows to tighten the timeouts rather than having to use biggest one.

It would also allow one to set a long timeout before running some long blocking operation.

I've build tested examples/rp and examples/rp325x and ran the watchdog example on rp2040.

### DIFF
--- a/embassy-rp/src/watchdog.rs
+++ b/embassy-rp/src/watchdog.rs
@@ -111,6 +111,11 @@ impl Watchdog {
         self.enable(true);
     }
 
+    /// Stop the watchdog timer
+    pub fn stop(&mut self) {
+        self.enable(false);
+    }
+
     /// Trigger a system reset
     pub fn trigger_reset(&mut self) {
         self.configure_wdog_reset_triggers();


dobrowolski-lukasz:
I've only build tested changes to embassy-boot-rp and examples/boot/{application,bootloader}/rp.

lulf:
Seems a bit weird having to pass the timeout every time to feed(). Could it make sense to have a set_timeout() intead?

dobrowolski-lukasz:
> Seems a bit weird having to pass the timeout every time to feed(). Could it make sense to have a set_timeout() intead?

It's possible by going back to watchdog remembering timeout value of timeout to apply and adding a set_timeout() as you suggested but IHMO that is the weird way.
1. It's not how the HW works(AFAIK most watchdog use countdown).
2. It's weird when the user doesn't use same value every time.

Also, the user can trivially wrap watchdog if they don't want to pass timeout const around. Wrapping HW API is not weird anyway. One probably wants to feed watchdog in a single task anyway.

I'm not the maintainer so I can go with Your way if you demand it but I think it's not the best solution. Let me know if I should do it.

PR #5152, opened 2025-12-30
Reduce footprint by preventing 64bit multiplication

roel0:
This prevents an unneeded u64 multiplication 

Original footprint:
4.7KiB    embassy_stm32 embassy_stm32::rcc::_version::init_pll

After u32 cast:
836B    embassy_stm32 embassy_stm32::rcc::_version::init_pll

### DIFF
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -590,7 +590,7 @@ fn init_pll(instance: PllInstance, config: Option<Pll>, input: &PllInput, voltag
     };
 
     // Calculate the PLL VCO clock
-    let vco_freq = ref_freq * pll.mul;
+    let vco_freq = ref_freq * (pll.mul as u32);
     assert!(vco_freq >= vco_min && vco_freq <= vco_max);
 
     // Calculate output clocks.


psusi2:
Are you sure that this involved u64 multiplication?  It doesn't look like there should have been any.  The impl <Mul> block for Hertz has u32 suffixes on the number's it is multiplying by and Hertz is a wrapper around u32.  It does use a massive 512 arm match though to shift the multiplication factor by one.  I'm guessing that the compiler generated a ton of code for that.  I'm not sure why it uses this massive match expression instead of just adding 1 to the multiplied value.  Maybe that's the problem?

Dirbaio:
> @psusi2  I'm not sure why it uses this massive match expression instead of just adding 1 to the multiplied value

it generates a match because "multiplier = 1 + enum value" doesn't hold in all chips. 

build.rs could check if that's the case and generate a smaller impl, fall back to the match if not. This would help code size a lot. PRs welcome. :)

roel0:
> Are you sure that this involved u64 multiplication? It doesn't look like there should have been any. The impl block for Hertz has u32 suffixes on the number's it is multiplying by and Hertz is a wrapper around u32. It does use a massive 512 arm match though to shift the multiplication factor by one. I'm guessing that the compiler generated a ton of code for that. I'm not sure why it uses this massive match expression instead of just adding 1 to the multiplied value. Maybe that's the problem?

Yeah my explanation was wrong, the cast just prevents the complete match from being inlined it seems

Both Plln and Hertz have valid u32 impl

psusi2:
Weird... I was just checking on the playground what the compiler does with this kind of match expression math.  You would think that it would just take the two arguments and multiply them, but instead it generates a fragment of code for every single match arm, each one optimized to perform the specific constant multiplication, and uses a look up table to find which fragment to jump to based on the value of the enum.

psusi2:
Well look at that.  Simply lifting the operations out of the match expression got the compiler to optimize the code to add 1 then multiply and did away with the whole lookup table.

In other words, instead of having each match arm do arg * const, I did arg * match { => const, => const2, etc }.

PR #5154, opened 2025-12-31
cyw: wip

xoviat:


embassy-stm32/src/sdmmc/mod.rs
```diff
@@ -835,7 +854,7 @@ impl<'d> Sdmmc<'d> {
         self.wait_idle();
         self.clear_interrupt_flags();
 
-        regs.dlenr().write(|w| w.set_datalength(size_of_val(buffer) as u32));
+        regs.dlenr().write(|w| w.set_datalength(buffer.len() as u32));
```
xoviat: It actually appears that this must be set to at least 4 bytes even if two bytes are read, therefore, this change is wrong.

PR #5155, opened 2026-01-01
Add support for fractional PLL multipliers on STMH743ZI

northernpaws:
The STM32H743zi has an option for fractional PLL multipliers that's currently not exposed by the PLL configs or used by the VCO clock calculation. This commit adds support for the fractional multiplier for the STM32H743zi, but could also be applied to any other STM32 that supports them by adding them to the feature gates.


embassy-stm32/src/rcc/h.rs
```diff
@@ -826,7 +830,16 @@ fn init_pll(num: usize, config: Option<Pll>, input: &PllInput) -> PllOutput {
     // be chosen when the reference clock frequency is lower than 2 MHz.
     let wide_allowed = ref_range != Pllrge::RANGE1;
 
+    #[cfg(stm32h743)]
+    let vco_clk = match config.fracn {
+        Some(fracn) => {
+            Hertz::hz((ref_clk.0 as f32 * ((config.mul.to_bits() + 1) as f32 + (fracn as f32 / 8192.0))) as u32)
```
northernpaws, line 836: I'm not a fan of all the casting going on here, but it was the best method I was able to come up with for applying the fractional multiplication. Would love to see if there are any better ideas on how to approach it!

PR #5156, opened 2026-01-01
stm32wl5x low-power both cores (STOP2) also lptim time driver!

liebman:
It's working, but needs lots of cleanup and refactoring:

- LPTIM Time Driver (time-driver-lptim1, time-driver-lptim2, time-driver-lptim3 features) does not use RTC during STOP
- LPTIM1 supports STOP2
- LPTIM2&LPTIM3 suppot STOP1
- stm32wl5x (dual core) moved RCC_CONFIG to SharedData (like Clock)
- working on stm32wlex as well
- enable second core GPIO & DMA clocks so when primary goes to sleeps peripheral is still available
- clocking lptimer with LSI and LSE both work

TODO:
- [x] ~refactor time_driver*.rs into time_driver dir with mod.rs with gp16.rs and lp.rs implementations~
- [x] ~LPTIM2 & LPTIM3 broken on second core~
- [x] ~seperate low_power dual core examples for stm32wl5x~
- [ ] TIM1 & TIM2 broken on second core because Rtc initialization not called in `init_secondary_hw`
- [ ] If the second core will be using a low power timer as time driver the primary core has to enable.configure its clocks!  Maybe there is a way to allow the secondary to configure them.

Nice to Have:
- [ ] try to reduce crazy number of cfg low-power / not _lp-time-driver etc
- [ ] How can we restrict (at least a clear compiler error) using a lp timer outside of specific chips.
- [ ] make HSI work (in theory it should with the right tick frequency choice)

### DIFF
--- a/embassy-stm32/src/hsem/mod.rs
+++ b/embassy-stm32/src/hsem/mod.rs
@@ -312,6 +312,11 @@ impl<T: Instance> HardwareSemaphore<T> {
 #[cfg(all(any(stm32wb, stm32wl5x), feature = "low-power"))]
 pub(crate) fn init_hsem(cs: CriticalSection) {
     rcc::enable_and_reset_with_cs::<crate::peripherals::HSEM>(cs);
+    #[cfg(stm32wb)]
+    unsafe {
+        crate::rcc::REFCOUNT_STOP1 = 0;
+        crate::rcc::REFCOUNT_STOP2 = 0;
+    }
 }
 
 struct State {


embassy-stm32/src/rtc/mod.rs
```diff
@@ -219,6 +219,7 @@ impl Rtc {
         (RtcContainer::new(), RtcTimeProvider::new())
     }
 
+    #[cfg(not(all(feature = "low-power", feature = "_lp-time-driver")))]
```
liebman: I think this needs to be more like `any(not(feature="low-power"), feature = "_lp-time-driver"))`.  And more of the `#[cfg(not(feature = "low-power"))]` in this file need this too.

xoviat:
>  TIM1 & TIM2 broken on second core because Rtc initialization not called in init_secondary_hw

does this break existing functionality?

liebman:
> > TIM1 & TIM2 broken on second core because Rtc initialization not called in init_secondary_hw
> 
> does this break existing functionality?

Nope - just identifying why they won't work.

liebman:
And I think its ready now.

liebman:
that said I only have stm32wle5 and stm32wl55 to test on.

embassy-stm32/src/hsem/mod.rs
```diff
@@ -307,14 +309,9 @@ impl<T: Instance> HardwareSemaphore<T> {
     }
 }
 
-#[cfg(all(stm32wb, feature = "low-power"))]
+#[cfg(all(any(stm32wb, stm32wl5x), feature = "low-power"))]
 pub(crate) fn init_hsem(cs: CriticalSection) {
     rcc::enable_and_reset_with_cs::<crate::peripherals::HSEM>(cs);
-
-    unsafe {
```
liebman, line 314: @xoviat maybe this is an issue for stm32wb?  for stm32wl55 this is in stm32-data low-power:
        (r"^STM32WL55.*:HSEM", StopMode::Stop2),

liebman: yes - that looks to be an issue.  I'll add it back for stm32wb

liebman:
Ok - fixed.

PR #5159, opened 2026-01-03
Embassy-rp: Add set_baudrate() to BufferedUartTx

OBauck:
Also add USB-to-Uart example to rp235x.
This example can also be added to all mcu implementing embassy-usb and BufferedUartTx/Rx.

### DIFF
--- a/examples/rp235x/src/bin/usb_to_uart.rs
+++ b/examples/rp235x/src/bin/usb_to_uart.rs
@@ -1,0 +1,134 @@
+//! This example shows how to use USB (Universal Serial Bus) in the RP235x chip.
+//!
+//! This creates a USB serial port that sends/receives data on uart (USB-to-Uart).
+//!
+//! No specific hardware is specified in this example. If you connect pin 0 (TX) and 1 (RX) you should get the same data back.
+//! The Raspberry Pi Debug Probe (https://www.raspberrypi.com/products/debug-probe/) could be used
+//! with its UART port.
+
+#![no_std]
+#![no_main]
+
+use defmt::{Debug2Format, error, info};
+use embassy_executor::Spawner;
+use embassy_futures::join::join3;
+use embassy_futures::select::{Either, select};
+use embassy_rp::bind_interrupts;
+use embassy_rp::peripherals::{UART0, USB};
+use embassy_rp::uart::{BufferedInterruptHandler, BufferedUart, Config};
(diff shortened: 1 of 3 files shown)


CBJamo:
I wasn't involved in writing this code, but I think it was an intentional API decision to have set_baudrate only appear on BufferedUart.

If you make a BufferedUart, and then split it into Tx and Rx, a set_baudrate on either will effect both, which could be surprising to users who are only looking at the API.

OBauck:
Then I think it is enough to add a description to the function which clearly states this.
IMO having the option to set baudrate after splitting into Tx and Rx is higher priority.
Also it looks like BufferedUartTx/Rx in both embassy-nrf and embassy-mspm0 has this side effect that changing the baudrate on one of them will change it on both.

OBauck:
If you look at the example, then you can see why I want this function on BufferedUartTx (or Rx).
If you know any other way to solve this, then please share your suggestion.
I could use BufferedUart in the example instead of splitting it, but that would make the example much more complex.

i509VCB:
> it looks like BufferedUartTx/Rx in both embassy-nrf and embassy-mspm0 has this side effect that changing the baudrate on one of them will change it on both.

This is intended from discussions on Matrix about this. set_config can do the same thing as well.

OBauck:
Thank you for your comment! Here is a link to the Matrix discussion: https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$lasIdbxBVklguQpUnjKrweKsBxVtlF1JSzoSfK9JZ_A

Ragarnoy:
> Thank you for your comment! Here is a link to the Matrix discussion: https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$lasIdbxBVklguQpUnjKrweKsBxVtlF1JSzoSfK9JZ_A

Shouldn't this information be indicated in the rustdoc of the function ?

examples/rp235x/src/bin/usb_to_uart.rs
```diff
@@ -145,3 +130,24 @@ async fn main(_spawner: Spawner) {
...
+                return;
+            }
+            Ok(sent) => bytes_sent += sent,
+        }
+
+        if bytes_sent == bytes.len() {
+            return;
+        }
+    }
+}
```
i509VCB: Is this not just `embedded_io::Write::write_all`?

OBauck: image

PR #5160, opened 2026-01-03
Update dependencies for embassy-nrf and related (WIP)

BartMassey:
Updates `embassy-nrf` to use `document-features` 0.2.12 (suppresses a warning) and `embassy-time-queue` to use `heapless` 0.9 (fixes resolver issues in some builds). Bumps version to `0.9.1`.


BartMassey:
Am realizing now that these upgrades are arguable point releases for the other embassy crates, which would touch almost everything here. Advice on making this transition correctly is appreciated.

lulf:
I think ideally we'd upgrade heapless across all the crates, not just one of them, to avoid duplicate heapless copies where we can avoid it.

BartMassey:
Agree: might as well upgrade everything that upgrades while we're doing this.

I'm willing to take on the big task of upgrading all the dependencies of all the things if that's cool. TBH I'll probably do it with Claude unless there's some tooling already in place to help with this: it's a lot of error-prone tedium for the most part. Maybe I'll have Claude write the tooling?

I probably should forego upgrading any crates that have breaking changes (except `heapless`) beyond their point release for now? I don't know how big folks want this to get.

lulf:
* Use claude as much as you want, it will be reviewed like any other code
* Don't bump version numbers, we do this via the release tooling which automatically checks for semver breakages
* Don't worry about breaking changes. As above, the release tooling will deal with that and in general if heapless is used in a public API for a crate, we'll have to bump the minor version (for pre- 1.0 crates).
* Thanks for taking on the task :)

Dirbaio:
i'm going to close this since

- we're already using heapless 0.9 everywhere
- you can get the document-features update by doing `cargo update`, embassy-nrf is not restricting the max version so it doesn't need updating.

BartMassey:
Sincere apologies for flaking on this task. Thanks for closing it.

PR #5170, opened 2026-01-04
Fix of the buffered version of STM32 UART

aegunov:
The ringbuffer was incorrectly used in the write (..) methods.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -688,6 +688,10 @@ impl<'d> BufferedUartTx<'d> {
 
             let empty = state.tx_buf.is_empty();
 
+            if state.tx_buf.len() < buf.len() {
+                return Poll::Ready(Err(Error::BufferTooLong));
+            }
+
             if state.tx_buf.len() - state.tx_buf.available() < buf.len() {
                 return Poll::Pending;
             }


embassy-stm32/src/usart/buffered.rs
```diff
@@ -688,22 +688,29 @@ impl<'d> BufferedUartTx<'d> {
 
             let empty = state.tx_buf.is_empty();
 
-            let mut tx_writer = unsafe { state.tx_buf.writer() };
-            let data = tx_writer.push_slice();
-            if data.is_empty() {
-                state.tx_waker.register(cx.waker());
+            if state.tx_buf.len() - state.tx_buf.available() < buf.len() {
```
timokroeger, line 695: Wouldn’t this condition mean we get stuck when the transmit buffer is empty?
Example:
tx_buf.len() == 100
tx_buf.available() == 100
len of data to be sent: buf.len() == 4
`100 - 100 < 4` -> `true`, return pending

aegunov: state.tx_buf.available() is amount of bytes that available to read, it's not free space.

embassy-stm32/src/usart/buffered.rs
```diff
@@ -688,22 +688,29 @@ impl<'d> BufferedUartTx<'d> {
...
+            let mut written: usize = 0;
+            let mut tx_writer = unsafe { state.tx_buf.writer() };
+            while written != buf.len() {
+                let data = tx_writer.push_slice();
+                if data.is_empty() {
+                    state.tx_waker.register(cx.waker());
+                    return Poll::Pending;
+                }
+                let n = data.len().min(buf.len() - written);
+                data[..n].copy_from_slice(&buf[written..written + n]);
```
kpfleming, line 708: FYI: In testing in my firmware builds, using `copy_from_slice` (or any similar methods) increased the resulting binary size by almost 1KB (even with maximum optimizations enabled). Using an open-coded `for` loop to do the same thing reduced binary size significantly.

embassy-stm32/src/usart/buffered.rs
aegunov: Maybe it's better to add above these lines check like this:
```
            if state.tx_buf.len() < buf.len() {
                return Poll::Ready(Err(Error::BufferTooLong));
            }
```

embassy-stm32/src/usart/buffered.rs
xoviat: What was the speed comparison?

kpfleming: I didn't perform an extensive analysis, but I did not see any performance decrease from the change. This is apparently just a quirk of this particular platform in the compiler/library: for other Arm platforms these helper functions get fully optimized out... they just don't for Cortex-M.

aegunov: I know that `copy_from_slice` (`memcpy`) implementation is not very simple, in addition it's inlined (as I can see in the source) and it has many additional checks. In the buffered.rs file I found about 6 calls to `copy_from_slice`. But even those facts can't explain so significant increase in resulting size.

PR #5171, opened 2026-01-04
fix: stm32h755cm4  examples compilation

liebman:
@lulf this fixes one that was missed during the release 
It was hidden because there were some stm32wl55 examples that were not named properly in Cargo.toml (my fault)


lulf (approved):
Thanks!

PR #5176, opened 2026-01-05
feat: add basic CryptoCell RNG driver

nponsard:
This PR adds a CryptoCell RNG driver for the nrf91 series. I built it following the RNG driver for the other nrf chips, an example has been added for the nrf9151 to show how to use it. 

It needs an updated nrf-pac, PR here: https://github.com/embassy-rs/nrf-pac/pull/14

I tested the example on the nrf9160dk and Thingy:91 X (nrf9151).

Concerning the activation of the CryptoCell (`Cryptocell::enable()`), should this be set up by the user or the RNG driver ?

### DIFF
--- a/examples/nrf5340/Cargo.toml
+++ b/examples/nrf5340/Cargo.toml
@@ -25,6 +25,7 @@ panic-probe = { version = "1.0.0", features = ["print-defmt"] }
 embedded-storage = "0.3.1"
 usbd-hid = "0.9.0"
 serde = { version = "1.0.136", default-features = false }
+rand = { version = "0.9.0", default-features = false }
 
 [profile.release]
 debug = 2
--- a/examples/nrf5340/src/bin/cryptocell_rng.rs
+++ b/examples/nrf5340/src/bin/cryptocell_rng.rs
@@ -1,0 +1,34 @@
+#![no_std]
+#![no_main]
+
+use embassy_executor::Spawner;
+use embassy_nrf::cryptocell_rng::{self, CcRng};
+use embassy_nrf::{bind_interrupts, peripherals};
+use rand::Rng as _;
+use {defmt_rtt as _, panic_probe as _};
+
+bind_interrupts!(struct Irqs {
+    CRYPTOCELL => cryptocell_rng::InterruptHandler<peripherals::CC_RNG>;
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_nrf::init(Default::default());
+
+    let mut rng = CcRng::new(p.CC_RNG, Irqs);
+
+    // Async API
+    let mut bytes = [0; 4];
+    rng.fill_bytes(&mut bytes).await;
+    defmt::info!("Some random bytes: {:?}", bytes);
+
+    // Sync API with `rand`
+    defmt::info!("A random number from 1 to 10: {:?}", rng.random_range(1..=10));
+
+    let mut bytes = [0; 1024];
+    rng.fill_bytes(&mut bytes).await;
+    let zero_count: u32 = bytes.iter().fold(0, |acc, val| acc + val.count_zeros());
+    let one_count: u32 = bytes.iter().fold(0, |acc, val| acc + val.count_ones());
+    defmt::info!("Chance of zero: {}%", zero_count * 100 / (bytes.len() as u32 * 8));
+    defmt::info!("Chance of one: {}%", one_count * 100 / (bytes.len() as u32 * 8));
+}


lulf (commented):
For now I think the rng driver can do it, ideally if it can check if it's enabled already first.

Also, maybe the peripheral and types could be named RNG and Rng? It aligns with existing peripherals and doesn't conflict with any other rng for those chips.

nponsard:
> Also, maybe the peripheral and types could be named RNG and Rng? It aligns with existing peripherals and doesn't conflict with any other rng for those chips.

It would become a conflict if it gets extended to the nrf52840 that has both RNG and CryptoCell peripherals.

embassy-nrf/src/chips/nrf9160.rs
```diff
@@ -74,7 +75,7 @@ pub mod pac {
         EGU4_S as EGU4,
         EGU5_S as EGU5,
         FICR_S as FICR,
-        FPU_S as FPU,
+        FPU_NS as FPU,
```
Dirbaio, line 78: why this change?

nponsard: It came with the new generated PAC, `FPU_S` doesn't exist anymore. I saw that `FPU_NS` was used even in s mode on the nrf9120 so I made the same modification as a fix.

nponsard:
Alright, squashed my fixup commits, managed to get `nrf52840` supported and tested (copying `cryptocell_rng` from `examples/nrf9151/s/`, ran on the nrf52840 of my nrf9160dk).

Is there any more changes needed ?

nponsard:
Oh also I changed the `nrf-pac` commit to be the one in main since the svd update got merged.

nponsard:
Ah I just thought about the naming, should the struct be named `CcRng`, `CCRng` or `Ccrng` ?
I think the last one matches more [UpperCamelCase](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case)

embassy-nrf/src/cryptocell_rng.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        // In self.start() there are calls to set_enable() that resets the interrupt mask,
+        // self.enable_irq() needs to be called after self.start().
+        self.start();
+
+        self.enable_irq();
+
+        let on_drop = OnDrop::new(|| {
+            self.stop();
+            self.disable_irq();
```
0e4ef622, line 166: `disable_irq` should come before `stop`, since `stop` stops the RNG_CLK and "The registers of a cryptographic engine are only accessible when its clock is enabled."

embassy-nrf/src/cryptocell_rng.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    fn enable_irq(&self) {
+        pac::CC_HOST_RGF
+            .imr()
+            .write(|w| w.set_rng_mask(pac::cc_host_rgf::vals::RngMask::IRQENABLE));
+        self.r
+            .rng_imr()
+            .write(|w| w.set_ehr_valid_mask(pac::cc_rng::vals::EhrValidMask::IRQENABLE));
+    }
```
0e4ef622, line 129: Not sure if this is considered a pac bug or not, but the default value of IMR from nrf-pac is 0, which means enable all the interrupts. (According to the datasheet and SVD file, the reset value is all 1's.). You could use `modify` instead. Same issue exists in `disable_irq`.

embassy-nrf/src/cryptocell_rng.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    fn disable_irq(&self) {
+        self.r.rng_icr().write(|w| w.set_ehr_valid_clear(true));
+        pac::CC_HOST_RGF.icr().write(|w| w.set_rng_clear(true));
+        self.r
+            .rng_imr()
+            .write(|w| w.set_ehr_valid_mask(pac::cc_rng::vals::EhrValidMask::IRQDISABLE));
+    }
```
0e4ef622, line 140: did you forget `CC_HOST_RGF.IMR` here?

embassy-nrf/src/cryptocell_rng.rs
nponsard: Ah my bad, I used `write` too often when experimenting. Should be good now.

embassy-nrf/src/cryptocell_rng.rs
nponsard: I didn't think about that, this should be fixed now.

embassy-nrf/src/cryptocell_rng.rs
nponsard: Thanks, I now added the instruction to mask the rng interrupt

examples/nrf9151/s/src/bin/cryptocell_rng.rs
```diff
@@ -0,0 +1,38 @@
...
+    CRYPTOCELL => cryptocell_rng::InterruptHandler<peripherals::CC_RNG>;
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_nrf::init(Default::default());
+
+    // https://docs.nordicsemi.com/bundle/ps_nrf9151/page/cryptocell.html#ariaid-title96
+
+    pac::CRYPTOCELL.enable().write(|w| w.set_enable(true));
```
0e4ef622: is this still necessary?

nponsard: No :+1:

embassy-nrf/src/cryptocell_rng.rs
```diff
@@ -0,0 +1,373 @@
...
+            // We need to make sure we haven't already filled the whole slice,
+            // in case the interrupt fired again before the executor got back to the future.
+            if !state.ptr.is_null() && state.ptr != state.end {
+                // If the future was dropped, the pointer would have been set to null,
+                // so we're still good to mutate the slice.
+                // The safety contract of `CcRng::new` means that the future can't have been dropped
+                // without calling its destructor.
+
+                for i in 0..6 {
+                    let bytes = r.ehr_data(i).read().to_be_bytes();
```
0e4ef622: Why `to_be_bytes`? I saw you use `from_ne_bytes` elsewhere.

0e4ef622:
Mentioned on matrix earlier, but I tested this on an nrf5340 and blocking api works, but async doesn't. The CRYPTOCELL interrupt never fires for some reason. But if I read some registers, like CRYPTOCELL.ENABLE or the RNG_ISR register, then the interrupt fires, might be an undocumented erratum.

Interestingly, I found a library within TF-M for using the cryptocell, and it seems like they always [busy wait for RNG interrupts](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/main/lib/ext/cryptocell-312-runtime/host/src/hal/cc3x/cc_hal.c#105) specifically.

embassy-nrf/src/cryptocell_rng.rs
nponsard: `from_ne_bytes` comes from the rng driver on main, I tend to use `to_be_bytes` by default with no real reason. It shouldn't matter as it's random bits right ? Is there a performance impact ?

I can change it to `to_ne_bytes` if you want.

nponsard:
I can try to add nrf5340 support.

What's the best way to busy wait ? Is it a good idea to make the task sleep for short durations using `Timer::after_millis(1).await;` so other tasks can be executed ?

embassy-nrf/src/cryptocell_rng.rs
0e4ef622: There is at least a size impact, it's one extra instruction.

Dirbaio:
> Is it a good idea to make the task sleep for short durations using Timer::after_millis(1).await; so other tasks can be executed ?

it's better to do `loop { yield_now().await }`. It still lets other tasks run but loops as fast as possible.

Sleeping adds ltency, plus embassy-time is an optional dep.

0e4ef622:
I'd also be fine with excluding the async functionality from nrf53.

embassy-nrf/src/cryptocell_rng.rs
nponsard: Changed `to_be_bytes` occurrences to `to_ne_bytes`

nponsard:
I made it so only the async driver of the nrf5340 CryptoCell is behind a `async-ccrng-nrf5340` feature. Makes a bit of a mess in the imports that I tried to mitigate in e6118b27c140e20f3328a686db560b78d1a3881a

nponsard:
Do you need more changes on this PR ? 
Should I squash my commits to get it ready to merge ?

Dirbaio:
what's the point of the `async-ccrng-nrf5340` cargo feature? if it works it should be available without any feature gate, it if it doesn't work it should just not be there.

nponsard:
> what's the point of the `async-ccrng-nrf5340` cargo feature? if it works it should be available without any feature gate, it if it doesn't work it should just not be there.

Async on the nrf5340 works (tested on Thingy:91X), its just that it requires `embassy-futures` that is an optional dependency. Should the dependency be pulled for `_nrf5330-app` instead ?

Dirbaio:
i'd just do the simplest thing, depend on `embassy-futures` unconditionally

nponsard:
Alright, unconditional import it is, should be good now.

Dirbaio:
:rocket: 

thanks!

PR #5175, opened 2026-01-05
[MCXA] Add SPI driver

bogdan-petru:
This PR adds SPI driver support for the MCXA276 family, including:

## Features
- Full SPI master and slave support
- Blocking, interrupt-driven, and DMA modes
- Configurable clock polarity and phase (all SPI modes)
- Configurable bit order (MSB/LSB first)
- Hardware chip select support
- FIFO management with configurable watermarks

## New Files
- `embassy-mcxa/src/spi.rs` - Main SPI driver implementation

## Examples
- `spi_master_blocking.rs` - Blocking master example
- `spi_slave_blocking.rs` - Blocking slave example  
- `spi_master_dma.rs` - DMA master example
- `spi_slave_dma.rs` - DMA slave example
- `spi_interrupt_master.rs` - Interrupt-driven master example
- `spi_interrupt_slave.rs` - Interrupt-driven slave example
- `spi_b2b_master.rs` - Back-to-back master example
- `spi_b2b_slave.rs` - Back-to-back slave example

## Changes to Existing Files
- Updated `Cargo.toml` with SPI feature flags
- Added SPI clock helper functions in `clocks/periph_helpers.rs`
- Updated DMA module with SPI DMA request mappings
- Added SPI interrupt handlers in `interrupt.rs`
- Exported SPI module in `lib.rs`

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/61
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/60
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/59
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/54

PR #5178, opened 2026-01-05
[MCXA] remove unnecessary interrupt module

felipebalbi:
Instead of replicating all we get from `interrupt_mod!()` macro, we can just rely on it instead.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/34
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/28

### DIFF
--- a/embassy-mcxa/src/ostimer.rs
+++ b/embassy-mcxa/src/ostimer.rs
@@ -655,13 +655,6 @@ pub mod time_driver {
     }
 
     /// Install the global embassy-time driver and configure NVIC priority for OS_EVENT.
-    ///
-    /// # Parameters
-    /// * `priority` - Interrupt priority for the OSTIMER interrupt
-    /// * `frequency_hz` - Actual OSTIMER clock frequency in Hz (stored for future use)
-    ///
-    /// Note: The frequency parameter is currently accepted for API compatibility.
-    /// The embassy_time_driver macro handles driver registration automatically.
     pub fn init() {
         let _clock_freq = unsafe {
             enable_and_reset::<_OSTIMER0_TIME_DRIVER>(&OsTimerConfig {


jamesmunns:
Might want to coordinate with @bogdan-petru on this - it's likely to cause merge conflicts with #5175

embassy-mcxa/src/lib.rs
```diff
@@ -347,6 +346,40 @@ embassy_hal_internal::peripherals!(
     WWDT0,
 );
 
+#[rustfmt::skip]
+embassy_hal_internal::interrupt_mod!(
+    ADC0,
```
jamesmunns, line 350: Does it make sense to just grab all the interrupts now, instead of just the ones we support so far?

If we do it all now, we can probably close out https://github.com/OpenDevicePartnership/embassy-mcxa/issues/28

felipebalbi: Done

felipebalbi:
> Might want to coordinate with @bogdan-petru on this - it's likely to cause merge conflicts with #5175

If this gets merged before SPI driver, I can also handle the merge conflicts myself. No worries there :)

embassy-mcxa/src/ostimer.rs
```diff
@@ -661,7 +662,7 @@ pub mod time_driver {
     ///
     /// Note: The frequency parameter is currently accepted for API compatibility.
     /// The embassy_time_driver macro handles driver registration automatically.
-    pub fn init(priority: crate::interrupt::Priority, frequency_hz: u64) {
+    pub fn init(frequency_hz: u64) {
```
jamesmunns: I'm not sure what we need to be API compatible with. We should probably remove `frequency_hz` if we don't use it.

felipebalbi: removed, thanks :)

PR #5192, opened 2026-01-07
[MCXA] i2c: erase the Instance generic

felipebalbi:
We can erase the `Instance` generic by caching the `WaitCell` and `RegisterBlock` references in a new `Info` struct that can be recovered during the instantiation of the driver.

From that point on, `Instance` is insignificant, reducing the number of generics an application needs to keep track of.

### DIFF
--- a/embassy-mcxa/src/i2c/mod.rs
+++ b/embassy-mcxa/src/i2c/mod.rs
@@ -83,8 +83,6 @@ mod sealed {
     pub trait Sealed {}
 }
 
-impl<T: GpioPin> sealed::Sealed for T {}
-
 trait SealedInstance {
     fn info() -> Info;
 }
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/i2c/mod.rs
```diff
@@ -83,11 +83,8 @@ mod sealed {
     pub trait Sealed {}
 }
 
-impl<T: GpioPin> sealed::Sealed for T {}
-
 trait SealedInstance {
-    fn regs() -> &'static pac::lpi2c0::RegisterBlock;
-    fn wait_cell() -> &'static WaitCell;
+    fn info() -> Info;
```
jamesmunns, line 87: Same note as #5193 re: "Struct of references" vs "reference to struct with items", and this maybe being const-able.

felipebalbi: I will keep this in mind for a future refactor, but the `UnsafeCell` used for the registers makes it a little non-trivial. I figured it wasn't a big deal.

PR #5193, opened 2026-01-07
[MCXA] adc: small refactor

felipebalbi:
A small refactor to make the driver look more similar to other drivers
in this HAL.

### DIFF
--- a/embassy-mcxa/src/adc.rs
+++ b/embassy-mcxa/src/adc.rs
@@ -682,9 +682,10 @@ impl<T: Instance> Handler<T::Interrupt> for InterruptHandler<T> {
 mod sealed {
     /// Seal a trait
     pub trait Sealed {}
-}
 
-impl<I: GpioPin> sealed::Sealed for I {}
+    /// Sealed pin trait
+    pub trait SealedAdcPin<T: super::Instance> {}
+}
 
 struct Info {
     regs: &'static pac::adc0::RegisterBlock,
@@ -726,7 +727,7 @@ macro_rules! impl_instance {
 
 impl_instance!(0, 1, 2, 3);
 
-pub trait AdcPin<Instance>: GpioPin + sealed::Sealed + PeripheralType {
+pub trait AdcPin<T: Instance>: sealed::SealedAdcPin<T> + GpioPin + PeripheralType {
     /// The channel to be used
     fn channel(&self) -> u8;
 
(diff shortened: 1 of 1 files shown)


felipebalbi:
@MathisDeroo, FYI. Please give it a read, if you have some time 👍

MathisDeroo:
> @MathisDeroo, FYI. Please give it a read, if you have some time 👍

I will definitely have a look 😄

embassy-mcxa/src/adc.rs
```diff
@@ -678,15 +708,13 @@ macro_rules! impl_instance {
         $(
             paste!{
                 impl SealedInstance for crate::peripherals::[<ADC $n>] {
-                    fn ptr() -> &'static pac::adc0::RegisterBlock {
-                        unsafe { &*pac::[<Adc $n>]::ptr() }
-                    }
-
-                    fn wait_cell() -> &'static WaitCell {
+                    fn info() -> Info {
```
jamesmunns, line 711: VERY minor nit: there's probably a way to make this return a `&'static Info`, which contains the wait cell directly, as well as the pointer to the regs. This would mean that holding the `info` in the `Adc` instance is only 1 pointer instead of two, which doesn't matter a ton for ADC, but if we make other peripherals consistent, it might be worth it.

Edit: at that point, you can probably also convert `fn info()` into `const INFO`, to avoid a potential fn call (tho since it's so simple and inlined, it probably doesn't matter either way :D).

embassy-mcxa/src/adc.rs
```diff
@@ -699,34 +727,39 @@ macro_rules! impl_instance {
 
 impl_instance!(0, 1, 2, 3);
 
-pub trait AdcPin<Instance>: GpioPin + sealed::Sealed + PeripheralType {
-    const CHANNEL: u8;
+pub trait AdcPin<T: Instance>: sealed::SealedAdcPin<T> + GpioPin + PeripheralType {
+    /// The channel to be used
```
jamesmunns, line 731: Any reason this couldn't still be a const, if we're macro generating it anyway? Is this a consistency thing?

felipebalbi: It was a consistency thing, I can certainly revert this hunk if you think it's better. A small inlined function used only one, I figured the compiler won't even make the function call :)

jamesmunns: Just wondering mostly!

embassy-mcxa/src/adc.rs
felipebalbi: I have a solution to this, I can add it to this PR or send a separate PR for both ADC and I2C. What you prefer?

PR #5196, opened 2026-01-08
[MCXA] make all `struct Info` a static reference

felipebalbi:
This PR maintains a static reference to the driver's `struct Info` instead of passing the full struct around. Note that the UART driver has a few more changes because of some functions and traits were marked `pub` when they should not have been.

Cc https://github.com/OpenDevicePartnership/embassy-mcxa/issues/55

### DIFF
--- a/embassy-mcxa/src/lpuart/buffered.rs
+++ b/embassy-mcxa/src/lpuart/buffered.rs
@@ -55,7 +55,7 @@ pub struct BufferedLpuart<'a> {
 
 /// Buffered LPUART TX driver
 pub struct BufferedLpuartTx<'a> {
-    info: Info,
+    info: &'static Info,
     state: &'static State,
     _tx_pin: Peri<'a, AnyPin>,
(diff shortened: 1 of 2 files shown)


felipebalbi:
@bogdan-petru, @MathisDeroo please review if you have time. This is touching drivers you worked on.

MathisDeroo:
Hi @felipebalbi, what is the reason for implementing the Info struct and setting it as static in the ADC structure ? What are the advantages ?

felipebalbi:
> Hi @felipebalbi, what is the reason for implementing the Info struct and setting it as static in the ADC structure ? What are the advantages ?

For ADC there are two very minor benefits:

1. Size: we will pass around a single reference (4 bytes) instead of two (8 bytes)
2. Consistency: making drivers look similar helps solidify patterns we want to be followed. That is, when a new driver needs to be written, consistent patterns will tend to be repeated.

Both are very minor, though :)

PR #5206, opened 2026-01-09
[MCXA] code watchdog support

MathisDeroo:


embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,449 @@
...
+}
+
+/// CDOG (Code Watchdog) configuration structure.
+///
+/// Defines the behavior of the watchdog for various fault conditions
+/// and operational modes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Config {
+    /// The timeout period after which the watchdog will trigger
+    pub timout: FaultControl,
```
felipebalbi: typo: timeout

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,449 @@
...
+/// This handler is called when any cdog interrupt fires.
+/// When reset happens, the interrupt handler will never be reached.
+pub struct InterruptHandler;
+
+impl Handler<typelevel::CDOG0> for InterruptHandler {
+    unsafe fn on_interrupt() {
+        let cdog0 = unsafe { &*pac::Cdog0::ptr() };
+
+        // Safety: `_irq` ensures an Interrupt Handler exists.
+        unsafe { crate::pac::Interrupt::CDOG0.disable() };
```
felipebalbi: Why do you disable the interrupt here? I suppose you should clear the interrupt status, but keep the line enabled, no? Care to explain?

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,449 @@
...
+            flags.state_flag().bit(),
+            flags.addr_flag().bit(),
+            flags.por_flag().bit()
+        );
+
+        // Stop the cdog
+        cdog0.stop().write(|w| unsafe { w.stp().bits(0) });
+
+        //Clear all flags by writing 0
+        unsafe { cdog0.flags().write_with_zero(|w| w) };
```
felipebalbi, line 387: right, this should be enough. When we're running the interrupt handler, interrupts are already disabled.

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,449 @@
...
+    pub fn sub16(&mut self) {
+        self.secure_counter = self.secure_counter.wrapping_sub(16);
+        self.info.sub16().write(|w| unsafe { w.sb16().bits(1) });
+    }
+
+    /// Decrements the secure counter by 256.
+    pub fn sub256(&mut self) {
+        self.secure_counter = self.secure_counter.wrapping_sub(256);
+        self.info.sub256().write(|w| unsafe { w.sb256().bits(1) });
+    }
```
felipebalbi: I think the public API is a little extensive. Maybe providing only `add()` and `sub()` is sufficient, then internally you figure out if you use the 256 and 16 registers.

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,446 @@
...
+/// This handler is called when any cdog interrupt fires.
+/// When reset happens, the interrupt handler will never be reached.
+pub struct InterruptHandler;
+
+impl Handler<typelevel::CDOG0> for InterruptHandler {
+    unsafe fn on_interrupt() {
+        let cdog0 = unsafe { &*pac::Cdog0::ptr() };
+
+        // Safety: `_irq` ensures an Interrupt Handler exists.
+        unsafe { crate::pac::Interrupt::CDOG0.disable() };
```
felipebalbi: Unnecessary unsafe here. Make sure you're using the latest rustc. Run "rustup update stable" to get the latest.

You should get a build warning for this.

embassy-mcxa/src/cdog.rs
MathisDeroo: Agree @felipebalbi, we can remove the disable/enable interrupt in the IRQ

MathisDeroo:
@felipebalbi Feedback integrated and rustfmt check passed using the script provided

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,404 @@
...
+/// Determines what action the CDOG takes when a fault is detected.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum FaultControl {
+    /// Enable system reset on fault detection
+    EnableReset = 1,
+    /// Enable interrupt on fault detection
+    EnableInterrupt = 2,
+    /// Disable both reset and interrupt
+    DisableBoth = 4,
```
felipebalbi, line 42: you can annotate this with `#[default]`

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+/// Timer pause control during special conditions.
+///
+/// Controls whether the instruction timer continues running or pauses.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum PauseControl {
+    /// Keep timer running during IRQ or debug halt
+    RunTimer = 1,
```
felipebalbi, line 53: you can annotate this with `#[default]`

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,404 @@
...
+/// Lock control for CDOG configuration.
+///
+/// When locked, configuration registers cannot be modified.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum LockControl {
+    /// Lock configuration
+    Locked = 1,
+    /// Unlock configuration
+    Unlocked = 2,
```
felipebalbi, line 68: you can annotate this with `#[default]`

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+/// CDOG (Code Watchdog) configuration structure.
+///
+/// Defines the behavior of the watchdog for various fault conditions
+/// and operational modes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
felipebalbi: Then this one can derive `Default` too.

embassy-mcxa/src/cdog.rs
```diff
@@ -0,0 +1,404 @@
...
+            miscompare: FaultControl::DisableBoth,
+            sequence: FaultControl::DisableBoth,
+            state: FaultControl::DisableBoth,
+            address: FaultControl::DisableBoth,
+            irq_pause: PauseControl::RunTimer,
+            debug_halt: PauseControl::RunTimer,
+            lock: LockControl::Unlocked,
+        }
+    }
+}
```
felipebalbi: and this entire block can be removed.

embassy-mcxa/src/cdog.rs
MathisDeroo: Does annotating with default replace the Default implementation for the Config ? Rust automatically matches the DisableBoth with the correct FaultControl variables in Config ?

felipebalbi: `#[default]` needs to be added to the variant you want to be the default. Then you derive `Default` for the enums and the struct.

MathisDeroo: Got it thanks ! I've done the change

felipebalbi (approved):

PR #5207, opened 2026-01-09
[MCXA] Add I3C driver

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/73

### DIFF
--- a/embassy-mcxa/src/i3c/controller.rs
+++ b/embassy-mcxa/src/i3c/controller.rs
@@ -15,6 +15,8 @@ use crate::interrupt::typelevel;
 use crate::pac::i3c0::mctrl::{Dir as I3cDir, Type};
 use crate::peripherals::I3C0;
 
+const MAX_CHUNK_SIZE: usize = 255;
+
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 enum SendStop {
@@ -420,7 +422,7 @@ impl<'d, M: Mode> I3c<'d, M> {
             return Err(Error::InvalidReadBufferLength);
         }
 
-        for chunk in read.chunks_mut(256) {
+        for chunk in read.chunks_mut(MAX_CHUNK_SIZE) {
             match self.blocking_start(address, bus_type, Dir::Read, chunk.len() as u8) {
                 Err(e) => {
                     self.blocking_remediation(bus_type);
@@ -684,7 +686,7 @@ impl<'d> I3c<'d, Async> {
             self.blocking_remediation(bus_type);
         });
 
-        for chunk in read.chunks_mut(256) {
+        for chunk in read.chunks_mut(MAX_CHUNK_SIZE) {
             self.async_start(address, bus_type, Dir::Read, chunk.len() as u8)
                 .await?;
 


embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/// I2C interrupt handler.
+pub struct InterruptHandler;
+
+impl interrupt::typelevel::Handler<interrupt::typelevel::I3C0> for InterruptHandler {
+    unsafe fn on_interrupt() {
+        let status = info().regs().mintmasked().read();
+
+        // defmt::info!("STATUS {:08x}", status.bits());
```
jamesmunns: Remove or uncomment

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+                    .errwarn()
+                    .clear_bit_by_one()
+                    .rxpend()
+                    .clear_bit_by_one()
+                    .txnotfull()
+                    .clear_bit_by_one()
+            });
+        }
+
+        info().wait_cell().wake();
```
jamesmunns: I haven't checked for consistency, do we usually always wake, or only inside of the `if` body if something was triggered?

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+                    self.set_drive_strength(crate::gpio::DriveStrength::Double.into());
+                    self.set_function(crate::pac::port0::pcr0::Mux::$fn);
+                    self.set_enable_input_buffer();
+                }
+            }
+        }
+    };
+}
+
+// impl_pin!(P0_2, Mux10, PurPin); REVISIT: what is this for?
```
jamesmunns, line 169: Maybe CC @bogdan-petru if you know if the PurPin is typically required for i3c, or if it's situational/optional?

https://www.nxp.com/docs/en/application-note/AN13952.pdf seems to talk about an externally gated pull-up pin, or "Pull Up Resistor" pin.

examples/mcxa/src/bin/i3c-async.rs
```diff
@@ -0,0 +1,74 @@
...
+    struct Irqs {
+        I3C0 => InterruptHandler;
+    }
+);
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let mut config = Config::default();
+    config.clock_cfg.sirc.fro_lf_div = Div8::from_divisor(1);
+    // let mut firc = config.clock_cfg.firc.unwrap();
```
jamesmunns: Clean up or remove dead code

examples/mcxa/src/bin/i3c-blocking.rs
```diff
@@ -0,0 +1,66 @@
...
+    config.clock_cfg.firc = Some(firc);
+
+    let p = hal::init(config);
+
+    defmt::info!("I3C example");
+
+    // Note: P0_2 is connected to P1_8 on the FRDM_MCXA276 via a resistor, and
+    // defaults to SWO on the debug peripheral. Explicitly make it a high-z
+    // input.
+    let _pin = Input::new(p.P0_2, Pull::Disabled);
```
jamesmunns, line 24: @felipebalbi I wonder if we should make this a hal::config::Config parameter, or clear this back to Disabled by default in init and/r make users explicitly enable SWO mode with a pin method.

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// Shorthand for `Result<T>`.
+pub type Result<T> = core::result::Result<T, Error>;
+
+pub mod controller;
+
+/// Error information type
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
```
jamesmunns, line 17: We might want to start marking some/all Error types `#[non_exhaustive]`.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+enum SendStop {
+    #[default]
+    No,
+    Yes,
+}
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
```
jamesmunns, line 28: Probably want doc comments for all these pub types.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+    fn from(value: Dir) -> Self {
+        match value {
+            Dir::Write => Self::Dirwrite,
+            Dir::Read => Self::Dirread,
+        }
+    }
+}
+
+/// I3C controller configuration
+pub struct Config {
```
jamesmunns, line 70: We might also want `non_exhaustive` here.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+
+        /* -------------------------------------------------------------
+         * 1) Push‑Pull baud (PPBAUD)
+         *    Generated from fclk / 2
+         * ------------------------------------------------------------- */
+
+        let mut pp_src_hz = fclk / 2;
+
+        let mut pp_div = pp_src_hz / target_pp_hz;
+        if pp_div == 0 {
```
jamesmunns: nit: could be `.min(1)`

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+        let target_i2c_hz = <Speed as Into<u32>>::into(config.i2c_speed);
+
+        /* -------------------------------------------------------------
+         * 1) Push‑Pull baud (PPBAUD)
+         *    Generated from fclk / 2
+         * ------------------------------------------------------------- */
+
+        let mut pp_src_hz = fclk / 2;
+
+        let mut pp_div = pp_src_hz / target_pp_hz;
```
jamesmunns: Might want to guard against `/0` here (either with a Result, an explicit check, or a min floor.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+         * 2) Open‑Drain baud (ODBAUD)
+         *    Depends on ODHPP mode
+         * ------------------------------------------------------------- */
+
+        let odhpp_enabled = self.info.regs().mconfig().read().odhpp().bit_is_set();
+
+        let (od_baud, _od_src_hz) = if odhpp_enabled {
+            // OD rate derived from 2×PP clock
+            let mut div = (2 * pp_src_hz) / target_od_hz;
+            if div < 2 {
```
jamesmunns: more `.min(2)`

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+            .odstop()
+            .disable()
+            .odhpp()
+            .enable()
+        });
+
+        Ok(())
+    }
+
+    // REVISIT: not very readable
```
jamesmunns, line 189: Yeah, some quotes from the reference manual here would be nice to at least follow the math around.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+
+    fn blocking_wait_for_tx_fifo(&self) {
+        while self.info.regs().mdatactrl().read().txfull().is_full() {}
+    }
+
+    fn blocking_wait_for_rx_fifo(&self) {
+        while self.info.regs().mdatactrl().read().rxempty().is_empty() {}
+    }
+
+    fn status(&self) -> Result<()> {
```
jamesmunns: As always, do we need an error iterator here?

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+use embassy_hal_internal::PeripheralType;
+use maitake_sync::WaitCell;
+use paste::paste;
+
+use crate::clocks::ClockError;
+use crate::gpio::{GpioPin, SealedPin};
+use crate::{interrupt, pac};
+
+/// Shorthand for `Result<T>`.
+pub type Result<T> = core::result::Result<T, Error>;
```
jamesmunns: I'm personally not a big fan of result types like this (these days at least), I think style guides are moving away from them.

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+impl sealed::Sealed for Async {}
+impl Mode for Async {}
+
+macro_rules! impl_pin {
+    ($pin:ident, $fn:ident, $trait:ident) => {
+        impl_pin!(crate::peripherals, $pin, $fn, $trait);
+    };
+
+    ($perip:path, $pin:ident, $fn:ident, $trait:ident) => {
+        paste! {
```
jamesmunns, line 153: I don't think you use paste for anything here?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+        self.info
+            .wait_cell()
+            .wait_for(|| {
+                // enable TXNOTFULL interrupt
+                self.info
+                    .regs()
+                    .mintset()
+                    .write(|w| w.txnotfull().set_bit().errwarn().enable());
+                // if the TX FIFO isn't full, we can write bytes
+                self.info.regs().mstatus().read().txnotfull().is_notfull()
```
jamesmunns, line 599: Do we need to check for errors in these `wait_for`s? Not sure if this could lead us to deadlock if the i2c gets in some kind of fault state. We might want to wait_for_value and return a `Result<()>`

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,929 @@
...
+    async fn async_stop(&self, bus_type: BusType) -> Result<()> {
+        if !self.info.regs().mstatus().read().state().is_normact() {
+            Err(Error::InvalidRequest)
+        } else {
+            // NOTE: Section 41.3.2.1 states that "when sending STOP
+            // in I2C mode, MCONFIG[ODSTOP] and MCTRL[TYPE] must be
+            // 1".
+            if bus_type == BusType::I2c {
+                self.info.regs().mconfig().modify(|_, w| w.odstop().enable());
+                self.info
```
jamesmunns: I think this second call is the same in both arms? factor it out?

embassy-mcxa/src/i3c/mod.rs
felipebalbi: fixed

embassy-mcxa/src/i3c/mod.rs
felipebalbi: should have moved it inside the if. Fixed.

examples/mcxa/src/bin/i3c-async.rs
felipebalbi: fixed

embassy-mcxa/src/i3c/mod.rs
felipebalbi: added here.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: Do you mean `.max(1)`? That is, if `pp_div` is 0, then take the max.

examples/mcxa/src/bin/i3c-async.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: fixed

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: That's in 41.5.1, but it's not super useful. The function you see here is mostly adapted from the vendor SDK.

image

embassy-mcxa/src/i3c/mod.rs
felipebalbi: Need `paste` because of the `path` argument:

image

unless I'm missing a detail somewhere.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: Fixed

embassy-mcxa/src/i3c/mod.rs
jamesmunns: Ah, I missed that! This could work, it might not be worth it just to remove the use of paste though.

```rust
macro_rules! impl_pin {
    ($pin:path, $fn:ident, $trait:ident) => {
        impl sealed::Sealed for $pin {}

        impl $trait<crate::peripherals::I3C0> for $pin {
            fn mux(&self) {
                self.set_pull(crate::gpio::Pull::Disabled);
                self.set_slew_rate(crate::gpio::SlewRate::Fast.into());
                self.set_drive_strength(crate::gpio::DriveStrength::Double.into());
                self.set_function(crate::pac::port0::pcr0::Mux::$fn);
                self.set_enable_input_buffer();
            }
        }
    };
}

// impl_pin!(P0_2, Mux10, PurPin); REVISIT: what is this for?
impl_pin!(crate::peripherals::P0_17, Mux10, SclPin);
impl_pin!(crate::peripherals::P0_18, Mux10, SdaPin);
impl_pin!(crate::peripherals::P1_8, Mux10, SdaPin);
impl_pin!(crate::peripherals::P1_9, Mux10, SclPin);
// impl_pin!(P1_11, Mux10, PurPin); REVISIT: what is this for?
impl_pin!(crate::internal_peripherals::P1_30, Mux10, SdaPin);
impl_pin!(crate::internal_peripherals::P1_31, Mux10, SclPin);
```

embassy-mcxa/src/i3c/mod.rs
felipebalbi: removed

embassy-mcxa/src/i3c/mod.rs
felipebalbi: ugh, I'll just leave it for now.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: good catch. I do check the status bits after calling these, but we should stop waiting if any of the errors are triggered as well.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: I need to produce a single error variant to the caller. Not sure how an iterator would help for this case. I'll keep it as is and take a mental note.

embassy-mcxa/src/clocks/periph_helpers.rs
```diff
@@ -163,6 +163,74 @@ impl SPConfHelper for NoConfig {
...
+                let freq = clocks.ensure_clk_1m_active(&self.power)?;
+                (freq, Mux::ClkrootFunc5)
+            }
+            I3cClockSel::None => unsafe {
+                // no ClkrootFunc7, just write manually for now
+                clksel.write(|w| w.bits(0b111));
+                clkdiv.modify(|_r, w| w.reset().asserted().halt().asserted());
+                return Ok(0);
+            },
+        };
```
jamesmunns, line 230: Could you enforce a 25MHz max for this *post* div?

Like https://github.com/embassy-rs/embassy/pull/5238/commits/1e6f47a7ea58f09e878c553238bdafe134130a1b, but it's 25MHz in both normal/overdrive, so you don't need to worry about getting the voltage state.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[allow(dead_code)]
+enum BusType {
+    /// I3C SDR
+    #[default]
+    I3cSdr,
+    /// Legacy I2C
+    I2c,
+    /// I3C DDR
+    I3cDdr, // Support for DDR is still missing
```
jamesmunns: Is this true? Maybe remove if so, and make the bustype non-exhaustive?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+            self.blocking_start(address, bus_type, Dir::Read, chunk.len() as u8)?;
+
+            for byte in chunk.iter_mut() {
+                self.blocking_wait_for_rx_fifo();
+                *byte = self.info.regs().mrdatab().read().value().bits();
+            }
+        }
+
+        if send_stop == SendStop::Yes {
+            self.blocking_stop(bus_type)?;
```
jamesmunns, line 441: This is a question for both blocking and async:

Do we need to make "send a stop condition" something we handle on early-error/early-cancellation?

Should we always *attempt* to send a stop condition before returning?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+        read: &mut [u8],
+        bus_type: BusType,
+        send_stop: SendStop,
+    ) -> Result<(), Error> {
+        if read.is_empty() {
+            return Err(Error::InvalidReadBufferLength);
+        }
+
+        for chunk in read.chunks_mut(256) {
+            self.blocking_start(address, bus_type, Dir::Read, chunk.len() as u8)?;
```
jamesmunns: Not familiar with i3c (or even specifics enough of i2c), do you actually want to send a "start condition" onto the bus every 256 bytes? I'm not sure if there's a limitation with the hardware, but triggering a "start" pattern is likely to cause problems to devices that don't expect it.

jamesmunns: Also if the len is 256 here, `len` will be zero, this is probably a bug.

jamesmunns: Reading the datasheet, it seems like if you set RDTERM to zero, you need to manually cause a stop or something? I don't think we can split_last with a chunks iterator, but maybe we have RDTERM = 0 for all but the last chunk, then set it ONLY on the last chunk?

Idk, but I think this merits a double check you got the logic right!

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+            self.async_start(address, bus_type, Dir::Read, chunk.len() as u8)
+                .await?;
+
+            for byte in chunk.iter_mut() {
+                self.async_wait_for_rx_fifo().await?;
+                *byte = self.info.regs().mrdatab().read().value().bits();
+            }
+        }
+
+        if send_stop == SendStop::Yes {
```
jamesmunns, line 699: Same note as above: Do we have any cancellation (or early error return!) considerations for the async API?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+        self.async_wait_for_complete().await?;
+
+        if send_stop == SendStop::Yes {
+            self.async_stop(bus_type).await?;
+        }
+
+        Ok(())
+    }
+
+    // Public API: Async
```
jamesmunns, line 764: We should probably discuss cancellation safety for all public APIs! "do we guarantee a STOP is sent if a START is sent successfully", etc.!

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+    /// Write to address from buffer asynchronously.
+    pub fn async_write<'a>(
+        &mut self,
+        address: u8,
+        write: &'a [u8],
+    ) -> impl Future<Output = Result<(), Error>> + use<'_, 'a, 'd> {
+        self.async_write_internal(address, write, BusType::I3cSdr, SendStop::Yes)
+    }
+
+    /// Write to address from bytes and read from address into buffer asynchronously.
```
jamesmunns:
```suggestion
    /// Write to address from bytes and then read from address into buffer asynchronously.
```

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+    pub async fn i2c_async_write_read(&mut self, address: u8, write: &[u8], read: &mut [u8]) -> Result<(), Error> {
+        self.async_write_internal(address, write, BusType::I2c, SendStop::No)
+            .await?;
+        self.async_read_internal(address, read, BusType::I2c, SendStop::Yes)
+            .await
+    }
+}
+
+fn calculate_error(cur_freq: u32, desired_freq: u32) -> u32 {
+    if cur_freq > desired_freq {
```
jamesmunns: could use https://doc.rust-lang.org/stable/std/primitive.u32.html#method.abs_diff here, e.g.

```rust
let delta = cur_freq.abs_diff(desired_freq);
delta * 100 / desired_freq
```

jamesmunns: nit: Also maybe move this function out from between impls and up or down?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -0,0 +1,932 @@
...
+
+impl<'d, M: Mode> embedded_hal_02::blocking::i2c::Transactional for I3c<'d, M> {
+    type Error = Error;
+
+    fn exec(
+        &mut self,
+        address: u8,
+        operations: &mut [embedded_hal_02::blocking::i2c::Operation<'_>],
+    ) -> Result<(), Error> {
+        if let Some((last, rest)) = operations.split_last_mut() {
```
jamesmunns: could use:

```rust
let Some(...) = ... else {
  return Ok(());
};

// ...
```

to avoid some rightward drift

embassy-mcxa/src/i3c/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    InvalidWriteBufferLength,
+    /// Invalid read buffer length.
+    InvalidReadBufferLength,
+    /// User provided an invalid configuration
+    InvalidConfiguration,
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
+/// I2C interrupt handler.
```
jamesmunns:
```suggestion
/// I3C interrupt handler.
```

examples/mcxa/src/bin/i3c-async.rs
```diff
@@ -0,0 +1,71 @@
...
+        // ~~~~~~~~ //
+        // I2C mode //
+        // ~~~~~~~~ //
+
+        // RSTDAA
+        i3c.async_write(0x7e, &[0x06]).await.unwrap();
+
+        i3c.i2c_async_write_read(0x48, &[0x00], &mut buf).await.unwrap();
+        let raw = f32::from(i16::from_be_bytes(buf) / 16);
+        let temp_i2c = raw * 0.0625;
```
jamesmunns, line 50: Are you dividing by 16 twice? (you do /16 above, then `* 0.0625`, which is the same as another div16)

examples/mcxa/src/bin/i3c-async.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        Timer::after_micros(100).await;
+
+        // ENTDAA
+        i3c.async_write(0x7e, &[0x07]).await.unwrap();
+
+        // P3T1755 temperature register = 0x00
+        i3c.async_write_read(0x48, &[0x00], &mut buf).await.unwrap();
+        let raw = f32::from(i16::from_be_bytes(buf) / 16);
+        let temp_i3c = raw * 0.0625;
```
jamesmunns: Same

embassy-mcxa/src/i3c/controller.rs
felipebalbi: We definitely don't need to send a stop always because of repeated starts, for example. In case of error, I haven't considered. Maybe a Stop is the right thing to do here, not sure. Let me think about this for a while.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: > Not familiar with i3c (or even specifics enough of i2c), do you actually want to send a "start condition" onto the bus every 256 bytes? I'm not sure if there's a limitation with the hardware, but triggering a "start" pattern is likely to cause problems to devices that don't expect it.

for this particular controller, yes. It can only handle up to 256 bytes per chunk, then you must send a repeated start. I did test with `RDTERM` set to zero, and it just would not work at all. It received zeros before I had the chance to send that Stop.

Something else to consider is a chunk of exactly 256 bytes. Maybe that's what 0 means, actually. "Read up to 256 bytes" and that's why I got random zeros when attempting to set `RDTERM` to 0. Another thing to test more carefully.

examples/mcxa/src/bin/i3c-async.rs
felipebalbi: yes. The temperature data is shifted left by 4 bits, first I'm getting rid of that. Then each bit represents 0.0625C, hence the multiplication.

examples/mcxa/src/bin/i3c-async.rs
felipebalbi: See previous.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: The only reason for that is that I'm not sure what's the best API here. Maybe I expose `BusType` and get rid of the `i2c_*()` APIs? Then we need to make sure `BusType` is (potentially) sufficient for other I3C masters as well for whenever we come up with an `embedded-hal` I3C trait.

examples/mcxa/src/bin/i3c-async.rs
jamesmunns: Ah, fun, just a coincidence then :)

embassy-mcxa/src/i3c/mod.rs
felipebalbi: fixed

embassy-mcxa/src/i3c/controller.rs
felipebalbi: I'll get rid of the `i2c_*` APIs, I think it doesn't look too bad to pass one extra argument. We can change it later if we learn anything different.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: Added a remediation function to handle this for both blocking and async.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: nice, done.

embassy-mcxa/src/clocks/periph_helpers.rs
felipebalbi: done

embassy-mcxa/src/i3c/controller.rs
felipebalbi: FYI only, I'll play with this again next week. Meanwhile, all the other comments were either sorted out or replied to ;)

felipebalbi: Just tested this and it appears that max chunk size is 255, not 256. I'll update the PR.

felipebalbi: pushed :)

PR #5213, opened 2026-01-11
Upstream embassy-neorv32

kurtjd:
This PR begins upstreaming [embassy-neorv32](https://github.com/kurtjd/neorv32-rs/tree/main/embassy-neorv32), a HAL for the open-source [NEORV32](https://github.com/stnolting/neorv32) RISC-V microcontroller.

While the NEORV32 is quite niche compared to some of the other supported families here, I think it would be cool to have upstream support for a RISC-V microcontroller, at least to help serve as a reference for others targeting RISC-V chips.

This initial PR includes UART and GPIO peripherals (with DMA and SysInfo pulled in since UART is dependent on them), but future PRs will be made for upstreaming more peripherals that I currently have supported (not included here to keep this relatively small for ease of reviewing).

### DIFF
--- a/embassy-neorv32/src/dma.rs
+++ b/embassy-neorv32/src/dma.rs
@@ -1,0 +1,366 @@
+//! Direct Memory Access (DMA)
+use core::marker::PhantomData;
+use core::pin::Pin;
+use core::sync::atomic::{AtomicBool, Ordering, fence};
+use core::task::{Context, Poll};
+
+use embassy_hal_internal::{Peri, PeripheralType};
+use embassy_sync::waitqueue::AtomicWaker;
+
+use crate::interrupt::typelevel::{Binding, Handler, Interrupt};
+use crate::peripherals::DMA;
+
+const U23_MAX: u32 = 0xff_ffff;
+
+/// DMA interrupt handler binding.
+pub struct InterruptHandler<T: Instance> {
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Instance> Handler<T::Interrupt> for InterruptHandler<T> {
+    unsafe fn on_interrupt() {
+        // Acking the interrupt clears both the ERROR and DONE flags.
+        // In poll, we can check the BUSY flag to know if we are done,
+        // but still need a way to check for bus error.
+        //
+        // So we cache the ERROR flag before clearing it.
+        let err = T::info().reg.ctrl().read().dma_ctrl_error().bit_is_set();
+        T::info().err_flag.store(err, Ordering::Release);
+        T::info().reg.ctrl().modify(|_, w| w.dma_ctrl_ack().set_bit());
+
+        T::info().waker.wake();
+    }
+}
+
+/// DMA error.
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    /// The NEORV32 configuration does not support DMA.
+    NotSupported,
+    /// Indicates a bus error occurred during transfer.
+    BusError,
+}
+
+enum DataConfig {
+    ConstantByte,
+    ConstantWord,
+    IncrementingByte,
+    IncrementingWord,
+}
+
+impl From<DataConfig> for u32 {
+    fn from(config: DataConfig) -> Self {
+        match config {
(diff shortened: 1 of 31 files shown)


kurtjd:
A note on the DMA driver: Although the DMA hardware is single-channel only, I decided to not make DMA a free-standing driver (similar to SysInfo) because I want the borrow checker's help in ensuring only one `Transfer` exists at a time (which is accomplished by the `Transfer` struct holding a mutable reference to the `Dma` struct).

Otherwise, I would have to rely on runtime mechanisms to enforce this (unless anyone else has other ideas) which I'd prefer not to do.

embassy-neorv32/src/dma.rs
```diff
@@ -0,0 +1,362 @@
...
+
+impl<'d, 't> Future for Transfer<'d, 't> {
+    type Output = Result<(), Error>;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.dma.info.waker.register(cx.waker());
+
+        if self.dma.busy() {
+            Poll::Pending
+        } else if self.dma.info.err_flag.load(Ordering::Acquire) {
+            Poll::Ready(Err(Error::BusError))
```
felipebalbi, line 288: does the DMA tell you how many bytes have been transferred? Might be a good idea to pass that along in the error case.

embassy-neorv32/src/dma.rs
```diff
@@ -0,0 +1,362 @@
...
+pub trait Word: SealedWord {}
+
+impl SealedWord for u8 {
+    fn cfg_constant() -> DataConfig {
+        DataConfig::ConstantByte
+    }
+
+    fn cfg_increment() -> DataConfig {
+        DataConfig::IncrementingByte
+    }
```
felipebalbi, line 318: `#[inline(always)]`?

embassy-neorv32/src/interrupts.rs
```diff
@@ -0,0 +1,154 @@
...
+                unsafe impl $crate::interrupt::typelevel::Binding<$crate::interrupt::typelevel::$irq, $handler> for $name {}
+            )*
+        )*
+    };
+}
+
+/// Generate a standard `mod interrupt` for a RISC-V HAL.
+#[macro_export]
+macro_rules! interrupt_mod {
+    ($($irqs:ident),* $(,)?) => {
```
felipebalbi, line 52: why can't you use the one from `embassy-hal-internal`?

embassy-neorv32/src/sysinfo.rs
```diff
@@ -0,0 +1,241 @@
...
+/// SoC configuration.
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct SocConfig(u32);
+
+impl SocConfig {
+    #[inline(always)]
+    fn supported(&self, i: u32) -> bool {
+        self.0 & (1 << i) != 0
+    }
```
felipebalbi: nit: predicates are more commonly named `is_<SOMETHING>`.

embassy-neorv32/src/dma.rs
kurtjd: Unfortunately, it does not :(

embassy-neorv32/src/interrupts.rs
kurtjd: `embassy-hal-internal` only currently supports `cortex-m`.

Ideally I would like to add RISC-V support there but interrupts for RISC-V are a bit more platform dependent and would need to think some more on what that would look like. For example on the NEORV32 the CPU contains custom core interrupts for the peripherals but on other platforms those might be external interrupts controlled via PLIC.

embassy-neorv32/src/dma.rs
kurtjd: Yeah, likely a good call here. I've been wishy-washy on adding `#[inline(always)` in some places because sometimes I feel like it's best to let compiler decide but probably pretty cut-and-dry here.

embassy-neorv32/src/sysinfo.rs
kurtjd: Sounds good, will make that change. Thanks!

PR #5215, opened 2026-01-12
[MCXA]: Don't read OsTimer before initialization

jamesmunns:
If we attempt to read the OsTimer prior to it being initialized, some kind of fault will occur.

https://github.com/embassy-rs/embassy/pull/5179 introduced a regression by enabling the `defmt-timestamp-uptime` feature, which caused prints in the SPLL setup to get the timestamp, causing a crash.

As a mitigation, we just return a time of zero until the OsTimer has actually been initialized.

### DIFF
--- a/embassy-mcxa/src/lib.rs
+++ b/embassy-mcxa/src/lib.rs
@@ -467,15 +467,16 @@ pub fn init(cfg: crate::config::Config) -> Peripherals {
     // Configure clocks
     crate::clocks::init(cfg.clock_cfg).unwrap();
 
+    // Initialize embassy-time global driver backed by OSTIMER0
+    // NOTE: As early as possible, but MUST be before clocks!
+    crate::ostimer::init(cfg.time_interrupt_priority);
+
     unsafe {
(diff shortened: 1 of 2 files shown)


embassy-mcxa/src/lib.rs
```diff
@@ -467,16 +467,17 @@ pub fn init(cfg: crate::config::Config) -> Peripherals {
     // Configure clocks
     crate::clocks::init(cfg.clock_cfg).unwrap();
 
+    // Initialize embassy-time global driver backed by OSTIMER0
+    // NOTE: As early as possible, but MUST be before clocks!
```
jannic, line 471: before? or after?

jamesmunns: Oops, after! I'll fix it in my other open PR!

jamesmunns: Fixed by https://github.com/embassy-rs/embassy/pull/5218/commits/cbdf1c521b3e9228c5000932598a719659930a3a in #5218.

PR #5218, opened 2026-01-12
[MCXA]: Allow swapping of main_clk

jamesmunns:
Implements the remaining clocking functionality (other than deep sleep) on the MCX-A, including the ability to swap core clock frequency.

### DIFF
--- a/embassy-mcxa/src/clocks/config.rs
+++ b/embassy-mcxa/src/clocks/config.rs
@@ -112,6 +112,7 @@ impl Div8 {
 /// ```
 #[non_exhaustive]
 pub struct ClocksConfig {
+    /// Clocks that are used to drive the main clock, including the AHB and CPU core
     pub main_clock: MainClockConfig,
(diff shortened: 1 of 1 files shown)


jamesmunns:
Screenshot 2026-01-13 at 3 28 51 PM

Very very slow CPU clock wiggles, using clkout in the `cpu-clocks` example

embassy-mcxa/src/clocks/config.rs
```diff
@@ -112,6 +112,7 @@ impl Div8 {
 /// ```
 #[non_exhaustive]
 pub struct ClocksConfig {
+    pub main_clock: MainClockConfig,
```
felipebalbi, line 116: do you want to add a comment here, like the others?

jamesmunns: Already added in https://github.com/embassy-rs/embassy/pull/5218/commits/b93e47cbece2d128982e4cf24daffda845c4a432 !

PR #5223, opened 2026-01-13
[embassy-time-driver]: Be specific that `Driver::now()` impls shouldn't fail

jamesmunns:
CC https://github.com/embassy-rs/embassy/pull/5215

@Dirbaio suggested adding this to the docs, opening this half as a ready-to-go PR, half as a discussion if this is a reasonable change to make (esp w/o a breaking change).


jamesmunns:
FWIW/for the record: I think you can pretty easily make a case this isn't a breaking change:

1. If it would have faulted (e.g. MCX-A, nrf54), it wasn't going to work anyway
2. If it wouldn't have faulted, but did read garbage data for the "now" time, the implementation PROBABLY wouldn't have been monotonic, if the counter started from zero when initialized

So IMO, I don't think we particularly need to worry about this being a breaking change or not, unless someone comes along and yells at us :)

PR #5228, opened 2026-01-13
stm32: BusOFF, BusPassive, BusWarning removed from `curr_error` in `FDCAN` and `BXCAN`

NeoHylde:
**Changes:**
- `curr_error` doesn't read PSR.BO/EW/EP.

**Problem:**
When awaiting can::read() / try_read(), if the CAN controller enters ErrorWarning or ErrorPassive, the future resolves immediately if the RxFifo's are empty and the CAN controller is in ErrorPassive or ErrorWarning.

For async functions this can result in tight polling loops that:
- hog cpu
- starve other async tasks

Root cause is treating a bus error state (BusPassive, BusWarning, BusOff) as an error. Error states reflect the TEC or REC registers exceeding certain numbers, while the LEC is the error last detected on the bus. Additionally, LEC may be "reset" on read, EW, EP, and BO do not.

**Fix:**
Read no longer reads BO, EW, EP registers in PSR registers.

**Possible Future Changes**
I would recommend seperating the error checks from the can read function, and creating a dedicated error checking function that is exposed through properties.

Meaning you can detect errors outside of reads, giving a bit more versatility.

### DIFF
--- a/embassy-stm32/src/can/enums.rs
+++ b/embassy-stm32/src/can/enums.rs
@@ -4,8 +4,6 @@
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum BusError {
-    /// No error occurred
-    NoError,
     /// Bit stuffing error - more than 5 equal bits
     Stuff,
     /// Form error - A fixed format part of a received message has wrong format
@@ -23,8 +21,6 @@ pub enum BusError {
     Crc,
     /// A software error occured. Exclusive to BXCAN.
     Software,
-    /// No CAN bus event was detected since the last CPU read access to the PSR. Exclusive to FDCAN.
-    NoChange,
 }
 
 /// Bus error modes.
--- a/embassy-stm32/src/can/fd/peripheral.rs
+++ b/embassy-stm32/src/can/fd/peripheral.rs
@@ -100,14 +100,14 @@ impl Registers {
 
     fn reg_to_error(value: u8) -> Option<BusError> {
         match value {
-            0b000 => Some(BusError::NoError),
+            // 0b000 => None,
             0b001 => Some(BusError::Stuff),
             0b010 => Some(BusError::Form),
             0b011 => Some(BusError::Acknowledge),
             0b100 => Some(BusError::BitRecessive),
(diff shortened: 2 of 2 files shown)


xoviat:
> When awaiting can::read() / try_read(), if the CAN controller enters ErrorWarning or ErrorPassive, 

What behavior does this pull request enact? If the can controller is in ErrorWarning and the user calls read with this PR, what happens?

NeoHylde:
> > When awaiting can::read() / try_read(), if the CAN controller enters ErrorWarning or ErrorPassive,
> 
> What behavior does this pull request enact? If the can controller is in ErrorWarning and the user calls read with this PR, what happens?

Before PR:
When awaiting a CAN read, if the controller was in BusWarning or BusPassive, curr_error() returned Some(err), where err reflected the current bus state (read from the BO/EW/EP status bits).

Which meant calling read() would return immediately instead of awaiting, even though no new CAN error had occurred.

After PR:
When awaiting a CAN read, if the controller is in BusWarning or BusPassive, curr_error() wont report bus states as errors. Instead, it only returns Some(err) when the LEC (Last Error Code) register updates a new protocol error.

If LEC reads NoError or NoChange, curr_error() returns None, allowing read() to await until a frame is received or a new error occurs.

This also applies to try_read since in principal returning a bus state as an error isn't correct. Hope this clarifies my thinking!

xoviat:
I think a better solution, if possible, is to add a `async fn wait_ready()` that the user can call after they receive a bus error.

NeoHylde:
> I think a better solution, if possible, is to add a `async fn wait_ready()` that the user can call after they receive a bus error.

I'm not sure if that approach would be effective.

By definition, a can node in Warning or Passive state still participates in bus activity; it just transmits different error flags. Requiring the user to call a `wait_ready()` function means that the node must now ignore incoming messages until the EW/EP flags are cleared, even though it's still in a valid operating state.
 
Additionally, curr_error also prioritizes status errors over actual errors which could help diagnose the can node.

Since `can::Properties` already exposes a `bus_error_state` function, removing bus states from `BusError` properly separates actual errors from bus states, while preserving correct can behaviour.

PR #5231, opened 2026-01-14
stm32/timer: add set_period, improve psc/arr calc.

Dirbaio:
- add set_period{,_ms,_us,_secs}. Allows setting non-integer hertz frequency.
- Add option to choose rounding mode if the exact requested period/freq can't be hit exactly (towards slower or towards faster timer).
- Fix panic when requested frequency is greater than the timer frequency.

### DIFF
--- a/embassy-stm32/src/timer/complementary_pwm.rs
+++ b/embassy-stm32/src/timer/complementary_pwm.rs
@@ -2,7 +2,7 @@
 
 use core::marker::PhantomData;
 
-use super::low_level::{CountingMode, OutputPolarity, Timer};
+use super::low_level::{CountingMode, OutputPolarity, RoundTo, Timer};
 use super::simple_pwm::PwmPin;
 use super::{AdvancedInstance4Channel, Ch1, Ch2, Ch3, Ch4, Channel, TimerComplementaryPin};
 use crate::Peri;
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/timer/complementary_pwm.rs
```diff
@@ -181,15 +181,85 @@ impl<'d, T: AdvancedInstance4Channel> ComplementaryPwm<'d, T> {
...
+    /// Set the PWM period in seconds.
+    ///
+    /// The actual period may differ from the requested value due to hardware
+    /// limitations. The timer will round towards a slower (longer) period.
+    ///
+    /// Note: that the period will not be applied in the timer until an update event
+    /// occurs.
+    pub fn set_period_secs(&mut self, secs: u32) {
+        let secs = if self.inner.get_counting_mode().is_center_aligned() {
+            secs / 2
```
diondokter: Actually, for seconds that are not multiple of 2, this is quite a rounding error...

Dirbaio: nice catch. fixing

Dirbaio: fix0red

diondokter (approved):

embassy-stm32/src/timer/low_level.rs
```diff
@@ -978,3 +1152,191 @@ impl<'d, T: AdvancedInstance4Channel> Timer<'d, T> {
...
+                            period_clocks,
+                            period_clocks,
+                            max_arr_bits
+                        );
+                    }
+                }
+
+                // Test mutations: verify the solution is optimal
+                // Try all combinations of psc +/- 1 and arr +/- 1
+                let mutations: [(i32, i64); 8] = [(-1, 0), (1, 0), (0, -1), (0, 1), (-1, -1), (-1, 1), (1, -1), (1, 1)];
```
0e4ef622, line 1277: this is not a great way to test for this. also i can make it fail by adding (847, -55360) to the list.

Specifically for period_clocks=10_000_000, max_arr_bits=16, we get psc=15, arr=62499. But a better solution is psc=1000, arr=10000.

Dirbaio: True it doesn't guarantee optimality. Updated comment.

Guaranteeing optimality would require searching all divisors of `period_clocks` which is obviously too expensive.

PR #5238, opened 2026-01-15
[MCXA]: Configure VDD_CORE power and flash wait states

jamesmunns:
The previous clocks PR will currently allow us to exceed some clocking frequency limits, without boosting the internal VDD_CORE LDO when necessary, or increasing the internal flash wait states. This PR fully rectifies that for the main "active" mode, and includes some pieces of "low power" mode (used in deep sleep), to be extended later.

CC @bogdan-petru - I would appreciate a look at this to ensure that I'm handling the VDD/LDO_CORE items correctly here, especially any of the TODOs.

### DIFF
--- a/embassy-mcxa/src/clocks/mod.rs
+++ b/embassy-mcxa/src/clocks/mod.rs
@@ -861,7 +861,7 @@ impl ClockOperator<'_> {
             if div_freq > self.active_limits().fro_hf_div {
                 return Err(ClockError::BadConfig {
                     clock: "fro_hf_root",
-                    reason: "exceeds max",
+                    reason: "exceeds max frequency",
                 });
             }
 


embassy-mcxa/src/clocks/mod.rs
```diff
@@ -1434,13 +1545,38 @@ impl ClockOperator<'_> {
...
+                reason: "Exceeds main_clock frequency",
             });
         }
 
+        // BEFORE we switch, update the flash wait states to the appropriate levels
+        //
+        // TODO: This calculation doesn't consider low power mode yet!
+        let wait_states = levels
+            .iter()
+            .find(|(fmax, _ws)| clk.frequency <= *fmax)
```
jamesmunns: Ah, this is incorrect, this is checking `main_clk`'s frequency, not the post-ahbdiv frequency. I'll fix this quickly.

jamesmunns: Fixed

jamesmunns:
Oh @felipebalbi - "22.3.2 Peripheral clock summary" has *another* table of clock limits, this time for functional clock limits. Would you like me to address that in another follow-up PR, or move this back to draft and add those checks?

felipebalbi:
> Oh @felipebalbi - "22.3.2 Peripheral clock summary" has _another_ table of clock limits, this time for functional clock limits. Would you like me to address that in another follow-up PR, or move this back to draft and add those checks?

We can probably handle that as a separate PR, no point delaying $subject.

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -755,6 +851,14 @@ impl ClockOperator<'_> {
                 });
             }
 
+            let div_freq = base_freq / d.into_divisor();
+            if div_freq > self.active_limits().fro_hf_div {
+                return Err(ClockError::BadConfig {
+                    clock: "fro_hf_root",
+                    reason: "exceeds max",
```
felipebalbi: nit: could you be explicit here? Maybe "exceeds max frequency"?

jamesmunns:
Merging to continue making progress, @bogdan-petru or @MathisDeroo happy to take feedback if there's anything wrong, and I can open up a follow-up PR!

PR #5239, opened 2026-01-15
[MCXA]: Fix warning when the `defmt` feature isn't selected

jamesmunns:
I noticed a warning in the cdog code when defmt isn't active.

CC @MathisDeroo


MathisDeroo:
> I noticed a warning in the cdog code when defmt isn't active.
> 
> CC @MathisDeroo

Thanks for the catch @jamesmunns !

PR #5243, opened 2026-01-17
stm32: Add non-async EXTI Input configuration

michelrandahl:
## Changes and motivation
This PR adds a 'blocking' version of STM32 EXTI input for easy use with for example RTIC or with custom interrupt handling.

This PR is replacing my old [draft PR](https://github.com/embassy-rs/embassy/pull/4028).

The PR introduces the following:
- Move exti.rs into its own module directory exti/mod.rs
- A module exti/low_level.rs that contains common PAC and low level configuration functionality used by both the async and the new blocking Exti Input.
- A module exti/blocking.rs implementing and exposing a non-async version of ExtiInput
- A new examples project stm32l4-rtic demonstrating how to use Embassy and RTIC together.

## Bug fix
The old ExtiInputFuture constructor contained, what I assume is a bug, in that the cfg flag for clearing pending bits was missing the `exti_u3` value. This motivated me to implement the two macros in the new low_level.rs module to reduce repetition.

The possible bug in the old code for reference:
https://github.com/embassy-rs/embassy/blob/a6cad0682895fee6abd6b19045017d3e13a09d80/embassy-stm32/src/exti.rs#L278

## Testing
I have tested the additions with my STM32L4R5ZI-P nucleo-board, old examples as well as new examples.

### DIFF
--- a/embassy-stm32/src/exti/mod.rs
+++ b/embassy-stm32/src/exti/mod.rs
@@ -225,7 +225,7 @@ impl<'d> ExtiInput<'d, Blocking> {
     ///
     /// # Returns
     /// A new `ExtiInput` instance with interrupts enabled
-    pub fn new<T: GpioPin + ExtiPin>(
+    pub fn new_blocking<T: GpioPin + ExtiPin>(
         pin: Peri<'d, T>,
         _ch: Peri<'d, T::ExtiChannel>, // Consumed for ownership tracking
         pull: Pull,
--- a/examples/stm32f3/src/bin/button_events.rs
+++ b/examples/stm32f3/src/bin/button_events.rs
@@ -13,6 +13,7 @@ use defmt::*;
 use embassy_executor::Spawner;
 use embassy_stm32::exti::{self, ExtiInput};
 use embassy_stm32::gpio::{Level, Output, Pull, Speed};
+use embassy_stm32::mode::Async;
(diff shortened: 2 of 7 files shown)


michelrandahl:
@lulf Thanks for the feedback, that is a good idea. I have now adjusted the implementation so ExtiInput has the type paremeter PeriMode.
Will look into fixing failing examples later this week.

lulf:
> @lulf Thanks for the feedback, that is a good idea. I have now adjusted the implementation so ExtiInput has the type paremeter PeriMode. Will look into fixing failing examples later this week.

Looks good. One note, is that the HALs often use `new_blocking` for the constructor of the blocking variant. Both to make it clear, and less obvious (in the docs as well) than having to do that `::<Blocking>` to differentiate between async and blocking.

michelrandahl:
> > @lulf Thanks for the feedback, that is a good idea. I have now adjusted the implementation so ExtiInput has the type paremeter PeriMode. Will look into fixing failing examples later this week.
> 
> Looks good. One note, is that the HALs often use `new_blocking` for the constructor of the blocking variant. Both to make it clear, and less obvious (in the docs as well) than having to do that `::<Blocking>` to differentiate between async and blocking.

ah yes thats a good idea. That also keeps the implementation more backwards compatible and saves me the hassle of having to update a lot of examples :).

PR #5250, opened 2026-01-18
Fix PSRAM max_select timing (convert microseconds to femtoseconds and divide by 64 cycles)

altundag:
This PR fixes the `max_select` timing (https://github.com/embassy-rs/embassy/issues/5269) by correcting the units, improving memory performance.


sourcebox:
The updated calculation seems to be correct.

PR #5254, opened 2026-01-19
Add a way to create optional peripherals with `embassy-hal-internal`

diondokter:
With this PR you can do `(option)Foo` in the peripheral definition to get a `Foo` peripheral that's an option in the peripherals struct. This adds the ability to make peripherals unavailable based on configuration.

I tried it so you could do `P1_30?` but couldn't figure it out.

This is an alternative to introducing a feature flag.

The pros:
- Easy clean solution
- No need for feature flags

The cons:
- Not zero-sized (but only when used)
- Slightly more complex peripheral macros

@jamesmunns added you for review of the mcxa part.

### DIFF
--- a/embassy-mcxa/src/gpio.rs
+++ b/embassy-mcxa/src/gpio.rs
@@ -322,12 +322,8 @@ impl GpioPin for AnyPin {}
 
 macro_rules! impl_pin {
     ($peri:ident, $port:expr, $pin:expr, $block:ident) => {
-        impl_pin!(crate::peripherals, $peri, $port, $pin, $block);
-    };
-
-    ($perip:path, $peri:ident, $port:expr, $pin:expr, $block:ident) => {
         paste! {
-            impl SealedPin for $perip::$peri {
+            impl SealedPin for crate::peripherals::$peri {
                 fn pin_port(&self) -> usize {
                     $port * 32 + $pin
                 }
@@ -378,15 +374,15 @@ macro_rules! impl_pin {
                 }
             }
 
-            impl GpioPin for $perip::$peri {}
+            impl GpioPin for crate::peripherals::$peri {}
 
-            impl From<$perip::$peri> for AnyPin {
-                fn from(value: $perip::$peri) -> Self {
+            impl From<crate::peripherals::$peri> for AnyPin {
+                fn from(value: crate::peripherals::$peri) -> Self {
                     value.degrade()
                 }
             }
 
-            impl $perip::$peri {
+            impl crate::peripherals::$peri {
                 /// Convenience helper to obtain a type-erased handle to this pin.
                 pub fn degrade(&self) -> AnyPin {
                     AnyPin::new(self.port(), self.pin(), self.gpio(), self.port_reg(), self.pcr_reg())
@@ -459,8 +455,8 @@ impl_pin!(P1_26, 1, 26, Gpio1);
 impl_pin!(P1_27, 1, 27, Gpio1);
 impl_pin!(P1_28, 1, 28, Gpio1);
 impl_pin!(P1_29, 1, 29, Gpio1);
-impl_pin!(crate::internal_peripherals, P1_30, 1, 30, Gpio1);
-impl_pin!(crate::internal_peripherals, P1_31, 1, 31, Gpio1);
+impl_pin!(P1_30, 1, 30, Gpio1);
+impl_pin!(P1_31, 1, 31, Gpio1);
 
 impl_pin!(P2_0, 2, 0, Gpio2);
 impl_pin!(P2_1, 2, 1, Gpio2);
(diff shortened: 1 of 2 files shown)


jamesmunns:
CC @i509VCB since you were asking about this too!

jamesmunns:
There's some stuff in mcxa that needs to be tweaked still, ~I'll open a PR!~

Pushed an update :)

jamesmunns:
@diondokter is it possible to make the syntax just like:

```rust
Option<P1_30>,
```

instead of this?

```rust
(option)P1_30,
```

Overall it looks good to me, I don't have *that* strong of an opinion on the syntax, and this does clear an annoying papercut!

diondokter:
> @diondokter is it possible to make the syntax just like:
> 
> ```rust
> Option<P1_30>,
> ```
> 
> instead of this?
> 
> ```rust
> (option)P1_30,
> ```
> 
> Overall it looks good to me, I don't have _that_ strong of an opinion on the syntax, and this does clear an annoying papercut!

Tried that too! But alas, the macro wasn't happy. Doing it manually with a token tree also doesn't work since the tt stops at angle brackets apparently.
I could make it work with a proc macro :P

Ok, so that's approval on the mcxa side.
I'll wait for an approval from dirbaio or lulf for the hal-internal side.

i509VCB:
5 bytes of size isn't too bad all things considered if I make the oscillator pins (HOSC, LOSC and ROSC compensation) in mspm0.

embassy-hal-internal/src/macros.rs
```diff
@@ -32,10 +32,34 @@ macro_rules! peripherals_definition {
...
+macro_rules! wrap_some {
+    (option, $e:expr) => {
+        Some($e)
+    };
+    (,$e:expr) => {
+        $e
+    };
+}
+
 /// Define the peripherals struct.
```
jamesmunns, line 59: I know there's not a LOT of docs here already, but it might be worth showing an example with/without the `(option)` here, AND making it clear that you're paying 1 byte per `(option)`, since the peripherals are no longer purely ZSTs.

Dirbaio:
For this problem, we're currently using feature flags and completely removing the peripheral from the struct in some places. examples:

- timers/rtcs used by the Time driver (nrf, stm32, rp)
- NFC pins as GPIO (nrf)
- RST pin as GPIO (nrf)

I don't have a strong opinion in either direction, but we should be consistent and do this the same way everywhere.

jamesmunns:
CC @felipebalbi - any opinion on making this a cfg vs a runtime check?

Thinking about it: this is a hardware-level change, so a feature flag is probably reasonable.

If nobody feels strongly about it, we could adopt that in MCXA as well.

i509VCB:
> For this problem, we're currently using feature flags and completely removing the peripheral from the struct in some places. examples:
> 
>     * timers/rtcs used by the Time driver (nrf, stm32, rp)
> 
>     * NFC pins as GPIO (nrf)
> 
>     * RST pin as GPIO (nrf)
> 
> 
> I don't have a strong opinion in either direction, but we should be consistent and do this the same way everywhere.

Timer for time driver obviously has to be not given.

Reset is fine to cfg guard imo.

Oscillators depend completely on the runtime config. You could of course do a cfg forcing a oscillator setup or crystal setup, but that's more complicated in the HAL.

jamesmunns:
Lemme do an alternative PR for MCXA using the feature flag approach, at least as a conversation piece.

jamesmunns:
@i509VCB not sure about the MSPM0, but at least in the MCX-A, the SOSC pins are fixed to two specific pins, which means we don't necessarily need any runtime input.

Do you think you could do something like https://github.com/embassy-rs/embassy/pull/5255 ?

i509VCB:
> @i509VCB not sure about the MSPM0, but at least in the MCX-A, the SOSC pins are fixed to two specific pins, which means we don't necessarily need any runtime input.
> 
> Do you think you could do something like #5255 ?

Yes mspm0 uses fixed pins. There is a 32k and "high speed" oscillator. ROSC only makes sense to enable by the user acted init imo? Or should ROSC be a alternative cfg? 

Oscillator pins could be used as GPIO of course, so keep that in mind

felipebalbi:
> CC @felipebalbi - any opinion on making this a cfg vs a runtime check?
> 
> 
> 
> Thinking about it: this is a hardware-level change, so a feature flag is probably reasonable.
> 
> 
> 
> If nobody feels strongly about it, we could adopt that in MCXA as well.

Given the rest of the conversation, I suppose feature flag is very reasonable. No strong opinions either way, except that feature flags already exist and folks are already used to them.

jamesmunns:
I'll probably go ahead and close this in favor of #5255 - I think there *could* still be a case where it makes sense to make this option-able, and where there might be motivation to do this at runtime instead of with features. However: that day is not today and the MCXA doesn't need it for now.

Thank you @diondokter for sketching this out (and other folks for discussing!), I could see us picking this back up in the future if we ever need it :)

PR #5257, opened 2026-01-19
mspm0: cfg guard pins used for reset and swd

i509VCB:


i509VCB:
@Iooon is there a reason why L1306 uses PA20 (an SWD pin) in the ADC example? The thermistor on dev board uses PA25.

I changed to PA25 here.

Iooon:
> @Iooon is there a reason why L1306 uses PA20 (an SWD pin) in the ADC example? The thermistor on dev board uses PA25.
> 
> I changed to PA25 here.

Yeah no problem as it is an example this should be better.

PR #5259, opened 2026-01-20
wpan: simplify sequencer

xoviat:


leftger (approved):
Much cleaner than my implementation

PR #5261, opened 2026-01-20
stm32/wba: Add AES, SAES, and PKA cryptographic drivers

leftger:
## Summary

- Add AES driver (v3b) with ECB, CBC, CTR, GCM, CCM, and GMAC modes
- Add SAES driver (v1a) for secure AES with hardware key derivation (DHUK/BHK)
- Add PKA driver (v1a) for ECC operations: ECDSA sign/verify, ECDH key agreement, point validation
- All drivers use blocking mode with interrupt-driven completion
- GCM/CCM implementations follow ST HAL reference for proper phase handling

Tested on STM32WBA65 hardware.


leftger:
I was trying to see what the failed status checks were but I wasn't able to determine the root cause. Any help is appreciated @xoviat .

Issue #5260, opened 2026-01-20
stm32 i2c v1 async respond_to_write not working as expected


ChocolateLoverRaj:
I am using embassy from git (because the latest released version doesn't have i2c peripheral mode) to program a stm32f103c8 to emulate a MCP23017. On the controller side of i2c I am using a esp32c3. I am using 400KHz (configured on both the esp32c3 and the stm32) for the i2c frequency.

On the esp32c3, I'm doing
```rs
embedded_hal_async::i2c::I2c::write(&mut mcp23017, 0x20, &[10, 20])
        .await
        .unwrap();
```
as a test.

On the stm32, I have this:
```rs
bind_interrupts!(struct Irqs {
    I2C1_EV => i2c::EventInterruptHandler<peripherals::I2C1>;
    I2C1_ER => i2c::ErrorInterruptHandler<peripherals::I2C1>;
});
```
```rs
let mut i2c = I2c::new(p.I2C1, p.PB6, p.PB7, Irqs, p.DMA1_CH6, p.DMA1_CH7, {
    let mut config = i2c::Config::default();
    config.frequency = khz(400);
    config
})
.into_slave_multimaster(SlaveAddrConfig::basic({
    let base_address = 0x20;
    let least_significant_bits = 0b000;
    base_address | least_significant_bits
}));

loop {
    info!("Ready to receive I2C commands");
    let command = match i2c.listen().await {
        Ok(command) => command,
        Err(e) => {
            warn!("I2C error: {}", e);
            continue;
        }
    };
    match command.kind {
        SlaveCommandKind::Read => {}
        SlaveCommandKind::Write => {
            // info!("I2C write command started");
            let mut buffer = [Default::default(); 512];
            let bytes_read = i2c.respond_to_write(&mut buffer).await.unwrap();
            info!("Received {} bytes: {}", bytes_read, &buffer[..bytes_read]);
        }
    }
}
```

With the async  functions, it  results in this:
```
[...]

PR #5266, opened 2026-01-20
stm32 i2c v1 - expose async receive_byte and transmit_byte functions

ChocolateLoverRaj:
This is part of my efforts to use a stm32 to emulate a MCP23017. Related: 
- https://github.com/stm32-rs/stm32f1xx-hal/discussions/558
- #5260 

Emulating a MCP23017 involves processing reads and writes where the amount of data read or written is not known in advance. This is why I (am pretty sure) I need single byte functions to process read and write requests.

Some things that I think need to be reviewed:
- Is it ok to expose thse functions as `pub`?
- Did I implement the async properly (I copied and pasted from the async `listen` fn)? 
- Is it okay  to not handle zero byte reads?

I am using these changes in my MCP23017 emulator project and they work.

### DIFF
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -12,7 +12,7 @@ use core::iter;
 use core::marker::PhantomData;
 
 #[cfg(i2c_v1)]
-pub use _version::ReceiveResult;
+pub use _version::{ReceiveResult, TransmitResult};
 pub use config::*;
 use embassy_hal_internal::Peri;
 use embassy_sync::waitqueue::AtomicWaker;
--- a/embassy-stm32/src/i2c/mod.rs
+++ b/embassy-stm32/src/i2c/mod.rs
@@ -11,6 +11,8 @@ use core::future::Future;
 use core::iter;
 use core::marker::PhantomData;
 
+#[cfg(i2c_v1)]
+pub use _version::ReceiveResult;
 pub use config::*;
 use embassy_hal_internal::Peri;
 use embassy_sync::waitqueue::AtomicWaker;
--- a/embassy-stm32/src/i2c/v1.rs
+++ b/embassy-stm32/src/i2c/v1.rs
@@ -764,7 +764,7 @@ impl Duty {
 
 /// Result of attempting to send a byte in slave transmitter mode
 #[derive(Debug, PartialEq)]
-enum TransmitResult {
+pub enum TransmitResult {
     /// Byte sent and ACKed by master - continue transmission
     Acknowledged,
     /// Byte sent but NACKed by master - normal end of read transaction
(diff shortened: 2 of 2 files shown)


xoviat:
Do the changes that @HybridChild proposed work?

ChocolateLoverRaj:
I didn't test @HybridChild 's proposed changes yet, but either way for my use case I need a byte-by-byte way of sending and receiving, without discarding any extra data when receiving or sending default data when sending.

HybridChild:
Hi @ChocolateLoverRaj

Before we add new public API surface, I'd like to understand if the existing functions can work for your use case. My [stm32-i2c-v1-slave-mode](https://github.com/HybridChild/embassy/tree/stm32-i2c-v1-slave-mode) branch should address the issues you reported in #5260 - they now correctly return the actual number of bytes transferred.

Could you try the following pattern for your MCP23017 emulation?

```rust
loop {
    let command = i2c.listen().await?;

    match command.kind {
        SlaveCommandKind::Write => {
            let mut buffer = [0u8; 32];
            let bytes_received = i2c.respond_to_write(&mut buffer).await?;

            // First byte is register address, rest is data
            let register = buffer[0];
            let data = &buffer[1..bytes_received];

            // Process write to register...
        }
        SlaveCommandKind::Read => {
            // Prepare response based on previously selected register
            let response = get_register_data(current_register);
            let bytes_sent = i2c.respond_to_read(&response).await?;
        }
    }
}
```

With the fixes, `bytes_received` should now return the actual count (e.g., `2` instead of `512`), and the buffer should contain the correct data without duplicates.

Is there a specific aspect of MCP23017 emulation that requires processing bytes individually during the transaction, rather than receiving the complete write and then processing it?

ChocolateLoverRaj:
> Is there a specific aspect of MCP23017 emulation that requires processing bytes individually during the transaction, rather than receiving the complete write and then processing it?

Technically with an MCP23017 you can make a read or write operation go on forever. Depending on how you configure the MCP23017's registers, after each read or write the register address pointer will either stay the same, toggle between A and B, or cycle through all registers.

With the existing API exposed by embassy, when the controller writes, I could program the stm32 to receive all of the  bytes in one function call and then process all of the bytes after that. When the controller reads, I would have to pre-compute values to  create the buffer, `respond_to_read` with that buffer, and then update the register address pointer based on how many of those bytes were actually transferred.

*Practically*, I think I could implement a mostly spec-compliant MCP23017 when processing buffers of multiple bytes at a time. Because practically, whoever is using this MCP23017 is not going to be doing read and write transactions that are larger than 22 bytes. However, there is one practical use case I can think of, which is to have a continuous i2c read transaction to continuously poll the state of GPIO pins. I won't be doing that because I will be using interrupts.

To fully match the data sheet for MCP23017, there are two issues with embassy's existing API:
- `respond_to_read` uses padding bytes instead of letting the caller call the function again if the i2c controller still wants even more bytes after the buffer is sent
- `respond_to_write` acknowledges and  discards extra bytes instead of letting the  caller call the function again to process remaining bytes

Overall I think being able to respond to a transaction byte-by-byte, or part-by-part would be valuable to the `pub` API and currently it's not possible to do that.

HybridChild:
I'm wondering if both issues you mention could be addressed by simply using a larger buffer?

- **For writes**: If you use a buffer of, say, 128 bytes, `respond_to_write` will receive all bytes in one call (for any practical transaction size). The "discard extra bytes" behavior only triggers when the master writes more than your buffer size.

- **For reads**: Similarly, you can pre-compute a buffer with the expected cycling register values. Since the register cycling pattern is deterministic based on how the device is configured, you can fill a buffer with multiple cycles worth of data. The padding only kicks in if the master reads beyond that.

You mentioned that practically, transactions won't exceed 22 bytes so a 64-128 byte buffer would comfortably handle that with room to spare.

The byte-by-byte API adds complexity to the public interface. I'd like to understand if there's a concrete use case that genuinely can't be solved with a reasonably-sized buffer. The "infinite continuous read" scenario you mentioned is quite unusual (typically you'd do repeated short transactions), and you said you won't be using it anyway.

Could you try your MCP23017 emulation with a larger buffer and see if it meets your needs?

ChocolateLoverRaj:
Yes, using a buffer with more than enough bytes that would be practical (`32` is enough  for my use case) does meet my needs, because I will not be using the "infinite continuous read" feature. So I would be fine with closing this PR if it is determined to be an unnecessary feature.

leftger:
I think at this point it's very likely this is no longer needed. I'm closing this PR.

PR #5267, opened 2026-01-20
[MCXA] set gpios to disabled state on Drop

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/111

### DIFF
--- a/embassy-mcxa/src/lpuart/buffered.rs
+++ b/embassy-mcxa/src/lpuart/buffered.rs
@@ -212,7 +212,9 @@ impl<'a> BufferedLpuart<'a> {
         Ok(())
     }
 
-    /// Create a new full duplex buffered LPUART
+    /// Create a new full duplex buffered LPUART.
+    ///
+    /// Any external pin will be placed into Disabled state upon Drop.
     pub fn new<T: Instance>(
         inner: Peri<'a, T>,
         tx_pin: Peri<'a, impl TxPin<T>>,
@@ -239,7 +241,9 @@ impl<'a> BufferedLpuart<'a> {
         Ok(Self { tx, rx })
     }
 
-    /// Create a new buffered LPUART instance with RTS/CTS flow control
+    /// Create a new buffered LPUART instance with RTS/CTS flow control.
+    ///
+    /// Any external pin will be placed into Disabled state upon Drop.
     pub fn new_with_rtscts<T: Instance>(
         inner: Peri<'a, T>,
         tx_pin: Peri<'a, impl TxPin<T>>,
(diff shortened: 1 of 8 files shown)


embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -613,6 +613,28 @@ impl<'a> BufferedLpuartRx<'a> {
...
 }
 
+// ============================================================================
+// DROP TRAIT IMPLEMENTATIONS
+// ============================================================================
+
+impl<'a> Drop for BufferedLpuartTx<'a> {
+    fn drop(&mut self) {
+        self._tx_pin.set_as_disabled();
+        // if let Some(cts_pin) = &self._cts_pin {
```
jamesmunns: Dead code

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -613,6 +613,28 @@ impl<'a> BufferedLpuartRx<'a> {
...
+        //     cts_pin.set_as_disabled();
+        // }
+    }
+}
+
+impl<'a> Drop for BufferedLpuartRx<'a> {
+    fn drop(&mut self) {
+        self._rx_pin.set_as_disabled();
+        // if let Some(rts_pin) = &self._rts_pin {
```
jamesmunns: Dead code

felipebalbi: fixed

embassy-mcxa/src/lpuart/buffered.rs
felipebalbi: fixed

felipebalbi:
> * We might want to document on the HAL types (e.g. `Lpuart`/`UartRx`) that the pins WILL be set to the `Disabled` state on drop. This isn't consistently the case with other HALs, e.g. embassy-rp doesn't really do anything with them at drop.

I agree. Done

> * We might (later) want to have a specific "drop the hal peripheral but leave the pins in some other state", for example dropping a `UartTx`, but leaving the pin in an output-high state, if we don't want a bus to go idle (this is just a random example), but just documenting that we DON'T do that now is enough IMO.

Right, this will be a per-driver thing. Potentially involves board layout matters too. We may want to eventually allow an application to provide hints to the HAL.

> * We currently don't disable/reset the peripherals themselves on drop, e.g. by re-gating the functional clocks and/or resetting the peripherals. This is _probably_ fine for now, but will be more relevant once we start caring about low power stuff.

yeah, I wanted to ask: do we want to make that decision now? That is, make sure all drivers gate clocks on Drop? It would certainly improve power consumption.

> * There's maybe a case for peripherals holding a `Flex` instead of an `AnyPin`, since then they get the `Drop` for free. That being said, idk if there would be some other trade off there. Something to think about in the future I guess.

It's been mostly a pattern that every HAL follows, holding a `Peri<'d, AnyPin>`. We can change that to a `Flex` if we desire, but then we would miss the opportunity for per-driver pin state. Everything would always go to `Disabled`.

PR #5276, opened 2026-01-21
[MCXA]: Implement WakeGuard

jamesmunns:
This PR adds basic "reference counting guard" for peripherals with a high-power-only clock source. Right now, this doesn't actually inhibit or affect any clock calculations, but is just "plumbing".

### DIFF
--- a/embassy-mcxa/src/clocks/mod.rs
+++ b/embassy-mcxa/src/clocks/mod.rs
@@ -153,7 +153,9 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
 // Structs/Enums
 //
 
-#[must_use = "Wake Guard must be kept!"]
+/// A guard that will inhibit the device from entering deep sleep while
+/// it exists.
+#[must_use = "Wake Guard must be kept in order to prevent deep sleep"]
 pub struct WakeGuard {
     _x: (),
 }
@@ -162,14 +164,14 @@ impl WakeGuard {
     /// Create a new wake guard, that increments the "live high power token" counts.
     ///
     /// This is typically used by HAL drivers (when a peripheral is clocked from an
-    /// active-mode-only source) to inihibit sleep, OR by application code to prevent
+    /// active-mode-only source) to inhibit sleep, OR by application code to prevent
     /// deep sleep as well.
     pub fn new() -> Self {
         _ = LIVE_HP_TOKENS.fetch_add(1, Ordering::AcqRel);
         Self { _x: () }
     }
 
-    /// Helper method to potentially create a guard if necessary for a clock
+    /// Helper method to potentially create a guard if necessary for a clock.
     pub fn for_power(level: &PoweredClock) -> Option<Self> {
         match level {
             PoweredClock::NormalEnabledDeepSleepDisabled => Some(Self::new()),


embassy-mcxa/src/clocks/mod.rs
```diff
@@ -151,6 +153,53 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for WakeGuard {
+    fn drop(&mut self) {
+        let old = LIVE_HP_TOKENS.fetch_sub(1, Ordering::AcqRel);
+        // Ensure we didn't just underflow.
+        assert!(old != 0);
```
felipebalbi, line 201: Is this correct? Shouldn't you `assert!(old > 0)` instead? I mean, once you reach `0`, sleep would be allowed, but maybe we want to catch the possibility of a double decrement of the `AtomicUsize`? Then again, I suppose that would be just an academic possibility given that you cannot `Drop` twice.

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -151,6 +153,53 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
 // Structs/Enums
 //
 
+#[must_use = "Wake Guard must be kept!"]
```
felipebalbi: Can we be more specific here?

```suggestion
#[must_use = "Wake Guard must be kept while preventing DeepSleep"]
```

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -151,6 +153,53 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+pub struct WakeGuard {
```
felipebalbi, line 159: I'm wondering if we will ever need to carry around some `T` driver specific data. Then we could:

```rust
pub struct WakeGuard<T> {
    data: T
}

pub type WakeToken = WakeGuard<()>;
```

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -151,6 +153,53 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+}
+
+impl WakeGuard {
+    /// Create a new wake guard, that increments the "live high power token" counts.
+    ///
+    /// This is typically used by HAL drivers (when a peripheral is clocked from an
+    /// active-mode-only source) to inihibit sleep, OR by application code to prevent
+    /// deep sleep as well.
+    pub fn new() -> Self {
+        _ = LIVE_HP_TOKENS.fetch_add(1, Ordering::AcqRel);
```
felipebalbi, line 170: Would `Release` suffice?

(and use `Acquire` for the eventual `load` operation)

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -151,6 +153,53 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+
+impl Default for WakeGuard {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for WakeGuard {
+    fn drop(&mut self) {
+        let old = LIVE_HP_TOKENS.fetch_sub(1, Ordering::AcqRel);
```
felipebalbi, line 199: Would `Release` suffice?

(and use `Acquire` for the eventual `load` operation)

embassy-mcxa/src/clocks/mod.rs
jamesmunns: We want this check to be != zero, because `fetch_sub` returns the OLD value. If we observe `old = 0`, that means that we JUST underflowed from `0` to `usize::MAX`.

embassy-mcxa/src/clocks/mod.rs
jamesmunns: I think we could do things that way, but it's a little backwards from what we have. Right now, we have a `Uart` that holds a `WakeGuard`, rather than a `WakeGuard<Uart>`. Since we only *sometimes* actually have a WakeGuard, I think the current API makes more sense, unless we want to have separate constructors to create a low-power-friendly driver.

embassy-mcxa/src/clocks/mod.rs
jamesmunns: It's *possible* that we could just use `Release` or `Relaxed` here, but on cortex-m I don't think it would make a practical difference, so I'd prefer to keep AcqRel at least until we have the full impl and can reason about it.

FWIW: This is a `fetch_add`, which means we are doing a "load/modify/store". Using `Release` here would mean "`Relaxed` load" + "`Release` store", which I'm not sure is what we want. If we don't have the "Acq" part of the `AcqRel`, I'm not sure we're guaranteed to see ALL previous writes that were made with `Release` ordering.

embassy-mcxa/src/clocks/mod.rs
jamesmunns: Same

embassy-mcxa/src/clocks/mod.rs
jamesmunns: Addressed

embassy-mcxa/src/trng.rs
```diff
@@ -42,6 +42,7 @@ pub struct Trng<'d, M: Mode> {
 
 impl<'d, M: Mode> Trng<'d, M> {
     fn new_inner(_peri: Peri<'d, TRNG0>, config: Config) -> Self {
+        // No clock: No WakeGuard!
         _ = unsafe { enable_and_reset::<TRNG0>(&NoConfig) };
```
jamesmunns, line 46: @MathisDeroo or @bogdan-petru: Using TRNG as an example, the reference manual "24.5 Module operation in low power modes" says that the TRNG is "Static" in Deep Sleep mode.

Is it valid to:

* Enable and clock the TRNG (or any other peripheral "Static" in deep sleep)
* Use the TRNG in active mode
* Go to deep sleep mode
* Wake up from deep sleep mode
* Continue to use the TRNG in active mode (without re-initalizing)

OR, do we need to:

* Enable and clock the TRNG (or any other peripheral "Static" in deep sleep)
* Use the TRNG in active mode
* Disable and gate the TRNG BEFORE going to deep sleep
* Go to deep sleep mode
* Wake up from deep sleep mode
* Re-Enable/re-clock/re-initialize the TRNG
* Use the TRNG in active mode

Just checking whether ALL peripherals listed as "static" are safe to "hibernate" while we're in deep sleep and they are unclocked, or if re-initialization is necessary.

jamesmunns: Spoke with @MathisDeroo, we believe the first case is true (we can just pause activity in deep sleep), we'll update if this is not the case.

PR #5279, opened 2026-01-21
[MCXA] reset-reason: use SSRS register

felipebalbi:
Instead of using the read-only SRS reset reason register, we can use SSRS, which allows us to follow up with a clear operation, erasing old reasons from the register.

Because we can still get more than one bit set, the Iterator is kept.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/117

### DIFF
--- a/embassy-mcxa/src/reset_reason.rs
+++ b/embassy-mcxa/src/reset_reason.rs
@@ -17,51 +17,54 @@ static RESET_REASON: AtomicU32 = AtomicU32::new(0);
 pub fn reset_reason() -> ResetReasonRaw {
     let regs = unsafe { &*crate::pac::Cmc::steal() };
 
-    let mut reason = RESET_REASON.load(Ordering::Acquire);
+    let reason = critical_section::with(|_| {
+        let mut r = RESET_REASON.load(Ordering::Relaxed);
 
-    if reason == 0 {
-        // Read status
-        let srs = regs.srs().read().bits();
+        if r == 0 {
+            // Read status
+            r = regs.srs().read().bits();
 
-        // Clear status
-        regs.ssrs().write(|w| {
-            w.wakeup()
-                .clear_bit_by_one()
-                .por()
-                .clear_bit_by_one()
-                .warm()
-                .clear_bit_by_one()
-                .fatal()
-                .clear_bit_by_one()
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/reset_reason.rs
```diff
@@ -8,7 +8,46 @@
 /// (CMC).
```
jamesmunns, line 16: Probably want to update the function doc comment that this function is now side-effectful and will only return the reset reason the *first* time you call it, rather than *every* time you call it.

The module level doc comment also needs to be updated, since it says `SRS` not `SSRS`.

felipebalbi: Is it worth adding a `OnceCell` or `StaticCell` to the HAL to keep the reset reason around?

jamesmunns: Offhand: Probably not? Just good for people to know what they're getting. We can always add it later if we feel like it, or just read from SRS?

felipebalbi: Added a `OnceLock` and updated the module doc.

embassy-mcxa/src/reset_reason.rs
```diff
@@ -3,13 +3,67 @@
...
+static RESET_REASON: AtomicU32 = AtomicU32::new(0);
 
 /// Reads the most recent reset reason from the Core Mode Controller
 /// (CMC).
 pub fn reset_reason() -> ResetReasonRaw {
     let regs = unsafe { &*crate::pac::Cmc::steal() };
-    let srs = regs.srs().read().bits();
-    ResetReasonRaw(srs)
+
+    let mut reason = RESET_REASON.load(Ordering::Acquire);
```
jamesmunns: Very pedantic nit: You might want to do this in a critical section, and use Relaxed ordering instead (`c-s::with` guarantees acquire ordering on entry and release ordering on exit).

TECHNICALLY you could race with this, and store over the reset reason back to zero. Instead of being clever, let's just smash it with a CS hammer.

felipebalbi: done

jamesmunns (approved):

PR #5284, opened 2026-01-22
nrf/pwm: fix put pins into idle state on drop

### LINKED ISSUE #5272: [embassy-nrf] PWM peripheral using power after drop
Hi,

I've been investigating the power consumption of app an found the PWM peripheral to be the likely culprit, without ever instantiating PWM idle current dropped as low as **0.05mA**,  with PWM(dropped long before measurement)  idle current only went as low as **0.39mA**

Minimal Example: https://github.com/mdrssv/embassy-nrf-pwm-power

mdrssv:
Fixes #5272

Had to drop:
```
pin.conf().write(|_| ());
```
because in my use-case the gpio is low active, I assume writing `()` into the register zeroes it thus setting the level to low

### DIFF
--- a/embassy-nrf/src/pwm.rs
+++ b/embassy-nrf/src/pwm.rs
@@ -975,26 +975,19 @@ impl<'a> Drop for SimplePwm<'a> {
 
         self.disable();
 
-        if let Some(pin) = &self.ch0 {
-            pin.set_low();
-            pin.conf().write(|_| ());
-            r.psel().out(0).write_value(DISCONNECTED);
-        }
-        if let Some(pin) = &self.ch1 {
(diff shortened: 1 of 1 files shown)


lulf (commented):
Thanks! Maybe it's worth having a default here, and make another constructor for providing the per-channel idle levels?

mdrssv:
High-z also solves the problem without the need to pass any additional state around

0e4ef622:
Might want to make the same change to `SequencePwm` too.

lulf:
bender run

lulf (approved):

mdrssv:
Removed unused import CI should pass now

PR #5288, opened 2026-01-22
stm32n6: add psram example and changes needed to xspi

tarfu:
This adds a psram example/test for the STM32N6570-DK.

Relies on https://github.com/embassy-rs/stm32-data/pull/723 to rewrite raw pointer access in the n6.rs to pac access.

### DIFF
--- a/examples/stm32n6/src/bin/xspi_psram.rs
+++ b/examples/stm32n6/src/bin/xspi_psram.rs
@@ -43,6 +43,7 @@ use embassy_time::Timer;
 use {defmt_rtt as _, panic_probe as _};
 
 /// APS256XX Commands
+#[allow(dead_code)]
 mod aps256_cmd {
     pub const READ: u32 = 0x00; // Synchronous Read (used by C HAL for memory-mapped)
     pub const READ_LINEAR_BURST: u32 = 0x20; // Linear Burst Read


TomDeRybel:
Hello Tarfu,
I was reading your code and, regarding the issue with 8-bit sub-words, there is something that may be at play here. (I'm presuming that you did not encounter this information before.)

In RM0477 (for the STM32H7RS), there is an interesting table on p.1030, Chapter 24.5: "Address alignment and data number", Table 213: "Address alignment cases"
In RM0486 (For the STM32N6), the same table appears on p.1326, Chapter 28.5: "Address alignment and data number", Table 199: "Address alignment cases"

(At a quick glance, the tables for both chips appear identical.)

In this table, it is shown that for memory-mapped operations using octa/hexa DTR, for writes, the address must be even/aligned and the number of bytes even/4. Otherwise, errors occur. This behavior may explain why byte-wise access doesn't work reliably.

I hope this is of help,

Best regards,
Tom

tarfu:
Thanks for pointing this out, have to see how we could incorporate this into the xspi peripheral in a good way if it is identical for all chips with xspi.

My qoal is still a fsbl with relative easily changeable flash chip. But right now I'm pretty stuck and right now shifted a bit to first have a working probe-rs flash-algorithm for the n6-dk.

PR #5293, opened 2026-01-25
stm32: FMC SDRAM Overhaul, NAND and NOR/PSRAM/SRAM Support

northernpaws:
This PR intends to overhaul the STM32 FMC driver to resolve several issues with the current driver, and to fill it out with a variety of missing features. In particular, the FMC driver in it’s current state doesn’t support the NAND and NOR/PSRAM/SRAM banks which are essential for features advertised for some of the high-performance STM32 devices, such as driving an 8080 parallel display with the FMC.

Some of the issues that this PR aims to resolve:
- The `stm32-fmc` crate backing the current driver appears to be abandoned, and is severely feature-incomplete.
- The `stm32-fmc` crate also supplies its own set of registers that are not backed by metapac, so their correctness is not validated across STM32 families.
- Only a single SDRAM bank can be used/configured, while the FMC allows for dual banks.
- Several unvalidated fields are typed as u32’s when their corresponding registers are u8’s.
- Several configuration fields are loosely-typed (such as memory data width), whereas their corresponding register fields are strongly typed.
- The `stm32-fmc` crate configures global SDRAM settings that apply to both banks in the configuration for a single bank.
- The `fmc_sdram_constructor` only allows configuring the pin alternate functions for SDRAM, and doesn’t allow for multi-protocol buses like the FMC supports.
- **There is no support for the NOR/PSRAM/SRAM and NAND controllers**

With this PR I aim to:
- [x] ~~Integrate the FMC driver functionality into the `embassy-stm32` crate as a new driver re-written from scratch, removing the need for an external unmaintained, and feature-incomplete dependency.~~
- [x] ~~Use the STM32 metapac to configure the registers, taking advantage of the strong typing and compile-type correctness safety offered by it.~~
- [x] ~~Integrate support for dual SDRAM banks to allow for full use of the FMC.~~
- [x] ~~Correctly type the SDRAM timing fields so they can’t allow values out-of-bounds.~~
- [x] ~~Strongly-type the SDRAM configuration fields where possible (i.e. with enums), so invalid configuration are not possible.~~
- [x] ~~Change SDRAM register path on newer `fmc_v4` MCUs.~~
- [x] ~~Add a NAND controller driver.~~
- [ ] Re-write the `fmc_sdram_constructor` functionality to allow for configuring the pins for the various kinds of devices/protocols the FMC supports, including multiple functions on the same bus.
- [x] ~~Add a NOR/PSRAM/SRAM driver.~~
- [ ] Add some drivers for functionality derived from the NOR/PSRAM/SRAM driver, such as Intel 8080-style buses for displays advertised by ST.
- [x] ~~Add mapping-aware base address resolution for memory-mapped addresses, so that the driver can derive the correct address if the end-user has enabled swapping of the SDRAM and NOR/PSRAM/SRAM banks.~~
 - [ ] Ensure memory bank base addresses are correct across all STM32 lines.
 - [ ] Gate SDRAM 32-bit memory data width based on availability on some STM32 lines. 

I’ll update that list with any other improvements or additions as I come across them!

Things that will need to be tested:
- [ ] SDRAM driver (I can do this on some of my boards)
- [ ] NAND driver (I currently do not have a NAND device to test with)
- [ ] NOR/PSRAM/SRAM driver (I can do this to an extent)

This should hopefully get the state of the STM32 FMC driver on-par (or better!) then the FMC drivers for HALs from other frameworks/languages, greatly improving on this one area where the Embassy STM32 HAL is currently lacking compared to others.


northernpaws:
I got most of the SDRAM driver written over the last few hours, but there are still a few areas I think it could be improved.

I need to update most of the examples to support the new driver, so reworking the `fmc_sdram_constructor` functionality is next on my list because it breaks the new SDRAM driver layout as it's currently written, and doesn't allow for initializing multiple FMC banks of different types (i.e. SDRAM _and_ an SRAM).

embassy-stm32/src/fmc/v1.rs
northernpaws, line 1: Found there were enough differences with `fmc_v1x3` used at least on the `STMF4`s that it was worth it to split some of the FMC types for bank mapping out into versioned modules to make it a bit simpler to declare. Those versioned types are imported into the root crate and then the root interface can just run off a couple `#[cfg(...)]` predicates and still be clear to read.

Dirbaio:

Issue #5281, opened 2026-01-21
`embassy_stm32::init` is a bit overzealous (setting pin modes, interrupt handlers..)


tnibler:
The situation: bootloader has setup external flash to execute the application from, which uses GPIO pins. When the application calls `init`, [all pins are set to analog mode](https://github.com/embassy-rs/embassy/blob/4bb5ccaa10abdeee72f243c77d30e97b8915f850/embassy-stm32/build.rs#L966) which is a bit of a problem for said SPI flash.

More generally, there's hundreds of chips that people will run in all kinds of ways, so a single piece of generated code serving everyone equally seems like a big ask (see #4597, #3741, #5072).

The simplest option would be to split the function up and allow users to only call the pieces they need, should the combined `init` doesn't work for them.

Another question (though that's just taste really) is how much `init` really needs to do. Personally if I really wanted to set all pins to analog, I wouldn't mind writing this loop myself, so I'm not sure it needs to be hardcoded by the library:

```
    crate::pac::#port.moder().modify(|w| {
        w.set_moder(0, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(1, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(2, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(3, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(4, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(5, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(6, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(7, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(8, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(9, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(10, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(11, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(12, crate::pac::gpio::vals::Moder::ANALOG);
        w.set_moder(15, crate::pac::gpio::vals::Moder::ANALOG);
    });
```

xoviat:
you can add a configuration option to disable this.

PR #5294, opened 2026-01-25
stm32: Add cfg to disable GPIOs inited to analog

tnibler:
This makes initialization of all GPIO pins to analog optional through a cargo feature.

Related to #5281

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -220,7 +220,7 @@ proptest-state-machine = "0.3.0"
 
 
 [features]
-default = ["rt", "gpio-init-analog"]
+default = ["rt"]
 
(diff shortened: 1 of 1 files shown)


embassy-stm32/Cargo.toml
```diff
@@ -220,7 +220,7 @@ proptest-state-machine = "0.3.0"
 
 
 [features]
-default = ["rt"]
+default = ["rt", "gpio-init-analog"]
```
diondokter: This keeps the previous behavior.

But do we want that? IMO this is better as an opt-in feature instead of opt-out.
But it's not my call to make.

tnibler: I personally don't think we want this :D didn't want to break defaults on my own though

PR #5295, opened 2026-01-25
stm32: docs clarifying I2S clear/reset losing WSCLK sync

tnibler:
STM32Cube has the same functions and they confuse people there too. They just stop DMA at an unknown point, and you have no way of knowing what your buffers contain, since the WS clock isn't linked to sample order anymore.


liarokapisv:
This is in principle possible to solve but requires keeping stride / num of channels metadata in the dma ring buffer. I plan to address this when I have the time.

tnibler:
I've tried a little bit but couldn't figure it out :/ On overrun we know the next sample will be the same channel as the last one that fit in the FIFO, so that's easy. 

For a manual ringbuffer reset/clear, if `remaining_transfers` at that instant is odd the channels get swapped right? Seems to be accurate most of the time at least. Not sure when/how to read that count though without getting racey, I assume that's the issue.

PR #5296, opened 2026-01-25
Add basic MCAN driver for MSPM0

magmastonealex:
This commit implements a simple driver for the MCAN peripheral on MSPM0, supporting both blocking and non-blocking mode.

An example is included for the MSPM0G3107 (the only part I have access to for testing), but this _should_ work on all parts with a CANFD/MCAN peripheral.

As discussed in Matrix - this also configures SYSPLL in a slightly silly way to allow for providing a valid functional clock without requiring an external oscillator. In testing, this is fine at moderate bitrates.

This driver is intended to be a starting point, and does not yet include a variety of useful features for a well-rounded driver. In particular it does not include:

 - Async operation (in progress, but not yet implemented here)
 - CAN-FD
 - Filtering
 - TX confirmation
 - More complex clocking
 - Bitrate calculation
 - Automatic bus-off recovery

I hope to work towards implementing these features, but I can't commit to any sort of timeline.


embassy-mspm0/Cargo.toml
```diff
@@ -69,16 +69,17 @@ critical-section = "1.2.0"
 micromath = "2.0.0"
 
 # mspm0-metapac = { version = "" }
-mspm0-metapac = { git = "https://github.com/mspm0-rs/mspm0-data-generated/", tag = "mspm0-data-d9f54366c65cec47957065f42fe04542b852a8cd" }
-rand_core = "0.9"
+mspm0-metapac = { git = "https://github.com/mspm0-rs/mspm0-data-generated/", tag = "mspm0-data-566cf58760f60a9126c957ae9b6b3364e46ed595" }
+bitfield = "0.19.4"
```
magmastonealex, line 74: `bitfield` is also used by `embassy-net-adin1110` and I couldn't find other similar bitfield crates in use within `embassy`. This is used to construct some of the fields for message RAM - I'm open to using a different crate if preferable.

embassy-mspm0/build.rs
```diff
@@ -710,6 +719,25 @@ fn generate_pin_trait_impls() -> TokenStream {
...
+        .iter()
+        .find(|p| p.name == "SYSCTL")
+        .expect("no SYSCTL peripheral");
+
+    match sysctl.version {
+        Some("g350x_g310x_g150x_g110x") | Some("g351x_g151x") => {
+            cfgs.enable("sysctl_syspll");
+        }
+        _ => {}
+    }
```
magmastonealex, line 738: This feels a little strange to me. Open to suggestions on how better to identify if SYSPLL is available.

embassy-mspm0/src/lib.rs
```diff
@@ -173,6 +176,134 @@ impl Default for Config {
...
+/// Read the PLL startup calibration values from the FACTORY region,
+/// given an expected f_loopin frequency.
+/// Returns SYSPLLPARAM0, SYSPLLPARAM1.
+#[cfg(sysctl_syspll)]
+fn load_pll_values(f_loopin: u32) -> Option<(u32, u32)> {
+    // TODO: should these be found via the PAC instead of hard-coded?
+    // From looking at the G series TRM, these addresses are constant,
+    // but it still feels wrong.
+    // Taken from table "Table 1-135. FACTORYREGION_TYPEG Registers"
+    // (constants identical to Table 1-116. FACTORYREGION_TYPEA Registers for PLL.)
```
i509VCB, line 188: Longer term I do want to expose the factory region in this crate in a `factory` module. Haven't gotten to it yet.

embassy-mspm0/build.rs
i509VCB: We do have metadata for a clock tree from sysconfig, just haven't plumbed it up yet.

magmastonealex:
CI fails due to:

```
error[E0432]: unresolved import `rand_core`
  --> /ci/code/embassy-mspm0/src/trng.rs:13:5
   |
13 | use rand_core::{TryCryptoRng, TryRngCore};
   |     ^^^^^^^^^ use of unresolved module or unlinked crate `rand_core`
   |
   = help: if you wanted to use a crate named `rand_core`, use `cargo add rand_core` to add it to your `Cargo.toml`
```

~I think this has been the case since #5172 ?~ No - it's because a bad rebase on my end dropped rand_core from Cargo.toml. Fixed now.

examples/mspm0g3107/README.md
```diff
@@ -0,0 +1,25 @@
...
+cargo run --bin <module-name>
+```
+for example
+```
+cargo run --bin can
+```
+
+## Checklist before running examples
+
+The MSPM0G3107 does not have a Launchpad or official development board, so you likely need to modify examples to update pin numbers or peripherals to match the specific MCU or board you are using.
```
i509VCB: Might be worth mentioning G3507 will be quite similar to this.

i509VCB:
> CI fails due to:
> 
> ```
> error[E0432]: unresolved import `rand_core`
>   --> /ci/code/embassy-mspm0/src/trng.rs:13:5
>    |
> 13 | use rand_core::{TryCryptoRng, TryRngCore};
>    |     ^^^^^^^^^ use of unresolved module or unlinked crate `rand_core`
>    |
>    = help: if you wanted to use a crate named `rand_core`, use `cargo add rand_core` to add it to your `Cargo.toml`
> ```
> 
> I think this has been the case since #5172 ?

Weird, the pull request clearly adds that to Cargo.toml?

magmastonealex:
Ugh, yeah. I accidentally deleted `rand_core = "0.9"` from Cargo.toml and didn't notice. Let me add that back in.

examples/mspm0g3107/README.md
magmastonealex: I've updated this line to mention that the examples should also run there with minimal changes.

I decided to create a separate examples folder for this specific MCU since I don't have access to others to test against.

embassy-mspm0/build.rs
magmastonealex: Is this approach reasonable for now, and just plan to improve it down the line?

i509VCB: For now this is fine.

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+
+    /// CAN timings to use for standard CAN. (CAN-FD support to come later.)
+    pub timing: CanTimings,
+
+    /// If remote frames should be accepted into the RX queue.
+    pub accept_remote_frames: bool,
+
+    /// If extended ID frames should be accepted into the RX queue.
+    pub accept_extended_ids: bool,
+}
```
i509VCB, line 127: I have a feeling that the code could calculate some of the timings for you to get a specific bitrate and back calculate from clock config. See the UART driver for an idea of this.

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+            timing: CanTimings::from_values(20, 2, 13, 2).unwrap(),
+        }
+    }
+}
+
+#[non_exhaustive]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Config Error
+pub enum InitializationError {
```
i509VCB: Use `ConfigError` to match other drivers.

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+            if let Some(frame) = self.get_frame() {
+                return Ok(frame);
+            }
+
+            // BusOff will never recover by itself in the current implementation - bail so the caller can deal with this.
+            if self.get_error_counters().bus_off {
+                return Err(BusError::BusOff);
+            }
+
+            cortex_m::asm::delay(100);
```
i509VCB: Required due to timing bugs in hardware?

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+
+// Implementations for the embedded_can traits.
+// I am really not sure that these are good abstractions for real applications.
+// They do not support "confirmable" mesage sending (i.e the docs specifically state that
+// transmit() only enqueues frames, there is no feedback mechanisim to confirm if/when a frame
+// was actually sent)
+// They also do not support monitoring the peripheral's status or recovering from bus-off.
+// I will implement them regardless to play nice with the overall ecosystem,
+// but similar to embassy-stm32, I'm going to offer a HAL-specific API which provides
+// more useful functionality, and add an async variant at some point.
```
i509VCB, line 364: I feel this is a question that could be an issue on the `embedded_can` crate. I am not 100% familar with CAN, but would confirmation be something that hardware reports or is that part of a "CAN stack".

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+use core::marker::PhantomData;
+
+use embassy_hal_internal::PeripheralType;
+
+use crate::Peri;
+use crate::can::frame::MCanFrame;
+use crate::can::msgram::{McanMessageRAM, MessageRAMAccess};
+use crate::gpio::{AnyPin, PfType};
+use crate::mode::{Blocking, Mode};
+use crate::pac::canfd::{Canfd as Regs, vals as CanVals};
```
i509VCB: Usually we don't rename path elements. Leaving it as `vals` is fine.

embassy-mspm0/src/can/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+        }
+    }
+
+    fn new_inner<T: Instance>(
+        peri: Peri<'d, T>,
+        rx: Peri<'d, impl RxPin<T>>,
+        tx: Peri<'d, impl TxPin<T>>,
+        config: Config,
+    ) -> Result<Self, InitializationError> {
+        // Note: use new_pin! when in tree.
```
i509VCB: Well we are in tree now

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(C)]
+pub(super) struct ExtendedFilter {
+    // a bit awkward to split this into two structs, but this is hidden from downstream consumers anyways.
+    pub ef1: ExtendedFilter1,
+    pub ef2: ExtendedFilter2,
+}
```
i509VCB, line 258: Does this need to be packed?

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(C)]
+pub(super) struct RxBufferElement {
+    // a bit awkward to split this into two structs, but this is hidden from downstream consumers anyways.
+    pub hdr: MsgHeader,
+    pub rxhdr: RxHeader,
+    pub data: [u8; MAX_DATA_LEN],
+}
```
i509VCB, line 281: Another question around packed and C repr.

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(C)]
+pub(super) struct TxBufferElement {
+    // a bit awkward to split this into two structs, but this is hidden from downstream consumers anyways.
+    pub hdr: MsgHeader,
+    pub txhdr: TxHeader,
+    pub data: [u8; MAX_DATA_LEN],
+}
```
i509VCB, line 301: Packed question again

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(C)]
+pub(super) struct TxEventElement {
+    // a bit awkward to split this into two structs, but this is hidden from downstream consumers anyways.
+    pub hdr: MsgHeader,
+    pub event: EventHeader,
+}
```
i509VCB, line 320: Another packed struct question

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+#[repr(C)]
+pub(super) struct McanMessageRAM {
+    filters: [StandardFilter; 0],
+    extended_filters: [ExtendedFilter; 0],
+    rxfifo0: [RxBufferElement; NUM_RX_ELEMENTS],
+    rxfifo1: [RxBufferElement; 0], // Note: while we're not using most of these features yet, I've included their offsets and structures anyways to save the next person some pain.
+    rxbuffers: [RxBufferElement; 0],
+    txevents: [TxEventElement; NUM_TX_EVENTS],
+    txfifo0: [TxBufferElement; NUM_TX_ELEMENTS],
+}
```
i509VCB, line 343: Another packed struct question

embassy-mspm0/src/can/msgram.rs
```diff
@@ -0,0 +1,534 @@
...
+            rxfifo0: core::mem::offset_of!(McanMessageRAM, rxfifo0) / 4,
+            rxfifo1: core::mem::offset_of!(McanMessageRAM, rxfifo1) / 4,
+            rxbuffers: core::mem::offset_of!(McanMessageRAM, rxbuffers) / 4,
+            txevents: core::mem::offset_of!(McanMessageRAM, txevents) / 4,
+            txfifo: core::mem::offset_of!(McanMessageRAM, txfifo0) / 4,
+        }
+    };
+}
+
+macro_rules! impl_ram_access {
```
i509VCB, line 470: I wonder if https://github.com/google/zerocopy could do a lot of this macro magic for you? Not going to demand a move to zerocopy at the moment.

embassy-mspm0/src/can/mod.rs
magmastonealex: It certainly can! I have listed it as an excluded feature for now to get something minimal working. 

The algorithm is a bit complex, which is why I left it out, but there are other implementations out there (even in embassy-stm32).

I can include it if you want - but my original intention was to follow up and add it in later since it’s another ~250 lines of code.

embassy-mspm0/src/can/mod.rs
magmastonealex: It depends on the hardware as far as I know - the MCAN IP does support hardware reports of confirmed successful transmission, but I don’t think all IPs do. Some higher level protocols might include protocol-layer ACKs as well, since the CAN ACK just means _some_ device received it correctly, not that the destination received it.

I don’t neccessarily see it as an issue within embedded_can - confirmable messages and asynchronous support are important to the specific applications that I have built, but general hardware support may not be there to expect every MCU to support them.

It’s more that our peripheral _can_ do more, so I want to expose methods to make more use of what is there.

embassy-mspm0/src/can/msgram.rs
magmastonealex: Yes, it should have been - will update. Got lucky that each element in these structs is exactly 32 bits, so C padding rules resulted in no additional padding between elements.

embassy-mspm0/src/can/msgram.rs
magmastonealex: It might be able to implement the same logic, but I don’t see support within `zerocopy` for volatile reads and writes, which should be used here IMO because the peripheral is also modifying or reading from RAM.

embassy-mspm0/src/can/msgram.rs
magmastonealex: Oh - but if I make this packed, then the compiler forces me to use raw pointers and `read_unaligned` everywhere - despite the alignment actually being the same as it as before since it’s two 32-bit values on a 32-bit platform.

I wonder if it would be easier to instead just assert on sizes and offsets somewhere.

embassy-mspm0/src/can/mod.rs
magmastonealex: Removed. Didn’t mean to leave this in. Thanks for catching it.

embassy-mspm0/src/can/mod.rs
magmastonealex: Ooops - leftover from out of tree. Updated.

embassy-mspm0/src/can/mod.rs
magmastonealex: Indeed 🙂 Removed.

embassy-mspm0/src/can/msgram.rs
magmastonealex: I’ve added a commit which asserts on this at compile time via the const SIZE.

It’s a bit hacky, but I also don’t love using the packed repr and needing to use raw pointers everywhere.

embassy-mspm0/src/can/msgram.rs
magmastonealex: This one actually _doesn’t_ need to be packed and alignment doesn’t matter because we tell the peripheral the offsets within msgram for each array.

embassy-mspm0/src/can/mod.rs
magmastonealex: Good point, updated.

embassy-mspm0/src/lib.rs
```diff
@@ -173,6 +176,134 @@ impl Default for Config {
...
+            cortex_m::asm::delay(16);
+        }
+    }
+
+    // We will set the pre-divider to 1, and the VCO divider to 4, causing:
+    // Reference clk: 32MHz.
+    // Divided by pre-divider: 32MHz
+    // Multiplied by VCO divider: 128MHz
+    // Divided by clk1div: 32MHz
+    // Produces a CANFD functional clock of 32MHz (matching MCLK sourced from SYSOSC.)
```
bespsm, line 226: Page 5 here: https://www.ti.com/lit/an/slaaet4/slaaet4.pdf?ts=1771398227621

```The recommendation frequency of the CAN_CLK is 20MHz, 40MHz and 80MHz.```

can we adapt to it?

Comment (i509VCB) [reply embassy-mspm0/src/lib.rs:226]: Until we have proper clock config we kind of just hard code some mode to work in.

File: embassy-mspm0/src/can/msgram.rs (hunk 12 above)
Comment (magmastonealex) [reply embassy-mspm0/src/can/msgram.rs:281]: Discussion above - packed makes accessing elements very frustrating even though they are actually aligned. I’ve worked around this with compile-time assertions via the const SIZE to ensure elements are the right size and right offset.

**Comment by i509VCB** [comment]
Need to review a 2nd time, bug me if I forget

File: examples/mspm0g3107/Cargo.toml [hunk 18]
```diff
@@ -0,0 +1,28 @@
+[package]
+edition = "2024"
+name = "embassy-mspm0-g3519-examples"
```
bespsm: is `g3519` correct here?

i509VCB: Probably not.

bespsm:
It partially works on `g3507` hardware. Meaning, CAN analyzer receives the frames (0x12 0x34 0x56) but MCU doesn't see incoming frames from the analyzer side. Any guess @magmastonealex @i509VCB  ?

i509VCB:
> It partially works on `g3507` hardware. Meaning, CAN analyzer receives the frames (0x12 0x34 0x56) but MCU doesn't see incoming frames from the analyzer side. Any guess @magmastonealex @i509VCB ?

I will try to test .

examples/mspm0g3107/Cargo.toml
magmastonealex: Definitely an oversight. Pushed a fix with the correct part number now.

magmastonealex:
> It partially works on `g3507` hardware. Meaning, CAN analyzer receives the frames (0x12 0x34 0x56) but MCU doesn't see incoming frames from the analyzer side. Any guess @magmastonealex @i509VCB ?

Can you clarify if you are seeing any error frames on the bus, or if the frame is being acknowledged by the MCU but then not displayed in the log output?

What are the periodic "CAN error counters:" messages showing while this is occurring?

If you're receiving frames on a CAN analyzer, then it's _probably_ not a timing issue, but in testing I've found the internal RC oscillator is _quite_ inaccurate and doesn't work super reliably for CAN across temperature - on my own projects I'm using an external oscillator which is not yet supported in main.

bespsm:
> > It partially works on `g3507` hardware. Meaning, CAN analyzer receives the frames (0x12 0x34 0x56) but MCU doesn't see incoming frames from the analyzer side. Any guess @magmastonealex @i509VCB ?
> 
> Can you clarify if you are seeing any error frames on the bus, or if the frame is being acknowledged by the MCU but then not displayed in the log output?
> 
> What are the periodic "CAN error counters:" messages showing while this is occurring?
> 
> If you're receiving frames on a CAN analyzer, then it's _probably_ not a timing issue, but in testing I've found the internal RC oscillator is _quite_ inaccurate and doesn't work super reliably for CAN across temperature - on my own projects I'm using an external oscillator which is not yet supported in main.

unfortunately I didn't check CAN error counters. Hopefully I'll find time to test M0 with launchpad and different CAN transceiver, using RC and external oscillator.

added on 21.04: I tested CAN interface with another CAN transceiver, and in both cases (with external and internal oscillator) firmware works fine. I tried also TI examples with 500 kbit/sec, works fine

PR #5297, opened 2026-01-25
embassy-stm32-wpan BLE split

arthurbabu:
Hi ! 

I want to use BLE on an STM32WB55. But i faced an issue with current implementation, `ble.read()` function must be called continuously to catch BLE events or responses to commands sent. But i want to send BLE notifications asynchronously to BLE events.

My first try was to use an `embassy::select` on `ble.read()` and a channel (with data to send into). But, this leads to hardfault when `ble.read()` future is dropped at the bad time (HCI commands corruption)

So, in this pull request i implemented a split function for BLE. It's not perfect because it's still user's responsibility to ensure `ble.read()` function is called before send any other command. I think it can be implemented at the driver level (await a read before sending other command ?)

### DIFF
--- a/embassy-stm32-wpan/src/wb55/sub/ble.rs
+++ b/embassy-stm32-wpan/src/wb55/sub/ble.rs
@@ -89,7 +89,7 @@ impl<'a> Ble<'a> {
             },
             BleRx {
                 ipcc_ble_event_channel: self.ipcc_ble_event_channel,
-            }
+            },
         )
     }
 
@@ -162,7 +162,6 @@ impl<'a> hci::Controller for Ble<'a> {
     }
 }
 
-
 impl<'a> BleTx<'a> {
     /// `TL_BLE_SendCmd`
     pub async fn tl_write(&mut self, opcode: u16, payload: &[u8]) {
@@ -227,7 +226,7 @@ impl<'a> hci::Controller for BleTx<'a> {
     }
 
     async fn controller_read_into(&mut self, _buf: &mut [u8]) {
-        panic!("BleTx cannot read!"); 
+        panic!("BleTx cannot read!");
     }
 }
 


xoviat:
> But, this leads to hardfault when ble.read() future is dropped at the bad time (HCI commands corruption)

Does this still occur with #5299?

arthurbabu:
Yes, i pull latest embassy/main branch and i still have : 
```
6.428619 [INFO ] [BLE][TX] Completed, packet size 243 (com_card src/system/ble_nus.rs:254)
Firmware exited unexpectedly: Multiple
Core 0
    Frame 0: HardFault_ @ 0x0801e9b0
      cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cortex-m-rt-0.7.5/src/lib.rs:1103:1
    Frame 1: HardFault <Cause: Escalated BusFault <Cause: Imprecise data access error>> @ 0x08007da1

    Frame 2: write_volatile<embassy_stm32_wpan::wb55::unsafe_linked_list::LinkedListNode> @ 0x08007da1 inline
      .rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:2205:9
    Frame 3: remove_node @ 0x0000000008007d9a inline
    comcard/dependencies/embassy/embassy-stm32-wpan/src/wb55/unsafe_linked_list.rs:0
    Frame 4: remove_head @ 0x0000000008007d82
     comcard/dependencies/embassy/embassy-stm32-wpan/src/wb55/unsafe_linked_list.rs:157:13
    Frame 5: embassy_stm32_wpan::wb55::tables::TL_SYS_TABLE @ 0x2003003e> @ 0x000000002003003e

Error: Multiple
```

PR #5301, opened 2026-01-26
microchip: Add UART driver

kurtjd:
Adds a UART driver for the microchip HAL. This does not currently support MODEM/flow control signals as I do not have a means of testing that, but can be added in a future PR.

Note: Pin de-configuration on drop needs to be revisited, but that will be handled in a future PR which will handle this for all current drivers.

### DIFF
--- a/embassy-microchip/src/uart.rs
+++ b/embassy-microchip/src/uart.rs
@@ -1,0 +1,1089 @@
+//! UART driver.
+
+use core::future::poll_fn;
+use core::marker::PhantomData;
+use core::sync::atomic::{AtomicU8, Ordering};
+use core::task::Poll;
+
+use embassy_hal_internal::{Peri, PeripheralType};
+use embassy_sync::waitqueue::AtomicWaker;
+use mec17xx_pac::uart0::regs::DataLsr;
+
+use crate::gpio::{AnyPin, Pin, SealedPin};
+use crate::interrupt::typelevel::{Binding, Interrupt};
+use crate::pac::uart0::Uart0;
+use crate::{interrupt, pac, peripherals};
+
+// The datasheet does not appear to specify this, but the 16550 has a 16-byte FIFO
+// and empirical testing confirms this is also the case for the MEC UART.
+//
+// This is useful to know since UART does not support DMA, allowing us to interrupt less frequently.
+const FIFO_SZ: usize = 16;
+
+// The HW gives us a convenient scratch register which we use to hold interrupt flags
+mod int_flag {
+    pub(crate) const RX_AVAILABLE: u8 = 1 << 0;
+    pub(crate) const TIMEOUT: u8 = 1 << 1;
+    pub(crate) const TX_EMPTY: u8 = 1 << 2;
+}
+
+/// UART interrupt handler.
+pub struct InterruptHandler<T: Instance> {
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandler<T> {
+    unsafe fn on_interrupt() {
+        let intid = InterruptType::try_from(T::info().reg.data().int_id().read().intid());
+        match intid {
+            // In case of RxAvailable, We need a way to know that the FIFO level is triggered in task,
+            // but also need to disable the interrupt to clear it (which clears it in intid).
+            //
+            // In case of LineStatus, We don't use a separate flag from RxAvailable because
+            // the RX task will keep reading bytes until it hits the byte that produced the error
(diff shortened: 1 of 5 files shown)


embassy-microchip/src/uart.rs
```diff
@@ -0,0 +1,1084 @@
...
+    CharacterTimeout,
+    TxEmpty,
+    // MODEM is not currently supported
+    _Modem,
+    Unknown,
+}
+
+impl InterruptType {
+    // Note: This always assumes we are in FIFO mode (the driver does not support non-FIFO mode)
+    fn from_bits(value: u8) -> Self {
```
felipebalbi: `impl TryFrom<u8> for InterruptType`?

embassy-microchip/src/uart.rs
```diff
@@ -0,0 +1,1084 @@
...
+
+enum RxFifoTrigger {
+    _1,
+    _4,
+    _8,
+    _14,
+}
+
+impl RxFifoTrigger {
+    fn as_bits(&self) -> u8 {
```
felipebalbi: `impl From<RxFifoTrigger> for u8`?

embassy-microchip/src/uart.rs
```diff
@@ -0,0 +1,1084 @@
...
+}
+
+/// UART configuration.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config {
+    /// The clock source which the baud clock is derived from.
+    ///
+    /// This will affect which baud rates can be accurately represented.
+    pub clk_src: ClkSrc,
```
felipebalbi, line 228: Do you really want to expose the clock source directly? Can you extrapolate clock source from the requested baud rate instead?

embassy-microchip/src/uart.rs
```diff
@@ -0,0 +1,1084 @@
...
+    /// Note: If multi stop bits is enabled, the number of stop bits corresponds to chosen word length.
+    pub word_len: WordLen,
+    /// Enable multiple stop bits.
+    ///
+    /// If enabled, the number of stop bits will be dependent on chosen word length.
+    ///
+    /// If disabled, the number of stop bits will always be 1.
+    pub use_multi_stop: bool,
+    /// Parity selection.
+    pub parity: Option<Parity>,
```
felipebalbi: This should be a mandatory argument, no? It just so happens that `Parity::None` is the default. I think using an `Option` is unnecessary.

embassy-microchip/src/uart.rs
kurtjd: Originally had this, but in my mind it feels "wrong" in the sense that these aren't u8 representations. For example RxFifoTrigger, in my mind `From<RxFifoTrigger> for u8` would produce a `14`, `8`, etc and not the bit representation in the register. Though since these are private probably no real risk of confusion.

embassy-microchip/src/uart.rs
kurtjd: Same as above.

embassy-microchip/src/uart.rs
kurtjd: Hm I had thought about this. I think I can check if the provided baud rate results in the 16th bit of the divisor being set, which should map to a a clock source. I'll experiment.

embassy-microchip/src/uart.rs
kurtjd: Personally I think an `Option<T>` makes more sense than a `T::None` variant, but I can make the change. Saw this pattern used in the Nuvoton UART HAL and thought it made sense but don't feel strongly.

kurtjd: Done!

embassy-microchip/src/uart.rs
kurtjd: Done.

embassy-microchip/src/uart.rs
kurtjd: Done.

embassy-microchip/src/uart.rs
kurtjd: After thinking about it some more, I think it makes sense still to expose the clock source in this case. If only one of two internal clock sources were available, we could determine which to choose based on the provided baud rate, but the interface allows for an external clock:

image
image

We could just not support external clock, but I think it's worth exposing clock source as a config option for a bit more flexibility.

kurtjd: Though this was just my reasoning, if typically HALs prefer not to expose this kinda detail, don't mind changing it.

felipebalbi: Sounds good to me. Let's get this queued up.

Issue #4662, opened 2025-09-14
Add WIZnet W6300 net driver


KingCol13:
The WIZnet [W6300](https://docs.wiznet.io/Product/iEthernet/W6300/overview) has a quad SPI interface, which is interesting since afaict the network bandwidth bottleneck is usually SPI. Their graph indicates them getting >80Mbps at 37.5MHz QSPI clock speed with iperf.

An example could also be added for the [W6300-EVB-Pico2](https://docs.wiznet.io/Product/iEthernet/W6300/w6300-evb-pico2) or [W6300-EVB-Pico](https://docs.wiznet.io/Product/iEthernet/W6300/w6300-evb-pico).

PR #5305, opened 2026-01-26
embassy-net-wiznet: add W6300 driver

lhalf:
Related to #4662 

Tested with this device:
https://wiznet.io/products/evaluation-boards/w6300-evb-pico2
Here:
https://github.com/lhalf/w6300-evb-pico2-json

The registers are similar to the W6100 driver, so there isn't significant change.

The main differences are the instruction and dummy phase in bus_read and bus_write, also the SpiDevice must be a pio_program::Spi device. I will add a new example for reference.

As the W6100 driver also does - I've left the MAC filter disabled but this does not affect the operation of this device (although it will cause more noise on the SPI bus).

For the above reason, and the fact that this is only supporting the Single SPI mode I would consider this driver experimental. To move towards QSPI I think we would need an embassy_rp::qspi implementation.

### DIFF
--- a/embassy-net-wiznet/README.md
+++ b/embassy-net-wiznet/README.md
@@ -8,6 +8,8 @@ See [`examples`](https://github.com/embassy-rs/embassy/tree/main/examples/rp) di
 
 - W5500
 - W5100S
+- W6100
+- W6300 (Single SPI only)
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
very cool! thank you

PR #5309, opened 2026-01-26
wpan: use pac to disable/enable clocks

xoviat:


xoviat:
@leftger can you approve this?

PR #5310, opened 2026-01-26
stm32f1 i2c loopback example

ChocolateLoverRaj:
This example demonstrates using async i2c master and slave.


xoviat:
can you run cargo +nightly fmt?

leftger:
@ChocolateLoverRaj Can you run fmtall.sh and make sure it passes CI? Once that happens we can get this merged :)

PR #5311, opened 2026-01-26
[MCXA] Add i2c target driver

felipebalbi:
New driver added. While doing that, also removed `Result` type alias as per @jamesmunns' suggestion.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/64
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/31
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/79
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/80

### DIFF
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -104,7 +104,6 @@ impl Default for Address {
 
 /// Enable or disable feature
 #[derive(Clone, Default)]
-#[non_exhaustive]
 pub enum Status {
     #[default]
     Disabled,


embassy-mcxa/src/i2c/mod.rs
```diff
@@ -11,19 +11,33 @@ use crate::clocks::{ClockError, Gate};
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ConfigError {
+    /// Invalid Address
+    InvalidAddress,
+}
 
 /// Error information type
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum Error {
```
jamesmunns: We might want to make this (and all other Error types in general) `#[non_exhaustive]` to avoid breaking changes when folks start using this.

jamesmunns: Also, if there are differing error types for the different I2C variants (controller, target) x (blocking, async, dma) and *maybe* x (setup, i/o), we should probably start breaking them up. This means users won't need to match on "impossible" error variants.

embassy-mcxa/src/i2c/mod.rs
```diff
@@ -49,7 +63,7 @@ pub struct InterruptHandler<T: Instance> {
 
 impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandler<T> {
     unsafe fn on_interrupt() {
-        if T::info().regs().mier().read().bits() != 0 {
+        if T::info().regs().mier().read().bits() != 0 || T::info().regs().sier().read().bits() != 0 {
```
jamesmunns: Would it make sense two have two different error handlers, one for controller, and one for target?

This might help with the split error types, and save a little effort in the error handler to not check errors that don't match.

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                        w.addrcfg().variant(if (0x00..=0x7f).contains(&addr) {
+                            Addrcfg::AddressMatch0_7Bit
+                        } else {
+                            Addrcfg::AddressMatch0_10Bit
+                        })
+                    });
+                }
+
+                Address::Dual(addr0, addr1) => {
+                    if ((0x00..=0x7f).contains(&addr0) ^ (0x00..=0x7f).contains(&addr1))
```
jamesmunns, line 293: Might want to add a comment to explain the use of XOR here, I'm *guessing* there is a requirement that addresses are BOTH 7-bit or 10-bit? Either way, it's not immediately clear what this is protecting against.

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                    .clear_bit_by_one()
+                    .fef()
+                    .clear_bit_by_one()
+            });
+
+            Ok(())
+        })
+    }
+
+    // REVISIT: allow selection of various clock sources
```
jamesmunns: Can we revisit this now?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+        (
+            PoweredClock::NormalEnabledDeepSleepDisabled,
+            Lpi2cClockSel::FroLfDiv,
+            const { Div4::no_div() },
+        )
+    }
+
+    /// Resets both TX and RX FIFOs dropping their contents.
+    fn reset_fifos(&self) {
+        critical_section::with(|_| {
```
jamesmunns, line 357: Maybe comment why a CS is required here but not other checks, like `clear_status`?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    // Public API: Blocking
+
+    /// Block waiting for new events
+    pub fn blocking_listen(&mut self) -> Result<Request, Error> {
+        self.clear_status();
+
+        // Wait for Address Valid
+        while self.info.regs().ssr().read().avf().bit_is_clear()
```
jamesmunns: Maybe turn this into a loop to avoid three reads of the SSR register?

```rust
loop {
    let ssr = self.info.regs().ssr().read();
    let avf = ssr.avf().bit_is_set();
    let sarf = ssr.sarf().bit_is_set();
    let gcf = ssr.gcf().bit_is_set();
    
    if avf || sarf || gcf {
        break;
    }
}
```

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    /// Respond to read request
+    pub fn blocking_respond_to_read(&mut self, buf: &[u8]) -> Result<usize, Error> {
+        let mut count = 0;
+
+        self.clear_status();
+
+        for byte in buf.iter() {
+            // Wait until we can send data
+            while self.info.regs().ssr().read().tdf().bit_is_clear()
```
jamesmunns: Same note re: triple read, we can also return it here to not need two more reads on line 364, e.g.

```rust
let ssr = loop {
    let ssr = self.info.regs().ssr().read();
    //
    if tdf || sdf || rsf {
        break ssr;
    }
};
```

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                    Ok(Request::Read(addr >> 1))
+                } else {
+                    Ok(Request::Write(addr >> 1))
+                }
+            }
+            _ => Err(Error::Other),
+        }
+    }
+
+    /// Respond to read request
```
jamesmunns: Maybe comment what the returned value here means? Number of bytes sent?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    /// Respond to write request
+    pub fn blocking_respond_to_write(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        let mut count = 0;
+
+        self.clear_status();
+
+        for byte in buf.iter_mut() {
+            // Wait until we have data to read
+            while self.info.regs().ssr().read().rdf().bit_is_clear()
```
jamesmunns: Same note re: repeated read

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+            } else {
+                self.info.regs().stdr().write(|w| unsafe { w.data().bits(*byte) });
+                count += 1;
+            }
+        }
+
+        Ok(count)
+    }
+
+    /// Respond to write request
```
jamesmunns: Same note about return value

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    /// Asynchronously wait for new events
+    pub async fn async_listen(&mut self) -> Result<Request, Error> {
+        self.clear_status();
+
+        self.info
+            .wait_cell()
+            .wait_for(|| {
+                self.enable_ints();
+                self.info.regs().ssr().read().avf().bit_is_set()
```
jamesmunns, line 592: More duped reads

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                .rdie()
+                .enabled()
+                .tdie()
+                .enabled()
+        });
+    }
+
+    // Public API: Async
+
+    /// Asynchronously wait for new events
```
jamesmunns, line 584: Discuss cancellation safety, do we need to disable anything if cancelled?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                    Ok(Request::Read(addr >> 1))
+                } else {
+                    Ok(Request::Write(addr >> 1))
+                }
+            }
+            _ => Err(Error::Other),
+        }
+    }
+
+    /// Respond to read request
```
jamesmunns: Discuss cancellation, return value meaning

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+
+        self.clear_status();
+
+        for byte in buf.iter() {
+            // Wait until we can send data
+            self.info
+                .wait_cell()
+                .wait_for(|| {
+                    self.enable_ints();
+                    self.info.regs().ssr().read().tdf().bit_is_set()
```
jamesmunns: Repeated reads

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+            } else {
+                self.info.regs().stdr().write(|w| unsafe { w.data().bits(*byte) });
+                count += 1;
+            }
+        }
+
+        Ok(count)
+    }
+
+    /// Respond to write request
```
jamesmunns: Discuss cancellation, return value

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+                *byte = self.info.regs().srdr().read().data().bits();
+                count += 1;
+            }
+        }
+
+        Ok(count)
+    }
+}
+
+impl<'d, M: Mode> Drop for I2c<'d, M> {
```
jamesmunns, line 693: Add commment on I2c that pins are disabled on drop

examples/mcxa/src/bin/i2c-target-async.rs
```diff
@@ -0,0 +1,56 @@
...
+async fn main(_spawner: Spawner) {
+    let mut config = Config::default();
+    config.clock_cfg.sirc.fro_lf_div = Div8::from_divisor(1);
+
+    let p = hal::init(config);
+
+    defmt::info!("I2C target example");
+
+    let mut config = target::Config::default();
+    config.address = target::Address::Single(42);
```
jamesmunns: Might want to use hex here for readability

examples/mcxa/src/bin/i2c-target-async.rs
```diff
@@ -0,0 +1,56 @@
...
+            Ok(request) => {
+                defmt::info!("Received event {}", request);
+                match request {
+                    target::Request::Read(_addr) => {
+                        buf.fill(0x55);
+                        let count = target.async_respond_to_read(&buf).await.unwrap();
+                        defmt::info!("T [R]: {:02x} -> {:02x}", _addr, buf[..count]);
+                    }
+                    target::Request::Write(_addr) => {
+                        buf.fill(0);
```
jamesmunns: Is it necessary to fill if this will be overwritten?

examples/mcxa/src/bin/i2c-target-blocking.rs
```diff
@@ -0,0 +1,49 @@
...
+    let mut config = Config::default();
+    config.clock_cfg.sirc.fro_lf_div = Div8::from_divisor(1);
+
+    let p = hal::init(config);
+
+    defmt::info!("I2C target example");
+
+    let mut config = target::Config::default();
+    config.address = target::Address::Single(42);
+    // config.address = target::Address::Dual(42, 43);
```
jamesmunns: Dead code, if you want to keep it for reference, add a note to the user that they can pick any of these.

examples/mcxa/src/bin/i2c-target-blocking.rs
```diff
@@ -0,0 +1,49 @@
...
+    // config.address = target::Address::Range(40..50);
+    let mut target = target::I2c::new_blocking(p.LPI2C3, p.P3_27, p.P3_28, config).unwrap();
+    let mut buf = [0u8; 32];
+
+    loop {
+        match target.blocking_listen() {
+            Err(e) => {
+                defmt::error!("Failed to listen for events: {}", e);
```
jamesmunns: Do we want to panic to prevent busy looping if there is some fatal error?

examples/mcxa/src/bin/i2c-target-async.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    let mut config = target::Config::default();
+    config.address = target::Address::Single(42);
+    let mut target = target::I2c::new_async(p.LPI2C3, p.P3_27, p.P3_28, Irqs, config).unwrap();
+    let mut buf = [0u8; 32];
+
+    loop {
+        match target.async_listen().await {
+            Err(e) => {
+                defmt::error!("Failed to listen for events: {}", e);
```
jamesmunns: panic/break on error?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -0,0 +1,574 @@
...
+
+use super::{Async, Blocking, ConfigError, Error, Info, Instance, InterruptHandler, Mode, SclPin, SdaPin};
+use crate::clocks::periph_helpers::{Div4, Lpi2cClockSel, Lpi2cConfig};
+use crate::clocks::{PoweredClock, WakeGuard, enable_and_reset};
+use crate::gpio::{AnyPin, SealedPin};
+use crate::interrupt::typelevel::Interrupt;
+
+/// I2C target addresses.
+#[derive(Clone)]
+#[non_exhaustive]
```
jamesmunns: For some of these, if there WILL NEVER be another variant, we can remove non_exhaustive. For items that retain non_exhaustive, it might be good to say *why* we are reserving the right to extend this in the future.

embassy-mcxa/src/i2c/mod.rs
felipebalbi: done both

embassy-mcxa/src/i2c/mod.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: I think the easy way would be to expose the valid clock sources, let the user select them, then we compute the timings based on that. I'll add a commit implementing it for reference.

embassy-mcxa/src/i2c/target.rs
felipebalbi: clear status is atomic, a single status. Resetting the fifos involved read-modify-write. We don't want accidental preemption mucking things up. Comment added

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: There's nothing to do on cancellation.

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: nothing for cancellation, comment updated for return value.

embassy-mcxa/src/i2c/target.rs
felipebalbi: that's already discussed in the constructor.

examples/mcxa/src/bin/i2c-target-async.rs
felipebalbi: done

examples/mcxa/src/bin/i2c-target-async.rs
felipebalbi: done

examples/mcxa/src/bin/i2c-target-blocking.rs
felipebalbi: Added a comment

examples/mcxa/src/bin/i2c-target-blocking.rs
felipebalbi: done

examples/mcxa/src/bin/i2c-target-async.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: Do we need to be explicit here? "Hey, we may add new errors in the future, that's why it's `non_exaustive`" sounds previous obvious, no?

embassy-mcxa/src/i2c/target.rs
felipebalbi: done

embassy-mcxa/src/i2c/target.rs
felipebalbi: Updated pushed, let me know what you think.

jamesmunns (approved):
Looks good to me, I sent you a DM about potentially how to split up the Error types, but feel free to merge with or without that @felipebalbi

felipebalbi:
> Looks good to me, I sent you a DM about potentially how to split up the Error types, but feel free to merge with or without that @felipebalbi

cool, I'll integrate and start working on the error split immediately.

PR #5312, opened 2026-01-26
add new async digital traits for BlockingAsync

ChocolateLoverRaj:
currently these new traits are only in the git repo of embedded-hal and not published, so this requires using the creates from git

for my project I need these traits so I'm using the git version of embedded-hal and  I need an adapter. Once the new traits are in a published version I can remove the changes to `Cargo.toml` and make this PR no longer a draft. If this is approved I can also implement yielding versions of this.


Dirbaio:

Issue #4597, opened 2025-08-28
embassy-stm32: Don't bind DMA interrupt handlers by default


wojciech-graj:
Interrupt handlers for DMA are automatically bound by the build script. This means that a user cannot supply their own interrupt handler, as there are now two symbols defined for this interrupt.

The error below is produced during building.
```
warning: Linking globals named 'DMA1_CHANNEL4': symbol multiply defined!
```

`build.rs` currently does this without a way to toggle it off.
https://github.com/embassy-rs/embassy/blob/1f945bcebd3b0018b3b5541ede767583df3545d8/embassy-stm32/build.rs#L1888C1-L1896C14
```rust
            quote! {
                #[cfg(feature = "rt")]
                #[crate::interrupt]
                unsafe fn #irq () {
                    #(
                        <crate::peripherals::#channels as crate::dma::ChannelInterrupt>::on_irq();
                    )*
                }
            }
```

Ideally the user should be required to use `bind_interrupts!` to bind these themselves.

everdrone:
Also experiencing this

nullcrimson:
+1 I also have the same issue, would be nice to have the ability to explicitely define DMA interrupts with `bind_interrupts!`

PR #5315, opened 2026-01-27
Allow users to define custom DMA interrupt handler functions

everdrone:
Addresses #4597

- Allows users to define custom DMA interrupt handlers.
- Users now need to manually bind DMA interrupt handlers using `bind_interrupts!`.
- Conforms with the rest of `embassy-stm32` irq semantics.

### DIFF
--- a/embassy-stm32/src/dma/mod.rs
+++ b/embassy-stm32/src/dma/mod.rs
@@ -72,36 +72,32 @@ pub(crate) trait SealedChannel: StoppablePeripheral {
 /// This trait provides the interrupt type and handler for a DMA channel.
 /// It is automatically implemented for all DMA channel peripherals.
 #[allow(private_bounds)]
-pub trait ChannelInterrupt {
+pub trait ChannelInterrupt: Into<AnyChannel> {
     /// The interrupt type for this DMA channel.
     type Interrupt: interrupt::typelevel::Interrupt;
 
     #[doc(hidden)]
     #[cfg_attr(not(feature = "rt"), allow(unused))]
     unsafe fn on_irq();
+
+    /// Degrade this channel to a type-erased [`AnyChannel`], verifying interrupt binding.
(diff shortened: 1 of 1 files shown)


Dirbaio:
What happens if the user forgets to do the `bind_interrupts!`? Does it compile but hang at runtime?

It should fail to compile, same as when you forget the `bind_interrupts` for an interrupt for UART or SPI or whatever.

embassy-stm32/src/dma/mod.rs
```diff
@@ -76,8 +85,45 @@ pub(crate) trait ChannelInterrupt {
...
+    fn degrade(self, _irq: impl interrupt::typelevel::Binding<Self::Interrupt, InterruptHandler<Self>>) -> AnyChannel {
+        self.into()
+    }
+}
+
+/// Asserts at compile time that a DMA channel's interrupt is bound.
+///
+/// This function is used by the `new_dma!` macro to verify interrupt bindings.
+#[inline(always)]
+pub fn assert_dma_binding<T: ChannelInterrupt, I: interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>>>(
```
Dirbaio: shouldn't be pub

embassy-stm32/src/dma/mod.rs
```diff
@@ -76,8 +85,45 @@ pub(crate) trait ChannelInterrupt {
 #[allow(private_bounds)]
 pub trait Channel: SealedChannel + PeripheralType + Into<AnyChannel> + 'static {}
 
+/// Extension trait providing `degrade()` for DMA channels with interrupt binding verification.
+pub trait ChannelExt: Channel + ChannelInterrupt {
```
Dirbaio: is there a way this could be in ChannelInterrupt? this way we'd avoid having an extra trait.

lulf:
Great addition, thanks :rocket: ! This could've used a CHANGELOG.md update for embassy-stm32.

PR #5317, opened 2026-01-27
fix(sdmmc): respect peripheral bus width configuration in SD card initialization

vincentberthier:
# Summary

  - Fix bus width mismatch when using Sdmmc::new_1bit() that caused CRC errors during SD card data transfers
  - Add memory barriers (DSB) around IDMA operations for proper CPU/DMA memory ordering

Related to #4723 

# Problem

  When creating an SDMMC peripheral with Sdmmc::new_1bit(), the driver would still send ACMD6 to configure the SD card for 4-bit mode. This caused a mismatch between:

  - Peripheral configuration: 1-bit mode (only D0 line used)
  - Card configuration: 4-bit mode (expecting D0-D3 lines)

  Result: The card sends data on lines D1-D3 that the peripheral isn't listening to, causing CRC errors (DCRCFAIL) during data transfers.

# Solution

  The fix queries the peripheral's configured bus width before sending ACMD6 and only enables 4-bit mode if both:
  1. The peripheral was initialized with 4 data lines (new_4bit/new_8bit)
  2. The card supports 4-bit mode (SCR register indicates support)

  ```rust
  // Before (always tries 4-bit if card supports it)
  let (bus_width, acmd_arg) = if !self.info.scr.bus_width_four() {
      (BusWidth::One, 0)
  } else {
      (BusWidth::Four, 2)
  };
```
```rust
  // After (respects peripheral configuration)
  let (bus_width, acmd_arg) = match configured_bus_width {
      BusWidth::Four if self.info.scr.bus_width_four() => (BusWidth::Four, 2),
      _ => (BusWidth::One, 0),
  };
```

# Additional Changes

  DSB (Data Synchronization Barrier) instructions are added around IDMA operations to ensure proper memory ordering between CPU and DMA. This prevents potential race conditions where:
  - DMA might access stale buffer data
  - CPU might read data before DMA has completed writing

# Test Plan

  - Tested on STM32H753ZI with Sdmmc::new_1bit() - SD card initialization succeeds
  - Verified block read/write operations complete without CRC errors
  - Should be tested with new_4bit() to ensure 4-bit mode still works for hardware supporting it

# Breaking Changes

  None. This is a bug fix that makes new_1bit() work as documented.

### DIFF
--- a/embassy-stm32/src/sdmmc/mod.rs
+++ b/embassy-stm32/src/sdmmc/mod.rs
@@ -5,10 +5,10 @@ use core::default::Default;
 use core::future::poll_fn;
 use core::marker::PhantomData;
 use core::slice;
-use cortex_m::asm::dsb;
 use core::task::Poll;
 
 use aligned::{A4, Aligned};
+use cortex_m::asm::dsb;
 use embassy_hal_internal::{Peri, PeripheralType};
 use embassy_sync::waitqueue::AtomicWaker;
 use sdio_host::Cmd;


xoviat:
can you run cargo +nightly fmt?

vincentberthier:
> can you run cargo +nightly fmt?

Done sorry.

PR #5318, opened 2026-01-28
MAX30102 support tested on STM32WBA6 and STM32F302R8

cesardtamayo:
Adding i2c support examples for STM32WBA6 and STM32F302R8 using MAX30102 heart rate sensor. 

@Dirbaio , @leftger - I'm going to move the Max driver to it's own crate and once it's published, i can use it from the i2c examples. Does that sound good?


cesardtamayo:
Duplicate work. Will probably re-submit as simpler i2c example.

PR #5319, opened 2026-01-28
stm32: replace assert_dma_binding with dma_into

xoviat:


Dirbaio:
i'm not sure if this is an imprpovement from the user's point of view.

Before:

```rust
ch.degrade(irqs)
```

After:

```rust
use embassy_stm32::dma::dma_into;
dma_into(ch, irqs)
```

it's nicer if it's a method, and the previous name was clearer imo

PR #5321, opened 2026-01-28
[MCXA] i3c: add Instance trait

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/121

### DIFF
--- a/embassy-mcxa/src/i3c/controller.rs
+++ b/embassy-mcxa/src/i3c/controller.rs
@@ -6,14 +6,13 @@ use embassy_hal_internal::Peri;
 use embassy_hal_internal::drop::OnDrop;
 use embassy_hal_internal::interrupt::InterruptExt;
 
-use super::{Async, Blocking, Error, Info, InterruptHandler, Mode, SclPin, SdaPin};
+use super::{Async, Blocking, Error, Info, Instance, InterruptHandler, Mode, SclPin, SdaPin};
 use crate::clocks::periph_helpers::{Div4, I3cClockSel, I3cConfig};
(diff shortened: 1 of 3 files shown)


embassy-mcxa/src/i3c/mod.rs
jamesmunns: We probably need to update this machinery so that these specific pins are implemented for I3C0, but allow for `I3Cn`.

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -540,11 +540,11 @@ impl<'d> I3c<'d, Async> {
     /// Create a new asynchronous instance of the I3C controller bus driver.
     ///
     /// Any external pin will be placed into Disabled state upon Drop.
-    pub fn new_async(
-        peri: Peri<'d, I3C0>,
+    pub fn new_async<T: Instance>(
+        peri: Peri<'d, T>,
         scl: Peri<'d, impl SclPin<I3C0>>,
```
jamesmunns: This should probably also be `impl SclPin<T>`

embassy-mcxa/src/i3c/mod.rs
felipebalbi: Argh, good catch.

embassy-mcxa/src/i3c/controller.rs
felipebalbi: done

embassy-mcxa/src/i3c/mod.rs
felipebalbi: done

PR #5322, opened 2026-01-28
[MCXA] i2c: split errors into Setup and I/O variants

felipebalbi:
Instead of having a single error type, make dedicated types. This should allow users to have simpler error matching.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/122

### DIFF
--- a/embassy-mcxa/src/i2c/controller.rs
+++ b/embassy-mcxa/src/i2c/controller.rs
@@ -14,37 +14,6 @@ use crate::gpio::{AnyPin, SealedPin};
 use crate::interrupt;
 use crate::interrupt::typelevel::Interrupt;
 
-/// Error information type
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[non_exhaustive]
-pub enum Error {
-    /// Setup errors
-    Setup(SetupError),
-    /// I/O errors
-    IO(IOError),
-    /// Other internal errors or unexpected state.
-    Other,
-}
-
-impl From<Error> for SetupError {
-    fn from(value: Error) -> Self {
-        match value {
-            Error::Setup(setup) => setup,
-            _ => Self::Other,
-        }
-    }
-}
-
-impl From<Error> for IOError {
-    fn from(value: Error) -> Self {
-        match value {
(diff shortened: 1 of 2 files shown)


embassy-mcxa/src/i2c/controller.rs
```diff
@@ -19,8 +19,48 @@ use crate::interrupt::typelevel::Interrupt;
...
 pub enum Error {
+    /// Setup errors
+    Setup(SetupError),
+    /// I/O errors
+    IO(IOError),
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
+impl From<Error> for SetupError {
```
jamesmunns: Same note as i3c about From

embassy-mcxa/src/i2c/controller.rs
```diff
@@ -740,22 +778,22 @@ impl<'d, M: Mode> embedded_hal_02::blocking::i2c::Transactional for I2c<'d, M> {
         if let Some((last, rest)) = operations.split_last_mut() {
             for op in rest {
                 match op {
-                    embedded_hal_02::blocking::i2c::Operation::Read(buf) => {
-                        self.blocking_read_internal(address, buf, SendStop::No)?
-                    }
-                    embedded_hal_02::blocking::i2c::Operation::Write(buf) => {
-                        self.blocking_write_internal(address, buf, SendStop::No)?
-                    }
+                    embedded_hal_02::blocking::i2c::Operation::Read(buf) => self
```
jamesmunns: Same note as I3c, we should probably use IOError as the Error type for the e-hal impl

embassy-mcxa/src/i2c/target.rs
jamesmunns, line 1: Same note re: Errors for target as controller

embassy-mcxa/src/i2c/controller.rs
felipebalbi: That's how it was before, but then I didn't get the free `Into` counterpart.

embassy-mcxa/src/i2c/controller.rs
felipebalbi: if I do that, then `Error` can be completely removed as it will be unused. I'll do that.

PR #5324, opened 2026-01-28
[MCXA] i3c: split errors into Setup and I/O variants

felipebalbi:


embassy-mcxa/src/i3c/controller.rs
```diff
@@ -6,16 +6,97 @@ use embassy_hal_internal::Peri;
...
+    AddressOutOfRange(u8),
+    /// Invalid write buffer length.
+    InvalidWriteBufferLength,
+    /// Invalid read buffer length.
+    InvalidReadBufferLength,
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
+impl From<Error> for IOError {
```
jamesmunns: Should this actually be `From<IOError> for Error`?

embassy-mcxa/src/i3c/controller.rs
```diff
@@ -813,15 +894,15 @@ impl<'d, M: Mode> embedded_hal_02::blocking::i2c::Read for I3c<'d, M> {
     type Error = Error;
 
     fn read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Error> {
-        self.blocking_read(address, buffer, BusType::I2c)
+        self.blocking_read(address, buffer, BusType::I2c).map_err(Error::IO)
     }
 }
 
 impl<'d, M: Mode> embedded_hal_02::blocking::i2c::Write for I3c<'d, M> {
     type Error = Error;
```
jamesmunns: Should these use e-hal impls use `IOError` instead?

embassy-mcxa/src/i3c/controller.rs
felipebalbi: Removed

embassy-mcxa/src/i3c/controller.rs
felipebalbi: Done

PR #5329, opened 2026-01-29
[MCXA]: Add optional perf counter feature

jamesmunns:
Add perf counters that can be used for debugging interrupt/wake counts.

In the future, we will likely also count WFE and deep sleep entries.

### DIFF
--- a/embassy-mcxa/src/perf_counters.rs
+++ b/embassy-mcxa/src/perf_counters.rs
@@ -1,3 +1,12 @@
+//! # Performance counters
+//!
+//! This module contains simple performance counters, intended to aid debugging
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/perf_counters.rs
```diff
@@ -0,0 +1,158 @@
+#[cfg_attr(not(feature = "perf"), allow(unused_imports))]
```
felipebalbi, line 10: Minor nit: a one liner `//! Performance counters` would probably be desirable, but we need to document the entire HAL anyway, can be done in a follow-up PR.

PR #5336, opened 2026-01-30
[MCXA]: Implement lower power light-sleep

jamesmunns:
This PR fills a few missing gaps for low power operation, currently only down to Light Sleep (WFE with the clock gated).

I've achieved 700uA in light sleep (with no peripherals), vs the claimed 580uA in the datasheet. This is likely "close enough" for now, I'll probably begin implementing Deep Sleep after this, as there are likely to be more gaps necessary to reach the target current goals.

### DIFF
--- a/examples/mcxa/src/bin/power-wfe-gated.rs
+++ b/examples/mcxa/src/bin/power-wfe-gated.rs
@@ -18,9 +18,9 @@
 use embassy_executor::Spawner;
 use embassy_mcxa::clocks::PoweredClock;
 use embassy_mcxa::clocks::config::{
-    CoreSleep, Div8, FlashSleep, Fro16KConfig, MainClockConfig, MainClockSource, VddDriveStrength, VddLevel,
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/clocks/mod.rs
```diff
@@ -1812,14 +1779,72 @@ impl ClockOperator<'_> {
...
         }
 
+        // TODO: is there a nicer way to "steal" the SCB peripheral? This is an emulation
+        // of `SCB::set_sleepdeep()`, but that requires `&mut SCB`, but `PTR` only gives us the
+        // `RegisterBlock`?
+        fn scb_set_sleepdeep() {
+            let scb = unsafe { &mut *cortex_m::peripheral::SCB::PTR.cast_mut() };
+            unsafe {
+                scb.scr.modify(|w| w | (0x1 << 2));
+            }
```
felipebalbi: does the following work?

```
let mut cp = unsafe { cortex_m::Peripherals::steal() };
cp.SCB.set_sleepdeep();
```

PR #5337, opened 2026-01-30
[MCXA] CTIMER: Add Ctimer PWM driver

felipebalbi:
Still some work left on this, for now creating as a draft PR.

@jamesmunns, are there any elegant ways to ensure `match_ch` and `pin` agree on the CTimer channel being used?

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/69
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/113

### DIFF
--- a/embassy-mcxa/src/ctimer/mod.rs
+++ b/embassy-mcxa/src/ctimer/mod.rs
@@ -1,0 +1,424 @@
+//! CTimer driver.
+
+use core::marker::PhantomData;
+
+use embassy_hal_internal::{Peri, PeripheralType};
+use maitake_sync::WaitCell;
+
+use crate::clkout::Div4;
+use crate::clocks::periph_helpers::{CTimerClockSel, CTimerConfig};
+use crate::clocks::{ClockError, Gate, PoweredClock, WakeGuard, enable_and_reset};
+use crate::gpio::{GpioPin, SealedPin};
+use crate::{interrupt, pac};
+
+pub mod pwm;
+
+/// Error information type
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum Error {
+    /// Clock configuration error.
+    ClockSetup(ClockError),
+
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
+/// CTimer configuration
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub struct Config {
+    /// Powered clock configuration
+    pub power: PoweredClock,
+    /// CTimer clock source
+    pub source: CTimerClockSel,
+    /// CTimer pre-divider
+    pub div: Div4,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            power: PoweredClock::NormalEnabledDeepSleepDisabled,
+            source: CTimerClockSel::FroLfDiv,
+            div: const { Div4::no_div() },
+        }
+    }
+}
+
+/// CTimer core driver.
+#[derive(Clone)]
+pub struct CTimer<'d> {
+    info: &'static Info,
+    _freq: u32,
+    _wg: Option<WakeGuard>,
+    _phantom: PhantomData<&'d mut ()>,
+}
+
+impl<'d> CTimer<'d> {
+    /// Create a new instance of the CTimer core cdriver.
+    pub fn new<T: Instance>(_peri: Peri<'d, T>, config: Config) -> Result<Self, Error> {
+        // Enable clocks
+        let conf = CTimerConfig {
+            power: config.power,
+            source: config.source,
+            div: config.div,
+            instance: T::CLOCK_INSTANCE,
+        };
+
+        let parts = unsafe { enable_and_reset::<T>(&conf).map_err(Error::ClockSetup)? };
+
+        let inst = Self {
+            info: T::info(),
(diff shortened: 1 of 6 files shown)


embassy-mcxa/src/ctimer/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+        self.ch2
+    }
+
+    /// Consume `self` and return Channel 3
+    pub fn ch3(self) -> Channel<'d> {
+        self.ch3
+    }
+
+    /// Split `self` into all four channels
+    pub fn split(self) -> (Channel<'d>, Channel<'d>, Channel<'d>, Channel<'d>) {
```
jamesmunns: Any reason not to just make the fields public? Then we don't need all these destructor methods or the `ChannelsRef` type at all

embassy-mcxa/src/ctimer/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+            source: config.source,
+            div: config.div,
+            instance: T::CLOCK_INSTANCE,
+        };
+
+        let parts = unsafe { enable_and_reset::<T>(&conf).map_err(Error::ClockSetup)? };
+
+        let inst = Self {
+            info: T::info(),
+            _freq: parts.freq,
```
jamesmunns, line 74: Check that `freq` != 0? I think this is only possible if the user selects a "Disabled" clock source, not sure if you expose that

embassy-mcxa/src/ctimer/pwm.rs
```diff
@@ -0,0 +1,221 @@
...
+    max_period: u16,
+}
+
+impl<'d> Pwm<'d> {
+    /// Create Pwm driver with a single pin as output.
+    ///
+    /// Upon `Drop`, the external `pin` will be placed into `Disabled`
+    /// state.
+    pub fn new_single_output<T: Instance>(
+        pwm_ch: Channel<'d>,
```
jamesmunns: You'll probably need the `Channel` to remember it's `Instance` number, or use the instance ptr/ref to ensure that the pwm and match channel belong to the same instance.

We could enforce this statically if we retain the `T: Instance` on the `Channel` struct, because then you would take two `Channel<'d, T>`, which would have to be the same `T`. If it's desirable to type erase already at the Channel stage, we'd need to check it at runtime.

embassy-mcxa/src/ctimer/mod.rs
felipebalbi: made fields public. Indeed, it looks nicer :-p

embassy-mcxa/src/ctimer/pwm.rs
felipebalbi: I think I can add something like `CTIMERn_CHm` zst, make them peripherals, then I can add a `ValidMatchConfig` trait and implement it for the tuple `(T, CTIMERn_CHm, Pi_j)`, then I can have that pair give me the match channel number.

Something like:

```rust
trait SealedValidMatchConfig {
    fn number() -> usize;
}

#[allow(private_bounds)]
trait ValidMatchConfig: SealedValidMatchConfig { }

// ...

imp<'d> Pwm<'d> {
    pub fn new_single_output<T: Instance, MATCH: PwmSlice<T>, PIN: OutputPin<T>>(
        period_ch: Peri<'d, impl PwmSlice<T>>,
        match_ch: Peri<'d, MATCH>,
        pin: Peri<'d, PIN>,
        config: Config,
    ) -> Result<Self, PwmError>
    where (T, MATCH, PIN): ValidMatchConfig)
    {
         let match_ch= <(T, MATCH, PIN) as SealedValidMatchConfig>::number();
        // ...
    }
}
```

What do you think?

felipebalbi: Updated the PR according to the description in the previous comment. I still think there are ways to improve the API slightly. Maybe I don't need the `CTimer` object at all? But in that case, how could we guarantee that clocks are kept on until the last user of the `p.CTIMERn` peripheral is done using it?

I mean, a user can use 2 channels in CTIMER0 for PWM and the other two for Capture functionality (useful as a simple Tachometer decoder for regular fans).

felipebalbi: With the current setup, I can also do things like this:

```rust
    let led_ctimer = CTimer::new(p.CTIMER2.reborrow(), Irqs, Default::default()).unwrap();
    let mut red_pwm = Pwm::new(
        led_ctimer.clone(),
        p.CTIMER2_CH3,
        p.CTIMER2_CH0,
        p.P3_18,
        Default::default(),
    )
    .unwrap();
    let mut green_pwm = Pwm::new(led_ctimer, p.CTIMER2_CH2, p.CTIMER2_CH1, p.P3_19, Default::default()).unwrap();
```

by deriving `Clone` on the `CTimer` object. Not sure how I feel about that. It's certainly useful, but then I consume two channels for match when I could use a single one. That being said, using a different channel for match on different PWMs lets the PWMs run at different periods as well.

UPDATE:

I could also add `SinglePwm`, `DualPwm`, and `TriplePwm` drivers for those cases when you want to base period the be the same, then a single match channel shared for the various outputs would work.

felipebalbi: Hmm, so it looks like adding `SinglePwm`, `DualPwm`, and `TriplePwm` is the way to go. Then `DualPwm`, and `TriplePwm` could provide `split()` method to split the channels into some `SinglePwm` representation to allow changing duty cycle on a single channel in isolation.

Issue #5340, opened 2026-01-30
e-net-tuntap: Allow setting hardware address


chrysn:
The current version produces a constant hardware address, which may cause trouble when attaching to a bridge.

I'll come up with a PR, likely along the lines of "carry the MAC address in self, with a default as currently is, and a `.set_hardware_address()` method".

As part of touching the module, I'd consider doing some general cleanup in that module (using cstrings, and importing constants from the anyway used libc), and maybe even using the tun-rs crate as a backend (as it supports a wider range of platforms than just Linux) – would such a change be appreciated?

Dirbaio:
Of course, all good ideas 👍

PR #5341, opened 2026-01-30
e-net-tuntap: allow setting hardware address; minor refactor

chrysn:
Closes: https://github.com/embassy-rs/embassy/issues/5340

The requirements to hardware addresses are not spelled out because I'm not sure what they are, but given that this module will likely not be touched on minor smoltcp updates, this might be for the better.

From the small refactorings on the top: Using the libc constants removes some previously present unused constants that might have been hints at planned behavior changes. Were those concrete plans or just things that have been played with and eventually were not needed?

(If you prefer to keep affairs separate, I can take the refactoring out and file separately.)

### DIFF
--- a/embassy-net-tuntap/src/lib.rs
+++ b/embassy-net-tuntap/src/lib.rs
@@ -11,21 +11,6 @@ use async_io::Async;
 use embassy_net_driver::{Capabilities, Driver, HardwareAddress, LinkState};
 use log::*;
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):

Dirbaio:
> The requirements to hardware addresses are not spelled out because I'm not sure what they are,

can be anything as long as the multicast bit is not set.

PR #5343, opened 2026-01-31
Use async OutputPin for shared Spi bus

ChocolateLoverRaj:
This is a breaking change and imo that's okay.

Similar to #5312, Once the async `OutputPin` is in a published version, we can  remove the  `patch.crates-io` and mark this as not a  draft.


Dirbaio:

PR #5345, opened 2026-01-31
Improve chrono integration into embassy-rp

fakusb:
Two bugs are fixed here:
- embassy-rp did not actually compile with the chrono-feature because chrono calls e.g. Monday `Mon` instead of `Monday`
- chrono did not  derive a trait needed for defmt before v0.4.43

In theory, it would be nice if the crate would be build with the chrono feature enabled in the CI, but I do neither know how I would achieve this, nor if this is desired.
Furthermore, the chrono feature for embassy-rp changes the public api (e.g. the examples do not compile anymore), but this is true regardless of this MR.

### DIFF
--- a/embassy-rp/src/datetime/datetime_chrono.rs
+++ b/embassy-rp/src/datetime/datetime_chrono.rs
@@ -23,6 +23,7 @@ pub enum Error {
 ///
 /// # Errors
 /// Returns error if DateTime is before 1970-01-01.
+#[cfg(feature = "_rp235x")]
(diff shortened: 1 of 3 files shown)


Dirbaio:
can you add the previously-failing feature combinations here so they don't regress again? https://github.com/embassy-rs/embassy/blob/main/embassy-rp/Cargo.toml#L13

fakusb:
I'll have a look into fixing the warnings later today or this week.

fakusb:
Fixing the unused warnings in the several feature combinations needed a bit of conditional compilation.
It's not pretty, but it represents the current state of the chrono feature.

PR #5363, opened 2026-02-03
embassy-rp::dma: Refactor change confusing names

dobrowolski-lukasz:
Rename some methods that have confusing name:
embassy-rp::dma::Channel read_repeated -> read_discard
embassy-rp::dma::Channel write_repeated -> write_zeros
embassy_rp::pio::StateMachineRx dma_pull_repeated -> dma_pull_discard
embassy_rp::pio::StateMachineTx dma_push_repeated -> dma_push_zeros

I believe that those names are confusing not only to me. I distinctly remember reading a comment somewhere(reddit?) that clearly showed somebody mistaking those for chained DMA transfers. I can try and find it if need be.

Tests: I've cargo checked code and rp/rp235x examples.

### DIFF
--- a/embassy-rp/src/pio/mod.rs
+++ b/embassy-rp/src/pio/mod.rs
@@ -478,7 +478,7 @@ impl<'d, PIO: Instance, const SM: usize> StateMachineTx<'d, PIO, SM> {
     }
 
     /// Prepare a repeated DMA transfer to TX FIFO.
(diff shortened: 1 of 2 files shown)


Dirbaio:
makes a lot of sense. I was thinking the same when working on the `Channel` refactor the other day :)

PR #5366, opened 2026-02-03
embassy-rp::dma: Rm pub(crate) from Channel::{regs,number}

dobrowolski-lukasz:
I'm not 100% sure of this change. Feedback is welcome.

I don't see the need  for `pub(crate)` for `Channel::regs(&self)` or `Channel::number(&self)`.
Former has single use that I've replaced with new `pub fn write_addr(&self)` and later has no uses.
I feel that those would only be needed to be used inside dma module or outside of embassy-rp crate itself. Making `pub(crate)` useless for in both cases.

Tests: I've cargo checked code and rp/rp235x examples.

### DIFF
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -69,8 +69,13 @@ impl<'d> Channel<'d> {
     }
 
     /// Get the channel register block.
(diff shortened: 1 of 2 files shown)


Dirbaio:
looks good. the only use was in PIO yep

PR #5368, opened 2026-02-03
Allow linker to trim large STM32 PLL initialization functions when PLL is unused

WillaWillNot:
Currently, most of the STM32 RCC initialization functions call their family's equivalent of `init_pll()` while passing user PLL configuration instructions wrapped in an Option. When it's passed None, `init_pll()` returns quickly after doing a few register writes to make sure the PLL is disabled. But since the function is called at all, the linker has to put it in the final binary. This is a problem because PLL initialization code is quite large, and this bloats the size of very thin binaries that deliberately avoid using the PLL but still need to call `embassy_stm32::init()` (e.g. bootloaders.)

This change moves the None check to before calling `init_pll()` so that link time optimization can trim the whole function if the user disables the PLL. For example, if you add `lto = "fat"` to the `[profile.release]` in /examples/stm32u5/Cargo.toml, then run `cargo bloat --release --bin blinky`, the `.text` section goes from 12.0kB before this change to 10.6kB afterwards.

### DIFF
--- a/embassy-stm32/src/rcc/wba.rs
+++ b/embassy-stm32/src/rcc/wba.rs
@@ -150,7 +150,6 @@ pub(crate) unsafe fn init(config: Config) {
 
     let pll_input = PllInput { hse, hsi };
 
-    let pll1 = init_pll(config.pll1, &pll_input, config.voltage_scale);
     let pll1 = config.pll1.map_or_else(
         || {
             pll_enable(false);


examples/stm32f0/src/bin/i2c_slave_async.rs
```diff
@@ -29,7 +29,7 @@ const BUFFER_SIZE: usize = 32;
 
 bind_interrupts!(struct Irqs {
     I2C1 => i2c::EventInterruptHandler<peripherals::I2C1>, i2c::ErrorInterruptHandler<peripherals::I2C1>;
-    DMA1_CHANNEL2_3 => dma::InterruptHandler<peripherals::DMA1_CH2>, dma::InterruptHandler<peripherals::DMA1_CH3>;
+    DMA1_CH2_3_DMA2_CH1_2 => dma::InterruptHandler<peripherals::DMA1_CH2>, dma::InterruptHandler<peripherals::DMA1_CH3>;
```
pilotchute, line 32: This change might not be required in this PLL clock init() flow update.

pilotchute: Otherwise this is a really good change, a more than 10% change in program size is impressive. Thanks for finding and fixing it!

PR #5379, opened 2026-02-04
[MCXA] CTimer: Add SinglePwm, DualPwm, and TriplePwm to share the period channel

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/127

### DIFF
--- a/embassy-mcxa/src/ctimer/mod.rs
+++ b/embassy-mcxa/src/ctimer/mod.rs
@@ -224,9 +224,7 @@ trait SealedInputPin {
 }
 
 /// Seal a trait
-trait SealedOutputPin<T: Instance> {
-    fn number(&self) -> usize;
-}
+trait SealedOutputPin<T: Instance> {}
 
 /// CTimer input pin.
 #[allow(private_bounds)]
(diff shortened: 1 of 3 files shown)


asasine (commented):
That this solution is limited to 3 (or any arbitrary N) that has to be implemented manually feels like an incomplete solution to me.

The solution I was envisioning was approximately some `FixedPeriodPwm<'pch>` that immutably borrows the period channel for `'pch` and any number of `FixedPeriodPwm<'pch>` may exist but none of them can modify the period channel via the borrow checker's pointer aliasing rules.

felipebalbi:
> That this solution is limited to 3 (or any arbitrary N) that has to be implemented manually feels like an incomplete solution to me.

The CTimer has 4 channels. One *must* be used as period. You can use more than one CTimer instance and get more channels like that.

That is, this solution is not limited to an arbitrary number. It's limited to what the HW is capable of providing.

> The solution I was envisioning was approximately some `FixedPeriodPwm<'pch>` that immutably borrows the period channel for `'pch` and any number of `FixedPeriodPwm<'pch>` may exist but none of them can modify the period channel via the borrow checker's pointer aliasing rules.

Even with that, you can only share the period within the same CTimer instance. The result would be equal: 3 PWMs per CTimer instance. Also, there's a limitation to how many pins you can use. In practice, you can use at most 15 PWM channels if you're dedicating all existing 20 CTimer channels (and 15 very specific GPIOs) for PWM output.

PR #5380, opened 2026-02-04
Fix N6 EXTI

jake-taf:
The EXTI mapping looks like this for the STM32N6, but notice how it skips a few ports (I-M) after H. When the EXTI driver sets pin PN4 for example, it tries to set 13 (A=0, Z=25) when it should set 8. Because we're setting the wrong value, we never end up getting the interrupt.

This patch adjusts the port number if it's port N or above on the N6.

```Rust
A:0
B:1
C:2
D:3
E:4
F:5
G:6
H:7
N:8
O:9
P:10
Q:12
```


Dirbaio:
ST forgot the alphabet :skull_and_crossbones:

PR #5381, opened 2026-02-04
N6 LTDC Support

jake-taf:
Adds support for the LTDC on the N6. It uses a different signal name (`LO` vs `GLOBAL`) which trips it up. With this change I have the LTDC working on the N6 Disco kit.

I'm not sure if this is the best way to handle it (duplicates the whole macro), so if there's a nicer way let me know.

### DIFF
--- a/embassy-stm32/src/ltdc.rs
+++ b/embassy-stm32/src/ltdc.rs
@@ -144,6 +144,7 @@ pub enum PixelFormat {
     /// AL88 (8-bit alpha, 8-bit luminance)
     AL88 = Pf::AL88 as u8,
 }
+/// Pixel format
 #[repr(u8)]
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]


Dirbaio:
the inconsistency should be fixed in stm32-data https://github.com/embassy-rs/stm32-data

jake-taf:
@Dirbaio I think it might be a case of extra interrupts rather than just the name? The N6 reference manual mentions duplicating the interrupts for unsecure and secure for a total of four interrupt output pins: global working, global error, secure global, secure global error.

In [the N6 chip data](https://github.com/embassy-rs/stm32-data-generated/blob/0c284a3084be274c180ad0dc467c984d935023f6/data/chips/STM32N655A0.json#L3912) it has `LTDC_UP`, `LTDC_UP_ERR`, `LTDC_LO`, `LTDC_LO_ERR`. When N6 support was added, there were changed made that expect `LTDC_LO`, not `LTDC_GLOBAL` in [stm32-data PR 612](https://github.com/embassy-rs/stm32-data/pull/612/changes#diff-6c69ecd684108b7568158626fa2711235c8471215d59144b72af59b8aec0dda8)

@tarfu Might have a better understanding than me

jake-taf:
I do need to add some changes to stm32-data as a twin PR, hence why this is a draft. I will try to have that up shortly (N6 LTDC v1.3 has a different PixelFormat, Clutwr address, and Layer address)

Dirbaio:
stm32-data maps interrupts through this concept of "irq signals". Signals are "logical". ie when an error happens the ERR signal is triggered, when an update happens the UP signal is triggered (or whatever). Then, each logical signal is mapped to a physical interrupt.

The goal is to handle chips with different interrupt mapping, where different signals sometimes go to different physical interrupts and sometimes to the same one.

The "GLOBAL" signal is a fallback that stm32-data-gen does when it doesn't "know" the signals one peripheral has. It assumes there's only 1 signal named "GLOBAL", wired to 1 physical interrupt.

Now that ST has decided to split irqs for one family (N6) we should update the data to be consistent about it: split signals, ensure there's no fallback "GLOBAL" signal. ie:

N6:
- signal ERR -> irq LTDC_ERR
- signal UP -> irq LTDC_UP
- signal LO -> irq LTDC_LO

others:
- signal ERR -> irq LTDC
- signal UP -> irq LTDC
- signal LO -> irq LTDC

and no GLOBAL signal anywhere.

Then the driver can just bind the LO interrupt signal and it'll do the right thing on all families.

jake-taf:
I think I see what you mean. Would you mind giving some more details (e.g. which files to take a look at). I can then look into making the needed changes.

Probably in [stm32-data-gen/src/interrupts.rs](https://github.com/embassy-rs/stm32-data/tree/main/stm32-data-gen/src/interrupts.rs) around [here](https://github.com/embassy-rs/stm32-data/blob/2846f7ae645cef28b8b9728927e3482fcfd74f6e/stm32-data-gen/src/interrupts.rs#L379-L431) and [here](https://github.com/embassy-rs/stm32-data/blob/2846f7ae645cef28b8b9728927e3482fcfd74f6e/stm32-data-gen/src/interrupts.rs#L516-L519)?

Dirbaio:
Yes, there. Also you probably have to add LTDC with the 3 signals here https://github.com/embassy-rs/stm32-data/blob/2846f7ae645cef28b8b9728927e3482fcfd74f6e/stm32-data-gen/src/interrupts.rs#L552

the logic is roughly: if a peri has a single irq, all signals are mapped to that. if it has multiple, it tries to deduce the mapping from the irq names. 

It's a bit of a mess yeah...

tarfu:
The UP IRQ is specific to the secure domain; as long as you ignore the secure domain, the following diff should be enough if I haven't overlooked something.  Looking at the old PR I probably just forgot to fix that after getting it to generate.

The UP signal is a bit of a headache, especially if you want to map ERR/ER to LO_ERR and UP_ERR depending on the domain.

```
diff --git stm32-data-gen/src/interrupts.rs stm32-data-gen/src/interrupts.rs
index dd847d916..cc3c969f4 100644
--- stm32-data-gen/src/interrupts.rs
+++ stm32-data-gen/src/interrupts.rs
@@ -341,6 +341,15 @@ impl ChipInterrupts {
                 }
 
                 for (p, mut ss) in peri_signals.into_iter() {
+                    // Normalize LTDC error signal: N6 uses "ERR" (from LTDC_LO_ERR), standardize to "ER"
+                    if p == "LTDC" {
+                        for s in ss.iter_mut() {
+                            if s == "ERR" {
+                                *s = "ER".to_string();
+                            }
+                        }
+                    }
+
                     let known = valid_signals(&p, chip_name);
 
                     // If we have no interrupt_signals for the peri, assume it's "global" so assign it all known ones
@@ -405,6 +414,15 @@ impl ChipInterrupts {
                         }
                     }
 
+                    // For LTDC on non-N6 chips: prefer the base LTDC interrupt over LTDC_ER
+                    // so that both ER and LO signals map to the same LTDC interrupt
+                    if irqs.len() != 1 && p.name == "LTDC" && signal == "ER" {
+                        // Keep LTDC (without _ER suffix) if available
+                        if irqs.contains("LTDC") {
+                            irqs.retain(|irq| irq == "LTDC");
+                        }
+                    }
+
                     // If there's a duplicate irqs in a signal other than "global", keep the non-global one.
                     if irqs.len() != 1 && signal != "GLOBAL" {
                         irqs.retain(|irq| !globals.contains(irq));
@@ -514,9 +532,6 @@ fn valid_signals(peri: &str, chip_name: &str) -> Vec<String> {
     // Special chip-specific signal mappings
     // Format: (peripheral_prefix, chip_pattern, signals)
     const CHIP_SPECIFIC_SIGNALS: &[(&str, &str, &[&str])] = &[
-        // LTDC signals: STM32N6 supports all signals, others only basic ones
-        ("LTDC", "STM32N6", &["GLOBAL", "ER", "LO", "ERR"]),
-        ("LTDC", "*", &["GLOBAL", "ER"]), // Default for all other LTDC devices
         // DCMIPP signals: STM32N6 supports CSI, others only basic signals
         ("DCMIPP", "STM32N6", &["GLOBAL", "CSI"]),
         ("DCMIPP", "*", &["GLOBAL"]), // Default for all other DCMIPP devices
@@ -586,6 +601,7 @@ fn valid_signals(peri: &str, chip_name: &str) -> Vec<String> {
         ("ADF", &["FLT0"]),
         ("RAMECC", &["ECC"]),
         ("RAMCFG", &["BKP", "ECC"]),
+        ("LTDC", &["ER", "LO"]),
     ];
 
     for (prefix, signals) in IRQ_SIGNALS_MAP {
```

jake-taf:
Related PR for stm32-data (work in progress) https://github.com/embassy-rs/stm32-data/pull/728

jake-taf:
The pixel format is the only difference for the drivers between lcdtft1_v1_3 and lcdtft1_v1_0/1/2. I've used `#[cfg(not(ltdc_v1_3))]` and `#[cfg(ltdc_v1_3)]` to select that. For some reason when ltdc_v1_3 is active so is ltdc_v1, so I wasn't able to gate it with `#[cfg(ltdc_v1)]`. Let me know if there's a better way.

xoviat:
@jake-taf can you update the pac and make the build pass?

PR #5382, opened 2026-02-04
Clarify that zerocopy channels wait for space

shepmaster:
While there is `try_send`, it doesn't look like that's the primary way of using the channel.

### DIFF
--- a/embassy-sync/src/zerocopy_channel.rs
+++ b/embassy-sync/src/zerocopy_channel.rs
@@ -12,7 +12,7 @@
 //!
 //! This module provides a bounded channel that has a limit on the number of
 //! messages that it can store, and if this limit is reached, trying to send
-//! another message will result in an error being returned.
+//! another message will wait until a message is received from the channel.
 
 use core::cell::RefCell;
 use core::future::{Future, poll_fn};


Dirbaio:
the new wording is also wrong technically because try_send exists. perhaps "will wait until there's space or return an error, depending on the method" and check each method documents what it does in case it's full?

shepmaster:
I noticed that the documentation [for the *type*](https://docs.embassy.dev/embassy-sync/git/default/zerocopy_channel/struct.Channel.html) and [the *module*](https://docs.embassy.dev/embassy-sync/git/default/zerocopy_channel/index.html) disagree on the phrasing. Looking broader, this issue also exists between the non-zero-copy channel type and module as well.

Another option is to borrow the phrasing from `Channel`, which calls out `send` specifically:

> Once the buffer is full, attempts to `send` new messages will wait until a message is received from the channel.

The current docs also mention "return an error", but that feels incorrect for the zero-copy channel's `try_send` which uses an `Option` and doesn't really feel like "returning an error".

Additionally, should I adjust *both* `channel` and `zerocopy_channel` with whatever we decide?

Dirbaio:
you raise good points. feel free to pick any wording you think is best. I just wanted to point out saying it waits is not fully correct either as some methods wait, some methods fail.

> Additionally, should I adjust both channel and zerocopy_channel with whatever we decide?

sounds good, it's nice to be consistnet.

Dirbaio:
#6661

PR #5390, opened 2026-02-05
[MCXA]: Deep Sleep proof of concept

jamesmunns:
Implement an initial proof-of-concept version of deep sleep. This achieves ~150uA current, while DMA and OsTimer remain enabled.

deep-sleep-pmprot

This has some unfortunate "you have to hold it right" items, but it's a first step and a good baseline to check against when implementing HAL API changes.

This PR also improves the example docs to include the required hardware rework necessary to achieve these numbers.

### DIFF
--- a/embassy-mcxa/src/clocks/mod.rs
+++ b/embassy-mcxa/src/clocks/mod.rs
@@ -791,6 +791,8 @@ impl ClockOperator<'_> {
             self.scg0.firccsr().modify(|w| {
                 w.set_fircsten(Fircsten::DISABLED_IN_STOP_MODES);
                 w.set_fircerr_ie(FircerrIe::ERROR_NOT_DETECTED);
+                w.set_firc_fclk_periph_en(false);
+                w.set_firc_sclk_periph_en(false);
                 w.set_fircen(false);
             });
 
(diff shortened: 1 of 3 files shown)


embassy-mcxa/src/clocks/mod.rs
```diff
@@ -1886,6 +1899,42 @@ impl ClockOperator<'_> {
     }
 }
 
+/// This method ACTUALLY enables deep sleep. See `CorePower::DeepSleep` for
+/// more context
+///
+/// ## SAFETY
+///
+/// This method is ONLY sound if you are ONLY using peripherals or clocks that are
+/// marked as `PoweredClock::AlwaysEnabled` (or are disabled). We do NOT currently
```
felipebalbi, line 1908: Maybe I'm reading this wrong, but shouldn't this be `PoweredClock::NormalEnabledDeepSleepDisabled`?

felipebalbi (approved):

embassy-mcxa/src/clocks/mod.rs
jamesmunns: "AlwaysEnabled" means "isn't turned off in low power mode". Right now we don't turn clocks back on, so if we went to sleep, which could turn off SPLL, and we don't start it back up, then make a peripheral that uses SPLL, it might not work properly.

Since "AlwaysEnabled" clocks never start+stop, they're okay to use for now.

PR #5400, opened 2026-02-06
[MCXA]: Custom Executor

jamesmunns:
Introduce an optional custom executor allowing deep-sleep operation

This executor automatically goes to deep sleep when possible, and respects
WakeGuards that keep the system in (light) WFE sleep.

All of the "building blocks" for this should be public, allowing non-executor
users to perform setup if they desire.

### DIFF
--- a/embassy-mcxa/Cargo.toml
+++ b/embassy-mcxa/Cargo.toml
@@ -24,7 +24,7 @@ flavors = [
 ]
 
 [dependencies]
-cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
+cortex-m = { version = "0.7", features = ["critical-section-single-core", "inline-asm"] }
 # If you would like "device" to be an optional feature, please open an issue.
 cortex-m-rt = { version = "0.7", features = ["device"] }
 critical-section = "1.2.0"
@@ -43,6 +43,9 @@ maitake-sync = { version = "0.2.2", default-features = false, features = ["criti
 nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "560e74384e20940a3c25e0435a6fbdc66644b558", features = ["rt", "mcxa256"] }
 nb = "1.1.0"
 paste = "1.0.15"
+
+# Custom executor support
+embassy-executor = { version = "0.9.1", path = "../embassy-executor" }
 
 # `time` dependencies
 embassy-time = { version = "0.5.0", path = "../embassy-time" }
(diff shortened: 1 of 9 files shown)


embassy-mcxa/src/clocks/mod.rs
```diff
@@ -162,6 +163,170 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+#[inline(always)]
+pub fn active_wake_guards(_cs: &CriticalSection) -> bool {
+    // Relaxed is okay: we are in a critical section
+    LIVE_HP_TOKENS.load(Ordering::Relaxed) != 0
+}
+
+/// Attempt to go to deep sleep if possible.
+///
+/// If we successfully went and returned from deep sleep, this function returns a `true`.
+/// If we were unsuccessful due to active `WaitGuard`s
```
felipebalbi: missing a ", this function returns `false`" here?

embassy-mcxa/src/clocks/mod.rs
```diff
@@ -162,6 +163,170 @@ pub fn with_clocks<R: 'static, F: FnOnce(&Clocks) -> R>(f: F) -> Option<R> {
...
+/// mode after a low power (e.g. deep sleep, power-off) state.
+///
+/// ## Safety
+///
+/// This should only be called in a critical section, immediately after waking up.
+unsafe fn restart_active_only_clocks(_cs: &CriticalSection) {
+    use core::cell::Ref;
+    use core::ops::Deref;
+
+    use nxp_pac::scg::vals::Fircvld;
```
felipebalbi: how about `crate::pac::scg::vals::Fircvld`? Unlikely, but if we ever rename the pac, this won't have to be changed.

embassy-mcxa/Cargo.toml
```diff
@@ -105,3 +108,5 @@ rt = []
 
 # Enable perf counters
 perf = []
+
+custom-executor = []
```
felipebalbi, line 115: maybe call it `low-power-executor`?

jamesmunns: Leaving it as-is for now, in case we want to do other fancy things too :)

jamesmunns:
ppk-20260209T191218

PPK view with the `power-deepsleep-gating` example, showing a 1s interval with a brief running time, 100ms of light WFE sleep due to an active WakeGuard, and then 900ms of Deep Sleep.

PR #5402, opened 2026-02-06
[MCXA]: Apply CoreSleep settings for all modes

jamesmunns:
On WfeUngated mode, force settings back to defaults.

Some of these settings appear not to be set back to default on a warm reset, causing problems when debugging after flashing with a deep sleep using application


felipebalbi:
I can confirm this PR solves the issue was I facing. Thanks @jamesmunns

jamesmunns:
For reference, it was SOD that was the problem, it's only cleared on a "cold" reset, e.g. a power on reset. CKMODE is reset on a "warm" reset, which includes hitting the reset button.

PR #5405, opened 2026-02-07
stm32: trace log when dropping flex pin

GrantM11235:


xoviat:
Is this not sufficient? 

 https://github.com/embassy-rs/embassy/commit/739f7b7c849c54d8324533c3bf78ba3b9656c5aa

PR #5409, opened 2026-02-07
[embassy-rp] Custom multicore-ready async executors

sourcebox:
This PR tries to address #1634 so that interrupt executors can be woken up across cores. I have tested this with a real-world application on a RP2350 (Pico 2), but not on a RP2040. 

The code is based on the suggestions from @Dirbaio in the Matrix room. There's probably better documentation required. I also did not update any examples to use this executors.

### DIFF
--- a/embassy-rp/Cargo.toml
+++ b/embassy-rp/Cargo.toml
@@ -169,8 +169,8 @@ embassy-futures = { version = "0.1.2", path = "../embassy-futures" }
 embassy-hal-internal = { version = "0.4.0", path = "../embassy-hal-internal", features = ["cortex-m", "prio-bits-2"] }
 embassy-embedded-hal = { version = "0.6.0", path = "../embassy-embedded-hal" }
 embassy-usb-driver = { version = "0.2.0", path = "../embassy-usb-driver" }
-embassy-executor = { version = "0.9.1", path = "../embassy-executor", optional = true }
-embassy-executor-macros = { version = "0.7.0", path = "../embassy-executor-macros", optional = true }
+embassy-executor = { version = "0.10.0", path = "../embassy-executor", optional = true }
+embassy-executor-macros = { version = "0.8.0", path = "../embassy-executor-macros", optional = true }
 defmt = { version = "1.0.1", optional = true }
 log = { version = "0.4.14", optional = true }
 nb = "1.1.0"


embassy-rp/src/executor.rs
```diff
@@ -0,0 +1,255 @@
...
+                let irq = Irq(context as u16);
+
+                // STIR is faster, but is only available in v7 and higher.
+                #[cfg(not(armv6m))]
+                {
+                    let mut nvic: NVIC = core::mem::transmute(());
+                    nvic.request(irq);
+                }
+
+                #[cfg(armv6m)]
```
orukusaki: These cfg flags are not available in `embassy-rp`. Use `#[cfg(feature = "_rp235x")]` / `#[cfg(feature = "rp2040")]` here instead?

orukusaki:
Hi, I've tested this on an `rp2040` and got it working 🥇 
A couple of issues to resolve:
1. the cfg flags `#[cfg(armv6m)]` need changing for something else
2. I had to remove the "executor-thread" and "executor-interrupt" features from `embassy-executor`, and that resulted in `#[embassy_executor::main]` becoming unavailable.

sourcebox:
> * the cfg flags `#[cfg(armv6m)]` need changing for something else

I asked @Dirbaio at Matrix about it because I was also wondering. He said that this comes from `build.rs`. Do you have an idea to replace them?

> * I had to remove the "executor-thread" and "executor-interrupt" features from `embassy-executor`, and that resulted in `#[embassy_executor::main]` becoming unavailable.

Yes, therefore I re-exported the `main` macro for now until there's a solution available within `embassy-executor`.
This has to be done first, otherwise there's no way to access the macro.

sourcebox:
Sorry, I think you already said which flags to use, `#[cfg(feature = "_rp235x")]` / `#[cfg(feature = "rp2040")]`

orukusaki:
Aha, yes I have it working using 
```rust
#[embassy_rp::executor::main(executor="embassy_rp::executor::Executor")]
//..
}
```
Is that the intended way, or is there a tidier version?

sourcebox:
> Is that the intended way, or is there a tidier version?

You need the `executor=...` part in the macro, but you can of course simplify it with the `use` statements:

```rust
use embassy_rp::executor::{Executor, main};

#[main(executor = "Executor")]
    // 
}

sourcebox:
@orukusaki I removed the export of the main macro again. You can use it from `embassy-executor` when no arch features are selected. 

You then have to use the macro like this:

`#[main(executor = "Executor", entry = "cortex_m_rt::entry")]`

`embassy-mcxa` has an example of this.

Please check again if it works for you.

orukusaki:
Yes, I've checked and doing it that way works as expected on my rp2040 👍

PR #5411, opened 2026-02-07
DAC - Add sawtooth waveform for G4

usbalbin:
Add methods for sawtooth waveform.

~~Not yet tested on real hardware~~

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -476,7 +476,7 @@ impl<'d, M: PeriMode> DacChannel<'d, M> {
         dac.set_wave(wave);
         trigger.map(|idx| {
             dac.info.regs.cr().modify(|reg| {
-               reg.set_tsel(dac.idx, idx);
+                reg.set_tsel(dac.idx, idx);
             });
 
             // Set in case Sawtooth wave form is used
@@ -543,7 +543,7 @@ impl<'d, M: PeriMode> DacChannel<'d, M> {
     }
 
     /// Software trigger this channels sawtooth waveform step
-    /// 
+    ///
     /// NOTE: This only works with [SOFTWARE] as reset_trigger source
     #[cfg(stm32g4)]
     pub fn trigger_step(&mut self) {


embassy-stm32/src/dac/mod.rs
```diff
@@ -328,6 +330,27 @@ impl<'d, T: Instance, C: Channel, M: PeriMode> DacChannel<'d, T, C, M> {
...
     }
 
+    /// Set the sawtooth mode this channel.
+    ///
+    /// See [Self::set_sawtooth_reset_value], [Self::set_sawtooth_step_value] and [Self::set_sawtooth_step_direction]
+    /// for setting the reset value, step size and -direction.
+    ///
+    /// This method disables the channel, so you may need to re-enable afterwards.
+    #[cfg(stm32g4)]
+    pub fn set_sawtooth_mode(&mut self, reset_source: TriggerSel, step_source: StepTriggerSel) {
```
usbalbin: @xoviat I see a lot of things have changed since I did this. When rebasing this would you prefer if I turned  this into something like a `fn new_sawtooth(reset_trigger, step_trigger) -> Dac<Blocking>`?

xoviat: You can do whatever you want. The only reason this wasn't merged earlier is that it was not taken out of draft. Draft means I won't look at it, but allows you to run CI on your work.

xoviat: Actually you probably should add the step triggers to stm32-data first.

usbalbin:
~~Tested and does *not* yet work~~.

usbalbin:
Tested and seems to work

```rust
let mut ref_dac = DacChannel::new_sawtooth(
    p.DAC1,
    SOFTWARE,
    SOFTWARE,
    p.PA4,
);
ref_dac.set_sawtooth_reset_value(2048);
ref_dac.set_sawtooth_step_direction(dac::StepDirection::Decrement);
ref_dac.set_sawtooth_step_value(10);

loop {
    ref_dac.trigger();
    for _ in 0..1000 {
        embassy_time::Timer::after_millis(1).await;
        ref_dac.trigger_step();
    }
}
```

image

xoviat:
waiting on embassy-rs/stm32-data#745

xoviat:
bender run

PR #5412, opened 2026-02-08
Fix i2c repeated start on mspm0

magmastonealex:
I noticed that repeated start conditions were not being respected, which caused errors talking to devices which expect that to read registers with an offset (eeproms, sensors, etc).

Because modify() was used, and set_stop() was never set to false, once a stop bit had been requested, all transactions will send a stop bit.

This just changes the logic to properly set set_stop(). Verified with logic analyzer before/after.

Before: 

Stop bit present for repeated-start transaction

After:

Screenshot From 2026-02-07 19-59-09

There might be an edge case I'm not thinking of which explains why it was done the other way - so I'm open to being told this is wrong - but it reliably fixed sensor communication for me.

### DIFF
--- a/embassy-mspm0/src/i2c.rs
+++ b/embassy-mspm0/src/i2c.rs
@@ -474,9 +474,7 @@ impl<'d, M: Mode> I2c<'d, M> {
             w.set_cblen(length as u16);
             w.set_start(false);
             w.set_ack(send_ack_nack);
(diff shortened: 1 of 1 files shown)


i509VCB:
Weird merge queue did nothing?

i509VCB:
bender run

PR #5414, opened 2026-02-08
stm32-hrtim

usbalbin:
I have now tested the [g474/hrtim](https://github.com/embassy-rs/embassy/pull/5414/changes#diff-0f7acc969e80b1717a722f96f4447e5176dff7761e7e287b156636efd2992598) example

image

image

image

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -185,8 +185,7 @@ critical-section = "1.1"
 
 vcell = "0.1.3"
 nb = "1.0.0"
-#stm32-hrtim = { version = "0.1.0", optional = true }
-stm32-hrtim = { git = "https://github.com/usbalbin/stm32-hrtim", branch = "simplify", optional = true }
+stm32-hrtim = { version = "0.2.0", optional = true }
 stm32-fmc = "0.4.0"
 cfg-if = "1.0.0"
 embedded-io = { version = "0.7.1" }


xoviat:
There's no reason to have two hrtim mods or depend on an external crate for this. You can copy the entire crate into the hrtim mod and replace the current code with your API.

usbalbin:
The [stm32-rs/stm32g4xx-hal](https://github.com/stm32-rs/stm32g4xx-hal) crate currently depends on stm32-hrtim, so having it as a separate crate would reduce duplicated code. But I can reconsider if you feel strongly about it

usbalbin:
Also, stm32-hrtim currently uses the https://github.com/stm32-rs/stm32-rs based pacs and a quite type heavy api

xoviat:
To be frank, the intention for embassy-stm32 is to replace the stm32-rs crates outright, because we can share code across the entire stm32 family and repository with other microcontrollers. This has been done previously first when we removed our dependency on the stm32-rs pac and replaced it with our own metapac, and then when we removed our dependency on bxcan and replaced it with our own. embassy-stm32 is not (or should not be) dependent on an executor or embassy-time, which means that it can be used in other frameworks (RTIC).

However, I understand if you don't share this agenda and therefore would like to keep the crate separate for now. If that is the case, there should still be one hrtim mod that either wraps or exposes a unified api, that can be potentially upgraded or replaced when or if we choose to integrate the dependency.

usbalbin:
> [...] because we can share code across the entire stm32 family and repository with other microcontrollers

That is really amazing and can not be understated! Getting to share share efforts across families is great

> embassy-stm32 is not (or should not be) dependent on an executor or embassy-time, which means that it can be used in other frameworks (RTIC).

That is really great too

We might have a new project at work which would really benefit from some quite niche hrtim and mixed signal stuff which the stm32-rs/stm32g4xx-hal does really well. However the project would also benefit from quite a bit of the awesome async communication features which embassy-stm32 does really well. So I'd like to give embassy a go for that assuming the mixed signal features can be supported(most already are AFAICT) . But I am still quite invested in the stm32-rs stuff so if at all possible I would really appreciate if it was possible to keep it separate for now. But I am very much open to that changing in the future.

I'll try and replace the current hrtim module for now.

embassy-stm32/src/hrtim/bridge_converter.rs
```diff
@@ -0,0 +1,172 @@
...
+        frequency: Hertz,
+        prescaler: PSCL,
+        hr_control: &mut HrPwmControl,
+    ) -> Self
+    where
+        P1: Output1Pin<TIM>,
+        P2: Output2Pin<TIM>,
+        HrPwmBuilder<TIM, PSCL, PreloadSource, P1, P2>: HrPwmBuilderExt<TIM, PSCL, P1, P2>,
+        HrOut1<TIM, PSCL>: HrOutput<TIM, PSCL>,
+        HrOut2<TIM, PSCL>: HrOutput<TIM, PSCL>,
```
usbalbin: I'll try to see what I can do about this mess...

embassy-stm32/src/hrtim/bridge_converter.rs
```diff
@@ -0,0 +1,172 @@
...
+    HrTim<TIM, PSCL, capture::NoDma, capture::NoDma>: HrTimer,
+{
+    /// Create a new HRTIM bridge converter driver.
+    pub fn new<P1, P2>(
+        timer: TIM,
+        pin1: P1,
+        pin2: P2,
+        frequency: Hertz,
+        prescaler: PSCL,
+        hr_control: &mut HrPwmControl,
```
usbalbin, line 42: This is currently used as a unique token proving no one else is messing with the common registers such as the ones used to start/stop timers. Maybe we could use bit banding or something to avoid having to pass around the token/or to avoid a critical section?

embassy-stm32/src/lib.rs
```diff
@@ -101,7 +101,7 @@ pub mod flash;
 pub mod fmc;
 #[cfg(hash)]
 pub mod hash;
-#[cfg(hrtim)]
+#[cfg(all(hrtim, feature = "stm32-hrtim"))]
 pub mod hrtim;
```
usbalbin, line 105: I made this optional so as not to force the extra dependency on those that does not use the hrtim. Is that ok? If so, then how do I best enable the feature in the CI?

xoviat: `stm32-hrtim` should be a dependency on chips with hrtim.

usbalbin: Just to be clear, do we want to force everyone using an MCU with the hrtim peripheral to bring in the stm32-hrtim dependency? Or do you want to make it optional?

xoviat: > do we want to force everyone using an MCU with the hrtim peripheral to bring in the stm32-hrtim dependency

Yes

usbalbin:
I have now tested the G474 example on a Nucleo-G474RE, see screenshots in PR description. I do not have access to the F334 so I will not be able to test that.

xoviat:
Looks good. Unmark it as draft and I'll merge it.

xoviat:
Actually, can you use a released version from crates.io?

usbalbin:
> Actually, can you use a released version from crates.io?

Yes, no problem

Is it ok if I then rebase to resolve the merge conflict on the changelog?

xoviat:
Yeah, I'm not strict about commit history. I just need a merge button and a green check to merge.

usbalbin:
Still works after rebase and with released stm32-hrtim

image

xoviat:
@usbalbin can you resolve the docs build failure?

usbalbin:
Yes, sure!

PR #5420, opened 2026-02-09
[MCXA] Buffered lpuart changes

diondokter:
Small refactor from stuff I encountered during HIL test debugging

### DIFF
--- a/embassy-mcxa/src/lpuart/buffered.rs
+++ b/embassy-mcxa/src/lpuart/buffered.rs
@@ -18,8 +18,7 @@ use crate::interrupt::{self};
 
 /// State for buffered LPUART operations
 pub struct State {
-    tie_waker: AtomicWaker,
-    tcie_waker: AtomicWaker,
+    tx_waker: AtomicWaker,
     tx_buf: RingBuffer,
     rx_waker: AtomicWaker,
     rx_buf: RingBuffer,
@@ -36,8 +35,7 @@ impl State {
     /// Create a new state instance
     pub const fn new() -> Self {
         Self {
-            tie_waker: AtomicWaker::new(),
-            tcie_waker: AtomicWaker::new(),
+            tx_waker: AtomicWaker::new(),
             tx_buf: RingBuffer::new(),
             rx_waker: AtomicWaker::new(),
             rx_buf: RingBuffer::new(),
@@ -459,7 +457,7 @@ impl<'a> BufferedLpuartTx<'a> {
     pub fn write(&mut self, buf: &[u8]) -> impl Future<Output = Result<usize>> {
         // Wait for space in the buffer
         poll_fn(move |cx| {
-            self.state.tie_waker.register(cx.waker());
+            self.state.tx_waker.register(cx.waker());
 
             let mut writer = unsafe { self.state.tx_buf.writer() };
             let written = writer.push(|slice| {
@@ -487,7 +485,7 @@ impl<'a> BufferedLpuartTx<'a> {
     pub fn flush(&mut self) -> impl Future<Output = Result<()>> {
         // Wait for TX buffer to empty and transmission to complete
         poll_fn(|cx| {
-            self.state.tcie_waker.register(cx.waker());
+            self.state.tx_waker.register(cx.waker());
 
             let tx_empty = self.state.tx_buf.is_empty();
             let fifo_empty = self.info.regs().water().read().txcount() == 0;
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -20,7 +20,6 @@ use crate::interrupt::{self};
 pub struct State {
     tx_waker: AtomicWaker,
     tx_buf: RingBuffer,
-    tx_done: AtomicBool,
```
diondokter, line 23: Never used

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -456,35 +454,35 @@ impl<'a> BufferedLpuartTx<'a> {
...
-                    Poll::Pending
-                }
-            })
-            .await?;
+            let mut writer = unsafe { self.state.tx_buf.writer() };
+            let written = writer.push(|slice| {
+                let write_len = slice.len().min(buf.len());
+                slice[..write_len].copy_from_slice(&buf[..write_len]);
+                write_len
+            });
```
diondokter, line 467: We now push a whole slice, as big as possible instead of doing it byte by byte

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -695,6 +684,11 @@ impl<'a> BufferedLpuartRx<'a> {
 
 impl<'a> Drop for BufferedLpuartTx<'a> {
     fn drop(&mut self) {
+        #[cfg(feature = "defmt")]
+        if !self.state.tx_buf.is_empty() {
+            defmt::warn!("Dropping BufferedLpuartTx while TX is still active. Consider flushing first");
```
jamesmunns, line 689: Would it make sense to have/use a blocking flush method here? We probably want to fully flush or stop on drop.

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -456,35 +454,35 @@ impl<'a> BufferedLpuartTx<'a> {
...
+                    w.set_tie(true);
+                });
+            });
 
-        Ok(written)
+            if written != 0 {
+                Poll::Ready(Ok(written))
+            } else {
+                Poll::Pending
+            }
```
diondokter, line 480: Instead of waiting for all bytes to be written, we return once *some* has been written, in line with the embedded-io traits (which eventually call this function)

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -496,18 +494,9 @@ impl<'a> BufferedLpuartTx<'a> {
...
                 Poll::Ready(Ok(()))
             } else {
-                // Enable appropriate interrupt
-                cortex_m::interrupt::free(|_| {
-                    if !tx_empty {
-                        self.info.regs().ctrl().modify(|w| w.set_tie(true));
-                    } else {
-                        self.info.regs().ctrl().modify(|w| w.set_tcie(true));
-                    }
-                });
```
diondokter, line 506: The interrupts manage themselves. So no need to take a CS and enable them

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -695,6 +684,11 @@ impl<'a> BufferedLpuartRx<'a> {
 
 impl<'a> Drop for BufferedLpuartTx<'a> {
     fn drop(&mut self) {
+        #[cfg(feature = "defmt")]
+        if !self.state.tx_buf.is_empty() {
+            defmt::warn!("Dropping BufferedLpuartTx while TX is still active. Consider flushing first");
+        }
```
diondokter, line 690: The source of my HIL test woes. I didn't flush and the uart got dropped. Hard to spot, so now it has a warning.

I yearn for the time we get async drops...

embassy-mcxa/src/lpuart/buffered.rs
diondokter: Maybe? Dropping is not necessarily wrong though. It can just be surprising because you don't think about it.

Hmmm, I'll take a look at what other embassy HALs do

diondokter: Both STM and nRF just stop on drop, no flush (also no warning)

jamesmunns: @diondokter feel free to merge if you want to leave it as-is. If we're consistent with other HALs we're probably fine.

PR #5422, opened 2026-02-09
[MCXA] Erase DMA channel type by using AnyChannel

Wassasin:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/103


felipebalbi (approved):
LGTM, I'm just not sure if we should keep the dedicated `Lpuart[RT]xDma` structs. In any case, the unnecessary generic is removed.

PR #5424, opened 2026-02-09
[MCXA] ctimer: add capture driver

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/105
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/116
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/124

### DIFF
--- a/embassy-mcxa/src/ctimer/capture.rs
+++ b/embassy-mcxa/src/ctimer/capture.rs
@@ -10,7 +10,8 @@ use nxp_pac::ctimer::vals::{
     Cap0fe, Cap0i, Cap0re, Cap1fe, Cap1i, Cap1re, Cap2fe, Cap2i, Cap2re, Cap3fe, Cap3i, Cap3re,
 };
 
-use super::{AnyChannel, CTimer, CTimerChannel, Info, InputPin, Instance};
+use super::{AnyChannel, CTimer, CTimerChannel, Channel, Info, InputPin, Instance};
+use crate::clocks::WakeGuard;
 use crate::gpio::{AnyPin, SealedPin};
 use crate::inputmux::{SealedValidInputMuxConfig, ValidInputMuxConfig};
 use crate::interrupt;
@@ -18,6 +19,7 @@ use crate::interrupt::typelevel::Interrupt;
 
 /// Capture error.
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum CaptureError {
     /// Other
     Other,
@@ -45,7 +47,7 @@ pub enum Edge {
 
 /// Timestamp capture
 ///
-/// Timestamp value in microseconds.
+/// Timestamp value in ticks.
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct Timestamp(u32);
@@ -69,11 +71,13 @@ pub struct TicksDiff(pub i32);
 impl TicksDiff {
     #[inline]
     pub fn to_period(self, tick_hz: u32) -> f32 {
+        assert!(tick_hz != 0);
         self.0 as f32 / tick_hz as f32
     }
 
     #[inline]
     pub fn to_frequency(self, tick_hz: u32) -> f32 {
+        assert!(self.0 != 0);
         tick_hz as f32 / self.0 as f32
     }
 
@@ -158,6 +162,7 @@ pub struct Capture<'d> {
     ch: Peri<'d, AnyChannel>,
     pin: Peri<'d, AnyPin>,
     source_freq: u32,
+    _wg: Option<WakeGuard>,
 }
 
 impl<'d> Capture<'d> {
@@ -183,6 +188,7 @@ impl<'d> Capture<'d> {
             ch: ch.into(),
             pin: pin.into(),
             source_freq: ctimer._freq,
+            _wg: ctimer._wg.clone(),
         };
 
         inst.set_configuration(&config)?;
@@ -201,7 +207,7 @@ impl<'d> Capture<'d> {
     fn set_configuration(&mut self, config: &Config) -> Result<(), CaptureError> {
         self.info.regs().ccr().modify(|w| {
             match self.ch.number() {
-                0 => match config.edge {
+                Channel::Zero => match config.edge {
                     Edge::Both => {
                         w.set_cap0re(Cap0re::CAPORE_1);
                         w.set_cap0fe(Cap0fe::CAPOFE_1);
@@ -213,7 +219,7 @@ impl<'d> Capture<'d> {
                         w.set_cap0fe(Cap0fe::CAPOFE_1);
                     }
                 },
-                1 => match config.edge {
+                Channel::One => match config.edge {
                     Edge::Both => {
                         w.set_cap1re(Cap1re::CAP1RE_1);
                         w.set_cap1fe(Cap1fe::CAP1FE_1);
@@ -225,7 +231,7 @@ impl<'d> Capture<'d> {
                         w.set_cap1fe(Cap1fe::CAP1FE_1);
                     }
                 },
-                2 => match config.edge {
+                Channel::Two => match config.edge {
                     Edge::Both => {
                         w.set_cap2re(Cap2re::CAP2RE_1);
                         w.set_cap2fe(Cap2fe::CAP2FE_1);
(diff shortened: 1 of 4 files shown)


embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+
+use super::{AnyChannel, CTimer, CTimerChannel, Info, InputPin, Instance};
+use crate::gpio::{AnyPin, SealedPin};
+use crate::inputmux::{SealedValidInputMuxConfig, ValidInputMuxConfig};
+use crate::interrupt;
+use crate::interrupt::typelevel::Interrupt;
+
+/// Capture error.
+#[derive(Debug)]
+pub enum CaptureError {
```
jamesmunns, line 23: non-exhaustive?

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// Falling edge
+    FallingEdge,
+    /// Both edges
+    #[default]
+    Both,
+}
+
+/// Timestamp capture
+///
+/// Timestamp value in microseconds.
```
jamesmunns: is this *ticks* or *microseconds*?

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+    fn with_ticks(self, ticks: u32) -> Self {
+        Self(ticks)
+    }
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, PartialOrd, Ord, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct TicksDiff(pub i32);
+
+impl TicksDiff {
```
jamesmunns, line 71: div by 0 checks?

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+                        w.set_cap3fe(Cap3fe::CAP3FE_1);
+                    }
+                    Edge::RisingEdge => {
+                        w.set_cap3re(Cap3re::CAP3RE_1);
+                    }
+                    Edge::FallingEdge => {
+                        w.set_cap3fe(Cap3fe::CAP3FE_1);
+                    }
+                },
+                _ => unreachable!(),
```
jamesmunns: Should we have a channel enum or something so we don't need an unreachable here?

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+        _irq: impl crate::interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
+        config: Config,
+    ) -> Result<Self, CaptureError>
+    where
+        (T, CH, PIN): ValidInputMuxConfig,
+    {
+        pin.mux();
+        <(T, CH, PIN) as SealedValidInputMuxConfig>::mux();
+
+        let mut inst = Self {
```
jamesmunns, line 186: We don't hold the ctimer, wouldn't this run the Drop impl (if any), and potentially drop WakeGuards? IMO we should either whole the whole ctimer, or do a manual destructure here to ensure we take all the important parts

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -0,0 +1,351 @@
...
+                T::info()
+                    .timestamp(2)
+                    .store(T::info().regs().cr(2).read().cap(), Ordering::Release);
+            }
+
+            if ir.cr3int() {
+                w.set_cap3i(Cap3i::CAP3I_0);
+                T::info()
+                    .timestamp(3)
+                    .store(T::info().regs().cr(3).read().cap(), Ordering::Release);
```
jamesmunns, line 350: Seems unlikely, but IF we had the event trigger immediately when `.cap()` still reads `0`, wouldn't we hang forever? Do we want to always write *at least* 1 to ensure this is not the case?

embassy-mcxa/src/ctimer/mod.rs
```diff
@@ -131,6 +125,12 @@ macro_rules! impl_instance {
                 static INFO: Info = Info {
                     regs: pac::$peri,
                     wait_cell: WaitCell::new(),
+                    timestamp: [
```
jamesmunns: Could do `[const { AtomicU32::new(0) }; 4]`, and/or have a const for "number of channels"

embassy-mcxa/src/inputmux.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::clocks::{disable, enable, enable_and_reset};
+use crate::pac;
+use crate::peripherals::INPUTMUX0;
+
+pub(crate) trait SealedValidInputMuxConfig {
+    fn mux() {}
+}
+
+#[allow(private_bounds)]
+
```
jamesmunns:
```suggestion
```

felipebalbi:
> LGTM! Maybe CC @diondokter to see if we need to update the HIL wiring diagram to make sure we have a CTimer PWM/Capture pair available for testing

I think it's best to use a raw GPIO and route it to a Capture. It's much easier to control what we're expecting from the capture, but I hear you.

embassy-mcxa/src/ctimer/mod.rs
felipebalbi: nice, done both :)

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: done

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: ugh, updated this comment.

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: added `assert!()`

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: cloned the wakeguard

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: We can, yes.  Done for both capture and pwm

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: That will never happen. We enable the counter from `new()` and only much later do we `await` for a capture. Several instructions are guaranteed to run between those two times. I guess the only real possibility would involve an extremely slow external clock, which we don't (yet?) support with the capture driver. I'll add a comment explaining this.

embassy-mcxa/src/inputmux.rs
felipebalbi: fixed. Also added missing rustdoc comment.

felipebalbi:
> > LGTM! Maybe CC @diondokter to see if we need to update the HIL wiring diagram to make sure we have a CTimer PWM/Capture pair available for testing
> 
> I think it's best to use a raw GPIO and route it to a Capture. It's much easier to control what we're expecting from the capture, but I hear you.

I verified @diondokter's wiring. `P2_4` can be used as a capture with `P1_8` being an output GPIO.

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -308,6 +312,15 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
         let ir = T::info().regs().ir().read();
         T::info().regs().ir().write(|w| w.0 = ir.0);
 
+        // If the interrupt triggered, we want to know which channel
```
jamesmunns, line 315: I would rather panic loudly than leave the potential for the system to permanently hang if the *very unlikely* thing does happen.

jamesmunns: An alternative would be to reduce the max range to 31 bits, and use the top bit (or just 0xFFFF_FFFF) as a sentinel flag for "actually zero".

felipebalbi: I'll send another PR panicking if timestamp == 0

PR #5425, opened 2026-02-09
[MCXA] ctimer: capture: Avoid situation where timestamp can be zero

felipebalbi:
Keep track of interrupt triggered flags, instead of waiting for a non-zero timestamp.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/131

### DIFF
--- a/embassy-mcxa/src/ctimer/capture.rs
+++ b/embassy-mcxa/src/ctimer/capture.rs
@@ -284,20 +284,14 @@ impl<'d> Capture<'d> {
                     }
                 });
 
-                let flags = self.info.irq_flags().load(Ordering::Acquire);
                 let n: usize = self.ch.number().into();
-                (flags & (1 << n)) != 0
+                let mask = 1 << n;
+                (self.info.irq_flags().fetch_and(!mask, Ordering::AcqRel)) != 0
             })
             .await
             .map_err(|_| CaptureError::Other)?;
 
         let timestamp = self.info.regs().cr(self.ch.number().into()).read().cap();
-
-        let mut flags = self.info.irq_flags().load(Ordering::Acquire);
-        let n: usize = self.ch.number().into();
-        flags &= !(1 << n);
-        self.info.irq_flags().store(flags, Ordering::Release);
-
         Ok(Timestamp(timestamp))
     }
 }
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -284,12 +284,20 @@ impl<'d> Capture<'d> {
                     }
                 });
 
-                self.info.timestamp(self.ch.number().into()).load(Ordering::Acquire) != 0
+                let flags = self.info.irq_flags().load(Ordering::Acquire);
+                let n: usize = self.ch.number().into();
+                (flags & (1 << n)) != 0
```
jamesmunns: you probably want to use something like:

```rust
let mask = 1 << n;
(self.info.irq_flags().fetch_and(!mask, Ordering::AcqRel) & mask) != 0
```

This way you retrieve and clear the flag in a single CAS operation, and don't need the additinal load/stores below either.

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -312,45 +320,32 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
...
             if ir.cr3int() {
                 w.set_cap3i(Cap3i::CAP3I_0);
-                T::info()
-                    .timestamp(3)
-                    .store(T::info().regs().cr(3).read().cap(), Ordering::Release);
+                flags |= 1 << 3;
             }
         });
 
+        T::info().irq_flags().store(flags, Ordering::Release);
```
jamesmunns: you probably want `fetch_or` here or in each `if` arm. You can use `Acquire` here instead of `AcqRel` because you don't care about the returned (old) value

felipebalbi: done

embassy-mcxa/src/ctimer/capture.rs
felipebalbi: done

embassy-mcxa/src/ctimer/capture.rs
```diff
@@ -312,44 +314,29 @@ impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandl
...
 
             if ir.cr3int() {
                 w.set_cap3i(Cap3i::CAP3I_0);
-                T::info()
-                    .timestamp(3)
-                    .store(T::info().regs().cr(3).read().cap(), Ordering::Release);
+                mask |= 1 << 3;
             }
         });
+        T::info().irq_flags().fetch_or(mask, Ordering::Acquire);
```
jamesmunns:
```suggestion
        T::info().irq_flags().fetch_or(mask, Ordering::Release);
```

Sorry, I was wrong here, we want the store to be Release, but we don't care about the load (so allow it to be Relaxed)

felipebalbi: done :)

Issue #4614, opened 2025-08-30
stm32n6 support


JRazek:
So, embassy-stm32 is missing the whole family stm32n6. Are there plans on adding it soon?

Dirbaio:
There was some work in stm32-data here, for the rcc registers: https://github.com/embassy-rs/stm32-data/pull/612

It needs:
- in stm32-data:
  - Add registers for other core blocks (syscfg, exti, flash, maybe others)
- in embassy-stm32:
  - Add features to Cargo.toml
  - Add RCC implementation
  - Fix whatever build issues show up

There's no one working on it AFAIK. PRs welcome.

ionspin:
I'll give it a look over the upcoming weekend, I've got a stm32n6 devkit lying around.

ionspin:
Hey @Dirbaio since Matrix is down at the moment, I'll ask here. This is my first time working with stm32 chips so I might be asking dumb questions. I'm looking at metadata generated for stm32n657x0 (metadata_0774.rs) and I noticed `rcc` is set to `None` for `DCMI` peripheral, which is not correct. 

Could you give me a pointer to what actually generates the `rcc`  field for peripherals, and why it would be working for `CSI` and `DCMIPP` and not `DCMI`?

Also is there some additional documentation describing the process of adding a new family from start to finish? I'm not quite clear on which parts are generated from headers and svd files and which parts need to manually created.

Thanks!

Dirbaio:
> I noticed rcc is set to None for DCMI peripheral, which is not correct. Could you give me a pointer to what actually generates the rcc field for peripherals, and why it would be working for CSI and DCMIPP and not DCMI?

It searches the rcc for fields named DCMIEN, DCMIRST. The issue is most likely the fields are named something different because ST is often inconsistent, the fix is renaming them.

relevant code is this https://github.com/embassy-rs/stm32-data/blob/main/stm32-data-gen/src/rcc.rs

> Also is there some additional documentation describing the process of adding a new family from start to finish? I'm not quite clear on which parts are generated from headers and svd files and which parts need to manually created.

There's nothing special that needs to be done in stm32-data for a new familiy. The process I follow is basically trying to build embassy-stm32 then fix compile errors. If it complains registers for some peripheral are missing, add them in stm32-data following the "add a new peripheral" instructions in the readme.

jobobob:
Is there any update on this?
@ionspin are you still planning to work on it?
I would be happy to test something, I have a board around.

[...]

PR #5427, opened 2026-02-09
STM32N6: Add the UART peripherals and implement the clock mux

lsartory:
Add the UART, USART and LPUART peripherals for the STM32N6.
The clock multiplexer is now also implemented, where the IC settings are separate from the peripherals. This change was required because some peripherals share the same IC.

Drafted until stm32-data PR [#729](https://github.com/embassy-rs/stm32-data/pull/729) is merged.
Related to issue #4614.


lsartory:
Thanks!

PR #5433, opened 2026-02-10
[MCXA]: Add optional gpio-based debugging for the custom executor

jamesmunns:
This adds a GPIO-based debugging interface for the executor.

This allows us to see when the executor is deciding to WFE, and when it has control
flow returned to it after sleeping. This mimics the C SDK examples. We can make this
a feature in the future if we want to gate it out when unused in production.

Examples:

## Waking from deep sleep

Screenshot 2026-02-10 at 5 04 02 PM

## Going to deep sleep

Screenshot 2026-02-10 at 5 05 07 PM

### DIFF
--- a/embassy-mcxa/src/executor.rs
+++ b/embassy-mcxa/src/executor.rs
@@ -1,12 +1,31 @@
 use core::marker::PhantomData;
-use core::sync::atomic::{AtomicBool, AtomicU8, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU8, AtomicU32, Ordering};
 
 use embassy_executor::{Spawner, raw};
+use embassy_hal_internal::Peri;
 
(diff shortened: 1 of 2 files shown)


jamesmunns:
We do have a sort of interesting double-dip both coming in and out of sleep, it's likely that the `__pender` is setting SEV more often than it should be. I'll look at that in a follow-up PR.

bugadani:
Why don't you use rtos-tracing events for this?

jamesmunns:
@bugadani this is specifically for debugging our custom executor, we want to see when the executor is deciding to WFE, rather than when polling is occurring or not. I don't *think* we can see exactly what we want here using rtos-trace, since we're instrumenting our executor directly. @ me on chat if you think I'm wrong, I could be!

PR #5437, opened 2026-02-10
stm32: Fix docs

usbalbin:


usbalbin:
@xoviat when I run `cargo doc --features stm32g474re,dual-bank`, I still get

```
warning: unresolved link to `Can::enable_non_blocking`
   --> src/can/fdcan.rs:183:24
    |
183 |     /// You must call [FdCan::enable_non_blocking] to use the peripheral.
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^ the struct `Can` has no field or associated item named `enable_non_blocking`
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `Mode::ZeroOverhead`
   --> src/cordic/mod.rs:200:42
    |
200 |     /// you may want to set Cordic into [Mode::ZeroOverhead] mode, and add extra arguments with [Self::extra_config]
    |                                          ^^^^^^^^^^^^^^^^^^ no item named `Mode` in scope

warning: `embassy-stm32` (lib doc) generated 2 warnings
```

`Mode::ZeroOverhead` was as far as I understand removed in https://github.com/embassy-rs/embassy/commit/c9f759bb21782eb0487c96a59500310d1283694c so that whole line should probably be updated/removed?

The other one is less clear, other than `FdCan` being renamed to `Can` I can not find what `enable_non_blocking` is supposed to be other than a method on a private trait in [bxcan](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/can/bxcan/registers.rs#L99)... So maybe a copy paste error from [here](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/can/bxcan/mod.rs#L183)?

Edit (sorry saw your message in the chat just now)

usbalbin:
Anyways, here are some less warnings for what it's worth :)

PR #5438, opened 2026-02-10
[MCXA] Add SPI controller driver

felipebalbi:
This PR is a refactored/cleaned-up version of #5175. We have removed support for DMA for now, also removed target support. Those two features will be added in subsequent PRs, but this lets us merge the minimal support as is.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/59
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/60

### DIFF
--- a/embassy-mcxa/src/spi/controller.rs
+++ b/embassy-mcxa/src/spi/controller.rs
@@ -1,0 +1,772 @@
+//! LPSPI Controller Driver.
+
+use core::marker::PhantomData;
+
+use embassy_embedded_hal::SetConfig;
+use embassy_hal_internal::Peri;
+pub use embedded_hal_1::spi::{MODE_0, MODE_1, MODE_2, MODE_3, Mode, Phase, Polarity};
+use nxp_pac::lpspi::vals::{Cpha, Cpol, Lsbf, Master, Mbf, Pcspol, Pincfg, Prescale, Rrf, Rtf, Rxmsk, Txmsk};
+
+use super::{Async, Blocking, Info, Instance, MisoPin, Mode as IoMode, MosiPin, SckPin};
+use crate::clocks::periph_helpers::{Div4, LpspiClockSel, LpspiConfig};
+use crate::clocks::{ClockError, PoweredClock, WakeGuard, enable_and_reset};
+use crate::gpio::AnyPin;
+use crate::interrupt;
+use crate::interrupt::typelevel::Interrupt;
+
+// LPSPI has a 4-word FIFO.
+const LPSPI_FIFO_SIZE: u8 = 4;
+
+/// Errors exclusive to HW initialization
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SetupError {
+    /// Clock configuration error.
+    ClockSetup(ClockError),
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
+/// I/O Errors
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum IoError {
+    /// Receive error.
+    ///
+    /// Indicates FIFO overflow condition has happened.
+    ReceiveError,
+    /// Transmit error.
+    ///
+    /// Indicated FIFO underrun condition has happened.
+    TransmitError,
+    /// Other internal errors or unexpected state.
+    Other,
+}
+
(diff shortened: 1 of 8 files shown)


felipebalbi:
PS: this needs some work on the error types. I'll get that done tomorrow. Currently, there's an assumption that we're always correctly using the thing and we never fall into transfer error 🤣

embassy-mcxa/src/spi/controller.rs
```diff
@@ -0,0 +1,717 @@
...
+    /// Bit order
+    pub bit_order: BitOrder,
+    /// Clock configuration
+    pub clock_config: ClockConfig,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            frequency: 1_000,
```
Wassasin: Default frequency of 1kHz seems a bit low.

embassy-mcxa/src/spi/controller.rs
```diff
@@ -0,0 +1,717 @@
...
+        });
+
+        for word in data {
+            // Wait until we have data in the RxFIFO.
+            while self.info.regs().fsr().read().rxcount() == 0 {}
+
+            *word = self.info.regs().rdr().read().data() as u8;
+        }
+
+        self.blocking_flush()
```
Wassasin: Flush is superfluous when reading.

embassy-mcxa/src/spi/controller.rs
```diff
@@ -0,0 +1,717 @@
...
+                })
+                .await
+                .map_err(|_| IoError::Other)?;
+
+            // dummy data
+            self.info.regs().tdr().write(|w| w.set_data(0));
+            *word = self.info.regs().rdr().read().data() as u8;
+        }
+
+        self.async_flush().await
```
Wassasin: Flush should not be needed to read.

embassy-mcxa/src/spi/controller.rs
felipebalbi: Fixed

embassy-mcxa/src/spi/controller.rs
felipebalbi: Fixed

embassy-mcxa/src/spi/controller.rs
felipebalbi: Changed to 1MHz

embassy-mcxa/src/spi/controller.rs
```diff
@@ -0,0 +1,770 @@
...
+        Ok(())
+    }
+}
+
+impl<'d, M: IoMode> Spi<'d, M> {
+    fn check_status(&mut self) -> Result<(), IoError> {
+        let status = self.info.regs().sr().read();
+
+        if status.ref_() {
+            self.info.regs().sr().write(|w| w.set_ref_(true));
```
Wassasin, line 232: The manual states that the transfer needs to be terminated and the FIFO queue should be emptied before clearing this flag. I am unsure whether we need to comply in this case; whether leaving it intact is actually an issue. Something to consider.

embassy-mcxa/src/spi/controller.rs
```diff
@@ -0,0 +1,770 @@
...
+            Err(IoError::ReceiveError)
+        } else if status.tef() {
+            self.info.regs().sr().write(|w| w.set_tef(true));
```
Wassasin, line 235: Transfer should also be stopped before clearing this flag.

embassy-mcxa/src/spi/controller.rs
Wassasin: Image

embassy-mcxa/src/spi/controller.rs
Wassasin: In this case I am not sure what "ending the transfer" actually entails as the FIFO queue is empty. So my bet is you can disregard this case.

embassy-mcxa/src/spi/controller.rs
Wassasin: Calling 'RX FIFO reset' meaning `cr.modify(|w| w.set_rrf(true))` here should be sufficient.

However we are the controller in this case, it should be impossible to overflow here.

felipebalbi: Added an Rx FIFO reset. There's no explicit "stop transfer" on SPI. That part is more of a concern for when we add DMA support, I suppose.

embassy-mcxa/src/spi/controller.rs
felipebalbi: Yeah, see previous comment. Additionally, we're transferring one byte at a time; when we get here, transfer is already "stopped".

PR #5441, opened 2026-02-11
fix(stm32/can): drain all rx frames per interrupt in buffered modes

kvogt:
The buffered CAN rx interrupt handler was reading only one frame per interrupt and clearing the interrupt flag before draining the FIFO. On a heavily loaded bus, there can be other frames in the hardware FIFO that get stuck until the next frame arrived to trigger a new interrupt.

Fix: drain all pending frames from the FIFO before clearing the interrupt flag. This ensures no frames are left behind regardless of how many frames arrive between interrupts.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -650,8 +650,14 @@ impl RxMode {
                 T::registers().regs.ir().write(|w| w.set_rfn(fifonr, true));
                 loop {
                     match self.try_read::<T>(ns_per_timer_tick) {
-                        Some(result) => {
-                            let _ = buf.rx_sender.try_send(result);
+                        Some(Ok(envelope)) => {
+                            let _ = buf.rx_sender.try_send(Ok(envelope));
+                        }
+                        Some(Err(err)) => {
+                            // bus error states can persist; emit once and return to avoid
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/can/fdcan.rs
```diff
@@ -641,20 +641,32 @@ impl RxMode {
...
-                    let _ = buf.rx_sender.try_send(result);
+                loop {
+                    match self.try_read::<T>(ns_per_timer_tick) {
+                        Some(result) => {
+                            let _ = buf.rx_sender.try_send(result);
+                        }
+                        None => break,
+                    }
                 }
+                T::registers().regs.ir().write(|w| w.set_rfn(fifonr, true));
```
xoviat: The issue with this is that it clears interrupts after reading the frames. Thus, if a frame is received between reading the last frame and clearing the interrupt, a new interrupt will not be generated. Therefore, the interrupt should be cleared first and then all frames should be read.

kvogt: Good catch. I have fixed this [here](https://github.com/embassy-rs/embassy/pull/5441/changes/89ced88427dac556199365174287d9bb27818b83).

xoviat (approved):

PR #5444, opened 2026-02-11
Add development environment management with Devenv

hatimthayyil:
[Devenv](https://devenv.sh/) is a Nix-based environment manager. It provides a relatively simple way to configure the development environment. To bring in the Rust toolchain, you would do `languages.rust.enable = true;` It sets up the correct environment variables behind the scenes (RUST_SRC etc). It also allows to bring in packages from the Nixpkgs repository.

- Add Nix/Devenv files for the project reproducible development environemnt.
- Add flip-link to binary
- Add treefmt for nixfmt and rustfmt

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+export DIRENV_WARN_TIMEOUT=20s
+
+eval "$(devenv direnvrc)"
+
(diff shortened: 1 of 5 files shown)


lulf (commented):
This feels like having a too narrow user base to warrant putting it in the repo IMHO

PR #5447, opened 2026-02-11
Embassy-rp: Support custom DMA transfers

MTelthoe:
Support custom DMA transfers, by configuring the DMA registers directly.

### DIFF
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -69,6 +69,13 @@ impl<'d> Channel<'d> {
     }
 
     /// Get the channel register block.
+    #[cfg(feature = "unstable-pac")]
+    pub fn regs(&self) -> pac::dma::Channel {
+        pac::DMA.ch(self.number as _)
+    }
+
+    /// Get the channel register block.
+    #[cfg(not(feature = "unstable-pac"))]
     fn regs(&self) -> pac::dma::Channel {
         pac::DMA.ch(self.number as _)
     }


embassy-rp/src/dma.rs
```diff
@@ -64,12 +64,12 @@ impl<'d> Channel<'d> {
     }
 
     /// Get the channel number.
-    fn number(&self) -> u8 {
+    pub fn number(&self) -> u8 {
```
dobrowolski-lukasz: This is already being addressed in https://github.com/embassy-rs/embassy/pull/5388

embassy-rp/src/dma.rs
```diff
@@ -64,12 +64,12 @@ impl<'d> Channel<'d> {
...
         self.number
     }
 
     /// Get the channel register block.
-    fn regs(&self) -> pac::dma::Channel {
+    pub fn regs(&self) -> pac::dma::Channel {
```
dobrowolski-lukasz, line 73: I think this should not be public unless `unstable-pac` feature is enabled for the `embassy-rp` as the pac can change major versions with minor version change of `embassy-rp`.

Do note that if you use `unstable-pac` the pac becomes available under `embassy-rp::pac` so it's at best a convinience function. I still think it would be a nice to have.

embassy-rp/src/dma.rs
```diff
@@ -250,7 +250,7 @@ pub struct Transfer<'a> {
 }
 
 impl<'a> Transfer<'a> {
-    fn new(channel: Channel<'a>) -> Self {
+    pub fn new(channel: Channel<'a>) -> Self {
```
dobrowolski-lukasz: I don't think Transwer type is that useful when not using the library transfers. What kind of use case do you have?

embassy-rp/src/dma.rs
MTelthoe: You are right, I will add this for convince.

embassy-rp/src/dma.rs
MTelthoe: I want to use one DMA channel to read from flash and another DMA channel that transfers this data into the sha256 accelerator. Because the data I want to get a hash for is too large to store in memory I am chaining the first and second DMA channel in a "ping pong" style.

I am currently using the stream ctrl register and the fifo empty flag to determine, whether all data has been read from the flash, but the issue is, that there might be data, that still needs to be written to the hash accelerator. Because the DMA channels do trigger each other in an endless cycle(chaining), I can not use the busy flag to check whether everything has been transfered or not. But the DMA Channels will trigger an interrupt after each completed transfer sequence. I want to use that interrupt to determine, if the data has been transferred completely.

I could check the interrupt register in a busy loop, but I think using the async Transfer for that would be better.

The changes of #5388 would be pretty useful for this as well.

dobrowolski-lukasz: What You wrote in your comment is the reason why I think `Transfer` is not useful when doing custom DMA handling and hence there is no sense in adding `pub` to `Transfer::new` function.

In my use case I've implemented custom type that I use instead of `Transfer` (It's even named the same). I also can't use busy bit to check if the future is done.

What I'm trying to say is that this change not useful, can mislead users(if it's pub then it should be useful somehow) and hence should be removed.

embassy-rp/src/dma.rs
dobrowolski-lukasz: I think the feature gate You've added works on the whole function not just the `pub` keyword.
You have to check how to put the feature gate in a way that the function is still available for private code when `unstable-pac` feature is disabled. Or figure out something equivalent.

MTelthoe: The function is available without the `pub` keyword, if the `unstable-pac` feature is disabled. That is why I added the same function without the `pub` keyword with the inverted feature gate.
This is the only reason to achieve the change in visibility with features, that I know of.

embassy-rp/src/dma.rs
MTelthoe: I agree, I added a custom type in my application code.

embassy-rp/src/dma.rs
dobrowolski-lukasz: Ah, I've missed the other function. Sorry.
I've tried to find some solution but quick search didn't give me any usable answer. :/

dobrowolski-lukasz:
Only one change is remaining in this PR.
Please squash your commits into a single one to improve readability.

PR #5451, opened 2026-02-11
stm32f1: Add config option to remap JTAG pins

GrantM11235:
I am open to suggestions for naming and documentation.

## Alternatives

Use cargo features to configure this. That would also allow us to remove the unusable pin singletons from the Peripherals struct.

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -338,7 +338,7 @@ pub struct Config {
 
     /// Allows JTAG pins to be used for GPIO
     #[cfg(stm32f1)]
-    pub swj_cfg: gpio::SwjCfg,
+    pub swj: gpio::SwjCfg,
 }
 
 impl Default for Config {
@@ -368,7 +368,7 @@ impl Default for Config {
             #[cfg(peri_ucpd2)]
             enable_ucpd2_dead_battery: false,
             #[cfg(stm32f1)]
-            swj_cfg: Default::default(),
+            swj: Default::default(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


GrantM11235:
Another alternative: replace the pin singletons in the Peripherals struct with `Debugger<PA15>` etc, and offer functions like `fn disable_jtag(pa15: Debugger<PA15>, pb3: Debugger<PB3>, pb4: Debugger<PB4>) -> (PA15, PB3, PB4)` to change the mapping at runtime. That's what [stm32f1xx-hal did](https://docs.rs/stm32f1xx-hal/latest/stm32f1xx_hal/afio/struct.MAPR.html#method.disable_jtag)

embassy-stm32/src/gpio.rs
```diff
@@ -937,6 +937,47 @@ pub(crate) unsafe fn init(_cs: CriticalSection) {
...
+#[cfg(stm32f1)]
+/// SWJ Config
+#[derive(Clone, Copy, Debug, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum SwjCfg {
+    /// Full SWJ (JTAG-DP + SW-DP) (Reset state)
+    ///
+    /// PA13, PA14, PA15, PB3, and PB4 cannot be used
+    #[default]
+    Reset = 0x0,
```
Dirbaio: naming of the variants could be improved.  Currently you have to read the docs to tell what each does. 

SwdAndJtag
SwdAndJtagNoRst
SwdOnly
Disabled

or something like that.

Dirbaio:
> replace the pin singletons in the Peripherals struct with Debugger<PA15> etc, and offer functions like fn disable_jtag(pa15: Debugger<PA15>, pb3: Debugger<PB3>, pb4: Debugger<PB4>) -> (PA15, PB3, PB4)

we don't do typestates here :D

ChocolateLoverRaj:
> we don't do typestates here :D

why not?

PR #5455, opened 2026-02-12
[MCXA] i2c: controller: add DMA support

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/46

### DIFF
--- a/embassy-mcxa/src/i2c/controller.rs
+++ b/embassy-mcxa/src/i2c/controller.rs
@@ -763,7 +763,7 @@ impl<'d> I2c<'d, Dma<'d>> {
             scl,
             sda,
             config,
-            super::Dma {
+            Dma {
                 tx_dma,
                 rx_dma,
                 tx_request_number,


felipebalbi:
Wondering what folks (@Wassasin , @diondokter , @jamesmunns) think about taking the DMA request number as an argument instead of a generic when configuring the channels.

We're trying to make sure we don't need to keep the `Instance` generic around, which means the request number needs to be cached from the constructor, when we _do_ have `Instance` available.

Additionally, @jamesmunns suggested that we try to make `Dma` a typestate in the `I2c` struct instead of creating a new `I2cDma` driver. Here's a first cut, let me know what you guys think. If you think it's okay, I can patch the UART driver accordingly.

jamesmunns:
@felipebalbi opened https://github.com/felipebalbi/embassy/pull/5 for some suggestions on using a little bit of type state to make some parts of this nicer :)

felipebalbi:
> @felipebalbi opened [felipebalbi#5](https://github.com/felipebalbi/embassy/pull/5) for some suggestions on using a little bit of type state to make some parts of this nicer :)

nice, no more `unwrap()` for the DMA channels.

PR #5459, opened 2026-02-13
[MCXA] spi: add DMA support

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/61

### DIFF
--- a/embassy-mcxa/src/spi/controller.rs
+++ b/embassy-mcxa/src/spi/controller.rs
@@ -1,6 +1,7 @@
 //! LPSPI Controller Driver.
 
 use core::marker::PhantomData;
+use core::sync::atomic::{Ordering, fence};
 
 use embassy_embedded_hal::SetConfig;
 use embassy_futures::join::join;
@@ -574,8 +575,6 @@ impl<'d> Spi<'d, Dma<'d>> {
         })
         .await;
 
-        // Ensure DMA writes are visible to CPU
-        cortex_m::asm::dsb();
         // Cleanup
         self.info.regs().der().modify(|w| {
             w.set_rdde(false);
@@ -588,6 +587,11 @@ impl<'d> Spi<'d, Dma<'d>> {
             self.mode.tx_dma.disable_request();
             self.mode.tx_dma.clear_done();
         }
+
+        // Ensure all writes by DMA are visible to the CPU
+        // TODO: ensure this is done internal to the DMA methods so individual drivers
+        // don't need to handle this?
+        fence(Ordering::Acquire);
 
         Ok(())
     }
@@ -609,6 +613,11 @@ impl<'d> Spi<'d, Dma<'d>> {
                 .tx_dma
                 .setup_write_to_peripheral(data, peri_addr, EnableInterrupt::Yes);
 
+            // Ensure all writes by CPU are visible to the DMA
+            // TODO: ensure this is done internal to the DMA methods so individual drivers
+            // don't need to handle this?
+            fence(Ordering::Release);
+
             // Enable SPI TX DMA request
             self.info.regs().der().modify(|w| w.set_tdde(true));
 
@@ -627,8 +636,6 @@ impl<'d> Spi<'d, Dma<'d>> {
         })
         .await;
 
-        // Ensure DMA writes are visible to CPU
-        cortex_m::asm::dsb();
         // Cleanup
         self.info.regs().der().modify(|w| w.set_tdde(false));
         unsafe {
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+
+        Ok(())
+    }
+}
+
+trait AsyncEngine {
+    fn async_read_internal(&mut self, data: &mut [u8]) -> impl Future<Output = Result<(), IoError>>;
+    fn async_write_internal(&mut self, data: &[u8]) -> impl Future<Output = Result<(), IoError>>;
+    fn async_transfer_internal(&mut self, read: &mut [u8], write: &[u8]) -> impl Future<Output = Result<(), IoError>>;
+    fn async_transfer_in_place_internal(&mut self, data: &mut [u8]) -> impl Future<Output = Result<(), IoError>>;
```
diondokter: Just FYI, no need to change, but you can leave these functions as `async fn` just fine and have the implementations in this form. That'll work just fine but reads a bit nicer

embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+impl<'d> Spi<'d, Dma<'d>> {
+    /// Create a SPI driver in async mode.
+    pub fn new_async_with_dma<T: Instance>(
+        _peri: Peri<'d, T>,
+        sck: Peri<'d, impl SckPin<T> + 'd>,
+        mosi: Peri<'d, impl MosiPin<T> + 'd>,
+        miso: Peri<'d, impl MisoPin<T> + 'd>,
+        tx_dma: Peri<'d, impl TxDma<T>>,
+        rx_dma: Peri<'d, impl RxDma<T>>,
+        _irq: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
```
diondokter, line 490: Do we really need interrupts for the DMA version? We seem to piggyback on the DMA interrupts.

The only thing I've found using this is the async_flush function. But is that function even relevant when using DMA? If not, we can move the function to the async-only impl block and remove/disable interrupts for the DMA version. Saves a teeny tiny bit of perf and it's less setup.

embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+            if self.mode.rx_dma.is_done() {
+                core::task::Poll::Ready(())
+            } else {
+                core::task::Poll::Pending
+            }
+        })
+        .await;
+
+        // Ensure DMA writes are visible to CPU
+        cortex_m::asm::dsb();
```
diondokter: This is good, but we need the compiler to not reorder stuff too. It's unlikely with all the volatile operations around it, but better be safe.

We can add a compiler_fence, but even nicer is just using a normal fence which also a barrier instruction (DMB instead of DSB, but that should still be strong enough I think).

Use `Acquire` because we, the cpu, need to acquire the memory the DMA has written into `data`
```suggestion
        fence(Ordering::Acquire);
```

embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+            if self.mode.tx_dma.is_done() {
+                core::task::Poll::Ready(())
+            } else {
+                core::task::Poll::Pending
+            }
+        })
+        .await;
+
+        // Ensure DMA writes are visible to CPU
+        cortex_m::asm::dsb();
```
diondokter: Similar to above, we want to use a fence, but this time with the `Release` ordering to make our `data` available to the DMA peripheral.

However, we need to do that *before* the DMA is started. This current line is after the DMA is completed. So this fence can be put just before `Enable SPI TX DMA request` maybe?

embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+                core::task::Poll::Ready(())
+            } else {
+                core::task::Poll::Pending
+            }
+        });
+
+        join(tx_transfer, rx_transfer).await;
 
+        // Ensure DMA writes are visible to CPU
+        cortex_m::asm::dsb();
```
diondokter: `Acquire` fence here too and also a `Release` fence just before the DMA starts (since we're doing a transfer in both directions)

embassy-mcxa/src/spi/controller.rs
```diff
@@ -467,11 +473,412 @@ impl<'d> Spi<'d, Async> {
...
+                core::task::Poll::Ready(())
+            } else {
+                core::task::Poll::Pending
+            }
+        });
+
+        join(tx_transfer, rx_transfer).await;
+
+        // Ensure DMA writes are visible to CPU
+        cortex_m::asm::dsb();
```
diondokter: You know the drill. Fence before and after

embassy-mcxa/src/spi/controller.rs
jamesmunns: I think we do want it still, the DMA transfer will complete when all bytes are put into the FIFO, we will need to wait for all bytes to be flushed out "on the wire". We could maybe have a simpler interrupt handler that only does this, but for now I think it's reasonable to remain for consistency.

PR #5465, opened 2026-02-16
docs: mention which crate has cargo size and nm

tachiniererin:
There's a cargo-size crate which is extremely unhelpful when trying to get the task done that cargo size from cargo-binutils does.


diondokter (approved):

diondokter:
Thanks!

PR #5467, opened 2026-02-16
[MCXA] Rework Lpuart such that it is more consistent wrt Embassy

Wassasin:


embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -15,29 +15,18 @@ use crate::pac::lpuart::vals::{
...
-// ============================================================================
-
-use crate::dma::{
-    Channel as DmaChannelTrait, DMA_MAX_TRANSFER_SIZE, DmaChannel, DmaRequest, EnableInterrupt, RingBuffer,
-};
-
-// ============================================================================
-// MISC
-// ============================================================================
+pub use blocking::*;
```
jamesmunns: I have a weak preference to not use `*` imports, they can cause odd semver problems especially around shadowing.

embassy-mcxa/src/lpuart/mod.rs
jamesmunns, line 1: It would be nice to have some module level docs that explain how the type states are used, and why the ring buffer dma variant *isn't* a type state.

Wassasin: Lets do this in a separate PR.

Issue #3590, opened 2024-11-28
STM32 ring-buffer


elagil:
This is a follow-up to the discussion in https://github.com/embassy-rs/embassy/pull/3336 by @liarokapisv

Feel free to add further people to the discussion.

Remaining issues/ideas with the current ring-buffer implementation include
- `clear()` on overrun causes issues with higher-level drivers that handle data in groups of samples. For example, a `SAI` driver may handle data in groups of two samples (left/right). The ring-buffer has no concept of grouping, so on `clear()` it will cause random read/write index offsets - in the case of the `SAI` example, causing left/right switching.
- Usefulness of `write_immediate()` is not clear, should it be removed?
- The ring buffer may pause automatically on overrun. The ring buffer shall then be explicitly restarted by a user when new data is available.

PR #5469, opened 2026-02-16
stm32/i2s: fix frame misalignment after DMA ring buffer overrun recovery

HybridChild:
## Summary

Fix I2S frame misalignment after DMA ring buffer overrun recovery.

When an overrun occurs, `ReadableDmaRingBuffer::reset()` snaps the read index to the current DMA write position, which can land mid-frame. For stereo 32-bit I2S this means subsequent reads have swapped left channels or mixed high/low words — producing noise until the next overrun happens to land on a frame boundary by chance.

**Fix:** Add frame alignment support to `ReadableDmaRingBuffer`. When configured with an alignment value, `read_raw()`:
- Skips forward to the next frame boundary after overrun recovery
- Rounds down read lengths to whole frames, so the read index never drifts off a frame boundary

The I2S driver sets alignment automatically at construction based on the configured `Format` (2 half-words for 16-bit channels, 4 for 32-bit). `set_alignment()` asserts that the DMA buffer length is a multiple of the alignment value.

## Files changed

- `embassy-stm32/src/dma/ringbuffer/mod.rs` — alignment field, skip + round-down logic in `read_raw()`, tests
- `embassy-stm32/src/dma/dma_bdma.rs` — `set_alignment()` pass-through on `ReadableRingBuffer`
- `embassy-stm32/src/i2s.rs` — compute frame size and set alignment on RX ring buffer, update docs

### DIFF
--- a/embassy-stm32/src/dma/gpdma/ringbuffered.rs
+++ b/embassy-stm32/src/dma/gpdma/ringbuffered.rs
@@ -85,6 +85,13 @@ impl<'a, W: Word> ReadableRingBuffer<'a, W> {
         self.channel.start();
     }
 
+    /// Set the frame alignment for the ring buffer.
+    ///
+    /// See [`ReadableDmaRingBuffer::set_alignment`] for details.
+    pub fn set_alignment(&mut self, alignment: usize) {
+        self.ringbuf.set_alignment(alignment);
+    }
+
     /// Clear all data in the ring buffer.
     pub fn clear(&mut self) {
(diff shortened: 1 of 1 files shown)


elagil:
Nice! I assume this should close my old issue here: https://github.com/embassy-rs/embassy/issues/3590, at least partly.

HybridChild:
> Nice! I assume this should close my old issue here: #3590, at least partly.

Yes. This should fix the part about sample groups getting misaligned. As long as you remember to `set_alignment()`

PR #5470, opened 2026-02-16
[MCXA] Implement Drop for Lpuart

Wassasin:
Moved `buffered::State` to toplevel module and added a refcount. When constructing `Lpuart(R|T)x` the refcount is incremented, and on drop decremented. If the last refcount is dropped, clean up the peripheral.

Other changes are because `fn init` now takes the instance type T instead of the `Info` block. Also removed `initialized` from State, and the interrupt handler does not look at that anymore. Immediately returning from the handler is definitely wrong. Maybe needs some more safety guard when accessing the ring buffers in the interrupt handler.

### DIFF
--- a/embassy-mcxa/src/lpuart/mod.rs
+++ b/embassy-mcxa/src/lpuart/mod.rs
@@ -63,14 +63,14 @@ impl TxRxRefMask {
 
     /// Atomically signal that either the Tx or Rx has been dropped.
     ///
-    /// Returns `true` if all parts have been dropped.
-    pub fn unalive(&self, value: TxRxRef) -> bool {
+    /// Returns `true` if after this call all parts are inactive.
+    pub fn set_inactive_fetch_last(&self, value: TxRxRef) -> bool {
         let value = value as u8;
-        self.0.fetch_and(!value, Ordering::AcqRel) == value
+        self.0.fetch_and(!value, Ordering::AcqRel) & !value == 0
     }
 
     /// Atomically signal that either the Tx or Rx has been created.
-    pub fn alive(&self, value: TxRxRef) {
+    pub fn set_active(&self, value: TxRxRef) {
         let value = value as u8;
         self.0.fetch_or(value, Ordering::AcqRel);
     }
@@ -78,9 +78,8 @@ impl TxRxRefMask {
     /// Atomically determine if either channels have been created, but not dropped. Clears the state.
     ///
     /// Should only be relevant when any of the parts have been leaked using [core::mem::forget].
-    pub fn check_alive_reset(&self) -> bool {
-        // 'and' with zero clears the value.
-        self.0.fetch_and(0, Ordering::AcqRel) != 0
+    pub fn fetch_any_alive_reset(&self) -> bool {
+        self.0.swap(0, Ordering::AcqRel) != 0
     }
 }
 
@@ -766,7 +765,7 @@ fn disable_peripheral(info: &'static Info) {
 
 impl<M: Mode> Drop for LpuartTx<'_, M> {
     fn drop(&mut self) {
-        if self.state.tx_rx_refmask.unalive(TxRxRef::Tx) {
+        if self.state.tx_rx_refmask.set_inactive_fetch_last(TxRxRef::Tx) {
             disable_peripheral(self.info);
         }
     }
@@ -774,7 +773,7 @@ impl<M: Mode> Drop for LpuartTx<'_, M> {
 
 impl<M: Mode> Drop for LpuartRx<'_, M> {
     fn drop(&mut self) {
-        if self.state.tx_rx_refmask.unalive(TxRxRef::Rx) {
+        if self.state.tx_rx_refmask.set_inactive_fetch_last(TxRxRef::Rx) {
             disable_peripheral(self.info);
         }
     }
@@ -789,7 +788,7 @@ impl<'a, M: Mode> Lpuart<'a, M> {
         config: Config,
     ) -> Result<Option<WakeGuard>> {
         // Check if the peripheral was leaked using [core::mem::forget], and clean up the peripheral.
-        if T::state().tx_rx_refmask.check_alive_reset() {
+        if T::state().tx_rx_refmask.fetch_any_alive_reset() {
             disable_peripheral(T::info());
         }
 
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -739,15 +782,17 @@ impl<'a, M: Mode> Lpuart<'a, M> {
...
 impl<'a, M: Mode> LpuartTx<'a, M> {
-    fn new_inner(
-        info: &'static Info,
+    fn new_inner<T: Instance>(
         tx_pin: Peri<'a, AnyPin>,
         cts_pin: Option<Peri<'a, AnyPin>>,
         mode: M,
         wg: Option<WakeGuard>,
     ) -> Self {
+        T::state().tx_rx_refcount.fetch_add(1, Ordering::AcqRel);
```
jamesmunns: Should we be checking if the refcount is <= 2 here? What happens if someone forgets an LpuartTx/Rx and the drop is never called?

We could also be more specific and use one bit for "Tx exists" and one bit for "Rx exists" instead of just counting as an integer, and use fetch_or/fetch_and instead of fetch_add/fetch_sub

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -757,15 +802,17 @@ impl<'a, M: Mode> LpuartTx<'a, M> {
...
 impl<'a, M: Mode> LpuartRx<'a, M> {
-    fn new_inner(
-        info: &'static Info,
+    fn new_inner<T: Instance>(
         rx_pin: Peri<'a, AnyPin>,
         rts_pin: Option<Peri<'a, AnyPin>>,
         mode: M,
         _wg: Option<WakeGuard>,
     ) -> Self {
+        T::state().tx_rx_refcount.fetch_add(1, Ordering::AcqRel);
```
jamesmunns: Same note as LpuartTx

jamesmunns:
@Wassasin I think we should probably do something like this in all the public constructors:

```rust
// was a handle leaked? If so, we still have exclusive access since we have the
// T peripheral
if T::state().tx_rx_refcount.load(Acquire) != 0 {
    // Wait for TX operations to complete
    wait_for_tx_complete(info);

    // Clear all status flags
    clear_all_status_flags(info);

    // Disable the module - clear all CTRL register bits
    info.regs().ctrl().write(|w| w.0 = 0);
    
    // Force-clear the reference count before creating new handle(s)
    T::state().tx_rx_refcount.store(0, Release);
}
```

Wassasin:
I implemented a bitmask solution. Note that embassy-stm32 does not take leaking into account too, so we might want to fix it there as well.

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -40,7 +40,48 @@ struct State {
...
+    /// Atomically signal that either the Tx or Rx has been dropped.
+    ///
+    /// Returns `true` if all parts have been dropped.
+    pub fn unalive(&self, value: TxRxRef) -> bool {
+        let value = value as u8;
+        self.0.fetch_and(!value, Ordering::AcqRel) == value
+    }
+
+    /// Atomically signal that either the Tx or Rx has been created.
+    pub fn alive(&self, value: TxRxRef) {
```
jamesmunns: maybe `set_alive` or `set_active`?

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -40,7 +40,48 @@ struct State {
...
+
+impl TxRxRefMask {
+    pub const fn new() -> Self {
+        Self(AtomicU8::new(0))
+    }
+
+    /// Atomically signal that either the Tx or Rx has been dropped.
+    ///
+    /// Returns `true` if all parts have been dropped.
+    pub fn unalive(&self, value: TxRxRef) -> bool {
```
jamesmunns: maybe `set_unalive` or `set_inactive`?

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -40,7 +40,48 @@ struct State {
...
+        let value = value as u8;
+        self.0.fetch_or(value, Ordering::AcqRel);
+    }
+
+    /// Atomically determine if either channels have been created, but not dropped. Clears the state.
+    ///
+    /// Should only be relevant when any of the parts have been leaked using [core::mem::forget].
+    pub fn check_alive_reset(&self) -> bool {
+        // 'and' with zero clears the value.
+        self.0.fetch_and(0, Ordering::AcqRel) != 0
```
jamesmunns: could probably just use `swap` here

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -40,7 +40,48 @@ struct State {
...
+    pub const fn new() -> Self {
+        Self(AtomicU8::new(0))
+    }
+
+    /// Atomically signal that either the Tx or Rx has been dropped.
+    ///
+    /// Returns `true` if all parts have been dropped.
+    pub fn unalive(&self, value: TxRxRef) -> bool {
+        let value = value as u8;
+        self.0.fetch_and(!value, Ordering::AcqRel) == value
```
jamesmunns: The comment and/or this line is maybe a little misleading or confusing.

This function returns true if ONLY the `value` mask was set. Maybe a comment like:

> This function clears the given `value`, and returns true if that was the
> only bit that was previously set, meaning that all parts are now inactive/unalive.

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -689,6 +788,11 @@ impl<'a, M: Mode> Lpuart<'a, M> {
         enable_rx_rts: bool,
         config: Config,
     ) -> Result<Option<WakeGuard>> {
+        // Check if the peripheral was leaked using [core::mem::forget], and clean up the peripheral.
+        if T::state().tx_rx_refmask.check_alive_reset() {
```
jamesmunns: We might need this check in the `LpuartRx::new` and `LpuartTx::new` (and similar) constructors.

Wassasin: `Lpuart::init` is called from all constructors already.

PR #5472, opened 2026-02-16
[MCXA] Change Lpuart AtomicCell to maitake_sync::WaitCell

Wassasin:


embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -266,49 +264,49 @@ impl<'a> LpuartTx<'a, Buffered> {
...
-            if tx_empty && fifo_empty && tc_complete {
-                Poll::Ready(Ok(()))
-            } else {
-                Poll::Pending
-            }
-        })
+        Ok(self
+            .state
+            .tx_waker
+            .wait_for_value(|| {
```
jamesmunns: This could just be `wait_for` and return `tx_empty && fifo_empty && tc_complete`

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -413,38 +411,39 @@ impl<'a> LpuartRx<'a, Buffered> {
...
-                if to_copy > 0 {
-                    buf[read..read + to_copy].copy_from_slice(&data[..to_copy]);
-                    read += to_copy;
-                }
-                to_copy
-            });
+                let mut reader = unsafe { self.state.rx_buf.reader() };
+                let available = reader.pop(|data| {
+                    let to_copy = core::cmp::min(data.len(), buf.len() - read);
+                    if to_copy > 0 {
```
jamesmunns: here you do the copy conditionally, but above you do the copy unconitionally. Probably good to pick one.

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -413,38 +411,39 @@ impl<'a> LpuartRx<'a, Buffered> {
...
-                Poll::Ready(Ok(0))
-            }
-        })
-        .await
+                if read > 0 {
+                    Some(read)
+                } else if available == 0 {
+                    None
+                } else {
+                    Some(0)
```
jamesmunns: This is *probably* impossible? We can also probably just rewrite this to return `to_copy` if `to_copy` != 0 and get rid of the `let read` entirely

Issue #5449, opened 2026-02-11
Main function rustdoc error


psusi2:
It appears that #[embassy-executor::main] emits a #[doc(hidden)] before your main so it does not show up in rustdoc, but not before the __embassy_main() function that it adds.  It seems like this #[doc(hidden)] is in the wrong place.  I am also not sure if there is an attribute that the macro can add to cause it to show up in rustdoc as main instead of the ____embassy_main_task that it is renamed to.  If so, that would be nice.

psusi2:
It looks like #[embassy-executor::task] is what emits the #[doc(hidden)] to hide the function that it transforms the original function into, and then it emits the right stuff for rustdoc to display the original function.  #[embassy-executor::main] first transforms main itself though, and then calls #[embassy-executor::task], which transforms it again, hides the second transformation, and documents the first transformation as if it were the original.

PR #5473, opened 2026-02-16
Fix task macros to work right with rustdoc (#5449)

psusi2:
The main and task macros significantly transform the function. This makes it show up with the wrong name and/or return type, as well as not being an async fn in rustdoc.  Fix them to emit two versions of the fn, one with the original that is #[cfg(doc)] and the transformed version for #[cfg(not(doc))].  This way they appear in rustdoc correctly.

### DIFF
--- a/embassy-executor-macros/src/macros/main.rs
+++ b/embassy-executor-macros/src/macros/main.rs
@@ -221,12 +221,19 @@ For example: `#[embassy_executor::main(entry = ..., executor = \"some_crate::Exe
     }
 
     let result = quote! {
+        #[cfg(not(doc))]
         #[::embassy_executor::task()]
         #[allow(clippy::future_not_send)]
         async fn __embassy_main(#fargs) #out {
             #f_body
         }
 
+        #[cfg(doc)]
+        async fn main(#fargs) #out {
+            #f_body
+        }
+
+        #[cfg(not(doc))]
         #entry
         #main_attrs
         fn main() -> #main_ret {
--- a/embassy-executor-macros/src/macros/task.rs
+++ b/embassy-executor-macros/src/macros/task.rs
@@ -132,6 +132,7 @@ pub fn run(args: TokenStream, item: TokenStream) -> TokenStream {
     let task_ident = f.sig.ident.clone();
     let task_inner_ident = format_ident!("__{}_task", task_ident);
 
+    let orig_output = f.sig.output.clone();
     let task_inner_future_output = match &f.sig.output {
         ReturnType::Default => quote! {-> impl ::core::future::Future<Output = ()>},
         // Special case the never type since we can't stuff it into a `impl Future<Output = !>`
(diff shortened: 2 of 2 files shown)


Dirbaio:
I don't think this is actually desirable. If you apply the `task` macro the resulting function is no longer an `async fn`, it's a funciton that returns a `SpawnToken`, which is a completely different thing. You can't call it like you can call an `async fn`. Showing `async fn` in the docs is a lie.

Task functions show up as `pub(crate) fn foo() -> Result<SpawnToken<impl Sized>, SpawnError>` in rustdoc which is correct, I wouldn't change anything there.

About main, I don't think we can show anything correct in docs? You can't call main from the program (doing so would have fun reentrancy concerns), if you can't call it why do you need it documented?

psusi2:
> Task functions show up as `pub(crate) fn foo() -> Result<SpawnToken<impl Sized>, SpawnError>` in rustdoc which is correct, I wouldn't change anything there.

It is only correct from the perspective of the compiler, not from the human reading the docs and looking at the code.  As a human when I read the docs, I expect to see a function that -> !, like I have in my code, rather than this weird SpawnToken thing, and also to see that it is async rather than a regular fn.

> About main, I don't think we can show anything correct in docs? You can't call main from the program (doing so would have fun reentrancy concerns), if you can't call it why do you need it documented?

So that your comments explaining what main does show up in the docs, like it normally would without this macro.

Dirbaio:
> It is only correct from the perspective of the compiler, not from the human reading the docs

it *is* correct for the human as well. If you're a human reading rustdocs for some crate that has a `#[task] pub async fn`, you want to know that it's a `#[task]` because that affects how you call it. if it's an `async fn` you have to do `foo().await`, if it's a task you have to do `spawner.spawn(foo().unwrap())`.

psusi2:
Well, yes, it should be noted that because of the macro it isn't *really* an async fn, but it does *look* like one.  I'm not sure if there is a way to make the macro show up in rustdoc maybe with a link explaining its significance.

But the current state of throwing out your documentation for the fn and instead documenting something else is not really acceptable.

Dirbaio:
rustdoc's main purpose is documenting apis for e.g. libraries so that people who haven't written that code know how to call it. you're thinking about it from the other end, as a tool for documenting code for the person writing it. When the two purposes are in conflict the former should take priority. For example, for the person writing the code a task "looks like an async fn" indeed, but for the person calling it it's a function returning a SpawnToken, so the docs should show that. 

Happy to merge PRs that improve the docs output while respecting this principle.

psusi2:
Hrm... Now I'm wondering why the macros even need to work like this.  It looks like it turns your function into an async block insde of a non async fn that has a static memory pool to hold the returned future from the async block.  Instead of moving your async code to an async block, why not keep it as an async fn and then you actually could use it as a normal async fn. albeit one with a different name than the original that you use with spawner.spawn?

Dirbaio:
it's confusing that the macros create another function with a different name.  In general I don't think it's an option to change how the macros work at this stage.

psusi2:
So for some more context, yea, docs on a library are meant to explain how to use the library, but not so much the case for docs on a binary are they?  I'm hoping to use rustdoc on my binary project to satisfy ISO 62304 requirements to have a Software Detailed Design document.  As such I'm more focused on documenting how the software works.

Now that I think about it, while in general one might be able to use the same async fn as both a top level task, and not as one, don't people normally try to keep those two roles separate in their mind and think of some functions as top level tasks, and others as appearing further down to call stack?  So I would think having some clear indicator in rustdoc which fns are tasks would be nice in the general case.

Specifically for embassy, it's top level tasks can't really be called directly, but can only be passed to spawn right?  So would it be sufficient to have a notation in the docs stating that this is a top level task, and as such, it only *appears* to be an async fn, but you can't actually call it like one; it is only for passing to spawn()?

I'm still trying to find a nice looking way to add something like that to the docs in the macro while respecting the docs on the original fn.

psusi2:
Ok, how about this?  It adds a nice bold Task: prefix to the function  name so it sticks out on the functions page, and a footnote when you read the function that informs you that:

"This function is marked with #[task] so you can not call it as a normal async fn as the signature would suggest. The macro transforms it into a non async fn with a different return type. You can only pass this fn to spawn()"

Dirbaio:
I appreciate the insistence but as i said, rustdoc's goal is to document the public api for callers, and task functions *are* functions returning SpawnTokens for all intents and purposes. What docs are showing currently is the correct thing to show. Special-casing the macros to show something else makes the docs worse even if we try to compensate it in the dooc comment.

psusi2:
Ok, fine... you like the signature to not match the code.  But in the case of main, it isn't even mentioned in rustdoc and instead the internal __prefix function is, so this still needs a fix.

PR #5478, opened 2026-02-17
Change LED output pin from PA0 to PB14

WeGoToMars:
Change to match the actual MSPM0L2228 port


i509VCB:
While I do have a prerelease L2228 board (it is black, Rev E1), the release board (in red, Rev A) also uses PA0. SLAU928A (board datasheet, schematic section) shows PA0 is the red LED.

Screenshot_20260216_195455


Is there another rev I'm not aware of?

PR #5482, opened 2026-02-17
[MCXA] Improve clock requirement enforcement

jamesmunns:
This PR fixes some issues in clock enforcement discovered while testing low power UART.

It also improves the lpuart error handling a bit, though we still need to actively check for errors during a transfer.

### DIFF
--- a/embassy-mcxa/src/lpuart/mod.rs
+++ b/embassy-mcxa/src/lpuart/mod.rs
@@ -223,8 +223,13 @@ fn configure_control_settings(info: &'static Info, config: &Config) {
             w.set_pe(false);
         };
 
-        // Allow the lpuart to wake from deep sleep
-        w.set_dozeen(Dozeen::ENABLED);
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -222,6 +223,9 @@ fn configure_control_settings(info: &'static Info, config: &Config) {
             w.set_pe(false);
         };
 
+        // Allow the lpuart to wake from deep sleep
+        w.set_dozeen(Dozeen::ENABLED);
```
felipebalbi: What's the power impact here? Should this be enabled through `Config` passed to the constructor?

jamesmunns: I'll probably set this conditionally whether the peripheral is set to `AlwaysEnabled`, that way we don't have to worry about edge cases like "doze disabled, but periph should work in sleep mode".

PR #5488, opened 2026-02-18
[MCXA] Misc lpuart improvements

jamesmunns:
Just a couple things I saw along the way

### DIFF
--- a/embassy-mcxa/src/lpuart/buffered.rs
+++ b/embassy-mcxa/src/lpuart/buffered.rs
@@ -98,6 +98,11 @@ impl<'a> Lpuart<'a, Buffered> {
     /// Create a new full duplex buffered LPUART.
     ///
     /// Any external pin will be placed into Disabled state upon Drop.
+    ///
+    /// ## SAFETY
+    ///
+    /// You must NOT call `core::mem::forget` on `Lpuart` if `rx_buffer` or `tx_buffer` are not
+    /// `'static`. This will cause memory corruption.
     pub fn new_buffered<T: Instance>(
         _inner: Peri<'a, T>,
         tx_pin: Peri<'a, impl TxPin<T>>,
@@ -116,6 +121,11 @@ impl<'a> Lpuart<'a, Buffered> {
     /// Create a new buffered LPUART instance with RTS/CTS flow control.
     ///
     /// Any external pin will be placed into Disabled state upon Drop.
+    ///
+    /// ## SAFETY
+    ///
+    /// You must NOT call `core::mem::forget` on `Lpuart` if `rx_buffer` or `tx_buffer` are not
+    /// `'static`. This will cause memory corruption.
     pub fn new_buffered_with_rtscts<T: Instance>(
         _inner: Peri<'a, T>,
         tx_pin: Peri<'a, impl TxPin<T>>,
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -265,28 +265,14 @@ impl<'a> LpuartTx<'a, Buffered> {
 
     /// Write data asynchronously
     pub async fn write(&mut self, buf: &[u8]) -> Result<usize> {
```
felipebalbi: maybe you'd like to get rid of the `Result` alias, while you're working on LPUART?

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -305,26 +291,50 @@ impl<'a> LpuartTx<'a, Buffered> {
     }
 
     /// Try to write without blocking
-    pub fn try_write(&mut self, buf: &[u8]) -> Result<usize> {
-        let mut writer = unsafe { self.state.tx_buf.writer() };
-        let mut written = 0;
+    ///
+    /// May return 0 if the provided buf is zero, or there are no bytes available
+    pub fn try_write(&mut self, buf: &[u8]) -> usize {
```
felipebalbi, line 326: `try` usually implies the operation can fail. But this one can't. Is there are more fitting identifier here?

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -458,14 +474,15 @@ impl<T: Instance> crate::interrupt::typelevel::Handler<T::Interrupt> for Buffere
 
         let ctrl = regs.ctrl().read();
         let stat = regs.stat().read();
-        let has_fifo = regs.param().read().rxfifo() > 0;
+        let param = regs.param().read();
+        let has_rx_fifo = param.rxfifo() > 0;
```
jamesmunns, line 518: This is a C SDK-ism: the MCXA2xx *always* has a FIFO, but I'm leaving this as-is for now.

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -136,6 +139,8 @@ macro_rules! impl_instance {
                     fn info() -> &'static Info {
                         static INFO: Info = Info {
                             regs: pac::[<LPUART $n>],
+                            int_disable: crate::interrupt::typelevel::[<LPUART $n>]::disable,
```
jamesmunns, line 142: @felipebalbi this is a new pattern for how we could disable ints and the mrcc on drop of the peripheral. Interested to hear what you think about this.

embassy-mcxa/src/lpuart/buffered.rs
```diff
@@ -520,10 +533,22 @@ impl<T: Instance> crate::interrupt::typelevel::Handler<T::Interrupt> for Buffere
...
+                } else {
+                    0
+                }
+            };
 
             // Send data while TX buffer is ready and we have data
-            while regs.stat().read().tdre() == Tdre::NO_TXDATA {
+            for _ in 0..to_pop {
                 if let Some(byte) = reader.pop_one() {
+                    defmt::info!("boop");
```
felipebalbi: `defmt` can be disabled. Maybe wrap with a `cfg` check? Or remove it, as it looks like debugging code.

jamesmunns: Removed this, I shouldn't have left it in.

embassy-mcxa/src/lpuart/buffered.rs
jamesmunns: It's "try" because it might not write anything, I can't think of a better name honestly.

embassy-mcxa/src/lpuart/mod.rs
felipebalbi: I think there aren't many other ways on how to achieve this if we really want to erase the `Instance` (which I agree we do). I don't have any issues with the proposed pattern.

PR #5489, opened 2026-02-18
Disable the RCC for FDCAN when the Info reference counter for an instance reaches 0

kevswims:
Without this, the RCC is never disabled for the CAN and if the can is re-initialized later the RCC reference counter eventually overflows the uint8 that stores it

In our application we drop and re-initialize the CAN peripherals often to change modes (listen only, transmit). After 255 mode changes, the RCC enable was panicking when the u8 reference counter overflowed.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -12,7 +12,7 @@ use embassy_sync::waitqueue::AtomicWaker;
 use crate::can::fd::peripheral::Registers;
 use crate::gpio::{AfType, OutputType, Pull, SealedPin as _, Speed};
 use crate::interrupt::typelevel::Interrupt;
-use crate::rcc::{self, RccPeripheral};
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/can/fdcan.rs
```diff
@@ -951,6 +951,9 @@ impl Info {
                     let rx_pin = crate::gpio::AnyPin::steal(mut_state.rx_pin_port.unwrap());
                     rx_pin.set_as_disconnected();
                     self.interrupt0.disable();
+                    // We don't have the `T` here but since the RCC enable/disable is the same for
```
kevswims: If there's a less weird way to handle the lack of access to `T` here I'm open. This feels hacky but it works just fine.

xoviat: The `RccInfo` can be stored in the struct and then called as an object.

kevswims: Ahh, I tried that earlier but got stuck somewhere. I got it figured out now and implemented.

PR #5490, opened 2026-02-18
[MCXA] cdog: refactor code watchdog driver

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/139

Trying something new for the `cdog`. Some of the changes are highly desirable (now we can support both `cdog` instances), others are up for discussion. Maybe it looks too "abusive" of rust operators. Let me know


embassy-mcxa/src/cdog.rs
```diff
@@ -170,150 +185,249 @@ impl<'d> Watchdog<'d> {
...
+        self.info.regs().stop().write(|w| w.set_stp(check));
+        let reload = self.info.regs().reload().read().rload();
+        self.info.regs().reload().write(|w| w.set_rload(reload));
+        self.info.regs().start().write(|w| w.set_strt(check));
+
+        Ok(())
+    }
+}
+
+impl AddAssign<u32> for SecureCounter {
```
jamesmunns, line 323: I'd prefer to have these as explicit `increment` and `decrement` methods, rather than operator overloading.

examples/mcxa/src/bin/cdog.rs
```diff
@@ -21,41 +22,54 @@ async fn main(_spawner: Spawner) {
...
-        timeout: FaultControl::EnableInterrupt,
-        miscompare: FaultControl::EnableInterrupt,
-        sequence: FaultControl::EnableInterrupt,
-        state: FaultControl::EnableInterrupt,
-        address: FaultControl::EnableInterrupt,
-        irq_pause: PauseControl::PauseTimer,
-        debug_halt: PauseControl::PauseTimer,
-        lock: LockControl::Unlocked,
-    };
+    let mut config = hal::cdog::Config::default();
```
jamesmunns, line 25: I have a weak preference to actually hide these settings, and expose these as async methods, or have different Mode type states.

felipebalbi: I did consider that, yeah. However, we don't get a warning interrupt with this one. We need to feed *before* the interrupt triggers. There are three possibilities, actually: either you want an interrupt, or a reset, or neither.

So, let's say we want an interrupt, you seem to imply an API similar to what follows:

```rust
loop {
    cdog.instruction_counter().update(200);
    let status = cdog.wait_for_interrupt_and_get_status().await;
    do_something_with(status);
}
```

Sadly, that won't work. The whole idea of the code watchdog is that if the interrupt triggers, it should imply unauthorized code modification (an attack) which should be treated as sort of a critical situation. That is, we *don't* want the interrupt to ever trigger. Maybe we could `split()` the instruction counter into a writer and a waiter and dedicate a task to wait for that interrupt forever, such that applications can decide what to do if that interrupt ever comes. Or, the application could just bind their own interrupt handler instead?

PR #5492, opened 2026-02-18
STM32H7RS: Add internal flash config option for SYSCFG control

kevswims:
This allows controlling whether or not the application is responsible for enabling and setting up the SYSCFG peripheral. For programs running from internal flash such as bootloader, this needs to be setup. Programs running from external flash don't need to enable SYSCFG since it should have already been setup by the bootloader

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -339,14 +339,6 @@ pub struct Config {
     /// Allows JTAG pins to be used for GPIO
     #[cfg(stm32f1)]
     pub swj: gpio::SwjCfg,
-
-    /// Sets if SYSCFG should be enabled and reset or not.
-    /// 
-    /// If running from external flash, the bootloader should handle initializing the SYSCFG peripheral.
-    /// 
-    /// Defaults to true (bootloader or single internal flash applications)
-    #[cfg(stm32h7rs)]
-    pub internal_flash: bool,
 }
 
 impl Default for Config {
@@ -377,8 +369,6 @@ impl Default for Config {
             enable_ucpd2_dead_battery: false,
             #[cfg(stm32f1)]
             swj: Default::default(),
-            #[cfg(stm32h7rs)]
-            internal_flash: true,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/lib.rs
```diff
@@ -369,6 +377,8 @@ impl Default for Config {
             enable_ucpd2_dead_battery: false,
             #[cfg(stm32f1)]
             swj: Default::default(),
+            #[cfg(stm32h7rs)]
+            internal_flash: true,
```
kevswims: I could go either way on this default. Setting this to true makes sure all the examples work since they run from internal flash. This default will also break any applications that are running from external flash.

kevswims:
This was the commit that originally changed this behavior: https://github.com/embassy-rs/embassy/commit/13fc222ef45e16980aca221044ae41d106065ce1

There might be other clock and IO setup that should be put behind this config as well.

xoviat:
cargo +nightly fmt

Dirbaio:
does it *hurt* if the application reenables syscfg if it's already enabled?  If it does, maybe we can simply *not* reset syscfg?

Asking because it'd be better if can make it Just Work for both cases without needing a config option.

kevswims:
> does it _hurt_ if the application reenables syscfg if it's already enabled? If it does, maybe we can simply _not_ reset syscfg?
> 
> Asking because it'd be better if can make it Just Work for both cases without needing a config option.

If this is the only thing that will ever be different then that makes a lot of sense. I have a gut feeling there are probably other RCC things that applications running from external flash shouldn't be touching as well but maybe those are all fine?

I really could go either way on this.

xoviat:
cc @meowcakes

kevswims:
@Dirbaio I implemented an enable without reset for the RCC. Is this what you were thinking?

I left the original commit but if this looks good I'll drop it out.

Dirbaio:
> there are probably other RCC things that applications running from external flash shouldn't be touching

It might be the case. However, I wouldn't add a for-now-useless flag that complicates the public api just in case it *maybe* becomes useful in the future. We can always add it later when we discover it's actually needed.

PR #5493, opened 2026-02-18
[MCXA] Add part numbers

felipebalbi:
Add all MCXA2xx part numbers with pin definitions based on selected part number feature.

Manually reviewed datasheet to build this one. Clearly this is unsustainable long term, some of this will have to be autogenerated in the future.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/142


diondokter:
Oh, hmmm yeah, this is not great.
I guess it's ok for now...

What we probably need to do is replicate part of embassy-stm32.

We'll need some json/yaml that describes the peripherals and the pins all of the chips have.
Then we parse that and generate the pin impls in the build.rs and we create further cfg gates for each of the peripherals.

diondokter (approved):
Approve is for the structure. I haven't checked if all the cfgs are correct

diondokter:
I guess that's something I could pick up after the ADC work?
But we'll need to discuss where you want that all to live. In the HAL? In nxp-pac? Or in a new repo 'nxp-data'? That last one mirrors the stm32 setup.

This would open up the path of potentially unifying the different nxp HALs too

felipebalbi:
> Oh, hmmm yeah, this is not great. I guess it's ok for now...
> 
> What we probably need to do is replicate part of embassy-stm32.
> 
> We'll need some json/yaml that describes the peripherals and the pins all of the chips have. Then we parse that and generate the pin impls in the build.rs and we create further cfg gates for each of the peripherals.

IIRC, `nxp-pac` has support for that metadata. Still need to figure out if it does all we need.

felipebalbi:
> I guess that's something I could pick up after the ADC work? But we'll need to discuss where you want that all to live. In the HAL? In nxp-pac? Or in a new repo 'nxp-data'? That last one mirrors the stm32 setup.
> 
> This would open up the path of potentially unifying the different nxp HALs too

FYI https://github.com/embassy-rs/nxp-pac/blob/main/data/metadata/MIMXRT106x.json

@i509VCB, what is the output generated from the metadata? Are you generating pin function marker traits?

diondokter:
Oh cool! Then nxp-pac is already capable of more than I knew

felipebalbi:
> Approve is for the structure. I haven't checked if all the cfgs are correct

I'll wait until tomorrow to give @Wassasin and @jamesmunns a chance to read through.

jamesmunns:
It feels like if we are going to lean on nxp-pac metadata and stm32-metapac style generation, we probably want to focus on that instead of adding configs now that we're going to have to remove later.

We can prioritize this if we need to, I think it would be good to incrementally get nxp-pac metadata generating blocks, like GPIO, and go from there. I think it will take us a bit to reach parity, but I think we can start by using both.

i509VCB:
> > I guess that's something I could pick up after the ADC work? But we'll need to discuss where you want that all to live. In the HAL? In nxp-pac? Or in a new repo 'nxp-data'? That last one mirrors the stm32 setup.
> > This would open up the path of potentially unifying the different nxp HALs too
> 
> FYI https://github.com/embassy-rs/nxp-pac/blob/main/data/metadata/MIMXRT106x.json
> 
> @i509VCB, what is the output generated from the metadata? Are you generating pin function marker traits?

Marker traits and singletons.

Although I dislike having to hand write these at the moment.

diondokter:
> Although I dislike having to hand write these at the moment.

Do we have a good data source? I don't think the SVDs have the pin data.
We'd need something like those stm cube json's if we don't want to do everything by hand.

felipebalbi:
> It feels like if we are going to lean on nxp-pac metadata and stm32-metapac style generation, we probably want to focus on that instead of adding configs now that we're going to have to remove later.

the part number features will still have to exist. The only difference will be how we define which pins exist on which part number. We could, for example, use per-package features instead of explicitly choosing each chip part number. I can switch to that for now (manually).

> We can prioritize this if we need to, I think it would be good to incrementally get nxp-pac metadata generating blocks, like GPIO, and go from there. I think it will take us a bit to reach parity, but I think we can start by using both.

Fair enough, let me start looking at the metadata supported by `nxp-pac`.

felipebalbi:
Closing this one as we're moving in another direction.

PR #5496, opened 2026-02-19
embassy-rp: Fix race in dma irq handler

dobrowolski-lukasz:
This fixes a race that could arise in following situation: 
Two channels are active CH1 and CH2.
CH1 finishes and triggers an IRQ.
Handlers for CH1 and CH2 are ran.
Just after handler for CH2 fetched ints0 CH2 finishes and fires it's IRQ.
The currently running handler decides that there is no work to be done for CH2 and clears CH2 flag anyway.

This would result in CH2 IRQ being cleared without being handled.

Fix it by only clearing the flag if we did the work.

### DIFF
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -29,9 +29,10 @@ impl<T: ChannelInstance> interrupt::typelevel::Handler<T::Interrupt> for Interru
 
         let ints0 = pac::DMA.ints(0).read();
         if ints0 & (1 << channel) != 0 {
+            pac::DMA.ints(0).write_value(1 << channel);
+
             CHANNEL_WAKERS[channel].wake();
         }
-        pac::DMA.ints(0).write_value(1 << channel);
     }
 }
 


sourcebox:
I also encountered this problem and fixed it in a local branch. You still have a race condiction, though because the flag can be set again between the waker call and clearing it. To fix it, you have to swap the order of the calls: clear the flag, then call the waker.

Dirbaio:
> the flag can be set again between the waker call and clearing it

for this to happen the dma irq would have to happen twice without the async task getting polled in between, which can't happen, right? the task won't start another dma transfer until it gets polled and sees the previous transfer finished.

sourcebox:
> for this to happen the dma irq would have to happen twice without the async task getting polled in between, which can't happen, right? the task won't start another dma transfer until it gets polled and sees the previous transfer finished.

What I can say is, it seems to happen in my use case where 5 channels were running and I was able to fix it by changing the order.

dobrowolski-lukasz:
I suspect that sourcebox's issue happens due to IRQ handlers being ran on core 0 and task(s) on core 1.
If the IRQ handler itself is interrupted just after waking the task and the core 1 being idle the task would have enough time to setup a (ultra short) DMA transfer that completes and is cleared by now resumed IRQ handler. -> task hanged awaiting on Transfer future.
I think this would be possible if sourcebox's program has a long running code in IRQ context (InterruptExecutor) on core 0 using a higher priority than DMA IRQ.

dobrowolski-lukasz:
Seems the original version failed in CI for some reason. I'm looking into it now.

sourcebox:
> I think this would be possible if sourcebox's program has a long running code in IRQ context (InterruptExecutor) on core 0 using a higher priority than DMA IRQ.

It's correct that I have a long running interrupt executor on core 0 with a priority `Priority::P0`. Overall, the dma channels are handling several thousand requests per second. When I start the firmware without my fix, the task hangs after less than 1 second. With the fix, I could not trigger it within hours.

Dirbaio:
trying merge again to make sure the failure wasn't a fluke

dobrowolski-lukasz:
Failing test was removed(https://github.com/embassy-rs/embassy/pull/5500) because of fails on completely unrelated changes.

@lulf Could You re-trigger the CI?

PR #5501, opened 2026-02-20
[MCXA] clocks: fix SIRC clock switch conditional

felipebalbi:
We need wait for as long as the switch did *NOT* happen.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/154


jamesmunns (approved):
Good catch!

PR #5508, opened 2026-02-23
[MCXA] Improve adc interface

diondokter:
All config is now done upfront and supports pretty much all hardware features.

There are two downsides to this approach:
- If you want to change the config, you need to drop the peripheral and rebuild it
  - This is because we don't have the instance generic on the struct, so we can't know which configs are valid.
- You can only use a pin once. The hardware supports using any and all pins in any and all commands. But that doesn't work well with Rust's ownership. Open to suggestions.

### DIFF
--- a/embassy-mcxa/src/adc.rs
+++ b/embassy-mcxa/src/adc.rs
@@ -904,34 +904,49 @@ pub trait AdcPin<T: Instance>: sealed::SealedAdcPin<T> + GpioPin + PeripheralTyp
         let channel = self.channel();
         AnyAdcPin {
             channel,
-            pin: GpioPin::degrade(self),
+            pin: Some(GpioPin::degrade(self)),
             _phantom: PhantomData,
         }
     }
 }
 
-#[doc(hidden)] // Not really part of the public API
 /// A type-erased ADC pin
 pub struct AnyAdcPin<T> {
     channel: u8,
-    pin: AnyPin,
+    pin: Option<AnyPin>,
     _phantom: PhantomData<T>,
 }
 
 impl<T> AnyAdcPin<T> {
     #[inline]
     fn mux(&self) {
-        // Set to digital GPIO with input buffer disabled and no pull-ups.
-        // TODO also clear digital output value?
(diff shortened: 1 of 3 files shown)


diondokter:
Oh, I should add in more docs still

embassy-mcxa/src/adc.rs
```diff
@@ -499,32 +636,86 @@ impl<'a, M: Mode> Adc<'a, M> {
...
+
+            let (cmdl, cmdh) = match index {
+                0 => (adc.cmdl1(), adc.cmdh1()),
+                1 => (adc.cmdl2(), adc.cmdh2()),
+                2 => (adc.cmdl3(), adc.cmdh3()),
+                3 => (adc.cmdl4(), adc.cmdh4()),
+                4 => (adc.cmdl5(), adc.cmdh5()),
+                5 => (adc.cmdl6(), adc.cmdh6()),
+                6 => (adc.cmdl7(), adc.cmdh7()),
+                _ => unreachable!(),
```
felipebalbi, line 682: how about patching the PAC to merge these into an array? Then you could use `(adc.cmdl(index), adc.cmdh(index))`.

embassy-mcxa/src/adc.rs
```diff
@@ -741,10 +880,48 @@ pub trait AdcPin<T: Instance>: sealed::SealedAdcPin<T> + GpioPin + PeripheralTyp
...
+        let channel = self.channel();
+        AnyAdcPin {
+            channel,
+            pin: GpioPin::degrade(self),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+pub struct AnyAdcPin<T> {
```
felipebalbi, line 914: Maybe add a one-liner rustdoc stating this type-erased adc pin?

embassy-mcxa/src/adc.rs
diondokter: Hmmm yeah, this was inherited from the existing code.

I'll merge now and then update the pac

PR #5512, opened 2026-02-23
[MCXA] `bbqueue`-based buffered lpuart

jamesmunns:
This PR implements a `bbqueue` based interface for buffered lpuart operations. Important features include:

* Interrupt based state machine for loading/reloading buffers on the way in/out
* Ability to use just one half, if desired
* User provided buffer for custom sizes, operation tries to be reasonable with grant sizes regardless of the size provided
* async read, write, (write) flush methods
* Interrupts are automatically started/stopped when there is something/nothing to do
* TX only gets one interrupt per DMA transfer completion, up to 1024 bytes for a 4KiB buffer for example
* RX gets an interrupt when EITHER the full DMA has completed (e.g. the bbqueue grant is consumed), OR if the RX line becomes idle for more than one character. This means data usually becomes available immediately on completion of receiving data externally, or a bit sooner for large messages.

There are probably a bunch of improvements to be had, and we might want to carve out the bbqueue parts to be reusable for other drivers if we want, but this seemed like a good place to start!

This is also low-power friendly, though at 4mbaud we do have to keep a relatively high power clock enabled. I'll include some power and timing graphs before I mark this ready to review.

### DIFF
--- a/embassy-mcxa/src/lpuart/bbq.rs
+++ b/embassy-mcxa/src/lpuart/bbq.rs
@@ -106,8 +106,6 @@ pub struct BbqConfig {
     pub stop_bits_count: StopBits,
     /// RX IDLE configuration
     pub rx_idle_config: IdleConfig,
-    /// Swap TXD and RXD pins
-    pub swap_txd_rxd: bool,
 }
 
 impl Default for BbqConfig {
@@ -119,7 +117,6 @@ impl Default for BbqConfig {
             msb_first: MsbFirst::LSB_FIRST,
             stop_bits_count: StopBits::ONE,
             rx_idle_config: IdleConfig::IDLE_1,
-            swap_txd_rxd: false,
             power: PoweredClock::AlwaysEnabled,
             source: LpuartClockSel::FroLfDiv,
             div: Div4::no_div(),
@@ -140,7 +137,6 @@ impl From<BbqConfig> for super::Config {
             msb_first,
             stop_bits_count,
             rx_idle_config,
-            swap_txd_rxd,
         } = value;
 
         // User selectable
@@ -153,11 +149,11 @@ impl From<BbqConfig> for super::Config {
         cfg.msb_first = msb_first;
         cfg.stop_bits_count = stop_bits_count;
         cfg.rx_idle_config = rx_idle_config;
-        cfg.swap_txd_rxd = swap_txd_rxd;
 
         // Manually set
         cfg.tx_fifo_watermark = 0;
         cfg.rx_fifo_watermark = 0;
+        cfg.swap_txd_rxd = false;
 
         cfg
     }


embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1025 @@
...
+            self.state.rxdma.get().read()
+        };
+
+        // Re-magic the mut slice from the storage we have now reclaimed by dropping the
+        // rx_queue.
+        let rx_buffer = unsafe { core::slice::from_raw_parts_mut(ptr.as_ptr(), len) };
+
+        // Now, if this was the last part of the lpuart, we are responsible for peripheral
+        // cleanup.
+        if (state & !(STATE_TXGR_ACTIVE | STATE_TXDMA_PRESENT)) == STATE_INITED {
```
jamesmunns: This is wrong - copy/paste error

embassy-mcxa/src/dma.rs
```diff
@@ -614,6 +614,20 @@ impl DmaChannel<'_> {
         pac::EDMA_0_TCD0.tcd(self.channel.index())
     }
 
+    /// set a manual callback to be called AFTER the DMA interrupt is processed. Will be called in the DMA interrupt
+    /// context.
+    pub(crate) unsafe fn set_callback(&mut self, f: fn()) {
```
jamesmunns, line 567: todo: we need a "clear_callback" method, `LpuartBbqRx::teardown` should call it.

examples/mcxa/src/bin/lpuart_bbq_rx.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let p = hal::init(cfg);
+
+    defmt::info!("LPUART DMA example starting...");
+
+    // Create UART configuration
+    let config = Config {
+        baudrate_bps: 4_000_000,
+        power: PoweredClock::NormalEnabledDeepSleepDisabled,
+        rx_fifo_watermark: 0,
```
felipebalbi: should the watermark be in bytes? That is, we handle the "off-by-one" internally?

embassy-mcxa/src/lpuart/mod.rs
```diff
@@ -870,7 +894,7 @@ impl<'a, M: Mode> Lpuart<'a, M> {
         clear_all_status_flags(T::info());
         configure_flow_control(T::info(), enable_tx_cts, enable_rx_rts, &config);
         configure_bit_order(T::info(), config.msb_first);
-        enable_transceiver(T::info(), enable_rx, enable_tx);
+        enable_transceiver(T::info(), enable_tx, enable_rx);
```
felipebalbi, line 907: Is this a bugfix? Do we need this to be merged as a separate PR, while this is still in draft?

jamesmunns: Yes, this is a bugfix, feel free to split off now if you want, or I can do it tomorrow!

examples/mcxa/src/bin/lpuart_bbq_rx.rs
jamesmunns: Honestly I really want to rip out the current `Config` interface. I might do that for the bbq variant in this PR, and come back for the other variants later.

But yeah, the config struct has the register values, which are (N - 1) of where they trigger (the hardware triggers when the depth EXCEEDS the watermark). It's not *wrong*, it's just a little foot-gun'y, and honestly we shouldn't be exposing it to the user anyway.

embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1025 @@
...
+    /// The TX half of the LPUART
+    pub tx: LpuartBbqTx<'a>,
+    /// The RX half of the LPUART
+    pub rx: LpuartBbqRx<'a>,
+}
+
+impl<'a> LpuartBbq<'a> {
+    /// Create a new LpuartBbq with both transmit and receive halves
+    pub fn new<T: BbqInstance>(
+        _inner: Peri<'a, T>,
```
jamesmunns: todo: review this 'a lifetime. the `status` check ensures we can't recreate the bbq unless we call `teardown`, but TECHNICALLY we could teardown a bbq lpuart and make a blocking lpuart, and things would probably get *weird*.

We could make this `'static` like the other stuff. I probably also want to check the pins too, we might be able to actually make dupes atm.

embassy-mcxa/src/dma.rs
```diff
@@ -1374,8 +1404,9 @@ impl DmaChannel<'_> {
             w.set_dreq(Dreq::ERQ_FIELD_CLEAR);
             w.set_esg(Esg::NORMAL_FORMAT);
             w.set_majorelink(false);
-            w.set_eeop(false);
-            w.set_esda(false);
+            // TODO(AJM): Determine if these are necessary or if I can revert this change
+            w.set_eeop(true);
+            w.set_esda(true);
```
Wassasin: * EEOP -> end-of-packet processing -> terminates the major loop transfer if the peripheral sends a End-Of-Packet. You would only want this if you want to force the CPU to process every packet immediately.
* ESDA -> Enable Store Destination Address -> writes last dest address to SLAST_SDA for either retrieval or for the next transfer. For a single shot transfer setting this is fine, as long as we reconfigure everything. For a generic circular transfer changing this is not what we want. For EOP this is especially interesting as it sets "where you left off" so to speak.

Unsure whether EEOP is currently doing something for you, but enabling it generally for all call to setup_read_from_peripheral seems OK to me.

I think ESDA is not doing anything for you as you are using DADDR. Also it does not seem to be generally applicable for all peripheral transfers.

embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1367 @@
...
+        // Now, drop the queue in place. This is sound because as the LpuartBbqTx, we have exclusive
+        // access to the "producer" half, and by disabling the interrupt and notching out the state
+        // bits, we know the ISR will no longer touch the consumer part.
+        //
+        // Also, take the DmaChannel by ownership this time.
+        let tx_dma = unsafe {
+            core::ptr::drop_in_place(self.state.tx_queue.get());
+            // Defensive coding: purge the tx_queue just in case. This doesn't zero the
+            // whole buffer, only the tracking pointers.
+            core::ptr::write_bytes(self.state.tx_queue.get(), 0, 1);
```
Wassasin, line 648: Perhaps add a function to BBQueue to do this. Calling [core::ptr::write_bytes] on a external lib type is spicey.

embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1367 @@
...
+        // Now, drop the queue in place. This is sound because as the LpuartBbqRx, we have exclusive
+        // access to the "consumer" half, and by disabling the interrupt and notching out the state
+        // bits, we know the ISR will no longer touch the producer part.
+        //
+        // Also, take the DmaChannel by ownership this time.
+        let rx_dma = unsafe {
+            core::ptr::drop_in_place(self.state.rx_queue.get());
+            // Defensive coding: purge the rx_queue just in case. This doesn't zero the
+            // whole buffer, only the tracking pointers.
+            core::ptr::write_bytes(self.state.rx_queue.get(), 0, 1);
```
Wassasin, line 932: Again spicey.

embassy-mcxa/src/lpuart/bbq.rs
jamesmunns: Yeah, we *could* (I should check what the drop impl for a queue does, it should probably zero the pointers anyway), though to be fair: the type is ACTUALLY `UnsafeCell<MaybeUninit<Queue>>`, and once we have taken the value out by ownership, it's fundamentally `UnsafeCell<MaybeUninit.uninit>`, so writing all zeroes to it is clearly sound, even if it is a bit spicy/smelly.

embassy-mcxa/src/dma.rs
jamesmunns: Removed

embassy-mcxa/src/lpuart/bbq.rs
jamesmunns: I'll leave as-is until I update bbqueue

embassy-mcxa/src/lpuart/bbq.rs
jamesmunns: same notes as above

embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1486 @@
...
+    /// Number of data bits
+    pub data_bits_count: DataBits,
+    /// MSB First or LSB First configuration
+    pub msb_first: MsbFirst,
+    /// Number of stop bits
+    pub stop_bits_count: StopBits,
+    /// RX IDLE configuration
+    pub rx_idle_config: IdleConfig,
+    /// Swap TXD and RXD pins
+    pub swap_txd_rxd: bool,
```
felipebalbi: Just curious, what is the need for this?

embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -0,0 +1,1486 @@
...
+//
+// SAFETY: Should only be called by the `on_interrupt` function in ISR context, with
+// the shared BbqState properly initialized in the INITED state
+unsafe fn handler(info: &'static Info, state: &'static BbqState) {
+    let regs = info.regs();
+    let ctrl = regs.ctrl().read();
+    let stat = regs.stat().read();
+
+    // Just clear any errors - TODO, signal these to the consumer?
+    // For now, we just clear + discard errors if they occur.
```
felipebalbi, line 1345: Yeah, we need to think about how to handle the errors at some point. This could leave the UART in a wedged state.

embassy-mcxa/src/lpuart/bbq.rs
jamesmunns: This existed in the original lpuart config, but you're right we don't do anything with it in the bbq variant. I'll remove it.

embassy-mcxa/src/lpuart/bbq.rs
jamesmunns: I don't *think* this could leave us wedged: we clear the errors unconditionally. But we don't signal the error to the user at all.

One option is to hard-stop receiving when we encounter an error, so we stop filling the queue until the user consumes the queue, THEN signal the error condition, so they know "when" it happened. Then after they "get" that notification, we auto-restart receiving. The tradeoff is we lose any data until the error is "acked".

PR #5517, opened 2026-02-24
[MCXA] i2c/i3c: let application choose clock source

felipebalbi:
Instead of attemping to guess which clock source to use, let the application tell us which one to use, just like SPI and I2C Target drivers.


jamesmunns:
CC @diondokter, flaky adc-compare test :/

PR #5524, opened 2026-02-25
[MCXA] Clean up DMA setup_* methods

Wassasin:


embassy-mcxa/src/dma.rs
```diff
@@ -227,30 +210,50 @@ impl Word for u32 {
...
+                half_transfer_interrupt: false,
+                complete_transfer_interrupt: false,
+            }
+        }
+    }
+
+    impl From<EnableComplete> for TransferOptions {
+        fn from(_value: EnableComplete) -> Self {
+            TransferOptions {
+                half_transfer_interrupt: false,
```
felipebalbi: Nice, a half-transfer interrupt. We can probably use this to implement circular transfers.

embassy-mcxa/src/dma.rs
```diff
@@ -221,36 +177,50 @@ impl Word for u32 {
...
-            complete_transfer_interrupt: true,
+/// Typical variants of [TransferOptions] to be used as shorthands.
+pub mod transfer_opts {
+    use crate::dma::{Priority, TransferOptions};
+
+    /// Short-hand to specify that no options should be configured.
+    pub struct NoInterrupt;
+
+    /// Short-hand to specify that only the complete transfer interrupt should be triggered.
+    pub struct EnableComplete;
```
jamesmunns: Instead of using marker types and doing `into`s for this, would it make sense to have methods or consts instead? e.g.

```rust
impl TransferOptions {
    pub fn enable_complete() -> Self { ... }
    // or
    pub const ENABLE_COMPLETE: Self = ...;
}
```

That way you get

```rust
.setup_write_to_peripheral(data, peri_addr, TransferOptions::enable_complete())?;
// or
.setup_write_to_peripheral(data, peri_addr, TransferOptions::ENABLE_COMPLETE)?;
// instead of
.setup_write_to_peripheral(data, peri_addr, EnableComplete.into())?;
```

jamesmunns:
@Wassasin you might want to look at the tests, there's a DMA test timing out, and the lpuart test hardfaults: https://ci.embassy.dev/jobs/0773122cb134

Wassasin:
Seem spurious, I'll look at it once I've received a new EVK. For now this can be merged as this does not seem to be caused by the changes in this PR.

jamesmunns:
Note for @Wassasin - the CI failure was spurious due to a TRNG test failing, I've reported to NXP to see what we need to do to fix that.

PR #5528, opened 2026-02-25
Map I2S_SDI and I2S_SDO to I2sSdPin

everdrone:
`I2sSdPin` should be implemented for `I2S_SDI` and `I2S_SDO` to use the interface in txonly or rxonly mode


leftger (approved):
LGTM 🚀

PR #5532, opened 2026-02-26
[MCXA] embassy-mcxa: add hash + sgi modules

alamfarjadf:
Add hardware accelerated hashing support for MCXA family with SGI (Secure Generic Interface) support.

Currently supports three modes: 1) blocking, 2) Streaming multi-part and 3) async via DMA.

Only supports CNSA 2.0 compliant hash sizes (i.e. 384 and 512).

### DIFF
--- a/embassy-mcxa/src/chips/mcxa2xx.rs
+++ b/embassy-mcxa/src/chips/mcxa2xx.rs
@@ -774,7 +774,9 @@ pub(crate) mod peripheral_gating {
     impl_cc_gate!(LPSPI0, mrcc_glb_acc0, mrcc_glb_rst0, lpspi0, LpspiConfig);
     impl_cc_gate!(LPSPI1, mrcc_glb_acc0, mrcc_glb_rst0, lpspi1, LpspiConfig);
 
+    // The following peripherals are no reset peripherals
     impl_cc_gate!(WWDT0, mrcc_glb_acc0, wwdt0, Clk1MConfig);
+    impl_cc_gate!(SGI0, mrcc_glb_acc1, sgi0, Clk1MConfig);
 }
 
 pub(crate) mod clock_limits {
--- a/embassy-mcxa/src/chips/mcxa5xx.rs
+++ b/embassy-mcxa/src/chips/mcxa5xx.rs
@@ -296,7 +296,7 @@ mod inner_periph {
         // SAU,
         // SCG0,
         // SCN_SCB,
-        // SGI0,
+        SGI0,
         // SMARTDMA0,
         // SPC0,
         // SYSCON,
@@ -412,7 +412,7 @@ mod inner_interrupt {
         // RTC,
         // RTC_1HZ,
         // SCG0,
-        // SGI,
+        SGI,
         // SLCD,
         // SMARTDMA,
         // SPC0,
@@ -738,8 +738,10 @@ pub(crate) mod peripheral_gating {
     // impl_cc_gate!(LPSPI0, mrcc_glb_acc0, mrcc_glb_rst0, lpspi0, LpspiConfig);
     // impl_cc_gate!(LPSPI1, mrcc_glb_acc0, mrcc_glb_rst0, lpspi1, LpspiConfig);
 
+    // The following peripherals are no reset peripherals 
     impl_cc_gate!(WWDT0, mrcc_glb_acc0, wwdt0, Clk1MConfig);
     impl_cc_gate!(WWDT1, mrcc_glb_acc0, wwdt1, Clk1MConfig);
+    impl_cc_gate!(SGI0, mrcc_glb_acc4, sgi0, Clk1MConfig);
 }
 
 pub(crate) mod clock_limits {
--- a/embassy-mcxa/src/lib.rs
+++ b/embassy-mcxa/src/lib.rs
@@ -41,6 +41,9 @@ mod all_chips {
     pub mod crc;
     pub mod ctimer;
     pub mod dma;
+    #[path = "sgi/hash.rs"]
+    pub mod hash;
+    pub mod sgi;
     #[cfg(feature = "custom-executor")]
     pub mod executor;
     pub mod gpio;
--- a/embassy-mcxa/src/sgi/hash.rs
+++ b/embassy-mcxa/src/sgi/hash.rs
@@ -1,0 +1,643 @@
+// Hash functionality using SGI hardware
+
+use super::sgi::{SGIError, Sgi};
+
+// ============================================================================
+// Public constants re-exported by `crate::sgi`
+//
+// These are used as self-documenting `HashOptions` field values.
+// ============================================================================
+
+/// Maximum SHA-384/512 block size in bytes.
+pub const MAX_BLOCK_SIZE: usize = 128;
+
+/// Output buffer size used by the SGI SHA2 engine (max digest size = SHA-512).
+pub const SGI_HASH_OUTPUT_SIZE: u32 = 64;
+
+// SGI_SHA2_CTRL register bitfields
+pub const SGI_SHA2_CTRL_MODE_AUTO: u32 = 1 << 1;
+pub const SGI_SHA2_CTRL_MODE_NORMAL: u32 = 0 << 1;
+pub const SGI_SHA2_CTRL_SIZE_SHA384: u32 = 2 << 2;
+pub const SGI_SHA2_CTRL_SIZE_SHA512: u32 = 3 << 2;
+pub const SGI_SHA2_CTRL_HASH_RELOAD: u32 = 1 << 13;
+pub const SGI_SHA2_CTRL_NO_HASH_RELOAD: u32 = 0 << 13;
+pub const SGI_SHA2_CTRL_NO_AUTO_INIT: u32 = 1 << 15;
+pub const SGI_SHA2_CTRL_AUTO_INIT: u32 = 0 << 15;
+
+
+use crate::dma::{DmaChannel, Transfer};
+
+const SHA384_DIGEST_LEN: usize = 48;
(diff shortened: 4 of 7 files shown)


jamesmunns:
@alamfarjadf if you run `./fmtall.sh` from the top of the `embassy` folder, it should fix your formatting woes!

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+    HashingNotComplete,
+}
+
+// ============================================================================
+// SGI driver interface and implementations
+
+static SGI_EXISTS: AtomicBool = AtomicBool::new(false);
+/// SGI (Secure Generic Interface) hardware abstraction
+pub struct Sgi {
+    base_addr: u32,
```
felipebalbi, line 195: You need to add an `Instance` trait that gives you the base address by taking it from the PAC. Then implement the trait to the valid peripheral singleton(s). Look at the i2c, spi, and i3c drivers for reference.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+    Timeout,
+    HardwareError,
+    DmaError,
+    HashingNotComplete,
+}
+
+// ============================================================================
+// SGI driver interface and implementations
+
+static SGI_EXISTS: AtomicBool = AtomicBool::new(false);
```
felipebalbi, line 192: this should not be required.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+}
+
+impl Sgi {
+    /// Create a new SGI instance with the default base address
+    pub fn new() -> Result<Self, SGIError> {
+        if SGI_EXISTS.swap(true, Ordering::SeqCst) {
+            return Err(SGIError::SGIInstanceAlreadyExists); // Another instance already exists
+        }
+        Ok(Self { base_addr: SGI0_BASE })
+    }
```
felipebalbi, line 230: This should look more like:

```rust
pub fn new<T: Instance>(
    _peri: Peri<'d, T>,
    _irq: _irq: impl interrupt::typelevel::Binding<T::Interrupt, InterruptHandler<T>> + 'd,
    config: Config,
) -> Result<Self, SetupError> {
    // enable clocks
    let parts = unsafe { enable_and_reset::<T>(&conf).map_err(SetupError::ClockSetup)? };
    let mut inst = Self {
        info: T::info(),
    };

    inst.set_configuration(&config)?;

    Ok(inst)
}
```

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    #[inline(always)]
+    fn read_reg(&self, offset: u32) -> u32 {
+        unsafe { core::ptr::read_volatile((self.base_addr + offset) as *const u32) }
+    }
+
+    #[inline(always)]
+    fn write_reg(&self, offset: u32, value: u32) {
+        unsafe { core::ptr::write_volatile((self.base_addr + offset) as *mut u32, value) }
+    }
```
felipebalbi, line 244: These are given to you by the PAC.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    #[inline(always)]
+    fn status(&self) -> u32 {
+        self.read_reg(SGI_STATUS_OFFSET)
+    }
+
+    pub fn set_auto_dma_handshake(&mut self, input_fifo_enable: bool, output_fifo_enable: Option<bool>) {
+        let mut value = 0u32;
+        if input_fifo_enable {
+            value |= SGI_AUTO_DMA_CTRL_IFE;
```
felipebalbi, line 254: PAC provides safer and more readable accessors for reading, writing, and modifying a register.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+                SGI_CTRL_OFFSET,
+                SGI_CTRL_DATOUT_RES_TRIGGER_UP | SGI_CTRL_START,
+            );
+            self.wait_until_sha2_not_busy()?;
+            self.read_dataout_block(output, idx + (16 * (i + 1)))?; // Each block is 16 bytes
+        }
+        Ok(())
+    }
+
+    pub fn update_partial_output(&mut self, options: HashOptions, output: &mut [u8]) -> Result<(), SGIError> {
```
felipebalbi, line 579: all `pub` functions should have a sensible rustdoc comment.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+                w.set_esg(Esg::NORMAL_FORMAT);
+                w.set_majorelink(false);
+                w.set_eeop(false);
+                w.set_esda(false);
+                w.set_bwc(Bwc::NO_STALL);
+                w.set_start(Start::CHANNEL_STARTED);
+            });
+        }
+
+        info!("DMA started for {} bytes (CITER=1, NBYTES={})", length, length_u32);
```
felipebalbi, line 443: the HAL can be compiled without `defmt` support. Look at other drivers. How was that solved?

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+pub const SGI_CTRL2_BYTES_ORDER: u32 = 1 << 22;
+pub const SGI_CTRL2_GCM_INXOR: u32 = 1 << 23;
+
+// SHA output sizes
+pub const SGI_HASH_OUTPUT_SIZE: usize = 64;
+pub const MAX_BLOCK_SIZE: usize = 128; // Maximum block size for SHA-384/SHA-512
+
+// SGI_AUTO_DMA_CTRL register bitfields
+pub const SGI_AUTO_DMA_CTRL_IFE: u32 = 1 << 0; // Input FIFO DMA handshake enable - for non SHA operations only.
+pub const SGI_AUTO_DMA_CTRL_OFE: u32 = 1 << 8; // Output FIFO DMA handshake enable - for non SHA operations only.
```
felipebalbi, line 167: None of these constants are needed and, even if they were, they shouldn't be `pub` because you don't want anything outside this driver fiddling with your registers.

In any case, use the PAC accessors.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+pub struct Sgi {
+    base_addr: u32,
+}
+
+impl Drop for Sgi {
+    fn drop(&mut self) {
+        // Ensure the global singleton lock is always released, even if callers
+        // return early via `?` before reaching an explicit `release()`.
+        //
+        // Keep Drop best-effort and non-blocking.
```
felipebalbi, line 203: no, no. You want to block until you don't have pending transfers anywhere. Make sure DMA has been stopped and any in-flight transfers, completed before allowing drop to return.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    pub fn is_sha2_busy(&self) -> bool {
+        (self.status() & SGI_STATUS_SHA2_BUSY) != 0
+    }
+
+    pub fn wait_until_sha2_not_busy(&mut self) -> Result<(), SGIError> {
+        let start_time = Instant::now();
+        const MAX_WAIT_TIME_USEC: u64 = 1000; // 1 millisecond timeout for normal SHA2 operations.
+        while self.is_sha2_busy() {
+            if is_expired(start_time, MAX_WAIT_TIME_USEC) {
```
felipebalbi, line 483: blocking implementation should, really, wait forever if it never completes. You should not add timeout handling here. Add the bare minimum, read status registers to make sure there are no errors, etc. But don't add a SW timeout.

embassy-mcxa/src/hash/sgi.rs
```diff
@@ -0,0 +1,650 @@
...
+        let dataout_offset = SGI_DATOUT_START_OFFSET as usize;
+        const TOTAL_DATAOUT_REGS: usize = 4; // DATOUTA, DATOUTB, DATOUTC, DATOUTD
+        for i in 0..TOTAL_DATAOUT_REGS {
+            let word = self.read_reg(dataout_offset as u32 + (i as u32 * 4));
+            output[idx + (i * 4)..idx + (i * 4) + 4].copy_from_slice(&word.to_be_bytes());
+        }
+        Ok(())
+    }
+
+    pub fn interrupt_is_operation_done(&self) -> Result<bool, SGIError> {
```
felipebalbi, line 503: no interrupt handler?

alamfarjadf:
> @alamfarjadf if you run `./fmtall.sh` from the top of the `embassy` folder, it should fix your formatting woes!

Thank you!!

embassy-mcxa/src/hash/sgi.rs
alamfarjadf: This is the PDONE bit in INT_STS register, not an actual NVIC interrupt which isn't setup yet. For async hash, the FIFO processes the blocks in a continuous streaming mode, so once the last DMA interrupt comes in, the hashing will be done in a bounded amount of time. But for improvement, an NVIC interrupt should be configured for future operations (i.e. AES).

embassy-mcxa/src/hash/sgi.rs
alamfarjadf: For DMA transfers, the DmaHasher instance will maintain the SGI block until either the hash completes or the DMA errors out. DmaHasher has a sgi_instance member, which will not release the lock.

The '?' operators releasing the SGI singleton apply for blocking/streaming modes.

Issue #5513, opened 2026-02-23
[MCXA] Add support for MCX A5xx


felipebalbi:
This is a tracking issue for the work required to add support for the new MCX A5xx family. Work on [nxp-pac](https://github.com/embassy-rs/nxp-pac) and [embassy](https://github.com/embassy-rs/embassy) will happen in the near future.

FYI: the MCX A55x is a pre-production device.  Device samples and evaluation board quantities are limited.  Please contact NXP sales for updates on availability.

PR #5536, opened 2026-02-27
[MCXA] Prep a "for now" MCXA5xx split

jamesmunns:
CC #5513, https://github.com/embassy-rs/nxp-pac/issues/36

We probably won't have the "metapac" ready to go before we need to start working on MCXA5xx support. This PR sets up a provisional project structure, assuming we will add mcxa577 support to the nxp-pac the same way we're support mcxa256 currently, and we can start tweaking modules on a case-by-case basis to allow the HAL to support both.

Some solutions are a little creative, I'm open to suggestions.


embassy-mcxa/src/lib.rs
```diff
@@ -6,576 +6,70 @@
 // Allow functions with too many args - we have a lot of HAL constructors like this for now
 #![allow(clippy::too_many_arguments)]
 
-// //! ## Feature flags
-// #![doc = document_features::document_features!(feature_label = r#"<span class="stab portability"><code>{feature}</code></span>"#)]
-
-pub mod adc;
```
diondokter, line 12: Typically you'd keep these here and gate them with a `cfg`.

https://github.com/embassy-rs/embassy/blob/c05fc75206966b8b5421cfc5f9268c2a4a240ccc/embassy-stm32/src/lib.rs#L55-L67

The way you've done it will work, but maybe it's a bit overindexing on the amount of differences there will be

embassy-mcxa/src/chips/mcxa2xx.rs
```diff
@@ -0,0 +1,722 @@
...
+        WUU0,
+        WWDT0,
+    );
+}
+
+// Use cortex-m-rt's #[interrupt] attribute directly; PAC does not re-export it.
+
+/// Initialize HAL with configuration (mirrors embassy-imxrt style). Minimal: just take peripherals.
+/// Also applies configurable NVIC priority for the OSTIMER OS_EVENT interrupt (no enabling).
+pub fn init(cfg: crate::config::Config) -> Peripherals {
```
diondokter, line 466: We're not gonna share an init?

Granted, this is a very annoying part in embassy-stm32, but fully separating them is maybe not the best

embassy-mcxa/src/lib.rs
jamesmunns: I can do that, if we want. I figured this would later be replaced with metadata info, so I didn't want to spend too much fine-grained effort splitting out the features for now, really just "exclusive" and "all".

embassy-mcxa/src/chips/mcxa2xx.rs
jamesmunns: *for now*, I think split is more reasonable, otherwise we need to basically feature gate block-by-block. I think once we have some basics like gpio and ostimer set up and working on both, we could move to a shared init.

jamesmunns:
@diondokter re: my thought process on this, I'm of the opinion that this setup is a "temporary jig", since we'll eventually move to the metapac approach. So it is definitely a little "worse is better", and "these are for-now decisions, not for-ever decisions".

PR #5537, opened 2026-02-27
Add sync methods to embassy-net

chemicstry:
This adds sync alternatives to all embassy-net sockets in the `try_xyz` flavor. This allows using embassy-net in mixed sync/async environments.


Dirbaio:
Looks good overall. 2 things:

- The naming for the "not ready yet" errors is a bit inconsistent. It's named `Exhausted`, `BufferFull` or `WouldBlock` depending on the place. 
- Adding it to the error enums means the non-`try` methods can still potentially return them, users are forced to handle them uselessly.

Not sure what's the cleanest solution. One is to add `enum TryError<T>{ WouldBlock, Other(T) }`, and wrap the non-try errors with it everywhere. If you have other ideas i'm all ears.

chemicstry:
The naming comes from `smoltcp`, I thought there is a reason for it, but maybe not. The only alternative I can think of is `nb` crate, but it seems that ecosystem is moving away from it? If not `nb`, I think your suggested `TryError` would be best. Let me know how to proceed.

Dirbaio:
yes, let's avoid `nb`.

chemicstry:
@Dirbaio this should be ready

PR #5539, opened 2026-02-27
[MCXA] crc: add Instance trait

felipebalbi:
Instead of assuming the existence of a unique singleton named CRC0, add an Instance trait and depend on that instead.


Wassasin:
Github HTTP server hickup seemingly, restarted queue.

PR #5542, opened 2026-02-27
[MCXA] Check errors for TRNG

jamesmunns:
In https://ci.embassy.dev/jobs/caa63977af64, we are timing out. This is a quick fix to try and check errors more proactively in case that is the cause of the timeout in the HIL test.

### DIFF
--- a/embassy-mcxa/src/trng.rs
+++ b/embassy-mcxa/src/trng.rs
@@ -6,14 +6,13 @@ use core::sync::atomic::{AtomicU32, Ordering};
 use embassy_hal_internal::Peri;
 use embassy_hal_internal::interrupt::InterruptExt;
 use maitake_sync::WaitCell;
-use nxp_pac::trng::regs::IntStatus;
-use nxp_pac::trng::vals::IntStatusEntVal;
 
 use crate::clocks::enable_and_reset;
 use crate::clocks::periph_helpers::NoConfig;
 use crate::interrupt::typelevel;
 use crate::interrupt::typelevel::Handler;
-use crate::pac::trng::vals::TrngEntCtl;
+use crate::pac::trng::regs::IntStatus;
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/trng.rs
```diff
@@ -437,15 +461,19 @@ pub enum Error {
 
     /// Buffer argument is invalid
     InvalidBuffer,
+    HardwareFail,
```
felipebalbi, line 466: missing rustdoc?

embassy-mcxa/src/trng.rs
```diff
@@ -351,12 +358,29 @@ impl<'d> Trng<'d, Async> {
...
+                    Some(Err(Error::HardwareFail))
+                } else if status.intg_flt() {
+                    Some(Err(Error::IntegrityError))
+                } else {
+                    Some(Err(Error::ErrorStatus))
+                }
             })
             .await
-            .map_err(|_| Error::ErrorStatus)
+            .map_err(|_| Error::ErrorStatus)?
```
felipebalbi: Is this one of those `Result<Result<T, E>, E>` cases? How about calling `flatten()` here?

embassy-mcxa/src/trng.rs
```diff
@@ -1,10 +1,13 @@
 //! True Random Number Generator
 
 use core::marker::PhantomData;
+use core::sync::atomic::{AtomicU32, Ordering};
 
 use embassy_hal_internal::Peri;
 use embassy_hal_internal::interrupt::InterruptExt;
 use maitake_sync::WaitCell;
+use nxp_pac::trng::regs::IntStatus;
+use nxp_pac::trng::vals::IntStatusEntVal;
```
felipebalbi:
```suggestion
use crate::pac::trng::regs::IntStatus;
use crate::pac::trng::vals::IntStatusEntVal;
```

PR #5552, opened 2026-03-01
Do not flush the wrong fifo

bugadani:
When disabling an OUT endpoint, the previous implementation flushed a TX FIFO, but TX FIFOs are assigned to IN endpoints.

I think this closes (maybe not, though?) #5521 as the PR fixes an identical-looking issue on the ESP32-S3, and the F405 uses a different version of the same IP.


kaspar030:
I confirm this fixes our issue on Linux where the esp32-s3 usb ethernet would not properly come up. With this fix, it works like a charm.

PR #5554, opened 2026-03-02
Make nxp-pac a non-optional build-dep in embassy-nxp.

wt:
The nxp-pac is use unconditionally in the build.rs. Because of that, it should not be an optional build dep.


lulf (approved):
Thanks!

PR #5561, opened 2026-03-02
[MCXA] WIP: Update with regenerated MCXA256 PAC

jamesmunns:


embassy-mcxa/src/lib.rs
```diff
@@ -29,7 +29,7 @@ mod mcxa2xx_exclusive {
     pub mod reset_reason;
     pub mod rtc;
     pub mod spi;
-    pub mod trng;
+    // pub mod trng;
```
jamesmunns, line 32: @felipebalbi I'll probably need help with this

jamesmunns:
Ah, this actually has changes from https://github.com/embassy-rs/nxp-pac/pull/45, I need to update the pac version

embassy-mcxa/src/lib.rs
felipebalbi: sure thing, do you want me to push trng update directly to your branch or send a PR?

PR #5562, opened 2026-03-02
[MCXA] add DMA functions for un-aligned and SW start transfers (SGI use)

alamfarjadf:
Two new public functions are added. 
'write_to_peripheral_mixed' allows bytes/half words to full word (u32) writes (will be useful for future symmetric SGI operations)
'write_to_peripheral_mixed_software_start' allows all the bytes to be transferred in one loop (used for filling the SHA2 AUTO FIFO in DMA mode.

### DIFF
--- a/embassy-mcxa/src/dma.rs
+++ b/embassy-mcxa/src/dma.rs
@@ -1136,7 +1136,6 @@ impl DmaChannel<'_> {
         let ssize = S::size();
         let dsize = D::size();
         let sbytes = ssize.bytes();
-        let dbytes = dsize.bytes();
         let total_bytes = buf.len() * sbytes;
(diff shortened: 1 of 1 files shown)


Wassasin:
In #5524 the DMA API and internals were deduplicated. That API should now be suitable to achieve the same as this PR. Let me know whether you agree!

alamfarjadf:
> In #5524 the DMA API and internals were deduplicated. That API should now be suitable to achieve the same as this PR. Let me know whether you agree!

Agreed. 5524 addresses the mismatched peripheral write as well as SW start. Built an example with #5524 and it is functioning as intended. This PR is now redundant and closing it.

Issue #5556, opened 2026-03-02
features for embassy-nxp


wt:
The embassy-nxp uses the `lpc55-core0` feature to enable the `lpc55s69_cm33_core0` feature in the nxp-pac. Considering that there are multiple lpc55 family boards. I would like the make change that feature name to `lpc55s69_core0`.

This will also make room for `lpc55s16`, which is my ultimate goal.

- [ ] Move feature `lpc55-core` to `lpc55s69-core0`.

wt:
@eva-cosma @i509VCB I think y'all are probably the best folks to be in this discussion. Are you okay with my proposal above? I'm happy to generate a PR with that change.

i509VCB:
I get the feeling that it should maybe be lpc55s69_cm33_core0

For lpc55s16 there is no 2nd core and requires no qualification of which core or type it is.

wt:
Sounds fine to me. I am happy to make that change. I'm going to start a checklist in the original comment.

@i509VCB Is there a reason to include the cm33. AFAIK, the s69 only comes in that variety. Could we shorten to `lpc55s69-core0`?

eva-cosma:
Do `lpc55s69-core0`, that's fine. Though I would encourage you to explore which features are shared and aren't shared between the s69 and s16

PR #5566, opened 2026-03-03
Add LPC55S16 support.

wt:
This is a a draft of my work to add lpc55s16 support. This currently depends on https://github.com/embassy-rs/nxp-pac/pull/30 to land and this embassy-nxp dependencies to be updated afterwards.

### DIFF
--- a/embassy-nxp/src/lib.rs
+++ b/embassy-nxp/src/lib.rs
@@ -154,10 +154,10 @@ pub fn init(_config: config::Config) -> Peripherals {
         pac::CCM.ccgr6().modify(|v| v.set_cg0(1));
     }
 
-    #[cfg(any(feature = "lpc55-core0", rt1xxx))]
+    #[cfg(any(lpc55, rt1xxx))]
     gpio::init();
 
-    #[cfg(feature = "lpc55-core0")]
+    #[cfg(lpc55)]
     {
         pint::init();
         pwm::Pwm::reset();
@@ -166,7 +166,7 @@ pub fn init(_config: config::Config) -> Peripherals {
     #[cfg(feature = "_time_driver")]
     time_driver::init();
 
-    #[cfg(feature = "lpc55-core0")]
+    #[cfg(lpc55)]
     dma::init();
 
     peripherals


wt:
With the landing of https://github.com/embassy-rs/nxp-pac/pull/30, this change is also ready to be reviewed for landing.

embassy-nxp/Cargo.toml
```diff
@@ -75,11 +77,13 @@ unstable-pac = []
 # This feature is unfortunately a hack around the fact that cfg_aliases cannot apply to the buildscript
 # that creates the aliases.
 _rt1xxx = []
+_lpc55 = []
 
 # A timer driver is enabled.
 _time_driver = ["dep:embassy-time-driver", "dep:embassy-time-queue-utils"]
 
 #! ### Chip selection features
 lpc55-core0 = ["nxp-pac/lpc55s69_cm33_core0"]
```
i509VCB, line 86: this should probably be `lpc55s69-core0`?

embassy-nxp/Cargo.toml
```diff
@@ -38,13 +40,13 @@ embassy-time-queue-utils = { version = "0.3.0", path = "../embassy-time-queue-ut
...
 ## Chip dependencies
-nxp-pac = { version = "0.1.0", optional = true, git = "https://github.com/i509VCB/nxp-pac", rev = "af5122e1cbe1483833c5d2e5af96b26a34ed5d62"}
+nxp-pac = { version = "0.1.0", optional = true, git = "https://github.com/i509VCB/nxp-pac", rev = "a7a6985f3402df08adfbe8fcc34fdb3cae003921"}
 
 imxrt-rt = { version = "0.1.7", optional = true, features = ["device"] }
 
 [build-dependencies]
 cfg_aliases = "0.2.1"
-nxp-pac = { version = "0.1.0", git = "https://github.com/i509VCB/nxp-pac", rev = "af5122e1cbe1483833c5d2e5af96b26a34ed5d62", features = ["metadata"] }
+nxp-pac = { version = "0.1.0", git = "https://github.com/i509VCB/nxp-pac", rev = "a7a6985f3402df08adfbe8fcc34fdb3cae003921", features = ["metadata"] }
```
i509VCB: probably use embassy link now

embassy-nxp/src/usart/lpc55.rs
```diff
@@ -929,6 +933,18 @@ macro_rules! impl_usart_rxd_pin {
     };
 }
 
+macro_rules! impl_usart_cts_pin {
```
i509VCB, line 979: `#[allow(unused)]` to these macros (or CI complains)

embassy-nxp/Cargo.toml
wt: I'm not renaming those in this change. That is what the #5556 was meant to capture. I don't see that as related to this change.

embassy-nxp/Cargo.toml
wt: I don't know what this means. Where do I find info? Is this something that should happen with a separate PR?

embassy-nxp/src/usart/lpc55.rs
wt: I will look over these.

wt: `#[allow(unused)]` is inappropriate here. These are intended to be used for any board that has CTS/RTS/SCK pins defined in their metadata on the pac. I will add `#[cfg(feature = "lpc55s16")]` instead so that they are only enabled for lpc55s16 until there are other boards that need it.

wt: Instead of using the `#[cfg(feature = "lpc55s16")]`, I added some magic to the build.rs to automatically enable those macros when the board has USART CTS/RTS/SCK pins defined in the metadata. This means that when the nxp-pac dep is updated, we automatically get CTS/RTS/SCK macros enabled.

wt:
I moved this back to draft as I found a bug in nxp-pac that prevents this change from working. See https://github.com/embassy-rs/nxp-pac/pull/56

wt:
The nxp-pac change was landed and I updated this change to use that new commit id, so it's ready for review again.

embassy-nxp/Cargo.toml
wt: Oh, you mean to change the org of the repo. Done.

embassy-nxp/Cargo.toml
i509VCB: ok

i509VCB (approved):

PR #5569, opened 2026-03-03
[MCXA] Unlock clocks in `init`

jamesmunns:
Contrary to the datasheet, the `syscon->clkunlock` register was set to "locked" on boot. Manually unlock it to allow writes to MRCC configuration registers.

It's possible the rom bootloader is setting this.


diondokter (approved):
Lol

PR #5583, opened 2026-03-03
Use inverted duty cycle for NRF54L15 PWM

### LINKED ISSUE #5582: Flashing PWM example to NRF54L15-DK leaves LEDs solid on with no apparent breathing
This is because those LEDs are connected in an open-drain configuration.  The PWM polarity needs to be inverted.  I will be sending a PR.

whmountains:
Closes #5582


lulf (approved):
Thanks!

PR #5589, opened 2026-03-04
embassy-usb-driver: remove dependency on `embedded-io-async`

yodaldevoid:
`EndpointError` is no longer being used as the error type for any `embedded-io-async` impls in `embedded-usb` which was the original reason for the dependency. It has been previously expressed that we would like to remove this dependency so that `embassy-usb-driver` doesn't need to be released each time `embedded-io-async` updates.

### DIFF
--- a/embassy-usb-driver/CHANGELOG.md
+++ b/embassy-usb-driver/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased - ReleaseDate
 
 - Add `EndpointOut::read_data()` and `EndpointIn::write_data()` provided methods.
+- Remove dependency on `embedded_io_async`
 
 ## 0.2.0 - 2025-07-16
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
The point of not depending on embedded-io is to avoid breaking changes, but removing it is *also* a breaking change. This is why I left it as a TODO. We should leave it there until we have an actual reason to make a breaking change(e.g. an actual design deficiency in the traits that can't be fixed backwards-compatibly) and release 0.3.

flaviojs:
Instead of removing you could make it optional with a feature, so users of `embedded-io-async` and `embassy` can use the implementation. (you can't implement for types of other crates)

According to [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) you can use prefix `_`, `unstable-`, or `unstable_` to mark a feature private, meaning it is not included in semver guarantees (changing the code is never a breaking change).

Dirbaio:
> Instead of removing you could make it optional with a feature

This is also a breaking change. 

*When* we do a breaking change we'll just remove the embedded-io dep. The problem is we simply want to delay doing a breaking change as much as possible, and *just* removing it is not worth doing the breaking change.

PR #5586, opened 2026-03-04
[MCXA5] Add support for ROSC/OSC32K

jamesmunns:
This is an initial attempt to enable OSC32K/ROSC. Experimentally, this *seems* to work on the FRDM-MCXA577 board, there are still some TODOs waiting for feedback from NXP on.

PR #5590, opened 2026-03-04
[MCXA5] Enable CTIMER

felipebalbi:
Pins pulled directly from `Pinout.xlsx` attached to the reference manual.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/168

### DIFF
--- a/embassy-mcxa/src/clocks/periph_helpers/mod.rs
+++ b/embassy-mcxa/src/clocks/periph_helpers/mod.rs
@@ -291,7 +291,6 @@ impl SPConfHelper for OsTimerConfig {
 pub enum CTimerClockSel {
     /// FRO12M/FRO_LF/SIRC clock source, passed through divider
     /// "fro_lf_div"
-    #[cfg(feature = "mcxa2xx")]
     FroLfDiv,
     /// FRO180M/FRO_HF/FIRC clock source, passed through divider
     /// "fro_hf_div"
@@ -357,10 +356,16 @@ impl SPConfHelper for CTimerConfig {
         };
 
         let (freq, variant) = match self.source {
-            #[cfg(feature = "mcxa2xx")]
             CTimerClockSel::FroLfDiv => {
                 let freq = clocks.ensure_fro_lf_div_active(&self.power)?;
-                (freq, CtimerClkselMux::CLKROOT_FUNC_0)
+
+                // TODO: fix PAC names for consistency
+                #[cfg(feature = "mcxa2xx")]
+                let mux = CtimerClkselMux::CLKROOT_FUNC_0;
+                #[cfg(feature = "mcxa5xx")]
+                let mux = CtimerClkselMux::I0_CLKROOT_SIRC_DIV;
+
+                (freq, mux)
             }
             #[cfg(feature = "mcxa2xx")]
             CTimerClockSel::FroHfDiv => {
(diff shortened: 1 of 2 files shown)


embassy-mcxa/src/clocks/periph_helpers/mod.rs
```diff
@@ -263,3 +264,157 @@ impl SPConfHelper for OsTimerConfig {
...
+    /// "fro_hf_div"
+    #[cfg(feature = "mcxa2xx")]
+    FroHfDiv,
+    /// SOSC/XTAL/EXTAL clock source
+    #[cfg(feature = "mcxa2xx")]
+    #[cfg(not(feature = "sosc-as-gpio"))]
+    ClkIn,
+    /// FRO16K/clk_16k source
+    #[cfg(feature = "mcxa2xx")]
+    Clk16K,
```
jamesmunns, line 305: If #5586 lands first we can un-gate this

embassy-mcxa/src/clocks/periph_helpers/mod.rs
```diff
@@ -263,3 +264,157 @@ impl SPConfHelper for OsTimerConfig {
...
+//
+// CTimer
+//
+
+/// Selectable clocks for `CTimer` peripherals
+#[derive(Debug, Clone, Copy)]
+pub enum CTimerClockSel {
+    /// FRO12M/FRO_LF/SIRC clock source, passed through divider
+    /// "fro_lf_div"
+    #[cfg(feature = "mcxa2xx")]
```
jamesmunns: We should have frolf now, that's based on the FRO12M clock

embassy-mcxa/src/clocks/periph_helpers/mod.rs
```diff
@@ -263,3 +264,157 @@ impl SPConfHelper for OsTimerConfig {
...
+
+        let div = self.div.into_divisor();
+        let expected = freq / div;
+
+        // 22.3.2 peripheral clock max functional clock limits
+        let power = match self.power {
+            PoweredClock::NormalEnabledDeepSleepDisabled => clocks.active_power,
+            PoweredClock::AlwaysEnabled => clocks.lp_power,
+        };
+        #[cfg(feature = "mcxa2xx")]
```
jamesmunns, line 421: I wonder if we should store these somewhere chip specific instead of splatting them across all the drivers. Doesn't have to be fixed this PR, but I'm open to ideas here.

embassy-mcxa/src/ctimer/mod.rs
```diff
@@ -69,7 +68,10 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             power: PoweredClock::NormalEnabledDeepSleepDisabled,
+            #[cfg(feature = "mcxa2xx")]
```
jamesmunns: Same note about frolf

embassy-mcxa/src/clocks/periph_helpers/mod.rs
felipebalbi: nice, fixed.

embassy-mcxa/src/clocks/periph_helpers/mod.rs
felipebalbi: we should wait until that goes in :-)

embassy-mcxa/src/ctimer/mod.rs
felipebalbi: fixed

embassy-mcxa/src/clocks/periph_helpers/mod.rs
felipebalbi: I guess we could add it to `chips/mcxa*.rs` with a function generic over `Gate`?

PR #5593, opened 2026-03-04
[MCXA] Enable PORT5 and GPIO5 clocks

felipebalbi:
Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/176


jamesmunns:
Note: we pulled this because we realized the interrupt handler for GPIOs needs to be fixed, and are maybe discussing de-magicking the GPIOs and moving to Instances and Irq tokens like other HAL drivers.

felipebalbi:
> Note: we pulled this because we realized the interrupt handler for GPIOs needs to be fixed, and are maybe discussing de-magicking the GPIOs and moving to Instances and Irq tokens like other HAL drivers.

Patched it in with the magic still in place. Moving to instances will require more careful consideration. If we move to `Instance` now, we will have to patch every GPIO use:

```diff
- let pin = Input::new(P0_0, //...);
+ let pin = Input::new(GPIO0, P0_0, //...);
```

That's because we need the `T: Instance` to grab the data we need from `T`. I'll need a couple more days thinking about this before we can move forward with `Instance`s. Meanwhile, GPIO5 is actually working now. Verified `Input`, `Output`, and their interrupts.

PR #5594, opened 2026-03-04
[MCXA5] Enable DMA

felipebalbi:
Had to add `#![allow(dead_code)]` to the dma driver until the drivers using DMA are enabled again, otherwise we would end up with a lot of build warnings on MCXA5. Subsequent PRs shall fix that.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/174

### DIFF
--- a/embassy-mcxa/src/chips/mcxa5xx.rs
+++ b/embassy-mcxa/src/chips/mcxa5xx.rs
@@ -69,16 +69,25 @@ mod inner_periph {
         CTIMER4_CH3,
 
         // DBGMAILBOX,
-        // DMA0,
-        // DMA_CH0,
-        // DMA_CH1,
-        // DMA_CH2,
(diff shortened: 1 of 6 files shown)


embassy-mcxa/src/chips/mcxa5xx.rs
```diff
@@ -69,16 +69,25 @@ mod inner_periph {
...
+        DMA_CH5,
+        DMA_CH6,
+        DMA_CH7,
+        // Need more work on the DMA driver before we can enable these
+        // DMA_CH8,
+        // DMA_CH9,
+        // DMA_CH10,
+        // DMA_CH11,
+        EDMA0_TCD0,
+        // Need more work on the DMA driver before we can enable this
```
jamesmunns, line 88: Ah, you did leave a note, 👍

PR #5604, opened 2026-03-05
fix: stm32/adc/g4: correct VREF_CALIB_MV

yodaldevoid:
Based on the STM32G491CC datasheet (DS13122 rev 4) section 3.18.2, the VREFINT was calibrated at 3.0 V.


xoviat:
bender run

PR #5605, opened 2026-03-05
fix: stm32/adc/v3: correct VREF_CALIB_MV for H5 and H7RS chips

yodaldevoid:
See the following datasheets:
- STM32H533xx datasheet (DS14539 rev 1) section 5.3.5
- STM32H7S3A8 datasheet (DS14359 rev 6) section 6.3.6


yodaldevoid:
The WB family of chips, which uses the v3 ADC, is still incorrect after this patch. See https://github.com/embassy-rs/embassy/issues/5606.

xoviat:
bender run

yodaldevoid:
Are those build errors spurious? I don't see any way for these changes to have broken anything unless a specific ADC test is looking for a specific value.

PR #5608, opened 2026-03-05
change: stm32/timer/input_capture: use timer word size for all outputs

yodaldevoid:
The input capture driver was written before the word size was available as a type for all timers.


xoviat:
@Dirbaio wanted to eventually type-erase the high-level timers, which would break this. Maybe for now this is fine.

yodaldevoid:
I would argue that timers should still know their word size, even if the exact peripheral is type erased. Without knowing the word size you can't properly account for timer roll over.

xoviat:
bender run

PR #5612, opened 2026-03-06
[MCXA] make SGI drivers compatible with embassy-rs peripheral style

alamfarjadf:
Add hardware accelerated hashing support for MCXA family with SGI (Secure Generic Interface) support.

Currently supports three modes: 1) blocking, 2) Streaming multi-part and 3) async via DMA.

Only supports CNSA 2.0 compliant hash sizes (i.e. 384 and 512).


embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+
+// SGI_SHA2_CTRL register bitfields
+pub const SGI_SHA2_CTRL_MODE_AUTO: u32 = 1 << 1;
+pub const SGI_SHA2_CTRL_MODE_NORMAL: u32 = 0 << 1;
+pub const SGI_SHA2_CTRL_SIZE_SHA384: u32 = 2 << 2;
+pub const SGI_SHA2_CTRL_SIZE_SHA512: u32 = 3 << 2;
+pub const SGI_SHA2_CTRL_HASH_RELOAD: u32 = 1 << 13;
+pub const SGI_SHA2_CTRL_NO_HASH_RELOAD: u32 = 0 << 13;
+pub const SGI_SHA2_CTRL_NO_AUTO_INIT: u32 = 1 << 15;
+pub const SGI_SHA2_CTRL_AUTO_INIT: u32 = 0 << 15;
```
diondokter: This should come from the PAC, not done manually.
And if this really is necessary, it shouldn't be pub

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
+// Hash functionality using SGI hardware
+
+use super::sgi::{SGIError, Sgi};
+
+// ============================================================================
+// Public constants re-exported by `crate::sgi`
+//
+// These are used as self-documenting `HashOptions` field values.
+// ============================================================================
+
```
diondokter: These type of comments are unwanted. They get out of date fast during normal refactoring

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+/// Options for SGI hash command
+#[derive(Debug, Clone, Copy)]
+pub struct HashOptions {
+    pub hash_mode: u32,       // SHA-384, SHA-512, etc.
+    pub op_mode: u32,         // 0 = Normal, 1 = Automatic
+    pub init: u32,            // Initialize hash
+    pub reload: u32,          // Load hash
+    pub output_enable: bool,  // Enable output
+    pub byte_order_big: bool, // Big endian byte order
```
diondokter: This is public facing. Use enums and other types to make user's life more easy and prevent mistakes.
Also use doc comments so the user can read the docs too

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+// ============================================================================
+
+/// A non-blocking SGI SHA2 DMA operation context.
+///
+/// This handle borrows the SGI instance for the duration of the operation.
+///
+/// SGI singleton ownership is enforced by the `Sgi` constructor consuming the
+/// Embassy `Peri` token.
+pub struct DmaHasher<'a, 'd> {
+    sgi_instance: &'a mut Sgi<'d>,
```
diondokter: This should keep using the `Peri` construct, just like the comments say

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+
+/// Start an SGI SHA2 DMA hash operation and return a context handle.
+/// This configures the SGI engine and software-starts the eDMA channel, then
+/// returns immediately. Completion is awaited via [`DmaHasher::finalize`].
+pub fn hash_dma_start<'a, 'd>(
+    sgi: &'a mut Sgi<'d>,
+    dma_ch: &'a mut DmaChannel<'d>,
+    hash_size: HashSize,
+    input: &[u8],
+) -> Result<DmaHasher<'a, 'd>, SGIError> {
```
diondokter: This should be an associated function for `DmaHasher`

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+
+        Ok(())
+    }
+}
+
+/// Start an SGI SHA2 DMA hash operation and return a context handle.
+/// This configures the SGI engine and software-starts the eDMA channel, then
+/// returns immediately. Completion is awaited via [`DmaHasher::finalize`].
+pub fn hash_dma_start<'a, 'd>(
+    sgi: &'a mut Sgi<'d>,
```
diondokter: Use `Peri`

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+            prev_result: [0xffu8; SGI_HASH_OUTPUT_SIZE as usize],
+            processed_bytes: dma_len_bytes,
+            total_input_len: input.len(),
+        })
+    }
+}
+
+/// Blocking hash function for input data of any size up to 512 bytes. This internally handles padding and block processing, and returns the final hash output in `hash_result`.
+pub fn hash_blocking<'d>(
+    sgi: &mut Sgi<'d>,
```
diondokter: Use Peri

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+    hash_size: HashSize,
+    input: &[u8],
+    hash_result: &mut [u8],
+) -> Result<(), SGIError> {
```
diondokter: Drivers are always an instance of a struct. Please do that here as well.

Target API should be something like:
```rust
let mut hasher = Hasher::new(p.sgi);
hasher.hash_blocking(...);
```

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+
+    // Add 128-bit length in bits (big-endian) to last 16 bytes
+    let bit_len = (input.len() as u128) * 8;
+    let len_offset = padded_len - 16;
+    buffer[len_offset..len_offset + 16].copy_from_slice(&bit_len.to_be_bytes());
+
+    Some(padded_len)
+}
+
+fn process_multi_block_update<'d>(hasher: &mut SGIHasher, sgi: &mut Sgi<'d>, input: &[u8]) -> Result<(), SGIError> {
```
diondokter: Use Peri for Sgi. But ideally, see comment above about what a driver should look like

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+    }
+    Ok(())
+}
+
+pub struct SGIHasher {
+    ctx: SgiShaCtx,
+}
+
+impl SGIHasher {
+    pub fn new() -> Self {
```
diondokter: This should take a `Peri<'d, Sgi>`

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+        self.ctx.options.op_mode = mode.to_sgi_op_mode();
+        self.ctx.options.init = SGI_SHA2_CTRL_AUTO_INIT;
+        self.ctx.options.reload = SGI_SHA2_CTRL_NO_HASH_RELOAD;
+        self.ctx.first_block = true;
+        Ok(())
+    }
+
+    /// Update the hash state with the provided input data. This can be called multiple times to process streaming data. Input limit per call is maximum 512 bytes,
+    /// but it's recommended to keep it to 128 bytes (one block) for better performance and to avoid auto mode FIFO filling which can be slower.
+    pub fn update<'d>(&mut self, sgi: &mut Sgi<'d>, input: &[u8]) -> Result<(), SGIError> {
```
diondokter: The Sgi should be part of the driver instance, not passed in at every function

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+}
+
+impl SGIHasher {
+    pub fn new() -> Self {
+        Self { ctx: SgiShaCtx::new() }
+    }
+
+    /// Initialize the hasher with the specified hash size and mode. This must be called before any update() calls.
+    /// Leave option as None (unless there are very specific performance/memory concerns).
+    pub fn init(&mut self, hash_size: HashSize, hash_mode: Option<HashMode>) -> Result<(), SGIError> {
```
diondokter: The new function should init the driver already. Now the user can forget to init and have a bad time

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+        }
+
+        let final_block_len = calculate_padded_length(self.ctx.total_len) - self.ctx.processed_len; // Calculate how many bytes are in the final block (including padding)
+
+        if remaining_data_len > final_block_len {
+            return Err(SGIError::InvalidSize); // Remaining data can't exceed the final block length, otherwise we would need to process another block before finalizing
+        }
+
+        #[cfg(feature = "defmt")]
+        defmt::info!(
```
diondokter: Don't use info. That's way too spammy. Use trace

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+        };
+
+        let remaining_final_block_len = if self.ctx.options.op_mode == SGI_SHA2_CTRL_MODE_NORMAL {
+            MAX_BLOCK_SIZE
+        } else {
+            final_block_len
+        }; // In AUTO mode we process the entire final block at once, so we don't have any remaining data to process after the first pass.
+
+        #[cfg(feature = "defmt")]
+        defmt::info!(
```
diondokter: Same, use trace

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,635 @@
...
+
+        sgi.read_hash_output(self.ctx.options, hash_result)?;
+        Ok(())
+    }
+}
+
+/// The following is a wrapper for SHA-512 ROTKH calculation which always returns 48 bytes by truncating the full SHA-512 digest.
+/// Hardware SHA-512 RKTH - always returns 48 bytes for RKTH calculation
+/// Hardware SHA-512 RKTH - always returns 48 bytes for RKTH calculation.
+pub fn hsm_sha512_rkth<'d>(sgi: &mut Sgi<'d>, input: &[u8]) -> Option<[u8; 48]> {
```
diondokter: Same as before, this should be part of a driver struct

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+impl Instance for peripherals::SGI0 {}
+
+// ============================================================================
+// SGI0 (Secure Generic Interface) constants
+// ============================================================================
+
+// CTRL field values.
+const CRYPTO_OP_SHA2: u8 = 4;
+const DATOUT_RES_TRIGGER_UP: u8 = 2;
+
```
diondokter: Use the PAC

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+            irq_enabled,
+            _wake_guard: wake_guard,
+        })
+    }
+
+    /// Helper to check if SGI IRQ is enabled.
+    #[inline(always)]
+    pub fn irq_enabled(&self) -> bool {
+        self.irq_enabled
+    }
```
diondokter: In embassy philosophy, users shouldn't care about interrupts unless there are really good reasons to. Are there here?

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+// ============================================================================
+// SGI structs and helpers
+
+/// SGI driver configuration.
+#[derive(Debug, Clone, Copy)]
+pub struct Config {
+    /// Enable the SGI interrupt in NVIC.
+    ///
+    /// The driver currently polls for completion, so this is optional, but can be enabled for async/DMA flows to avoid busy-waiting on the completion bit.
+    pub enable_interrupt: bool,
```
diondokter: This functionality should be type-stated. (async vs blocking version of the driver)

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+    /// Enables DMA handshake for datain/dataout registers.
+    pub fn set_auto_dma_handshake(&mut self, input_fifo_enable: bool, output_fifo_enable: Option<bool>) {
+        let output_enable = output_fifo_enable.unwrap_or(false); // Default to false if not provided
+
+        // Reserved bits are written as 0.
+        self.periph.sgi_auto_dma_ctrl().write(|w| {
+            w.set_ife(input_fifo_enable);
+            w.set_ofe(output_enable);
+        });
+    }
```
diondokter, line 249: Why is this pub? Seems like something the user shouldn't care about

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        // The hardware uses an inverted convention here in our existing driver:
+        // - `byte_order_big = true` => clear BYTES_ORDER
+        // - `byte_order_big = false` => set BYTES_ORDER
+        self.periph
+            .sgi_ctrl2()
+            .modify(|w| w.set_bytes_order(((!byte_order_big) as u8).into()));
+        Ok(())
+    }
```
diondokter, line 298: Seems like this should be a config given at init

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        });
+        Ok(())
+    }
+
+    /// Configures byte order for SGI IO.
+    pub fn sgi_byte_order_big(&self, byte_order_big: bool) -> Result<(), SGIError> {
+        if self.is_busy() || self.is_sha2_busy() {
+            return Err(SGIError::Busy);
+        }
+
```
diondokter, line 289: The driver should be structured so this error will never be encountered by the user.
Make sure all operations are either blocking or async.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        // Reserved bits are written as 0.
+        self.periph.sgi_auto_dma_ctrl().write(|w| {
+            w.set_ife(input_fifo_enable);
+            w.set_ofe(output_enable);
+        });
+    }
+
+    /// Configures the appropriate hash operation parameters (e.g. auto vs. normal mode, hash size, byte order) and enables the hash engine.
+    pub fn init_sgi_sha(&mut self, options: HashOptions) -> Result<(), SGIError> {
```
diondokter: This should be part of the config in the `new` constructor

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+            self.fill_sha2_fifo_normal(data)?;
+        }
+        // Keep as a single write to the CTRL register.
+        self.periph.sgi_ctrl().write(|w| {
+            // crypto_op selects the kernel. For SHA2 this is 0b100 (4).
+            w.set_crypto_op(CRYPTO_OP_SHA2.into());
+            w.set_start((true as u8).into());
+        });
+        Ok(())
+    }
```
diondokter, line 315: Don't just start a hash. Have 2 functions, one async and one blocking that wait for the hash to be done

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+            28 => self.periph.sgi_key3a().write(|w| w.set_key3a(word)),
+            29 => self.periph.sgi_key3b().write(|w| w.set_key3b(word)),
+            30 => self.periph.sgi_key3c().write(|w| w.set_key3c(word)),
+            31 => self.periph.sgi_key3d().write(|w| w.set_key3d(word)),
+            _ => {
+                // Caller bounds word_index.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        }
+    }
```
diondokter, line 338: Send a PR to nxp-pac to convert these registers into an array of registers so we can do direct indexing

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+
+    /// CPU driven FIFO filling for normal mode. Caller is responsible for starting the operation after filling the FIFO, and ensuring data size does not exceed FIFO capacity.
+    pub fn fill_sha2_fifo_normal(&mut self, data: &[u8]) -> Result<(), SGIError> {
```
diondokter: Probably shouldn't be pub

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        }
+        for (i, chunk) in data.chunks_exact(4).enumerate() {
+            let word = u32::from_be_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]);
+            self.write_fifo_word_normal(i, word);
+        }
+        Ok(())
+    }
+
+    /// Wrapper for normal/auto mode FIFO filling. Normal mode requires FIFO be filled prior to SHA2 operation start.
+    pub fn fill_sha2_fifo(&mut self, options: HashOptions, data: &[u8], length: usize) -> Result<(), SGIError> {
```
diondokter: I'm gonna skip the next couple of functions but the same comment applies: This doesn't look like an API the user should use. So make it private

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+            } else {
+                dma_ch.setup_write_to_peripheral(&data[..length], peri_address, software_start, options)
+            }
+        }
+        .map_err(|_| SGIError::InvalidSize)?;
+
+        let transfer = Transfer::new(dma_ch.reborrow());
+
+        #[cfg(feature = "defmt")]
+        defmt::info!("DMA started for {=usize} bytes (software-start, DSIZE=32)", length);
```
diondokter: Trace, not info

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        Ok(transfer)
+    }
+
+    /// Checks for SGI peripheral errors (SHA2, key read, key unwrap). This should be called after checking that the operation is done,
+    /// since some errors only get latched/reported once busy is de-asserted. Clears any errors after reading.
+    pub fn sgi_operation_error(&self) -> Result<(), SGIError> {
```
diondokter: Really think about the API you're presenting to the user.

If this needs to be called every time after an operation is done, then the code should be written so it either happens automatically or the user has no choice not to call this.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+    /// SGI status busy bit indicates the peripheral is currently processing data and cannot accept new commands or data.
+    /// This is independent from the SHA2 busy bit, which specifically indicates the hash engine is active.
+    pub fn is_busy(&self) -> bool {
+        self.status().busy()
+    }
+
+    /// Checks if the SHA2 engine is busy processing data. This is independent from the BUSY bit, which may be de-asserted in error conditions.
+    pub fn is_sha2_busy(&self) -> bool {
+        self.status().sha2_busy()
+    }
```
diondokter, line 469: The user shouldn't care. The driver should just work, see earlier review comments

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        let start_time = Instant::now();
+        const MAX_WAIT_TIME_USEC: u64 = 1000; // 1 millisecond timeout for normal SHA2 operations.
+        while self.is_sha2_busy() {
+            if is_expired(start_time, MAX_WAIT_TIME_USEC) {
+                // 1 millisecond timeout
+                #[cfg(feature = "defmt")]
+                defmt::error!("Timeout waiting for SGI SHA2 operation to complete");
+                return Err(SGIError::Timeout);
+            }
+            core::hint::spin_loop();
```
diondokter: This doesn't really do anything on embedded. Can be removed

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        ];
+
+        for (i, word) in words.into_iter().take(TOTAL_DATAOUT_REGS).enumerate() {
+            output[idx + (i * 4)..idx + (i * 4) + 4].copy_from_slice(&word.to_be_bytes());
+        }
+        Ok(())
+    }
+
+    ///Checks if PDONE interrupt is set, which indicates SHA2 operation completion. Also checks for errors.
+    pub fn interrupt_is_operation_done(&self) -> Result<bool, SGIError> {
```
diondokter: Interrupt handling should be done by the driver, not the user

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+
+    /// Await SHA2 completion using the SGI operation-done interrupt.
+    ///
+    /// This is intended for async/DMA flows, to avoid busy-spinning on `sha2_busy`.
+    ///
+    /// Notes:
+    /// - NVIC enable is controlled by `Config.enable_interrupt` in `Sgi::new`.
+    /// - Hardware error paths may deassert busy without generating an interrupt; this
+    ///   function treats that as completion and returns the decoded error.
+    pub async fn wait_sha2_complete_irq(&mut self) -> Result<(), SGIError> {
```
diondokter: Starting and waiting for completion should be one async function

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+        // If the SGI NVIC line is disabled, the waiter will never be woken by an ISR.
+        // Fall back to a bounded busy-bit wait to avoid hanging forever.
+        if !self.irq_enabled {
+            self.wait_until_sha2_not_busy()?;
+            self.sgi_operation_error()?;
+            self.sgi_error()?;
+            return Ok(());
+        }
```
diondokter: Async should not block, especially not unexpectedly. This should be an error

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+    /// Issue stop command to halt the SHA2 operation.
+    pub fn sgi_stop_sha2_cmd(&self) {
+        self.periph
+            .sgi_sha2_ctrl()
+            .modify(|w| w.set_sha2_stop((true as u8).into()));
+    }
+
+    /// Reads the hash output from the SGI dataout registers into the provided buffer. This should be called after confirming operation completion,
+    /// and it checks for errors before reading. It handles both normal and auto mode completion, including triggering additional reads for the larger SHA-512 outputs.
+    pub fn read_hash_output(&mut self, options: HashOptions, output: &mut [u8]) -> Result<(), SGIError> {
```
diondokter: When would you want to do a hash and then not read the output?

I assume the answer is never and if I'm right, then the API should reflect that.
The api should do start, wait for completion and then read the output in one go.

diondokter (changes requested):

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: These were left over from older non-embassy style code. Cleaned up now.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Removed.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Now uses Enums to make intent more clear and easier to use.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Now holds a `peri` construct

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: It's now an `impl` for `DmaHasher.` Also, it no longer returns a `DmaHasher` instance. Instead, it's now called `start_and_finalize` and completes the whole operation start to finish in an async maner.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Updated. New `BlockingHasher` struct added.

`let mut blocking_hasher = BlockingHasher::new(p.SGI0.reborrow());`

And,

`blocking_hasher.hash_blocking(HashSize::Sha384, &input_data[..308], &mut hash_result)`

See examples/mcxa5xx/bin/hasher.rs

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Updated, now takes in `Peri<'d, SGI0>`

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Not for streaming mode. The idea for streaming mode with SGI is that a localized `Ctx` can be maintained with a "running" hash of whatever the input may be. So, when something like an encrypted firmware payload is getting validated, something like `AesInstance` (not implemented yet) decrypts and then `StreamingHasher` updates the chunk hash.

If streaming mode holds the `peri` internally, not sure how another secure instance can use it without dropping the instance first (e.g. a MAC or AES). Since there is only one SGI, it needs to be managed by some common thread. Open to ideas.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Agreed, no more `init()`, `new()` takes the necessary inputs and constructs the instance.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Changed all debug prints to trace.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: It's now part of `BlockingHasher`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: The PAC takes those configurations as direct `u8` values, not configurable individually. Updated as an Enum to make all the options clear.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: No, they don't need to. This is used internally and has been made private.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: There are two versions now, `new()` and `new_blocking()`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: It's now `pub(crate)`, to be used in future with AES. User doesn't care currently (or in future).

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: It's initialized as part of `DefaultOptions` for hashing as `ByteOrder::BigEndian`, unless it's DMA mode. Then it's configured automatically as `ByteOrder::LittleEndian`, then switched to `ByteOrder::BigEndian` for reading the result.

Either way, user does not need to configure this, this is a `pub(crate)` now.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: This is leftover old code, removed.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: This controls the configuration of the SHA engine, which changes depending on mode/size/init/reload, which isn't always known at construction time (e.g. whether the finalize only had less than one block or had already processed muti-block through the engine).

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Yes, there are 3 modes: blocking, streaming and DMA (see examples/mcxa5xx/src/bin/hasher.rs) User wouldn't use this. Changed to `pub(crate)`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Opened issue with nxp-pac https://github.com/embassy-rs/nxp-pac/issues/55

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Changed to private.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Removed.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: It is done by the driver, see examples.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Changed to one async function to finalize the hash and update results.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Good catch! Not sure why I added this. Removed. Now returns `InterruptNotEnabled` in this case.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Two cases: 1) The operation errored out. or 2) You are updating in streaming mode, so you only read back the registers and don't publish full result.

The hashing APIs do, see examples. The user does not use this; this is used by the DMA/Blocking/Streaming hash drivers.

This is now `pub(crate)`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Changed to private.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: The user does not call this, this is used by the hashing drivers themselves to interface with SGI itself. See examples.

Changed to `pub(crate)`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: User does not care, used internally by hash driver itself (and in future AES and MAC).

Changed to `pub(crate)`

embassy-mcxa/src/lib.rs
```diff
@@ -39,6 +39,8 @@ mod all_chips {
     #[cfg(feature = "executor-platform")]
     pub mod executor;
     pub mod gpio;
+    #[path = "sgi/hash.rs"]
+    pub mod hash;
```
felipebalbi: This should be unnecessary. User can `use hal::sgi::hash` directly.

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+const SHA384_DIGEST_LEN: usize = 48;
+const SHA512_DIGEST_LEN: usize = 64;
+
+#[inline(always)]
+fn required_digest_len(hash_size: HashSize) -> Result<usize, SGIError> {
+    match hash_size {
+        HashSize::Sha384 => Ok(SHA384_DIGEST_LEN),
+        HashSize::Sha512 => Ok(SHA512_DIGEST_LEN),
+    }
+}
```
felipebalbi: This looks like it could be a `From` implementation.

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+pub enum HashReload {
+    NoReload,
+    Reload,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ByteOrder {
+    BigEndian,
+    LittleEndian,
+}
```
felipebalbi, line 65: We need rustdoc comments on these.

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+                // block is CPU-fed below and switches back to big-endian explicitly.
+                byte_order: ByteOrder::LittleEndian,
+            },
+            remainder: [0u8; MAX_BLOCK_SIZE],
+            prev_result: [0xffu8; SGI_HASH_OUTPUT_SIZE as usize],
+            processed_bytes: 0,
+            total_input_len: input.len(),
+        };
+
+        let digest_len = required_digest_len(state.options.hash_size)?;
```
felipebalbi:
```rust
let digest_len = state.options.hash_size.into();
```

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+            return Err(SGIError::BufferTooSmall);
+        }
+
+        let mut sgi = Sgi::new(
+            state.peri.reborrow(),
+            SgiConfig {
+                interrupt: SgiInterrupt::AsyncEnabled,
+            },
+        )
+        .map_err(|_| SGIError::HardwareError)?;
```
felipebalbi: I think this should be left for the user. The user should be the one taking ownership of the `Sgi` and passing it to the hasher.

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+            sgi.init_sgi_sha(state.options)?;
+
+            // Arm operation-done interrupt before starting the hash operation.
+            sgi.enable_operation_done_interrupt();
+            sgi.start_sgi_hash(state.options, dma_input)?;
+            state.transfer = Some(sgi.fill_sha2_fifo_dma_start(dma_ch, dma_input, dma_len_bytes)?);
+
+            // After starting, subsequent operations must not auto-init unless re-init'd.
+            state.options.init = HashInit::NoInit;
+        }
```
felipebalbi: This looks like it could be using `chunks_exact(MAX_BLOCK_SIZE)`.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+}
+
+#[interrupt]
+fn SGI() {
+    unsafe { on_interrupt() }
+}
+
+/// SGI Error types
+#[derive(Debug, Clone, Copy)]
+pub enum SGIError {
```
felipebalbi: This should be named `SgiError` instead.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+        SGI_OP_DONE.store(true, Ordering::Release);
+    }
+    sgi.sgi_int_status_clr().write(|w| w.set_int_clr(true));
+    SGI_WAKER.wake();
+}
+
+#[interrupt]
+fn SGI() {
+    unsafe { on_interrupt() }
+}
```
felipebalbi: We probably want interrupts to be explicitly bound, instead.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+/// Errors exclusive to SGI peripheral initialization.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum SetupError {
+    /// Clock configuration error.
+    ClockSetup(ClockError),
+}
+
+static SGI_OP_DONE: AtomicBool = AtomicBool::new(false);
+static SGI_WAKER: AtomicWaker = AtomicWaker::new();
```
felipebalbi: We don't want `AtomicWaker`s here. Use a `WaitCell` instead.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+
+unsafe fn on_interrupt() {
+    let sgi = crate::pac::SGI0;
```
felipebalbi: You need an `Instance` abstracting this.

embassy-mcxa/src/sgi/hash.rs
```diff
@@ -0,0 +1,693 @@
...
+}
+
+pub struct StreamingHasher {
+    ctx: SgiShaCtx,
+}
+
+impl StreamingHasher {
+    /// Create and initialize a streaming hasher with the specified hash size and mode.
+    /// Leave `hash_mode` as `None` unless there are very specific performance or memory concerns.
+    pub fn new(hash_size: HashSize, hash_mode: Option<HashMode>) -> Result<Self, SGIError> {
```
felipebalbi: You should take the peripheral here, instead of expecting the user to pass the SGI on every other call.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+    HardwareError,
+    DmaError,
+    HashingNotComplete,
+    InterruptNotEnabled,
+}
+
+/// SGI (Secure Generic Interface) hardware abstraction
+pub struct Sgi<'d> {
+    _peri: Peri<'d, peripherals::SGI0>,
+    periph: pac_sgi::Sgi,
```
felipebalbi: This should be recovered through an `Instance` trait.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+    InvalidSize,
+    Timeout,
+    HardwareError,
+    DmaError,
+    HashingNotComplete,
+    InterruptNotEnabled,
+}
+
+/// SGI (Secure Generic Interface) hardware abstraction
+pub struct Sgi<'d> {
```
felipebalbi: This needs a `Mode` as well. Only `Async` and `Dma` enable interrupts. your `irq_enabled` flag is unnecessary.

embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,708 @@
...
+    _peri: Peri<'d, peripherals::SGI0>,
+    periph: pac_sgi::Sgi,
+    irq_enabled: bool,
+    _wake_guard: Option<WakeGuard>,
+}
+
+impl Drop for Sgi<'_> {
+    fn drop(&mut self) {
+        // Keep Drop best-effort and non-blocking.
```
felipebalbi: Drop is expected to be blocking. You should block until you are 100% sure you can drop the thing.

felipebalbi (changes requested):

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Added rustdoc comments for all the `pub enum`.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: The comment was misleading, removed. At this point, we are here because errored out or timed out or successful. SGI automatically drops the busy states and Interrupts in the event of an error. If we timed out, de-asserting the SHA command stops all operations. In success case, everything is already cleared.

It's not intended to be non-blocking, it just happens to be in this case.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Implemented Blocking and Async modes.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Updated to be similar pattern as TRNG.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: StreamingHasher is a slight exception, sharing my response from other review comment:

> Not for streaming mode. The idea for streaming mode with SGI is that a localized Ctx can be maintained with a "running" hash of whatever the input may be. So, when something like an encrypted firmware payload is getting validated, something like AesInstance (not implemented yet) decrypts and then StreamingHasher updates the chunk hash.

> If streaming mode holds the peri internally, not sure how another secure instance can use it without dropping the instance first (e.g. a MAC or AES). Since there is only one SGI, it needs to be managed by some common thread. Open to ideas.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Updated, now similar pattern as TRNG.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Agreed, updated signature and examples.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Updated.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Agreed, added.

embassy-mcxa/src/lib.rs
alamfarjadf: Removed the path description.

embassy-mcxa/src/sgi/hash.rs
alamfarjadf: Updated with chunks_exact; much less verbose now.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Updated to similar style as TRNG.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Updated to `use maitake_sync::WaitCell;`

Similar to TRNG.

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: Updated to `SgiError`

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: This is now resolved. NXP-PAC now uses indexed access for SGI key/data registers.

https://github.com/embassy-rs/nxp-pac/pull/85

diondokter:
Posted a new PR that supersedes this one

PR #5622, opened 2026-03-09
[MCXA5] Enable SPI

felipebalbi:
While at that, also (finally) enable I2C DMA on MCXA5.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/164

### DIFF
--- a/embassy-mcxa/src/perf_counters.rs
+++ b/embassy-mcxa/src/perf_counters.rs
@@ -12,7 +12,7 @@ use core::sync::atomic::{AtomicU32, Ordering};
 
 use paste::paste;
 macro_rules! define_counters {
-    ($($name:ident),*) => {
+    ($( $(#[$attr:meta])* $name:ident),*) => {
         #[cfg_attr(not(feature = "perf"), allow(dead_code))]
         static PERF_COUNTERS: Counters = Counters::new();
 
@@ -20,6 +20,7 @@ macro_rules! define_counters {
             const fn new() -> Self {
                 Self {
                     $(
+                        $(#[$attr])*
                         #[cfg(feature = "perf")]
                         $name: AtomicU32::new(0),
                     )*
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/perf_counters.rs
```diff
@@ -190,6 +190,14 @@ define_counters!(
     interrupt_spi0_wake,
     interrupt_spi1,
     interrupt_spi1_wake,
+    interrupt_spi2,
```
jamesmunns, line 201: Do you want to gate these for 5xx only?

embassy-mcxa/src/spi/controller.rs
```diff
@@ -339,7 +339,7 @@ impl<'d, M: IoMode> Spi<'d, M> {
 
         for word in data {
             // Wait until we have at least one byte space in the TxFIFO.
-            while self.info.regs().fsr().read().txcount() - fifo_size == 0 {}
+            while fifo_size - self.info.regs().fsr().read().txcount() == 0 {}
```
jamesmunns: 👀 what happened here?

felipebalbi: no clue yet how come I missed this on mcxa2. It's clearly an underflow of `fifo_size`.

embassy-mcxa/src/perf_counters.rs
felipebalbi: Macro needs to be updated to accept attributes, then. Let me give that a go.

embassy-mcxa/src/spi/controller.rs
felipebalbi: Unless the fifo is larger on mcxa5, I'll check again shortly

felipebalbi: 4-byte FIFO, just like MCXA2. I'll debug this a little more.

felipebalbi: Resolved, this is not needed at all. Brain fart.

embassy-mcxa/src/perf_counters.rs
felipebalbi: Macro updated.

PR #5623, opened 2026-03-09
feat: add slaac support

### LINKED ISSUE #4582: IPv6 SLAAC
No DHCPv6, just some basic SLAAC to get us started. IPv6 is almost unusable in this state unless you constantly want to keep up. Read some RAs and grab yourself an address, this is really basic stuff. <https://datatracker.ietf.org/doc/html/rfc4862>

While we’re at it, some better Builder API for configuring both IPv4 and IPv6 in `embassy_net::Config` would be convenient in this context.

By the way: It’s 2025. Our microcontrollers are about as fast as desktop computers when IPv6 was invented in the 90s. Embedded devices can have some IPv6 too.

bergzand:
This adds Stateless Address Auto-configuration to embassy-net.The functionality is gated behind a 'slaac' feature.

With this change, a link-local ipv6 address is automatically added to a link, based on the hardware address. I've decided to also guard this behind the slaac feature and not behind proto-ipv6 to not change current behaviour with this PR.  

As the smoltcp doesn't yet support RDNNS (DNS servers via slaac), the DNS entries are left empty in the configuration.

This depends on two PRs in smoltcp:
- https://github.com/smoltcp-rs/smoltcp/pull/1129
- https://github.com/smoltcp-rs/smoltcp/pull/1130

Fixes #4582


bergzand:
> Could you add this configuration to CI? (Adding a flavor in Cargo.toml). This is probably why the build is passing.

I've added two configurations for slaac, one combined with and one without dhcpv4.

bergzand:
With both dependencies merged, I've bumped the `rev` for smoltcp in the `Cargo.toml` with https://github.com/embassy-rs/embassy/pull/5623/commits/d6727a7eb437e092bd6386769556dd1dececa61a. I couldn't spot anything breaking between the two revisions of smoltcp.

bergzand:
I'm not really sure where the latest CI build failure originates from, whether it is related to this PR and where to start looking for fixing it.

lulf:
bender run

akeamc:
> I'm not really sure where the latest CI build failure originates from, whether it is related to this PR and where to start looking for fixing it.

I think I found the cause! The git rev of `smoltcp` inside of embassy-stm32-wpan hasn't been bumped:

https://github.com/embassy-rs/embassy/blob/d6727a7eb437e092bd6386769556dd1dececa61a/embassy-stm32-wpan/Cargo.toml#L36

Running the ci locally, I got the following errors after updating that rev:

<details>
<summary>irrefutable pattern in let chain</summary>

```
error: leading irrefutable pattern in let chain
   --> /home/ake/Documents/embassy-rs/embassy/embassy-net/src/lib.rs:932:20
    |
932 |                 if let IpCidr::Ipv6(ip6_address) = addr
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this pattern will always match
    = help: consider moving it outside of the construct
    = note: `-D irrefutable-let-patterns` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(irrefutable_let_patterns)]`

error: irrefutable `if let` pattern
   --> /home/ake/Documents/embassy-rs/embassy/embassy-net/src/lib.rs:945:28
    |
945 |                         if let IpAddress::Ipv6(gateway) = r.via_router {
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this pattern will always match, so the `if let` is useless
    = help: consider replacing the `if let` with a `let`

error: could not compile `embassy-net` (lib) due to 2 previous errors
     features: ["defmt", "dns", "medium-ethernet", "multicast", "proto-ipv6", "slaac", "tcp", "udp"]
```
</details>

bergzand:
> I think I found the cause! The git rev of `smoltcp` inside of embassy-stm32-wpan hasn't been bumped:

Thanks!

I'll take a look at the irrefutable pattern later this weekend. I guess that's caused when smoltcp is compiled with only IPv6 (and without IPv4), so `IpCidr` enum only has `Ipv6`.

bergzand:
I've refactored the code a bit and bumped the embassy-net revision used by embassy-stm32-wpan, the CI failures are gone now. Thanks @akeamc for pointing out the issue.

spookyvision:
super happy to see this moving forward! I've been playing around with it a little and it seems the link local address is not exposed in any way?

bergzand:
> super happy to see this moving forward! I've been playing around with it a little and it seems the link local address is not exposed in any way?

Correct. Currently Embassy-net exposes only a single address (one for IPv4, one for IPv6). I didn't want to change that here, so had to pick one. The first address configured via slaac makes most sense here.

spookyvision:
> The first address configured via slaac makes most sense here.

sorry for bad wording on my part; I was under the impression that there is *only* the LL address, and that this cannot be accessed. I made a small patch to grab it, but from your reply it's clear I should be doing something different. Am I supposed to somehow retrieve interface data from `smoltcp` instead? (wasn't able to figure out how, which is how I ended up with patching `embassy-net`)


```diff
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -757,7 +757,14 @@ impl Inner {
             ConfigV6::None => None,
             ConfigV6::Static(c) => Some(c),
             #[cfg(feature = "slaac")]
-            ConfigV6::Slaac => None,
+            ConfigV6::Slaac => match self.get_link_local_address() {
+                IpCidr::Ipv4(_) => None,
+                IpCidr::Ipv6(cidr) => Some(StaticConfigV6 {
+                    address: cidr,
+                    gateway: None,
+                    dns_servers: Vec::new(),
+                }),
+            },
         };
     }
```

spookyvision:
@bergzand are you on Matrix (/IRC/Discord/Zulip)? I have a few further questions, but don't want to hijack this PR

PR #5633, opened 2026-03-10
Add USB host support

leftger:
## Summary

- Adds USB host mode support to Embassy with a full vertical slice: traits → DWC2 OTG host driver → enumeration stack → CDC ACM host class → STM32 integration → working examples
- New crates: `embassy-usb-host-driver` (host driver traits), `embassy-usb-host` (enumeration + CDC ACM class)
- Host mode in `embassy-usb-synopsys-otg` behind `host` feature; STM32 integration behind `usb-host` feature
- Examples for STM32F429ZI (OTG_FS) and STM32WBA65RI (OTG_HS)

### DIFF
--- a/embassy-usb-host/src/descriptor.rs
+++ b/embassy-usb-host/src/descriptor.rs
@@ -410,7 +410,7 @@ impl From<EndpointDescriptor> for EndpointInfo {
 mod test {
     use heapless::Vec;
 
-    use super::{ConfigurationDescriptor, EndpointDescriptor, USBDescriptor};
+    use super::{ConfigurationDescriptor, EndpointDescriptor};
 
     #[test]
     fn test_parse_extended_endpoint_descriptor() {
--- a/.github/ci/test.sh
+++ b/.github/ci/test.sh
@@ -37,3 +37,4 @@ cargo test --manifest-path ./embassy-stm32/Cargo.toml --no-default-features --fe
 
 cargo test --manifest-path ./embassy-net-adin1110/Cargo.toml
 cargo test --manifest-path ./embassy-usb-dfu/Cargo.toml --features dfu
+cargo test --manifest-path ./embassy-usb-host/Cargo.toml
--- a/embassy-usb-host/Cargo.toml
+++ b/embassy-usb-host/Cargo.toml
@@ -28,6 +28,9 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 defmt = { version = "1", optional = true }
 log = { version = "0.4.14", optional = true }
 
+[dev-dependencies]
+heapless = "0.9.2"
+
 [features]
 defmt = ["dep:defmt", "embassy-usb-driver/defmt"]
 log = ["dep:log"]


embassy-usb-synopsys-otg/src/host.rs
```diff
@@ -0,0 +1,975 @@
...
+        // Wait for host mode to take effect (~25ms)
+        // We poll cmod bit in gintsts
+        let mut timeout = 250_000u32;
+        while !r.gintsts().read().cmod() {
+            timeout -= 1;
+            if timeout == 0 {
+                break;
+            }
+            cortex_m_nop();
+        }
```
bugadani: I wonder how many changes it will take to make this work on an ESP32-S3. At the very least, a function called `cortex_m_nop` makes me a bit worried 😅

leftger: You're absolutely right! I should have pulled this out into the driver layer. It helps to have it as a draft 🤣

i404788:
Hey, just be aware there is another longstanding USB Host PR with more devices supported: https://github.com/embassy-rs/embassy/pull/4321 you may want to combine forced

leftger:
Thanks so much for pointing out that PR. It seems that PR had a better foundation. I am refactoring my PR to bring in their changes but add a couple of missing details.

embassy-usb-synopsys-otg/src/host.rs
leftger: I replaced it with a spin_loop(). Hope that's OK.

xoviat:
bender run

bugadani:
@i404788 would you mind if I rebased/updated your WIP esp-hal commit for this PR?

i404788:
@bugadani Feel free to, haven't checked the commit log here yet but seems like it would need an update anyway

bugadani:
Thanks, I'll look into that part.

That being said, I'm a bit confused by the crate structure in this PR. `embassy-usb-driver` provides host and device driver interfaces. `embassy-usb-host` implements the host stack and classes. embassy-usb implements the usb device stack and device classes, but also has some files related to usb host (`handlers/hub.rs`, `handlers/kbd.rs`), but also doesn't depend on `embassy-usb-host`. So which of these is the actual host stack, or how does embassy-usb's host code, and embassy-usb-host relate to each other?

i404788:
Hmm could be an artifact of the previous PR which had host/device mode in one crate (embassy-usb) but different submodules, seems like this PR originally started with splitting the host functionality into `embassy-usb-host`   :thinking:

It seems like `embassy-usb-host/src/class` & `embassy-usb/src/handlers/`  serve the same purpose (drivers for connected devices); it seems like it also has some parsing (potentially duplicating `embassy-usb/src/host/descriptor.rs`). Probably should be either merged again, or the host functionality further extracted into `embassy-usb-host`

embassy-usb-synopsys-otg/src/host.rs
```diff
@@ -0,0 +1,1054 @@
+//! USB Host mode driver for Synopsys DWC2 OTG controllers.
+
+use core::cell::UnsafeCell;
+use core::future::poll_fn;
+use core::marker::PhantomData;
+use core::sync::atomic::{AtomicBool, AtomicU8, Ordering};
```
bugadani: Please use portable_atomic, ESP32-S2 has no atomic compare-exchange

embassy-usb-synopsys-otg/Cargo.toml
```diff
@@ -17,6 +17,7 @@ target = "thumbv7em-none-eabi"
 
 [dependencies]
 critical-section = "1.1"
+portable-atomic = { version = "1.5" }
```
bugadani: Nothing leaks AI involvement better than outdated dependencies 😅 portable-atomic is at 1.13 currently, but I think it's fine to declare `"1"`. Either way it's nothing important, just funny :)

leftger: I'm not gonna deny it. I took your comment seriously about the weird structure of joining 2 disparate host approaches and wanted to consolidate it into something more cohesive. I totally understand If you want me to redo the whole thing "human style". I'm OK with that too. I will keep testing this branch to make sure it doesn't break anything.

bugadani: >  I totally understand If you want me to redo the whole thing "human style".

I personally don't really care either way, and I also don't have any powers to make you do anything anyway :)

bugadani:
Okay so now this PR has:
- a new host stack crate (embassy-usb-host)
- a driver interface in embassy-usb-driver
- implementations
- examples
- and some refactor in embassy-usb

To me it looks like that the host stack is intended to be completely separate with embassy-usb-host, I don't know why/if any of the embassy-usb changes are necessary. Are they? Can they be either reverted, or submitted separately if they are a considerable improvement to embassy-usb?

leftger:
Yes very valid point. I agree. I'll remove the `embassy-usb` changes from this PR and submit them as a separate one thereafter.

embassy-usb/src/config_descriptor.rs
```diff
@@ -0,0 +1,550 @@
+//! Parsed USB descriptor types used by the device stack.
```
bugadani: I believe this file is dead code at this point

leftger: just saw that too

embassy-usb-driver/src/lib.rs
```diff
@@ -4,6 +4,31 @@
...
+    Low,
+    /// 12 Mbit/s
+    Full,
+    /// 480 Mbit/s
+    High,
+}
+
+impl Speed {
+    /// Provides the default max_packet_size for a given port speed
+    pub const fn max_packet_size(&self) -> u16 {
```
bugadani: Since `Speed` is copy, this could be
```suggestion
    pub const fn max_packet_size(self) -> u16 {
```

I've seen cases where going by value resulted in better generated code

embassy-usb-driver/src/lib.rs
```diff
@@ -111,6 +136,18 @@ pub struct EndpointInfo {
     pub interval_ms: u8,
 }
 
+impl EndpointInfo {
+    /// Manually create an interface info (you should prefer deriving from parsed `EndpointDescriptor`s)
+    pub fn new(addr: EndpointAddress, ep_type: EndpointType, max_packet_size: u16) -> Self {
```
bugadani: This addition seems unnecessary. You can construct `EndpointInfo` anyway, this constructor just defaults `interval_ms` to 0. But it introduces another API that needs to be kept stable, changing either the struct or the constructor would be a breaking change anyway.

If structs in the driver need constructors, they would be better formulated with a builder pattern, but that complexity seems unneeded.

leftger:
@xoviat It's ready now. I tried it on a USB mouse and it worked fine.

embassy-usb-host/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        // GET_DESCRIPTOR(Device, 8 bytes)
+        let setup = bytes_to_setup(&control::get_device_descriptor(8));
+        let n = ch.control_in(&setup, &mut desc_buf[..8]).await?;
+
+        if n < 8 {
+            return Err(EnumerationError::InvalidDescriptor);
+        }
+
+        let max_packet_size0 = desc_buf[7];
```
bugadani: Should the packet size be validated before we start using it? Legal values are 8, 16, 32, or 64.

embassy-usb-host/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+        // Step 2: SET_ADDRESS
+        let addr = self.next_address;
+        self.next_address = if self.next_address >= 127 {
+            1
+        } else {
+            self.next_address + 1
+        };
+
+        let setup = bytes_to_setup(&control::set_address(addr));
+        ch.control_out(&setup, &[]).await?;
```
bugadani: > After successful completion of the Status stage, the device is allowed a SetAddress() recovery interval of 2 ms. At the end of this interval, the device must be able to accept Setup packets addressed to the new address.

Should the host wait for 2ms after this SetAddress?

embassy-usb-host/src/control.rs
```diff
@@ -0,0 +1,378 @@
...
+                            continue;
+                        } else {
+                            return Err(HostError::RequestFailed);
+                        }
+                    }
+                }
+            }
+        };
+
+        self.device_set_address(new_device_address).await?;
```
bugadani, line 315: Same here, device should be allowed 2ms, maybe?

embassy-usb-host/src/control.rs
```diff
@@ -0,0 +1,378 @@
...
+        })
+    }
+}
+
+impl<D: channel::Direction, C> ControlChannelExt<D> for C where C: UsbChannel<channel::Control, D> {}
+
+/// Extension trait for [`UsbHostDriver`] providing a convenience enumeration method.
+pub trait UsbHostBusExt: UsbHostDriver {
+    /// Enumerate the root port device without selecting a configuration.
+    async fn enumerate_root_bare(
```
bugadani: Why are these enumerate implementations not used in `UsbHost::enumerate`?

embassy-usb-driver/src/host.rs
```diff
@@ -0,0 +1,352 @@
...
+/// ## Virtual USB Channels
+/// These contain the required information to send a packet correctly to a device endpoint.
+/// The information is carried with the channel on creation (see [`UsbHostDriver::alloc_channel`]) and can be changed with [`UsbChannel::retarget_channel`].
+///
+/// It is up to the HAL's driver how to implement concurrent requests, some hardware IP may allow for multiple hardware channels
+///  while others may only have a single channel which needs to be multiplexed in software, while others still use DMA request linked-lists.
+/// Any of these are compatible with the UsbChannel with varying degrees of sync primitives required.
+pub trait UsbChannel<T: channel::Type, D: channel::Direction> {
+    /// Send IN control request
+    async fn control_in(&mut self, setup: &SetupPacket, buf: &mut [u8]) -> Result<usize, ChannelError>
```
bugadani, line 326: I assume the `usize` return value is the number of returned bytes, but it would be nice if that was spelled out

embassy-usb-host/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+            .map_err(|_| EnumerationError::InvalidDescriptor)?;
+        let total_len = config_header.total_len as usize;
+
+        if total_len > config_buf.len() {
+            return Err(EnumerationError::InvalidDescriptor);
+        }
+
+        // Get full configuration descriptor
+        let setup = bytes_to_setup(&control::get_config_descriptor(0, total_len as u16));
+        let n = ch.control_in(&setup, &mut config_buf[..total_len]).await?;
```
bugadani, line 197: I believe `n` should be checked, it should be equal to total_len.

USB 2.0, Section 9.6.3:

> When the host requests the configuration descriptor, all related interface and endpoint descriptors are returned (refer to Section 9.4.3)

USB 2.0, Section 9.4.3:

> The wLength field specifies the number of bytes to return. If the descriptor is longer than the wLength field, only the initial bytes of the descriptor are returned. If the descriptor is shorter than the wLength field, the device indicates the end of the control transfer by sending a short packet when further data is requested. 

You've already read the config descriptor header, which has the exact size in it. If, then, the device returns a different number of bytes, maybe it should be treated as an error.

embassy-usb-host/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+        let max_packet_size0 = desc_buf[7];
+        trace!("EP0 max packet size: {}", max_packet_size0);
+
+        // Step 2: SET_ADDRESS
+        let addr = self.next_address;
+        self.next_address = if self.next_address >= 127 {
+            1
+        } else {
+            self.next_address + 1
+        };
```
bugadani: Should the host maintain a 128-bit bitmask instead to see which addresses are free to use?

leftger:
I plugged one of these into the Host USB-C port with your suggested changes and it worked 🤣 
image

bugadani:
But did it work before? 🙃

leftger:
Yes it worked fine before too on keyboard/mouse. I left my usb-a to usb-c adapter at work and I only had this flight stick to test against (usb-c to usb-c cable)

embassy-usb-host/src/descriptor.rs
```diff
@@ -0,0 +1,468 @@
...
+            ep_type: value.ep_type(),
+            max_packet_size: value.max_packet_size,
+            interval_ms: value.interval,
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use heapless::Vec;
```
bugadani, line 411: Try running the tests once, please. At the very least, heapless is not a dependency of embassy-usb-host.

PR #5641, opened 2026-03-11
Permit writes that don't fill the entire requested buffer in `{UdpSocket, IcmpSocket}::send_to_with`

### LINKED ISSUE #5636: embassy-net: Variable-length writes with `UdpSocket::send_to_with`
This might be a rather niche use case, but I'm sending Protobuf over UDP and would like to take advantage of the zero-copy functionality that embassy-net provides. The problem that I'm facing is that `UdpSocket::send_to_with` requires the user to know the exact size beforehand, while I only know the maximum size.

I've noticed that `smoltcp::udp::Socket` already provides a method `send_with` that does exactly what I need: the closure returns the number of bytes written. Could `UdpSocket::send_to_with` maybe also do that? It would be a breaking change though, unless we introduce a second method.

What do you think? I'd be happy to open a PR!

akeamc:
This PR resolves #5636.

~~The `send_to_with` methods took a `&mut self` for no apparent reason (`send_to` doesn't and works essentially in the same way), so they have been updated to take `&self` instead. Or were the `&mut self`s intentional?~~ Nevermind, found the reason: https://github.com/smoltcp-rs/smoltcp/pull/964#issuecomment-2260564841 Shouldn't `Udp::recv_from` also take a mutable reference then?

I've tested the UDP implementation on an ESP32 with a W5500 and it appears to work!


Dirbaio:
> Shouldn't Udp::recv_from also take a mutable reference then?

Probably.

PR #5587, opened 2026-03-04
prepare release of `embassy-sync` 0.8.0

yodaldevoid:
`embassy-sync` is one of the few Embassy crates that I can find that have not been released since the `embedded-io` 0.7 update.

Relates to #5212

PR #5644, opened 2026-03-11
Prepare embassy-usb-synopsys-otg release

bugadani:
This PR is the result of me running `cargo embassy-devtool prepare-release embassy-usb-synopsys-otg`. I don't know why the devtool bumped embassy-stm32 and embassy-boot-stm32 for this.


lulf:
It's also being prepared in https://github.com/embassy-rs/embassy/pull/5587 so please just wait for that. I'm not sure why embassy-stm32 is bumped though, will look into it.

PR #5649, opened 2026-03-12
chore: update changelogs for release

lulf:
Used claude code to check for missing entries and improve structure for the more complex ones.


ivmarkov:
@lulf Leaving a mark here as well: I'm not sure `embassy-net-driver-channel` needs the 0.4 bump, because I can't see `embassy-sync` being part of its public API in the first place. 

Might be missing something as I'm just doing visual inspection, but the channel driver API is small enough anyway.

PR #5650, opened 2026-03-12
chore: rename executor-thread and executor-interrupt features

lulf:
- Rename features to to signal they are for single core use only.
- Add deprecation warning when using old feature names.


xoviat:
bender run

Dirbaio:
I don't think this is quite right, `executor-thread` works fine for std multicore. The "must be single core" restriction is more about the cortex-m platform (and probably other platforms). 

btw I think we should do #5099 before release

PR #5651, opened 2026-03-12
[MCXA] GPIO Interrupt binding

Guacimo:
# MCXA GPIO Interrupt binding
Replaced the static interrupt binding of the MCXA gpio module with the `bind_interrupts` mechanism.
This is done for both `Input` and `Flex`.

## Implementation
The structs now use the typestate pattern to differentiate being initialized with the Irq bound or not. This mirrors what the MCXA `adc` module does. In order to use the pins asynchronously the `new_async` constructors need to be used with the interrupt binding as argument. 
Enabling the gpio interrupts is now done in the `new_async()` constructor. This means the gpio interrupts are only enabled when they are actually being used by the gpio module. The chip init now calls `cortex_m::interrupt::enable();` because other drivers depend on interrupts being enabled.

## Validation
The examples have also been updated to use the `bind_interrupts` mechanism. 
Right now I am unable to test on the MCXA577 but I did test the example on the MCXA276.
The MCXA577 example has also been updated but only validated by running `cargo check`.

### DIFF
--- a/examples/mcxa5xx/src/bin/button_async.rs
+++ b/examples/mcxa5xx/src/bin/button_async.rs
@@ -8,7 +8,7 @@ use hal::gpio::{Input, Pull};
 use {defmt_rtt as _, embassy_mcxa as hal, panic_probe as _};
 
 bind_interrupts!(struct Irqs {
-    GPIO1 => gpio::InterruptHandler<peripherals::GPIO1>;
+    GPIO3 => gpio::InterruptHandler<peripherals::GPIO3>;
 });
 
 #[embassy_executor::main]


jamesmunns:
Let's see what CI thinks about this

bender run

embassy-mcxa/src/gpio.rs
```diff
@@ -448,13 +449,27 @@ macro_rules! impl_pin {
...
+impl sealed::Sealed for Async {}
+impl Mode for Async {}
+
+pub struct Blocking {}
+impl sealed::Sealed for Blocking {}
+impl Mode for Blocking {}
+
 /// A flexible pin that can be configured as input or output.
-pub struct Flex<'d> {
+pub struct Flex<'d, M: Mode> {
```
diondokter: This is correct, but is it desired? Pins are used a lot, do we want everyone to have to type in blocking all the time?

Maybe it makes sense to have a default generic here?
`M: Mode = Blocking`

Alternatively the async one could be a separate type...

embassy-mcxa/src/gpio.rs
```diff
@@ -576,7 +594,27 @@ impl<'d> Flex<'d> {
 }
 
 /// Async methods
-impl<'d> Flex<'d> {
+impl<'d> Flex<'d, Async> {
+    /// Wrap the pin in AsyncFlex with Async support.
```
diondokter: Maybe put some pointers to the extra functions you get with async

jamesmunns:
bender run

jamesmunns:
@Guacimo for some reason I can't push to your PR, but you'll need to run `./fmtall.sh` in the root of the repo to clear the formatting issues.

embassy-mcxa/src/gpio.rs
```diff
@@ -39,72 +40,76 @@ const PORT_COUNT: usize = 6;
...
+            }
+        )*
+    };
 }
 
+impl_instance!(0);
+impl_instance!(1);
+impl_instance!(2);
+impl_instance!(3);
+impl_instance!(4);
```
felipebalbi: Looks like your macro allows for:

```rust
impl_instance!(0, 1, 2, 3, 4);
```

embassy-mcxa/src/gpio.rs
```diff
@@ -444,14 +449,32 @@ macro_rules! impl_pin {
...
+pub struct Blocking {}
+impl sealed::Sealed for Blocking {}
+impl Mode for Blocking {}
+
 /// A flexible pin that can be configured as input or output.
-pub struct Flex<'d> {
+pub struct Flex<'d, M: Mode = Blocking> {
     pin: Peri<'d, AnyPin>,
     _marker: PhantomData<&'d mut ()>,
+    _mode: PhantomData<M>,
```
felipebalbi: You don't need a new field, right? How about:

```suggestion
    _phantom: PhantomData<&'d mut M>,
```

felipebalbi (approved):
LGTM, couple minor nits below.

embassy-mcxa/src/gpio.rs
Guacimo: I have added the default generic as you suggested. If you think implementing it as a different type I can do that as well. I thought this would be cleaner because implementing a second `Flex` and `Input` type with all supported traits would create a lot of code duplication.

embassy-mcxa/src/gpio.rs
Guacimo: I have added a line to the documentation that provides a few of the async function examples.

embassy-mcxa/src/gpio.rs
Guacimo: I have added your change!

embassy-mcxa/src/gpio.rs
Guacimo: I have added your change!

felipebalbi:
@Guacimo, it looks like you need to patch the tests as well.

Guacimo:
> @Guacimo, it looks like you need to patch the tests as well.

Yes, I will attempt to fix them tomorrow. Some of the examples are also failing and I am not 100% sure how to deal with the `puart_bbq_rx` example which type erases the pin and then wants to asynchronously `wait_for_low` afterwards. This is disallowed after this pull request as I cannot statically verify the correct interrupt was bound after type erasure.

Guacimo:
I have fixed the test that was failing, and also fixed the remaining examples that were failing.
In order to make the `puart_bbq_rx` example work I have added a mechanism that allows for constructing an async `Input` and an async `Flex` from an `AnyPin`. 
You can now downcast to an `AnyPin` using the `degrade_async(Irq)` function. This will force binding the correct Irq and then allows constructing an `Input` or a `Flex` using the `async_from_anypin(Anypin)` functions. These check if the `irq_bound` flag is set in the `AnyPin` if not, they return the `NoIrqBound` error.

PR #5653, opened 2026-03-12
[MCXA] Pre-split RTC

jamesmunns:
This is a setup step for adding 5xx RTC support, which will not share (most) code from the 2xx impl. Doing this to keep the diff simple


Dirbaio:
bender run

PR #5654, opened 2026-03-12
[MCXA] i2c: improve documentation

felipebalbi:
Make the documentation a little more extensive, if repetitive at times.

### DIFF
--- a/embassy-mcxa/Cargo.toml
+++ b/embassy-mcxa/Cargo.toml
@@ -140,5 +140,7 @@ mcxa2xx = ["nxp-pac/mcxa256"]
 mcxa5xx = ["nxp-pac/mcxa577"]
 
 [dev-dependencies]
+embassy-executor = { version = "0.10.0", path = "../embassy-executor", features = ["platform-cortex-m", "executor-interrupt", "executor-thread"] }
+embedded-storage = "0.3.1"
 panic-halt = "1.0.0"
-embassy-executor = { version = "0.10.0", path = "../embassy-executor", features = ["arch-cortex-m", "executor-interrupt", "executor-thread"] }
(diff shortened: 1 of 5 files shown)


felipebalbi:
Glad to modify, remove, update anything here.

embassy-mcxa/src/i2c/controller.rs
```diff
@@ -1,4 +1,58 @@
...
+//!
+//! ### Error Types
+//!
+//! - `SetupError`: Errors related to hardware initialization, such as
+//! clock configuration issues.
+//! - `IOError`: Errors during I2C operations, including FIFO errors,
+//! arbitration loss, and invalid buffer lengths.
+//!
+//! ## Example
+//! ```rust,no_run
```
jamesmunns, line 33: Are doc tests checked in CI?

felipebalbi: I don't think they are. But, should I add a `ignore` here?

diondokter:
I mean it looks fine. But these types of docs go out of date *fast* usually

embassy-mcxa/src/i2c/controller.rs
felipebalbi: Tests should compile now.

felipebalbi:
> I mean it looks fine. But these types of docs go out of date _fast_ usually

Was the original better? I can remove the changes to the public interface, keep the module-level update and all the ones that make sure doc code blocks compile.

felipebalbi:
I'm queueing this one. If it turns out we have "too much" information on the public methods, we can always delete the extra docs later. Meanwhile, all rustdoc examples compile fine.

PR #5656, opened 2026-03-12
examples/rp add multicore_multiprio example

orukusaki:
Adds an example to demonstrate using the new Multicore Executors added in https://github.com/embassy-rs/embassy/pull/5409


sourcebox:
As all examples share the same `Cargo.toml` either all of them have to be updated or this one has to go into a separate subdirectory so it can have a separate configuration.

The main question to be answered is probably whether the user should always use the custom executors to avoid any possible issues.

orukusaki:
I see what you mean. Another approach might be to copy what was done for  `mcxa5xx/src/bin/power-deepsleep-big-jump.rs` etc and put cfg flags in the example

sourcebox:
When we discussed it on Matrix, @lulf suggested using a subdirectory, @jamesmunns pointed me to the MCXA example with the cfg flags. 

@Dirbaio: what's your opinion on this?

orukusaki:
This way seems more simple to me. I think it would be good to have the custom executors be the default, there's no reason to _not_ use them. That would require updating all the existing examples though

sourcebox:
Updating all examples is a bit of work, needs to be done also for the RP235x, I guess. So we should be sure to go that route before doing that.

Dirbaio:
updated all examples to use the custom executor.

Issue #4982, opened 2025-12-03
Rp2350 won't enable interrupts at start


ohm69:
Applications for Rpi Pico 2350 depend on the interrupts being enabled by the bootstrap. Not all bootstraps do that, especially multiboot loaders like uf2loader by pelrun.
C SDK have fixed that issue in version 2.2 (see [this PR](https://github.com/raspberrypi/pico-sdk/pull/2491/files)).

A workaround for that is placing following code at the beginning of an app, but it shouldn't be application task to set up initial system state reliably.
>     unsafe {
>         cortex_m::interrupt::enable();
>     }

Reproduction steps:
- use the default bootstrap
- upload an application built with embassy
- check for the interrupt state
- use the bootstrap that disable interrupts
- upload an application built with embassy
- check for the interrupt state

PR #5660, opened 2026-03-13
[embassy-rp] Enable interrupts in init function

sourcebox:
Fix for https://github.com/embassy-rs/embassy/issues/4982

### DIFF
--- a/embassy-rp/CHANGELOG.md
+++ b/embassy-rp/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Complete missing Gpin/GpoutPin impls for RP235x
 - Improve PIO clock divider math
 - Fix chrono compilation
+- Enable interrupts in `init`
 - Update embassy-sync to 0.8.0
 - Update embassy-embedded-hal to 0.6.0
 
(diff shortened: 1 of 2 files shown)


Dirbaio:
This is risky, the user might have good reasons to run `init` with interrupts disabled.

- It might be inside a critical section, we'd be "prematurely ending" it which can cause unsoundness.
- RTIC's init() runs with interrupts disabled because the resource sharing stuff isn't fully initialized yet, forcing enabling interrupts can probably cause unsoundness too.

sourcebox:
So what would be the alternative to make it independent of the bootloader? Because the risks you describe are probably also present when using a bootloader that enables them by default.

Dirbaio:
> So what would be the alternative to make it independent of the bootloader?

I don't think it's possible to fix. The current convention is irqs are enabled at boot. Libraries can't enable irqs without risking unsoundness. If you have a broken bootloader that jumps to the app with irqs disabled, it's up to you, the user, to unsafely enable irqs at start of main when you know it can't cause unsoundness.

> Because the risks you describe are probably also present when using a bootloader that enables them by default.

No, because the firmware will disable irqs if it's necessary for soundness, it doesn't rely on irqs being disabled on boot.

Dirbaio:
Maybe we can do it in pre_init. That shouldn't cause any unsoundness. 

Disable+unpend all irqs individually in NVIC, then enable global interrupts.

sourcebox:
Ok, I close this PR so we can check a better solution.

PR #5661, opened 2026-03-13
[MCXA5] Add RTC driver

felipebalbi:
While at that, also patch mcxa2 rtc slightly to remove `Rtc` prefix from its public config and datetime types.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/177


felipebalbi:
bender run

PR #5663, opened 2026-03-13
stm32: cordic: refactor API, fix RRDY bug, re-enable CORDIC on U5A5/U5A9

okhsunrog:
## Summary

- **API refactor**: Move `arg_count` and `res_count` into `Config` with builder methods. Calc methods read from config instead of taking bool params. Add `set_access_counts()` for lightweight mode switching. Add `#[inline]` to all calc methods. Remove `extra_config()`.
- **Bug fix**: Fix pre-existing RRDY stale data bug in `reconfigure()` that caused wrong blocking results and async DMA hangs on H5/U5.
- **Re-enable CORDIC on STM32U5A5/U5A9**, previously disabled due to the above bug.

## Benchmark (STM32G474RE, O3 + fat LTO)

| Mode | Cycles/call | Overhead vs bare |
|---|---|---|
| **New API** (config-based, no per-call CSR write) | **22** | **3 cycles** |
| **Old API** (3 CSR modify() per call) | **35** | **16 cycles** |
| **Bare registers** (WDATA + 2×RDATA) | **19** | baseline |

13 cycles saved per call (81% overhead reduction). All results verified correct with `black_box` preventing compiler from optimizing away reads.

## RRDY bug details

`reconfigure()` performs a dummy CORDIC calc to reset ARG2 to +1, then drains RRDY. However, the subsequent CSR writes (`set_func`, `set_precision`, `set_scale`, and in the new API also `set_nres`) could re-assert RRDY because the secondary result from the dummy calc was never read (NRES was One during drain). This caused:

- **Blocking**: first calc read stale dummy result instead of new computation
- **Async DMA on H5/U5**: GPDMA received a premature read request from the stale RRDY, corrupting the transfer sequence and causing a hang

Fix: add `clean_rrdy_flag()` after applying user config in `reconfigure()`.

This was the root cause of the CORDIC crash on U5A5/U5A9 that led to CORDIC being disabled on those chips. With the fix, CORDIC is re-enabled.

## Test plan

- [x] HIL tests pass on G491 (regular DMA), H563 (GPDMA), U585 (GPDMA), U5A5 (GPDMA, re-enabled)
- [x] Blocking correctness verified on G474 and H562 with known values
- [x] Async DMA verified on H562 with various binary layouts (previously caused hangs)
- [x] Benchmark verified on G474 with O3 + fat LTO


okhsunrog:
bender run

xoviat:
Ok to merge?

okhsunrog:
@xoviat still debugging some issues, figuring out what exactly breaks HIL

PR #5674, opened 2026-03-16
[MCXA] GPIO Interrupt binding (take two)

jamesmunns:
Closes https://github.com/embassy-rs/embassy/pull/5651


jamesmunns:
This is #5651, but I can't push to that branch for some reason. Opening as a replacement PR.

PR #5676, opened 2026-03-16
[MCXA] lpuart-bbq improvements

jamesmunns:
Adding some missing methods and impls

### DIFF
--- a/embassy-mcxa/src/lpuart/bbq.rs
+++ b/embassy-mcxa/src/lpuart/bbq.rs
@@ -449,7 +449,7 @@ impl LpuartBbq {
     }
 
     /// Borrow split parts.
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -448,6 +448,12 @@ impl LpuartBbq {
         self.tx.blocking_flush();
     }
 
+    /// Borrow split parts.
+    pub fn split(&mut self) -> (&mut LpuartBbqRx, &mut LpuartBbqTx) {
```
felipebalbi: In the rest of this HAL, `split()` methods are consuming with `split_ref` being the borrowing counterpart. Would you like to match that?

jamesmunns: Sure! I'll switch to just `split ref` for now, splitting by ownership is more of a thing

PR #5686, opened 2026-03-17
stm32: Make StorageDevice construction infallible and expose initialization methods

peterkrull:
Examples have not been updated to match, but I wanted to get some feedback first. I believe @xoviat worked the most on this.

### DIFF
--- a/embassy-stm32/src/sdmmc/sd.rs
+++ b/embassy-stm32/src/sdmmc/sd.rs
@@ -128,19 +128,18 @@ pub struct StorageDevice<'a, 'b, T: Addressable> {
 /// Card Storage Device
 impl<'a, 'b> StorageDevice<'a, 'b, Card> {
     /// Create a new SD card
-    pub async fn new_sd_card(sdmmc: &'a mut Sdmmc<'b>, cmd_block: &mut CmdBlock, freq: Hertz) -> Result<Self, Error> {
-        let mut s = Self {
+    ///
+    /// Remember to call [`initialize`](`StorageDevice::initialize`)
+    /// before using it, or after inserting an SD card.
+    pub fn new_sd_card(sdmmc: &'a mut Sdmmc<'b>) -> Self {
+        Self {
             info: Card::default(),
             sdmmc,
-        };
-
-        s.acquire(cmd_block, freq).await?;
-
-        Ok(s)
+        }
     }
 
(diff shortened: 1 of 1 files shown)


xoviat:
What is the purpose of this?

peterkrull:
Without exposing the initialize function, I am not able to unplug and plug an SD card in without having to drop and reconstruct the StorageDevice. This is however awkward due to the borrowed Sdmmc, since there is no easy way to reclaim it without relying on the dropped borrow becoming available again, which is not easy outside of simple scopes. Exposing the initializer makes this much easier. And making the constructor infallible ensures I can obtain the StorageDevice first, and then worry about the connected status of the card afterwards.

Maybe the purpose of the StorageDevice is to represent a verified connected device. But then I think there should at least be a way to explicitly reclaim the inner Sdmmc. It is currently not possible to simply move the Sdmmc out due to StorageDevice implementing Drop.

xoviat:
```rust
let sdmmc = Sdmmc::new();

loop {
    // wait for sd card request

    let mut storage = loop {
        if let Ok(storage) = StorageDevice::new_sd_card(&mut sdmmc, &mut cmd_block, mhz(15)).await {
            break storage;
        }
    };


    // do stuff


     // drop occurs here
}
```

peterkrull:
> without relying on the dropped borrow becoming available again, which is not easy outside of simple scopes

But no matter what I  do, I cannot seem to get the pattern below to work. At least not in edition 2024, which does not like the looping re-borrowing.
```rust
struct StorageTask {
    block: DataBlock,
    storage: StorageDevice<'static, 'static, Card>
}

impl StorageTask {
    async fn new(sdmmc: &'static mut Sdmmc<'static>) -> Self {
        let storage = loop {
            // [Error] `*sdmmc` was mutably borrowed here in the previous iteration of the loop
            if let Ok(storage) = StorageDevice::new_sd_card(sdmmc, ...).await {
                break storage;
            }
        };

        StorageTask {
            block: DataBlock::new(),
            storage
        }
    }

    async fn run(&mut self) {
        if self.storage.read_block(0, &mut self.block).await.is_err() {
            // If card was disconnected, how should I recover?
            // I cant get &mut Sdmmc back, and there is no reinit
        }
    }
}
```

Unless the intent is to only keep the `Sdmmc` around persistently, and construct the `StorageDevice` only right when it is needed?

PR #5688, opened 2026-03-17
update sdmmc example and add functions

xoviat:
cc @peterkrull

closes #5686

### DIFF
--- a/examples/stm32f4/src/bin/sdmmc.rs
+++ b/examples/stm32f4/src/bin/sdmmc.rs
@@ -4,9 +4,12 @@
 use defmt::*;
 use embassy_executor::Spawner;
 use embassy_stm32::sdmmc::Sdmmc;
-use embassy_stm32::sdmmc::sd::{CmdBlock, DataBlock, StorageDevice};
+use embassy_stm32::sdmmc::sd::{Card, CmdBlock, DataBlock, StorageDevice};
 use embassy_stm32::time::{Hertz, mhz};
 use embassy_stm32::{Config, bind_interrupts, dma, peripherals, sdmmc};
+use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+use embassy_sync::channel::Channel;
+use embassy_time::{Duration, Timer};
 use {defmt_rtt as _, panic_probe as _};
 
 /// This is a safeguard to not overwrite any data on the SD card.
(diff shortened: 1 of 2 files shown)


peterkrull:
I do not think this really answers my concern. Part of my issue with the current API is that it forces a sort of type-state pattern, but using mutable borrows, where the StorageDevice represents a card which was once connected and initialized. It forces the user to adopt this pattern, and write a `init, use, drop, loop` like this example, even if it would suffice to just have a public initialize/acquire method. It makes it extremely awkward to create a self-contained type wrapping the StorageDevice/Sdmmc ([like this one](https://github.com/embassy-rs/embassy/pull/5686#issuecomment-4078147701)), which is able to reset itself.

And that is where I think, if the StorageDevice can exist for a disconnected card (that was once initialized), why not just allow me to construct it infallibly, and allow me to initialize whenever I need to.

xoviat:
I've added two new functions which allow the behavior you're seeking, Are they satisfactory?

peterkrull:
Yes that is what I am seeking. I do not have an opinion on whether the example stays, but the comments describing errors should at least be removed.

examples/stm32f4/src/bin/sdmmc.rs
```diff
@@ -18,6 +21,43 @@ bind_interrupts!(struct Irqs {
...
 
+pub enum StorageRequest {
+    WriteRequest(u32, &'static DataBlock),
+    ReadRequest,
+}
+
+pub async fn run_storage<'a>(mut sdmmc: Sdmmc<'a>, channel: &'static Channel<NoopRawMutex, StorageRequest, 3>) {
+    loop {
+        let storage = loop {
+            // [Error] `*sdmmc` was mutably borrowed here in the previous iteration of the loop
```
peterkrull: this one

examples/stm32f4/src/bin/sdmmc.rs
```diff
@@ -58,9 +98,12 @@ async fn main(_spawner: Spawner) {
 
     let mut cmd_block = CmdBlock::new();
 
-    let mut storage = StorageDevice::new_sd_card(&mut sdmmc, &mut cmd_block, mhz(24))
-        .await
-        .unwrap();
+    let mut storage = loop {
+        // [Error] `*sdmmc` was mutably borrowed here in the previous iteration of the loop
```
peterkrull: and this one

PR #5693, opened 2026-03-19
stm32/dma: Fix ring buffer issues

HybridChild:
## Summary

Seven correctness and robustness fixes to the DMA ring buffer implementation, plus two code clarity improvements, identified through a review of `dma_bdma.rs` and `ringbuffer/mod.rs`.

---

### 1. Fix `else if` masking TCIF when both HTIF and TCIF fire simultaneously

The interrupt handler used an `else if` chain for HTIF and TCIF. If the handler was delayed (e.g. by a higher-priority interrupt), both flags can be set when it fires. Only HTIF was cleared; TCIF remained pending and was handled in a second interrupt invocation.

In the window between the two invocations, `dma_sync` observed `laps_completed == 0` with `pos < self.pos` — meaning the DMA had wrapped but the complete count hadn't been incremented yet. This triggers the `cap - 1` fallback position, causing `read_exact` to return an unnecessary extra `Poll::Pending` and under-reporting available data by the items already written into the new DMA lap.

**Fix:** Replace `else if` with separate `if` blocks in both the DMA and BDMA branches so both flags are acknowledged and `complete_count` is incremented in a single interrupt invocation.

---

### 2. Defer `read_index` advance until after the copy in `read_raw`

The alignment-skip advance (`read_index.advance(skip)`) happened before the data copy and before the second `len()` call. If that second `len()` returned `Err(Overrun)`, `read()` called `reset()` which corrected `read_index` — so there was no data corruption. However, `read_index` was left in a partially-advanced state between the two operations, and callers of `read()` had no way to distinguish an overrun detected before any copying from one detected after data was already written into `buf`.

**Fix:** Compute `skip` without touching `read_index`. Pass it as an extra offset to `read_buf`. Commit `skip + readable` atomically in a single `advance()` at the end. `read_index` is now either fully committed or unchanged on error.

---

### 3. Eliminate the second `sync_len()` call in `read_raw`

`read_raw` called `self.sync_len(dma)` twice: once to determine available data, and once after the copy to compute the `remaining` return value. The second call read the NDTR hardware register again and invoked `reset_complete_count()` a second time mid-read, potentially consuming a wrap count that arrived during the copy. On the next `dma_sync`, this caused `laps_completed == 0` even though a lap had completed, unnecessarily widening the window for the `cap - 1` fallback.

**Fix:** Snapshot `write_index` after the first `sync_len()`. After committing the advance, derive `remaining` from `write_snapshot.diff(&self.read_index)`. This is a conservative underestimate if the DMA wrote more data during the copy, but `remaining` is already documented as an estimate and the change eliminates the second hardware access and `complete_count` reset.

---

### 4. Assert minimum buffer size in `set_alignment`

`set_alignment(n)` asserted `cap % n == 0` but allowed `cap == n` (exactly one frame in the buffer). With a single-frame buffer, HTIF fires at the halfway point of that frame and TCIF fires at the end — there is no window in which the CPU can safely read a complete frame while the DMA fills the next one.

**Fix:** Add `assert!(cap >= 2 * alignment)` so the buffer always holds at least two frames, ensuring HTIF and TCIF fire on frame boundaries with a safe read window between them.

---

---

### 5. Expose `write_pos()` on `WritableDmaRingBuffer` for TX frame-alignment recovery

After a TX write overrun, `WritableDmaRingBuffer::reset()` records `DMA_pos_at_reset` in `write_index.pos`. There was no way for callers to read this value, forcing them to re-read the NDTR hardware register after `write()` returned the error.

The NDTR re-read has an inherent race: DMA advances by 1–3 words between the `reset()` call inside `inspect_err` and the external NDTR read. For I2S Data24Channel32 (4 words per stereo frame), this 1-word error causes incorrect frame-alignment padding. In the worst case a pad of N positions is applied to a `write_index.pos` that is already N-1 positions ahead of the aligned boundary, landing the write pointer at an odd offset and swapping the 16-bit hi/lo words of every sample — producing sustained distortion until the next TX ring-buffer reset.

**Fix:** Add `pub fn write_pos(&self) -> usize` to `WritableDmaRingBuffer`, exposed through `WritableRingBuffer` in `dma_bdma.rs`. Callers can now read `write_index.pos` immediately after a write error (before any subsequent writes) and get the exact DMA position recorded at reset time with no timing uncertainty.

---

### 5b. Fix `read_latest` resetting `read_index` when `diff == 0`

`read_latest` used `diff <= 0` as the overrun/desync guard, which reset `read_index` to `write_index` when `diff == 0` — i.e. when no data is simply available yet. This was incorrect: an empty buffer is not an error condition and should not disturb `read_index`.

**Fix:** Split `diff <= 0` into `diff < 0 || diff > cap` (reset + return 0) and `diff == 0` (return 0 without touching `read_index`). Add early return for `available == 0`.

---

### 5d. Eliminate the second `sync_len()` call in `write_raw`

`write_raw` called `self.sync_len(dma)` twice: once to determine available space, and once after the copy to compute the `remaining` return value. The second call invoked `reset_complete_count()` again mid-write, potentially consuming a wrap count that arrived during the copy loop — the same problem as fix 3 for `read_raw`.

The second call also relied on `write_index` being un-advanced at the time it ran, requiring `available - writable` arithmetic to compensate. This implicit coupling was fragile and non-obvious.

**Fix:** Snapshot `read_index` after the first `sync_len()`. Advance `write_index` immediately after the copy, then derive `remaining` from `write_index.diff(cap, &read_snapshot)`. The `.max(0)` guard used in `read_raw`'s analogous fix is not needed here because `write_index >= read_snapshot` is guaranteed by the first `sync_len()` not returning `Overrun`, and `advance` only increases it further.

---

### 5c. Fix `WritableDmaRingBuffer::new` using out-of-range `pos`

The constructor initialised `write_index` as `{ complete_count: 0, pos: len }` (where `len == cap`) to represent a full buffer (one buffer-length ahead of `read_index`). The value `pos: cap` violates the invariant `pos < cap`. It happened to work because `as_index` wraps via `% cap`, but `write_pos()` exposes the raw `pos` field, so it returned `cap` rather than the correct `0`.

**Fix:** Use `{ complete_count: 1, pos: 0 }`, which is the canonical representation of the same linear position and keeps `pos` within `[0, cap)`.

---

### 6. Rename `DmaRingBuffer::len` to `sync_len`

`len()` implied a cheap, pure query — the standard Rust convention. In reality the method takes `&mut self`, reads the NDTR hardware register, and atomically resets `complete_count`. Callers had no indication from the name that a hardware side effect was involved.

**Fix:** Rename to `sync_len` across `ReadableDmaRingBuffer`, `WritableDmaRingBuffer`, and all call sites in `dma_bdma.rs` and `gpdma/ringbuffered.rs`.

---

### 7. Add doc comments to `DmaIndex` methods

`DmaIndex` is the core bookkeeping type for the ring buffer — its methods (`reset`, `as_index`, `dma_sync`, `advance`, `normalize`, `diff`) were undocumented. In particular, the ordering constraint in `dma_sync` and the wrap-around race it guards against were only captured in a terse inline comment.

**Fix:** Add doc comments to all `DmaIndex` methods explaining their purpose, preconditions, and — for `dma_sync` — the race condition and why the two reads must happen in the order they do. Also rename the local variable `count_diff` to `laps_completed` for clarity.

---

## Files changed

- `embassy-stm32/src/dma/dma_bdma.rs` — separate `if` blocks for HTIF/TCIF in both DMA and BDMA branches of `on_irq`; `write_pos()` accessor on `WritableRingBuffer`; `sync_len` rename
- `embassy-stm32/src/dma/ringbuffer/mod.rs` — deferred `read_index` advance, single `sync_len()` call with write/read snapshot in both `read_raw` and `write_raw`, minimum buffer size assertion in `set_alignment`; `write_pos()` accessor on `WritableDmaRingBuffer`; fix `new()` initial `write_index` to `{complete_count:1, pos:0}`; `sync_len` rename; `DmaIndex` doc comments; `count_diff` → `laps_completed`
- `embassy-stm32/src/dma/gpdma/ringbuffered.rs` — `sync_len` rename

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -16,6 +16,7 @@ DMA:
 - fix: stm32/dma: fix read_latest incorrectly resetting read_index when diff == 0 (no data available)
 - feat: stm32/dma: expose write_pos() on WritableRingBuffer for timing-safe TX frame alignment after overrun
 - fix: stm32/dma: fix WritableDmaRingBuffer::new using out-of-range pos (cap) instead of {complete_count:1, pos:0}
+- fix: stm32/dma: eliminate second sync_len() call in write_raw to prevent consuming a lap count mid-copy
 
 ADC:
 - feat: stm32/adc: add `VrefInt::calibrated_value()` for additional chips
--- a/embassy-stm32/src/dma/ringbuffer/mod.rs
+++ b/embassy-stm32/src/dma/ringbuffer/mod.rs
@@ -473,16 +473,27 @@ impl<'a, W: Word> WritableDmaRingBuffer<'a, W> {
         .await
     }
 
+    /// Write as many elements as currently fit, returning `(written, remaining_capacity)`.
+    /// Does not block; call `write` for automatic reset-on-overrun or `write_exact` to write all.
     fn write_raw(&mut self, dma: &mut impl DmaCtrl, buf: &[W]) -> Result<(usize, usize), Error> {
         fence(Ordering::Release);
 
         let writable = self.sync_len(dma)?.min(buf.len());
+        // Snapshot read_index before the copy so that the returned `remaining`
+        // reflects the buffer state at the time of the write rather than
+        // requiring a second hardware NDTR access and complete_count reset mid-copy.
+        let read_snapshot = self.read_index;
+
         for i in 0..writable {
             self.write_buf(i, buf[i]);
         }
-        let available = self.sync_len(dma)?;
+
         self.write_index.advance(self.cap(), writable);
-        Ok((writable, available - writable))
+
+        let remaining = self
+            .cap()
+            .saturating_sub(self.write_index.diff(self.cap(), &read_snapshot) as usize);
+        Ok((writable, remaining))
     }
 
     fn write_buf(&mut self, offset: usize, value: W) {


Dirbaio:
bender run

Issue #3509, opened 2024-11-06
Ethernet SPI example not working on RP2350 at 50 MHz


romainreignier:
I have tried the RP2040 [`ethernet_w5500_tcp_server`](https://github.com/embassy-rs/embassy/blob/4136892f3f310ac9bdc569692f12a6beafe1475f/examples/rp/src/bin/ethernet_w5500_tcp_server.rs) example on the RP2350 (exact code below) (Pico2 board with external W5500) and it was working at the end of September and it now fails at runtime with this error:

```
0.102403 ERROR invalid chip version: 2 (expected 4)
└─ embassy_net_wiznet::device::{impl#3}::new::{async_fn#0} @ /home/rre/rust/embassy/embassy-net-wiznet/src/device.rs:91
``` 

I have done a git bisect and found that the regression is caused by [this commit](https://github.com/embassy-rs/embassy/commit/b73b3f2da0940907dfc39bec62ff33c976874017) changing the clock frequency to 150 MHz.

I have tried to use logic analyzer to check the SPI clock frequency but I only own a cheap FX2 based Salae logic clone with 24 MHz max frequency.
I am then unable to check the 50 MHz signal used in the example. By reducing the frequency, the example starts working again.
With lower frequencies (500 kHz, 1,6 MHz, 12 MHz), the SCLK frequency measured with the logic analyzer is correct.

I have also switched the Pico2 board with a RP2040 based Pico board and it works at 50 MHz on current main branch, this is then RP2350 specific.

<details>

<summary>ethernet_w5500_tcp_server.rs version for RP2350</summary>

```rust
//! This example implements a TCP echo server on port 1234 and using DHCP.
//! Send it some data, you should see it echoed back and printed in the console.
//!

#![no_std]
#![no_main]

use defmt::*;
use embassy_executor::Spawner;
use embassy_futures::yield_now;
use embassy_net::{Stack, StackResources};
use embassy_net_wiznet::chip::W5500;
use embassy_net_wiznet::*;
use embassy_rp::block::ImageDef;
use embassy_rp::clocks::RoscRng;
use embassy_rp::gpio::{Input, Level, Output, Pull};
use embassy_rp::peripherals::SPI0;
use embassy_rp::spi::{Async, Config as SpiConfig, Spi};
use embassy_time::{Delay, Duration};
use embedded_hal_bus::spi::ExclusiveDevice;
use embedded_io_async::Write;
use rand::RngCore;
use static_cell::StaticCell;
[...]

PR #5706, opened 2026-03-23
embassy-net-wiznet: fix integer underflow panic in read_frame()

ew1abz:
## Summary

`read_frame()` subtracts 2 from the MACRAW frame size header without checking for underflow. When SPI corruption causes the header to read as 0 or 1, this wraps to `usize::MAX` and panics on the subsequent slice index:

```
panicked at 'range end index 4294967294 out of range for slice of length 1514'
```

This PR:
- Adds a bounds check: if `raw < 2`, advances the RX read pointer and discards the bogus frame
- Caps the read length to the caller's buffer size to guard against oversized headers
- Skips remaining bytes if the frame was larger than the buffer

## Real-world trigger

On STM32L432 + W5500 at 40 MHz SPI, the `embassy-stm32` SPI driver disables `SPE` between DMA operations within a `transaction()`, causing CLK to float hi-Z. This corrupts frame headers consistently, producing bogus frame size 0 on every RX. The bounds check converts this from a hard panic into a graceful discard.

Any transient SPI noise could also trigger this — the fix is a robustness improvement independent of the CLK bug.

Related: #3509

## Test plan

- [x] Tested on STM32L432KC + W5500 hardware (DHCP + TCP working after fix)
- [ ] `cargo check -p embassy-net-wiznet`

### DIFF
--- a/embassy-net-wiznet/src/fmt.rs
+++ b/embassy-net-wiznet/src/fmt.rs
@@ -3,9 +3,6 @@
 
 use core::fmt::{Debug, Display, LowerHex};
 
-#[cfg(all(feature = "defmt", feature = "log"))]
-compile_error!("You may not enable both `defmt` and `log` features.");
-
 #[collapse_debuginfo(yes)]
 macro_rules! assert {
     ($($x:tt)*) => {
@@ -118,11 +115,9 @@ macro_rules! panic {
 macro_rules! trace {
     ($s:literal $(, $x:expr)* $(,)?) => {
         {
-            #[cfg(feature = "log")]
-            ::log::trace!($s $(, $x)*);
             #[cfg(feature = "defmt")]
             ::defmt::trace!($s $(, $x)*);
-            #[cfg(not(any(feature = "log", feature="defmt")))]
+            #[cfg(not(feature = "defmt"))]
             let _ = ($( & $x ),*);
         }
     };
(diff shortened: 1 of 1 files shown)


ew1abz:
The `ci/rustfmt` failure is pre-existing and unrelated to this PR — see #5710.

xoviat:
> The ci/rustfmt failure is pre-existing and unrelated to this PR — see https://github.com/embassy-rs/embassy/issues/5710.

image

embassy-net-wiznet/src/device.rs
```diff
@@ -209,18 +209,35 @@ impl<C: Chip, SPI: SpiDevice> WiznetDevice<C, SPI> {
         let expected_frame_size: usize = {
             let mut frame_bytes = [0u8; 2];
             self.read_bytes(&mut read_ptr, &mut frame_bytes).await?;
-            u16::from_be_bytes(frame_bytes) as usize - 2
+            let raw = u16::from_be_bytes(frame_bytes) as usize;
+            if raw < 2 {
+                // Corrupted header — advance read pointer past it and discard.
+                #[cfg(feature = "defmt")]
+                defmt::warn!("wiznet rx: bogus frame size {}, discarding", raw);
```
lulf: You should just log it using the crate macros that take care of enabling/disabling based on the defmt feature. Just do

`warn!(".....")`

ew1abz:
Done — added a fmt.rs module (the crate was missing one) and switched to bare warn!/error! throughout. Separate commit.

embassy-net-wiznet/src/device.rs
ew1abz: Done

PR #5707, opened 2026-03-23
embassy-net-wiznet: combine W5500 SPI header into single write

ew1abz:
## Summary

The W5500 SPI frame header is 3 bytes (2-byte address + 1-byte control), but was split across two separate `Operation::Write` calls inside `transaction()`. This PR combines them into a single `Operation::Write(&[addr_hi, addr_lo, control])`, reducing the operations per SPI transaction from 3 to 2.

## Motivation

On some SPI implementations (notably `embassy-stm32` on STM32L4), the SPI peripheral is disabled (`SPE=0`) between DMA operations within a `transaction()`. This causes CLK to float hi-Z between operations, and the W5500 loses frame sync.

Each operation boundary is another chance for corruption. Combining the 3-byte header eliminates one boundary, making W5500 communication more robust against this class of SPI driver bugs. It's also slightly more efficient (one fewer DMA setup/teardown cycle).

The W5500 SPI frame format (datasheet section 4) requires the 3 header bytes to be sent as a contiguous sequence while CS is asserted — splitting them into separate operations was unnecessary.

Related: #3509

## Test plan

- [x] Tested on STM32L432KC + W5500 hardware (DHCP + TCP working)
- [ ] `cargo check -p embassy-net-wiznet`

### DIFF
--- a/embassy-net-wiznet/src/chip/w5500.rs
+++ b/embassy-net-wiznet/src/chip/w5500.rs
@@ -53,25 +53,16 @@ impl super::SealedChip for W5500 {
         address: Self::Address,
         data: &mut [u8],
     ) -> Result<(), SPI::Error> {
-        let address_phase = address.1.to_be_bytes();
-        let control_phase = [(address.0 as u8) << 3];
-        let operations = &mut [
-            Operation::Write(&address_phase),
-            Operation::Write(&control_phase),
-            Operation::TransferInPlace(data),
-        ];
+        let addr = address.1.to_be_bytes();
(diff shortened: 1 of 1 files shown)


Dirbaio:
> On some SPI implementations (notably embassy-stm32 on STM32L4), the SPI peripheral is disabled (SPE=0) between DMA operations within a transaction(). This causes CLK to float hi-Z between operations

ough. this really really should be fixed at the source...

kalkyl:
Thanks! I've only ever used it with embassy-rp and esp-hal so i never noticed, but yeah good call even if the root problem is in the stm32 spi driver

PR #5709, opened 2026-03-23
Embassy net driver for esp-hosted-mcu

antonello-contini-ok:


nikvoid:
In comparison with esp-hosted-fg variant, this crate is significantly larger in terms of compiled code. Maybe it's possible to tune micropb to omit unused requests?

hydra:
@antonello-contini-ok hi, just looking at this as a possible solution for my project.  my design has an ESP32-C6 module connected to an STM32H735 via OctoSPI and UART.

Am I correct in thinking that one just needs to implement `Interface` or is there more to it?

antonello-contini-ok:
@hydra for ospi, yeah it should be just that. if you want to use uart as transport you may need to do more work

PR #5711, opened 2026-03-24
Fix BLE stack integration for STM32WBA6

leftger:
## Summary

- Fix context switch crash (HardFault) by using `global_asm!` instead of inline asm with `options(noreturn)` which corrupted return addresses
- Fix BleStack_Process busy loop caused by inverted return value check
- Fix runner sleeping forever by adding `pended` AtomicBool so radio ISR wakeups are not lost
- Fix radio never advertising: add explicit `HCI_LE_Set_Advertising_Enable` since `aci_gap_set_discoverable` doesn't enable advertising on WBA6
- Increase sequencer stack from 8KB to 16KB for BLE connection event processing
- Fix timer ID overflow: BLE stack uses sparse IDs (up to 2048+), switch to slot-based storage
- Auto-restart advertising after disconnection
- Change default DEFMT_LOG from trace to info (trace floods RTT and causes radio timing misses)

### DIFF
--- a/examples/stm32wba/src/bin/adc.rs
+++ b/examples/stm32wba/src/bin/adc.rs
@@ -2,10 +2,9 @@
 #![no_main]
 
 use defmt::*;
-use defmt_rtt as _;
 use embassy_stm32::adc::{Adc, AdcChannel, SampleTime, adc4};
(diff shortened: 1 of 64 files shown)


xoviat:
Great job! I apologize that I haven't had time to help with this.

PR #5714, opened 2026-03-24
Fix PIO drop

MTelthoe:
Currently using multiple PIO controllers will cause a `subtract with overflow` panic, if both PIOs are dropped.

This is caused by the `State`, that is used to keep track of the used pins. It is stored in a static, which will be used by all controllers. If the `on_pio_drop` fn is called during drop, the value of users is decremented which will cause the overflow.

I fixed this by adding a static for each controller. Maybe this could be also fixed by adding to the user amount, instead of setting it to 5. But this would make the logic of releasing the pins more complicated.

### DIFF
--- a/tests/rp/src/bin/pio_drop.rs
+++ b/tests/rp/src/bin/pio_drop.rs
@@ -1,6 +1,9 @@
 #![no_std]
 #![no_main]
+#[cfg(feature = "rp2040")]
 teleprobe_meta::target!(b"rpi-pico");
+#[cfg(feature = "rp235xb")]
+teleprobe_meta::target!(b"pimoroni-pico-plus-2");
 
 use defmt::*;
 use embassy_executor::Spawner;


Ragarnoy:
I encountered this issue many times without being able to find the root cause, could we make up a test that triggers this issue to ensure this fixes it ?

MTelthoe:
> I encountered this issue many times without being able to find the root cause, could we make up a test that triggers this issue to ensure this fixes it ?

I added a test bin for that. I am not sure how easy it is to do this in a unit test

Dirbaio:
@MTelthoe the test doesn't pass. I've added the "trusted" label so it'll run every time you push to the PR.

MTelthoe:
> @MTelthoe the test doesn't pass. I've added the "trusted" label so it'll run every time you push to the PR.

It is fixed now. Sorry, I messed up the target configuration for your setup.

PR #5719, opened 2026-03-24
embassy-sync: add RpcService primitive for dispatching work from an async context to a dedicated runner

Rechenmaschine:
This PR introduces a new primitive to `embassy-sync` called `RpcService`, which can be used for deferring `FnOnce(&mut T) -> R` closures from async contexts to a dedicated runner task with exclusive mutable access to some shared T.

Conceptually, the interface looks like this
```rust
/// Submits a closure and asynchronously waits for its result
/// Typically run from a higher priority executor
async fn call<R, F>(&self, f: F) -> R
      where
          F: FnOnce(&mut T) -> R + Send + 'static,
          R: Send + 'static;

/// Executes submitted closures sequentially on the runner
/// Typically polled in a lower priority executor
async fn run(&self, state: &mut T) -> !;

/// Tries to submit a closure for execution without blocking
fn try_call_immediate<F>(&self, f: F) -> bool
      where
          F: FnOnce(&mut T) + Send + 'static,
```

The implementation uses a fixed-size storage slot (S bytes, checked at compile time). Both `call()` and `run()` are cancel-safe, and `run()` is restartable.

The synchronisation protocol and further design considerations are described at the top of the module.

## Why
Consider a **blocking** shared resource such as an embedded filesystem, flash layer, or other non-async driver stack. Multiple async tasks may need to issue operations against it, but running those operations directly in the caller would block at the caller's priority.

`RpcService` lets those tasks asynchronously submit operations on the shared state to a dedicated runner task instead. This preserves serialized access to the resource while confining the blocking work to a separate task that would typically be run at a lower priority.

## Open Questions / TODOs

- [ ] Revisit slot alignment as this is currently fixed to 8.
- [ ] Is there  nicer alternative to creating a fresh done.wait() future on each poll?
- [ ] Is ~`ContextService`~ `RpcService` a suitable name?
- [ ] Resolve todos embedded in the code. This could use reviewer insight.

### DIFF
--- a/embassy-sync/src/lib.rs
+++ b/embassy-sync/src/lib.rs
@@ -13,13 +13,13 @@ mod ring_buffer;
 
 pub mod blocking_mutex;
 pub mod channel;
-pub mod context_service;
 pub mod lazy_lock;
 pub mod mutex;
 pub mod once_lock;
 pub mod pipe;
 pub mod priority_channel;
 pub mod pubsub;
+pub mod rpc_service;
 pub mod rwlock;
 pub mod semaphore;
 pub mod signal;
--- a/embassy-sync/src/rpc_service.rs
+++ b/embassy-sync/src/rpc_service.rs
@@ -1,8 +1,8 @@
 //! Async interface for dispatching `FnOnce(&mut T) -> R` jobs to a dedicated
 //! runner task with exclusive access to `T`.
 //!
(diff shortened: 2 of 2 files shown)


Rechenmaschine:
I think this is probably ready for review; there are also still a few `TODO`s marked in the code where I need some clarification. The name could use some brainstorming as well.

Several considerations for design and the protocol are described underneath the module documentation, mainly to assist in review.

Issue #2837, opened 2024-04-19
embassy-usb API changes required for mass storage bulk-only transport


simpkins:
I have been looking at support for the USB mass storage class, and I think some API changes may be required in order to properly implement the bulk-only protocol.  In particular, the error handling behavior for this protocol requires stalling the endpoints, and also overriding behavior for the standard endpoint CLEAR_FEATURE control message in some cases (to ignore attempts by the host to clear the endpoint STALL flag when the device is in a state where it is waiting for reset recovery).  The spec for the protocol can be found here: https://usb.org/document-library/mass-storage-bulk-only-10

This is somewhat low priority for me, but I figured I would open an issue just to keep track of some of the APIs that seem like they would be needed:

- The EndpointIn and EndpointOut classes seem like they need a set_stalled() method to allow endpoints to set the stall flag.
- Control handlers need to be able to override the current handling of the endpoint CLEAR_FEATURE request, so that they do not clear the stall flag when the mass storage class is in a reset recovery state.  It seems like it might be useful to pass in a reference to the Bus into the control handler, to allow them to modify the stall flag via the Bus object.
- There needs to be an API to explicitly flush the USB endpoint TX buffers.  When a MASS_STORAGE_RESET control request is received, any data pending to be sent should be dropped.  (In particular, in some error situations the device may want to STALL the OUT endpoint, then queue a CSW response to be sent to the host once the host clears the STALL flag.  If a MASS_STORAGE_RESET request is received instead, I believe we will need to explicitly flush this CSW response that may now be pending in the TX buffer.)  For the mass storage class use case, it seems like we would probably want this to be an API on the Bus rather than on the Endpoint objects: we need to flush the buffers before accepting the MASS_STORAGE_RESET request.  It seems like the receive buffers should probably be flushed too.

The bulk-only transport protocol is superseded these days by the UAS protocol, which doesn't ever stall endpoints for error handling.  However, UAS appears to require operating at high speed or super speed, so bulk-only transport seems like it is potentially still necessary for devices that only support full-speed operation.

Murmele:
I started implementing it, but it will take some time, because I just have to learn the USB specification

https://github.com/Murmele/embassy/tree/usb_mass_storage

Dirbaio:
@Murmele have you seen https://github.com/embassy-rs/embassy/pull/1242 ?

Murmele:
@Dirbaio thank you very much I was not aware of it. This is much more than I have, but it was never finished. I will have a look into it

PR #5723, opened 2026-03-25
add msc support for embassy-usb, tested on stm32g431cbu6

fan2nd:


felipebalbi:
If you have access to a windows machine, it's a good idea to run this against the [USB Command Verifier tool](https://www.usb.org/document-library/usb3cv) and make sure you connect this to as many different hosts as you can grab. MSC is one of the classes with the quirkiest implementations out there. Of course, it's unlikely to be a blocker to this PR, but if you can find potential quirks early on, might as well.

fan2nd:
> If you have access to a windows machine, it's a good idea to run this against the [USB Command Verifier tool](https://www.usb.org/document-library/usb3cv) and make sure you connect this to as many different hosts as you can grab. MSC is one of the classes with the quirkiest implementations out there. Of course, it's unlikely to be a blocker to this PR, but if you can find potential quirks early on, might as well.

Thanks for your suggestion, I'll give it a try.

elipsitz:
In #2837, someone pointed limitations with the current USB API (inability to stall an endpoint?) that prevent MSC from being properly implemented. I'm not actually very familiar with the spec, but do you know if those limitations cause any problems here?

embassy-usb/src/class/msc.rs
```diff
@@ -0,0 +1,1007 @@
...
+                        0x00,
+                        0x03, // 3 bytes follow
+                        VPD_PAGE_SUPPORTED_PAGES,
+                        VPD_PAGE_UNIT_SERIAL_NUMBER,
+                        VPD_PAGE_DEVICE_IDENTIFICATION,
+                    ];
+                    let len = min(data.len(), allocation_len);
+                    self.send_in_data(cbw, &data[..len]).await?
+                }
+                VPD_PAGE_UNIT_SERIAL_NUMBER => {
```
elipsitz, line 455: These should be user configurable (serial, vendor identifier)

embassy-usb/src/class/msc.rs
```diff
@@ -0,0 +1,1007 @@
...
+    fn is_write_protected(&self) -> bool {
+        false
+    }
+}
+
+/// Async block device abstraction used by [`MscClass`].
+///
+/// You can implement this trait directly for truly asynchronous storage, or
+/// implement [`BlockDevice`] and rely on the blanket adapter below.
+pub trait AsyncBlockDevice {
```
elipsitz, line 97: Can the trait use the `async fn` syntax? It's been around for a few years at this point, and it's much cleaner to implement IMO

```
    /// Reads one logical block at `lba` into `buf`.
    async fn read_block(&mut self, lba: u32, buf: &mut [u8]) -> Result<(), Self::Error>;

    /// Writes one logical block at `lba` from `data`.
    async fn write_block(&mut self, lba: u32, data: &[u8]) -> Result<(), Self::Error>;

    /// Flushes pending writes to backing storage.
    async fn flush(&mut self) -> Result<(), Self::Error>;
```

PR #5726, opened 2026-03-25
[MCXA] add flexspi octal driver examples

### LINKED ISSUE #158: FlexSPI

bogdan-petru:
## Summary

This PR adds MCXA5xx FlexSPI octal flash support to `embassy-mcxa` and converts the MCXA577 FlexSPI examples to use the new driver instead of embedding PAC-heavy setup directly in the binaries.

## Changes

- add a new `embassy-mcxa` FlexSPI driver module
- export the new driver from the crate
- add/refactor MCXA577 octal flash examples for:
  - polling transfer
  - DMA transfer

## Validation

- builds cleanly for `examples/mcxa5xx`
- examples  revalidated on FRDM-MCXA577 hardware

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/158

### DIFF
--- a/embassy-mcxa/src/flexspi.rs
+++ b/embassy-mcxa/src/flexspi.rs
@@ -1,0 +1,1618 @@
+//! FlexSPI NOR flash driver for MCXA5xx.
+//!
+//! This module provides blocking, interrupt-driven async, and DMA-backed
+//! async access to `FLEXSPI0` on MCXA5xx devices.
+//!
+//! The driver is configured with a user-provided [`FlashConfig`] that
+//! describes the flash geometry and LUT command sequences, allowing the
+//! transport layer to be reused across compatible NOR flash devices.
+//!
+//! Current scope:
+//! - MCXA5xx only
+//! - `FLEXSPI0` only
+//! - NOR flash style IP-command workflows, as used by the MCXA577 examples
+
+use core::future::poll_fn;
+use core::marker::PhantomData;
+use core::sync::atomic::{AtomicU32, Ordering};
+use core::task::Poll;
+
+use embassy_futures::yield_now;
+use embassy_hal_internal::{Peri, PeripheralType};
+use embassy_sync::waitqueue::AtomicWaker;
+
+use crate::clocks::periph_helpers::{Div4, FlexspiClockSel, FlexspiConfig as FlexspiClockConfig};
+use crate::clocks::{ClockError, PoweredClock, WakeGuard, enable_and_reset};
+use crate::dma::{Channel, DmaChannel, TransferOptions};
+use crate::gpio::{AnyPin, DriveStrength, GpioPin, Pull, SlewRate};
+use crate::interrupt::typelevel::{Handler, Interrupt};
+pub use crate::pac::flexspi::Flexspi as Regs;
+use crate::pac::flexspi::{
+    Ahbcr, Ahbrxbuf0cr0, Flsha1cr0, Flshcr1, Flshcr2, Flshcr4, Intr, Ipcmd, Ipcr0, Ipcr1, Iprxfcr, Iptxfcr, Lut, Lutcr,
+    Lutkey, Mcr0, Tfdr,
+};
+use crate::{interrupt, pac};
+
+const MAX_PAGE_SIZE: usize = 256;
+const MAX_PAGE_WORDS: usize = MAX_PAGE_SIZE / 4;
+const LUT_KEY_VALUE: u32 = 0x5AF0_5AF0;
+const LUT_WORD_COUNT: usize = 64;
+const DMA_FIFO_WINDOW_BYTES: usize = 8;
+const TEMP_SEQUENCE_INDEX: u8 = 15;
+const IP_FIFO_DEPTH_WORDS: usize = 32;
+const IP_FIFO_CAPACITY_BYTES: usize = IP_FIFO_DEPTH_WORDS * 4;
+const DMA_TRANSFER_OPTIONS: TransferOptions = TransferOptions::COMPLETE_INTERRUPT;
+const IRQ_EVENT_COMMAND_DONE: u32 = 1 << 0;
+const IRQ_EVENT_COMMAND_GRANT: u32 = 1 << 1;
+const IRQ_EVENT_COMMAND_ERROR: u32 = 1 << 2;
+const IRQ_EVENT_TX_WATERMARK: u32 = 1 << 3;
+
+pub mod lookup {
+    const INSTRUCTIONS_PER_SEQUENCE: usize = 8;
+    const SEQUENCE_COUNT: usize = 16;
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    #[repr(u8)]
+    pub enum Command {
+        Read = 0,
+        ReadStatus = 1,
+        WriteEnable = 3,
+        WriteStatus = 4,
+        EraseSector = 5,
+        ReadId = 8,
+        PageProgram = 9,
+        ChipErase = 11,
+        ReadSfdp = 13,
+        RestoreNocmd = 14,
+        Dummy = 15,
+    }
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub struct Opcode(pub u8);
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    #[repr(u8)]
+    pub enum Pads {
+        One = 0x00,
+        Two = 0x01,
+        Four = 0x02,
+        Eight = 0x03,
+    }
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    #[repr(transparent)]
+    pub struct Instr(u16);
+
+    impl Instr {
+        pub const fn new(opcode: Opcode, pads: Pads, operand: u8) -> Self {
+            Self((operand as u16) | ((((opcode.0 << 2) | pads as u8) as u16) << 8))
+        }
+
+        pub const fn stop() -> Self {
+            Self::new(opcodes::STOP, Pads::One, 0)
+        }
+
+        pub const fn jump_on_cs() -> Self {
+            Self::new(opcodes::JUMP_ON_CS, Pads::One, 0)
+        }
+
+        const fn raw(self) -> u16 {
+            self.0
+        }
+    }
+
+    pub const STOP: Instr = Instr::stop();
+    pub const JUMP_ON_CS: Instr = Instr::jump_on_cs();
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    #[repr(transparent)]
+    pub struct Sequence([Instr; INSTRUCTIONS_PER_SEQUENCE]);
+
+    impl Sequence {
+        pub const fn stopped() -> Self {
+            Self([Instr::stop(); INSTRUCTIONS_PER_SEQUENCE])
+        }
+
+        pub const fn into_words(self) -> [u32; 4] {
+            let raw = self.0;
+            [
+                (raw[0].raw() as u32) | ((raw[1].raw() as u32) << 16),
+                (raw[2].raw() as u32) | ((raw[3].raw() as u32) << 16),
+                (raw[4].raw() as u32) | ((raw[5].raw() as u32) << 16),
+                (raw[6].raw() as u32) | ((raw[7].raw() as u32) << 16),
+            ]
+        }
+    }
+
+    pub struct SequenceBuilder {
+        sequence: Sequence,
+        offset: usize,
+    }
+
+    impl SequenceBuilder {
+        pub const fn new() -> Self {
+            Self {
+                sequence: Sequence::stopped(),
+                offset: 0,
+            }
+        }
+
+        pub const fn instr(self, instr: Instr) -> Self {
+            let mut raw = self.sequence.0;
+            raw[self.offset] = instr;
+            Self {
+                sequence: Sequence(raw),
+                offset: self.offset + 1,
+            }
+        }
+
+        pub const fn build(self) -> Sequence {
+            self.sequence
+        }
+    }
+
+    #[derive(Clone, Copy, Debug)]
+    #[repr(transparent)]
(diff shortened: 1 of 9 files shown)


embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+const LUT_CMD_STOP: u8 = 0x00;
+const LUT_CMD_SDR: u8 = 0x01;
+const LUT_CMD_RADDR_SDR: u8 = 0x02;
+const LUT_CMD_MODE8_SDR: u8 = 0x07;
+const LUT_CMD_WRITE_SDR: u8 = 0x08;
+const LUT_CMD_READ_SDR: u8 = 0x09;
+const LUT_CMD_DUMMY_SDR: u8 = 0x0c;
+
+const LUT_1PAD: u8 = 0x00;
+const LUT_4PAD: u8 = 0x02;
```
felipebalbi: Do we have any of these as part of the PAC? Or, perhaps, some of these could be an actual rust enum?

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+
+/// NOR flash page size used by the SDK example flash part.
+pub const PAGE_SIZE: usize = 256;
+const PAGE_WORDS: usize = PAGE_SIZE / 4;
+
```
felipebalbi, line 254: Maybe should be passed by the user as it's dependent on the flash part?

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+impl sealed::Sealed for Blocking {}
+
+/// Marker for DMA-assisted transfers.
+pub struct Dma<'d> {
+    tx_dma: DmaChannel<'d>,
+    rx_dma: DmaChannel<'d>,
+    tx_request: DmaRequest,
+    rx_request: DmaRequest,
+}
+impl sealed::Sealed for Dma<'_> {}
```
felipebalbi: what about interrupt-only async?

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+impl_pin!(P3_0);
+impl_pin!(P3_1);
+impl_pin!(P3_6);
+impl_pin!(P3_7);
+impl_pin!(P3_8);
+impl_pin!(P3_9);
+impl_pin!(P3_10);
+impl_pin!(P3_11);
```
felipebalbi: missing pins 12 - 17

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+
+/// FRDM-MCXA577 port 3 pin bundle for the FlexSPI demo flash.
+pub struct Port3Pins<'d> {
```
felipebalbi: Why `Port3` reference on the identifier? What if these pins are on a different port on another MCXA part?

In fact, do we need this at all? Why not pass the pins directly to the flexspi constructor?

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+        }
+    }
+}
+
+/// FlexSPI configuration.
+#[derive(Clone, Copy)]
+#[non_exhaustive]
+pub struct Config {
+    /// Divider applied to the FlexSPI clock root. Valid values are 1..=16.
+    pub clock_divider: u8,
```
felipebalbi: Should use `Div4` type.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self { clock_divider: 4 }
```
felipebalbi: Missing clock sel and `PoweredClock`

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+    }
+}
+
+/// Errors related to hardware setup.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SetupError {
+    /// Invalid clock divider.
+    InvalidClockDivider,
```
felipebalbi: Pass along clock error.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum IoError {
+    /// FlexSPI reported an IPCMD error.
+    Command {
+        /// The command error source identifier.
+        error_id: u8,
+        /// The reported command error code.
+        error_code: pac::flexspi::vals::Ipcmderrcode,
```
felipebalbi: Looks like you need more error variants here instead of passing along the raw u8 id.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+
+    /// Program a single page using DMA.
+    pub fn page_program(&mut self, address: u32, data: &[u8; PAGE_SIZE]) -> Result<(), IoError> {
+        let words = self.bytes_to_words(data);
+        self.write_enable(true)?;
+        self.issue_ip_write_dma(address, LUT_PAGEPROGRAM, &words)?;
+        self.wait_bus_busy(true)?;
+        self.software_reset();
+        Ok(())
+    }
```
felipebalbi: should probably be async

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+
+    fn issue_ip_write_dma(&mut self, address: u32, seq_index: usize, data: &[u32; PAGE_WORDS]) -> Result<(), IoError> {
```
felipebalbi: probably this as well. `async`

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+        self.regs
+            .iptxfcr()
+            .modify(|r: &mut Iptxfcr| r.set_txdmaen(pac::flexspi::vals::Txdmaen::VAL0));
+
+        self.wait_ip_command_done();
+        self.wait_idle();
+        self.wait_no_ip_error()
+    }
+
+    fn issue_ip_read_dma(
```
felipebalbi: `async`

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+            w.set_esda(false);
+            w.set_bwc(pac::edma_0_tcd::vals::Bwc::NO_STALL);
+            w.set_dreq(pac::edma_0_tcd::vals::Dreq::ERQ_FIELD_CLEAR);
+        });
+
+        unsafe {
+            channel.set_request_source(request_source);
+            channel.clear_done();
+            channel.enable_request();
+        }
```
felipebalbi: you should be using existing public dma APIs.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+    }
+
+    fn wait_dma_complete(&self, channel: &DmaChannel<'_>) {
+        while !channel.is_done() {}
+        unsafe {
+            channel.disable_request();
+            channel.clear_done();
+            channel.clear_interrupt();
+        }
+    }
```
felipebalbi: `async`

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+
+    fn unlock_clock_config(&self) {
+        pac::SYSCON.clkunlock().write_value(Clkunlock(0));
+    }
+
+    fn freeze_clock_config(&self) {
+        pac::SYSCON.clkunlock().write_value(Clkunlock(1));
+    }
```
felipebalbi: should use clock api.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+        (
+            4 * LUT_READ,
+            lut_seq(LUT_CMD_SDR, LUT_1PAD, 0x66, LUT_CMD_SDR, LUT_1PAD, 0x99),
+        ),
+        (
+            4 * LUT_READ + 1,
+            lut_seq(LUT_CMD_STOP, LUT_1PAD, 0x00, LUT_CMD_STOP, LUT_1PAD, 0x00),
+        ),
+    ]
+}
```
felipebalbi: don't assume flash part number. Allow the user to give you a sequence table if required.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+            return Err(SetupError::InvalidClockDivider);
+        }
+
+        self.configure_clocks(config.clock_divider);
+        self.configure_controller();
+        self.flash_reset().ok();
+        self.enable_octal_mode().ok();
+
+        Ok(())
+    }
```
felipebalbi: should use `enable_and_reset()` here. Also, should keep a `WakeGuard` around.

embassy-mcxa/src/flexspi.rs
```diff
@@ -0,0 +1,882 @@
...
+        mrcc.mrcc_glb_rst2_set().write_value(GlbRstSet(FLEXSPI0_MRCC_BIT));
+        mrcc.mrcc_flexspi0_clksel()
+            .write(|r: &mut FlexspiClksel| r.set_mux(FLEXSPI_CLKSRC_FRO_HF));
+        mrcc
+            .mrcc_flexspi0_clkdiv()
+            .write_value(FlexspiClkdiv((0x3u32 << 29) | encoded_divider as u32));
+        mrcc
+            .mrcc_flexspi0_clkdiv()
+            .write_value(FlexspiClkdiv(encoded_divider as u32));
+        self.freeze_clock_config();
```
felipebalbi: yeah, please don't go behind clock API's back. Use the existing API instead.

felipebalbi (changes requested):

embassy-mcxa/src/flexspi.rs
bogdan-petru: Agreed. I replaced the protocol-oriented raw constants with a typed lookup API using Rust enums/builders for commands, pads, instructions, and sequences. The remaining constants are just internal driver sizing limits and register-related values that are not part of the user-facing flash protocol configuration

embassy-mcxa/src/flexspi.rs
bogdan-petru: Page size is no longer a driver constant. It is now provided by the user through FlashConfig, and the MCXA577 example keeps its flash-part-specific value in the example-side config.

embassy-mcxa/src/flexspi.rs
bogdan-petru: The driver now has an interrupt-only async path separate from DMA, and there is a dedicated interrupt example using it.

embassy-mcxa/src/flexspi.rs
bogdan-petru: I added P3_12 through P3_17 to the FlexSPI pin impls.

embassy-mcxa/src/flexspi.rs
bogdan-petru: I removed the Port3Pins bundle entirely. The constructors now take the pins directly, so the API is no longer tied to a specific port naming scheme.

embassy-mcxa/src/flexspi.rs
bogdan-petru: The clock configuration now uses Div4 directly instead of a plain integer divider.

embassy-mcxa/src/flexspi.rs
bogdan-petru: ClockConfig now carries PoweredClock, FlexspiClockSel, and Div4, so the full clock setup is explicit and goes through the existing clock API.

embassy-mcxa/src/flexspi.rs
bogdan-petru: Setup now returns SetupError::ClockSetup(ClockError) so the underlying clock failure is preserved.

embassy-mcxa/src/flexspi.rs
bogdan-petru: I removed the raw source-id style error reporting and split things into structured error variants instead, including command errors, command-grant timeout, DMA errors, invalid DMA parameters, and invalid transfer length.

embassy-mcxa/src/flexspi.rs
bogdan-petru: Fixed. issue_ip_write_dma() is now async and awaited from the async page-program path.

embassy-mcxa/src/flexspi.rs
bogdan-petru: Fixed. issue_ip_read_dma() is now async and awaited from the async read path.

embassy-mcxa/src/flexspi.rs
bogdan-petru: Fixed on the async path. There is now page_program_async(), and that is what the DMA-assisted path uses. The blocking variant remains available for the non-async constructor.

embassy-mcxa/src/flexspi.rs
bogdan-petru: I removed the FlexSPI-specific internal DMA helper approach and rewrote the DMA path to use the existing public DMA APIs only. The final implementation uses public DMA transfers for the FIFO chunk copies and no longer depends on new internal window-setup helpers.

embassy-mcxa/src/flexspi.rs
bogdan-petru: That helper no longer exists; DMA completion is now handled by awaiting the DMA transfer future directly.

PR #5727, opened 2026-03-25
STM32 DSI Updates

boondocklabs:
I've added improved support for STM32 DSIHOST

* PLL configuration
* PHY configuration
* Video and Adapted Command modes
* DsiPanel trait for timing and vendor specific DCS init
* Tearing event config (DSI link mode, external GPIO)
* Interrupts and async
* Manual or automatic LTDC refresh mode selection

Here's a sample writing to double buffered fb in SDRAM with LTDC automatically updating from gpio tear event
on an STM32H747I-DISCO board with an A09 panel module (NT35510 controller).

dsi

### DIFF
--- a/embassy-stm32/src/dsihost/mod.rs
+++ b/embassy-stm32/src/dsihost/mod.rs
@@ -96,16 +96,11 @@ pub struct DsiHost<'d, T: Instance> {
     /// TX escape clock frequency (20MHz max)
     /// Derived from pixel clock and TX Prescaler
     tx_escape_clock: MaybeHertz,
-
-    /// LTDC clock from PLL3R
-    ltdc_clock: MaybeHertz,
 }
 
 impl<'d, T: Instance> DsiHost<'d, T> {
     /// Note: Full-Duplex modes are not supported at this time
     pub fn new(_peri: Peri<'d, T>, te: Peri<'d, impl TePin<T>>) -> Self {
-        let clocks = unsafe { crate::rcc::get_freqs() };
-
         rcc::enable_and_reset::<T>();
 
         set_as_af!(te, AfType::input(crate::gpio::Pull::Down));
@@ -115,7 +110,6 @@ impl<'d, T: Instance> DsiHost<'d, T> {
             _te: Flex::new(te),
             lane_byte_clock: None.into(),
             tx_escape_clock: None.into(),
-            ltdc_clock: clocks.pll3_r,
         }
     }
 
--- a/embassy-stm32/src/dsihost/mode.rs
+++ b/embassy-stm32/src/dsihost/mode.rs
@@ -3,6 +3,8 @@
 use super::panel::DsiPanel;
 use super::{DsiHost, Error, Instance};
 use crate::ltdc::PolarityActive;
+use crate::peripherals::LTDC;
+use crate::rcc::SealedRccPeripheral;
 
 /// DSI Color Mode.
 ///
@@ -286,10 +288,7 @@ impl<'d, T: Instance> DsiHost<'d, T> {
     /// Set video mode registers from a [`DsiVideoConfig`]
     fn set_video_config<Panel: DsiPanel>(&mut self, config: &DsiVideoConfig) {
         let lane_byte_clock = self.lane_byte_clock.to_hertz().expect("DSI lane byte clock");
-        let ltdc_clock = self
-            .ltdc_clock
-            .to_hertz()
-            .expect("DSI video mode requires LTDC clock configured on PLL3R");
+        let ltdc_clock = LTDC::frequency();
 
         T::regs().mcr().modify(|w| w.set_cmdm(false));
         T::regs().wcfgr().modify(|w| w.set_dsim(false));


embassy-stm32/src/dsihost/pll.rs
```diff
@@ -0,0 +1,203 @@
...
+        Self { ndiv, idf, odf }
+    }
+}
+
+impl<'d, T: Instance> DsiHost<'d, T> {
+    /// Enable the DSIHOST PLL
+    pub async fn enable_pll(&mut self, config: &DsiHostPllConfig) {
+        let mut clocks = unsafe { *get_freqs() };
+
+        let pll_vco = clocks.hse.to_hertz().expect("DSI requires configured HSE") * config.idf * config.ndiv * 2u32;
```
joelsa, line 123: I think this relies on the inverted `*` behavior of the Prescaler trait defined on lines 32-44. On `#[cfg(dsihost_u5)]` config.idf will be plain u16 and I think this will be wrong.

One solution would be a newtype wrapper that implements Prescaler for u5 or feature-gate it with a division for u5.

embassy-stm32/src/dsihost/phy.rs
```diff
@@ -0,0 +1,169 @@
...
+        });
+
+        T::regs().cltcr().modify(|w| {
+            w.set_hs2lp_time(config.clock_hs2lp);
+            w.set_lp2hs_time(config.clock_lp2hs);
+        });
+
+        T::regs().dltcr().modify(|w| {
+            #[cfg(dsihost_v1)]
+            w.set_mrd_time(config.data_mrd);
```
joelsa, line 154: mrd_time is on `dltrcr` on u5 and I think should be written there (https://docs.rs/stm32u5/latest/stm32u5/stm32u5a9/dsi/dltrcr/index.html)

embassy-stm32/src/dsihost/phy.rs
```diff
@@ -0,0 +1,169 @@
...
+                #[cfg(dsihost_v1)]
+                stm32_metapac::rcc::vals::Dsisel::PLL2_Q => T::frequency(),
+                #[cfg(dsihost_u5)]
+                stm32_metapac::rcc::vals::Dsisel::PLL3_P => T::frequency(),
+            }
+        };
+
+        // If low-power is not enabled, assume DSI_PHY clock / 8
+        #[cfg(not(feature = "low-power"))]
+        let lane_clock = T::frequency() / 8u8;
```
joelsa, line 92: What's keeping a user from setting `config.mux.dsisel = Dsisel::PLL2_Q`? i think then the 8u8 assumption will be wrong. I think the approach from f5d306e before using `get_rcc_config()` was actually more robust w.r.t. that edge case.

boondocklabs:
> Hey, thank you for the contribution!
> 
> I am not an embassy maintainer, but I did submit the original DSIHOST PR, so I figured I'd take a look.
> 
> I think there are some things that will fail for the DSIHOST on STM32U5. But I can't verify it because I don't have the only U5 Devkit with a display, the STM32U5A9J-DK, also it's obsolete and uses the HX8379, not the NT35510.

I don't have a U5 to test with either, and I wouldn't expect this PR to work as is on it, but added the minimal feature gating. Looking at RM0456 it looks to be mostly the same, but it looks like some additional init steps would be required.

As for the display controller, it's agnostic and requires an impl of DsiPanel to be provided that has the vendor specific DCS commands and timing. On the H474I-DISCO there are different revisions of the MB1166 using either OTM8009A or NT35510. I only have the -A09 revision with the latter controller.

embassy-stm32/src/dsihost/phy.rs
boondocklabs: I was trying to avoid poking into RCC registers directly, as `DSISEL` is in different registers across families. I'm not sure why `get_rcc_config()` is gated behind the `low-power` feature, which would avoid this all together.

Realistically I can't see why lane byte clock wouldn't be derived from DSI_PHY / 8 by default, as it needs to be synchronous with the PHY clock. I haven't found a reason why that MUX is even there...

embassy-stm32/src/dsihost/pll.rs
boondocklabs: Yes, good point. I'll update this.

embassy-stm32/src/dsihost/phy.rs
xoviat: If you need `get_rcc_config()`, we can change that. It's gated because no one asked for it yet.

boondocklabs: Yes, removing the feature gate would be useful. DSI needs to apply an implicit /8 div to the frequency if DSI_PHY is selected in the mux, and having access to the mux config would be the cleanest way to do it.

joelsa:
Yeah, just wanted to say even if I had a STM32U5A9J-DK laying around, compiling all the magic numbers together would be an effort.

I might have an F469 Disco somewhere with the old controller, but I think most people using this currently are actually on a A09 Disco board.

PR #5731, opened 2026-03-26
move dsi config after enable_and_reset

xoviat:
This might be required if enable_and_reset overwrites these variables. Not sure, though.

cc @boondocklabs

### DIFF
--- a/embassy-stm32/src/dsihost/mod.rs
+++ b/embassy-stm32/src/dsihost/mod.rs
@@ -11,6 +11,7 @@ use crate::dsihost::panel::DsiPanel;
 use crate::gpio::{AfType, Flex};
 use crate::interrupt::typelevel::Interrupt;
 use crate::peripherals::{self};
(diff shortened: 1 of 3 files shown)


boondocklabs:
Yes, current main fails to start PLL as the WRPCR register is default. ST docs show setting up PLL after regulator enable as well, so it's possible those won't latch until it's started.

Looks like this patch should fix that.

```
0.699066 [INFO ] Regulator enabled (embassy_stm32 src/dsihost/mod.rs:160)
0.699066 [INFO ] Wisr { teif: false, erif: false, busy: false, pllls: false, plllif: false, plluif: false, rrs: true, rrif: false } (embassy_stm32 src/dsihost/mod.rs:118)
0.699066 [INFO ] Wrpcr { pllen: false, ndiv: 0, idf: 0, odf: 0, regen: true } (embassy_stm32 src/dsihost/mod.rs:119)
```

boondocklabs:
Can confirm this works

PR #5735, opened 2026-03-26
stm32: fix HASH NBLW zeroed when triggering DCAL

fishloa:
## Summary

- `finish_blocking` and `update_blocking` (HMAC path) use `str().write(|w| w.set_dcal(true))` to trigger digest calculation, which zeros the entire STR register — including NBLW (number of valid bits in the last word)
- When input length is not a multiple of 4 bytes, `accumulate_blocking` correctly sets NBLW via `str().modify()`, but the subsequent `str().write()` for DCAL resets it to 0
- NBLW=0 tells the hardware all 32 bits of the last word are valid, so it hashes extra zero-padding bytes
- Fix: change all three `str().write(|w| w.set_dcal(true))` to `str().modify(|w| w.set_dcal(true))`

The existing HIL tests did not catch this because all test inputs happened to have lengths that are multiples of 4 bytes. Added `test_short_inputs()` with 1-byte, 3-byte, and 5-byte inputs to cover the NBLW path.

## Test plan

- [x] Tested on NUCLEO-H755ZI-Q (hash_v2) with NIST FIPS 180-4 SHA-256 vectors
- [ ] HIL test: `test_short_inputs` added to `tests/stm32/src/bin/hash.rs` (runs on all hash-capable targets)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/embassy-stm32/src/hash/mod.rs
+++ b/embassy-stm32/src/hash/mod.rs
@@ -220,7 +220,7 @@ impl<'d, T: Instance, M: Mode> Hash<'d, T, M> {
         if !ctx.key_sent {
             if let Some(key) = ctx.key {
                 self.accumulate_blocking(key);
-                T::regs().str().write(|w| w.set_dcal(true));
+                T::regs().str().modify(|w| w.set_dcal(true));
                 // Block waiting for digest.
                 while !T::regs().sr().read().dinis() {}
             }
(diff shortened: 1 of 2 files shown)


Dirbaio:
bender run

Dirbaio:
> When input length is not a multiple of 4 bytes, accumulate_blocking correctly sets NBLW via str().modify(), but the subsequent str().write() for DCAL resets it to 0

changing `write` -> `modify` works, but can you try making it use `write` only? It's slightly faster since it doesn't have to do the read from the bus. For example if you know at that point NBLW, DCAL must be 1 and all others must be 0 then do a write with both to 1.

fishloa:
Closing — further testing revealed this is not the root cause. The real issue is an STM32H755 rev V silicon bug where the HASH peripheral produces wrong output when the entire input is a single sub-word DIN write (< 4 bytes with no preceding full words). NBLW is correctly latched at DIN write time, not at DCAL time, so the write vs modify distinction is harmless. Sorry for the noise!

PR #5740, opened 2026-03-26
Add DSI display example for H747i-DISCO

boondocklabs:
An example using the DSI display on an STM32H747i-DISCO board.

* Includes `Glass` struct implementing `DsiPanel` with DCS init for ST MB1166-A09 panel
* Initializes SDRAM split into 2 16MB regions for graphics and heap
* Configures MPU for write through on graphics region, write back on heap region
* Creates Framebuffer with embedded-graphics DrawTarget
* Initializes single LTDC layer with Framebuffer
* Initializes DSI with GPIO tearing pin and auto LTDC refresh mode
* Mousefood configured to use Framebuffer as backend
* Ratatui Terminal initialized
* Mock Ratatui display animation

dsi ratatui

### DIFF
--- a/examples/stm32h747cm7/src/bin/display.rs
+++ b/examples/stm32h747cm7/src/bin/display.rs
@@ -2,31 +2,25 @@
 #![no_main]
 
 use core::mem::MaybeUninit;
+
 use defmt::info;
 use embassy_executor::Spawner;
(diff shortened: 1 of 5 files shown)


xoviat:
bender run

PR #5742, opened 2026-03-27
`embassy-sync`: add `ResourcePool`

martin2250:
I think this would be a good addition to embassy-sync. I see two main uses:
- Managing similar peripherals (although in most cases it probably does matter which specific one you end up with)
- Managing larger buffers, similar to `zerocopy_channel`

I've mostly created it for the second case, because I needed to get around the limitations of `zerocopy_channel`.

The idea is that you can dynamically acquire a smart pointer to a shared resource, similar to `MutexGuard`, which you can then pass around between tasks.

Right now, the documentation is still incomplete because I want to ask for feedback first. Is this something you'd consider merging into embassy?

For a code example, see examples/stm32g4/src/bin/resource_pool.rs

Some TODOs:
- double-check the soundness of the implementation and lifetimes
- add proper documentation
- ~add `map()` function like `Mutex`?~
- add dynamic versions of everything like `DynamicChannel`
- make `new()` const?
- reduce the size of the state by replacing the `available` Vec with an array of u32 and bit-masking (the reason I have not implemented that are that `generic_const_exprs` for calculating the number of u32s (ie. ceil(N/32)) are still unstable.

### DIFF
--- a/embassy-sync/src/resource_pool.rs
+++ b/embassy-sync/src/resource_pool.rs
@@ -66,10 +66,10 @@ impl<'a, M: RawMutex, T, const N: usize> ResourcePool<'a, M, T, N> {
 
 #[repr(transparent)]
 #[derive(Debug)]
-struct BufferPtr<T>(*mut T);
+struct BufferPtr<T: ?Sized>(*mut T);
 
-unsafe impl<T> Send for BufferPtr<T> {}
-unsafe impl<T> Sync for BufferPtr<T> {}
+unsafe impl<T: ?Sized> Send for BufferPtr<T> {}
+unsafe impl<T: ?Sized> Sync for BufferPtr<T> {}
 
 struct State<const N: usize> {
     available: Vec<usize, N>,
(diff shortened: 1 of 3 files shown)


embassy-sync/src/resource_pool.rs
diondokter, line 1: Make sure to add tests & more docs

embassy-sync/src/resource_pool.rs
```diff
@@ -0,0 +1,192 @@
...
+
+#[repr(transparent)]
+#[derive(Debug)]
+struct BufferPtr<T: ?Sized>(*mut T);
+
+unsafe impl<T: ?Sized> Send for BufferPtr<T> {}
+unsafe impl<T: ?Sized> Sync for BufferPtr<T> {}
+
+struct State<const N: usize> {
+    available: Vec<usize, N>,
```
diondokter, line 84: Using some sort of ringbuffer would be a lot more efficient. Now at every pop there's a memmove being done

embassy-sync/src/resource_pool.rs
```diff
@@ -0,0 +1,192 @@
...
+        unsafe { &mut *self.store.buf.0.add(self.index) }
+    }
+}
+
+impl<'guard, 'buffer, M: RawMutex, T, const N: usize> ResourceGuard<'guard, 'buffer, M, T, N> {
+    /// maps the value contained to another, referencing the original value. Does not take "self" to avoid shadowing any functions of the wrapped type.
+    pub fn map<U: ?Sized>(
+        orig: Self,
+        fun: impl FnOnce(&mut T) -> &mut U,
+    ) -> MappedResourceGuard<'guard, 'buffer, M, T, U, N> {
```
diondokter, line 127: This is a weird API to me... When would you use this?
Especially the `&mut U` is weird to me. Where would the U be stored? In the closure? But then you get lifetime issues.

diondokter (commented):
Implementation looks fine to me. Got a couple of questions and notes though.

Whether this PR is actually accepted, well I'm not the endboss :P

embassy-sync/src/resource_pool.rs
martin2250: Is a memmove of a single `usize` much of a concern (*)? I'm only doing push and pop at the end of the `Vec`, so no shifting has to be done. If you're suggesting using a ringbuffer in the sense of only keeping track of head + tail, that would not work since the buffers are not guaranteed to be returned in-order.

* assuming we're talking about the usize here, maybe I'm missing something

embassy-sync/src/resource_pool.rs
martin2250: embassy is already doing this with MutexGuard. It allows you to "narrow-down" a reference, eg. from a slice to only a sub-slice.

The main use I see here would be taking a buffer (eg. `&mut [u8; 1024]`), reading some data into it and then passing on a slice (`&buffer[..n_bytes_read]`)  to the next part of your software. This works because the rust compiler will use fat pointers for pointers to slices (eg *mut/const [u8])

embassy-sync/src/resource_pool.rs
martin2250: yes, absolutely!

embassy-sync/src/resource_pool.rs
diondokter: Oh, right! Yeah I had a ringbuffer use in my head, but you're not rotating through the numbers at all. No it's fine, you're right

embassy-sync/src/resource_pool.rs
diondokter: Hmmm I guess. It's fine to have and the impl is fine. Just doesn't seem that useful to me

Dirbaio:

d3d9:
I have been using this implementation for a while now and I really like the concept.
Would appreciate if this could become part of embassy-sync!

Dirbaio:
If you'd like it to get reviewed, please mark it as ready to review.

martin2250:
I'll add some more docs in the coming days and then hand it in 👍🏻

martin2250:
force-pushed, since I rebased the changes to the current embassy/main.

I've added some docs and basic tests for the drop-return mechanism.

For the review: many parts have been adapted from `Mutex` and `zerocopy_channel::Channel`, you can look there for reference.

I'll mark the PR as ready.

@d3d9 I'm curious what you're using it for. Since I was only using it for buffers, I've implemented a more narrowly-focused version here: https://github.com/martin2250/embedded-buffer-pool/. Maybe that ends up being more useful?

d3d9:
> I'm curious what you're using it for. Since I was only using it for buffers, I've implemented a more narrowly-focused version here: https://github.com/martin2250/embedded-buffer-pool/. Maybe that ends up being more useful?

I'm also using it for a buffer pool; thank you for letting me know

PR #5743, opened 2026-03-27
[MCXA] lpuart: bias computation towards maximal OSR

felipebalbi:
We want to find the lowest possible error between requested baudrate and actual baudrate. In case of a tie, we should bias towards the largest OSR for robustness against noise.

A quick example shows that we can achieve e.g. 1MBaud with the following (osr, sbr) pairs (filtered for 0% error):

(4, 48)
(6, 32)
(8, 24)
(12, 16)
(16, 12)
(24, 8)
(32, 6)

The previous version would return (4, 48) (the first possible solution) while the version in this PR produces (32, 6).


felipebalbi:
bender run

PR #5749, opened 2026-03-30
[MCXA] Initial switch to metapac

diondokter:
Depends on: https://github.com/embassy-rs/nxp-pac/pull/62

The metadata isn't used yet. That's for follow-up PRs

### DIFF
--- a/embassy-mcxa/Cargo.toml
+++ b/embassy-mcxa/Cargo.toml
@@ -45,7 +45,7 @@ embedded-io = "0.7"
 embedded-io-async = { version = "0.7.0" }
 heapless = "0.9"
 maitake-sync = { version = "0.3.0", default-features = false, features = ["critical-section", "no-cache-pad"] }
-nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "4e4b287ef8437311f18f6524c8a5a5c5d7d11236", features = ["rt"] }
+nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "e1800907450711cf70b38db7ac81f58bbc37a918", features = ["rt"] }
 nb = "1.1.0"
 paste = "1.0.15"
 


embassy-mcxa/src/clocks/operator.rs
```diff
@@ -8,24 +8,31 @@ use config::{
...
 use super::types::{Clock, ClockError, Clocks, PoweredClock};
 use crate::chips::{ClockLimits, clock_limits};
 use crate::pac;
-use crate::pac::cmc::vals::CkctrlCkmode;
-use crate::pac::scg::vals::{
+use crate::pac::cmc::CkctrlCkmode;
+#[cfg(feature = "mcxa2xx")]
+use crate::pac::mrcc2xx::Mrcc;
+#[cfg(feature = "mcxa5xx")]
+use crate::pac::mrcc5xx::Mrcc;
```
felipebalbi: it seems like this difference could be handled at the PAC side?

embassy-mcxa/src/clocks/operator.rs
```diff
@@ -8,24 +8,31 @@ use config::{
...
     ActiveCfgBgmode, ActiveCfgCoreldoVddDs, ActiveCfgCoreldoVddLvl, LpCfgBgmode, LpCfgCoreldoVddLvl, Vsm,
 };
-use crate::pac::syscon::vals::{
-    AhbclkdivUnstab, FrohfdivHalt, FrohfdivReset, FrohfdivUnstab, FrolfdivHalt, FrolfdivReset, FrolfdivUnstab,
-    Pll1clkdivHalt, Pll1clkdivReset, Pll1clkdivUnstab,
-};
+#[cfg(feature = "mcxa2xx")]
+use crate::pac::syscon2xx as syscon;
+#[cfg(feature = "mcxa5xx")]
+use crate::pac::syscon5xx as syscon;
```
felipebalbi: likewise, maybe handle at the PAC side? That is, all chips are mutually exclusive anyway, just expose the correct one based on which PAC feature is selected.

embassy-mcxa/src/rtc/mcxa5xx.rs
```diff
@@ -8,7 +8,7 @@ use maitake_sync::WaitCell;
 use crate::clocks::{WakeGuard, with_clocks};
 use crate::interrupt::typelevel::{Handler, Interrupt};
 use crate::pac;
-use crate::pac::rtc::vals::Swr;
+use crate::pac::rtc5xx::Swr;
```
felipebalbi, line 11: Another case that could be handled at the PAC side, maybe?

embassy-mcxa/src/clkout.rs
```diff
@@ -7,12 +7,16 @@
...
 
 use crate::clocks::config::VddLevel;
 pub use crate::clocks::periph_helpers::Div4;
 use crate::clocks::{ClockError, PoweredClock, WakeGuard, with_clocks};
 use crate::gpio::{AnyPin, SealedPin};
-use crate::pac::mrcc::vals::{ClkdivHalt, ClkdivReset, ClkdivUnstab, ClkoutClkselMux as Mux};
+#[cfg(feature = "mcxa2xx")]
+use crate::pac::mrcc2xx as mrcc;
+#[cfg(feature = "mcxa5xx")]
+use crate::pac::mrcc5xx as mrcc;
```
felipebalbi: I'll stop repeating, but it really seems like these should be easy to handle at the PAC.

diondokter: It's not a question whether we can, it's a question whether we should...
And maybe we should yeah

diondokter: It differs a bit per peripheral. For mrcc, yes!
But for rtc? Well they're very different and so it makes sense for them to have different names

diondokter: Alright, you can now override the rust module name in the metapac

embassy-mcxa/Cargo.toml
```diff
@@ -45,7 +45,7 @@ embedded-io = "0.7"
 embedded-io-async = { version = "0.7.0" }
 heapless = "0.9"
 maitake-sync = { version = "0.3.0", default-features = false, features = ["critical-section", "no-cache-pad"] }
-nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "366d44d45f080d6aba0fd27597f17044c028d813", features = ["rt"] }
+nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "4e4b287ef8437311f18f6524c8a5a5c5d7d11236", features = ["rt"] }
```
Wassasin: Should be `e1800907450711cf70b38db7ac81f58bbc37a918` now right?

PR #5752, opened 2026-03-30
stm32: disable mco pin on drop

### LINKED ISSUE #5737: embassy-stm32: Dropping MCO doesn't disable pin
The MCO peripheral routes the internal clock to a GPIO pin.

The MCO doesn't own its pin, but merely uses it in `new()`, so you can override the MCO output by reborrowing the pin

```rust
let mco = Mco::new(r.mco.reborrow(), r.mco_pin.reborrow(), McoSource::HSI48, {
    let mut c = McoConfig::default();
    c.prescaler = McoPrescaler::DIV4; // 12 MHz
    c.speed = Speed::VeryHigh;
    c
});
Flex::new(r.mco_pin.reborrow()).set_as_analog(); // this overrides the MCO functionality, despite mco still being in scope
```

This seems to violate the idea that a peripheral owns its resources while it is in scope.

Further, since the MCO does not own the pin, and has no `Drop` implementation, so dropping an MCO does not disable the peripheral and the pin continues outputting the clock.

Compare this to I2C:

```rust
let i2c = I2c::new(
    r.i2c.reborrow(),
    r.scl.reborrow(),
    r.sda.reborrow(),
    r.dma_tx.reborrow(),
    r.dma_rx.reborrow(),
    Irqs,
    {
        let mut config = i2c::Config::default();
        config.gpio_speed = Speed::Medium;
        config.frequency = Hertz::khz(100);
        config.scl_pullup = false;
        config.scl_pullup = false;
        config
    },
);

Flex::new(r.scl.reborrow()).set_as_analog(); // this does NOT compile
```

xoviat:
fixes #5737

### DIFF
--- a/embassy-stm32/src/rcc/mco.rs
+++ b/embassy-stm32/src/rcc/mco.rs
@@ -1,8 +1,6 @@
-use core::marker::PhantomData;
-
 use embassy_hal_internal::PeripheralType;
 
-use crate::gpio::{AfType, OutputType, Speed};
+use crate::gpio::{AfType, Flex, OutputType, Speed};
 use crate::pac::RCC;
 #[cfg(any(
     rcc_f2,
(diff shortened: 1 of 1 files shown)


BlakeFreer:
thanks!

PR #5772, opened 2026-04-01
Add USB host UAC (audio class) support

AlexCharlton:
Fresh off the back of @leftger's exciting [USB host addition](https://github.com/embassy-rs/embassy/pull/5633), here's support for hosting USB Audio Class devices.

Some notes:
- This is far from a complete UAC host implementation, but it does the straightforward thing of connecting to a device and sending audio data to it.
- This has been tested on two Beringer devices: the Wing and X32, and while I haven't tried to tailor the implementation to these devices specifically, I'm sure there are other devices out there that will fail due to unforeseen quirks. For these tested devices, however, it works well!
- [Here's a usage example](https://github.com/AlexCharlton/mpfs-hal/blob/usbh/examples/src/bin/embassy-usb-host-uac.rs)
- The handler implementation mirrors the `UsbHostHandler` (at least insofar as it contains `try_register`), but does not implement it directly as I'm not so sure the `wait_for_event` method is quite the right fit for this kind of isochronous protocol. I'm sure there's a proper interface that could be created to abstract over all the possible USB host types, but I'm a bit skeptical this would have much value given the thing interacting with a given host handler needs to be tailored to that class anyway.
  - That said, I might suggest taking the `wait_for_event` out of `UsbHostHandler` and putting it into a `trait UsbHostPoller: UsbHostHandler`, so at least `try_register` can be shared across handlers.
-  I snuck in a fix and an enhancement, that I'm happy to pull out into separate PRs if desired:
    - [`embassy-usb-host/src/class/hub.rs`](https://github.com/embassy-rs/embassy/pull/5772/changes#diff-6cd3a47ff9e73ab9430c1ee931b6eef3e0dc3ce4623f0595ca2b6d5e18c7c22dR276): without these `#[derive(Debug)]`s, the crate doesn't compile with non-defmt logging
    - [`embassy-usb-host/src/lib.rs`](https://github.com/embassy-rs/embassy/pull/5772/changes#diff-bad4534467968cee8c7099408d592be28892560ebfd57c68a1540ad9e6580610R40): `EnumerationError::ConfigBufferTooSmall` was added, since otherwise a user has no way of reacting to or knowing why a call to `enumerate` has failed if the buffer they passed is smaller than what the device requires.

### DIFF
--- a/embassy-usb-host/src/class/uac/codes.rs
+++ b/embassy-usb-host/src/class/uac/codes.rs
@@ -395,7 +395,15 @@ pub mod control_selector {
 }
 
 pub mod format_type {
-    use bitflags::bitflags;
+
+    macro_rules! bitflags {
+        ($($tt:tt)*) => {
+            #[cfg(feature = "defmt")]
+            defmt::bitflags! { $($tt)* }
+            #[cfg(not(feature = "defmt"))]
+            bitflags::bitflags! { #[derive(Debug, Clone, PartialEq)] $($tt)* }
+        };
+    }
 
     pub const UNDEFINED: u8 = 0x00;
     pub const I: u8 = 0x01;
@@ -407,7 +415,6 @@ pub mod format_type {
     pub const EXT_III: u8 = 0x83;
 
     bitflags! {
-        #[derive(Debug, Copy, Clone, PartialEq)]
         pub struct Type1: u32 {
             const PCM = 1 << 0;
             const PCM8 = 1 << 1;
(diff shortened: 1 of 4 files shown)


embassy-usb-host/src/class/uac/codes.rs
leftger, line 1: I don't have the knowledge to verify the codes (great job getting this all in here lol). I'm a big fan of the `device-driver` crate to autogenerate stuff like this. Maybe it's outside of the scope of this PR but for maintainability's sake it'd be cool to find some way to save these into a YAML/TOML structure. @diondokter might be able to voice a better opinion.

AlexCharlton: Ah, interesting idea!

leftger (approved):

embassy-usb-host/src/class/uac/codes.rs
leftger: I feel like the added functionality should precede any potential TOMLfication of structs/enums. Great work!

PR #5773, opened 2026-04-02
Enable full-duplex I2S for spi_v2 devices using I2SEXT peripherals

### LINKED ISSUE #4698: Enable "dual" full-duplex I2S for "v2_i2s" devices
Currently, `embassy_stm32` doesn't support full-duplex I2S operation in some cases where the hardware does support it.  For example, the [STM32F405RG datasheet](https://www.st.com/en/microcontrollers-microprocessors/stm32f405rg.html) says that full duplex I2S is supported, but the `I2S::new_full_duplex()` method isn't enabled in `embassy_stm32` because that chip uses the older `v2_i2s` SPI interface.

Those devices implement a different style of full-duplex I2S, leveraging an extra "I2SEXT" peripheral to instantiate two SPI interfaces, one in each direction.  This approach is implemented, somewhat incompletely, in the [`stm32_i2s_v12x` crate](https://docs.rs/stm32_i2s_v12x/latest/stm32_i2s_v12x/).

It seems like the "dual" approach to full-duplex I2S could be implemented within the current API, as a variation on `new_full_duplex`.  Or as a minor variation, where the I2SEXT peripheral is provided explicitly.  Unfortunately, it doesn't look like the I2SEXT peripherals are captured by the `stm32-metapac` crate, so the `stm32-data` repo would have to be updated to generate code for those peripherals.

bifurcation:
Fixes https://github.com/embassy-rs/embassy/issues/4698

There are two classes of changes in this PR: Ones that will be permanent, and ones that are temporary until stm32-metapac gets updated to reflect information about I2SEXT peripherals.

Changes of the latter type are reflected in a few traits at the top of the file. I have mimicked traits such as WsPin and RxDma, which is nice because it preserves a bunch of the design patterns below. It seems like metapac just needs to add these traits and implement them as appropriate for a device, but I don't grok metapac well enough to make this change right now. The I2sRegs trait should probably be absorbed into spi::Instance.

The permanent changes are of the following character:

* We add a method I2S::new_full_duplex() which is feature-gated to spi_v2 (though TBH I'm not sure if this is the only version of SPI that should be covered, or if every spi_v2 platform has I2SEXT peripherals)
* We add a field to struct I2S to hold the register block for the I2SEXT peripheral, if used
* In new_inner, if an I2SEXT peripheral is being used, we initialize it to the same I2S settings as the main SPI peripheral, in the opposite direction, and always in slave mode.
* Instead of referring to a single register block self.spi.info.regs, we distinguish between register blocks used in the tx and rx directions. These are different when an I2SEXT peripheral is in use, and the same otherwise. This does result in some duplication when I2SEXT is not in use; it might be possible to refactor to avoid.

Marking this as draft because of the above-mentioned issues, and because of the unresolved dependency on metapac changes. I have tested this on a real device using `stm32f405rg` using SPI to talk to an `WM8960` and it works reliably, but I have not tested not more generally.

Replaces #4746

### DIFF
--- a/embassy-stm32/src/i2s.rs
+++ b/embassy-stm32/src/i2s.rs
@@ -51,7 +51,7 @@ impl I2sExtRegs for crate::peripherals::SPI3 {
 
 /// Provides the DMA request number for routing a DMA channel to an I2SEXT
 /// peripheral.
-pub trait RxDmaExt<T: Instance>: crate::dma::ChannelInstance {
+pub trait RxDmaExt<T: Instance>: crate::dma::Channel {
     /// Get the DMA request number needed to use this channel with I2SEXT.
     fn request(&self) -> crate::dma::Request;
 
@@ -264,7 +264,7 @@ pub struct Writer<'s, 'd, W: Word>(&'s mut WritableRingBuffer<'d, W>);
 impl<'s, 'd, W: Word> Writer<'s, 'd, W> {
     /// Write data to the I2S ringbuffer.
     /// This appends the data to the buffer and returns immediately. The data will be transmitted in the background.
-    /// If thfre’s no space in the buffer, this waits until there is.
+    /// If thfre's no space in the buffer, this waits until there is.
     pub async fn write(&mut self, data: &[W]) -> Result<(), Error> {
         self.0.write_exact(data).await?;
         Ok(())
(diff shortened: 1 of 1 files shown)


bifurcation:
This PR is also based on an older commit of `embassy-stm32`, so it probably needs a fairly substantial rebase.

xoviat:
@bifurcation hang on a minute, I'm going to rework the traits a bit...

bifurcation:
@xoviat No rush!  My fork is working for me right now, so we can do thus on whatever schedule you like.  I just refreshed the PR with my latest state since I saw your ping.

xoviat:
@bifurcation I just merged #5774, which should have the traits that you need to complete this.

leftger:
@bifurcation I think this PR has potential. I know xoviat said they introduced some traits. If you can update your PR to incorporate those traits I'm sure we can get this merged in.

xoviat:
This functionality is already implemented.

PR #5777, opened 2026-04-02
Add slot API to zerocopy channel

wojciech-graj:
The current zerocopy channel API is suboptimal, because the only way
to enforce we had previously reserved a slot for sending or receiving
when calling (send/receive)_done, was to panic in such a case. With
the new API, this would be impossible.

The proposed new API makes the send/receive methods return a slot. The
sending slot has the send_done method on it instead of sending
automatically when dropped, to prevent sends of partial data.

The important changes are in
embassy-sync/src/zerocopy_channel.rs and
embassy-net-driver-channel/src/lib.rs, while the rest of the files
just had to be edited to use the new API.

### DIFF
--- a/cyw43/src/bluetooth.rs
+++ b/cyw43/src/bluetooth.rs
@@ -164,6 +164,14 @@ pub(crate) fn read_firmware_patch_line(p_btfw_cb: &mut CybtFwCb, hfd: &mut HexFi
     0
 }
 
+async fn bt_toggle_intr(bus: &mut impl Bus) {
+    trace!("bt_toggle_intr");
+    let old_val = bus.bp_read32(HOST_CTRL_REG_ADDR).await;
+    // TODO: do we need to swap endianness on this read?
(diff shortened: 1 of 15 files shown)


Dirbaio:
I like the idea! 2 things:

- It's inconsistent that the sender needs a method call to finish sending, and the receiver automatically finishes the receiving on drop. I'd make both require a method call, drop is a bit too implicit imo.
- It should be possible to make the slots have 1 lifetime instead of 2 by storing a reference to the state, not the whole Sender/Receiver.

wojciech-graj:
> It's inconsistent that the sender needs a method call to finish sending, and the receiver automatically finishes the receiving on drop. I'd make both require a method call, drop is a bit too implicit imo.

Fair enough, neither approach is great, but it makes sense to at least keep it consistent. Done.

> It should be possible to make the slots have 1 lifetime instead of 2 by storing a reference to the state, not the whole Sender/Receiver.

Done, I forgot lifetime coercion is a thing.

PR #5787, opened 2026-04-03
Add wake_at() methods to waker registrations

regicidalplutophage:
I've been making a construct which is a little bit like a timer which we can await before we determined when it should return. Waker had to be registered upon the initial await. One caveat is that once registered it's private so the functionality of scheduling its waking has to be a method.

It might also be preferable just make the waker field public.

### DIFF
--- a/embassy-sync/Cargo.toml
+++ b/embassy-sync/Cargo.toml
@@ -27,14 +27,15 @@ src_base_git = "https://github.com/embassy-rs/embassy/blob/$COMMIT/embassy-sync/
 target = "thumbv7em-none-eabi"
 
 [features]
+schedule-wake = ["dep:embassy-time", "dep:embassy-time-driver"]
 defmt = ["dep:defmt"]
 log = ["dep:log"]
 std = []
 turbowakers = []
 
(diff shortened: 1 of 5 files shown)


embassy-sync/Cargo.toml
```diff
@@ -33,6 +33,9 @@ std = []
 turbowakers = []
 
 [dependencies]
+embassy-time = {version = "0.5.1", path = "../embassy-time"}
```
leftger: You're adding a new dependency to this crate. Is this the right place for this new functionality? Maybe you could gate this feature?

leftger (changes requested):
I think it's a neat feature. My gut tells me you could feature-gate this at a minimum. The fact this crate hasn't needed these dependencies tells me you might need to reconsider where this belongs.

leftger (approved):

Dirbaio:
The use case for this seems a bit niche, and it's something the user can do on their own easily, WakerRegistration is not that complex. Therefore I don't think it's worth making embassy-sync depend on embassy-time.

PR #4484, opened 2025-07-30
[st32f4/adc] first prototype of read_seq function

tommasoclini:
I have written a basic prototype of a `read_seq` async function similar to the one found in the stm32u5 adc driver, it is useful whenever using a `RingBufferedAdc` is too much data, and I've also found it incredibly easy to mess up the channel order, that happens to me whenever I connect with a debugger for example.

I have also added an example called `adc_dma_seq` to the stm32f4 examples, I've tested on an stm32f446re.

I'm new to working on low level hal code so feel free to critique as much as you can.

PR #5789, opened 2026-04-03
embassy-stm32/adc/v2: fix configure_sequence and add async DMA reads

fuchskurt:
embassy-stm32/adc/v2: fix configure_sequence and add async DMA reads

- Fix configure_sequence to correctly distribute channels across
  SQR1/SQR2/SQR3 instead of writing all ranks to SQR3
- Add read() for one-shot async DMA multi-channel reads
- Add into_seq_reader/SeqReader for repeated reads without
  reprogramming the sequence each time

Based on PR #4484 by @tommasoclini


xoviat:
can you chat https://app.element.io/#/room/#embassy-rs:matrix.org ?

PR #5791, opened 2026-04-03
embassy-stm32/adc/v2: fix configure_sequence

fuchskurt:
- Fix configure_sequence to correctly distribute channels across SQR1/SQR2/SQR3 instead of writing all ranks to SQR3


xoviat:
bender run

PR #5797, opened 2026-04-04
support for nxp mimxrt1064

jgemmell124:


i509VCB (approved):
Basically looks the same as RT1062

jgemmell124:
> Basically looks the same as RT1062

RT 1060/62/64 have identical pins, but their memory regions are different

PR #5800, opened 2026-04-04
STM32 HSEM improvements, use HSEM for dual core init sync

boondocklabs:
Adds HSEM functionality tested on H747 which was previously not functional due to interrupts never being enabled.

* Added async `fast_lock()` for 1-step fast locking between cores without using a process id
* Added `blocking_notify()` for simple 1-step inter core notification
* Added `blocking_listen()` to busy wait poll the interrupt flag for an HSEM channel
* Added async `listen()` to wait for a channel interrupt without taking the lock for inter core notifications
* Avoid resetting HSEM in RCC, as it results in race/deadlock when two cores are trying to initialize
* Use HSEM notifications in `init_primary()` and `init_secondary()` in place of `INIT_DONE_FLAG` which avoids d-cache coherency issues. H7 is not hardware cache coherent, and atomics don't behave well even in an MPU configured non-cacheable region (bus faults)

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -396,8 +396,8 @@ pub fn init(config: Config) -> Peripherals {
 mod dual_core {
     use core::cell::UnsafeCell;
     use core::mem::MaybeUninit;
-    use core::sync::atomic::{AtomicUsize, Ordering};
 
+    use hsem::HardwareSemaphoreChannel;
     use rcc::Clocks;
 
     use super::*;
@@ -419,7 +419,6 @@ mod dual_core {
     /// This static must be placed in the same position for both cores. How and where this is done is left to the user.
     #[repr(C)]
     pub struct SharedData {
-        init_flag: AtomicUsize,
         clocks: UnsafeCell<MaybeUninit<Clocks>>,
         config: UnsafeCell<MaybeUninit<SharedConfig>>,
         #[cfg(feature = "low-power")]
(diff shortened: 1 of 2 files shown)


diondokter:
Do all dual core STMs have an HSEM? (And is HSEM implemented for them all?)

boondocklabs:
> Do all dual core STMs have an HSEM? (And is HSEM implemented for them all?)

It looks like only some H7, WB, and WL are dual core, and all have HSEM. Unfortunately I don't have any W parts to test with.

embassy-stm32/src/lib.rs
```diff
@@ -438,24 +435,36 @@ mod dual_core {
...
 
         rcc::set_freqs_ptr(shared_data.clocks.get());
         #[cfg(feature = "low-power")]
         rcc::set_rcc_config_ptr(shared_data.rcc_config.get());
         let p = init_hw(config);
 
-        unsafe { *shared_data.config.get() }.write(config.into());
-
-        shared_data.init_flag.store(INIT_DONE_FLAG, Ordering::SeqCst);
+        let mut hsem = HardwareSemaphoreChannel::<'_, peripherals::HSEM>::new(1);
```
xoviat: Can you also change this to get_hsem?

xoviat (changes requested):
Please fix the HIL failure.

xoviat (approved):

embassy-stm32/src/lib.rs
TechnoPorg, line 472: After this commit, I don't think the documentation about it returning `None` is still accurate. It looks as though there's now no difference between `try_init_secondary` and `init_secondary`.

PR #5803, opened 2026-04-05
stm32: use exhaustive enum for refcounts

wojciech-graj:
This greatly simplifies the refcount-related code. Since
sizeof(Option<#[repr(u8)] enum>) == sizeof(u8), the code is
equivalent, but the compiler can optimize out bounds checks.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -774,7 +774,7 @@ fn main() {
         }
     }
 
-    let mut refcount_idxs = HashMap::new();
+    let mut refcount_idxs = BTreeSet::new();
 
     for p in METADATA.peripherals {
         if !singletons.contains(&p.name.to_string()) {
@@ -825,11 +825,10 @@ fn main() {
 
             let needs_refcount = *rcc_field_count.get(&(en_reg.register, en_reg.field)).unwrap() > 1;
             let refcount_idx = if needs_refcount {
-                let next_refcount_idx = refcount_idxs.len() as u8;
-                let refcount_idx = *refcount_idxs
-                    .entry((en_reg.register, en_reg.field))
-                    .or_insert(next_refcount_idx);
-                quote! { Some(#refcount_idx) }
+                let refcount_idx = format_ident!("{}_{}", en_reg.register, en_reg.field);
+                let quoted = quote! { Some(RefcountIndex::#refcount_idx) };
+                refcount_idxs.insert(refcount_idx);
+                quoted
             } else {
                 quote! { None }
             };
@@ -877,8 +876,17 @@ fn main() {
     g.extend({
         let refcounts_len = refcount_idxs.len();
         let refcount_zeros: TokenStream = refcount_idxs.iter().map(|_| quote! { 0u8, }).collect();
+        let repr = (!refcount_idxs.is_empty()).then(|| quote! { #[repr(u8)] });
+        let refcount_idxs = refcount_idxs.iter();
         quote! {
             pub(crate) static mut REFCOUNTS: [u8; #refcounts_len] = [#refcount_zeros];
+
+            #repr
+            #[allow(non_camel_case_types)]
+            #[derive(Clone, Copy)]
+            pub(crate) enum RefcountIndex {
+                #(#refcount_idxs),*
+            }
         }
     });
 
--- a/embassy-stm32/src/rcc/mod.rs
+++ b/embassy-stm32/src/rcc/mod.rs
@@ -38,6 +38,7 @@ mod _version;
 pub use _version::*;
 use stm32_metapac::RCC;
 
+use crate::_generated::RefcountIndex;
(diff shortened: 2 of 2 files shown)


embassy-stm32/src/rcc/mod.rs
```diff
@@ -270,38 +270,33 @@ impl RccInfo {
     /// Safety:
     /// - `reset_offset_and_bit`, if set, must correspond to valid xxxRST bit
     /// - `enable_offset_and_bit` must correspond to valid xxxEN bit
-    /// - `refcount_idx`, if set, must correspond to valid refcount in `_generated::REFCOUNTS`
     /// - `stop_mode` must be valid
     pub(crate) const unsafe fn new(
         reset_offset_and_bit: Option<(u8, u8)>,
         enable_offset_and_bit: (u8, u8),
-        refcount_idx: Option<u8>,
+        refcount_idx: Option<Refcount>,
```
diondokter: Maybe add a None option to the enum? Then we can ditch the option, which saves an extra byte

embassy-stm32/build.rs
```diff
@@ -877,8 +876,16 @@ fn main() {
...
         let refcounts_len = refcount_idxs.len();
         let refcount_zeros: TokenStream = refcount_idxs.iter().map(|_| quote! { 0u8, }).collect();
+        let refcount_idxs = refcount_idxs.iter();
         quote! {
             pub(crate) static mut REFCOUNTS: [u8; #refcounts_len] = [#refcount_zeros];
+
+            #[repr(u8)]
+            #[allow(non_camel_case_types)]
+            #[derive(Clone, Copy)]
+            pub(crate) enum Refcount {
```
diondokter: Let's maybe make the name `RefcountIndex`. That's more clear I think

diondokter (commented):
Cool, I like it!

embassy-stm32/src/rcc/mod.rs
wojciech-graj: Not necessary: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e294ac169cdbdec30d083aee697d9b82

embassy-stm32/src/rcc/mod.rs
```diff
@@ -270,51 +270,37 @@ impl RccInfo {
...
-            {
-                *refcount += 1;
-                if *refcount > 1 {
-                    return;
-                }
-            } else {
-                panic!("refcount_idx out of bounds: {}", refcount_idx)
+        if let Some(refcount_idx) = self.refcount_idx {
+            let refcount_idx = refcount_idx as usize;
+            let refcount = &mut unsafe { (*core::ptr::addr_of_mut!(crate::_generated::REFCOUNTS))[refcount_idx] };
```
diondokter: While this code is being touched, let's use the new syntax instead of the macro.
I think this should be correct:
```suggestion
            let refcount = &mut unsafe { *&raw mut crate::_generated::REFCOUNTS)[refcount_idx] };
```

embassy-stm32/src/rcc/mod.rs
```diff
@@ -388,21 +374,12 @@ impl RccInfo {
...
-            {
-                *refcount -= 1;
-                if *refcount > 0 {
-                    return;
-                }
-            } else {
-                panic!("refcount_idx out of bounds: {}", refcount_idx)
+        if let Some(refcount_idx) = self.refcount_idx {
+            let refcount_idx = refcount_idx as usize;
+            let refcount = &mut unsafe { (*core::ptr::addr_of_mut!(crate::_generated::REFCOUNTS))[refcount_idx] };
```
diondokter: Same here:
```suggestion
            let refcount = &mut unsafe { *&raw mut crate::_generated::REFCOUNTS)[refcount_idx] };
```

embassy-stm32/src/rcc/mod.rs
diondokter: Ah hmm, that's true!

wojciech-graj:
@diondokter Do you know why the CI is failing? It's only the fdcan tests, and it doesn't seem like this should've been caused by this PR.

diondokter:
Not sure! But it broke on multiple devices. So that hints that it has to do with this PR...
Do you have a board you could run the FDCAN test on?

diondokter:
bender run

diondokter:
Yeah, it's still happening on a retry. And a different PR is running the test just fine, so it must be something to do with this PR!

wojciech-graj:
Should be fixed now. It was quite an insidious order of operations issue.

```diff
-            let refcount = &mut unsafe { (*&raw mut crate::_generated::REFCOUNTS)[refcount_idx] };
+            let refcount = unsafe { &mut (*&raw mut crate::_generated::REFCOUNTS)[refcount_idx] };
```

diondokter:
Let's try again 😁

PR #5805, opened 2026-04-05
Forbid zero length MultiWakerRegistration

wojciech-graj:
A MultiWakerRegistration with zero length is useless, and this lets us get rid of a panic that otherwise doesn't get optimized out.

### DIFF
--- a/embassy-sync/src/waitqueue/multi_waker.rs
+++ b/embassy-sync/src/waitqueue/multi_waker.rs
@@ -35,7 +35,7 @@ impl<const N: usize> MultiWakerRegistration<N> {
             self.wake();
         }
 
-        if self.wakers.push(w.clone()).is_err() {
+        if self.wakers.push(w.clone()).is_err() && N == 0 {
             // This can't happen unless N=0
             // (Either `wakers` wasn't full, or it was in which case `wake()` empied it)
             panic!("tried to push a waker to a zero-length MultiWakerRegistration")


diondokter:
Ok, so the code is correct, but typically embassy tends to avoid unsafe like this.
A lot of people work on the codebase and so the invariants can get broken after an accidental refactor.

I see two solutions:
- Encode the variant in the typesystem somehow (probably in the direction of having a `NonZeroVec`)
- Leave the assert in the new function and then just ignore the error

wojciech-graj:
> Ok, so the code is correct, but typically embassy tends to avoid unsafe like this. A lot of people work on the codebase and so the invariants can get broken after an accidental refactor.

Makes sense. Might be good to document general guidelines like this somewhere.

> Leave the assert in the new function and then just ignore the error

I went with this option, creating a wrapper around the Vec seems overkill.

Dirbaio:
A zero-length MultiWakerRegistration is useless when used directly, but someone might want to write generic code that sometimes needs wakers and sometimes doesn't depending on some condition. Forbidding N=0 at compile time would prevent that.

I think it should be possible to rearrange the code so the panic is optimized out when N != 0 but still allowing N = 0

Dirbaio:
@wojciech-graj what's your motivation for this? 

- Speed
- Removing the panic to make `panic_never` happy
- Removing the panic to reduce code size
  - you might want to use `panic_immediate_abort` for this instead

wojciech-graj:
>  Removing the panic to make `panic_never` happy

This. I'll admit it's a bit ambitious, but so far I've actually managed to get rid of a good deal of embassy's panics that cropped up in my project.

I've re-arranged things, let me know if this looks better.

Dirbaio:
uuuuh good luck with `panic-never`, you'll need it :)

PR #5809, opened 2026-04-06
Support QSPI for Wiznet W6300

### LINKED ISSUE #4662: Add WIZnet W6300 net driver

KingCol13:
Closes #4662 - builds on @lhalf's work.

This is quite messy but I'm making an early draft PR for visibility and potential feedback on the design. There's some parts that should be split out into other PRs.

# Design

- Add `QspiDevice` and `QspiBus` traits in the style of `embedded_hal_async`.
    - The difference with the regular `Spi*` traits is the `WriteSingleLine` operation and no `Transfer` or `TransferInPlace`.
    - These should probably be upstreamed to `embedded_hal_async` if they make sense.
    - This is my first time writing a QSPI driver so I'm not sure whether QSPI is standardised enough for a trait to exist - after a quick search it looked like most QSPI operates like this though.
- Add a QSPI PIO program for `embassy-rp`, implementing `QspiBus`.
- Add a `WiznetSpiBus` trait to `embassy-net-wiznet` - this has a blanket implementation for `embedded_hal_async::spi::SpiDevice` for backwards compatibility.
    - This has the downside of allowing users to pass QSPI interfaces to chips that don't support QSPI.

# Current problems

- QSPI PIO program
    - Likely to break at low frequencies since the operations can interrupt each other - I'm fairly certain I can fix this with interrupts.
    - Changing the program for each operation is quite slow - I think I can fix this by combining the programs into a single program and using `exec_jmp` - `onewire.rs` has an example of using `exec_jmp`.

# Performance

I get these results with a TCP benchmark (which spends 40% of its time in `smoltcp`, mostly doing copies and calculating checksums):

SPI mode:

| SPI Frequency (MHz) | down (KB/s) | up (KB/s) | updown (KB/s) |
| ------------------- | ----------- | --------- | ------------- |
| 15                  | 905         | 1224      | 498           |

invalid chip version >15MHz SPI.

In QSPI mode:

| QSPI Frequency (MHz) | down (KB/s) | up (KB/s) | updown (KB/s) |
| ------------------- | ----------- | --------- | ------------- |
| 25                  | 1310        | 1736      | 875           |
| 15                  | 1257        | 1622      | 809           |

Maybe a better benchmark could use raw sockets and UDP to avoid a lot of the `smoltcp` overhead. Looking at flamegraphs I think I can get about 15% more performance from combining into one PIO program.


KingCol13:
I've fixed the interruption problem with an interrupt but the performance has been badly degraded - I will try to use dummy DMA transfers to synchronise, as is done in the regular SPI PIO program.

lhalf (approved):

embassy-embedded-hal/src/qspi/traits.rs
```diff
@@ -0,0 +1,385 @@
...
+    DelayNs(u32),
+}
+
+/// QSPI device trait.
+///
+/// `QspiDevice` represents ownership over a single QSPI device on a (possibly shared) bus, selected
+/// with a CS (Chip Select) pin.
+///
+/// See the [module-level documentation](self) for important usage information.
+pub trait QspiDevice<Word: Copy + 'static = u8>: ErrorType {
```
bugadani, line 261: IMHO this should probably be called HalfDuplexSpi but I saw the MultiSpi recommendation, too.

bugadani (dismissed):
Let's make some movement here, half-duplex SPI sorely needs a standard-ish trait set :)

embassy-embedded-hal/src/qspi/traits.rs
KingCol13: I've been planning to try to upstream some sort of MultiHalfDuplexSpi trait into embedded-hal but I've just been getting distracted. These were the relevant issues:

- (Multipin serial spi interface trait)[https://github.com/rust-embedded/embedded-hal/issues/597]
- (Half-duplex SPI)[https://github.com/rust-embedded/embedded-hal/issues/144]

From memory the differences were just:
- single line write
- maybe no transfer operation

I'll have a go at this soon given the interest

bugadani: > From memory the differences were just:
>
> * single line write
> * maybe no transfer operation

I think it's a bit more nuanced that this, there are hardware that implement the common memory protocol directly in one transfer (command-address-dummy-data phases, all with separately configurable line widths). That hardware can still implement this version of the traits, but will need to provide a custom version of ExclusiveDevice that collapses multiple operations into one. While it may not be a formal standard, it may be worth optimising for, the rest of the operations can be special cases where there are no command and/or address and/or dummy bits.

Dirbaio:

PR #5815, opened 2026-04-07
multiwaker: use unsafe to optimize checks

xoviat:
No changes to existing behavior, but allows the compiler to optimize bounds checks.

closes #5805.


leftger (approved):

Dirbaio:
I don't think we should use unsafe, it's too dangerous. Is there really no way to rearrange the code (without unsafe) so the compiler does optimize it? The vec is obviously empty at that point and N is known at compile time, so I'm surprised the compiler is not able to optimize it.

PR #5821, opened 2026-04-07
stm32/adc: cleanup enum variants

xoviat:
there's no reason for a separate variant,  given that `adc.read(..)` is just `adc.configured_sequence(..).read(..)`.


embassy-stm32/src/adc/v2.rs
xoviat: I think this should be unconditionally `CONTINUOUS`.

xoviat:
bender run

Issue #5133, opened 2025-12-23
[embassy_sync] Why ThreadModeRawMutex and ThreadModeMutex  is implemented only for "cortex_m" chips.


gobftald:
I think there is mechanism also for RISCV processors (e.g. ESP32xx) to check if they are in ISR (or exception) state/context. Checking the MIE bit in the mstatus state.

PR #5822, opened 2026-04-08
embassy-sync: add ThreadModeRawMutex support for riscv32

cong-or:
#### Fix ThreadModeRawMutex behavior on RISC-V (#5133)

`ThreadModeRawMutex` only works on `cortex_m` right now, so it can’t be used on `riscv32` targets.  
The RISC-V executor is thread-only for now, so `in_thread_mode()` just returns `true`.

If we add an interrupt executor for RISC-V in the future, this will need proper tracking—but that’s a separate PR.

## Testing

- Builds cleanly on `riscv32imac-unknown-none-elf`
- All host tests pass

### DIFF
--- a/embassy-sync/src/blocking_mutex/raw.rs
+++ b/embassy-sync/src/blocking_mutex/raw.rs
@@ -89,7 +89,7 @@ unsafe impl RawMutex for NoopRawMutex {
 
 // ================
 
-#[cfg(any(cortex_m, doc, feature = "std"))]
+#[cfg(any(cortex_m, riscv32, doc, feature = "std"))]
 mod thread_mode {
     use super::*;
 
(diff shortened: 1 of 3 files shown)


embassy-sync/src/blocking_mutex/raw.rs
```diff
@@ -142,10 +142,14 @@ mod thread_mode {
...
         return Some("main") == std::thread::current().name();
 
-        #[cfg(not(feature = "std"))]
+        #[cfg(cortex_m)]
         // ICSR.VECTACTIVE == 0
         return unsafe { (0xE000ED04 as *const u32).read_volatile() } & 0x1FF == 0;
+
+        #[cfg(riscv32)]
+        // No interrupt executor on RISC-V yet, so always in thread mode.
+        return true;
```
bugadani, line 151: That's false, esp-hal/esp-rtos provides an interrupt executor for risc-v chips.

cong-or: @bugadani Would an AtomicBool that the interrupt executor sets on entry/exit work? Or does esp-hal already have a mechanism for tracking interrupt context that this should hook into?

bugadani: There is no standard interrupt controller on RISC-V chips, and embassy-sync shouldn't rely on anything ESP32-specific. I'm not aware of an implementation-independent way to check if code is running in a thread context

cong-or: Wouldn't an AtomicBool set/cleared by the executor itself be implementation-independent? It doesn't need to know anything about the interrupt controller — just that the executor entered/exited interrupt context.

bugadani:
IMHO things that require platform-specific implementation are best bundled with that platform-specific implementation. I don't see a nice way to make this mergeable.

PR #5845, opened 2026-04-12
Fix hub port index returned by HubInterrupt

### LINKED ISSUE #5842: embassy-usb-host: `HubInterrupt::take_port_change` generates wrong port
While making a driver for a OHCI host (USB 1.0) I encountered what appears to be a bug in `embassy-usb-host`.

I was using `HubHandler` to interact with the root hub of a 2-port OHCI host using my driver, which currently has the control channel and the "Status Change Endpoint" interrupt channel implemented.
The interrupt channel said there was a change in the first port with `[0b0000_0010]`, but what followed was a control IN for `get_port_status(3, &mut [u8])`.
There is no port 3... and there is in fact something connected to the first port during tests.

Debugging led me to how `HubInterrupt::take_port_change` generates the port number:
```rust
                (bit + idx * 8 + 1) as u8
```
which produces the value 2 with `[0b0000_0010]`, which is increased to 3 in the control channel.

### Fix

Changing the sign to '-' produces the expected 0, which becomes the intended `get_port_status(1, &mut [u8])`.
```rust
                (bit + idx * 8 - 1) as u8
```

### Note
If the hub status change bit is set `0b1` (first bit) then none of the port events are seen.
Probably either `HubHandler::wait_for_event` needs to handle/discard it or `HubInterrupt` needs to skip it while getting port changes.

flaviojs:
Fixes #5842.

Also adds hub change stuff.
No idea how you want to handle them so I just logged them and left a FIXME in `wait_for_event`.

### DIFF
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -170,11 +170,12 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> UsbHostHandler for HubHandler<H, 
 }
 
 impl<H: UsbHostDriver, const MAX_PORTS: usize> HubHandler<H, MAX_PORTS> {
+    #[allow(dead_code)]
     async fn hub_feature(&mut self, set: bool, feature: HubFeature) -> Result<(), HostError> {
         let setup = SetupPacket {
             request_type: RequestType::OUT | RequestType::TYPE_CLASS | RequestType::RECIPIENT_DEVICE,
             request: if set { SET_FEATURE } else { CLEAR_FEATURE },
-            value: 0,
+            value: feature as u16,
             index: 0,
             length: 0,
         };
@@ -325,6 +326,20 @@ bitflags! {
     struct HubStatusChange: u16 {
         const LOCAL_POWER = 1 << 0;
         const OVERCURRENT = 1 << 1;
+    }
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for HubStatus {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "HubStatus({=u16:b})", self.bits());
+    }
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for HubStatusChange {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "HubStatusChange({=u16:b})", self.bits());
     }
 }
 
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -128,22 +128,26 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> UsbHostHandler for HubHandler<H, 
                     self.device_address, status, change
                 );
 
-                // FIXME clear unknown changes? add an event?
+                if !change.is_empty() {
+                    return Err(HostError::Other("Unhandled hub status change"));
+                }
             }
             while let Some(port) = hub_changes.take_port_change() {
                 trace!("HUB {}: port {} changed, requesting status", self.device_address, port);
 
-                let (status, change) = self.get_port_status(port).await?;
+                let (status, mut change) = self.get_port_status(port).await?;
                 debug!(
                     "HUB {}: port {} status: {:?} change: {:?}",
                     self.device_address, port, status, change
                 );
 
                 if change.contains(PortStatusChange::RESET) {
+                    change.toggle(PortStatusChange::RESET);
                     self.port_feature(false, PortFeature::ChangeReset, port, 0).await?;
                 }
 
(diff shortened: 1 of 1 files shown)


embassy-usb-host/src/class/hub.rs
```diff
@@ -119,10 +119,21 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> UsbHostHandler for HubHandler<H,
...
+            if hub_changes.take_hub_change() {
+                trace!("HUB {}: hub changed, requesting status", self.device_address);
+
+                let (status, change) = self.get_hub_status().await?;
+                debug!(
+                    "HUB {}: hub status: {:?} change: {:?}",
+                    self.device_address, status, change
+                );
+
+                // FIXME clear unknown changes? add an event?
```
leftger: I think you should throw some sort of Error with unknown changes, but that's just my thought. It'll let users know to either raise an issue or add the missing functionality themselves. Others can chime in on this with better ideas.

embassy-usb-host/src/class/hub.rs
```diff
@@ -159,6 +170,35 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> UsbHostHandler for HubHandler<H,
...
+            value: 0,
+            index: 0,
+            length: 4,
+        };
+        let mut buf = [0u16; 2];
+        self.control_channel.control_in(&setup, buf.as_mut_bytes()).await?;
+        Ok((
+            HubStatus::from_bits_truncate(buf[0]),
+            HubStatusChange::from_bits_truncate(buf[1]),
+        ))
```
leftger, line 193: Similar fns. Could they be consolidated?

embassy-usb-host/src/class/hub.rs
```diff
@@ -208,15 +248,27 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> HubHandler<H, MAX_PORTS> {
...
 impl HubInterrupt<'_> {
+    fn take_hub_change(&mut self) -> bool {
+        if let Some(b) = self.0.get_mut(0) {
+            if *b & 1 != 0 {
+                *b &= !1;
+                return true;
+            }
+        }
+        false
+    }
```
leftger, line 258: Adding a comment on what this does would be useful. Also that return true could be consolidated with the false statement at the end by defining a bool and setting the desired result and finally returning the bool regardless. Easier to find issues down the line.

embassy-usb-host/src/class/hub.rs
```diff
@@ -208,15 +248,27 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> HubHandler<H, MAX_PORTS> {
...
-            .find(|(idx, v)| v.trailing_zeros() >= if *idx != 0 { 0 } else { 1 })
+            .find(|(_, v)| v.trailing_zeros() < u8::BITS)
             .map(|(idx, v)| {
                 let bit = v.trailing_zeros() as usize;
+                if idx == 0 && bit == 0 {
+                    panic!("the hub change must be taken before a port change is taken");
+                }
                 *v &= !(1 << bit);
-                (bit + idx * 8 + 1) as u8
+                (bit + idx * 8 - 1) as u8
```
leftger, line 279: Adding a comment why the `-` operation needs to be done would be useful.

leftger (changes requested):
Overall a good PR. I'd appreciate addressing the changes before getting this merged.

embassy-usb-host/src/class/hub.rs
flaviojs: The first port is in idx 0 bit 1.
Whoever made this code decided to start port numbers at 0.
On the other hand the usb spec starts port numbers at 1.

I'm guessing it was programmer convenience/habit that resulted in 0-based numbers.
Using the type `NonZero<u8>` for ports would make it closer to the usb spec (1-based).

leftger: That makes sense. Could you just add this as a comment right above this section of code?

embassy-usb-host/src/class/hub.rs
flaviojs: Different argument type or result type, so no, I don't think so.

embassy-usb-host/src/class/hub.rs
flaviojs: done

embassy-usb-host/src/class/hub.rs
flaviojs: done

embassy-usb-host/src/class/hub.rs
flaviojs: done

flaviojs:
Merged main to revolve the new conflicts.

leftger (approved):

PR #5849, opened 2026-04-13
[MCXA] use metadata

diondokter:


embassy-mcxa/src/spi/mod.rs
```diff
@@ -86,23 +88,45 @@ impl_instance!(0, 1);
...
+#[allow(private_bounds)]
+pub trait Pcs2Pin<Instance>: GpioPin + sealed::SealedSpiPin<Instance> + PeripheralType {
+    fn mux(&self);
+}
+
+/// Peripheral chip select or data pin 3 during parallel data transfers pin trait.
+#[allow(private_bounds)]
+pub trait Pcs3Pin<Instance>: GpioPin + sealed::SealedSpiPin<Instance> + PeripheralType {
     fn mux(&self);
 }
```
felipebalbi, line 131: we're not really using the HW-managed chip select lines

embassy-mcxa/build.rs
```diff
@@ -0,0 +1,391 @@
...
+    for peripheral in METADATA.peripherals {
+        if peripheral.driver_name.is_empty() {
+            continue;
+        }
+
+        let cfg_name = match peripheral.driver_name.split_once("::") {
+            Some((path, _block)) => path,
+            None => peripheral.driver_name,
+        }
+        .replace("/", "_");
```
felipebalbi, line 33: maybe also downcase?

felipebalbi (commented):
Aren't you missing updates to the tests directory?

embassy-mcxa/build.rs
diondokter: Done

embassy-mcxa/src/spi/mod.rs
diondokter: True, but either we add these trait shims that aren't used yet or we need to filter this kind of stuff in the build.rs. I.E. these pins will show up in some place and this is the most correct/complete way to do it.

diondokter:
Tested the mcxa2xx tests and they all run locally.

The work is not done, but this PR is blocking other work. So I propose we merge in a (fully functioning) partial state.

Wassasin (approved):

PR #5860, opened 2026-04-15
embassy-net: Avoid looping if the link is down

bugadani:
When the network device is down, smoltcp may return a request for immediate re-polling (see https://github.com/esp-rs/esp-hal/issues/4973) which, because the link is down, will not do anything useful. This PR only queries for a re-poll request when the link is up, which can prevent the runner task from consuming all CPU time in this case.

### DIFF
--- a/embassy-net/CHANGELOG.md
+++ b/embassy-net/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Avoid looping forever if the network interface is down.
+
 ## 0.9.0 - 2026-03-10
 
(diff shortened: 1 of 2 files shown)


lulf (commented):
Hmm, could you say more why you think this is a correct fix. Wouldn't this just skip polling the link state, not waking whatever is waiting for link up? You say 'may' request immediate repolling, meaning that's not always the case?

bugadani:
> You say 'may' request immediate repolling, meaning that's not always the case?

I don't know if it's always the case, but we have a reproducer where it is.

> Wouldn't this just skip polling the link state, not waking whatever is waiting for link up?

[link_state](https://github.com/embassy-rs/embassy/blob/6b015329e6784d084c1b589ba31af7a7dee13a7e/embassy-net-driver/src/lib.rs#L69-L73) must wake the task if the link state changes.

PR #5861, opened 2026-04-15
STM32H7RS HSLV and compensation cells

TomDeRybel:
Added HSLV support, as well as compensation cell support with their errata. 

The XSPI flash example was updated to reflect the API changes, and how to use the errata work-around functions. With the changes, the flash can now work properly at its maximum speed of 200 MHz (in octospi).

HSLV and the compensation cells were also tested on GPIO pins, at 1V8, using MCO2, with an oscilloscope. The waveform rise/fall times are now on the order of 1.2 ns, with good symmetry. (Using VeryFast pin mode.)

Two things require attention:
- I don't know if any critical sections are required in this code, this is not a subject I grasp enough to decide on.
- Is there a better way to implement the two busy loop time-outs? The code is in the RCC set-up, so I don't know if a proper delay is already available?

### DIFF
--- a/embassy-stm32/src/rcc/h.rs
+++ b/embassy-stm32/src/rcc/h.rs
@@ -5,6 +5,8 @@ use stm32_metapac::rcc::vals::Xspisel;
 
 use crate::pac;
 #[cfg(stm32h7rs)]
+use crate::pac::flash::regs::Optkeyr;
+#[cfg(stm32h7rs)]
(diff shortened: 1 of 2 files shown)


xoviat:
@TomDeRybel can you run ./fmtall.sh?

TomDeRybel:
Just ran the ./fmtall.sh script and it fixed what ailed.

PR #5864, opened 2026-04-15
Refactor stm32u5 clock initialization in `rcc.init` function for safer configuration

RecursiveError:
The current RCC/U5 implementation has several issues related to clock and power configuration sequencing. This PR addresses them with the following changes:

 - Early VOS modification:
VOS was being changed at the very beginning of the configuration process. It is now updated only after a safe clock source has been selected.

 - MSIS modified before SYSCLK inital switch:
MSIS was modified before switching the system clock, which can lead to issues when the current clock depends on MSIS (e.g., through PLL). Fix: MSIS is now updated only after a safe clock configuration is established.

- MSIS used as default clock source:
MSIS was used as the initial clock source, which is unsafe since it can run up to 48 MHz and requires prior VOS validation.
Fix: the initial clock source is now always HSI, as it is a stable and universally safe source.

- Late HPRE validation:
The HPRE prescaler limits were checked only after being applied, and only after the system clock switch.
Fix: prescalers are now applied and validated before switching SYSCLK, ensuring that even in case of failure, the system remains in a safe clock configuration.

### DIFF
--- a/embassy-stm32/src/rcc/u5.rs
+++ b/embassy-stm32/src/rcc/u5.rs
@@ -178,9 +178,32 @@ impl Default for Config {
 }
 
 pub(crate) unsafe fn init(config: Config) {
-    // Set the requested power mode
-    PWR.vosr().modify(|w| w.set_vos(config.voltage_range));
+    // Configure the clock to a safe default state before starting configuration:
+    
+    // 1 - Set power mode to Range1
+    PWR.vosr().modify(|w| w.set_vos(VoltageScale::Range1));
     while !PWR.vosr().read().vosrdy() {}
+
+    //2 - set flash WS to 4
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rcc/u5.rs
```diff
@@ -318,12 +341,7 @@ pub(crate) unsafe fn init(config: Config) {
         }
     }
 
-    let hsi = config.hsi.then(|| {
-        RCC.cr().modify(|w| w.set_hsion(true));
-        while !RCC.cr().read().hsirdy() {}
-
-        HSI_FREQ
-    });
+    let hsi = config.hsi.then(|| HSI_FREQ);
```
xoviat, line 344: Should hsi be disabled after clock configuration is done?

PR #5868, opened 2026-04-16
Address GPDMA race condition in ringbuffered mode

everdrone:
`dma_sync` called `reset_complete_count` and `get_remaining_transfers` as two separate steps and the DMA ISR could fire between them. This would cause `DmaUnsynced` errors to be triggered.

This PR switches from ping-pong (2-LLI) to single LLI circular mode (in `linked_list.rs`).
`ReadableRingBuffer` and `WritableRingBuffer` now use `Table<1>` instead of `Table<2> + new_ping_pong`.

### DIFF
--- a/embassy-stm32/src/dma/gpdma/ringbuffered.rs
+++ b/embassy-stm32/src/dma/gpdma/ringbuffered.rs
@@ -35,8 +35,7 @@ impl<'a> DmaCtrlImpl<'a> {
     }
 
     /// Compute remaining transfers from hardware and LLI state.
-    /// Must be called with interrupts disabled (inside a critical section).
-    fn compute_remaining(&self) -> usize {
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/dma/gpdma/ringbuffered.rs
```diff
@@ -13,36 +13,71 @@ use crate::dma::word::Word;
...
+        Self {
+            channel,
+            cached_remaining: 0,
+        }
+    }
 
-        let lli_count = state.lli_state.count.load(Ordering::Acquire);
+    /// Compute remaining transfers from hardware and LLI state.
+    /// Must be called with interrupts disabled (inside a critical section).
+    fn compute_remaining(&self) -> usize {
```
xoviat: If this must be called in a critical section, then it should accept cs: CriticalSection as a parameter.

PR #5880, opened 2026-04-17
Prepare embassy-usb-driver, usb-host and synopsys releases

### LINKED ISSUE #5881: blockers for usb host release
- [ ] https://github.com/embassy-rs/embassy/blob/b36e742c364a596cf703a64eb3b4642fb34417bf/embassy-usb-driver/src/host.rs#L338 is also a rather strange API and probably shouldn't be in the first release. What's the point of such an API if it can fail?
- [ ] https://github.com/embassy-rs/embassy/blob/b36e742c364a596cf703a64eb3b4642fb34417bf/embassy-usb-driver/src/host.rs#L353 this should only be available if the channel is the control type right?
- [ ] Last bit of feedback is this trait in general https://github.com/embassy-rs/embassy/blob/b36e742c364a596cf703a64eb3b4642fb34417bf/embassy-usb-driver/src/host.rs#L322 is missing documentation about NAK handling and cancellation. Once you ship this, we risk ending up with a mess of implementation differences
- [ ] Also, poll frequency is missing for interrupt transfers, but I suppose that's just a nice to have. I'm not going to argue for that haha

bugadani:
Closes #5881 - because merging this will count as accepting that a first release is no longer blocked :)


leftger:
Thank you so much @bugadani for going through this.

bugadani:
dates need fixing before merging

PR #5882, opened 2026-04-17
Rework SetupPacket - move from driver to embassy-usb-host

bugadani:
RequestType was implemented using `bitflags` which is not the right tool for the job for defining bitfields. This PR removes the dependency and hand-implements the type with proper one-of-many semantics on RequestType fields.


embassy-usb-driver/src/host.rs
```diff
@@ -27,85 +27,163 @@ pub enum ChannelError {
...
+    Device = 0,
+    /// The request is intended for an interface.
+    Interface = 1,
+    /// The request is intended for an endpoint.
+    Endpoint = 2,
+    /// The recipient of the request is unspecified.
+    Other = 3,
+    /// Any reserved recipient value (4..=31).
+    Reserved = 4,
 }
```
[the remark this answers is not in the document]
bugadani: IDK if we should add 20-some reserved variants just for this

embassy-usb-driver/src/host.rs
```diff
@@ -27,85 +27,163 @@ pub enum ChannelError {
...
+    /// Called `wLength` in USB spec (Table 9-2).
     pub length: u16,
 }
 
 impl SetupPacket {
-    /// Get a reference to the underlying bytes of the setup packet.
-    pub fn as_bytes(&self) -> &[u8] {
-        // Safe because we know that the size of SetupPacket is 8 bytes.
-        unsafe { core::slice::from_raw_parts(self as *const _ as *const u8, core::mem::size_of::<Self>()) }
+    /// Serialize this SETUP packet to its 8-byte wire format.
```
Dirbaio: do you think it's possible to move SetupPacket out of embassy-usb-driver? ie make the driver work with just bytes.

The usb-host crate can have the nicer public api with parsing/unparsing.

Less api surface means less potential for breaking changes.

bugadani: Sure why not

bugadani: I guess it's a bit weird that driver implementations need to bit-fiddle with the setup packet bytes, though 😅

bugadani: What if we did something in the middle? Keep SetupPacket in the driver, but let the trait method take `&[u8; 8]`? That way implementations could use SetupPacket to wrap/parse what they need out of the bytes, but it would be a utility instead of an integral part of the API.

embassy-usb-host/src/class/uac/mod.rs
```diff
@@ -366,7 +367,7 @@ impl<H: UsbHostDriver> UacHandler<H> {
 
         // Now get the full string with the correct length
         let packet = SetupPacket {
-            request_type: RequestType::IN | RequestType::TYPE_STANDARD | RequestType::RECIPIENT_DEVICE,
+            request_type: RequestType::device_to_host(ControlType::Standard, Recipient::Device),
```
Dominaezzz: It's a bit odd to see `device_to_host` and also `Recipient::Device` in the same line. Makes me wonder who the real recipient is. Though I'm not sure how to improve this since `in` is a keyword.

bugadani: TBH I find the names better than "request_in/request_out", although from a host perspective in/out is less bad than from a device's perspective :) I can rename these into `RequestType::r#in/out` if that's better (or `request_in/request_out` though a bit redundant), but I'm not sure if it is better.

Maybe `control_in/control_out` could work?

Dominaezzz: Oh actually, what about `RequestType::new(Direction::In, ControlType::Standard, Recipient::Device)` ?

bugadani: What if we drop the constructor completely and construct the struct directly? It's a bit wordy, but there's no confusion about parameter positions.

embassy-usb-driver/src/host.rs
Dominaezzz: The `SetupPacket` would still be part of the `embassy-usb-driver`'s API surface even as a utility right? The breakage potential is still the same I think.

I do wonder if a `embassy-usb-util`/`embassy-usb-common` library that both `embassy-usb` and `embassy-usb-host` could use might make sense here. A whole library for just one struct feels silly though :sweat_smile: but it would solve the API surface problem and the bit-fiddling problem.

embassy-usb-host/src/class/uac/mod.rs
Dominaezzz: Works!

xoviat:
bender run

embassy-usb-driver/src/host.rs
bugadani: It's not like setup packets will suddenly change anyway, but it'd be possible to add v2, v3, ... versions of utilities until a breaking release could clear out the accumulated garbage. In the end it's not a big deal.

PR #5891, opened 2026-04-18
Fix endianness of HubStatus, HubStatusChange, PortStatus, PortStatushange

flaviojs:
According to the USB spec, all the data is little endian.

In big endian systems, like mine, the wrong value was being interpreted.

There might be other places with endian issues, but I don't see code to test/check/handle endianness automatically so I'm only fixing what I detected.

### DIFF
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -10,7 +10,7 @@ use bitflags::bitflags;
 use embassy_time::Timer;
 use embassy_usb_driver::host::{HostError, RequestType, SetupPacket, UsbChannel, UsbHostDriver, channel};
 use embassy_usb_driver::{Direction, EndpointInfo, EndpointType, Speed};
-use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+use zerocopy::{FromBytes, Immutable, KnownLayout};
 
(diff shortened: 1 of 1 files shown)


embassy-usb-host/src/class/hub.rs
```diff
@@ -188,8 +188,8 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> HubHandler<H, MAX_PORTS> {
         let mut buf = [0u16; 2];
         self.control_channel.control_in(&setup, buf.as_mut_bytes()).await?;
         Ok((
-            HubStatus::from_bits_truncate(buf[0]),
-            HubStatusChange::from_bits_truncate(buf[1]),
+            HubStatus::from_bits_truncate(u16::from_le(buf[0])),
+            HubStatusChange::from_bits_truncate(u16::from_le(buf[1])),
         ))
```
bugadani: This is conceptually smelly to me. The real problem is that from the point of view of this code, `buf.as_mut_bytes()` flips byte order - i.e. it's reading little-endian bytes into big-endian u16 variables, so to me it doesn't seem correct to try to read into `[u16; 2]` directly. A better solution would be to get rid of `as_mut_bytes`, read into `[u8; 4]` and use `u16::from_le_bytes` to convert.

As for other places where endian issues may exist, perhaps getting rid of zerocopy altogether would help.

flaviojs: done

PR #5901, opened 2026-04-19
stm32/hsem: fix duplicate Instance impl after GTZC NVIC merging

leftger:
## Summary
- Fixes CI build failure introduced by stm32-data#775 (GTZC sub-block decomposition with NVIC merging)
- The NVIC merge exposes both `HSEM` (GLOBAL) and `HSEM_S` interrupt entries for WBA chips, causing `foreach_interrupt!` to generate conflicting `impl Instance` blocks (E0119)
- Filters the macro by `GLOBAL` signal, matching the pattern used by other drivers (rng, cryp, hash, tsc, etc.)
- Includes the PAC update from #5900

### DIFF
--- a/embassy-stm32/Cargo.toml
+++ b/embassy-stm32/Cargo.toml
@@ -209,11 +209,11 @@ aligned = "0.4.3"
 heapless = "0.9.1"
 
 # stm32-metapac = { version = "21" }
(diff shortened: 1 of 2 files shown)


leftger:
@xoviat this fixes your PR.

PR #5905, opened 2026-04-20
[MCXA] Dac driver

Guacimo:
# Dac driver
This PR adds a basic DAC driver for the mcxa dac. 
It currently only supports data mode and does not yet work with the mcxa5xx though I do want to add that to this PR.

Right now this PR does not work at all because the new metapac does not contain the DAC peripheral.

## Testing
A HIL test has been added to test the DAC driver. The Uart test has been rewired in order to free up pin `P2_2` which is the only pin the DAC can use.  The test covers five voltages and checks if the output does not deviate more than 0.2v using an ADC.


xoviat:
bender run

Dirbaio:

Guacimo:
Hi! The pull request has been replaced by https://github.com/embassy-rs/embassy/pull/5982.

PR #5907, opened 2026-04-20
usb-host: TimeoutConfig tweaks

bugadani:
This addresses the respective concern in #5881 and makes the struct more useable, and extendable without introducing a breaking change.

There is some argument for introducing a third, status timeout, but it doesn't seem too important.

### DIFF
--- a/embassy-usb-driver/src/host.rs
+++ b/embassy-usb-driver/src/host.rs
@@ -278,19 +278,23 @@ pub mod pipe {
     impl IsOut for InOut {}
 }
 
-/// Specify the timeout of a pipe
+/// Timeouts applied to a control pipe's NAK-retry behaviour.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


embassy-usb-driver/src/host.rs
```diff
@@ -278,19 +278,23 @@ pub mod pipe {
     impl IsOut for InOut {}
 }
 
-/// Specify the timeout of a pipe
+/// Timeouts applied to a control pipe's NAK-retry behaviour.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
```
bugadani, line 284: This is inconsistent with SplitInfo, which doesn't implement Default, therefore can't use this style... But builders are annoying if they aren't needed, so TimeoutConfig shouldn't use the pattern, probably :/

leftger (approved):

PR #5908, opened 2026-04-20
[MCXA] Bump nxp-pac and use PascalCase for all enum variants

Wassasin:
We updated chiptool to use PascalCase for all enum variants. [nxp-pac has been updated](https://github.com/embassy-rs/nxp-pac/pull/77) to have a more clear pipeline from SVD to generated code, and uses this new chiptool version.

Hence we update the HAL to use these new PascalCase variants.

### DIFF
--- a/embassy-mcxa/src/spi/controller.rs
+++ b/embassy-mcxa/src/spi/controller.rs
@@ -1165,22 +1165,22 @@ impl<'d> AsyncEngine for Spi<'d, Dma<'d>> {
 /// Baud = src_hz / (prescaler.divisor() * (SCKDIV + 2))
 pub(super) fn compute_baud_params(src_hz: u32, baud_hz: u32) -> (Prescale, u8) {
     if baud_hz == 0 {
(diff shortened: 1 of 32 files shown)


embassy-mcxa/src/spi/controller.rs
```diff
@@ -1165,22 +1165,22 @@ impl<'d> AsyncEngine for Spi<'d, Dma<'d>> {
...
     let prescalers = [
-        Prescale::DIVIDEBY1,
-        Prescale::DIVIDEBY2,
-        Prescale::DIVIDEBY4,
-        Prescale::DIVIDEBY8,
-        Prescale::DIVIDEBY16,
-        Prescale::DIVIDEBY32,
-        Prescale::DIVIDEBY64,
-        Prescale::DIVIDEBY128,
+        Prescale::Divideby1,
```
diondokter: `Divideby1` should be `DivideBy1`

diondokter (changes requested):

PR #5909, opened 2026-04-20
USB host: remove retarget_pipe

bugadani:
`enumerate_port` now takes a `&mut usb_driver` because it needs to allocate an address.

cc #5881


bugadani:
@flaviojs as you are the closest to USB hubs, can you try and review or test these changes to make sure I don't break anything? Alternatively, what example should we add to make sure we don't break your use cases?

leftger:
I should get another nucleo board to test this on. If flaviojs hasn't gotten to it by the end of the week I can test it out myself.

bugadani:
I can probably set up a test case myself, there is some justification for that in my setup :)

bugadani:
image

ESP32-S3 (host) -> Hub (1) -> ESP32-S3's USB Serial/JTAG printing bootloader messages.

flaviojs:
> @flaviojs as you are the closest to USB hubs, can you try and review or test these changes to make sure I don't break anything? Alternatively, what example should we add to make sure we don't break your use cases?

Uff, I don't have a working OHCI driver yet, I'm still making it, so I'm probably not the right person to check this out, but I'll have a look.

embassy-usb-host/src/class/hub.rs
```diff
@@ -214,23 +212,49 @@ impl<H: UsbHostDriver, const MAX_PORTS: usize> HubHandler<H, MAX_PORTS> {
...
         &mut self,
+        bus: &mut UsbHost<H>,
+        config_buffer: &mut [u8],
         port: u8,
         speed: Speed,
-        new_device_address: u8,
-    ) -> Result<EnumerationInfo, HostError> {
+    ) -> Result<(EnumerationInfo, usize), EnumerationError> {
         self.port_feature(true, PortFeature::Reset, port, 0).await?;
+        // USB 2.0 §7.1.7.5: TDRSTR ≥ 10 ms. Match the 50 ms margin used in similar drivers.
```
Dominaezzz, line 244: Probably out of scope but rather than hard code a duration here, this could just wait for the port change event on the interrupt pipe.

bugadani: I'm happy to review a PR that implements this ;)

flaviojs:
### UsbRoute
An additional use case to consider might be EHCI with one or more companion OHCI/UHCI.

An EHCI host is supposed to have one or more companion host OHCI/UHCI.
Each port can be owned by the EHCI or owned by a companion OHCI/UHCI.
A driver (that is aware of them) can choose to redirect full/low speed devices to the companion OHCI/UHCI that handles that port, or if the driver detects a high speed device in a companion, it can choose to redirect the high speed device to EHCI.
Not sure what the benefits of doing so nor the restrictions of not doing so (I'll cross that bridge after OHCI).

linux mostly ignores this relationship and treats them independently (assuming the other exists).
Ideally implementers should be able to decide if they want to kick the device between EHCI and OHCI/UHCI.
Not sure if it's possible right now (haven't used embassy enough to judge).

As a library user, I would also find value in being able to get a UsbRoute from the various handlers, but I don't have a specific use case right now.

### get_address
I was wondering how I was gonna get the address for `enumerate_port` with a private `get_address`, glad that's settled.

It's error prone to alloc an address and then manually free.
It might be a good idea to have an `UnclaimedAddress` helper or similar that gives back the address on drop if the address has not been claimed (claiming consumes the helper).
If that cannot be implemented, it could panic on drop if the address has not been claimed or freed (freeing consumes the helper).

### enumerate_port
When `reset_port` was replaced with `enumerate_port` the useful-looking help text was discarded, it's a good idea to copy the parts that apply to `enumerate_port`.

I can't comment on the inner workings of this function because I'm just starting to handle this part of my driver.

bugadani:
> An additional use case to consider might be EHCI with one or more companion OHCI/UHCI.

I believe this has no embassy-usb-driver-related implications.

> It's error prone to alloc an address and then manually free.

It is, but again this has no embassy-usb-driver impications.

> When reset_port was replaced with enumerate_port the useful-looking help text was discarded

I've re-added it.

leftger:
@bugadani do you want me to merge this at this point? Seems like almost everything has been addressed.

bugadani:
Please give me until midnight CEST to think it through.

leftger:
Take your time. Let me know when you think it's in a good spot to merge.

bugadani:
So the reservation I had was related to repurposing pipes - but it's not really relevant here as the removed API wouldn't have worked in general. This is good to go.

PR #5913, opened 2026-04-20
usb-host: more completeness-related changes

bugadani:
- enums are non_exhaustive where they are expected to grow
- added `reset_data_toggle`
- added DeviceEvent::Overcurrent, handled in synopsys-otg
- updated embassy-rp to respect documented behaviour in driver trait impl

Claude wants to implement supend/resume/remote wakeup but I think it can be a later non-breaking addition using a subtrait (`SuspendResume: UsbHostDriver`) in case not every host IP supports it. It is also more of a low-power concern than a strictly required feature. Other missing features can also be implemented this way, whether or not they really make sense as an optional capability.

Claude also has some problems around missing queue-based isochronous APIs but I don't know if any changes are needed here, considering we already have an UAC class driver. Claude is wrong in its reasoning claiming that adding iso-specific methods to the UsbPipe would be an additive change, but oh well, one can't expect everything to be correct.

### DIFF
--- a/embassy-rp/src/usb/host.rs
+++ b/embassy-rp/src/usb/host.rs
@@ -705,6 +705,10 @@ impl<'d, T: Instance, E: pipe::Type, D: pipe::Direction> UsbPipe<E, D> for Chann
     fn set_timeout(&mut self, _: TimeoutConfig) {
         // Not yet implemented for RP2040.
     }
+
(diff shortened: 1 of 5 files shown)


leftger (approved):
Thank you for going over these changes. Much better handling TBH.

PR #5914, opened 2026-04-20
cyw: add error type

xoviat:
allows better error handling with context.


xoviat:
confirmed on hardware

PR #5916, opened 2026-04-21
fix: use device ID as mac addr ble

Tavish-Srivastava:


leftger (approved):
Thanks for checking on this!

PR #5918, opened 2026-04-21
feat(stm32/rcc): implement CPU reboot reason read

jack-greenberg:
This only works for STM32H74x/75x (chips covered under RM0433). Let me know if I am using the features incorrectly!

For testing, I am not sure the best way to make a HitL test because it involves resetting the CPU, which would cause the debug probe to complain... Open to other ideas as well!

Reference:
Screenshot from Reference Manual 0433 covering the logic for determining reboot reason

Link to stm32h7xx-hal's implementation, which is similar: https://docs.rs/stm32h7xx-hal/latest/src/stm32h7xx_hal/rcc/reset_reason.rs.html


jack-greenberg:
Gauh, I need to figure out what the register name is actually supposed to be--seems to be different between chips in the h7 family :( This is a tomorrow problem...

PR #5921, opened 2026-04-21
usb-host: ensure classes don't outlive the host

bugadani:
Branch is based on #5909 

cc https://github.com/embassy-rs/embassy/issues/5881#issuecomment-4286647731


Dirbaio (approved):
lifetimes are nice
I like lifetimes
yay lifetimes

leftger:
Let me know when you're OK with merging this. Appreciate the lifetime additions!

bugadani:
This is ready enough, there is work to be done on the synopsys driver but I think that's an invariant :)

PR #5927, opened 2026-04-21
rp: i2c: Add support 10-bit addressing

felipebalbi:


embassy-rp/src/i2c.rs
```diff
@@ -724,6 +727,36 @@ impl<'d, T: Instance, M: Mode> embedded_hal_1::i2c::I2c for I2c<'d, T, M> {
     }
 }
 
+impl<'d, T: Instance, M: Mode> embedded_hal_1::i2c::I2c<embedded_hal_1::i2c::TenBitAddress> for I2c<'d, T, M> {
```
Dirbaio: this impl can probably be generic for the two addr types like the async one below.

felipebalbi: I'll fix that in a second.

felipebalbi: I suppose it's okay now.

PR #5936, opened 2026-04-22
Add FDCAN missing Transceiver Delay Compensation setup

### LINKED ISSUE #5273: Transceiver Delay Compensation configuration feature is missing in CANFD for STM32
Setting Transceiver Delay Compensation is necessary to use CAN FD with higher bitrates.  But the feature to calculate the params and to set them up is missing in `embassy_stm32::can`.  Is there any way to set the registers directly?

Also, there's a flag `transceiver_delay_compensation` in `embassy_stm32::can::config::DataBitTiming` but it's not in use. The feature seems to be a TODO item.  Is anyone working on it currently?  If no one is working, I'd be happy to add the enhancement to the CAN module.

naokiiwakami:
closes #5273.

The fix was tested using STM32C092KCT and STM32G0B1CET.

### DIFF
--- a/embassy-stm32/src/can/fdcan.rs
+++ b/embassy-stm32/src/can/fdcan.rs
@@ -246,19 +246,27 @@ impl<'d> CanConfigurator<'d> {
         self.config = self.config.set_nominal_bit_timing(nbtr);
     }
 
-    /// Configures the bit timings for VBR data calculated from supplied bitrate. This also sets confit to allow can FD and VBR
+    /// Configures the bit timings for VBR data calculated from supplied bitrate. This also sets config to allow can FD and VBR
     pub fn set_fd_data_bitrate(&mut self, bitrate: u32, transceiver_delay_compensation: bool) {
         let bit_timing = util::calc_can_timings(self.properties.kernel_input_clock(), bitrate).unwrap();
-        // Note, used existing calcluation for normal(non-VBR) bitrate, appears to work for 250k/1M
-        let nbtr = crate::can::fd::config::DataBitTiming {
-            transceiver_delay_compensation,
+        // Note, used existing calculation for normal(non-VBR) bitrate, appears to work for 250k/1M
+        let tdc_offset = if transceiver_delay_compensation {
+            // sets at the end of tseg1
+            (1 + u8::from(bit_timing.seg1)) * u16::from(bit_timing.prescaler) as u8
+        } else {
+            0
+        };
+        let dbtr = crate::can::fd::config::DataBitTiming {
             sync_jump_width: bit_timing.sync_jump_width,
             prescaler: bit_timing.prescaler,
             seg1: bit_timing.seg1,
             seg2: bit_timing.seg2,
+            transceiver_delay_compensation,
+            tdc_offset,
+            tdc_filter_window_length: 0,
         };
         self.config.frame_transmit = FrameTransmissionConfig::AllowFdCanAndBRS;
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/can/fdcan.rs
```diff
@@ -246,19 +246,27 @@ impl<'d> CanConfigurator<'d> {
...
+    /// Configures the bit timings for VBR data calculated from supplied bitrate. This also sets config to allow can FD and VBR
     pub fn set_fd_data_bitrate(&mut self, bitrate: u32, transceiver_delay_compensation: bool) {
         let bit_timing = util::calc_can_timings(self.properties.kernel_input_clock(), bitrate).unwrap();
-        // Note, used existing calcluation for normal(non-VBR) bitrate, appears to work for 250k/1M
-        let nbtr = crate::can::fd::config::DataBitTiming {
-            transceiver_delay_compensation,
+        // Note, used existing calculation for normal(non-VBR) bitrate, appears to work for 250k/1M
+        let tdc_offset = if transceiver_delay_compensation {
+            // sets at the end of tseg1
+            (1 + u8::from(bit_timing.seg1)) * u16::from(bit_timing.prescaler) as u8
```
anlif: Can this overflow?
prescaler is in [1, 1024] and
seg1 is in [1, 16] (looking at `util::calc_can_timings`).

Overflows if (1+seg1) * prescaler > 255.

There's already an unwrap at calc_can_timings, in this case it might be worth doing a checked addition and multiplication if the overflow could cause issues.

xoviat:
bender run

nathanaelg:
> closes #5273.
> 
> The fix was tested using STM32C092KCT and STM32G0B1CET.

I'm struggling to get bit rate switching working at all on STM32C092GC, do you have a minimal example?

I'm trying to do something like:

```rust
let mut can = can::CanConfigurator::new(p.FDCAN1, p.PA11, p.PA12, CanIrqs);
can.properties().set_extended_filter(
    can::filter::ExtendedFilterSlot::_0,
    can::filter::ExtendedFilter::accept_all_into_fifo1(),
);
can.set_bitrate(1_000_000);
can.set_fd_data_bitrate(2_000_000, true);
let (can_tx, can_rx, _can_prop) = can.start(can::OperatingMode::NormalOperationMode).split();
```

but it seems to just do everything at the 1Mbit rate.

naokiiwakami:
I'm using STM32C092KCT in this project https://github.com/naokiiwakami/can-monitor/tree/main
The module monitors a CANFD bus via a serial interface. It also can send test frames with and without BRS.
See around line 247 in main.rs for the configuration and function process_command() for TX methods.

The configuration in your code looks fine to me.  So the issue may be in the part to send frames.  You need to send a FD frame with BRS enabled explicitly in order to switch rate.

I also vaguely remember that a node sends 1Mbps data even if it's configured to use BRS if other nodes do not support FD CAN.  I think it's from the FD CAN spec, but  I may be wrong.  You may want to check that if everything in the code looks good.

Hope it helps.

nathanaelg:
Thanks, that was a good place to start - does this look right to you? https://github.com/embassy-rs/embassy/blob/29a92acfecaeee33f5fd89982d248166d0b56820/embassy-stm32/src/can/frame.rs#L394 Seems like calling `FdFrame::new_extended`/`FdFrame::new_standard`/`FdFrame::new_remote` you end up with `Header::new` instead of `Header::new_fd`? And that doesn't have a `brs` flag passed in...

naokiiwakami:
right

PR #5937, opened 2026-04-22
stm32/low-power: update L0 config

fe1es:
Set `LPSDSR` to `LowPowerMode` to reduce stop-mode current further.

Remove the stm32l0 cfg gate in `exit_stop` altogether. The previous code re-initialized RCC on every RTC wake from stop mode, which is unnecessary. I guess the intent was to reconfigure the clocks, but the path was incomplete anyway; it ran only on RTC wakes, not EXTI. Users can call `rcc::reinit` to reconfigure clocks explicitly if needed.

Tested on `stm32l072`

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -170,10 +170,10 @@ mod platform {
 
         #[cfg(stm32l0)]
         {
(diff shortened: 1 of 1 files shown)


xoviat:
#5416

PR #5945, opened 2026-04-23
Add CP210x host driver

bugadani:
Claude-assisted code, based on the relevant application note and the existing embassy-usb-host code. Tested only by reading data from a CP2102N device. A slight wrinke with this code is that the config descriptor needs to be kept around to open ports, and that ports can be opened multiple times - neither of which is ideal, but I don't find maintaining state for potentially 8 ports, or a user-defined MAX_PORTS generic parameter to be a particularly good solution either.


leftger (approved):
I appreciate your concerns but I think this is a good addition for the time being.

PR #5948, opened 2026-04-23
wpan: use bind_interrupts for wba

xoviat:


Tavish-Srivastava (approved):

leftger (approved):
Thanks for the cleanup xoviat!

PR #5949, opened 2026-04-23
DescriptorVisitor trait and helper method for USB Host

cryptographix:
Added a `DescriptorVisitor` trait and a `visit_descriptors` method to `ConfigurationDescriptor`. The method iterates through the descriptors and calls callbacks on the trait (which has default empty impl) such that a class implementing this trait receives the descriptors in order and can use heuristics to identify the device, check compatibility and stash relevant information from, for example, class-specific interfaces and endpoints.


leftger:
I think you addressed @bugadani 's comments as far as I can tell. I'll wait a day in case he has any further concerns. Overall this seems like a good PR.

embassy-usb-host/src/descriptor.rs
```diff
@@ -189,6 +193,68 @@ impl<'a> ConfigurationDescriptor<'a> {
...
+    }
+}
+
+/// Callback-based visitor for a configuration's descriptor tree.
+/// Implement only the methods you care about.
+/// Return `Ok(false)` to stop iteration early without an error, or `Err(e)` to stop with one.
+pub trait DescriptorVisitor<'a> {
+    type Error;
+
+    fn on_configuration(&mut self, _c: &ConfigurationDescriptor<'a>) -> Result<bool, Self::Error> {
```
bugadani: You've taken an idea and went with it in a slightly different direction :)

By error handling, I meant that potential errors from descriptor parsing should be surfaced, in my opinion. The visitor here is less interesting - it can maintain its own internal error if it needs to. I think these methods really only should return `bool` to control the flow.

embassy-usb-host/src/descriptor.rs
```diff
@@ -189,6 +193,68 @@ impl<'a> ConfigurationDescriptor<'a> {
...
+            return Ok(());
+        }
+        let mut current_iface: Option<InterfaceDescriptor<'a>> = None;
+        for (_, bytes) in self.iter_descriptors() {
+            if bytes.len() < 2 {
+                continue;
+            }
+            match bytes[1] {
+                descriptor_type::INTERFACE => {
+                    if let Ok(iface) = InterfaceDescriptor::try_from_bytes(bytes) {
```
bugadani: I think I should have asked first: does it make sense to continue parsing if `try_from_bytes` fails?

bugadani (commented):
Almost :)

Looking at this, PR, we have some cleaning up to do outside of this PR:

- What's the point of the `USBDescriptor` trait
- Why do we have two interface descriptor structs? The one in UAC looks like what we should have in `src/descriptor.rs`. The only thing that uses the lifetimed buffer right now seems to be HID, and it looks like another good candidate for this visitor pattern.

My recommendation is:

- The error handling is backwards to what I meant, but which one lands isn't really important to me. Decide, and move on :)
- Merge this pr
- Remove `USBDescriptor`, nothing needs it, and what could use it is duplicated in all `try_from_bytes` implementations. The associated error type makes it awkward anyway.
- Redo HID to not rely on `InterfaceDescriptor<'_>::iter_descriptors`
- Replace the current `InterfaceDescriptor<'_>` with `uac::descriptors::InterfaceDescriptor`

embassy-usb-host/src/descriptor.rs
cryptographix: Understood, returning bool would be simpler.

`on_other` is generally responsible for parsing non-basic descriptors (class-specific, vender, iad), and may fail. 

Maybe remove the errors from the standard callbacks (cfg/if/ep) only, and keep for `on_other`.

embassy-usb-host/src/descriptor.rs
cryptographix: Guess not. I was trying to be fault-tolerant here, but for the basic descriptors, if not parsable something serious is wrong and we should probably not accept the device. Will correct.

embassy-usb-host/src/descriptor.rs
cryptographix: Done. Reworked the error handling are per discussion.

PR #5951, opened 2026-04-23
MSC host driver

bugadani:
Claude-generated driver, Claude-generated fixes to make synopsys-otg not fail horribly. embedded-fatfs almost works, it just trips up on the MBR in my setup.

### DIFF
--- a/embassy-usb-host/Cargo.toml
+++ b/embassy-usb-host/Cargo.toml
@@ -31,6 +31,7 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 
 defmt = { version = "1", optional = true }
 log = { version = "0.4.14", optional = true }
+block-device-driver = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
 env_logger = "0.9"
@@ -38,3 +39,4 @@ env_logger = "0.9"
 [features]
 defmt = ["dep:defmt", "embassy-usb-driver/defmt", "heapless/defmt"]
 log = ["dep:log"]
+block-device-driver = ["dep:block-device-driver"]
--- a/embassy-usb-host/Cargo.toml
+++ b/embassy-usb-host/Cargo.toml
@@ -35,6 +35,8 @@ block-device-driver = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
 env_logger = "0.9"
+# Enable critical-section implementation for std, for tests
+critical-section = { version = "1.1", features = ["std"] }
 
 [features]
 defmt = ["dep:defmt", "embassy-usb-driver/defmt", "heapless/defmt"]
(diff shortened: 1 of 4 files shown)


elipsitz:
Perhaps there's an opportunity here to share some of the SCSI structs with the not-yet-merged device-side implementation (https://github.com/embassy-rs/embassy/pull/5928)?

bugadani:
There may be, but I'm not sure how valuable actually is at this point to introduce a shared crate at this point.

elipsitz:
Ah, right, they're in two separate crates.

leftger (commented):
I just have a question: would it be prudent to implement traits for something besides `block-device-driver`? I actually don't mind either way I'm just thinking if there's other downstream dependents that could use alternate traits.

bugadani:
I find this question always hard to answer, for two reasons:

## Should we impl `block-device-driver` here, or should we rely on an external adapter?

I'm not entirely sure we want to implement `block-device-driver`, even. Considering it requires an adapter type, that adapter type can be implemented in a compatibility crate. Such adapter can be implemented in three different places: here, in block-device-*, or in a third crate that doesn't exist yet.

In embassy-usb-host, we would either end up with
- a bunch of ecosystem-compat features, or
- dependencies that people may not need

One other thing: block-device-driver is already implemented by embassy-stm32, so pulling it into this project wouldn't be that weird :)

## What other traits should we implement?

While usb msc doesn't really provide a byte stream, block-device-adapters does the conversion and implements embedded-io-async traits for its adapter type. Apart from that, I'm not sure any relevant, de-facto standard abstractions exist yet.

After a short crates.io research:

- https://docs.rs/embedded-sdmmc/latest/embedded_sdmmc/blockdevice/trait.BlockDevice.html is blocking
- https://docs.rs/block_device/0.1.3/block_device/trait.BlockDevice.html is blocking
- what block device related crate I can see on crates.io are linux abstractions
- https://docs.rs/usbd-storage/2.0.0/usbd_storage/ is blocking, and usb-device specific anyway
- https://github.com/qiuchengxuan/exfat/blob/main/src/io/mod.rs#L20-L48 looks interesting, but also a bit questionable if we have to force the API shape of a random different crate

🤷‍♂️

bugadani:
cc @MabezDev because you made yourself relevant in this space :P

leftger:
I'll wait until Monday at noon (US Pacific Time) for any feedback from @MabezDev , otherwise I think we can merge.

leftger:
Any objections to merge?

MabezDev:
Sorry, I meant to reply here. I don't have enough skin in this game to know if these are the perfect traits for this class, but its probably more than good enough for a first run, and at the very least @bugadani has had some success with them.

Thanks for working on USB host btw @leftger, I'm excited to see USB host support land in esp-hal soon :D.

PR #5957, opened 2026-04-24
use stm32wb-hci in wba and factor out high-level logic

xoviat:
factor out high-level logic into separate mod and re-use parsing from stm32wb-hci. move parsing out of sequencer, and cleanup misc. low-level code.


leftger (approved):
Everything looks good from my end. This is the right approach. Thanks for looking into this @xoviat! You can merge this without issue once you've got everything set up.

xoviat:
@leftger please check OueslatiGhaith/stm32wb-hci#16 and see if I missed any events.

PR #5962, opened 2026-04-26
stm32: include STM32WBA in low-power stop refcount tracking

leftger:
## Summary
- Adds `stm32wba` to the cfg gate in `increment_minimum_stop_refcount_with_cs` and `decrement_minimum_stop_refcount_with_cs` so that enabling/disabling peripherals on STM32WBA properly increments/decrements the STOP mode refcounts.
- This was an oversight — WBA support predates the stop refcount mechanism (introduced for WL/WB in late 2025) and was never added to the cfg gate.
- Without this fix, peripheral enable/disable on WBA has no effect on stop refcounts, meaning the low-power executor cannot determine whether active peripherals should block STOP mode entry.

### DIFF
--- a/embassy-stm32/src/rcc/mod.rs
+++ b/embassy-stm32/src/rcc/mod.rs
@@ -408,7 +408,7 @@ impl RccInfo {
 
     #[allow(dead_code)]
     fn increment_minimum_stop_refcount_with_cs(&self, _cs: CriticalSection) {
-        #[cfg(all(any(stm32wl, stm32wb), feature = "low-power"))]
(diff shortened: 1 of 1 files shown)


xoviat:
Does entering stop2 require re-initialization of peripherals?

leftger:
According to RM0515: 
Screenshot 2026-04-27 at 2 15 02 PM

Now in this implementation: No, individual peripheral re-initialization is not needed, because the refcount mechanism makes it so peripherals that can't survive STOP2 (and would need re-init) must be dropped/disabled before STOP2 is entered.

PR #5966, opened 2026-04-26
usb-host: propagate BusState MAX generic through BusHandle and bus()

leftger:
## Summary

Follow-up to #5965. That PR added a `const MAX: u8 = 127` generic to `BusState`, but the consuming APIs (`BusHandle`, `bus()`, `HubHandler`, `Default for BusState`) still hardcoded the default `BusState` (= `BusState<127>`), so a `BusState::<200>::new()` couldn't actually be passed to `bus()` and the generic was effectively inert.

This PR threads `MAX` through:

- `BusHandle<'d, A, const MAX: u8 = 127>` — defaulted const generic, `state` field is now `&'d BusState<MAX>`.
- `bus<'d, C, const MAX: u8>(driver, state: &'d BusState<MAX>) -> (BusController<'d, C>, BusHandle<'d, C::Allocator, MAX>)` — `MAX` is inferred from the argument, no turbofish required at call sites.
- `impl<'d, A, const MAX: u8> UsbHostAllocator<'d> for BusHandle<'d, A, MAX>` — propagates `MAX`.
- `impl<const MAX: u8> Default for BusState<MAX>` — was previously only for the default type.
- `HubHandler<'d, A, const MAX_PORTS: usize, const MAX: u8 = 127>` — added so it can hold a `BusHandle<'d, A, MAX>`.

## Backwards compatibility

All existing call sites compile unchanged:

- `static BUS_STATE: BusState = BusState::new();` — uses the type-level default of 127.
- `bus(driver, &BUS_STATE)` — `MAX` is inferred from the `&BusState<MAX>` argument.
- Existing `BusHandle` / `HubHandler` uses without naming `MAX` get 127 via the type default.

### DIFF
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -18,8 +18,8 @@ use crate::descriptor::{DEFAULT_MAX_DESCRIPTOR_SIZE, InterfaceDescriptor, USBDes
 use crate::handler::{BusRoute, EnumerationInfo, HandlerEvent, RegisterError};
 use crate::{BusHandle, EnumerationError};
 
-pub struct HubHandler<'d, A: UsbHostAllocator<'d>, const MAX_PORTS: usize> {
-    bus: BusHandle<'d, A>,
+pub struct HubHandler<'d, A: UsbHostAllocator<'d>, const MAX_PORTS: usize, const MAX: u8 = 127> {
+    bus: BusHandle<'d, A, MAX>,
     interrupt_channel: A::Pipe<pipe::Interrupt, pipe::In>,
     control_channel: A::Pipe<pipe::Control, pipe::InOut>,
     desc: HubDescriptor,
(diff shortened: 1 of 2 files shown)


leftger:
@bugadani let me know what you think

Dirbaio:
do you think it could be made a runtime value instead of a const generic? this way we avoid infecting everything with it.

If it's needed to size arrays, we can do like embassy-net: make a ["resources" struct](https://github.com/embassy-rs/embassy/blob/e9b30942b874a785aec7f07f082d709a2e685b7d/embassy-net/src/lib.rs#L89) that does have the const generics, but that [gets type-erased](https://github.com/embassy-rs/embassy/blob/e9b30942b874a785aec7f07f082d709a2e685b7d/embassy-net/src/lib.rs#L330) when you construct the stack.

flaviojs:
It can also be make into a variable in the struct.
If you know of any hardware that has a usb max address based on some configuration (chosen at runtime), then it's a preferred solution over generics.

bugadani:
This feels pointless, the spec has a hard limit of 127 addresses (+ address 0 for the root). We're trying to optimize away some of the 16 bytes of bitmap necessary to track those addresses by making the codebase considerably worse.

flaviojs:
So that's where the 127 comes from, then I guess the other PR should be reverted.
I would still like it to use `usize` instead of `u64` though.

bugadani:
> I would still like it to use usize instead of u64 though.

I would like to see some codegen difference between the two, but otherwise why not.

flaviojs:
Codegen difference after reverting, with `cargo asm --target powerpc-unknown-linux-gnu BusState`:

u64 array (has extra code for non-native u64 inlined)
```
0 "embassy_usb_host::BusState::alloc_address" [168]
1 "embassy_usb_host::BusState::free_address" [113]
```

usize array (no extra code)
```
0 "embassy_usb_host::BusState::alloc_address" [153]
1 "embassy_usb_host::BusState::free_address" [98]
```

bugadani:
A tad surprising, but results are results :)

flaviojs:
I'll make a PR after the revert is merged (I noticed the new branch).

Issue #5843, opened 2026-04-12
embassy-nrf: nrf54l15 Incorrect Timer Channel assertion


cwriter:
I'm encountering an issue with the nrf54l15 timer channels:

## Issue
The [documentation](https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/timer.html) states that the timers have 6 and 8 capture/compare channels, respectively. However, the embassy-nrf mentions the existence 4/6 channels in a comment, which it later also asserts.
https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/timer.rs#L223

The default timer implementation for the nrf54l15 also assumes 4 channels, as for any other nrf device, not specifying them as extended:
https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/chips/nrf54l15_app.rs#L626

This leads to a crash in the [assertion ](https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/timer.rs#L225) when attempting to use any channel > 4.

The macro rule hardcodes these values (4/6) based on the "extended" flag:
https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/timer.rs#L45

## Fix
Is there a reason to reserve the high-channels for an internal `embassy` facility? Otherwise, would it make sense to always be explicit about the number of CC channels for every nrf chip, given that future iterations might further diversify the possible implemented values?

PR #5967, opened 2026-04-26
nrf: nrf54 timer channel fix

cwriter:
Propses fix for #5843 by explicitly specifying the channel numbers.

### Summary ###
The patch assigns a channel number of 6 for all `timer_impl!` in the chip definition of nrf54-type chips, except for TIMER10, which has 8 channels (see the [online documentation](https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/timer.html))

I've kept the patch as minimal as possible intentionally. Specifically, I did not touch the timer_impl! extended case, even though I'm not sure using a boolean flag makes sense compared to simply specifying the explicit number of channels in other nrf-chips as well.


lulf (approved):
Thanks!

PR #5970, opened 2026-04-27
stm32/i2c: Fix 10-bit slave address detection in determine_matched_address()

### LINKED ISSUE #5969: stm32/i2c: Incorrect 10-bit slave address detection in determine_matched_address()
When testing I2C slave mode with a 10-bit address using `embassy-stm32`, I found that `listen()` returns an incorrect 7-bit address. After reviewing the source code and reference manual, I believe there are two issues in the current address detection logic in embassy-stm32 (`61faf6f911c0ad8e99f5ec972d7e1fa6a8302c4d`).

### Problem 1: Incorrect bit shifting for 10-bit address detection

The current [`determine_matched_address`](https://github.com/embassy-rs/embassy/blob/61faf6f911c0ad8e99f5ec972d7e1fa6a8302c4d/embassy-stm32/src/i2c/v2.rs#L1622C5-L1634C6) function reads `addcode()` from `self.info.regs.isr().read()` and right-shifts by 3 bits before comparing with `0b11110`. However, `addcode()` returns the `ADDCODE` field from the ISR register, which is **7 bits**. According to the reference manual (STM32 RM0456, Section 65.9.7 I2C_ISR), for 10-bit address matching, the `ADDCODE` field contains `11110XX` (the 5 MSBs of the 10-bit address followed by the 2 MSBs of the address). To properly detect a 10-bit address match, it should be right-shifted by **2 bits** instead of 3.

### Problem 2: Cannot retrieve the full 10-bit address after detection

Even after fixing the address type detection, retrieving the full 10-bit address remains problematic. 
According to the STM32 reference manual (RM0456), during 10-bit address matching in slave mode, the address bytes are consumed by the hardware as part of the address recognition process. They do not trigger RXNE nor are they stored in I2C_RXDR. Therefore, after an ADDR event, the software cannot retrieve the full 10-bit address from RXDR. The only way to retrieve the configured 10-bit address is likely through `self.info.regs.oar1().read().oa1()`.

### Reproduction Code

```rust
#[embassy_executor::task]
async fn task1_i2c(i2c: embassy_stm32::i2c::I2c<'static, embassy_stm32::mode::Async, i2c::Master>) {
    let mut i2c = i2c.into_slave_multimaster(SlaveAddrConfig { 
        addr: i2c::OwnAddresses::OA1(Address::TenBit(391)), 
        general_call: false 
    });
    let mut buf = [0u8; 10];
    loop {
        match i2c.listen().await {
            Ok(slave) => {
                defmt::info!("slave command: {:?}", slave);
                i2c.respond_to_write(&mut buf).await.unwrap();
                let s = str::from_utf8(&buf[0..4]).unwrap();
                defmt::info!("buf: {:?}", s);
            },
            Err(e) => {
                defmt::error!("failed to listen: {:?}", e);
                Timer::after_secs(1).await;
            },
        }
    }
}
```
### Expected vs Actual Behavior
Expected output: `SlaveCommand { kind: Write, address: TenBit(391) }`
Actual output: `SlaveCommand { kind: Write, address: SevenBit(121) }`

chenfengyuan:
Fixes https://github.com/embassy-rs/embassy/issues/5969
### Problem

When using I2C slave mode with a 10-bit address, `listen()` returns an incorrect 7-bit address. For example, configuring address `TenBit(391)` results in `SevenBit(121)`.

### Root Cause

Two issues in `determine_matched_address()`:

1. **Incorrect bit shift**: The `addcode()` field from ISR is 7 bits. For 10-bit address detection, comparing `matched >> 3 == 0b11110` is wrong - it should be `matched >> 2 == 0b11110` (since `0b11110XX >> 2 = 0x1E`).

2. **Wrong address retrieval**: The current code attempts to read the second address byte from RXDR via `slave_read_internal()`. According to RM0456, the hardware consumes both address bytes during address matching and does not store them in RXDR. The correct approach is to read the configured address from the OAR1 register.

### Solution

- Change the bit shift from 3 to 2 for 10-bit address detection
- Retrieve the address from `self.info.regs.oar1().read().oa1()` instead of reading from RXDR

### Testing
```rust
#[embassy_executor::task]
async fn task1_i2c(i2c: embassy_stm32::i2c::I2c<'static, embassy_stm32::mode::Async, i2c::Master>) {
    let mut i2c = i2c.into_slave_multimaster(SlaveAddrConfig { addr: i2c::OwnAddresses::OA1(Address::TenBit(391)), general_call: false });
    let mut buf = [0u8;10];
    loop {
        match i2c.listen().await {
            Ok(slave) => {
                defmt::info!("slave command: {:?}", slave);
                i2c.respond_to_write(&mut buf).await.unwrap();
                let s = str::from_utf8(&buf[0..4]).unwrap();
                defmt::info!("buf: {:?}", s);
            },
            Err(e) => {
                defmt::error!("failed to listen: {:?}", e);
                Timer::after_secs(1).await;
            },
        }
    }
}
```
With this fix, `listen()` correctly returns `TenBit(391)` instead of `SevenBit(121)`.

### Related

STM32 RM0456

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -1622,12 +1622,9 @@ impl<'d, M: Mode> I2c<'d, M, MultiMaster> {
     fn determine_matched_address(&self) -> Result<Address, Error> {
         let matched = self.info.regs.isr().read().addcode();
 
-        if matched >> 3 == 0b11110 {
-            // is 10-bit address and we need to get the other 8 bits from the rxdr
-            // we do this by doing a blocking read of 1 byte
-            let mut buffer = [0];
-            self.slave_read_internal(&mut buffer, self.timeout())?;
-            Ok(Address::TenBit((matched as u16) << 6 | buffer[0] as u16))
+        if matched >> 2 == 0b11110 {
+            let address = self.info.regs.oar1().read().oa1();
+            Ok(Address::TenBit(address))
         } else {
             Ok(Address::SevenBit(matched))
         }


xoviat:
Does this impact #5903?

chenfengyuan:
> Does this impact #5903?

No, my commit is a descendant of that merge, so there's no impact.

xoviat:
cc @tomchuk

tomchuk:
Looks good to me. 👍 

1. This is totally independent from my previous #5903 (v1 master mode 10-bit vs v2 slave 10-bit)
2. Bit shift is now correct: `ADDCODE` is `11110XX`, shifted `>> 2` isolates the 5-bit `11110` prefix.
3. Verified in the RM that `OAR1` is the correct register - the hardware consumes both address bytes during matching, and OA1 is the only register that can hold a 10-bit address (OA2 is 7-bit only, per 65.9.4)

PR #5973, opened 2026-04-27
Change the BusState array from u64 to usize.

flaviojs:
In 64-bit systems, the same code will be generated.
In 32-bit systems, the extra code for non-native u64 will be avoided.

[Codegen comparisson](https://github.com/embassy-rs/embassy/issues/5966#issuecomment-4322607411)

--

Also clarified why the addresses only go up to 127.

### DIFF
--- a/embassy-usb-host/src/lib.rs
+++ b/embassy-usb-host/src/lib.rs
@@ -79,8 +79,10 @@ impl core::error::Error for EnumerationError {}
 ///
 /// Holds the in-use USB device addresses (1–127) and an async mutex used to
 /// serialise enumerations on a single bus.
+///
+/// The behavior of addresses greater than 127 is undefined (USB 2.0 §9.4.6).
 pub struct BusState {
-    addr_bitmap: BlockingMutex<CriticalSectionRawMutex, RefCell<[u64; 2]>>,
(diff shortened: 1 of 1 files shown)


embassy-usb-host/src/lib.rs
```diff
@@ -79,16 +79,18 @@ impl core::error::Error for EnumerationError {}
 ///
 /// Holds the in-use USB device addresses (1–127) and an async mutex used to
 /// serialise enumerations on a single bus.
+///
+/// The behavior of addresses greater than 127 is undefined (USB 2.0 §9.4.6).
```
bugadani: We don't issue addresses greater than 127, is this necessary?

flaviojs: If I understood WHY we don't issue addresses greater than 127 my original PR would not have been created and later reverted, I'm merely documenting it.

Also, it's perfectly possible that some hardware does support addresses greater than 127 or has specific uses for them and this structure might need to change to accommodate that use case, you never know what vendors will do.

flaviojs: I'll try rewording it to improve clarity.

xoviat:
bender run

xoviat:
bender run

xoviat:
bender run

xoviat:
bender run

bugadani (approved):

Issue #4318, opened 2025-06-18
Embassy-USB: Device_Qualifier request should produce a Error for USB2.0 FS-Device


sehicfai:
I am currently working on a USB driver implementation for a PSoC6 (USB Full Speed device). During testing I encountered that the usb device would accept the "Get Device Qualifier" request which is sent by the host by sending an ACK. 

According to the USB 2.0 Specification chapter 9.6.2 Device_Qualifier: "If a full-speed only device (with a device descriptor version number equal to 0200H) receives a GetDescriptor() request for a device_qualifier, it must respond with a request error (STALL). The host must not make a request for an other_speed_configuration descriptor unless it first successfully retrieves the device_qualifier descriptor." 

There is currently no check implemented if the device support USB High Speed [Current implementation:](https://github.com/embassy-rs/embassy/blob/main/embassy-usb/src/lib.rs)

`descriptor_type::DEVICE_QUALIFIER => {
                InResponse::Accepted(&self.device_qualifier_descriptor)
            }`

My idea to fix this would be [Changes:](https://github.com/embassy-rs/embassy/commit/978d818c9a7442ef6ce91c7f2ad5dc90a6b8f6f5)

bugadani:
As the relevant PR has been merged, perhaps this issue can be closed now.

PR #5981, opened 2026-04-28
USB: reject DEVICE_QUALIFIER req for non-HS device

andreasWallner:
As discussed in the chat recently: There is currently an issue in the stack that it will [respond ](https://github.com/embassy-rs/embassy/blob/ddbe69377fc1797bdf4ed0a4a0b7d312388e0a06/embassy-usb/src/lib.rs#L801) to a DEVICE_QUALIFIER descriptor request. 

The USB 2.0 spec (April 27, 2000) in 9.6.2, page 264 states:

> If a full-speed only device (with a device descriptor version number equal to 0200H) receives a GetDescriptor() request for a device_qualifier, it must respond with a request error.

In the current implementation this response is always sent.

# Thoughts on the actual implementation:

- In #4318 there was a proposal to add a `const` property to the `Driver` to state what the max speed of the hardware is.
- In the chat the point was that we should rather go for a function on the `Driver` so that drivers can be configured differently, depending on the hardware, even if the same driver can be used (Synopsys IP with same interface, different config).
- Some side-discussion in the chat later pointed me to the `max_speed` setting, and it seemed to make most most sense to me to use that property. Biggest reason to me would be that if a user states a `max_speed` of full-speed it shouldn't matter that the hardware is capable of doing high speed. So the actual hardware property seems less important compared to the user intent.

If a user really sets `max_speed` = high-speed, and the hardware only supports full-speed - is this something that really needs to be cought? If yes the problem here is that during the config generation there are no APIs that would allow for that to be fallible...

I'm happy to adapt this PR to any of the other options, etc - if you prefer that.

### DIFF
--- a/embassy-usb/src/builder.rs
+++ b/embassy-usb/src/builder.rs
@@ -20,7 +20,7 @@ pub enum UsbVersion {
     TwoOne = 0x0210,
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 /// The fastest speed the device configuration supports.
 pub enum UsbDeviceSpeed {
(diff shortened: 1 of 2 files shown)


bugadani (approved):

andreasWallner:
thanks for the really quick turnaround!

PR #5992, opened 2026-04-29
usb/synopsys-otg: implement remote wakeup

ghost:
## Summary

`Bus::remote_wakeup()` in `embassy-usb-synopsys-otg` unconditionally
returns `Err(Unsupported)`, meaning USB remote wakeup has never worked
on any chip using the Synopsys OTG IP regardless of configuration. This
includes STM32 via `embassy-stm32` and ESP32-S2/S3 via `esp-hal`, both
of which delegate to this crate.

This PR implements remote wakeup directly in `embassy-usb-synopsys-otg` by:

1. Ungating the PHY clock and HCLK stopped during suspend
   (`PCGCCTL.STPPCLK` / `PCGCCTL.GATEHCLK`)
2. Asserting the K-state resume signal on D+/D- (`DCTL.RWUSIG`) for 10 ms
   per USB 2.0 spec §7.1.7.7: TDRSMUP (1–15 ms required)
3. Deasserting the signal

`GATEHCLK` is only cleared on HS cores as it is not present on FS-only
peripherals. `embassy-time` is an unconditional dependency of this crate
so no feature flag or fallback is needed.

Tested on STM32F401RC (OTG FS) via the [Keychron Q6 HE RMK firmware](https://github.com/fuchskurt/keychron_q6_he_rmk) and on ESP32-S3

## References

- USB 2.0 spec §7.1.7.7 (resume signaling timing, TDRSMUP: 1–15 ms)
- STM32F4 RM0368 §22.5.2 (Peripheral states — Suspended state)
- STM32F4 RM0368 §22.8 (OTG low-power modes)
- STM32F4 RM0368 §22.16.5 (OTG_FS_PCGCCTL)

### DIFF
--- a/embassy-usb-synopsys-otg/src/lib.rs
+++ b/embassy-usb-synopsys-otg/src/lib.rs
@@ -1145,7 +1145,25 @@ impl<'d, const MAX_EP_COUNT: usize> embassy_usb_driver::Bus for Bus<'d, MAX_EP_C
     }
 
     async fn remote_wakeup(&mut self) -> Result<(), Unsupported> {
-        Err(Unsupported)
+        let r = self.instance.regs;
+
+        // Re-enable PHY clock gated during suspend.
+        // See RM0368 §22.8 "OTG low-power modes" (STPPCLK / GATEHCLK).
+        r.pcgcctl().modify(|w| {
+            w.set_stppclk(false);
+            // GATEHCLK is only present on HS cores.
+            if self.instance.phy_type.high_speed() {
+                w.set_gatehclk(false);
+            }
+        });
+
+        // Assert resume K-state on D+/D-.
+        // USB 2.0 spec §7.1.7.7: TDRSMUP requires 1–15 ms of resume signaling.
+        r.dctl().modify(|w| w.set_rwusig(true));
+        embassy_time::Timer::after_millis(10).await;
+        r.dctl().modify(|w| w.set_rwusig(false));
+
+        Ok(())
     }
 }
 


leftger:
@bugadani I think the ESP32s could benefit from this too so I think @fuchskurt is correct in it maybe belonging to `embassy-usb-synopsys-otg`. Let me know what you think bugadani.

bugadani:
Obviously the cortex_m specific code doesn't really fit in the synopsys crate, but the rest does. How's one supposed to test this?

> STM32F4 RM0368 §22.16.5 (remote wakeup sequence)

That's "OTG_FS power and clock gating control register (OTG_FS_PCGCCTL)".

ghost:
> How's one supposed to test this?

That's why I was hesitant to open the PR directly against `embassy-usb-synopsys-otg`. I can only verify it on the STM32 devices through my implementation of the [Keychron Q6 HE RMK firmware](https://github.com/fuchskurt/keychron_q6_he_rmk). I don’t have a way to test this on ESP32 or other platforms using the same IP. 

That said, the risk of introducing a regression should be low: the previous implementation unconditionally returned `Err(Unsupported)`, so these platforms would effectively have treated this as a no-op already.

> That's "OTG_FS power and clock gating control register (OTG_FS_PCGCCTL)".

Thanks, I’ve updated the references accordingly.

bugadani:
I mean, I can test this on an ESP32, but I need to know what setup I need and what I need to expect.

ghost:
I put together a quick test case in a [esp-hal fork](https://github.com/fuchskurt/esp-hal) under `examples/async/embassy_usb_otg_wakeup`. It triggers remote wakeup by pressing the boot button (GPIO0) on an ESP32-S3 devboard. Note that to actually test this fix the example must be built against a version of `embassy-usb-synopsys-otg` that includes this PR's adapted implementation, with the upstream stub the result will always be `Err(Unsupported)`.
>
> To reproduce:
> 1. Flash the example to an ESP32-S3 devboard
> 2. Connect via USB and open a serial monitor to see log output
> 3. Enable autosuspend on the host:
>    ```bash
>    echo auto > /sys/bus/usb/devices/<device>/power/control
>    ```
>    You can find your device path with `lsusb` and then `ls /sys/bus/usb/devices/`
> 4. Wait a few seconds for the host to suspend the device, the serial monitor should print `Suspended`
> 5. Verify wakeup is armed by the host:
>    ```bash
>    cat /sys/bus/usb/devices/<device>/power/wakeup
>    ```
>    This should print `enabled`. If it prints `disabled`, the host will not respond to remote wakeup regardless of what the device does.
> 6. Press the boot button (GPIO0)
> 7. **Expected with this fix:** serial monitor prints `Triggering remote wakeup` -> `Remote wakeup OK` -> `Active`, and the host wakes up
> 8. **Expected without this fix:** serial monitor prints `Triggering remote wakeup` -> `Remote wakeup error: Unsupported`, host stays suspended

bugadani:
It took me a while to get this working on Windows, but it is working :)

ghost:
Thank you for taking the time, glad it worked on ESP! Sorry for not including the Windows steps earlier, I overlooked that. The extra configuration is required because Embassy strictly follows the specification and only drives the resume signal after the host has armed remote wakeup via SET_FEATURE(DEVICE_REMOTE_WAKEUP). Other firmware stacks often skip this check and drive the K-state unconditionally, which is why they appear to work out of the box without any host-side setup.

Since placing this in embassy-usb-synopsys-otg seems like the better approach, should I update this PR to move the implementation there? If so, should I upstream the example to esp-hal as well?

bugadani:
>  should I update this PR to move the implementation there?

Yes please!

> If so, should I upstream the example to esp-hal as well?

I don't think there is a need for that, thank you :)

ghost:
Done!

bugadani (approved):

leftger (approved):

PR #5996, opened 2026-04-29
embassy-sync: fix `PriorityChannel::poll_ready_to_send` always returns Pending

Rechenmaschine:
Due to what looks like a typo, the poll condition is false in (essentially) every reachable state and thus returns `Poll::Pending` when it shouldn't.

`PriorityChannel::poll_ready_to_send` doesn't seem to be used anywhere across embassy, so its effect has likely remained invisible – but it is reachable from the public API.


lulf:
bender run

PR #6000, opened 2026-04-30
embassy-sync: fix OnceLock init memory ordering

Rechenmaschine:
OnceLock is missing ordering guarantees on the atomic `init` flag

The relaxed store to `init` in `init()`/`get_or_init()` and relaxed load from `init` in `try_get()` does not enforce a happens-before relation

Seeing `init == true` then does not guarantee that the initialized value in `data` is already visible to a concurrent reader so `try_get()` could create a reference to memory that is still uninitialized from the readers pov.

`LazyLock` already uses Acquire/Release for the same pattern.

Miri detects the bug when testing:

```rust
fn main() {
    let lock = Arc::new(OnceLock::<u64>::new());
    let l = lock.clone();
    let r = thread::spawn(move || while l.try_get().is_none() {});
    lock.init(42).unwrap();
    r.join().unwrap();
}
```

```
error: Undefined Behavior: Data race detected between (1) non-atomic write on thread `main` and (2) retag read of type `std::mem::MaybeUninit<u64>` on thread `unnamed-1` at alloc308+0x10
--> embassy-sync/src/once_lock.rs:158:9
    |
158 |         (*self.data.as_ptr()).assume_init_ref()
    |         ^^^^^^^^^^^^^^^^^^^^^ (2) just happened here
    |
help: and (1) occurred earlier here
   --> src/main.rs:10:5
    |
 10 |     lock.init(42).unwrap();
    |     ^^^^^^^^^^^^^
...
```

### DIFF
--- a/embassy-sync/src/once_lock.rs
+++ b/embassy-sync/src/once_lock.rs
@@ -77,7 +77,7 @@ impl<T> OnceLock<T> {
 
     /// Try to get a reference to the underlying value if it exists.
     pub fn try_get(&self) -> Option<&T> {
-        if self.init.load(Ordering::Relaxed) {
+        if self.init.load(Ordering::Acquire) {
             Some(unsafe { self.get_ref_unchecked() })
         } else {
             None
@@ -92,7 +92,7 @@ impl<T> OnceLock<T> {
             // If the value is not set, set it and return Ok.
             if !self.init.load(Ordering::Relaxed) {
                 self.data.set(MaybeUninit::new(value));
-                self.init.store(true, Ordering::Relaxed);
+                self.init.store(true, Ordering::Release);
                 Ok(())
 
             // Otherwise return an error with the given value.
@@ -113,7 +113,7 @@ impl<T> OnceLock<T> {
             // If the value is not set, set it.
             if !self.init.load(Ordering::Relaxed) {
                 self.data.set(MaybeUninit::new(f()));
-                self.init.store(true, Ordering::Relaxed);
+                self.init.store(true, Ordering::Release);
             }
         });
 


diondokter:
The acquire is good. But I'm not sure the release ones are necessary since they're in a critical section

Rechenmaschine:
> But I'm not sure the release ones are necessary since they're in a critical section

Is there something in the contract of critical section that would guarantee a reader observing `init=true` also observes the writes that happend before the write to `init` in the writer?

I saw that [critical-section's contract for acquire/release](https://docs.rs/critical-section/latest/src/critical_section/lib.rs.html#192-194) mentions 

> It must provide ordering guarantees at least equivalent to a `Ordering::Acquire` on a memory location shared by all critical sections, on which the release call will do a `Ordering::Release` operation.

but as i understand it, this ordering guarantee only forms an edge between threads that both want to enter the CS. And  `try_get` doesn't enter the CS, so it doesn't get these guarantees. It only reads the init flag, and a relaxed store there doesn't pair with the acquire load, so some thread calling `try_get` might still see init=true before the data is observable (thus reading uninitialised data).

I unfortunately don't have the embedded hardware on hand currently to replicate this, but on std miri still flags flags a data race if I don't include the release stores.

Rechenmaschine:
It's also possible to actually produce the error on std (target aarch64) with this test https://gist.github.com/Rechenmaschine/b740aee94804f45a50ae98fa5e5555f0

Same deal here, without the `release` ordering it falls through.

diondokter:
That might be a problem with the critical-section std implementation.
But maybe not...

In any case, it's not a great harm to have the release ordering and it does seem to help with edge cases, so let's just merge

PR #6001, opened 2026-04-30
cyw43-pio: fix deterministic single-bit RX corruption on RP2350

tralamazza:
## Summary

Fixes a deterministic single-bit corruption on `cyw43-pio` backplane reads on RP2350 (Pico 2 W).

## Root cause

`cmd_read` (and `cmd_read_status`) drove TX and RX over a single DMA channel, sequentially:

```rust
self.sm.set_enable(true);
self.sm.tx().dma_push(&mut self.dma, &cmd, false).await;  // TX first
self.sm.rx().dma_pull(&mut self.dma, read, false).await;  // RX armed only after TX completes
```

The PIO state machine is enabled before RX DMA is armed. Output produced in that gap fills the 4-deep RX FIFO. By the time the SM samples the final bit of the read, the FIFO is already saturated and the last bit
is dropped, manifesting as a cleared LSB on the last word.

## Fix

Use two DMA channels (TX + RX) and arm them concurrently with `embassy_futures::join`, so RX is ready before the SM produces any samples. This mirrors the pico-sdk pattern.

```rust
let (rx, tx) = self.sm.rx_tx();
let rx_fut = rx.dma_pull(&mut self.dma_rx, read, false);
let tx_fut = tx.dma_push(&mut self.dma_tx, &cmd_arr, false);
embassy_futures::join::join(tx_fut, rx_fut).await;
```

## Reproduction

Hardware: Raspberry Pi Pico 2 W.

```bash
cd examples/rp235x
cargo run --bin blinky_wifi
```

Any `cyw43`-using binary on RP2350 in debug mode should reproduce. Release builds are fast enough to sometimes avoid the race.

## Symptom

Init panics during firmware load with a checksum mismatch on a 16-byte sample read back from the chip:

```text
[DEBUG] loading fw
[DEBUG] FW sample @00000000 checksum exp=a1f2d985 got=a1f24d24 match=false
[ERROR] FW expected: [00, 00, 00, 00, 71, 14, 00, 00, 9d, 13, 00, 00, 9d, 13, 00, 00]
        actual:      [00, 00, 00, 00, 71, 14, 00, 00, 9d, 13, 00, 00, 9c, 13, 00, 00]
                                                                      ^^
panicked at cyw43/src/lib.rs:458: firmware checksum mismatch
```

The error is deterministic and always in the same position (LSB of the last word of every 16-byte read). This was verified in 2 different firmware blobs.

### DIFF
--- a/cyw43-pio/CHANGELOG.md
+++ b/cyw43-pio/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Fix deterministic single-bit corruption on RP2350 backplane reads by using two DMA channels (TX + RX). The single single-channel pattern raced with the PIO state machine in debug builds, causing the LSB of the last sampled word in `cmd_read` to be cleared.
+- Breaking: `PioSpi::new` now takes two `dma::Channel` arguments (TX and RX) instead of one.
+
 ## 0.10.0 - 2026-03-10
 
 - Updated to use new DMA `Channel` driver struct from embassy-rp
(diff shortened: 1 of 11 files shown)


edisno:
Also affects RP2040 (pico w), I get the same "FW expected" on latest embassy git. :(

Update: Your fix .. fixes it.

Dirbaio:
very nice find! thanks

PR #6003, opened 2026-04-30
stm32/ipcc: optimize and cleanup

xoviat:
The existing logic in wpan/wb/mac assumed that the channel was blocked until the macevent was read. Previously, this was changed so that in other applications, the other core would be free to send more data. However, this broke this use case so I added a `close` flag to allow the user to determine whether they want the channel closed.

I also optimized the logic so that if the IPCC channel has data present, the interrupt is bypassed and the data is read/written directly. 

confirmed on hardware.

### DIFF
--- a/embassy-stm32/src/ipcc.rs
+++ b/embassy-stm32/src/ipcc.rs
@@ -122,6 +122,35 @@ impl Default for Config {
     }
 }
 
+struct ActiveTxInterrupt {
+    core: CoreId,
+    index: u8,
+}
+
+impl ActiveTxInterrupt {
+    fn configure_interrupt(&mut self, enabled: bool) {
+        critical_section::with(|_| {
+            IPCC::regs()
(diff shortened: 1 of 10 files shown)


embassy-stm32/src/ipcc.rs
```diff
@@ -163,27 +192,16 @@ impl<'a> IpccTxChannel<'a> {
...
             compiler_fence(Ordering::SeqCst);
 
             if !regs.cpu(core.to_index().into()).sr().read().chf(self.index as usize) {
-                // If bit is set to 1 then interrupt is disabled; we want to disable the interrupt
-                critical_section::with(|_| {
-                    regs.cpu(core.to_index().into())
-                        .mr()
-                        .modify(|w| w.set_chfm(self.index as usize, true))
-                });
-
```
xoviat, line 186: These changes require the irq handler not to disable the irq until chf is cleared, which it does.

PR #6004, opened 2026-04-30
embassy-sync: remove unsound zerocopy_channel `Sender::clear`

Rechenmaschine:
`Sender::clear` resets the channel front/back/full indices without checking whether the peer endpoint still holds an outstanding slot.  

We can use this in safe code to obtain two aliasing &mut T to the same buffer element

```rust
fn main() {
    let mut buf = [0u32; 2];
    let mut channel = Channel::<CriticalSectionRawMutex, u32>::new(&mut buf);
    let (mut sender, mut receiver) = channel.split();
    let mut slot = sender.try_send().unwrap();
    *slot = 42;
    slot.send_done();

    // receiver holds &mut u32 to buf[0]
    let r = receiver.try_receive().unwrap();

    // clear channel state while r is still alive
    sender.clear();
    // channel now believes it's empty, but r still points to buf[0]

    // sender takes what it thinks is a "fresh" &mut but now aliases buf[0]
    let s = sender.try_send().unwrap();

    // r and s alias
    assert_ne!(*r as *const u32, *s as *const u32);
}
```

This PR simply removes `Sender::clear`. In previous commits, I tried implementing a `try_clear()` instead, but this modified the Drop impl, and caused problems in other crates.

Note: `Receiver::clear` does not seem to be directly exploitable the same way, since `try_receive()` retruns None on an empty channel. `Channel::clear(&mut self)`needs exclusive access and is thus also not affected.

### DIFF
--- a/embassy-sync/src/zerocopy_channel.rs
+++ b/embassy-sync/src/zerocopy_channel.rs
@@ -24,12 +24,6 @@ use crate::blocking_mutex::Mutex;
 use crate::blocking_mutex::raw::RawMutex;
 use crate::waitqueue::WakerRegistration;
 
-/// Error returned by [`Sender::try_clear`] when the [`Receiver`]
-/// still holds an outstanding [`ReceiveSlot`].
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct TryClearError;
-
 /// A bounded zero-copy channel for communicating between asynchronous tasks
 /// with backpressure.
 ///
@@ -65,7 +59,6 @@ impl<'a, M: RawMutex, T> Channel<'a, M, T> {
                 front: 0,
                 back: 0,
                 full: false,
-                outstanding_receive: false,
                 send_waker: WakerRegistration::new(),
                 receive_waker: WakerRegistration::new(),
             })),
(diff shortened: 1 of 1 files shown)


Rechenmaschine:
Modifying the Drop impl causes some problems downstream. Probably simplest to just remove `Sender::clear` since it doesn't seem to be used in embassy... is this ok?

xoviat:
Or mark it unsafe?

Rechenmaschine:
> Or mark it unsafe?

Thanks for your suggestion! I considered this, but found it difficult to formulate a useful contract that is both useful and checkable at the call site.

One might formulate
> Safety: Any `ReceiveSlot` alive at the time of this call must be dropped before the next `send` / `try_send` / `poll_send` call on this `Sender`.

which is the most permissive I could come up with, but isn't a clean precondition and thus becomes awkward to verify since it includes an oblication about future behaviour.

A cleaner precondition is
> Safety: no `ReceiveSlot` may be alive when calling this.

It's a bit weaker, but easier to reason about. Still, I don't know how useful this is in practice since after a split the sending side typically has no knowledge of what the receiver is doing.

You do raise a good point though, because if I just remove the function now there may well be a PR sometime in the future that simply puts it back, becuase it was proposedly 'forgotten'.

Imo the second condition would be most sensible if we go this route, but maybe you have another idea?

xoviat:
Tbh, I actually think it should be removed because a sender should not be able to read from a channel, which is what this does.

xoviat (approved):

PR #6006, opened 2026-04-30
Don't let bootrom flash routines nuke PSRAM (QMI snapshot + cache clean) 

Citingmarc:
Hey there!

The RP2350 bootrom flash helpers ('connect_internal_flash' -> 'flash_exit_xip' -> 'flash_range_erase/program' -> 'flash_flush_cache') was silently reset QMI window 1 back to default XIP timings. On boards with PSRAM wired to CS1 (like the Pimoroni Pico LiPo 2 XL W I'm using) it would yeet whatever the was configured via embassy_rp::psram::Psram::new(...), and subsequent CPU accesses to the PSRAM aperture would return garbage.

I was hitting it to save a small KV store but then long-lived bytes in PSRAM went to `0xCC` right after the first save.

So what I did was inside 'flash::in_ram', snapshot the five 'QMI.mem(1)' registers before the bootrom call and restore them after, gated on _rp235x as to not affect RP2040. Followed by a compiler_fence(SeqCst) so subsequent PSRAM accesses see the restored config. I don't think this should affect regular usage, at least it solved my issue on this particular board.

Hope this helps.

### DIFF
--- a/embassy-rp/src/flash.rs
+++ b/embassy-rp/src/flash.rs
@@ -955,8 +955,37 @@ pub(crate) unsafe fn in_ram(operation: impl FnOnce()) -> Result<(), Error> {
         // Wait for completion of any background reads
         while pac::XIP_CTRL.stream_ctr().read().0 > 0 {}
 
+        // On RP235x the bootrom flash routines reset QMI window 1
+        // (CS1) to default XIP timings, which clobbers any custom
+        // configuration applied for an attached PSRAM. Snapshot the
+        // window-1 registers and restore them afterwards so PSRAM
+        // access keeps working across flash erase/program.
+        // See pico-sdk#1983.
+        #[cfg(feature = "_rp235x")]
+        let qmi_m1 = {
+            let m = pac::QMI.mem(1);
+            (
+                m.timing().read(),
+                m.rfmt().read(),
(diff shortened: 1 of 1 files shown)


Citingmarc:
turns out there was more. After that fix I was still getting PSRAM corruption sometimes. It looked like *sometimes* bytes that survived the QMI restore would come back wrong after saving. Took a while to pin down but I think I got it down to the XIP cache. The bootrom's flash_flush_cache invalidates the unified cache by set/way, which silently drops any CPU writes that were in the way to PSRAM. So the QMI restore was fine, the data was just gone.

I fixed it by adding a xip_cache_clean_all()-equivalent loop before operation() to put every dirty line back to PSRAM first, so the bootrom's invalidate is harmless. Maintenance writes go to 0x1BFFC000 (top 16 KiB of the maintenance window) from what I could find on the forum.

Reproduced both bugs on my board by using 256 KiB of PSRAM a deterministic pattern then 100 blocking_erase + blocking_write iterations. The fix made the mismatch data count drop from ~14k/iter to 0. Also sanity-checked against the uncached PSRAM alias (0x15000000) the whole time, which stayed clean throughout, and that's what pointed me at the cache rather than the QMI registers in the first place.

PR #6008, opened 2026-05-01
Fix spurious crash of stm32wl during latency change

shilga:
Current guess is that changing latency while prefetcher is enabled is not a good idea. This change just switches up latency change and prefetch enable

### DIFF
--- a/embassy-stm32/src/rcc/l.rs
+++ b/embassy-stm32/src/rcc/l.rs
@@ -185,7 +185,6 @@ pub(crate) unsafe fn init(config: Config) {
     #[cfg(stm32wl)]
     {
         // Set max latency
-        FLASH.acr().modify(|w| w.set_prften(true));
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rcc/l.rs
```diff
@@ -383,11 +382,14 @@ pub(crate) unsafe fn init(config: Config) {
 
     #[cfg(stm32l1)]
     FLASH.acr().write(|w| w.set_acc64(true));
-    #[cfg(not(stm32l5))]
-    FLASH.acr().modify(|w| w.set_prften(true));
+
     FLASH.acr().modify(|w| w.set_latency(latency));
     while FLASH.acr().read().latency() != latency {}
 
+    // enable prft after switching latency
```
jamesmunns:
```suggestion
    // enable prefetch (prft) after switching latency
```

PR #6010, opened 2026-05-01
wpan/wb: allow sending acl data

xoviat:
again, untested. the locking mechanism should work, but I am not sure.


xoviat:
bender run

PR #6013, opened 2026-05-01
[MCXA5] fix flexspi and DMA bugs

### LINKED ISSUE #188: FlexSPI and DMA bugs
Stress testing has uncovered a few bugs on flexspi and dma drivers.

felipebalbi:
Spent some time writing some stress tests for flexspi and ended up finding a couple bugs. One of which was actually in the DMA driver.

The DMA changes could use some careful review, but this is best I could get to guarantee that we wouldn't have any races with `Drop` and completion.

Closes https://github.com/OpenDevicePartnership/embassy-mcxa/issues/188


embassy-mcxa/src/dma.rs
```diff
@@ -1421,20 +1421,56 @@ impl<'a> Transfer<'a> {
     }
 
     /// Abort the transfer.
+    ///
+    /// Runs the entire shutdown sequence under `critical_section` so that
+    /// the cancelled transfer's completion IRQ can never fire. This is
```
diondokter, line 1426: If we're only worried about the transfer completion IRQ, could we maybe just mask that one instead of masking all interrupts? Should be a bit nicer for perf, especially with the spinloop present.

It says the spinloop is only microseconds, but let's say it's 10us running on 200mhz, that's 2000 clocks/instructions

embassy-mcxa/src/flexspi.rs
```diff
@@ -689,7 +683,6 @@ impl<'d, T: Instance> InnerFlexSpi<'d, T> {
         self.write_enable()?;
         self.issue_ip_command(address, self.flash.erase_sector_seq as usize, 0, None)?;
         self.wait_bus_busy()?;
-        self.software_reset();
```
diondokter, line 692: You haven't explained why this is removed, but if you're sure, it's fine

embassy-mcxa/src/flexspi.rs
```diff
@@ -1103,7 +1092,11 @@ impl<'d, T: Instance> InnerFlexSpi<'d, T> {
 
             let chunk_len = (8 * tx_watermark).min(data.len() - offset);
             for (index, chunk) in data[offset..offset + chunk_len].chunks(4).enumerate() {
-                let mut word = [0u8; 4];
+                // Pad the trailing partial word with 0xFF so that, for NOR
+                // flash program commands, any FIFO bytes the controller
+                // shifts out past `IDATSZ` are no-ops (a 0x00 pad would
+                // permanently zero the bits after the data).
+                let mut word = [0xFFu8; 4];
```
diondokter: Typically with flash a write is a write. Even if it's with 0xFF. So saying it's a no-op is probably incorrect. i.e. an embedded-storage norflash impl using this shouldn't claim to have a write size of <4 even if it works on the flash you tested with. That's partially a shortcoming of those traits

examples/mcxa5xx/src/bin/dma-stress.rs
diondokter, line 1: How long does this stress test take? If it's short (like max 2 secs) it could make sense to turn it into a test for mxca2xx so it's run on the HIL setup

embassy-mcxa/src/flexspi.rs
felipebalbi: I removed it because it felt wrong to reset the controller so frequently, then I tested to make sure there were no regressions after removing.

examples/mcxa5xx/src/bin/dma-stress.rs
felipebalbi: it's pretty quick: debug takes about 6 seconds, release takes about 1.14 seconds. I'll port it to mcxa2 and add both example and test versions.

embassy-mcxa/src/flexspi.rs
felipebalbi: removed the misleading chunk of the comment.

embassy-mcxa/src/dma.rs
felipebalbi: now using the critical section only to mask the interrupts and clear pending bit with NVIC.

felipebalbi:
bender run

PR #6016, opened 2026-05-02
Rework USBDescriptor.

### LINKED ISSUE #5988: USBDescriptor is used in non fixed-size descriptor
So, I was looking at how/where I should add descriptor->bytes conversion, maybe in `USBDescriptor`, for the OHCI root hub (software generated descriptors).

I noticed the "fixed-size" comment and I know it's used in the not-fixed-size descriptor `HubDescriptor`.
I also see that `ControlPipeExt` assumes the size is fixed.

two questions:
1) how about changing `HubDescriptor::SIZE` to `HubDescriptor::MAX_SIZE` and allow variable size descriptors?
2) is it ok to add a descriptor->buffer function directly to `USBDescriptor` or should it use a different trait?

flaviojs:
*(remade)* Take 2

Closes #5988

### DIFF
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -84,7 +84,9 @@ impl<'d, A: UsbHostAllocator<'d>, const MAX_PORTS: usize> HubHandler<'d, A, MAX_
             enum_info.split(),
         )?;
 
-        let desc = control_channel.request_descriptor::<HubDescriptor, 64>(0, true).await?;
+        let desc = control_channel
+            .request_descriptor::<HubDescriptor, { HubDescriptor::BUF_SIZE }>(0, true)
+            .await?;
 
         let mut hub = HubHandler {
             bus: bus.clone(),
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -5,6 +5,7 @@
 //! Requires the USB driver to support Interrupt IN pipes.
 
 use core::num::NonZeroU8;
+use core::ops::Deref;
 
 use bitflags::bitflags;
 use embassy_time::Timer;
@@ -62,7 +63,7 @@ impl<'d, A: UsbHostAllocator<'d>, const MAX_PORTS: usize> HubHandler<'d, A, MAX_
             .iter_interface()
             .find(|v| {
                 matches!(
-                    v,
+                    v.deref(),
                     InterfaceDescriptor {
                         interface_class: 0x09,
                         interface_subclass: 0x0,
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -380,7 +380,7 @@ struct HubDescriptor {
 }
 
 impl USBDescriptor for HubDescriptor {
-    const SIZE: usize = core::mem::size_of::<Self>();
+    const BUF_SIZE: usize = core::mem::size_of::<Self>();
     const DESC_TYPE: u8 = 0x29;
     type Error = ();
 
--- a/embassy-usb-host/src/class/hub.rs
+++ b/embassy-usb-host/src/class/hub.rs
@@ -12,10 +12,11 @@ use embassy_time::Timer;
 use embassy_usb::control::Request;
 use embassy_usb_driver::host::{HostError, SplitInfo, SplitSpeed, UsbHostAllocator, UsbPipe, pipe};
 use embassy_usb_driver::{Direction, EndpointInfo, EndpointType, Speed};
-use zerocopy::{FromBytes, Immutable, KnownLayout};
 
 use crate::control::{ControlPipeExt, ControlType, Recipient, RequestType, SetupPacket};
-use crate::descriptor::{DEFAULT_MAX_DESCRIPTOR_SIZE, InterfaceDescriptor, USBDescriptor};
+use crate::descriptor::{
+    DEFAULT_MAX_DESCRIPTOR_SIZE, DescriptorError, InterfaceDescriptor, USBDescriptor, VariableSizeDescriptor,
+};
 use crate::handler::{BusRoute, EnumerationInfo, HandlerEvent, RegisterError};
 use crate::{BusHandle, EnumerationError};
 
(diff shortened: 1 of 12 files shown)


embassy-usb-host/src/handler.rs
```diff
@@ -112,7 +112,7 @@ impl EnumerationInfo {
         let mut cfg_len = 0;
         for i in 0..self.device_desc.num_configurations {
             let cfg_desc_short = channel
-                .request_descriptor::<ConfigurationDescriptor, { ConfigurationDescriptor::SIZE }>(i, false)
+                .request_descriptor::<ConfigurationDescriptor, { ConfigurationDescriptor::MIN_SIZE }>(i, false)
```
bugadani: I think this shows the idea isn't a great one - if you only read the minimum possible size, you may be missing data. On the other hand, if the descriptors are generally unlimited in length, using MAX_SIZE would be equally incorrect, even if the device only returns as much data as the descriptor has - you would immediately trigger a stack overflow.

So my question is: what is the actual problem you are trying to solve, besides a bit of semantic unclarity around the current SIZE constant?

flaviojs: The original problem I am trying to solve is how to write the OHCI root hub descriptors to a buffer (without duplicating code and using descriptor structs).

The secondary problem that I'm trying to solve is USBDescriptor assuming descriptors are fixed-size while also being used for variable-size descriptors too.
In the original code SIZE is already misleading, it's actually MIN_SIZE for fixed-size descriptors and MAX_SIZE for variable-size descriptors.

For fixed-size descriptors MIN_SIZE is the minimum amount of data that is needed to parse the descriptor, so it's actually the correct buffer size and the rest of the data can be discarded.
For variable-size descriptors MAX_SIZE is the maximum supported size, so it's actually the correct buffer size.
RANGE expresses the sizes that are considered valid, regardless of the type of descriptor.

Would it help to add a convenience BUF_SIZE constant that would be MIN_SIZE in fixed-descriptors and MAX_SIZE in variable-size descriptors?

bugadani: > In the original code SIZE is already misleading, it's actually MIN_SIZE for fixed-size descriptors and MAX_SIZE for variable-size descriptors.
> For fixed-size descriptors MIN_SIZE is the minimum amount of data that is needed to parse the descriptor, so it's actually the correct buffer size and the rest of the data can be discarded.
> For variable-size descriptors MAX_SIZE is the maximum supported size, so it's actually the correct buffer size.
RANGE expresses the sizes that are considered valid, regardless of the type of descriptor.

To me this sounds like what we have is perfectly fine and anything more other than documenting this, is overcomplication.

flaviojs: Well, with only SIZE it is impossible to do the size checks (that rely on RANGE) that I'm doing in DescriptorError automatically... manual checks are error prone, it's very easy to make a mistake and hard to spot it.

Though, I can easily ditch MIN_SIZE and MAX_SIZE since they are convenience consts.
I guess you would want a convenience BUF_SIZE, or even name it the more misleading SIZE?

bugadani: > Well, with only SIZE it is impossible to do the size checks (that rely on RANGE) that I'm doing in DescriptorError automatically... 

Do you consider that a good change in the first place? A range is annoying enough to work with, as your code shows. MIN/MAX constants are okay to work with, but they are only necessary in a minority of cases, and even then I think they have limited utility. In general you aren't interested in the "there is more data than I need" case - what does it matter that you have trailing bytes in a buffer? Speaking of misleading, what does it mean that a fixed size descriptor's MAX_SIZE is a different value from it's MIN_SIZE?

An error type having any logic is already quite unorthodox, and you're, in my opinion, trying to generalize something that probably doesn't need to be generalized into a single concept. The end result is difficult to read, and probably equally difficult to work with. What does the `check` fuction do? If you look at `Self::check(buf_len, len, desc_type, T::RANGE, &[T::DESC_TYPE, subtype])`, can you tell two weeks from now what this is supposed to mean?

A descriptor knows if it can be deserialized from, or serialized to a number of bytes, it doesn't need any associated constants at all for that. SIZE carries the maximum amount of bytes a descriptor might need, so that the user can allocate a buffer that is big enough - in this function, it is a MAX_SIZE. The USB device itself will tell you if a descriptor you're trying to read is shorted than this size.

You can think the current code misleading and it's a fair conclusion. But what you're trying to turn it into is overcomplicated in its current form. Perhaps because you're trying to do too much in a single PR - changes like preferring DescriptorError over `()` are obviously good, and serializing a descriptor into a buffer is something you need while being independent of the questionable parts. If you send those in separately, then those changes won't show up here as part of this PR, so maybe the shape of this one can be more clear.

flaviojs: > Speaking of misleading, what does it mean that a fixed size descriptor's MAX_SIZE is a different value from it's MIN_SIZE?

As stated in the docs for RANGE, it means the fixed-size descriptor allows extended bytes to be present (like the 2 extra bytes in `test_parse_extended_endpoint_descriptor`).

> What does the check fuction do? If you look at Self::check(buf_len, len, desc_type, T::RANGE, &[T::DESC_TYPE, subtype]), can you tell two weeks from now what this is supposed to mean?

I'm pretty sure I will know exactly what it does just by reading the function comment and the argument names of `check`. (I'm not very good at wording/naming things so others might not)
*(note)* I only added those functions to DeviceError because I couldn't add them to the trait (maybe I should have make a separate struct for them, dunno)

> If you send those in separately, then those changes won't show up here as part of this PR, so maybe the shape of this one can be more clear.

Ok, so I should only add `write_to` without the useful automatic checks in a separate PR.
Should I leave the PR open (as a draft) and try to rework this PR after the other PR is merged or just close this PR?

leftger: I believe @bugadani is correct in his assessment. Personally I would try to implement the smaller changes into separate PRs and if your endgame objective can be brought into this PR then I think that would serve as a good middle ground for implementation.

flaviojs:
I tried just adding `write_to` and hated the result, this definitely needs to go first, so...

Take 2 - remade this without `write_to`, this time using my usual workflow (individual reviewable commits, while rebasing and reediting commits as needed)
I hope what I wanted is clearer now, what do you think? @bugadani @leftger

Note: rebased/reedited commits tend to show up as confusing history in github, so it's better to squash-merge this

Sidenote: I would prefer a separate `embassy-usb-descriptor` crate that gathered all known descriptors, and could be used to implement usb devices, usb hosts, and other programs (like a usb trafic/log parser), but I understand not wanting to deal with all the dependency coordination

leftger:
@flaviojs I agree with your sentiment behind `embassy-usb-descriptor`. Makes more sense from an organizational standpoint if the descriptors are agnostic to the actual IP stack. I think it'd be a good idea to bring up in the Matrix chat room.

embassy-usb-host/src/lib.rs
```diff
@@ -3,6 +3,19 @@
...
+        let mut max = $first;
+        $(
+            if max < $next {
+                max = $next;
+            }
+        )*
+        max
+    }};
+}
+
```
leftger, line 18: I'm a bit iffy on macros, but it seems useful for the use-case. Is there no other utility that could do this? If not then I think it's fine to include it for now.

leftger (commented):
@bugadani I think Flavio did most of what we asked. Just pinging you in case you have anything else to comment on. Seems good to merge from my POV.

embassy-usb-host/src/lib.rs
flaviojs: I'm not aware of anything that could be used.

bugadani:
```rust
pub const fn const_max<const N: usize>(n: [usize; N]) -> usize {
    let mut max = 0;
    let mut i = 0;
    while i < n.len() {
        if n[i] > max {
            max = n[i];
        }
        i += 1;
    }

    max
}
```

[godbolt](https://rust.godbolt.org/z/oPxca87Ko)

Whether it's better than the macro or not, I can't tell, and I don't mind the macro anyway.

leftger:
@flaviojs do you wanna clean up the commit history? I promised myself I'd start enforcing squash commit PRs. That's all that I'd ask before merging

bugadani:
I'm not awake enough to form a proper opinion just yet, but this feels better :)

embassy-usb-host/src/class/hub.rs
```diff
@@ -363,33 +367,93 @@ impl HubInterrupt<'_> {
...
+    pub device_removable: [u8; 16],
+    /// Port power control mask bitmap (variable size).
+    ///
+    /// All bits should be 1 for compatibility with USB 1.0 code.
+    pub port_power_ctrl_mask: [u8; 16],
+}
+
+impl HubDescriptor {
+    /// Maximum number of ports that we support (at most 255).
+    pub const SUPPORTED_PORTS: u8 = (16 * u8::BITS as usize - 1) as u8;
```
bugadani: Isn't this 127? Why have code this complex if the end result is different than the documented value? And what's the purpose of the const assert a few lines below?

leftger:
Ok and maybe address the change @bugadani preferred

bugadani:
TBH I'm actually finding the current embassy-usb/embassy-usb-host split a bit unnecessary, so perhaps instead of adding one more crate to the mix that has the common parts, we can try and merge them.

embassy-usb-host/src/class/hub.rs
flaviojs: The important part of that code is 16, it's supposed to be the length of `HubDescriptor::device_removable`, which I failed to figure out how to get in const context.

The assert a bit below is for bit 0, which is reserved and should always exist.
So at least 1 byte is needed, which means at least 7 ports are supported in the descriptor.
I believe compile-time checks to be an advantage of rust, the assert is for that.

--

I think I figured out how to get the size of `HubDescriptor::device_removable` in const context, so I'll change `SUPPORTED_PORTS` and merge the assert with it.

flaviojs:
> do you wanna clean up the commit history? I promised myself I'd start enforcing squash commit PRs. That's all that I'd ask before merging

Dunno how the history ends up in a squash commit PRs, but I think the history is fine like this.

embassy-usb-host/src/class/hub.rs
```diff
@@ -394,16 +394,20 @@ pub struct HubDescriptor {
...
 
 impl HubDescriptor {
-    /// Maximum number of ports that we support (at most 255).
-    pub const SUPPORTED_PORTS: u8 = (16 * u8::BITS as usize - 1) as u8;
+    /// Maximum number of ports that this descriptor supports (at most 255).
+    pub const SUPPORTED_PORTS: u8 = {
+        const ZEROED: HubDescriptor = unsafe { core::mem::zeroed() };
+        // bit 0 is reserved and should always exist, so 1 byte is required
+        const _: () = assert!(ZEROED.device_removable.len() > 0);
+        (ZEROED.device_removable.len() * u8::BITS as usize - 1) as u8
```
bugadani: wtf

embassy-usb-host/src/class/hub.rs
```diff
@@ -363,33 +367,97 @@ impl HubInterrupt<'_> {
...
+}
+
+impl HubDescriptor {
+    /// Maximum number of ports that this descriptor supports (at most 255).
+    pub const SUPPORTED_PORTS: u8 = {
+        const ZEROED: HubDescriptor = unsafe { core::mem::zeroed() };
+        // bit 0 is reserved and should always exist, so 1 byte is required
+        const _: () = assert!(ZEROED.device_removable.len() > 0);
+        (ZEROED.device_removable.len() * u8::BITS as usize - 1) as u8
+    };
```
bugadani:
```suggestion
    /// Maximum number of ports that this descriptor supports.
    pub const SUPPORTED_PORTS: u8 = 127;
```

Is anything more necessary? Or, if you really want the spec-maximum of 255, adjust the arrays as well

flaviojs: I guess this is a difference in style?

I prefer to have one source of truth instead of having multiple places that have to be kept in sync, in this case the source of truth is the size of `device_removable` (the amount of ports you can fit in there).
I also prefer to enforce spec restrictions at compile time, in this case for bit 0, which is the assert.

Also note that in a future PR I will increase the supported ports to 255, because that is what OHCI allows, even if only 127 addresses are available (you could easily have address 10 used by a device in port 200).

Maintainers can edit the PR, feel free to change this as you see fit.

*(edit)* to clarify, the purpose of SUPPORTED_PORTS is to serve as the source of truth for the byte reader, and also give cognitive context (the name) for the descriptor user

flaviojs: > Or, if you really want the spec-maximum of 255, adjust the arrays as well

Noticed this addition after refreshing, I will change to the spec maximum so SUPPORTED_PORTS does not need to exist.

flaviojs:
@leftger done

leftger (approved):

PR #6023, opened 2026-05-04
don't track settings.json

xoviat:
this stops git from tracking this file, allowing changes to it without showing it as modified.


diondokter:
Maybe don't rename the file, but do add it to the ignore?
Then the file will just be there as it is right now and you'll have a working setup by default. But any changes to it will be ignored unless forced.

xoviat:
that does nothing

PR #6031, opened 2026-05-05
refactor: improve timer handling

Tavish-Srivastava:
Verified on hardware

### DIFF
--- a/embassy-stm32-wpan/src/wba/runner.rs
+++ b/embassy-stm32-wpan/src/wba/runner.rs
@@ -101,11 +101,14 @@ pub async fn ble_runner() -> ! {
 
     loop {
         // Wait for either a sequencer event or a timer expiry
-        select(
-            util_seq::wait_for_event(),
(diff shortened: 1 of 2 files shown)


xoviat (commented):
I don't think I should merge this PR because it's special-casing the unset timer.  Specifically, see gp16.rs L260 for comparison.

embassy-stm32-wpan/src/wba/runner.rs
```diff
@@ -101,11 +101,14 @@ pub async fn ble_runner() -> ! {
...
-        )
-        .await;
+        match linklayer_plat::earliest_timer_deadline() {
+            Some(deadline) => {
+                select(util_seq::wait_for_event(), Timer::at(deadline)).await;
+            }
+            None => {
+                util_seq::wait_for_event().await;
+            }
+        }
```
xoviat, line 111: There's also no need for this. A timer set to u64::max will never expire, and does not require the branch.

PR #6037, opened 2026-05-05
wpan: cleanup controller

xoviat:
Tested on local hw, and disconnection (0x22) still occurs. However, no other observed regressions.

### DIFF
--- a/embassy-stm32-wpan/src/wba/controller.rs
+++ b/embassy-stm32-wpan/src/wba/controller.rs
@@ -28,13 +28,13 @@ use stm32wb_hci::vendor::CommandHeader;
 use stm32wb_hci::{Event, event};
 
 use crate::bluetooth::error::BleError;
-use crate::host_if::{MAX_BLE_PKT_SIZE, TASK_BLE_HOST_MASK, TASK_LINK_LAYER_MASK, TASK_PRIO_BLE_HOST};
-use crate::linklayer_plat::{
+use crate::runner::BLE_INIT;
+use crate::wba::host_if::{MAX_BLE_PKT_SIZE, TASK_BLE_HOST_MASK, TASK_LINK_LAYER_MASK, TASK_PRIO_BLE_HOST};
+use crate::wba::linklayer_plat::{
     EVENT_CHANNEL, HARDWARE_AES, HARDWARE_PKA, HARDWARE_RNG, run_radio_high_isr, run_radio_sw_low_isr,
 };
-use crate::runner::BLE_INIT;
-use crate::util_seq;
 use crate::wba::ll_sys::init_ble_stack;
+use crate::wba::util_seq;
 
 /// High interrupt handler.
 pub struct HighInterruptHandler {}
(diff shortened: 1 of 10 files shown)


embassy-stm32-wpan/src/wba/ll_sys/ll_sys_startup.rs
```diff
@@ -108,6 +108,7 @@ pub unsafe extern "C" fn ll_sys_ble_cntrl_init(host_callback: hst_cbk) {
 #[cfg(feature = "wba-ble")]
 /// Complete the BLE link layer initialization
 /// This should be called after the sequencer is running
+#[allow(dead_code)]
 pub fn complete_ble_link_layer_init() {
```
xoviat: Looks like this method is never called. Should it be called?

leftger:
I'll try this out in a couple of hours

embassy-stm32-wpan/src/wba/ll_sys/ll_sys_startup.rs
leftger: It's already covered by ll_sys_ble_cntrl_init() and ll_sys_dependencies_init(). The ST HAL has no equivalent either. You can basically remove it and that should be it.

embassy-stm32-wpan/src/wba/controller.rs
```diff
@@ -327,29 +318,31 @@ impl bt_hci::controller::Controller for AtomicController {
...
             let Poll::Ready(slot) = controller.receiver.poll_receive(cx) else {
                 return Poll::Pending;
             };
 
-            // Must copy to extend the lifetime
-            buf[..*&slot.len()].copy_from_slice(&slot);
+            self.pending_evt.store(true, Ordering::Release);
 
-            Poll::Ready(*&slot.len())
+            Poll::Ready(unsafe { core::slice::from_raw_parts(&slot.0 as *const _ as *const u8, slot.1) })
```
leftger, line 341: This works in the common case but has a soundness corner:                                                                                                                                                                          
  - The returned ControllerToHostPacket<'a> claims to borrow from _buf, but it actually borrows from a static channel slot. The lifetime is laundered.
  - If a caller ever holds two outstanding packets (calls read() again before dropping the previous result), the second call advances the receiver and receive_done()s the first slot — the sender can then overwrite it under the still-live first packet. UB.
  - Normal bt_hci callers do read→parse→drop and never trip this, but the unsafe block deserves a SAFETY comment spelling out the invariant.

xoviat: I looked at the event loop of trouble to verify this before I made this optimization. wb55 has the same optimization and works fine. as for other callers, it's probably going to be just us in wpan? `ControllerToHostPacket` is is pre-parsed and only parsed events will be returned to the user.

xoviat: See https://github.com/embassy-rs/trouble/blob/c103e97ff7d333c2cb76b643477e3c4141d0ff74/host/src/host.rs#L1142.

leftger: OK thanks for the clarification 😄. Then the code is fine then.

leftger (approved):

PR #6039, opened 2026-05-05
[MCXA] Normalize RAW ADC

SanjeevKumm:
Normalize the raw ADC result based on the configured resolution. In 12-bit mode, the LPADC stores results left-justified in the 16-bit RESFIFO D field (bits [15:3]), so right-shift by 3 to produce the actual 12-bit value (0–4095). In 16-bit mode, the full 16-bit range is used and no shift is needed.

### DIFF
--- a/embassy-mcxa/src/adc.rs
+++ b/embassy-mcxa/src/adc.rs
@@ -389,7 +389,7 @@ pub struct Conversion {
     pub loop_channel_index: u8,
     /// The trigger that triggered the command to run
     pub trigger_id_source: u8,
-    /// The raw value from the ADC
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/adc.rs
felipebalbi, line 577: according to the documentation for `Conversion`, we return the raw value. This PR is changing that statement. We can certainly change, but the documentation has to be updated as well.

felipebalbi (changes requested):

felipebalbi (approved):
LGTM!

Let's get tests to run here to make sure there are no regressions.

felipebalbi:
bender run

SanjeevKumm:
bender run

felipebalbi:
bender run

PR #6044, opened 2026-05-06
wip: add volatile ptr type

xoviat:
mostly copied from the volatile crate

### DIFF
--- a/embassy-hal-internal/src/volatile_ptr/slice.rs
+++ b/embassy-hal-internal/src/volatile_ptr/slice.rs
@@ -50,16 +50,13 @@ impl<'a, T, A> VolatilePtr<'a, [T], A> {
     /// let subslice = volatile.index(1..);
     /// assert_eq!(subslice.index(0).read(), 2);
     /// ```
-    pub fn index<I>(self, index: I) -> VolatilePtr<'a, <I as SliceIndex<[T]>>::Output, A>
+    pub fn index(self, index: usize) -> VolatilePtr<'a, <usize as SliceIndex<[T]>>::Output, A>
     where
-        I: SliceIndex<[T]> + SliceIndex<[()]> + Clone,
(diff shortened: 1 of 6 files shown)


Dirbaio:
why is this needed?

If it's for the backup memory we can do it by providing functions to read/write bytes or slices that use regular raw pointers internally with read_volatile/write_volatile. There's no need to actually return a volatile pointer to the user. It keeps the API simpler, there's less concepts.

xoviat:
> It keeps the API simpler, there's less concepts.

I would disagree with this, because the volatile pointer is essentially a rust slice with restricted compiler operations. In fact it would be much nicer to restrict the compiler operations for this memory, but I can't do that.

embassy-hal-internal/src/volatile_ptr/slice.rs
```diff
@@ -153,17 +150,34 @@ impl<'a, T, A> VolatilePtr<'a, [T], A> {
...
         T: Copy,
         A: Writable,
     {
         let len = self.pointer.len();
         assert_eq!(len, src.len(), "destination and source slices have different lengths");
 
-        // intrinsics::volatile_copy_nonoverlapping_memory(self.pointer.as_mut_ptr(), src.as_ptr(), len);
+        unsafe {
+            let src = core::slice::from_raw_parts(src as *const _ as *const u8, size_of_val(src));
+            let dst = self.pointer.as_mut() as *mut _ as *mut u8;
```
usbalbin, line 163: Correct me if I am wrong, but in cases where dst is the one we need to watch out the alignment for. If dst has the correct alignment but prefix is non-empty then that will mess up the alignment once we start with the middle part.

xoviat: You are right about that. This code is probably wrong.

PR #6055, opened 2026-05-07
stm32/sdmmc: DLYB-tuned SDR50/SDR104, optional CMD23/ACMD23 fast path

tarfu:
- New `Sdmmc::new_4bit_with_vswitch_dlyb` constructor (N6 SDMMC2 has no CKIN pin → needs DLYB tap tuning for SDR50+).
- DLYB driver in `sdmmc/dlyb.rs` (gated on `cfg(dlybsd)`), DLL-locked tap sweep with CMD13 probe.
- `switch_function` now negotiates SDR104 when `freq > 100 MHz` and DLYB is owned; SDR50 otherwise.
- `Config::use_cmd23` / `use_acmd23`: opt-in fast path. CMD23 is silently AND-ed with `SCR.CMD_SUPPORT[1]`; `StorageDevice::supports_cmd23()` lets users probe.`


embassy-stm32/src/sdmmc/dlyb.rs
```diff
@@ -0,0 +1,74 @@
...
+const POLL_LIMIT: u32 = 100_000;
+
+#[allow(private_bounds)]
+pub trait DlybInstance<T: Instance>: SealedDlybInstance<T> + PeripheralType + 'static {}
+pub(crate) trait SealedDlybInstance<T: Instance> {
+    fn regs() -> crate::pac::dlybsd::Dlybsd;
+    fn release_dll_reset();
+}
+
+use crate::peripherals::{DLYB_SDMMC1, DLYB_SDMMC2, SDMMC1, SDMMC2};
```
xoviat, line 17: @tarfu Looks like this broke the docs build. Perhaps foreach_interrupt or foreach_peripheral should be used?

tarfu: I'll take a look in the evening.

tarfu: @xoviat I totally forgot to push my change -.- Sorry for the inconvenience and the work you needed to do yourself.

PR #6057, opened 2026-05-07
[MCXA] fix OSC32K enum variant casing to match updated nxp-pac

SanjeevKumm:
fix OSC32K enum variant casing to match updated nxp-pac


felipebalbi (approved):

felipebalbi:
bender run

PR #6060, opened 2026-05-08
[MCXA] I2C controller fixes and loopback test suite

felipebalbi:
Three real bug fixes for the embassy-mcxa LPI2C controller driver and a
comprehensive in-chip loopback test example used to find them. Validated on
MCXA577 at DEFMT_LOG=off and DEFMT_LOG=trace.

Driver fixes

1. NACK recovery race — Two coupled bugs in the async NACK recovery path:

 - async_read_internal armed an OnDrop(remediation) before async_start. On
   AddressNack, async_start's status_and_act already issued the STOP, then ?
   propagated Err and the OnDrop ran remediation() again. Two STOPs in quick
   succession corrupted controller state for the next transaction. Fixed by
   moving the OnDrop inside the chunk loop, after async_start.await?, with a
   per-chunk defuse() before the next chunk. Applied to both interrupt-mode and
   DMA-mode read paths.

 - remediation() issued STOP via stop(), which busy-waits on
   is_tx_fifo_empty_or_error. After an address NACK with empty TX FIFO and
   autostop disabled, the master is idle when STOP is queued, raising FEF; the
   wait loop exits early on that error, leaving the STOP cmd stuck in the TX
   FIFO and FEF set in MSR. The next transaction then started with a stale STOP
   in flight and reported FifoError/ArbitrationLoss on the next START. Fixed by
   resetting the FIFOs after the STOP attempt and clearing MSR flags.

2. FastPlus MCCR0 timing — The hard-coded FastPlus values (CLKLO=0x04,
CLKHI=0x03, SETHOLD=0x03, DATAVD=0x02) yielded an SCL period of only 9
functional clocks, putting the bus around 1.33 MHz on a typical 12 MHz
functional clock — above the 1 MHz Fast Mode+ spec and unreliable on
MCXA577. Loosened to (0x06, 0x05, 0x05, 0x03) for 13 functional clocks per
period (920 kHz on 12 MHz), well within spec.

3. Compute MCCR0/PRESCALE from the actual peripheral clock — The bus-timing
table implicitly assumed a 12 MHz LPI2C input clock. Switching the peripheral
clock source, or running on a board with a different default, would silently
produce out-of-spec or completely wrong SCL rates.  Replaced the table with
compute_mccr0_params(src_hz, target_hz), ported from NXP's fsl_lpi2c.c
LPI2C_MasterSetBaudRate: for each prescaler 0..=7, pick the period that
minimizes |error| to target_hz, then derive CLKHI/CLKLO with a tBUF clamp and
SETHOLD/DATAVD as ~½ and ~¼ of the SCL period. The peripheral input frequency
from enable_and_reset() is now stored on the I2c struct so SetConfig::set_config
can recompute on runtime speed changes.

New example: i2c-loopback-async

In-chip controller↔target loopback for validating the I2C controller
driver. Target task on LPI2C3 (P3_21/P3_20) mirrors i2c-target-async; controller
on LPI2C0 (P0_21/P0_20). Wire jumpers P0_20 <-> P3_20 (SDA), P0_21 <-> P3_21
(SCL) with pull-ups.

Test cases (added incrementally across commits):

 - t_basic_rw — 100 iters of W/R 2 bytes, expect 0x55.

 - t_lengths — write/read/write_read for L ∈ {1, 2, 4, 8, 16, 32}.

 - t_burst — 500 iters of W2 / R2 / WR1+R2 mix.

 - t_edges — bad-address NACK, post-NACK recovery, zero-length read, recovery
   after a NACKed READ (exercises the read-side OnDrop fix).

 - t_speed_sweep — 50 iters of W2/R2/WR1+R2 at Standard, Fast, and FastPlus;
   exercises the SetConfig path.

 - t_soak — 2000 randomized op iterations (W/R/WR mix, lengths 1..8) using a
   deterministic xorshift PRNG.

### DIFF
--- a/embassy-mcxa/src/i2c/controller.rs
+++ b/embassy-mcxa/src/i2c/controller.rs
@@ -74,7 +74,7 @@ use crate::dma::{Channel, DMA_MAX_TRANSFER_SIZE, DmaChannel, TransferOptions};
 use crate::gpio::{AnyPin, SealedPin};
 use crate::interrupt;
 use crate::interrupt::typelevel::Interrupt;
-use crate::pac::lpi2c::{Alf, Cmd, Dmf, Dozen, Epf, McrRrf, McrRtf, Msr, MsrFef, MsrSdf, Ndf, Pltf, Stf};
+use crate::pac::lpi2c::{Alf, Cmd, Dmf, Dozen, Epf, McrRrf, McrRtf, Msr, MsrFef, MsrSdf, Ndf, Pltf, Prescale, Stf};
 
 /// Errors exclusive to HW initialization
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


felipebalbi:
There's still a DMA-only bug on FastPlus. That's pre-existing on today's HEAD, not introduced by this PR. Still debugging that one.

PR #6061, opened 2026-05-09
Add WritableDescriptor.

flaviojs:
Add WritableDescriptor.
Implement WritableDescriptor for complete descriptors.
(not implemented for partial descriptors DeviceDescriptorPartial, AudioControlHeaderDescriptor, and UnitDescriptor)

Implement TerminalType<->u16 conversion with the trait From.

Fix length check in HubDescriptor::match_bytes_len.
Fix length of DeviceDescriptorPartial.


flaviojs:
If you prefer, I can implement the other descriptors now instead of just the ones I need.

leftger:
Yeah that's fine you can add the other descriptors if you'd like

flaviojs:
Done, updated the PR description accordingly.

flaviojs:
So, linking the length of DeviceDescriptorPartial with the length of DeviceDescriptor in #6016 was a bad idea.
It stopped being able to read the partial descriptor.

I'm including the fix here because the test uses the `write_to_bytes` capability that this PR adds.

PR #6063, opened 2026-05-10
rp/dma: Make DMA channel number public

mvirkkunen:
Since the registers are public, it would make sense to also make the channel number public. It's needed to e.g. configure the `CHAIN_TO` channel register and to use the channel abort register.

### DIFF
--- a/embassy-rp/src/dma.rs
+++ b/embassy-rp/src/dma.rs
@@ -65,7 +65,7 @@ impl<'d> Channel<'d> {
     }
 
     /// Get the channel number.
-    fn number(&self) -> u8 {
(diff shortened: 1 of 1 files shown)


Dirbaio:
There's not much point into tying it with unstable-pac, We can make it unconditionally public.

mvirkkunen:
All right, I thought it was hidden for a reason! I'll update the PR.

Dirbaio:
`unstable-pac`'s purpose is to allow doing a major bump of the PAC in a minor bump of embassy-rp. So it gates everything returning PAC registers. The channel number is just a number.

PR #6064, opened 2026-05-10
MPS3-AN536 example

thejpster:
Adds an example for the MPS3-AN536 board emulated by QEMU.

A future PR could pull out the Generic Timer implementation, but you also need to turn on the timer’s interrupt in the interrupt controller and that varies depending on which interrupt controller you have.


thejpster:
Note on physical timers vs virtual timers. Physical timers count wall time. Virtual timers only count time while that guest OS is scheduled by a hypervisor - time doesn’t tick whilst other VMs are running. Without a hypervisor running and adjusting a special offset value they will be the same.

Dirbaio:
can you add armv8r-none-eabihf to rust-toolchain.toml to make ci happy?

Dirbaio (approved):
very cool! thanks

PR #6071, opened 2026-05-11
Support SMP on Armv8-R in MPS3-AN536 example.

jonathanpallant:
Sets up two thread-mode executors, with their own timer queues and their own timers.

Requires an [unreleased copy of qemu-run](https://github.com/knurling-rs/defmt/pull/1055) with extra arg support.

I don't love the core1 assembly code in the `lib.rs` file - I am looking at moving that into `aarch32-rt`, because it's mostly the same for every chip. The only thing that's unique is the specific memory address you spin on - it cannot be in RAM because RAM is not initialised on start-up, and anything else is memory-map specific.


jonathanpallant:
qemu-run 0.3 is out, which supports the `--arg smp=2` option we need.

lulf (approved):
Probably not plenty of reviewers for this PR. I think it looks OK, and low risk, so merging.

PR #6076, opened 2026-05-12
wpan/wba: fill rng bytes async and refactor into platform mod

xoviat:
- make rng generation async so that it doesn't block interrupts.
- refactor to new platform struct that reduces boilerplate on init

confirmed on hw: no regressions on mainstack and trouble still works.


leftger (approved):
Ran it on my HW and it looks good!

PR #6080, opened 2026-05-12
Initial Silicon Labs EFR32 support

andresv:
Adds `embassy-silabs`, a new HAL crate targeting Silicon Labs EFR32 / EFM32 series microcontrollers. Currently tested on EFR32MG26 with blinky and logging.

## What's included

* GPIO (Output, Input) with embedded-hal 1.0 impls.
* `embassy-time` driver backed by TIMER0, TIMER1 at 1 MHz (32-bit counter, mirroring embassy-stm32's gp16.rs).
* Clock bring-up (rcc)
* Blinky example

## PAC generation pipeline
Mirrors the stm32-data / stm32-data-generated split used by `embassy-stm32`. Three companion repos:

[andresv/silabs-data-source](https://github.com/andresv/silabs-data-source)
[andresv/silabs-data](https://github.com/andresv/silabs-data)
[andresv/silabs-data-generated](https://github.com/andresv/silabs-data-generated)

# Scope
This is a starting point and currently only GPIO + time driver is tested.

I thought I open PR early one, maybe somebody else is also interested but never wanted to bring up whole PAC generation.


embassy-silabs/src/time_driver.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+pub(crate) fn init(cs: CriticalSection) {
+    DRIVER.init(cs);
+}
+
+#[cfg(feature = "rt")]
+#[unsafe(no_mangle)]
+unsafe extern "C" fn TIMER0() {
```
i509VCB: I would think you want to use `#[interrupt]` from the PAC over a no_mangle function attribute?

andresv: Fixed

lulf (commented):
No objections on a new HAL from me. However, just to make sure: do you solemnly swear to maintain and ensure PRs are reviewed for this HAL?

andresv:
Yes, it will be used in production. In nasty mixed environment where rust code is mixed with silabs sdk c code.

PR #6081, opened 2026-05-12
[MCXA] Flexspi refactor & improvements

diondokter:
Depends on: https://github.com/embassy-rs/nxp-pac/pull/104

I wanted to look into XIP. Got infuriated by the flexspi driver, so I made some improvements.
Best reviewed commit by commit.

- Added more type info to the pins and the constructors always take the exact right pins now
- Removed instance generic
- Added Async/Blocking modes to properly split out the async mechanism

I can't test the chipselect index stuff on my board since it's hard soldered on.
The datasheet is a bit vague on how exactly it works. I'm 70% sure it works for chips not connected to A_SS0_B.
But to let people know they might be using an untested feature, it prints a defmt warning to be sure.


embassy-mcxa/src/flexspi.rs
```diff
@@ -349,7 +349,17 @@ macro_rules! impl_flexspi_instance {
...
 }
 
-pub trait Pin<T: Instance>: GpioPin + sealed::Sealed + PeripheralType {
+/// Flexspi port A
+pub struct A;
+/// Flexspi port B
+pub struct B;
+
+/// Marker trait indicating a type is a port
+pub trait Port {}
```
felipebalbi: Do you need a `SealedPort` or something similar to prevent applications from creating their own ports?

felipebalbi (approved):

PR #6082, opened 2026-05-12
stm32: add try_until to hal_internal and yield on timeout

### LINKED ISSUE #3540: STM32 Sdmmc blocks on write_block
The function `write_block` in the sdmmc module currently blocks while waiting for the SD card after the block has been written. This can be several milliseconds depending of the SD card.
The relevant section of code is:
https://github.com/embassy-rs/embassy/blob/6c4b3d82b637fce5ab6efdc312d7852381d8ddeb/embassy-stm32/src/sdmmc/mod.rs#L1310-L1319
I added a yield and this seems to work well as a quick fix, but I'll admit I'm well out of my depth here to know if there may be undesirable side effects. I guess ideally one would await that entire section and wake on D0 going high - but I don't know if that would be feasible.
```rust
use embassy_futures::yield_now();
...
// Try to read card status (ACMD13)
while timeout > 0 {
    match self.read_sd_status().await {
        Ok(_) => return Ok(()),
        Err(Error::Timeout) => (yield_now().await), // Try again
        Err(e) => return Err(e),
    }
    timeout -= 1;
}
Err(Error::SoftwareTimeout)
```

Attached is the output of a logic analyser (1ms/div) showing the effect of the change (A different type of SD card was used for the two tests which seem to have different wait times).
SD_Card_Blocking
SD_Card_NotBlocking

xoviat:
closes #3540 
closes #6078

### DIFF
--- a/embassy-stm32/src/lib.rs
+++ b/embassy-stm32/src/lib.rs
@@ -18,6 +18,7 @@ include!(concat!(env!("OUT_DIR"), "/_macros.rs"));
 // Utilities
 mod macros;
 pub mod time;
+mod try_until;
 /// Operating modes for peripherals.
 pub mod mode {
     trait SealedMode {}
@@ -871,7 +872,6 @@ fn init_hw(config: Config) -> Peripherals {
 }
 
 /// Performs a busy-wait delay for a specified number of microseconds.
-#[allow(unused)]
 pub(crate) fn block_for_us(us: u64) {
     cortex_m::asm::delay(unsafe { rcc::get_freqs().sys.to_hertz().unwrap().0 as u64 * us / 1_000_000 } as u32);
 }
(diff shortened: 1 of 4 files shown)


embassy-hal-internal/src/try_until.rs
```diff
@@ -0,0 +1,46 @@
+//! Provides a function to try until something is true
+
+#[cfg(all(not(feature = "time"), feature = "block-for-us"))]
+unsafe extern "Rust" {
+    fn __block_for_us(micros: u64);
```
Dirbaio: this is a bit hacky. i'd suggest keeping these utilities in embassy-stm32 so there's no dep problems that require extern to workaround.

xoviat: I was going to use this in the cyw crate as well.

Dirbaio: cyw43 requires embassy-time so the blocking fallback is not useful there. I'd just add the `try_until` function to cyw43 even if it's a bit of duplication.

It's very little duplication, deduplicating it is not worth it adding a dep on hal-internal *and* a Cargo feature to hal-internal *and* extern.

PR #6091, opened 2026-05-13
Implement is_bootsel_pressed for the rp2350.

### LINKED ISSUE #4769: Add BOOTSEL button poll to Pico 2 / rp235x
Currently the embassy-rp crate provides a function to poll the bootsel button press for the Pico 1 but not the Pico 2. If doing that is technically feasible, it would be a nice feature for early prototyping.

nathdobson:
This PR includes an implementation of is_bootsel_pressed for the rp2350 and examples to test this functionality. This feature should cover Pico 2 devices and similar boards using the rp2350 with a bootsel button. I've tested this change on a Pico W board and a Pico 2 W board.

The only real change between the rp2040 and rp2350 is the pin number and the value of the Oeover::Disable mask. I've tried to move higher-level logic from read_cs_status to is_bootsel_pressed to avoid the risk of function calls in read_cs_status.

Resolves #4769


examples/rp235x/Cargo.toml
```diff
@@ -11,7 +11,7 @@ embassy-embedded-hal = { version = "0.6.0", path = "../../embassy-embedded-hal",
 embassy-sync = { version = "0.8.0", path = "../../embassy-sync", features = ["defmt"] }
 embassy-executor = { version = "0.10.0", path = "../../embassy-executor", features = ["platform-cortex-m", "executor-thread", "executor-interrupt", "defmt"] }
 embassy-time = { version = "0.5.1", path = "../../embassy-time", features = ["defmt", "defmt-timestamp-uptime"] }
-embassy-rp = { version = "0.10.0", path = "../../embassy-rp", features = ["defmt", "unstable-pac", "time-driver", "critical-section-impl", "rp235xa", "binary-info"] }
+embassy-rp = { version = "0.10.0", path = "../../embassy-rp", features = ["unstable-pac", "time-driver", "critical-section-impl", "rp235xa", "binary-info"] }
```
Dirbaio: unintentional change?

Dirbaio:
Can you extend this test to cover rp235x? https://github.com/embassy-rs/embassy/blob/main/tests/rp/src/bin/bootsel.rs

examples/rp235x/Cargo.toml
nathdobson: Fixed.

nathdobson:
I've added the test (by removing the restriction to the rp2040).

lulf (approved):

lulf:
@nathdobson The bootsel HIL test is not working on rp2350

nathdobson:
I think the teleprobe_meta::target change should fix the issue. It looks like both versions of the test had been sent to the pico 1:
```
 INFO  teleprobe::client > === rpi-pico out/tests/rpi-pico/bootsel: OK
 ERROR teleprobe::client > === rpi-pico out/tests/pimoroni-pico-plus-2/bootsel: FAILED: ...
```

PR #6094, opened 2026-05-13
wpan: cleanup and remove timers on llo

xoviat:
confirmed on hw; no observed regressions on wba6

### DIFF
--- a/embassy-stm32/src/rcc/wba.rs
+++ b/embassy-stm32/src/rcc/wba.rs
@@ -21,6 +21,7 @@ pub const HSE_FREQ: Hertz = Hertz(32_000_000);
 #[derive(Clone, Copy, Eq, PartialEq)]
 pub struct Hse {
     pub prescaler: HsePrescaler,
+    pub trim: Option<u8>,
 }
 
 #[derive(Clone, Copy)]
@@ -190,6 +191,10 @@ pub(crate) unsafe fn init(config: Config) {
             w.set_hsepre(hse.prescaler);
         });
         while !RCC.cr().read().hserdy() {}
+
+        hse.trim.map(|trim| {
+            RCC.ecscr1().modify(|w| w.set_hsetrim(trim));
+        });
 
         HSE_FREQ
     });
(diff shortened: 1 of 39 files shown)


examples/stm32wba6/src/bin/ble_gatt_server.rs
```diff
@@ -78,54 +73,10 @@ struct AppState {
...
-    });
-
-    config.rcc.ahb_pre = AHBPrescaler::Div1;
-    config.rcc.apb1_pre = APBPrescaler::Div1;
-    config.rcc.apb2_pre = APBPrescaler::Div1;
-    config.rcc.apb7_pre = APBPrescaler::Div1;
-    config.rcc.ahb5_pre = AHB5Prescaler::Div4;
-    config.rcc.voltage_scale = VoltageScale::Range1;
-    config.rcc.sys = Sysclk::Pll1R;
-    config.rcc.mux.rngsel = mux::Rngsel::Hsi;
```
xoviat, line 116: This appears to have a different configuration than other examples.

embassy-stm32/src/rcc/wba.rs
```diff
@@ -101,6 +102,52 @@ impl Config {
...
+        // VCO = 16MHz * 12 = 192MHz, PLLR = 192 / 2 = 96MHz system clock
+        rcc.pll1 = Some(Pll {
+            source: PllSource::Hse,
+            prediv: PllPreDiv::Div2,  // 32MHz / 2 = 16MHz to PLL input
+            mul: PllMul::Mul12,       // 16MHz * 12 = 192MHz VCO
+            divr: Some(PllDiv::Div2), // 192MHz / 2 = 96MHz system clock
+            divq: None,
+            divp: Some(PllDiv::Div12), // 192MHz / 12 = 16MHz for peripherals
+            frac: Some(0),
+        });
```
xoviat, line 137: Not sure whether this configuration will work for wba.

leftger: You can change it to what you think is best.

xoviat: This comment was made purely from the observation that this configuration is different in the examples. I have not worked this up in CubeMX.

examples/stm32wba6/src/bin/ble_gatt_server.rs
leftger: Might have been an oversight from my end or Tswish's end.

leftger:
I'll double-check with my HW in about an hour

leftger:
Works on my HW!

leftger (approved):

PR #6097, opened 2026-05-13
Synopsys-OTG: type-erase endpoint counts

bugadani:
This PR allows type-erasing the USB instance in HAL drivers without forcing them to use the same capacity endpoint/channel state buffer (i.e. without wasting memory for the "smaller" peripheral). This is done by introducing a type-erased state struct that holds slices and references, and also by moving the endpoint allocation buffers to the global state. This is the foundation for https://github.com/esp-rs/esp-hal/pull/5560

### DIFF
--- a/embassy-usb-synopsys-otg/src/host.rs
+++ b/embassy-usb-synopsys-otg/src/host.rs
@@ -85,14 +85,38 @@ impl<const CH_COUNT: usize> HostState<CH_COUNT> {
             inited: AtomicBool::new(false),
         }
     }
+
+    /// Borrow this [`HostState`] as an [`OtgHostState`] for [`OtgHostInstance`].
+    pub fn as_otg_host_state(&self) -> OtgHostState<'_> {
+        OtgHostState {
+            channels: self.channels.as_slice(),
+            port_waker: &self.port_waker,
+            port_event: &self.port_event,
+            port_speed: &self.port_speed,
+            inited: &self.inited,
+        }
+    }
+}
+
+/// Type-erased view of [`HostState`] for [`OtgHostInstance`], [`OtgHost`], [`on_host_interrupt`], and pipes.
+///
+/// Build from [`HostState::as_otg_host_state`].
+#[derive(Clone, Copy)]
+pub struct OtgHostState<'d> {
+    channels: &'d [ChannelState],
+    port_waker: &'d AtomicWaker,
+    port_event: &'d AtomicU8,
+    port_speed: &'d AtomicU8,
+    inited: &'d AtomicBool,
 }
 
 /// Hardware-dependent host configuration.
-pub struct OtgHostInstance<'d, const CH_COUNT: usize> {
+#[derive(Copy, Clone)]
+pub struct OtgHostInstance<'d> {
     /// The USB peripheral registers.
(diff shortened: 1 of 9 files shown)


embassy-usb-synopsys-otg/src/host.rs
```diff
@@ -85,14 +85,45 @@ impl<const CH_COUNT: usize> HostState<CH_COUNT> {
...
+            inited: &self.inited,
+        }
+    }
+}
+
+/// Type-erased view of [`HostState`] for [`OtgHostInstance`], [`OtgHost`], [`on_host_interrupt`], and pipes.
+///
+/// Build from [`HostState::as_otg_host_state`].
+#[derive(Clone, Copy)]
+pub struct OtgHostState<'d> {
```
Dirbaio: The naming is very confusing: we have `HostState` and `OtgHostState`. There's no way to tell the difference from the name.

- Do you think it's possible to make the erased `OtgHostState` private? Maybe we can copy what `embassy-net` does: only the [non-erased struct](https://github.com/embassy-rs/embassy/blob/6fa07cf2aa341528b0e36b6311b4e6ecfc61b0fc/embassy-net/src/lib.rs#L89) is public, and only [fn new()](https://github.com/embassy-rs/embassy/blob/6fa07cf2aa341528b0e36b6311b4e6ecfc61b0fc/embassy-net/src/lib.rs#L327) is generic, it erases things internally before storing.
- If this trick doesn't work here we should find better names.

embassy-usb-synopsys-otg/src/host.rs
```diff
@@ -85,14 +85,45 @@ impl<const CH_COUNT: usize> HostState<CH_COUNT> {
...
+    channels: &'d [ChannelState],
+    port_waker: &'d AtomicWaker,
```
Dirbaio: This is 4 pointers. It would be more efficient to store a single pointer to a "flags" struct.

embassy-usb-synopsys-otg/src/host.rs
Dirbaio: Is this guaranteed to be the same as `state.channel_count()`? If so we could remove it.

bugadani: In the current stm32 impl, I believe the code is written to allocate memory for the biggest instance, so... yes, if I rework that. Which I'm hesitating to do, because I generally don't have the devices.

embassy-usb-synopsys-otg/src/host.rs
bugadani: I'll have to think about this. It'd mean coupling esp-hal more tightly to embassy-usb, but I guess we're already tightly coupled.

bugadani: @dirbaio I can offer you `xStateStorage` for the generic type, and `xState` for the non-generic, or `xState` and `DynXState` respectively.

bugadani:
I'm done with the shape-related changes, the only remaining open concern is the naming.

STM32 channel/endpoint counts are strictly speaking equivalent to the old code (nothing has more than 12 endpoints, to the old code always boiled down to it, correct or not), but I guess using the endpoint count for host channels is questionable.

PR #6098, opened 2026-05-13
wpan/platform: expose borrow_aes, borrow_pka, fill_random_bytes

leftger:
## Summary

Promotes `Platform::borrow_aes` and `Platform::borrow_pka` from `pub(crate)` to `pub`, and adds a new `pub async fn fill_random_bytes` that reads from the RNG pipe fed by `run_rng`.

## Motivation

STM32WBA chips have a single AES, single PKA, and single RNG peripheral. When BLE is initialized with `Platform::new_full`, the platform takes ownership of all three. Applications that want to do their own crypto (e.g. AES-GCM for a secure transport, P-256 ECDSA for signing, etc.) currently have no way to access those peripherals — they would need a second instance, which the hardware doesn't have.

The accessors already exist internally (called from the C link-layer callbacks for AES-128 CMAC and P-256 LE Secure Connections). They guard access with a critical-section mutex, which is exactly the synchronization an application needs to share the peripheral with the BLE stack. Making them `pub` lets applications reuse this machinery instead of re-implementing it externally.

`fill_random_bytes` complements this for the RNG case: `run_rng` already holds the RNG mutex permanently and pumps bytes into a pipe for the BLE stack to consume; exposing a pipe reader lets applications draw from the same well.

### DIFF
--- a/embassy-stm32-wpan/src/wba/platform.rs
+++ b/embassy-stm32-wpan/src/wba/platform.rs
@@ -141,7 +141,24 @@ impl Platform {
         }
     }
 
-    pub(crate) fn borrow_aes<R>(&self, f: impl FnOnce(&mut Aes<'static, AesPeriph, Blocking>) -> R) -> R {
+    /// Fill `buf` with random bytes from the BLE platform's RNG pipe.
+    ///
+    /// The pipe is fed by [`Self::run_rng`] using the same hardware RNG that
+    /// backs the BLE controller, so applications sharing this `Platform` for
+    /// crypto don't need a second `Rng` instance.
+    pub async fn fill_random_bytes(&self, buf: &mut [u8]) {
+        let mut filled = 0;
+        while filled < buf.len() {
+            let n = self.pipe.read(&mut buf[filled..]).await;
+            filled += n;
+        }
+    }
+
+    /// Borrow the shared AES peripheral for the duration of `f`.
+    ///
+    /// Panics if the platform was built with [`Self::new_basic`] (no AES).
+    /// Held under a critical section, so `f` should be short.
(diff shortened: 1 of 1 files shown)


embassy-stm32-wpan/src/wba/platform.rs
```diff
@@ -141,7 +141,24 @@ impl Platform {
...
+    /// The pipe is fed by [`Self::run_rng`] using the same hardware RNG that
+    /// backs the BLE controller, so applications sharing this `Platform` for
+    /// crypto don't need a second `Rng` instance.
+    pub async fn fill_random_bytes(&self, buf: &mut [u8]) {
+        let mut filled = 0;
+        while filled < buf.len() {
+            let n = self.pipe.read(&mut buf[filled..]).await;
+            filled += n;
+        }
+    }
```
xoviat, line 155: This is not a valid implementation of this function, because it can potentially starve the Platform. The implementation that I thought of would wait until the pipe is full, then read at most 64 bytes from the pipe.

leftger: OK that's fine if you wanna edit this PR and add on top of it I'm ok with that. If not that's fine and I can address this in a bit. I do need to share the PKA,AES,etc peripherals because I'll be using it to encrypt other communication buses.

xoviat: I understand that you need to share them, and I agree with that (that was the direction that I was heading, anyway). Neither am I objecting to this function, just the implementation of it. I have not confirmed #6094 on hardware yet. If you want, we can merge this and then I can pull upstream into that PR, and then I'll fix this function later. Up to you.

leftger: OK let's merge this and you can fix this function. Thanks again!

PR #6099, opened 2026-05-13
Usb HID host fixes

enchant97:
## Changes
1. During enumeration it is possible to get a `InvalidDescriptor` error due to the device minimum packet size being 8 bytes. This PR changes the default to the minimum (8 bytes) until the actual device min is known.
2. binterval is not used for HID interrupt leading to integer overflow error.

### DIFF
--- a/embassy-usb-host/src/lib.rs
+++ b/embassy-usb-host/src/lib.rs
@@ -277,7 +277,7 @@ impl<'d, A: UsbHostAllocator<'d>> BusHandle<'d, A> {
         let ep0_info = EndpointInfo {
             addr: EndpointAddress::from_parts(0, UsbDirection::In),
             ep_type: EndpointType::Control,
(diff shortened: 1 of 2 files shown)


embassy-usb-host/src/lib.rs
```diff
@@ -277,7 +277,7 @@ impl<'d, A: UsbHostAllocator<'d>> BusHandle<'d, A> {
         let ep0_info = EndpointInfo {
             addr: EndpointAddress::from_parts(0, UsbDirection::In),
             ep_type: EndpointType::Control,
-            max_packet_size: route.device_speed().max_packet_size(),
+            max_packet_size: 8,
```
leftger: Can you set this as a variable such as:
```
max_packet_size: DEFAULT_MAX_PACKET_SIZE,
```

PR #6104, opened 2026-05-13
Octospi build rs

xoviat:
Fix build.rs to support impl. for multiple pintraits.


embassy-stm32/build.rs
```diff
@@ -1375,54 +1388,173 @@ fn main() {
...
-        (("octospim_alt", "P1_IO0"), quote!(crate::ospi::D4Src<{ crate::ospi::OCTOSPIM_P1_LOW }>)),
-        (("octospim_alt", "P1_IO1"), quote!(crate::ospi::D5Src<{ crate::ospi::OCTOSPIM_P1_LOW }>)),
-        (("octospim_alt", "P1_IO2"), quote!(crate::ospi::D6Src<{ crate::ospi::OCTOSPIM_P1_LOW }>)),
-        (("octospim_alt", "P1_IO3"), quote!(crate::ospi::D7Src<{ crate::ospi::OCTOSPIM_P1_LOW }>)),
-        (("octospim", "P1_DQS"), quote!(crate::ospi::DQSSrc<{ crate::ospi::OCTOSPIM_P1_CTRL }>)),
-        (("octospim", "P1_NCS"), quote!(crate::ospi::NSSSrc<{ crate::ospi::OCTOSPIM_P1_CTRL }>)),
-        (("octospim", "P1_CLK"), quote!(crate::ospi::SckSrc<{ crate::ospi::OCTOSPIM_P1_CTRL }>)),
-        (("octospim", "P1_NCLK"), quote!(crate::ospi::NckSrc<{ crate::ospi::OCTOSPIM_P1_CTRL }>)),
+        (
+            ("octospim", "P1_IO0"),
```
hydra, line 1443: this is the problem I was trying to fix with `_alt` you can't add `("octospim", "P1_IO0"),` to the HashMap twice. See line 1391 above.  Same key, different value, thus only the last-added value is present.

hydra (changes requested):

embassy-stm32/build.rs
```diff
@@ -1532,7 +1679,9 @@ fn main() {
         (("lcd", "VLCD"), quote!(crate::lcd::VlcdPin)),
         (("dac", "OUT1"), quote!(crate::dac::DacPin<Ch1>)),
         (("dac", "OUT2"), quote!(crate::dac::DacPin<Ch2>)),
-    ].into();
+    ] {
+        signals.entry(key).or_default().push(value);
```
hydra, line 1683: oh wait, I missed that you changed it from a hashmap... carry on!

xoviat: My hashmap is different from yours.

hydra: yup, I see you're adding multiple values now, probably more maintainable this way, nice fix.  I didn't like my `_alt` approach, it felt too much like a hack.

PR #6107, opened 2026-05-14
executor: add platform-riscv64 support

DrSquid:
Adds a `platform-riscv64` Cargo feature to `embassy-executor` for RISC-V 64-bit targets, sharing implementation with `platform-riscv32` via a single `platform/riscv.rs` thread module (WFI-based; interrupt executor unsupported). Both feature gates point at the same source file; instruction-set selection is driven by `--target`, matching how the Cortex-M and Cortex-A/R platform modules already cover multiple sub-architectures from one file.

### DIFF
--- a/embassy-executor/CHANGELOG.md
+++ b/embassy-executor/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Added `platform-riscv64` for RISC-V 64-bit targets (thread executor only, uses `WFI`; shares implementation with `platform-riscv32`).
+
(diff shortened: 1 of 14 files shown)


lulf (commented):
Thanks!

I'm not sure if the tests/riscv64 should be included:

* CI doesn't run them
* Example is already compiled to check that it compiles.

DrSquid:
You are welcome! 

Sure, no problem removing tests/riscv64 or would you prefer me adding it to the CI even though the example covers it.

lulf:
> You are welcome!
> 
> Sure, no problem removing tests/riscv64 or would you prefer me adding it to the CI even though the example covers it.

I would just remove the tests/riscv64

DrSquid:
test/riscv64 removed.

lulf (approved):

PR #6108, opened 2026-05-14
STM32N6 fix PLL output frequency calculation

antonello-contini-ok:
According to chapter 14.6.5 of RM0486, PLL post-dividers POSTDIV1 and POSTDIV2 both contribute to output frequency but the implemented formula only accounts for POSTDIV1.

image

This PR proposes to include POSTDIV2 in the output frequency calculation


xoviat:
cc @tarfu can you approve this?

tarfu:
@xoviat LGTM, I wrongly dropped the p2 when changing from `in_clk.0 / m / n / p1 / p2` in which `m / n` was wrong and needed to be `m * n`.

This never caught me as In all my example divp2 is Div1

PR #6109, opened 2026-05-14
[MCXA] Fix set_as_input: clear PIDR to re-enable pin input

SanjeevKumm:
set_as_disabled() sets PID=1 in the Pin Input Disable Register, which permanently disables input reads (always returns 0). set_as_input() must clear PID back to 0 to restore input functionality.


felipebalbi:
bender run

felipebalbi (approved):

PR #6110, opened 2026-05-14
Add string descriptors.

flaviojs:
Add StringDescriptorZero.
Add StringDescriptor.
Add english language identifiers.
Fix doc links.

### DIFF
--- a/embassy-usb-host/src/descriptor.rs
+++ b/embassy-usb-host/src/descriptor.rs
@@ -3,11 +3,13 @@
 
 use embassy_usb_driver::host::HostError;
 use embassy_usb_driver::{Direction, EndpointInfo, EndpointType};
+use heapless::Vec;
 
 /// Standard descriptor type constants.
 pub mod descriptor_type {
     pub const DEVICE: u8 = 0x01;
     pub const CONFIGURATION: u8 = 0x02;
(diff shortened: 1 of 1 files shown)


embassy-usb-host/src/descriptor.rs
```diff
@@ -16,6 +18,45 @@ pub mod descriptor_type {
...
+        /// English (United Kingdom)
+        pub const UK: u16 = super::ENGLISH | (0x02 << 10);
+        /// English (Australian)
+        pub const AUS: u16 = super::ENGLISH | (0x03 << 10);
+        /// English (Canadian)
+        pub const CAN: u16 = super::ENGLISH | (0x04 << 10);
+        /// English (New Zealand)
+        pub const NZ: u16 = super::ENGLISH | (0x05 << 10);
+        /// English (Ireland)
+        pub const EIRE: u16 = super::ENGLISH | (0x06 << 10);
```
leftger, line 42: Is EIRE an acceptable name? I know it's Gaelic for Ireland but wouldn't it be better to just put IRELAND?

leftger: Nevermind 🤦‍♂️ It shows up as `EIRE` in the USB spec.

leftger (approved):

PR #6111, opened 2026-05-14
embassy-sync: lockless `AtomicWaker`; add `CriticalSectionWaker`

felipebalbi:
Replace the previous `AtomicWaker` (a typedef-style alias for `GenericAtomicWaker<CriticalSectionRawMutex>`) with a lockless single-slot waker, ported from `futures::task::AtomicWaker` and adapted for `no_std`.

Coordination uses a small atomic state machine (WAITING / REGISTERING / WAKING) so that neither `register()` nor `wake()` enters a critical section. On Cortex-M this means `wake()` does not disable interrupts, which avoids unnecessary jitter.

The previous typedef-style wrapper around
`GenericAtomicWaker<CriticalSectionRawMutex>` is preserved verbatim under a new name, `CriticalSectionWaker`, and lives in its own `critical_section_waker.rs` module. Callers that want the old critical-section semantics (`wake_by_ref` + restore, no atomic state machine) can switch to it without changes to their call sites beyond the type name. `GenericAtomicWaker<M>` itself is unchanged.

Behavioral note for `AtomicWaker`: the new `wake()` consumes the waker (takes it out of the slot before invoking it) instead of waking by reference and restoring it. Tasks always re-register their waker on each poll, so this is observably equivalent for the single-waiter contract that this type is designed for.

Measured on FRDM-MCXA266 (MCXA266, Cortex-M33 @ 180 MHz, FIRC HF, WfeUngated), GPIO IRQ wake path with a scope:

| edge               | mutex   | lockless | delta   |
|--------------------|---------|----------|---------|
| wake() body (mean) | 1920 ns | 1630 ns  | -290 ns |

### DIFF
--- a/embassy-sync/src/waitqueue/atomic_waker.rs
+++ b/embassy-sync/src/waitqueue/atomic_waker.rs
@@ -156,29 +156,25 @@ impl AtomicWaker {
 
     /// Wake the registered waker, if any.
     ///
-    /// The waker is consumed (taken out of the slot) before being invoked.
-    /// The task is expected to re-register its waker on the next poll, as is
-    /// already the convention for users of [`CriticalSectionWaker`].
+    /// The stored waker is invoked via [`Waker::wake_by_ref`] and remains
+    /// registered, matching the semantics of
+    /// [`CriticalSectionWaker`](super::CriticalSectionWaker).
     pub fn wake(&self) {
-        if let Some(w) = self.take() {
-            w.wake();
-        }
-    }
-
-    fn take(&self) -> Option<Waker> {
         match self.state.fetch_or(WAKING, AcqRel) {
             WAITING => {
                 // SAFETY: We hold the WAKING flag; no other call can touch
                 // the waker cell while we own it.
-                let w = unsafe { (*self.waker.get()).take() };
+                unsafe {
+                    if let Some(w) = &*self.waker.get() {
+                        w.wake_by_ref();
+                    }
+                }
                 self.state.swap(WAITING, Release);
-                w
             }
             _ => {
                 // Either a register is in progress (it will see the WAKING
                 // bit on its CAS-back and wake the supplied waker itself),
                 // or another wake is in flight.
-                None
             }
         }
     }
--- a/embassy-sync/src/waitqueue/atomic_waker.rs
+++ b/embassy-sync/src/waitqueue/atomic_waker.rs
@@ -96,9 +96,15 @@ const WAKING: usize = 0b10;
 /// enter a critical section, which avoids unnecessary jitter on higher-priority
 /// interrupts and improves wake latency for time-critical async drivers.
 ///
-/// If a waker is registered, registering another waker replaces the previous
-/// one without waking it. Intended to wake a single task from an interrupt
-/// handler.
+/// If a waker is registered, registering a different waker replaces the
+/// previous one and wakes it, giving the displaced task a chance to
+/// re-register itself. This matches the semantics of
+/// [`WakerRegistration`](super::WakerRegistration); note that
+/// [`GenericAtomicWaker`] and
+/// [`CriticalSectionWaker`](super::CriticalSectionWaker) instead replace the
+/// previous waker *without* waking it.
+///
+/// Intended to wake a single task from an interrupt handler.
 ///
 /// On targets without 32-bit atomic CAS support (e.g. Xtensa S2, AVR, RV32I),
 /// `AtomicWaker` is a type alias for
@@ -128,7 +134,12 @@ impl AtomicWaker {
         }
     }
 
-    /// Register a waker. Overwrites the previous waker, if any.
+    /// Register a waker.
+    ///
+    /// If a different waker was previously registered, it is replaced and
+    /// woken so the displaced task can re-register itself if it is still
+    /// interested. This matches
+    /// [`WakerRegistration::register`](super::WakerRegistration::register).
     ///
     /// If a [`wake`](Self::wake) call is in flight when this is called, the
     /// supplied waker is woken directly to schedule a re-poll.
@@ -139,13 +150,21 @@ impl AtomicWaker {
             .unwrap_or_else(|x| x)
         {
             WAITING => {
+                // The waker evicted from the cell (if any), to be woken
+                // *after* we release REGISTERING so the wake cannot re-enter
+                // while we still own the cell.
+                let evicted: Option<Waker>;
+                // The newly-registered waker, taken back out in the handoff
+                // path because a racing `wake()` transferred wake duty to us.
+                let pending_wake: Option<Waker>;
+
                 // SAFETY: We hold the REGISTERING flag; no other call can
                 // touch the waker cell while we own it.
                 unsafe {
-                    match &*self.waker.get() {
-                        Some(old) if old.will_wake(waker) => {}
-                        _ => *self.waker.get() = Some(waker.clone()),
-                    }
+                    evicted = match &*self.waker.get() {
+                        Some(old) if old.will_wake(waker) => None,
+                        _ => (*self.waker.get()).replace(waker.clone()),
+                    };
 
                     // Try to release the cell by clearing REGISTERING. This
                     // CAS only succeeds if no `wake()` raced us; otherwise the
@@ -153,26 +172,37 @@ impl AtomicWaker {
                     // CAS fails — that failure is precisely the handoff
                     // signal from `wake()`.
                     let res = self.state.compare_exchange(REGISTERING, WAITING, AcqRel, Acquire);
-                    match res {
-                        Ok(_) => {}
+                    pending_wake = match res {
+                        Ok(_) => None,
                         Err(_) => {
                             // A concurrent `wake()` observed our REGISTERING
                             // bit and intentionally left the WAKING bit set
                             // instead of touching the waker cell (which we
                             // own). The wake duty has been transferred to us:
-                            // take the waker out, clear both bits in a single
-                            // `swap` so a new `register()` / `wake()` cycle
-                            // can begin, and finally drive the wake.
+                            // take the waker out and clear both bits in a
+                            // single `swap` so a new `register()` / `wake()`
+                            // cycle can begin. We drive the wake outside the
+                            // unsafe block, below.
                             //
                             // This is what makes the "missed wake" race
                             // impossible: the wake is delivered exactly once,
                             // either by `wake()` itself (no race) or by
                             // `register()` here (race), but never dropped.
-                            let waker = (*self.waker.get()).take().unwrap();
+                            let w = (*self.waker.get()).take().unwrap();
                             self.state.swap(WAITING, AcqRel);
-                            waker.wake();
+                            Some(w)
                         }
-                    }
+                    };
+                }
+
+                // Cell ownership has been released. It is now safe to invoke
+                // user wakers, which may re-enter `register()` / `wake()` on
+                // this same `AtomicWaker`.
+                if let Some(w) = pending_wake {
+                    w.wake();
+                }
+                if let Some(w) = evicted {
+                    w.wake();
                 }
             }
             WAKING => {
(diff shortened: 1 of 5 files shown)


felipebalbi:
I was considering just making the turbo waker the default, it looks like all the bits are already in stable, but I'm not entirely sure.

(and it looks like xtensa doesn't like this waker heh)

xoviat:
thumbv6 strikes again! You should make the atomicwaker used conditional on the target.

felipebalbi:
> thumbv6 strikes again! You should make the atomicwaker used conditional on the target.

Yeah, completely missed that. How about it being conditional on `target_has_atomic...`?

embassy-sync/src/waitqueue/atomic_waker.rs
```diff
@@ -41,26 +48,128 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
...
+    /// already the convention for users of [`CriticalSectionWaker`].
     pub fn wake(&self) {
-        self.waker.wake();
+        if let Some(w) = self.take() {
+            w.wake();
+        }
+    }
+
+    fn take(&self) -> Option<Waker> {
+        match self.state.fetch_or(WAKING, AcqRel) {
```
diondokter, line 238: Can't this put the state in an undefined state? 0b11?
I see this is sort of handled by the register function, but that wakes the future, something that the take function doesn't necessarily imply

Should maybe be a compare exchange too

felipebalbi: I can change to compare exchange, sure, but the states are not mutually exclusive here, they are used as flags. Here's what `fetch_or` is giving us:

 1. If state was `WAITING`: we now own the `waker` cell (state = `WAKING`), drive the wake, swap back to `WAITING`, and we're done.
 2. If state was `REGISTERING`: state becomes `REGISTERING | WAKING`. We do not touch the `waker` cell and we do not  clear our bit. We hand the wake duty off to the in-flight `register()`.

Now, in `register()` we sort it all out: it CASs `REGISTERING` to `WAITING` to release. If a wake snuck in, that CAS fails (because state is now 0b11), the Err branch fires, and register itself calls `waker.wake()` before clearing both bits with swap(`WAITING`). The wakeup is delivered only once.

With compare and exchange we could have a hung task instead:

With the same register/wake race, the CAS fails and `wake()` silently returns. But `register()`'s CAS succeeds (state was `REGISTERING` switches cleanly to `WAITING`). It never learns there was a pending wake. Wakeup lost. Task hung.

Am I missing something here? Maybe I need to add a comment with these details?

diondokter: Ok, it might help to mention in docs for the constants that they're supposed to be flags. They don't look like flags since you have a waiting/0 const too.

Given that they are flags, they are handled soundly I think.

But a failed `take` meaning the waker can be woken is a little weird I think. (Weird as in surprising, not wrong)
But maybe this is the best way to do it? Idk, I don't see a great alternative tbh

felipebalbi: okay, added more docs now. Should be pretty clear how the flags are being used, glad to update if needed.

xoviat:
bender run

felipebalbi:
bender run

(after setting trusted label)

diondokter (approved):

felipebalbi:
@Dirbaio , @xoviat any concerns with this PR?

xoviat:
I am +1 to merge. However, I am not going to click the merge button on this because 1) you can do it and 2) this is core functionality and I don't want to be partially responsible for breaking it if it turns out this was a mistake.

felipebalbi:
> I am +1 to merge. However, I am not going to click the merge button on this because 1) you can do it and 2) this is core functionality and I don't want to be partially responsible for breaking it if it turns out this was a mistake.

I'll give dirbaio a chance to read as well :-)

embassy-sync/src/waitqueue/atomic_waker.rs
```diff
@@ -41,26 +52,202 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
...
+                            // `swap` so a new `register()` / `wake()` cycle
+                            // can begin, and finally drive the wake.
+                            //
+                            // This is what makes the "missed wake" race
+                            // impossible: the wake is delivered exactly once,
+                            // either by `wake()` itself (no race) or by
+                            // `register()` here (race), but never dropped.
+                            let waker = (*self.waker.get()).take().unwrap();
+                            self.state.swap(WAITING, AcqRel);
+                            waker.wake();
```
bugadani: This should probably be documented in the changelog - previously, it was enough to register `AtomicWaker` once, now you need to register it every time before futures are polled. Callers getting this wrong will silently stop getting scheduled.

embassy-sync/src/waitqueue/atomic_waker.rs
```diff
@@ -41,26 +52,202 @@ impl<M: RawMutex> GenericAtomicWaker<M> {
...
+            .compare_exchange(WAITING, REGISTERING, Acquire, Acquire)
+            .unwrap_or_else(|x| x)
+        {
+            WAITING => {
+                // SAFETY: We hold the REGISTERING flag; no other call can
+                // touch the waker cell while we own it.
+                unsafe {
+                    match &*self.waker.get() {
+                        Some(old) if old.will_wake(waker) => {}
+                        _ => *self.waker.get() = Some(waker.clone()),
```
bugadani: Since you're changing semantics, should this wake the evicted waker?

embassy-sync/src/waitqueue/atomic_waker.rs
felipebalbi: I'm not sure I follow. AFAICT re-registering on every poll was always a requirement and that is not changing. Additionally, `Future::poll()` can be called by any executor holding any `Context`, no? This matches, in my understanding, the documentation to `poll` where it states:

> Note that on multiple calls to poll, only the Waker from the Context passed to the most recent call should be scheduled to receive a wakeup.

I'm reading that as an implication that the waker could be different from call to call and, therefore, it should be re-registered.

Additionally, here are some examples of drivers re-registering the waker on every call:

- embassy-rp ADC: https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/adc.rs#L206-L220
- embassy-rp GPIO: https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/gpio.rs#L332-L364
- embassy-rp DMA: https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/dma.rs#L284-L294
- embassy-stm32 RNG (note the "Need to check condition after register to avoid a race" comment): https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/rng.rs#L178-L195
- embassy-stm32 USART (Tx): https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L660-L670
- embassy-stm32 USART (Rx): https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L886-L929
- embassy-nrf TWIM: https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/twim.rs#L361-L384
- embassy-mspm0 i2c-target: https://github.com/embassy-rs/embassy/blob/main/embassy-mspm0/src/i2c_target.rs#L478-L499

Furthermore, is this not similar, in nature, to what atomic_waker_turbo does with its uncondtional store on register?

Maybe I'm missing the point you're trying to make?

bugadani: The problem is that the _old_ implementation didn't need you to re-register the waker in subsequent polls, because the waker wasn't taken and dropped. Some of the callers (which may be out of tree!) may expect this (arguably a bug at call site), so this currently silent change in behaviour should be documented.

felipebalbi: Let me try to understand what you're saying here:

1. old implementation did not require re-registration of the waker in subsequent polls.
2. old implementation did not take the waker and drop it

Let me try to counter both:

## old implementation did not require re-registration of the waker in subsequent polls.

For (1): that contract is imposed by Future::poll itself, not by embassy-sync. Its docs are explicit:
 
Note that on multiple calls to poll, only the Waker from the Context passed to the most recent call should be scheduled to receive a wakeup.

The only conformant way to honor that is to re-register on every poll. A caller that registers once and reuses across polls is already broken against std::future::Future, independent of which waker implementation backs the slot.

Worth noting too: the new wake() is also non-destructive on the common path. It uses wake_by_ref and never touches the waker cell - only the state atomic flips. So the slot survives across wake() calls exactly like the old AtomicWaker::wake(). The only place a waker is taken is the register() race-handoff branch, which is what (2) is really about, see below.

## old implementation did not take the waker and drop it

For (2): we're not dropping the waker on the floor. The race-handoff branch consumes the just-installed waker (the caller's brand-new cx.waker() from the current register() call) and fires it. The wake is delivered exactly once. If we were to set the UnsafeCell back to the waker instead, here's what would happen:

- t=0: race branch puts the waker back, fires it.
- t=ε: re-poll runs (because we just fired the wake), calls register(), overwrites the put-back waker with a fresh cx.waker().
 
i.e. putting the waker back just to have it immediately erased. Zero observable difference.
 
I can change the code to put the waker back into the cell, but it would require either an extra clone or a restructure to wake_by_ref + set, and the put-back waker would be immediately overwritten by the next register(cx.waker()) in the upcoming re-poll. So it's strictly wasted work - a stray wake() between the handoff and the re-poll is a no-op in either case (Waker::wake is idempotent on an already-scheduled task).


If I'm still missing anything, let me know.

bugadani: I haven't seen anyone argue so hard against a changelog entry before 😅

felipebalbi: Lol, sorry. I can add the changelog. Just trying to understand what you're really saying 😅

Anyway, I'll send an extra commit adding a changelog entry in a little bit. Just want to make sure I understand your point such that the changelog isn't incorrect.

felipebalbi: Added another CHANGELOG entry 👍

embassy-sync/src/waitqueue/atomic_waker.rs
Dirbaio: Agreed, it'd be better to do this, for consistency with WakerRegistration.

felipebalbi: Almost done with this change. Carrying out the same change for `GenericAtomicWaker` to make sure the semantics is the same everywhere.

bugadani: > Carrying out the same change for GenericAtomicWaker to make sure the semantics is the same everywhere.

Please don't change semantics of existing public types. You're retaining the type for compatibility in this PR, so it would be best to keep it _unchanged_.

felipebalbi: alright, two more commits added, one patching only `AtomicWaker` and another patching `GenericAtomicWaker`. Everything should agree on wake semantics. Tests updated too. If you prefer to keep `GenericAtomicWaker` as it was, all we have to do is drop or revert the top commit.

felipebalbi: > > Carrying out the same change for GenericAtomicWaker to make sure the semantics is the same everywhere.
> 
> Please don't change semantics of existing public types. You're retaining the type for compatibility in this PR, so it would be best to keep it _unchanged_.

missed this comment. Let me drop the top-most commit.

felipebalbi: done, only AtomicWaker follows the new semantics.

felipebalbi:
pico is having a hard time:

```
 ERROR teleprobe::client > === rpi-pico out/tests/rpi-pico/i2c: FAILED: HTTP request failed with status code: 400: Bad Request
```

xoviat:
bender run

driveraid:
When is there a release with this? I'm using embassy on a multi-core chip where this would help spinlock contention.

diondokter:
@driveraid a release happens when it happens. There's no set schedule

PR #6112, opened 2026-05-15
STM32N6 calculate PLL outputs even when previously initialized

antonello-contini-ok:
Currently, if PLLs are already initialized (e.g. by FSBL) and embassy configuration matches the hardware one, initialization is skipped and PLL output frequencies are not calculated.

However, this causes ICs depending on that PLL to not have their frequency set and a panic if a mux tries to use those ICs.

This PR proposes to calculate the PLL output frequencies based on the values provided by the configuration since it already matches with the registers

### DIFF
--- a/embassy-stm32/src/rcc/n6.rs
+++ b/embassy-stm32/src/rcc/n6.rs
@@ -695,10 +695,60 @@ fn disable_pll(pll_index: usize) {
     cfgr1.modify(|w| w.set_pllbyp(false));
 }
 
-fn init_pll(pll_config: Option<Pll>, pll_index: usize, input: &PllInput) -> PllOutput {
+fn init_pll(pll_config: Option<Pll>, pll_index: usize, input: &PllInput, skip_initialization: bool) -> PllOutput {
     let cfgr1 = RCC.pllcfgr1(pll_index);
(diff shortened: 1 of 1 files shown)


xoviat:
This should be moved outside of the `is_new_pll_config` branch:

```rust
            *out = pll.map_or_else(
                || {
                    disable_pll(n);
                    PllOutput::default()
                },
                |c| init_pll(Some(c), n, &pll_input),
            );
```

so that `init_pll` is always called. Then, the `is_new_pll` logic can be moved inside of `init_pll`.

antonello-contini-ok:
> This should be moved outside of the `is_new_pll_config` branch:
> 
> ```rust
>             *out = pll.map_or_else(
>                 || {
>                     disable_pll(n);
>                     PllOutput::default()
>                 },
>                 |c| init_pll(Some(c), n, &pll_input),
>             );
> ```
> 
> so that `init_pll` is always called. Then, the `is_new_pll` logic can be moved inside of `init_pll`.

i've extracted the calculation of the pll output frequency, now init_pll is called only once. Moving that branch inside init_pll would have required changing the function signature to pass the muxes sources calculated before. this seems more clean to me.

PR #6113, opened 2026-05-15
stm32/pka: add async API

fbrozovic:
Pka now follows a dual blocking/async pattern. It's generic over Mode, with Pka::new (async) and Pka::new_blocking. The 20 op methods split per mode: async keeps the natural name, blocking gets a _blocking suffix. Bodies share private prepare_*/read_* helpers so register/RAM setup isn't duplicated.

Callers in embassy-stm32-wpan and the wba/wba6 PKA examples are updated for the rename.


fbrozovic:
@leftger I've only tested this functionality on H7S. It would be great if you could re-test on WBA/WBA6

leftger:
I'll test this in 2-3 hours

leftger:
Works on WBA6 too :)

leftger (approved):

PR #6114, opened 2026-05-15
embassy-time driver for ARM CMSDK

robamu:


xoviat:
Why is this a separate crate? It should be included in embassy-time under a feature gate.

robamu:
i can move it there

embassy-time/src/lib.rs
```diff
@@ -23,6 +23,8 @@ mod driver_mock;
 #[cfg(feature = "mock-driver")]
 pub use driver_mock::MockDriver;
 
+#[cfg(feature = "cmsdk")]
+pub mod cmsdk;
```
xoviat: `driver_cmsdk`

xoviat (changes requested):

xoviat (approved):

PR #6117, opened 2026-05-15
Feature/stm32 h7 bkpsram configuration

tlanser-plexus:
Added BKPSRAM register configuration for STM32H7 parts, as they use different registers than other STM32 controllers that were already implemented. Registers and configuration settings are based on the RM0433 Reference Manual from STMicro.

Depends on the following stm32-data PR to provide memory map location and Retention enum for the STM32H7 series parts:
https://github.com/embassy-rs/stm32-data/pull/792


Dirbaio:
bender run

PR #6121, opened 2026-05-16
stm32/cryp: fix on h7s and improve DMA payload

fbrozovic:


xoviat:
Looks like HIL failed

fbrozovic:
I changed the implementation so it should work with unaligned buffers as well.
I don't have a NUCLEO-H755ZI to test on, but I've tested both aligned and unaligned buffers on a NUCLEO-H7S3L8 and got the following results, both with 32k payloads:

- 4-byte aligned: 290us (~113 MB/s)
- misaligned (1-byte): 565us (~58 MB/s)

PR #6124, opened 2026-05-16
enforce that embassy-executor-timer-queue is only once in the dependency tree

robamu:
Does it ever make sense to have two different versions, possibly with different ITEM_SIZES, in the dependency tree? I had that issue when mixing git and crates-io dependencies of embassy-executor/embassy-time.

this led to a silent failure because embassy-executor used crate-io v0.1 with item size 0, while the item size 6 was only activated for the git version of embassy-executor-timer-queue


Dirbaio:

Issue #6126, opened 2026-05-16
embassy_nrf: nrf54: time_driver_grtc occupying special timer channel


cwriter:
Hi

I've run into an issue on the nrf54l15 using the GRTC time driver, reducing the potential functionality.

The nrf54l15's GRTC provides a single channel with periodic triggering (somewhat equivalent to the normal `TIMERXX.CC[n].SHORTS.CLEAR`:
This CC[0] makes use of the `GRTC.INTERVAL` register to add the INTERVAL value back to the CC channel on `EVENTS_COMPARE[0]`, making it a suitable low-power source for PPI-driven workloads (without CPU intervention).

The current implementation of the GRTC time driver uses CC n=0 without making use of the `INTERVAL` register. Namely, embassy only requires a single wakeup, following by a reprogramming of the next scheduled timer event.

Is it possible to reconfigure without triggering a breaking change?

PR #6127, opened 2026-05-16
nrf_ppi: time_driver_grtc: Switch to CC[11] to keep CC[0] free

cwriter:
A proposal to fix the issue outlined in #6126 .

I'm not sure about the implications onto other chips. From what I understand, the `_grtc` feature is currently available only on `nrf54*` ICs. However, this change could be a breaking change potentially, especially if existing applications already use CC[11] differently.

I'm not sure if adding a feature flag to switch to CC[11] makes more sense, but this could potentially clutter the feature flag namespace for a technically small change.

Also, I cannot test other chips, and I'm not sure about the implications on the nRF53. Any feedback is welcome

### DIFF
--- a/embassy-nrf/src/time_driver.rs
+++ b/embassy-nrf/src/time_driver.rs
@@ -23,6 +23,22 @@ fn rtc() -> pac::rtc::Rtc {
     pac::RTC1
 }
 
+// The nRF54 series (which has the GRTC) uses asymmetric channels
+// Namely, CC[0] is the only channel which supports "periodic interval" (page 298 in the datasheet),
+// which supports periodic alarms by adding the value in the INTERVAL register when EVENTS_COMPARE[0] triggers
+// Because the time driver does not make use of periodic alarms (but always reprograms it), it does not need to make use of this.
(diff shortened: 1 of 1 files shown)


embassy-nrf/src/time_driver.rs
```diff
@@ -23,6 +23,22 @@ fn rtc() -> pac::rtc::Rtc {
...
 }
 
+// The nRF54 series (which has the GRTC) uses asymmetric channels
+// Namely, CC[0] is the only channel which supports "periodic interval" (page 298 in the datasheet),
+// which supports periodic alarms by adding the value in the INTERVAL register when EVENTS_COMPARE[0] triggers
+// Because the time driver does not make use of periodic alarms (but always reprograms it), it does not need to make use of this.
+// This selects the last CC channel (CC[11]) instead of CC[0] to keep CC[0] free for application use,
+// specifically for PPI-driven events that aim to avoid CPU intervention in periodic tasks.
+#[cfg(feature = "_grtc")]
+fn time_driver_cc_n() -> usize {
```
Dirbaio: it's more idiomatic to use a const.

PR #6131, opened 2026-05-18
wpan/bluetooth: fix advertising name and use random static address on WBA

leftger:
- gap/advertiser: extract_local_name and extract_service_uuids_16 now keep the AD-type byte in the returned slice. aci_gap_set_discoverable expects local_name / service_uuid_list to start with the AD-type byte (with the length including it); previously the name byte was stripped, so phones saw the device as "N/A".

- gap_init: add AddressType { Public, RandomStatic } on GapInitParams, default to RandomStatic, and skip the PUBADDR config write in that mode. PUBADDR (offset 0x00) is rejected by WBA stack-basic when the address isn't a real IEEE-registered public address.

- bluetooth/mod: mask the UID-derived BD address into a valid random static address (top two bits of byte 5 set to 0b11) and program it via hci_le_set_random_address after init_gap_and_hal.

- gap/types: default AdvParams.own_addr_type to Random to match.

### DIFF
--- a/embassy-stm32-wpan/src/bluetooth/gap/aci_gap.rs
+++ b/embassy-stm32-wpan/src/bluetooth/gap/aci_gap.rs
@@ -57,6 +57,27 @@ unsafe extern "C" {
     /// Set device in non-discoverable mode (stop advertising)
     #[link_name = "ACI_GAP_SET_NON_DISCOVERABLE"]
     fn aci_gap_set_non_discoverable() -> tBleStatus;
+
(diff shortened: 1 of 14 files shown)


leftger:
`ble_peripheral_connect` now properly broadcasts `Embassy_Peripheral` when using the nRF Connect app

leftger:
I've fixed a couple more previously buggy ble examples on wba6

xoviat (approved):

PR #6132, opened 2026-05-18
stm32/bksram: improved volatile memcpy

xoviat:
This certainly generates a lot more code in the binary than the previous impl. I can't say whether it's worth it or not.

### DIFF
--- a/embassy-stm32/src/backup_sram.rs
+++ b/embassy-stm32/src/backup_sram.rs
@@ -1,5 +1,7 @@
 //! Battary backed SRAM
 
+use core::{ptr, slice};
+
 use embassy_hal_internal::Peri;
 
 use crate::_generated::{BKPSRAM_BASE, BKPSRAM_SIZE};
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/backup_sram.rs
```diff
@@ -29,50 +31,73 @@ impl BackupMemory {
...
-    /// Write single byte to backup sram
+    /// Write bytes to backup sram
     ///
     /// Address is relative start of backup sram
+    #[inline]
     pub fn read(&mut self, address: usize, dst: &mut [u8]) {
         assert!(address + dst.len() <= self.size());
-        let p = unsafe { self.as_ptr().add(address) };
+        unsafe {
+            let (start, mid, end) = slice::from_raw_parts(self.as_ptr().add(address), dst.len()).align_to::<u64>();
```
Dirbaio, line 58: I think this is UB. `&` references are marked as "dereferenceable" to LLVM, so lust the act of creating them gives permission to the compiler to read from them however it likes, aligned or not. ie the compiler is allowed to assume `&` references point to "normal memory".

For this to work you really have to ensure you only create raw pointers pointing to the bkpsram, never references.

(What's the motivation for this optimization? do we have evidence the old code is too slow for some use case?)

PR #6137, opened 2026-05-19
Fix mac address handling

Tavish-Srivastava:
Fix random static address masking and add public address support for testing
  
  Moves the |= 0xC0 random static address bit-mask from init_with_role into init_gap_and_hal's
  RandomStatic branch so it only applies when actually using a random address. Removes the redundant
   le_set_random_address call (the WBA stack sets the address internally during BleStack_Init).
  Fixes the address log to read from config slot 0x2E — the slot the stack actually uses — instead
  of the unrelated params.bd_addr value.

  Adds HCI::new_with_gap_params to allow callers to supply custom GapInitParams, enabling a fixed
  known address for testing. All wba5 and wba6 BLE examples now have a single const ADDR_TYPE toggle
   to switch between random and public address modes.


Tavish-Srivastava:
Tested on hardware (wba65)

leftger (approved):
OK I see where my assumptions went wrong. This does fix the shortcomings my code had. LGTM 🚀

PR #6144, opened 2026-05-20
stm32wba: fix HSE init clobbering HSION; add Config::new_wpan_lsi()

leftger:
## Summary

- **Bug fix**: `RCC.cr().write()` when enabling HSE zeroed the entire CR register, clearing HSION immediately after `hsi_enable()` set it. Any config with both HSI and HSE enabled (including `new_wpan()`) silently disabled HSI, breaking RNG and any other HSI-clocked peripheral. Fixed by using `modify()` instead of `write()`.
- **Defensive assert**: added a check before the HSI disable that catches the misconfiguration `rngsel=Hsi` + `hsi=false` at init time with a clear message instead of a silent runtime failure.
- **New constructor**: `Config::new_wpan_lsi()` for STM32WBA boards without an LSE crystal. Mirrors `new_wpan()` but uses LSI1 as the RTC and 2.4 GHz radio sleep timer clock source. `RADIOSTSEL` is set to `_RESERVED_2` (hardware value `0x02` = LSI) pending a named `Lsi` variant in the metapac — see stm32-data PR [#795](https://github.com/embassy-rs/stm32-data/pull/795).

### DIFF
--- a/embassy-stm32-wpan/src/wba/ll_sys_if.rs
+++ b/embassy-stm32-wpan/src/wba/ll_sys_if.rs
@@ -405,7 +405,7 @@ pub unsafe extern "C" fn ll_sys_sleep_clock_source_selection() {
     let radiostsel = RCC.bdcr().read().radiostsel();
     let slp_clk_src = match radiostsel {
         Radiostsel::Lse => link_layer::_SLPTMR_SRC_TYPE_E_RTC_SLPTMR as u8,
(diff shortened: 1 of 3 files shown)


leftger:
This PR has the latest change from the stm32 metapac. It's important enough in the sense that in previous BLE examples I had used the LSE for the clock domains. @Tavish-Srivastava and I are working on a specific WBA variant that uses the LSI for the BLE stack. This PR addresses that issue.

PR #6148, opened 2026-05-20
Add StringDescriptorLossy.

flaviojs:
This descriptor wrapper allows reading invalid data just like [String::from_ut1f16_lossy](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16_lossy).
USB mandates valid UNICODE, but you never know what vendors will do in practice.

### DIFF
--- a/embassy-usb-host/src/descriptor.rs
+++ b/embassy-usb-host/src/descriptor.rs
@@ -1117,6 +1117,29 @@ impl StringDescriptor {
         }
         Self { string }
     }
+
+    /// Alternate to [USBDescriptor::try_from_bytes] that allows invalid UNICODE data.
+    ///
+    /// Invalid data will be replaced with [`U+FFFD` REPLACEMENT CHARACTER](char::REPLACEMENT_CHARACTER) (�).
+    fn try_from_bytes_lossy(bytes: &[u8]) -> Result<Self, DescriptorError> {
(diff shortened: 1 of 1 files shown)


leftger:
While I do applaud defensive programming in most of its forms I'm hesitant to accept this on the premise of the possibility of something occurring. If there is evidence of at least one product NOT providing UNICODE characters then I would be willing to merge this in.

flaviojs:
Recently I saw this in someone's linux dmesg log:
```
[    1.316201] usb 3-1: New USB device found, idVendor=057e, idProduct=0305, bcdDevice= 1.00
[    1.316229] usb 3-1: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[    1.316244] usb 3-1: Product: BCM2045�
```
It's the reason that led me to create this wrapper, which I will use in my code (either from embassy or from my code).
I'm not sure if it's actually an invalid character, but it reminded me of why we have lossy utf functions in rust in the first place.

I should also point out that USB includes the capability of setting descriptors, and I certainly don't trust vendors to validate the data they receive in hardware, specially if they are the ones providing the official driver (a supposedly trusted source).

leftger:
OK this answered my question. Seems acceptable to implement this.

leftger (approved):

PR #6151, opened 2026-05-20
Add FMAC peripheral

usbalbin:
Work in progress for adding support for the FMAC peripheral.

I have looked at the Cordic impl for some inspiration, please let me know if I am on the right track.

My use case and end goal for this is to be able to feed the FMAC from an ADC triggered by a timer and then fire an interrupt every time the FMAC is done computing its result. This would be for the control loop of a power converter where minimal overhead is needed. I am of course happy to try adding support for other use cases as well.

So far only support for stm32g4, tested on a nucleo-g474

### DIFF
--- a/embassy-stm32/src/fmac.rs
+++ b/embassy-stm32/src/fmac.rs
@@ -1,0 +1,321 @@
+//! Filter Math Accelerator
+
+pub use dsp_fixedpoint::Q16;
+use embassy_hal_internal::{Peri, PeripheralType};
+
+use crate::{peripherals, rcc};
+
+/// FMAC driver
+pub struct Fmac<'d, T: Instance> {
+    peri: Peri<'d, T>,
+}
+
+/// FMAC instance
+trait SealedInstance {
+    /// Get access to FMAC registers
+    fn regs() -> crate::pac::fmac::Fmac;
+
+    /// Read value from RDATA
+    fn read_result(&self) -> u16 {
+        Self::regs().rdata().read().res()
+    }
+
+    fn read_q16(&self) -> Q16<15> {
+        Q16::new(self.read_result() as i16)
+    }
+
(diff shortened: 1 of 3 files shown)


xoviat:
Looks good!

embassy-stm32/src/fmac.rs
```diff
@@ -0,0 +1,321 @@
...
+    pub read_method: AccessMethod,
+    /// Set how the input will be written
+    pub write_method: AccessMethod,
+}
+
+/// The input buffer is full
+#[derive(Clone, Debug)]
+pub struct ErrorInputFull;
+
+impl<'d, T: Instance> Fmac<'d, T> {
```
usbalbin, line 135: Notes to self:

TODO: @xoviat's comment from matrix chat

> I don't think you would be able to use the regular sequence for anything else then. So maybe the best way to do this is to have a new FromAdc struct in the fmac mod that takes ownership of the fmac and the adc struct.
And then you can add a pub(crate) function to the g4 adc that you can call from the FromAdc struct.
What do you think?

I think that makes a lot of sense. I think the FMAC should conceptually own the ADC and have complete control of it without anyone else interfering.

Maybe add an unsafe-ish method for setting offset of adc. Note owever, we are only allowed to write ADC_OFRy.OFFSET register while the ADC has ADSTART = 0 and JADSTART = 0. So having a dynamic setpoint will be tricky. (RM0440 21.7.17)

Use DMA to connect ADC and FMAC with regular conversion done(or whatever it might be called), as DMA trigger.

leftger:
I think you forgot to run `fmtall.sh`. Do that, push, and you should be good 😄

leftger:
@usbalbin is this ready for review?

usbalbin:
It is quite MVP at this point. No DMA, no async etc. But it is usable for testing

PR #6153, opened 2026-05-21
stm32wba: BLE bonded-connection support with controller privacy (RPA)

leftger:
## Summary

Adds everything needed to make MITM-protected bonding with iOS-compatible RPA reconnect work on the STM32WBA BLE stack.

### 1. NVM persistence + LE SC fixes
- Rewrite `BLEPLAT_NvmStore` to write the whole registered NVM cache buffer, matching the ST reference. The previous version mis-interpreted the `size` argument (u64 words, not bytes) and effectively never persisted bond data.
- Fix `flash_erase_page` for BANK_2 addresses on STM32WBA65RI (2 MB flash); the hardcoded `bker=false` was BANK_1-only.
- Fix `NvmCacheBuffer` size — `CFG_BLE_NVM_SIZE_MAX = 256` is u64 word units, so the buffer is 2048 bytes, not 256. The 8× undersized buffer was overflowing into `DYN_ALLOC_BUFFER` and silently corrupting bond state.
- Add `set_nvm_base_address`, `register_nvm_cache`, `load_nvm_from_flash` helpers in `linklayer_plat.rs` and wire them into `init_ble_stack`.
- Export `erase_bond_nvm_flash` to let applications clear stale NVM bonds before stack init (reflashing does not erase the NVM page).
- Dispatch deferred `BLEPLATCB_PkaComplete` from embassy-task context after each `seq_resume()`. Calling it re-entrantly from within the sequencer corrupts the BLE stack state machine and silently makes LE SC pairing time out at 30 s.

### 2. Security manager additions
- Bind `aci_gap_get_bonded_devices`, `aci_gap_add_devices_to_list`, `aci_gap_configure_filter_accept_list`, `hci_le_set_address_resolution_enable`.
- Add `SecurityManager::configure_filter_and_resolving_list()` — clears and repopulates both the Filter Accept List and the resolving list from bonded devices (`aci_gap_add_devices_to_list` mode `0x05`), matching the ST `BLE_Privacy_Peripheral` reference flow.
- Add `SecurityManager::bonded_device_count()` and `SecurityManager::log_bonded_devices()` for diagnostics.
- Fix `aci_gap_init`'s `privacy_enabled` parameter: the stack accepts only `0x00` (disabled) or `0x02` (Controller Privacy); `params.privacy_enabled as u8` produced `0x01` for `true`, which is undefined and silently left SMP in a "no privacy" state.
- Fix symbol name: `ACI_GAP_CLEAR_SECURITY_DB` (was incorrectly `ACI_GAP_CLEAR_SECURITY_DATABASE`).
- Add `GapInitParams::skip_authentication_setup` to avoid double-configuring auth when the application uses `SecurityManager` directly.

### 3. Controller privacy (RPA) advertising
- Bind `ACI_GAP_SET_UNDIRECTED_CONNECTABLE` and `ACI_GAP_UPDATE_ADV_DATA`.
- Add `AdvParams::privacy_undirected` flag: when set, `advertiser::configure` uses the `set_undirected_connectable` + `update_adv_data` path required by the ST privacy-peripheral flow instead of `aci_gap_set_discoverable`.
- Add `GAP_RESOLVABLE_PRIVATE_ADDR` constant for `own_address_type`.

### 4. HCI event handling improvements
- Add `ControllerAdapter::read_hci_event` — reads HCI events with a vendor-event fallback parser so that `GapPairingComplete` with `reason=0` (success) is not silently dropped (stm32wb-hci rejects reason 0).
- Add `parse_event_with_fallback` and `vendor_event_is_hal_firmware_warning` utilities in `util.rs`.
- Fix `set_event_mask` / `le_set_event_mask` to use `bt-hci` byte-slice constructors (`FromHciBytes`) instead of the removed `EventFlags::all()` / `LeEventFlags::all()` API.
- Update stm32wb-hci rev to `08112f`; stm32-bindings to `0.2.0`.

### 5. Examples
- **`ble_bonding`** (stm32wba6): updated to use controller privacy flow — `configure_filter_and_resolving_list` + `privacy_undirected` advertising — for iOS RPA reconnect without re-pairing.
- **`ble_advertiser_lp`** (stm32wba6-lp): new low-power BLE advertiser that uses the STOP-mode executor between advertising ticks.


embassy-stm32-wpan/src/wba/controller.rs
```diff
@@ -201,6 +201,41 @@ impl<'d, T: Runtime> ControllerAdapter<'d, T> {
...
+            };
+
+            match crate::util::parse_event_with_fallback(evt.kind.0, evt.data) {
+                Ok(event) => return Ok(event),
+                Err(_) => {
+                    if crate::util::vendor_event_is_hal_firmware_warning(
+                        evt.data,
+                        crate::bluetooth::security::SecurityManager::FW_ERROR_SMP_UNEXPECTED_LTK,
+                    ) {
+                        warn!("HAL firmware warning 0x06: SMP unexpected LTK request (bond lookup failed)");
```
xoviat, line 226: Can you post the event bytes that stm32wb-hci couldn't parse for this?

leftger: I'll update to your latest `stm32wb-hci` rev and send you the unexpected LTK request warning in a couple of hours. I need to get some rest lol.

leftger:
I've gotten this to work. Bonded connections persist (generally). There's some edge cases but I can have a persisted bonded connection to the Nucleo board that (occasionally) reconnects. I'm happy pushing this @xoviat

PR #6157, opened 2026-05-22
Add panic-free time functions and document panics

flaviojs:
I noticed the panic info in the `Timer::after` docs, but this is not enough for me, I need every individual panicky place documented.

`cargo asm` was used to inquire which places panic, and documentation added.

I also added `try_*` constructors and `checked_*` mutators, following the convention in Instant.
I also added `try_*` solo functions, 

I failed to compile the tests with feature "mock-driver", so instead I used the "std" driver.

### DIFF
--- a/embassy-time/src/instant.rs
+++ b/embassy-time/src/instant.rs
@@ -125,7 +125,11 @@ impl Instant {
     }
 
     /// Duration between this Instant and another Instant
-    /// Panics on over/underflow.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the computed duration overflows.
+    /// Avoid panics with [`Instant::checked_duration_since()`] or [`Instant::saturating_duration_since()`].
     pub fn duration_since(&self, earlier: Instant) -> Duration {
         Duration {
             ticks: unwrap!(self.ticks.checked_sub(earlier.ticks)),
@@ -133,6 +137,8 @@ impl Instant {
     }
 
     /// Duration between this Instant and another Instant
+    ///
+    /// This is a panic-free [`Instant::duration_since()`].
     pub fn checked_duration_since(&self, earlier: Instant) -> Option<Duration> {
         if self.ticks < earlier.ticks {
             None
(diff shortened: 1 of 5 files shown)


flaviojs:
Ah, I want to make a panic-free `block_for` but don't know what naming convention to follow here.
Suggestion?

Sidenote: I want to use panic-free code. I found that panic locations produce much more assembly than what I expected (it adds up fast), a panic-free implementation is more awkward to code but produces better and smaller assembly.

flaviojs:
Converting to draft, I noticed `cargo asm` didn't pick up some panicky functions that were probably optimized away. Will work on it.

Would still appreciate your feedback about this though.

flaviojs:
I opted for try_* in solo functions.

This is a minor semver change (instead of major) because the new `WithTimeout::try_with_timeout` has a default implementation.

Done.

embassy-time/src/instant.rs
```diff
@@ -216,6 +248,11 @@ impl SubAssign<Duration> for Instant {
 impl Sub<Instant> for Instant {
     type Output = Duration;
 
+    /// Computes `Instant - Instant`. [Read more](Sub)
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the computed duration overflows in [`Instant::duration_since`].
```
Dirbaio, line 255: in what situation can `instant - instant` overflow?

Dirbaio: ah duh, if RHS > LHS. ignore me

lulf (approved):

lulf (commented):
In what situations should one use the checked_ vs try_ prefix? I noticed that there is `checked_reset() -> Option`, but `try_every() -> Option` and it felt a little inconsistent.

flaviojs:
@lulf I followed the convention already present in Instant:
* `try_*` constructors
* `checked_*` mutators

Without feedback, I ended up opting for `try_*` in solo functions.

Feel free to specify what convention should be used.

flaviojs:
@lulf since you're the only one that seems to care about the naming:
Would naming all the new functions/methods `try_*` satisfy you in terms of consistency?

lulf:
@flaviojs The convention we should follow is that of the core language library, ideally. But looking into that, I've not found any documented convention here, so I was merely asking the question. I think your argument for the current approach is sound, so no objection from me on merging this.

flaviojs:
Cool, I rebased with main and applied changes so this can be merged.

Note that #6192 converted `Duration` arguments to `impl Into<Duration>`, but block_for was forgotten.
It's out of scope for this PR so I left the argument unchanged.

flaviojs:
@lulf 1-week reminder poke
The merge failure looks like a transient error unrelated to this PR.

PR #6159, opened 2026-05-22
Add support for packetmeta to embassy-net

datdenkikniet:
Currently, we can't do packetmeta stuff with `embassy-net` because the plumbing does not exist.

Add the plumbing and the wiring so that we can

### DIFF
--- a/embassy-net/src/driver_util.rs
+++ b/embassy-net/src/driver_util.rs
@@ -1,6 +1,6 @@
 use core::task::Context;
 
-use embassy_net_driver::{Capabilities, Checksum, Driver, RxToken, TxToken};
+use embassy_net_driver::{Capabilities, Checksum, Driver, PacketMeta, RxToken, TxToken};
 use smoltcp::phy::{self, Medium};
 use smoltcp::time::Instant;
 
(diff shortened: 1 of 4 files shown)


embassy-net/src/driver_util.rs
```diff
@@ -114,4 +118,28 @@ where
...
+}
+
+#[allow(unused, reason = "meta isn't used if no features are enabled")]
+fn into_embassy_net_meta(meta: phy::PacketMeta) -> PacketMeta {
+    let out_meta = PacketMeta::default();
+    #[cfg(feature = "packetmeta-id")]
+    {
+        out_meta.id = meta.id;
+    }
+    out_meta
```
lulf, line 144: Why are not the  Into/From traits used here?

lulf (commented):
Nice feature, just left a minor comment on ideomatic rust.

embassy-net/src/driver_util.rs
datdenkikniet: `PacketMeta` has to be defined in `embassy-net-driver` (since that's where this value is actually relevant), which would have to take a new dependency on `smoltcp` to implement this conversion (potentially behind a feature flag, but still). 

I interpreted the existence of this adapter file (where the conversion now lives) as the fact/a hint that we want `embassy-net-driver` to be as lightweight as possible, and have `embassy-net` contain the glue for `embassy-net-driver` and `smoltcp` interop.

PR #6164, opened 2026-05-22
stm32: fix uid mod

xoviat:
closes the hal issue embassy-rs/stm32-data#800

transferred the issue to the pac


xoviat:
confirmed on hardware

PR #6168, opened 2026-05-23
[time] Add experimental embedded-hal-timer traits

diondokter:
Hey, I'm trying to get some timer and alarm traits into embedded-hal.
This adds those to embassy-time.

I want to add more impls for the HAL hardware timers too, but will do that in different PRs.

To consider:
- Is `HalTimer` a good name? `Timer` is already taken for something else
- Should it be behind an 'unstable' feature flag?
  - The traits will likely get a breaking change at some point...
- Currently when the timer hasn't started any elapsed function will return `u64::MAX` and alarm invocations will return `Err(Overflow)`. Do we think that's good behavior?
  - We could drop requiring calling the start function, but I think that'd lead to bad drivers that don't call start

### DIFF
--- a/embassy-time/src/eh_timer.rs
+++ b/embassy-time/src/eh_timer.rs
@@ -1,0 +1,125 @@
+use core::cell::Cell;
+
+use critical_section::Mutex;
+use embassy_time_driver::TICK_HZ;
+use embedded_hal_timer::{Alarm, OverflowError, Timer};
+
+use crate::{Duration, Instant};
+
+/// Embedded-hal implementation on top of embassy-time
+#[derive(Debug)]
+pub struct HalTimer(Mutex<Cell<Option<Instant>>>);
+
+impl HalTimer {
+    /// Construct a new timer that's not running.
+    /// Make sure to call [Self::start].
+    pub fn new() -> Self {
+        Self(Mutex::new(Cell::new(None)))
+    }
+
+    fn get_instant(&self) -> Option<Instant> {
+        critical_section::with(|cs| self.0.borrow(cs).get())
+    }
+}
+
+impl Timer for HalTimer {
+    fn start(&mut self) {
+        critical_section::with(|cs| self.0.borrow(cs).set(Some(Instant::now())));
+    }
+
+    fn tickrate(&self) -> u64 {
+        TICK_HZ
+    }
+
+    fn elapsed_ticks(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_ticks()).unwrap_or(u64::MAX))
+    }
+
+    fn elapsed_nanos(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_nanos()).unwrap_or(u64::MAX))
+    }
+
+    fn elapsed_micros(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_micros()).unwrap_or(u64::MAX))
+    }
+
+    fn elapsed_millis(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_millis()).unwrap_or(u64::MAX))
+    }
+
+    fn elapsed_secs(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_secs()).unwrap_or(u64::MAX))
+    }
+
+    fn max_ticks(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_ticks()
+    }
+
+    fn max_nanos(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_nanos()
+    }
+
+    fn max_micros(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_micros()
+    }
+
+    fn max_millis(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_millis()
+    }
+
+    fn max_secs(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_secs()
+    }
+}
+
+impl Alarm for HalTimer {
+    async fn wait_until_ticks(&mut self, value: u64) -> Result<(), OverflowError> {
+        let end_time = self
+            .get_instant()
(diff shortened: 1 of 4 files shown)


embassy-time/src/eh_timer.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use critical_section::Mutex;
+use embassy_time_driver::TICK_HZ;
+use embedded_hal_timer::{Alarm, OverflowError, Timer};
+
+use crate::{Duration, Instant};
+
+/// Embedded-hal implementation on top of embassy-time
+#[derive(Debug)]
+pub struct HalTimer(Mutex<Cell<Option<Instant>>>);
```
bugadani: Why is this not simply an `Option<Instant>`? Nothing seems to need the extra critical section/interior mutability complexity.

embassy-time/src/eh_timer.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    fn elapsed_secs(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_secs()).unwrap_or(u64::MAX))
+    }
+
+    fn max_ticks(&self) -> u64 {
+        (Instant::MAX - Instant::now()).as_ticks()
+    }
+
+    fn max_nanos(&self) -> u64 {
```
bugadani, line 51: What's the use of these functions? There's a very good chance that the returned value will (after an interruption) actually be more than the value should be at the time of use.

bugadani:
Also this is just a generic wishlist item, but it would be nice if the abstraction could support [this](https://github.com/embassy-rs/embassy/blob/823abdbe8512721b4797536b9d3151c94c4bbad0/embassy-stm32/src/eth/generic_phy.rs#L132-L134) use case in some way.

embassy-time/src/eh_timer.rs
jannic: Indeed, this is an inherent TOCTOU race.

Also, depending on TICK_HZ, the multiplication inside as_nanos() may overflow the u64 long before the tick counter inside Instant overflows.

embassy-time/src/eh_timer.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn tickrate(&self) -> u64 {
+        TICK_HZ
+    }
+
+    fn elapsed_ticks(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_ticks()).unwrap_or(u64::MAX))
+    }
+
+    fn elapsed_nanos(&self) -> Result<u64, OverflowError> {
+        Ok(self.get_instant().map(|i| i.elapsed().as_nanos()).unwrap_or(u64::MAX))
```
jannic: This sometimes returns too large values: The first timer tick may happen immediately after calling start(). So it counts one tick even though only a fraction of a tick actually elapsed.
As the Timer trait documents these functions as rounding down, this additional tick needs to be subtracted.
(The [RFC](https://github.com/tweedegolf/embedded-hal-timer/blob/main/rfc.md) also says "The returned elapsed time is rounded down, so the contract is that at least this amount of time has passed. This is congruent with DelayNs and other timers in the ecosystem.")

embassy-time/src/eh_timer.rs
diondokter: Yeah I discovered this too while implementing. (But apparently you guys think this is more of a problem than I do and that's fair!)

It really is there just for utility. It allows a driver to assert there's enough time available for the intended usecase instead of having to find out when it already went wrong.
But for these 'continuously running' virtual timers the value steadily decreases. Maybe that means we simply can't have these max functions...

So, should we just remove them?

embassy-time/src/eh_timer.rs
diondokter: Ah right! Yeah this was based on an older impl where start was immutable. I guess this can just be an option now yeah

embassy-time/src/eh_timer.rs
bugadani: Why does OverflowError even exist in this API? Why is wanting to sleep beyond the end of time an error? The timer can measure the ticks necessary, even if the driver has to manage the overflow - the current Instant will not make sense, sure, but is that an issue when we are beyond the end of time anyway? If someone is lazy and implements that case as a call to `pending().await`, that's also a valid resolution.

Yes, `elapsed_*` may be unreliable beyond the end of time, and OverflowError is fair for those functions.

So, yes, I think some part of the API needs a bit more work, and removing the max allowed duration getters is probably a good idea, but it still fells like a bandaid on top of a perhaps underdone concept.

diondokter: The goal is that the trait is implementable on really simple hardware timers too (so without virtualization).

And I think I'd rather have an error rather than 'undefined' after the timer can't support operations anymore

bugadani: Then this abstraction will become a lot of headache for library users, if they will have to manage range extension themselves.

What's the intended use, and why doesn't `DelayNs` already cover that use case?

diondokter: If the library needs long running timers, then yes. They should document what the requirements are of the timer.
That's the same with SPI or I2c where the driver specifies through docs what the mode/speed/other settings need to be to be capable of communicating with the external device.

There's an RFC text btw:
https://github.com/tweedegolf/embedded-hal-timer/blob/main/rfc.md#why-do-we-need-this

bugadani: > If the library needs long running timers, then yes. 

And how does the current API actually support this? Functions can't enforce a minimum supported delay amount, `Alarm` has no operation that can wait for "at most" a specified amount of time so you can't loop with smaller timers, and there is no way to measure long time with these timers either because they will just report overflows.

diondokter: The behavior has changed according to the discussions of the WG meeting last week.
If the timer is designed to never overflow (in practice), then it should just return `u64::MAX`.

embassy-time/src/eh_timer.rs
diondokter: Yes, this is interesting. I wonder if we need to be that strict. After all, a valid implementation could do a really long blocking operation in the start function and it's undefined whether the timer was zeroed at the start or the end.

I'll post this as an issue on the timer repo so this isn't lost.

PR #6169, opened 2026-05-23
embassy_net::udp: add UdpMetadata newtype

datdenkikniet:
Without this additional type, converting between `embassy_net::udp::UdpMetadata` and `smoltcp::socket::udp::UdpMetadata` is a right pain due to the differing `PacketMeta` types. To fix it, we create an equivalent `UdpMetadata` newtype with the `embassy_net_driver::PacketMeta` inside of it, and remove the smoltcp re-export.

Converting the type is done because right now there is a type-disconnect between the `UdpMetadata.meta` (which is a `smoltcp::phy::PacketMeta`) used in `embassy_net::udp`, and the `PacketMeta` used in `embassy_net_driver` (which is a `embassy_net_driver;:PacketMeta`).

### DIFF
--- a/embassy-net/src/udp.rs
+++ b/embassy-net/src/udp.rs
@@ -4,12 +4,68 @@ use core::future::{Future, poll_fn};
 use core::mem;
 use core::task::{Context, Poll};
 
+pub use embassy_net_driver::PacketMeta;
 use smoltcp::iface::{Interface, SocketHandle};
 use smoltcp::socket::udp;
-pub use smoltcp::socket::udp::{PacketMetadata, UdpMetadata};
+pub use smoltcp::socket::udp::PacketMetadata;
 use smoltcp::wire::IpListenEndpoint;
 
-use crate::{Stack, TryError};
+use crate::{IpAddress, IpEndpoint, Stack, TryError};
+
+/// Metadata for a sent or received UDP packet.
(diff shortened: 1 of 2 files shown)


embassy-net/src/udp.rs
```diff
@@ -4,12 +4,68 @@ use core::future::{Future, poll_fn};
...
+    /// will be sent. Incoming datagrams always have this set. On outgoing datagrams, if it is not
+    /// set, and the socket is not bound to a single address anyway, a suitable address will be
+    /// determined using the algorithms of RFC 6724 (candidate source address selection) or some
+    /// heuristic (for IPv4).
+    pub local_address: Option<IpAddress>,
+    /// Additional metadata for the packet.
+    pub meta: PacketMeta,
+}
+
+impl UdpMetadata {
```
datdenkikniet, line 32: These are not `From` impls because:

1. `From<udp::UdpMetadata> for UdpMetadata` conflicts with the other `From` implementation (which is mirrored from `smoltcp::UdpMetadata`)
2. `From<UdpMetadata> for udp::UdpMetadata` exposes a weird to-smoltcp conversion that makes little sense to expose in the public API. The reverse cannot be applied due to point 1, and it's also not useful for anything.

lulf (approved):

PR #6173, opened 2026-05-24
docs: add Docker-based PDF build target

pgray:
Adds `make pdf` that produces a book-style PDF from the same AsciiDoc sources as `make all`. Pipeline runs entirely in Docker:

  1. asciidoctor/docker-asciidoctor : AsciiDoc -> DocBook XML (expands include:: across the docs/examples symlink)
  2. embassy-pandoc (Dockerfile.pandoc, FROM pandoc/latex:latest) : DocBook -> PDF via xelatex

Output defaults to /tmp/embassy-book.pdf; override with PDF_OUT. MAIN_FONT / SANS_FONT / MONO_FONT swap fonts on the command line — DejaVu (default) and IBM Plex are both pre-installed in the image.

Polish via pandoc-header.tex: microtype, sans-family chapter heads (titlesec), code blocks rendered as mdframed boxes with a left bar and no page-split, and a cleaner title page.

attached an example:
[embassy-book.pdf](https://github.com/user-attachments/files/28201176/embassy-book.pdf)

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,3 @@
+book/
+build/
+texput.log
--- a/docs/Dockerfile.pandoc
+++ b/docs/Dockerfile.pandoc
@@ -1,0 +1,24 @@
+# pandoc/latex with extra fonts for xelatex Unicode coverage and nice
+# typography. Used by `make pdf` to build the Embassy Book PDF.
+#
+# Bundled font families (all usable via MAIN_FONT / SANS_FONT / MONO_FONT):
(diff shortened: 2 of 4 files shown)


Dirbaio:
Thanks for the PR! However, I'm not convinced we should have tooling in the repo to generate PDF docs if the way we distribute the docs is HTML in the website. (and I'm not sure we should distribute PDF, I think HTML is a better format than PDF for docs)

pgray:
Oh ok. Thanks for reviewing. It made it super easy to read on an e-reader. Cheers!

PR #6182, opened 2026-05-27
can: gate TimingCalcError Debug behind !defmt

Jacke-debug:
can: use unwrap!() macro in set_bitrate for bxcan and fdcan

When the defmt feature is enabled, Result::unwrap() requires E: Debug,
but defmt builds prefer defmt::unwrap!() which requires E: defmt::Format
instead, avoiding the core::fmt::Debug machinery in the binary.


Dirbaio:
The better fix is to use the `unwrap!()` macro from `fmt.rs` which uses `defmt::unwrap!()` if `defmt` is enabled.

PR #6186, opened 2026-05-27
stm32/gpio: remove most critical sections

xoviat:
this should allow the optimizer to have a complete view of the situation.

### DIFF
--- a/embassy-stm32/src/reg.rs
+++ b/embassy-stm32/src/reg.rs
@@ -10,6 +10,13 @@ pub trait AtomicModify<T: Sized> {
     ///
     /// Call `set_xxx(false)` inside the closure
     fn clear_bits(&self, f: impl FnOnce(&mut T));
+    /// Atomically mask and unmask bits
+    ///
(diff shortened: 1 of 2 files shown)


xoviat:
bender run

embassy-stm32/src/reg.rs
```diff
@@ -10,6 +10,13 @@ pub trait AtomicModify<T: Sized> {
     ///
     /// Call `set_xxx(false)` inside the closure
     fn clear_bits(&self, f: impl FnOnce(&mut T));
+    /// Atomically mask and unmask bits
+    ///
+    /// Values may exist briefly in an intermediate state.
+    fn clear_set_bits(&self, f: impl Fn(&mut T)) {
```
Dirbaio, line 16: this does two read-modify-write cycles. are you sure it's faster?

also the correctness concern of "the register may have some garbage value in between" is a bit serious imo

xoviat:
assembly output was not good

PR #6193, opened 2026-05-28
Add RNG peripheral test for N6

william-spinelli:
Add test program for RNG peripheral on N6

Bump stm32-metapac to stm32-data-a1906cf83a69ba308d630c2cf1c1ba545c9ff101 which adds RNG v3 to perimap for N6

### DIFF
--- a/examples/stm32n6/src/bin/rng.rs
+++ b/examples/stm32n6/src/bin/rng.rs
@@ -1,0 +1,32 @@
+#![no_std]
+#![no_main]
+
+use defmt::*;
+use embassy_executor::Spawner;
+use embassy_stm32::rng::Rng;
+use embassy_stm32::{Config, bind_interrupts, peripherals, rng};
+use {defmt_rtt as _, panic_probe as _};
+
+bind_interrupts!(struct Irqs {
+    RNG => rng::InterruptHandler<peripherals::RNG>;
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let mut core_peri = unsafe { cortex_m::Peripherals::steal() };
+    core_peri.SCB.invalidate_icache();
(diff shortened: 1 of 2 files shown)


xoviat:
> Add test program for RNG peripheral on N6

are you going to add this? if the pr is not ready to merge, then it should be marked as draft

william-spinelli:
I had an issue with formatting and then I force pushed without the file...  Sorry about that.  Now everything is there :)

examples/stm32n6/src/bin/rng.rs
```diff
@@ -0,0 +1,32 @@
...
+
+bind_interrupts!(struct Irqs {
+    RNG => rng::InterruptHandler<peripherals::RNG>;
+});
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let mut core_peri = unsafe { cortex_m::Peripherals::steal() };
+    core_peri.SCB.invalidate_icache();
+    core_peri.SCB.enable_icache();
```
xoviat, line 18: Should we do this in embassy_stm32::init?

william-spinelli: Not always I guess, because there could be some conditions where you do not want to enable it.  Inside a bootloader where you are writing the flash is what comes to my mind.  But even then, you could have it enabled and invalidate it properly.
In any case if we move it to `embassy_stm32::init`, we should probably add a flag in `embassy_stm32::Config` to allow the user to enable or not enable the instruction cache.
We also need to gate this flag with some features since not all architecture support it.  As far as I know all ST MCU with M7 and M55 cores support instruction cache directly.  U5 (and probably many other families) support it via ST ICACHE peripheral.
Let me know what you think and thanks for all your help!

PR #6194, opened 2026-05-28
embassy-net: fix compilation with packetmeta-id feature

datdenkikniet:
Apparently I didn't do my due-dilligence and this just didn't compile at all.

Is there an easy way to add features to the stuff that is checked by CI? I don't understand the embassy-dev-tool, and can't figure out where (if anywhere) to add features that should be included in the CI run.

### DIFF
--- a/embassy-net-driver/Cargo.toml
+++ b/embassy-net-driver/Cargo.toml
@@ -21,6 +21,11 @@ target = "thumbv7em-none-eabi"
 [package.metadata.docs.rs]
 features = ["defmt"]
 
+[package.metadata.embassy]
+build = [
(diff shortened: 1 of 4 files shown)


lulf (commented):
My bad, I should've checked that it was added to CI. In the Cargo.toml, you have the `package.metadata.embassy` section, the build key contains a list of feature + target combinations. Add a combination with this feature enabled.

datdenkikniet:
Sorry if spammy, but would appreciate if you could take another look at this @lulf . All issues should be resolved now :)

lulf (approved):
Sorry for the delay, looks good!

PR #6195, opened 2026-05-28
embassy-net: change IP address info to `info` log level

chrysn:
The log levels used when new IP configuration is applied are inconsistent: The "DOWN" messages come at `info!`, whereas the "UP" and their details come in `debug!`.

This PR changes them all to info, for I think that's useful: Especially in testing and prototyping, the full infrastructure one would need to find their devices properly (multicast / registering / fixed-addresses) is often not set up, and users would just look at the debug output for the addresses -- even when they're not so deep in "why is it doing what it is doing" mode that they'd enable `debug` log level.

We've had this as a custom patch in Ariel OS for some time (it's the only patch left on embassy-net), but I think it's useful to have also directly in embassy, plus there's the consistency aspect.

### DIFF
--- a/embassy-net/src/lib.rs
+++ b/embassy-net/src/lib.rs
@@ -787,15 +787,15 @@ impl Inner {
 
         #[cfg(feature = "proto-ipv4")]
         if let Some(config) = &self.static_v4 {
-            debug!("IPv4: UP");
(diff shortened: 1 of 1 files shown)


lulf (approved):
I agree. It's also not something that will be printed a lot, so I think it's fair to log this at the info level.

Issue #6129, opened 2026-05-16
embassy-stm32 breaks at opt-level = 2 on SPI communication


bananasmoothii:
I am using an STM32F446RE with a MPU9250 accelerometer & gyro in SPI mode. The `dev` profile works fine but `release`doesn't, until I add this in Cargo.toml:
```toml
[profile.release.package.embassy-stm32]
opt-level = 1 # doesn't work above 2
```
The issue is wrong SPI readings: the first reading gives the device ID and is sometimes 0, sometimes 255, sometimes 166...

Note that I pass a `Spi<'static, Async, Master>` to code that only handles blocking call, but creating a blocking Spi with `Spi::new_blocking` didn't help.

Note: I use the [mpu9250](https://github.com/bananasmoothii/mpu9250-forked/tree/embedded-hal-1) crate but setting `opt-level = 1` to it didn't change anything

[source code at revision](https://github.com/bananasmoothii/micromouse/tree/7c121db9c02da63dd3ee7a0a96c0edccaa0c40b0)

xoviat:
can you contribute a minimal reproducer?

bananasmoothii:
No problem, I just made [a new branch](https://github.com/bananasmoothii/micromouse/tree/minimal-embassy-stm32-optimization-break) without all the fluff

xoviat:
@bananasmoothii does #6196  resolve this?

bananasmoothii:
@xoviat Indeed, thanks !

Issue #5607, opened 2026-03-05
STM32N6 SPI Hanging


jake-taf:
# **Issue**

The N6 (STM32N6570-DK Discovery kit) SPI occasionally hangs on reads. This can be observed to occur after running the below test code snippet for 10-900 seconds (based on testing with two disco kits). It happens even when nothing is connected to the SPI bus.

The issue occurs because the RXP flag (to signal data has been received) is never raised (I'm unsure why), and the driver code waits in an infinite loop for it. The call tree is:
- [blocking_read](https://github.com/embassy-rs/embassy/blob/3651d8ef249dc20d30df7382237f2451d889c011/embassy-stm32/src/spi/mod.rs#L596)
   - [transfer_word](https://github.com/embassy-rs/embassy/blob/3651d8ef249dc20d30df7382237f2451d889c011/embassy-stm32/src/spi/mod.rs#L1393)
      - [spin_until_rx_ready](https://github.com/embassy-rs/embassy/blob/3651d8ef249dc20d30df7382237f2451d889c011/embassy-stm32/src/spi/mod.rs#L1306)

### Steps performed by `blocking_read`

1. Disable SPI (clear CR1.SPE). The embassy documentation says this is needed to `avoid overrun causing the SPI RX state machine to get stuck...?`
2. Set the word size (CFG1.DSIZE = 0x7 (8-bit), skipped if already done)
3. Enable SPI (set CR1.SPE)
4. Flush RX FIFO (read RX data register until RXP is cleared)
5. Wait until we have room in the TX FIFO (SR.TXP flag set) (or TI frame format error, Mode Fault, CRC error, overrun)
6. Write zero to the TX data register
7. Master transfer start (set CR1.CSTART)
8. Wait until the RX FIFO has data (SR.RXP flag set) (or TI frame format error, Mode Fault, CRC error, overrun) <mark>[Trapped here, as the RXP flag is never set]</mark>
9. Read data from the RX data register

### Test code

The following code was used to trigger the issue. Effectively it spams the SPI with writes and reads. Note that SPI always sends data both ways simultaneously. A write is actually (send data, discard received) and a read is (send zero, receive data).

This code writes two randomly selected bytes followed by a read.

```Rust
let mut spi_config = spi::Config::default();
spi_config.frequency = Hertz(100_000);
let spi = spi::Spi::new_blocking(
  p.SPI5, // SPI
  p.PE15, // SCK
  p.PG2,  // MOSI
[...]

PR #6196, opened 2026-05-29
stm32/spi: add fences and cleanups

### LINKED ISSUE #6197: stm32: spi: pauses between words
Hardware: [nucleo-h723zg](https://www.st.com/en/evaluation-tools/nucleo-h723zg.html) 

In a usecases of ours we have a fast spi transfer (write and read, 8mhz) of roughly 40 bytes. It's very performance critical. 

We have two options: dma, which introduces roughly 10us of setup/teardown overhead, or blocking, which causes pauses between words.

I think the second one, the pauses, should be considered a bug, because there's no reason a pause should exist between words on a peripheral that has fifos.

Upon code review it's [quite easy to see why](https://github.com/embassy-rs/embassy/blob/664d4ead36bb24a63955ca649bcec66c6e70bf6d/embassy-stm32/src/spi/mod.rs#L612). It does:

```
for word in words {
    transfer(word);
}
```

Meaning, it uses the fifo as a single word buffer without overlap, which stalls the peripheral between every word.

Ideally, one wants to fill the write fifo entirely and then drain the read fifo while keeping the write fifo full.

If that's not feasible due to `embassy-stm32`'s generic nature, at least one word of overlap would already suffice in our usecases, as the transfer happens with interrupts disabled:

```
write(word[first])
for (prev, current) in words.windows(2) {
    transfer(write: current, read: previous)
}
read(words[last])
```

What's the reason for the current implementation without fifos? Simplicity? Is there a reason against utilizing fifos?

xoviat:
closes #6197.
possible fix for #6129
possible fix for #5607

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -609,10 +609,7 @@ impl<'d, M: PeriMode, CM: CommunicationMode> Spi<'d, M, CM> {
         // Memory barrier after flush RX fifo to ensure register writes complete
         fence(Ordering::SeqCst);
 
-        for word in words.iter_mut() {
-            *word = transfer_word(self.info.regs, W::default())?;
-        }
-        Ok(())
+        transfer_words(self.info.regs, words, &[])
     }
 
     /// Blocking in-place bidirectional transfer.
@@ -629,10 +626,7 @@ impl<'d, M: PeriMode, CM: CommunicationMode> Spi<'d, M, CM> {
         // Memory barrier after flush RX fifo to ensure register writes complete
         fence(Ordering::SeqCst);
 
-        for word in words.iter_mut() {
-            *word = transfer_word(self.info.regs, *word)?;
-        }
-        Ok(())
+        transfer_words(self.info.regs, words, words)
     }
 
     /// Blocking bidirectional transfer.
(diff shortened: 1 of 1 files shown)


xoviat:
Needs a fence on the blocking read

embassy-stm32/src/spi/mod.rs
```diff
@@ -563,6 +564,10 @@ impl<'d, M: PeriMode, CM: CommunicationMode> Spi<'d, M, CM> {
         self.set_word_size(W::CONFIG);
         self.info.regs.cr1().modify(|w| w.set_spe(true));
         flush_rx_fifo(self.info.regs);
+
+        // Memory barrier after DMA setup to ensure register writes complete before command
```
Finomnis: This is not a dma write, comment is probably wrong?

xoviat: This is effectively a copy of the fences in the sdmmc mod, with some additional fences added to the blocking methods. The relevant question is whether it fixes the problem.

Finomnis: I know, but the comment says "dma setup", this is a function that does not use dma

xoviat:
bender run

embassy-stm32/src/spi/mod.rs
Finomnis: Sorry, didn't mean to be pedantic.

Finomnis:
In case this becomes a general rework PR, `pub fn set_config(&mut self, config: &Config)` probably wants to return `Result<(), Infallible>`, not `Result<(), ()>`.

Although this is technically an API breaking change

xoviat:
@Finomnis do you see any other issues? If you want, we can go to a fifo-2 if all chips with a fifo can fit two words.

Finomnis:
So far not that I know of... This was the only issue I encountered so far. Using embassy-stm32 with rtic because I have some hard realtime requirements, and so far embassy-stm32 seems to work great :)

Much snappier than Zephyr

PR #6198, opened 2026-05-29
stm32: spi: Prevent stalling through overlapping transfers

### LINKED ISSUE #6197: stm32: spi: pauses between words

Finomnis:
The device currently stalls between every read and its following write, because the fifo only gets filled with a single word.

This change introduces one word of overlap to prevent this stall.

Fixes #6197.

Collides with #6196 and I will rebase it onto main once #6196 is merged.

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -600,9 +600,25 @@ impl<'d, M: PeriMode, CM: CommunicationMode> Spi<'d, M, CM> {
         self.set_word_size(W::CONFIG);
         self.info.regs.cr1().modify(|w| w.set_spe(true));
         flush_rx_fifo(self.info.regs);
+
+        // Overlap write and read to avoid stalling the peripheral
+        #[cfg(not(any(spi_v1, spi_v2)))]
+        if let Some((last, words)) = words.split_last_mut() {
+            write_word(self.info.regs, W::default())?;
+            for word in words {
+                *word = transfer_word(self.info.regs, W::default())?;
+            }
+            *last = read_word(self.info.regs)?;
+        }
+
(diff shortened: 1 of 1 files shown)


Finomnis:
It seems SPI v1/v2 only has a single word fifo, so this new overlap is not compatible with them and might lose rx words.

Need to verify and maybe add a cfg switch.

What's the maintainers feedback on this?

**EDIT:** I updated it so that previous implementation gets used for v1 and v2.

PR #6200, opened 2026-05-29
Add start_circular_dma for fire-and-forget DAC waveform output

Jacke-debug:
- Adds embassy-stm32/src/dac/ringbuffered.rs with RingBufferedDacChannel, following the existing ringbuffered.rs pattern in the timer, ADC, and UART modules.

- DacChannel::into_ring_buffered_8bit and into_ring_buffered_12right consume the channel and return a RingBufferedDacChannel<u32>. Buffers are [u32] because the DAC peripheral requires 32-bit DMA writes, sample values sit in the lower 8 or 12 bits. A static waveform is achieved with write_immediate + start; dynamic waveforms stream via write_exact in a loop.

- Removes start_circular_dma and start_circular_write, which are superseded by this API.

- Includes examples/stm32g4/src/bin/dac_dma_circular.rs: a 64-sample 12-bit sine at ~10 kHz on PA4, clocked by TIM8_TRGO, with no CPU involvement after setup.

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -307,6 +307,42 @@ impl<'d> DacChannel<'d, Async> {
         )
     }
 
+    /// Start a circular DMA output from a static buffer, running indefinitely.
+    ///
+    /// Unlike [`write`](Self::write), this is non-blocking and does not return a future.
+    /// DMA is configured in circular mode and started immediately, so the waveform begins
+    /// as soon as the trigger source fires. Because no `Transfer` guard is returned, DMA
+    /// is never stopped by `Drop` — the excitation runs until the channel itself is dropped.
+    ///
+    /// The buffer must be `'static` because DMA will continuously read it after this
+    /// function returns. Suitable for use in non-async contexts.
+    #[cfg(not(gpdma))]
+    pub fn start_circular_dma(&mut self, data: ValueArray<'static>) {
+        self.info.regs.cr().modify(|w| {
+            w.set_en(self.idx, true);
+            w.set_dmaen(self.idx, true);
+        });
+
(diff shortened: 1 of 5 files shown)


xoviat:
Last time something like this was attempted, I made the author add a ringbuffered.rs (this was the timer mod). I think I should do the same here: we have ringbuffered.rs in the timer mod, the adc mod, the uart mod, and we should also add it to the dac mod. A one-shot circular transfer is much less useful.

Jacke-debug:
> Last time something like this was attempted, I made the author add a ringbuffered.rs (this was the timer mod). I think I should do the same here: we have ringbuffered.rs in the timer mod, the adc mod, the uart mod, and we should also add it to the dac mod. A one-shot circular transfer is much less useful.

The ring buffer uses [u32] buffers rather than [u16] for 12-bit samples, which is slightly wasteful (2× memory). This is a consequence of a limitation in WritableRingBuffer<W>: it enforces the same data width for both memory and peripheral. The DAC however requires 32-bit DMA writes to its registers, so we need W = u32 to get the right peripheral transfer size.

The existing DacChannel::write() handles this correctly, it accepts &[u16] from the user but configures DMA with MW=u16 and PW=u32 (separate memory and peripheral widths). Making the ring buffer behave the same way would require either extending WritableRingBuffer::new to accept a separate peripheral data size, or adding a new variant. That's a broader DMA layer change affecting other users of WritableRingBuffer...

xoviat:
> Making the ring buffer behave the same way would require either extending WritableRingBuffer::new to accept a separate peripheral data size, or adding a new variant. That's a broader DMA layer change affecting other users of WritableRingBuffer...

Backwards compatibility is not a valid concern in embassy-stm32 for methods not implementing a trait defined outside of the crate. However, I'll merge this and fix this myself, or you can.

PR #6201, opened 2026-05-29
stm32/spi: optimize transfer_words

xoviat:
autouse the full fifo capacity and use builtin iterators.

### DIFF
--- a/embassy-stm32/src/spi/mod.rs
+++ b/embassy-stm32/src/spi/mod.rs
@@ -1312,35 +1312,52 @@ fn check_error_flags(sr: regs::Sr, ovr: bool) -> Result<(), Error> {
     Ok(())
 }
 
+fn check_tx_ready(regs: Regs, ovr: bool) -> Result<bool, Error> {
+    let sr = regs.sr().read();
+
+    check_error_flags(sr, ovr)?;
+
+    #[cfg(not(any(spi_v4, spi_v5, spi_v6)))]
+    if sr.txe() {
+        return Ok(true);
+    }
+    #[cfg(any(spi_v4, spi_v5, spi_v6))]
+    if sr.txp() {
+        return Ok(true);
+    }
+
+    Ok(false)
+}
+
 fn spin_until_tx_ready(regs: Regs, ovr: bool) -> Result<(), Error> {
     loop {
-        let sr = regs.sr().read();
-
-        check_error_flags(sr, ovr)?;
-
-        #[cfg(not(any(spi_v4, spi_v5, spi_v6)))]
-        if sr.txe() {
-            return Ok(());
-        }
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/spi/mod.rs
```diff
@@ -1414,61 +1431,47 @@ fn finish_dma(regs: Regs) {
...
+                    regs.cr1().modify(|reg| reg.set_cstart(true));
+                }
 
-            if let Some(word_in) = read_ptr(&mut r) {
-                *word_in = transfer_word(regs, word_out)?;
-            } else {
-                transfer_word(regs, word_out)?;
+                w += 1;
             }
-        }
```
Finomnis, line 1458: @xoviat What if we get preempted exactly here, and while this thread is paused the SPI device keeps transmitting? Won't we hit an RX overflow?

Finomnis: ONE way to prevent this would be if we explicitly check that tx + rx fifo content together is less then the RX fifo capacity, if that's possible

Finomnis: Could be fixed by defining a `safe_overlap` amount based on peripheral version and then only pushing into the tx fifo when `w < r + safe_overlap`.

Finomnis: Or another possibility: We initially write `fifo_size` words into the TX fifo, and then write a new word whenever we get an rx word. That way it's guaranteed that we never get an overflow.

xoviat: do you know what the fifo sizes are?

Finomnis: Not for sure... Ai says without querying it at runtime or adding it to spi32-data the only sane thing we can assume is:
- 1byte for v1/v2
- 4 bytes for v3-v6

There seem to be larger ones but only for some chips, not for an entire peripheral version.

from ChatGPT 5.5 extended thinking, because I don't feel like crawling through tons of reference manuals right now, so take it with a grain of salt :)

Finomnis: If you think it's worth the extra effort then we can add a credits counter to track how many bytes are in flight, otherwise the same thing to do imo is 0 words overlap for v1/v2, 1 word overlap otherwise

xoviat: can you discuss more on matrix?

Finomnis: Imo the extra effort of adding a credits counter for 4 bytes is not worth it

Finomnis: Sure, let's do matrix

PR #6202, opened 2026-05-29
Add embassy-boot example for RP2350

tnibler:
What it says on the tin basically^^

The linker scripts are the main difference from the 2040 version since there's not boot2, the rest is similar.


lulf (approved):
Tin looks good, thanks!

PR #6209, opened 2026-05-29
stm32/hsem: replace flags with atomic modify

xoviat:
cc @plorefice 

Does this work or have I made a mistake?

### DIFF
--- a/embassy-stm32/src/hsem/mod.rs
+++ b/embassy-stm32/src/hsem/mod.rs
@@ -208,7 +208,7 @@ impl<'a, T: Instance> HardwareSemaphoreChannel<'a, T> {
             .write(|w| w.set_isc(self.index.into(), true));
 
         // Wait for the semaphore interrupt flag
-        while !T::regs().misr(core.to_index().into()).read().misf(self.index.into()) {}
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/hsem/mod.rs
```diff
@@ -202,28 +202,32 @@ impl<'a, T: Instance> HardwareSemaphoreChannel<'a, T> {
...
-        T::state().flag_for(self.index).store(false, Ordering::Release);
+        // Clear the interupt
+        T::regs()
+            .icr(core.to_index().into())
+            .write(|w| w.set_isc(self.index.into(), true));
 
-        let _irq = self.clear_and_enable_interupt(core);
         // Wait for the semaphore interrupt flag
-        while !T::state().flag_for(self.index).load(Ordering::Relaxed) {}
+        while !T::regs().misr(core.to_index().into()).read().misf(self.index.into()) {}
```
plorefice: It doesn't work out of the box (tested on the blinky example for STM32H755), to make it work I needed to change this line to the following:

```suggestion
        while !T::regs().isr(core.to_index().into()).read().isf(self.index.into()) {}
```

However, I'm not sure this is the right fix, I'm not overly familiar with this peripheral.

PR #6216, opened 2026-05-30
Adc to fmac

usbalbin:
Based on #6151

This could use some input on a lot of things, one being the separation of concerns. I will do my questions in the review thingy.

### DIFF
--- a/embassy-stm32/src/adc/mod.rs
+++ b/embassy-stm32/src/adc/mod.rs
@@ -59,8 +59,8 @@ dma_trait!(RxDma, Instance);
 pub struct Exten;
 
 pub struct RegularAdcTrigger<T: Instance> {
-    _trigger: u8,
-    _edge: Exten,
+    pub(crate) _trigger: u8,
+    pub(crate) _edge: Exten,
     _typ: PhantomData<T>,
 }
 
@@ -145,7 +145,7 @@ pub trait BasicAdcRegs {
     type SampleTime: Copy;
 }
 
-trait AdcRegs: BasicAdcRegs {
+pub(crate) trait AdcRegs: BasicAdcRegs {
     const HAS_ERRATA: bool = false;
     fn enable(&self);
     fn start(&self);
@@ -177,7 +177,7 @@ pub trait BasicInstance {
     type Regs: AdcRegs;
 }
 
-trait SealedInstance: BasicInstance {
+pub(crate) trait SealedInstance: BasicInstance {
     fn regs() -> Self::Regs;
     #[cfg(not(any(adc_f1, adc_v1, adc_l0, adc_f3v3, adc_f3v2, adc_g0)))]
     #[allow(unused)]
--- a/embassy-stm32/src/dma/dma_bdma.rs
+++ b/embassy-stm32/src/dma/dma_bdma.rs
@@ -999,15 +999,33 @@ impl<'d> Channel<'d> {
         }
     }
 
-    /// Create a read DMA transfer (peripheral to memory).
-    pub unsafe fn read<'a, W: Word>(
+    /// Create a read DMA transfer (peripheral to memory), using raw pointers.
+    unsafe fn read_inner<'a, MW: Word, PW: Word>(
         &'a mut self,
         request: Request,
-        peri_addr: *mut W,
-        buf: &'a mut [W],
+        peri_addr: *mut PW,
+        buf: *mut [MW],
         options: TransferOptions,
+        mem_inc: Increment,
     ) -> Transfer<'a> {
-        self.read_raw(request, peri_addr, buf, options)
+        let mem_len = buf.len();
+
+        self.configure(
+            request,
+            Dir::PeripheralToMemory,
+            peri_addr as *const u32,
+            buf as *mut MW as *mut u32,
+            mem_len,
+            mem_inc,
+            MW::size(),
+            PW::size(),
+            options,
+        );
+        self.start();
+        Transfer {
+            _wake_guard: self.info().wake_guard(),
+            channel: self.reborrow(),
+        }
     }
 
     /// Create a read DMA transfer (peripheral to memory), using raw pointers.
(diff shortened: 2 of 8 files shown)


embassy-stm32/src/fmac/from_adc.rs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            // Ensure no conversions are ongoing
+            ADC::regs().stop(false);
+            ADC::regs().configure_sequence([((adc_ch.channel(), adc_ch.is_differential()), sample_time)].into_iter());
+
+            ADC::regs().enable();
+
+            // Configure DMA once, reused across all subsequent read() calls.
+            ADC::regs().configure_dma(ConversionMode::Repeated(Some((trigger._trigger, trigger._edge))));
+        }
```
usbalbin: Should we have the ADC related code in the adc module so it can be reused for CORDIC and other peripherals that may want adc readings?

embassy-stm32/src/fmac/from_adc.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        let dma_request = dma_ch.request();
+        let mut dma_channel = dma::Channel::new(dma_ch, irq_not_used);
+        let transfer = unsafe {
+            dma_channel.read_repeated(
```
usbalbin: From what I can tell, we need to keep this around or the transfer will be cancelled. What is the best way to do that? It borrows the dma_channel which is a problem

embassy-stm32/src/adc/mod.rs
```diff
@@ -59,8 +59,8 @@ dma_trait!(RxDma, Instance);
 pub struct Exten;
 
 pub struct RegularAdcTrigger<T: Instance> {
-    _trigger: u8,
-    _edge: Exten,
+    pub(crate) _trigger: u8,
```
usbalbin, line 62: These changes would probably not be needed if we move some things to the adc module as suggested in https://github.com/embassy-rs/embassy/pull/6216/changes/a8da24ee732c8b7b78675838cfacb2a8db39371f#r3329269321

embassy-stm32/src/dma/dma_bdma.rs
```diff
@@ -1038,6 +1028,46 @@ impl<'d> Channel<'d> {
...
+        &'a mut self,
+        request: Request,
+        peri_addr: *mut W,
+        buf: &'a mut [W],
+        options: TransferOptions,
+    ) -> Transfer<'a> {
+        self.read_raw(request, peri_addr, buf, options)
+    }
+
     /// Create a write DMA transfer (memory to peripheral).
```
xoviat, line 1041: The only change to this file should be this function, placed write before `write`:

```
    /// Create a read DMA transfer (peripheral to memory), writing the same value repeatedly.
    pub unsafe fn read_repeated<'a, W: Word>(
        &'a mut self,
        request: Request,
        repeated: &'a mut W,
        count: usize,
        peri_addr: *mut W,
        options: TransferOptions,
    ) -> Transfer<'a> {
        assert!(count > 0 && count <= 0xFFFF);

        self.configure(
            request,
            Dir::PeripheralToMemory,
            peri_addr as *const u32,
            repeated as *const W as *mut u32,
            count,
            Increment::None,
            W::size(),
            W::size(),
            options,
        );
        self.start();
        Transfer {
            _wake_guard: self.info().wake_guard(),
            channel: self.reborrow(),
        }
    }
```

usbalbin: Are we allowed to have references to peripherals? The "mem" in the peri->mem is the FMAC

```
repeated: &'a mut W,
```

embassy-stm32/src/fmac/from_adc.rs
xoviat: This function should go in the g4.rs file in the adc folder initially (it can be moved out as the shape becomes more defined), and should return a struct similar to the configured sequence struct (except it will be slightly different) and that struct can be defined in the g4.rs file as well.

embassy-stm32/src/fmac/from_adc.rs
xoviat: use `unchecked_extend_lifetime()` and store the transfer in the struct.

embassy-stm32/src/dma/dma_bdma.rs
xoviat: Good point. make it `read_raw_repeated` and accept a pointer instead.

embassy-stm32/src/fmac/from_adc.rs
usbalbin: Should the new type be public, and the user pass it to FromAdc::new, or does the user just pass a regular Adc which FromAdc::new turns into the new type?

usbalbin: (from matrix chat)
> Albin Hedman: I think we can have a struct called ConfiguredTransfer ? defined in g4.rs that **contains the dma transfer** (with unchecked_extended_lifetime) and FromAdc can have one? struct field that is just that.

I guess that forces us into `user just pass a regular Adc which FromAdc::new turns into the new type?`

xoviat: The idea is that `FromAdc` takes a `&mut Adc` in it's constructor, which then calls `pub(crate) adc.configure_transfer`, defined in `g4.rs`, which the returns `ConfiguredTransfer { Transfer<'_> }`, which then the `FromAdc` constructor returns `FromAdc { transfer: ConfiguredTransfer { transfer: dma::Transfer } }`

xoviat: The `dma::Transfer` represents a configured dma transfer and `ConfiguredTransfer` represents a configured adc transfer (which must contain a configured dma transfer) and then `FromAdc` represents any register reads/writes to the FMAC registers.

usbalbin:
This has not yet been tested

PR #6220, opened 2026-05-31
Cleanup FMAC

usbalbin:
I get lots of weird lifetime issues

### DIFF
--- a/embassy-stm32/src/adc/g4.rs
+++ b/embassy-stm32/src/adc/g4.rs
@@ -474,9 +474,9 @@ impl<'d, T: DefaultInstance> Adc<'d, T> {
             dma_channel
                 .read_raw_repeated(
                     dma_request,
-                    T::regs().data(),
-                    1,
                     dst,
+                    1,
+                    T::regs().data(),
                     dma::TransferOptions {
                         priority: dma::Priority::VeryHigh,
                         circular: true,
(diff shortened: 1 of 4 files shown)


xoviat:
@usbalbin fixed

embassy-stm32/src/adc/g4.rs
```diff
@@ -445,14 +445,17 @@ impl<'d, T: DefaultInstance> Adc<'d, T> {
...
+        irq: impl crate::interrupt::typelevel::Binding<D::Interrupt, crate::dma::InterruptHandler<D>> + 'd,
+        sequence: impl ExactSizeIterator<Item = (&'adc mut AnyAdcChannel<'ch, T>, SampleTime)>,
         trigger: RegularAdcTrigger<T>,
-        dma_ch: embassy_hal_internal::Peri<'d, D>,
         dst: *mut u16,
-        irq: impl crate::interrupt::typelevel::Binding<D::Interrupt, crate::dma::InterruptHandler<D>> + 'd,
-    ) -> ConfiguredTransfer<'d, T::Regs> {
+    ) -> ConfiguredTransfer<'adc, T::Regs>
+    where
+        'ch: 'adc,
```
usbalbin, line 458: So we decouple the adc's potentially shorter lifetime from the others and let the result only live as long as the adc?

usbalbin:
Tested on a nucleo-g474 and the values after the filter does look a lot like the ones at the ADC(which given a stable input, is a good thing). I used the temp channel as input just for simplicity but I do not know if there is some embassy way to convert it into something useful

usbalbin:
Thanks!

PR #6221, opened 2026-05-31
rp/gpio: make Flex::is_set_as_output public

tari:
This status is useful for Flex pins, indirectly exposed by `toggle_set_as_output`, and the doc comment was wrong so it looks like this was unintentionally private.


leftger:
@CBJamo this might be another PR you should look at. Dirbaio used it as a good example of an exception to unifying inconsistencies across HALs

PR #6222, opened 2026-05-31
mspm0: add HFXT + MCLK source config (G350x)

ozongzi:
Replaces the `// TODO: OSC configuration` placeholder in Config with a `clock` submodule that brings up the external HFXT crystal on PA5/PA6 and optionally switches MCLK over to HSCLK <- HFCLK.

Scope:
- Gated on the existing `mspm0g350x` family cfg flag.
- Bench-validated on MSPM0G3507PM with a 40 MHz crystal driving MCLK.
- Same register layout per TI TRM applies to MSPM0G3506; not bench- verified there, so reviewers can widen the gate at their discretion.

Out of scope (left for follow-ups):
- SYSPLL routing as an HSCLK source.
- MDIV and UDIV dividers.
- Other families.

The `embassy-time` driver runs on LFCLK (32.768 kHz, independent of MCLK), so switching MCLK does not skew Timer durations.

A `clock_hfxt` example in `examples/mspm0g3507` demonstrates the new Config fields with a 1 Hz blink.

### DIFF
--- a/embassy-mspm0/src/clock.rs
+++ b/embassy-mspm0/src/clock.rs
@@ -1,0 +1,158 @@
+//! Clock-tree configuration for MSPM0 devices.
+//!
+//! `embassy-mspm0` boots with the chip's reset-state clock tree: MCLK is
+//! sourced from the internal SYSOSC at 32 MHz. For applications that need
+//! tighter frequency accuracy (no ±1 % SYSOSC drift) or higher CPU rates
+//! (up to 80 MHz with SYSPLL on supported devices, up to 40 MHz from HFXT)
+//! this module brings up the external crystal and switches MCLK over to it.
+//!
+//! Coverage in this version is intentionally narrow:
+//!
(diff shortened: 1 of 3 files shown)


embassy-mspm0/src/clock.rs
```diff
@@ -0,0 +1,158 @@
...
+fn enable_hfxt(cfg: &HfxtConfig) {
+    let s = pac::SYSCTL;
+
+    // HFXIN/HFXOUT are dedicated alternate functions on PA5/PA6. Writing the
+    // default (cleared) PINCM disables both the digital input buffer and
+    // pull resistors so the analog crystal path is selected. PA5 → PINCM 9,
+    // PA6 → PINCM 10. These indices are fixed on G350x and don't go through
+    // the gpio_pincm() lookup, which translates pin-number IDs.
+    pac::IOMUX.pincm(9).write_value(Default::default());
+    pac::IOMUX.pincm(10).write_value(Default::default());
```
i509VCB, line 112: These would probably differ by chip (G518x now exists). I do think there is a better way to do this potentially.

PR #6225, opened 2026-05-31
stm32: timer: add generate_capture_compare_event for advanced timers

Jacke-debug:
Exposes CCxG in EGR on AdvancedInstance4Channel timers, allowing software generation of capture/compare events. The bit is automatically cleared by hardware.

### DIFF
--- a/embassy-stm32/src/timer/low_level.rs
+++ b/embassy-stm32/src/timer/low_level.rs
@@ -1337,11 +1337,11 @@ impl<'d, T: AdvancedInstance4Channel> Timer<'d, T> {
         self.regs_advanced().egr().write(|r| r.set_bg(n, true));
     }
 
-    /// Generate a software capture/compare event on channel `n` (0-based).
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/timer/low_level.rs
```diff
@@ -1337,6 +1337,13 @@ impl<'d, T: AdvancedInstance4Channel> Timer<'d, T> {
         self.regs_advanced().egr().write(|r| r.set_bg(n, true));
     }
 
+    /// Generate a software capture/compare event on channel `n` (0-based).
+    ///
+    /// Sets CCxG in EGR. The bit is automatically cleared by hardware.
+    pub fn generate_capture_compare_event(&self, n: usize) {
+        self.regs_advanced().egr().write(|r| r.set_ccg(n, true));
```
xoviat: Take the 'Channel' enum as a param?

PR #6226, opened 2026-06-01
wpan: use async pka

xoviat:
Not tested on hardware, but by inspection, it should work.


leftger (approved):
Thanks for fixing this 😄

xoviat:
`ble_peripheral_connect` and `ble_secure_connect` observered no regressions.

xoviat:
ble_secure_connect does not work on git main or this

PR #6228, opened 2026-06-01
[MCXA] add irq trace option

diondokter:
This allows us to trace interrupts that are made with `bind_interrupts` in a similar fashion as how we can trace the executor.

### DIFF
--- a/embassy-mcxa/Cargo.toml
+++ b/embassy-mcxa/Cargo.toml
@@ -147,6 +147,9 @@ rt = []
 # Enable perf counters
 perf = []
 
+# Enable irq tracing
+trace = []
+
 # Enable the custom executor platform implementation, which allows for low-power modes.
 #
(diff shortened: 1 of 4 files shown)


embassy-mcxa/Cargo.toml
```diff
@@ -147,6 +147,9 @@ rt = []
 # Enable perf counters
 perf = []
 
+# Enable irq tracing
+trace = []
```
felipebalbi, line 151: Maybe call the feature trace-irq for clarity? Not a strong opinion, though.

felipebalbi (approved):

embassy-mcxa/Cargo.toml
diondokter: It could enable more tracing in the future in theory, so that's why I kept it general

PR #6236, opened 2026-06-02
embassy-rp: Add runtime baud rate reconfiguration for PIO UART

HenriqueDomiciano:
Adding the possibility to change PioUartRx and PioUartTx RP2xxx baud rate in runtime, by changing the clock divisor on state machine..

It's my first PR so some support may be needed.

### DIFF
--- a/embassy-rp/src/pio_programs/uart.rs
+++ b/embassy-rp/src/pio_programs/uart.rs
@@ -77,6 +77,18 @@ impl<'d, PIO: Instance, const SM: usize> PioUartTx<'d, PIO, SM> {
     pub async fn write_u8(&mut self, data: u8) {
         self.sm_tx.tx().wait_push(data as u32).await;
     }
+
+
+    /// Change Baud Rate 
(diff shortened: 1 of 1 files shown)


HenriqueDomiciano:
Opening the pull request because i needed on my project and i think it can be useful for other users to, if you guys think it's worth it your time and follows the expected architecture would be pleased to help

wowkster:
Good addition, but I think `set_baudrate` might be more inline with other APIs and would match the [native UART driver](https://github.com/embassy-rs/embassy/blob/905ca9debeac4037eac6cb50d4ee0389022a5482/embassy-rp/src/uart/mod.rs#L1052). Merging the imports and some more concise doc comments would probably also be appreciated.

HenriqueDomiciano:
@wowkster Did the following 
1. Change the function name 
2. Made the doc comment more concise
3. Regarding the merge in imports, the rustfmt of the embassy devtools does not allow me to use the imports merged i tried to set them but every time i ran it, with the imports merged, it just broke.

PR #6238, opened 2026-06-02
stm32: Fix half-duplex usart

chemicstry:
This PR fixes multiple issues:
- When reconfiguring baudrate via `UartTx::set_config`, the duplex config is lost and driver reverts to full-duplex mode
- Using `RingBufferedRx` simultaneously with TX would discard bytes in the TX FIFO
- Unrelated typo in `config.de_deassertion_time`

I would also like to consult about the `OutputConfig` struct which currently can not express some configuration options like push-pull with pull down (useful with half-duplex and invert_tx). Adding all permutations of `OutputType` and `Pull` seems unnecessary, so I propose to remove `OutputConfig` struct and let user specify `AfType` directly for all rx, tx, rts, cts, de pins. This would also allow selecting IO speed. The downside is that this would allow invalid config options like rx pin in output mode. Another option would be to make `OutputConfig` into a struct with mode, speed, pull.

### DIFF
--- a/embassy-stm32/src/usart/mod.rs
+++ b/embassy-stm32/src/usart/mod.rs
@@ -1780,6 +1780,14 @@ fn configure(
         return Err(ConfigError::RxOrTxNotEnabled);
     }
 
+    // The `duplex` config field is private and defaults to `Full`, so a user-supplied
+    // `Config` passed to `set_config`/`reconfigure` cannot express half-duplex. Combine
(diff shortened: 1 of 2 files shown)


xoviat:
> Adding all permutations of OutputType and Pull seems unnecessary

It's the best solution that can be merged at this point without a larger rethink. I am not necessarily opposed to this rethink, but I am focused on other things at the moment.

PR #6243, opened 2026-06-03
stm32/wba: bundle low-power, HSEM layout

leftger:
## Summary
- improve WBA low-power sleep timing/stop-entry robustness in the core low-power path
- add WBA 16-channel HSEM layout support for correct semaphore addressing

## Included source branches/commits
- `feat/wba-main-lowpower-hsem` (low-power + HSEM commits)

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -196,6 +196,9 @@ mod platform {
         });
         #[cfg(stm32wba)]
         crate::pac::PWR.sr().modify(|w| w.set_cssf(true));
+        #[cfg(stm32wba)]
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/low_power.rs
```diff
@@ -323,6 +333,8 @@ fn configure_pwr(cs: CriticalSection) {
 
     get_scb().clear_sleepdeep();
     platform::clear_flags();
+    #[cfg(stm32wba)]
+    platform::clear_wakeup_flags();
```
xoviat: Why does this need another function than `clear_flags`?

leftger: You're absolutely right. Let me fix it.

PR #6247, opened 2026-06-03
stm32: consolidate PR1-PR3 for PKA, SAES, and RNG API updates

leftger:
## Summary
- consolidate PR1-PR3 into one branch for coordinated cryptography/peripheral API review
- extend PKA ECC point support to 640-bit, expose SAES key-mode + async parity APIs, and add configurable RNG initialization policy APIs
- remove the newly added RNG low-power disable helper for now based on review feedback (needs refcount design first)
- keep commit history linear and scoped to `embassy-stm32`

## Included source branches
- `feat/stm32-pr1-pka-ecc-640bit`
- `feat/stm32-pr2-saes-keymode-async`
- `feat/stm32-pr3-rng-config-api`

### DIFF
--- a/embassy-stm32/src/rng.rs
+++ b/embassy-stm32/src/rng.rs
@@ -268,12 +268,6 @@ impl<'d, T: Instance> Rng<'d, T> {
         let _ = self.next_u32();
     }
 
-    #[cfg(not(rng_v1))]
-    /// Disable RNG and wait until it is idle (safe before low-power entry).
-    pub fn disable_for_low_power(&mut self) {
-        T::regs().cr().modify(|reg| reg.set_rngen(false));
-    }
-
     /// Try to recover from a seed error.
     pub fn recover_seed_error(&mut self) {
         self.reset();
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/rng.rs
```diff
@@ -157,6 +268,12 @@ impl<'d, T: Instance> Rng<'d, T> {
         let _ = self.next_u32();
     }
 
+    #[cfg(not(rng_v1))]
+    /// Disable RNG and wait until it is idle (safe before low-power entry).
+    pub fn disable_for_low_power(&mut self) {
+        T::regs().cr().modify(|reg| reg.set_rngen(false));
+    }
```
xoviat: This will not be sufficient without modification of the refcounts. I think this needs some thought.

leftger: OK I'll remove this from the PR.

leftger: It's been removed from the PR.

PR #6248, opened 2026-06-03
embassy-net: Add auto-icmp-echo-reply to default features

adamgreig:
The feature [was added to smoltcp](https://github.com/smoltcp-rs/smoltcp/pull/1106) in the recent 0.13.0 release, and embassy-net was updated to use 0.13.0 in f0ff2ac131d2979f0bae0caa5ad7d28e228ff269. Support for the feature was earlier added to embassy-net in 3540e528f2490569bf6cb3cb134a0c69b0ca5adb.

In smoltcp the feature is enabled by default to preserve the previous behaviour of replying to ICMP echo requests, but possible to disable using `default-features=false`. When the equivalent feature was added to embassy-net it wasn't enabled by default, so the behaviour changed to default not reply to pings. I think that was probably unintentional and we should match the smoltcp behaviour of replying by default.


lulf (approved):
Got bitten by this too, but didn't realize smoltcp had it enabled by default. Agreed we should do the same in embassy-net then.

PR #6252, opened 2026-06-03
Fix PIO Common lifetime issue with cyw43::PioSpi

### LINKED ISSUE #6251: `cyw43_pio::PioSpi::new` does not tie `PioSpi` lifetime to the borrowed `Common`
I believe there may be a lifetime/resource-ownership bug in `cyw43_pio::PioSpi::new`.

The current constructor has this general shape:

```rust
pub fn new(
    common: &mut Common<'d, PIO>,
    sm: StateMachine<'d, PIO, SM>,
    ...
) -> PioSpi<'d, PIO, SM, DMA>
```

The returned `PioSpi` is parameterized by `'d`, but not by the lifetime of the actual `&mut Common<'d, PIO>` borrow passed to `new()`.

That means the API appears to allow a value like this to escape:

```rust
fn make_spi(...) -> PioSpi<'static, PIO0, 0, DMA_CH0> {
    let mut pio = Pio::new(...);

    PioSpi::new(&mut pio.common, pio.sm0, ...)
}
```

At the type level, the returned `PioSpi` is `'static`. But operationally, it appears to depend on the `Common` value inside `pio`, which is dropped when `make_spi()` returns.

I ran into this through `cyw43` on a Pico 2 W. If I create the `PioSpi` inside a bringup function, pass it to `cyw43::new(...)`, spawn the CYW43 runner, and then return only a wrapper around `cyw43::Control`, subsequent calls such as this hang:

```rust
control.gpio_set(0, false).await;
```

However, if I move the exact same call into the bringup function before the local `Pio` value is dropped, it succeeds.

That suggests that `PioSpi` remains dependent on the continued validity of the `Common`/PIO ownership state, but the current type signature does not encode that dependency.

I would expect the returned type to be tied either to the borrow of `Common`, for example conceptually:

```rust
pub fn new<'a>(
    common: &'a mut Common<'d, PIO>,
    ...
) -> PioSpi<'a, 'd, PIO, SM, DMA>
```

or for `PioSpi` to own whatever PIO resource guard/state is required to remain valid after construction.

As written, `PioSpi::new()` appears to treat `Common` as construction-only input, while runtime behavior suggests that dropping `Common` later can invalidate the transport. This makes it possible to construct code that compiles cleanly but later hangs in CYW43 operations.

marti4d:
Currently, `cyw43::PioSpi::new()` borrows a `&mut Common<'d, PIO>` and logically requires that the `Common` and the underlying `PIO` outlive it, but it doesn't currently enforce that - the lifetime of the new `PioSpi` only depends on `'d` - the lifetime of the internal borrows of `PIO`.

This patch is a bit of a conservative approach - It holds a `&mut Common` the entire time it's alive. This fixes the issue, although a more permissive patch could be imagined that programs it with a `&mut Common` but then only holds a `&Common` afterwards. But expressing that might be awkward, and I don't know how common it might be to do something like that.

Fixes #6251

### DIFF
--- a/cyw43-pio/src/lib.rs
+++ b/cyw43-pio/src/lib.rs
@@ -15,12 +15,13 @@ use fixed::FixedU32;
 use fixed::types::extra::U8;
 
 /// SPI comms driven by PIO.
-pub struct PioSpi<'d, PIO: Instance, const SM: usize, DMA: Channel> {
+pub struct PioSpi<'a, 'd, PIO: Instance, const SM: usize, DMA: Channel> {
     cs: Output<'d>,
     sm: StateMachine<'d, PIO, SM>,
     irq: Irq<'d, PIO, 0>,
     dma: Peri<'d, DMA>,
     wrap_target: u8,
+    common: core::marker::PhantomData<&'a mut Common<'d, PIO>>,
 }
 
 /// The default clock divider that works for Pico 1 and 2 W. As well as the RM2 on rp2040 devices.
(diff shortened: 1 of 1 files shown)


marti4d:
Nevermind -- I guess I made this against an older version of the crate. It still has the same issue, though.

Issue #6254, opened 2026-06-03
`read_latest` reads 0 elements when ring buffer is full


pufmat:
Here is an example with ring buffered ADC:
```rs
let mut adc: RingBufferedAdc<_> = ... ;
let mut buffer = [0u16; 128];

adc.start();
loop {
    Timer::after_secs(1).await;
    info!("{}", adc.read_latest(&mut buffer));
}
```
Assuming 1 second is enough to collect enough samples, it should always read 128 latest elements. However, current behavior is different, and it reads 0 elements, because it detects overrun:
https://github.com/embassy-rs/embassy/blob/3d7b6d1c287742638f832f1f675a13866414f697/embassy-stm32/src/dma/ringbuffer/mod.rs#L281-L284

PR #6255, opened 2026-06-03
stm32/dma: Fix `read_latest` in ring buffer

pufmat:
This is my proposed fix for issue #6254. 

When an overrun is detected it sets read and write indexes such that the entire buffer can be read.

~~Since `read_latest` and `read_raw` were doing nearly the same thing, to deduplicate code, I modified `read_raw` so that it can be reused by `read_latest` function. I chose `read_raw` implementation over original `read_latest` because in my opinion it was cleaner and easier to reason about.~~ Their difference is actually important here.

Tested on STM32F411CE.

### DIFF
--- a/embassy-stm32/src/dma/ringbuffer/mod.rs
+++ b/embassy-stm32/src/dma/ringbuffer/mod.rs
@@ -167,13 +167,9 @@ impl<'a, W: Word> ReadableDmaRingBuffer<'a, W> {
     /// Error is returned if the portion to be read was overwritten by the DMA controller,
     /// in which case the rinbuffer will automatically reset itself.
     pub fn read(&mut self, dma: &mut impl DmaCtrl, buf: &mut [W]) -> Result<(usize, usize), Error> {
-        fence(Ordering::Acquire);
-
-        let available = self.sync_len(dma).inspect_err(|_e| {
+        self.read_raw(dma, buf).inspect_err(|_e| {
             self.reset(dma);
-        })?;
-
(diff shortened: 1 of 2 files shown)


leftger:
Hey @pufmat — thanks for catching this. As the original author, you're right that "return 0, catch up next call" doesn't match the doc's "only the most recent data is returned." Fix aligns implementation with contract.

I traced the bookkeeping and the `complete_count += 1` is correct: after `sync_len()` returns `Overrun`, setting `read_index = write_index` and bumping `write_index.complete_count` by 1 restores `diff = cap` exactly — matching the fact that the ring holds one iteration of samples. The end-of-function advance by `cap` then lands `read == write`, so the next `sync_len` reports only fresh post-read samples. Makes sense.

Two asks:

1. **Add a test for overrun + `alignment > 1`.** After the fix, `read_index.pos = cap - NDTR`, which isn't guaranteed frame-aligned. Existing overrun test uses `alignment = 1`; the alignment test doesn't hit the overrun path. I worked through cap=16, alignment=4, `write_index.pos=6` by hand and the front_skip/tail-discard does return intact frames newest-first, but it's unverified.

2. **Drop the `read_raw` refactor.** In agreement with your edited description, the dedup motivation didn't pan out. Since `read_latest` still has its own implementation, extracting `sync_len` out of `read_raw` and moving the `fence` up is no-op churn. Easier to review and backport if scoped to `read_latest` + the test.

leftger (approved):

PR #6260, opened 2026-06-04
embassy-mcxa: lpuart: add LpuartBbq::split() for separate TX/RX halves

### LINKED ISSUE #5816: Implement split function for LPUART BBQ
Implement split function for LPUART BBQ

bramsdell-ms:
Adds a `split()` method to `LpuartBbq` so the TX and RX halves can be moved into separate tasks, matching the pattern used by other embassy UART wrappers.

Closes #5816.

Commit by @SanjeevKumm, opening on his behalf.

### DIFF
--- a/embassy-mcxa/src/lpuart/bbq.rs
+++ b/embassy-mcxa/src/lpuart/bbq.rs
@@ -168,8 +168,6 @@ impl From<BbqConfig> for super::Config {
 
 /// A `bbqueue` powered buffered Lpuart
 pub struct LpuartBbq {
-    // TODO: Don't just make these pub, we don't *really* handle dropping/recreation
-    // of separate parts at the moment.
     /// The TX half of the LPUART
     tx: LpuartBbqTx,
     /// The RX half of the LPUART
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/lpuart/bbq.rs
```diff
@@ -477,6 +477,11 @@ impl LpuartBbq {
         (rx, tx)
     }
 
+    /// Split the LpuartBbq into separate TX and RX halves
+    pub fn split(self) -> (LpuartBbqTx, LpuartBbqRx) {
+        (self.tx, self.rx)
+    }
+
```
diondokter, line 482: If we do this, we first need to implement drop for them properly: https://github.com/bramsdell-ms/embassy/blob/f5a07b63f89b6da64e238b5acec85f5c043224ff/embassy-mcxa/src/lpuart/bbq.rs#L171-L172

There are teardown functions which seem to drop everything safely. So if you impl Drop for the types (which is seemingly also missing from the combined uart!) and have them call teardown, I think that would make it work

diondokter (changes requested):

embassy-mcxa/src/lpuart/bbq.rs
bramsdell-ms: Done in ff5305a. Added `Drop` for both `LpuartBbqTx` and `LpuartBbqRx` by extracting the side-effect body of each `teardown(self)` into a `teardown_inner(&mut self)` helper. The owned `teardown(self)` now calls the helper, builds `BbqHalfParts` (using `core::ptr::read` + `core::mem::forget` to safely move out the non-Copy `_*_pins` / `_wg` fields), and `Drop` simply calls the helper and discards what comes back so the field destructors can run as usual.

For the combined `LpuartBbq` I went with no explicit `Drop`: its fields' new `Drop` impls cover everything, and `LpuartBbq::teardown(self)` continues to work via destructuring because the wrapper itself still has no `Drop`. Let me know if you'd prefer a no-op `Drop` there for documentation purposes.

I also updated the `new_tx_half` / `new_rx_half` doc comments to reflect the new behavior (buffer is still leaked on Drop, but peripheral / DMA / pin / clock gate are now released) and fixed a TX/RX copy-paste in the RX-half doc.

Built clean for both `mcxa5xx` and `mcxa2xx` features.

diondokter (approved):
Thanks, looks good

PR #6262, opened 2026-06-04
embassy-mcxa: i2c: enable open-drain on I2C pins and recover from bus errors

bramsdell-ms:
Two related fixes for the LPI2C controller:

- `impl_lpi2c_pin!` now sets `ODE=1`, so SDA/SCL are driven open-drain as the spec requires (they were push-pull before, which fights a slow-rising bus or a peer pulling low).
- Adds `recover_from_error()` and calls it from the async paths after a NACK/FifoError: flushes the FIFOs and re-arms the per-event interrupts so the next transaction starts clean instead of leaving the peripheral wedged.

Commit by @gjpmsft, opening on his behalf.

### DIFF
--- a/embassy-mcxa/src/i2c/mod.rs
+++ b/embassy-mcxa/src/i2c/mod.rs
@@ -153,6 +153,10 @@ macro_rules! impl_lpi2c_pin {
                 self.set_drive_strength(crate::gpio::DriveStrength::Double.into());
                 self.set_function(crate::pac::port::Mux::$fn);
                 self.set_enable_input_buffer(true);
+                // I2C requires open-drain so multiple devices can share the bus.
+                self.pcr_reg().modify(|w| {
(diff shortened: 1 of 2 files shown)


embassy-mcxa/src/i2c/mod.rs
```diff
@@ -153,6 +153,10 @@ macro_rules! impl_lpi2c_pin {
                 self.set_drive_strength(crate::gpio::DriveStrength::Double.into());
                 self.set_function(crate::pac::port::Mux::$fn);
                 self.set_enable_input_buffer(true);
+                // I2C requires open-drain so multiple devices can share the bus.
+                self.pcr_reg().modify(|w| {
+                    w.set_ode(crate::pac::port::Ode::Ode1);
+                });
```
diondokter, line 159: Yikes! Weird how we never noticed that.

diondokter (approved):

PR #6263, opened 2026-06-04
embassy-mcxa: i2c/target: improve async target robustness

bramsdell-ms:
Four small fixes for the async LPI2C target driver, each useful on its own:

1. **use per-operation interrupt enables in async mode** — the target was leaving a broad interrupt mask enabled, which caused spurious wakeups and races where stale events were serviced under the wrong waker. Each `await` now enables only the interrupts it needs and clears them on completion.
2. **drain unconsumed RX FIFO bytes after `respond_to_write`** — if the host writes more bytes than the caller's buffer holds, the leftovers used to sit in the FIFO and corrupt the next read. Drain them before returning.
3. **add `preload_tx()` helper** — factors the "fill TX before the host starts clocking" sequence (including the `SCR.RTF = NowEmpty` reset) used by both read paths.
4. **pad TX with zeros until controller ends read** — if the host reads past what's been prepared, the peripheral used to stall SCL until something fed TX. Now we pad with zeros until the controller issues STOP / repeated START, matching what most masters expect.

Each commit builds and is independently reviewable.

### DIFF
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -414,21 +414,6 @@ impl<'d, M: Mode> I2c<'d, M> {
         });
     }
 
-    /// Pre-load the TX FIFO with default data so the hardware has
-    /// something to send if the controller starts a read before the
-    /// firmware is ready with a real response.
-    pub fn preload_tx(&mut self, data: &[u8]) {
-        // Clear the TX FIFO first
-        critical_section::with(|_| {
-            self.info.regs().scr().modify(|w| {
-                w.set_rtf(ScrRtf::NowEmpty);
-            });
-        });
-        for &byte in data {
-            self.info.regs().stdr().write(|w| w.set_data(byte));
-        }
-    }
-
     fn clear_status(&self) {
         self.info.regs().ssr().write(|w| {
             w.set_rsf(true);
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -1033,6 +1033,29 @@ impl<'d> AsyncEngine for I2c<'d, Async> {
             }
         }
 
+        // The controller may read more bytes than the buffer contains.
+        // Keep feeding zeros until the controller NACKs or sends STOP,
+        // otherwise the hardware clock-stretches indefinitely.
+        loop {
+            self.info
+                .wait_cell()
+                .wait_for(|| {
+                    self.enable_tx_ints();
+                    let ssr = self.info.regs().ssr().read();
+                    ssr.tdf() || ssr.sdf() || ssr.rsf()
+                })
+                .await
+                .map_err(|_| IOError::Other)?;
+
+            let ssr = self.info.regs().ssr().read();
+            if ssr.sdf() || ssr.rsf() {
+                self.reset_fifos();
+                break;
+            } else {
+                self.info.regs().stdr().write(|w| w.set_data(0x00));
+            }
+        }
+
         Ok(count)
     }
 
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -859,7 +859,8 @@ impl<'d, M: AsyncMode> I2c<'d, M>
 where
     Self: AsyncEngine,
 {
-    fn enable_ints(&self) {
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/i2c/target.rs
```diff
@@ -1050,6 +1050,13 @@ impl<'d> AsyncEngine for I2c<'d, Async> {
             }
         }
 
+        // Drain any remaining bytes from the RX FIFO that the caller
+        // didn't consume. Prevents FIFO overflow and bus lock-up
+        // (SCL clock-stretch) on subsequent transactions.
+        while self.info.regs().ssr().read().rdf() {
+            let _ = self.info.regs().srdr().read().data();
+        }
+
```
diondokter: This is one way to do it...
Another (maybe better?) way is to return an error.

Because this change just silently truncates the bytes.
Whatever we end up choosing, we should reflect that in the docs of the public function.

@felipebalbi ?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -1033,6 +1033,29 @@ impl<'d> AsyncEngine for I2c<'d, Async> {
             }
         }
 
+        // The controller may read more bytes than the buffer contains.
+        // Keep feeding zeros until the controller NACKs or sends STOP,
+        // otherwise the hardware clock-stretches indefinitely.
```
diondokter, line 1030: This behavior should be documented in the public-facing function

diondokter (changes requested):

embassy-mcxa/src/i2c/target.rs
```diff
@@ -414,6 +414,21 @@ impl<'d, M: Mode> I2c<'d, M> {
...
+        // Clear the TX FIFO first
+        critical_section::with(|_| {
+            self.info.regs().scr().modify(|w| {
+                w.set_rtf(ScrRtf::NowEmpty);
+            });
+        });
+        for &byte in data {
+            self.info.regs().stdr().write(|w| w.set_data(byte));
+        }
+    }
```
felipebalbi: not sure this is a good idea. The target can clock stretch if it's not ready. Did you find any error without this?

embassy-mcxa/src/i2c/target.rs
felipebalbi: I agree, let's document it with proper rustdoc comments.

embassy-mcxa/src/i2c/target.rs
felipebalbi: We could enrich the return value here:

```rust
pub enum WrieResult {
    Complete(usize),
    WantsMore(usize),
}
```

Then we can correctly tell the application layer that the controller wants more data.

embassy-mcxa/src/i2c/target.rs
bramsdell-ms: Yeah, this looks like it got caught in the gears on my end as a speculative fix that didn't pan out, my bad, removing it.

embassy-mcxa/src/i2c/target.rs
bramsdell-ms: Added rustdoc on the public `async_respond_to_read` documenting the zero-padding in commit e99d0fbc3.

embassy-mcxa/src/i2c/target.rs
bramsdell-ms: Thanks both. I went back and instrumented our downstream consumer and could not reproduce a wedge or any observable benefit from this drain in practice — all our callers use fixed-size HID buffers sized to spec, so we don't hit the buf-full-with-leftover case. The original ADO commit message described a theoretical concern but didn't record a reproducer.

Given that, rather than ship silent truncation now and design the right API around a use case we can't reproduce, I'm dropping this commit from the PR. If we (or anyone) find a real reproducer later, we can come back with Felipe's `WriteResult { Complete(usize), WantsMore }` enum (or equivalent `Result<usize>` call-again semantics) — both designs would also need to stop draining on the buf-full path so bytes remain in the FIFO for the follow-up call.

Force-pushed the rebase: branch is now 4 commits (per-op interrupt enables, add preload_tx, pad TX with zeros, rustdoc) plus the preload_tx removal commit on top from the other thread.

felipebalbi (approved):
LGTM

felipebalbi:
bender run

PR #6278, opened 2026-06-05
[MCXA] i2c/target: report end-of-transfer via ReadStatus/WriteStatus

felipebalbi:
The unconditional zero-padding loop added to async_respond_to_read in 26a8afb30 ("pad TX with zeros until controller ends read") waits for SDF/RSF before returning. In back-to-back transactions, the firmware falls behind: a subsequent async_listen returns the stale first address-match, and async_respond_to_write then waits for RDF that will never fire because the bus has already advanced to the READ phase, so the target deadlocks.

Replace the per-op usize return with richer status enums modelled on embedded-mcu-hal::i2c::target:

  ReadStatus  := Complete(n) | NeedMore(n) | EarlyStop(n)
  WriteStatus := Stopped(n)  | Restarted(n) | BufferFull(n)

Each respond_to_* call now services the transfer to a clean termination point and tells the caller why it ended. NeedMore signals "controller wants more bytes than the buffer holds" so the caller can decide what to send next (more real data, fill bytes, or accept the bus stretching until something else terminates the transfer). No automatic padding.

Applied consistently to blocking, async, and DMA modes:

  * Blocking: busy-loop on tdf|sdf|rsf; map exit cause to a variant.
  * Async: drop the padding loop; one post-buffer wait_for to map to Complete / NeedMore / EarlyStop.
  * DMA: rewrite read_dma_chunk / write_dma_chunk with a dual-wake poll_fn that registers on both the DMA channel's wait_cell and the I2C peripheral's wait_cell. Switch the TX-side TCD from NO_INTERRUPTS to COMPLETE_INTERRUPT so DMA exhaustion can wake the task and surface NeedMore / BufferFull when the controller wants past the chunk. Outer respond_to_* aggregates per-chunk outcomes.

Test:

  tests/mcxa2xx/src/bin/i2c.rs now runs the controller script twice
  back-to-back against the same target peripheral, reborrowed first for
  new_async then for new_async_with_dma, racing the target loop and the
  controller script via embassy_futures::select. A 250 ms warm-up at
  the top of the controller script ensures the target side has armed
  its interrupts before the first START.

  Before: timeout (hang) on the second transaction.
  After:  Test OK in ~511 ms on FRDM-MCX-A266.

All in-tree examples migrated to the new return types.

### DIFF
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -40,11 +40,11 @@
 //!             target::Request::Read(addr) => {
 //!                 // Controller wants to read from us at `addr`
 //!                 buf.fill(0x55);
-//!                 let _count = i2c.blocking_respond_to_read(&buf).unwrap();
+//!                 let _status = i2c.blocking_respond_to_read(&buf).unwrap();
 //!             }
 //!             target::Request::Write(_addr) => {
 //!                 // Controller wants to write to us at `addr`
-//!                 let _count = i2c.blocking_respond_to_write(&mut buf).unwrap();
+//!                 let _status = i2c.blocking_respond_to_write(&mut buf).unwrap();
 //!             }
 //!             target::Request::Stop(_addr) => {
 //!                 // Controller issued a STOP condition for `addr`
@@ -53,7 +53,7 @@
 //!                 // Controller issued a General Call (broadcast write
 //!                 // to address 0x00). Drain the payload via the
 //!                 // normal write-response path.
-//!                 let _count = i2c.blocking_respond_to_write(&mut buf).unwrap();
+//!                 let _status = i2c.blocking_respond_to_write(&mut buf).unwrap();
 //!             }
 //!             target::Request::SmbusAlert => {
 //!                 // Controller issued an SMBus Alert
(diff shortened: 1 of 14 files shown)


embassy-mcxa/src/i2c/target.rs
```diff
@@ -117,6 +119,48 @@ impl From<crate::dma::InvalidParameters> for IOError {
...
+    Complete(usize),
+    /// Buffer was fully consumed but the controller is still asking for
+    /// more bytes (it ACKed the last byte). Caller should call
+    /// `respond_to_read` again with additional bytes, or accept that the
+    /// bus will clock-stretch (with TXDSTALL enabled) until something
+    /// else terminates the transfer.
+    NeedMore(usize),
+    /// Controller issued an early STOP or repeated START before the
+    /// buffer was exhausted.
+    EarlyStop(usize),
```
diondokter, line 141: Do you ever care about this? I think this enum only needs two variants, one for there was enough and one for needmore.
Or am I missing something?

diondokter (approved):
Looks good on a quick pass. Didn't test the changes though

embassy-mcxa/src/i2c/target.rs
felipebalbi: from a "raw" i2c perspective, unlikely to be useful. But there are other protocols built on top of i2c (smbus, pmbus, hidi2c, etc) and those may care. I'm just being thorough, really.

Issue #6143, opened 2026-05-19
net: panic in send_to_with on buffer full


KingCol13:
In this code:

https://github.com/embassy-rs/embassy/blob/6fa07cf2aa341528b0e36b6311b4e6ecfc61b0fc/embassy-net/src/udp.rs#L364-L378

the option `f` can be taken from twice, causing a panic.

`send_to_with` calls [`send_with`](https://github.com/smoltcp-rs/smoltcp/blob/e347a1e2d3ac33c5ce2c0c114e24b85ae23c4897/src/socket/udp.rs#L341), which in turn calls [`enqueue_with_infallible`](https://github.com/smoltcp-rs/smoltcp/blob/e347a1e2d3ac33c5ce2c0c114e24b85ae23c4897/src/storage/packet_buffer.rs#L122). Normally `enqueue_with_infallible` checks the metadata buffer has space before calling the passed closure, however [in this case](https://github.com/smoltcp-rs/smoltcp/blob/e347a1e2d3ac33c5ce2c0c114e24b85ae23c4897/src/storage/packet_buffer.rs#L153) when padding is added to the payload ring buffer to wrap around, an extra padding metadata is also added. If this causes the metadata buffer to be full then the passed closure will be called and `enqueue_with_infallible` will fail, resulting in `send_with` returning `SendError::BufferFull` and causing the `send_to_with` future to be pending. The next time it is polled and the closure is called we attempt to take the option a second time, leading to a panic.

There is identical code for ICMP.

# Reproducible example

I've written an example [here](https://github.com/KingCol13/embassy/blob/send_to_with-panic-reproducer/examples/rp235x/src/bin/ethernet_w6300_udp_panic.rs). I could not reproduce the issue on linux, I suspect the buffers are being completely emptied at each syscall, which would avoid the problem.

# Potential fixes

I'm not sure how best to fix this, things I've come up with so far:
- Constrain `f` some more by either making it `Clone` or `Fn` rather than just `FnOnce`
- Modifying the contract for smoltcp's `Socket::send_with` to only call `f` once it knows there is enough space in all buffers, this could be done fairly easily by re-ordering these lines. https://github.com/smoltcp-rs/smoltcp/blob/e347a1e2d3ac33c5ce2c0c114e24b85ae23c4897/src/storage/packet_buffer.rs#L157-L161. This would make sense for performance since then we avoid unnecessary writes, which is the point of having this zero-copy function in the first place.

Dirbaio:
[...]

PR #6280, opened 2026-06-06
possible fix for UDP send_to_with

robamu:
I am still trying to understand how this code dealt with the Buffer Full condition before.
Wouldn't that always cause a panic because the second take unwrap on a later poll call panics? I managed to make this function panic when trying to maximize the data rate on a Z7
and sending UDP frames in a permanent loop.

### DIFF
--- a/embassy-net/src/udp.rs
+++ b/embassy-net/src/udp.rs
@@ -408,10 +408,10 @@ impl<'a> UdpSocket<'a> {
     /// If the socket's send buffer is too small to fit `max_size`, this method will return `Err(SendError::PacketTooLarge)`
     ///
     /// When the remote endpoint is not reachable, this method will return `Err(SendError::NoRoute)`
-    pub async fn send_to_with<T, F, R>(&mut self, max_size: usize, remote_endpoint: T, f: F) -> Result<R, SendError>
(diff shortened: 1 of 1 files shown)


Dirbaio:
The existing code is correct. If you look at `send_with` in `smoltcp`, it first tries allocating buffer space, and only if successful it calls the closure. So in case of BufferFull error the `f.take()` can't have run, so it'll still be available for the retry.

If you're seeing crashes here it has to be due to something else. It might memory corruption, or the critical section impl not synchronizing things properly.

robamu:
Thanks for the clarification. I guess I will have to do some more debugging. This only occurs when I try do send as much data as possible without any delay in between..

robamu:
When looking at this bug again,
I looked a bit more through the smoltcp code. I found this segment:

https://github.com/smoltcp-rs/smoltcp/blob/94ce96c1a13bba592c81433a115e95206f90ad33/src/storage/packet_buffer.rs#L157

Wouldn't that code allow to call f while still returning the Full error?

KingCol13:
The mechanism for this bug is described in #6143

Dirbaio:
fixed by https://github.com/smoltcp-rs/smoltcp/pull/1150

PR #6284, opened 2026-06-07
C5 - Simple example running

usbalbin:


examples/stm32c5/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+embedded-hal-async = { version = "1.0" }
+embedded-io-async = { version = "0.7.0" }
+embedded-nal-async = "0.9.0"
+panic-probe = { version = "1.0.0", features = ["print-defmt"] }
+heapless = { version = "0.9", default-features = false }
+critical-section = "1.1"
+micromath = "2.0.0"
+stm32-fmc = "0.4.0"
+dsp-fixedpoint = { version = "0.1.0", default-features = false }
+embedded-storage = "0.3.1"
```
leftger: I'd advise against introducing any unused crates. I know @xoviat previously advised against introducing third party crates unless needed.

leftger (changes requested):
I'd like to see a GPIO/toggling/blinky action versus a simple "hello". It's armv8m so we know the CPU works fine. It's the peripheral bringup that is the best proof-of-concept in proving `stm32-data-generated` gave you the proper PAC.

examples/stm32c5/Cargo.toml
usbalbin: Yeah, I will clean this up. So far this is just a "look it sort of works". Same with the memory.x which I assume should not be needed once everything is ok

usbalbin:
I'll close this since this seems to have been merged in some other PR

PR #6288, opened 2026-06-08
net: add DHCPv4 NTP server (option 42) support

MathiasKoch:
## Summary

Adds support for reading the NTP servers DHCP option (option 42) in `embassy-net`, behind a new opt-in `dhcpv4-ntp` feature.

When enabled, the obtained `StaticConfigV4` gains an `ntp_servers: Vec<Ipv4Address, 4>` field, populated from the DHCP lease and surfaced through the existing `Stack::config_v4()`.

### Why a dedicated feature / receive buffer

smoltcp deliberately does not parse this option into its `dhcpv4::Config` — per [smoltcp-rs/smoltcp#972](https://github.com/smoltcp-rs/smoltcp/pull/972), the maintainers' position is that adding a field for every possible option does not scale, and the supported way to read additional options is the raw reply exposed via `set_receive_packet_buffer` / `Config::packet`.

This PR follows that intended mechanism:

- `StackResources` gains a fixed-size `dhcp_rx_buffer` (`DHCP_RX_BUFFER_SIZE = 576`, the minimum DHCP message size every server must respect per RFC 2131), only allocated when the feature is enabled — mirroring how `dhcpv4-hostname` allocates `HostnameResources`.
- The buffer is handed to the DHCP socket once at creation.
- Option 42 is decoded from the retained packet when a lease is configured.

An undersized buffer never corrupts the IP configuration (address/gateway/DNS come from smoltcp's parsed `Config`); it only causes the extra options to be dropped.

### Notes

- The NTP server list is capped at 4 (matching common NTP client practice); extra servers are dropped, as smoltcp does for DNS servers.
- Setting a receive packet buffer makes smoltcp report the config as changed on every received DHCP packet (including renewals), so `apply_static_config()` re-runs each renewal. This only affects stacks that opt into `dhcpv4-ntp`.

### DIFF
--- a/embassy-net/Cargo.toml
+++ b/embassy-net/Cargo.toml
@@ -20,6 +20,7 @@ build = [
     {target = "thumbv7em-none-eabi", features = ["defmt", "dhcpv4", "dns", "medium-ethernet", "tcp", "udp"]},
     {target = "thumbv7em-none-eabi", features = ["defmt", "dhcpv4", "dhcpv4-hostname", "dns", "medium-ethernet", "tcp", "udp"]},
     {target = "thumbv7em-none-eabi", features = ["defmt", "dhcpv4", "dhcpv4-hostname", "dns", "medium-ethernet", "slaac", "tcp", "udp"]},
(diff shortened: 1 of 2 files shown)


MathiasKoch:
@Dirbaio Any chance you could take a look at this, when you find the time?

PR #6290, opened 2026-06-08
[Executor] Expand trace

### LINKED ISSUE #6235: [executor] Exposing metadata to trace?
Hi all, I'm using the trace function of the executor.

I also want to be able to capture info about the metadata of tasks (names, prios & deadlines).
This is not possible right now and there are 2 separate problems:

1. The trace functions get a task id which you're supposed to treat opaquely. There's no stable way to go from task id to metadata of that task.
2. Even if we were able to get to the metadata, we'd get no updates when the values change.

What would solve my issue:
- Change the trace functions to get `TaskRef`s instead of the `u32` task ids. If people still want the IDs, they can get them from the `TaskRef`.
- Add extra trace functions (one for each metadata entry) that get called when the values change. These functions would only get a `TaskRef` and no executor ID (since that info is not present or relevant at that point)

This would change the behavior of the trace slightly since you'd be able to observe a task before `TaskNew` is called. Personally I think that's fine, but we'd have to change the docs.

Do we think this is a good change?

diondokter:
Closes #6235 

Metadata is exposed to the trace & ids have been formalized a bit

- The `Metadata` struct is no longer exposed to the used, instead the `MetadataRef` is, which will call trace callbacks when the properties are changed
- There are now types for executor id and task id. Internally they are a `usize` instead of a `u32` so it keeps fully working on 64-bit targets
- The callbacks now get the taskref instead of just the ID.
  - An implementor can get the id through the taskref
  - The metadata can now also be accessed
- The downside is that the implementor of the trace callbacks now needs to pull in the executor crate, but I don't think that's an issue in practice.

### DIFF
--- a/embassy-executor/src/metadata.rs
+++ b/embassy-executor/src/metadata.rs
@@ -53,6 +53,7 @@ impl MetadataRef {
     #[cfg(feature = "metadata-name")]
     pub fn set_name(&self, name: &'static str) {
         self.metadata().set_name(name);
+        #[cfg(feature = "_any_trace")]
         crate::raw::trace::task_name_set(self.task, name);
     }
 
@@ -66,6 +67,7 @@ impl MetadataRef {
     #[cfg(feature = "scheduler-priority")]
     pub fn set_priority(&self, priority: u8) {
         self.metadata().set_priority(priority);
+        #[cfg(feature = "_any_trace")]
         crate::raw::trace::task_priority_set(self.task, priority);
     }
 
@@ -81,6 +83,7 @@ impl MetadataRef {
     #[cfg(feature = "scheduler-deadline")]
     pub fn set_deadline(&self, instant_ticks: u64) {
         self.metadata().set_deadline(instant_ticks);
+        #[cfg(feature = "_any_trace")]
         crate::raw::trace::task_deadline_set(self.task, instant_ticks);
     }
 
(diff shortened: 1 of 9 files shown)


diondokter:
Ugh ok, let's run CI locally first haha

lulf (commented):
The API, types etc. look fine to me. I'm less sure about the implications of &'static self , but I suppose it doesn't really matter given executor is always static?

diondokter:
> I'm less sure about the implications of &'static self

For the enqueue function? Well pretty much all other functions in the executor use &'static self too. I think it wasn't static because it strictly didn't need it (yet).
Enqueue is only called through `spawn` which already uses &'static self and through `wake_task` with a pointer (so that has any lifetime)

diondokter:
If you really want, I can change it btw. The executor ID function doesn't *need* the static lifetime

lulf (approved):
I don't think it's necessary to change it, I suppose it even reads more clear/is less confusing when explicitly stated like this.

embassy-executor/src/metadata.rs
```diff
@@ -146,3 +132,103 @@ impl Metadata {
...
+    ///
+    /// This method does NOT check whether the deadline has already passed.
+    #[cfg(feature = "scheduler-deadline")]
+    fn set_deadline(&self, instant_ticks: u64) {
+        self.deadline.set(instant_ticks);
+    }
+}
+
+/// A unique ID for a task
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
bugadani, line 217: These should probably derive `defmt::Format` if they also derive Debug

PR #6296, opened 2026-06-08
feat!: cryptocell power management

nponsard:
This refactors how the CryptoCell is handled in `embassy-nrf` to allow powering down the CryptoCell when none of its components are used, improving power consumption. 

This has a breaking change as `cryptocell_rng` is moved into a `cryptocell` module.

This allows the chip to go on System ON IDLE, for the nrf9151 that makes it idle at less than 10 μA after using CcRng instead of ~500μA before this change.

### DIFF
--- a/embassy-nrf/src/cryptocell/mod.rs
+++ b/embassy-nrf/src/cryptocell/mod.rs
@@ -1,0 +1,56 @@
+//! CryptoCell drivers
+//!
+//! If you want to interact with the CryptoCell, you can activate it using [`activate()`].
+//! This function will return a handle that will keep the CryptoCell active until it is dropped.
+
+#![macro_use]
+
+use core::sync::atomic::{AtomicU16, Ordering};
+
+use crate::pac;
+
+pub mod rng;
+
+static ACTIVE_USERS: AtomicU16 = AtomicU16::new(0);
+
+/// Use this function to activate the CryptoCell subsystem.
+///
+/// CryptoCell is active until the handle is dropped.
+#[must_use]
+pub fn activate() -> CryptoCellActivationHandle {
+    ACTIVE_USERS.fetch_add(1, Ordering::Relaxed);
+
+    // Make sure the CryptoCell is enabled.
+    pac::CRYPTOCELL.enable().write(|w| w.set_enable(true));
+
+    CryptoCellActivationHandle::new()
+}
+
+fn release() {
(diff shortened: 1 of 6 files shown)


nponsard:
`CcRng` has been added not that long ago so I felt that moving it into a `cryptocell` module won't cause much issues. If it is an issue, I can try to change it to not do a breaking change.

embassy-nrf/src/cryptocell/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+pub mod rng;
+
+static ACTIVE_USERS: AtomicU16 = AtomicU16::new(0);
+
+/// Use this function to activate the CryptoCell subsystem.
+///
+/// CryptoCell is active until the handle is dropped.
+#[must_use]
+pub fn activate() -> CryptoCellActivationHandle {
+    let _ = ACTIVE_USERS.fetch_update(Ordering::Release, Ordering::Acquire, |x| Some(x + 1));
```
Dirbaio: would be better to use fetch_add, fetch_sub. Better conveys intent and I think they generate slightly shorter asm.

embassy-nrf/src/cryptocell/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    pac::CRYPTOCELL.enable().write(|w| w.set_enable(true));
+
+    CryptoCellActivationHandle::new()
+}
+
+fn release() {
+    // If we released the last one, turn off CRYPTOCELL
+    if let Ok(deps) = ACTIVE_USERS.fetch_update(Ordering::Release, Ordering::Acquire, |x| Some(x - 1))
+        && deps == 1
+    {
```
Dirbaio: this is racy. 

- Main thread calls release
- release decrements user count, sees it went from 1 to 0
- interrupt fires at this very instant. after the decrement but before set_enable(false)
- interrupt calls activate
- activate increments user count, writes set_enable(true)
- interrupt returns
- release does set_enable(false)
- oops now we got ACTIVE_USERS=1 but ENABLE=0

I think the only way to fix this is to use critical sections. With that the atomic ordering can be Relaxed.

nponsard: Ah this is indeed an issue, thank you for spotting this.

nponsard:
Ah wait let me squash first

PR #6298, opened 2026-06-08
Support esp-hosted-mcu

### LINKED ISSUE #5418: Embassy-net support for esp-hosted-mcu
It seems that Espressif has deprecated the esp-hosted-fg firmware, recommending instead to move to the esp-hosted-mcu project. Unfortunately, the protocol is incompatible with the previous esp-hosted-fg, so I've started working on esp-hosted-mcu support for embassy-net.

For now, I've forked the embassy-net-esp-hosted crate and integrated the new protocol, as well as added a basic signaling system for events coming from the ESP chip (disconnection and heartbeat deadline).

Is this something you'd be interested in merging? If so, what would be the preferred approach for contribution?

bugadani:
Supersedes and closes #5709, closes #5418 supports both in the same crate, optionally both, in which case the version used is determined at runtime (via observing the protobuf endpoint names).

Includes trimming down the generated code
   - Message IDs/payload variants done, for a massive code size decrease
   - Types left uncommented to reduce future churn

Code size implications:

Host: ESP32-H2, release mode

```toml
[profile.release]
codegen-units = 1
debug         = 2
lto           = 'fat'
opt-level     = 's'
```

```
git main @ 65ce9488d2af544d2124b00a8b08544fea7e624a
   text	   data	    bss	    dec	    hex	filename
 476716	   2352	 103232	 582300	  8e29c	esp32h2-hosted

dev (mcu only)
   text	   data	    bss	    dec	    hex	filename
 471278	   2352	 102400	 576030	  8ca1e	esp32h2-hosted

dev (fg only)
   text	   data	    bss	    dec	    hex	filename
 454254	   2352	 102176	 558782	  886be	esp32h2-hosted

dev (mcu + fg)
   text	   data	    bss	    dec	    hex	filename
 488166	   2368	 102440	 592974	  90c4e	esp32h2-hosted
```


bugadani:
I'll not touch OTA in this PR, but I suspect we should set link state to be down when the hosted device is rebooting 😅

Issue #5498, opened 2026-02-20
embassy-rp: Setting state machine JMP pin does not work properly for pins > 31


wgxer:
On the **RP235xB**, setting a pin that exclusively exist in the upper range (GPIO 31-47) as `JMP` pin using [set_jmp_pin](https://github.com/embassy-rs/embassy/blob/embassy-rp-v0.9.0/embassy-rp/src/pio/mod.rs#L733) doesn't work as expected. It seems like it sets `JMP` pin in [set_config](https://github.com/embassy-rs/embassy/blob/embassy-rp-v0.9.0/embassy-rp/src/pio/mod.rs#L782) without shifting the pin number. I would expect it to do something similar to what it does for [other pins](https://github.com/embassy-rs/embassy/blob/embassy-rp-v0.9.0/embassy-rp/src/pio/mod.rs#L864).

phansel:
@wgxer is this still the case?

PR #6306, opened 2026-06-09
Fix RP2350 PIO GPIOBase unsoundness & make uart stopbit valid

### LINKED ISSUE #6303: RP2350 PIO GPIOBASE soundness issues
When configuring a pio sm, the sm's pins are checked if it fits in a GPIOBASE, however when configuring multiple SMs to the same PIO block this check is not done for all SMs at the same time, but only each individually cause the GPIOBASE to be in the state of the last SM created, causing potentially broken pin configuration. This is mainly an issue when you use pins below 32 in one SM and above 32 in another, as this will be a silent failure (each individual check passes), and it's possible to use it (with GPIOBASE set for both) but that is not possible with the current algo.

One way to fix it is to scan all `sm`s when doing `set_config`, however this will be quite different from the current code, as it needs to recompute all of the sms for each set_config. The other option is to allow gpiobase to be set manually, and calculate everything based on that

i404788:
This should resolve https://github.com/embassy-rs/embassy/issues/6303 as well as improve the PioUart reliability slightly


wowkster:
I was about to make a related PR to fix #5498 (code can be found [here](https://github.com/wowkster/embassy/commit/3815917d7e821db3f10735f74b5f566fed95351f)), but it looks like this change would actually resolve that issue and in a cleaner way. I'm a fan of this approach since it's much closer to what [pico-sdk does internally](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_pio/include/hardware/pio.h#L809) and makes the configuration a lot more predictable.

Dirbaio:
Can you fix CI?

CBJamo:
This LGTM. It's a clean improvement and fixes a hole in my logic, thanks for catching it.

leftger:
@CBJamo welcome as maintainer! Since you approved this do you wanna merge it?

PR #6308, opened 2026-06-09
embassy_rp: WS2812 pio program improvements

tact1m4n3:
- Use `smart-leds-trait` crate instead of `smart-leds` (as advised in the smart-leds crate README)
- Implement `SmartLedsWriteAsync` trait
- Refactor to remove the const generic N from the struct, moving it directly to the `write_slice` method
- Rename `write` -> `write_slice` (to distinguish from the write method of `SmartLedsWriteAsync`)

Please let me know if these changes are reasonable. If so, should I also write something in the CHANGELOG?


Ragarnoy:
Tested, works, and code lgtm

PR #6310, opened 2026-06-09
C5 - Work on RCC config

usbalbin:
Not yet tested


xoviat:
Note: flash should be available now, but it's not cleaned up. I decided to leave the cleanup to you, whenever you have time.

usbalbin:
I have tested both hsi and hsi_div3 examples and they print

> Sys clocked by HSI/3 running at 48MHz!

and

>Sys clocked by HSI running at 144MHz!

as expected

PR #6315, opened 2026-06-10
sQSPI soft-peripheral driver for nRF54L

lulf:
Add a Rust driver for the sQSPI ('soft' QSPI) running on the nRF54L coprocessor (VPR)

* Add config to embassy_nrf::init() to tune FLPR reset/init behavior.
* Add sqspi module using same register interface as PAC for the 'RAM-based' register block.
* Add example running on the nrf54l15 devkit against the external flash. Add license note of binary running on the FLPR core.

Written with help of claude code, but tuned, reviewed and tested on the nrf54l15 devkit.


embassy-nrf/src/sqspi/mod.rs
```diff
@@ -0,0 +1,966 @@
...
+//! # Example
+//! ```ignore
+//! use embassy_nrf::sqspi::{self, Config};
+//! use static_cell::StaticCell;
+//! use core::mem::MaybeUninit;
+//!
+//! bind_interrupts!(struct Irqs { VPR00 => sqspi::InterruptHandler<peripherals::VPR>; });
+//!
+//! static FW: &[u8] = include_bytes!("sqspi_firmware.bin");
+//! static RAM: StaticCell<[MaybeUninit<u8>; 0x4000]> = StaticCell::new();
```
i509VCB: ConstStaticCell should work here.

examples/nrf54l15-app/src/bin/sqspi.rs
```diff
@@ -0,0 +1,107 @@
+#![no_std]
+#![no_main]
+
+//! Example for the `embassy-nrf` **sQSPI** driver on the nRF54L15-DK.
+//!
+//! This drives an external MX25R64 (e.g. the chip on the nRF52840-DK) in Quad
```
i509VCB: AI gave a bad chip example. (Using nRF52 as an example in something only available on nRF54)

PR #6318, opened 2026-06-10
stm32/adc4: add AUTOFF/LFTRIG; fix RingBufferedAdc stop/start as suspend/resume

leftger:
## Summary

Three related improvements to the STM32 ADC4 driver (STM32WBA / STM32U5) motivated by a low-power VBUS detection use case: LPTIM1-triggered ADC4 sampling with an analog watchdog wakeup from Stop mode.

### 1. `set_autoff` and `set_lftrig` on `Adc<T>` (`#[cfg(stm32wba)]`)

`AUTOFF` (PWR.AUTOFF) powers the ADC down after each conversion and wakes it automatically on the next external trigger. `LFTRIG` (CFGR2.LFTRIG) adjusts the ADC sampling window for trigger rates below ~1 kHz (e.g. LPTIM from LSE). Both are needed for correct, power-efficient operation when the ADC is triggered at a very low duty cycle.

### 2. Fix `AdcRegs::stop(disable: bool)` for Adc4

`stop()` accepted a `disable: bool` parameter but ignored it (`_disable`), always issuing ADDIS and fully disabling the ADC. This made it impossible to use `RingBufferedAdc::stop()` / `start()` as a lightweight suspend/resume pair. Fixed to:
- `disable=false` → ADSTP only; ADC stays enabled, registers can be reconfigured, `start()` restarts without re-enabling
- `disable=true` → ADSTP then ADDIS; matches previous behaviour

### 3. Fix `RingBufferedAdc::stop()` to halt the ADC hardware

Previously `stop()` only paused the DMA ring buffer, leaving the ADC running and issuing DMA requests into a paused channel. Now it calls `self.regs.stop(false)` first. `start()` and `stop()` are documented as a suspend/resume pair.

## Test plan

- [x] Builds for `stm32wba65ri` (thumbv8m.main-none-eabihf)
- [x] Builds for `stm32f429zi` (thumbv7em-none-eabihf) — confirms no regression on non-WBA targets using `stop()`

### DIFF
--- a/embassy-stm32/src/adc/adc4.rs
+++ b/embassy-stm32/src/adc/adc4.rs
@@ -292,7 +292,12 @@ impl AdcRegs for crate::pac::adc::Adc4 {
         // subsequent `start()` call restarts conversion without the ADEN startup
         // sequence.  This is the correct mode for RingBufferedAdc::stop() /
         // start() used as a suspend/resume pair.
-        if disable && (cr.aden() || cr.adstart()) {
+        //
+        // TODO: align `disable` flag semantics with other ADC impls (c0, f1, l1,
+        // g4, v1-v4, f3 all handle this differently); ideally all impls expose a
+        // consistent enable/disable API so callers aren't surprised when porting.
+        if disable && cr.aden() {
+            // ADSTART=1 implies ADEN=1, so checking aden() alone is sufficient.
             self.cr().modify(|w| w.set_addis(true));
             while self.cr().read().aden() {}
         }
--- a/embassy-stm32/src/adc/adc4.rs
+++ b/embassy-stm32/src/adc/adc4.rs
@@ -275,19 +275,30 @@ impl AdcRegs for crate::pac::adc::Adc4 {
         });
     }
 
-    fn stop(&self, _disable: bool) {
+    fn stop(&self, disable: bool) {
         let cr = self.cr().read();
+
+        // Stop any ongoing regular conversion via ADSTP (self-clearing).
+        // Must be done before ADDIS per RM: software shall not set ADDIS while
+        // ADSTART=1 or JADSTART=1.
         if cr.adstart() {
             self.cr().modify(|w| w.set_adstp(true));
             while self.cr().read().adstart() {}
         }
 
-        if cr.aden() || cr.adstart() {
+        // Disable only when the caller requests it (disable=true).  Keeping the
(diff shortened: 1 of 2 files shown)


leftger:
@pufmat if you can take a look at this I'd appreciate it. You've been helping a lot with DMA so I figured you'd want to check it out.

leftger:
It's draft for now but if pufmat says it looks good I'll un-draft it.

embassy-stm32/src/adc/adc4.rs
```diff
@@ -275,19 +275,30 @@ impl AdcRegs for crate::pac::adc::Adc4 {
...
         }
 
-        if cr.aden() || cr.adstart() {
+        // Disable only when the caller requests it (disable=true).  Keeping the
+        // ADC enabled (disable=false) after ADSTP leaves it in a suspended state
+        // where CFGR1/CHSELR/AWD registers can be safely reconfigured, and a
+        // subsequent `start()` call restarts conversion without the ADEN startup
+        // sequence.  This is the correct mode for RingBufferedAdc::stop() /
+        // start() used as a suspend/resume pair.
+        if disable && (cr.aden() || cr.adstart()) {
```
pufmat: I think `cr.adstart()` is not required here, because ADSTART can be 1 only when ADEN is also 1.

pufmat:
Some notes for this PR or for improvements in future PRs:
- Other implementations currently handle disable flag differently:
  - c0, f1, l1, g4, v2, v3, v4 - ignores, always disables ADC
  - f3 - always disables ADC, flag only disables regulator
  - v1 - ignores, does not disable ADC, exposes `power_down`
- All ADCs should probably allow user to manually enable/disable them.

leftger:
Thanks for the review! I'll try and make it consistent with the other devices. I agree that aligning consistency is important across all devices. Any embassy user should expect similar behavior when switching between ST devices :)

embassy-stm32/src/adc/adc4.rs
leftger: Done.

leftger:
I'm planning to make a PR soon that'll align ADC stop/disable behavior. I'll loop you in on it as soon as I make the PR.

xoviat:
> All ADCs should probably allow user to manually enable/disable them.

The idea is that you can drop the ADC, and this will disable it.

PR #6320, opened 2026-06-11
stm32/adc: split AdcRegs::stop(disable) into stop() and power_down()

leftger:
## Summary

Addresses the API consistency concern raised in #6318 (cc @pufmat).

The `stop(&self, disable: bool)` method on `AdcRegs` was ambiguous: 7 of 10
implementations ignored the `disable` flag entirely, each doing something
different regardless. The flag was only meaningful in `f3`, `l1`, and `adc4`,
but with different semantics in each. This made the trait a poor contract when
porting between STM32 families.

Replace with two explicit methods:

- **`stop()`** — halt any ongoing conversion, leaving the ADC enabled and ready
  to restart. All configuration registers remain valid; `start()` resumes without
  re-running the enable sequence. This is the correct primitive for
  `RingBufferedAdc::stop()` / `start()` suspend-resume and for reconfiguration
  between reads.
- **`power_down()`** — stop conversions then fully disable the ADC hardware.
  Intended for `Drop` impls and explicit user-facing power management.

## Changes

- `AdcRegs` trait: replace `fn stop(&self, disable: bool)` with `fn stop(&self)` + `fn power_down(&self)` (each with a doc comment)
- All 10 `AdcRegs` impls updated: `v1`, `v2`, `v3`, `v4`, `g4`, `c0`, `f1`, `f3`, `l1`, `adc4`
- The two `Drop` impls (`Adc` and `RingBufferedAdc`) updated to call `power_down()`
- All 20 other call sites updated to `stop()`
- Existing public `Adc::power_down()` on `v1` and `v3` now delegate to the trait method
- `f3` `stop()` retains ADDIS (F3 RM requires ADEN=0 before register writes); `power_down()` additionally disables the voltage regulator

## Notes

- For legacy CR2-based ADCs (`f1`, `v2`, `l1`) where ADON serves as both enable
  and conversion trigger, `stop()` clears SWSTART/DMA/interrupts without touching
  ADON; `power_down()` additionally clears ADON. This enables correct
  suspend-resume on those families.
- `g4`'s previous `stop()` always cleared ADEN — that logic has moved to
  `power_down()`.

## Test plan

- [x] `stm32wba65ri` (adc_wba / adc4)
- [x] `stm32f429zi` (adc_v2)
- [x] `stm32g491re` (adc_g4 + adc_v3)
- [x] `stm32f030r8` (adc_v1)
- [x] `stm32f303vc` (adc_f3v1)
- [x] `stm32f103rb` (adc_f1)
- [x] `stm32h743bi` (adc_v4)
- [x] `stm32l151rb` (adc_f3v2 / l1)


leftger:
Aligned this PR with the maintainer direction discussed in chat (xoviat): prefer resumable/suspend semantics for normal operation, and reserve full power teardown for ownership/drop paths.

## What changed

- Kept the `AdcRegs` split as explicit semantics:
  - `stop()`: stop conversion, keep ADC enabled (suspend/resume use case)
  - `power_down()`: full ADC disable (teardown use case)
- Removed public method-based teardown APIs from user-facing ADC types:
  - removed `Adc::power_down()` in `adc/v1.rs`
  - removed `Adc::power_down()` in `adc/v3.rs`
- `Drop` paths still perform full teardown:
  - `Adc` drop uses `regs().power_down()` + `rcc::disable::<T>()`
  - `RingBufferedAdc` drop uses `regs().power_down()`

## Rationale

The method-based `power_down()/power_up()` style makes it too easy to call operational methods while the peripheral is in a powered-down state, which does not generalize well across peripherals.

This PR now follows a clearer contract:
- suspend/resume behavior is explicit and lightweight (`stop`/`start`)
- teardown is tied to ownership/lifetime (`Drop`), which naturally preserves low-power/refcount correctness for Stop modes.

## Low-power / Stop mode intent

- Suspend (`stop`) does **not** imply relinquishing peripheral ownership.
- Teardown (`Drop`) is the point where RCC-level disable/refcount transitions happen.
- This keeps Stop eligibility behavior predictable while preserving fast resume paths.

## Verification

- `cargo check --manifest-path embassy-stm32/Cargo.toml --features stm32wba65ri` passes.

PR #6324, opened 2026-06-12
refactor(tuntap): use an `OwnedFd` internally

ROMemories:
This refactors the internals of `TunTap` to leverage an `OwnedFd`: this removes the need for a manual `Drop` implementation, as [`OwnedFd` automatically closes the file descriptor on drop](https://doc.rust-lang.org/stable/std/os/fd/struct.OwnedFd.html), and makes the lifecycle and safety of the file descriptor a bit easier to reason about.

I've run Clippy (with the pedantic group enabled) and checked the lints relevant to this change, but I haven't run tests on this.

### DIFF
--- a/embassy-net-tuntap/src/lib.rs
+++ b/embassy-net-tuntap/src/lib.rs
@@ -2,7 +2,7 @@
 #![doc = include_str!("../README.md")]
 use std::io;
 use std::io::{Read, Write};
-use std::os::fd::AsFd;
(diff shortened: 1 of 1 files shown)


ROMemories:
Thanks a lot for merging this quickly! Could you do a quick release of this crate? I don't think there are other changes pending, and that'd make my cargo-vet audit of it easier.

PR #6325, opened 2026-06-12
stm32: ETH PTP timestamps

jordens:
This builds on the existing `embassy-net-driver::PacketMeta` packet-id plumbing and lets the STM32 Ethernet (v2/v2a) driver associate hardware RX/TX descriptor timestamps with packet metadata. Builds without `ptp` do not pay memory or code cost.

This PR focuses on the packet and timestamp handling and does not yet implement the higher layers of MAC PTP peripheral register setup (later PR) or protocol layer (statime shim).

### DIFF
--- a/embassy-stm32/src/eth/v2/descriptors.rs
+++ b/embassy-stm32/src/eth/v2/descriptors.rs
@@ -94,6 +94,7 @@ impl TDes {
         self.tdes3.get() & EMAC_DES3_OWN == 0
     }
 
+    #[cfg(feature = "ptp")]
     fn timestamp(&self) -> Option<PtpTimestamp> {
         (self.tdes3.get() & EMAC_TDES3_TTSS != 0).then(|| PtpTimestamp {
             seconds: self.tdes1.get(),
@@ -107,6 +108,7 @@ pub(crate) struct TDesRing<'a> {
     buffers: &'a mut [Packet<TX_BUFFER_SIZE>],
     packets: TxPacketStateRing<'a>,
     index: usize,
+    #[cfg(feature = "ptp")]
     completion_index: usize,
 }
 
@@ -136,6 +138,7 @@ impl<'a> TDesRing<'a> {
             buffers,
             packets,
             index: 0,
+            #[cfg(feature = "ptp")]
             completion_index: 0,
         }
     }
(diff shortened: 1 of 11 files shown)


jordens:
@xoviat that was fast, thanks. yolo!

cc @datdenkikniet given you did the work on `packetmeta-id`.

embassy-stm32/src/eth/v2/descriptors.rs
```diff
@@ -272,26 +376,78 @@ impl<'a> RDesRing<'a> {
...
+        // not interpret TSA unless the hardware says the status word is valid.
+        let has_timestamp =
+            descriptor.rdes3.get() & EMAC_RDES3_RS1V != 0 && descriptor.rdes1.get() & EMAC_RDES1_TSA != 0;
+        if !has_timestamp {
+            return Some((None, false));
+        }
+
+        let next = (index + 1) % self.descriptors.len();
+        let context = &self.descriptors[next];
+        if context.context_available() {
```
datdenkikniet, line 416: As a note/question/point of interest: we observed that enabling PTP on the RX path always causes immediate RX pipeline stalls, because (I believe) of the following chain of events:

1. DMA receives a packet, writes it to `index`
2. MAC generates context descriptor, overwrites `index + 1`
3. DMA tries to continue reception at `index + 1`, which is now that context descriptor and therefore stalls

Is this behaviour that you have also observed, and if so, is it an acceptable drawback/problem for you? I assume the problem is reduced if you use the "only timestamp PTP specific packets", but we never tried that (and were just disappointed that this thing seems to happen)

jordens: I agree that the DMA handoff is delicate.
We've been testing this on a full H743 app, with low RX and 3 MB/s TX traffic load, but with PTP event snapshots only, not every RX packet. There were no pipeline stalls. The full stack worked fine all the way to a well disciplined PTP clock with statime including some convoluted PPS-timer-capture setup.
This driver treats the timestamp context descriptor as consuming a real RX ring slot. If a context descriptor reaches the current RX index, it is returned to DMA without being exposed as a packet. If a packet advertises an RX timestamp, the driver waits for the following context descriptor and then returns both descriptors to DMA when the packet is popped.

PR #6328, opened 2026-06-12
stm32/rcc:  fix refcount handling

xoviat:
only adjust the refcounts if the bit is changing state

### DIFF
--- a/embassy-stm32/src/rcc/mod.rs
+++ b/embassy-stm32/src/rcc/mod.rs
@@ -297,33 +297,32 @@ impl RccInfo {
     }
 
     // TODO: should this be `unsafe`?
-    pub(crate) fn enable_and_reset_with_cs(&self, cs: CriticalSection) -> Result<(), ()> {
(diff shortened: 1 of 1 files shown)


xoviat:
bender run

embassy-stm32/src/rcc/mod.rs
```diff
@@ -297,116 +297,97 @@ impl RccInfo {
...
 
-                if !self.is_enabled_by_other_core() {
+                if (self.enable_ptr().read_volatile() & (1u32 << self.enable_bit)) == 0 {
                     let val = reset_ptr.read_volatile();
                     reset_ptr.write_volatile(val | 1u32 << self.reset_bit);
 
                     trace!("rcc: reset 0x{:x}:{}", self.enable_offset, self.enable_bit);
+                } else {
+                    trace!("rcc: skip reset 0x{:x}:{}", self.enable_offset, self.enable_bit);
                 }
```
xoviat, line 326: Maybe I should return here?

PR #6334, opened 2026-06-13
STM32N6: Add voltage scale and add a module to handle RIF attributes

lsartory:
Improved support for the STM32N6.
Related to issue #4614.

Please note: the `promote_master()` and associated functions were removed from the RCC code.
I could not find any reference to them, but I assume they were used in third-party projects.
See the updated examples (e.g. `lcd.rs`) using the `rif::RifMaster` and `rif::RifPeripheral` structs for proper usage.


xoviat:
What about making `RifMaster` and `RifPeripheral` pac enums, and re-exporting them from the pac?

lsartory:
> What about making `RifMaster` and `RifPeripheral` pac enums, and re-exporting them from the pac?

Sure, that would make it cleaner.
Can you please give me some pointers on how to do it?

xoviat:
> Can you please give me some pointers on how to do it?

I am sorry: just realized that this wouldn't work for this.

xoviat:
bender run

lsartory:
> > Can you please give me some pointers on how to do it?
> 
> I am sorry: just realized that this wouldn't work for this.

Thanks for the merge!
I'm still not 100 % satisfied by the current implementation though, as it is entirely relative to the STM32N6.
We will need to find a way to auto-generate the whole RIF thing if / when ST releases a new variant using it.

PR #6338, opened 2026-06-13
stm32/rtc:  conditionally re-init rtc

ant32:
entering init mode causes the subseconds to be loss and the clock is frozen for the duration. I used claude to research and double check the work but the code itself was hand written. I confirmed from the stm32l5 reference manual that only the initial time and date calendar values, including the time format and the prescaler configuration need init mode.

RTC initialization and configuration is on page 1407 https://www.st.com/resource/en/reference_manual/rm0438-stm32l5-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf

### DIFF
--- a/embassy-stm32/src/rtc/v3.rs
+++ b/embassy-stm32/src/rtc/v3.rs
@@ -10,10 +10,10 @@ impl super::Rtc {
     /// It this changes the RTC clock source the time will be reset
     pub(super) fn configure(&mut self, async_psc: u8, sync_psc: u16) {
         let prer = RTC::regs().prer().read();
-        if (prer.prediv_s() == sync_psc
+        if prer.prediv_s() == sync_psc
             && prer.prediv_a() == async_psc
             && RTC::regs().cr().read().fmt() == Fmt::TwentyFourHour
-            && RTC::regs().icsr().read().inits())
+            && RTC::regs().icsr().read().inits()
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/rtc/v3.rs
```diff
@@ -9,19 +9,26 @@ impl super::Rtc {
     /// Applies the RTC config
     /// It this changes the RTC clock source the time will be reset
     pub(super) fn configure(&mut self, async_psc: u8, sync_psc: u16) {
-        self.write(true, |rtc| {
+        let prer = RTC::regs().prer().read();
+        let needs_init = prer.prediv_s() != sync_psc
+            || prer.prediv_a() != async_psc
+            || RTC::regs().cr().read().fmt() != Fmt::TwentyFourHour;
```
xoviat: Why not just return here if the RTC does not need to be initialized?

ant32: I am fine with that. I don't have a board to test it today. I am also not super familiar if we need any of the specific registers in a known state for the rest of the rtc library. The STMicroelectronics/stm32l5xx-hal-driver only checks if it is [initialized](https://github.com/STMicroelectronics/stm32l5xx-hal-driver/blob/0718e81a6cf945f488ed9ed1eb5d514290fa3fc6/Src/stm32l5xx_hal_rtc.c#L322).
```rust
RTC::regs().icsr().read().inits()
```

xoviat: The condition should be if `if (prer.prediv_s() == sync_psc && prer.prediv_a() == async_psc &&  RTC::regs().cr().read().fmt() == Fmt::TwentyFourHour && RTC::regs().icsr().read().inits()) { return }`, to verify that the RTC is initialized with the correct format and prescalers. If this is not the case, then we should re-initialize the RTC.

PR #6339, opened 2026-06-13
Move ioctl buffer from stack to State

bugadani:
We're probably going to have bigger messages (related to AP scanning, e.g.), and the only constraint is that a serialized message must fit into the 1600 byte transfer. It's probably better to allocate that buffer once, rather than hoping we don't overflow the stack.

### DIFF
--- a/embassy-net-esp-hosted/src/rpc/ioctl_ctx.rs
+++ b/embassy-net-esp-hosted/src/rpc/ioctl_ctx.rs
@@ -1,24 +1,24 @@
 use micropb::{MessageDecode, MessageEncode, PbEncoder};
 
+use crate::MAX_IOCTL_SIZE;
 use crate::control::Error;
 use crate::ioctl::Shared;
 
 /// One encode → ioctl → decode round-trip through [`Shared`].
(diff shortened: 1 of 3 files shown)


embassy-net-esp-hosted/src/rpc/ioctl_ctx.rs
```diff
@@ -1,24 +1,24 @@
...
     }
 
     pub async fn exchange(&mut self, msg: &mut (impl MessageDecode + MessageEncode)) -> Result<(), Error> {
-        // Theoretical max overhead is 29 bytes. Biggest message is OTA write with 256 bytes.
-        let mut buf = [0u8; 256 + 29];
-        let buf_len = buf.len();
+        let buf_len = self.ioctl_buffer.len();
 
-        let mut encoder = PbEncoder::new(&mut buf[..]);
+        let mut encoder = PbEncoder::new(&mut self.ioctl_buffer[..]);
```
bugadani, line 21: It's a shame PbWriter is not implemented for `&mut [u8; N]`

PR #6344, opened 2026-06-14
esp-hosted: Implement scanning

bugadani:
- FG is limited to 8 APs to not bloat the message struct up too much
  - Should the scan API maybe be feature-gated? We could break this into logical features (wifi sta (optionally with scan), ota, wifi ap, bluetooth, etc.). Slightly annoying to keep the generated proto up-to-date, but code size and ram use could be tuned.
- MCU iterates one by one so that we don't limit the results there
- The big deserialized message is now stored in State

### DIFF
--- a/embassy-net-esp-hosted/src/proto/mod.rs
+++ b/embassy-net-esp-hosted/src/proto/mod.rs
@@ -35,6 +35,11 @@ fn main() {
         ".CtrlMsg_Req_VendorIEData.payload",
         micropb_gen::Config::new().max_bytes(64),
     );
+    // Scan result is extremely big
+    g.configure(
+        ".CtrlMsg_Resp_ScanResult.entries",
+        micropb_gen::Config::new().max_len(8),
+    );
 
     g.compile_protos(
         &["src/proto/fg/esp_hosted_config.proto"],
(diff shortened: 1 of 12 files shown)


embassy-net-esp-hosted/src/proto/mod.rs
```diff
@@ -56,6 +61,21 @@ fn main() {
...
+    g.configure(".wifi_ap_record.second", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.pairwise_cipher", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.group_cipher", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.ant", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.bitmask", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.country", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.he_ap", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.bandwidth", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.vht_ch_freq1", micropb_gen::Config::new().skip(true));
+    g.configure(".wifi_ap_record.vht_ch_freq2", micropb_gen::Config::new().skip(true));
```
bugadani: This mess brings the Rpc message size down from 3k+ bytes to 700-some. It might still be better to make the scan api optional, though, limiting to 8 networks is maybe a bit strict.

embassy-net-esp-hosted/src/rpc/mcu.rs
```diff
@@ -190,6 +192,27 @@ impl RpcBackend for McuBackend {
...
     }
 
+    async fn scan<const N: usize>(&self, ctx: &mut IoctlCtx<'_>, result: &mut Vec<Network, N>) -> Result<(), Error> {
+        let req = Rpc_Req_WifiScanStart {
+            block: true,
+            ..Default::default()
+        };
+        exchange!(ctx, ReqWifiScanStart, RespWifiScanStart, req);
+
+        // TODO: query results one-by-one?
```
bugadani: There is a different API returning results one by one, but it's more opportunity for leaking memory if this is cancelled (need to clear explicitly if not all records are fetched). But this API is a bit memory-leaky anyway.

PR #6336, opened 2026-06-13
stm32/i2cv2: allow zero-length transfers

s7g4:
#### Description
This PR allows zero-length transfers in the STM32 I2C v2 driver (`stm32/i2cv2`). This supports sending the address, verifying the ACK/NACK status, and generating proper STOP/RESTART conditions without transferring any data bytes. This enables address-only "device probe" transactions (I2C bus scanning) which the I2C spec permits.

Fixes #2298 

**Key Design Decisions:**
1. **Always Initiate Address Phase**: Removed the `is_first_group` guard around `master_write` and `master_read` inside `execute_write_group`, `execute_read_group`, and their async variants. This ensures the master always programs the START/RESTART condition and sends the target address (SAD+W or SAD+R) even for zero-length groups that are not the first in a transaction.
2. **Deterministic Sequencing**: Bypasses the byte data transfer loops for zero-length groups, instead waiting for transfer completion (`wait_tc`) and cleanly transitioning to a STOP condition (on last group) or returning to allow a RESTART condition (on intermediate groups).
3. **Delegation**: Modified `write_vectored` and `blocking_write_vectored` to delegate empty buffers to `write_internal` to execute a zero-length address-only transaction.
4. **Documentation**: Added doc examples to `transaction` and `blocking_transaction` showing how to perform I2C bus scanning via zero-length writes.

#### Testing
The changes have been validated through local compilation, Clippy, and formatting checks:
- Compiled `embassy-stm32` cleanly for STM32L4 (`stm32l476vg`, blocking & async paths) and STM32H7 (`stm32h743zi`, async DMA-heavy paths).
- Verified zero clippy warnings/errors (`cargo clippy --features stm32l476vg,time-driver-any,exti -- -D warnings`).
- Formatting verified with `cargo fmt --all --check`.
- Verified the state machine sequencing for all cases by inspection:
  - Single zero-length write (only op): START + SAD+W → ACK/NACK → STOP
  - Zero-length write as last of N: RESTART + SAD+W → ACK/NACK → STOP
  - Zero-length write as intermediate: RESTART + SAD+W → ACK/NACK → wait for next group (matches non-empty intermediate group sequencing with data phase skipped)
  - Zero-length write as first of N: START + SAD+W → ACK/NACK → wait for next group
  - Equivalent read-side sequencing verified for `Stop::Automatic`/`Stop::Software`

> [!WARNING]
> **Physical Hardware Validation**:
[...]

PR #6353, opened 2026-06-15
Fix/i2cv2 zero length transfers

### LINKED ISSUE #2298: stm32/i2cv2: allow zero-length transfers.
I2C has checks for zero-length transfers, which return an error. These should instead do the right thing (send the address, check for ACK, then do a stop condition).

This is useful for I2C scanning all addresses.

anlif:
Fixes #2298. Supersedes #6336, which was closed for not being hardware tested.

Zero-length transfers returned `Error::ZeroLengthTransfer` instead of sending
the address, check ACK/NACK, issue a STOP, so bus scanning was impossible.

## What changed

**Master.** An empty operations slice is a no-op returning `Ok(())`. A zero-length group always
emits the address phase, not only when it is the first group, and waits for TC so a following
group's RESTART cannot race it. `write_vectored` with an empty outer slice no longer errors.

**Slave.** An empty `respond_to_write` accepts zero bytes. An empty `respond_to_read` sends
`0xFF` filler: I2C cannot encode "nothing to send" once the slave has ACKed an address with R=1
and become the transmitter, and `0xFF` releases SDA on every bit so the master can still form
its STOP. Both previously returned without clearing ADDR — which occurs in exactly one place in
`v2.rs`, inside `slave_start` — leaving SCL held low and the bus unusable for every device on
it. `blocking_respond_to_write` reported `Ok(0)` while doing so. The async forms tripped
`assert!(mem_len > 0)` in the DMA layer first. embedded-hal's `transaction` contract does not
define zero-length behaviour, so the chosen semantics are documented on all four methods.

**The first commit is two unrelated pre-existing bugs**, kept separate so they survive a revert
of the zero-length work. `write_dma_internal` re-enabled TCIE only on its reload path: when the
write group is not the first group, the previous group leaves TC set, enabling TCIE fires the
interrupt immediately, `on_interrupt` disables it again, and the TC ending the data phase never
wakes the future. And `write_dma_internal_slave` did not enable ADDRIE, so a master RESTART
mid-DMA was ignored until the timeout expired.

Neither involves a zero-length transfer. Both reproduce on `main` with
`transaction(addr, [Read(&mut [0u8; 1]), Write(&[0xAB])])` — against a pristine bus the very
first attempt times out:

```
### A: async [Read(1), Write(1)] - no zero-length involved
### A0 begin
### A0: Timeout          <-- origin/main
### A0: Ok, read 0xff    <-- this branch
```

## Hardware test

NUCLEO-H503RB master ↔ NUCLEO-H533RE async DMA slave at `0x42`, I2C2 on PB10 (SCL) / PB3 (SDA),
50 kHz, internal pull-ups. 14 master-side cases per round covering zero-length reads, writes and
empty operation slices on both the blocking and async paths, plus ordinary
write / read / write_read as regression guards. Slave-side cases read SCL/SDA out of `GPIOB.IDR`
so a stretched bus is observed directly rather than inferred from a timeout.

## Notes for reviewers

- `Error::ZeroLengthTransfer` is now produced nowhere in `v2.rs` or `v1.rs`. I left the variant
  in place.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -12,6 +12,18 @@ use stm32_metapac::i2c::vals::{Addmode, Oamsk};
 use super::*;
 use crate::pac::i2c;
 
+/// Bytes a slave transmits when it is read but has nothing to send.
+///
+/// I2C has no encoding for "nothing to send": once a slave ACKs an address with R=1 it is the
+/// transmitter and must drive the MSB of a data byte at the next SCL low. `0xFF` leaves SDA
+/// released for every bit, which is both what an idle bus reads as and what lets the master
+/// generate its STOP or repeated START at any point — driving a `0` bit would hold SDA low and
+/// prevent the master from forming a STOP at all.
+///
+/// A master that reads past this length under-runs the slave, which also releases SDA, so the
+/// wire keeps reading `0xFF`. The length only bounds how much is queued, not what is seen.
+const SLAVE_READ_FILLER: [u8; 16] = [0xFF; 16];
+
 impl From<AddrMask> for Oamsk {
     fn from(value: AddrMask) -> Self {
         match value {
@@ -600,10 +612,6 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
         addr: impl Into<Address>,
         operations: &mut [Operation<'_>],
     ) -> Result<(), Error> {
-        if operations.is_empty() {
-            return Err(Error::ZeroLengthTransfer);
-        }
-
         let address = addr.into();
         let timeout = self.timeout();
 
@@ -661,11 +669,13 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
 
         if total_bytes == 0 {
             // Handle empty write group - just send address
-            if is_first_group {
-                Self::master_write(self.info, address, 0, Stop::Software, false, !is_first_group, timeout)?;
-            }
+            Self::master_write(self.info, address, 0, Stop::Software, false, !is_first_group, timeout)?;
             if is_last_group {
+                self.wait_tc(timeout)?;
                 self.master_stop();
+                self.wait_stop(timeout)?;
+            } else {
+                self.wait_tc(timeout)?;
             }
             return Ok(());
         }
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/i2c/v2.rs
```diff
@@ -596,15 +596,33 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
...
+    ///
+    /// for addr in 0x08..0x78 {
+    ///     // A zero-length write transaction will send the address and check for ACK/NACK
+    ///     // without transferring any data bytes.
+    ///     let mut ops = [Operation::Write(&[])];
+    ///     if i2c.blocking_transaction(addr, &mut ops).is_ok() {
+    ///         defmt::info!("Device found at address 0x{:02x}", addr);
+    ///     }
+    /// }
+    /// # } }
```
xoviat: This comment is too specific for this use case.

embassy-stm32/src/i2c/v2.rs
```diff
@@ -1238,15 +1256,34 @@ impl<'d, IM: MasterMode> I2c<'d, Async, IM> {
...
+    /// let mut i2c: I2c<'_, _, _> = ...;
+    ///
+    /// for addr in 0x08..0x78 {
+    ///     // A zero-length write transaction will send the address and check for ACK/NACK
+    ///     // without transferring any data bytes.
+    ///     let mut ops = [Operation::Write(&[])];
+    ///     if i2c.transaction(addr, &mut ops).await.is_ok() {
+    ///         defmt::info!("Device found at address 0x{:02x}", addr);
+    ///     }
+    /// }
```
xoviat: as is this comment

xoviat (changes requested):

embassy-stm32/src/i2c/v2.rs
anlif: agreed, will remove

embassy-stm32/src/i2c/v2.rs
anlif: yep

anlif:
Think I found a bug in the implementation for empty read + write transactions. Trying to implement a loopback test example to verify, and fix it.
- [x] Verify and fix empty+non-empty transaction group

Dirbaio:

anlif:
Hey! Been on vacation a while, but I'll have a new look at this tomorrow. Have a couple of new H5 micros that I'm currently testing i2c on, so should be pretty quick to get this verified before it's ready for review.

xoviat (approved):
Thanks!

PR #6364, opened 2026-06-16
Add I2C reset function

kevswims:
Adds a reset method to the I2C to be able to recover from spurious start conditions on the bus.

### DIFF
--- a/embassy-stm32/src/i2c/v2.rs
+++ b/embassy-stm32/src/i2c/v2.rs
@@ -139,9 +139,14 @@ impl<'d, M: Mode, IM: MasterMode> I2c<'d, M, IM> {
         self.info.regs.cr2().write(|w| w.set_stop(true));
     }
 
+    /// Resets the I2C peripheral state machine while preserving configuration.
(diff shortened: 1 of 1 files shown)


xoviat:
With Claude's help I determined that this function basically copies the existing `soft_reset()` function in this file, but unlike `soft_reset()`, this implementation is unsound because the register read is not guaranteed to always wait 3 APB cycles. Basically you need to check what calls `soft_reset()` and determine whether it needs to be exposed publicly, or whether it can be accessed indirectly with an existing public method.

kevswims:
@xoviat thanks for the review. I had written this against an older embassy release that did not have the soft_reset function. I took a look through and it does not look like any of the current calls to soft_reset will deal with the issue I ran into. 

I was seeing the bus hang indefinitely when a device on it was reset it would trigger a spurious start condition which the I2C peripheral never clears. All operations after that will just timeout until the PE reset sequence occurs. I'll add some documentation to the soft_reset function and expose it publicly.

PR #6371, opened 2026-06-17
[MCXA] Add some sgi functionality

diondokter:
Closes #5612

Based on open PR and did some small cleanups.
I don't like the driver, but let's just get it merged and fix later as needed


embassy-mcxa/src/sgi/mod.rs
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    /// Fills SHAFIFO register using DMA in software start mode, handles misalignment and awaits completion in async manner.
+    /// Returns immediately after starting the DMA transfer; caller should await completion and check for errors.
+    fn fill_sha2_fifo_dma_start<'a, 'd_dma>(
+        &mut self,
+        dma_ch: &'a mut DmaChannel<'d_dma>,
+        data: &[u8],
+        length: usize,
+    ) -> Result<Transfer<'a>, SgiError> {
```
felipebalbi, line 408: technically this is only useful when we have `Dma`, but we can fix this later.

felipebalbi (approved):

embassy-mcxa/src/sgi/mod.rs
alamfarjadf: > technically this is only useful when we have `Dma`, but we can fix this later.

Do we not have DMA HAL anymore?

In an older branch it was working fine on MCXA hardware.

PR #6377, opened 2026-06-17
C5 peri examples

usbalbin:


usbalbin:
I have now tested this.

I have used the CAN peripheral at work in a firmware and it seems to work fine on an actual bus(so far). The `can` example however, is only tested in loopback mode so far. The usart examples are verified with a logic analyzer

usbalbin:
@xoviat this should be ready now

PR #6379, opened 2026-06-17
C5 timer

usbalbin:
I do not get any output in the example - Likely fixed in https://github.com/embassy-rs/stm32-data/pull/865 (missing pin alternate functions)


usbalbin:
@xoviat  This has been tested by looking at the LED of the nucleo when running the example. I have not verified with a scope yet

PR #6380, opened 2026-06-18
C5: USB and HSE

usbalbin:
~~Needs https://github.com/embassy-rs/stm32-data/pull/865 for missing interrupts and the alternate function pins~~ - Done


usbalbin:
Tested `usb_serial` example and works fine

PR #6382, opened 2026-06-18
Add MCXA577 USB device support

bogdan-petru:
Adds USB device support for MCXA577's ChipIdea/EHCI USBHS controller, forced to full-speed operation, plus `embassy-usb` HID mouse and CDC ACM serial examples.

Depends on embassy-rs/nxp-pac#108 for generated MCXA577 USBHS/USBPHY PAC register blocks.

This currently pins `nxp-pac` to `bogdan-petru/nxp-pac@1a3fd38855dc1d9c81c3ecdf7be2188e363180f0` so it builds before the PAC PR merges. After embassy-rs/nxp-pac#108 merges, update this PR to the upstream `embassy-rs/nxp-pac` revision.


embassy-mcxa/src/usb/clock.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+/// Enable the USB controller and PHY clock gates in MRCC, select clock sources,
+/// and bring up the USBPHY for the EHCI device controller.
+///
+/// # Safety
+/// Must be called once during USB initialization with exclusive access to the
+/// MRCC and USBPHY. The SOSC / PHY reference clock must already be running.
+pub(crate) unsafe fn init_clocks_and_phy(cfg: &PhyConfig) {
```
felipebalbi: This function shouldn't exist. You should use the clock configuration helpers like every other driver.

embassy-mcxa/src/usb/clock.rs
```diff
@@ -0,0 +1,176 @@
+//! USB clock gating and USBPHY bring-up for the MCXA5xx USBHS controller.
+//!
+//! This mirrors the NXP MCUXpresso SDK `USB_DeviceClockInit` / `USB_EhciPhyInit`
+//! sequence, reduced to what a full-speed device needs.
+
+// Some bit constants and the deinit path document the hardware but are not all
+// exercised by the current full-speed device bring-up.
+#![allow(dead_code)]
```
felipebalbi: This should not be needed. At least keep the constants commented out. I also wonder if these constants aren't exposed through methods in the PAC already. Do we need to patch the PAC?

embassy-mcxa/src/ostimer.rs
```diff
@@ -182,7 +182,7 @@ impl Driver for OsTimer {
 }
 
 #[allow(non_snake_case)]
-#[interrupt]
+#[crate::pac::interrupt]
```
felipebalbi: looks like an unrelated change.

embassy-mcxa/Cargo.toml
```diff
@@ -59,15 +62,15 @@ grounded = { version = "0.2.1", features = ["cas", "critical-section"] }
 heapless = "0.9"
 maitake-sync = { version = "0.3.0", default-features = false, features = ["critical-section", "no-cache-pad"] }
 nb = "1.1.0"
-nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "a46946e7f5ba3b5d8136f2962400300d54a99d6c", features = ["rt"] }
+nxp-pac = { git = "https://github.com/bogdan-petru/nxp-pac.git", rev = "1a3fd38855dc1d9c81c3ecdf7be2188e363180f0", features = ["rt"] }
```
felipebalbi, line 69: You need to send a PR to the upstream PAC, it seems.

embassy-mcxa/src/usb/ehci.rs
```diff
@@ -0,0 +1,173 @@
...
+
+// ---- USBCMD bits ----
+/// Run/Stop.
+pub(crate) const USBCMD_RS: u32 = 1 << 0;
+/// Controller reset.
+pub(crate) const USBCMD_RST: u32 = 1 << 1;
+/// Setup tripwire.
+pub(crate) const USBCMD_SUTW: u32 = 1 << 13;
+/// Add dTD tripwire.
+pub(crate) const USBCMD_ATDTW: u32 = 1 << 14;
```
felipebalbi: it really looks like all the constants in this file should be provided by the PAC with accessor methods. Why aren't they?

felipebalbi (changes requested):

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+                r.clear_endptcomplete(complete);
+            }
+            for ep in 0..EP_COUNT {
+                let out = EndpointSlot::new(ep, Direction::Out);
+                let in_ = EndpointSlot::new(ep, Direction::In);
+                if setup & (1 << ep) != 0 || complete & out.endpoint_bit() != 0 {
+                    out.waker().wake();
+                }
+                if complete & in_.endpoint_bit() != 0 {
+                    in_.waker().wake();
```
felipebalbi: This wake loop only wakes the OUT-direction waker on a SETUP (`setup & (1 << ep)` -> `out.waker()`); the IN waker is only woken on an IN completion. But an in-flight EP0 IN transfer (the data phase of a control-IN, or the IN ZLP in `accept`/`accept_set_address`) relies on its own waker firing to re-run the `abort_on_setup` check in `ep_transfer`. When the host abandons that transfer and sends a new SETUP, hardware doesn't complete the pending IN dTD, so the IN waker never fires and the future never aborts. That's the 'control pipe gets stuck' case the `ControlPipe` docs warn about, and it shows up during ordinary enumeration retries. I think a SETUP needs to wake both directions of the affected endpoint (EP0 at minimum).

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+impl<T: Instance> interrupt::typelevel::Handler<T::Interrupt> for InterruptHandler<T> {
+    unsafe fn on_interrupt() {
+        let r = regs();
+        let status = r.usbsts() & r.usbintr();
+        // Acknowledge all handled sources (write-1-to-clear).
+        r.clear_usbsts(status);
+
+        if status & USBSTS_URI != 0 {
+            FLAG_RESET.store(true, Ordering::Relaxed);
+            BUS_WAKER.wake();
```
felipebalbi, line 644: On a bus reset we only wake `BUS_WAKER`, so any class endpoint with an outstanding `read`/`write` is never woken. The reset handler in `Bus::poll` flushes the endpoints, but `ENDPTFLUSH` doesn't write the dTD token back to memory or raise a completion interrupt, and `embassy-usb`'s reset handling doesn't call `endpoint_set_enabled(false)`, so nothing ever re-polls the endpoint future and the class task wedges. We should wake all `EP_WAKERS` here (and on suspend) so in-flight transfers re-evaluate and unwind. Note that after a flush the dTD reads back inactive with `remaining == len`, so the future would currently return `Ok(0)` rather than an error, so it's worth deciding what the right post-reset result is.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+}
+static EP_BUFFERS: DmaCell<EpBuffers> = DmaCell::new(EpBuffers {
+    buf: [[0; FS_MAX_PACKET as usize]; EP_COUNT * 2],
+});
+
+async fn ep_read(index: usize, buf: &mut [u8]) -> Result<usize, EndpointError> {
+    let slot = EndpointSlot::new(index, Direction::Out);
+    let bounce = ep_buffer_ptr(slot);
+    let cap = FS_MAX_PACKET as usize;
+    let n = ep_transfer(slot, bounce, cap, false).await?;
```
felipebalbi: This always primes a 64-byte transfer regardless of the endpoint's negotiated max packet size (the value stored in `EP_MAX_PACKET` is never used here). A ChipIdea dTD stays active until the byte count reaches 0 or a short packet arrives, so for any OUT endpoint with `mps < 64` a single full-size packet leaves the dTD active and `read` blocks until enough packets accumulate to reach 64. That breaks the 'one packet per read' contract for interrupt-OUT / HID-OUT endpoints and for a control pipe configured with `max_packet_size_0 < 64`. The shipped examples happen to use 64-byte OUT endpoints so they don't hit it. We should prime against the endpoint's configured max packet size instead of `FS_MAX_PACKET`.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+        #[cfg(feature = "defmt")]
+        defmt::trace!(
+            "usb: stall ep{=usize} dir={=u8} stalled={=u8}",
+            slot.index,
+            slot.dir_num(),
+            stalled as u8
+        );
+        r.modify_endptctrl(slot.index, |v| {
+            let bit = endpoint_stall_bit(slot);
+            if stalled { v | bit } else { v & !bit }
```
felipebalbi: When clearing a stall (`stalled == false`) we only clear the stall bit; we don't reset the data toggle. USB 2.0 section 9.4.5 requires `CLEAR_FEATURE(ENDPOINT_HALT)` to reset the toggle to DATA0, and embassy-usb services that purely through `endpoint_set_stalled(addr, false)`. Without also asserting `RXR`/`TXR` here (as `configure_endpoint` already does), the device and host toggles desync after a halt/clear-halt cycle and subsequent transfers get silently dropped. That's the standard bulk error-recovery path, so it will bite. NXP's own `USB_DeviceEhciEndpointUnstall` resets the toggle on unstall.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+    }
+
+    let r = regs();
+    let prime_bit = slot.endpoint_bit();
+
+    dma_write_barrier();
+
+    // Prime the endpoint. We allow only one outstanding dTD per endpoint
+    // direction, so there is no need to use the add-dTD tripwire (ATDTW).
+    r.set_endptprime(prime_bit);
```
felipebalbi, line 1244: Once the endpoint is primed here, the transfer isn't cancel-safe: if the future is dropped before completion (e.g. wrapped in `with_timeout`/`select`, or a control future dropped when `bus.poll()` wins the `select` in `run_until_suspend` on a suspend/resume), the dTD stays ACTIVE and the controller still owns the bounce buffer. The next transfer on this endpoint direction rewrites the descriptor/buffer and re-primes, potentially while the controller is still DMA-ing the abandoned one. A drop guard that flushes the endpoint (and waits for `ENDPTFLUSH` to drain) before releasing the dTD would make this safe.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+                };
+                if r.usbcmd() & USBCMD_SUTW != 0 {
+                    break bytes;
+                }
+            };
+            r.modify_usbcmd(|v| v & !USBCMD_SUTW);
+            // Acknowledge the setup status.
+            r.clear_endptsetupstat(stat);
+            r.modify_endptctrl(0, |v| v & !(EPCTRL_TXS | EPCTRL_RXS));
+            FLAG_SETUP.store(false, Ordering::Relaxed);
```
felipebalbi: Clearing `ENDPTSETUPSTAT` after the SUTW copy loop leaves a race: a second SETUP arriving in the window between the tripwire confirming this packet and this `clear_endptsetupstat(stat)` re-asserts the status bit, which the write-1-to-clear then discards. The next `setup()` sees `endptsetupstat & 1 == 0` and parks forever. The ChipIdea-recommended sequence clears `ENDPTSETUPSTAT` before the SUTW read loop precisely so a racing SETUP re-sets the bit and is preserved. I'd also drop the `FLAG_SETUP.store(false)` here and treat the hardware status bit as the single source of truth.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+    let r = regs();
+    let prime_bit = slot.endpoint_bit();
+
+    dma_write_barrier();
+
+    // Prime the endpoint. We allow only one outstanding dTD per endpoint
+    // direction, so there is no need to use the add-dTD tripwire (ATDTW).
+    r.set_endptprime(prime_bit);
+    // Wait until the controller has acknowledged the prime.
+    while r.endptprime() & prime_bit != 0 {}
```
felipebalbi: This is a synchronous busy-wait on a hardware bit in an async path (there are similar ones at the flush inside the `poll_fn` below, and `while r.endptprime() != 0 {}` in the reset handler). On a single-threaded executor, if the bit never clears (transfer primed while `RS=0`, host-held reset signaling, a clock/PLL fault) it wedges every task, not just USB, including any watchdog feeder. These steady-state/reset-path spins should be bounded and unwind to an error rather than hang. The init-time spins in `clock.rs` are fine since they match the existing clock bring-up style; this is specifically about the loops on the live traffic path.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+        r.set_usbintr(0);
+        r.clear_usbsts(ALL_USBSTS);
+        UsbInterrupt::disable();
+
+        // SAFETY: dropping the bus is the terminal owner path for the USBHS
+        // peripheral. Embassy's USB device graph drops endpoint/class handles
+        // with the bus, so no further endpoint transfer should touch the gated
+        // controller registers after this point.
+        unsafe {
+            clock::deinit_clocks_and_phy();
```
felipebalbi: Two concerns with tearing the clocks/PHY down in `Bus::drop`. First, it reaches into `MRCC0` directly instead of going through the clocks subsystem, the same boundary issue as the bring-up but on teardown. Second, Rust doesn't guarantee the `Endpoint`s are dropped before the `Bus` in the embassy-usb device aggregate, so a suspended `ep_transfer` future can be polled after the clock is gated and fault when it reads `ENDPTCTRL`. The SAFETY note asserts this won't happen but nothing enforces it. I'd move clock/PHY teardown to the `Driver`/`Peri` lifetime (matching nrf/stm32, which never deinit from `Bus`) and limit `Bus::drop` to stopping the controller.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+            config.phy.pll_div_sel,
+            config.phy.d_cal,
+            config.phy.txcal45dp,
+            config.phy.txcal45dm
+        );
+        // SAFETY: we own the USB peripheral and bring up its clocks/PHY once.
+        unsafe {
+            clock::init_clocks_and_phy(&config.phy);
+        }
+        reset_controller();
```
felipebalbi, line 852: `Driver::new` resets the per-instance `alloc_*` bitmaps but not the module statics. After a `Bus` is dropped the `Peri` is free again, and a second `new` starts with stale `EP_TYPE`/`EP_MAX_PACKET`, possibly-latched `FLAG_*`, leftover QH/dTD tokens, and registered `EP_WAKERS`. That will produce spurious reset/setup events and mis-typed endpoints on re-init. Either zero the USB statics here (an explicit `reset_state()`), or gate construction on a one-shot claim and still re-zero the side tables.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+
+// =========================================================================
+// Instance trait
+// =========================================================================
+
+trait SealedInstance {}
+
+/// USB peripheral instance.
+#[allow(private_bounds)]
+pub trait Instance: SealedInstance + PeripheralType + 'static {
```
felipebalbi: The `Instance`/`SealedInstance` machinery doesn't really do anything here: `SealedInstance` is empty, registers come from the hardwired `regs() -> UsbRegs(USB1)` free function, and the `type UsbInterrupt = USB1_HS` alias below is what the `Bus`/enable/disable paths actually use instead of `T::Interrupt`, so the generic `T` can't select anything. It also drops the `+ Send` bound the other embassy-mcxa `Instance`s carry. I'd either make this real by tying `SealedInstance` to `clocks::Gate` with a `regs()` accessor (which would also fold the clock bring-up into the standard path), or drop the trait entirely and take `Peri<'d, USB1>` concretely.

embassy-mcxa/src/usb/clock.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub txcal45dp: u32,
+    /// `TXCAL45DM` trim value.
+    pub txcal45dm: u32,
+    /// PLL divider select for the reference clock (2 = 24 MHz reference).
+    pub pll_div_sel: u32,
+}
+
+impl Default for PhyConfig {
+    fn default() -> Self {
+        Self::frdm_mcxa577()
```
felipebalbi: Making `frdm_mcxa577()` the `Default` means `Config::default()` bakes one specific board's analog trims into the generic HAL, so any other MCXA5xx board that uses the default silently gets FRDM-MCXA577 calibration. Board calibration like this usually lives in the example/BSP, with the HAL default being the reset/nominal values (or no default at all). The raw `u32` fields are also silently masked with `& 0xF` at use, so an out-of-range trim is swallowed rather than rejected; narrower types or a validation step would catch that.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+// mutable access. The controller may access the memory concurrently as DMA, so
+// all synchronization with hardware is done with explicit barriers.
+unsafe impl<T> Sync for DmaCell<T> {}
+
+// ---- DMA-visible controller structures (statically allocated) ----
+//
+// MCXA577 SRAM is treated as non-cacheable by the current MCX-A target setup.
+// If this driver is reused on a target with data-cacheable SRAM, these
+// descriptors and bounce buffers must either be placed in a non-cacheable
+// section or wrapped with explicit cache clean/invalidate operations.
```
felipebalbi: This non-cacheable assumption is load-bearing but completely unguarded: the DMA structures rely on it with only `dsb()` barriers, no clean/invalidate, no dedicated linker section, and no assertion. If the part is ever ported, or an app enables a data cache covering `QH_LIST`/`DTD_LIST`/`EP_BUFFERS`, you get silent descriptor/payload corruption with no fault. I'd put these in a dedicated non-cacheable section and/or add a `debug_assert!` on the region attribute, and keep clean/invalidate stubs next to the barriers so a port can't miss them.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+                FLAG_SETUP.store(false, Ordering::Relaxed);
+                let complete = r.endptcomplete();
+                r.clear_endptcomplete(complete);
+                while r.endptprime() != 0 {}
+                r.set_endptflush(ALL_ENDPOINT_BITS);
+                r.set_deviceaddr(0);
+                init_control_qh(self.control_max_packet_size);
+                return Poll::Ready(Event::Reset);
+            }
+            if FLAG_RESUME.swap(false, Ordering::Relaxed) {
```
felipebalbi: Draining the three event flags in fixed priority (reset, then resume, then suspend) can reorder events: if a real suspend->resume happens between two polls, both flags are set and we report Resume before Suspend, leaving the stack `suspended` while the bus is live until the next resume. Modeling the bus event off the observed `PORTSC1` state and emitting edges in occurrence order would avoid this; if we keep it simple, it's worth a comment that suspend/resume edges are best-effort.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+        }
+        if status & USBSTS_SLI != 0 {
+            FLAG_SUSPEND.store(true, Ordering::Relaxed);
+            BUS_WAKER.wake();
+        }
+        if status & USBSTS_PCI != 0 {
+            // Port change: treat as resume notification.
+            FLAG_RESUME.store(true, Ordering::Relaxed);
+            BUS_WAKER.wake();
+        }
```
felipebalbi, line 670: Mapping every port-change interrupt to Resume is pretty loose: PCI also fires on connect, port-enable changes, and speed changes, so we'll emit spurious `Resume` events (e.g. at initial attach before any suspend). It'd be more accurate to gate this on actually being suspended, or to derive suspend/resume from the `PORTSC1` suspend bit rather than the raw PCI latch.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+                    out.waker().wake();
+                }
+                if complete & in_.endpoint_bit() != 0 {
+                    in_.waker().wake();
+                }
+            }
+        }
+
+        // Mask further interrupts are not needed: sources are level and cleared above.
+        T::Interrupt::unpend();
```
felipebalbi: Calling `unpend()` at the end of the handler is risky for level/W1C sources: if a new event re-latches the NVIC pending bit while the handler is running, this clears it before its `USBSTS` flag was sampled, dropping the wake. For level-triggered sources the write-1-to-clear ack above is sufficient and this is redundant at best. I'd drop it (the nrf driver relies on level + W1C); if masking is ever needed, mask `USBINTR` instead of poking NVIC pending.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+            clock::deinit_clocks_and_phy();
+        }
+    }
+}
+
+impl embassy_usb_driver::Bus for Bus {
+    async fn enable(&mut self) {
+        let r = regs();
+        // Enable the interrupt sources we handle.
+        r.set_usbintr(USBSTS_UI | USBSTS_UEI | USBSTS_PCI | USBSTS_URI | USBSTS_SLI);
```
felipebalbi, line 1511: We enable `USBSTS_UEI` but there's no dedicated handling for it: error interrupts only get noticed because every dTD sets IOC, so `UI` co-fires and the completion path reads the error bits. A `UEI` without a coincident `UI` (any future non-IOC dTD) would wake nobody. There's also no error counter and the only visibility is a `defmt`-gated `warn!`, so a field failure is invisible without defmt. A real `UEI` branch plus a couple of `AtomicU32` counters (errors, resets, suspend/resume) would help a lot for diagnosis.

embassy-mcxa/src/usb/clock.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let mut locked = false;
+    for _ in 0..4_000_000 {
+        if phy_read!(PLL_SIC) & PLL_LOCK != 0 {
+            locked = true;
+            break;
+        }
+        cortex_m::asm::nop();
+    }
+    assert!(locked, "USB PHY PLL did not lock");
```
felipebalbi: This panics inside an infallible `Driver::new` if the PLL doesn't lock (wrong board voltage/SOSC). 'Fail loud' is defensible, but since this is the one failure a misconfigured board will actually hit, a fallible `new` returning `Result` would let applications report or degrade instead of panicking during init.

embassy-mcxa/src/usb/clock.rs
```diff
@@ -0,0 +1,176 @@
...
+/// # Safety
+/// Must be called once during USB initialization with exclusive access to the
+/// MRCC and USBPHY. The SOSC / PHY reference clock must already be running.
+pub(crate) unsafe fn init_clocks_and_phy(cfg: &PhyConfig) {
+    use crate::pac::mrcc::{ClkdivHalt, ClkdivReset, ClkdivUnstab, PhyClkselMux, UsbClkselMux};
+    use crate::pac::{MRCC0, SPC0};
+
+    // Match the FRDM-MCXA577 SDK USB init delay setting before enabling the
+    // USB clocks. Voltage policy remains owned by the clock driver.
+    SPC0.active_vdelay().write(|w| w.set_active_vdelay(0x0500));
```
felipebalbi: Writing `SPC0.active_vdelay` from the USB PHY bring-up pokes a system-power-controller register, even though the comment just above says voltage policy stays with the clock driver, and the examples also set the Vdd level via `clock_cfg`, so it's configured in two places. Any USB-required SPC/voltage timing would be better expressed once, in the power/clock configuration path that owns it.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+                "usb: transfer error ep{=usize} dir={=u8} token=0x{=u32:08x}",
+                slot.index,
+                slot.dir_num(),
+                token
+            );
+            return Poll::Ready(Err(EndpointError::Disabled));
+        }
+        // Remaining bytes are in token[30:16]; transferred = requested - remaining.
+        let remaining = (token >> DTD_TOKEN_TOTAL_SHIFT) & DTD_TOKEN_TOTAL_MASK;
+        Poll::Ready(Ok(len - remaining as usize))
```
felipebalbi: `len - remaining` can underflow if hardware ever reports `remaining > len` (debug panic, or release wrap surfacing as a bogus `BufferOverflow`). The error mask is checked first so a clean completion should have `remaining <= len`, but `len.saturating_sub(remaining as usize)` is cheap insurance.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+    async fn accept(&mut self) {
+        if control_setup_pending() {
+            return;
+        }
+        // Status stage: send a zero-length IN packet.
+        // EP0 IN owned by control pipe.
+        let _ = ep_zlp(0, Direction::In, true).await;
+    }
+
+    async fn reject(&mut self) {
```
felipebalbi, line 1453: `reject()` stalls EP0 without flushing any in-flight EP0 transfer first, so a primed dTD can be left active on the control endpoint when we stall. Once the SETUP-abort wake issue is sorted this is mostly moot, but flushing before stalling would keep EP0 state clean.

embassy-mcxa/src/usb/mod.rs
```diff
@@ -0,0 +1,1263 @@
...
+
+        // SAFETY: enabling the controller interrupt.
+        unsafe {
+            UsbInterrupt::unpend();
+            UsbInterrupt::enable();
+        }
+
+        // Attach: set Run/Stop.
+        r.modify_usbcmd(|v| v | USBCMD_RS);
+        self.inited = true;
```
felipebalbi: Minor: `inited` is written here and in the first `poll()`, but the stack always polls before it calls `enable()` (it only enables in response to `PowerDetected`), so this write is effectively dead and the field really means 'PowerDetected already emitted'. Renaming it (e.g. `power_detected`) and dropping this line would make the intent clearer.

felipebalbi (changes requested):

embassy-mcxa/src/usb/clock.rs
bogdan-petru: Clock-source selection and the PHY reference divider moved into UsbhsConfig: SPConfHelper in clocks/periph_helpers.rs, and the driver now brings the peripheral up via clocks::enable_and_reset::<USB1>().

embassy-mcxa/src/usb/clock.rs
bogdan-petru: PAC was patched to expose USBPHY ([embassy-rs/nxp-pac#108](https://github.com/embassy-rs/nxp-pac/issues/108)), and the PHY sequence now uses typed accessors (set_SFTRST, set_PLL_REG_ENABLE, PLL_LOCK, set_D_CAL).

embassy-mcxa/src/ostimer.rs
bogdan-petru: reverted

embassy-mcxa/Cargo.toml
bogdan-petru: https://github.com/embassy-rs/nxp-pac/issues/108 adds the MCXA577 USBHS/USBPHY registers.

embassy-mcxa/src/usb/ehci.rs
bogdan-petru: The MMIO register masks were removed from this file — those registers now go through PAC accessors. What remains is only the dQH/dTD RAM data-structure layout (capability/token bit positions), which the controller DMAs from memory and the PAC doesn't model as registers. The module header now documents that distinction.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: SETUP now wakes both the OUT and IN wakers of the endpoint, so an in-flight EP0 IN transfer re-runs its abort_on_setup check and unwinds with Disabled when the host abandons it. The handler deliberately doesn't clear ENDPTSETUPSTAT, so the woken future still observes the pending SETUP.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Reset and suspend now wake all EP_WAKERS and bump a transfer-abort epoch; outstanding read/write futures observe the epoch change, flush their endpoint, and return Err(EndpointError::Disabled) rather than Ok(0) or wedging. That's the post-reset result we settled on since the transfer state is unknowable after a flush.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: OUT/control priming now uses the endpoint's stored max packet size (endpoint_max_packet), restoring one-packet-per-read for mps < 64.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: endpoint_set_stalled(_, false) now asserts RXR/TXR along with clearing the stall bit, resetting the toggle to DATA0.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: ep_transfer now arms a TransferGuard before awaiting; if the future is dropped mid-flight, the guard flushes the endpoint (waiting for ENDPTFLUSH to drain) before the dTD/buffer can be reused. The prime→arm window contains no await point, so there's no gap.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Ack is now issued before the SUTW copy loop. A racing SETUP re-asserts the status bit and is preserved for the next setup(). FLAG_SETUP is gone; the hardware bit is the single source of truth everywhere (control_setup_pending()).

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Fixed — the live-path spins are bounded via wait_register_clear (prime-ack, flush, the reset-path prime drain) and unwind to EndpointError::Disabled (or log-and-continue in the reset path). The SUTW tripwire retry in setup() is also bounded now. Init-time spins in Driver::new remain simple loops, matching the existing clock bring-up style as you noted.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Driver::new now calls reset_static_state(), which clears the flags and counters, resets EP_TYPE/EP_MAX_PACKET, reinitializes the QH/dTD lists and bounce buffers, bumps the abort epoch, and wakes any stale wakers, so a second new() starts clean.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Dropped it entirely. Driver, InterruptHandler, and new() now take Peri<'d, USB1> and the USB1_HS interrupt concretely; the clock bring-up folds into the standard Gate/enable_and_reset path. The examples' bind_interrupts! lost the type parameter accordingly.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Bus::drop now only stops the controller (clear RS, mask+ack interrupts, NVIC disable); the direct MRCC0 access and the PHY teardown are gone, which removes the drop-ordering hazard. Following nrf/stm32 as you suggested, I've left the clocks gated-on for the borrowed USB1 lifetime rather than gating off on drop (documented on Bus::drop; Driver::new re-runs enable_and_reset, so re-init is clean).

embassy-mcxa/src/usb/clock.rs
bogdan-petru: PhyConfig::default() is now nominal/reset values; frdm_mcxa577() stays as a named constructor and the examples opt into it explicitly. The trim fields were narrowed to u8 and are range-validated in init_phy (rejected as InitError::PhyConfig) instead of silently masked.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Driver::new now debug-asserts that SCB CCR.DC is clear, so enabling a data cache without adding real maintenance fails loudly. dma_clean/dma_invalidate stubs sit at every prime/read-back site as the single porting hook. I initially tried a dedicated link section, but since no linker script placed it, that was a placement hazard rather than a guarantee, so the statics stay in regular (non-cacheable on this part) SRAM with their explicit alignment.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Resume is now only flagged on an edge out of an observed suspended state (so a suspend→resume pair always emits in order), and a reset clears any pending suspend/resume flags in both the ISR and Bus::poll, so the stack can't be left suspended on a live bus.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: The PCI branch now emits Resume only when SUSPENDED.swap(false) observes a true→false edge, so connect/port-enable/speed changes at attach no longer produce spurious Resume events.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: The W1C USBSTS ack is the sole acknowledgment in the handler; the only remaining unpend is the stale-pending clear in Bus::enable before enable(), which doesn't have this hazard.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: There's a dedicated UEI branch now: it counts the error and wakes all transfer waiters (without aborting healthy endpoints — the failing transfer reports through its own dTD error bits), so a UEI without a coincident UI no longer wakes nobody. Added AtomicU32 counters for errors/resets/suspends/resumes, exposed via diagnostic_counters() so failures are visible without defmt.

embassy-mcxa/src/usb/clock.rs
bogdan-petru: Driver::new returns Result<Self, InitError> (PhyPllLock, PhyConfig, Clock(...)). The PLL-lock poll is bounded and returns an error instead of panicking; the examples handle the Result at the call site.

embassy-mcxa/src/usb/clock.rs
bogdan-petru: Removed. The PHY bring-up touches only USB1_HS_PHY registers now; voltage/active_vdelay policy lives solely in the power/clock configuration path.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Fixed - len.saturating_sub(remaining as usize).

embassy-mcxa/src/usb/mod.rs
bogdan-petru: reject() flushes both EP0 directions (and clears the dTD/QH overlays) before asserting the stall bits.

embassy-mcxa/src/usb/mod.rs
bogdan-petru: Renamed to power_detected and the dead write in enable() removed; it's now written only where the initial PowerDetected event is emitted.

PR #6388, opened 2026-06-19
fixed RTC wr protection dis bits

yourongtang:
corrected bits in RTC STATUS register for WE


gjpmsft:
@felipebalbi  FYI

felipebalbi:
bender run

PR #6391, opened 2026-06-20
embassy-rp/cyw43-pio: fix input_sync_bypass for pins >= 32 (GPIOBASE offset)

cedrivard:
## Problem

`Pio::Pin::set_input_sync_bypass` builds the bitmask from the absolute pin
number (`1 << self.pin()`), but `input_sync_bypass` is indexed *relative to the
PIO's `GPIOBASE`* (0 or 16 on RP235x). On RP235xB with `GPIOBASE = 16`, a pin
>= 32 (e.g. GPIO 38) shifts by the full pin number: it panics in debug, and in
release wraps to the wrong bit (38 -> bit 6).

This breaks `cyw43-pio` on boards that wire the CYW43 data line to a high pin.
On the SparkFun IoT RedBoard RP2350 (RP2350B; CYW43 on GPIO 36/37/38),
`cyw43::new()` hangs: without a working sync bypass the PIO samples stale data
and the link never comes up.

## Fix

**1. `embassy-rp` — offset the mask by GPIOBASE.** Shift by `pin - offset`,
reading the offset (0/16) from the `GPIOBASE` register instead of assuming a
fixed value (resolving the 16..31 ambiguity from #5116).

**2. `cyw43-pio` — call it after `set_config`.** `GPIOBASE` is only written by
`StateMachine::set_config`, but the call was at pin creation, before
`set_config`, so the register read returned a stale 0. Moving it after
`sm.set_config(&cfg)` makes the offset correct.

## Testing

Verified on hardware (SparkFun IoT RedBoard RP2350, RP2350B): `cyw43::new()`
succeeds and WiFi associates, replacing a manual `INPUT_SYNC_BYPASS` write I'd
been carrying downstream.

Supersedes #5116 (same root cause); credit to @diogo464 for the original report
and the ordering insight, and @Dirbaio for the GPIOBASE review note.


CBJamo:
I don't like that this requires users to call `set_input_sync_bypass` after `set_config`, especially since making that mistake will not cause an obvious error, it'll just be broken. If it's possible, we should find a way to enforce that with the API. 

A brute-force solution would be to make `set_input_sync_bypass` panic if called before `set_config`, at least that way we could give a helpful error message, but if there's a way to make it a compile-time error that would be much preferred.

xoviat:
@CBJamo can you approve the tagged rp prs?

cedrivard:
@CBJamo Agreed, the ordering shouldn't be the user's problem. Pushed a version that removes it: `Config` now carries the bypass set and `set_config` applies it, right after it establishes `GPIOBASE`.

```rust
cfg.set_input_sync_bypass(&[&pin_io]);
sm.set_config(&cfg);
```

Offset is correct by construction, and bypass pins join the existing `<32`/`>=16` window check. `cyw43-pio` uses this and no longer touches the pin after `set_config`.

`Pin::set_input_sync_bypass` stays (it's the only way to clear a bypass) but now panics when the pin is outside the PIO's current `GPIOBASE` window, pointing at the `Config` method.

Retested on hardware (RP2350B, CYW43 on GPIO 36/37/38): WiFi associates.

embassy-rp/src/pio/mod.rs
```diff
@@ -631,6 +652,7 @@ pub struct Config<'d, PIO: Instance> {
     // PINCTRL
     pins: PinConfig,
     in_count: u8,
+    input_sync_bypass: u64,
```
CBJamo: Nitpick: Let's name this `input_sync_bypass_mask`, hopefully that will guide a user who tries to construct it by hand in the right direction.

Question: Should this be in `PinConfig`? I also wonder why `in_count` isn't in there, changing that can be done another time, but if you saw a reason while working in this file it'd be good to note it.

CBJamo (changes requested):
I like this much better, just a nitpick about naming and and I think the changes are good to go.

Would you also add a line to the changelog in embassy-rp? You'll have to rebase or there will be a merge conflict.

embassy-rp/src/pio/mod.rs
cedrivard: Renamed to `input_sync_bypass_mask`, and added the changelog line.

On `PinConfig`: I kept both out of it because `PinConfig` currently maps 1:1 onto the per-state-machine PINCTRL register (all seven fields are PINCTRL fields), and `set_config` shifts them as a unit against GPIOBASE.

Neither candidate belongs to that register: `in_count` is a SHIFTCTRL field (and RP2350-only), and INPUT_SYNC_BYPASS is a block-wide register shared by all four state machines rather than per-SM state.

Moving them in would break the "this struct = PINCTRL" invariant, so `Config` felt like the honest home.

Happy to revisit `in_count` separately if you'd like PinConfig to become "all pin-related config" instead.

CBJamo: Ah, I thought there'd be a reason, I just didn't immediately catch it. I agree with your reasoning.

CBJamo (approved):

PR #6402, opened 2026-06-23
N6 ADC Support

jake-taf:
Adds support for ADC on the N6. The N6 is mostly follows the U3, but has a 32-bit data register and different calibration steps. This has been tested on the disco N6 and a custom N6 board with blocking reads.

Twin ticket: https://github.com/embassy-rs/stm32-data/pull/868

> [!NOTE]  
> It's still a work in progress, as I need to refactor and clean it up

## ~~REFSC Fun~~

```Rust
RIFSC.risc_seccfgr(2).modify(|w| w.set_cfg(0, true));
```

~~Without the above line blocking and DMA reads return 2000. With the above line blocking reads return the correct value for the voltage, but DMA reads just return 0. I'm not sure if it's a driver issue or a security setting preventing DMA from working. I think it's security related though, since DMA could read the ADC data register fine when seccfgr wasn't set?~~

Fixed: DMA had to be marked secure

## Todo

- [x] Squash commits
- [x] Formatting
- [x] Update blocking reads to use a 32-bit pointer and convert it to 16-bit data
- [x] Update DMA reads to use blanking unless FIFO packing is requested
- [x] Is MAX_ADC_CLK_FREQ correct?
- [x] Clean up calibration (better comments, refactor, remove unnecessary delays)

Let me know if there's anything else that looks off.

### DIFF
--- a/embassy-stm32/src/adc/mod.rs
+++ b/embassy-stm32/src/adc/mod.rs
@@ -12,7 +12,7 @@
 #[cfg_attr(adc_l0, path = "v1.rs")]
 #[cfg_attr(adc_v2, path = "v2.rs")]
 #[cfg_attr(any(adc_v3, adc_g0, adc_h5, adc_h7rs, adc_u0), path = "v3.rs")]
-#[cfg_attr(any(adc_v4, adc_u5, adc_u3), path = "v4.rs")]
+#[cfg_attr(any(adc_v4, adc_u5, adc_u3, adc_n6), path = "v4.rs")]
 #[cfg_attr(adc_g4, path = "g4.rs")]
 #[cfg_attr(adc_c0, path = "c0.rs")]
 mod _version;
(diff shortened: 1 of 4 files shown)


jake-taf:
DMA works now. The fix was to mark the transfer as secure.

jake-taf:
@xoviat I have DMA and blocking read working with the methods we discussed. If this looks good I'll mark both as ready.

- Right now I just handled those two, as it's all I need for now. I can add more in future PR as needed.
- I gated some things behind the N6 (like 32-bit pointers), as I wasn't sure how other chips would handle them. Maybe that's not necessary?
- I found DMA only works when the buffer passed in is a multiple of 32-bits (e.g. `[u16; 1]` breaks, `[u16; 2]` works). The N6 will give a DMA user error otherwise. It truncates correctly and doesn't overwrite the unused elements, so maybe that's expected.

xoviat:
I'll look at this next week. Sorry for any delays.

jake-taf:
Not a problem :)

PR #6403, opened 2026-06-23
embassy-rp: Add interpolator driver + examples

### LINKED ISSUE #2042: rp2040 Interpolator Support
In my project I'm making heavy use of the rp2040's interpolators, which are currently unsupported in `embassy_rp`. When moving from `rp2040_hal` to `embassy_rp`, I reimplemented support by taking the code from `rp2040_hal::sio` and modifying it to work with `rp_pac` as its base, rather than `rp2040_pac`, but keeping the interface the same.
I'm happy to tidy it up and submit as a PR if:
1. Anyone is interested in having interpolator support in `embassy_rp`
2. People feel this is the right approach, rather than creating a different implementation from scratch.

orukusaki:
Adds a basic, but functionally complete, driver for the Interpolator peripheral in the rp2040 and rp235x chips.
Examples taken from the datasheet. 
Tested using RP Pico and Pico 2 boards respectively.

Fixes #2042

### DIFF
--- a/embassy-rp/src/lib.rs
+++ b/embassy-rp/src/lib.rs
@@ -37,6 +37,7 @@ mod float;
 pub mod gpio;
 pub mod i2c;
 pub mod i2c_slave;
+pub mod interpolator;
 pub mod multicore;
 #[cfg(feature = "_rp235x")]
 pub mod otp;
@@ -310,6 +311,9 @@ embassy_hal_internal::peripherals! {
 
     WATCHDOG,
     BOOTSEL,
+
+    INTERP0,
+    INTERP1,
 }
 
 #[cfg(feature = "_rp235x")]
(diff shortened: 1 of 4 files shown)


embassy-rp/src/lib.rs
```diff
@@ -447,7 +451,10 @@ embassy_hal_internal::peripherals! {
     BOOTSEL,
 
     POWMAN,
-    TRNG
+    TRNG,
+
+    INTERP0,
```
Dirbaio: Each core has their own private instances of INTERP0, INTERP1. We can't simply model them as global singletons because it causes two problems:

- It prevents using the INTERPx on the two cores at the same time
- It is Send, therefore you can instantiate it in one core, do things to it (e.g. configure it), then send it to the other core. it will appear that the configuration has been "undone" because you'll be talking to the *other* interpolator.

Image

This means we can't use the regular singletons. Unfortunately there's no precedent in embassy-rp for how to handle this special case.

I would do the following:

- Define the singleton structs manually, make them `!Send`.
- Make functions to "take" INTERP0/1 for core 0/1 at runtime. They check the current core number matches, then they check a global flag to see if it's already been taken. Panic if either check fails.

orukusaki:
Thanks for the feedback, I have implemented your suggestion. Please let me know if it's not how/where you would expect it to be done.

Dirbaio (approved):
looks great now. Thanks!

PR #6405, opened 2026-06-23
Fix embassy-boot async FirmwareUpdater incompatible buffer size for ed25519-dalek/salty

### LINKED ISSUE #6330: embassy-boot async FirmwareUpdater uses incompatible buffer size for ed25519-dalek
There's a very small buffer size being used by the `verify_and_mark_updated` method -

https://github.com/embassy-rs/embassy/blob/0f831da2b372cf7491a4685f645b89c5ed1ce769/embassy-boot/src/firmware_updater/asynch.rs#L106-L108

This is causing a `Flash(NotAligned)` error on the rp2040 due to the 4 byte read size -

https://github.com/embassy-rs/embassy/blob/0f831da2b372cf7491a4685f645b89c5ed1ce769/embassy-rp/src/flash.rs#L40-L41

An easy fix would be to increase this value to the same size as the message (i.e. 64) if that would be acceptable? 

A more involved fix would be to set it based on the existing `max` code -

https://github.com/embassy-rs/embassy/blob/0f831da2b372cf7491a4685f645b89c5ed1ce769/embassy-boot/src/firmware_updater/asynch.rs#L296-L298

N.B. I'm not using it but it looks like salty would have the same issue -

https://github.com/embassy-rs/embassy/blob/0f831da2b372cf7491a4685f645b89c5ed1ce769/embassy-boot/src/firmware_updater/asynch.rs#L126-L128

chrisprice:
Fixes https://github.com/embassy-rs/embassy/issues/6330. I've opted for the simple fix I described but happy to put more work in if you'd prefer.

### DIFF
--- a/embassy-boot/src/firmware_updater/asynch.rs
+++ b/embassy-boot/src/firmware_updater/asynch.rs
@@ -103,7 +103,7 @@ impl<'d, DFU: NorFlash, STATE: NorFlash> FirmwareUpdater<'d, DFU, STATE> {
             let public_key = VerifyingKey::from_bytes(_public_key).map_err(into_signature_error)?;
             let signature = Signature::from_bytes(_signature);
 
-            let mut chunk_buf = [0; 2];
+            let mut chunk_buf = [0; 64];
             let mut message = [0; 64];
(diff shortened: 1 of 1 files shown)


leftger (approved):
LGTM

Issue #6331, opened 2026-06-13
Please create an UART "data available" function


rhuys:
Sometimes you need a simple read operation from an interface, in this example, UART.

let mut uart = uart::Uart::new_blocking(...)
//snip

let mut buffer = [0u8; 16];
uart.blocking_read(&mut buffer); 
//  if you receive for example amessage of 15 bytes, this function will block the entire application.

Please include these methods (or something similar)

// check if there is any data on the fifo
uart.has_rx_data() -> bool ;    // return true if there is data on the fifo

// read data until a certain character is reached, or timeout occured (in ms)
uart.blocking_read_until(buffer, '\n', 200);

lulf:
PRs welcome (which HAL are you using?)

carlsuarez:
Hey @lulf  — I'm on STM32. PR is up: #6406

It adds both pieces from the issue to the non-buffered UART (with delegating methods on `Uart`): `has_rx_data()` for a non-blocking availability check, and a delimiter-based `blocking_read_until()`. Semantics and design notes are in the PR description.

I left the `timeout_ms` out on purpose — the blocking path has no time source, so a wall-clock timeout would pull `embassy-time` into a blocking call. Between `has_rx_data()` (caller composes a timeout with their own timer) and the existing `read_until_idle()`, that need seems covered. One open question, also in the PR: would you rather `blocking_read_until` stay delimiter-based as written, be an idle-line `blocking_read_until_idle` mirroring the async version, or both? Happy to adjust.

rhuys:
what generic crate do you recommend for blocking serial streams?  (with traits that are implemented by the drivers)

HenriqueDomiciano:
Guys, took the initiative to add this feature to embassy-rp, hope is the expected behavior and is useful, i added to UartRx struct but to adapt to UART struct must be easy 
Edit : Added to the UART struct

rhuys:
Hi @HenriqueDomiciano would love to test your contribution, can you give me advice how to ?  (not very familiar with the community and how to contribute to the code base)

HenriqueDomiciano:
Hi @rhuys on the PR I left a comment with some tests I executed, the name of the function changed but I will provide an example later, so it's clear how to use it, also I added tests to the PR so it's possible to see how it works from there. 

If you need to test just run the example i pushed on the latest commit, just running on the examples/rp 
`cargo run --bin uart_blocking_read_until`

I did an change on the function signature, but i think it improves usage. 

[...]

PR #6406, opened 2026-06-24
stm32/usart: add has_rx_data() and blocking_read_until() to the non-buffered UART

carlsuarez:
## Summary

Adds two small, additive methods to the non-buffered STM32 UART (`UartRx`, with
delegating methods on `Uart`) so callers of the blocking API don't stall on short
messages, per #6331:

- `has_rx_data() -> bool` — non-blocking check for whether a byte is available.
- `blocking_read_until(buf, delim) -> Result<usize, Error>` — blocking read that
  stops at a delimiter or when the buffer is full.

Both are inherent methods; no existing signatures change.

## `has_rx_data()`

A non-destructive peek of the RX-not-empty flag, cfg-gated for the `sr` (usart_v1/v2)
and `isr` (usart_v3/v4) register families. It does not consume data and does not clear
pending error flags (parity/framing/noise/overrun), so a following read still observes
them. On the `sr` families it also checks `buffered_sr`, so a byte latched while error
flags are still draining isn't missed.

Implemented as an inherent method rather than an `embedded_io::ReadReady` impl, to stay
consistent with the existing, deliberate decision not to expose embedded-io semantics on
the non-buffered RX (the doc note in `usart/mod.rs`). The buffered types keep their
`ReadReady` impls.

## `blocking_read_until()`

Reads one byte at a time until `delim` is received or `buffer` is full, returning the
number of bytes written. The delimiter is stored in the buffer and included in the count,
so the read bytes are always `&buffer[..n]`. A return of `buffer.len()` is ambiguous
(buffer filled vs. delimiter as the final byte); callers can disambiguate via
`buffer[n - 1]`, and this is noted in the method docs.

### On the `timeout_ms` from the issue

I left the millisecond timeout out deliberately: the blocking path has no time source, so
a wall-clock timeout would pull `embassy-time` into a blocking call. Two existing pieces
cover that need instead — `has_rx_data()` lets a caller bound the read with their own
timer, and `read_until_idle()` already returns early via hardware idle-line detection.

## Testing

HIL test added in `tests/stm32/src/bin/usart.rs` for `has_rx_data()`: asserts no data
before TX, data after TX, that peeking doesn't consume the byte, and that a following
`blocking_read` returns it. Happy to add a delimiter test for `blocking_read_until()` if
useful.

## Open question

Would you prefer `blocking_read_until` delimiter-based (as here), an idle-line
`blocking_read_until_idle` mirroring the async `read_until_idle`, or both? Glad to adjust.

Part of #6331


xoviat:
this was requested for rp, not stm32.

PR #6410, opened 2026-06-24
don't set split pins as analog

### LINKED ISSUE #6409: STM32H7 - Using PC2_C with ADC breaks OctoSPI.
This code breaks OctoSPI communications when using PC2_C as analog in and PC2 as OCTOSPIM_P1_IO2.
```
        let _ = adc.blocking_read(&mut p.PC2_C, SampleTime::Cycles325);
```

PC2 and PC2_C are separate pins on the STM32H735.

Image

After some investigation it seems that the code inside the adc implementation reconfigures the PC2 pin as analog, but it should not.

```
    fn reborrow_adc<'a>(&'a mut self) -> BorrowedAdcChannel<'a, T> {
        self.setup(); // <-- this reconfigures PC2 as analog, but PC2 is being used as OctoSPI_P1_IO2.

        BorrowedAdcChannel {
            channel: self.channel(),
            is_differential: self.is_differential(),
            _marker: PhantomData,
        }
    }
```

xoviat:
closes #6409.

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -1907,7 +1907,7 @@ fn main() {
                     adc_pairs.entry(ch).or_insert((None, None)).0.replace(pin_name.clone());
 
                     g.extend(quote! {
-                    impl_adc_pin!( #peri, #pin_name, #ch);
+                        impl_adc_pin!( #peri, #pin_name, #ch);
                     })
                 }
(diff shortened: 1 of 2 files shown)


hydra:
confirmed that this fixes the problem I was having.

https://github.com/MakerPnP/machine/commit/2d34d6351f88c2323ee6bb4c380a55595988c504

PR #6416, opened 2026-06-25
synopsys-otg: fix losing events, fix parking forever

### LINKED ISSUE #6312: embassy-usb-synopsys-otg host: HCINT coalescing in ISR loses CHH after NAK — periodic IN transfer future parks forever
**Crate/version:** `embassy-usb-synopsys-otg` 0.4.0 (current `main` is identical for `src/host.rs`)
**Hardware:** ESP32-S3 full-speed OTG host (via esp-hal's `embassy_usb_host` driver), device = HID boot-keyboard foot pedal on an interrupt IN endpoint

## Summary

The host ISR collapses all HCINT bits seen in one interrupt entry into a **single** result code and W1C-clears everything it read:

```rust
// src/host.rs (on_host_interrupt)
let hcint = r.hcint(ch).read();
let result = if hcint.xfrc() { CH_RESULT_COMPLETE }
    else if hcint.stall() { CH_RESULT_STALL }
    // ...
    else if hcint.nak() { CH_RESULT_NAK }
    else if hcint.chh() { CH_RESULT_HALTED }
    else { CH_RESULT_NONE };
// Clear all channel interrupts
r.hcint(ch).write_value(hcint);
if result != CH_RESULT_NONE {
    state.channels[ch].result.store(result, Ordering::Release);
    state.channels[ch].waker.wake();
}
```

On a periodic IN, the core auto-halts the channel after a NAK, so `NAK|CHH` routinely land in the **same** ISR entry (the device NAKs every frame, ~1 kHz). NAK wins the priority chain and the CHH is cleared and never reported. The NAK handler then:

```rust
// do_in_transfer
if result == CH_RESULT_NAK {
    if is_periodic {
        // For periodic endpoints, the hardware may start halting the
        // channel after NAK. Explicitly halt and wait for completion
        // (CHH) before reconfiguring, ...
        self.halt_channel();
        let _halt = self.wait_for_result().await; // expect CHH
    }
    ...
```

`halt_channel()` writes `CHENA|CHDIS` without checking whether the channel is still enabled. When the channel was already auto-halted (its CHH consumed above), no new CHH is generated, and `wait_for_result().await` parks the transfer future **forever** — `read()` never returns `Ok` or `Err`. Only a physical disconnect (port event path stores `CH_RESULT_HALTED` into all channels) unparks it. Linux's dwc2 guards this exact case: `dwc2_hc_halt()` returns early when `!(hcchar & HCCHAR_CHENA)`, with a comment explaining that halting a non-enabled channel must be avoided.

A second interleaving of the same design needs no hardware timing assumption at all: the result mailbox is a single overwriting `AtomicU8`. If ISR #1 stores NAK and ISR #2 stores the auto-halt's CHH **before the task polls**, the NAK is overwritten, `wait_for_result` returns `CH_RESULT_HALTED`, and `do_in_transfer` maps it to `Err(Disconnected)` with the device still attached.

## Observed on hardware

ESP32-S3 + a PCsensor HID foot pedal (3553:b001, FS, interrupt IN, NAK-polled every frame since `bInterval` isn't applied): `HidHost::read()` futures parked permanently within minutes under input-report bursts (counters frozen, no error, everything else on the executor healthy), plus occasional phantom `Disconnected` errors with the device attached. Both disappear behind downstream mitigation (read timeout + control-pipe probe + bus reset + re-enumerate), but the driver-level loss remains.

## Suggested fix

1. Accumulate channel events instead of overwriting: store a bitmask with `fetch_or` (or per-event flags) and let `wait_for_result` drain it, so NAK and CHH from one or two ISR entries both survive.
2. In the periodic-NAK path (or `halt_channel` itself), check `HCCHAR.CHENA` first and skip the halt-ack wait when the channel is already disabled — mirroring Linux dwc2.

Happy to test patches on ESP32-S3 hardware.

bugadani:
Fixes #6312

### DIFF
--- a/embassy-usb-synopsys-otg/CHANGELOG.md
+++ b/embassy-usb-synopsys-otg/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- Fixed: Periodic IN transfer future can no longer park forever.
+- Fixed: Later events no longer override earlier ones.
+
 ## 0.4.0 - 2026-05-28
 
 - **Breaking:** type-erased `State`/`HostState`, non-generic `OtgInstance`/`OtgHostInstance`, endpoint allocation in `State`, const generics removed from device/host drivers. Static state now needs to be constructed as `StateStorage::new()`/`HostStateStorage::new()` and then `as_state()`/`as_host_state()` must be called to obtain the state reference.
--- a/embassy-usb-synopsys-otg/src/host.rs
+++ b/embassy-usb-synopsys-otg/src/host.rs
@@ -10,25 +10,34 @@ use embassy_usb_driver::host::{
     DeviceEvent, HostError, PipeError, SplitInfo, UsbHostAllocator, UsbHostController, UsbPipe, pipe,
 };
 use embassy_usb_driver::{EndpointInfo, EndpointType, Speed};
-use portable_atomic::{AtomicBool, AtomicU8, Ordering};
+use portable_atomic::{AtomicBool, AtomicU8, AtomicU16, Ordering};
 
 use crate::PhyType;
 use crate::otg_v1::{Otg, vals};
 
-// Channel transfer result codes stored atomically.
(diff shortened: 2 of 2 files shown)


Dirbaio (approved):
💯

PR #6429, opened 2026-06-29
USB Host (STM32): fixes for multi device support via hub

ijager:
Two small fixes for using USB Host with an USB Hub

- USB Hubs with 7 or fewer ports report `wMaxPacketLength 1` for their status-change interrupt endpoint. This caused a panic in `calc_receive_len_bits`.
- Host channel allocation index does not necessarily coincide with the device endpoint address when having multiple devices attached via HUB.

Tested on STM32G0B with Nedis USB-A hub (UHUBU2420BK)


leftger (approved):

leftger:
This looks good to me, but xoviat is the maintainer for STM32 code specifically. From the usb host aspect this PR is good. Xoviat should be back in a week so I can ping them to look at this

PR #6430, opened 2026-06-29
USB Host (STM32): Restore shared control channel state before each control transfer.

ijager:
All control pipes share host channel register slot 0, but its device address and BTABLE receive-buffer descriptor are written only at allocation, so allocating a second device's control pipe re-points slot 0 at that device.

For example first we enumerate a Hub, then the devices behind it. After that the control pipe is configured for the device. So if we want to read the Hub status we first need to reconfigure the correct address and buffer.

Therefore we re-assert the device address and receive descriptor from each pipe's own state at the start of every control transfer.

Tested on STM32G0B with Nedis USB-A hub (UHUBU2420BK)


leftger (approved):

leftger:
Xoviat maintains `embassy-stm32`. They will be back next week. I'll ping them at that point concerning this PR. With respect to USB host functionality this checks out

PR #6435, opened 2026-07-02
embassy-mcxa: FlexCAN

bjackson312006:
This PR adds classic-mode FlexCAN support to `embassy-mcxa`. CAN FD is not implemented yet, but should be straightforward to implement in the future.

Most additions are in the new `flexcan` module. This module contains the `flexcan/classic` submodule (for classic-mode specific stuff), as well as some mode-agnostic files directly in the `flexcan/` root. There is also a stub `flexcan/fd` module for CAN FD, but it is not implemented yet and is not exposed publicly. Since Classic CAN and CAN FD are pretty different from one another, this should make it easy to partition them cleanly while still letting them access some shared functionality via top-level `flexcan/` files like `control.rs`, `filter.rs`, etc.
___
### Async Handling
I didn't implement separate "Blocking"/"Async" modes here. Since FlexCAN is driven by interrupts/buffers at the hardware level, a true `Blocking` mode wouldn't really have been possible. So, everything is primarily async, with some non-async alternatives (e.g., `try_send()`, `try_receive()`) that are still asynchronous at the hardware level.

Here's what the chain of events for sending/receiving:
- Sending:
  - When you call `FlexCan::send(frame)` or `FlexCan::try_send(frame)`, it jumps to the first available TX buffer, drops your frame in, activates the buffer, and then returns success. So, when the function returns Ok, it is less of a "frame confirmed transmitted" signal, and more of a "frame successfully handed off to hardware" signal. Basically, once you hand off your frame to the hardware, the hardware takes it from there and the CPU gets back to doing other stuff.
  - I think this preferable to async-awaiting until the transmission success interrupt fires, since it lets tasks get on with their day (and hand off multiple frames to TX at once) without having to wait for transmission success. I think this is how `embassy-stm32` handles CAN TX as well, unless I'm misunderstanding their code. Nonetheless, let me know if this is a bad idea.
  - Note: There are 32 TX message buffers, so you can only have 32 pending TX frames in the hardware at once. If the TX message buffers are all full, `FlexCan::send(frame)` will .await until a buffer opens up, and `FlexCan::try_send(frame)` will return an `SendError::TxMailboxFull` immediately. You have to be sending messages at very high frequency for this to be a problem though. In my testing, I only ran into `TxMailboxFull` errors after a `FlexCan::send(frame)` of ~70us (albeit on a very uncrowded bus).
- Receiving:
  - I've set up the Enhanced RX FIFO hardware (which has a capacity of 12 frames) to handle all RX messages. When a frame arrives in the hardware FIFO, it triggers an interrupt, which then loads the frame into a software queue (an embassy `Channel`) and flags the hardware to pop the FIFO. When you call `FlexCan::receive()` or `FlexCan::try_receive()`, it dequeues a frame from the `Channel` and hands it back to you. If the `Channel` is empty, `FlexCan::receive()` will .await until a new message arrives, and `FlexCan::try_receive()` will return immediately with a `ReceiveError::NoMessages`.
  - The size of the internal `Channel` can be configured via the `EMBASSY_MCXA_FLEXCAN_CLASSIC_RX_QUEUE_SIZE` env var, and I put some docs informing users that they can do this on all the public rx-related functions. Right now, `EMBASSY_MCXA_FLEXCAN_CLASSIC_RX_QUEUE_SIZE` defaults to 4 frames, but you could probably get away with even less if you need to. I configured it to 1 frame when testing and didn't run into `ReceiveError::NoMessages` until my RX frequency got very low (~70us again). 
  - The reason I chose using an internal `Channel` rather than requiring users to pass one in themselves was mainly just for out-of-the-box convenience. I figured that most people would not care about the default memory overhead for a 4-frame queue, and if they did, they could just configure `EMBASSY_MCXA_FLEXCAN_CLASSIC_RX_QUEUE_SIZE` to their liking. However, let me know if including an internal `Channel` like this is bad practice for HALs.
___
### Summary of Changes/Modifications in this PR

Existing files that I modified:
- `clocks/periph_helpers.rs` - Added the FlexCAN peripheral clock and instance info. Basically just copied what was already here for LPUART but adapted to FlexCAN.
- `executor.rs` - Small changes due to inconsistency with the PAC that was causing build errors.
- `lib.rs` - Make `flexcan` public
- `build.rs` - Added `flexcan` support to the existing codegen/build system stuff. Basically just copied what was already here for other peripherals but adapted to FlexCAN.
- `Cargo.toml` - Updated the revs for `nxp-pac` to the version after my PRs were merged.

New files I added:
- `flexcan/mod.rs` - Contains the traits and macros (e.g., `impl_can_instance!()`, `impl_flexcan_pin!()`) needed for the `flexcan` module to integrate with embassy-mcxa's `build.rs`.
- `flexcan/id.rs` - Basically just a bunch of newtypes for `embedded-can`'s ID types. I couldn't just re-export the types directly from `embedded-can` since those types don't have defmt impls.
- `flexcan/filter.rs` - Defines everything for configuring RX filters. The actual hardware-level filter register writing happens in `mailbox.rs` (via `rx::setup()`, but `filter.rs` sets up the public-facing interface for filter configuration. A large part of this file is the `filters!()` macro, which allows you to configure your filters via a declarative syntax that's checked for errors at compile-time. I had to make this a macro because I couldn't guarantee that a normal `const fn` would actually be evaluated at compile-time, which wasn't ideal since I'm using `panic!()` to throw my compiler errors. If your filter config isn't known until runtime (I'm not sure how common this is) and you can't use the `filters!()` macro, you can use the `FilterConfig::try_new()` function.
- `flexcan/control.rs` - Internal module for controlling the FlexCAN peripheral. It provides the raw PAC access, as well as some helper functions to abstract certain things away from pure register writes.
- `flexcan/classic/timing.rs` - Internal module for handling the bit timing configuration/baudrate stuff. I basically just translated this logic into Rust from NXP's Zephyr `fsl_flexcan.c` HAL since they probably know what they're doing. My doc at the top of `timing.rs` links to the `fsl_flexcan.c` driver I was translating from and lists the original license (BSD-3-Clause).
- `flexcan/classic/mod.rs` - This is the main public API for classic FlexCAN. The main entry-point to the driver (the `classic::FlexCan` struct) is hosted here, alongside the public configuration types and the internal `Info` struct. This file also contains the interrupt handler for classic mode.
- `flexcan/classic/meta.rs` - This file contains any metaprogramming and macro stuff relevant to `flexcan::classic`. I figured having a dedicated file for this stuff would be cleaner than scattering it around the rest of the module. This file contains the compile-time evaluation for the `EMBASSY_MCXA_FLEXCAN_CLASSIC_RX_QUEUE_SIZE` env var. It also contains a bunch of macros for rustdocs that are used in multiple places  (e.g., so I wouldn't have to copy-paste identical docs for `FlexCan::send()` and `FlexCanTx::send()`).
- `flexcan/classic/mailbox.rs` - This is the main "low-level" file for the `classic` module, and handles the actual memory areas for TX frames and RX frames. I've decided to partition the FlexCAN registers so that the entire message buffer area (80h - 27Fh) is dedicated to TX frames, and all RX frames are routed through the Enhanced RX FIFO. I think this cleanly separates the responsibilities/functions of these memory areas, and leaves us with plenty of hardware frame buffers (32 TX buffers and 12 RX buffers) to work with.
- `flexcan/classic/frame.rs` - Defines the `Frame` struct and all the corresponding `embedded_can::Frame` impls.
- `flexcan/classic/example.md` - A rustdoc containing some example code for initializing and using a `FlexCan` instance, for use in the `classic/mod.rs` module docs.
- `flexcan/fd/mod.rs` - This is just private stub module for CAN FD support. It contains some notes/considerations for whoever starts work on adding CAN FD (whether it is me or someone else)
___
### Testing
If interested, [here](https://github.com/bjackson312006/flexcan-test) is the repo I used to test/verify my FlexCAN stuff


examples/mcxa2xx/src/bin/flexcan-classic-async.rs
felipebalbi, line 1: This should be converted into an actual example under examples/mcxa2xx and examples/mcxa5xx

embassy-mcxa/src/flexcan/classic/meta.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+    pub(in crate::flexcan::classic) use env_var_name;
+
+    // Default size of the queue, in messages/Frames.
+    // This is a macro instead of a `const`` so we can use it in docs.
+    macro_rules! rx_queue_size_default {
+        () => {
+            4
+        };
+    }
```
felipebalbi: really sounds like this should be a const generic. Or, probably preferably, have the user give you ownership of the buffer. Did you have a look at how other embassy HALs are implementing CAN support?

embassy-mcxa/src/flexcan/classic/meta.rs
```diff
@@ -0,0 +1,163 @@
...
+                "Note: During a BusOff event, this function will asynchronously wait until\n",
+                "the bus recovers. This is due to the behavior mentioned above: The TX mailbox\n",
+                "doesn't drain during BusOff (and will eventually fill up), causing this\n",
+                "function to wait until after recovery when buffers start becoming available again.\n",
+                "\n",
+                "Unless explicitly disabled, FlexCAN will recover from BusOff automatically. However,\n",
+                "if you need to be notified immediately when a BusOff event occurs, see the `try_send()`\n",
+                "and `error_mode()` functions.",
+            )
+        };
```
felipebalbi: I'd prefer to avoid macros for rustdocs. Feels like an unnecessary obfuscation.

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+    pub fn tx_mailbox_full_count(&self) -> u32 {
+        functions::tx_mailbox_full_count(self.info)
+    }
+}
+
+/// FlexCAN driver instance, in Classic CAN mode.
+pub struct FlexCan<'d> {
+    tx: FlexCanTx<'d>,
+    rx: FlexCanRx<'d>,
+}
```
felipebalbi: You still want to have the Mode generic here:

- registers can be polled for blocking implementation
- FlexCAN supports DMA

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+        // Setup the interrupts
+        T::Interrupt::unpend();
+        unsafe {
+            T::Interrupt::enable();
+        }
+
+        info.control.unfreeze();
+
+        let tx = FlexCanTx::new(info, _tx, _wake_guard);
+        let rx = FlexCanRx::new(info, _rx);
```
felipebalbi: wake guard should be held by both TX and RX halves. You can clone it. 

If you don't, we could fall into a situation where deep sleep is not prevented when someone holds *only* a receive half of the FlexCAN peripheral.

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+    #[doc = doc_send!()]
+    pub async fn send(info: &Info, frame: &Frame) {
+        use core::future::poll_fn;
+        use core::sync::atomic::Ordering;
+        use core::task::Poll;
+
+        use nb::Error::{Other, WouldBlock};
+
+        let message = tx::TxMessage::from(frame);
+        poll_fn(|cx| {
```
felipebalbi: it's in the DEVGUIDE that the entire embassy-mcxa uses maitake_sync. This driver should use WaitCell, not AtomicWaker.

embassy-mcxa/src/flexcan/fd/mod.rs
felipebalbi, line 1: looks like you should remove this file.

embassy-mcxa/src/flexcan/control.rs
```diff
@@ -0,0 +1,150 @@
...
+        unsafe { crate::pac::common::Reg::from_ptr((pac::Can::as_ptr(&can) as *mut u8).wrapping_add(0x2000usize) as _) }
+    }
+
+    /// PAC extension function: Accesses the ID area inside the Enhanced RX FIFO.
+    ///
+    /// `ID` is a 32-bit region starting at 2004h.
+    pub(in crate::flexcan) const fn pac_fifoid(&self) -> crate::pac::common::Reg<pac::Id, crate::pac::common::R> {
+        let can = self.regs();
+        unsafe { crate::pac::common::Reg::from_ptr((pac::Can::as_ptr(&can) as *mut u8).wrapping_add(0x2004usize) as _) }
+    }
```
felipebalbi: Is this missing from the PAC? Do we need to patch the PAC?

embassy-mcxa/src/executor.rs
felipebalbi, line 1: Did you find a build error here? Looks like we need to update mcxa CI.

felipebalbi (changes requested):

examples/mcxa2xx/src/bin/flexcan-classic-async.rs
bjackson312006: Fixed, added the example files (`flexcan-classic-async.rs` and `flexcan-classic-blocking.rs`) to both those directories and deleted the markdown file

embassy-mcxa/src/flexcan/classic/meta.rs
bjackson312006: Fixed by removing the env var and just having users pass in a `RxQueue` type (which just wraps around a Channel).

Originally, I didn't use const generics due to the guidelines set in the DEVGUIDE. Technically, the new `RxQueue` type uses a `N` const generic, but this is purely just to configure the size of the inner `Channel` and doesn't propagate any further than that. The driver stores the channel via a `SendDynamicSender` and a `SendDynamicReceiver` which is what embassy-stm32's CAN HAL does

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed

embassy-mcxa/src/flexcan/fd/mod.rs
bjackson312006: Removed

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed, I added separate `Async` and `Blocking` modes. The Async mode is generally unchanged from what the single-mode driver looked like.

The blocking mode gets rid of the ISR and  replaces `send()`/`receive()` with `blocking_send()`/`blocking_receive()`, which just do blocking polls on the registers until they're able to complete their action. Blocking mode still provides `try_send()` and `try_receive()` for those who don't want to block, as well as `blocking_send_timeout()` and `blocking_receive_timeout()` for those who need to bound their blocking.

embassy-mcxa/src/executor.rs
bjackson312006: There was a build error (the PAC no longer used screaming case for those variants after being regenerated)

embassy-mcxa/src/flexcan/control.rs
bjackson312006: This is missing from the PAC. I don't know what the intended scope of the PAC is, but if memory areas like this are meant to be included in it, then yes it would make sense to patch this into the PAC. I can make a PR for that  on nxp-pac if needed.

Note: I don't think the [raw SVD](https://raw.githubusercontent.com/nxp-mcuxpresso/mcux-soc-svd/da354fa1807150dcbe1e9bddbd92ab9b11fc6454/MCXA256/MCXA256.xml) includes this memory area

embassy-mcxa/src/flexcan/classic/meta.rs
bjackson312006: Are you sure this would be better? These macros exist because the driver can be used via a single struct (`FlexCan`) or split into two (`FlexCanTx`, `FlexCanRx`), meaning there are several functions that are identical across multiple structs. Because of this, the rustdocs for those functions need to be identical to each other. Rather than just copy-pasting the docs, it made sense to write them once in a macro so they could be re-used as needed. Otherwise, whenever making changes to the docs, you'd have to track down every place they're used and make changes for each one individually.

Note: `meta.rs` itself no longer exists due to the other changes.

embassy-mcxa/src/flexcan/control.rs
bjackson312006: Update: I've made a [PR](https://github.com/embassy-rs/nxp-pac/pull/110) on nxp-pac to add in the missing accessors

felipebalbi: okay, merged the PAC PR, thank you.

embassy-mcxa/src/flexcan/classic/meta.rs
felipebalbi: Yeah, don't obfuscate. Look at the existing drivers, they duplicate the documentation when necessary. You really don't want the obfuscation of this macro.

bjackson312006: Alright updated

embassy-mcxa/src/flexcan/classic/mailbox.rs
```diff
@@ -0,0 +1,569 @@
...
+            w.set_erfwmi(pac::Erfwmi::WatermarkYes); // Clear the Enhanced FIFO Watermark Indication Flag (write-1-to-clear)
+            w.set_erfda(true); // Clear the Enhanced FIFO Data Available Flag (write-1-to-clear)
+        });
+
+        // Make it so every RX triggers an interrupt rather than batching them
+        info.control.regs().erfier().modify(|m| {
+            m.set_erfdaie(true); // Enable the Enhanced FIFO Data Available Interrupt
+            m.set_erfwmiie(false); // Disable the Enhanced RX FIFO Watermark Indication Interrupt (since we're not batching)
+            m.set_erfovfie(true); // Enable the Enhanced RX FIFO Overflow Interrupt Enable
+        });
```
felipebalbi, line 466: **RX overflow interrupt is enabled but never acknowledged (interrupt storm).**

`erfovfie` is enabled here, but `on_interrupt` (`asynchronous.rs`, ~L233-290) only drains via the `ERFDA` path and clears the BusOff flags — it never reads or write-1-clears `ERFOVF`. Because `ERFOVF` is a sticky W1C level source, the first RX FIFO overflow (exactly what `rx_dropped_count` is meant to measure) latches the NVIC line high permanently and the handler re-enters forever, starving the executor and any task-fed watchdog. This is the DEVGUIDE's "leaving the enable bit set means the ISR re-fires forever" case.

Please W1C `erfovf` in the handler and count it, e.g.:
```rust
let erfsr = can.erfsr().read();
if erfsr.erfovf() {
    can.erfsr().write(|w| w.set_erfovf(true));
    info.rx_dropped_count.fetch_add(1, Ordering::Relaxed);
}
```

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+    pub fn tx_mailbox_full_count(&self) -> u32 {
+        self.info.tx_mailbox_full_count.load(Ordering::Acquire)
+    }
+}
+
+/// FlexCAN driver instance, in Classic CAN mode.
+pub struct FlexCan<'d, M: Mode> {
+    tx: FlexCanTx<'d, M>,
+    rx: FlexCanRx<'d, M>,
+}
```
felipebalbi, line 623: **missing `Drop`/teardown for `FlexCan` / `FlexCanTx` / `FlexCanRx` or the pins.**

Construction ends with `unfreeze()` and (async) `T::Interrupt::enable()`, so a dropped driver leaves the peripheral unfrozen and ACK-ing on the bus, the NVIC IRQ enabled, the pins still muxed, and the MRCC gate on (only the `WakeGuard` vote is released). Combined with the RX-overflow issue, a dropped async driver keeps receiving into an unread channel and can storm with no owner. Compare `lpuart` (`TxPins`/`RxPins` `Drop` -> `set_as_disabled()`) and `i2c`.

Since the split halves have no guaranteed drop order (see the DEVGUIDE "Teardown" section), reset only each half's own pin in its `Drop`, and tie the shared teardown (re-freeze, `Interrupt::disable()`, clear `erfdaie`/`erfovfie`/`boffmsk`/`boffdonemsk`, gate the clock via the `clocks` subsystem) to the last/owning handle.

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    /// Recombines a `FlexCanTx` and `FlexCanRx` (previously obtained from `split()`) back into a `FlexCan`.
+    ///
+    /// If the two halves come from different peripherals, this returns `Err((tx, rx))`,
+    /// where `tx` and `rx` are the same `tx`/`rx` you passed in. This means you can try again if you need to.
+    pub fn join(tx: FlexCanTx<'d, M>, rx: FlexCanRx<'d, M>) -> Result<Self, (FlexCanRx<'d, M>, FlexCanTx<'d, M>)> {
+        if !core::ptr::eq(tx.info, rx.info) {
+            return Err((rx, tx));
+        }
+        Ok(Self { tx, rx })
```
felipebalbi, line 641: **`join()` error tuple is reversed, and the doc contradicts the code.**

The doc says `Err((tx, rx))`, but the return type is `(FlexCanRx, FlexCanTx)` and the body returns `Err((rx, tx))`. `split()` (L347) returns `(FlexCanTx, FlexCanRx)`, so please make the error `(FlexCanTx, FlexCanRx)` / `Err((tx, rx))` to match both the argument order and `split()`. Worth fixing the type (not just the doc), since it's a breaking detail post-release.

embassy-mcxa/src/flexcan/classic/asynchronous.rs
```diff
@@ -0,0 +1,291 @@
...
+/// FlexCAN interrupt handler.
+/// Construct this in a `bind_interrupts!` block to route an IRQ (e.g., CAN0, CAN1) here.
+///
+/// Note: This is only required (and only relevant) for the async driver.
+pub struct InterruptHandler<T: Instance> {
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Instance> Handler<T::Interrupt> for InterruptHandler<T> {
+    unsafe fn on_interrupt() {
```
felipebalbi, line 316: **ISR skips the crate's perf-counter instrumentation.**

Every other async driver's `SealedInstance` carries `PERF_INT_INCR`/`PERF_INT_WAKE_INCR` and the handler calls them (see `i2c/controller.rs:131,146`); `flexcan` has neither, so the CAN IRQ is invisible to the crate's interrupt instrumentation. Please add the two consts to `classic::SealedInstance`, wire them in `impl_can_instance!`, and call them at entry and before each `wake()`.

embassy-mcxa/src/flexcan/classic/frame.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Returns this frame's payload bytes (empty for remote frames).
+    pub fn data(&self) -> &[u8] {
+        match self.kind {
+            FrameKind::RemoteFrame => &[],
+            FrameKind::DataFrame => &self.data[..self.length],
+        }
+    }
+}
+
+impl embedded_can::Frame for Frame {
```
felipebalbi, line 93: **consider implementing `embedded_can::nb::Can`.**

`Frame` implements `embedded_can::Frame`, but no driver type implements `embedded_can::nb::Can`, so a generic CAN consumer can't drive this peripheral (cf. `lpuart` implementing `embedded_hal_nb`/`embedded_io`). The blocking `try_send`/`try_receive` already have nb-shaped semantics, so this is close to mechanical for `FlexCan<Blocking>`. (No standardized async CAN trait exists at these versions — nothing to add there.)

embassy-mcxa/src/flexcan/classic/timing.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub(crate) phase_seg_2: u8,
+
+    /// Propagation Segment.
+    pub(crate) prop_seg: u8,
+}
+
+/// Errors that may occur when configuring timing/bitrate.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum TimingError {
```
felipebalbi, line 89: **`TimingError` is not `#[non_exhaustive]`.**

It's the one public error missing the attribute, and it's re-exported through `InitError::TimingError`, so adding a variant later would be a breaking change.

embassy-mcxa/src/flexcan/classic/frame.rs
```diff
@@ -0,0 +1,112 @@
...
+pub(in crate::flexcan) enum FrameKind {
+    /// A "normal" CAN frame. Corresponds to RTR bit = 0.
+    DataFrame,
+
+    /// A Remote CAN frame. Corresponds to RTR bit = 1.
+    RemoteFrame,
+}
+
+/// A CAN frame.
+pub struct Frame {
```
felipebalbi, line 21: **`Frame` has no `defmt::Format` (or `Debug`).**

`id.rs` states every publicly exposed type "should probably support `defmt`", and both examples work around this by logging `frame.id()` rather than the frame itself. Please add `#[cfg_attr(feature = "defmt", derive(defmt::Format))]` (and `Debug`), or soften the claim in `id.rs`.

embassy-mcxa/src/flexcan/classic/timing.rs
```diff
@@ -0,0 +1,230 @@
...
+//! ```
+//!
+//! Note: This module targets the Enhanced CAN Bit Timing registers (rather than the older CBT[BTF] and CTRL1 timing registers).
+//! This is because all modern MCXA chips with FlexCAN support the Enhanced CAN Bit Timing registers, and they're better than the older timing registers.
+
+// These are a bunch of constants taken from the hal_nxp C code.
+// They're not even that necessary in this project since we have a PAC (i.e., we're not manually doing a bunch of shifting and masking in registers),
+// but having the `MAX` values is kind of nice as iteration bounds, and mirroring the layout keeps this file easy to compare against the upstream driver.
+
+#![allow(dead_code)] // need this since some of the constants are considered unused by rust-analyzer even though they're used by other constants
```
felipebalbi: **module-wide `#![allow(dead_code)]` over a bit-mask/shift const wall.**

Per the DEVGUIDE, a const block behind a blanket `allow(dead_code)` is a signal to patch the PAC. These are only used to derive the `MAX_*` loop bounds, so please drop the unused `*_MASK`/`*_SHIFT` pairs and keep just the bounds (localizing or removing the `allow`).

embassy-mcxa/src/flexcan/classic/blocking.rs
```diff
@@ -0,0 +1,285 @@
...
+#[non_exhaustive]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum SendErrorWithTimeout {
+    /// A traditional `SendError`.
+    SendError(SendError),
+
+    /// The function call exceeded the user-provided timeout.
+    Timeout,
+}
```
felipebalbi: **`SendErrorWithTimeout`/`ReceiveErrorWithTimeout` duplicate one shape.**

Both are `enum { Inner(E), Timeout }`. A single `pub enum WithTimeout<E> { Error(E), Timeout }` (`#[non_exhaustive]`, conditional `defmt`) used as `WithTimeout<SendError>` / `WithTimeout<ReceiveError>` keeps `Timeout` off the non-timeout methods while removing the duplication.

embassy-mcxa/src/flexcan/classic/asynchronous.rs
```diff
@@ -0,0 +1,291 @@
...
+            let _ = can.esr1().read(); // make sure the clear lands before returning
+        }
+
+        // Handle when BusOff autorecovery has finished
+        if esr1.boffdoneint() == crate::pac::can::Boffdoneint::BusOffDone {
+            // Acknowledge the flag (write 1 to clear)
+            can.esr1()
+                .write(|w| w.set_boffdoneint(crate::pac::can::Boffdoneint::BusOffDone));
+            let _ = can.esr1().read(); // Make surethe clear lands before returning
+        }
```
felipebalbi, line 392: **recovery never wakes `tx_waker`.**

A sender parked in `send()` (all 32 buffers full) currently relies on "a pending buffer retransmits and completes after auto-recovery" to be re-woken, since neither `boffint` nor `boffdoneint` calls `tx_waker.wake()`. Per the DEVGUIDE ("wake all waiters on global events"), add `info.tx_waker.wake()` in both branches so parked senders re-check on the global event.

embassy-mcxa/src/flexcan/classic/asynchronous.rs
```diff
@@ -0,0 +1,291 @@
...
+            };
+
+            // Push the frame into the software RX queue.
+            let dropped = match rx_sender {
+                Some::<SendDynamicSender<'static, Frame>>(sender) => sender.try_send(frame).is_err(),
+                None => true,
+            };
+            if dropped {
+                // if the software queue is full, drop the frame, and increment the `rx_dropped_count` counter.
+                info.rx_dropped_count.fetch_add(1, Ordering::Acquire);
```
felipebalbi: **atomic ordering.**

I believe the `fetch_add(1, Ordering::Acquire)` on this counter (and the `tx_mailbox_full_count` paths) should be `Relaxed` — `Acquire` on an RMW orders the load side, and these counters gate no other memory. Separately, the `Relaxed` on `tx_remote` across the ISR/thread boundary is sound on single-core but worth a one-line justification.

embassy-mcxa/src/flexcan/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+pub mod filter;
+pub mod id;
+
+use embassy_hal_internal::PeripheralType;
+
+use crate::gpio::AnyPin;
+use crate::interrupt::typelevel::Interrupt;
+
+/// Peripheral identity.
+pub trait Instance: PeripheralType + 'static + Send {
```
felipebalbi: **base `flexcan::Instance` is unsealed, unlike every other `Instance` in the crate** (note the absent `#[allow(private_bounds)]`).

`classic::Instance` *is* sealed, so it's safe in practice, but please seal the base trait too for consistency. The two-layer `flexcan::Instance` + `classic::Instance` split is only justified by a future `fd` module, and there's no `fd/` stub in this PR — so either add the stub or a comment noting the intent.

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+/// Configuration settings for a Classic-mode FlexCAN driver instance.
+pub struct FlexCanConfig<'a> {
+    /// This setting allows you to enable/disable the
+    /// FlexCAN's Protocol Exception feature. This may be
+    /// useful if you want this FlexCAN to coexist on a FD-enabled
+    /// bus while staying in Classic mode.
+    ///
+    /// Note: This feature is only available on CAN0. If you attempt to enable this
+    /// feature on an instance where it isn't supported, you will get an error
+    /// when calling `FlexCan::new()`.
```
felipebalbi: nit: `FlexCan::new()` doesn't exist — the constructors are `new_async()` / `new_blocking()`. Same dangling reference on the `clock_source` field doc at L137.

embassy-mcxa/src/flexcan/classic/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    /// This error indicates that you attempted enabling the
+    /// `protocol_exception` feature in `FlexCanConfig` on a
+    /// peripheral where it is not supported.
+    ///
+    /// Note: Protocol Exception is only supported on CAN0.
+    ProtocolExceptionUnsupported,
+
+    /// You have attempted to configure an invalid bitrate.
+    /// See the `TimingError` struct docs for more info.
```
felipebalbi: nit: `TimingError` is an enum, not a struct — and it lives in the private `timing` module, so although it's returned via this public variant it can't be named by users. Consider `pub use timing::TimingError;`.

embassy-mcxa/src/flexcan/classic/mailbox.rs
```diff
@@ -0,0 +1,569 @@
+//! This module models the FlexCAN mailbox for Classic CAN (not FD).
+//!
+//! TX/outgoing messages are handled in the FlexCAN message buffer, which uses the memory area 80h - 27Fh (see page 1545 of the datasheet). This memory area is 512 bytes in total.
```
felipebalbi: nit: datasheet citations here (and in `control.rs` / `periph_helpers.rs`) give page numbers but name neither the manual nor its revision — and this driver builds for two families with different manuals. Prefer `Doc title, Rev. X, section N` (section numbers survive re-pagination).

embassy-mcxa/src/flexcan/classic/timing.rs
```diff
@@ -0,0 +1,230 @@
...
+//! A lot of the code in this module was translated into Rust from the `fsl_flexcan.c` source file inside `hal_nxp` Zephyr module.
+//! As of June 2026, that source file is hosted here: <https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/drivers/flexcan/fsl_flexcan.c>
+//!
+//! Here is the original copyright notice from that source file:
+//!
+//! ```text
+//! Copyright 2020-2021, 2026 NXP
+//!
+//! SPDX-License-Identifier: BSD-3-Clause
+//! ```
```
felipebalbi, line 26: nit: BSD-3-Clause requires retaining the full condition list + "AS IS" disclaimer, not just the copyright line + SPDX tag; an SPDX identifier alone generally isn't sufficient in source form. Consider reproducing the full notice, pinning the source link to a specific revision (not `master`), and adding a crate-level third-party notice. Worth a maintainer's eye.

examples/mcxa2xx/src/bin/flexcan-classic-async.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[embassy_executor::main]
+async fn main(spawner: Spawner) {
+    let p = embassy_mcxa::init(Config::default());
+
+    // Create and configure a `FlexCan` instance for CAN0.
+    let can0 = FlexCan::new_async(
+        p.CAN0,
+        p.P1_11,
+        p.P1_2,
```
felipebalbi, line 40: nit: these pins are passed positionally with no hint that `P1_11` is RXD and `P1_2` is TXD, and no note that a CAN transceiver / specific board wiring is assumed. Trailing `// CAN0 RXD` / `// CAN0 TXD` comments would help. Also: the mcxa2xx and mcxa5xx example binaries are byte-identical — worth a note that they're intentionally kept in sync (or differentiating the pins per board).

embassy-mcxa/src/flexcan/classic/asynchronous.rs
```diff
@@ -0,0 +1,291 @@
...
+            can.esr1().write(|w| w.set_boffint(true));
+            let _ = can.esr1().read(); // make sure the clear lands before returning
+        }
+
+        // Handle when BusOff autorecovery has finished
+        if esr1.boffdoneint() == crate::pac::can::Boffdoneint::BusOffDone {
+            // Acknowledge the flag (write 1 to clear)
+            can.esr1()
+                .write(|w| w.set_boffdoneint(crate::pac::can::Boffdoneint::BusOffDone));
+            let _ = can.esr1().read(); // Make surethe clear lands before returning
```
felipebalbi: nit: small typo sweep — `Make surethe` (here, L288), `occured` (L134, L218), `indefinietely` (`blocking.rs` L137 et al.), `relavent`/`recieving`/`handeled`/`clearning` (`mailbox.rs`), `deafults`/`ususally` (`control.rs`), `coutner` (`mod.rs` `Info` doc).

embassy-mcxa/src/flexcan/control.rs
```diff
@@ -0,0 +1,117 @@
...
+use nxp_pac::can as pac;
+
+/// Contains a bunch of random helpers for controlling/configuring the FlexCAN peripheral.
+#[allow(dead_code)]
+pub(crate) struct Control {
+    regs: pac::Can,
+}
+
+// u_Note: default MCR for CAN0: 11011000100100000000000000001111
+// u_Note: default MCR for CAN1: 11011000100100000000010000001111
```
felipebalbi: nit: leftover personal markers `u_Note` (here, L12-13) and `u_TODO` (`clocks/periph_helpers.rs`, in the `CanClockSel` note). Convert to standard `// NOTE:` / `// TODO:` (or remove) so they surface in normal tooling and don't read as unfinished.

embassy-mcxa/src/flexcan/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    type Interrupt: Interrupt;
```
felipebalbi, line 29: nit: `type Interrupt` is undocumented; the reference driver documents it (`i2c/mod.rs:36` -> "Interrupt for this I2C instance."). A one-line doc keeps it consistent.

embassy-mcxa/src/flexcan/classic/blocking.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.tx.try_send(frame)
+    }
+
+    /// Receives a CAN message.
+    ///
+    /// If the hardware RX FIFO is empty, this will block until a frame arrives. If you need to
+    /// avoid blocking, see `try_receive()` (or, just use the `Async` mode).
+    pub fn blocking_receive(&self) -> Frame {
+        self.rx.blocking_receive()
+    }
```
felipebalbi, line 93: nit: this doc says "Like `receive()`", but in blocking mode the method is `blocking_receive()` (`receive()` is async-only). Copy-paste drift from the async module; same at L259.

embassy-mcxa/src/flexcan/classic/blocking.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    /// Like `receive()`, but returns immediately with `ReceiveError::NoMessages` if the RX FIFO is empty.
+    pub fn try_receive(&self) -> Result<Frame, ReceiveError> {
+        self.rx.try_receive()
+    }
```
felipebalbi, line 98: nit: this doc says "Like `receive()`", but the method is `try_receive()` (`receive()` is async-only). Copy-paste drift from the async module; same at L259.

felipebalbi (changes requested):
A few more comments after a more careful review

embassy-mcxa/src/flexcan/classic/mailbox.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix rx overflow interrupt storm, and split rx_dropped_count into rx_dropped_count_channel and rx_dropped_count_fifo`)

Note: `rx_dropped_count` isn't meant to measure the ERFOVF overflows (it is specific to the software Channel/`RxQueue`). For this reason I decided to add separate `rx_dropped_count_channel` and `rx_dropped_count_fifo` trackers in this commit to replace `rx_dropped_count`

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: added teardown for flexcan`).

This isn't 100% like LPUART's teardown due to the TX and RX pins are quite closely coupled in CAN. For example, even if the RX half of FlexCAN is dropped, the remaining TX half still needs the RX pin for error detection. In the other direction, even if the TX half of FlexCAN is dropped, the remaining RX half still needs the TX pin to send ACKs. So if I were to drop each pin individually as each half dropped, the other half would stop working.

There were probably several potential ways to handle this:
- First, it would've been possible to just keep the LPUART-like behavior and leave docs warning that, once one half is dropped, the remaining half is not guaranteed to work correctly. This would've been more consistent with the rest of embassy-mcxa, but I don't think this behavior would have been appropriate for FlexCAN since it basically willingly introduces a janky unstable error state into the FlexCAN driver. Technically, since FlexCAN has a sleep mode, I could do some state tracking to automatically but FlexCAN in sleep mode when TX is dropped but RX remains. However, this would require putting the peripheral in freeze mode, and generally seems like an inappropriate action to force onto the user via an automatic `Drop` routine. More importantly, there is no analogy to sleep mode for the opposite state (where RX is dropped and TX remains), which makes this idea impractical.
- Second, it could've been possible to just get rid of the TX/RX half functionality and just expose the single combined `FlexCan`. I think this is similar to how the SPI and I2C/I3C drivers work (?) since those protocols don't really make sense to model as separate halves, but IMO CAN is in a limbo between fully independent TX/RX (like LPUART) and fully codependent/transactional (like SPI/I2C/I3C as far as I understand them). It also just wouldn't have made sense to get rid of the TX/RX splitting functionality since it goes against embassy conventions (e.g., embassy-stm32 CAN)
- Thirdly, it could've been possible to do something like `embassy-stm32` where the independent TX/RX halves are borrowed from the central combined CAN (`split()` is `&mut self` IIRC). If I'm understanding embassy-stm32's API correctly, the borrowed TX/RX halves never own the TX/RX pins directly, and dropping the TX/RX halves doesn't mess with the TX/RX pins (since they are ultimately managed by the central CAN). It's only when the central CAN gets dropped that the TX/RX pins are disabled (or more accurately for embassy-stm32 from what I can tell, put into a sleep mode). I didn't really want to use this pattern since it seems like it introduces a possible condition where the central CAN could be dropped, but the TX/RX halves could still be fully alive/in scope (since the halves have different lifetimes to the central CAN). I think this would allow you to continue using the TX/RX halves even though the peripheral is asleep/non-functional, which does not seem ideal. So assuming I am understanding how embassy-stm32 works correctly, I did not really want to follow that pattern either.

So ultimately, I settled on something that is somewhat similar to embassy-mcxa's LPUART, but with both pins getting disabled only when the last remaining TX/RX half is dropped. In other words, both halves still own their respective pins via `Peri`, but the underlying `AnyPin` state is tracked inside the static `Info` so they can be disabled during the final teardown rather than each half's individual teardown. It may be a good idea to review this commit individually since it is somewhat hefty on its own.

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fixed join() return type and docs`)

embassy-mcxa/src/flexcan/classic/asynchronous.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: added PERF_INT_INCR and PERF_INT_WAKE_INCR to SealedInstance and the ISR just like the other embassy-mcxa async drivers do`)

embassy-mcxa/src/flexcan/classic/frame.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: added embedded_can::nb::Can and embedded_can::blocking::Can impls`)

embassy-mcxa/src/flexcan/classic/timing.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: cleanup derives for public types (added non_exhaustive to TimingError and added defmt::Format plus common derives for Frame)`)

embassy-mcxa/src/flexcan/classic/frame.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: cleanup derives for public types (added non_exhaustive to TimingError and added defmt::Format plus common derives for Frame)`)

embassy-mcxa/src/flexcan/classic/timing.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: remove unused consts in timing.rs and get rid of the blanket allow_unused for the module`)

embassy-mcxa/src/flexcan/classic/blocking.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: remove SendErrorWithTimeout and ReceiveErrorWithTimeout, and add the new single WithTimeout struct`)

embassy-mcxa/src/flexcan/classic/asynchronous.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: add tx_waker.wake() for busoff recovery`).

Note: The `tx_waker.wake()` call was already added for `esr1.boffint()` in an earlier commit (I think the teardown commit), but I added to the `esr1.boffdoneint()` block here (plus `T::PERF_INT_WAKE_INCR()`)

xoviat:
pull upstream changes; we've had a compiler update.

embassy-mcxa/src/flexcan/classic/asynchronous.rs
bjackson312006: Fixed previously via the RX interrupt commit (`mcxa/flexcan: fix rx overflow interrupt storm, and split rx_dropped_count into rx_dropped_count_channel and rx_dropped_count_fifo`)

embassy-mcxa/src/flexcan/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: seal flexcan::instance, plus update to new formatting (affecting both the flexcan driver and the flexcan examples for mcxa2xx/mcxa5xx)`)

examples/mcxa2xx/src/bin/flexcan-classic-async.rs
bjackson312006: I haven't made any changes for this specific comment since it seems like these edits are unneeded based on the conventions in the other mcxa2xx/mcxa5xx examples (ex. most/all assume a specific board wiring and don't explicitly document what parameters are TX pins vs RX pins). For the RXD/TXD distinction specifically, you can just look at the rust-analyzer param annotations or what traits (RxPin/TxPin) the pins implement. Similarly, having identical mcxa2x/mcxa5xx examples doesn't seem worthy of a note since there are no hardware or pin differences between the two platforms relevant to the FlexCAN examples, and other existing examples are also identical between the two platforms (e.g., `capture.rs`, `dma-stress.rs`, `dma_scatter_gather_builder.rs`, etc)

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/mailbox.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/timing.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/asynchronous.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/control.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/mod.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/blocking.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

embassy-mcxa/src/flexcan/classic/blocking.rs
bjackson312006: Fixed (see commit `mcxa/flexcan: fix nits`)

felipebalbi (approved):

PR #6444, opened 2026-07-03
add spi and spi_dma examples for stm32u5

RaresAndrei0:
This PR adds two simple SPI examples demonstrating how to use the protocol using a BMP390 sensor. Includes standard blocking SPI and async SPI using GPDMA.

Tested on the STM32U545RE board.


examples/stm32u5/src/bin/adc.rs
```diff
@@ -2,9 +2,10 @@
 #![no_main]
 
 use defmt::*;
+use defmt_rtt as _;
 use embassy_stm32::adc::{self, Adc, AdcChannel, AdcConfig, SampleTime, adc4};
 use embassy_stm32::{bind_interrupts, dma, peripherals};
-use {defmt_rtt as _, panic_probe as _};
+use panic_probe as _;
```
eva-cosma: I think your formatter's version is different from the rust_toolchain.toml. Try to run `cargo fmt` or `cargo +nightly fmt` again, or maually revert these changes

examples/stm32u5/src/bin/spi.rs
```diff
@@ -0,0 +1,56 @@
+//This example uses the BMP390 barometric pressure sensor, for simplicity we only read the chip ID
```
eva-cosma: Use module-level doc comments (`//!`) and write more details about the hardware requirements for this example (which pins are connected and where). Alternatively, move these details to a README.md if you think images could help with the explanations.

examples/stm32u5/src/bin/spi.rs
```diff
@@ -0,0 +1,56 @@
...
+use embassy_stm32::gpio::{Level, Output, Speed};
+use embassy_stm32::spi::{Config, Spi};
+use embassy_stm32::time::Hertz;
+use panic_probe as _;
+
+#[entry]
+fn main() -> ! {
+    info!("Device has started!");
+
+    //Initialize peripherals
```
eva-cosma:
```suggestion
    // Initialize peripherals
```
Space after `//`. This applies to all of your comments.

examples/stm32u5/src/bin/spi.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use cortex_m_rt::entry;
+use defmt::*;
+use defmt_rtt as _;
+use embassy_stm32::gpio::{Level, Output, Speed};
+use embassy_stm32::spi::{Config, Spi};
+use embassy_stm32::time::Hertz;
+use panic_probe as _;
+
+#[entry]
```
eva-cosma: Why do we use the cortex_m entry? Can we not use the embassy entry?

examples/stm32u5/src/bin/spi.rs
```diff
@@ -0,0 +1,56 @@
...
+fn main() -> ! {
+    info!("Device has started!");
+
+    //Initialize peripherals
+    let p = embassy_stm32::init(Default::default());
+
+    //Set spi frequency
+    let mut spi_config = Config::default();
+    spi_config.frequency = Hertz(1_000_000);
```
eva-cosma: Comments should explain WHY not WHAT is happening. Code should be clear enough to be self-documenting as to what is happening (e.g. in this example, its clear that we are setting the SPI configuration to 1 MHz. WHY are we setting it to 1MHz though? That should be explained by the comment instead.

eva-cosma (changes requested):

RaresAndrei0:
Ran cargo fmt and it returned some errors so i tried cargo +nightly fmt. After commiting, the fmt checker failed again with the code 123. Checked the command from the github script and ran it without the "--check" flag, after that the build was succesful. Is there an explanation for this difference in behavior? As I want to know for the future.

examples/stm32u5/README.md
```diff
@@ -0,0 +1,18 @@
+# STM32 Examples
```
eva-cosma:
```suggestion
# STM32U5 Examples
```

examples/stm32u5/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+STM32 D10/PC_9 -> BMP390_CSB
+
+STM32 D11/PA_7 -> BMP390_SDI
+
+STM32 D12/PA_6 -> BMP390_SDO
+
+STM32 D13/PA_5 -> BMP390_SCK
+
+![Figure 1](../images/pinout1.png)
```
eva-cosma: Add thr other pinouts as well, and put the images at the start

examples/stm32u5/README.md
```diff
@@ -0,0 +1,18 @@
+# STM32 Examples
+
+# Spi (blocking & DMA) 
+These examples use the BMP390 barometric pressure sensor, we send a read request to register 0x00 and we wait for the chip ID response (0x60).
```
eva-cosma, line 11: Mention the nucleo board this was tested on

eva-cosma (changes requested):
Didn't we have a change in the Cargo.toml for setting the feature flag for the stm32u545? If yes, we may need to do more complex changes. If not, all is good besides some nitpicks

RaresAndrei0:
> Didn't we have a change in the Cargo.toml for setting the feature flag for the stm32u545? If yes, we may need to do more complex changes. If not, all is good besides some nitpicks

Before pushing, the only thing changed in Cargo.toml was the board name, flashed and ran the examples, made sure they work, then restored the Cargo.toml to the default.

PR #6447, opened 2026-07-04
fix: add `const` to eligible fns and `deny(clippy::missing_const_for_fn)`

vic1707:
Found a few fns in `embassy-sync` who could become `const fn`s, figure I could check all crates.
Added `const` to all eligible functions and denied `missing_const_for_fn` across all crates. I believe more `const` is better in embedded and in general. 
Happy to split into per-crate PRs or drop the `#![deny()]` attr if preferred.
Hope this helps!

### DIFF
--- a/cyw43-pio/src/lib.rs
+++ b/cyw43-pio/src/lib.rs
@@ -2,6 +2,7 @@
 #![allow(async_fn_in_trait)]
 #![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
+#![deny(clippy::missing_const_for_fn)]
 
 use core::slice;
 
--- a/cyw43/src/lib.rs
+++ b/cyw43/src/lib.rs
@@ -5,6 +5,7 @@
 #![deny(unused_must_use)]
 #![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
+#![deny(clippy::missing_const_for_fn)]
 
 // This mod MUST go first, so that the others see its macros.
 pub(crate) mod fmt;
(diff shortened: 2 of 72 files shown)


Dirbaio:
- I don't think we should enable `missing_const_for_fn`. There might be functions in the public API that *currently* can be const but we might want to make them non-const in the future, which would be a breaking change. If we indiscriminately constify everything that can be const now, we paint ourselves into a corner.
- Many of the functions you've marked as const will clearly never ever be used in const. 

Also the diff is huge so it's impossible to review. Please send smaller PRs, each marking a small set of related functions as const, only for functions you've verified make sense to be const and is realistic that someone might want to use them in const. One example is the math functions in embassy-time, but there might be more.

vic1707:
I can do the smaller PRs and remove the `#![deny()]` no problem 👍 

but  I'd like to know how you would decide if a function "makes sense to be const"?
Because of the lint's existence I always viewed `const` as a purely beneficial attribute, a bonus, not something we could not want 🤔 

I suppose [the `joinN` functions qualify as such](https://github.com/embassy-rs/embassy/pull/6447/changes#diff-1a914b97776219f2aeb47a4a6453dd03b32b81171f71a98dc4fbc213ae99b062L171)?

vic1707:
@Dirbaio can you enlighten me on this so I can make the proper PRs?

Dirbaio:
I can't think of a general rule, it has to be analyzed on a case by case basis.

For example, the join functions can be made const, but there's no benefit in doing so. Futures are something you create and run at runtime, it would be *extremely* rare to want to join two futures at compile-time and then store the joined future in a const/static and then run it at run-time. 

There's a very slight cost in marking things const: it's a promise in the public API. You can't take it back without a breaking change.

So in this particular case, since there's zero benefit and a very slight cost, imo it's best to not mark them as const.

PR #6448, opened 2026-07-05
adc/g4: fix DIFSEL not set for injected channels on STM32G4

Jacke-debug:
adc/g4: fix DIFSEL not set for injected channels on STM32G4

write_injected_sequence ignored the is_differential flag, so differential pin pairs configured as injected channels never had their DIFSEL bit set and were sampled single-ended instead.

Also switched configure_sequence to update DIFSEL via modify() per channel rather than building a fresh register and write_value()-ing it, the old approach zeroed bits for any channel outside the current regular sequence, which could clobber DIFSEL bits set by the injected path.

### DIFF
--- a/embassy-stm32/src/adc/g4.rs
+++ b/embassy-stm32/src/adc/g4.rs
@@ -1,5 +1,3 @@
-#[cfg(stm32g4)]
-use pac::adc::regs::Difsel as DifselReg;
 #[allow(unused)]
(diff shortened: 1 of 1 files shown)


Jacke-debug:
@xoviat is there something you would like me to  update?

xoviat:
Teleprobe is having some issues. When they are resolved, this will be merged.

PR #6450, opened 2026-07-06
nRF54LM20 USB HS fixes

HaoboGu:
1. On nRF54LM20, the DWC2 core is clocked from an oscillator supplied by the
USB power domain and cannot run without VBUS. Nordic documents this: "On USBHS, we cannot
access the DWC2 register until VBUS is detected and valid" (Zephyr,
udc_dwc2_nrf_usbhs_nrf54l.h). So, we track the endpoint enabled state in `EpState` to avoid accessing registers before is powered.

2. Current `RX_FIFO_EXTRA_SIZE_WORDS = 30` is enough for 64 byte USB-FS packet, but USB-HS has 512-byte wire maximum packet. Increase this value to make nRF54LM20 USB work on 512 byte packet.

### DIFF
--- a/embassy-nrf/src/usb/usbhs.rs
+++ b/embassy-nrf/src/usb/usbhs.rs
@@ -23,7 +23,8 @@ use crate::interrupt::typelevel::Interrupt;
 use crate::{interrupt, pac};
 
 const MAX_EP_COUNT: usize = 16;
-const RX_FIFO_EXTRA_SIZE_WORDS: u16 = 30;
+// USB-HS needs more FIFO space.
+const RX_FIFO_EXTRA_SIZE_WORDS: u16 = 256;
 const FIFO_DEPTH_WORDS: u16 = 3040;
 
 static BUS_WAKER: AtomicWaker = AtomicWaker::new();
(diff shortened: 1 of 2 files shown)


bugadani:
Can you connect a logical high level to VBUS, like its usually done on current ESPs? If we were to change esp-hal to support VBUS detection, would we run into the same issue, or is this an nRF-specific system issue? I'm slightly hesitant introducing more state unconditionally, although I'm not strongly against.

HaoboGu:
> Can you connect a logical high level to VBUS, like its usually done on current ESPs? If we were to change esp-hal to support VBUS detection, would we run into the same issue, or is this an nRF-specific system issue? I'm slightly hesitant introducing more state unconditionally, although I'm not strongly against.

The nRF54LM20A VBUS pin feeds a dedicated regulator (VREGUSB) that powers the USB PHY, and without it the DWC2 core can't be brought up at all. So setting a logical high level to VBUS doesn't work. ESPs won't run into this issue.

I know it's kind of strange to introduce the workaround for nRF54LM20A to the general DWC2 driver. An alternative is to introduce something like

```rust
pub struct OtgInstance<'d> {
    // ...

    /// Whether the DWC2 register file is currently safe to access.
    /// `None` = always accessible (STM32, ESP32, ...).
    pub core_accessible: Option<fn() -> bool>,
}
```

And check it in nRF54LM20's usb implementation if `core_accessible` is not `None`.

bugadani:
I'm conflicted. I don't have a hard problem accepting this - I'll test and merge soon, but if there was a way to tune the code without extra memory for us, or for an indirect function call for you, that would be best.

PR #6452, opened 2026-07-06
DAC blocking example

RaresAndrei0:
Added a simple DAC implementation using blocking. 

Tested on the STM32U545RE board using an LED.


examples/stm32u5/src/bin/dac.rs
```diff
@@ -0,0 +1,34 @@
...
+    // We need the internal LSI(Low Speed Internal oscillator) for DAC sync
+    config.rcc.ls = LsConfig::default_lsi();
+
+    // changing the mux to point to our clock(LSI)
+    config.rcc.mux.dac1sel = mux::Dacsel::Lsi;
+    let p = embassy_stm32::init(config);
+
+    info!("Board connected!");
+
+    let mut dac = DacChannel::new_blocking(p.DAC1, p.PA4);
```
eva-cosma, line 29: PA4 = what? An led on the nucleo board or a random gpio pin?

examples/stm32u5/src/bin/dac.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    dac.set(100);
```
eva-cosma, line 32: 100/255. Add a comment explaining the  led should be half-lit or smth like that

eva-cosma (changes requested):

examples/stm32u5/src/bin/dac.rs
RaresAndrei0: It is written at the top of the file in the comment. Adding a more explicit comment about the DAC pins on the STM32U5.

PR #6453, opened 2026-07-06
fix: handle missing connect event in older version

lulf:
Older esp-hosted-fg versions do not emit a connected event, so the driver will stop working with these versions. Gate the legacy behavior behind a feature flag.

Debugged/fix produced with claude.

### DIFF
--- a/embassy-net-esp-hosted/src/control.rs
+++ b/embassy-net-esp-hosted/src/control.rs
@@ -202,19 +202,19 @@ impl<'a> Control<'a> {
         // treating the ConnectAP response as the connection result.
         #[cfg(feature = "esp-hosted-fg")]
         {
-            let probe = with_timeout(
-                FW_VERSION_PROBE_TIMEOUT,
-                self.backend.get_fw_version(&mut self.ioctl),
-            )
-            .await
-            .unwrap_or(Err(Error::Timeout));
+            let probe = with_timeout(FW_VERSION_PROBE_TIMEOUT, self.backend.get_fw_version(&mut self.ioctl))
+                .await
+                .unwrap_or(Err(Error::Timeout));
             self.sync_connect = match probe {
                 Ok(version) => {
                     debug!("fw version: {:?}", version);
                     false
                 }
                 Err(e) => {
-                    info!("fw version probe failed ({:?}), assuming old esp-hosted-fg with synchronous connect", e);
+                    info!(
+                        "fw version probe failed ({:?}), assuming old esp-hosted-fg with synchronous connect",
+                        e
+                    );
                     true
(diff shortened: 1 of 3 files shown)


bugadani:
Perhaps it would be best to not use a newer version of embassy-net-esp-hosted with very old versions of esp-hosted-fg. This sort of feature-flag based workaround for protocol differences will likely be very unwieldy. What happes if you enable this feature flag but only with esp-hosted-mcu? What happens if you OTA update to esp-hosted-fg 1.0.0.0.0? The PR also desyncs the shared connect_state from the embassy-net-driver-channel's link state.

Perhaps as an alternative, the fix could be gated on the firmware version in runtime.

lulf:
> Perhaps it would be best to not use a newer version of embassy-net-esp-hosted with very old versions of esp-hosted-fg. 

That would be the ideal, but from what I can tell, this fix should allow it to 'work' with newer versions, so that one can first upgrade embassy, then upgrade the esp32 fw. 

> This sort of feature-flag based workaround for protocol differences will likely be very unwieldy. What happes if you enable this feature flag but only with esp-hosted-mcu?

Yes, but I think it is unavoidable when having devices in the field in some situations. Guarding previous behavior behind a flag like this I think is OK, especially when the old behavior was changed.

> What happens if you OTA update to esp-hosted-fg 1.0.0.0.0? The PR also desyncs the shared connect_state from the > embassy-net-driver-channel's link state.
> 
> Perhaps as an alternative, the fix could be gated on the firmware version in runtime.

Perhaps, I can check if this is feasible based on the boot up event.

bugadani:
> Perhaps, I can check if this is feasible based on the boot up event.

Unlikely, but there is a call to read the firmware version, at least on the newest FG. I think we can fall back to "old firmware quirk" mode if:

- this call returns an error (i.e. it doesn't exist on old old firmware)
- AND `esp-hosted-fg` is enabled

lulf:
> > Perhaps, I can check if this is feasible based on the boot up event.
> 
> Unlikely, but there is a call to read the firmware version, at least on the newest FG. I think we can fall back to "old firmware quirk" mode if:
> 
>     * this call returns an error (i.e. it doesn't exist on old old firmware)
> 
>     * AND `esp-hosted-fg` is enabled

Verified that this works for me, though I had to use a timeout since there is no response for an unknown command on the old fg firmware it seems. PR updated

bugadani (approved):

bugadani:
Thanks, I find this a better solution, as you have no way to forget enabling the quirkaround. I didn't expect the device to fail to answer to an incorrect/unknown command, though :/

PR #6476, opened 2026-07-09
stm32/usb: disable ULPI clock for OTG_HS FS host

joelsa:
I couldn't get OTG_HS using to work in order to receive keyboard events on an STM32H7.

Upon inspection, I found that `Bus::init` disables the ULPI clock when using the internal FS PHY, but `new_fs_host` only calls `common_init` and thus leaves the clock enabled.

On OTG_HS, the enabled ULPI clock makes the core wait for an external ULPI PHY that is not there. The core then never switches to host mode (`GINTSTS.CMOD` stays 0), `HPRT.PCSTS` never gets set, and the host never sees the downstream device.

This PR disables the ULPI clock bits in `new_fs_host` as well, matching `Bus::init`. This lets OTG_HS with the internal FS PHY switch to host mode and enumerate devices correctly.

### DIFF
--- a/embassy-stm32/src/usb/otg.rs
+++ b/embassy-stm32/src/usb/otg.rs
@@ -723,6 +723,21 @@ mod host_impl {
 
             super::super::common_init::<T>();
 
+            // The ULPI clock has to be disabled for an internal PHY.
+            // Left enabled, the OTG_HS core waits on an absent external ULPI PHY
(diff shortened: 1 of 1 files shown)


leftger (approved):
This looks good to me on the USB_HOST side but STM32 is more @xoviat 's domain. I'll let them be the final arbiter.

PR #6478, opened 2026-07-10
Update, sort the list of Embassy compatible HALs.

inferiorhumanorgans:
* Add `embassy-mcxa`, `embassy-mspm0`, `ra-hal`
* Format the external HAL list similarly to the in-tree HAL list
* Move description to after both lists


lulf (approved):
Thanks!

lulf:
bender run

PR #6480, opened 2026-07-10
stm32: GPDMA: fix linked-list table ownership and expose descriptor API

liebman:
make more features of gpdma available:
- Fix soundness issue where Channel::linked_list() took Table by value, dropping the descriptor memory while DMA was still reading from it. It now borrows &'a Table<N> so the compiler ties the table's lifetime to the transfer.
- Remove the unused const ITEM_COUNT generic from LinkedListTransfer -- item count is tracked at runtime in lli_state, the generic just added noise to type signatures.
- Make LinearItem::link_to(), unlink(), and transfer_count() public so users can build custom descriptor topologies (not just sequential/circular).
- Add TransferCompleteMode enum (EachBlock, EachLinkedListItem, LastLinkedListItem) to control when TC/HT events fire, with per-item support via LinearItem::set_transfer_complete_mode().
- Set SAP/DAP port assignments in LinearItem for GPDMA, matching the existing non-linked-list Channel::configure() logic.
- Add Channel::restart_linked_list() for restarting a chain from item 0.
- Re-export Table, LinearItem, RunMode from embassy_stm32::dma for ergonomic imports.

### DIFF
--- a/embassy-stm32/src/dma/gpdma/linked_list.rs
+++ b/embassy-stm32/src/dma/gpdma/linked_list.rs
@@ -126,7 +126,7 @@ impl LinearItem {
     /// Link to the next linear item at the given address.
     ///
     /// Enables channel update bits.
-    fn link_to(&mut self, next: u16) {
+    pub fn link_to(&mut self, next: u16) {
         let mut llr = regs::ChLlr(0);
 
         llr.set_ut1(true);
@@ -145,12 +145,12 @@ impl LinearItem {
     /// Unlink the next linear item.
     ///
     /// Disables channel update bits.
-    fn unlink(&mut self) {
+    pub fn unlink(&mut self) {
         self.llr = regs::ChLlr(0);
     }
 
     /// The item's transfer count in number of words.
-    fn transfer_count(&self) -> usize {
+    pub fn transfer_count(&self) -> usize {
         let word_size: WordSize = self.tr1.ddw().into();
         self.br1.bndt() as usize / word_size.bytes()
     }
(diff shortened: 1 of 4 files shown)


embassy-stm32/src/dma/gpdma/linked_list.rs
```diff
@@ -145,12 +163,22 @@ impl LinearItem {
...
     }
 
+    /// Set the transfer complete event mode (`TR2.TCEM`) for this item.
+    ///
+    /// In linked-list mode, the channel's TR2 register is overwritten by
+    /// each LLI when `UT2` is set, so TCEM must be configured per-item.
+    ///
+    /// See [`TransferCompleteMode`](super::TransferCompleteMode) for details.
+    pub fn set_transfer_complete_mode(&mut self, mode: super::TransferCompleteMode) {
+        self.tr2.set_tcem(mode.into());
```
xoviat, line 177: Should this be added to new_read and new_write?

liebman: Not sure, but maybe it makes sense, the current behavior just sets `let tr2 = regs::ChTr2(0)` and the 0 makes the default `EachBlock` in a hidden way.  so we could keep this "default" and make it explicit.  Or add an extra parameter to all the constructors.  the question then would be do we need to add it to `Table::new_circular` and `Table::new_ping_pong` (and update the current callers of these) as well?

xoviat: I'll take care of it later.

Issue #2783, opened 2024-04-06
stm32: DMA can't be used with DAC on some families (g4, others)


adamgreig:
## The bug

The DAC `write()` method on an STM32G4 immediately faults, because the DMA is configured for an 8- or 16-bit PSIZE which is not supported by the DAC registers and generates a bus fault.

![image](https://github.com/embassy-rs/embassy/assets/47219/09c2ba3c-9c63-495b-a1ea-51bf6a0a55f8)

The same note about 32-bit access is present in all the other reference manuals I checked (F0, F1, F4, F7, H7) but it seems on those families no bus error is generated (see below), so there's no bug.

The DAC driver's `write()` method takes either a `&[u8]` or a `&[u16]` and sets up a DMA transfer with that buffer. The DMA driver sets up MSIZE and PSIZE based on the data type to either both be 8-bit or both be 16-bit. As soon as the DMA transfer is started, the DMA generates a bus fault and stops.

## Possible fix 1

If we change the DAC driver's `write()` method to always take a `&[u32]`, the correct DMA transfer is generated and everything works by-the-book on all platforms. However, the user buffer must be 2x or 4x the size, so it's a very wasteful approach.

## Possible fix 2

Currently the DMA driver doesn't support PSIZE != MSIZE. If it did, on G4, we could run a transfer where MSIZE=8 or 16 bit, and PSIZE=32 bit, and the BDMA behaviour is to fill the extra bits in with 0s:

![image](https://github.com/embassy-rs/embassy/assets/47219/5288eb8f-2674-444a-8c1c-a1a7bec158c1)

This works great for G4 and other BDMA platforms: the user passes a `&[u8]` or `&[u16]` and each word is zero-extended to 32 bits before being written to the DAC as a 32-bit access. However, there's no API for doing this at the moment. There's one other catch: (non-B) DMA has a completely different behaviour when MSIZE != PSIZE: it packs multiple memory reads into a single peripheral write, so a user-provided `&[u16]` would have two entries packed into a single u32 which is written to the peripheral, and it uses NDTR as the number of peripheral writes instead of the size of the data in memory:

![image](https://github.com/embassy-rs/embassy/assets/47219/7251f6f8-5f46-4a97-9938-cd12653a1ccf)

[...]

PR #6487, opened 2026-07-12
stm32: fix dac for not-gpdma

xoviat:


xoviat:
bender run

adamgreig:
Conclusions from discussion today on Matrix:

- On F2, F4, F7, H7 (DMA but not BDMA), the DMA peripheral will always repack if you have a u16 read and u32 write and therefore must also be in FIFO mode; there's no way to do a u16 read, zero-extend, and write as u32
- On every other platform (and H7 BDMA), BDMA will zero extend in this case
- On H7, only DMA implements `dac::Dma` trait, so we only get the F4-like behaviour (packing, no zero extending)
- The DAC claims to need 32 bit writes on all platforms
    - On G4 and probably others, DMA faults if configured for 16-bit writes
    - On F4 and H7 and probably F2 and F7, 16-bit DMA writes work perfectly
- We'd like to keep the actual word size in memory 16-bit to avoid doubling all the buffers
- So in conclusion: `cfg(dma)` for F2/F4/F7/H7 should use 16-bit PSIZE in DMA, causing it to have equal read/write, not try to pack, perform an "invalid" but successful 16-bit write to DAC, while `cfg(not(dma))` for every other platform, which uses BDMA, is configured with 32-bit PSIZE, causing it to zero-extend and perform a valid 32-bit write to DAC

adamgreig:
cc #2783 #6277 #6200

PR #6488, opened 2026-07-12
nrf/qdec: write readclracc after reportrdy

michelrandahl:
Bug discovered while playing around with my micro:bit v2 MCU (nRF52833) and an Alps Alpine EC11E15244G1 encoder.


lulf (approved):
Thanks!

PR #6493, opened 2026-07-13
stm32: cleanup ptp and optimize eth performance

xoviat:
misc cleanups and volatile read optimizations.


embassy-stm32/src/eth/ptp.rs
```diff
@@ -69,223 +182,196 @@ impl PtpTimestamp {
...
+        return None;
     }
+
+    slots[slot_index(slots, packet_id)].load(packet_id)
 }
 
-pub(crate) use imp::PtpTimestampSink;
-#[cfg(feature = "ptp")]
-pub use imp::PtpTimestampStore;
+fn slot_index(slots: &[TimestampSlot], packet_id: u32) -> usize {
```
jordens, line 378: In hindsight I'd probably want to inline this.

xoviat:
bender run

jordens (commented):
I just stumbled over this by accident. Thanks for the cleanup!
Prefering to cfg-gate each item over cfg-muxing zero size impls is fine by me. More of a personal preference thing.

xoviat:
@jordens The reason for this is that I'm trying to implement two network stacks using one ethernet port with vlan, and there's no way to split the stacks without copying using the existing net-driver api. Therefore, I needed to add vlan to packetmeta so that I could inspect the vlan before consuming the token, and once I did that I thought I might as well implement hardware vlan tag insertion/stripping and therefore I needed to dive into this code for the first time.

The first problem that I found is that the existing code does not make full use of all of the buffers because it has one index  that tracks both the descriptors and the buffer. However, there are more descriptors than buffers because some of them are context descriptors. Were you aware of this issue?

jordens:
Yes, a receive context descriptor consumes a ring entry without using its associated buffer. The driver cannot predict which next entry DMA will turn into a context descriptor and therefore every DMA-owned entry must initially have a unique valid buffer. The one-to-one mapping is intentional and from what I read correct; a separate buffer index alone would not improve capacity. Contexts do reduce effective ring depth during bursts of timestamped packets.

jordens:
I suspect that also means that the vlan tag metadata should not lightly be mixed up with the ptp timestamp metadata.

embassy-stm32/src/eth/ptp.rs
xoviat: This is a misunderstanding of the rust compiler. The only time you might consider #[inline] is if the function is called outside of the crate. The rust compiler can inline functions inside of the crate without the attribute.

xoviat:
bender run

embassy-stm32/src/eth/ptp.rs
jordens: I meant literally inline the modulo. Pure readability.

PR #6500, opened 2026-07-14
feature: add support for GPDMA on the DAC peripheral on STM32U5

RaresAndrei0:
Implemented GPDMA support, tested using a simple u8 transfer to a LED

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -336,6 +336,27 @@ impl<'d> DacChannel<'d, Async> {
             w.set_dmaen(self.idx, false);
         });
     }
+
+    #[cfg(gpdma)]
+    pub async fn write<W: Word>(&mut self, buffer: &[W]) {
+        let _scoped_wake_guard = self.info.rcc.wake_guard();
+
+        // tsel and ten needed to connect the timer to the GPDMA transfer
+        self.info.regs.cr().modify(|w| {
+            w.set_tsel(self.idx, 5);
+            w.set_ten(self.idx, true);
+            w.set_en(self.idx, true);
+            w.set_dmaen(self.idx, true);
+        });
+
+        let dma = self.dma.as_mut().unwrap();
+        let tx_options = crate::dma::TransferOptions { ..Default::default() };
+        let tx_dst = W::dma_ptr(self.info.regs, self.idx);
+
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/dac/mod.rs
```diff
@@ -336,6 +336,27 @@ impl<'d> DacChannel<'d, Async> {
...
         });
     }
+
+    #[cfg(gpdma)]
+    pub async fn write<W: Word>(&mut self, buffer: &[W]) {
+        let _scoped_wake_guard = self.info.rcc.wake_guard();
+
+        // tsel and ten needed to connect the timer to the GPDMA transfer
+        self.info.regs.cr().modify(|w| {
+            w.set_tsel(self.idx, 5);
```
eva-cosma: This should be change-able either via feature flags or some other mechanism.

eva-cosma: Also, is there an enum for this "5"? The pac should have generated something... probably

RaresAndrei0: The "5" stands for the TIM6's trigger output.
Will add a flag so that the tsel can be easily changed when the write function is called, such that the chosen timer will fall into the user's hands.

RaresAndrei0:
Will improve the function by adding a TriggerTimer enum such that it will become a bit more verbose and easy to use(Typing "Tim6" instead of passing a "5" to the function)

Also will change the DacChannel::new, adding a timer parameter so the GPDMA timer will be declared in the ::new function and will be tied to the DAC instead of declaring and configuring it in main. Should also make the code cleaner and easier to use, this can work for now :)

embassy-stm32/src/dac/mod.rs
```diff
@@ -342,6 +336,31 @@ impl<'d> DacChannel<'d, Async> {
...
+        // tsel and ten needed to connect the timer to the GPDMA transfer
+        self.info.regs.cr().modify(|w| {
+            w.set_tsel(self.idx, trgo_timer);
+            w.set_ten(self.idx, true);
+            w.set_en(self.idx, true);
+            w.set_dmaen(self.idx, true);
+        });
+
+        let dma = self.dma.as_mut().unwrap();
+        let tx_options = crate::dma::TransferOptions { ..Default::default() };
```
xoviat: The default option is to pack, which is wrong. This needs no packing.

xoviat:
See #6513

PR #6502, opened 2026-07-14
executor: emit rtos-trace system_idle only at real sleep sites

basil-huber:
Previously, executors would call rtos-trace's system_idle callback after polling.
For InterruptExecutors this is incorrect: the executor finishes polling and returns to the preempted (lower-priority) context, so execution continues and the system is not idle. Only the thread-mode executor reaching its wait site means the system is truly idle.

This misleads notably SystemView
(https://www.segger.com/products/development-tools/systemview/), connected via the systemview-target crate
(https://crates.io/crates/systemview-target), which misinterprets the events and produces a corrupted timeline.

Stop emitting `system_idle` in `executor_idle`, and add a public `Executor::trace_system_idle()` that each thread-mode run loop calls at its genuine wait site instead. It is a no-op unless a trace feature is enabled, and must not be called from an interrupt executor.

The behaviour of the following executors stays the same:
- embassy-executor/src/platform/cortex_m.rs: Executor
- embassy-executor/src/platform/cortex_ar.rs: Executor
- embassy-rp/src/executor.rs: Executor

The following executors do NOT call `system_idle` anymore:
- embassy-executor/src/platform/cortex_m.rs: InterruptExecutor
- embassy-rp/src/executor.rs: InterruptExecutor
- embassy-stm32/src/executor.rs: InterruptExecutor
- embassy-executor/src/platform/spin.rs: Executor
- embassy-executor/src/platform/wasm.rs: Executor

The following executors call `system_idle` only before actually idling:
- embassy-executor/src/platform/riscv32.rs: Executor
- embassy-executor/src/platform/std.rs: Executor
- embassy-executor/src/platform/avr.rs: Executor
- embassy-stm32/src/executor.rs: Executor
- embassy-mcxa/src/executor.rs: Executor

Note: these differences only affect the `rtos-trace` feature.


embassy-executor/src/platform/std.rs
```diff
@@ -72,6 +72,7 @@ mod thread {
                     break;
                 }
 
+                self.inner.trace_system_idle();
```
bugadani, line 75: This call is in conflict with your "in case there are multiple executors" goal. On std you can run any number of executors on any number of threads.

embassy-rp/src/executor.rs
```diff
@@ -123,6 +123,7 @@ mod thread {
             loop {
                 unsafe {
                     self.inner.poll();
+                    self.inner.trace_system_idle();
```
bugadani, line 126: I think this has the same issue as std - you can run a thread executor on both cores, both need to go idle for the system to be considered idle.

embassy-executor/src/platform/std.rs
basil-huber: That's a good point. And I don't see a way around this w/o adding complexity.
And I guess already the complexity added as is, is quite big if it's only for the STM32s.
I wonder if I should not simply drop this PR.

Dirbaio:
closing in favor of #6925 which contains a rebased + reworked version.

Note rtos-trace has been removed, see #6920 . Sorry about that :sweat: .

PR #6505, opened 2026-07-14
stm32/qspi: add checks for address size

TechnoPorg:
This catches errors where the address gets silently truncated because of a mismatch with the address size.

This also makes the capitalization of the 24- and 32-bit variants of the AddressSize enum consistent with itself. The other *SPIs are also inconsistent, though: hspi uses all caps, whereas xspi uses all lowercase, and ospi uses a mix of both. Since the enum is the same between all of them, perhaps they could be combined into a single definition?

### DIFF
--- a/embassy-stm32/src/qspi/enums.rs
+++ b/embassy-stm32/src/qspi/enums.rs
@@ -136,9 +136,21 @@ pub enum AddressSize {
     /// 16-bit address
     _16Bit,
     /// 24-bit address
-    _24bit,
+    _24Bit,
     /// 32-bit address
-    _32bit,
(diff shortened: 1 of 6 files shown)


xoviat:
You need to update the examples.

TechnoPorg:
Will do, thanks. What's the best way to run CI locally to catch this sort of problem in the future?

xoviat:
> What's the best way to run CI locally to catch this sort of problem in the future?

install cargo-embassy-devtool and run it. However, the CI is pretty fast so I usually use search to find problems when changing stuff and leave the CI to pick up the other problems.

PR #6506, opened 2026-07-14
stm32/qspi: move address size to TransferConfig

### LINKED ISSUE #6498: stm32: Cannot reconfigure address size for Qspi peripheral
I'm currently writing a driver for the MX35LF1GE4AB NAND flash ([datasheet](https://www.mxic.com.tw/Lists/Datasheet/Attachments/9075/MX35LF1GE4AB,%203V,%201Gb,%20v2.0.pdf)), which has a quad SPI interface.

One "delightful" thing about this chip is that it has a variable address width per operation (see Table 1 on page 11). So, for example, PAGE READ uses a 24-bit address, but GET FEATURE uses an 8-bit address. This matches what's seen in the timing diagrams for each of those operations.

The Qspi driver currently only allows setting the address width once, in the initial `qspi::Config`, making it impossible to interact with this device without reinitializing the bank every time or directly changing the ADSIZE register via the PAC.

Would a PR moving address_size from `qspi::Config` to `qspi::TransferConfig` be welcome? If so, I will send one in within the next day or so!

TechnoPorg:
Closes #6498.

This now matches the behaviour of ospi/hspi/xspi, which store the address size in TransferConfig. I have tested the patch with the NAND flash driver that motivated the opening of the above issue, and it now works as expected.

### DIFF
--- a/embassy-stm32/src/qspi/mod.rs
+++ b/embassy-stm32/src/qspi/mod.rs
@@ -32,6 +32,8 @@ pub struct TransferConfig {
     pub instruction: u8,
     /// Flash memory address
     pub address: Option<u32>,
+    /// Address size (8/16/24/32-bit)
+    pub address_size: AddressSize,
     /// Number of dummy cycles (DCYC)
     pub dummy: DummyCycles,
 }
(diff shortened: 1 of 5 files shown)


TechnoPorg:
Just realized I'll probably have to fix examples 😅

PR #6512, opened 2026-07-14
nrf: add System OFF support for nRF54L series

erdemsimsek:
Adds System OFF support for the nRF54L series to `embassy-nrf`.

- `power::set_system_off()` now covers nRF54L by writing
  `REGULATORS.SYSTEMOFF`, same as the existing nRF91 path (the nRF54L has
  no dedicated POWER.SYSTEMOFF register). This matches Zephyr's
  implementation in `soc/nordic/common/poweroff.c`

Note: on nRF54L all RAM is retained in System OFF by reset default
(https://docs.nordicsemi.com/r/bundle/ps_nrf54l15/page/memconf.html), so power consumption stays above the datasheet floor. RAM retention control is intentionally left out of this PR.  I'll open a
separate issue to discuss the API for it.

Tested on hardware (nRF54L15-DK): flashed the example with probe-rs.

Steps to test on hardware:
- Navigate into `embassy/examples/nrf54l15-app`
- Build and flash the application `cargo run --release --bin system_off`
- Terminate the probe-rs connection as it is keeping the debug port active and prevents the uC going into `system off` state
- Unplug/plug the power cable
- `led0` is on and after 5 seconds `led0` turns off as nRF54L15 goes into `system off` mode
- Press `button0` to wake-up, it will be cold boot.

### DIFF
--- a/embassy-nrf/CHANGELOG.md
+++ b/embassy-nrf/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## Unreleased - ReleaseDate
 
+- added: System OFF support for the nRF54L series.
+
 ## 0.11.0 - 2026-06-16
 
 - added: support for the SQSPI peripheral for nRF54.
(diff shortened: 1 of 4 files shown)


lulf (approved):
Thank you!

PR #6513, opened 2026-07-14
stm32/dac: impl. write for gpdma and other cleanups

xoviat:
closes #6500.

### DIFF
--- a/embassy-stm32/src/dac/mod.rs
+++ b/embassy-stm32/src/dac/mod.rs
@@ -9,10 +9,11 @@ use core::slice;
 #[cfg(stm32g4)]
 use dac::vals;
 use embassy_hal_internal::PeripheralType;
+use embassy_hal_internal::drop::OnDrop;
 use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
 pub use ringbuffered::RingBufferedDacChannel;
 
-use crate::dma::{ChannelAndRequest, word as dma};
+use crate::dma::{ChannelAndRequest, no_packing, word as dma};
 use crate::mode::{Async, Blocking, Mode as PeriMode};
 #[cfg(any(dac_v3, dac_v4, dac_v5, dac_v6, dac_v7))]
 use crate::pac::dac;
@@ -182,7 +183,7 @@ impl<'d> DacChannel<'d, Async> {
     ) -> Self {
         pin.set_as_analog();
         Self::new_inner::<T, C>(
-            peri,
+            Some(peri),
             None,
             new_dma!(dma, _irq),
             #[cfg(any(dac_v3, dac_v4, dac_v5, dac_v6, dac_v7))]
@@ -208,7 +209,7 @@ impl<'d> DacChannel<'d, Async> {
     ) -> Self {
         pin.set_as_analog();
         Self::new_inner::<T, C>(
-            peri,
+            Some(peri),
             Some(trigger.signal()),
             new_dma!(dma, _irq),
             #[cfg(any(dac_v3, dac_v4, dac_v5, dac_v6, dac_v7))]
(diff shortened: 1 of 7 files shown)


RaresAndrei0:
I have a STM32U545RE and I can't seem to make a successful transfer using GPDMA using this version of write, I might need some guidance. 
From my research, the GPDMA transfer on my board won't start unless connected to a hardware timer trigger, and it seems your PR completely ditched the timer trgo select on the control register.
Of course, I can be setting up the transfer the wrong way...

RaresAndrei0 (approved):

RaresAndrei0:
Thanks for the help, seems like the GPDMA works now

embassy-stm32/src/dma/dma_bdma.rs
```diff
@@ -178,6 +178,11 @@ pub struct TransferOptions {
     pub word_swap: bool,
 }
 
+/// Return the no packing value
+pub const fn no_packing() -> bool {
+    false
+}
+
```
eva-cosma, line 185: I think this could be an enum instead

enum Packing {
 Disabled,
 Enabled
}

embassy-stm32/src/dma/gpdma/mod.rs
```diff
@@ -350,6 +351,11 @@ pub struct TransferOptions {
     pub trigger: Option<TriggerConfig>,
 }
 
+/// Return the no packing value
+pub const fn no_packing() -> Pam {
+    Pam::ZeroExtendOrLeftTruncate
+}
+
```
eva-cosma, line 358: Any reason we need this and we couldn't use Pam::ZeroExtendOrLeftTruncate directly?

embassy-stm32/src/dac/mod.rs
```diff
@@ -443,14 +473,17 @@ impl<'d> DacChannel<'d, Blocking> {
 
 impl<'d, M: PeriMode> DacChannel<'d, M> {
     fn new_inner<T: Instance, C: Channel>(
-        _peri: Peri<'d, T>,
+        peri: Option<Peri<'d, T>>,
         trigger: Option<u8>,
```
eva-cosma, line 477: Could you comment what the u8 value is supposed to represent? What valid values would be, etc

eva-cosma: Could be worthwhile to comment the other parameters as well, since its non-obvious when something should be None, vs Some, what the values should be, etc

embassy-stm32/src/dac/mod.rs
```diff
@@ -308,39 +308,69 @@ impl<'d> DacChannel<'d, Async> {
...
+            w.set_en(self.idx, true);
+            w.set_dmaen(self.idx, true);
         });
+
+        let _guard = OnDrop::new(|| {
+            self.info.regs.cr().modify(|w| {
+                w.set_en(self.idx, false);
+                w.set_dmaen(self.idx, false);
+            })
+        });
```
eva-cosma, line 358: As a personal preference, I would think its more obvious if we just add these disable commands at the end of the function, since there is no other place this would get dropped (if I am not missing something)

embassy-stm32/src/dac/mod.rs
```diff
@@ -308,39 +308,69 @@ impl<'d> DacChannel<'d, Async> {
...
+        let tx_options = crate::dma::TransferOptions {
+            circular: true,
+            half_transfer_ir: false,
+            complete_transfer_ir: false,
+            packing: no_packing(),
+            ..Default::default()
+        };
+
+        // Initiate the correct type of DMA transfer depending on what data is passed
+        let tx_f = unsafe { dma.write_raw(W::dma_buf(data), W::dma_ptr(self.info.regs, self.idx), tx_options) };
```
eva-cosma, line 371: Safety comment

eva-cosma (changes requested):

PR #6517, opened 2026-07-15
mspm0: add OPA driver with buffer and non-inverting PGA modes

ozongzi:
Adds an OPA driver for MSPM0 parts that include the peripheral: G150x, G350x, L130x, and L134x.

The driver currently supports the two modes:

- Buffer (TRM 17.2.7.2)
- Non-inverting PGA with x2–x32 gain (TRM 17.2.7.3.2)

```rust
let mut adc = Adc::new_async(p.ADC1, Irqs, Default::default());
let mut opa = Opa::new(p.OPA1, Default::default());

let mut out = opa.pga_ext(p.PB19, p.PA16, Gain::X4);

let raw = adc.irq_read(&mut out, Conversion::default()).await;
out.set_gain(Gain::X16);
```

`_ext` methods drive the corresponding `OPAx_OUT` pin. `_int` methods leave the output internal and only expose it to the ADC. The returned output handle implements `AdcChannel`, keeps the amplifier enabled while borrowed by the ADC, and disables it on drop.

The non-inverting input may be an `OPAx_INy+` pin or one of the internal sources exposed by `NonInvertingInput`: DAC12, DAC8, VREF, or ground. Pin implementations are generated from the metapac metadata, including the correct PSEL value for each input.

PGA gain can be changed while the amplifier is running. This is useful for auto-ranging and avoids rebuilding the output handle for every gain change.

OPA outputs use fixed internal ADC connections that are not currently present in mspm0-data:

- G150x/G350x: OPA0 → ADC0 channel 13, OPA1 → ADC1 channel 13
- L130x/L134x: OPA0 → ADC0 channel 12, OPA1 → ADC0 channel 13

These mappings are generated from a small family table in `build.rs`, based on TI's `ADC12_internalConnections.js`.

Chopping is available through `Config` and defaults to disabled. Standard chopping introduces output ripple; `AdcAveraging` is intended for use with ADC hardware averaging.

### DIFF
--- a/embassy-mspm0/src/opa.rs
+++ b/embassy-mspm0/src/opa.rs
@@ -6,6 +6,9 @@
 //!
 //! - **Buffer** (unity gain follower): [`Opa::buffer_ext`] / [`Opa::buffer_int`]
 //! - **Non-inverting PGA** (x2..x32): [`Opa::pga_ext`] / [`Opa::pga_int`]
+//! - **Non-inverting PGA about a reference**: [`Opa::pga_biased_ext`] /
+//!   [`Opa::pga_biased_int`], which drive the bottom of the gain ladder from a
+//!   [`LadderBottom`] source instead of grounding it
 //!
 //! `_ext` variants drive the OPAx_OUT pin; `_int` variants keep the output
 //! off the pins and only route it to the ADC. Both can be sampled by the ADC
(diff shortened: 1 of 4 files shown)


embassy-mspm0/src/opa.rs
```diff
@@ -0,0 +1,495 @@
...
+/// ```text
+/// Vout = gain * Vin + (1 - gain) * Vladder
+/// ```
+///
+/// Grounding it gives the plain `Vout = gain * Vin`, which also means the
+/// input's own DC is multiplied by the gain: at x32 an input sitting 50 mV
+/// away from where it needs to be moves the output by 1.6 V. Driving the
+/// ladder bottom from the DAC12 instead makes the DC operating point of the
+/// output a free variable, settable independently of the gain, which is what
+/// lets a high gain be used on a signal whose DC is not already placed for it.
```
i509VCB, line 168: This section of the documentation has the tone of a Q2 financial report from \<mega name corporation here>.

The documentation here can be simplified by quite a bit.

i509VCB:
Architecture discussed on matrix would probably be my preference here.

Ideally so we can do change gain on the fly without exposing it when unneeded.

PR #6518, opened 2026-07-15
net-driver: implement ptp and vlan support

xoviat:
This has no breaking changes.

### DIFF
--- a/embassy-net-driver/src/lib.rs
+++ b/embassy-net-driver/src/lib.rs
@@ -70,6 +70,15 @@ pub trait Driver {
     where
         Self: 'a;
 
+    /// Poll the driver for timestamps and return a pair of (id, `Timestamp`)
+    ///
+    /// Ids may be reused, and therefore this method should be called before calling
+    /// `receive` or `transmit` to avoid deducing an incorrect association.
+    #[allow(unused_variables)]
+    fn poll_timestamp(&mut self, cx: &mut Context) -> Option<(u8, Timestamp)> {
+        None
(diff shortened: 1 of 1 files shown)


diondokter (approved):
Approve on the timestamp

Dirbaio:
as we discussed on Matrix, I don't think VLAN should be part of the net driver trait.

About PTP timestamps: smoltcp uses the packetmeta id mechanism for this. We're doing something different here, is there a good reason? How is embassy-net going to use the timestamp to provide the upper layers of PTP?

xoviat:
> as we discussed on Matrix, I don't think VLAN should be part of the net driver trait.

What pushed me over the line on this is that this change is not breaking, whereas adding buf(&self) -> &[u8] was. Further, it's not like vlan is something that will change. This is a standard that will be around for a very long time. The problem with the current trait is that it's not possible to look ahead without consuming the token, as I have said. 

So the solutions to this are to add a breaking change, or add non-breaking change that only drivers that support vlan need to implement. There are no changes that will support all drivers existing currently with no work on the driver side, and copying the entire packet is not acceptable.

xoviat:
I will discuss PTP after we have agreed on the vlan issue.

xoviat:
If you're open to breaking changes, then I think the best shape of the trait for the `RxToken` is to have `fn buf(&mut self) -> &mut [u8]` and `consume(self)` methods (no vlan method). However, I'm not sure that it's worth the churn to add this. Up to you.

PR #6523, opened 2026-07-15
stm32/can: add loopback examples for classic CAN, FD, and split TX/RX

VictorVulturescu:
Adds three CAN bus examples for the STM32U5: one using classic CAN frames, one using CAN-FD frames, and one driving RX and TX from separate tasks via split().

All three were hardware-tested on a NUCLEO-U545RE-Q development board (stm32u545re feature, STM32U545REIx chip) and successfully compile against this directory's pinned stm32u5g9zj feature (STM32U5G9ZJTxQ chip).

### DIFF
--- a/examples/stm32u5/src/bin/can.rs
+++ b/examples/stm32u5/src/bin/can.rs
@@ -5,12 +5,11 @@
 // Tested on a NUCLEO-U545RE-Q; builds under this directory's stm32u5g9zj feature.
 
 use defmt::{error, info};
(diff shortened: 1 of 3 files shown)


xoviat:
Install cargo-embassy-devtool from git and run `cargo embassy-devtool fmt`. Or run `./fmtall.sh`.

xoviat:
now unmark this as draft and I'll merge it.

PR #6524, opened 2026-07-16
mcxa: enable DMA0 channels 8-11 on MCXA5xx

bramsdell-ms:
## Summary

The MCXA5xx (e.g. MCXA577) exposes **12 DMA0 channels (0-11)**, but the `embassy-mcxa` driver only implemented channels 0-7 — a leftover from the 8-channel MCXA2xx port. This PR enables DMA0 channels 8-11 on MCXA5xx.

## Changes

- **Bump `nxp-pac`** to pick up the `EDMA_0_TCD` `Tcd12` fix (embassy-rs/nxp-pac#112), so the TCD register array actually has 12 slots. Without it, channels 8-11 would index past the 8-slot array at runtime.
- **Add `impl_channel!` and `impl_dma_interrupt_handler!` for `DMA0_CH8`-`DMA0_CH11`** (`mcxa5xx` only). NVIC wiring is derived dynamically from each channel's interrupt, so no additional plumbing is needed.
- **Replace the hardcoded `[_; 8]` `STATES`/`CALLBACKS` array sizes** with a `DMA0_CHANNELS` const (12 on `mcxa5xx`, 8 otherwise).

### Incidental: FlexCAN clock-gate build fix

The `nxp-pac` bump also pulls in FlexCAN clock-gate metadata that `embassy-mcxa` did not yet handle (unrelated to DMA, but required for the crate to compile against current `nxp-pac` `main`). Minimal fix:
- Add a `CanConfig` gate-config placeholder mirroring the existing `DacConfig`.
- Map the `CANn` peripheral name to its `flexcann` MRCC register field in the gate codegen.

There is no CAN HAL driver yet; this only keeps the build green.

## Not included

**DMA1 (CH0-3)** remains unimplemented and is intentionally left for a follow-up, as it requires decoupling the flat channel index from the per-controller TCD/state indexing.

## Testing

Builds clean for both `--features mcxa5xx` and `--features mcxa2xx` on `thumbv8m.main-none-eabihf`.

### DIFF
--- a/embassy-mcxa/Cargo.toml
+++ b/embassy-mcxa/Cargo.toml
@@ -59,7 +59,7 @@ grounded = { version = "0.2.1", features = ["cas", "critical-section"] }
 heapless = "0.9"
 maitake-sync = { version = "0.3.0", default-features = false, features = ["critical-section", "no-cache-pad"] }
 nb = "1.1.0"
-nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "a46946e7f5ba3b5d8136f2962400300d54a99d6c", features = ["rt"] }
+nxp-pac = { git = "https://github.com/embassy-rs/nxp-pac.git", rev = "8512ecedcd0fe9c2713d06f3bbd1b7f93805f306", features = ["rt"] }
 paste = "1.0.15"
 
 rand-core-06 = { package = "rand_core", version = "0.6" }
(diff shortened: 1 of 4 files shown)


bramsdell-ms:
This is more elegantly fixed in #6525 (includes DMA1 support.) I'll leave this here just in case, but if that PR merges, this one is obsolete.

PR #6525, opened 2026-07-16
[MCXA] Add DMA1 support

diondokter:
Adds DMA0_8..=11 and DMA1_0..=3 support. This is now generated from the build.rs.

@bramsdell-ms I've tried adding DMA1 to the HAL. Implementation should work, but it doesn't... But I don't really have time to debug.

To reproduce, go to the mcxa5xx examples, dma_mem_to_mem and then change dma0_ch0 to dma1_ch0. Run the example and you'll see a hardfault because of some pointer write.

### DIFF
--- a/embassy-mcxa/build.rs
+++ b/embassy-mcxa/build.rs
@@ -39,11 +39,6 @@ fn main() {
         }
 
         cfgs.enable(&driver_to_cfg_name(peripheral.driver_name));
-
-        // Also enable cfgs for each peripheral
-        let peripheral_gate = format!("peripheral_{}", peripheral.name);
-        cfgs.declare(&peripheral_gate);
-        cfgs.enable(&peripheral_gate);
     }
 
     let generated = [
--- a/embassy-mcxa/build.rs
+++ b/embassy-mcxa/build.rs
@@ -39,6 +39,11 @@ fn main() {
         }
 
         cfgs.enable(&driver_to_cfg_name(peripheral.driver_name));
+
+        // Also enable cfgs for each peripheral
+        let peripheral_gate = format!("peripheral_{}", peripheral.name);
+        cfgs.declare(&peripheral_gate);
+        cfgs.enable(&peripheral_gate);
     }
 
     let generated = [
@@ -56,6 +61,7 @@ fn main() {
         generate_ctimer_pin_impls(),
         generate_lpuart_pin_impls(),
         generate_flexspi_pin_impls(),
+        generate_dma_impls(),
     ];
 
     let out_dir = &PathBuf::from(env::var_os("OUT_DIR").unwrap());
@@ -158,6 +164,10 @@ fn generate_cc_gates() -> TokenStream {
     let mut generated = TokenStream::new();
 
     for peripheral in METADATA.peripherals {
+        if peripheral.name.to_ascii_lowercase().starts_with("can") {
+            continue;
+        }
+
         let Some(gate) = peripheral.gate.as_ref() else {
             continue;
         };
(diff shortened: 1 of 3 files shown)


embassy-mcxa/build.rs
```diff
@@ -158,6 +164,10 @@ fn generate_cc_gates() -> TokenStream {
     let mut generated = TokenStream::new();
 
     for peripheral in METADATA.peripherals {
+        if peripheral.name.to_ascii_lowercase().starts_with("can") {
+            continue;
+        }
+
```
diondokter, line 165: Skipping CAN because something in nxp-pac broke. I'm guessing/hoping #6435 fixes that

bramsdell-ms: So I patched this in #6524 -- there seems to be a missing config as well as a naming inconsistency

diondokter: Yep! I saw.

Didn't notice your PR. I was guessing you'd be asleep.
Your PR works, but my PR is a more proper durable implementation

bramsdell-ms: Agree completely that your solution is better -- if we can both extend DMA0 and add DMA1 I'll take it.

bramsdell-ms:
@diondokter Found it — the hardfault happens because `dma::init()` never brings up the **DMA1** instance.

`init()` only enables the DMA0 clock gate + releases DMA0 reset (`enable_and_reset::<DMA0>`) and configures **DMA0's** management page (`DMA0.mp_csr()`). DMA1 is left unclocked and held in reset, so the first write to any `EDMA_1_TCD` register (e.g. the TCD reset in `setup_typical`) triggers a bus fault → hardfault. That's exactly the "pointer write" fault you saw when switching `dma0_ch0` → `dma1_ch0`.

The DMA1 gate is already generated by the new build.rs (`impl_cc_gate!(DMA1, mrcc_glb_acc0, mrcc_glb_rst0, dma1, NoConfig)`), so `enable_and_reset::<DMA1>` just works. We only need to enable/reset it and configure its `MP_CSR`, mirroring the DMA0 setup (mcxa5xx only):

```diff
diff --git a/embassy-mcxa/src/dma.rs b/embassy-mcxa/src/dma.rs
--- a/embassy-mcxa/src/dma.rs
+++ b/embassy-mcxa/src/dma.rs
@@ use crate::pac::{self, Interrupt};
+#[cfg(feature = "mcxa5xx")]
+use crate::peripherals::DMA1;
 use crate::peripherals::DMA0;
@@ pub(crate) fn init() {
         w.set_gmrc(true);
     });
 
+    // Enable DMA1 clock, release reset, and configure its management page.
+    // Without this, any access to the DMA1 TCD registers faults because the
+    // peripheral is unclocked and held in reset.
+    #[cfg(feature = "mcxa5xx")]
+    {
+        let _ = unsafe { enable_and_reset::<DMA1>(&NoConfig) };
+
+        pac::DMA1.mp_csr().modify(|w| {
+            w.set_edbg(true);
+            w.set_erca(true);
+            w.set_halt(Halt::NormalOperation);
+            w.set_gclc(true);
+            w.set_gmrc(true);
+        });
+    }
+
     // Enable all DMA request lines for non-secure access.
     #[cfg(all(feature = "mcxa5xx", feature = "dma-ipd-req"))]
     for sec_gp_i in 0..=9 {
```

Verified on an **FRDM-MCXA577**: with this patch, `dma_mem_to_mem` using `DMA1_CH0` passes both mem-to-mem and memset — no hardfault:

```
[INFO ] Destination Buffer (after): [1, 2, 3, 4]
[INFO ] PASS: mem_to_mem verified.
[INFO ] Memset Buffer (after): [3735928559, 3735928559, 3735928559, 3735928559]
[INFO ] PASS: memset verified.
[INFO ] === All DMA tests complete ===
```

Both `mcxa5xx` and `mcxa2xx` still build (the DMA1 block is `#[cfg(feature = "mcxa5xx")]`-gated). Happy to open a PR against your branch if that's easier.

diondokter:
Ah well found! Yes please open a PR. If you do that within half an hour I can merge it all just before I go to bed 😇

embassy-mcxa/build.rs
bjackson312006: I am intending to fix this directly in nxp-pac (see [this PR](https://github.com/embassy-rs/nxp-pac/pull/110)) as part of #6435. The naming convention used in build.rs where `gate.bit == peripheral.name.to_lowercase()` is valid for most, but not all, peripherals. Among the latter are CAN0 and CAN1, which is what caused the build errors. The nxp-pac PR adds a `bit` member to the `Gate` struct and JSON schema that defaults to `peripheral.name.to_lowercase()` when unspecified (i.e., the vast majority of cases), but can be manually overwritten in the pac JSONs for outliers (e.g., CAN0/CAN1)

bramsdell-ms:
Done — opened diondokter/embassy#3 against your mcxa_dma1 branch with the fix. 🚀

embassy-mcxa/build.rs
diondokter: Ok, then let's fix it in the hal when that PR lands

diondokter:
Hmm fmt failed. Sorry, but I need to go to bed now. I'll fix it tomorrow. Or you need @felipebalbi to fix and merge it

bramsdell-ms:
Formatting slipped through — the imports weren't run through nightly rustfmt, which is why CI is red. One-line fix in diondokter/embassy#4 (reorders \DMA0\/\DMA1\ imports). \cargo +nightly fmt --check\ is clean with it. No rush if you're already off — it'll be there in the morning. 🌙

Issue #6526, opened 2026-07-16
`embassy-executor` hangs on `embassy-stm32` with `features = [ "stm32g031f8" ], default-features = false`


datdenkikniet:
It seems that something in `embassy-stm32` is not quite working as intended with `default-features = false`.

Perhaps this is intentional, but it's rather confusing.

To demonstrate: 
1. Go to `examples/stm32g0`, 
2. Set up `embassy-stm32` with `default-features = false` and `stm32g031f8` as the target MCU (perhaps also works with other g0 parts)
3. Run the `blinky` example
4. Observe that it hangs

xoviat:
The rt feature is required to run on hardware; it's disabled for some hosted tests.

datdenkikniet:
Is there something that can be done to inform the user? Is there a way for the executor (or HAL? I don't know which crate should be responsible for making sure the executor doesn't hang if `rt` isn't enabled) can detect that whatever it needs from the `rt` feature is missing? Even just documenting this would be good.  The way I discovered it was by comparing my application to that in the examples, but it still took a while because I did not even consider the fact that disabling something in the HAL would brick the executor in the way that it did.

It works for a little while and then (once the first executor interrupt occurs? I'm not sure) it hangs and does nothing, and it's pretty non-obvious that it's the `rt` feature that does the trick.

xoviat:
Just to clarify, it's possible to run embassy-stm32 on your PC with the `rt` feature disabled. Obviously, this isn't that useful because your PC has no stm32 peripherals, but it can be used to run tests in a virtual environment. The reason that it hangs as soon as an interrupt happens is that no-rt disables interrupts, since your PC does not allow access to the interrupt table (that's kernel mode) and even if it did, your PC doesn't have stm32 interrupts and any attempt to use them would crash your PC.

The first code that runs is init, so maybe we can put a waring there if defmt logs are enabled.

PR #6533, opened 2026-07-16
print warning if building embassy-stm32 for bare-metal arm but withou…

datdenkikniet:
…t rt

Not activating the `rt` feature will cause all interrupt handlers to be omitted, making anything that relies on interrupts hang/fail/do nothing.

I only applied this to `embassy-stm32` for now, because I'm not sure if this holds for other HAL crates too. 

Additionally, I'm not sure whether this should be a warning or just an error: are there use-cases where one wants to do this?

One solution/helpful hand towards resolving #6526

### DIFF
--- a/embassy-stm32/build.rs
+++ b/embassy-stm32/build.rs
@@ -42,6 +42,13 @@ fn main() {
     let mut cfgs = common::CfgSet::new();
     common::set_target_cfgs(&mut cfgs);
 
+    if std::env::var("CARGO_FEATURE_RT").is_err()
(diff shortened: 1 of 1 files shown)


datdenkikniet:
Thank you for the quick merge.

Sadly I discovered that this isn't as useful as I thought it would be: seems like the warnings are only printed in fairly specific circumstances: [ref](https://internals.rust-lang.org/t/how-fail-on-cargo-warning-warnings-from-build-rs/23590/9), so users of `embassy` outside of the repo won't actually see the warning... I will try to find a better way

PR #6537, opened 2026-07-17
example/stm32: add DAC DMA

RaresAndrei0:
DAC using DMA example tested on the stm32u545re board.

### DIFF
--- a/embassy-stm32/src/timer/low_level.rs
+++ b/embassy-stm32/src/timer/low_level.rs
@@ -690,6 +690,11 @@ impl<'d, T: BasicInstance> Timer<'d, T> {
     pub fn regs_basic(&self) -> crate::pac::timer::TimBasic {
         unsafe { crate::pac::timer::TimBasic::from_ptr(T::regs()) }
     }
+
+    /// Set Timer Master Mode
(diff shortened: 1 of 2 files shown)


xoviat:
is this still draft?

RaresAndrei0:
Yes, ran into an issue regarding the MMS update, I don't think using pac:: inside an example is intended, but set_master_mode does not work on my board...(I use the TIM6 timer and if I understood correctly, only some timers are being accessed using regs_gp16(), and the TIM6 is not one of them)

set_master_mode is using set_mms by accessing regs_gp16(), on my board it works by accessing regs_basic(), so i tried to create a new function inside timer/low_level.rs but for some reason i got the message "stm32u5 is disabled", I will continue to look for a way to update it, if it is possible.

RaresAndrei0:
Small update, as per the comments on the BasicInstance:
Note: This works even if the timer is more capable, because registers for the less capable timers are a subset. This allows writing a driver for a given set of capabilities, and having it transparently work with more capable timers.

Given this, I moved the set_master_mode fn from GeneralInstance4Channel to BasicInstance, and changed regs_gp16 to regs_basic, now it works but I need some feedback

PR #6538, opened 2026-07-17
stm32/adc: fix continuous adc on h5

pufmat:
Found and tested on STM32H523CE.

I also added `#[cfg(not(any(...)))]` to `w.set_cont(false);` to prevent setting conversion mode twice, but don't know on which ADCs `w.set_cont(false);` is actually correct. I think it should always be set based on `conversion_mode`.


xoviat:
The compiler can optimize out setting the coversion mode twice inside of a modify.

PR #6546, opened 2026-07-18
mspm0: new adc driver

i509VCB:
To be more similar to stm32.

Some new additions too which involved some larger changes to the driver.

### DIFF
--- a/embassy-mspm0/src/adc.rs
+++ b/embassy-mspm0/src/adc.rs
@@ -1,16 +1,23 @@
+//! Analog to Digital Converter (ADC)
+
 #![macro_use]
 
 use core::future::poll_fn;
+use core::hint::unreachable_unchecked;
 use core::marker::PhantomData;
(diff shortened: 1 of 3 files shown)


embassy-mspm0/src/adc.rs
```diff
@@ -80,314 +117,252 @@ pub enum Vrsel {
...
+            _mode: PhantomData,
         }
-        this
     }
-}
 
-impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
-    const SINGLE_CHANNEL: u8 = 0;
+    /// Read an ADC pin.
+    pub fn blocking_read<'a>(&mut self, channel: impl BorrowedChannel<'a, T>, conversion: Conversion) -> u16
```
i509VCB: One thing I should check again is why I need to clarify lifetimes here but stm32 has no such additional bounds.

PR #6553, opened 2026-07-20
nrf: nvmc: remove unsound usage of `core::slice::from_raw_parts`

datdenkikniet:
The docs for `raw_from_slice` explicitly state that the starting address must be non-0 ([ref](https://doc.rust-lang.org/beta/core/slice/fn.from_raw_parts.html#safety)).

In a bit of code I ran I just got a `HardFault` when trying to execute this `read` function. The example code seemed to Just Work even without the fix, but this is obviously unsound.

Replaced with `ptr::copy_nonoverlapping`, which doesn't have the non-0 requirement.


leftger (approved):
LGTM.

PR #6554, opened 2026-07-20
chore: Make most blinky examples a bit more uniform

kripton:
I was running the `blinky` example on a Raspberry Pi Pico (RP2040) and a Pico 2 (RP2350). I realized that the blinking frequency was really different. At first, I thought about a difference in CPU clocks but I thought it would be embarrassing for embassy to be this off regarding timing. Looking at the code, I saw that it was just that the examples were different regarding timing.
This PR makes most examples use the same interval (500ms off, 500ms on). I didn't touch the examples using `.toggle` (those could be changed to `on` and `off` to make the examples even more unified) and I also didn't touch the "low power" examples since it would defeat their purpose, demonstrating the CPU going to sleep.
I didn't test all the examples on real boards since I do not own all of them, I compile-tested most of them, though. Hoping for CI / GH actions to catch any problems I caused.

### DIFF
--- a/examples/mcxa2xx/src/bin/blinky.rs
+++ b/examples/mcxa2xx/src/bin/blinky.rs
@@ -20,17 +20,17 @@ async fn main(_spawner: Spawner) {
         defmt::info!("Toggle LEDs");
 
         red.toggle();
-        Timer::after_millis(250).await;
+        Timer::after_millis(500).await;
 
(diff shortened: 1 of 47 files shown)


leftger (approved):
This seems fine. It's a pretty large changeset so I'll defer to others to see if they want to merge it.

PR #6555, opened 2026-07-20
adc: power down ADC before configuring DIFSEL

Jacke-debug:
DIFSEL is only writable while ADEN=0 (RM0440 §21.7.21), but Adc::new() already enables the ADC before into_ring_buffered/ setup_injected_conversions run configure_sequence. Disable, write DIFSEL, re-enable, calibration is retained across ADEN=0.


xoviat:
See #6556; does that work?

PR #6556, opened 2026-07-20
stm32: fix adc/g4

xoviat:
closes #6555.


embassy-stm32/src/adc/mod.rs
```diff
@@ -153,6 +153,8 @@ trait AdcRegs: BasicAdcRegs {
     fn power_down(&self);
     fn wait_done(&self) -> bool;
     fn configure_dma(&self, conversion_mode: ConversionMode);
+    /// Configure the sequence. If the ADC is capable of differential channels,
+    /// this method must disable the ADC before configuring the sequence.
```
xoviat: *if required by hardware.

PR #6564, opened 2026-07-21
embassy-boot: Allow restarting of firmware updates

rgilton:
At present it's tricky to make `FirmwareUpdater` restart operations from offset 0 without re-instantiating it.  This patch adds a `reset` method that makes this easier.


lulf (approved):
Thanks!

PR #6565, opened 2026-07-21
stm32/i2s: cleanup mod

xoviat:
regs is already a pointer, so there's no need to store two indirections.

### DIFF
--- a/embassy-stm32/src/i2s.rs
+++ b/embassy-stm32/src/i2s.rs
@@ -2,8 +2,6 @@
 
 #![macro_use]
 
-use core::ptr;
-
 use embassy_futures::join::join;
 use stm32_metapac::spi::vals;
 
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/i2s.rs
```diff
@@ -614,32 +612,32 @@ impl<'d, W: Word> I2S<'d, W> {
 
     /// Write data directly to the raw I2S ringbuffer.
     /// This can be used to fill the buffer before starting the DMA transfer.
-    pub async fn write_immediate(&mut self, data: &[W]) -> Result<(usize, usize), Error> {
+    pub fn write_immediate(&mut self, data: &[W]) -> Result<(usize, usize), Error> {
```
xoviat, line 615: Also don't make this async.

PR #6567, opened 2026-07-21
stm32/usb: fix compiling without time enabled

pufmat:
Removes usb host dependency on `embassy_time`, and instead uses `wait` utility where applicable.

Quite important:
- These changes disable timeout in `read` and `write` when time feature is not enabled. I'm not sure if this is acceptable, but I also don't know how it can be implemented without time feature.
- I haven't tested it because I do not have a board that supports usb host (without hardware modification). While the changes are small, I would be glad if someone could test it.

So if these changes are not right, please treat this PR as an issue instead.

### DIFF
--- a/embassy-stm32/src/usb/usb_host.rs
+++ b/embassy-stm32/src/usb/usb_host.rs
@@ -6,7 +6,6 @@ use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use core::task::Poll;
 
 use embassy_sync::waitqueue::AtomicWaker;
-use embassy_time::{Duration, Instant, Timer};
 use embassy_usb_driver::host::{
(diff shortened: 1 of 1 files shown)


TomDeRybel:
Hello @pufmat,

I was looking at the changes and have a question: to my eye, all the new wait() delays look 10x longer than the comments suggest?

Eg: `block_for_us(100_0000); // 100 ms` --> To my naive eye, this reads as 1000 ms?

I did a little test, as a sanity check, and get this:

`
info!("Number 1: {}", 100_000);
`
`
info!("Number 2: {}", 100_0000);
`

`
0.046703 [INFO ] Number 1: 100000
`
`
0.046706 [INFO ] Number 2: 1000000
`

I'm likely missing something?

Best regards,
Tom

pufmat:
You are right, I accidentally typed one zero too much, sorry for that.

PR #6568, opened 2026-07-21
stm32/adc: add h5 injected adc and trigger support

pufmat:
Tested on STM32H523CE.

### DIFF
--- a/embassy-stm32/src/adc/v3.rs
+++ b/embassy-stm32/src/adc/v3.rs
@@ -13,9 +13,12 @@ use pac::adccommon::vals::Presc;
 
 #[allow(unused_imports)]
 use crate::adc::SealedAdcChannel;
-use crate::adc::{Adc, Averaging, ConversionMode, Instance, Resolution, SampleTime, Temperature, Vbat, VrefInt};
+use crate::adc::{Adc, Averaging, ConversionMode, InjectedRegs, Instance, Resolution, SampleTime, Temperature, Vbat, VrefInt, Exten};
 use crate::wait::block_for_us;
 use crate::{Peri, pac, rcc};
+
+mod injected;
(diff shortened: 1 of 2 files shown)


embassy-stm32/src/adc/v3.rs
```diff
@@ -428,6 +435,141 @@ impl super::AdcRegs for crate::pac::adc::Adc {
     }
 }
 
+#[cfg(adc_h5)]
+impl InjectedRegs for crate::pac::adc::Adc {
+    fn configure_injected_sequence(&self, sequence: impl ExactSizeIterator<Item = ((u8, bool), Self::SampleTime)>) {
+        use crate::pac::adc::regs::Jsqr;
+
+        let mut difsel = 0u32;
```
xoviat, line 467: Do we need to set aden = false here?

pufmat: Yes, according to RM0481 due to DIFSEL:
> The software is allowed to write these bits only when the ADC is disabled (ADCAL = 0,
JADSTART = 0, JADSTP = 0, ADSTART = 0, ADSTP = 0, ADDIS = 0 and ADEN = 0).

But I saw other PR that fixes it, so I didn't disable ADC here.

xoviat: #6555 is the wrong fix because only the adcs that support difsel need to be powered down first; and power down is meant to be called from the destructor. I have submitted #6556 instead, which is the approach that I think we should take.

pufmat: Okay.

pufmat:
Note: In order to test injected ADC merge locally with #6556, otherwise it won't work as ADC stays disabled.

PR #6569, opened 2026-07-21
fix(stm32/qei): support full counting range on 32-bit timers

Jacke-debug:
Config::auto_reload and Qei::count()/reset() previously used the 16-bit register view unconditionally, so on 32-bit timers the ARR and counter were truncated to u16 and reset() only cleared the lower 16 bits, leaving the upper bits stale.

Make Config and AdvancedConfig generic over the timer instance and use T::Word for auto_reload/count(), and add
low_level::Timer::get_counter() to read the counter as T::Word.

### DIFF
--- a/embassy-stm32/src/timer/qei.rs
+++ b/embassy-stm32/src/timer/qei.rs
@@ -6,13 +6,17 @@ use super::low_level::Timer;
 pub use super::{Ch1, Ch2};
 use super::{GeneralInstance4Channel, TimerPin};
 use crate::Peri;
+use crate::dma::word::Word;
 use crate::gpio::{AfType, Flex, Pull};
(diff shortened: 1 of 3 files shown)


embassy-stm32/src/timer/qei.rs
```diff
@@ -208,13 +220,13 @@ impl<'d, T: GeneralInstance4Channel> Qei<'d, T> {
     }
 
     /// Get count.
-    pub fn count(&self) -> u16 {
-        self.inner.regs_gp16().cnt().read().cnt()
+    pub fn count(&self) -> T::Word {
+        self.inner.get_counter()
```
xoviat: This should return a u32.

PR #6574, opened 2026-07-22
embassy-usb: Clippy fixes

GrantM11235:
This is mostly non-function changes, except

- Changing a few instances of `#[repr(packed)]` to `#[repr(C, packed)]`, which is necessary for correctness.
- Making `AudioSource::new` and `Speaker::new` return `Self` with the parts inside instead of a tuple, which is an ergonomic improvement IMO.

Note that there is still a clippy warning, but it is a [false positive](https://github.com/rust-lang/rust-clippy/issues/17373). Coincidentally, it looks like the fix was merged less than two days ago.

<details>

<summary>False positive</summary>

```
warning: this `if` has identical blocks
   --> src/class/uac1/source.rs:686:39
    |
686 |               if alternate_setting == 0 {
    |  _______________________________________^
687 | |                 info!(
688 | |                     "AudioSourceControlHandler: Audio streaming interface({}) deactivated!",
689 | |                     iface
690 | |                 );
691 | |             } else if alternate_setting == 1 {
    | |_____________^
    |
note: same as this
   --> src/class/uac1/source.rs:691:46
    |
691 |               } else if alternate_setting == 1 {
    |  ______________________________________________^
692 | |                 info!(
693 | |                     "AudioSourceControlHandler: Audio streaming interface({}) activated!",
694 | |                     iface
695 | |                 );
696 | |             }
    | |_____________^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#if_same_then_else
    = note: `#[warn(clippy::if_same_then_else)]` on by default
```

</details>

### DIFF
--- a/embassy-usb/src/lib.rs
+++ b/embassy-usb/src/lib.rs
@@ -239,7 +239,6 @@ struct Inner<'d, D: Driver<'d>> {
 }
 
 impl<'d, D: Driver<'d>> UsbDevice<'d, D> {
-    #[allow(clippy::too_many_arguments)]
     pub(crate) fn build(
         driver: D,
         config: Config<'d>,
--- a/embassy-usb/src/lib.rs
+++ b/embassy-usb/src/lib.rs
@@ -239,6 +239,7 @@ struct Inner<'d, D: Driver<'d>> {
 }
 
 impl<'d, D: Driver<'d>> UsbDevice<'d, D> {
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn build(
         driver: D,
         config: Config<'d>,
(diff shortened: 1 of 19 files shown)


embassy-usb/src/lib.rs
```diff
@@ -239,6 +239,7 @@ struct Inner<'d, D: Driver<'d>> {
 }
 
 impl<'d, D: Driver<'d>> UsbDevice<'d, D> {
+    #[allow(clippy::too_many_arguments)]
```
bugadani: Either don't add this, or add a reason why this is a good thing. My preference is keeping the warning.

GrantM11235: Okay, I reverted that. I was able to avoid the warning for `UsbDevice::build` by inlining the method, which I think is cleaner anyway. I don't know what to do about `InterfaceAltBuilder::endpoint_out` and `endpoint_in`, so I just left them with the warning for now.

bugadani (approved):

PR #6578, opened 2026-07-22
stm32/adc: add async read for injected adc

pufmat:
Implements interrupt driven read function for injected ADC convention which waits for end of sequence. This way reading injected convention no longer requires manually implementing interrupt handler, but is still possible if needed.

I had to duplicate some code because otherwise I would have to implement ADC interrupt handler for more STMs, but I don't want to do that until I get some feedback on the actual feature. In case this PR is ok, it can be merged, and I will implement interrupt handlers in a separate PR or I can add them in this PR.

Tested on STM32H523CE.


pufmat:
Also tested on STM32F411CE.

I haven't tested STM32G4 part as I do not have any board with it.

PR #6580, opened 2026-07-22
Implement deep sleep in embassy-mspm0

CordlessCoder:
I'm generally following the approach embassy-stm32 took where it makes sense - the testing is done on a custom MSPM0L1305 board since that's what I have access to.

I used an LLM to assist me in navigating the TRMs for all the MSPM0 product families, to review the code as I write it and for first drafts of some code and documentation.

### DIFF
--- a/embassy-mspm0/src/adc.rs
+++ b/embassy-mspm0/src/adc.rs
@@ -211,7 +211,7 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
         });
 
         r.ctl1().modify(|w| {
-            w.set_sc(vals::Sc::START);
+            w.set_sc(vals::Sc::Start);
         });
 
         // Wait for conversion
@@ -297,7 +297,7 @@ impl<'d, T: Instance> Adc<'d, T, Async> {
         });
 
         r.ctl1().modify(|w| {
-            w.set_sc(vals::Sc::START);
+            w.set_sc(vals::Sc::Start);
         });
 
         Self::wait_for_conversion().await;
@@ -344,7 +344,7 @@ impl<'d, T: Instance> Adc<'d, T, Async> {
         });
 
         r.ctl1().modify(|w| {
-            w.set_sc(vals::Sc::START);
+            w.set_sc(vals::Sc::Start);
         });
 
         Self::wait_for_conversion().await;
@@ -432,20 +432,20 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
         r.gprcm(0).rstctl().write(|w| {
             w.set_resetstkyclr(true);
             w.set_resetassert(true);
-            w.set_key(vals::ResetKey::KEY);
+            w.set_key(vals::ResetKey::Key);
         });
 
         r.gprcm(0).pwren().modify(|reg| {
             reg.set_enable(true);
-            reg.set_key(vals::PwrenKey::KEY);
+            reg.set_key(vals::PwrenKey::Key);
         });
 
         // Wait for power up
         cortex_m::asm::delay(16);
 
         r.gprcm(0).clkcfg().write(|w| {
-            w.set_key(vals::ClkcfgKey::KEY);
-            w.set_sampclk(vals::Sampclk::SYSOSC);
+            w.set_key(vals::ClkcfgKey::Key);
+            w.set_sampclk(vals::Sampclk::Sysosc);
         });
 
         // FIXME: Consider clock config
@@ -453,20 +453,20 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
         r.ctl0().write(|w| {
             w.set_enc(false);
             // TODO: power down config
-            w.set_pwrdn(vals::Pwrdn::MANUAL);
-            w.set_sclkdiv(vals::Sclkdiv::DIV_BY_4);
+            w.set_pwrdn(vals::Pwrdn::Manual);
+            w.set_sclkdiv(vals::Sclkdiv::DivBy4);
         });
 
         r.clkfreq().write(|w| {
-            w.set_frange(vals::Frange::RANGE24TO32);
+            w.set_frange(vals::Frange::Range24to32);
         });
 
         r.ctl1().write(|w| {
-            w.set_trigsrc(vals::Trigsrc::SOFTWARE);
-            w.set_sc(vals::Sc::STOP);
-            w.set_conseq(vals::Conseq::SEQUENCE);
-            w.set_sampmode(vals::Sampmode::AUTO);
-            w.set_avgn(vals::Avgn::DISABLE);
+            w.set_trigsrc(vals::Trigsrc::Software);
+            w.set_sc(vals::Sc::Stop);
+            w.set_conseq(vals::Conseq::Sequence);
+            w.set_sampmode(vals::Sampmode::Auto);
+            w.set_avgn(vals::Avgn::Disable);
             w.set_avgd(0);
         });
 
@@ -477,7 +477,7 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
             w.set_rstsampcapen(false);
             w.set_dmaen(false);
             w.set_fifoen(false);
-            w.set_sampcnt(vals::Sampcnt::MIN);
+            w.set_sampcnt(vals::Sampcnt::Min);
             w.set_startadd(0);
             w.set_endadd(0);
         });
@@ -510,7 +510,7 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
                 // TODO: More parameters
                 w.set_avgen(false);
                 w.set_bcsen(false);
-                w.set_trig(vals::Trig::AUTO_NEXT);
+                w.set_trig(vals::Trig::AutoNext);
                 w.set_wincomp(false);
             });
         }
@@ -576,17 +576,17 @@ trait SealedBorrowedChannel<'a, T> {
 
 const fn to_res(resolution: Resolution) -> vals::Res {
     match resolution {
-        Resolution::Bits12 => vals::Res::BIT_12,
-        Resolution::Bits10 => vals::Res::BIT_10,
-        Resolution::Bits8 => vals::Res::BIT_8,
+        Resolution::Bits12 => vals::Res::Bit12,
+        Resolution::Bits10 => vals::Res::Bit10,
+        Resolution::Bits8 => vals::Res::Bit8,
     }
 }
 
 const fn from_res(res: vals::Res) -> Resolution {
     match res {
-        vals::Res::BIT_12 => Resolution::Bits12,
-        vals::Res::BIT_10 => Resolution::Bits10,
-        vals::Res::BIT_8 => Resolution::Bits8,
+        vals::Res::Bit12 => Resolution::Bits12,
+        vals::Res::Bit10 => Resolution::Bits10,
+        vals::Res::Bit8 => Resolution::Bits8,
         // SAFETY: The HAL will never program an invalid valid.
         vals::Res::_RESERVED_3 => unsafe { unreachable_unchecked() },
     }
@@ -594,8 +594,8 @@ const fn from_res(res: vals::Res) -> Resolution {
 
 const fn convert_stime(stime: SampleTimeComparator) -> vals::Stime {
     match stime {
-        SampleTimeComparator::Scomp0 => vals::Stime::SEL_SCOMP0,
-        SampleTimeComparator::Scomp1 => vals::Stime::SEL_SCOMP1,
+        SampleTimeComparator::Scomp0 => vals::Stime::SelScomp0,
+        SampleTimeComparator::Scomp1 => vals::Stime::SelScomp1,
     }
 }
 
--- a/embassy-mspm0/src/adc.rs
+++ b/embassy-mspm0/src/adc.rs
@@ -525,15 +525,13 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
     #[inline]
     fn wait_for_conversion() -> impl Future<Output = ()> {
         let r = T::info().regs;
-        let state = T::state();
 
         poll_fn(move |cx| {
-            state.waker.register(cx.waker());
-
-            if !r.ctl0().read().enc() {
-                Poll::Ready(())
-            } else {
+            if r.ctl0().read().enc() {
+                cx.waker().wake_by_ref();
                 Poll::Pending
+            } else {
+                Poll::Ready(())
             }
         })
     }
--- a/embassy-mspm0/src/executor.rs
+++ b/embassy-mspm0/src/executor.rs
@@ -1,0 +1,101 @@
+//! MSPM0-specific `embassy-executor` platform.
+//!
+//! This module provides an `embassy-executor` platform specific for MSPM0 chips that integrates [`low_power::sleep()`](crate::low_power::sleep) in the main loop.
+//! Read the `embassy-executor` README for information about what "executor platforms" are and how they work.
+//!
+//! To use it:
+//! - Enable the `executor-thread` feature on this crate.
+//! - **Do not** enable features `platform-cortex-m`, `executor-thread` or `executor-interrupt` in the `embassy-executor` crate.
+//! - Tell the `main` macro to use this executor like this:
+//!
+//! ```rust,no_run
+//! #[embassy_executor::main(executor = "embassy_mspm0::executor::Executor", entry = "cortex_m_rt::entry")]
+//! async fn main(spawner: Spawner) {
+//!     let p = embassy_mspm0::init(Config::default());
+//!     // ...
+//! }
+//! ```
+
+#[unsafe(export_name = "__pender")]
+#[cfg(feature = "executor-thread")]
+fn __pender(_context: *mut ()) {
+    thread::SIGNAL_WORK_THREAD_MODE.store(true, core::sync::atomic::Ordering::SeqCst);
+}
+
+#[cfg(feature = "executor-thread")]
+pub use thread::*;
+#[cfg(feature = "executor-thread")]
+mod thread {
+    use core::marker::PhantomData;
+    use core::sync::atomic::{AtomicBool, Ordering};
+
+    use embassy_executor::{Spawner, raw};
+
+    const THREAD_PENDER: usize = usize::MAX;
+
+    /// Set by the pender to signal pending work; checked before sleeping since `WFI` ignores `SEV`.
+    pub(crate) static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
+
+    /// Thread-mode executor that deep-sleeps on idle via [`low_power::sleep`](crate::low_power::sleep).
+    ///
+    /// This is the simplest and most common kind of executor. It runs on
+    /// thread mode (at the lowest priority level), and uses the `WFE` ARM instruction
+    /// to sleep when it has no more work to do. When a task is woken, a `SEV` instruction
+    /// is executed, to make the `WFE` exit from sleep and poll the task.
+    pub struct Executor {
+        inner: raw::Executor,
+        not_send: PhantomData<*mut ()>,
+    }
+
+    impl Executor {
+        /// Create a new Executor.
+        pub fn new() -> Self {
+            Self {
+                inner: raw::Executor::new(THREAD_PENDER as *mut ()),
+                not_send: PhantomData,
+            }
+        }
+
+        /// Run the executor.
+        ///
+        /// The `init` closure is called with a [`Spawner`] that spawns tasks on
+        /// this executor. Use it to spawn the initial task(s). After `init` returns,
+        /// the executor starts running the tasks.
+        ///
+        /// To spawn more tasks later, you may keep copies of the [`Spawner`] (it is `Copy`),
+        /// for example by passing it as an argument to the initial tasks.
+        ///
+        /// This function requires `&'static mut self`. This means you have to store the
+        /// Executor instance in a place where it'll live forever and grants you mutable
+        /// access. There's a few ways to do this:
+        ///
+        /// - a [StaticCell](https://docs.rs/static_cell/latest/static_cell/) (safe)
+        /// - a `static mut` (unsafe)
+        /// - a local variable in a function you know never returns (like `fn main() -> !`), upgrading its lifetime with `transmute`. (unsafe)
+        ///
+        /// This function never returns.
+        pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+            init(self.inner.spawner());
+
+            loop {
+                unsafe {
+                    self.inner.poll();
+
+                    critical_section::with(|cs| {
+                        if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
+                            SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
+                        } else {
+                            crate::low_power::sleep(cs);
+                        }
+                    });
+                }
+            }
+        }
+    }
+
+    impl Default for Executor {
+        fn default() -> Self {
+            Self::new()
+        }
+    }
+}
--- a/embassy-mspm0/src/executor.rs
+++ b/embassy-mspm0/src/executor.rs
@@ -4,7 +4,7 @@
 //! Read the `embassy-executor` README for information about what "executor platforms" are and how they work.
 //!
 //! To use it:
-//! - Enable the `executor-thread` feature on this crate.
+//! - Enable the `executor-thread` and/or `executor-interrupt` feature on this crate.
 //! - **Do not** enable features `platform-cortex-m`, `executor-thread` or `executor-interrupt` in the `embassy-executor` crate.
 //! - Tell the `main` macro to use this executor like this:
 //!
@@ -17,9 +17,36 @@
 //! ```
 
 #[unsafe(export_name = "__pender")]
-#[cfg(feature = "executor-thread")]
-fn __pender(_context: *mut ()) {
-    thread::SIGNAL_WORK_THREAD_MODE.store(true, core::sync::atomic::Ordering::SeqCst);
+#[cfg(any(feature = "executor-thread", feature = "executor-interrupt"))]
+fn __pender(context: *mut ()) {
+    // `context` is either `THREAD_PENDER`, or an interrupt number passed to `InterruptExecutor::start`.
+    let context = context as usize;
+
+    #[cfg(feature = "executor-thread")]
+    // Try to optimize away the branch when only thread mode is enabled.
+    if !cfg!(feature = "executor-interrupt") || context == thread::THREAD_PENDER {
+        thread::SIGNAL_WORK_THREAD_MODE.store(true, core::sync::atomic::Ordering::SeqCst);
+        return;
+    }
+
+    #[cfg(feature = "executor-interrupt")]
+    {
+        use cortex_m::interrupt::InterruptNumber;
+        use cortex_m::peripheral::NVIC;
+
+        #[derive(Clone, Copy)]
+        struct Irq(u16);
+
+        // SAFETY: `context` was an `InterruptNumber` when passed to `InterruptExecutor::start`.
+        unsafe impl InterruptNumber for Irq {
+            fn number(self) -> u16 {
+                self.0
+            }
+        }
+
+        // MSPM0 is Cortex-M0+, which has no STIR.
+        NVIC::pend(Irq(context as u16));
+    }
 }
 
 #[cfg(feature = "executor-thread")]
@@ -31,7 +58,7 @@ mod thread {
 
     use embassy_executor::{Spawner, raw};
 
-    const THREAD_PENDER: usize = usize::MAX;
+    pub(super) const THREAD_PENDER: usize = usize::MAX;
 
     /// Set by the pender to signal pending work; checked before sleeping since `WFI` ignores `SEV`.
     pub(crate) static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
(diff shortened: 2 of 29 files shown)


CordlessCoder:
I believe I fixed the race condition that caused the time driver to jump backwards when driven by LFCLK.

CordlessCoder:
To reproduce the timer jump bug I used this snippet after adjusting TICK_HZ and switching the clock source from MFCLK to LFCLK.
```rs
let mut last = Instant::now();
loop {
    let n = Instant::now();
    defmt::assert!(n >= last, "time went backwards: {}us -> {}us", last.as_micros(), n.as_micros());
    last = n;
}
```
It reliably panicked after 2 seconds(a single overflow cycle).
After the changes I left it running for 15 minutes without getting a panic.

This was tested on the MSPM0L1305RHB, but I'm quite confident about the fix translating to other MCUs.

CordlessCoder:
I'm measuring current consumption in sleep/standby that matches the datasheet on my MSPM0L1305, so the sleep modes work.
I'll start working on integrating sleep into the executor.

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -277,13 +286,53 @@ impl Driver for TimxDriver {
...
+                return calc_now(period, counter);
+            }
+
+            // The counter and period are incoherent.
+            let ris = regs.cpu_int(0).ris().read();
+            if ris.l() || ris.ccu0() {
+                // A period event is latched but the interrupt hasn't run yet, so period is stale
+                // and counter is fresh. `calc_now`'s XOR handles this.
+                // NOTE: You must NOT wait for the interrupt here: now() is called inside critical
+                // sections, where it would deadlock.
```
i509VCB: ~~This is an invalid assumption. `Instant::now()` could be called anywhere at any time.~~

Misread

i509VCB:
The assembly output of now() is quite a bit of code, I feel there is some room for optimization:

```
.section .text._embassy_time_now,"ax",%progbits
        .globl  _embassy_time_now
        .p2align        2
        .type   _embassy_time_now,%function
        .code   16
        .thumb_func
_embassy_time_now:
        .fnstart
        .cfi_startproc
        .save   {r4, r5, r7, lr}
        push {r4, r5, r7, lr}
        .cfi_def_cfa_offset 16
        .cfi_offset lr, -4
        .cfi_offset r7, -8
        .cfi_offset r5, -12
        .cfi_offset r4, -16
        .setfp  r7, sp, #8
        add r7, sp, #8
        .cfi_def_cfa r7, 8
        ldr r1, .LCPI141_0
        ldr r2, .LCPI141_1
        b .LBB141_2
.LBB141_1:
        @MEMBARRIER
        ldr r4, [r2, #16]
        cmp r0, r4
        beq .LBB141_7
.LBB141_2:
        ldr r0, [r2, #16]
        @MEMBARRIER
        ldr r3, [r1]
        uxth r4, r3
.LBB141_3:
        ldr r3, [r1]
        uxth r5, r3
        cmp r4, r5
        mov r3, r4
        beq .LBB141_1
        ldr r3, [r1]
        uxth r4, r3
        cmp r5, r4
        mov r3, r5
        beq .LBB141_1
        ldr r3, [r1]
        uxth r5, r3
        cmp r4, r5
        mov r3, r4
        beq .LBB141_1
        ldr r3, [r1]
        uxth r4, r3
        cmp r5, r4
        mov r3, r5
        bne .LBB141_3
        b .LBB141_1
.LBB141_7:
        movs r2, #1
        ands r2, r0
        lsrs r1, r3, #15
        cmp r2, r1
        bne .LBB141_9
        lsls r2, r1, #15
        eors r2, r3
        lsrs r3, r0, #17
        lsls r0, r0, #15
        movs r1, #0
        b .LBB141_11
.LBB141_9:
        lsrs r1, r0, #17
        lsls r0, r0, #15
        movs r4, #129
        lsls r4, r4, #1
        ldr r5, .LCPI141_2
        ldr r5, [r5]
        tst r5, r4
        beq .LBB141_12
        lsls r2, r2, #15
        eors r2, r3
        movs r3, #0
.LBB141_11:
        adds r0, r0, r2
        adcs r1, r3
.LBB141_12:
        pop {r4, r5, r7, pc}
        .p2align        2
.LCPI141_0:
        .long   1074288640
.LCPI141_1:
        .long   embassy_mspm0::time_driver::driver::DRIVER
.LCPI141_2:
        .long   1074286640
```

embassy-mspm0/src/time_driver/tim.rs
CordlessCoder: I'll make the comment clearer - what I mean here is that it *can* be called from a critical section, so interrupts *may* be blocked. The function itself doesn't depend on interrupts being disabled, it's tolerant to that being the case.

CordlessCoder:
I suspect a lot of the bloat is the "torn read protection" loops - I'll double-check if TI's SDK actually does something similar. If it doesn't - they can be plain reads and that gets rid of the loops.

Edit: Removing them seems to break things :(

i509VCB:
I would like to split the timer changes into a different pull request, along with the rhb packages.

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -17,6 +17,16 @@ use crate::{peripherals, tim};
 #[cfg(any(time_driver_timg12, time_driver_timg13))]
 compile_error!("TIMG12 and TIMG13 are not supported by the time driver yet");
 
+// Only TIMG0 and TIMG1 remain clocked in STANDBY, so they are the only timers that can wake the
+// core from deep sleep via the time driver. Reject a `low-power` build on any other timer.
+// TODO: Or maybe allow using them, but disable STANDBY? STOP0 still work.
```
i509VCB: I do think we should allow timers which cannot run in STANDBY by leaking an appropriate wake guard in time driver init.

Would also be worth documenting that specific timers for time driver use would not allow certain sleep modes.

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -278,12 +288,29 @@ impl Driver for TimxDriver {
...
 
         let period = self.period.load(Ordering::Relaxed);
-        // Ensure the compiler does not read the counter before the period.
-        compiler_fence(Ordering::Acquire);
-
         let counter = regs.counterregs(0).ctr().read() as u16;
 
-        calc_now(period, counter)
+        // When the counter's top bit equals period parity, the two are coherent.
+        if (((counter >> 15) as u32) & 1) == (period & 1) {
```
i509VCB: I assume this is the likely branch?

If so I would like to tell the compiler that the alternative is a cold path (so put it before the read of RIS). Would need Rust 1.95 for `core::hint::cold_path();` which is being done in https://github.com/embassy-rs/embassy/pull/6610

For now

i509VCB:
Okay the timers were fixed in #6613

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+/// sources are a wake-capable IO event, NRST, or SWD activity. Does not return as the wake results
+/// in a reset.
+/// You can respond to the reset on boot using [`ResetCause::BorWakeFromShutdown`](crate::ResetCause).
+///
+/// # Safety
+/// This is irreversible in place: CPU and peripheral state are lost and this call never returns.
+//
+// From the TRM: SYSCTL "Operating Modes": set `PMODECFG.DSLEEP = SHUTDOWN`, arm `SLEEPDEEP`,
+// then `WFI`. This is identical across every MSPM0 family.
+pub unsafe fn shutdown(_cs: CriticalSection) -> ! {
```
i509VCB: Does this need to be unsafe?

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+    // Never dropped — the chip resets rather than resuming here.
+    let _prefetch = PrefetchSuspend::new();
+
+    let mut scb = unsafe { cortex_m::Peripherals::steal() }.SCB;
+    scb.set_sleepdeep();
+    cortex_m::asm::dsb();
+
+    loop {
+        cortex_m::asm::wfi();
+    }
```
i509VCB: I think we should do the following:

```rust
cortex_m::asm::wfi();

// SAFETY: Setting DSLEEP to SHUTDOWN means WFI will never return.
unsafe { unreachable_unchecked() }
```

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -171,14 +194,14 @@ impl TimxDriver {
             w.set_ccu0(true);
         });
 
-        // Allow the counter to start counting.
+        // Allow the counter to start counting (from the preset value 1, kept by CVAE = NOCHANGE).
```
i509VCB: Unneeded documentation change. Above explains it.

embassy-mspm0/src/time_driver/tim.rs
i509VCB: Why do we need this clear if we start the counter at 1?

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -159,6 +173,15 @@ impl TimxDriver {
         regs.counterregs(0).cc(0).write_value(0x8000 as u32);
         regs.counterregs(0).load().write_value(u16::MAX as u32);
 
+        // Preset the counter to 1 (kept on enable by CVAE = NOCHANGE above) so it never sits at 0 at
+        // boot. Enabling with CVAE = ZEROVAL would zero the counter, and while it sits at 0 (LFCLK
+        // has not ticked yet) the zero (Z) event stays asserted; the handler, armed below, would then
+        // fire on Z once or twice before the counter leaves 0 and count that many spurious periods,
+        // leaving the timebase ~2 s ahead and cutting the first sleep short. (Only reproduces on
+        // bare-metal boots -- a debugger's startup latency lets the counter move off 0 first, hiding
+        // it.) Starting at 1 means the counter is 0 only on real overflows, so Z only fires then.
```
i509VCB: This documentation is excessive to explain starting 1 tick late so that we don't immediately increment time since boot by 1 second (instead approx 30 microseconds).

I'm sure you didn't write this due to the presence of the dashes and excessive parenthesis.

embassy-mspm0/src/uart/mod.rs
```diff
@@ -990,6 +1029,18 @@ fn set_baudrate_inner(regs: Regs, clock: u32, baudrate: u32) -> Result<(), Confi
...
+// compute the divisors at compile time and keep the division out of their binary. On this target the
+// baud maths is the *only* thing pulling in the 64-bit division helpers — a UART build links ~94 bytes
+// of `__udivmoddi4`/`__aeabi_uldivmod` plus ~110 bytes of `fixed` monomorphisation that a timer-only
+// build does not. The larger win is that an unrepresentable clock/baud pair becomes a compile error.
+//
+// Feasible on stable: `fixed` 1.30 already has const `checked_div`/`checked_mul`/`checked_add`/`frac`/
+// `from_bits`/`to_bits`. What has to change: `from_num`/`checked_to_num` -> `from_bits`/`to_bits` with
+// explicit shifts and range checks; the `/` for `half` -> `from_bits(32)`; the `for` loops -> `while`
+// with indexing; `?` -> `match`; `trace!` out of the pure path; and the `x3_invalid` register read
+// becomes a parameter.
```
i509VCB: Unneeded, did AI write this?

Unfortunately with runtime reclocking being possible this does not seem possible to truly remove. The ability to precompute some of these things is a larger discussion worth its own issue since it would also be applicable to other HALs.

embassy-mspm0/src/adc.rs
```diff
@@ -525,15 +525,23 @@ impl<'d, T: Instance, M: Mode> Adc<'d, T, M> {
     #[inline]
     fn wait_for_conversion() -> impl Future<Output = ()> {
         let r = T::info().regs;
-        let state = T::state();
 
+        // Busy-poll ENC by self-waking; never let the executor deep-sleep while a software-triggered
```
i509VCB: This will be better in the future when we have more options for clock source for the ADC (ULPCLK or HFCLK).

What we should do is set up an appropriate wake guard when doing an async read if our clock source happens to be SYSOSC which needs to be kept alive. 


Otherwise the system will never truly go to a lower sleep mode if possible because we will effectively be doing a `yield_now` in a loop.

embassy-mspm0/src/executor.rs
```diff
@@ -0,0 +1,83 @@
...
+    use core::sync::atomic::{AtomicBool, Ordering};
+
+    use embassy_executor::{Spawner, raw};
+
+    const THREAD_PENDER: usize = usize::MAX;
+
+    /// Set by the pender to signal pending work; checked before sleeping since `WFI` ignores `SEV`.
+    pub(crate) static SIGNAL_WORK_THREAD_MODE: AtomicBool = AtomicBool::new(false);
+
+    /// Thread-mode executor that deep-sleeps on idle via [`low_power::sleep`](crate::low_power::sleep).
```
i509VCB, line 66: Copy the main docs for the executor, and add a mention that this executor will manage entry/exit of low power modes. Do not need to include the last comment about extra steps needing raw::Executor to program custom behavior.

embassy-mspm0/src/executor.rs
```diff
@@ -0,0 +1,83 @@
...
+                inner: raw::Executor::new(THREAD_PENDER as *mut ()),
+                not_send: PhantomData,
+            }
+        }
+
+        /// Run the executor, calling `init` with a [`Spawner`] for the initial task(s).
+        ///
+        /// Never returns. See [`embassy_executor`]'s thread executor for how to obtain the
+        /// `&'static mut self` this requires.
+        pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
```
i509VCB: Copy the whole documentation so that the users is not confused about 2 unique executor types.

embassy-mspm0/src/trng.rs
```diff
@@ -447,6 +435,12 @@ impl TrngInner<'_> {
 
     #[cfg(feature = "rt")]
     async fn async_read_u32(&mut self) -> Result<u32, Error> {
+        // The TRNG is clocked from MCLK, which only runs in RUN/SLEEP, so a read forbids all deep
+        // sleep while it is in flight (floor Stop0 → the executor idles in SLEEP, where MCLK and the
+        // TRNG stay alive and the ready IRQ wakes it). Operation-scoped: the TRNG is read on demand,
+        // not a background service, so this is dropped when the read completes.
+        let _guard = SleepLevel::floor_for_clock_hz(mclk_frequency()).map(WakeGuard::new);
+
```
i509VCB: I think this could be written more simply.

```
TRNG is clocked from MCLK which only runs in RUN/SLEEP. If sampling from the TRNG make sure modes below SLEEP are not entered. 
```

embassy-mspm0/src/lib.rs
```diff
@@ -194,6 +198,9 @@ pub fn init(config: Config) -> Peripherals {
             w.set_mfpclken(true);
         });
 
+        // BOR0. The alternate thresholds do not reset the device correctly in STANDBY (errata
+        // PMCU_ERR_03), and nothing here raises the level afterwards, so deep sleep needs no
+        // save/restore around it. A future API for the higher thresholds would have to add one.
```
i509VCB: This should just be a TODO mentioning the errata and that in the future when BOR is configurable that we need to deal with it before going to sleep.

embassy-mspm0/src/gpio.rs
```diff
@@ -1050,14 +1160,36 @@ fn irq_handler(gpio: gpio::Gpio, port: Port) {
 
     let bits = gpio.cpu_int().mis().read().0;
 
+    // One snapshot for the whole batch, so every pin is classified against the same instant.
```
i509VCB: AI speak

embassy-mspm0/src/gpio.rs
```diff
@@ -1050,14 +1160,36 @@ fn irq_handler(gpio: gpio::Gpio, port: Port) {
...
         let id = ((port as u8) * 32) + i as u8;
-        let _ = GPIO_WAIT_MAP.wake(&id, ());
-
-        // Notify the future that an edge event has occurred by masking the interrupt for this pin.
-        gpio.cpu_int().imask().modify(|w| {
-            w.set_dio(i as usize, false);
-        });
+        let (rise, fall, mask) = want_masks(id);
+
+        // The status bit does not say which edge fired, so the level standing now is the evidence.
```
i509VCB: This is extra wording for no reason.

Something along the lines of "GPIO_ERR_01 requires that we detect level changes to determine if an edge event happened."

embassy-mspm0/src/gpio.rs
```diff
@@ -1050,14 +1160,36 @@ fn irq_handler(gpio: gpio::Gpio, port: Port) {
...
+        let fired = if level.dio(i as usize) { rise } else { fall };
+
+        if fired.load(Ordering::Relaxed) & mask != 0 {
+            // Withdrawing both requests is what tells the waiting future its edge arrived.
```
i509VCB: "Clear the mask for this pin since the edge arrived."

embassy-mspm0/src/gpio.rs
```diff
@@ -1050,14 +1160,36 @@ fn irq_handler(gpio: gpio::Gpio, port: Port) {
...
+
+            // Notify the future that an edge event has occurred by masking the interrupt for this pin.
+            gpio.cpu_int().imask().modify(|w| {
+                w.set_dio(i as usize, false);
+            });
+        } else {
+            // The other edge, or a wait that has since been cancelled. Discard it and stay armed.
+            // Waking at all is what errata GPIO_ERR_01 case 2 needs - the detector has no clock in
+            // STANDBY1, so it only observes the release if the release is itself a wake event - and
+            // leaving the status bit set would re-enter here forever.
```
i509VCB: "Did not receive the correct event or no event is being waited. Per GPIO_ERR_01 clear this event otherwise entering STANDBY1 will mean the next wake up event is not detected."

embassy-mspm0/src/gpio.rs
```diff
@@ -1050,14 +1160,36 @@ fn irq_handler(gpio: gpio::Gpio, port: Port) {
...
+            // Withdrawing both requests is what tells the waiting future its edge arrived.
+            rise.fetch_and(!mask, Ordering::Relaxed);
+            fall.fetch_and(!mask, Ordering::Relaxed);
+
+            let _ = GPIO_WAIT_MAP.wake(&id, ());
+
+            // Notify the future that an edge event has occurred by masking the interrupt for this pin.
+            gpio.cpu_int().imask().modify(|w| {
+                w.set_dio(i as usize, false);
+            });
```
i509VCB, line 1134: We don't read RIS anymore in the future?

embassy-mspm0/src/gpio.rs
```diff
@@ -382,6 +401,72 @@ impl<'d> Flex<'d> {
...
+        let (rise, fall, mask) = want_masks(pin_port);
+
+        // Published before fast wake is armed, so the interrupt can never see an edge without
+        // knowing which direction was asked for.
+        if matches!(polarity, Polarity::RISE | Polarity::RISE_FALL) {
+            rise.fetch_or(mask, Ordering::Relaxed);
+        }
+        if matches!(polarity, Polarity::FALL | Polarity::RISE_FALL) {
+            fall.fetch_or(mask, Ordering::Relaxed);
+        }
```
i509VCB: I don't think these ors have strong enough ordering.

One thread could be reading and then another thread preempts, reads and writes causing us to lose one bit. I think this needs AcqRel.

embassy-mspm0/src/gpio.rs
```diff
@@ -382,6 +401,72 @@ impl<'d> Flex<'d> {
...
+        let bit = usize::from(self.pin_port % 32);
+
+        critical_section::with(|_cs| {
+            self.block.fastwake().modify(|w| w.set_din(bit, false));
+            self.block.cpu_int().imask().modify(|w| w.set_dio(bit, false));
+        });
+        self.block.cpu_int().iclr().write(|w| w.set_dio(bit, true));
+
+        rise.fetch_and(!mask, Ordering::Relaxed);
+        fall.fetch_and(!mask, Ordering::Relaxed);
```
i509VCB, line 427: Same here, I think this needs to be AcqRel.

embassy-mspm0/src/gpio.rs
```diff
@@ -336,36 +338,53 @@ impl<'d> Flex<'d> {
...
-        let key = pin.pin_port();
+        let (rise, fall, mask) = want_masks(key);
         let result = GPIO_WAIT_MAP
             .wait_for(key, || {
-                if pin.block().cpu_int().ris().read().dio(pin.bit_index()) {
+                // The interrupt withdraws both requests when it accepts an edge, so their absence is
+                // the completion signal. Testing the request rather than the status bit also closes
+                // the window before the interrupt is first unmasked: an edge landing there is still
+                // held in the status bit, and unmasking below re-enters the handler to classify it.
+                if (rise.load(Ordering::Relaxed) | fall.load(Ordering::Relaxed)) & mask == 0 {
```
i509VCB, line 367: Note for ordering: Relaxed is fine here since the interrupt handler will preempt if needed (or we will wait).

embassy-mspm0/src/gpio.rs
```diff
@@ -336,36 +338,53 @@ impl<'d> Flex<'d> {
...
                 block.polarity15_0().modify(|w| {
-                    w.set_dio(pin.bit_index(), polarity);
+                    w.set_dio(bit, Polarity::RISE_FALL);
                 });
             };
         });
 
+        // Armed before the stale events are cleared, so the write below also discards anything
+        // latched by the arming itself.
+        let _arm = EdgeArm::new(block, key, polarity);
```
i509VCB, line 357: I think we should use `OnDrop` from embassy-hal-internal here and do the setup and then have the drop closure do the deinit in here.

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+//! the chip shallower than a given [`SleepLevel`]; the low-power executor then idles into the deepest
+//! mode no guard blocks.
+//!
+//! Each family implements the SYSCTL power-mode sequence from the device TRM (chapter
+//! "System Control (SYSCTL)" -> "Operating Modes"), cross-checked against TI driverlib's
+//! `DL_SYSCTL_setPowerPolicy*`. The families are split into three behaviors:
+//! - `full.rs` — STOP0/1/2 + STANDBY0/1: all G families and the supported L families.
+//! - `c110x.rs` — STOP0/2 + STANDBY0/1 (no STOP1); STOP0 also clears `USELFCLK`: C-series.
+//! - `h321x.rs` — STOP0/2 + STANDBY0/1 (no STOP1): H321x.
+//!
```
i509VCB: There is no value here in my opinion top say we copied driverlib. I think this highlighted section is needed in the docs.

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+pub use crate::sysctl::SleepLevel;
+
+/// `SLEEP_BLOCKS[i]` counts the guards forbidding `SleepLevel::LEVELS[i]` and every deeper mode.
+static SLEEP_BLOCKS: [AtomicU32; 5] = [
+    AtomicU32::new(0),
+    AtomicU32::new(0),
+    AtomicU32::new(0),
+    AtomicU32::new(0),
+    AtomicU32::new(0),
+];
```
i509VCB: Do these need to be AtomicU32? I would think you cannot have more than 255 things preventing a sleep mode from being entered. AtomicU8 is probably enough and will save 15 bytes of RAM.

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Only the instruction prefetcher is touched; the cache and literal caching stay as configured, as
+/// in TI's own sequence.
+///
+/// `CPU_ERR_02` and `CPU_ERR_03` are checked in every shipping revision of the L series
+/// (`SLAZ741E`, Rev C/D), the G series (`SLAZ738`, Rev B/C/D) and the H series (`SLAZ763B`, Rev A);
+/// `PMCU_ERR_13` in the L and H ones. Nothing gates on silicon revision because they hold for all
+/// of them. No errata sheet exists for the C series, so this is applied there on the assumption
+/// that the same CPUSS behaves the same way — the cost is a handful of register accesses either
+/// side of a `WFI`, and it is inert on parts that do not need it.
+struct PrefetchSuspend(pac::cpuss::regs::Ctl);
```
i509VCB, line 166: OnDrop can probably implement the drop half.

embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+        let saved = pac::CPUSS.ctl().read();
+        let mut disabled = saved;
+        disabled.set_prefetch(Prefetch::DISABLE);
+        pac::CPUSS.ctl().write_value(disabled);
+
+        // CPU_ERR_02: the write above only lands once any pending flash access retires, so force
+        // one that has to reach SYSCTL. TI's reference writes a scratch byte first; a read is
+        // enough, and leaves the shutdown storage — which survives SHUTDOWN and may hold caller
+        // state — untouched. The advisory names `SHUTDNSTORE` only because it is a harmless target;
+        // the metapac does not model it on H321x, where any other SYSCTL read does the same job.
```
i509VCB:
```
CPU_ERR_02 means the prefetch will not be disabled until pending flash access is finished. Reading any SYSCTL register after disabling prefetch will complete the pending flash access. 
```

embassy-mspm0/src/sysctl/mod.rs
```diff
@@ -23,6 +23,147 @@ mod inner;
 
 pub use inner::ClkOutSource;
 
+/// Deep-sleep modes, shallowest (most capable) to deepest (least capable).
+///
+/// The argument to [`WakeGuard::new`] and the vocabulary the low-power executor idles into. Not every
+/// family implements every mode (only G/L have STOP1); a mode a family lacks is rounded to the next
+/// shallower one it does have.
```
i509VCB: > The argument to [`WakeGuard::new`] and the vocabulary the low-power executor idles into

Is just fluff that tells the user nothing. This type's documentation can probably be simpler.

embassy-mspm0/src/sysctl/mod.rs
```diff
@@ -23,6 +23,147 @@ mod inner;
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum SleepLevel {
+    /// STOP0.
+    Stop0,
+    /// STOP1; rounded to STOP0 on families without it.
+    Stop1,
```
i509VCB: Should this even be a variant on chip families that don't support it?

embassy-mspm0/src/sysctl/mod.rs
```diff
@@ -23,6 +23,147 @@ mod inner;
...
+}
+
+impl SleepLevel {
+    pub const LEVELS: [SleepLevel; 5] = [
+        SleepLevel::Stop0,
+        SleepLevel::Stop1,
+        SleepLevel::Stop2,
+        SleepLevel::Standby0,
+        SleepLevel::Standby1,
+    ];
```
i509VCB: I see no reason for this to be public

embassy-mspm0/src/sysctl/mod.rs
```diff
@@ -23,6 +23,147 @@ mod inner;
...
+        } else if clock_hz > 0 {
+            Some(Self::Standby1) // STOP2/STANDBY0 (LFCLK) suffice; only STANDBY1 unclocks PD0
+        } else {
+            None // clock-agnostic
+        }
+    }
+}
+
+// Boundary checks for `floor_for_clock_hz`, verified at compile time.
+// `core::assert!` bypasses the crate's defmt-routing `assert!` shim, which is not const-evaluable.
```
i509VCB: Can we not do a `use core::assert;` in the scope? Also easier to say "crate::fmt cannot be used in const."

embassy-mspm0/src/sysctl/mod.rs
```diff
@@ -23,6 +23,147 @@ mod inner;
...
+/// idles into the deepest mode still permitted, or a plain `WFI` if even [`SleepLevel::Stop0`] is
+/// blocked. Guards are refcounted per level and compose: the shallowest blocked mode wins. Drop
+/// releases it.
+///
+/// Always available so drivers can hold one unconditionally; without the `low-power` feature it is a
+/// no-op (the `level` is ignored and nothing is tracked).
+#[must_use]
+pub struct WakeGuard {
+    #[cfg(feature = "low-power")]
+    level: SleepLevel,
```
i509VCB, line 179: If low-power is not enabled, it becomes possible for someone to do `WakeGuard {}` to create an instance outside `new`. If low-power is not enabled you need a field that contains a unit value.

embassy-mspm0/src/time_driver/mod.rs
```diff
@@ -1,4 +1,4 @@
 // TODO: Alternative TIMB implementation
 #[path = "tim.rs"]
 mod driver;
-pub use driver::*;
+pub(crate) use driver::*;
```
i509VCB, line 4: `time_driver` is not visible to the public API, so the fact that this is pub_use will not matter.

embassy-mspm0/src/uart/mod.rs
```diff
@@ -28,9 +28,18 @@ pub enum ClockSel {
...
+    ///
+    /// Required to receive after a deep sleep. An asynchronous fast clock request forces the
+    /// MCLK/ULPCLK tree back to the SYSOSC base frequency, so a UART on this source recovers a usable
+    /// baud clock; LFCLK is not restored by the request and MFCLK only comes back if it was already
+    /// enabled. Pair with [`Config::low_power_rx_wake`].
+    ///
+    /// The baud divisors are sized for the full ULPCLK rate, so an idle or in-flight transfer needs
+    /// the chip awake — a receiver holds a sleep floor of `Stop0` unless `low_power_rx_wake` is set,
+    /// and a transmitter holds one for the duration of each flush.
+    BusClk,
```
i509VCB, line 34: This is power domain dependent which the HAL does not do yet. I don't see a good reason to add this right now.

embassy-mspm0/src/uart/mod.rs
```diff
@@ -158,6 +167,29 @@ pub struct Config {
...
+    ///   an LFCLK-sourced receiver keeps its 32 kHz baud clock and never speeds up.
+    ///
+    /// A guard at [`SleepLevel::Standby1`](crate::sysctl::SleepLevel) is held for as long as the
+    /// receiver lives, capping the idle at STANDBY0. STANDBY1 is excluded deliberately: it unclocks
+    /// PD0, and the request then needs 241 µs typical / 284 µs maximum to deliver the first 32 MHz
+    /// edge — over two bit times at 9600 baud, so the start bit and the first data bits are lost.
+    /// From STANDBY0 the same delay is short enough to catch the start bit at ordinary baud rates.
+    ///
+    /// Only affects the async [`BufferedUart`] receiver; ignored by the blocking driver and by TX.
+    pub low_power_rx_wake: bool,
```
i509VCB, line 173: Mucho texto.

I also think this option is going to cause confusion. A user will listen with rx, let power go down to lowest mode possible and wonder why UART traffic doesn't wake the device.

embassy-mspm0/src/executor.rs
```diff
@@ -0,0 +1,83 @@
+//! MSPM0-specific `embassy-executor` platform.
+//!
+//! This module provides an `embassy-executor` platform specific for MSPM0 chips that integrates [`low_power::sleep()`](crate::low_power::sleep) in the main loop.
+//! Read the `embassy-executor` README for information about what "executor platforms" are and how they work.
+//!
+//! To use it:
+//! - Enable the `executor-thread` feature on this crate.
+//! - **Do not** enable features `platform-cortex-m`, `executor-thread` or `executor-interrupt` in the `embassy-executor` crate.
```
i509VCB, line 8: We should implement the `executor-interrupt` feature.

embassy-mspm0/src/gpio.rs
```diff
@@ -382,6 +401,72 @@ impl<'d> Flex<'d> {
...
+/// an asynchronous fast clock request instead, which restores ULPCLK from SYSOSC for long enough to
+/// run the detector (TRM "GPIO Fast Wake"). It is inert in RUN mode, where the request is ignored
+/// because MCLK already runs from SYSOSC at its base frequency.
+///
+/// Dropping disarms the pin completely, so a cancelled wait leaves neither a clock request nor an
+/// interrupt behind.
+///
+/// NOTE: `SYSOSCCFG.BLOCKASYNCALL` has to stay clear. With fast wake enabled the GPIO expects a
+/// handshake from SYSCTL, and if the request is blocked the pin receives no clock at all. It resets
+/// to 0 and nothing here sets it; anything that ever does must account for this.
```
i509VCB: This feels like too much text.

embassy-mspm0/src/low_power/mod.rs
CordlessCoder: Technically the user could leak a bunch of WakeGuard's but I don't think it's unreasonable to panic in that case, I can make these narrower and panic if one overflows.

embassy-mspm0/src/gpio.rs
CordlessCoder: I don't believe that's the case, AcqRel would only cause these to form happens-before relationships, which we don't care about because the order that the bits are set in doesn't matter, as long as they're set.

embassy-mspm0/src/sysctl/mod.rs
CordlessCoder: But then each peripheral needs to care about what sleep modes are supported on what family, when all it actually cares about is "don't go into STOP1 or lower".

embassy-mspm0/src/sysctl/mod.rs
CordlessCoder: No, a use doesn't shadow a crate macro_use.

CordlessCoder:
The core executor and sleep mode changes have been split off into #6646, this PR will probably be superseded by multiple smaller PRs implementing low-power support for individual peripherals.

embassy-mspm0/src/low_power/mod.rs
CordlessCoder: This is used in two places, so the setup half would need to be duplicated.

embassy-mspm0/src/uart/mod.rs
CordlessCoder: This is the only clock source that UART can issue an async fast clock request from, so the only one that will work for a deep sleep wake.
I2C already exposes BusClk as a clock option, even though it also depends on power domains.
https://github.com/embassy-rs/embassy/blob/2d98e96dd42a56b94bed34e97361abd4ab59391f/embassy-mspm0/src/i2c.rs#L28

embassy-mspm0/src/sysctl/mod.rs
CordlessCoder: Retention has turned out to be a lot more complex, it'll be addressed in separate peripheral-specific PRs.

embassy-mspm0/src/uart/mod.rs
CordlessCoder: Power domains have been implemented in mspm0-metapac and are waiting to be merged - power domain support will be implemented in peripheral specific PRs.

CordlessCoder:
It doesn't make much sense to keep this PR around since I've decided to split all of the changes off into smaller PRs.

PR #6594, opened 2026-07-23
feat: embassy-net: DHCPv4 server

### LINKED ISSUE #2845: embassy-net feature request DHCP server support.
I've been looking at embassy as an option for an embedded system with wifi, the system is intended to be able to operate in both AP mode for setup purposes and for local use in offline locations and to be connected to an existing wifi network.

However a roadblock I have run into is a lack of DHCP server support. It is possible to use link-local addresses (as the example does) but this does not provide a good client experience.

kripton:
This PR adds a basic DHCPv4 server (dhcpd) to embassy-net.

An example is added for the RP235x to showcase using the DHCP server with USB CDC NCM (Tested on a Pico 2 board)
Another example is added for the RP2040 to showcase using the DHCP server in combination with the cyw43 WiFi AP (Tested on a Pico W board).

Fixes #2845

This is my first production code in Rust, so please DO comment and propose enhancements or request changes if something could or should be improved. But please be gentle :) Thanks!

### DIFF
--- a/embassy-net/Cargo.toml
+++ b/embassy-net/Cargo.toml
@@ -52,7 +52,7 @@ features = ["defmt", "tcp", "udp", "raw", "dns", "icmp", "dhcpv4", "proto-ipv6",
 target = "thumbv7em-none-eabi"
 
 [package.metadata.docs.rs]
-features = ["defmt", "tcp", "udp", "raw", "dns", "icmp", "dhcpv4", "proto-ipv6", "medium-ethernet", "medium-ip", "medium-ieee802154", "multicast", "dhcpv4-hostname", "packetmeta-id"]
+features = ["defmt", "tcp", "udp", "raw", "dns", "icmp", "dhcpv4", "proto-ipv6", "medium-ethernet", "medium-ip", "medium-ieee802154", "multicast", "dhcpv4-hostname", "packetmeta-id", "dhcpd"]
 
 [features]
 default = ["auto-icmp-echo-reply"]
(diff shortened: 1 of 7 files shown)


xoviat:
`./fmtall.sh`

kripton:
> `./fmtall.sh`

Thanks, done, all checks green now :)

kripton:
One thing we could discuss: For the leases, a `Vec` of a custom struct is used. I thought about switching that to a `HashMap` with the client_id/MAC address as key. However, it would have required a custom struct for the values (since we need to store the IPv4 address and the time the lease was last refreshed), adding one layer of complexity for minimal gain. So, I left it as a `Vec`


And: DhcpdOptions fields `domain_name` and `mtu` are not yet implemented

bugadani:
https://crates.io/crates/edge-dhcp exists, and I think improving that if necessary should be preferred over a separate implementation in embassy-net. Not everything has to be part of embassy-net, nor should we cause ecosystem fragmentation if we don't have to.

kripton:
> https://crates.io/crates/edge-dhcp exists, and I think improving that if necessary should be preferred over a separate implementation in embassy-net. Not everything has to be part of embassy-net, nor should we cause ecosystem fragmentation if we don't have to.

Great, thanks for the hint. I did some research before starting the implementation but was not able to find any no-std implementation of a DHCP server and was not aware of edge-dhcp. Also, no one replied in #2845, so I assumed there is just nothing yet.

I agree that we should not duplicate implementations and I will check if edge-dhcp works fine with embassy-net and fits my use cases. If it does, I will close this PR.

xoviat:
I'm sorry about this, but I have to agree. There's no point in duplicating work across the ecosystem and edge-net already exists. I've created #6598 to clarify this.

kripton:
No worries, it was a nice exercise anyways. Good to get into the topic of embedded Rust.

Someone mentioned Leasehund here: https://github.com/embassy-rs/embassy/issues/2845#issuecomment-5079766638 and API-wise this seems to be much closer to what I did and would expect (compared to edge-dhcp, which is nice but different to use).

Closing this

PR #6599, opened 2026-07-24
2d gpdma support

liebman:
Adds support for GPDMA 2D addressing (block repeat with per-burst and per-block address offsets) through a new `TwoDItem` descriptor and `TwoDConfig`, gated behind a `gpdma2d` cfg that build.rs enables when any DMA channel has `supports_2d`.

The main structural change: `LinearItem` was previously a concrete struct with its own `Table`. Now there's a `LinkedListItem` trait that both `LinearItem` and `TwoDItem` implement, and `Table<T, N>` is generic over the item type. This lets the same table/linking infrastructure work for both linear and 2D transfers without duplication.

The common 5-word descriptor prefix (TR1, TR2, BR1, SAR, DAR) is factored into a shared `Item` struct that both item types embed as their first #[repr(C)] field, preserving the hardware-required memory layout. `Item` owns construction (Item::new), config application (apply_config), and transfer_count -- so when new common config knobs are added, there's one place to handle them.

On the channel side, `configure_linked_list_raw` replaces the old `configure_linked_list` to accept raw table parameters (base address, offset, count, transfer count, is_2d flag), so both item types can share the same channel setup without the channel needing to know the concrete item type. 2D items set UT3/UB2 in the initial LLR so the DMA loads the extra TR3/BR2 words.

Build.rs picks up `supports_2d` from stm32-data's DMA channel metadata and emits `dma_channel_2d_impl!` for those channels, implementing the new `TwoDChannelInstance` marker trait for compile-time 2D capability checking.

### DIFF
--- a/embassy-stm32/CHANGELOG.md
+++ b/embassy-stm32/CHANGELOG.md
@@ -24,6 +24,7 @@ DMA:
 - fix: stm32/dma: auto-set `TR1.PAM = Pack` on GPDMA when source and destination widths differ, instead of silently zero-extending one beat per destination beat
 - fix: stm32/dma: compute GPDMA `BR1.BNDT` from the memory-side width regardless of direction, fixing destination overrun on reads with peripheral width > memory width
 - feat: stm32/dma: GPDMA: allow access to construct custom LinkedList chains for scatter/gather DMA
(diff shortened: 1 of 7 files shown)


xoviat:
bender run

PR #6602, opened 2026-07-25
stm32/aes: add aes_v2 driver, sharing a common cipher layer with aes_v3b

theatrus:
Adds a driver for the aes_v2 peripheral revision (STM32G0, G4, L5, U0, WL), which embassy previously did not support (the existing driver targets aes_v3b only, on H5/WBA).

To avoid duplicating the cipher layer, the register-agnostic code is factored into a shared aes/common.rs (following the can/common.rs precedent): the Error/Direction/KeySize types, the Cipher trait and its ECB/CBC/CTR/GCM/GMAC/ CCM implementations, the Context, and the GCM/CCM state machine (op_start/op_aad/op_payload/op_finish). The six register primitives that actually differ between the revisions (CCF/error clear, CHMOD encoding, DINR/DOUTR/KEYR/IVR access) are selected with #[cfg(aes_v2)]/#[cfg(aes_v3b)].

aes/v2.rs and aes/v3b.rs are thin shells over common, owning only their instance wiring and constructors:

- v2 is blocking by default. On parts whose AES engine has a dedicated interrupt vector (L5, WL) it also offers an interrupt-driven async API, gated on an InterruptableInstance trait implemented via foreach_interrupt!; where the engine shares its vector or has none (U0, G0, G4) only the blocking API compiles in. new_blocking() takes no interrupt binding.

- v3b keeps its existing public API unchanged (Aes<'d,T,M>, new_blocking(irq), the DMA new()/DmaIn/DmaOut, InterruptHandler, Instance::Interrupt); only its method bodies now delegate to common.

The public module is embassy_stm32::aes for both revisions, selected by cfg following the adc pattern.

One fix, in shared code: setup() now clears CR.NPBLB when starting an operation. The driver writes that field only when a final partial block needs padding, but the hardware keeps the last value, so a leftover count from an earlier message corrupted the next message's authentication tag. On a NUCLEO-U083RC, a CCM message that followed a GCM message with a partial final block produced a wrong tag until NPBLB was cleared. This also applies to v3b.

examples/stm32u0/src/bin/aes.rs runs known-answer tests for every mode: NIST SP 800-38A vectors for ECB/CBC/CTR, SP 800-38D for GCM and GMAC, and an ST HAL vector for CCM, covering partial final blocks and multi-call AAD and payload. All 24 checks pass on a NUCLEO-U083RC.


xoviat:
resolve the merge conflict, please.

theatrus:
> resolve the merge conflict, please.

Resolved, the library got ahead of me :)

theatrus:
Note I don't have L5 hardware, so I am unable to verify the v2aes block there (or with proper interrupts).

leftger:
Thanks for the addition!

PR #6585, opened 2026-07-22
nxp: fix default features build dep

xoviat:

PR #6619, opened 2026-07-26
embassy-nxp: migrate to nxp-pac 8512ece

VictorVulturescu:
Migrates embassy-nxp to the latest revision of nxp-pac (8512ece). This entails the following changes:

 - METADATA.interrupts now has the type: (&str, u32) (build.rs was modified to accommodate this change)
 - enum constants became PascalCase instead of SCREAMING_CASE
 - a few methods have been renamed since the previous rev ex: set_matchn_l become set_matc_hn_l
 - in Cargo.toml, all dependencies were bumped to the rev = "8512ece...". Also the build dep was changed to have default-features = "false" (overlap with pr: #6585), to take advantage of the lib.rs file in the new rev, so the build works on non ELF hosts (MacOS and Windows)
 
Tested on LPCXpresso55S69-EVK dev-board, with the official examples: blinky_nop, blinky_embassy_time, button_executor, pwm, usart_blocking, usart_async, which all work properly.

### DIFF
--- a/embassy-nxp/Cargo.toml
+++ b/embassy-nxp/Cargo.toml
@@ -40,13 +40,13 @@ embassy-time-queue-utils = { version = "0.3.2", path = "../embassy-time-queue-ut
 embedded-io = { version = "0.7.1" }
 embedded-hal-02 = { package = "embedded-hal", version = "0.2.6", features = ["unproven"] }
 ## Chip dependencies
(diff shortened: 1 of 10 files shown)


eva-cosma (approved):

diondokter (approved):
Looks alright to me, this is very similar to changes we made in embassy-mcxa.
But I'm not a maintainer for this HAL, so I'll let someone else hit the merge button.

PR #6624, opened 2026-07-26
mcxa: support clock configuration for all I3C instances

ayodejiige:
`I3cConfig::pre_enable_config()` was hard-coded to I3C0's MRCC registers, so
instantiating the driver on I3C1-I3C3 programmed instance 0's clock and left the
intended peripheral unclocked.

Adds an `I3cInstance` enum carried on `I3cConfig`, selected by `match` instead of
hard-coding instance 0. The instance is declared at the type level via
`SealedInstance::CLOCK_INSTANCE` and populated by the `impl_i3c_instance!` macro, so
the application never sets it (`I3cConfig::instance` is `pub(crate)`).

### DIFF
--- a/embassy-mcxa/src/clocks/periph_helpers.rs
+++ b/embassy-mcxa/src/clocks/periph_helpers.rs
@@ -691,6 +691,24 @@ pub enum I3cClockSel {
     None,
 }
 
+/// Which instance of the `I3c` is this?
+///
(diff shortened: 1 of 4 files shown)


felipebalbi (approved):
Looks good to me.

felipebalbi:
bender run

PR #6628, opened 2026-07-26
C5: Fix ADC_DAC limits panic for all examples

usbalbin:
Set ADC_DAC prescaler to a safer default to prevent every single example except `adc` (which was fine) from panicking (sorry my bad)

### DIFF
--- a/embassy-stm32/src/rcc/c5.rs
+++ b/embassy-stm32/src/rcc/c5.rs
@@ -64,7 +64,7 @@ impl Config {
             apb1_pre: APBPrescaler::Div1,
             apb2_pre: APBPrescaler::Div1,
             apb3_pre: APBPrescaler::Div1,
-            adcdac_pre: Adcdacpre::Div1,
+            adcdac_pre: Adcdacpre::Div4,
             mux: super::mux::ClockMux::default(),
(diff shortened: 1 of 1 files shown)


leftger:
I don't have any fundamental disagreement with changing defaults. My question is if this change is a safe or prudent default. In my own examples I like to provide the user with the best "bang for the buck" where I showcase the most performant (fastest) capabilities. If this change is a normal and typical default then that's fine, but if the cause is the examples themselves I'd think it'd make more sense to provide an explanation as to why the defaults have been changed. You're the only one working on C5 so this is up to your direction TBH.

usbalbin:
ADC max speed is 36MHz. HSI is 144MHz(144/4=36), HSIDIV3=48MHz. HSIDIV3 is the default system clock source. A Div4 makes the ADC work no matter the system clock source, however not as fast as it could given a HSIDIV3 sysclock. Given that this (as implemented before this PR) causes panics even if not using the ADC. For example something like a blinky example(not using ADC) running at 144MHz would have to set adcdacsel to Div4 just to prevent panics. Therefore I choose Div4 as the default as it never panics.

I am happy to change it though

PR #6630, opened 2026-07-26
Document and fix possible 1s offset in MSPM timer logic.

CordlessCoder:


embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -162,13 +171,23 @@ impl TimxDriver {
...
         regs.counterregs(0).ctrctl().modify(|w| {
             w.set_en(true);
         });
+
+        // Enabling the counter with CVAE = ZEROVAL zeroes it, which can latch a zero event. Discard
+        // anything latched up to here before arming the interrupt, or that event would be counted as
+        // a wrap that never happened and put the whole timebase a half period off.
+        regs.cpu_int(0).iclr().write(|w| {
+            w.set_z(true);
+            w.set_l(true);
```
i509VCB: We should not need to clear load? We do not enable that interrupt.

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -273,34 +294,30 @@ impl Driver for TimxDriver {
...
+        // out the same way: calc_now(X even, 0x8000) = (X + 1) << 15.
+        //
+        // The counter can only ever be the fresher of the two reads, never the staler one -- both
+        // boundary events assert on the very TIMCLK cycle the counter reaches the boundary value, so
+        // a latched event and an already-advanced counter always arrive together.
+        //
+        // NOTE: You must NOT wait for the interrupt here: now() is called inside critical sections,
+        // where it would deadlock.
+        calc_now(period, counter)
     }
```
i509VCB, line 324: The timer driver will go backwards if this is changed. Yes parity should be handling this, but I guess it does not.

embassy-mspm0/src/time_driver/tim.rs
```diff
@@ -162,13 +171,22 @@ impl TimxDriver {
...
-        unsafe { <T as tim::Instance>::Interrupt::IRQ.enable() };
-
         // Allow the counter to start counting.
         regs.counterregs(0).ctrctl().modify(|w| {
             w.set_en(true);
         });
+
+        // Enabling the counter with CVAE = ZEROVAL zeroes it, which can latch a zero event. Discard
+        // anything latched up to here before arming the interrupt, or that event would be counted as
+        // a wrap that never happened and put the whole timebase a half period off.
```
i509VCB:
```suggestion
        // Enabling the counter with CVAE = ZEROVAL zeroes it, which can latch a zero event. Discard
        // anything latched up to here before arming the interrupt, or that event would be counted as
        // a wrap that never happened and put the whole timebase a one period off.
```

PR #6632, opened 2026-07-27
Relax RISC-V executor mem ordering

robamu:


embassy-executor/src/platform/riscv.rs
```diff
@@ -62,13 +64,11 @@ mod thread {
...
-                        if SIGNAL_WORK_THREAD_MODE.load(Ordering::SeqCst) {
-                            SIGNAL_WORK_THREAD_MODE.store(false, Ordering::SeqCst);
-                        }
-                        // if not, wait for interrupt
-                        else {
+                        // if there is work to do, loop back to polling. otherwise wait for interrupt
+                        //
+                        // Use Acquire ordering here to establish a happens-before relationship
+                        // with the store in __pender.
+                        if !SIGNAL_WORK_THREAD_MODE.swap(false, Ordering::Acquire) {
```
bugadani: This will take a critical section on `imc` targets, which can do the load-store atomically usually.

robamu: hmm. that is unfortunate.

robamu: so even if the A extension is not supported, the CPU can do a load-store atomically? is this guaranteed?

robamu: well, it is a critical section anyway. maybe we can even use relaxed everywhere here?

robamu: it is definitely more relaxed than before

bugadani:
Wouldn't `Relaxed` be sufficient? We are reading/writing a single variable. Quoting Rust Atomics and Locks:

> Relaxed still guarantees consistency on a single atomic variable, but does not promise anything about the relative order of operations between different variables.

robamu:
I think in this case, it does not matter because the interrupts are disabled while the variable is read and written in the executor runner.

If I understand https://mara.nl/atomics/memory-ordering.html#release-and-acquire-ordering correctly, if we want to ensure that anything before the pender store happens before everything after the load of the same variable, then ordering becomes important.

robamu:
maybe the same relaxation can be done for AVR?

PR #6637, opened 2026-07-27
stm32: preserve rx'd UART data that doesn't fit in buffer

wojciech-graj:
The BufferedInterruptHandler used to discard excess data if its buffer was full. This meant that flow control, if enabled, was unused, because the hardware was having its internal buffer cleared.

We now check whether the buffer has capacity for new data before reading the bytes out of the hardware buffer (thereby removing them from the buffer). Interrupts indicating that there is new data in the hardware RX buffer are disabled until the buffer has space.

### DIFF
--- a/embassy-stm32/src/usart/buffered.rs
+++ b/embassy-stm32/src/usart/buffered.rs
@@ -16,6 +16,7 @@ use super::{
     clear_interrupt_flags, configure, half_duplex_set_rx_tx_before_write, rdr, reconfigure, send_break, set_baudrate,
     sr, tdr,
 };
+use crate::atomic::AtomicModify;
 use crate::gpio::{AfType, Flex, Pull};
 use crate::interrupt::{self, InterruptExt};
 use crate::time::Hertz;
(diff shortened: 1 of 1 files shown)


embassy-stm32/src/usart/mod.rs
```diff
@@ -2196,6 +2203,16 @@ fn clear_interrupt_flags(r: Regs, sr: regs::Isr) {
...
+#[cfg(not(usart_v4))]
+fn set_read_interrupt_enable(r: Regs, value: bool) {
+    r.cr1().modify(|w| w.set_rxneie(value));
+}
+
+#[cfg(usart_v4)]
+fn set_read_interrupt_enable(r: Regs, value: bool) {
+    r.cr3().modify(|w| w.set_rxftie(value));
+}
+
```
xoviat: calls to modify outside of the irq handler that are reading/writing to the same registers need to use set_bits/clear_bits.

wojciech-graj: Done

PR #6644, opened 2026-07-28
[MCXA] Make flash driver work for mcxa5

diondokter:
Uses the existing driver, but expanded where required.
Included the same example from the 2xx and that runs just fine


felipebalbi (approved):
Did not test on actual HW but looks good to me from a review standpoint.

diondokter:
Works on my pc, so ship it!

PR #6646, opened 2026-07-28
MSPM0: Deep-sleep modes and executor

CordlessCoder:
This PR does not currently implement the peripheral guards needed to make peripherals safe to use under the low-power executor. This is being implemented in PR #6580.

### DIFF
--- a/embassy-mspm0/src/low_power/mod.rs
+++ b/embassy-mspm0/src/low_power/mod.rs
@@ -1,0 +1,204 @@
+//! Low-power (deep-sleep) support.
+//!
+//! Deep-sleep depth is gated by [`WakeGuard`](crate::sysctl::WakeGuard), which drivers hold to keep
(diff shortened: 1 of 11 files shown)


embassy-mspm0/src/low_power/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    let _prefetch = PrefetchSuspend::new();
+
+    let mut scb = unsafe { cortex_m::Peripherals::steal() }.SCB;
+    scb.set_sleepdeep();
+    cortex_m::asm::dsb();
+
+    cortex_m::asm::wfi();
+
+    unsafe { core::hint::unreachable_unchecked() }
```
i509VCB, line 161: Probably should add my suggested safety comment in the last review I did so that readers know WFI will truly never return.

CordlessCoder:
This PR is the base for #6652, so that will be unblocked once this is merged.

PR #6647, opened 2026-07-28
stm32-wpan: fix ub due to bytecasting status codes from ipc.

Dirbaio:
AssociateConfirm.status in the tests is reporting 0xE1/0xE9 which is not
in the enum. Bytecasting forces creating the enum with a byte value it shouldn't
have, which is UB. The defmt format impl was jumping to nowhere due to overflowing a jump table.

Actually adding all the missing enum cases is TODO, but at least it's not UB now.


xoviat:
Thanks! I don't think I wrote this, but I would always get UB sometimes during a compiler update.

PR #6654, opened 2026-07-29
embassy-nxp: ADC blocking read driver with example

sanyaswee:
This PR adds the ADC driver with blocking_read implementation and a usage example. It also uses the newer PAC revision, so the migration PR #6619 should be merged first


sanyaswee:
Commit history seems messed up, I will fix it today

embassy-nxp/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 // This mod MUST go first, so that the others see its macros.
 pub(crate) mod fmt;
 
+pub mod adc;
```
eva-cosma, line 8:
```suggestion
#[cfg(lpc55)]
pub mod adc;
```

This will fix half of your CI build issues

examples/lpc55s69/src/bin/adc_blocking.rs
```diff
@@ -0,0 +1,28 @@
...
+use embassy_executor::Spawner;
+use embassy_nxp::adc::{Adc, Config};
+use embassy_time::Timer;
+use panic_halt as _;
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_nxp::init(Default::default());
+
+    let config = Config::default(); // corresponds to Config::new(Resolution::Bits16, Averaging::None);
```
eva-cosma:
```suggestion
    // The default configuration corresponds to Config::new(Resolution::Bits16, Averaging::None);
    let config = Config::default();
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+                        });
+                    }
+                    _ => unreachable!(),
+                }
+            }
+        }
+    };
+}
+
+// https://www.nxp.com/docs/en/data-sheet/LPC55S6x.pdf
```
eva-cosma, line 289: Where in this document did you find these?

I also found them in UM11126 Table 745 (pages 760-761)

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn configure_iocon(&mut self);
+}
+
+/// Macro to implement the AdcPin trait for pins
+macro_rules! impl_adc_pin {
+    // pin     => peripheral struct
+    // port    => gpio port number
+    // pin_num => pin number
+    // channel => u8 channel number
+    // ctype   => channel side (0=A, 1=B)
```
eva-cosma:
```suggestion
    // pin     => pin peripheral struct (e.g.`PIO1_8`)
    // port    => pin port number (e.g. `1` for `PIO1_8`)
    // pin_num => pin number (e.g. `8` for `PIO1_8`)
    // channel => ADC channel number (as u8)
    // ctype   => channel side (0=A, 1=B)
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+    ($pin:ident, $port:expr, $pin_num:expr, $channel:expr, $ctype:expr) => {
```
eva-cosma: `ctype` could be renamed to something like multiplexer_channel or something like that.
`port` => `pin_port` ( to be consistent with `pin_num`)
`channel` => `adc_channel`

Also for the `pin_port` we could use gpio::Bank struct, and for `ctype`/`multiplexer_channel` we could also do an enum so we don't have an out-of-bounds value.

(Note: for the naming suggestions, if you have other suggestions, feel free to suggest)

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+                        iocon.pio1($pin_num).modify(|w| {
+                            w.set_func(0.into());
+                            w.set_mode(0.into());
+                            w.set_digimode(0.into());
+                            w.set_asw(1.into())
+                        });
+                    }
+                    _ => unreachable!(),
+                }
+            }
```
eva-cosma: Is there a way for us to have this code not inside of the macro? It creates a bigger-than-needed macro. Take a look at how GPIO does it. Furthermore, see if you can't just have a type constraint to gpio::Pin, so in the end all you end up with is 

```rust
pub trait AdcPin : gpio::Pin {
 ctype ...
 configure_iocon ...
} // <-- note, maybe a default implementation for configure_iocon?

// ...

impl<'d> crate::adc::AdcPin for crate::Peri<'d, crate::peripherals::$pin>  {
  ctype ...
}
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+        Self { resolution, averaging }
+    }
+
+    /// The default config
+    pub const fn default() -> Self {
+        Self {
+            resolution: Resolution::Bits16,
+            averaging: Averaging::None,
+        }
+    }
```
eva-cosma: Implement the default trait

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,289 @@
...
+
+/// The main struct
+pub struct Adc<'d> {
+    _peri: Peri<'d, ADC0>,
+    config: Config,
+}
+
+impl<'d> Adc<'d> {
+    /// Creation and initialization of ADC
+    /// adc parameter should be reaplced with Peri once nxp-pac implements it
```
eva-cosma: Did we not merge for this already? Comment still relevant?

eva-cosma (changes requested):

embassy-nxp/src/adc/lpc55.rs
sanyaswee: Table 3 starting on page 8, in the last column they have ADC channels described for supported pins

embassy-nxp/src/adc/lpc55.rs
sanyaswee: You are right, I forgot to remove this comment

embassy-nxp/src/adc/lpc55.rs
sanyaswee: If the issue is with macro size I can move this function out of it and just call it from the macro. Regarding the default implementation, I don't think it is necessary because it depends on pin port and number, which is different every time. Regarding a type binding, that's a good idea, I will try to implement it

eva-cosma: The idea behind is that all of the things that are pin-dependant you implement them as getters.

So if you need the pin bank and pin number you have
```
fn get_pin_bank() -> gpio::Bank { $pin_bank } 
fn get_pin_number() -> u8 { $pin_number } 
```

And from the the configure_iocion could be
```
pub trait AdcPin {
  fn get_pin_bank() -> gpio::Bank;
  fn get_pin_number() -> u8;
  fn configure_iocon(&self) {
  let pin_bank  = self.get_pin_bank();
  let pin_number = self.get_pin_number();
  // ....
  }
}
```

embassy-nxp/src/adc/lpc55.rs
eva-cosma: Can you add this in the comment?

embassy-nxp/src/adc/lpc55.rs
eva-cosma: Also, pin banks and pin number are already implemented by the gpio::Pin trait, so we can reuste that

eva-cosma (changes requested):

embassy-nxp/src/adc/lpc55.rs
sanyaswee: Okay, I'll add this

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,297 @@
...
+    // pin          => pin peripheral struct (e.g.`PIO1_8`)
+    // adc_channel  => ADC channel number (as u8)
+    // channel_side => channel side (0=A, 1=B)
+    ($pin:ident, $adc_channel:expr, $channel_side:expr) => {
+        impl<'d> crate::adc::AdcPin for crate::peripherals::$pin {
+            fn channel(&self) -> u8 {
+                $adc_channel
+            }
+
+            fn ctype(&self) -> u8 {
```
eva-cosma:
```suggestion
            fn channel_side(&self) -> ChannelSide {
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+/// Trait that provides channel numbers for pins that support ADC
+pub trait AdcPin: crate::gpio::Pin {
+    /// Channel number
+    fn channel(&self) -> u8;
+
+    /// Channel side (A / B), 0 = A, 1 = B
+    fn ctype(&self) -> u8;
```
eva-cosma:
```suggestion
    /// Channel side (A / B), 0 = A, 1 = B
    fn channel_side(&self) -> ChannelSide;
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,297 @@
...
+        let iocon = pac::IOCON;
+        let bank = self.bank();
+        let pin = self.pin();
+        match bank {
+            crate::gpio::Bank::Gpio0 => {
+                iocon.pio0(pin.into()).modify(|w| {
+                    w.set_func(0.into());
+                    w.set_mode(0.into());
+                    w.set_digimode(0.into());
+                    #[cfg(not(feature = "lpc55s16"))]
```
eva-cosma:
```suggestion
                    #[cfg(feature = "lpc55s69")]
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,297 @@
...
+                    w.set_asw(1.into())
+                });
+            }
+            crate::gpio::Bank::Gpio1 => {
+                iocon.pio1(pin.into()).modify(|w| {
+                    w.set_func(0.into());
+                    w.set_mode(0.into());
+                    w.set_digimode(0.into());
+                    #[cfg(not(feature = "lpc55s16"))]
```
eva-cosma:
```suggestion
                    #[cfg(feature = "lpc55s69")]
```

embassy-nxp/src/adc/lpc55.rs
```diff
@@ -0,0 +1,297 @@
...
+                    w.set_asw(1.into())
+                });
+            }
+        }
+    }
+}
+
+#[repr(u8)]
+enum ChannelSide {
```
eva-cosma: Doc comment explaining how the each ADC channel has 2 pins as inputs, enabling to either read from side A, side B, or have a combination of them (I think one did A - B, and the other did something else fancy, read the user manual and see what they said there)

eva-cosma (changes requested):
I think these are the last nitpicks I have on this topic

eva-cosma (approved):

PR #6661, opened 2026-07-30
sync: fix "will return an error" docs wording.

Dirbaio:


embassy-sync/src/channel.rs
```diff
@@ -14,8 +14,8 @@
 //! passed from exception mode e.g. out of an interrupt handler.
 //!
 //! This module provides a bounded channel that has a limit on the number of
-//! messages that it can store, and if this limit is reached, trying to send
-//! another message will result in an error being returned.
+//! messages that it can store. If this limit is reached, trying to send
+//! another message wither waits or returns an error depending on the function.
```
shepmaster: “Either”

Dirbaio: oops

PR #6662, opened 2026-07-30
C5/H5: flash

usbalbin:
Adds
* bank swap - taken from flash/h50.rs
* Change erase bank selection from physical to logical to match how blocking_read and blocking_write work


usbalbin:
Tested on Nucleo-C5A3 and Nucleo-H563

xoviat:
bender run

usbalbin:
Note that the C5 tests were done against https://github.com/embassy-rs/stm32-data/pull/884

PR #6671, opened 2026-07-30
embassy-time: make Instant arithmetic const

vic1707:
followup to #6447


xoviat:
Should these functions have #[inline] tags?

vic1707:
didn't since they weren't I guess I can 👍

PR #6673, opened 2026-07-31
embassy-stm32: stop hardcoding LSESYSEN on u0

sebastianmueller01:
on u0, `lse_sysen` was hardcoded to `lse_en`, so LSESYSEN was forced on whenever LSE was enabled. that ignores `LseConfig::peripherals_clocked` and on my dev board I found that this led to 350nA of increased current draw


Dirbaio:
bender run

PR #6675, opened 2026-07-31
ci: update docserver and enable monitor

xoviat:
the docsever will pause cargo-batch if the system is running low on memory.


xoviat:
bender run

xoviat:
bender run

PR #6684, opened 2026-08-01
Clear a stale suspend when enumeration is done

HaoboGu:
Suspended bits needs to be clear when enumeration is done, otherwise if `Bus::poll()` runs late enough that `ENUMDNE` and `USBSUSP` are both latched, the suspend bit remains, resulting in enumeration failure.

The fix is from TinyUSB:

https://github.com/hathach/tinyusb/blob/master/src/portable/synopsys/dwc2/dcd_dwc2.c#L1197-L1198


bugadani (approved):

bugadani:
Could you rebase your branch? I think the old base is causing the CI issue

PR #6685, opened 2026-08-01
Add HAL API guidelines document.

Dirbaio:
Add a document explaining API patterns and guidelines for writing Embassy HALs.

Portions taken from mcxa DEVGUIDE.md by @jamesmunns, mostly everything that's not mcxa-specific.

Goal is:
1. We currently have many inconsistencies between HALs, it would be great to make them consistent. I have this LLM-made [list of iconsistencies](https://gist.github.com/Dirbaio/7a7caee9129f9a964153d189a4edd195) I plan to review and fix in the future.
2. to have a document contributors can reference when writing code and reviewers when reviewing.

I've made the document by looking at existing HALs, so nothing in it should be too surprising. When HALs disagree I've picked what's more common. Only two exceptions are:
- gpio: added `is_input`, `is_output`, `is_disconnected` to `Flex`. This is a wanted feature, see #6221
- uart: specified split by ref returns `(UartTx, UartRx)`, not `(&mut UartTx, &mut UartRx)`. This is smething that's always bugged me, it forces the unsplit uart to physically contain both halves, duplicating info.

It documents:
- Patterns (peripheral singletons, interrupts, sealed traits, init, etc)
- Naming of things
- Embedded-hal and other trait implementations
- APIs of common peripherals (gpio, uart, spi, i2c, etc)

### DIFF
--- a/API-GUIDELINES.md
+++ b/API-GUIDELINES.md
@@ -1,0 +1,1655 @@
+# Embassy HAL developer's guide and API guidelines
+
+## Naming
+
+Sometimes the same thing has different names depending on vendor, chip family, etc. We follow this algorithm to choose names, in order of decreasing priority.
+
+1. Be consistent within each HAL.
+2. Be consistent with the vendor documentation
+3. Be consistent with the industry-wide accepted names.
+
+Examples:
+
+- Nordic calls I2C "TWI", probably for trademark reasons. Even though the industry-wide name is "I2C" we call it "TWI" in `embassy-nrf`, because `2` takes priority over `3`.
+- We use "new" terms for SPI/I2C roles like "controller/peripheral/target" or "legacy" terms like "master/slave" matching what the vendor uses in their documentation, because `2` takes priority over `3`.
+- ST itself is inconsistent in DMA naming. STM32F4, F7, H7 uses "DMA stream" and "DMA channel" to refer to what other STM32 families call "DMA channel" and "DMA request number" respectively. `1` takes priority over `2`, so `embassy-stm32` uses "DMA channel" and "DMA request number" since it's both the most common in STM32 documentation and industry-wide. 
+
+## The `Cargo.toml` file
+
+This section describes the notable components of the `Cargo.toml` package
+manifest.
+
+### `package.metadata`
+
+As an embassy crate, we have a couple of embassy-specific metadata sections.
+
+* `package.metadata.embassy`
+    * This section is used for determining how to build the crate for embassy's
+      CI process.
+* `package.metadata.embassy_docs`
+    * This section is used for determining how to generate embassy's API docs.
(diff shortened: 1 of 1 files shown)


leftger (approved):

API-GUIDELINES.md
```diff
@@ -0,0 +1,1655 @@
...
+}
+```
+
+`AtomicWaker` holds a single waker and needs no lock, which is what a driver
+that has one waiter per peripheral instance wants. When the waker has to live
+inside state you already guard with a critical section or a `Mutex<RefCell<…>>`
+(a queue's head/tail pointers, a shared buffer), use
+`embassy_sync::waitqueue::WakerRegistration` instead — it is the same thing
+without the atomics, since the surrounding lock already provides the
+synchronization. Do not reach for third-party wait primitives; `embassy-sync` is
```
jamesmunns, line 964: This is probably the one part of the whole document I disagree with, especially as `embassy-sync` doesn't (yet) offer any intrusive/unbounded synchronization primitives allowing for avoiding "waker churn", and particularly for high-gpio count MCUs (or very low RAM MCUs) avoiding one waker per GPIO (sometimes in the hundreds on some packages).

I think it would be reasonable to instead say:

* Be consistent in the sync primitives you use in a HAL
* Prefer `embassy-sync`, document in a per-HAL development guide if you're *not* using embassy-sync, and note why.

IMO it would also be worth discussing the concept of waker churn in this section, or documenting it elsewhere (embassy-sync docs, faq, somewhere), and linking to it here. You can probably find one of my tirades in the matrix chat history and paste that somewhere.

Dirbaio: Sorry, I couldn't disagree more :sweat_smile: 

- In general, about policy: Maintainers should look for the good of the project. If the part you maintain depends on another part and that part is not up to the task, you should fix that, not give up on it and use something else. Using something else is *double bad* for the project: embassy-sync stays unfixed, *plus* now embassy-mcxa is inconsistent with the rest of the project. So, *of course* Embassy HALs should use the Embassy waker utils. If those utils are not good *we should make them good*. This is not a general rule against taking external dependencies, if a HAL needs something super niche and complex that no other HAL will ever need it is fine to take an external dependency. Waker utils is *not* that, every HAL needs them.
- For the particular case of maitake in embassy-mcxa:
   - waker churn is not a problem in HALs due to `&mut self`. You simply can't have two tasks waiting on the same thing so WakerRegistration/AtomicWaker works fine. 
   - Maitake WaitCell *also* wakes+replaces the old waker, so it also has waker churn. There is simply *no reason* to use maitake WaitCell instead of embassy-sync in embassy-mcxa.
   - For the gpio WaitMap the solution would be something with `bind_interrupts!` so only statics for used pins are created. It would be faster, smaller and more consistent with the rest of Embassy.

PR #6686, opened 2026-08-01
[MCXA] dma: Peripheral Paced Scatter-Gather

ayodejiige:
Extend the `ScatterGatherBuilder` DMA API to support peripheral paced mode, alongside a `LPUART2` loopback example that exercises it.


examples/mcxa5xx/src/bin/dma_scatter_gather_builder_uart_loopback.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let tx_fut = async {
+        defmt::info!("TX: gather chain starting");
+        let t = tx_sg.build_borrowed(&mut tx_ch).unwrap();
+        let res = t.await;
+        defmt::info!("TX: transfer done, result ok={}", res.is_ok());
+        res
+    };
+
+    let (rx_res, tx_res) = join(rx_fut, tx_fut).await;
```
felipebalbi, line 122: Run something like this in a loop. We need to know that this API remains robust even after running for many days.

examples/mcxa5xx/src/bin/dma_scatter_gather_builder_uart_loopback.rs
```diff
@@ -0,0 +1,162 @@
...
+        tx_sg.segment_count(),
+        rx_sg.segment_count()
+    );
+
+    // Enable LPUART2 DMA request lines directly via PAC — the HAL does not
+    // yet provide a method for this on `Lpuart`. This is only used for the purpose of this example.
+    hal::pac::LPUART2.baud().modify(|w| {
+        w.set_tdmae(true);
+        w.set_rdmae(true);
+    });
```
felipebalbi, line 103: I don't think this will survive for long if we put all the onus on the users. How about an API where the user passes a slice of slices and the driver ensures safe use of the DMA block?

felipebalbi (changes requested):

examples/mcxa5xx/src/bin/dma_scatter_gather_builder_uart_loopback.rs
ayodejiige: I was following the existing ScatterGather pattern but with peripheral paced transfers enabled. Is what you are suggesting akin to having something like `read_from_peripheral_scatter` and `write_to_peripheral_gather` as methods under `DmaChannel` which takes in slices?

PR #6687, opened 2026-08-01
Fix a few USB issues

bugadani:
Claude was asked to audit embassy-usb-synopsys-otg, then another one to come up with a test for the discovered issues. This PR reflects the end result of three turns of ping-ponging, which means the windows device plugged/unplugged sounds have burnt out of my ears.

The driver now uses embassy-time to bound some previously-infinite loops. Some of the suspend-related fixes could not be fully tested - either needs iso endpoints, vbus detection, or Linux (per-port suspend).

Also:

- Deduplicates four open-coded copies of the same FIFO and endpoint handshakes; the isochronous end-of-frame handler and `init_fifo` now share the new helpers.
- Closes #6684, which fixes a subset of the suspend defect.

<details>
<summary>Claude report</summary>

# Fix five device-mode defects in the Synopsys DWC2 driver

An audit of `embassy-usb-synopsys-otg` found five defects, all confirmed on hardware
(ESP32-S3, full speed) with a purpose-built fixture and host runner. Each fix is verified
by a test that fails without it.

## Defects and fixes

- **Clear-halt never reset the data toggle**, so every packet after a `ClearFeature(ENDPOINT_HALT)` was dropped as a retransmission (USB 2.0 §9.4.5) — `endpoint_set_stalled` now sets `SD0PID_SEVNFRM` for bulk and interrupt endpoints.
- **Enabling an endpoint never reset the data toggle either**, so the first packet after a `SET_INTERFACE` or `SET_CONFIGURATION` was lost (USB 2.0 §9.1.1.5) — same fix in `endpoint_set_enabled`.
- **Stalling an IN endpoint mid-transfer wedged it permanently**, leaving `EPENA` set with no path to clear it — `endpoint_set_stalled` now runs the endpoint-disable sequence and flushes the TX FIFO first.
- **Endpoint disable never waited for the core to acknowledge**, and OUT endpoints were disabled without global OUT NAK — new `abort_in_endpoint` / `abort_out_endpoint` helpers do it properly, with a real 10 ms `embassy_time` deadline instead of an unbounded spin.
- **A suspend latched before a bus reset wedged enumeration**, parking `embassy-usb` in `wait_resume` where it stops servicing SETUP — `USBSUSP` is now cleared in both the `usbrst` and `enumdne` branches, and masked out of the `GINTSTS` snapshot. (#6684)
- **A bus reset never woke parked endpoint futures**, so classes missed the `Disabled` error and carried stale state across re-enumeration — `configure_endpoints` now wakes both wakers and discards data buffered before the reset.
- **A disconnect during a control IN transfer left EP0's `EPENA` set permanently**, making the device unenumerable across any number of resets — the reset path now aborts any IN endpoint still showing `EPENA`.

## Verified

Data toggle both directions, stall-mid-transfer, 500 alternate-setting churn cycles, bus reset
waking a parked read, no SETUP replay across reset, forced `USBSUSP` latch, global OUT NAK
against a second endpoint pair, and disconnect during a control IN transfer with a negative
control. Worst-case critical section on the disable path is 11 µs.

## Not yet verified

- Isochronous endpoints are excluded from the toggle reset by inspection only; no iso endpoint has been tested.
- No test yet proves the suspend fix does not *over*-clear and swallow a legitimate suspend — a device that never suspends looks healthy from the host, so this needs device-side counting.
- Replay of pre-reset OUT payload data (the SETUP half is covered).
- `abort_out_endpoint` costs ~1.2 ms per active OUT endpoint, which appears to be a frame-boundary property of the core. Harmless for one endpoint; untested for a composite device that disables several at once.

</details>

### DIFF
--- a/embassy-usb-synopsys-otg/CHANGELOG.md
+++ b/embassy-usb-synopsys-otg/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased - ReleaseDate
 
 - Changed: `embassy-time` is now an optional feature for device mode. Remote wakeup is now supported via `embassy-time`.
+- Fixed: Clearing an endpoint halt now resets the data toggle to DATA0, as does enabling an endpoint for a new configuration or alternate setting.
+- Fixed: Stalling an IN endpoint with a transfer in flight no longer leaves it permanently unusable.
+- Fixed: Disabling an endpoint now waits for the core to acknowledge, and OUT endpoints are disabled under global OUT NAK.
+- Fixed: A suspend latched before a bus reset no longer wedges enumeration.
+- Fixed: A bus reset now wakes endpoint futures so they observe the endpoint being disabled, and discards data received before the reset.
 - Fixed: Periodic IN transfer future can no longer park forever.
 - Fixed: Later events no longer override earlier ones.
 - Fixed: Correctly work with low-speed devices over high-speed hubs.
--- a/embassy-usb-synopsys-otg/src/lib.rs
+++ b/embassy-usb-synopsys-otg/src/lib.rs
@@ -17,9 +17,11 @@ use core::sync::atomic::{AtomicBool, AtomicU16, AtomicU32, Ordering};
 use core::task::Poll;
 
 use embassy_sync::waitqueue::AtomicWaker;
+#[cfg(feature = "embassy-time")]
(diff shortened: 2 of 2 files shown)


bugadani:
@HaoboGu can you try this PR please, and see that I didn't break anything? It includes your fix as well.

leftger (approved):

HaoboGu:
@bugadani I can verify that this branch works on nRF54LM20

Issue #2880, opened 2024-04-27
RP2040 USB hang when receiving a corrupted IN packet


danielstuart14:
We have a board that has an USB2422 HUB that connects both an RP2040 (B2 Revision) and an external USB port. Sometimes, when connecting a new device to the external port, the RP2040 will stop communicating with the O.S. (we are using CDC, device is still recognized, just won't accept any messages - ERR 13 on windows):
![image](https://github.com/embassy-rs/embassy/assets/8076962/6c5e70ac-8794-4505-a7cd-f0e75d71b2b1)

USB2422 Errata says that there is an issue that can happen with split transactions, where a packet may be corrupted in a very specific case, but shouldn't be a big issue as the device stack should just ignore this packet. We checked the RP2040 USB signals and it seems like one packet indeed sometimes looses some of its data:
![image](https://github.com/embassy-rs/embassy/assets/8076962/4976aeb4-63c9-4699-b519-b4b320ddae90)

The problem is that the RP2040 isn't able to recover from this issue, and just stops acknowledging  packets from the HOST (even the SETUP that it tries to send later):
![image](https://github.com/embassy-rs/embassy/assets/8076962/815dc419-06a6-49eb-a751-fbc438f15131)

If we try to send any new data through the CDC interface, this is what happens in the bus:
![image](https://github.com/embassy-rs/embassy/assets/8076962/99900d7b-e515-48f8-829f-f0aa13c97a70)

1 - The host sends the data, but no ACK/NACK is received
![image](https://github.com/embassy-rs/embassy/assets/8076962/6b39615e-6c19-4f67-a68e-75c9d66365f1)

2 - The host tries to receive some data (44us later):
![image](https://github.com/embassy-rs/embassy/assets/8076962/ff3cbd0f-e293-4dcf-aef8-b023dda5e66a)

3 - Device responds an ACK (9us later):
![image](https://github.com/embassy-rs/embassy/assets/8076962/e5bd64e6-174a-493e-aaad-37b1fde5362f)

4 - Nothing is sent for 30us and then the HOST retries...

The only things that we found that would make it work again are:
- Reseting the RP2040
- Reseting the HUB (by its reset pin)
- Disabling the RP2040 Composite Device in device manager and re-enabling it again.

For the two latter cases, we found out that the USB line reset is what makes it work again.
![image](https://github.com/embassy-rs/embassy/assets/8076962/e6683ab8-de98-455b-a4b9-7f5a2918c06e)

[...]

PR #6693, opened 2026-08-02
fix usb issues in embassy-rp

benedikt-bartscher:
inspired by https://github.com/embassy-rs/embassy/pull/6687

verified locally with an example project and some python scripts

### DIFF
--- a/embassy-rp/CHANGELOG.md
+++ b/embassy-rp/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 
 ## Unreleased - ReleaseDate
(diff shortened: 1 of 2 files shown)


bugadani:
I think it might be worth attempting to fix https://github.com/embassy-rs/embassy/issues/2880 or test if this PR has any effect on the issue.

benedikt-bartscher:
I don't think this has any effect on #2880 and i currently don't have an rp2040 lying around... maybe @danielstuart14 can verify this.

bugadani:
So you tested this with an RP2350?

benedikt-bartscher:
Yes, i did - currently trying to write some e2e tests for our downstream project. Embassy CI doesn't have any directly connected usb rp's, right?

LivingUserError (approved):

MTelthoe (approved):

PR #6696, opened 2026-08-03
 net: impl. timestamp polling

xoviat:
implemented on stack.

### DIFF
--- a/embassy-net/Cargo.toml
+++ b/embassy-net/Cargo.toml
@@ -109,7 +109,7 @@ std = ["xarxa/std"]
 alloc = ["xarxa/alloc"]
 ## Enable xarxa packetmeta-id feature
 packetmeta-id = [ "xarxa/packetmeta-id", "embassy-net-driver/packetmeta-id" ]
-packetmeta-timestamp = [ "xarxa/packetmeta-timestamp", "embassy-net-driver/packetmeta-timestamp", "packetmeta-id" ]
+packetmeta-timestamp = [ "xarxa/packetmeta-timestamp", "embassy-net-driver/packetmeta-timestamp" ]
 
 [dependencies]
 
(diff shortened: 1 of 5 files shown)


embassy-net/Cargo.toml
```diff
@@ -29,7 +29,7 @@ build = [
     {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp"]},
     {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ethernet", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp"]},
     {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ethernet", "medium-ieee802154", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp"]},
-    {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ethernet", "medium-ieee802154", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp", "packetmeta-id"]},
+    {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ethernet", "medium-ieee802154", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp", "packetmeta-id", "packetmeta-timestamp"]},
```
jordens, line 32:
```suggestion
    {target = "thumbv7em-none-eabi", features = ["defmt", "dns", "medium-ethernet", "medium-ieee802154", "medium-ip", "proto-ipv4", "proto-ipv6", "tcp", "udp", "packetmeta-timestamp"]},
```

I don't think the packetmeta-id is useful on its own here. Or -- IIUC -- you are creatively using the packet id as a vlan tag and need the feature on its own, then i would make that clearer. If you are: does vlan and ptp collide or is that broken due to id overload?

PR #6697, opened 2026-08-03
stm32: impl net ptp iface

xoviat:
requires the user to use the poll_tx_timestamp method and removes rx id support.


embassy-stm32/src/eth/v2/descriptors.rs
```diff
@@ -452,15 +448,19 @@ impl<'a> RDesRing<'a> {
     pub(crate) fn meta(&self) -> PacketMeta {
         let mut meta = PacketMeta::default();
 
-        meta.id = self.state.id(self.index);
+        meta.timestamp = match self.timestamp {
+            None
+            | Some(Timestamp {
+                seconds: 0,
+                quarter_nanos: 0,
```
jordens, line 462: When does this happen? If it's "clock not running" shouldn't we err or warn?

xoviat: Instead of storing Option<Option<Timestamp>>, I store Option<Timestamp> and assume that a zero timestamp is None.

jordens (approved):
LGTM (without having done a really thorough review).
Tested on hardware.

PR #6702, opened 2026-08-04
[MCXA] Fix MCXA I2C Target DMA Race Between TDF and STOP Detection

GBJin:
We've encountered a HID timeout due to race condition between DMA STOP detection and TDF

The timeout was caused by treating TX DMA completion as completion of the I2C transaction. DMA completion only means the final byte has reached the TX FIFO; the byte may still be transmitted, NACKed, and followed by STOP. Current implementation return  NeedMore  before STOP was visible, then clear that late STOP while starting a continuation. This left stale target state, and the next host write could be clock-stretched through RXSTALL  until the host’s ~200 ms timeout.
image

In plain word: The DMA finished before STOP was detected, so the MCU mistakenly waited for more data and held SCL low indefinitely.

Waveform shows MCU hold SCL low for 195 ms thus timeout occurs (200ms). 
image

What to fix?
• Waits for the I2C bus result after DMA finishes and check STOP, repeated START, errors, and a genuine request for more data.
• Preserves a late STOP instead of clearing it accidentally.
• Prevents stale DMA/FIFO data from affecting the next transaction.
• Resets the FIFO cleanly when a transaction ends or errors.

### DIFF
--- a/embassy-mcxa/src/i2c/target.rs
+++ b/embassy-mcxa/src/i2c/target.rs
@@ -889,7 +889,22 @@ impl<'d> I2c<'d, Dma<'d>> {
         let peri_addr = self.info.regs().stdr().as_ptr() as *mut u8;
         let chunk_len = data.len();
 
-        self.clear_status();
+        // Preserve and consume a terminal status that arrived after the
+        // previous chunk reported NeedMore instead of clearing it here.
+        let ssr = self.info.regs().ssr().read();
+        if ssr.fef() {
+            self.reset_fifos();
+            return Err(IOError::FifoError);
+        } else if ssr.bef() {
+            self.reset_fifos();
+            return Err(IOError::BitError);
+        } else if ssr.sdf() {
+            self.reset_fifos();
+            return Ok(TxChunkOutcome::Stopped(0));
+        } else if ssr.rsf() {
+            self.reset_fifos();
+            return Ok(TxChunkOutcome::Restarted(0));
+        }
 
         unsafe {
             // Clean up channel state
@@ -949,12 +964,11 @@ impl<'d> I2c<'d, Dma<'d>> {
             self.mode.tx_dma.clear_done();
         }
 
-        // DMA completion only means the final byte reached the TX FIFO. The
-        // controller may still NACK that byte and issue STOP, or it may ACK it
-        // and request another byte. Wait for the I2C peripheral to distinguish
-        // those outcomes before reporting NeedMore. Returning on DMA completion
-        // alone races STOP and can cause a continuation call to clear the
-        // latched STOP before observing it.
+        // DMA completion only means the final byte reached the TX FIFO. TDF
+        // can assert before the controller's ninth-bit NACK and STOP, so a
+        // NeedMore result is provisional. If STOP arrives after this function
+        // returns, the continuation call above consumes it before re-arming
+        // DMA.
         if !(ssr.fef() || ssr.bef() || ssr.sdf() || ssr.rsf()) {
             self.info
                 .wait_cell()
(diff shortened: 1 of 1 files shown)


embassy-mcxa/src/i2c/target.rs
```diff
@@ -889,7 +889,22 @@ impl<'d> I2c<'d, Dma<'d>> {
...
+            return Err(IOError::FifoError);
+        } else if ssr.bef() {
+            self.reset_fifos();
+            return Err(IOError::BitError);
+        } else if ssr.sdf() {
+            self.reset_fifos();
+            return Ok(TxChunkOutcome::Stopped(0));
+        } else if ssr.rsf() {
+            self.reset_fifos();
+            return Ok(TxChunkOutcome::Restarted(0));
```
felipebalbi: every branch resets the fifos. Could you move `reset_fifos()` just outside the conditionals?

embassy-mcxa/src/i2c/target.rs
```diff
@@ -918,11 +933,8 @@ impl<'d> I2c<'d, Dma<'d>> {
...
-        // Wait for any of:
-        //  - I2C end-of-transfer flag (sdf, rsf) -> controller terminated
-        //  - I2C error flag (fef, bef) -> bus problem
-        //  - DMA channel completion -> chunk exhausted; if controller still
-        //    clocking, caller may want to call again (NeedMore)
+        // First wait for either the controller to terminate the transfer or
+        // DMA to load the entire chunk into the peripheral.
         poll_fn(|cx| {
             let _ = self.mode.tx_dma.wait_cell().poll_wait(cx);
             let _ = self.info.wait_cell().poll_wait(cx);
```
felipebalbi: I think this would be easier to read if the `poll_fn()` would be split into two separate futures and combined with a `select()`. Something like this (originally reported by @jamesmunns, who also wrote the example below):

```rust
let periph = self.info.wait_cell().wait_for(|| {
    self.info.regs().sier().write(|w| {
        w.set_feie(true);
        w.set_beie(true);
        w.set_sdie(true);
        w.set_rsie(true);
    });
    let ssr = self.info.regs().ssr().read();
    ssr.fef() || ssr.bef() || ssr.sdf() || ssr.rsf()
});
let dma = self.mode.rx_dma.wait_cell().wait_for(|| {
    self.mode.rx_dma.is_done()
});
let _res = select(periph, dma).await;
```

felipebalbi (changes requested):

embassy-mcxa/src/i2c/target.rs
GBJin: Moved outside the conditionals.

embassy-mcxa/src/i2c/target.rs
GBJin: Refined as request.

embassy-mcxa/src/i2c/target.rs
```diff
@@ -918,30 +933,22 @@ impl<'d> I2c<'d, Dma<'d>> {
...
-            } else {
-                Poll::Pending
-            }
-        })
-        .await;
+            ssr.fef() || ssr.bef() || ssr.sdf() || ssr.rsf()
+        });
+        // DMA to load the entire chunk into the peripheral.
+        let dma_is_done = self.mode.tx_dma.wait_cell().wait_for(|| self.mode.tx_dma.is_done());
+        let _ = select(i2c_interrupt, dma_is_done).await;
```
felipebalbi, line 949: If you ignore the return value, you're throwing away the best information we get from `select()`: which future actually completed. `select()` will give you an `Either` which directly maps to its two arguments. When you get a `First` that means `i2c_interrupt` completed, when you get a `Second`, then `dma_is_done` completed. That's useful information for the code that follows, right? Or are you proposing that telling them apart is irrelevant?

felipebalbi (commented):
One detail/question follows.

embassy-mcxa/src/i2c/target.rs
GBJin: Good question!
After tracing the follow-up code logic, it's irrelevant to telling completion sources apart since we need the current hardware state (line 951 `let mut ssr = self.info.regs().ssr().read();` ), not which notification observed first.
Both the `i2c_interrupt ` and `dma_is_done ` is to decide whether wake the task.
The subsequent SSR read which reads the hardware status determines the transfer outcome.

Issue #3754, opened 2025-01-11
AVR examples?


MarkuBu:
I noticed that AVR support is listed in the official documentation

https://docs.embassy.dev/embassy-executor/git/std/index.html#architecture

How far has AVR support progressed? Are there any examples, e.g. for an Arduino Uno?

All I found is this https://github.com/niclashoyer/embassy-avr-blink

Has there been any progress since then?

lulf:
The only thing that have been added for AVR is support in the embassy-executor. AFAIK no one is actively working on AVR support.

There is no HAL inside embassy, but you might be able to pick a HAL from another crate on crates.io (I can see some atmega PAC and HALs there). There is also a PR here https://github.com/embassy-rs/embassy/pull/3696 for adding an example.

brainstorm:
Well, I have this toy example: https://github.com/brainstorm/embassy-as5600/

And it worked well on Espressif ICs around a year ago, but for AVR I was facing some challenges (panics). Fast forward to today, it does not seem to compile (it 100% compiled at least back then!):

```
% rustc --version
rustc 1.87.0-nightly (f04bbc60f 2025-02-20)

% cargo build
(...)
error[E0432]: unresolved import `core::sync::atomic::AtomicPtr`
   --> /Users/rvalls/.cargo/git/checkouts/embassy-c08a80187403f815/0edd45e/embassy-executor/src/raw/mod.rs:31:26
    |
31  | use core::sync::atomic::{AtomicPtr, Ordering};
    |                          ^^^^^^^^^
    |                          |
    |                          no `AtomicPtr` in `sync::atomic`
    |                          help: a similar name exists in the module: `AtomicI8`
    |
    = help: consider importing this struct instead:
            portable_atomic::AtomicPtr
note: found an item that was configured out
   --> /Users/rvalls/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:301:12
    |
301 | pub struct AtomicPtr<T> {
    |            ^^^^^^^^^
note: the item is gated behind the `ptr` feature
   --> /Users/rvalls/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:295:7
    |
295 | #[cfg(target_has_atomic_load_store = "ptr")]
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

Issue #2453, opened 2024-01-15
Compatibility with ATMEGA architecture


chuckberrypi:
The docs say that this project maintains `hal` crates for various architectures (RP2040, STM, NRF), but that "you can still use HALs from other projects with Embassy." Does this mean that I can use embassy with an ATMEGA32u4 microcontroller I have, or can I only use it with the architectures listed on the [embassy-executor](https://docs.rs/embassy-executor/latest/embassy_executor/#architecture) documentation? If I can use it with the ATMEGA32u4 chip, any tips on getting rid of the error I am getting: 
``` 
| #[embassy_executor::main]
|                     ^^^^ could not find \`main\` in \`embassy_executor\`
```

enbyted:
Embassy-executor has some platform-dependent code required to support it, kind of like any embedded operating system.

Arguably it should not be very difficult to create a thread-mode (i.e. endless loop in main function) executor for AVR.

Regarding the error you're getting - the main macro is only defined if you select a valid architecture --> that's because it needs to know which platform-specific executor to use. 

I'm not an embassy expert (just started using it a couple weeks ago on a toy project actually), but to me it looks like to add basic AVR support you'd need at least:
 - Add executor implementation to `embassy-executor/src/arch/` - look at the `cortex_m` implementation for reference. Probably skip the interrupt executor --> make it fail if enabled.
 - Implement `__pender` in the same file
 - Add the main macro support to `embassy-executor-macros`

The simplest executor you can make is one that will do:
```rust
pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
    init(self.inner.spawner());
    loop { unsafe {self.iner.pool(); } }
}
```

For testing you can just start with using `embassy_executor::raw::Executor` ([docs](https://docs.rs/embassy-executor/latest/embassy_executor/raw/index.html))

I did not even try to compile below code (never tried rust on AVR), but you could try something along those lines:
```rust
use embassy_executor::raw;
use static_cell::StaticCell;

static EXECUTOR: raw::Executor = StaticCell::new();

#[export_name = "__pender"]
/// Required by raw::Executor, but can do nothing
fn pender(_context: *mut ()) {}

pub fn main()
{
  let executor = EXECUTOR.init(raw::Executor::new(ptr::null_mut()));
  let _spawner = executor.spawner();
  
  loop { 
    unsafe { executor.pool(); } 
  }
}
```

jamesmunns:
[...]

Issue #570, opened 2022-01-09
AVR support?


jeremysalwen:
I noticed that this project reserved the crate embassy-avr: https://lib.rs/crates/embassy-avr  Does this foretell AVR support in the near future?

wbrickner:
Bump. Is there a way to do this? I'd love to use embassy for all my arduino projects. I don't know how to set this up myself unfortunately

cherrydev:
I too would like to see this. Unfortunately I'm pretty much a rust beginner so I'm not going to be much help. That said, the documentation says that you can use a 3rd party HAL with embassy and if you look at the esp-rs/esp-hal repo in the esp-hal-common module you'll see how it's implemented.

niclashoyer:
I hacked together a working embassy executor and a working time driver with a fixed frequency. Please be aware that this is just a proof of concept, as there are no async driver implementations for other peripherals yet.

Embassy executor: https://github.com/niclashoyer/embassy/blob/avr/embassy-executor/src/arch/avr.rs

Blink/hello world example (using `simavr` and async delays): https://github.com/niclashoyer/embassy-avr-blink

PR #6703, opened 2026-08-04
feat: add initial ATmega328P HAL support

nursude355:
## Summary

Adds initial Embassy-compatible HAL support for the ATmega328P.

The implementation currently includes:

- GPIO for ports B, C and D
- Timer1-based Embassy time driver
- blocking ADC, I2C, SPI and USART drivers
- Timer0 and Timer2 PWM
- bit-banged 1-Wire support
- ATmega328P examples, including async blinky
- 8 MHz and 16 MHz clock configurations

## Current validation

- Host cargo check passes
- Clippy with warnings denied passes
- All AVR examples pass cargo check for `avr-none`
- 8 MHz and 16 MHz configurations compile

## Remaining validation

- Final linking with avr-gcc
- Flashing and testing on real ATmega328P hardware
- Peripheral hardware validation

This draft PR is also intended to request feedback on whether the HAL
should remain ATmega328P-specific or evolve into a broader `embassy-avr`
crate.

### DIFF
--- a/embassy-atmega328p/Cargo.toml
+++ b/embassy-atmega328p/Cargo.toml
@@ -6,7 +6,8 @@ license = "MIT OR Apache-2.0"
 description = "Embassy-compatible hardware abstraction layer for the ATmega328P"
 keywords = ["embedded", "async", "avr", "atmega328p", "embedded-hal"]
 categories = ["embedded", "hardware-support", "no-std"]
-repository = "https://github.com/nursude355/embassy"
+repository = "https://github.com/embassy-rs/embassy"
+documentation = "https://docs.embassy.dev/embassy-atmega328p"
 publish = false
 
 [package.metadata.embassy]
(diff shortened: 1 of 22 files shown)


leftger:
I don't want to be a downer but the 328P is a significant departure from what Embassy currently supports. Cortex-M/Cortex-R/RISC-V are the common architectures. The 328P with 32KB of flash is absolutely tiny compared to what current microcontrollers offer. It's a 17-year-old device with decreasing use in industry. From a maintenance standpoint the ROI of bringing this into Embassy just isn't there. First and foremost `probe-rs` support for the ATMega328 is nonexistent. I'd consider debugger/flasher support before bringing this into Embassy.

joelsa:
I'd also like to point to the following issues:

- https://github.com/embassy-rs/embassy/issues/3754
- https://github.com/embassy-rs/embassy/issues/2453
- https://github.com/embassy-rs/embassy/issues/570

the existing naming schemes could give a hint towards answering this question:

> This draft PR is also intended to request feedback on whether the HAL
> should remain ATmega328P-specific or evolve into a broader embassy-avr
> crate.

PR #6704, opened 2026-08-04
embassy-stm32/low-power: add stm32f4 support

decaday:
Thanks to embassy already implementing time-driver pause and RTC handoff during stop mode, I have implemented low-power support for stm32f4.

With this PR, stm32f4 can also enter stop mode during idle by using `embassy_stm32::executor::Executor`.

I also wrote the `stm32f4-lp` example and added detailed comments.
The ~42uA stop power consumption on stm32f401 (as shown in the attached image) matches the expectations from [AN4365](https://www.st.com/resource/en/application_note/an4365-using-stm32f4-mcu-power-modes-with-best-dynamic-efficiency-stmicroelectronics.pdf), and wakeup works normally.

Additionally, if the `FPDS` bit in `PWR.CR1` is enabled, the power consumption can be further reduced to ~12uA, but it will extend the wakeup time. Since this behavior is not exclusive to stm32f4, I did not include it in this PR.
Later, I might open a new PR to introduce a new configuration option to control this, but that may require more discussion.

image

### DIFF
--- a/embassy-stm32/src/low_power.rs
+++ b/embassy-stm32/src/low_power.rs
@@ -164,6 +164,12 @@ mod platform {
             });
         }
 
+        #[cfg(stm32f4)]
+        crate::pac::PWR.cr1().modify(|w| {
+            w.set_pdds(crate::pac::pwr::vals::Pdds::StopMode);
+            w.set_lpds(true);
+        });
+
         #[cfg(stm32wb)]
         drop(mutex);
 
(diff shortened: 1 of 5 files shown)


embassy-stm32/src/low_power.rs
xoviat, line 253: Just put true for f4 here?

decaday: Applied and rebased. Thanks for the review!

decaday: I also retested the example, execution and power consumption looks still good.

xoviat:
note to self: f4 can tests need a wakeguard.

PR #6710, opened 2026-08-04
stm32: align gpdma with dma

xoviat:
pending hw confirmation


pufmat:
After changes 3 it no longer works. In UART case it receives nothing.

pufmat:
After changes 6.6 it works. Tested on STM32H523CE and STM32F411CE since there are changes affecting DMA/BDMA part.

xoviat:
bender run

